PR #9691, opened 2021-12-29
support cat send multi address and amount

HiveProject2021:
To support send multi address with different amount and memo at one time.


mariano54:
This is a good feature but we should implement a new RPC consistent with how normal chia TXs are performed.

Quexington:
Can you point this at the branch `pacr-dev` instead?  We also need a test added to `tests/wallet/rpc/test_wallet_rpc.py`.

PR #9688, opened 2021-12-28
plotting|tests: Extend the lifetime of plot data cache entries

xdustinface:
This PR extends the lifetime of plot data cache entries to 7 days. On current `main` cache entries are instantly dropped if a plot is not longer available. This is annoying because it needs to reload the cache data if you for example detach a drive (e.g. for some maintenance) longer than the refresh interval.

PR #9692, opened 2021-12-29
tests: Some renaming in plot manager tests to avoid `pytest` conflicts

xdustinface:
Having `Test` or `test_` prefixes obviously lets `pytest` think its a 
actual test class/method and it 
leads to warnings in test logs if it doesn't comply with its rules:

```
/home/runner/work/chia-blockchain/chia-blockchain/tests/plotting/test_plot_manager.py:45: 
PytestCollectionWarning: cannot collect test class 'TestDirectory' 
because it has a __init__ constructor (from: 
plotting/test_plot_manager.py)
```
or
```
/home/runner/work/chia-blockchain/chia-blockchain/tests/plotting/test_plot_manager.py:135: 
PytestCollectionWarning: cannot collect test class 'TestEnvironment' 
because it has a __init__ constructor (from: 
plotting/test_plot_manager.py)
```

Based on #9688.

### DIFF
--- a/tests/plot_sync/test_plot_sync.py
+++ b/tests/plot_sync/test_plot_sync.py
@@ -26,7 +26,7 @@ from chia.util.ints import uint8, uint64
 from chia.util.streamable import _T_Streamable
 from tests.block_tools import BlockTools
 from tests.plot_sync.util import start_harvester_service
-from tests.plotting.test_plot_manager import MockPlotInfo, TestDirectory
+from tests.plotting.test_plot_manager import Directory, MockPlotInfo
 from tests.plotting.util import get_test_plots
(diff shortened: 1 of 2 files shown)


altendky:
Looks like this got mixed with https://github.com/Chia-Network/chia-blockchain/pull/9688.

xdustinface:
> Looks like this got mixed with #9688.

Check out the last line of the PR description :) 

> Based on #9688.

Hence its still draft until the other is merged.

altendky (approved):
We shouldn't be importing `test_*` modules anyways, but that only addresses one of the cases here and is not a smell that is unique to this code.  We can take that on separately.

wjblanke (approved):
aok

PR #9697, opened 2021-12-30
Windows: start daemon without a window and detached from current console

### LINKED ISSUE #13175: [Bug] Windows CLI: Ctrl-C in console will cause Daemon to close despite it appearing to be running in the background
### What happened?

Start node with `chia start node`
After node and daemon are started type Ctrl-C
Daemon will exit

(happens in both cmd.exe and PowerShell)

The Daemon does not detach from the console so it receives keyboard interrupts from the console as long as it remains running.

See https://github.com/Chia-Network/chia-blockchain/blob/764a5a22761f10da1811ed8129224db04602c1c6/chia/cmds/start_funcs.py#L20

### Version

1.5.1 and earlier

Windows

CLI

### Relevant log output

```shell
N/A
```

jcteng:
fixed issue on: 
1. windows os , daemon gone after start node
2. gbk platform encoding issue on stdio pipe


emlowe:
Possibly this fixes #13175 - needs some testing.

emlowe:
Close and re-open for fixed CI

emlowe:
During testing, a console window was appearing when using DETACHED_PROCESS with CREATE_NO_WINDOW. This seems to be an instance of this defect (https://github.com/python/cpython/issues/85785) - setting DETACHED_PROCESS will always create a windows regardless of either CREATE_NO_WINDOW or STARTF_USESHOWWINDOW. Using CREATE_NO_WINDOW alone however, does work in that a new console is created (detaching from the previous console) and no windows are shown for this new console. Further testing ongoing

emlowe:
Testing of packaged builds on windows is good. Various other cases all working as expected (`chia run_daemon`)

emlowe:
This does fix #13175

chia/cmds/start_funcs.py
```diff
@@ -17,8 +17,18 @@
 
 def launch_start_daemon(root_path: Path) -> subprocess.Popen:
     os.environ["CHIA_ROOT"] = str(root_path)
-    # TODO: use startupinfo=subprocess.DETACHED_PROCESS on windows
-    process = subprocess.Popen([sys.argv[0], "run_daemon", "--wait-for-unlock"], stdout=subprocess.PIPE)
+    if sys.platform == "win32":
+        return subprocess.Popen(
+            [sys.argv[0], "run_daemon", "--wait-for-unlock"],
+            encoding="utf-8",
+            creationflags=subprocess.CREATE_NEW_PROCESS_GROUP | subprocess.CREATE_NO_WINDOW,
```
altendky, line 24: I think the `if` can be used just to set a variable for the creation flags and then we can retain a single `Popen()` call since everything else is the same.

altendky (changes requested):

chia/cmds/start_funcs.py
emlowe: Agreed and changed

altendky (approved):

arvidn (approved):

PR #9704, opened 2021-12-31
run tests in CI via coverage

altendky:
When running with `coverage` the `tests.core.full_node.test_full_node::TestFullNodeBlockCompression::test_block_compression` test ended up running into an apparent deadlock after some runs on a stdout flush in pytest (`_pytest._io.terminalwriter.TerminalWriter.flush()`).

https://github.com/pytest-dev/pytest/blob/9a992df3c93ae330765e81316f7fb65ef40692af/src/_pytest/_io/terminalwriter.py#L173-L174
```python
    def flush(self) -> None:
        self._file.flush()
```

For whatever reason this doesn't seem to happen with coverage==4.5.4, but that's a few years old.  See https://github.com/Chia-Network/chia-blockchain/pull/9705.  Here I instead use `multiprocessing.set_start_method("spawn")` to avoid created processes inheriting resources like file descriptors, including sockets.

A somewhat similar issue recently came up in some plot management tests.  A minimal example is available at https://github.com/xdustinface/chia-blockchain/commit/9fd950e5cf013c65579e2d4ec44b9dafffca7954.  In short, if the socket is opened in the fixture then closed in the test then the port is still in use.  If the socket is opened in the test, it can be closed and reopened without issue as would be expected.  After some debugging I reported this first against pytest-monitor (https://github.com/CFMTech/pytest-monitor/issues/53) and then after more debugging against memory-profiler (https://github.com/pythonprofilers/memory_profiler/issues/342#issue-1088426177).  memory-profiler also uses multiprocessing and launches a process during the test run, but not the fixture setup.  tl;dr, if you open a socket in the fixture then it is inherited by the process created by pytest-monitor/memory-profiler and even when you close the socket in the test, the subprocess keeps the port open and the test cannot rebind to that port.

Is there a reason we want to use multiprocessing's default start method in Linux of `fork`?  Or would it be ok, as is done here, to use `spawn` in all cases.  This is already the default on macOS and Windows.  https://docs.python.org/3.10/library/multiprocessing.html#contexts-and-start-methods

A bunch of CI reruns haven't turned up issues so far.  I am doing a local full sync on a test RPi, though I'm cheating and tried removing the `- 2` process count calculation to make it go faster (being offered configurably in https://github.com/Chia-Network/chia-blockchain/pull/9709).

Note that this does _not_ provide super useful coverage results since it does not combine all the runs together.  With the existing multi-workflow setup this is a bit annoying and I would probably enlist IPS support on that.  If we were to switch to a single workflow with a generated matrix then collection would be easier.  I already have that pair implemented (along with a bunch of debugging garbage) over in https://github.com/Chia-Network/chia-blockchain/pull/9320 and https://github.com/Chia-Network/chia-blockchain/pull/9324.  The basic reason to not switch is that the granularity of GitHub Actions re-runs is at the workflow so we presently don't have to rerun all jobs for just one flaky failure (someday... https://github.com/github/roadmap/issues/271). Anyways, it might be worth merging this if we are happy with the change so that we can get to wider spread testing of it sooner.  Combining the coverage results and reporting them well can be taken on separately.

Draft for:
- [x] Discussion of the multiprocessing start method change and any concerns around that
- [x] At least one successful full sync~, presently in progress on my RPi, maybe we want more~
- [x] Fix the worker process names
- [x] Run on a working node with a wallet etc
- [x] Run blueboxing
- [x] `spawn` was merged in https://github.com/Chia-Network/chia-blockchain/pull/10322
- [ ] The `spawn` change may need some additional work, no need to pile this on until we address that.


altendky:
That's one green, let's try again.

altendky:
I guess I should note that I have done a couple reruns just like I would with our existing CI runs.

altendky:
always run more

altendky:
and run again

altendky:
A bunch of failures on macOS related to various files not being available for compiling the timelord.  Let's see if it repeats.

mariano54 (commented):
I have no problem with this change but I'm not super well versed in fork vs spawn. Anyway the processes we use are long lived so it won't have a negative performance impact

arvidn:
I would expect spawn to perform better than fork. I actually made this patch myself a while ago (but I don't think I submitted a PR). The only concern I would have is compatibility. there are systems that don't support `posix_spawn()`. But maybe they are all very old or fringe enough for us to not care about. But I don't know.

It seems these systems supports posix_spawn:

[FreeBSD 8.0](https://www.freebsd.org/cgi/man.cgi?query=posix_spawn&sektion=3)
[OpenBSD 5.2](https://man.openbsd.org/posix_spawn.3)
[NetBSD 6.0](https://man.netbsd.org/NetBSD-8.0/posix_spawn.3)
[Illumos](https://illumos.org/man/3C/posix_spawn)
[smartos](https://smartos.org/man/3c/posix_spawn)
[Solaris](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwib3ZLAgbf1AhXtQvEDHX29AogQFnoECAkQAQ&url=https%3A%2F%2Fblogs.oracle.com%2Fsolaris%2Fpost%2Fposix_spawn-as-an-actual-system-call&usg=AOvVaw3YQUEqZnvhAayE2d_0CgYf) However, there appears to be a [python bug](https://bugs.python.org/issue44935) only fixed very recently.
[Haiku](https://git.haiku-os.org/haiku/commit/?id=ccd42320c45658052d620804bb8e05e5bc327706)

So, it seems like a safe change (but I wonder why it isn't the default in python).

arvidn:
one problem I have with code coverage tools is that I have never found one that is good at excluding "collateral" coverage. i.e. lines of code that are executed but whose results are not checked by any test.

in an ideal world, only executed lines that would cause a test to fail if changed, would be included in the code coverage. Those are the lines that are being tested.

aqk (dismissed):

aqk (approved):

PR #9708, opened 2022-01-01
plotting: Add missing locks in `PlotManager`

### LINKED ISSUE #9707: [Bug] ERROR    Exception: dictionary changed size during iteration
### What happened?

Errors in some remote harvesters' logs and corresponding farmer log.  A remote harvester showing this ERROR is either plotting (using bladebit) or receiving (NFT) plots over rsync (using plotman).  From time to time remote harvesters would lose connection to the farmer. 

### Version

1.2.11

Linux

CLI

### Relevant log output

```shell
ON REMOTE HARVESTER:  xxx.xxx.xxx.xxx is farmer LAN IP.
2021-12-31T23:23:46.652 harvester harvester_server        : ERROR    Exception: dictionary changed size during iteration <class 'RuntimeError'>, closing connection {'host': '192.168.1.174', 'port': 8447}. Traceback (most recent call last):
  File "/home/jh/chia-blockchain/chia/server/server.py", line 576, in api_call
    response: Optional[Message] = await asyncio.wait_for(wrapped_coroutine(), timeout=timeout)
  File "/usr/lib/python3.8/asyncio/tasks.py", line 494, in wait_for
    return fut.result()
  File "/home/jh/chia-blockchain/chia/server/server.py", line 573, in wrapped_coroutine
    raise e
  File "/home/jh/chia-blockchain/chia/server/server.py", line 566, in wrapped_coroutine
    result = await coroutine
  File "/home/jh/chia-blockchain/chia/harvester/harvester_api.py", line 185, in new_signage_point_harvester
    for try_plot_filename, try_plot_info in self.harvester.plot_manager.plots.items():
RuntimeError: dictionary changed size during iteration

2021-12-31T23:23:46.656 harvester harvester_server        : WARNING  Banning xxx.xxx.xxx.xxx for 10 seconds
2021-12-31T23:23:49.436 harvester harvester_server        : WARNING  Peer xxx.xxx.xxx.xxx is still banned, not connecting to it
2021-12-31T23:23:52.438 harvester harvester_server        : WARNING  Peer xxx.xxx.xxx.xxx is still banned, not connecting to it
2021-12-31T23:23:55.443 harvester harvester_server        : WARNING  Peer xxx.xxx.xxx.xxx is still banned, not connecting to it

ON FARMER: yyy.yyy.yyy.yyy is remote harvester IP
2021-12-31T21:23:18.934 farmer chia.farmer.farmer         : ERROR    Harvester 'yyy.yyy.yyy.yyy/zzzzzzzzzzzzzzzzzzzzzzz' did not respond: (version mismatch or time out 90s)
```

xdustinface:
Should fix #9707.

### DIFF
--- a/chia/plotting/manager.py
+++ b/chia/plotting/manager.py
@@ -166,11 +166,12 @@ class PlotManager:
         self._lock.release()
 
     def reset(self):
-        self.last_refresh_time = time.time()
-        self.plots.clear()
-        self.plot_filename_paths.clear()
-        self.failed_to_open_filenames.clear()
-        self.no_key_filenames.clear()
+        with self:
+            self.last_refresh_time = time.time()
+            self.plots.clear()
+            self.plot_filename_paths.clear()
+            self.failed_to_open_filenames.clear()
+            self.no_key_filenames.clear()
 
     def set_refresh_callback(self, callback: Callable):
         self._refresh_callback = callback  # type: ignore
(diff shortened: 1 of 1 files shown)


chia/plotting/manager.py
```diff
@@ -171,11 +171,12 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
         self._lock.release()
 
     def reset(self):
```
arvidn, line 173: are you confident this is never called from the main thread?
If it is, this really needs to be an async mutex, that puts the co-routine to sleep, rather than the whole thread.
mixing regular threading and mutexes with co-routines makes me nervous about accidentally lock up threads.

arvidn: it looks like it's called here: https://github.com/xdustinface/chia-blockchain/blob/1f83bf97d71e8660632bafc7f4e2d7c2bc004a60/chia/plotting/manager.py#L319
which is a separate thread. However I think it might warrant some warning comments. I can see someone accidentally calling this from the main thread and it slipping through reviews.

xdustinface: Whats the exact problem when calling it from the main thread if it eventually gets released? We lock from the main thread when populating the coroutine list for lookups here https://github.com/Chia-Network/chia-blockchain/blob/467478ef754d1bb43ceadfa5729e35029edc3f91/chia/harvester/harvester_api.py#L173

arvidn: the async mutexes don't block the thread when you acquire them, they de-schedule the co-routine and adds it to the wait-list of the mutex. Once it's its turn is it scheduled to run again. i.e. other co-routines can make progress during that time.

xdustinface: Sure but this is only used by the harvester which only has the farmer connection and no blockchain to maintain...its purpose is to process signage points so isn't it reasonable to just wait for the plot manager to be ready if we anyway can't do something else, no? I agree though that all this signage point processing code needs some love but this is not related to this PR i guess.

xdustinface:
Rebased on `main`.

arvidn (approved):

PR #9709, opened 2022-01-02
Make the hard coded 2 reserved cores configurable

altendky:
Removing the hard coded `- 2` took my 4-core RPi4 from 2 full cores and 14k BPH (Blocks Per Hour) sync rate to closer to 4 full cores and 23k BPH.  Linux anyways is pretty good about not starving user interaction from CPU time so attempting to use all cores should be fine in that regard anyways.  I don't know how well Windows handles this these days.  If there is concern, perhaps process priority would be better than forgoing entire cores all the time?  For example, there were some number of weeks where I had a full plotting load on my laptop with plotman's priority adjustment on them and I was also working full time on the laptop.  I only really bothered to pause the plotting when on video calls or otherwise wanting my system to focus on other loads.  In general, just working in PyCharm etc was still fine.

(gratuitous `htop` screenshot from around block 105k)
image

Draft for:
- [x] Self testing
  - [x] blockchain
  - [x] wallet
- [x] Discussion of the reasons for the existing value of 2
  - 2 specifically accounts for a 'real core' when dealing with hyperthreading.
- [x] Discussion of the reasons for not using all cores
  - The interest was in leaving a processor available for the OS to use for other things including remote connections etc.
- [x] Discussion of what the default value should be for this change
  - We're going to leave the default behavior unchanged for now.  Future changes to the default can be discussed including maybe after public feedback on the effects in various cases.
- [ ] ~Testing responsiveness on Windows with 0 reserved cores~
  - Not applicable since we are not changing the default
- [ ] ~Testing responsiveness on older macOS systems with 0 reserved cores~
  - Not applicable since we are not changing the default
- [x] Consider if we want to saturate to non-negative or maybe fail on negative values
  - Maybe it's just ok for someone to request the pool be larger than their core count if they decide to specify negative?

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -100,6 +100,7 @@ class Blockchain(BlockchainInterface):
         consensus_constants: ConsensusConstants,
         hint_store: HintStore,
         blockchain_dir: Path,
+        reserved_cores: int,
     ):
         """
         Initializes a blockchain with the BlockRecords from disk, assuming they have all been
@@ -112,7 +113,7 @@ class Blockchain(BlockchainInterface):
         cpu_count = multiprocessing.cpu_count()
         if cpu_count > 61:
             cpu_count = 61  # Windows Server 2016 has an issue https://bugs.python.org/issue26903
-        num_workers = max(cpu_count - 2, 1)
+        num_workers = max(cpu_count - reserved_cores, 1)
         self.pool = ProcessPoolExecutor(max_workers=num_workers)
         log.info(f"Started {num_workers} processes for block validation")
 
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -185,8 +185,9 @@ class FullNode:
         self.coin_store = await CoinStore.create(self.db_wrapper)
         self.log.info("Initializing blockchain from disk")
         start_time = time.time()
+        reserved_cores = self.config.get("reserved_cores", 2)
         self.blockchain = await Blockchain.create(
-            self.coin_store, self.block_store, self.constants, self.hint_store, self.db_path.parent
+            self.coin_store, self.block_store, self.constants, self.hint_store, self.db_path.parent, reserved_cores
         )
         self.mempool_manager = MempoolManager(self.coin_store, self.constants)
 
--- a/chia/util/initial-config.yaml
+++ b/chia/util/initial-config.yaml
@@ -333,6 +333,10 @@ full_node:
   # If node is more than these blocks behind, will do a short batch-sync, if it's less, will do a backtrack sync
   short_sync_blocks_behind_threshold: 20
 
+  # When creating process pools the process count will generally be the CPU count minus
+  # this reserved core count.
+  reserved_cores: 2
+
   # How often to initiate outbound connections to other full nodes.
   peer_connect_interval: 30
   # How long to wait for a peer connection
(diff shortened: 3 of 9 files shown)


mariano54:
I agree with this change (config) and also with reducing the number to 1 or 0. There are multiple reports of people having better sync experiences with a value of 0 so I support that.

I don't remember if it was my idea to do -2 but it might just have been an overly conservative number.

neurosis69:
On my RPi I tested more than 30 `long sync` events in between the heights of 1070016 and 1080000 (covering first dust storm) during the last few days.
Beside of droping the, for the `long sync` event unnecessary, coin_record indexes, using all 4 cpu cores brought the best performance boost.

Tested on RPi4, 8G, DB@SSD (USB3)

Sync time between height of 1070016 and 1080000
Initial time: 6h15m
Time without using any coin_record indexes: 5h30m
Time without coin_record indexes and using 4 cores: 5h08m

You may also want to like into that [mempool cpu core limitation](https://github.com/Chia-Network/chia-blockchain/blob/13ff7489b606d38b8294ed8c256d0177d39eb4bb/chia/full_node/mempool_manager.py#L104)?
Had no impact during `long sync` events, but may improve RPi performance during regular operation?

altendky:
This is around block 325k and is down to around 15k BPH and not keeping the average CPU usage quite as high.  This is presently a secondary system that isn't getting 8444 exposed publicly so maybe it doesn't have enough peers to keep syncing at "full" speed?  Or perhaps it is back and forth with the disk for db work, I don't know.  I am logging the following data once a minute throughout.  We should probably add more full-sync monitoring information.

```json
{"query_time":"2022-01-03T14:14:50Z","height":324768,"timestamp":"2021-05-23T14:05:07Z","sync":{"sync_mode":true,"sync_progress_height":324768,"sync_tip_height":1367270,"synced":false}}
```

image

neurosis69:
Here is a chart of 5 consecutive syncs on my RPi, all between Height 223648 and 228000, where 225696 marks the start of transactions. 

Testcases are
1: No changes from branch 1.2.11 (40m42s)
2: Dropped all coin_record indexes (38m47s)
3: Dropped all coin_record indexes + use all 4 cores (33m43s)
4: Dropped all coin_record indexes + increase coin_records lru * 100 (37m40s)
5: Dropped all coin_record indexes + use all 4 cores + increase coin_records lru * 100 (29m48s)

grafik

altendky:
Looks like @hoffmang9 did this back in 2020 https://github.com/Chia-Network/chia-blockchain/commit/7fb2e03d38310546c0e2fa7f338ca38721663dff so I'll give them a chance to weigh in if they like.

altendky:
We could account for "real core"s by checking for hyperthreading and making it 1 or 2 accordingly.  Then even with the existing intent we would still give the pi 3 cores rather than the 2 it gets now.  We could also set the process priorities probably?  There might be some caveats with it being a library-managed pool that might launch new processes behind the scenes.

Or maybe we're just comfortable with this and move forward.

hoffmang9:
We should check that 0 is ok on Windows and Mac (older machines especially.) assuming it is I don't see a problem with using 0.

mariano54 (approved):

mariano54:
The only feedback I have is that we should definitely change the default to 0. Too many people are suffering from the slow sync times.

altendky:
We'll leave the debate over the default value to another PR and this one will leave the default behavior the same as before.

A question though.  Should we limit the configurable value to non-negative numbers?  A negative number would result in more pool processes than reported cores.  Seems like something that wouldn't usually be wanted but...  some people like to try stuff.  It could be limited by saturating to 0 regardless of how negative it is configured or a negative value could be an error.  I tend to think a straight up error isn't good and saturating seems more likely to confuse someone that's playing around intentionally than to help someone avoid a mistake.

PR #9715, opened 2022-01-03
Add the ability to rename tokens form CLI

Quexington:


chia/cmds/wallet_funcs.py
```diff
@@ -162,10 +162,16 @@ async def add_token(args: dict, wallet_client: WalletRpcClient, fingerprint: int
     token_name = args["token_name"]
     try:
         asset_id_bytes: bytes = hexstr_to_bytes(asset_id)
-        response = await wallet_client.create_wallet_for_existing_cat(asset_id_bytes)
-        wallet_id = response["wallet_id"]
-        await wallet_client.set_cat_name(wallet_id, token_name)
-        print(f"Successfully added {token_name} with wallet id {wallet_id} on key {fingerprint}")
+        existing_info: Optional[Tuple[Optional[uint32], str]] = await wallet_client.cat_asset_id_to_name(asset_id_bytes)
+        if existing_info is None:
```
paninaro, line 166: Should this be:

```suggestion
        if existing_info is None or existing_info.wallet_id is None:
```

`cat_asset_id_to_name` will return existing_info with a `wallet_id` of None if the asset ID is in `DEFAULT_CATS`

chia/cmds/wallet_funcs.py
```diff
@@ -162,10 +162,16 @@ async def add_token(args: dict, wallet_client: WalletRpcClient, fingerprint: int
     token_name = args["token_name"]
```
paninaro, line 162: Maybe consider updating the name to help indicate to readers that this adds or renames a wallet.

Quexington: I was pondering this, but do you have any ideas?  `manage_token` is the best I can come up with but that sounds far too sophisticated for what this does.

paninaro: add_or_rename_token? or is that too wordy? manage_token would be ok too I think since manage doesn't have a single meaning.

Quexington: I just feel like if it's manage_token, people will wonder where to add a token.  And the functionality is so small and similar I don't think it's worth splitting into two commands.  Renaming can be thought of as overwriting one of your current tokens with an add :)

paninaro (approved):

PR #9716, opened 2022-01-03
New weight proof version

almogdepaz:
- [x] un-revert revert of chiavdf PR 87 (see revert PR 104)

Smaller proof format using compressed vdf outputs

Include proofs only in sampled segments

uses 140 sub epoch samples vs the older 20

interactive proving using local salt and peak hash

Added Wp capabilities in handshake to support both versions
so node will try using new version if peer is capable old version if not

added wp  cmd to cli to init Segment V2 database and checking if v2 db exists 

using V2 is opt in for nodes that are upgrading and True for new installations
 
this is opt in for running nodes

size of proofs:
   old version 13.0797 with 20 sub epochs sampled             
   new version - 10.4372 mb with 140 sub epochs sampled
  
validation times:
    old version: 20.8 sec
    new version: 60 sec

serialisation from bytes times:
   old version 1.3 sec
   new version 1.1 sec

creating V2 segments:
  10 sec for non blueboxed sub epochs
  1.5 sec for blueboxed (this is the same as  old version)


chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
...
     sub_epoch_segments: List[SubEpochChallengeSegment]  # sampled sub epoch
     recent_chain_data: List[HeaderBlock]
+
+
+@streamable
+@dataclass(frozen=True)
+class SubSlotDataV2(Streamable):
+    proof_of_space: Optional[ProofOfSpace]  # if challenge block
+    # VDF to signage point
+    cc_signage_point: Optional[VDFProof]  # if block
```
mariano54, line 113: what is the difference between `if challenge block`, and `if block`. Aren't we only verifying the challenge blocks?

chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
...
+    # VDF from signage to infusion point
+    cc_infusion_point: Optional[VDFProof]  # if block
+    signage_point_index: Optional[uint8]  # if block
+    cc_slot_end: Optional[VDFProof]  # if is end of slot
+    cc_sp_vdf_output: Optional[CompressedClassgroupElement]  # if is end of slot
+    cc_ip_vdf_output: Optional[CompressedClassgroupElement]  # if is end of slot
+    cc_slot_end_output: Optional[ClassgroupElement]  # if is end of slot
```
mariano54, line 120: Why not compressed?

chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
...
+@dataclass(frozen=True)
+class SubSlotDataV2(Streamable):
+    proof_of_space: Optional[ProofOfSpace]  # if challenge block
+    # VDF to signage point
+    cc_signage_point: Optional[VDFProof]  # if block
+    # VDF from signage to infusion point
+    cc_infusion_point: Optional[VDFProof]  # if block
+    signage_point_index: Optional[uint8]  # if block
+    cc_slot_end: Optional[VDFProof]  # if is end of slot
+    cc_sp_vdf_output: Optional[CompressedClassgroupElement]  # if is end of slot
```
mariano54, line 118: What do you mean "if is end of slot"? Shouldn't we always validate cc_sp_vdf_output if we have a challenge block?

chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
...
+    cc_ip_vdf_output: Optional[CompressedClassgroupElement]  # if is end of slot
+    cc_slot_end_output: Optional[ClassgroupElement]  # if is end of slot
+    icc_infusion_point: Optional[VDFProof]
+    icc_ip_vdf_output: Optional[CompressedClassgroupElement]
+
+    icc_slot_end: Optional[VDFProof]
+    icc_slot_end_output: Optional[ClassgroupElement]
```
mariano54, line 125: compressed?

chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
...
+    cc_sp_signature: Optional[G2Element]  # if challenge block (needed for ChallengeBlockInfo)
+    ip_iters: Optional[uint64]  # if block
```
mariano54, line 127: can we obtain this from PoSpace?

chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
...
+    total_iters: Optional[uint128]  # if block
```
mariano54, line 128: can we calculate this using slots and pospace?

chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
     sub_epochs: List[SubEpochData]
     sub_epoch_segments: List[SubEpochChallengeSegment]  # sampled sub epoch
     recent_chain_data: List[HeaderBlock]
+
+
+@streamable
+@dataclass(frozen=True)
+class SubSlotDataV2(Streamable):
```
mariano54, line 110: 1. If it's an empty subslot everything will be None right?
2. If we combine the Optionals into one structure, we can save some space (one byte per Optional) and maybe make it less complex

chia/wallet/wallet_node.py
```diff
@@ -1086,37 +1101,62 @@ async def fetch_and_validate_the_weight_proof(
     ) -> Tuple[bool, Optional[WeightProof], List[SubEpochSummary], List[BlockRecord]]:
         assert self.wallet_state_manager is not None
         assert self.wallet_state_manager.weight_proof_handler is not None
+        capabilities = peer.capabilities
```
mariano54, line 1104: same comments as full node.
Also the code here is very similar, maybe we can merge it?

mariano54 (commented):
Nice work. I made a lot of questions because I'm still trying to understand the code.
This is the first pass review and I will review `weight_proof_v2.py` after the response to the comments.

chia/types/weight_proof.py
almogdepaz: i dont think so because we only have pospace for the challenge blocks

chia/full_node/block_store.py
```diff
@@ -249,6 +263,30 @@ async def get_sub_epoch_challenge_segments(
             return challenge_segments
         return None
 
+    async def persist_sub_epoch_challenge_segments_v2(
```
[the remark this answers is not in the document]
almogdepaz: yes, it is saved by blockhash so reorgs should not be an issue, this is the same as the old version basically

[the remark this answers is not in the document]
almogdepaz: yes, it is saved by blockhash so reorgs should not be an issue, this is the same as the old version basically

chia/consensus/blockchain.py
```diff
@@ -687,9 +687,7 @@ def get_ses_heights(self) -> List[uint32]:
     def get_ses(self, height: uint32) -> SubEpochSummary:
         return self.__height_map.get_ses(height)
 
-    def height_to_hash(self, height: uint32) -> Optional[bytes32]:
-        if not self.__height_map.contains_height(height):
-            return None
+    def height_to_hash(self, height: uint32) -> bytes32:
```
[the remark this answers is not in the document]
almogdepaz: yes i checked, actually i think this is safer.

we always assume hight < peak is in hight_to_hash, in most places we assert and fail but in some places we don't do that.

if im not wrong about the hight_map implementation, it fails with an assertion if hight is not in hight_to_hash, maybe we should even change that to throw an exception ?

chia/full_node/full_node.py
```diff
@@ -879,11 +886,29 @@ async def _sync(self):
...
+            self.log.debug(f"capabilities {capabilities} ")
+            weight_proof_v2 = False
+            ses_response = None
+            seed = None
+            if capabilities is not None and (uint16(Capability.WP.value), "1") in capabilities:
+                weight_proof_v2 = True
+                self.log.info("using new weight proof format")
+            if weight_proof_v2:
+                request = full_node_protocol.RequestSubEpochSummary(heaviest_peak_height)
+                ses_response = await weight_proof_peer.request_sub_epoch_summary(request, timeout=10)
```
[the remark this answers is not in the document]
almogdepaz: 67

chia/full_node/full_node.py
```diff
@@ -1652,6 +1690,8 @@ async def respond_block(
         if self.weight_proof_handler is not None and record.sub_epoch_summary_included is not None:
             if self._segment_task is None or self._segment_task.done():
                 self._segment_task = asyncio.create_task(self.weight_proof_handler.create_prev_sub_epoch_segments())
+            if self.weight_proof_handler_v2 is not None:
```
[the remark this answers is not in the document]
almogdepaz: what do we do with old clients if we don't have these available ?

chia/types/weight_proof.py
almogdepaz: we are still validating vdfs from non challenge blocks in the case where its no blueboxed

chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
...
+            return True
+        return False
+
+
+@streamable
+@dataclass(frozen=True)
+class SubEpochChallengeSegmentV2(Streamable):
+    sub_epoch_n: uint32
+    sub_slot_data: List[SubSlotDataV2]
+    # only in first segment of each sub_epoch
```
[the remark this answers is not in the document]
almogdepaz: these fields are only present in the segment of each sub_epoch because we, we need them to validate all the hashes and vdfs.

we dont need them for the other parts of the subepoch because they are obtained from the previous  SubEpochChallengeSegmentV2

chia/types/blockchain_format/vdf.py
```diff
@@ -96,3 +100,64 @@ class CompressibleVDFField(IntEnum):
     ICC_EOS_VDF = 2
     CC_SP_VDF = 3
     CC_IP_VDF = 4
+
+
+def compress_future(
+    disc: str,
```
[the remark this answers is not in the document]
almogdepaz: look in verify_vdf() line 44, in the old vdf code when call the c code we cast to string, the type in the methods in chiavdf is string, that string is later cast to an int maybe thats a bug in chiavdf ?

chia/consensus/vdf_info_computation.py
```diff
@@ -152,3 +152,34 @@ def get_signage_point_vdf_info(
         sp_vdf_iters,
         sp_vdf_iters,
     )
+
+
+async def get_sp_input(
```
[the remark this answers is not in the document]
almogdepaz: actually this isent used, its old code i forgot to delete, thanks

chia/full_node/full_node.py
```diff
@@ -248,6 +253,7 @@ def sql_trace_callback(req: str):
         self.transaction_responses: List[Tuple[bytes32, MempoolInclusionStatus, Optional[Err]]] = []
 
         self.weight_proof_handler = None
+        self.weight_proof_handler_v2 = None
```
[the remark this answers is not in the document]
mariano54: Ah i see, since we still need to return the old ones. I think it's OK to do it this way then, but can we add a comment explaining that both are needed because we support returning both versions of the weight proof?

chia/full_node/full_node.py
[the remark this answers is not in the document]
mariano54: I forgot about that

chia/types/weight_proof.py
mariano54: Basically for some of these objects, if one of them is present, then the rest will also be present. So instead of having an Optional for each, you can group them up into another stremable, and just have one big Optional[OtherStreamable]

chia/types/weight_proof.py
mariano54: So are you saying that one subslot might be composed of many `SubSlotDataV2` objects, one for each block, if they are not blueboxed? If so we should put a comment for this

chia/types/weight_proof.py
mariano54: if we know the ip_iters we can calculate the total_iters right?

chia/types/weight_proof.py
almogdepaz: i dont think that simplify things because of edge cases and the fact that we replace the proofs for None in challenge slots that are not sampled, we would still need to have optionals inside the different structures

chia/types/weight_proof.py
almogdepaz: sure, how ?

you just stared validating a new subepoch (yo dont know the total_iters of the previous block), what is the total iters of the first block given we have ip_iters x ?

maybe if we include the total_iters in the start of each sub_epoch sample

chia/server/server.py
```diff
@@ -150,7 +150,10 @@ def __init__(
         self.on_connect: Optional[Callable] = None
         self.incoming_messages: asyncio.Queue = asyncio.Queue()
         self.shut_down_event = asyncio.Event()
-
+        self.capabilities: List[Tuple[uint16, str]] = [
+            (uint16(Capability.BASE.value), "1"),
+            (uint16(Capability.WP.value), "1"),
```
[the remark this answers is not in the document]
almogdepaz: this way we can iterate versions using the integer, iirc its a need you raised when we talked about capabilities.

if we just call it WP_v2 there is no need for the integer in the Tuple

chia/types/blockchain_format/vdf.py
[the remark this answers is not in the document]
almogdepaz: we need to change chiavdf for that, ill ask florin if there was a reason for the use of string

chia/types/weight_proof.py
almogdepaz: yes this works the same way as the old proofs in that respect

chia/types/weight_proof.py
almogdepaz: getting from the compressed form (B) to the regular ClassgroupElement means we need to validate the vdf (its almost the same code)

that means that if we use the B form for the slots we cant really validate the hashes without validating the slot vdfs, and that means we cant really use the optimization where we validate vdfs only for one challenge slot in each sub_epoch

chia/server/server.py
```diff
@@ -150,7 +150,10 @@ def __init__(
         self.on_connect: Optional[Callable] = None
         self.incoming_messages: asyncio.Queue = asyncio.Queue()
         self.shut_down_event = asyncio.Event()
-
+        self.capabilities: List[Tuple[uint16, str]] = [
```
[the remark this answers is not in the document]
almogdepaz: i remembered why i did this, streamable does not support enums and im not sure i want to touch streamable as part of this pr

chia/types/blockchain_format/vdf.py
```diff
@@ -96,3 +100,64 @@ class CompressibleVDFField(IntEnum):
...
+    constants: ConsensusConstants,
+    challenge: bytes32,
+    vdf_input: ClassgroupElement,
+    vdf_output: B,
+    proof: VDFProof,
+    number_of_iterations: uint64,
+) -> Tuple[bool, ClassgroupElement]:
+    if proof.witness_type + 1 > constants.MAX_VDF_WITNESS_SIZE:
+        raise Exception(f"invalid witness type")
+    x, y = verify_n_wesolowski_with_b(
```
mariano54, line 155: what is x, is that a bool for whether the proof was valid? If so, we should rename it. 
We can also use type annotations in the line above, to make it more clear:
```
is_valid: bool
output_y: ClassgroupElement
```

mariano54: What is returned if the proof is false?
Do we have any tests for this function?

chia/types/weight_proof.py
mariano54: Ah i see. Would it be worth doing that? It would save some data (128 bits for each of these).

chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
...
+    icc_ip_vdf_output: Optional[B]
+
+    icc_slot_end: Optional[VDFProof]  # if is end of slot
+    icc_slot_end_output: Optional[ClassgroupElement]  # if is end of slot
+    cc_sp_signature: Optional[G2Element]  # if challenge block (needed for ChallengeBlockInfo)
+    ip_iters: Optional[uint64]  # if block
+    total_iters: Optional[uint128]  # if block
+
+    def is_challenge(self) -> bool:
+        if self.proof_of_space is not None:
```
mariano54, line 131: This can be simplified:
`return self.proof_of_space is not None`
Also the same for `is_end_of_slot`

tests/block_tools.py
```diff
@@ -103,7 +103,7 @@
         "DISCRIMINANT_SIZE_BITS": 16,
         "SUB_EPOCH_BLOCKS": 170,
         "WEIGHT_PROOF_THRESHOLD": 2,
-        "WEIGHT_PROOF_RECENT_BLOCKS": 380,
+        "WEIGHT_PROOF_RECENT_BLOCKS": 600,
```
mariano54, line 106: Why are we increasing this?

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1758 @@
...
+from chia.util.block_cache import BlockCache
+from chia.util.hash import std_hash
+from chia.util.ints import uint8, uint32, uint64, uint128
+from chia.util.streamable import recurse_jsonify, dataclass_from_dict
+
+log = logging.getLogger(__name__)
+
+
+class WeightProofHandlerV2:
+    LAMBDA_L = 100
```
mariano54, line 46: Can we add a link to the flyclient paper and or explain where these values come from?

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1758 @@
...
+    C = 0.5
+    MAX_SAMPLES = 140
+
+    def __init__(
+        self,
+        constants: ConsensusConstants,
+        blockchain: BlockchainInterface,
+    ):
+        self.tip: Optional[bytes32] = None
```
mariano54, line 55: peak?

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1758 @@
...
+        summaries, sub_epoch_weight_list = _validate_sub_epoch_summaries(self.constants, weight_proof)
+        if summaries is None:
+            log.warning("weight proof failed to validate sub epoch summaries")
+            return False, uint32(0)
+        return True, self.get_fork_point(summaries)
+
+    async def validate_weight_proof(
+        self, weight_proof: WeightProofV2, seed: bytes32, skip_segments: bool = False
+    ) -> Tuple[bool, uint32, List[SubEpochSummary], List[BlockRecord]]:
+        valid, summaries, records = await validate_weight_proof_no_fork_point(
```
mariano54, line 95: Why do we have this method of validation without fork point? Is getting the fork point slow? If not, I think it's better that have less methods

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1758 @@
...
+    async def _create_proof_of_weight(self, tip: bytes32, seed: bytes32) -> Optional[WeightProofV2]:
+        """
+        Creates a weight proof object
+        """
+        tip_rec = self.blockchain.try_block_record(tip)
+        if tip_rec is None:
+            log.error("failed not tip in cache")
+            return None
+        log.info(f"create weight proof peak {tip} {tip_rec.height}")
+        recent_chain_task = asyncio.create_task(get_recent_chain(self.blockchain, tip_rec.height))
```
mariano54, line 131: We should use type annotations for these variables as much as possible

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1758 @@
...
+        rng = random.Random(seed)
+        last_ses_block, prev_prev_ses_block = await self.get_last_l(summary_heights, tip_rec.height)
+        if last_ses_block is None or prev_prev_ses_block is None:
+            log.error("failed getting chain last L")
+            return None
+        last_l_weight = uint128(last_ses_block.weight - prev_prev_ses_block.weight)
+        log.debug(f"total weight {last_ses_block.weight} prev weight {prev_prev_ses_block.weight}")
+        weight_to_check = _get_weights_for_sampling(rng, last_ses_block.weight, last_l_weight)
+        if weight_to_check is None:
+            log.debug("chain to light, will choose all sub epochs until cap is reached")
```
mariano54, line 151: typo: `chain is too light`

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1758 @@
...
+        if fork_point_index > 2:
+            # Two summeries can have different blocks and still be identical
+            # This gets resolved after one full sub epoch
+            height = ses_heights[fork_point_index - 2]
+        else:
+            height = uint32(0)
+
+        return height
+
+    async def _create_proof_of_weight(self, tip: bytes32, seed: bytes32) -> Optional[WeightProofV2]:
```
mariano54, line 122: How long does this take for example for a node in mainnet?

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1758 @@
...
+        return WeightProofV2(sub_epoch_data, compressed_sub_epochs, recent_chain)
+
+    async def get_last_l(
+        self, summary_heights: List[uint32], peak: uint32
+    ) -> Tuple[Optional[BlockRecord], Optional[BlockRecord]]:
+        summaries_n = len(summary_heights)
+        for idx, height in enumerate(reversed(summary_heights)):
+            if height <= peak:
+                if summaries_n - idx < 3:
+                    log.warning("chain to short not enough sub epochs ")
```
mariano54, line 214: typo: `too short`

chia/full_node/weight_proof_v2.py
almogdepaz: i did this to be compatible to the wallet 
https://github.com/Chia-Network/chia-blockchain/blob/9f48b5ac6c4899179498f076908739bcf6596078/chia/wallet/wallet_node.py#L1147

ill switch to the  normal one and just not take the fork_point

chia/full_node/weight_proof_v2.py
almogdepaz: if the database is fully populated with the sub_epoch objects this should take about 4 sec' 
the old one takes 2 sec'. 


i tried optimizing this as best i can but currently out of ideas

chia/full_node/weight_proof_v2.py
almogdepaz: Thanks

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1837 @@
...
+    slot_iters, slots = uint64(0), 0
+    output_cache: Dict[B, ClassgroupElement] = {}
+    sub_slot_data = segment.sub_slot_data[start_from:]
+    first_block = True
+    prev_cc_challenge = None
+    after_challenge_block = False
+    slot_after_challenge_block = False
+    deficit = 0
+    for idx, ssd in enumerate(sub_slot_data):
+        if ssd.is_challenge():
```
mariano54, line 905: What happens if none of them are challenge? Shouldn't we check that there are at least 12 challenge blocks per subepoch?

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1837 @@
...
+        raise Exception(f"failed cc infusion point vdf validation {ssd_idx}")
+    long_outputs[sub_slot_data.cc_ip_vdf_output] = ip_output
+
+    pospace_challenge = challenge
+    assert sub_slot_data.signage_point_index is not None
+    if sampled:
+        if is_overflow_block(constants, sub_slot_data.signage_point_index):
+            assert prev_challenge
+            pospace_challenge = prev_challenge
+        __validate_pospace(constants, sub_slot_data, curr_difficulty, pospace_challenge, ssi, long_outputs)
```
mariano54, line 1225: Shouldn't we also check that the iterations returned from the POSpace are correct? Anyone can make a Pospace with a trillion iterations

mariano54 (changes requested):
We are going to need tests to cover all the edge cases of the sub methods

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1837 @@
...
+    tip_height = recent_chain.recent_chain_data[-1].height
+    prev_block_record = None
+    deficit = uint8(0)
+    for idx, block in enumerate(recent_chain.recent_chain_data):
+        required_iters = uint64(0)
+        overflow = False
+        ses = False
+        height = block.height
+        for sub_slot in block.finished_sub_slots:
+            prev_challenge = challenge
```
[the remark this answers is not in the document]
almogdepaz: i tried asking about this, we have no repro and i dont really understand the change you did there, i am not sure what the bug is, i can change this anyway but this feels wrong.

tests/block_tools.py
almogdepaz: we need 3 sub epochs for the sampling since Benedikt asked for little changes, one of them is that last L is the weight of the last 2 sub epochs, if we use 2 like before we have an issue with the sample method since the chain weight would be equal to Last L

chia/full_node/weight_proof_v2.py
almogdepaz: we can add a check counting the number of challenges, anything out of place here would cause the vdf and hash validation to fail since everything depends on the challenges

chia/full_node/weight_proof_v2.py
almogdepaz: good point i missed that, added

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1837 @@
...
+        assert sub_slot_data.cc_sp_vdf_output
+        long_outputs[sub_slot_data.cc_sp_vdf_output] = sp_output
+    cc_ip_input = ClassgroupElement.get_default_element()
+    assert sub_slot_data.ip_iters
+    ip_vdf_iters = sub_slot_data.ip_iters
+    assert sub_slot_data.cc_infusion_point is not None
+    if not prev_ssd.is_end_of_slot() and not sub_slot_data.cc_infusion_point.normalized_to_identity:
+        assert prev_ssd.cc_ip_vdf_output
+        assert sub_slot_data.total_iters
+        assert prev_ssd.total_iters
```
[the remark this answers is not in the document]
almogdepaz: actually these asserts are just for lint purposes im not sure changing this to Exceptions would be more readable

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1837 @@
...
+    """
+    sub_slot_data = sub_slots[ssd_idx]
+    prev_ssd = None
+    if ssd_idx > 0:
+        prev_ssd = sub_slots[ssd_idx - 1]
+    assert sub_slot_data.signage_point_index is not None
+    assert sub_slot_data.proof_of_space
+    assert sub_slot_data.ip_iters
+    sp_info = None
+    if sub_slot_data.cc_signage_point:
```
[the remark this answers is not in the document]
almogdepaz: sp iters should be 0 in that case so i can just check that instead and throw if sub_slot_data.cc_signage_point is None

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1837 @@
...
+    icc_challenge: bytes32,
+    long_outputs: Dict[B, ClassgroupElement],
+) -> None:
+    """
+    validate vdfs from a block
+    """
+    sub_slot_data = sub_slots[sub_slot_idx]
+    prev_ssd = sub_slots[sub_slot_idx - 1]
+    # find next end of slot
+    if sub_slot_data.cc_signage_point:
```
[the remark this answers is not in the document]
almogdepaz: this code is similar to the code you commented on in _validate_challenge_sub_slot_data
so like there i would just change to check sp_iters !=0

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1837 @@
...
+    returns the result and the list of blocks records as bytes
+    """
+    constants, summaries = bytes_to_vars(constants_dict, summaries_bytes)
+    recent_chain: RecentChainData = RecentChainData.from_bytes(recent_chain_bytes)
+    sub_blocks = BlockCache({})
+    first_ses_idx = _get_ses_idx(recent_chain.recent_chain_data)
+    ses_idx = len(summaries) - len(first_ses_idx)
+    ssi: uint64 = constants.SUB_SLOT_ITERS_STARTING
+    diff: Optional[uint64] = constants.DIFFICULTY_STARTING
+    last_blocks_to_validate = 100
```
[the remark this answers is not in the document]
almogdepaz: this takes lots of cpu and runs in parallel to the sub epoch validation, increasing it will make the validation times longer, ill move this to the handler params for better visibilit.

chia/types/weight_proof.py
```diff
@@ -100,3 +103,63 @@ class WeightProof(Streamable):
...
+# this is used only for serialization purposes using streamable
+class SubEpochSegmentsV2(Streamable):
+    challenge_segments: List[SubEpochChallengeSegmentV2]
+
+
+@streamable
+@dataclass(frozen=True)
+class WeightProofV2(Streamable):
+    sub_epochs: List[SubEpochData]
+    sub_epoch_segments: List[bytes]  # list where each bytes item represents List[SubEpochChallengeSegmentV2]
```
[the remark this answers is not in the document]
almogdepaz: this saves lots of time on conversions since we pass  sub_epoch_segments into multiproc for validation and that excepts bytes.

these are the objects that represent the sub epoch's so they are not small

chia/types/weight_proof.py
almogdepaz: since we remove these field from all the parts where we dont do vdf validation i think it wont be significant i can do some runs tomorrow and find out exactly

almogdepaz: after checking it will save about 0.1 mb out of about 10 total i dont think we should change this

chia/consensus/blockchain.py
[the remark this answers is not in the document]
almogdepaz: im reverting this, we can add it in a separate pr so it will be clear

chia/full_node/weight_proof_v2.py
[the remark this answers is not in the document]
mariano54: We have to check this, otherwise the code is insecure and all the checks can be bypassed

chia/full_node/weight_proof_v2.py
```diff
@@ -0,0 +1,1837 @@
...
+        cc_ip_input,
+        sub_slot_data.cc_ip_vdf_output,
+        sub_slot_data.cc_infusion_point,
+        ip_vdf_iters,
+    )
+    if not ip_valid:
+        raise Exception(f"failed cc infusion point vdf validation {cc_challenge}")
+    long_outputs[sub_slot_data.cc_ip_vdf_output] = ip_output
+
+    if sub_slot_data.icc_infusion_point is not None:
```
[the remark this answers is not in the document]
mariano54: Maybe add some asserts? I'm not 100% confident that every single case is covered

chia/full_node/weight_proof_v2.py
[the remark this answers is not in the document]
almogdepaz: some numbers, each extra 100 blocks we validate adds about 10 sec to the validation

PR #9717, opened 2022-01-03
replaced npm install => npm ci

seeden:
used npm ci instead of npm install 
it will use pinned dependencies for npm packages from our package-lock.json

### DIFF
--- a/Install-gui.ps1
+++ b/Install-gui.ps1
@@ -22,7 +22,7 @@ try {
     Set-Location chia-blockchain-gui
 
     $ErrorActionPreference = "SilentlyContinue"
-    npm install --loglevel=error
+    npm ci --loglevel=error
     npm audit fix
     npm run build
     py ..\installhelper.py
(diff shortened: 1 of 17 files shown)


emlowe (dismissed):

ChiaMineJP:
There are global `npm install` in build scripts. While global npm package is useful in that it can be executed from terminal directly, exact package version cannot be specified in package-lock.json because it is for local `npm install`.
We should  fix the version of npm package even for global npm package.

### Option1) Specify npm version in build script. e.g. `npm install -g lerna@4.0.0`
This is a direct approach. But there is one concern. If there is already another version of the same npm package installed globally, the latest `npm install -g xxx@a.b.c` replaces existing one.

### Option2) Make it local npm dependency and execute 'locally installed' npm package
This approach is safer than option1. Besides, it may give us an opportunity for 'parallel build', meaning building installers for macos, debian, windows can be run at the same time. (We should make sure whether it is actually possible though. There might be something which prevent parallel build)

Regarding to Option2, I'll investigate deeper and report it back soon.

wjblanke:
I assume ChiaMine JP's PR makes this one redundant

wjblanke:
lol ok merging this correctly. thanks cmj!

PR #9719, opened 2022-01-03
Allow CATs to be sent in multiple ways at once

Quexington:
This is a recreation of https://github.com/Chia-Network/chia-blockchain/pull/9691.  Thanks @HiveProject2021 for the start.

The change adds the functionality to use the `/send_transaction_multi` for CATs as well as standard XCH.


ChivesProject:
The previous account have fork a chives version. 
Cannot sumbit file use github client.
So I change this github accout to submit the change.
I will integrate this function to /send_transaction_multi and rewrite the /send_transaction to send cat at one api.

PR #9728, opened 2022-01-05
Only install NodeJS on Ubuntu if it's not already available

### LINKED ISSUE #9727: [Bug] install-gui.sh failing on Ubuntu 20.04
### What happened?

I tried running the `install-gui.sh` script from a terminal after activating my python virtual environment and it failed on the line `sudo apt-get install -y npm nodejs libxss1`.

I tried to research the error, but I couldn't find anything helpful. Trying to install npm and nodejs isn't necessary since I already have npm and node installed through nvm, the Node Version Manager, so I think checking if they're already available would be a suitable solution.

I don't think it's recommended to install these programs through apt-get because the versions there are usually a bit old. For instance, the version of nodejs in apt-get is v14.18.2 while v16.x.x has been active LTS since October.

### Version

1.2.12.dev124 (built from 13ff748)

Linux

CLI

### Relevant log output

```shell
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help resolve the situation:

The following packages have unmet dependencies:
 nodejs : Conflicts: npm
 npm : Depends: node-gyp (>= 3.6.2~) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

Jahorse:
This change to `install-gui.sh` will skip the attempt to install the npm and nodejs packages through `apt-get` if the `npm` and `node` commands are already available.

Fixes https://github.com/Chia-Network/chia-blockchain/issues/9727

### DIFF
--- a/install-gui.sh
+++ b/install-gui.sh
@@ -33,7 +33,14 @@ if [ "$(uname)" = "Linux" ]; then
 					echo >&2 "Please install NODEJS&NPM manually"
 			}
 		else
-			sudo apt-get install -y npm nodejs libxss1
+			program_list=( "libxss1" )
+			if ! [ -x "$(command -v npm)" ]; then
+			    prorgam_list+=( "npm" )
+			fi
+			if ! [ -x "$(command -v node)" ]; then
+			    program_list+=( "nodejs" )
+			fi
+			sudo apt-get install -y "${program_list[@]}"
 		fi
 	elif type yum &&  [ ! -f "/etc/redhat-release" ] && [ ! -f "/etc/centos-release" ] && [ ! -f /etc/rocky-release ] && [ ! -f /etc/fedora-release ]; then
 		# AMZN 2


jack60612:
@Jahorse Could you try fixing the lint issues? [Link](https://github.com/Chia-Network/chia-blockchain/blob/main/CONTRIBUTING.md )
The link above should help you setup a linter. If it does not help let me know.

Jahorse:
I was able to use the provided link to setup black, flake8, and mypy, but it looks like the bash linter that runs in the CI job is coming from GitHub's Super Linter, which needs Docker to setup locally, and I don't have it installed. I looked at the linting failures and they looked straightforward to fix, so I tried to just fix them without the linter.

If it's still broken I'll get Docker set up and do it properly.

Jahorse:
I noticed that the "Test install scripts" job has failed to run `install-gui.sh` with the following error:
```
install-gui.sh: 36: Syntax error: "(" unexpected (expecting "fi")
```
That's a line I've updated, but I cannot replicate the error and it looks like valid bash to me.

emlowe:
We are making several related changes to the install script. Your PR is very much appreciated but likely will be included in a bigger PR where we fix several other items.
I'll leave this open until then
@ChiaMineJP Can you take a look at this, thanks

install-gui.sh
```diff
@@ -33,7 +33,14 @@ if [ "$(uname)" = "Linux" ]; then
...
 		else
-			sudo apt-get install -y npm nodejs libxss1
+			program_list=( "libxss1" )
+			if ! [ -x "$(command -v npm)" ]; then
+			    prorgam_list+=( "npm" )
+			fi
+			if ! [ -x "$(command -v node)" ]; then
+			    program_list+=( "nodejs" )
+			fi
+			sudo apt-get install -y "${program_list[@]}"
```
ChiaMineJP, line 43: Even if `nodejs` is already installed, `sudo apt-get` tries to run whatever.
Ideally if nodejs/npm are installed already, `apt-get` don't need to be run at all.

By the way, we're planning a bunch of updates to the `install-gui.sh`.
I tried to rebase the upcoming PR onto your branch to express our-thanks but turned out number of conflicts are required to be resolved.

So I've decided to manage the updates to `install-gui.sh` in https://github.com/Chia-Network/chia-blockchain/pull/9921
Much appreciate your work and your patience.

ChiaMineJP:
Hello.
As I wrote in [this issue ](https://github.com/Chia-Network/chia-blockchain/issues/9727), the install scripts was overhauled by another PR.
Much appreciate your work and patience.

PR #9735, opened 2022-01-05
Add coin announcement to create signed transaction rpc call

jack60612:
You can use this to easily link transactions together.
*  for pool block claim fees


chia/rpc/wallet_rpc_api.py
```diff
@@ -1185,14 +1185,30 @@ async def create_signed_transaction(self, request, hold_lock=True):
         if "coins" in request and len(request["coins"]) > 0:
             coins = set([Coin.from_json_dict(coin_json) for coin_json in request["coins"]])
 
+        coin_announcements = None
```
mariano54, line 1188: Can we use a type annotation here?

chia/rpc/wallet_rpc_client.py
```diff
@@ -180,17 +180,34 @@ async def get_farmed_amount(self) -> Dict:
         return await self.fetch("get_farmed_amount", {})
 
     async def create_signed_transaction(
-        self, additions: List[Dict], coins: List[Coin] = None, fee: uint64 = uint64(0)
+        self,
+        additions: List[Dict],
+        coins: List[Coin] = None,
+        fee: uint64 = uint64(0),
+        coin_announcements: bytes32 = None,
```
mariano54, line 187: This should be List[bytes32] right?

mariano54 (changes requested):

chia/rpc/wallet_rpc_api.py
jack60612: Added

chia/rpc/wallet_rpc_client.py
jack60612: Fixed

mariano54:
A few more comments:
- Undo changes to files not relevant to change: `.flake8`, `.isort.cfg`.
- Add a test to `test_full_node_rpc.py` to make sure it is working
- Fix lint until all the checks are green

chia/rpc/wallet_rpc_api.py
```diff
@@ -1185,14 +1185,30 @@ async def create_signed_transaction(self, request, hold_lock=True):
         if "coins" in request and len(request["coins"]) > 0:
             coins = set([Coin.from_json_dict(coin_json) for coin_json in request["coins"]])
 
+        coin_announcements: Set[bytes32] = None
```
mariano54, line 1188: I think you will have to change this to Optional[Set[bytes32]] because None is not a Set. 
Or initialize to set()

mariano54 (dismissed):

mariano54 (approved):

PR #9736, opened 2022-01-05
collect streamable metadata early, allow ignored fields

altendky:
Mariano was interested in caching the `HeaderBlock.header_hash` property but we couldn't add it as a dataclass field/attribute because then the streamable system would pick up on it.  There were various other options but they all felt to me like workarounds for streamable presuming that every annotation or field (depending which bit of streamable you consider) must be part of what gets streamed.  With this change we will be able to calculate the header hash one time at the time of creation and assign it as an attribute such that all other uses will avoid both recalculation and traversal through the descriptor/property protocols and function calls.

https://github.com/Chia-Network/chia-blockchain/blob/9cc908678b1255c9a520c322302ba35084676e08/chia/types/header_block.py#L43-L45

Draft for:
- [ ] Consider being able to indicate to streamable what callable to use for building an object when parsing.  Maybe this can help avoid `.__post_init__()` and the `.__setattr__()` games inside.  This will likely run into some issues with inheritance and avoiding the skipping of the `strictdataclass()`-added type checking.
- [x] Check for performance effects using `benchmarks/coin_store.py`

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -145,7 +145,7 @@ def unstreamed_field(*args, **kwargs):
     metadata = kwargs.setdefault("metadata", {})
     metadata[_field_metadata_key] = _FieldMetadata(ignore=True)
 
-    return dataclasses.fields(*args, **kwargs)
+    return dataclasses.field(*args, **kwargs)
 
 
 _T_Field = TypeVar("_T_Field", bound="_Field")
@@ -351,10 +351,20 @@ class Streamable:
     def parse(cls: Type[cls.__name__], f: BinaryIO) -> cls.__name__:  # type: ignore
         # Create the object without calling __init__() to avoid unnecessary post-init checks in strictdataclass
         obj: Streamable = object.__new__(cls)
-        fields: Iterator[str] = iter(getattr(cls, "__annotations__", {}))
+        try:
+            bare_fields = cls._chia_streamable.fields
+        except AttributeError:
+            bare_fields = []
+        fields: Iterator[str] = iter(field.name for field in bare_fields)
         values: Iterator = (parse_f(f) for parse_f in PARSE_FUNCTIONS_FOR_STREAMABLE_CLASS[cls])
         for field, value in zip(fields, values):
             object.__setattr__(obj, field, value)
+
+        # TODO: should we handle inheritance here?  Or should the post init super() for
+        #       itself?
+        post_init = getattr(cls, "__post_init__", None)
+        if post_init is not None:
+            post_init(obj)
 
         # Use -1 as a sentinel value as it's not currently serializable
         if next(fields, -1) != -1:
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -3,11 +3,12 @@
 from __future__ import annotations
 
 import dataclasses
+import functools
 import io
 import pprint
 import sys
 from enum import Enum
-from typing import Any, BinaryIO, Dict, List, Tuple, Type, Callable, Optional, Iterator
+from typing import Any, BinaryIO, ClassVar, Dict, List, Tuple, Type, TypeVar, Callable, Optional, Iterator
 
 from blspy import G1Element, G2Element, PrivateKey
 from typing_extensions import Literal
@@ -69,7 +70,7 @@ def dataclass_from_dict(klass, d):
         return tuple(klass_properties)
     elif dataclasses.is_dataclass(klass):
         # Type is a dataclass, data is a dictionary
-        fieldtypes = {f.name: f.type for f in dataclasses.fields(klass)}
+        fieldtypes = {f.name: f.annotation for f in klass._chia_streamable.fields}
         return klass(**{f: dataclass_from_dict(fieldtypes[f], d[f]) for f in d})
     elif is_type_List(klass):
         # Type is a list, data is a list
@@ -127,8 +128,55 @@ def recurse_jsonify(d):
 
 PARSE_FUNCTIONS_FOR_STREAMABLE_CLASS = {}
 
+_field_metadata_key = "_chia_streamable"
 
-def streamable(cls: Any):
+
+@dataclasses.dataclass(frozen=True)
+class _FieldMetadata:
+    ignore: bool = False
+
+
+@functools.wraps(wrapped=dataclasses.field)
+def unstreamed_field(*args, **kwargs):
+    metadata = kwargs.setdefault("metadata", {})
+    metadata[_field_metadata_key] = _FieldMetadata(ignore=True)
+
+    return dataclasses.fields(*args, **kwargs)
+
+
+_T_Field = TypeVar("_T_Field", bound="_Field")
+
+
+@dataclasses.dataclass(frozen=True)
+class _Field:
+    name: str
+    annotation: Any
+
+    @classmethod
+    def from_dataclass_field(cls: Type[_T_Field], field: dataclasses.Field) -> _T_Field:
+        return cls(name=field.name, annotation=field.type)
+
+
+_T_ClassMetadata = TypeVar("_T_ClassMetadata", bound="_ClassMetadata")
+
+
+@dataclasses.dataclass(frozen=True)
+class _ClassMetadata:
+    fields: Tuple[_Field, ...] = ()
+
+    @classmethod
+    def from_dataclass(cls: Type[_T_ClassMetadata], dataclass) -> _T_ClassMetadata:
+        fields = []
+        for dataclass_field in dataclasses.fields(dataclass):
+            metadata = dataclass_field.metadata.get(_field_metadata_key, _FieldMetadata())
+            if metadata.ignore:
+                continue
+            fields.append(_Field.from_dataclass_field(field=dataclass_field))
+
+        return cls(fields=tuple(fields))
+
+
+def streamable(cls: Type["Streamable"]) -> Type["Streamable"]:
     """
     This is a decorator for class definitions. It applies the strictdataclass decorator,
     which checks all types at construction. It also defines a simple serialization format,
@@ -179,12 +227,14 @@ def streamable(cls: Any):
 
     parse_functions = []
     try:
-        fields = cls1.__annotations__  # pylint: disable=no-member
+        metadata = _ClassMetadata.from_dataclass(dataclass=cls1)
     except Exception:
-        fields = {}
+        metadata = _ClassMetadata()
 
-    for _, f_type in fields.items():
-        parse_functions.append(cls.function_to_parse_one_item(f_type))
+    cls1._chia_streamable = metadata
+
+    for field in cls1._chia_streamable.fields:
+        parse_functions.append(cls.function_to_parse_one_item(field.annotation))
 
     PARSE_FUNCTIONS_FOR_STREAMABLE_CLASS[t] = parse_functions
     return t
@@ -259,6 +309,8 @@ def parse_str(f: BinaryIO) -> str:
 
 
 class Streamable:
+    _chia_streamable: ClassVar[_ClassMetadata]
+
     @classmethod
     def function_to_parse_one_item(cls: Type[cls.__name__], f_type: Type):  # type: ignore
         """
@@ -347,11 +399,11 @@ class Streamable:
 
     def stream(self, f: BinaryIO) -> None:
         try:
-            fields = self.__annotations__  # pylint: disable=no-member
+            fields = self._chia_streamable.fields
         except Exception:
-            fields = {}
-        for f_name, f_type in fields.items():
-            self.stream_one_item(f_type, getattr(self, f_name), f)
+            fields = ()
+        for field in fields:
+            self.stream_one_item(field.annotation, getattr(self, field.name), f)
 
     def get_hash(self) -> bytes32:
         return bytes32(std_hash(bytes(self)))
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -364,7 +364,7 @@ class Streamable:
         #       itself?
         post_init = getattr(cls, "__post_init__", None)
         if post_init is not None:
-            post_init(obj)
+            post_init(obj, parsed=True)
 
         # Use -1 as a sentinel value as it's not currently serializable
         if next(fields, -1) != -1:
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -70,7 +70,11 @@ def dataclass_from_dict(klass, d):
         return tuple(klass_properties)
     elif dataclasses.is_dataclass(klass):
         # Type is a dataclass, data is a dictionary
-        fieldtypes = {f.name: f.annotation for f in klass._chia_streamable.fields}
+        metadata = getattr(klass, "_chia_streamable", None)
+        if metadata is not None:
+            fieldtypes = {f.name: f.annotation for f in metadata.fields}
+        else:
+            fieldtypes = {f.name: f.type for f in dataclasses.fields(klass)}
         return klass(**{f: dataclass_from_dict(fieldtypes[f], d[f]) for f in d})
     elif is_type_List(klass):
         # Type is a list, data is a list
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -3,12 +3,11 @@
 from __future__ import annotations
 
 import dataclasses
-import functools
 import io
 import pprint
 import sys
 from enum import Enum
-from typing import Any, BinaryIO, ClassVar, Dict, List, Tuple, Type, TypeVar, Callable, Optional, Iterator
+from typing import Any, BinaryIO, ClassVar, Dict, List, Tuple, Type, Callable, Optional, Iterator
 
 from blspy import G1Element, G2Element, PrivateKey
 from typing_extensions import Literal
@@ -17,6 +16,7 @@ from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.byte_types import hexstr_to_bytes
 from chia.util.hash import std_hash
 from chia.util.ints import int64, int512, uint32, uint64, uint128
+from chia.util.streamablemetadata import _ClassMetadata
 from chia.util.type_checking import is_type_List, is_type_SpecificOptional, is_type_Tuple, strictdataclass
 
 if sys.version_info < (3, 8):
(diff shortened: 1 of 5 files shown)


mariano54:
We should check if this affects performance at all. Maybe run the coin store benchmarks?

mariano54:
This also seems to fix python 3.10 issues due to not using `__annotations__`.

altendky:
Full log below but the overall average here from 8 interlaced runs of the coin store benchmark against the v2 db is 310 seconds with this and 308 seconds without.  Run on my old Xeon v0 or whatever server.  So, a 0.6% slow down.  @mariano54, would you like me to run more including v1 to verify this and look into fixing it?  Or is this measurement a non-concern for you.

```
 ==== 0
     ---- after: /farm/dev/chia-blockchain
version 2
Building database 
85.0496s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9603s, MOSTLY REMOVALS additions: 600 removals: 140000
122.8009s, FULLBLOCKS additions: 400400 removals: 400000
2.1333s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
58.8546s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.4726s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 308.2713s
database size: 353.501 MB
     ---- before: /farm/dev/chia-blockchain-reference
version 2
Building database 
84.3573s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9106s, MOSTLY REMOVALS additions: 600 removals: 140000
122.9460s, FULLBLOCKS additions: 400400 removals: 400000
2.1285s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
58.8728s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.3847s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 307.5999s
database size: 353.501 MB
 ==== 1
     ---- after: /farm/dev/chia-blockchain
version 2
Building database 
84.4966s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
16.0584s, MOSTLY REMOVALS additions: 600 removals: 140000
123.2611s, FULLBLOCKS additions: 400400 removals: 400000
2.1718s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
58.8565s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.6733s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 308.5177s
database size: 353.501 MB
     ---- before: /farm/dev/chia-blockchain-reference
version 2
Building database 
84.6572s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9880s, MOSTLY REMOVALS additions: 600 removals: 140000
123.5316s, FULLBLOCKS additions: 400400 removals: 400000
2.1428s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
58.7345s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.3141s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 308.3683s
database size: 353.501 MB
 ==== 2
     ---- after: /farm/dev/chia-blockchain
version 2
Building database 
84.9157s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9869s, MOSTLY REMOVALS additions: 600 removals: 140000
123.6582s, FULLBLOCKS additions: 400400 removals: 400000
2.1401s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
59.2616s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.4983s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 309.4608s
database size: 353.501 MB
     ---- before: /farm/dev/chia-blockchain-reference
version 2
Building database 
84.4568s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9007s, MOSTLY REMOVALS additions: 600 removals: 140000
123.2219s, FULLBLOCKS additions: 400400 removals: 400000
2.1809s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
59.1271s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.4379s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 308.3254s
database size: 353.501 MB
 ==== 3
     ---- after: /farm/dev/chia-blockchain
version 2
Building database 
84.4480s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.8794s, MOSTLY REMOVALS additions: 600 removals: 140000
122.2898s, FULLBLOCKS additions: 400400 removals: 400000
2.1317s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
59.1197s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.3339s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 307.2024s
database size: 353.501 MB
     ---- before: /farm/dev/chia-blockchain-reference
version 2
Building database 
84.8716s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
16.0716s, MOSTLY REMOVALS additions: 600 removals: 140000
123.6902s, FULLBLOCKS additions: 400400 removals: 400000
2.1261s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
59.0698s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.3613s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 309.1905s
database size: 353.501 MB
 ==== 4
     ---- after: /farm/dev/chia-blockchain
version 2
Building database 
84.3712s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
16.0327s, MOSTLY REMOVALS additions: 600 removals: 140000
123.0725s, FULLBLOCKS additions: 400400 removals: 400000
2.1307s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
60.8545s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.3019s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 309.7633s
database size: 353.501 MB
     ---- before: /farm/dev/chia-blockchain-reference
version 2
Building database 
84.1937s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
16.0181s, MOSTLY REMOVALS additions: 600 removals: 140000
123.0506s, FULLBLOCKS additions: 400400 removals: 400000
2.2934s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
59.6895s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.3463s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 308.5917s
database size: 353.501 MB
 ==== 5
     ---- after: /farm/dev/chia-blockchain
version 2
Building database 
85.0485s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9613s, MOSTLY REMOVALS additions: 600 removals: 140000
123.4609s, FULLBLOCKS additions: 400400 removals: 400000
2.1147s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
59.0568s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.3156s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 308.9577s
database size: 353.501 MB
     ---- before: /farm/dev/chia-blockchain-reference
version 2
Building database 
84.7077s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9197s, MOSTLY REMOVALS additions: 600 removals: 140000
123.3466s, FULLBLOCKS additions: 400400 removals: 400000
2.1357s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
59.1997s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.4072s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 308.7167s
database size: 353.501 MB
 ==== 6
     ---- after: /farm/dev/chia-blockchain
version 2
Building database 
87.3444s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9927s, MOSTLY REMOVALS additions: 600 removals: 140000
124.1297s, FULLBLOCKS additions: 400400 removals: 400000
2.1772s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
61.7977s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.7060s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 315.1478s
database size: 353.501 MB
     ---- before: /farm/dev/chia-blockchain-reference
version 2
Building database 
84.3457s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9264s, MOSTLY REMOVALS additions: 600 removals: 140000
122.9270s, FULLBLOCKS additions: 400400 removals: 400000
2.0725s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
58.7056s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.3345s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 307.3118s
database size: 353.501 MB
 ==== 7
     ---- after: /farm/dev/chia-blockchain
version 2
Building database 
85.0659s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
16.1087s, MOSTLY REMOVALS additions: 600 removals: 140000
123.0103s, FULLBLOCKS additions: 400400 removals: 400000
2.1592s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
61.1586s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.4736s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 310.9763s
database size: 353.501 MB
     ---- before: /farm/dev/chia-blockchain-reference
version 2
Building database 
84.3242s, MOSTLY ADDITIONS additions: 400400.0 removals: 20000.0
15.9182s, MOSTLY REMOVALS additions: 600 removals: 140000
123.6314s, FULLBLOCKS additions: 400400 removals: 400000
2.1424s, GET RECORDS BY NAMES with spent 200 lookups found 40000 coins in total
58.8607s, GET RECORDS BY NAMES without spent 200 lookups found 20641 coins in total
23.2863s, GET COINS REMOVED AT HEIGHT 800 blocks, found 580000 coins in total
all tests completed in 308.1632s
database size: 353.501 MB
```

mariano54:
No, 0.6% is not significant and not worth spending time on. I would like to merge this soon as it helps with python 3.10

mariano54:
@richardkiss could you review this one? I am not so comfortable doing so.

tests/core/util/test_streamable.py
```diff
@@ -373,6 +374,64 @@ def test_parse_str(self):
...
+        @streamable
+        class Ignored(Streamable):
+            a: uint32
+            b: uint32
+            c: List[uint32]
+            d: List[List[uint32]]
+            e: Optional[uint32]
+            f: Optional[uint32]
+            g: Tuple[uint32, str, bytes]
+            m: uint32 = unstreamed_field(init=False)
```
richardkiss, line 404: It seems like we're initializing `m` to a value that is not a `uint32`. I don't understand, can you explain? @altendky

chia/util/streamable.py
```diff
@@ -295,11 +351,21 @@ def function_to_parse_one_item(cls: Type[cls.__name__], f_type: Type):  # type:
...
+        except AttributeError:
+            bare_fields = []
+        fields: Iterator[str] = iter(field.name for field in bare_fields)
         values: Iterator = (parse_f(f) for parse_f in PARSE_FUNCTIONS_FOR_STREAMABLE_CLASS[cls])
         for field, value in zip(fields, values):
             object.__setattr__(obj, field, value)
 
+        # TODO: should we handle inheritance here?  Or should the post init super() for
+        #       itself?
+        post_init = getattr(cls, "__post_init__", None)
```
richardkiss, line 365:
```suggestion
        post_init = getattr(cls, "__post_init__", lambda *args, **kwargs: 0)
```

then you can just call it no matter what.

richardkiss (commented):
This is all very scary and confusing. My guess is it comes about from 3.10 no longer supporting `__annotations__`. Is that right?

Is there any chance that using [`__init_subclass__`](https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__) could simplify this? We could build the parse and stream tables at that time and save it away in the subclass, and then possibly greatly simplify `parse` and `stream`.

richardkiss:
Oh, I see, this is for adding cacheable elements that aren't serialized. I see a couple options that might make it simpler:

- standard names for unserialized fields (for example, `unstremaed_` prefix `unstremaed_header_hash`)
- use a file-level `weakref.WeakKeyDictionary` that keeps the cache around (and so entries go away when the objects do)
```
   def header_hash(self):
        if self not in CACHE_DICT:
            CACHE_DICT[self] = calculate_header_hash_somehow(self)
        return CACHE_DICT[self]
```

altendky (commented):
Scary and confusing?  I thought the previous inconsistent use of `.__annotations__` was the scary part, as shown with the breakage in 3.10 (though I didn't look into what really broke).  `:]`  The caching of the header hash was the inspiration to look at this but it is very much a self-justified shift to a cleaner, more explicit, and more extensible design.  Or so I thought...

I'll set my perspective.  dataclasses provides three basic things.  A concise way to describe class attributes and get some boilerplate written for you (`.__init__`, `.__repr__`, etc), a form and featureset that encourages clean and flexible classes, and finally a framework for introspecting classes and their attributes including custom metadata.

Now some direct responses to your comment.

Using `.__init_subclass__` seems to me to be increasing the dependence on inheritance without a clear upside.  We already have a decorator that runs right after the class is created so there's little difference in timing etc.  But sure, we may well be able to shift even more to the time of class definition to simplify runtime behavior.  That seems like a separate task and independent of the approach taken here as long as something continues to happen around class creation time.

Embedding metadata into the attribute names doesn't strike me as good.  It's very implicit and forces awkward code.  What happens when we end up wanting another bit of metadata other than ignoring?  The dataclass' field's metadata is exactly intended for this usage.  "It is not used at all by Data Classes, and is provided as a third-party extension mechanism. Multiple third-parties can each have their own key, to use as a namespace in the metadata." (https://docs.python.org/3.10/library/dataclasses.html#dataclasses.field)

For the next one I'll start with: you just suggested a global!  `:]`  But anyways, this is all global what with being class attributes.  But the global cache suggestion does add _runtime_ global mutation.  It also leaves a function call in the middle of every access.  Neither of these particular aspects are terrible here but they don't seem obviously good.  Well, if we particularly feel the need to only conditionally calculate this attribute then yeah, some adjustment would be made.  I was under the impression the header hash was likely to be accessed multiple times on 'every' instance.  I guess this relates to the direct justification of this PR by another feature we want (the caching) but I think it doesn't really relate to the specific design change submitted here.

I did consider leaving the metadata in the dataclass fields but one of the early points Mariano made was that they recalled `dataclass.fields()` or such to be considered relevantly slow so I instead duplicated the data out to simple classes we could directly access when wanted.

Given this scares and confuses you but seems like a much tidier system to me, perhaps a call would be good to discuss it?  This all went in roughly how I imagined it based on other mechanisms I've built around attrs-defined classes.

altendky:
For the record from the Python 3.10 changelog:

https://docs.python.org/3.10/whatsnew/changelog.html#id116
> Enable `from __future__ import annotations` (PEP 563) by default. The values found in `__annotations__` dicts are now strings, e.g. `{"x": "int"}` instead of `{"x": int}`.

tests/core/util/test_streamable.py
richardkiss: Ah, got it, it's an extension of the `dataclasses` magic.

> Scary and confusing? I thought the previous inconsistent use of `.__annotations__` was the scary part, as shown with the breakage in 3.10 (though I didn't look into what really broke). `:]` The caching of the header hash was the inspiration to look at this but it is very much a self-justified shift to a cleaner, more explicit, and more extensible design. Or so I thought...

Apologies for this comment... it's more me reacting to the fact that I don't recognize this file anymore, and it's gotten so complicated that I don't feel like I can get my head around it enough to properly review this patch without setting aside significant time to study it. I fully expect given your history that the work you've done is likely well-reasoned and thoughtful of trade-offs. I just don't understand it yet.

chia/util/streamable.py
richardkiss: No biggie either way really.

chia/util/streamable.py
```diff
@@ -69,7 +70,11 @@ def dataclass_from_dict(klass, d):
         return tuple(klass_properties)
     elif dataclasses.is_dataclass(klass):
         # Type is a dataclass, data is a dictionary
-        fieldtypes = {f.name: f.type for f in dataclasses.fields(klass)}
+        metadata = getattr(klass, "_chia_streamable", None)
+        if metadata is not None:
```
richardkiss, line 74: Why not `isinstance(klass, Streamable)`? Is it so we can use the `_chia_streamable` tag in other non-`Streamable` objects?

chia/util/streamable.py
```diff
@@ -127,8 +132,55 @@ def recurse_jsonify(d):
...
+            metadata = dataclass_field.metadata.get(_field_metadata_key, _FieldMetadata())
+            if metadata.ignore:
+                continue
+            fields.append(_Field.from_dataclass_field(field=dataclass_field))
+
+        return cls(fields=tuple(fields))
+
 
-def streamable(cls: Any):
+def streamable(cls: Type["Streamable"]) -> Type["Streamable"]:
```
richardkiss, line 183: Super-pedantic, but doesn't this technically take any class type and turn it into a `Streamable`? As in
```suggestion
def streamable(cls: Type["Any"]) -> Type["Streamable"]:
```

chia/util/streamable.py
```diff
@@ -179,12 +231,14 @@ def streamable(cls: Any):
 
     parse_functions = []
     try:
-        fields = cls1.__annotations__  # pylint: disable=no-member
+        metadata = _ClassMetadata.from_dataclass(dataclass=cls1)
```
richardkiss, line 234: Ah, got it, you're generalizing `fields` here. How would you feel about moving the introspection utilities and types (like `_ClassMetadata` and `unstreamed_field` to their own file? That might make it more clear what their intended interface is.

chia/util/streamable.py
altendky: No, my intent was not to try to work well with not-`Streamable` instances.

I think at one point I just straight up did `klass._chia_streamable` because I thought it was always a `Streamable`...  but then a test failed that used it on a raw `dataclass`.  There are a few places in this change where there is more checking than I feel like I would have designed in but it aligns better with prior behavior.  I could certainly explore cleaning up the conditionals and try to set this all up to be exclusively presuming a proper `Streamable`.

I'll take another look at that unrepresentative test and double check my memory on the uses of `dataclass_from_dict()`.  Unless you want to retain this working for not-`Streamable` instances?  Otherwise the `isinstance()` is reasonable in this case I suppose.  Mostly I try to tend away from it.

altendky (commented):
Well, it would certainly be good for there to be more knowledge about the code.  I would be happy to have a discussion with you, undoubtedly I would learn something as well, or maybe I should take this up with someone else to pass on the torch?  If the latter, any particular suggestions?

chia/util/streamable.py
richardkiss: There's not actually a reason to cache this metadata, since any function that uses this could be dynamically "written" at class-definition time. See how `PARSE_FUNCTIONS_FOR_STREAMABLE_CLASS` are built up. Not sure if `_chia_streamable` is used elsewhere or not, but this might obviate the need for it.

chia/util/streamable.py
altendky: Uh...  so it does...  This is all a bit odd what with the classes first inheriting from `Streamable` then getting passed here where we create a `dataclass` from that then creating another class inheriting from that and `Streamable` (again) and then we always decorate with `dataclass` (again) after this as I recall...

My thinking was that we want this applied to a `Streamable` subclass so hinting that would encourage people to not apply it without having inherited from `Streamable`.  Perhaps with proper type checks we would even become comfortable removing the inheritance that is applied below?  I dunno, I should probably just not delve into all this now.  Maybe just an `isinstance()` with an error instead of reapplying the inheritance in this function?

So, back to what to hint, if anything.  As noted at the end of the docstring below, I think we want people to have applied the `Streamable` inheritance themselves so I think we would rather hint as it is to encourage that?

richardkiss: That does sound right that when we've created a new class intended to be streamable, we both subclass and decorate it. But maybe we shouldn't have been doing that?

tests/core/util/test_streamablemetadata.py
```diff
@@ -0,0 +1,65 @@
...
+    @streamable
+    class Ignored(Streamable):
+        a: uint32
+        b: uint32
+        c: List[uint32]
+        d: List[List[uint32]]
+        e: Optional[uint32]
+        f: Optional[uint32]
+        g: Tuple[uint32, str, bytes]
+        m: uint32 = unstreamed_field(init=False)
```
richardkiss, line 36: How about
```
        m: uint32 = unstreamed_field(lambda self: self.a + uint32(1))
```
?

altendky:
this is being dealt with in other ways

PR #9740, opened 2022-01-06
Made global npm modules be executed as local npm modules

ChiaMineJP:
Additional commits for https://github.com/Chia-Network/chia-blockchain/pull/9717
As I noted in the above PR, even if `npm ci` replaces `npm install`, global npm modules version is still variable.
To fix the version of global npm modules, I made a choice that let global npm modules be executed as local npm modules.

### DIFF
--- a/.github/workflows/build-linux-arm64-installer.yml
+++ b/.github/workflows/build-linux-arm64-installer.yml
@@ -110,11 +110,6 @@ jobs:
       run: |
         sh install.sh
 
-    - name: Setup Node 16.x
-      uses: actions/setup-node@v2.4.1
-      with:
-        node-version: '16.x'
-
     - name: Build arm64 .deb package
       run: |
         . ./activate
--- a/.github/workflows/build-linux-arm64-installer.yml
+++ b/.github/workflows/build-linux-arm64-installer.yml
@@ -110,6 +110,11 @@ jobs:
       run: |
         sh install.sh
 
+    - name: Setup Node 16.x
+      uses: actions/setup-node@v2.4.1
+      with:
+        node-version: '16.x'
+
     - name: Build arm64 .deb package
       run: |
         . ./activate
--- a/build_scripts/build_linux_deb.sh
+++ b/build_scripts/build_linux_deb.sh
@@ -24,9 +24,10 @@ fi
 echo "Chia Installer Version is: $CHIA_INSTALLER_VERSION"
 
 echo "Installing npm and electron packagers"
-npm install electron-packager -g
-npm install electron-installer-debian -g
-npm install lerna -g
+cd npm_linux_deb || exit
+npm ci
+PATH=$(npm bin):$PATH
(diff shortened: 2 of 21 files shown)


emlowe:
I like this idea, although we control the build systems but this is better really.

Do we want to change the user install script - `install-gui.sh`? It has some similar issues where it installs things globally I think.

ChiaMineJP:
I've found global npm install in `install-gui.sh` as below.
```bash
if [ "$UBUNTU_PRE_2004" = "True" ]; then
	echo "Installing on Ubuntu older than 20.04 LTS: Ugrading node.js to stable."
	UBUNTU_PRE_2004=true # Unfortunately Python returns True when shell expects true
	sudo npm install -g n
	sudo n stable
	export PATH="$PATH"
fi
```

Umm. If I remember correctly, `n` package is `nodejs`/`npm` version switcher like `nvm`.
In addition to fix the version of `n`, `sudo n stable`  also brings version variability.
I think improving `install-gui.sh` should be handled in another PR.
A big change will be inevitable if we go deep about `install-gui.sh`.

emlowe:
Yes that makes sense...
You are right,  `n` is a node package manager like `nvm`

ChiaMineJP:
Umm. 'Linux ARM64 installer on Python 3.8' not working.
I'll look into the root cause.

image

ChiaMineJP:
@emlowe @seeden @cmmarslender 
ARM64 installer issue has been now resolved. The root cause was corrupted npm cache folder.
Now all global npm modules can be installed and executed locally.
Please review and merge if it's good. (Currently base branch is set to `npm_install`)

emlowe (approved):
I like this.

.github/workflows/build-linux-arm64-installer.yml
```diff
@@ -110,6 +110,11 @@ jobs:
       run: |
         sh install.sh
 
+    - name: Setup Node 16.x
```
cmmarslender, line 113: This shouldn't be necessary - node should be preinstalled in the container we use (see line 17 of this file) `container: chianetwork/ubuntu-18.04-builder:latest`

That container comes from here https://github.com/Chia-Network/build-images/blob/main/ubuntu-18.04/Dockerfile#L34

cmmarslender (changes requested):

.github/workflows/build-linux-arm64-installer.yml
ChiaMineJP: Yep. This should be removed. Making a new commit.

ChiaMineJP:
Note: Failure on `LGTM analysis: Javascript` is already expected. This is resolved once LGTM see commits on `main` branch.
See https://github.com/Chia-Network/chia-blockchain/pull/9750

cmmarslender (approved):

seeden (approved):
LGTM thank you

PR #9746, opened 2022-01-07
Move BLS validation to other processes during sync

mariano54:
There are a few related components of this PR:

## Multiprocess BLS
During sync, blocks were validated in `block_body_validation` in the main process, which included BLS validation. This is arguably one of the slowest parts of block validation (at least when there are a lot of TX) so it can be optimized by moving this to the multiprocess prevalidation during sync. 

During normal operation, nothing is changed.

This yields a 13-36% sync time improvement in RPI certain situations (a lot of transactions), which is expected to be the norm eventually. Likely 30%+ improvement on machines with fast IO. Thank you @neurosis69 for the data and @altendky for identifying the optimization!
https://github.com/neurosis69/chia-sync-data

Furthermore, the BLS is done without the cache, to avoid the performance hit of repeatedly checking empty cache.

## Simplication of `blockchain.py`
This PR removes a block of code from `receive_block`, which is an essential method in the blockchain, and should not have clutter. The whole branch where npc_result is None, was not executed except for tests. This used to run the CLVM if it wasn't run before, but the full node always runs it before. This leads to the next change.

## Changing `test_blockchain.py`
Due to the above changes, the tests have to be changed to not use `receive_block` with NPCResult of None. Instead, we will do it the same way it is done in the full node, by first prevalidating and then calling `recieve_block` with the prevalidation result. A new convenience method is added for the tests, which does:
* prevalidation
* receive_block
* Checking consistency of DB before and after
* Checking that the error code is what we expect
* Checking that the RecieveBlockResult is what we expect

This adds some extra checking and validation in the tests, and also cleans up the test code slightly. Due to the extra checks, a few tests have been modified, because they were buggy and not testing properly.

## Return error CLVM error codes in prevalidation
Before this PR, prevalidation would run the CLVM but not check the error, and then proceed to validate the header. This was annoying because prevalidation would return an unrelated error that happened in header validation: INVALID_TRANSACTION_FILTER_HASH, so it was hard to see the exact error code. This is now changed to get the error if CLVM fails.


chia/consensus/blockchain.py
```diff
@@ -259,6 +259,7 @@ async def receive_block(
             npc_result,
             fork_point_with_peak,
             self.get_block_generator,
+            pre_validation_result is None or not pre_validation_result.validated_signature,
```
arvidn, line 262: it's not obvious what this bool means, just reading this code. I have to look up the `validate_block_body()` function to ensure it's correct. It might be a good idea to name this parameter.

arvidn: there's a valid return valid pre_validation_result that says it didn't validate signatures? (it seems a bit risky and unintuitive)

chia/consensus/multiprocess_validation.py
```diff
@@ -5,6 +5,8 @@
 from dataclasses import dataclass
 from typing import Dict, List, Optional, Sequence, Tuple, Union, Callable
 
+from chia.util import cached_bls
```
arvidn, line 8: my understanding is that the caching saves duplicate work between validating mempool entries and full blocks with those entries in them. while syncing, wouldn't the cache just slow us down?

chia/consensus/blockchain.py
```diff
@@ -628,14 +629,15 @@ async def validate_unfinished_block(
...
+            return PreValidationResult(uint16(error_code.value), None, None, False)
 
-        return PreValidationResult(None, required_iters, cost_result)
+        return PreValidationResult(None, required_iters, cost_result, False)
 
     async def pre_validate_blocks_multiprocessing(
         self,
         blocks: List[FullBlock],
         npc_results: Dict[uint32, NPCResult],
+        validate_signatures: bool,
```
arvidn, line 640: I think it would be safer to add `,*` before this, to force all callers to name the parameter. If there are calls to this function where the `batch_size` is passed in as positional, are you sure it couldn't be interpreted as a boolean?

chia/consensus/multiprocess_validation.py
mariano54: Yes you are right. We would compute each pairing individually and add them to the cache. If we avoid using cache here, we can batch verify which computes all the pairings together and is more efficient. I will make this change.

chia/consensus/blockchain.py
mariano54: It's meant to convey that signatures were not validated so we must validate them later. Otherwise we might forget to validate them in the future. Block body validation does the safe thing and validates signatures if the result is False.

arvidn: right. I realized later you added this field. I suppose it makes sense to be explicit about this, rather than deduce it at this point by looking at the error code (and possibly something else).

This line in particular could be made clearer though, I think. You could give this value a name, like:

```
validate_signatures = pre_validation_result is None or not pre_validation_result.validated_signature
```

mariano54: > it's not obvious what this bool means, just reading this code. I have to look up the `validate_block_body()` function to ensure it's correct. It might be a good idea to name this parameter.

OK i will clean that up

mariano54: That is only None in tests, that is annoying.. I think it's worth cleaning up though

chia/consensus/blockchain.py
mariano54: Good idea

chia/consensus/blockchain.py
altendky: In general, I would like to see named parameters roughly everywhere.  At one point I only half jokingly thought about setting my personal style guide to include `*` in every function def to force every parameter to be keyword only.

chia/consensus/blockchain.py
```diff
@@ -202,53 +215,12 @@ async def receive_block(
         if not genesis and (self.block_record(block.prev_header_hash).height + 1) != block.height:
             return ReceiveBlockResult.INVALID_BLOCK, Err.INVALID_HEIGHT, None, ([], {})
 
-        npc_result: Optional[NPCResult] = None
```
mariano54, line 205: All of this code was only being used for tests, since pre_validation_result is never None in full node.

chia/consensus/blockchain.py
```diff
@@ -202,53 +215,12 @@ async def receive_block(
...
-            if error is not None:
-                return ReceiveBlockResult.INVALID_BLOCK, error.code, None, ([], {})
-        else:
-            npc_result = pre_validation_result.npc_result
-            required_iters = pre_validation_result.required_iters
-            assert pre_validation_result.error is None
+        npc_result: Optional[NPCResult] = pre_validation_result.npc_result
+        required_iters = pre_validation_result.required_iters
+        if pre_validation_result.error is not None:
+            return ReceiveBlockResult.INVALID_BLOCK, Err(pre_validation_result.error), None, ([], {})
```
arvidn, line 221: this is more helpful than an assert if it happens in production for whatever reason. But during testing, I would expect an `assert` to be more helpful, since it also communicates that it's a programmer error. I guess `asserts` in python still are just exceptions, so they can still be quite easily swallowed and ignored by mistalke.

chia/consensus/blockchain.py
```diff
@@ -260,6 +232,8 @@ async def receive_block(
             npc_result,
             fork_point_with_peak,
             self.get_block_generator,
+            # If we did not already validate the signature, validate it now
+            validate_signature=pre_validation_result.validated_signature,
```
arvidn, line 236: didn't you mean to negate `pre_validation_result.validated_signature`?

arvidn: "validate_signature" sounds like asking the function (`validate_block_body()`) to perform the validation, and `validated_signature` sounds like it's saying whether the signature has been validated already.

chia/consensus/blockchain.py
```diff
@@ -580,7 +554,7 @@ async def validate_unfinished_block(
             not self.contains_block(block.prev_header_hash)
             and not block.prev_header_hash == self.constants.GENESIS_CHALLENGE
         ):
-            return PreValidationResult(uint16(Err.INVALID_PREV_BLOCK_HASH.value), None, None)
+            return PreValidationResult(uint16(Err.INVALID_PREV_BLOCK_HASH.value), None, None, False)
```
arvidn, line 557: it's hard to know what this `False` means here, without a named parameter or a named flag/enum.

chia/consensus/blockchain.py
```diff
@@ -606,8 +580,7 @@ async def validate_unfinished_block(
         )
 
         if error is not None:
-            return PreValidationResult(uint16(error.code.value), None, None)
-
+            return PreValidationResult(uint16(error.code.value), None, None, False)
```
arvidn, line 583: and here

chia/consensus/blockchain.py
```diff
@@ -625,20 +598,22 @@ async def validate_unfinished_block(
             npc_result,
             None,
             self.get_block_generator,
-            False,
+            validate_signature=False,  # Signature was already validated before calling this method, no need to val
```
arvidn, line 601: this comment looks truncated

chia/consensus/multiprocess_validation.py
```diff
@@ -102,11 +112,29 @@ def batch_pre_validate_blocks(
                 if error is not None:
                     error_int = uint16(error.code.value)
```
arvidn, line 113: if we get here, and set `error_int`, shouldn't we return?
as it is right now, it looks like we might overwrite this (true) error code with `Err.BAD_AGGREGATE_SIGNATURE` later

arvidn: it would be interesting to test this function with an invalid block but with a valid signature

chia/simulator/full_node_simulator.py
```diff
@@ -49,7 +50,13 @@ async def farm_new_transaction_block(self, request: FarmNewBlockProtocol):
             current_blocks = await self.get_all_full_blocks()
             if len(current_blocks) == 0:
                 genesis = self.bt.get_consecutive_blocks(uint8(1))[0]
-                await self.full_node.blockchain.receive_block(genesis)
+                pre_validation_results: Optional[
+                    List[PreValidationResult]
+                ] = await self.full_node.blockchain.pre_validate_blocks_multiprocessing(
```
arvidn, line 55: it seems to me the return type of this function should not be `Optional[...]`

altendky:
There was some skepticism about the accuracy or relevance of this data but here's what got collected.  Red is with this PR and blue without.  Both forced to four-core pools on my test RPi.  The dip right before 1M is a filtering artifact related to a stuck sync where all peers were reported to have zero height.  The node was otherwise responsive.  Another user reported a similar symptom combined with the `is_peak` on height 0 db issue which did not happen in my case.  They were running 1.2.11 so it seems maybe unrelated to this PR.  Restarting my node was sufficient to return to syncing.  I shared lots of notes with Mariano on keybase.

image

tests/blockchain/test_blockchain.py
```diff
@@ -2832,39 +2818,9 @@ async def test_get_blocks_at(self, empty_blockchain, default_1000_blocks):
...
-            await check_block_store_invariant(b)
+            await _validate_and_add_block(b, block)
 
         blocks = await b.get_block_records_at(heights, batch_size=2)
         assert blocks
         assert len(blocks) == 200
         assert blocks[-1].height == 199
-
-
-async def check_block_store_invariant(bc: Blockchain):
```
arvidn: I take it this was moved verbatim, right?

arvidn:
@altendky what are the units on the axis in that graph? I would guess y is the number of blocks processed per time-window, and x is seconds since you started syncing. Is that right?
If so, is the time window minutes?

chia/consensus/blockchain.py
mariano54: Either error will end up closing the connection so that stays the same. But an assert is actually not the right thing to do, because it's not programmer error, it's a malicious block I think.

chia/consensus/blockchain.py
mariano54: wow yes we need to negate it. What a huge mistake!

chia/consensus/blockchain.py
mariano54: Not sure if we can force named parameters in streamable

chia/consensus/multiprocess_validation.py
mariano54: Yes good point, the signature error will take precedence, kind of like if the signature was validated before the CLVM. It shouldn't cause the block to be valid though. If the `PreValidationResult` has an error, the block is completely invalid and will be thrown out. 

Still it does seem better to just add the error and continue instead of force validating which could be slow

chia/simulator/full_node_simulator.py
mariano54: What is that? What's wrong with `Optional[List[PreValidationResult]]`?

tests/blockchain/blockchain_test_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    result, err, _, _ = await blockchain.receive_block(block, results)
+    await check_block_store_invariant(blockchain)
+    if expected_error is None and expected_result != ReceiveBlockResult.INVALID_BLOCK:
+        # Not expecting any errors
+        if err is not None:
+            # Got an error
+            raise ValueError(err)
+    else:
+        # Expecting an error
+        if err != expected_error:
```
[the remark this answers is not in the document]
mariano54: Not necessarily. We might expect error A but we got error B. So that is not what we expected so the test should fail. 
The if statement handles the normal case where no errors are expected.

altendky:
Oops, that graph didn't get the axis labels.  Sorry.  It is blocks per hour (vertical) vs. block height (horizontal).

chia/consensus/blockchain.py
arvidn: it would be great to have a test that would catch this

chia/consensus/blockchain.py
arvidn: I see, this is a `dataclass` thing.

chia/simulator/full_node_simulator.py
arvidn: Virtuall all calls I saw to this function was followed by an

```
assert pre_validation_result is not None
```

Which to me suggests that (at least on all of those cases), the function cannot fail (or whatever a `None` return would mean). If the function was just made to never return `None`, we wouldn't need all those asserts. But I didn't look very deeply into what `None` means or under what conditions it's returned, it just looked a bit suspicious to me.

arvidn: I think I may have been looking at tests checking the return value.. sorry about that.

tests/blockchain/blockchain_test_utils.py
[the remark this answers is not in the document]
arvidn: right. if I'm reading the code correctly, that's exactly what this condition checks. Ensuring the expected error is exactly equal to the error we got.

What I'm saying is that I would expect this check to always be performed, even when
```
expected_error is None and expected_result != ReceiveBlockResult.INVALID_BLOCK
```

Right now, you don't check this in that case, which looks suspicious. Maybe it's correct, but it's not as obvious as if you would just always ensure we got the error (or lack thereof) as we expected.

tests/blockchain/blockchain_test_utils.py
```diff
@@ -0,0 +1,142 @@
...
+            # Did not get the right error, or did not get an error
+            raise ValueError(f"Expected {expected_error} but got {err}")
```
arvidn, line 89: is there a reason to not use `assert` here, rather than throwing `ValueError`?
in my mind at least, an `AssertionError` is a hard failure and means the test failed, whereas a `ValueError` might be expected for negative tests. I don't think these errors can ever mean the test still passes, right?

arvidn:
I'm a bit paranoid about tests that might pass despite its tests are failing. it makes me nervous when checks the tests itself is asserting might get mixed up with run-time errors that might happen during normal operation (and that might happen in negative tests), and vice versa.

I expect us to never catch `AssertionError`, but always let those bubble up and terminate the test, at least in tests. (I recognize we have an exception at the top of a task message loop, which makes sense).

mariano54:
> I'm a bit paranoid about tests that might pass despite its tests are failing. it makes me nervous when checks the tests itself is asserting might get mixed up with run-time errors that might happen during normal operation (and that might happen in negative tests), and vice versa.
> 
> I expect us to never catch `AssertionError`, but always let those bubble up and terminate the test, at least in tests. (I recognize we have an exception at the top of a task message loop, which makes sense).

I'll try to improve this.
Also, please do not benchmark yet as I have to fix a performance bug (the boolean issue arvid metioned).

chia/consensus/blockchain.py
mariano54: Roger that

tests/blockchain/test_blockchain.py
```diff
@@ -147,8 +153,13 @@ async def test_long_chain(self, empty_blockchain, default_1000_blocks):
...
                     block, "finished_sub_slots", [new_finished_ss_2] + block.finished_sub_slots[1:]
                 )
-                result, err, _, _ = await empty_blockchain.receive_block(block_bad_2)
-                assert err == Err.INVALID_NEW_DIFFICULTY
+                try:
+                    await _validate_and_add_block(
+                        empty_blockchain, block_bad_2, expected_error=Err.INVALID_NEW_DIFFICULTY
+                    )
+                except AssertionError:
+                    # This can also fail in prevalidation through an assertion error
```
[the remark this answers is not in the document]
mariano54: You're right, I was being lazy here. I will clean it up to actually raise errors instead of leaving those asserts there

tests/blockchain/blockchain_test_utils.py
mariano54: yeah you are right, makes sense to change to assertion error

tests/blockchain/blockchain_test_utils.py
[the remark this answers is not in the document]
mariano54: Oh yeah that is intentional, to simplify test code. I can say for example:

`_validate_and_add_block(blockchain, b, expected_result=ReceiveBlockResult.INVALID_BLOCK)`

The comment at the top of the function explains this. I will also add a comment at the corresponding place.

tests/blockchain/blockchain_test_utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+async def _validate_and_add_block_multi_error(
+    blockchain: Blockchain, block: FullBlock, expected_errors: List[Err], skip_prevalidation: bool = False
+) -> None:
+    # Checks that the blockchain returns one of the expected errors
+    try:
+        await _validate_and_add_block(blockchain, block, skip_prevalidation=skip_prevalidation)
+    except Exception as e:
+        assert isinstance(e, ValueError)
+        assert e.args[0] in expected_errors
```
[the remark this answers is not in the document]
mariano54: good point.

mariano54:
I have addressed all of your comments arvid, thank you for the very detailed and helpful review.

mariano54:
If you want to re-run the benchmark @neurosis69 that would be appreciated.

mariano54:
Looks like CIs are timing out due to the added validation time in tests

chia/simulator/full_node_simulator.py
```diff
@@ -49,7 +50,13 @@ async def farm_new_transaction_block(self, request: FarmNewBlockProtocol):
             current_blocks = await self.get_all_full_blocks()
             if len(current_blocks) == 0:
                 genesis = self.bt.get_consecutive_blocks(uint8(1))[0]
-                await self.full_node.blockchain.receive_block(genesis)
+                pre_validation_results: List[
+                    PreValidationResult
+                ] = await self.full_node.blockchain.pre_validate_blocks_multiprocessing(
+                    [genesis], {}, validate_signatures=True
+                )
+                assert pre_validation_results is not None
```
arvidn, line 58: I suppose this assert can be removed now, and the one on line 100

arvidn (dismissed):

PR #9284, opened 2021-11-16
Atari - blobs not text and SQLite checks

altendky:
This started with adding checks but they are much simpler with the switch to storing bytes into blobs rather than as hex strings.  If needed the checks could be separated out into a follow-up PR.

Draft for:
- [x] Merge https://github.com/Chia-Network/chia-blockchain/pull/9558 first as this contains it.
- [ ] Other things that have been merged will need reviewed and updated for the `TEXT` -> `BLOB` change as well.
- [ ] This isn't the priority right now.  Just getting it out of the list of PRs to merge.

PR #9748, opened 2022-01-07
Optimize get ancestors.

fchirica:


chia/data_layer/data_store.py
```diff
@@ -350,6 +399,42 @@ async def get_ancestors(self, node_hash: bytes32, tree_id: bytes32, *, lock: boo
...
+    async def get_ancestors_2(
+        self, node_hash: bytes32, tree_id: bytes32, generation: Optional[int] = None, lock: bool = True
+    ) -> List[InternalNode]:
+        nodes = []
+        if generation is None:
+            generation = await self.get_tree_generation(tree_id=tree_id, lock=False)
+        while True:
+            cursor = await self.db.execute(
+                """
+                SELECT * from node INNER JOIN (
```
fchirica, line 411: I left this query in non-recursive form for now, as SQL doesn't support aggregates with recursive.

chia/data_layer/data_store.py
```diff
@@ -81,6 +81,17 @@ async def create(cls, db_wrapper: DBWrapper) -> "DataStore":
...
+                """
+                CREATE TABLE IF NOT EXISTS ancestors(
+                    hash TEXT NOT NULL,
+                    ancestor TEXT,
+                    tree_id TEXT NOT NULL,
+                    generation INTEGER NOT NULL,
+                    PRIMARY KEY(hash, tree_id, generation)
+                )
+                """
+            )
```
altendky, line 94: This does not work as is, the `root` foreign key complains.  I didn't actually try to fix it but I think that may be able to be addressed by just moving the root insertion earlier than the ancestor insertion.

```suggestion
            # TODO: Add ancestor -> hash relationship, this might involve temporarily
            #       deferring the foreign key enforcement due to the insertion order
            #       and the node table also enforcing a similar relationship in the
            #       other direction.
            #       FOREIGN KEY(ancestor) REFERENCES ancestors(ancestor)
            await self.db.execute(
                """
                CREATE TABLE IF NOT EXISTS ancestors(
                    hash TEXT NOT NULL REFERENCES node,
                    ancestor TEXT,
                    tree_id TEXT NOT NULL,
                    generation INTEGER NOT NULL,
                    PRIMARY KEY(hash, tree_id, generation),
                    FOREIGN KEY(tree_id, generation) REFERENCES root(tree_id, generation),
                    FOREIGN KEY(ancestor) REFERENCES node(hash)
                )
                """
            )
```

altendky (changes requested):
Thanks for this!  It took me a few times thinking about the sparse nature of the ancestor entries to internalize why it worked, but it works out nicely for the data.  Makes the query a bit more complicated, but I think we can manage that.

chia/data_layer/data_store.py
```diff
@@ -350,6 +399,42 @@ async def get_ancestors(self, node_hash: bytes32, tree_id: bytes32, *, lock: boo
...
+                """
+                SELECT * from node INNER JOIN (
+                    SELECT ancestors.ancestor AS hash, MAX(ancestors.generation) AS generation
+                    FROM ancestors
+                    WHERE ancestors.hash == :hash
+                    AND ancestors.tree_id == :tree_id
+                    AND ancestors.generation <= :generation
+                    AND ancestors.ancestor is NOT NULL
+                    GROUP BY hash
+                ) asc on asc.hash == node.hash
```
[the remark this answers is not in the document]
altendky: So I looked a bit more and apparently `max()` and `min()` are special cased in SQLite and are guaranteed to do what you wanted here.  So I guess if we are ok with using an SQLite specific feature (which we probably are) then this is ok.

https://www.sqlite.org/lang_select.html#bareagg
> When the min() or max() aggregate functions are used in an aggregate query, all bare columns in the result set take values from the input row which also contains the minimum or maximum. So in the query above, the value of the "b" column in the output will be the value of the "b" column in the input row that has the largest "c" value. There is still an ambiguity if two or more of the input rows have the same minimum or maximum value or if the query contains more than one min() and/or max() aggregate function. Only the built-in min() and max() functions work this way.

chia/data_layer/data_store.py
```diff
@@ -350,6 +399,42 @@ async def get_ancestors(self, node_hash: bytes32, tree_id: bytes32, *, lock: boo
 
         return ancestors
 
+    async def get_ancestors_2(
```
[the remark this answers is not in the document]
fchirica: I kept original `get_ancestors` to test this function against it. Maybe rename with `get_ancestors_optimised`?

chia/data_layer/data_store.py
```diff
@@ -104,6 +115,21 @@ async def _insert_root(self, tree_id: bytes32, node_hash: Optional[bytes32], sta
             },
         )
```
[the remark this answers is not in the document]
fchirica: I've separated it into `_insert_ancestor_table` for getting table constraints to work.

chia/data_layer/data_store.py
```diff
@@ -81,6 +81,24 @@ async def create(cls, db_wrapper: DBWrapper) -> "DataStore":
...
+            await self.db.execute(
+                """
+                CREATE TABLE IF NOT EXISTS ancestors(
+                    hash TEXT NOT NULL REFERENCES node,
+                    ancestor TEXT,
+                    tree_id TEXT NOT NULL,
+                    generation INTEGER NOT NULL,
+                    PRIMARY KEY(hash, tree_id, generation),
+                    FOREIGN KEY(tree_id, generation) REFERENCES root(tree_id, generation),
+                    FOREIGN KEY(ancestor) REFERENCES node(hash)
```
arvidn, line 98: I think, basically, all fields of type `TEXT` should be `BLOB`. let's not store hashes in hex.

arvidn (changes requested):
it's not clear to me why we need the `get_ancestors` function at all. I'd love understand that. It would be even better if we could remove it.

chia/data_layer/data_store.py
```diff
@@ -139,7 +172,12 @@ async def _insert_node(
             if result_dict != values:
                 raise Exception(f"Requested insertion of node with matching hash but other values differ: {node_hash}")
 
-    async def _insert_internal_node(self, left_hash: bytes32, right_hash: bytes32) -> bytes32:
+    async def _insert_internal_node(
+        self,
+        left_hash: bytes32,
+        right_hash: bytes32,
+        tree_id: bytes32,
```
almogdepaz, line 179: tree_id not used here

chia/data_layer/data_store.py
altendky: Long since a plan, now not a priority to merge vs. functional changes.  https://github.com/Chia-Network/chia-blockchain/pull/9284  It will still get in at some point.  I added a note there to review other changes that have been made in the mean time so we get all the `TEXT` switched over.

PR #9752, opened 2022-01-07
Make announcements in RPC be objects instead of bytes

Quexington:
This is an additional change to the recent PR https://github.com/Chia-Network/chia-blockchain/pull/9735 that reworks how announcements are passed around.  With the coming `main_pac` branch we tend to treat them as objects rather than simply passing around their ID.  There's also commented code for when that merge happens to support puzzle announcements as well.


jack60612 (approved):
lgtm

PR #9753, opened 2022-01-07
misc hint'ish cleanup

altendky:


.github/workflows/upload-pypi-source.yml
```diff
@@ -49,6 +49,14 @@ jobs:
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install .[dev]
 
+    - name: Lint source with flake8
```
Yostra, line 52: do we need these here? They are ran by superlinter/pre-commit already

altendky: Shouldn't that be a separate discussion from PAC?

altendky: And specifically, mypy is run most correctly here.

Yostra: I guess, lets merge this

PR #9755, opened 2022-01-08
Fix npm install issue

ChiaMineJP:
This is just a test PR. Will remove anytime.


ChiaMineJP:
memo:
Gotcha. Issue is at cache.
Immediately uninstall `lerna` after installing it was OK. It's so weird since other than ARM64 there have been no issue.
image

PR #9349, opened 2021-11-22
Sql index changes

arvidn:
these changes are primarily not performance improvements, but makes the data base file smaller. Since there are fewer indices to maintain, there is likely a small performance gain as well.

PR #9768, opened 2022-01-09
revert allow-upgrades configuration option

arvidn:
It makes it significantly more complicated to do proper database migrations

### DIFF
--- a/benchmarks/utils.py
+++ b/benchmarks/utils.py
@@ -67,4 +67,4 @@ async def setup_db(name: str, db_version: int) -> DBWrapper:
     await connection.execute("pragma journal_mode=wal")
     await connection.execute("pragma synchronous=full")
 
-    return DBWrapper(connection, False, db_version)
+    return DBWrapper(connection, db_version)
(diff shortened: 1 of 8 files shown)


mariano54:
Will the people that already used it be in trouble for the migration?

arvidn:
I don't believe we ever released this feature, so it would be limited to people running from git `main`. And the extent of the "trouble" people will be in, is that the transition to `v2` will be slower, because the v2 transition tool currently assumes the v1 database is in the non-upgraded state.

So, there won't be any data loss, it's just some indices that will be re-created before starting the conversion.

arvidn:
The alternative to reverting this would be to make the transition code more sophisticated and detect (or load from settings) which indices to expect in the source DB file.

mariano54:
I see. Ok that's not a big issue then, it would be fine as long as it works.

arvidn:
The PR introducing this was here: https://github.com/Chia-Network/chia-blockchain/pull/9349
reviewed by @mariano54  and @emlowe so I'll add you as reviewers here too.

mariano54 (approved):

PR #9771, opened 2022-01-10
Some offer tx record changes

Quexington:
We were not creating an outgoing transaction when a maker notices that their offer has been taken.  This fixes that (as well as a few small changes I noticed in the process)


paninaro (commented):
lgtm

paninaro (approved):

PR #9778, opened 2022-01-11
[v2] remove unused counter column in hint table (for v2 schema)

arvidn:
apart from removing the index column, this also adds a constraint preventing duplicate (coin_id, hint) pairs in the table. Such duplicates probably don't come up a lot in real-life, but might during reorgs. We don't reverse and replay into the hints table.

### DIFF
--- a/chia/cmds/db_upgrade_func.py
+++ b/chia/cmds/db_upgrade_func.py
@@ -109,11 +109,9 @@ async def convert_v1_to_v2(in_path: Path, out_path: Path) -> None:
                 "block_record blob)"
             )
             await out_db.execute(
-                "CREATE TABLE IF NOT EXISTS sub_epoch_segments_v3("
-                "ses_block_hash blob PRIMARY KEY,"
-                "challenge_segments blob)"
+                "CREATE TABLE sub_epoch_segments_v3(" "ses_block_hash blob PRIMARY KEY," "challenge_segments blob)"
             )
-            await out_db.execute("CREATE TABLE IF NOT EXISTS current_peak(key int PRIMARY KEY, hash blob)")
+            await out_db.execute("CREATE TABLE current_peak(key int PRIMARY KEY, hash blob)")
 
             peak_hash, peak_height = await store_v1.get_peak()
             print(f"peak: {peak_hash.hex()} height: {peak_height}")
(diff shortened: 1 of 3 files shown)


chia/cmds/db_upgrade_func.py
```diff
@@ -237,22 +237,22 @@ async def convert_v1_to_v2(in_path: Path, out_path: Path) -> None:
             commit_in = HINT_COMMIT_RATE
             hint_start_time = time()
             hint_values = []
-            await out_db.execute("CREATE TABLE hints(id INTEGER PRIMARY KEY AUTOINCREMENT, coin_id blob,  hint blob)")
+            await out_db.execute("CREATE TABLE IF NOT EXISTS hints(coin_id blob, hint blob, UNIQUE (coin_id, hint))")
             await out_db.commit()
             async with in_db.execute("SELECT coin_id, hint FROM hints") as cursor:
```
Yostra, line 242: is there a way to check that upgrade has been finished successfully? (worry being if you start and stop full node before upgrade has been performed table will exist but it will be incomplete)

arvidn: the upgrade doesn't happen in-place, so to a new DB file. And if you "restart" if will fail because the target file already exists. There's no explicit record of the conversion completing successfully, maybe there should be. The problem you describe is not limited to the hints table though

arvidn: you say "start and stop full node", the conversion doesn't happen inside the full node, it's done as a manual command:
```
chia db upgrade
```

arvidn: I think I see what you mean. This should not say "IF NOT EXISTS". That table should never exist here, if it does, something's wrong and we need to fail. Fixed.

mariano54 (approved):

chia/cmds/db_upgrade_func.py
```diff
@@ -237,22 +235,22 @@ async def convert_v1_to_v2(in_path: Path, out_path: Path) -> None:
...
                     if commit_in == 0:
                         commit_in = HINT_COMMIT_RATE
-                        await out_db.executemany("INSERT INTO hints VALUES (?, ?, ?)", hint_values)
+                        await out_db.executemany("INSERT INTO hints VALUES(?, ?) ON CONFLICT DO NOTHING", hint_values)
                         await out_db.commit()
                         await out_db.execute("begin transaction")
                         hint_values = []
 
-            await out_db.executemany("INSERT INTO hints VALUES (?, ?, ?)", hint_values)
+            await out_db.executemany("INSERT INTO hints VALUES (?, ?)", hint_values)
```
neurosis69, line 253: "ON CONFLICT DO NOTHING" clause is missing here (ran into error on a test upgrade)

arvidn: good, point. this needs fixing and a test

arvidn: https://github.com/Chia-Network/chia-blockchain/pull/10041

PR #9780, opened 2022-01-11
sanity check the blockchain database on startup

arvidn:
If we can't find a peak, we also should not find any coins

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -214,6 +214,13 @@ class FullNode:
         time_taken = time.time() - start_time
         if self.blockchain.get_peak() is None:
             self.log.info(f"Initialized with empty blockchain time taken: {int(time_taken)}s")
+            num_unspent = await self.coin_store.num_unspent()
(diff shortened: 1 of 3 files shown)


chia/full_node/full_node.py
```diff
@@ -214,6 +214,13 @@ def sql_trace_callback(req: str):
         time_taken = time.time() - start_time
         if self.blockchain.get_peak() is None:
             self.log.info(f"Initialized with empty blockchain time taken: {int(time_taken)}s")
+            num_unspent = await self.coin_store.num_unspent()
+            if num_unspent > 0:
+                self.log.error(
+                    f"Inconsistent blockchain DB file! Could not found peak block but found {num_unspent} coins! "
```
hoffmang9, line 220: should be "Could not find peak"?

arvidn: oops

mariano54 (approved):
If this is fast to execute, then it makes sense.

PR #9786, opened 2022-01-12
plotting: Return `None` in `process_file` if the plot exists

xdustinface:
This avoids redundant insertions of the existing plots here 
https://github.com/Chia-Network/chia-blockchain/blob/0ba838b7a8ea2b5410d438ac70295df699a30dae/chia/plotting/manager.py#L463 
and avoids having them in the update call here 
https://github.com/Chia-Network/chia-blockchain/blob/0ba838b7a8ea2b5410d438ac70295df699a30dae/chia/plotting/manager.py#L464.

### DIFF
--- a/chia/plotting/manager.py
+++ b/chia/plotting/manager.py
@@ -341,7 +341,7 @@ class PlotManager:
                 return None
 
             if file_path in self.plots:
-                return self.plots[file_path]
+                return None
 
             entry: Optional[Tuple[str, Set[str]]] = self.plot_filename_paths.get(file_path.name)
             if entry is not None:


arvidn:
I agree. It's not so easy to tell whether this is correct or reasonable just by looking at this diff. Some comments (in the code) would be helpful

arvidn:
as far as I can tell, this looks correct. But the code seems a bit unnecessarily complex.

I think, ideally, `process_file()` would not depend on the existing list of plot files we know about. The name suggests that it will process the file, not that it will *maybe* process the file.

Filtering out plot files we already know about could be done before the call to `map()`, to just never call `process_file()` with known-good plots to begin with. I think that would be a simpler way of doing it (and maybe `process_file()` could even be made a free function, to clearly show it doesn't have dependencies on the `PlotManager` state).

Also, another question this raises for me is; what happens in the case where an existing good plot file is deleted? If we never process good plot files, do we detect them disappearing (before they pass the filter and we attempt to farm from it)?

arvidn (approved):

wjblanke:
maybe some more comments

xdustinface:
I have some refactoring of this part in the pipeline.

PR #9789, opened 2022-01-12
harvester: Reuse `plot_public_key` instead calculating it in each lookup

xdustinface:


mariano54 (approved):

jack60612 (approved):
Lgtm

wjblanke:
so i am worried about this one @dustinface . relic stores keys in locked memory and libsodium via the OS only gives you a finite amount of it. if we cache too many of these keys there will be segfaults due to memory allocation failures. or am i missing something?

PR #9790, opened 2022-01-12
Set reserved cores to 0

mariano54:
The sync speed benefits when switching to zero are significant, and many users are switching to it. This is especially clear in machines with not many cores like the RPI. It also seems to have a greater effect when combined with the other PR (bls optimization). 

Data can be seen here: 
https://github.com/neurosis69/chia-sync-data

The initial reason for having 2 here was that linux was lagging for some people during sync. However this is anectdotal evidence and might not be related to this change, as this was a long time ago there were many bugs present in past versions. 

Furthermore, as pointed out by @altendky operating systems should not freeze even when under full CPU load. Chia sync also currently does not take advantage of all CPUs even after this change. 

EDIT: wallet config has been removed, since it's no longer in use after the recent merge.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -183,7 +183,7 @@ class FullNode:
         self.coin_store = await CoinStore.create(self.db_wrapper)
         self.log.info("Initializing blockchain from disk")
         start_time = time.time()
-        reserved_cores = self.config.get("reserved_cores", 2)
+        reserved_cores = self.config.get("reserved_cores", 0)
         self.blockchain = await Blockchain.create(
             self.coin_store, self.block_store, self.constants, self.hint_store, self.db_path.parent, reserved_cores
         )
--- a/chia/util/initial-config.yaml
+++ b/chia/util/initial-config.yaml
@@ -342,7 +342,7 @@ full_node:
 
   # When creating process pools the process count will generally be the CPU count minus
   # this reserved core count.
-  reserved_cores: 2
+  reserved_cores: 0
 
   # How often to initiate outbound connections to other full nodes.
   peer_connect_interval: 30
--- a/chia/util/initial-config.yaml
+++ b/chia/util/initial-config.yaml
@@ -472,7 +472,7 @@ wallet:
 
   # When creating process pools the process count will generally be the CPU count minus
   # this reserved core count.
-  reserved_cores: 2
+  reserved_cores: 0
 
   logging: *logging
   network_overrides: *network_overrides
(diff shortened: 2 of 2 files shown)


hoffmang9 (dismissed):
AOK

altendky (changes requested):
There are defaults in the code reading this for when this is not present.  They should be updated as well.  Also to make sure it was considered, there's a separate setting for the wallet process pool.

The config should all be centralized and then we could write a test to check that the static default config file (with comments) matches the defaults for when the value is not specified.  But that is...  a whole other thing.

neurosis69:
> 
> RPI DUSTSTORM1 6:15 -> 5:08 
> RPI TRANSACTION_START: 0:40 -> 0:33
> 

The mentioned times here compare between my test scenarios AUTOTEST1 with AUTOTEST5.
Just please be aware, that for the test scenario AUTOTEST5 in addition to the cpu core change, I also droped all indexes from table coin_records. The missing indexes will be recreated automatically after the next node restart.

On RPI the drop of the Coin Record Indexes only (AUTOTEST3) reduced sync time from 6:15 -> 5:30.

mariano54:
> 

Thank you! I will fix that.

mariano54:
> > RPI DUSTSTORM1 6:15 -> 5:08
> > RPI TRANSACTION_START: 0:40 -> 0:33
> 
> The mentioned times here compare between my test scenarios AUTOTEST1 and AUTOTEST5. Just please be aware, that for the test scenario AUTOTEST5 in addition to the cpu core change, I also droped all indexes from table coin_records. The missing indexes will be recreated automatically after the next node restart.
> 
> On RPI the drop of the Coin Record Indexes only (AUTOTEST3) reduced sync time from 6:15 -> 5:30.

Thank you! I will fix that.

altendky (dismissed):
This lost the wallet reserved cores default in code at some point.  Either that should be added back or the change to the default config file should be removed.

mariano54:
> This lost the wallet reserved cores default in code at some point. Either that should be added back or the change to the default config file should be removed.

The wallet was completely changed with the new update so maybe it's not even relevant. ill check.

altendky (approved):
I think it might be a good idea to consider setting reduced priority for these processes.  I don't expect this will lock anyone out of their systems though.  I haven't had any significant issues that I remember on my Pi while running with four cores used.

mariano54:
Waiting on merge @wjblanke @emlowe

wjblanke:
I'll merge this since we are releasing in beta but we may need to swap the preference names if we have to override the 0 setting if an issue is found.

altendky:
@wjblanke i'm not sure what you mean by "swap the preference names"?  I understand your concern that 0 may cause some people trouble.

PR #9797, opened 2022-01-13
remove dead Seeder Guide link

dkackman:
This points to a non-existent document:
* [Chia seeder](https://github.com/Chia-Network/chia-blockchain/wiki/Chia-Seeder-User-Guide), which maintains a list of reliable nodes within the Chia network via a built-in DNS server.

(also did a search and couldn't find an alternative)


hoffmang9:
We actually should do the reverse and add some documents to the wiki for it to land on - but huge thanks for reminding us Don.

emlowe:
This link now points to a page - which we had prepared but forgot to create ...

PR #9798, opened 2022-01-13
Use get_type_hints()

altendky:
Draft for:
- [ ] This probably isn't a super efficient approach what with the function call.

### DIFF
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 import time
 import traceback
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, get_type_hints, List, Optional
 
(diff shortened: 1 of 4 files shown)


altendky:
It makes some things better anyways.

mariano54:
We switch from `get_type_hints` to `__annotations__` because it was super slow. Why not use `__dataclass_fields__`?

altendky:
Hence the reason for being draft.  `:]`

I think what we should do is do this once early and hold onto it in the form we want instead of spreading lots of low level api-change-prone accesses all over.  Just in a different way than the last attempt.  This was a quick check so I could see it help with the 3.10 effort and it seems to.  I'll follow up here, or in another PR, with something using `get_type_hints()` or `dataclasses.fields()` and caching the results more like other things are already being held for streamables.

hoffmang9:
Can we close this so that I can delete the branch?

PR #9802, opened 2022-01-13
Atari tree diff.

fchirica:
Given two tree roots, produces the differences as a list of
```
INSERT KEY VALUE
DELETE KEY VALUE
```

Depends on https://github.com/Chia-Network/chia-blockchain/pull/9748

### DIFF
--- a/chia/data_layer/data_store.py
+++ b/chia/data_layer/data_store.py
@@ -87,7 +87,8 @@ class DataStore:
                     hash TEXT NOT NULL,
                     ancestor TEXT,
                     tree_id TEXT NOT NULL,
-                    PRIMARY KEY(hash, tree_id)
+                    generation INTEGER NOT NULL,
+                    PRIMARY KEY(hash, tree_id, generation)
                 )
                 """
             )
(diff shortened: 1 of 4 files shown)


chia/data_layer/data_store.py
```diff
@@ -655,3 +761,53 @@ async def get_proof_of_inclusion_by_key(
...
+            return result
+
+    async def get_tree_diff(
+        self,
+        tree_id: bytes32,
+        root_1: Root,
+        root_2: Root,
+        *,
+        lock: bool = True,
+    ) -> List[DiffData]:
```
altendky, line 799: Since there's no meaning to the order and there shouldn't be duplicates, would a `set` be more appropriate?

tests/core/data_layer/test_data_store.py
```diff
@@ -232,6 +245,97 @@ async def test_get_ancestors(data_store: DataStore, tree_id: bytes32) -> None:
...
+                else:
+                    expected_diff.add(DiffData("delete", node.key, node.value))
+        if i == 200:
+            generation_start = await data_store.get_tree_generation(tree_id=tree_id)
+    root_start = await data_store.get_tree_root(tree_id, generation_start)
+    root_end = await data_store.get_tree_root(tree_id)
+    diffs = await data_store.get_tree_diff(tree_id, root_start, root_end)
+    assert len(diffs) == len(expected_diff)
+    for diff in diffs:
+        assert diff in expected_diff
```
altendky, line 337: This also looks like we mean for these to be sets.

chia/data_layer/data_store.py
```diff
@@ -655,3 +761,53 @@ async def get_proof_of_inclusion_by_key(
...
+                    node = await self.get_node(node_hash, lock=False)
+                    if isinstance(node, TerminalNode):
+                        result.append(node)
+                    else:
+                        queue.append(node.left_hash)
+                        queue.append(node.right_hash)
+
+            return result
+
+    async def get_tree_diff(
```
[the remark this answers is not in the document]
fchirica: yes, this is supposed to be proportional to `diff_size`, while doing `get_pairs` should be proportional to the tree size, as far as I understand. this is because when a tree matches in both old and new (by hash), we don't recurse over its subtree.

arvidn:
is this PR still relevant? It seems its dependency is stalled (#9748)

PR #9806, opened 2022-01-13
Exclude Files created by wheel.

jack60612:
Increases sanity.


altendky:
Could we exclude the entire build directory?

jack60612:
> Could we exclude the entire build directory?

Assumed you would want this but sure can do

jack60612:
> > Could we exclude the entire build directory?
> 
> Assumed you would want this but sure can do

Oh wait confused the directories, sorry about that.

altendky (approved):
It looks like we have a mess of different places we build stuff, but when I `venv/bin/python -m build --wheel .` I do get a `build/` directory and that should be ignored.  Thanks.

PR #9807, opened 2022-01-13
run block tool updates

emlowe:
Updates to Adam's run_block tool which is used by the retirement project explorer


arvidn:
I'm not very familiar with the details here, but this function should be exercised by CI, right?

mariano54:
> I'm not very familiar with the details here, but this function should be exercised by CI, right?

It's not used anywhere in the repo. It's only used for higher level applications. It's not clear to me that it belongs here. However I could see it as a useful thing for some people to see how to run a block. I would be interested in getting all transactions though, not just the CAT transactions.

tools/run_block.py
```diff
@@ -0,0 +1,187 @@
...
+                if condition[0] == ConditionOpcode.CREATE_COIN and len(condition) >= 4:
+                    # If only 3 elements (opcode + 2 args), there is no memo, this is ph, amount
+                    if type(condition[3]) != list:
+                        # If it's not a list, it's not the correct format
+                        continue
+
+                    # special retirement address
+                    if condition[3][0].hex() == "0000000000000000000000000000000000000000000000000000000000000000":
+                        if len(condition[3]) >= 2:
+                            memo = condition[3][1].decode("utf-8")
```
arvidn, line 122: it seems there's a risk this string might not be correct UTF-8. It might make sense to catch an exception here and maybe treat it as latin-1 or something like that, rather than aborting. This is, potentially, reading any untrusted data from the chain, right?

tools/run_block.py
```diff
@@ -0,0 +1,187 @@
...
+        constants.MAX_BLOCK_COST_CLVM,  # min(self.constants.MAX_BLOCK_COST_CLVM, block.transactions_info.cost),
+        cost_per_byte=constants.COST_PER_BYTE,
+        mempool_mode=False,
+    )
+    if npc_result.error is not None:
+        raise ConsensusError(Err(npc_result.error))
+
+    cat_list: List[CAT] = []
+    for npc in npc_result.npc_list:
+        _, puzzle, solution = get_puzzle_and_solution_for_coin(
```
arvidn, line 101: unrelated to this patch, but `get_puzzle_and_solution_for_coin()` seems to run the whole block program again, which isn't ideal. But I suppose that's mostly an implementation detail for now.

arvidn (dismissed):
I'll approve this, although I think at least the utf-8 decoding should be addressed

tools/run_block.py
```diff
@@ -0,0 +1,187 @@
...
+        "coin_name": npc.coin_name.hex(),
+        "conditions": [{"condition_type": c[0].name, "conditions": condition_list_to_dict(c)} for c in npc.conditions],
+        "puzzle_hash": npc.puzzle_hash.hex(),
+    }
+
+
+def run_generator(block_generator: BlockGenerator, constants: ConsensusConstants) -> List[CAT]:
+    npc_result = get_name_puzzle_conditions(
+        block_generator,
+        constants.MAX_BLOCK_COST_CLVM,  # min(self.constants.MAX_BLOCK_COST_CLVM, block.transactions_info.cost),
```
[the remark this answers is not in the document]
emlowe: Ok, i refactored the code a bit so I can do this

tools/run_block.py
emlowe: Added try/except here - good idead

tools/run_block.py
emlowe: Yes, I think this tool ends up running twice.

arvidn (approved):

PR #9765, opened 2022-01-09
Python 310

hoffmang9:
Attempting to get Python 3.10 working across the application.

The challenge is that 3.10 deprecated loop as a the loop parameter is deprecated from all APIs in 3.10.

We may have to go back to removing websockets as it is only used in daemon.

PR #9808, opened 2022-01-13
Update Dependencies for python 3.10. Part 1 of many

jack60612:
try to get python 3.10 ready.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ dependencies = [
     "setproctitle==1.2.2",  # Gives the chia processes readable names
     "sortedcontainers==2.4.0",  # For maintaining sorted mempools
     "websockets==8.1.0",  # For use in wallet RPC and electron UI
-    "click==8.0.3",  # For the CLI
+    "click==7.1.2",  # For the CLI
     "dnspythonchia==2.2.0",  # Query DNS seeds
     "watchdog==2.1.6",  # Filesystem event watching - watches keyring.yaml
     "dnslib==0.9.18",  # dns lib
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup
 
 dependencies = [
     "multidict==5.1.0",  # Avoid 5.2.0 due to Avast
-    "aiofiles==0.8.0",  # Async IO for files
+    "aiofiles==0.7.0",  # Async IO for files
     "blspy==1.0.9",  # Signature library
     "chiavdf==1.0.4",  # timelord and vdf verification
     "chiabip158==1.1",  # bip158-style wallet filters
--- a/setup.py
+++ b/setup.py
@@ -1,16 +1,16 @@
 from setuptools import setup
 
 dependencies = [
-    "multidict==5.1.0",  # Avoid 5.2.0 due to Avast
-    "aiofiles==0.7.0",  # Async IO for files
-    "blspy==1.0.8",  # Signature library
-    "chiavdf==1.0.3",  # timelord and vdf verification
+    "multidict==5.2.0",  # Avoid 5.2.0 due to Avast
+    "aiofiles==0.8.0",  # Async IO for files
+    "blspy==1.0.9",  # Signature library
+    "chiavdf==1.0.4",  # timelord and vdf verification
     "chiabip158==1.0",  # bip158-style wallet filters
-    "chiapos==1.0.7",  # proof of space
+    "chiapos==1.0.8",  # proof of space
     "clvm==0.9.7",
     "clvm_rs==0.1.17",
     "clvm_tools==0.4.3",
-    "aiohttp==3.7.4",  # HTTP server for full node rpc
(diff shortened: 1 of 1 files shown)


altendky (commented):
The required bits are updating in https://github.com/Chia-Network/chia-blockchain/pull/9765/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7.  At least click needs to be considered and tested first before the update to v8.

setup.py
```diff
@@ -1,38 +1,38 @@
 from setuptools import setup
 
 dependencies = [
-    "multidict==5.1.0",  # Avoid 5.2.0 due to Avast
-    "aiofiles==0.7.0",  # Async IO for files
-    "blspy==1.0.8",  # Signature library
-    "chiavdf==1.0.3",  # timelord and vdf verification
-    "chiabip158==1.0",  # bip158-style wallet filters
-    "chiapos==1.0.7",  # proof of space
+    "multidict==5.2.0",  # Avoid 5.2.0 due to Avast
```
hoffmang9, line 4: Note that we need to avoid 5.2.0, as the comment says, to not get an anti virus false positive.

hoffmang9 (changes requested):

setup.py
jack60612: Oops, I thought I removed that before pushing

setup.py
```diff
@@ -2,37 +2,37 @@
...
     "keyrings.cryptfile==1.3.4",  # Secure storage for keys on Linux (Will be replaced)
     #  "keyrings.cryptfile==1.3.8",  # Secure storage for keys on Linux (Will be replaced)
     #  See https://github.com/frispete/keyrings.cryptfile/issues/15
-    "PyYAML==5.4.1",  # Used for config file format
+    "PyYAML==6.0",  # Used for config file format
     "setproctitle==1.2.2",  # Gives the chia processes readable names
     "sortedcontainers==2.4.0",  # For maintaining sorted mempools
     "websockets==8.1.0",  # For use in wallet RPC and electron UI
-    "click==7.1.2",  # For the CLI
+    "click==8.0.3",  # For the CLI
```
altendky, line 29: I have run into issue with Click 8.  They may not be so applicable here but I would want to have thought through them first.  The one particular thing I ran into was that Windows expects programs to handle command line parsing including shell expansion where as in Linux (and macOS I believe) the shell is expected to do expansion and the process receives a 'list' of the arguments.  If I remember correctly, Click started respecting this difference which means it expands wildcards and processes quoting that it did not before.

```suggestion
    "click==7.1.2",  # For the CLI
```

altendky (dismissed):
Just formalizing the concern about Click.  I haven't looked over the other updates, including two other major version changes, to decide if we test them enough in CI to bank on the green results or any other considerations.

jack60612:
> Just formalizing the concern about Click. I haven't looked over the other updates, including two other major version changes, to decide if we test them enough in CI to bank on the green results or any other considerations.

Mostly made this to get us thinking but it is always great to update. ( if nothing breaks) 
I did go through the change logs on every one excluding click, oops and it seemed fine.

altendky:
Is dependabot not suggesting these?  The granularity of updating separately seems useful in case we need to bisect an issue.

altendky:
I would dismiss my change requested review but I don't see how to do that on a draft PR.  Looks like a GitHub UX hole.

jack60612:
> I would dismiss my change requested review but I don't see how to do that on a draft PR.  Looks like a GitHub UX hole.

Hehe now it is not a draft

jack60612:
> Is dependabot not suggesting these?  The granularity of updating separately seems useful in case we need to bisect an issue.

Dependabot is not but I went through the change logs of every dependency to insure that we are good.

hoffmang9 (dismissed):
This looks fine but note that under Python 3.10 the Daemon still does not start.
```
Daemon not started yet
Failed to create the chia daemon
```

hoffmang9:
Though I do like @altendky 's point about breaking this into 11 PRs for later debugging.

jack60612:
> Though I do like @altendky 's point about breaking this into 11 PRs for later debugging.

I will do that in a bit.

altendky:
Instead of breaking this up, should we just catch up on dependabot PRs until we get them all?  Or tell it to ignore versions we don't ever want to update to so they don't count against the aparrent 10 PR limit?

wjblanke:
Jack I'll take the chia owned updates to support python 3.10 but would like to avoid updating the others unless we have a good reason to do so. Can you remove those and then it can be merged.

jack60612:
> Jack I'll take the chia owned updates to support python 3.10 but would like to avoid updating the others unless we have a good reason to do so. Can you remove those and then it can be merged.

I guess its a start.

jack60612:
Forgot to remove aiofiles.

hoffmang9 (approved):
lgtm

Issue #9672, opened 2021-12-25
Wallet not synced stuck on 721957[Bug] 


feccie:
### What happened?

Can't sync a wallet although the blockchain itself is synced.
Have tried to delete both DBs - wallet and blockchain, and resynced again.
Tried deleting the wallet and importing it back.

I try both on 1.2.11 and 1.2.7 and both mode GUI and CLR are same results. Please help advice

### Version

1.2.11 and 1.2.7

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output

```shell
2021-12-26T00:22:01.341 wallet chia.wallet.wallet_node    : INFO     Requesting blocks 721955-721987
2021-12-26T00:22:01.838 wallet chia.wallet.wallet_state_manager: INFO     Adding coin: {'amount': 117,
 'parent_coin_info': '0xed17c426dacc4ea9a2c3b0c2f6329d6f868638a22760dfbba3300013288f91e5',
 'puzzle_hash': '0x9f1d2df27bc1784cc50bd4b2e257db130a32685af658c55267c208bf36afd1ec'} at 721957
2021-12-26T00:22:01.840 wallet chia.wallet.wallet_state_manager: INFO     Coins removed [Coin(parent_coin_info=<bytes32: 1a9c358095a4eb88e7b17cf76993a48858d2cac5610d8fc32298e55ed658b34b>, puzzle_hash=<bytes32: a58af351e4d45a265eb6fa4ac583dfc0a5e7fd9a4d7be8b74f085e07a767acad>, amount=118)] at height: 721957
2021-12-26T00:22:01.842 wallet chia.wallet.wallet_state_manager: INFO     Found created launcher. Creating pool wallet
2021-12-26T00:22:01.909 wallet chia.wallet.wallet_blockchain: ERROR    Error during db transaction: 'NoneType' object is not iterable
2021-12-26T00:22:01.913 full_node full_node_server        : INFO     Connection closed: 127.0.0.1, node id: 10c75f460903db542a91a097bc2f51052cba726729f7375c2404adf1eba5bd86
2021-12-26T00:22:01.913 full_node chia.full_node.full_node: INFO     peer disconnected {'host': '127.0.0.1', 'port': 8449}
2021-12-26T00:22:02.139 full_node chia.full_node.mempool_manager: INFO     It took 0.002931356430053711 to pre validate transaction
[...]

PR #9809, opened 2022-01-13
El.pool list fix

emlowe:
Sometimes people end up with `pool_list: null` in their config.yaml - this can happen when users manually delete items from the list to remove old NFTs, but they leave the `pool_list:` line. There may be other cases where this can happen as well.

In this case with `pool_list: null`, the wallet will stop syncing when it finds a plotnft because `load_pool_config` will throw `TypeError: 'NoneType' object is not iterable` which causes the entire sync to fail.

See the initial report in #9672


emlowe:
Messed up this PR - closing

PR #9827, opened 2022-01-15
Update pre-commit black to latest 21.12b0

altendky:
Draft for:
- [ ] Latest black isn't yet used by the GitHub super-linter.  Waiting for https://github.com/github/super-linter/commit/fc6c5b34d9cde631aff9bc918947b5e69ff146af to be released so we can update to that here.


altendky:
This is being dealt with elsewhere

PR #9835, opened 2022-01-16
Add ability to change payout address using cli.

jack60612:
Tested and working.

### DIFF
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -2,7 +2,6 @@ import asyncio
 import json
 import logging
 import time
-from dataclasses import replace
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
 import traceback
@@ -19,7 +18,7 @@ from chia.daemon.keychain_proxy import (
     connect_to_keychain_and_validate,
     wrap_local_keychain,
 )
-from chia.pools.pool_config import PoolWalletConfig, load_pool_config, update_pool_config
+from chia.pools.pool_config import PoolWalletConfig, load_pool_config
 from chia.protocols import farmer_protocol, harvester_protocol
 from chia.protocols.pool_protocol import (
     ErrorResponse,
@@ -406,7 +405,6 @@ class Farmer:
     async def update_pool_state(self):
         config = load_config(self._root_path, "config.yaml")
         pool_config_list: List[PoolWalletConfig] = load_pool_config(self._root_path)
-        new_pool_config_list: List[PoolWalletConfig] = []
         for pool_config in pool_config_list:
             p2_singleton_puzzle_hash = pool_config.p2_singleton_puzzle_hash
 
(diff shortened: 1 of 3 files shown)


Quexington (dismissed):
Looks like it works to me

Quexington (dismissed):
Still looks like it works to me :+1:

mariano54 (commented):
So this looks good for changing the config file. The question is, what will happen after that?

Let's say I'm farming on two machines A and B. I change the config on computer A. 
The expected behavior is that farmer A will contact the pool who will update it, which will tell B to update. Will this happen or will B complain and try to change it back?

jack60612:
> So this looks good for changing the config file. The question is, what will happen after that?
> 
> Let's say I'm farming on two machines A and B. I change the config on computer A. The expected behavior is that farmer A will contact the pool who will update it, which will tell B to update. Will this happen or will B complain and try to change it back?

They will both try to change it back.

mariano54:
> > So this looks good for changing the config file. The question is, what will happen after that?
> > Let's say I'm farming on two machines A and B. I change the config on computer A. The expected behavior is that farmer A will contact the pool who will update it, which will tell B to update. Will this happen or will B complain and try to change it back?
> 
> They will both try to change it back.

Maybe we should have some logic on the farmer side that if something changes on the server, we should also change it on our config file?

mariano54:
This is not necessarily part of this PR. 
However, if we don't solve this problem, we should at least warn the user when they change it,  that they should reconfigure this in all farmer machines that they have, so they don't have mismatched. (farmer not harvester)

jack60612:
> This is not necessarily part of this PR. However, if we don't solve this problem, we should at least warn the user when they change it, that they should reconfigure this in all farmer machines that they have, so they don't have mismatched. (farmer not harvester)

This has already been a pre existing issue, so i think i can get away with not having a warning till i fix the other issue with another pull.

mariano54:
Good point. However I still think a just adding an extra line to the print would be good to ensure that people are aware of this issue.

jack60612:
> Good point. However I still think a just adding an extra line to the print would be good to ensure that people are aware of this issue.

Already half way done adding it to this pull, give me a couple minutes.

jack60612:
Done

chia/farmer/farmer.py
```diff
@@ -496,12 +498,14 @@ async def update_pool_farmer_info() -> Tuple[Optional[GetFarmerResponse], Option
                                 farmer_info, farmer_is_known = await update_pool_farmer_info()
                                 if farmer_info is None and not farmer_is_known:
                                     self.log.error("Failed to update farmer info after POST /farmer.")
+                        if pool_config.payout_instructions.lower() != farmer_info.payout_instructions.lower():
```
mariano54, line 501: famer info might be None

chia/farmer/farmer.py
```diff
@@ -496,12 +498,14 @@ async def update_pool_farmer_info() -> Tuple[Optional[GetFarmerResponse], Option
...
+                            pool_config = replace(pool_config, payout_instructions=farmer_info.payout_instructions)
+                            just_updated = True
```
mariano54, line 503: If we change the config file, won't that trigger this to be just_updated = True?

mariano54 (approved):

xdustinface (approved):

PR #9836, opened 2022-01-16
Remove bintools from wallet api.

jack60612:


Quexington (dismissed):
Sure, I perhaps overthought this when I wrote it, we can always specify the byte hex.

mariano54:
Why are we converting to a Program @Quexington ? Shouldn't we just provide the bytes and force the client to do the conversion beforehand?

jack60612:
No it's useless, should I just removed it @mariano54

mariano54 (approved):
Would this mess up any callers of the RPC like GUI @Quexington ? If not, then we can merge

Quexington:
> Why are we converting to a Program @Quexington ? Shouldn't we just provide the bytes and force the client to do the conversion beforehand?
> 
> Edit: Is it even doing anything, it's parsing from bytes to program and then back to bytes?

I copied from other areas in the code although `SExp` probably also has a `__bytes__` I didn't think about it.

Quexington:
> Would this mess up any callers of the RPC like GUI @Quexington ? If not, then we can merge

I'm 90% positive that no part of our codebase calls this RPC including the GUI.  I'm even more confident that they're not doing it by manually specifying the announcements to assert.

mariano54:
Great let's merge it then.

jack60612:
> @jack60612 all commts must be signed. Can you reapply the commit, and then force push to this branch? `git push origin jack-kill-bintools --force`.

Sorry about that, my new laptop signed it wrong.

jack60612:
@mariano54 should be good now, github was just being a bit glitchy.

PR #9840, opened 2022-01-17
Pick TX type even when added coin has been spent

altendky:
While my wallet balance was correct but the reported farming rewards were low.  Several rewards were being categorized as regular transactions.  The code path for coins that had been spent assumed they were regular incoming transactions instead of considering whether they were rewards.  It doesn't seem great to have to duplicate this functionality between the two cases but this form seemed like the minimally intrusive change.  Given my limited experience here that seemed like a good way to share a fix.

This fix applies to syncing as far as I know so it will not correct incorrectly categorized transactions that were already processed into the db.

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -695,10 +695,15 @@ class WalletStateManager:
                 if record is None:
                     farmer_reward = False
                     pool_reward = False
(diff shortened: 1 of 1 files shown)


jack60612 (approved):
LGTM

Quexington (approved):
Yeah this makes sense to me

PR #9841, opened 2022-01-17
Add limit to number of plot NFT's displayed In CLI

jack60612:


mariano54:
What is the motivation for this change? I think the change should be that we shouldn't allow creation of more than 20, not display. Displaying as many as we have is fine.

jack60612:
> What is the motivation for this change? I think the change should be that we shouldn't allow creation of more than 20, not display. Displaying as many as we have is fine.

Oh, I was confused why you wanted to make me do this too, that makes more sense.

PR #9845, opened 2022-01-18
Atari test utility changes

Quexington:
These are two small fixes.  One to handle transactions without a spend bundle and one to handle an error that was ocurring on Windows.


altendky (dismissed):
Maybe `SYSTEMROOT` should be conditional on `sys.platform == "win32"`?

altendky (approved):

altendky:
Yes, I'm cheating on this one.  `:[`  It already built green and was just an unsigned commit.  The hazard is that it is now building for the first time against latest `atari` with the latest catch up merge from `main`.  If the resulting `atari` build fails I'll hold up any further merges until whatever issue is fixed and green on `atari`.

PR #9859, opened 2022-01-18
Use cache for validating signatures

mariano54:
We already validate the signature in unfinished block so there is no point in revalidating it. If we set `validated_signature` to True, then when we get the finished block, we will use this PrevalidationResult in `receive_block`, and will avoid revalidating it. 

Signature is always validated when we receive an unfinished block, so we can hardcode it to True here.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -1411,6 +1411,8 @@ class FullNode:
                 and unf_block.transactions_generator is not None
                 and unf_block.foliage_transaction_block == block.foliage_transaction_block
             ):
+                # We checked that the transaction block is the same, therefore all transactions and the signature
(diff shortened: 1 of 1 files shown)


arvidn (approved):
this looks right. it also looks like it was an oversight earlier, since (obviously) we should set `validated_signature` to true right after we validated it.

On line 1671, we just `return`.. is that the same as `return None`?

it also seems suspicious to return `None` on line 1682, *after* we have validated the signature. Do we not record that the signature was valid in that case?

mariano54:
Yes it was an oversight, but not in main, but in `ms.bls_multiproc` branch. (this pr is not to main)

yes `return` and `return None` are the same in Python.

Nothing is recorded until 1687 where we add the unfinished block to the memory cache.

PR #9864, opened 2022-01-19
Default connect_to_unknown_peers to True

altendky:


wjblanke (approved):
aok

jack60612 (approved):
looks good

chia/util/initial-config.yaml
```diff
@@ -460,6 +462,9 @@ wallet:
   initial_num_public_keys: 100
   initial_num_public_keys_new_wallet: 5
 
+  dns_servers:
```
mariano54, line 465: If we have an old config, the wallet will not start @altendky

altendky: Agreed.  That resulted in https://github.com/Chia-Network/chia-blockchain/pull/9880.  We still need to follow up to handle testnet as well.  Though I really don't like how 'testnet' is defined by a bunch of hardcoded values in the configure command code rather than having these things be defined in the config in a network-specific manner.

PR #9867, opened 2022-01-19
Restore chia wallet add_token CLI

altendky:


Quexington (approved):
Yeah makes sense :+1:

PR #9868, opened 2022-01-19
set BUILD_VDF_CLIENT=n as default on Darwin

cameroncooper:
BUILD_VDF_CLIENT does not have a default on Darwin which causes it to build and fail on M1.


hoffmang9:
If you are building the chiavdf wheel something else is wrong. Is this Python 3.10?

You should be getting prebuilt python binary wheels that already managed the ENV variable issue.

PR #9876, opened 2022-01-19
Atari catch up merge - main 2acf03e850fb883dc6ca5a554f5f827097157ac1

altendky:


chia/full_node/coin_store.py
```diff
@@ -85,6 +82,13 @@ async def create(cls, db_wrapper: DBWrapper, cache_size: uint32 = uint32(60000))
         self.coin_record_cache = LRUCache(cache_size)
         return self
 
+    async def num_unspent(self) -> int:
+        async with self.coin_record_db.execute("SELECT COUNT(*) FROM coin_record WHERE spent_index=0") as cursor:
```
Quexington, line 86: Not an expert in DB performance or if this even matters, but there is a `spent` field on the coin_record too that seems to be the natural tool for the job here.

chia/full_node/full_node.py
```diff
@@ -176,9 +176,7 @@ def sql_trace_callback(req: str):
 
         db_version: int = await lookup_db_version(self.connection)
 
-        self.db_wrapper = DBWrapper(
-            self.connection, self.config.get("allow_database_upgrades", False), db_version=db_version
-        )
+        self.db_wrapper = DBWrapper(self.connection, db_version=db_version)
```
Quexington, line 179: Is this a correct change?  I thought the "allow_database_upgrades" option was new and as such should probably be in here.

chia/util/db_wrapper.py
```diff
@@ -11,12 +11,10 @@ class DBWrapper:
 
     db: aiosqlite.Connection
     lock: asyncio.Lock
-    allow_upgrades: bool
     db_version: int
 
-    def __init__(self, connection: aiosqlite.Connection, allow_upgrades: bool = False, db_version: int = 1):
+    def __init__(self, connection: aiosqlite.Connection, db_version: int = 1):
```
Quexington, line 16: If my above comment was correct, this should also not be a change.

Quexington: There's a couple other instances spread throughout the PR

Quexington (approved):
A few comments, but none that seem critical to Atari functioning.

chia/full_node/coin_store.py
altendky: Introduced in https://github.com/Chia-Network/chia-blockchain/pull/9780 if you want to follow up on the concerns.

chia/full_node/full_node.py
altendky: Removed in https://github.com/Chia-Network/chia-blockchain/pull/9768.

chia/util/db_wrapper.py
altendky: https://github.com/Chia-Network/chia-blockchain/pull/9768

PR #9877, opened 2022-01-19
harvester: Drop plot existance check in `lookup_challenge`

xdustinface:
Saves a `Path.exists()` call in each lookup which obviously can be more 
expensive in some setups + this is anyway the job of 
the `PlotManager` and the existance is already implied due to the plot 
being in `PlotManager.plots`.

### DIFF
--- a/chia/harvester/harvester_api.py
+++ b/chia/harvester/harvester_api.py
@@ -172,21 +172,17 @@ class HarvesterAPI:
         total = 0
         with self.harvester.plot_manager:
             for try_plot_filename, try_plot_info in self.harvester.plot_manager.plots.items():
-                try:
-                    if try_plot_filename.exists():
-                        # Passes the plot filter (does not check sp filter yet though, since we have not reached sp)
-                        # This is being executed at the beginning of the slot
-                        total += 1
-                        if ProofOfSpace.passes_plot_filter(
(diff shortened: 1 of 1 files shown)


arvidn (commented):
I agree that it seems unnecessary to check for existence first, just to immediately open the file to read from it. The file may have been removed in between those two checks anyway. However, I don't think you can ignore the fact that the file may have been removed or have had permissions changed to prevent it from being opened. It seems incorrect to remove the `try`-`catch` around looking up the proof of space.

If a file is missing or fail for some reason, we should still attempt to look up proofs of the remaining files in the list, right?

xdustinface:
The code in question just populates a list of coroutines for the lookups here https://github.com/Chia-Network/chia-blockchain/blob/c6a46751b8e17fc5339360553f7330b4d46a3aad/chia/harvester/harvester_api.py#L185 The actual lookup is happening here https://github.com/Chia-Network/chia-blockchain/blob/c6a46751b8e17fc5339360553f7330b4d46a3aad/chia/harvester/harvester_api.py#L72-L146 and is still surrounded by a `try`/`catch`. Is there anything else why you think it should still be a `try`/`catch`?

mariano54 (approved):

PR #9880, opened 2022-01-19
Set default wallet: dns_servers in code

altendky:


jack60612 (approved):
Lgtm

fchirica (changes requested):
we should check we are running mainnet before defaulting to that DNS introducer, otherwise testnet wallets will get mainnet nodes.

emlowe:
Is the problem that on upgrade people won't have this wallet: dns_server setting at all in config.yaml ?

altendky:
Yes to the upgrade path bit.  Florin's comment makes sense so I'll take a look.  Though the concern seems just as applicable to the default config as to this.  (in other words, all the more important to get it fixed, thanks)

altendky:
@fchirica, what do you think of merging this PR in so that the wrong setup (not accounting for testnet) is properly complete, and reworking the whole config design for this separately?  This is completing a missing piece that belonged in https://github.com/Chia-Network/chia-blockchain/pull/9864.

altendky:
I guess the issue is that presently people that run main with an old config will get a `KeyError`.  Adding this lets them run mainnet at least.  I have looked a bit at fixing this properly and it seems a bit confused between the presence of the per-network configuration in the config file and then the `chia configure` command having a bunch of hardcoded values for switching to testnet.  I feel like there's going to be some actual design and maybe rework to make this be handled separately and having it not fail for mainnet might be a useful intermediate step.

jack60612:
> we should check we are running mainnet before defaulting to that DNS introducer, otherwise testnet wallets will get mainnet nodes.

i feel like this is the best solution for now until we can do some config migration/ upgrade thing. Maybe tie it in with the v2 db?

altendky:
This change is only relevant in the case where you otherwise get an exception.  Either the indexing would have succeeded and this does nothing, or it would have failed and this gives you mainnet introducers.  I think getting mainnet over an exception is good for people running mainnet.  It's worse for those running testnet and that's why I agree that the greater issue should be explored and addressed.

jack60612:
> This change is only relevant in the case where you otherwise get an exception. Either the indexing would have succeeded and this does nothing, or it would have failed and this gives you mainnet introducers. I think getting mainnet over an exception is good for people running mainnet. It's worse for those running testnet and that's why I agree that the greater issue should be explored and addressed.

could you check the config file to check if we are on mainnet or not, or would that be too heavy?

jack60612 (changes requested):
fair point, we need to distinguish between testnet and mainnet before providing a dns server.

altendky:
We have what strikes me as a weird situation where the config file has sections for mainnet and multiple testnets...  but instead of configuring to use testnet just setting a value to refer to one of those it instead overwrites various values throughout the config file.  It seems to me to be an odd hybrid situation with yet more places that configuration values get put.  Instead of just the default config file and everywhere that value gets used in code, you also have to think about whether it goes in the configure command.

I guess this is specifically for the wallet so it should probably by default have the same value as the full node.  I guess the full node just leaves the list empty if the port number is 8444.  not a beautiful mainnet check especially given the discussions about making sure we're entirely agnostic about port number.  Does an empty list result in testnet behavior?

https://github.com/Chia-Network/chia-blockchain/blob/ee9c7968a8ab4aadc3033d536b7e0e110119355f/chia/full_node/full_node.py#L294-L305

Maybe I'll get inspired to dive into this soon but I'm not promising to.

PR #9882, opened 2022-01-20
fix mempool issue with ASSERT_SECONDS_RELATIVE

arvidn:
Currently, the mempool rejects a spend bundle that spends an ephemeral coin with an ASSERT_SECONDS_RELATIVE 0 condition. Such condition is a no-op in the main blockchain validation (the consensus rules), so the mempool should also allow it.

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -343,14 +343,19 @@ class MempoolManager:
                 return None, MempoolInclusionStatus.FAILED, Err.UNKNOWN_UNSPENT
             elif name in additions_dict:
                 removal_coin = additions_dict[name]
-                # TODO(straya): what timestamp to use here?
+                # The timestamp and block-height of this coin being spent needs
+                # to be consistent with what we use to check time-lock
+                # conditions (below). All spends (including ephemeral coins) are
+                # spent simultaneously. Ephemeral coins with an
+                # ASSERT_SECONDS_RELATIVE 0 condition are still OK to spend in
+                # the same block.
                 assert self.peak.timestamp is not None
                 removal_record = CoinRecord(
                     removal_coin,
                     uint32(self.peak.height + 1),  # In mempool, so will be included in next height
                     uint32(0),
                     False,
-                    uint64(self.peak.timestamp + 1),
+                    self.peak.timestamp,
                 )
 
             assert removal_record is not None
@@ -361,7 +366,6 @@ class MempoolManager:
         removals: List[Coin] = [coin for coin in removal_coin_dict.values()]
 
         if addition_amount > removal_amount:
-            print(addition_amount, removal_amount)
             return None, MempoolInclusionStatus.FAILED, Err.MINTING_COIN
 
         fees = uint64(removal_amount - addition_amount)
--- a/tests/core/full_node/test_mempool.py
+++ b/tests/core/full_node/test_mempool.py
@@ -77,6 +77,14 @@ def event_loop():
     yield loop
 
 
+# TODO: this fixture should really be at function scope, to make all tests
+# independent. Right now, TestMempool::test_basic_mempool initializes these
+# nodes, and the remaining tests just build on top of them. This means
+# test_basic_mempool must alwasy be run in order to have most of the other tests
+# pass.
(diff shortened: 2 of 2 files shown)


tests/core/full_node/test_mempool.py
```diff
@@ -246,6 +254,63 @@ async def test_basic_mempool_manager(self, two_nodes):
...
+
+            conditions = {opcode: [ConditionWithArgs(opcode, [int_to_bytes(lock_value)])]}
+            tx1 = WALLET_A.generate_signed_transaction(
+                uint64(1000000), WALLET_A.get_new_puzzlehash(), coin_2, conditions.copy(), uint64(0)
+            )
+
+            ephemeral_coin: Coin = tx1.additions()[0]
+            tx2 = WALLET_A.generate_signed_transaction(
+                uint64(1000000), WALLET_A.get_new_puzzlehash(), ephemeral_coin, conditions.copy(), uint64(0)
+            )
```
arvidn, line 294: this test spends `coin_2`, giving it the condition being tested, then spends it again (`ephemeral_coin`) within the same spend bundle (i.e. it's ephemeral).

tests/core/full_node/test_mempool.py
```diff
@@ -246,6 +254,63 @@ async def test_basic_mempool_manager(self, two_nodes):
...
+            # (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 4, MempoolInclusionStatus.PENDING),
+            # (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 5, MempoolInclusionStatus.PENDING),
+            # genesis timestamp is 10000 and each block is 10 seconds
+            # (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, 10029, MempoolInclusionStatus.SUCCESS),
+            # (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, 10030, MempoolInclusionStatus.SUCCESS),
+            # (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, 10031, MempoolInclusionStatus.FAILED),
+            # (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, 10032, MempoolInclusionStatus.FAILED),
+        ],
+    )
+    async def test_ephemeral_timelock(self, two_nodes, opcode, lock_value, expected):
```
arvidn, line 283: this can be compared with the test for the block validation, here: https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L1701

chia/full_node/mempool_manager.py
```diff
@@ -343,14 +343,19 @@ async def add_spendbundle(
...
+                # the same block.
                 assert self.peak.timestamp is not None
                 removal_record = CoinRecord(
                     removal_coin,
-                    uint32(self.peak.height + 1),  # In mempool, so will be included in next height
+                    uint32(self.peak.height + 1),
                     uint32(0),
                     False,
-                    uint64(self.peak.timestamp + 1),
+                    uint64(self.peak.timestamp),
```
arvidn, line 358: this is the fix

mariano54 (commented):
This was done this way because each transaction block must increment the timestamp by at least 1 second. Why don't we instead change the timestamp in the call to `mempool_check_conditions_dict`  to use `peak_height + 1`?

mariano54 (dismissed):
LGTM if tests pass

arvidn:
some more investigation is necessary

arvidn:
@mariano54 the effect of the next block needing to be at least one second after the previous one is already ensured, without making the change you suggested, by the actual comparison in `mempool_assert_relative_time_exceeds()`. It's checking:

```
if timestamp < expected_seconds + unspent.timestamp:
   return Err.ASSERT_SECONDS_RELATIVE_FAILED
```

which means that timestamp of the peak block has to be *strictly* less than, the coin timestamp + the condition's parameter.

So, the change I made in response to your comment, to keep the +1 when creating ephemeral coins, but instead add +1 to the peak timestamp makes it work right for ephemeral coins, but not for coins spent in the previous block.
I believe my original change is correct.

mariano54 (approved):

Issue #9890, opened 2022-01-20
[Bug] 1.3 beta release from main wallet never syncs to node


paulhainsworth-chia:
### What happened?

[debug.log](https://github.com/Chia-Network/chia-blockchain/files/7908745/debug.log)

Wallet never syncs. 

### Version

1.2.12.dev143

### What platform are you using?

macOS

### What ui mode are you using?

GUI

PR #9893, opened 2022-01-20
Do not assert if pool_state is none

emlowe:
Possible fix for issues PaulH is seeing in main release #9890


mariano54:
Hmm this is sketchy. Why would pool state be None? We might be ignoring a valid transaction for the pool wallet

emlowe:
Was a quick test based on some feedback from Straya - it's quite possibly it's not the proper fix.

Quexington:
I had to make this change in the atari branch to support launchers for non-pool singletons but I did it a bit differently although I think semantically these end up being the same thing. https://github.com/Chia-Network/chia-blockchain/blob/c2768d386fbee4ea927604949e1abac6814eac0c/chia/wallet/wallet_state_manager.py#L864

Quexington:
> Hmm this is sketchy. Why would pool state be None? We might be ignoring a valid transaction for the pool wallet

The function will return none if there is no hint to a new state such as during an absorb.  This assertion is likely just for mypy.

mariano54:
> > Hmm this is sketchy. Why would pool state be None? We might be ignoring a valid transaction for the pool wallet
> 
> The function will return none if there is no hint to a new state such as during an absorb. This assertion is likely just for mypy.

I see, thank you for explaining. We should at least have a comment when checking for None that says that it's possible for it to be None in an absorb scenario.

mariano54 (approved):

wjblanke:
Paul did this fix your issue?

PR #9896, opened 2022-01-21
Test installers in Dockers

altendky:
This introduces testing across the Linux distributions that our `install.sh` explicitly covers with the exception of CentOS and Redhat.  I don't know if we have an explicitly supported version list for each but I picked a few that seemed to make sense.  This also reverts the addition of `=3.9*` on the `python3-venv` install for Debian added in https://github.com/Chia-Network/chia-blockchain/pull/9633/files#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fR85 since it does not work. When `python3-venv` on Debian is 3.10 you will have to `python3.9-venv` to get 3.9.  `python3-venv=3.9*` will do nothing when `python3-venv` is 3.9 and it will fail when it is anything else, such as in Debian Buster.  Additionally, a `-a` flag was added to `install.sh` to allow for automated installs.  For now, this is only used with `pacman` to pass `--noconfirm`.

Draft for:
- [x] Self reflection

### DIFF
--- a/.github/workflows/test-install-scripts.yml
+++ b/.github/workflows/test-install-scripts.yml
@@ -50,43 +50,53 @@ jobs:
         sh install-gui.sh
 
   test_scripts_in_docker:
-    name: Test Install Scripts
+    name: Test Install Scripts ${{ matrix.distribution.name }}
     runs-on: ${{ matrix.os }}
-    container: ${{ matrix.docker.url }}
+    container: ${{ matrix.distribution.url }}
     strategy:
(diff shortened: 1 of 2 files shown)


hoffmang9:
😎

wallentx (commented):
Nice!
:shipit:

hoffmang9 (approved):
AOK

cmmarslender (approved):

PR #9906, opened 2022-01-21
fix parameter name to blockchain_interface.persist_sub_epoch_challenge_segments

arvidn:


altendky:
`chia.util.block_cache.BlockCache` also needs updated.

https://github.com/Chia-Network/chia-blockchain/blob/567061919fcd5a4dbbe4f1bd34e55185f18b58b7/chia/util/block_cache.py#L85-L88

I will note additionally that while `chia.wallet.wallet_blockchain.WalletBlockchain` inherits from the `BlockchainInterface` it does not implement `.persist_sub_epoch_challenge_segments()` at all.  Though this is probably for yet another PR.

https://github.com/Chia-Network/chia-blockchain/blob/567061919fcd5a4dbbe4f1bd34e55185f18b58b7/chia/wallet/wallet_blockchain.py#L21

Let alone a PR to make it so that this interface is actually properly checked and probably convert it to a `typing.Protocol` to drop the inheritance.  As a note, I sometimes like to explicitly and directly check that something implements a protocol rather than just waiting until it is passed somewhere requiring the protocol.  I made https://github.com/altendky/qtrio/blob/43c7ff24c0be2f7a3df86eef3c6cc5dae2f7ffd3/qtrio/_util.py#L7 to do that.

arvidn:
@altendky making `BlockchainInterface` a protocol definitely sounds like the right thing to do. Do you think we should abandon this change?

altendky:
You can see a start at that over in https://github.com/Chia-Network/chia-blockchain/pull/9979.  It is pretty annoying that mypy doesn't always give a clear and complete explanation of exactly _how_ a given object doesn't satisfy a `Protocol`.  But, at least it checks it.

Should we abandon this PR here?  Maybe.  Mostly I just want the four relevant classes to agree.  At least in terms of this particular method.

PR #9907, opened 2022-01-22
move member function get_memos() out of SpendBundle

arvidn:
and into a free-function living in the wallet code. It does not belong to the full node. Also rename it to indicate that it's expensive (and dangerous) to call.

This function can only be run on trusted clvm code.


mariano54 (approved):
100% support this

jack60612 (approved):
Does not belong so i'm happy your moving it.

PR #9911, opened 2022-01-22
Coin Selection Refactor 

jack60612:
New algorithms to limit unnecessary spends and reduce fees.
Requires more thorough testing. 

- [x] Finalize exact match algorithm
- [x] Finalize smaller coin algorithm 
- [x] Finalize knapsack coin algorithm 
- [x] Test Algorithms  
- [x] Fix CI
- [x] Write tests for all 3 algorithms 
- [x] Add Unit tests


chia/wallet/wallet.py
```diff
@@ -276,16 +276,70 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
         unconfirmed_removals: Dict[bytes32, Coin] = await self.wallet_state_manager.unconfirmed_removals_for_wallet(
             self.id()
         )
+        # check for exact 1 to 1 coin match.
```
mariano54, line 279: This should be decomposed into another file and method, for easier code reading in the future.
For example, `coin_selection.py`, `find_exact_coin(coins .....`

chia/wallet/wallet.py
```diff
@@ -276,16 +276,70 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+                self.log.debug(
+                    f"Selected coin: {coinrecord.coin.name()} at height {coinrecord.confirmed_block_height}!"
+                )
+                break
+
+        # Check for exact match with all coins smaller than the amount.
+        # If the total of all the smaller coins are less than the amount we choose; the smallest coin available.
+        # If we have more, smaller coins than the amount we run the next algorithm.
+        if len(used_coins) == 0:
+            smaller_coins: Set = set()
```
mariano54, line 297: This section should also be its own function

chia/wallet/wallet.py
```diff
@@ -276,16 +276,70 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+            if coinrecord.coin.amount == amount:
+                sum_value += coinrecord.coin.amount
+                used_coins.add(coinrecord.coin)
+                self.log.debug(
+                    f"Selected coin: {coinrecord.coin.name()} at height {coinrecord.confirmed_block_height}!"
+                )
+                break
+
+        # Check for exact match with all coins smaller than the amount.
+        # If the total of all the smaller coins are less than the amount we choose; the smallest coin available.
```
mariano54, line 294: Fix grmmar here, remove semicolon. Choose the smallest coin available that is still greater than the amount correct? What if there are no coins big enough, do we fail here?

chia/wallet/wallet.py
```diff
@@ -276,16 +276,70 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+                    sum_value = smallest_coin.coin.amount
+                    self.log.debug(
+                        f"Selected coin: {smallest_coin.coin.name()} at height {smallest_coin.confirmed_block_height}!"
+                    )
+            elif sum_value == amount:
+                used_coins = smaller_coins
+
+        if len(used_coins) == 0:
+            sum_value = 0
+            # check for all other combinations of coins.
```
mariano54, line 330: It's not checking all combinations, it's adding one at a time until they are sufficient right? In what order are the coins sorted in? It's worth a comment because that is very important here. 

Is this the same as bitcoins algorithm?

Also probably should decompose this out into it's own little module.

mariano54 (commented):
Finally, when it's decomposed, then we can easily write tests for it to make sure it's 100% solid

chia/wallet/wallet.py
jack60612: @mariano54 that uses the same code as before.

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,53 @@
+import random
+from typing import Set, Optional, Tuple, List
+from chia.types.blockchain_format.coin import Coin
+from chia.util.ints import uint64
+from chia.wallet.wallet_coin_record import WalletCoinRecord
+
+
+# we use this to check if one of the coins exactly matches the target.
+def check_for_exact_match(coin_records: List[WalletCoinRecord], target: int) -> Optional[WalletCoinRecord]:
```
mariano54, line 9: target should be uint64

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,53 @@
+import random
+from typing import Set, Optional, Tuple, List
+from chia.types.blockchain_format.coin import Coin
+from chia.util.ints import uint64
+from chia.wallet.wallet_coin_record import WalletCoinRecord
+
+
+# we use this to check if one of the coins exactly matches the target.
+def check_for_exact_match(coin_records: List[WalletCoinRecord], target: int) -> Optional[WalletCoinRecord]:
+    for coinrecord in coin_records:
```
mariano54, line 10: use `camel_case` naming so `coin_record`, the same everywhere. For classes we can use ThisKindOfName

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,53 @@
...
+def check_for_exact_match(coin_records: List[WalletCoinRecord], target: int) -> Optional[WalletCoinRecord]:
+    for coinrecord in coin_records:
+        if coinrecord.coin.amount == target:
+            return coinrecord
+    return None
+
+
+# we use this to find the individual coin that is closest to the target amount.
+def find_smallest_coin(coin_records: List[WalletCoinRecord], target: int) -> Optional[WalletCoinRecord]:
+    smallest_value = float("inf")  # smallest coins value
```
mariano54, line 18: we can use constants.MAX_COIN_AMOUNT

chia/wallet/coin_selection.py
mariano54: same in the other functions

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,53 @@
...
+
+# we use this to check if one of the coins exactly matches the target.
+def check_for_exact_match(coin_records: List[WalletCoinRecord], target: int) -> Optional[WalletCoinRecord]:
+    for coinrecord in coin_records:
+        if coinrecord.coin.amount == target:
+            return coinrecord
+    return None
+
+
+# we use this to find the individual coin that is closest to the target amount.
```
mariano54, line 16: comment should also say: that is still greater than the target

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,53 @@
...
+    for coinrecord in coin_records:
+        if target < coinrecord.coin.amount < smallest_value:
+            # try to find a coin that is as close as possible to the amount.
+            smallest_value = coinrecord.coin.amount
+            smallest_coin = coinrecord
+    return smallest_coin
+
+
+# we use this to find the smallest set of coins.
+def knapsack_coin_algorithm(smaller_coins: Set[Coin], target: int) -> Tuple[Optional[Set[Coin]], uint64]:
```
mariano54, line 29: Why do we use Coin here but WalletCoinRecord above? It's preferable for it to be consistent.

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,53 @@
...
+    smallest_coin: Optional[WalletCoinRecord] = None
+    for coinrecord in coin_records:
+        if target < coinrecord.coin.amount < smallest_value:
+            # try to find a coin that is as close as possible to the amount.
+            smallest_value = coinrecord.coin.amount
+            smallest_coin = coinrecord
+    return smallest_coin
+
+
+# we use this to find the smallest set of coins.
```
mariano54, line 28: to find the set of coins which have total value closest to the target, but at least the target.

mariano54: We should also add a link to the algorithm paper or website.

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,53 @@
...
+    best_set_of_coins: Optional[Set[Coin]] = None
+    for i in range(1000):
+        # reset these variables every loop.
+        selected_coins: Set = set()
+        selected_coins_sum = 0
+        n_pass = 0
+        target_reached = False
+        while n_pass < 2 and not target_reached:
+            for coin in smaller_coins:
+                if (n_pass == 0 and bool(random.getrandbits(1))) or (coin not in selected_coins):
```
mariano54, line 40: Small comment explaining what pass 0 and pass 1 mean

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,53 @@
...
+    best_set_sum = float("inf")
+    best_set_of_coins: Optional[Set[Coin]] = None
+    for i in range(1000):
+        # reset these variables every loop.
+        selected_coins: Set = set()
+        selected_coins_sum = 0
+        n_pass = 0
+        target_reached = False
+        while n_pass < 2 and not target_reached:
+            for coin in smaller_coins:
```
mariano54, line 39: You cannot iterate through a set, you will have to do `.items()` I think

chia/wallet/coin_selection.py
mariano54: Also, the input coins have to be sorted by descending value according to the paper, we should point this out in the comment. We should either check that the ordering is correct and throw if it's not, OR sort the smaller coins after receiving them.

chia/wallet/wallet.py
```diff
@@ -294,6 +291,41 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
                 "Can't make this transaction at the moment. Waiting for the change from the previous transaction."
             )
 
+        # Use older coins first
+        valid_unspent.sort(reverse=True, key=lambda r: r.coin.amount)
```
mariano54, line 295: This is the wrong ordering for the Bitcoin knapsack algo

chia/wallet/wallet.py
```diff
@@ -265,27 +266,23 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
         # Try to use coins from the store, if there isn't enough of "unused"
         # coins use change coins that are not confirmed yet
         unconfirmed_removals: Dict[bytes32, Coin] = await self.wallet_state_manager.unconfirmed_removals_for_wallet(
             self.id()
         )
-        for coinrecord in unspent:
-            if sum_value >= amount and len(used_coins) > 0:
-                break
+
+        sum_value = 0
```
mariano54, line 275: Sum value is not used

chia/wallet/wallet.py
```diff
@@ -265,27 +266,23 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+        used_coins: Set = set()
+        valid_unspent: List = []
```
mariano54, line 277: List of what? Type annotation like `List[WalletCoinRecord]`

chia/wallet/wallet.py
```diff
@@ -294,6 +291,41 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+        if exact_match_coinrecord:
+            used_coins.add(exact_match_coinrecord.coin)
+            self.log.debug(
+                f"Selected coin: {exact_match_coinrecord.coin.name()} "
+                f"at height {exact_match_coinrecord.confirmed_block_height}!"
+            )
+
+        # Check for an exact match with all of the coins smaller than the amount.
+        # If we have more, smaller coins than the amount we run the next algorithm.
+        if not exact_match_coinrecord:
```
mariano54, line 308: else

chia/wallet/wallet.py
```diff
@@ -294,6 +291,41 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+        valid_unspent.sort(reverse=True, key=lambda r: r.coin.amount)
+
+        # check for exact 1 to 1 coin match.
+        exact_match_coinrecord: Optional[WalletCoinRecord] = check_for_exact_match(valid_unspent, amount)
+        if exact_match_coinrecord:
+            used_coins.add(exact_match_coinrecord.coin)
+            self.log.debug(
+                f"Selected coin: {exact_match_coinrecord.coin.name()} "
+                f"at height {exact_match_coinrecord.confirmed_block_height}!"
+            )
```
mariano54, line 304: Actually here we can just return right?

chia/wallet/wallet.py
```diff
@@ -294,6 +291,41 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+                used_coins = smaller_coins
+            elif smaller_coin_sum < amount:
+                smallest_coin: Optional[WalletCoinRecord] = find_smallest_coin(valid_unspent, amount)
+                if smallest_coin:
+                    used_coins.add(smallest_coin.coin)
+                    self.log.debug(
+                        f"Selected coin: {smallest_coin.coin.name()} at height {smallest_coin.confirmed_block_height}!"
+                    )
+            elif smaller_coin_sum > amount:
+                best_coin_set, sum_best_coin_set = knapsack_coin_algorithm(smaller_coins, amount)
```
mariano54, line 325: I think this algorithm is guaranteed to succeed because pass 2 ends up adding all the remaining coins which add up to > amount

mariano54 (changes requested):

chia/wallet/coin_selection.py
jack60612: You can if i am not mistaken. https://www.geeksforgeeks.org/iterate-over-a-set-in-python/

chia/wallet/wallet.py
jack60612: Changed it and it is now used to check if we have enough coins.

chia/wallet/coin_selection.py
jack60612: Fixed

chia/wallet/wallet.py
jack60612: Descending order is the correct order if i am not mistaken.

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,62 @@
...
+# we use this to check if one of the coins exactly matches the target.
+def check_for_exact_match(coin_list: List[Coin], target: uint64) -> Optional[Coin]:
+    for coin in coin_list:
+        if coin.amount == target:
+            return coin
+    return None
+
+
+# we use this to find an individual coin greater than the target but as close as possible to the target.
+def find_smallest_coin(greater_coin_list: List[Coin], target: uint64) -> Optional[Coin]:
```
mariano54, line 21: The coins passed in to the function are not greater, they might be less also right? So they should be named `coin_list`

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,62 @@
...
+    smallest_value = DEFAULT_CONSTANTS.MAX_COIN_AMOUNT  # smallest coins value
```
mariano54, line 22: Usually using DEFAULT_CONSTANTS directly in the code is not great because if you want to run a testnet with modified constants, for example, you want the whole code to use your new constants. The better way is passing them in (the wallet.wallet_state_manager.constants.). However it's not a big deal here since this constant will never change in forks or testnets

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,62 @@
...
+    for i in range(1000):
+        # reset these variables every loop.
+        selected_coins: Set = set()
+        selected_coins_sum = 0
+        n_pass = 0
+        target_reached = False
+        while n_pass < 2 and not target_reached:
+            for coin in smaller_coins_sorted:
+                # run 2 passes where the first pass selects coins a coin 50 percent of the time.
+                # the second pass runs only if the coin is not selected in the first pass.
```
mariano54, line 47: Not quite: The second pass only runs if pass fails to find a selection. 
Pass 2 basically adds the remaining coins (those that were unselected) one at a time until we get greater than target,

mariano54: Pass 1 and pass 2 are one trial together. The purpose is not to select different coins in pass 1 and pass 2. The purpose is that each trial is random because pass 1 is random. The purpose of pass 2 is to finish the set in the cases that pass 1 did not get enough value.

chia/wallet/wallet.py
```diff
@@ -265,35 +266,61 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+        if smaller_coin_sum == amount:
+            used_coins = smaller_coins
+        elif smaller_coin_sum < amount:
+            smallest_coin: Optional[Coin] = find_smallest_coin(valid_unspent, amount)
+            if smallest_coin:
+                used_coins.add(smallest_coin)
+                self.log.debug(f"Selected coin: {smallest_coin.name()}")
+        elif smaller_coin_sum > amount:
+            best_coin_set, sum_best_coin_set = knapsack_coin_algorithm(smaller_coins, amount)
+            if best_coin_set:
```
mariano54, line 321: kanpsack_coin_algorithm always returns a value, it cannot return None

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,62 @@
...
+        if target < coin.amount < smallest_value:
+            # try to find a coin that is as close as possible to the amount.
+            smallest_value = coin.amount
+            smallest_coin = coin
+    return smallest_coin
+
+
+# we use this to find the set of coins which have total value closest to the target, but at least the target.
+# coins should be sorted in descending order.
+def knapsack_coin_algorithm(smaller_coins: Set[Coin], target: uint64) -> Tuple[Optional[Set[Coin]], uint64]:
```
mariano54, line 34: I think it's better to return `Optional[Tuple[Set[Coin], uint64]]`, That is, the uint64 in the optional

chia/wallet/wallet.py
mariano54: The reason is that `smaller_coin_sum > amount` therefore pass2 will always succeed

chia/wallet/wallet.py
```diff
@@ -265,35 +266,61 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+        # Check for an exact match with all of the coins smaller than the amount.
+        # If we have more, smaller coins than the amount we run the next algorithm.
+        smaller_coin_sum = 0  # coins smaller than target.
+        smaller_coins: Set[Coin] = set()
+        for coin in valid_unspent:
+            if coin.amount < amount:
+                smaller_coin_sum += coin.amount
+                smaller_coins.add(coin)
+        if smaller_coin_sum == amount:
+            used_coins = smaller_coins
```
mariano54, line 313: `return smaller_coins`

chia/wallet/wallet.py
```diff
@@ -265,35 +266,61 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+        elif smaller_coin_sum < amount:
+            smallest_coin: Optional[Coin] = find_smallest_coin(valid_unspent, amount)
+            if smallest_coin:
+                used_coins.add(smallest_coin)
```
mariano54, line 317: `return set([smallest_coin])`

chia/wallet/wallet.py
mariano54: `assert best_coin_set is not None`
`return best_coin_set`

chia/wallet/wallet.py
```diff
@@ -265,35 +266,61 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+                self.log.debug(f"Selected coin: {smallest_coin.name()}")
+        elif smaller_coin_sum > amount:
```
mariano54, line 319: This can be an else instead of elif. There are no other remaining cases, the only cases are `=`. `<`, and `>`

chia/wallet/wallet.py
```diff
@@ -265,35 +266,61 @@ async def select_coins(self, amount, exclude: List[Coin] = None) -> Set[Coin]:
...
+            best_coin_set, sum_best_coin_set = knapsack_coin_algorithm(smaller_coins, amount)
+            if best_coin_set:
+                used_coins = best_coin_set
+
```
mariano54, line 323: No need to return here because the Else should catch everything

mariano54 (commented):
Much better

chia/wallet/wallet.py
jack60612: Do we not want to use the debug return at the end?

chia/wallet/coin_selection.py
jack60612: Ok i think i got it this time.

mariano54 (commented):
LGTM, now we need tests for each function, as well as tests for the higher level function, which go into each of the functions.

jack60612:
I ended up doing 1 test that used the wallet to make sure that everything could be tested together

tests/wallet/test_wallet.py
```diff
@@ -774,3 +774,81 @@ async def test_address_sliding_window(self, wallet_node_100_pk, trusted):
...
+        )
+
+        await wallet_2.push_transaction(tx)
+        await full_node_api.full_node.respond_transaction(tx.spend_bundle, tx.name)
+        assert len(tx.removals) == 1
+        assert sum([coin.amount for coin in tx.removals]) == 4 * 10 ** 12
+
+        for i in range(0, num_blocks):
+            await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph))
+        # check that the knapsack algorithm works.
```
mariano54, line 832: How do you know that knapsack is getting triggered and not finding indentical coin algo?

tests/wallet/test_wallet.py
```diff
@@ -774,3 +774,81 @@ async def test_address_sliding_window(self, wallet_node_100_pk, trusted):
...
+
+        tx = await wallet.generate_signed_transaction(
+            4 * 10 ** 12,
+            ph_2,
+            0,
+        )
+        await wallet.push_transaction(tx)
+        await full_node_api.full_node.respond_transaction(tx.spend_bundle, tx.name)
+
+        for i in range(0, num_blocks):
```
mariano54, line 815: Why are we creating blocks?

mariano54 (commented):
This test is fine, but we should also have a few smaller tests for each method, to check that it's doing the right thing.

Those will be easier to write because you don't need a full node or wallet. You can just create coins like this:

`[Coin(std_hash(b"a"), std_hash(b"b"), 120000)]`

and then pass in specific lists. What we want is to hit all of the cases in each method.
For example, finding a smallest coin, or not finding it.

Knapsack finding identical, knapsack not finding identical. A large amount of coins to check if knapsack is too slow or if it's fine, etc. These are called unit tests and they are smaller than integration tests, and allow you to test fine grained properties about functions.

tests/wallet/test_wallet.py
jack60612: I thought you needed to create blocks to confirm the transaction

tests/wallet/test_wallet.py
jack60612: We do not have any coins with a value of 2.5 xch

tests/wallet/test_coin_selection.py
```diff
@@ -0,0 +1,47 @@
...
+        return std_hash(b"a")
+
+    def test_exact_match(self, a_hash):
+        coin_list = [
+            Coin(a_hash, a_hash, uint64(220000)),
+            Coin(a_hash, a_hash, uint64(120000)),
+            Coin(a_hash, a_hash, uint64(22)),
+        ]
+        assert check_for_exact_match(coin_list, uint64(220000)) == coin_list[0]
+        assert check_for_exact_match(coin_list, uint64(22)) == coin_list[2]
```
mariano54, line 21: Check for a failure case. if the coin is not there, assert that it is None

tests/wallet/test_coin_selection.py
```diff
@@ -0,0 +1,47 @@
...
+    def test_smallest_individual_coin_selection(self, a_hash):
+        coin_list = [
+            Coin(a_hash, a_hash, uint64(340000)),
+            Coin(a_hash, a_hash, uint64(300000)),
+            Coin(a_hash, a_hash, uint64(200000)),
+            Coin(a_hash, a_hash, uint64(123331)),
+            Coin(a_hash, a_hash, uint64(120000)),
+            Coin(a_hash, a_hash, uint64(110000)),
+            Coin(a_hash, a_hash, uint64(300)),
+        ]
```
mariano54, line 32: Test for a failure case

tests/wallet/test_coin_selection.py
```diff
@@ -0,0 +1,47 @@
...
+        coin_list = {
+            Coin(a_hash, a_hash, uint64(200000)),
+            Coin(a_hash, a_hash, uint64(123331)),
+            Coin(a_hash, a_hash, uint64(120000)),
+            Coin(a_hash, a_hash, uint64(110000)),
+            Coin(a_hash, a_hash, uint64(300)),
+        }
+        knapsack = knapsack_coin_algorithm(coin_list, uint64(310000), DEFAULT_CONSTANTS.MAX_COIN_AMOUNT)
+        assert knapsack is not None
+        assert sum([coin.amount for coin in knapsack]) == 310000
```
mariano54, line 46: Test for a non exact match

mariano54: I would add a test with a large number of coins, make a for loop and add like values of 10000*n for in 1...1000

mariano54: Call the method 100 times and assert that it's not None each time, to see that it always has a guaranteed result if the coin values add up to more than the target

mariano54 (dismissed):

jack60612:
closing for ci

jack60612:
Superseded  by  #9975

PR #9918, opened 2022-01-24
Fix streamable for Python 3.10

altendky:
Use the `dataclasses.fields()` with a preference for the hints processed through `typing.get_type_hints()` to handle properly processing stringy hints to actual classes.  This makes `tests.core.util.test_streamable` pass locally with Python 3.10 while it fails without these changes.  Hopefully by using a couple higher level APIs this will be a bit less prone to detailed changes like this.

```python-console
Python 3.9.5 (default, Jun  3 2021, 15:18:23) 
[GCC 9.3.0] on linux
>>> class C:
...     x: int
... 
>>> class D(C):
...     pass
... 
>>> C.__annotations__
{'x': <class 'int'>}
>>> D.__annotations__
{'x': <class 'int'>}
```

```python-console
Python 3.10.1 (main, Jan 12 2022, 20:28:45) [GCC 9.3.0] on linux
>>> class C:
...     x: int
... 
>>> class D(C):
...     pass
... 
>>> C.__annotations__
{'x': <class 'int'>}
>>> D.__annotations__
{}
```

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -7,7 +7,7 @@ import io
 import pprint
 import sys
 from enum import Enum
-from typing import Any, BinaryIO, Dict, List, Tuple, Type, Callable, Optional, Iterator
+from typing import Any, BinaryIO, Dict, get_type_hints, List, Tuple, Type, Callable, Optional, Iterator
 
 from blspy import G1Element, G2Element, PrivateKey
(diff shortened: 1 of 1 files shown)


hoffmang9 (approved):
Woo

mariano54 (approved):

PR #9920, opened 2022-01-24
add catch to catch if farmer is not started when creating a plotnft.

jack60612:


chia/cmds/plotnft_funcs.py
```diff
@@ -64,7 +64,19 @@ async def create(args: dict, wallet_client: WalletRpcClient, fingerprint: int) -
     farmer_rpc_port = config["farmer"]["rpc_port"]
     # get total amount of plotnft's
     farmer_client = await FarmerRpcClient.create(self_hostname, uint16(farmer_rpc_port), DEFAULT_ROOT_PATH, config)
-    total_plot_nfts = len((await farmer_client.get_pool_state())["pool_state"])
+    try:
```
mariano54, line 67: Two things:
If we catch an exception we don't know if it's actually that the farmer is not started, or something else is wrong. We need to catch the correct exception for that. (See line 211). 

Also, if we want to add this, we should add it to all the farmer API calls consistently. For example check out line 211, maybe we want to do a wrapper function that catches these exceptions and use it in multiple places.

jack60612:
oops, ignore that.

chia/cmds/plotnft_funcs.py
jack60612: I made the exception more precise, i don't really think a wrapper function is necessary.
This is mostly a hotfix for an edgecase so I think it is fine.

PR #9921, opened 2022-01-24
Update to handle the NodeJS 16 dependency

altendky:
Draft for:
- [x] Consider using a local `n` install for all platforms, not just Ubuntu
- [ ] Make sure the instructions at the end of installation properly reflect each platform
- [ ] Generally lots of testing.
- [x] Maybe use NodeJS 16 instead of `stable`
- [ ] Should we wipe the `.n` directory before installing to it?

### DIFF
--- a/.github/workflows/test-install-scripts.yml
+++ b/.github/workflows/test-install-scripts.yml
@@ -65,10 +65,10 @@ jobs:
           type: arch
           url: "docker://archlinux:latest"
         - name: centos:7
-          type centos
+          type: centos
           url: "docker://centos:7"
         - name: centos:8
-          type centos
+          type: centos
           url: "docker://centos:8"
         - name: debian:buster
           type: debian
--- a/.github/workflows/test-install-scripts.yml
+++ b/.github/workflows/test-install-scripts.yml
@@ -124,9 +124,22 @@ jobs:
 
     - name: Prepare CentOS
       if: ${{ matrix.distribution.type == 'centos' }}
+      # Installing Git from yum brings git@1.x which doesn't work on actions/checkout.
+      # So install git@2.x from source
       run: |
-        yum update --assumeyes
-        yum install git sudo --assumeyes
+        if [ "$(rpm --eval %{centos_ver})" = "8" ]; then
+          sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-Linux-*;
+        fi
+        yum update -y
+        yum install -y sudo gcc autoconf make wget curl-devel expat-devel gettext-devel openssl-devel perl-devel zlib-devel
+        wget https://mirrors.edge.kernel.org/pub/software/scm/git/git-2.29.0.tar.gz
+        tar zxf git-2.29.0.tar.gz
+        pushd git-2.29.0
+        make configure
+        ./configure --prefix=/usr/local
+        make all
+        make install
+        popd
 
     - name: Prepare Debian
       if: ${{ matrix.distribution.type == 'debian' }}
--- a/.github/workflows/test-install-scripts.yml
+++ b/.github/workflows/test-install-scripts.yml
@@ -64,7 +64,12 @@ jobs:
         - name: arch:latest
           type: arch
           url: "docker://archlinux:latest"
-        # TODO: what CentOS version provides Python3.7-3.9?
+        - name: centos:7
+          type centos
+          url: "docker://centos:7"
+        - name: centos:8
+          type centos
+          url: "docker://centos:8"
         - name: debian:buster
           type: debian
           # https://packages.debian.org/buster/python/python3 (3.7)
@@ -116,6 +121,12 @@ jobs:
       if: ${{ matrix.distribution.type == 'arch' }}
       run: |
         pacman --noconfirm --refresh base --sync git sudo
+
+    - name: Prepare CentOS
+      if: ${{ matrix.distribution.type == 'centos' }}
+      run: |
+        yum update --assumeyes
+        yum install git sudo --assumeyes
 
     - name: Prepare Debian
       if: ${{ matrix.distribution.type == 'debian' }}
--- a/.gitignore
+++ b/.gitignore
@@ -90,6 +90,9 @@ win_code_sign_cert.p12
 # chia-blockchain wheel build folder
 build/
 
+# Temporal `n` (node version manager) directory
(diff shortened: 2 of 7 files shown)


ChiaMineJP:
We can close the issue/PR below after this PR is merged.
https://github.com/Chia-Network/chia-blockchain/issues/9727
https://github.com/Chia-Network/chia-blockchain/pull/9728

ChiaMineJP:
@wjblanke @altendky 
(This PR updates install scripts. Maybe more members should review the changed files)

# Changelog

### Changed
- Removed `sudo npm install n` and put `cd build_scripts/npm_global; npm ci` instead.
  - `n` is a nodejs version switcher
  - `n` is installed to `build_scripts/npm_global/node_modules`  if `npm>=7` is not available on system.
- If `npm>=7` is not available, it automatically install `npm>=7` to `<chia-blockchain root dir>/.n`.
  - So it won't override global npm/nodejs anymore.

### Fixed
- Fixed an issue where `install-gui.sh` failed if only `npm<7` was available on the system.
- Fixed an issue where `install-gui.sh` did not work as expected when an user runs the script not from a directory where the shell script file exists.
- Fixed an issue where `install-gui.sh` tried to run `sudo apt-get npm nodejs libxss1` even if `nodejs` is already installed.
- Fixed an issue where `install.sh` tried to run `sudo apt-get install bc -y` even if `bc` is already installed.

### Added
- Added `start-gui.sh`.
  - This automatically tries to load local nodejs environment if it's available.
  - If `npm>=7` is not available from system/local nodejs environment, asks users to install `npm>=7`
  - This checks whether GUI build files are available on startup. If not, tell user to build the GUI app before starting.
    - This is to help users who accidentally run `start-gui.sh` before running `install-gui.sh`.

ChiaMineJP:
Found out that Ubuntu18 which is ec2 instance(t2.medium) with 4GB RAM struggles to process GUI build.  Will do some researching how much RAM is sufficient to build GUI.

ChiaMineJP:
In my observation, `install-gui.sh` consumes 5.1GB at peak. So when running `install-gui.sh` user need to have at least 5.1GB free RAM.

|OS|RAM|intall-gui.sh result|
|-|-|-|
|Ubuntu18(t2.medium)|4GB|Sometimes runing out of memory, sometimes build process completes.|
|Ubuntu18(t2.large)|8GB|OK|
|AmazonLinux2 kernel5.10(t2.large)|8GB|OK|
|RHEL8.5 (t2.large)|8GB|OK (RHEL8.5 on ec2 did not have python installed by default. So I updated `install.sh` to install python which was missing in the script)|

emlowe:
I tested Ubuntu 18 and Rocky OS 8. Both worked fine, although those were not an entirely new OS install (I had installed previous versions of chia)

Centos 7 doesn't work as there is apparently no python 3.9 package and this causes a failure:
`No package python39 available`

ChiaMineJP:
@emlowe Thanks for testing. CentOS7, I'll check and fix the script. IIRC CentOS7 even do not have `sudo` installed by default so checking whether `sudo` exists is what I'm going to modify. But as far as installing `sudo`, I think I shouldn't because such an important security module should be installed by admin of the computer and should not be urged to install by non-admin user/application.
Will do my best to fix anyway.

emlowe:
I don't think I've ever installed Centos 7 and not had `sudo`.

ChiaMineJP:
IIRC CentOS7 in docker container used in CI runner for the rpm installer showed error that `sudo` was not available. I remember I was surprised to see that. I'll install CentOS7 on my local Hyper-V environment from CD installer and see whether `sudo` actually exists or not by default

ChiaMineJP:
I've found that installing Python on CentOS7 does not come easy. You cannot simply `sudo yum install python3.9` to install python3 because default package repository for CentOS7 seems not to have python(>=3.7) module(If you run `sudo yum install python3` it tries to install Python3.6).
I'll find the way to install chia-compatible version of Python.

altendky:
I didn't add any CentOS docker (CLI) installer tests exactly because they seemed to not provide new enough packages.  Isn't CentOS dead anyways?  Was Rocky, as you reference Earle, the replacement?  Anyways, not a clean install here but worked straight away on Ubuntu 20.04.  Thanks for all the work on this.  I guess I'll go look at adding Rocky to the (CLI) installer test set.

hoffmang9:
I had earlier made the call that CentOS 8 was as far back as we need to go. I think it's safe to return to that thought.

altendky:
For reference on CentOS installs.  https://github.com/Chia-Network/chia-blockchain/wiki/INSTALL#centosred-hatfedora

ChiaMineJP:
I've updated `install.sh` to be run without error on CentOS7. I'm testing CentOS8 now.

emlowe:
Centos moved from a downstream model (Now called Centos Linux) to an upstream model and rebranded as Centos Stream

Centos Linux 7 is supported until 2024
Centos Linux 8 is EOL

Centos Stream 8 is supported until 2024, but because of the change in the model, this release gets patches continuously **before** they go into RHEL - there I don't think anyone is using it as they used to since it is upstream of RHEL it is no longer a bug-for-bug proxy for RHEL. Centos Stream 8 has no minor patch versions, as patches are just rolled into it whenever

Centos Stream 9 is also supported until 2027

Rocky 8 is a community-driven response to this change to make a release that is downstream of RHEL and is bug-for-bug compatible with RHEL 8 and is supported until 2029

Here endeth the saga that is Centos.

ChiaMineJP:
@wjblanke @emlowe @altendky
I tested the installer and 90% of work has been done.
For now I need to solve a problem where at AmazonLinux2 the default version of sqlite3 available is 3.7, which causes an error creating partial index. (create index xxx on yyy _**where zzz**_).
Once this issue is resolved, I'll suggest you to merge this PR.
Just FYI.
See https://github.com/Chia-Network/chia-blockchain/pull/9921#issuecomment-1026351971 for detail

altendky:
Thanks for keeping this going.  Whatever you have added support for in `install.sh` should get added to the matrix in `test-install-scripts.yml`.

https://github.com/Chia-Network/chia-blockchain/blob/2314f47ecc1883cdf655e382a0cee7f501ce035c/.github/workflows/test-install-scripts.yml#L60-L107

ChiaMineJP:
@altendky Added centos7/8 docker urls to the install test.

emlowe (approved):
Looks good - nice work!

PR #9922, opened 2022-01-24
pool_wallet: Deterministic derivation of auth key, and only count poo…

mariano54:
…lWallets for indeces.

Problem: we have a hardcoded limit of 20 wallets/plotNFTs, and this includes CATs.  The reason is for syncing a 2nd machine and looking for keys: that we try index 0 to 19 for every possible wallet ID, so it's a lot of BLS operations which take time. There are two ways we can improve this. The first is just hardcoding the use of index 0 for all authentication keys. We are not rotating them anyway, and if we did, 20 wouldn't be enough. The second is to only increment indeces for pooling wallets. There is no need to skip index 3 if index 3 is a CAT wallet or something else. 

This change is fully backwards compatible, and will start assigning indeces from the highest pool wallet index the user has. The limit has been increased to 100 from 20 due to the auth key optimization. Also this can allow the user to have many CAT wallets but not have any effect on plotNFT limits

The auth key optimization also means that we will hopefully not have any more auth key errors while pooling, because the auth key can be found deterministically from the owner key.

TODO:

- [x] self review
- [x] add an automated test
- [x] manually test with real farmers
- [x] potentially remove auth key from configuration altogether.
- [ ] consider catching invalid auth key and doing a PUT farmer
- [x] test plotNFT limit on CLI and GUI

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -445,25 +445,22 @@ class PoolWallet:
         )
         return standard_wallet_record, p2_singleton_puzzle_hash, launcher_coin_id
 
-    async def get_pool_wallet_index(self) -> uint32:
+    async def _get_owner_key_cache(self) -> Tuple[PrivateKey, uint32]:
         if self._owner_sk_and_index is None:
-            owner_sk_and_index: Optional[Tuple[PrivateKey, uint32]] = find_owner_sk(
+            self._owner_sk_and_index = find_owner_sk(
                 [self.wallet_state_manager.private_key], (await self.get_current_state()).current.owner_pubkey
             )
-            assert owner_sk_and_index is not None
-            self._owner_sk_and_index = owner_sk_and_index
-        return self._owner_sk_and_index[1]
+        assert self._owner_sk_and_index is not None
+        return self._owner_sk_and_index
+
+    async def get_pool_wallet_index(self) -> uint32:
+        return (await self._get_owner_key_cache())[1]
 
     async def sign(self, coin_spend: CoinSpend) -> SpendBundle:
         async def pk_to_sk(pk: G1Element) -> PrivateKey:
-            if self._owner_sk_and_index is None:
-                owner_sk_and_index: Optional[Tuple[PrivateKey, uint32]] = find_owner_sk(
-                    [self.wallet_state_manager.private_key], (await self.get_current_state()).current.owner_pubkey
-                )
-                assert owner_sk_and_index is not None
-                self._owner_sk_and_index = owner_sk_and_index
-            assert self._owner_sk_and_index[0].get_g1() == pk
-            return self._owner_sk_and_index[0]
+            sk, _ = await self._get_owner_key_cache()
+            assert sk.get_g1() == pk
+            return sk
 
         return await sign_coin_spends(
             [coin_spend],
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -70,6 +70,7 @@ class PoolWallet:
     standard_wallet: Wallet
     wallet_id: int
     singleton_list: List[Coin]
+    _pool_wallet_index_cache: Optional[uint32]
     """
     From the user's perspective, this is not a wallet at all, but a way to control
     whether their pooling-enabled plots are being self-farmed, or farmed by a pool,
@@ -235,7 +236,7 @@ class PoolWallet:
 
         if make_new_authentication_key or existing_config is None:
             new_auth_sk: PrivateKey = master_sk_to_pooling_authentication_sk(
-                self.wallet_state_manager.private_key, uint32(self.wallet_id), uint32(0)
+                self.wallet_state_manager.private_key, await self.get_pool_wallet_index(), uint32(0)
             )
             auth_pk: G1Element = new_auth_sk.get_g1()
             payout_instructions: str = (await self.standard_wallet.get_new_puzzlehash(in_transaction=True)).hex()
@@ -332,6 +333,7 @@ class PoolWallet:
         this method.
         """
         self = PoolWallet()
+        self._pool_wallet_index_cache = None
         self.wallet_state_manager = wallet_state_manager
 
         self.wallet_info = await wallet_state_manager.user_store.create_wallet(
@@ -368,6 +370,7 @@ class PoolWallet:
         to do anything here.
         """
         self = PoolWallet()
+        self._pool_wallet_index_cache = None
         self.wallet_state_manager = wallet_state_manager
         self.wallet_id = wallet_info.id
         self.standard_wallet = wallet
@@ -448,11 +451,24 @@ class PoolWallet:
         )
         return standard_wallet_record, p2_singleton_puzzle_hash, launcher_coin_id
 
+    async def get_pool_wallet_index(self) -> uint32:
+        if self._pool_wallet_index_cache is not None:
+            return self._pool_wallet_index_cache
+
+        owner_sk_and_index: Optional[Tuple[PrivateKey, uint32]] = await find_owner_sk(
+            [self.wallet_state_manager.private_key], (await self.get_current_state()).current.owner_pubkey
+        )
+        assert owner_sk_and_index is not None
+        self._pool_wallet_index_cache = owner_sk_and_index[1]
+        return self._pool_wallet_index_cache
+
     async def sign(self, coin_spend: CoinSpend) -> SpendBundle:
         async def pk_to_sk(pk: G1Element) -> PrivateKey:
-            owner_sk: Optional[PrivateKey] = await find_owner_sk([self.wallet_state_manager.private_key], pk)
-            assert owner_sk is not None
-            return owner_sk
+            owner_sk_and_index: Optional[Tuple[PrivateKey, uint32]] = await find_owner_sk(
+                [self.wallet_state_manager.private_key], pk
+            )
+            assert owner_sk_and_index is not None
+            return owner_sk_and_index[0]
 
         return await sign_coin_spends(
             [coin_spend],
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -69,7 +69,8 @@ class PoolWallet:
     standard_wallet: Wallet
     wallet_id: int
     singleton_list: List[Coin]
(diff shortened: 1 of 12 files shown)


tests/pools/test_pool_rpc.py
```diff
@@ -372,6 +374,67 @@ async def test_create_multiple_pool_wallets(self, one_wallet_node_and_rpc, fee,
...
+                pool_list: List[Dict] = full_config["pool"]["pool_list"]
+                assert len(pool_list) == i + 3
+                if i == 0:
+                    # Ensures that the CAT creation does not cause pool wallet IDs to increment
+                    summaries_response = await client.get_wallets()
+                    print(summaries_response)
+                    for wallet in wallet_node_0.wallet_state_manager.wallets.values():
+                        if wallet.type() == WalletType.POOLING_WALLET:
+                            status: PoolWalletInfo = (await client.pw_status(wallet.id()))[0]
+                            print(await wallet.get_pool_wallet_index())
```
xdustinface, line 423:
```suggestion
```

chia/pools/pool_wallet.py
```diff
@@ -70,6 +69,7 @@ class PoolWallet:
     standard_wallet: Wallet
     wallet_id: int
     singleton_list: List[Coin]
+    _pool_wallet_index_cache: Optional[uint32]
```
xdustinface, line 72:
```suggestion
    _owner_sk_and_index: Optional[Tuple[uint32, PrivateKey]]
```

Wouldn't something like above (or just have two members) make sense to avoid lookups for each signature?

xdustinface (changes requested):
Looks good. See two nits and a question below :)

chia/pools/pool_wallet.py
mariano54: yeah that's a good change thanks

xdustinface:
@mariano54 I think we should factor out the cache generation to avoid that code duplicates, see 1f74ceaac04303669dc6aee38250b05d5b1a8bda.

tests/core/test_farmer_harvester_rpc.py
```diff
@@ -251,6 +248,7 @@ async def test_get_harvesters():
             ]
             config["pool"]["pool_list"] = pool_list
             save_config(root_path, "config.yaml", config)
+
```
xdustinface, line 251:
```suggestion
```

tests/pools/test_pool_rpc.py
```diff
@@ -206,9 +207,10 @@ async def test_create_new_pool_wallet_self_farm(self, one_wallet_node_and_rpc, f
         assert len(pool_list) == 1
         pool_config = pool_list[0]
         assert (
-            pool_config["authentication_public_key"]
-            == "0xb3c4b513600729c6b2cf776d8786d620b6acc88f86f9d6f489fa0a0aff81d634262d5348fb7ba304db55185bb4c5c8a4"
+            pool_config["owner_public_key"]
+            == "0xb286bbf7a10fa058d2a2a758921377ef00bb7f8143e1bd40dd195ae918dbef42cfc481140f01b9eae13b430a0c8fe304"
         )
+
```
xdustinface, line 213:
```suggestion
```

tests/pools/test_pool_rpc.py
```diff
@@ -372,6 +374,66 @@ async def test_create_multiple_pool_wallets(self, one_wallet_node_and_rpc, fee,
...
+        def wallet_is_synced():
+            return (
+                wallet_node_0.wallet_state_manager.blockchain.get_peak_height()
+                == full_node_api.full_node.blockchain.get_peak_height()
+            )
+
+        # Create some CAT wallets to increase wallet IDs
+        for i in range(5):
+            res = await client.create_new_cat_and_wallet(20)
+            await asyncio.sleep(2)
```
xdustinface, line 386: What are we sleeping here? Just asking because it looks suspicious to become a flaky CI test in case we are waiting for something here.

tests/pools/test_pool_rpc.py
```diff
@@ -372,6 +374,66 @@ async def test_create_multiple_pool_wallets(self, one_wallet_node_and_rpc, fee,
...
+                            )
+                            assert auth_sk is not None
+                            owner_sk = find_owner_sk(
+                                [wallet_0.wallet_state_manager.private_key], status.current.owner_pubkey
+                            )
+                            assert owner_sk is not None
+                            assert owner_sk != auth_sk
+
+                            summaries_response = await client.get_wallets()
+            print(summaries_response)
```
xdustinface, line 435:
```suggestion
```

tests/pools/test_pool_rpc.py
```diff
@@ -372,6 +374,66 @@ async def test_create_multiple_pool_wallets(self, one_wallet_node_and_rpc, fee,
...
+                await self.farm_blocks(full_node_api, our_ph_2, 2)
+                await time_out_assert(20, wallet_is_synced)
+
+                full_config: Dict = load_config(wallet_0.wallet_state_manager.root_path, "config.yaml")
+                pool_list: List[Dict] = full_config["pool"]["pool_list"]
+                assert len(pool_list) == i + 3
+                if i == 0:
+                    # Ensures that the CAT creation does not cause pool wallet IDs to increment
+                    summaries_response = await client.get_wallets()
+                    print(summaries_response)
```
xdustinface, line 419:
```suggestion
```

xdustinface: Im not sure if you want to keep all those `print` calls in the whole test. Looked to me like debugging thats why i proposed to drop them all.

And in this and the below case also the `client.get_wallets` seems redundant?

tests/pools/test_pool_rpc.py
xdustinface: Same here?

tests/wallet/test_wallet_user_store.py
```diff
@@ -0,0 +1,57 @@
...
+from chia.util.db_wrapper import DBWrapper
+from chia.wallet.util.wallet_types import WalletType
+
+from chia.wallet.wallet_user_store import WalletUserStore
+
+
+@pytest.fixture(scope="module")
+def event_loop():
+    loop = asyncio.get_event_loop()
+    yield loop
```
xdustinface, line 15:
```suggestion
```

tests/wallet/test_wallet_user_store.py
```diff
@@ -0,0 +1,57 @@
+import asyncio
```
xdustinface, line 1:
```suggestion
```

tests/wallet/test_wallet_user_store.py
```diff
@@ -0,0 +1,57 @@
...
+from chia.wallet.wallet_user_store import WalletUserStore
+
+
+@pytest.fixture(scope="module")
+def event_loop():
+    loop = asyncio.get_event_loop()
+    yield loop
+
+
+class TestWalletUserStore:
```
xdustinface, line 18: This class could be dropped which would save an indent for the whole test.

tests/wallet/test_wallet_user_store.py
```diff
@@ -0,0 +1,57 @@
...
+        try:
+            await store.init_wallet()
+            assert (await store.get_last_wallet()).id == 1
+            wallet = await store.create_wallet("CAT_WALLET", WalletType.CAT, "abc")
+            assert wallet is not None
+            assert (await store.get_last_wallet()).id == 2
+            wallet = await store.create_wallet("CAT_WALLET", WalletType.CAT, "abc")
+            wallet = await store.create_wallet("CAT_WALLET", WalletType.CAT, "abc")
+            wallet = await store.create_wallet("CAT_WALLET", WalletType.CAT, "abc")
+            assert (await store.get_last_wallet()).id == 5
```
xdustinface, line 38:
```suggestion
            for i in range(1, 5):
                assert (await store.get_last_wallet()).id == i
                wallet = await store.create_wallet("CAT_WALLET", WalletType.CAT, "abc")
                assert wallet.id == i + 1
```

tests/wallet/test_wallet_user_store.py
```diff
@@ -0,0 +1,57 @@
...
+
+            print(await store.get_all_wallet_info_entries())
```
xdustinface, line 40:
```suggestion
```

tests/wallet/test_wallet_user_store.py
```diff
@@ -0,0 +1,57 @@
...
+            for i in range(2, 6):
+                await store.delete_wallet(i, in_transaction=False)
+
+            print(await store.get_all_wallet_info_entries())
```
xdustinface, line 44:
```suggestion
```

xdustinface (changes requested):
Some more thoughts below + it would make sense to run `isort` for the new test file.

tests/pools/test_pool_rpc.py
mariano54: Full node needs to get the TX into the mempool, which we can't check because we don't know tx id here

tests/pools/test_pool_rpc.py
mariano54: good point thanks

tests/wallet/test_wallet_user_store.py
mariano54: Why canwe drop this?

xdustinface: The class just isn't used in any way, is it? You can just have the tests in there, see https://github.com/Chia-Network/chia-blockchain/blob/main/tests/farmer_harvester/test_farmer_harvester.py.

xdustinface (approved):

mariano54:
@wjblanke let's merge this so we can start getting some testing on it?

tests/wallet/test_wallet_user_store.py
```diff
@@ -0,0 +1,45 @@
...
+import pytest
+
+from chia.util.db_wrapper import DBWrapper
+from chia.wallet.util.wallet_types import WalletType
+
+from chia.wallet.wallet_user_store import WalletUserStore
+
+
+@pytest.mark.asyncio
+async def test_store(self):
```
xdustinface, line 12:
```suggestion
async def test_store():
```

Leftover form the class.

xdustinface (changes requested):

PR #9925, opened 2022-01-24
Get roots returns store ids

almogdepaz:


chia/data_layer/data_layer.py
```diff
@@ -137,12 +137,14 @@ async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
...
         return res.node_hash
 
-    async def get_roots(self, store_ids: List[str]) -> List[Optional[bytes32]]:
+    async def get_roots(self, store_ids: List[str]) -> List[Dict[str, Any]]:
         roots = []
         for id in store_ids:
-            res = await self.data_store.get_tree_root(tree_id=bytes32(hexstr_to_bytes(id)))
+            id_bytes = bytes32(hexstr_to_bytes(id))
+            res = await self.data_store.get_tree_root(tree_id=id_bytes)
             if res is None:
```
altendky, line 145: `.get_tree_root()` doesn't return `None`.  So this handling was already broken...  I did ask about why mypy didn't warn about this and it seems like it might be a bug as similar scenarios but with variables for example can be warned about.  I need to follow up on that and find or file an issue.

chia/data_layer/data_layer.py
```diff
@@ -137,12 +137,14 @@ async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
...
-                self.log.error(f"Failed to get root for {id}")
+                self.log.warning(f"Failed to get root for {id}")
+                roots.append({"id": id_bytes, "root": None})
                 continue
-            roots.append(res.node_hash)
+            roots.append({"id": id_bytes, "root": res.node_hash})
```
altendky, line 149: We've got the terms 'store id' and 'root hash' and we took different parts from each.  Maybe `"store"` and `"root"` or `"id"` and `"hash"`?  Or just always the full term like below.

altendky (changes requested):

chia/data_layer/data_layer.py
```diff
@@ -137,12 +137,14 @@ async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
             self.log.error(f"Failed to get root for {store_id.hex()}")
         return res.node_hash
 
-    async def get_roots(self, store_ids: List[str]) -> List[Optional[bytes32]]:
+    async def get_roots(self, store_ids: List[str]) -> List[Dict[str, Any]]:
```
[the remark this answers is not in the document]
almogdepaz: good point, will fix

chia/data_layer/data_layer.py
almogdepaz: id, hash

chia/data_layer/data_layer.py
```diff
@@ -135,14 +134,5 @@ async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
         res = await self.data_store.get_tree_root(tree_id=store_id)
         if res is None:
             self.log.error(f"Failed to get root for {store_id.hex()}")
+            return None
```
altendky, line 137: This is pointless but the whole roots/`None` path needs reviewed so we don't need to burn CI cycles I guess right now.

chia/rpc/data_layer_rpc_api.py
```diff
@@ -157,5 +157,9 @@ async def get_roots(self, request: Dict[str, Any]) -> Dict[str, Any]:
         # todo input checks
         if self.service is None:
             raise Exception("Data layer not created")
-        res = await self.service.get_roots(store_ids)
-        return {"hashes": res}
+        roots = []
+        for id in store_ids:
+            id_bytes = bytes32.from_hexstr(id)
+            res = await self.service.get_root(store_id=id_bytes)
+            roots.append({"id": id_bytes, "hash": res})
```
altendky, line 164: Not a requested change, just for consideration.  How do we decide between `List[TypedDict]` vs just `Dict[str, str]`?  Like this could just be a mapping from the id to the hash.

altendky (approved):

PR #9926, opened 2022-01-24
cli error handling atari

almogdepaz:
handle rpc client close

handle root not in store

### DIFF
--- a/chia/cmds/data_funcs.py
+++ b/chia/cmds/data_funcs.py
@@ -25,25 +25,25 @@ async def get_client(rpc_port: Optional[int]) -> Tuple[DataLayerRpcClient, int]:
 
 async def create_data_store_cmd(rpc_port: Optional[int], table_string: str) -> Optional[Dict[str, Any]]:
     # TODO: nice cli error handling
-
+    client = None
+    response = None
     try:
         client, rpc_port = await get_client(rpc_port)
         response = await client.create_data_store()
     except aiohttp.ClientConnectorError:
         print(f"Connection error. Check if data is running at {rpc_port}")
-        return None
     except Exception as e:
         print(f"Exception from 'data': {e}")
-        return None
-
-    client.close()
-    await client.await_closed()
+    if client is not None:
+        client.close()
+        await client.await_closed()
     return response
 
 
 async def get_value_cmd(rpc_port: Optional[int], store_id: str, key: str) -> Optional[Dict[str, Any]]:
     # TODO: nice cli error handling
-
+    client = None
+    response = None
     store_id_bytes = bytes32(hexstr_to_bytes(store_id))
     key_bytes = hexstr_to_bytes(key)
     try:
@@ -51,13 +51,11 @@ async def get_value_cmd(rpc_port: Optional[int], store_id: str, key: str) -> Opt
         response = await client.get_value(store_id=store_id_bytes, key=key_bytes)
     except aiohttp.ClientConnectorError:
         print(f"Connection error. Check if data is running at {rpc_port}")
-        return None
     except Exception as e:
         print(f"Exception from 'data': {e}")
-        return None
-
-    client.close()
-    await client.await_closed()
+    if client is not None:
+        client.close()
+        await client.await_closed()
     return response
 
 
@@ -67,19 +65,20 @@ async def update_data_store_cmd(
     changelist: Dict[str, str],
 ) -> Optional[Dict[str, Any]]:
     # TODO: nice cli error handling
+    client = None
+    response = None
     store_id_bytes = bytes32(hexstr_to_bytes(store_id))
     try:
         client, rpc_port = await get_client(rpc_port)
         response = await client.update_data_store(store_id=store_id_bytes, changelist=changelist)
     except aiohttp.ClientConnectorError:
         print(f"Connection error. Check if data is running at {rpc_port}")
-        return None
     except Exception as e:
         print(f"Exception from 'data': {e}")
-        return None
 
-    client.close()
-    await client.await_closed()
+    if client is not None:
+        client.close()
+        await client.await_closed()
     return response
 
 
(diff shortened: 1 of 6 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1338,13 +1338,18 @@ async def create_new_dl(self, request) -> Dict:
...
-        dl_tx, std_tx, launcher_id = await dl_wallet.generate_new_reporter(
-            bytes32.from_hexstr(request["root"]), fee=request.get("fee", uint64(0))
-        )
+        try:
+            dl_tx, std_tx, launcher_id = await dl_wallet.generate_new_reporter(
+                bytes32.from_hexstr(request["root"]), fee=request.get("fee", uint64(0))
+            )
+        except ValueError as e:
+            log.error(f"Error while generating new reporter {e}")
+            return {"success": False, "error": str(e)}
```
altendky, line 1347: Doesn't the framework handle this such that just a bare `raise` would be fine after logging?  Or maybe the framework takes care of the logging as well?  The log should have the traceback.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1338,13 +1338,18 @@ async def create_new_dl(self, request) -> Dict:
...
         await self.service.wallet_state_manager.add_pending_transaction(dl_tx)
         await self.service.wallet_state_manager.add_pending_transaction(std_tx)
 
         return {
+            "success": True,
```
altendky, line 1352: Doesn't the framework add this?

chia/rpc/wallet_rpc_client.py
```diff
@@ -473,6 +473,8 @@ async def cancel_offer(self, trade_id: bytes32, fee=uint64(0), secure: bool = Tr
     async def create_new_dl(self, root: bytes32, fee: uint64) -> Tuple[List[TransactionRecord], bytes32]:
         request = {"root": root.hex(), "fee": fee}
         response = await self.fetch("create_new_dl", request)
+        # if response["error"] is not None:
+        #     return response["error"]
```
altendky, line 477:
```suggestion
```

chia/rpc/data_layer_rpc_api.py
```diff
@@ -11,6 +12,8 @@
 # todo input assertions for all rpc's
 from chia.util.streamable import recurse_jsonify
 
+log = logging.getLogger(__name__)
```
altendky, line 15: Don't we generally pass in loggers rather than creating them locally?  Also, the changes here look all unused?

chia/data_layer/data_store.py
```diff
@@ -317,6 +317,8 @@ async def get_tree_root(self, tree_id: bytes32, *, lock: bool = True) -> Root:
                 "SELECT * FROM root WHERE tree_id == :tree_id AND generation == :generation",
                 {"tree_id": tree_id.hex(), "generation": generation},
             )
+            if cursor.rowcount < 1:
+                raise Exception("tree id not in store")
```
altendky, line 321: Unless the ID can be considered semi-private?

```suggestion
                raise Exception(f"tree id not in store: {tree_id.hex()}")
```

chia/data_layer/data_store.py
```diff
@@ -330,6 +332,8 @@ async def get_roots_between(
                 "AND generation >= :generation_begin AND generation < :generation_end ORDER BY generation ASC",
                 {"tree_id": tree_id.hex(), "generation_begin": generation_begin, "generation_end": generation_end},
             )
+            if cursor.rowcount < 1:
+                raise Exception("tree id not in store")
```
altendky, line 336: Unless the ID can be considered semi-private?

```suggestion
                raise Exception(f"tree id not in store: {tree_id.hex()}")
```

chia/rpc/data_layer_rpc_client.py
```diff
@@ -1,11 +1,11 @@
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from chia.rpc.rpc_client import RpcClient
 from chia.types.blockchain_format.sized_bytes import bytes32
 
 
 class DataLayerRpcClient(RpcClient):
-    async def create_data_store(self) -> Dict[str, Any]:
+    async def create_data_store(self) -> Optional[Dict[str, Any]]:
```
altendky, line 8: When would this return `None`?

https://github.com/Chia-Network/chia-blockchain/blob/94e961f41019be9a4c3a1154623a4a56c5708711/chia/rpc/data_layer_rpc_api.py#L59-L63

https://github.com/Chia-Network/chia-blockchain/blob/94e961f41019be9a4c3a1154623a4a56c5708711/chia/rpc/rpc_client.py#L44-L50

```suggestion
    async def create_data_store(self) -> Dict[str, Any]:
```

altendky (changes requested):
Yes, if there's a resource to be managed make a context manager.  It looks like it might be able to handle all the existing  `try`/`except` plus the client close in a finally.

chia/cmds/data_funcs.py
```diff
@@ -25,39 +25,39 @@ async def get_client(rpc_port: Optional[int]) -> Tuple[DataLayerRpcClient, int]:
...
+            await client.await_closed()
     return response
 
 
 async def get_value_cmd(rpc_port: Optional[int], store_id: str, key: str) -> Optional[Dict[str, Any]]:
     # TODO: nice cli error handling
-
+    client = None
+    response = None
     store_id_bytes = bytes32(hexstr_to_bytes(store_id))
```
emlowe, line 48: I believe we prefer `bytes32.from_hexstr()` over `bytes32(hexstr_to_bytes())` - several instances here

emlowe (changes requested):

arvidn:
with the `finally` block I think using a context manager is even more called for. I made this PR against your PR: https://github.com/Chia-Network/chia-blockchain/pull/9956

altendky:
@arvidn, I generally prefer [the `@contextmanager` decorator](https://docs.python.org/3.10/library/contextlib.html#contextlib.contextmanager).  Was there a specific need for the class/method approach here?  Regardless, your context manager does move it all to a single location which is the real value.  Thanks!  I will add to the todo list to consider consuming the rest of the remaining `try`/`except` into either this context manager or another that uses it.

altendky (approved):

emlowe (approved):

PR #9927, opened 2022-01-24
fix get_ancestors in data_layer.py

almogdepaz:


chia/data_layer/data_layer.py
```diff
@@ -126,7 +126,7 @@ async def get_keys_values(self, store_id: bytes32) -> List[TerminalNode]:
         return res
 
     async def get_ancestors(self, node_hash: bytes32, store_id: bytes32) -> List[InternalNode]:
-        res = await self.data_store.get_ancestors(store_id, node_hash)
+        res = await self.data_store.get_ancestors(node_hash, store_id)
```
altendky, line 129: Also maybe defining some specific types instead of `bytes32` meaning everything.  `class NodeHash(bytes32):: pass` and the like.  But that's a bigger question.  Thanks for catching this.

```suggestion
        res = await self.data_store.get_ancestors(node_hash=node_hash, tree_id=store_id)
```

If we ever get the PRs all in maybe we can handle the `tree` -> `store` rename everywhere without making a pile of conflicts.

altendky (changes requested):

altendky (approved):

PR #9929, opened 2022-01-24
remove one magic number from weight_proof

arvidn:


mariano54 (approved):
Good one, that will speed things up a lot

arvidn:
actually, `_num_processes` is still hard coded to `4` in this same file, so it's a small step in the right direction :)

hoffmang9 (approved):
lgtm

mariano54:
What if we do the same as in blockchain.py
```python
     cpu_count = multiprocessing.cpu_count()
        if cpu_count > 61:
            cpu_count = 61  # Windows Server 2016 has an issue https://bugs.python.org/issue26903
        num_workers = max(cpu_count - reserved_cores, 1)
        ```

arvidn:
I'd have to defer to @Yostra . This code came in with merging the new wallet code

PR #9930, opened 2022-01-24
Support for Python 3.10

altendky:
This is just a duplicate of https://github.com/Chia-Network/chia-blockchain/pull/9765 to start.  Feel free to create PRs against this one for 3.10 fixes that can't go into `main`.  If they can go into main then submit them against `main`.  After they get merged you can close and reopen this PR (just buttons at the bottom) and it will trigger a rebuild including merging with the latest `main` so we can see the results of your fixes.

If you know some explicit categories that need fixed or discussed, please add them to the list of reasons this is still draft.

Notable points:
- This drops testing of macOS with Python 3.7
- Wheel availability is significantly incomplete for macOS Arm with Python 3.8.
- Wheel availability is not being checked for macOS Arm or Intel for Python 3.7.
- Python versions used for installers is not being changed.

Draft for:
- [x] Working in 3.10
- [x] Wheels for 3.10 on https://pypi.chia.net/simple/
  - Attempting to add a check for their availability in https://github.com/Chia-Network/chia-blockchain/pull/11243
  - Create a miniupnp 3.10 wheel (currently does not exist in standard nor our aux repo)
  - [x] pycryptodome for m1
  - [x] https://github.com/Chia-Network/chia-blockchain/pull/11277
- [x] First simplify `install.sh` Ubuntu version tracking in https://github.com/Chia-Network/chia-blockchain/pull/11288
- [x] Break out PRs
  - https://github.com/Chia-Network/chia-blockchain/pull/11129
  - https://github.com/Chia-Network/chia-blockchain/pull/11131
  - https://github.com/Chia-Network/chia-blockchain/pull/11134
  - https://github.com/Chia-Network/chia-blockchain/pull/11272
  - https://github.com/Chia-Network/chia-blockchain/pull/11273
  - https://github.com/Chia-Network/chia-blockchain/pull/11274

Maybe followup:
- [ ] Use the same version for all installers?  3.9?  3.10?


setup.py
```diff
@@ -3,10 +3,10 @@
...
     "multidict==5.1.0",  # Avoid 5.2.0 due to Avast
     "aiofiles==0.7.0",  # Async IO for files
-    "blspy==1.0.8",  # Signature library
-    "chiavdf==1.0.3",  # timelord and vdf verification
-    "chiabip158==1.0",  # bip158-style wallet filters
-    "chiapos==1.0.7",  # proof of space
+    "blspy==1.0.9",  # Signature library
+    "chiavdf==1.0.4",  # timelord and vdf verification
+    "chiabip158==1.1",  # bip158-style wallet filters
+    "chiapos==1.0.8",  # proof of space
```
arvidn, line 9: bumping these dependencies seem unrelated to supporting python 3.10. are there other python-3.10 fixes in those dependencies?

arvidn (dismissed):
is it OK to drop python 3.8 from CI?

setup.py
altendky: I think these are updates for these to provide wheels.  I'm not sure if there are other changes (nor entirely sure about that).  They are being updated in https://github.com/Chia-Network/chia-blockchain/pull/9808 separately anyways and presumably that goes in first.

altendky:
I'm guessing Gene was dropping 3.8 to keep just two CI runs for macOS.  You get lower runner count and they are slower.  But, I don't know my way around the macOS community and what is available and actually used etc.  For example, why did we already not have 3.7 on macOS?

hoffmang9:
I was indeed trying to keep the total Mac runners limited. All of our sub repos had to be version bumped to build 3.10 wheels.

hoffmang9:
The last two MacOS releases are 3.8+ and Silicon requires 3.9+ (though Silicon was back ported into 3.8 later). Brew installs 3.9 by default right now. As such 3.7 really isn't a factor on any mostly up to date Mac.

altendky:
We'll want to hit up the loop parameter warnings being ignored at https://github.com/Chia-Network/chia-blockchain/pull/10240/files#diff-338af70c6877cd271f4d5d1c05bbfc692ad15c5dbff0f7ea3c0e578f3f59c850R19.  `The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10`

hoffmang9 (dismissed):
lgtm pending CI

cmmarslender:
Trying to install on 22.04, I'm getting the following error, even though I had pre-installed python 3.10:

```E: Unable to locate package python3.9-venv
E: Couldn't find any package by glob 'python3.9-venv'
E: Couldn't find any package by regex 'python3.9-venv'```

Seems like apt is attempting to install python 3.9 even though 3.10 should be supported?

File: install.sh [hunk 2]
```diff
@@ -132,9 +135,13 @@ if [ "$(uname)" = "Linux" ]; then
     sudo apt-get update
     sudo apt-get install -y python3.8-venv python3-distutils openssl
   elif [ "$UBUNTU_2100" = "1" ]; then
-    echo "Installing on Ubuntu 21.04 or newer."
+    echo "Installing on Ubuntu 21.04."
```
cmmarslender, line 138: I think this could also be 21.10? Should this be "Intalling on Ubuntu 21.x" or something along those lines?

altendky: This continues the existing pattern.  I think we can take up addressing all of these version claims separately.

cmmarslender (approved):
lgtm. Tested all of the following with Python 3.10, and everything worked fine with the install scripts

Linux arm64
Linux amd64
Mac arm64
Mac intel
Windows

PR #9931, opened 2022-01-24
simplify cost calculation. It's already computed in rust

arvidn:
the rust `run_generator` function already computes the cost of the clvm + conditions. It needs to in order to terminate parsing of conditions as early as possible if it exceeds the cost limit.

Right now, for backwards compatibility reasons, we subtract the cost of the conditions in the returned `NPCResult`, because it has a field counting *just* the cost of CLVM (for historical reasons).

This patch replaces the `clvm_cost` field with a `cost` field, that counts the total cost of the generator. Including clvm, conditions and the size. This turns `compute_cost_of_program()` into a no-op. I intend to remove it in a separate patch.

### DIFF
--- a/chia/consensus/cost_calculator.py
+++ b/chia/consensus/cost_calculator.py
@@ -1,10 +1,7 @@
 from dataclasses import dataclass
-from typing import Dict, List, Optional
+from typing import List, Optional
 
-from chia.consensus.condition_costs import ConditionCost
 from chia.types.blockchain_format.program import SerializedProgram
-from chia.types.condition_opcodes import ConditionOpcode
-from chia.types.condition_with_args import ConditionWithArgs
 from chia.types.name_puzzle_condition import NPC
 from chia.util.ints import uint16, uint64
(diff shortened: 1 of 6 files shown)


chia/consensus/cost_calculator.py
```diff
@@ -1,50 +1,23 @@
...
-        else:
-            # all other conditions are free, and we ignore unknown conditions in
-            # order to allow for future soft forks
-            pass
-    return uint64(total_cost)
+    cost: uint64  # The total cost of the block, including CLVM cost, cost of
+    # conditions and cost of bytes
 
 
 def calculate_cost_of_program(program: SerializedProgram, npc_result: NPCResult, cost_per_byte: int) -> uint64:
```
mariano54, line 19: maybe we can just remove this function

arvidn: yeah, I intend to. I just wanted to keep the PR small and easy to review

mariano54 (approved):

PR #9933, opened 2022-01-24
Atari wallet rpc enhancements

Quexington:
These are new RPCs at the request of @fchirica

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -730,6 +730,10 @@ class DataLayerWallet:
         singleton: Optional[SingletonRecord] = await self.wallet_state_manager.dl_store.get_singleton_record(coin_id)
         return singleton
 
+    async def get_singletons_by_root(self, launcher_id: bytes32, root: bytes32) -> List[SingletonRecord]:
+        singletons = await self.wallet_state_manager.dl_store.get_singletons_by_root(launcher_id, root)
+        return singletons
+
     ##########
     # WALLET #
     ##########
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -731,7 +731,9 @@ class DataLayerWallet:
         return singleton
 
     async def get_singletons_by_root(self, launcher_id: bytes32, root: bytes32) -> List[SingletonRecord]:
-        singletons = await self.wallet_state_manager.dl_store.get_singletons_by_root(launcher_id, root)
+        singletons: List[SingletonRecord] = await self.wallet_state_manager.dl_store.get_singletons_by_root(
+            launcher_id, root
+        )
         return singletons
 
     ##########
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -711,9 +711,18 @@ class DataLayerWallet:
         )
         return singleton
 
-    async def get_history(self, launcher_id: bytes32) -> List[SingletonRecord]:
+    async def get_history(
+        self,
+        launcher_id: bytes32,
+        min_generation: Optional[uint32] = None,
+        max_generation: Optional[uint32] = None,
+        num_results: Optional[uint32] = None,
+    ) -> List[SingletonRecord]:
         history: List[SingletonRecord] = await self.wallet_state_manager.dl_store.get_all_singletons_for_launcher(
-            launcher_id
+            launcher_id,
+            min_generation,
+            max_generation,
+            num_results,
         )
         return history
 
--- a/chia/data_layer/dl_wallet_store.py
+++ b/chia/data_layer/dl_wallet_store.py
@@ -116,7 +116,7 @@ class DataLayerStore:
         num_results: Optional[uint32] = None,
     ) -> List[SingletonRecord]:
         """
-        Returns all stored singletons with a specific launcher ID.
+        Returns stored singletons with a specific launcher ID.
         """
         query_params: List[Union[bytes32, uint32]] = [launcher_id]
         for optional_param in (min_generation, max_generation, num_results):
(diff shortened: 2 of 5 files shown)


fchirica (dismissed):
lgtm

tests/wallet/rpc/test_dl_wallet_rpc.py
```diff
@@ -152,7 +152,17 @@ async def is_singleton_generation(rpc_client, lid, generation) -> bool:
...
-            assert await client_2.dl_history(launcher_id) == [singleton_record, new_singleton_record]
+            assert await client_2.dl_history(launcher_id) == [new_singleton_record, singleton_record]
+
+            assert await client.dl_history(launcher_id, min_generation=uint32(1)) == [new_singleton_record]
+            assert await client.dl_history(launcher_id, max_generation=uint32(0)) == [singleton_record]
+            assert await client.dl_history(launcher_id, num_results=uint32(1)) == [new_singleton_record]
+            assert await client.dl_history(launcher_id, num_results=uint32(2)) == [
+                new_singleton_record,
+                singleton_record,
+            ]
```
altendky, line 163: I think there are several more permutations that could be covered here.

chia/data_layer/dl_wallet_store.py
```diff
@@ -108,11 +108,29 @@ async def add_singleton_record(self, record: SingletonRecord, in_transaction: bo
...
+    async def get_all_singletons_for_launcher(
+        self,
+        launcher_id: bytes32,
+        min_generation: Optional[uint32] = None,
+        max_generation: Optional[uint32] = None,
+        num_results: Optional[uint32] = None,
+    ) -> List[SingletonRecord]:
         """
-        Returns all stored singletons.
+        Returns all stored singletons with a specific launcher ID.
```
altendky, line 119:
```suggestion
        Returns stored singletons with a specific launcher ID.
```

altendky (changes requested):
Please use `.execute()` as a context manager to get the cursor and make sure we close it.  Also use dicts with value names in the select queries anyways to avoid just-be-really-careful-about-order situations when possible.

altendky (dismissed):
Good corner cases to check would include:

- Negative minimum generation
- Negative maximum generation
- Both negative
- Maximum less than minimum
- Number of results negative
- Number of results less than maximum minus minimum
- Number of results greater than maximum minus minimum

And yes, this gets a bit crazy going through all these permutations.  It might be able to be setup with a parametrized test.  Anyways, this is why APIs with parameters that can conflict and override each other are a bit smelly.  Not always the wrong choice, but smelly.  Also hard to document clearly.

chia/data_layer/data_layer_wallet.py
```diff
@@ -711,16 +711,31 @@ async def get_latest_singleton(self, launcher_id: bytes32) -> Optional[Singleton
...
+    async def get_history(
+        self,
+        launcher_id: bytes32,
+        min_generation: Optional[uint32] = None,
+        max_generation: Optional[uint32] = None,
+        num_results: Optional[uint32] = None,
+    ) -> List[SingletonRecord]:
         history: List[SingletonRecord] = await self.wallet_state_manager.dl_store.get_all_singletons_for_launcher(
-            launcher_id
+            launcher_id,
```
almogdepaz, line 722: maybe change to dl_wallet_store to avoid confusion with the data_store ?

chia/data_layer/dl_wallet_store.py
```diff
@@ -108,11 +108,29 @@ async def add_singleton_record(self, record: SingletonRecord, in_transaction: bo
             if not in_transaction:
                 self.db_wrapper.lock.release()
 
-    async def get_all_singletons_for_launcher(self, launcher_id: bytes32) -> List[SingletonRecord]:
+    async def get_all_singletons_for_launcher(
+        self,
+        launcher_id: bytes32,
+        min_generation: Optional[uint32] = None,
+        max_generation: Optional[uint32] = None,
+        num_results: Optional[uint32] = None,
```
almogdepaz, line 116: why do we need the ability  to limit number of results between generations ?

chia/rpc/wallet_rpc_api.py
```diff
@@ -1383,10 +1384,25 @@ async def dl_latest_singleton(self, request) -> Dict:
 
         raise ValueError("No DataLayer wallet has been initialized")
 
+    async def dl_singletons_by_root(self, request) -> Dict:
+        """Get the singleton records that contain the specified root"""
+        if self.service.wallet_state_manager is None:
+            raise ValueError("The wallet service is not currently initialized")
+
+        for wallet in self.service.wallet_state_manager.wallets.values():
```
almogdepaz, line 1392: i would do a get_data_wallet method so we can avoid for looping any time we need the wallet

chia/data_layer/dl_wallet_store.py
Quexington: This is just generally consistent with how we do paginations in other stores.

chia/rpc/wallet_rpc_api.py
Quexington: I have a local branch for this and other changes that I'll PR when general changes will break less other PRs.

chia/data_layer/data_layer_wallet.py
Quexington: I have a local branch for this and other changes that I'll PR when general changes will break less other PRs.

chia/data_layer/dl_wallet_store.py
altendky: As a general comment, not a change request here, I think an API with separate min/max and min/count requests might be good.  Then there doesn't have to be any prioritization of whether max or count wins or the associated documentation and test explosion.

fchirica (approved):

PR #9935, opened 2022-01-24
Remove wallet-only as a service group

Quexington:
Should we do this?  Running the full node is not a dependency of the wallet anymore.

### DIFF
--- a/chia/util/service_groups.py
+++ b/chia/util/service_groups.py
@@ -10,8 +10,7 @@ SERVICES_FOR_GROUP = {
     "timelord": "chia_timelord_launcher chia_timelord chia_full_node".split(),
     "timelord-only": "chia_timelord".split(),
     "timelord-launcher-only": "chia_timelord_launcher".split(),
-    "wallet": "chia_wallet chia_full_node".split(),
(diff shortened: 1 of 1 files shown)


mariano54:
The expectation for those that used the command in the future is that the `chia start wallet` starts both. Maybe the better approach  is to remove that one instead, and replace it with "wallet_and_node".

mariano54:
> > The expectation for those that used the command before is that the `chia start wallet` starts both. Maybe the better approach is to remove that one instead, and replace it with "wallet_and_node".
> 
> Doesn't that end with the same effect though where running `chia start wallet` only starts the wallet? I feel like either we should do it this way or leave it as it is.

I was suggesting having:
`chia start wallet_and_node`
`chia start wallet_only`

So it's clear what is what. But I'm not opinionated, we can merge this as is

Yostra:
We could have a single `chia start wallet` command that would then prompt a user with: `do you want to start a full node with the wallet? (Y/N)`

Yostra:
that can be a separate PR, I like this more than what we did previously.

Yostra (approved):

wjblanke (approved):
aok

PR #9938, opened 2022-01-24
Keep backwards compatibility on coin record RPCs

Quexington:
When the `spent` field was removed from `CoinRecord` it changed the JSON output of the RPCs that return coin records.  This patch forces a spent field onto responses from the full node api as well as strips those spent responses when parsing in the api client.

### DIFF
--- a/chia/rpc/full_node_rpc_client.py
+++ b/chia/rpc/full_node_rpc_client.py
@@ -14,6 +14,11 @@ from chia.util.byte_types import hexstr_to_bytes
 from chia.util.ints import uint32, uint64
 
 
+def coin_record_dict_backwards_compat(coin_record: Dict[str, Any]):
+    del coin_record["spent"]
+    return coin_record
+
+
 class FullNodeRpcClient(RpcClient):
     """
     Client to Chia RPC, connects to a local full node. Uses HTTP/JSON, and converts back from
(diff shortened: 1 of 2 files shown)


mariano54:
We should have a method that gets called here, otherwise it will be very difficult to keep all of the calls in sync. Take a look at this PR which does a similar thing:
https://github.com/Chia-Network/chia-blockchain/pull/9212/files

chia/rpc/full_node_rpc_client.py
```diff
@@ -101,10 +107,9 @@ async def get_coin_records_by_names(
             d["start_height"] = start_height
         if end_height is not None:
             d["end_height"] = end_height
-        return [
-            CoinRecord.from_json_dict(coin)
-            for coin in (await self.fetch("get_coin_records_by_names", d))["coin_records"]
-        ]
+
+        response = await self.fetch("get_coin_records_by_names", d)
+        return [CoinRecord.from_json_dict(coin_record_dict_backwards_compat(coin)) for coin in response["coin_records"]]
```
arvidn, line 112: I'm surprised `from_json_dict()` doesn't tolerate additional fields

arvidn (approved):

mariano54 (approved):

PR #9939, opened 2022-01-24
Add an RPC to update singletons atomically

Quexington:


chia/rpc/wallet_rpc_api.py
```diff
@@ -1395,6 +1398,35 @@ async def dl_update_root(self, request) -> Dict:
...
+                for launcher, root in request["updates"].items():
+                    record = await wallet.create_update_state_spend(
+                        bytes32.from_hexstr(launcher), bytes32.from_hexstr(root)
+                    )
+                    tx_records.append(record)
+                # Now that we have all the txs, we need to aggregate them all into just one spend
+                modified_txs: List[TransactionRecord] = []
+                aggregate_spend = SpendBundle([], G2Element())
+                for tx in tx_records:
+                    if tx.spend_bundle is not None:
```
almogdepaz, line 1420: what does it mean if if tx.spend_bundle is None ?
shouldn't we abort in that case ?

Quexington: No, quite often we create txs without spendbundles because we needed to aggregate them onto another tx record.  Al spend bundles on tx records will get pushed when the tx is pushed and sometimes it's redundant (or an incoming transaction in which case you shouldn't push anything).  If the spend bundle is none we just need to move on to the next transaction.

almogdepaz (approved):

chia/rpc/wallet_rpc_api.py
```diff
@@ -1395,6 +1398,35 @@ async def dl_update_root(self, request) -> Dict:
 
         raise ValueError("No DataLayer wallet has been initialized")
 
+    async def dl_update_multiple(self, request) -> Dict:
+        """Update multiple singletons with new merkle roots"""
+        if self.service.wallet_state_manager is None:
+            return {"success": False, "error": "not_initialized"}
+
+        for _, wallet in self.service.wallet_state_manager.wallets.items():
```
altendky, line 1406:
```suggestion
        for wallet in self.service.wallet_state_manager.wallets.values():
```

chia/rpc/wallet_rpc_api.py
```diff
@@ -1395,6 +1398,35 @@ async def dl_update_root(self, request) -> Dict:
...
+                    )
+                    tx_records.append(record)
+                # Now that we have all the txs, we need to aggregate them all into just one spend
+                modified_txs: List[TransactionRecord] = []
+                aggregate_spend = SpendBundle([], G2Element())
+                for tx in tx_records:
+                    if tx.spend_bundle is not None:
+                        aggregate_spend = SpendBundle.aggregate([aggregate_spend, tx.spend_bundle])
+                        modified_txs.append(dataclasses.replace(tx, spend_bundle=None))
+                modified_txs[0] = dataclasses.replace(modified_txs[0], spend_bundle=aggregate_spend)
```
altendky, line 1423: I appreciate the `.replace()` usage.  `:]`

PR #9945, opened 2022-01-25
Adding additional data to RPCs for metrics

cmmarslender:
Adds data to RPCs to support metrics from the full node. Thinking it might make sense to merge these in by service rather than a huge PR, so this one is probably at a good place for a first iteration

### DIFF
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -488,3 +488,23 @@ class BlockStore:
         heights = [int(row[0]) for row in rows]
 
         return heights
+
+    async def count_compactified_blocks(self) -> int:
+        async with self.db.execute("select count(*) from full_blocks where is_fully_compactified=1") as cursor:
+            row = await cursor.fetchone()
+
+        if row is None:
+            raise Exception("row was None when querying compact blocks")
+
+        [count] = row
+        return int(count)
+
+    async def count_uncompactified_blocks(self) -> int:
+        async with self.db.execute("select count(*) from full_blocks where is_fully_compactified=0") as cursor:
+            row = await cursor.fetchone()
+
+        if row is None:
+            raise Exception("row was None when querying uncompact blocks")
+
+        [count] = row
+        return int(count)
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -493,8 +493,7 @@ class BlockStore:
         async with self.db.execute("select count(*) from full_blocks where is_fully_compactified=1") as cursor:
             row = await cursor.fetchone()
 
-        if row is None:
-            raise Exception("row was None when querying compact blocks")
+        assert row is not None
 
         [count] = row
         return int(count)
@@ -503,8 +502,7 @@ class BlockStore:
         async with self.db.execute("select count(*) from full_blocks where is_fully_compactified=0") as cursor:
             row = await cursor.fetchone()
 
-        if row is None:
-            raise Exception("row was None when querying uncompact blocks")
+        assert row is not None
 
         [count] = row
         return int(count)
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -329,9 +329,9 @@ class FullNode:
             self.log.error(f"Exception in peer discovery: {e}")
             self.log.error(f"Exception Stack: {error_stack}")
 
-    def _state_changed(self, change: str):
+    def _state_changed(self, change: str, change_data: Dict[str, Any] = None):
         if self.state_changed_callback is not None:
-            self.state_changed_callback(change)
+            self.state_changed_callback(change, change_data)
 
     async def short_sync_batch(self, peer: ws.WSChiaConnection, start_height: uint32, target_height: uint32) -> bool:
         """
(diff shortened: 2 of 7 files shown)


chia/full_node/block_store.py
```diff
@@ -488,3 +488,23 @@ async def get_random_not_compactified(self, number: int) -> List[int]:
         heights = [int(row[0]) for row in rows]
 
         return heights
+
+    async def count_compactified_blocks(self) -> int:
```
mariano54, line 492: Is there a benchmark for this? Looks pretty slow. Does it work on both v1 and v2 schemas? This needs a test for sure

chia/full_node/full_node.py
```diff
@@ -1763,7 +1763,13 @@ async def respond_unfinished_block(
             await self.server.send_to_all_except([msg], NodeType.FULL_NODE, peer.peer_node_id)
         else:
             await self.server.send_to_all([msg], NodeType.FULL_NODE)
-        self._state_changed("unfinished_block")
+
+        state_changed_data = {}
```
mariano54, line 1767: You might want to add more data here like:
- size of the generator bytes
- transaction generator ref list
- timestamp
- k size

mariano54: Or would that go into the new block state change?

chia/full_node/hint_store.py
```diff
@@ -38,3 +38,13 @@ async def add_hints(self, coin_hint_list: List[Tuple[bytes32, bytes]]) -> None:
             [(None,) + record for record in coin_hint_list],
         )
         await cursor.close()
+
+    async def count_hints(self) -> int:
+        async with self.coin_record_db.execute("select count(*) from hints") as cursor:
+            row = await cursor.fetchone()
+
+        if row is None:
```
mariano54, line 46: Is this none with an empty table? If not, we should use:
`assert row is not None`

mariano54: Also needs a test

chia/rpc/full_node_rpc_api.py
```diff
@@ -140,14 +172,29 @@ async def get_blockchain_state(self, _request: Dict):
...
         if self.service.server is not None:
             is_connected = len(self.service.server.get_full_node_connections()) > 0
         else:
             is_connected = False
         synced = await self.service.synced() and is_connected
 
+        compact_blocks = 0
+        uncompact_blocks = 0
+        with log_exceptions(self.service.log):
+            compact_blocks = await self.service.block_store.count_compactified_blocks()
```
mariano54, line 190: We have to ensure this is fast, because `get_blockchain_state` might get called very often by people. If it's not fast, I suggest we move it somewhere else where it can get called more rarely.

cmmarslender: I'm fine moving some of this stuff to a different RPC call, was just trying to fit in to existing calls if the data seemed to make sense.. I did time this at one point, but forget how quick it was - was a few ms IIRC, but will redo that and document in this PR

chia/full_node/hint_store.py
cmmarslender: TBH I'm not sure when this is none - if the table is empty, the returned value is just `0` for the count, not `None` here - I think this is just a possible result from the general `execute` call on the DB - if None happens here, something is messed up with the DB.

chia/rpc/full_node_rpc_api.py
cmmarslender: Moved this + hint counts to a new RPC call.

The slowest I observed all three calls to be was about 12ms combined. most of the time, they were closer to the 1-2ms range - This is on a relatively quick machine with an SSD, so could be a bit slower on an HDD or something slower.

Either way, being it its own endpoint now probably alleviates concerns about this running when people may not need this data.

chia/full_node/block_store.py
cmmarslender: `"is_fully_compactified tinyint,"` exists on both schema versions
Benchmark data is in another comment on the PR (also, moved to a separate RPC method now that will alleviate any concerns about it running too much, but its a few ms)

chia/full_node/full_node.py
cmmarslender: Moved all of this to `block` event. Didn't see timestamp immediately on `class FullBlock` - whats the best source for the timestamp? Where do I pull k size from in `respond_block`?

mariano54: It's all in the block itself. Check out the method `respond_block`, that's responsible for adding a block to the chain. After adding it, we can update the state for metrics

chia/full_node/full_node.py
```diff
@@ -1566,7 +1566,26 @@ async def respond_block(
...
             await self.sync_store.clear_sync_info()  # Occasionally clear sync peer info
-        self._state_changed("block")
+
+        state_changed_data: Dict[str, Any] = {
+            "transaction_block": False,
+            "k_size": block.reward_chain_block.proof_of_space.size,
+        }
+
+        if block.transactions_info is not None:
+            state_changed_data["transaction_block"] = True
```
mariano54, line 1576: Might as well add the timestamp here

chia/full_node/full_node.py
```diff
@@ -1566,7 +1566,26 @@ async def respond_block(
         self.full_node_store.clear_unfinished_blocks_below(clear_height)
         if peak.height % 1000 == 0 and not self.sync_store.get_sync_mode():
             await self.sync_store.clear_sync_info()  # Occasionally clear sync peer info
-        self._state_changed("block")
+
```
mariano54, line 1569: Actually I think this can all go into peak_post_processing_2 at the bottom of it, since it is kind of post processing related

mariano54 (commented):
LGTM. I think we should add tests for the websocket stuff eventually to make sure it doesn't break, but this doesn't have to be done now.

chia/full_node/full_node.py
cmmarslender: Discussed in keybase - the `ReceiveBlockResult` isn't available in post processing 2, so we're leaving here for now

mariano54 (dismissed):

mariano54 (dismissed):

mariano54 (approved):

wjblanke (approved):
aok

PR #9946, opened 2022-01-25
Data propagation integration

fchirica:


altendky:
Kick CI

emlowe:
Florin, superlinter is complaining: `chia/data_layer/data_store.py:1075:23: E1121: Too many positional arguments for constructor call (too-many-function-args)`

chia/data_layer/data_store.py
```diff
@@ -821,3 +879,223 @@ async def get_operations(
                 previous_root = current_root
 
         return result
+
+    async def download_data(self, subscription: Subscription, target_hash: bytes32, *, lock: bool = True) -> bool:
```
almogdepaz, line 883: i think this probably belongs in its own file

chia/data_layer/data_layer.py
```diff
@@ -136,3 +141,137 @@ async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
...
+    async def fetch_and_validate(self, subscription: Subscription) -> None:
+        tree_id = subscription.tree_id
+        singleton_record: Optional[SingletonRecord] = await self.wallet_rpc.dl_latest_singleton(tree_id)
+        if singleton_record is None:
+            return
+        if singleton_record.generation == uint32(0):
+            return
+        old_root: Optional[Root] = None
+        if await self.data_store.tree_id_exists(tree_id=tree_id):
+            old_root = await self.data_store.get_tree_root(tree_id=tree_id)
```
almogdepaz, line 179: if you try/catch on get_tree_root(tree_id=tree_id) you can remove tree_id_exists and call the db only once

chia/data_layer/data_store.py
```diff
@@ -821,3 +879,223 @@ async def get_operations(
...
+            async for row in cursor:
+                tree_id = bytes32.fromhex(row["tree_id"])
+                mode_value = int(row["mode"])
+                ip = row["ip"]
+                port = uint16(row["port"])
+                subscriptions.append(Subscription(tree_id, DownloadMode(mode_value), ip, port))
+
+        return subscriptions
+
+    async def get_wallet_generation(self, tree_id: bytes32, *, lock: bool = True) -> int:
```
almogdepaz, line 1082: this should be done by the wallet, the wallet already manages history for the singleton

fchirica: this is just to track how much we've validated before - it compares against `singleton_record.generation`, which is the target, not the value produced by this.

chia/data_layer/data_layer.py
```diff
@@ -136,3 +141,137 @@ async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
...
+                launcher_id=tree_id, min_generation=uint32(wallet_current_generation + 1)
+            )
+        # No root hash changes in the new wallet records, so ignore.
+        if (
+            old_root is not None
+            and old_root.node_hash is not None
+            and to_check[0].root == old_root.node_hash
+            and len(set(record.root for record in to_check)) == 1
+        ):
+            await self.data_store.set_wallet_generation(tree_id, int(singleton_record.generation))
```
almogdepaz, line 199: is this really needed ?
we don't fetch anything in this path so not sure why we update the generation

chia/data_layer/data_layer.py
```diff
@@ -136,3 +141,137 @@ async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
...
+        if subscription.mode is DownloadMode.HISTORY:
+            to_check = await self.wallet_rpc.dl_history(
+                launcher_id=tree_id, min_generation=uint32(wallet_current_generation + 1)
+            )
+        # No root hash changes in the new wallet records, so ignore.
+        if (
+            old_root is not None
+            and old_root.node_hash is not None
+            and to_check[0].root == old_root.node_hash
+            and len(set(record.root for record in to_check)) == 1
```
almogdepaz, line 197: i think you can expect the wallet rpc to return unique values for dl_history

fchirica: see Quex's comment above.

chia/data_layer/data_layer.py
fchirica: all hashes are identical here - so increase generation so that we return earlier next time: at the condition with checking singleton's generation against our generation.

chia/data_layer/data_layer.py
```diff
@@ -136,3 +141,137 @@ async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
...
+            f"Target wallet generation: {singleton_record.generation}."
+        )
+
+        downloaded = await self.data_store.download_data(subscription, singleton_record.root)
+        if not downloaded:
+            raise RuntimeError("Could not download the data.")
+
+        root = await self.data_store.get_tree_root(tree_id=tree_id)
+        # Wallet root hash must match to our data store root hash.
+        if root.node_hash is None or root.node_hash != to_check[0].root:
```
almogdepaz, line 218: is the comment wrong ?
it looks like both these root hashes where retrieved from the data_store and not the wallet

chia/data_layer/data_store.py
```diff
@@ -821,3 +879,223 @@ async def get_operations(
...
+                                node = left_hash
+
+                        if node is not None:
+                            raise RuntimeError("Did not download full data.")
+                        await ws.send_str(json.dumps({"type": "close"}))
+
+                await self._insert_root(
+                    bytes32.from_hexstr(root_json["tree_id"]),
+                    bytes32.from_hexstr(root_json["node_hash"]),
+                    Status(root_json["status"]),
```
almogdepaz, line 978: we would only download changes that are confirmed on chain so status here should always be CONFIRMED

chia/data_layer/data_layer.py
fchirica: to_check is retrieved from dl_wallet:

```
singleton_record: Optional[SingletonRecord] = await self.wallet_rpc.dl_latest_singleton(tree_id)
if subscription.mode is DownloadMode.LATEST:
    to_check = [singleton_record]
if subscription.mode is DownloadMode.HISTORY:
    to_check = await self.wallet_rpc.dl_history(
        launcher_id=tree_id, min_generation=uint32(wallet_current_generation + 1)
    )
```

chia/data_layer/data_store.py
```diff
@@ -821,3 +879,223 @@ async def get_operations(
...
+                                        f"Did not received expected node. Expected: {node} Received: {hash.hex()}"
+                                    )
+                                if len(stack) > 0:
+                                    node = stack.pop()
+                                else:
+                                    node = None
+                            else:
+                                left_hash = row["left"]
+                                right_hash = row["right"]
+                                left_hash_bytes = hexstr_to_bytes(left_hash)
```
almogdepaz, line 955: use bytes32.from_hexstr, these need to be byte32 when passed to get_tree_hash

chia/data_layer/data_store.py
```diff
@@ -821,3 +879,223 @@ async def get_operations(
...
+                        msg = await ws.receive()
+                        msg_json = json.loads(msg.data)
+                        answer = msg_json["answer"]
+                        for row in answer:
+                            if row["is_terminal"]:
+                                key = row["key"]
+                                value = row["value"]
+                                hash = Program.to((hexstr_to_bytes(key), hexstr_to_bytes(value))).get_tree_hash()
+                                if hash.hex() == node:
+                                    await self._insert_node(
```
almogdepaz, line 925: switch to use  _insert_terminal_node and _insert_terminal_node and _insert_internal_node instead of _insert_node

fchirica: the root is inserted last - so worst case we'll have some orphan nodes, as no node will link any of them.

what concerns me more is calling the internal functions, instead of public ones. so I'm thinking to refactor it to a `insert_generation` and a cache, in this way at least we call them only in 1 place instead of all over the place.

fchirica: this code was intended to include pagination - i.e. for a big tree not fitting in memory, split it into multiple batches. then we won't be able to fit all of it into a cache. but I think we can worry about this later, for now I changed it to send the whole tree anyway.

almogdepaz: you will have orphan nodes which you might try to download again and then im not sure what will happen, maybe not in this pr but i dont think we should commit unvalidated nodes to the main tables

chia/data_layer/data_store.py
```diff
@@ -821,3 +879,223 @@ async def get_operations(
...
+                            "node_hash": target_hash.hex(),
+                        }
+                        await ws.send_str(json.dumps(request))
+                        msg = await ws.receive()
+                        root_json = json.loads(msg.data)
+                        node = root_json["node_hash"]
+                        internal_nodes = 0
+                        terminal_nodes = 0
+                        stack: List[str] = []
+                        add_to_db_cache: Dict[str, Any] = {}
```
almogdepaz, line 907: internal_nodes and terminal_nodes counters are unused

chia/data_layer/data_store.py
```diff
@@ -821,3 +879,223 @@ async def get_operations(
...
+            elif subscription.mode is DownloadMode.HISTORY:
+                async with aiohttp.ClientSession() as session:
+                    async with session.ws_connect(URL) as ws:
+                        request = {
+                            "type": "request_root",
+                            "tree_id": tree_id.hex(),
+                            "node_hash": target_hash.hex(),
+                        }
+                        await ws.send_str(json.dumps(request))
+                        msg = await ws.receive()
```
almogdepaz, line 991: do we need a timeout for all these ws.receive() calls ?

chia/data_layer/data_layer.py
```diff
@@ -136,3 +141,74 @@ async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
...
+            old_root = await self.data_store.get_tree_root(tree_id=tree_id)
+        to_check: List[SingletonRecord] = []
+        if subscription.mode is DownloadMode.LATEST:
+            to_check = [singleton_record]
+        if subscription.mode is DownloadMode.HISTORY:
+            to_check = await self.wallet_rpc.dl_history(
+                launcher_id=tree_id, min_generation=current_generation + 1
+            )  # type: ignore
+
+        downloaded = await self.data_store.download_data(subscription, singleton_record.root)
```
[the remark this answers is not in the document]
almogdepaz: we dont need to download if the root didn't change

chia/data_layer/data_layer.py
almogdepaz: after reading it i still think this should be handled by the wallet rpc, dont see why we would want to return something with duplicated values from the rpc

 we can add a todo if you dont like this added to this pr

chia/data_layer/data_layer.py
almogdepaz: so that is not the real generation we hold ? it a +1 used for earlier exit ?

chia/data_layer/data_store.py
almogdepaz: looking into it we will skip values that are already in the node table and just do nothing but this should be the case in #10064 ancestor table as well

chia/data_layer/data_layer.py
fchirica: if say the hash stored in data store is {deadbeef, generation=5} but on chain data is initially {deadbeef, generation=5}, then gets updated as  {deadbeef, generation=6},  {deadbeef, generation=7}, we will update here the data store generation to {deadbeef, generation=7}. then, next time it'll detect there's no real change and avoid the extra `dl_history` call (from 5 to 7).

chia/data_layer/data_store.py
fchirica: added those constants: ```session.ws_connect(URL, timeout=180, heartbeat=60)```

I think at this phrase we don't want failure at all - so I went for a big timeout.

chia/data_layer/data_layer.py
almogdepaz: sorry i got a bit confused there

chia/data_layer/data_layer.py
almogdepaz: thanks for clarifying

emlowe (dismissed):
Approving to get in for functional testing
Wouldn't mind some tests around subscribe/unsubscribe

almogdepaz (approved):

altendky:
Kick for a recent merge build.

emlowe (approved):
ok

PR #9947, opened 2022-01-25
Atari wallet fees

Quexington:
Fees already existed on the creation of a data layer but not on its update or report.

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -390,11 +390,23 @@ class DataLayerWallet:
 
         return dl_record, std_record, launcher_coin.name()
 
+    async def create_tandem_xch_tx(self, fee: uint64, announcement_to_assert: Announcement) -> TransactionRecord:
+        chia_tx = await self.standard_wallet.generate_signed_transaction(
+            uint64(0),
+            (await self.standard_wallet.get_new_puzzlehash()),
+            fee=fee,
+            negative_change_allowed=False,
+            coin_announcements_to_consume={announcement_to_assert},
+        )
+        assert chia_tx.spend_bundle is not None
+        return chia_tx
+
     async def create_update_state_spend(
         self,
         launcher_id: bytes32,
         root_hash: bytes32,
-    ) -> TransactionRecord:
+        fee: uint64 = uint64(0),
+    ) -> List[TransactionRecord]:
         singleton_record, parent_lineage = await self.get_spendable_singleton_info(launcher_id)
 
         inner_puzzle_derivation: Optional[
@@ -428,7 +440,10 @@ class DataLayerWallet:
                 "memos": [launcher_id, root_hash, next_inner_puzzle.get_tree_hash()],
             }
         ]
-        inner_sol: Program = self.standard_wallet.make_solution(primaries=primaries)
+        inner_sol: Program = self.standard_wallet.make_solution(
+            primaries=primaries,
+            coin_announcements={b"$"} if fee > 0 else None,
+        )
         db_layer_sol = Program.to([0, inner_sol, current_inner_puzzle])
         full_sol = Program.to(
             [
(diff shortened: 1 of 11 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1387,11 +1387,14 @@ async def dl_update_root(self, request) -> Dict:
 
         for _, wallet in self.service.wallet_state_manager.wallets.items():
```
emlowe, line 1388: How about using `.values()` and eliminate the `_`

chia/rpc/wallet_rpc_api.py
```diff
@@ -1387,11 +1387,14 @@ async def dl_update_root(self, request) -> Dict:
...
+                records = await wallet.create_update_state_spend(
+                    bytes32.from_hexstr(request["launcher_id"]),
+                    bytes32.from_hexstr(request["new_root"]),
+                    fee=uint64(request.get("fee", 0)),
                 )
-                await self.service.wallet_state_manager.add_pending_transaction(record)
-                return {"tx_record": record.to_json_dict_convenience(self.service.config)}
+                for record in records:
+                    await self.service.wallet_state_manager.add_pending_transaction(record)
+                return {"tx_record": records[0].to_json_dict_convenience(self.service.config)}
```
emlowe, line 1397: So we only return the "dl_tx" or the first item in this list but we call add_pending_transaction for all the records.

chia/data_layer/data_layer_wallet.py
```diff
@@ -542,6 +578,14 @@ async def create_report_spend(self, launcher_id: bytes32) -> Tuple[TransactionRe
             type=uint32(TransactionType.OUTGOING_TX.value),
             name=singleton_record.coin_id,
         )
+        if fee > 0:
+            chia_tx = await self.create_tandem_xch_tx(fee, expected_announcement, coin_announcement=False)
+            aggregate_bundle = SpendBundle.aggregate([dl_tx.spend_bundle, chia_tx.spend_bundle])
+            dl_tx = dataclasses.replace(dl_tx, spend_bundle=aggregate_bundle)
+            chia_tx = dataclasses.replace(chia_tx, spend_bundle=None)
+            txs: List[TransactionRecord] = [dl_tx, chia_tx]
```
emlowe, line 586: This List is always either one item (dl_tx) or two items (dl_tx, chia_tx) , correct?

chia/rpc/wallet_rpc_api.py
Quexington: Yes I keep forgetting I can do this :)

chia/rpc/wallet_rpc_api.py
Quexington: Yeah I didn't want to break the RPCs that were currently in use.  I don't think which tx record you get is that important, but we can fix it later if it is.

chia/data_layer/data_layer_wallet.py
Quexington: Correct, but the reason it's a list is to keep consistentcy with the equivalent wallet methods `generate_signed_transaction` which should be returning lists of tx records (CAT wallet does right now, Standard Wallet does not).

emlowe (dismissed):
ok

altendky:
Kick CI

emlowe (dismissed):
Ok for me

chia/rpc/data_layer_rpc_api.py
```diff
@@ -36,11 +37,18 @@ def process_change(change: Dict[str, Any]) -> Dict[str, Any]:
     }
 
 
+def get_fee(config: Dict[str, Any], request: Dict[str, Any]) -> uint64:
+    fee = request.get("fee")
+    if fee is None:
+        config_fee = config.get("fee", 0)
+        return uint64(config_fee)
```
altendky, line 44: This doesn't cover the case where the config has `None` for example from a line like `fee:` with no value I believe.

https://github.com/Chia-Network/chia-blockchain/pull/10042#discussion_r798788659

altendky (approved):
After a very quick scan I saw one minor hinting error and a spot where we don't handle a `None` from the config.  Maybe you could follow up in a separate PR for those.

PR #9949, opened 2022-01-25
Improve StructStream hints to support decimal and other parameters like int itself

altendky:
This is an effort to make the hints for `StructStream`, and consequently `uint32` etc, more accurate to their actual runtime behavior.  Any runtime behavior change is unintended and would be a bug in this PR.


tests/util/test_struct_stream.py
```diff
@@ -106,3 +107,15 @@ def roundtrip(v):
 
         roundtrip(int8(0x7F))
         roundtrip(int8(-0x80))
+
+    def test_uint32_from_decimal(self) -> None:
+        assert uint32(Decimal("137")) == 137
+
+    def test_uint32_from_float(self) -> None:
+        assert uint32(4.0) == 4
```
arvidn, line 115: will it still throw an exception if you try to pass an a value with actual decimals in it? e.g. `4.4`
I hope so, and it would be nice if the test would demonstrate that

altendky: This doesn't change runtime behavior.  `int()` doesn't complain.  This just uses `int()` as the first step so it also doesn't complain.  The hints shouldn't complain either.  `:]`

I'm happy to add tests showing both `float` and `decimal.Decimal` behavior for non-integer values.

arvidn: I guess I'm not quite familiar with what this patch is doing. but before this patch, is there no error if I try to create a `uint32` from a decimal number? the value can't be represented by a `uint32` so I would expect it to throw (just like it throws for negative values for example).

altendky: This is a hint-only change (plus tests to exercise the hints and show the behavior of these cases).  I got into this because of https://github.com/Chia-Network/chia-blockchain/pull/9934#discussion_r791310788 (which doesn't expand, thanks GitHub) where I complained in the review about `final_amount = uint64(int(amount * units["chia"]))` because the `int()` there is pointless (at runtime).  `StructStream.__new__()` already applies `int()` as the first step.  Note that `amount = Decimal(args["amount"])` was tripping up mypy because it resulted in the argument to `uint64()` being a `Decimal` while `StructStream.__new__()` is only hinted to accept `int`.

To your concerns, since the first line of `.__new__()` is `value = int(value)` and `int()` happily converts a non-integral float to an integer, after that we've lost the incoming `value` so we can't do anything to check it.  The non-integral nature of an input like `1.7` is already lost.  The only thing we can check is if the value that comes out of `int()` is something we don't like.  Of course, this behavior could be changed.  I'm just trying to hint for what it is.  If it should change, then sure, perhaps this PR shouldn't go through.  `:]`

```python-console
Python 3.9.5 (default, Jun  3 2021, 15:18:23) 
[GCC 9.3.0] on linux
>>> from chia.util.ints import uint32
>>> uint32(1)
1
>>> uint32(1.1)
1
>>> uint32(-1)
  File "/home/altendky/repos/chia-blockchain/chia/util/struct_stream.py", line 16, in __new__
    v1 = struct.unpack(cls.PACK, struct.pack(cls.PACK, value))[0]
struct.error: argument out of range

During handling of the above exception, another exception occurred:

  File "/home/altendky/repos/chia-blockchain/chia/util/struct_stream.py", line 21, in __new__
    raise ValueError(
ValueError: Value -1 of size 1 does not fit into uint32 of size 32
```

arvidn (approved):
This looks right. I was a bit surprised about accepting anything that's convertible to an `int`, but that is our current behavior. In the future, we may want to look into making this stricter. We already check sign (for unsigned) and size, to ensure we never lose data in the serialization. It would make sense to extend these checks to decimals too.

PR #9951, opened 2022-01-25
default new installs to use v2 blockchain database schema

arvidn:
Leave an option to initialize a v1 database


arvidn:
the filename of the database file currently have the version number in it. The option in this PR to create a v1 database doesn't attempt to change the filename in the config file. Should it?

aqk:
Yes, if they use `--v1-db`, it's reasonable to assume they want to set that DB as the active one in the config.
If you do add that, make sure to add a print statement like "setting DB version v1 in config"

arvidn:
@aqk well, I made that change, but I'm not printing any messages about it

wjblanke (approved):
aok

jack60612 (approved):
YAY

PR #9957, opened 2022-01-26
introduce restrictions on generators at a specific height

arvidn:
disallow division on negative numbers and disallow redundant leading zeros on integer condition arguments (produced by a generator).

I have successfully synced from scratch with the post-softfork rules enabled.

### DIFF
--- a/chia/consensus/block_creation.py
+++ b/chia/consensus/block_creation.py
@@ -136,6 +136,7 @@ def create_foliage(
                 constants.MAX_BLOCK_COST_CLVM,
                 cost_per_byte=constants.COST_PER_BYTE,
                 mempool_mode=True,
+                height=height,
             )
             cost = calculate_cost_of_program(block_generator.program, result, constants.COST_PER_BYTE)
 
--- a/chia/consensus/block_creation.py
+++ b/chia/consensus/block_creation.py
@@ -136,7 +136,6 @@ def create_foliage(
                 constants.MAX_BLOCK_COST_CLVM,
                 cost_per_byte=constants.COST_PER_BYTE,
                 mempool_mode=True,
-                height=height,
             )
             cost = calculate_cost_of_program(block_generator.program, result, constants.COST_PER_BYTE)
 
--- a/chia/consensus/default_constants.py
+++ b/chia/consensus/default_constants.py
@@ -54,6 +54,7 @@ testnet_kwargs = {
     "MAX_GENERATOR_SIZE": 1000000,
     "MAX_GENERATOR_REF_LIST_SIZE": 512,  # Number of references allowed in the block generator ref list
     "POOL_SUB_SLOT_ITERS": 37600000000,  # iters limit * NUM_SPS
+    "SOFT_FORK_HEIGHT": 2000000,
 }
 
 
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -1,10 +1,11 @@
 import logging
 import time
 from typing import Dict, List, Optional
-from clvm_rs import STRICT_MODE as MEMPOOL_MODE
+from clvm_rs import MEMPOOL_MODE, COND_CANON_INTS, NO_NEG_DIV
 
 from clvm.casts import int_from_bytes, int_to_bytes
 from chia.consensus.cost_calculator import NPCResult
+from chia.consensus.default_constants import DEFAULT_CONSTANTS
 from chia.full_node.generator import create_generator_args, setup_generator_args
 from chia.types.blockchain_format.program import NIL
 from chia.types.coin_record import CoinRecord
@@ -107,8 +108,13 @@ def add_cond(
     conds[op].append(ConditionWithArgs(op, args))
 
 
+def unwrap(x: Optional[uint32]) -> uint32:
+    assert x is not None
+    return x
+
+
 def get_name_puzzle_conditions(
-    generator: BlockGenerator, max_cost: int, *, cost_per_byte: int, mempool_mode: bool
+    generator: BlockGenerator, max_cost: int, *, cost_per_byte: int, mempool_mode: bool, height: Optional[uint32] = None
 ) -> NPCResult:
     block_program, block_program_args = setup_generator_args(generator)
     size_cost = len(bytes(generator.program)) * cost_per_byte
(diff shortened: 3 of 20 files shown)


chia/types/coin_spend.py
```diff
@@ -39,7 +40,11 @@ def hints(self) -> List[bytes]:
         generator = simple_solution_generator(bundle)
 
         npc_result = get_name_puzzle_conditions(
-            generator, INFINITE_COST, cost_per_byte=DEFAULT_CONSTANTS.COST_PER_BYTE, mempool_mode=False
+            generator,
+            INFINITE_COST,
+            cost_per_byte=DEFAULT_CONSTANTS.COST_PER_BYTE,
+            mempool_mode=False,
+            height=uint32(2000000),
```
arvidn, line 47: the `hints()` function is only called from the wallet, on our own coins, so this should be safe. I have another outstanding PR that moves this function into the wallet to make this clear

chia/consensus/default_constants.py
```diff
@@ -54,6 +54,7 @@
     "MAX_GENERATOR_SIZE": 1000000,
     "MAX_GENERATOR_REF_LIST_SIZE": 512,  # Number of references allowed in the block generator ref list
     "POOL_SUB_SLOT_ITERS": 37600000000,  # iters limit * NUM_SPS
+    "SOFT_FORK_HEIGHT": 2000000,
```
richardkiss, line 57: This is not a very forward-looking name. Maybe `SOFT_FORK_1_HEIGHT`?

arvidn: I'm hoping to delete this line in a few months.. thinking that we won't ever schedule more than one at a time

chia/full_node/mempool_check_conditions.py
```diff
@@ -107,16 +108,38 @@ def add_cond(
     conds[op].append(ConditionWithArgs(op, args))
 
 
+def unwrap(x: Optional[uint32]) -> uint32:
+    assert x is not None
```
richardkiss, line 112: `assert` can get optimized out if you go `python -O` (what a stupid idea) so you should consider `if x is None: raise ValueError`

arvidn: yeah, this was the best I could think of to satisfy mypy. the value is known to not be `None` at this point

chia/full_node/mempool_check_conditions.py
```diff
@@ -107,16 +108,38 @@ def add_cond(
...
+    return x
+
+
 def get_name_puzzle_conditions(
-    generator: BlockGenerator, max_cost: int, *, cost_per_byte: int, mempool_mode: bool
+    generator: BlockGenerator, max_cost: int, *, cost_per_byte: int, mempool_mode: bool, height: Optional[uint32] = None
```
xdustinface, line 117:
```suggestion
    generator: BlockGenerator, max_cost: int, *, cost_per_byte: int, mempool_mode: bool, height: uint32 = uint32(0)
```

Any reason to not just drop the `Optional` and default to `0`? I guess we can as well assume a zero height is invalid here. This way you would assert `> 0` and you could drop the `unwrap`.

chia/full_node/mempool_check_conditions.py
```diff
@@ -107,16 +108,38 @@ def add_cond(
...
     block_program, block_program_args = setup_generator_args(generator)
     size_cost = len(bytes(generator.program)) * cost_per_byte
     max_cost -= size_cost
     if max_cost < 0:
         return NPCResult(uint16(Err.INVALID_BLOCK_COST.value), [], uint64(0))
 
-    flags = MEMPOOL_MODE if mempool_mode else 0
+    # in mempool mode, the height doesn't matter, because it's always strict.
+    # But otherwise, height must be specified to know which rules to apply
+    assert mempool_mode or height is not None
```
xdustinface, line 127:
```suggestion
    assert mempool_mode or height > 0
```

chia/full_node/mempool_check_conditions.py
```diff
@@ -107,16 +108,38 @@ def add_cond(
...
+
+    # mempool mode also has these rules apply
+    assert (MEMPOOL_MODE & COND_CANON_INTS) != 0
+    assert (MEMPOOL_MODE & NO_NEG_DIV) != 0
+
+    if mempool_mode:
+        flags = MEMPOOL_MODE
+    elif unwrap(height) >= DEFAULT_CONSTANTS.SOFT_FORK_HEIGHT:
```
xdustinface, line 135:
```suggestion
    elif height >= DEFAULT_CONSTANTS.SOFT_FORK_HEIGHT:
```

chia/full_node/mempool_check_conditions.py
```diff
@@ -107,16 +108,38 @@ def add_cond(
     conds[op].append(ConditionWithArgs(op, args))
 
 
+def unwrap(x: Optional[uint32]) -> uint32:
+    assert x is not None
+    return x
+
+
```
xdustinface, line 115:
```suggestion
```

chia/wallet/util/compute_hints.py
```diff
@@ -8,6 +8,7 @@
 from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
 from chia.types.spend_bundle import SpendBundle
+from chia.util.ints import uint32
```
xdustinface, line 11:
```suggestion
```

chia/consensus/block_creation.py
```diff
@@ -136,6 +136,7 @@ def create_foliage(
                 constants.MAX_BLOCK_COST_CLVM,
                 cost_per_byte=constants.COST_PER_BYTE,
                 mempool_mode=True,
+                height=height,
```
xdustinface, line 139:
```suggestion
```

It's mempool mode here.

xdustinface (changes requested):

chia/full_node/mempool_check_conditions.py
arvidn: the reason is that I don't want it to be possible to call this function, not in mempool mode and also not specifying the height. 0 *is* a valid height, but it would work to default it to -1. However, that just seems like a poor-mans's optional.

tests/core/full_node/test_mempool.py
```diff
@@ -1790,11 +1791,31 @@ def test_invalid_condition_args_terminator(self):
...
+            (False, 2000000, -1, Err.GENERATOR_RUNTIME_ERROR.value),
+            (True, 0, 1, None),
+            (False, 1000000, 1, None),
+            (False, 2000000, 1, None),
+        ],
+    )
+    def test_div(self, mempool, height, operand, expected):
+
+        # op_div is disallowed on negative numbers in the mempool, and after the
+        # softfork#
```
xdustinface, line 1808:
```suggestion
        # softfork
```

xdustinface (dismissed):
Looks good. There is one more nit in a comment you could consider to change though.

xdustinface (approved):

PR #9962, opened 2022-01-26
simplify run_block()

arvidn:
we run the generator multiple times currently. once to list all the spent coins, then one more time for each spent coin, to extract the puzzle and solution.

This patch runs the generator once, and picks out all information we need directly.

### DIFF
--- a/tests/tools/test_run_block.py
+++ b/tests/tools/test_run_block.py
@@ -1,6 +1,6 @@
 import json
 from pathlib import Path
-from typing import List, Tuple
+from typing import List
 
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
 from chia.types.condition_opcodes import ConditionOpcode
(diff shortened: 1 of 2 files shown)


tests/tools/test_run_block.py
```diff
@@ -63,9 +63,14 @@ def test_block_retired_cat_with_memo():
...
     assert cat_list[0].tail_hash == "86bf9abe0600edf96b2e0fa928d19435b5aa756a9c9151c4b53c2c3da258502f"
     assert cat_list[0].memo == "Hello, please find me, I'm a memo!"
-    first_condition: Tuple[ConditionOpcode, List[ConditionWithArgs]] = cat_list[0].npc.conditions[0]
-    assert first_condition[0] == ConditionOpcode.CREATE_COIN
-    assert find_retirement(first_condition[1])
+    assert cat_list[0].npc.coin_name.hex() == "244854a6fadf837b0fbb78d19b94b0de24fd2ffb440e7c0ec7866104b2aecd16"
+    assert cat_list[0].npc.puzzle_hash.hex() == "4aa945b657928602e59d37ad165ba12008d1dbee3a7be06c9bd19b4f00da456c"
+    found = False
+    for cond in cat_list[0].npc.conditions:
+        if cond[0] != ConditionOpcode.CREATE_COIN:
```
arvidn, line 70: the order of the conditions is not defined

emlowe (approved):
Looks good to me - thanks for the improvment

aqk (approved):

PR #9963, opened 2022-01-26
Use native concurrency settings to cancel duplicate in progress jobs

cmmarslender:
Rather than a job needing to start to run the action, this will cancel duplicate older jobs natively/immediately, without needing any runner to be available.


hoffmang9 (approved):
Cool!

justinengland (approved):

altendky:
Thanks for the research.  Thanks for the change.  Thanks for the lesson...  `:]`

PR #9968, opened 2022-01-26
This patch enables fees for the plotnft commands - create, claim, joi…

aqk:
…n, and leave.


chia/pools/pool_wallet.py
```diff
@@ -770,12 +785,34 @@ async def claim_pool_rewards(self, fee: uint64) -> TransactionRecord:
...
+            my_address = await self.standard_wallet.get_new_puzzlehash()
+            fee_spend = await self.standard_wallet.generate_signed_transaction(
+                amount,
+                my_address,
+                fee,
+                None,
+                None,
+                None,
+                False,
+                {Announcement(the_coin_record_in_question.coin.name(), b"$").name()},
```
arvidn, line 809: is there an implicit `RESERVE_FEE` condition in here? we need that, right?

chia/pools/pool_wallet.py
```diff
@@ -448,14 +449,24 @@ async def create_new_pool_wallet_transaction(
...
+                return self.standard_wallet.secret_key_store.secret_key_for_public_key(pk)
             return owner_sk
 
+        spends = [coin_spend]
+        if fee > 0:
+            # TODO: test failure UX when fee is higher than available balance
+            get_max_send_amount = await self.standard_wallet.get_max_send_amount()
+            assert get_max_send_amount > fee
+            my_address = await self.standard_wallet.get_new_puzzlehash()
+            fee_spends = await self.standard_wallet._generate_unsigned_transaction(uint64(0), my_address, fee)
```
Quexington, line 466: We need to assert an announcement when creating fee spends so that a malicious farmer can't separate the fee from the thing being paid for.  I'm not sure why a function called `sign` should have fee logic inside of it?  Shouldn't you just be passing spends to this function and having it sign with the PKs and msgs it detects?

chia/pools/pool_wallet.py
richardkiss: It's optional, but probably a good idea (although I suppose you could argue it increases the cost slightly).

chia/pools/pool_wallet.py
```diff
@@ -448,14 +449,24 @@ async def create_new_pool_wallet_transaction(
         )
         return standard_wallet_record, p2_singleton_puzzle_hash, launcher_coin_id
 
-    async def sign(self, coin_spend: CoinSpend) -> SpendBundle:
+    async def sign(self, coin_spend: CoinSpend, fee: uint64) -> SpendBundle:
```
richardkiss, line 452: This makes no sense to me that a function called `sign` also generates an unsigned transaction. The fee should be included in the initial unsigned transaction.

chia/pools/pool_wallet.py
```diff
@@ -784,15 +821,15 @@ async def claim_pool_rewards(self, fee: uint64) -> TransactionRecord:
...
+            removals=full_spend.removals(),
             wallet_id=uint32(self.wallet_id),
             sent_to=[],
             memos=[],
             trade_id=None,
             type=uint32(TransactionType.OUTGOING_TX.value),
-            name=spend_bundle.name(),
+            name=full_spend.name(),
         )
         await self.wallet_state_manager.add_pending_transaction(absorb_transaction)
```
Quexington, line 834: You should be pushing both transaction records otherwise the balance will just disappear from the wallet without a trace.  You can use `dataclasses.replace` to set the spend bundle of the fee tx to `None` so that when you push it, it doesn't push a duplicate spend.

Also, out of the scope of this PR but `generate_signed_transaction` in the other wallets doesn't push transactions for you.  It only generates the transaction and then the RPC pushes it or whatever.  What this should probably return is a `List[TransactionRecord]` containing both the pool tx and the fee tx and leave it to the calling function to push them for you.  But that's for another day.

chia/wallet/wallet.py
```diff
@@ -307,8 +307,8 @@ async def _generate_unsigned_transaction(
         coins: Set[Coin] = None,
         primaries_input: Optional[List[AmountWithPuzzlehash]] = None,
         ignore_max_send_amount: bool = False,
-        coin_announcements_to_consume: Set[Announcement] = None,
-        puzzle_announcements_to_consume: Set[Announcement] = None,
+        coin_announcements_to_consume_as_bytes: Optional[Set[bytes32]] = None,
+        puzzle_announcements_to_consume_as_bytes: Optional[Set[bytes32]] = None,
```
Quexington, line 311: Pretty sure this is a bad merge, we switched to having these arguments be Announcements when we merged in the wallet changes.  There's lots of other code that depends on this.

Quexington: And all of the other changes in this file too.

chia/cmds/plotnft.py
```diff
@@ -121,13 +121,6 @@ def create_cmd(
     required=True,
     callback=validate_fee,
 )
-@click.option(
-    "--fee",
-    help="Fee Per Transaction, in Mojos. Fee is used TWICE: once to leave pool, once to join.",
-    type=int,
-    callback=validate_fee,
-    default=0,
-)
```
Quexington, line 130: This diff suggests we're removing the fee parameter rather than adding it.  That's not right right?

chia/cmds/plotnft_funcs.py
```diff
@@ -120,7 +119,10 @@ async def create(args: dict, wallet_client: WalletRpcClient, fingerprint: int) -
                     print(f"Do chia wallet get_transaction -f {fingerprint} -tx 0x{tx_record.name} to get status")
                     return None
         except Exception as e:
-            print(f"Error creating plot NFT: {e}")
+            print(
+                f"Error creating plot NFT: {e}\n  Please ensure both farmer and wallet are running:"
+                f"    chia start -r wallet farmer-only"
```
Quexington, line 124: I have a pending PR out to change wallet to only start the wallet and not the node.  I think just recommending `chia start -r farmer` here is better because it will start everything the farmer needs.

chia/pools/pool_wallet.py
Quexington: I think the wallet as part of `generate_signed_transaction` will add the RESERVE_FEE condition

chia/cmds/plotnft_funcs.py
aqk: That's a good idea, but doesn't this code use the wallet RPC API above at this line?

`tx = await wallet_client.get_transaction(str(1), tx_record.name)`

chia/wallet/wallet.py
aqk: I see - so all calls into Wallet.generate_signed_transaction have been changed to use Announcements?

chia/cmds/plotnft.py
aqk: Good eye! This removes a duplicate flag declaration that snuck into main earlier, so this is okay.

chia/pools/pool_wallet.py
aqk: That is an excellent point. The code that creates this CoinSpend is in the driver code `chia/pools/pool_puzzles.py:create_travel_spend` - could you or Quex add a fee parameter here, please?

chia/pools/pool_wallet.py
aqk: Yes, this is exactly the main reason I wanted you and Richard to review this - the code that follows attempts to join the two spends. But Richard points out the same improvement - it would be better if the fee were includedd in the initial, unsigned spend.

The code that creates this CoinSpend is in the driver code `chia/pools/pool_puzzles.py:create_travel_spend` - could Quex or Richard add a fee parameter here, please?

chia/pools/pool_wallet.py
aqk: Interesting. So you're saying that `self.wallet_state_manager.add_pending_transaction` doesn't take into consideration that the `spend_bundle` parameter contains spends concerning two different wallets, and I should make two TransactionRecords, one for `fee_spend`, and one for the claim spend?

But if I do that, doesn't the PoolWallet need the tx id of the combinded transaction, full_spend?

Ahhh, no, you are saying that I should submit a TransactionRecord for the fee_spend only, for the main wallet, and a combined tx record for the full transaction, but without the fee, for the PoolWallet, so the balances are computed correctly. Is that correct?

Quexington: Right now you have this `fee_spend` variable above and you just take its spend bundle and aggregate it with the absorb spend to create a new `full_spend` which you attach to the absorb transaction record.  This is all well and good, but you want to push the transaction record that the standard wallet generated for the fee so it correctly shows up in that wallet's history.  The problem is that you've already aggregated the spend bundle from the fee tx into the absorb tx (because you had to) and so if you try to push the fee tx it will fail since it conflicts with `full_spend` which has already been pushed.  For this reason, you need to set `spend_bundle=None` on the fee tx before you push it.

chia/wallet/wallet.py
Quexington: Correct

chia/cmds/plotnft_funcs.py
Quexington: It does but `chia start farmer` starts the wallet too https://github.com/Chia-Network/chia-blockchain/blob/00613bc259584c7f9c06c3f7aca2a52e1da72b2c/chia/util/service_groups.py#L7

chia/pools/pool_wallet.py
```diff
@@ -778,29 +794,54 @@ async def claim_pool_rewards(self, fee: uint64) -> TransactionRecord:
             self.log.info(
                 f"Farmer coin: {coin_record.coin} {coin_record.coin.name()} {coin_to_height_farmed[coin_record.coin]}"
             )
-        if len(all_spends) == 0:
+        if len(all_spends) == 0 or the_coin_record_in_question is None:
```
jack60612, line 797: Would you be ok with renaming the_coin_record_in_question to current_coin_record? it just seems cleaner

aqk: Good idea. Done.

chia/cmds/plotnft_funcs.py
aqk: Good idea

chia/pools/pool_wallet.py
aqk: @Quexington  Can you take a look at the current updated code?

chia/pools/pool_wallet.py
Quexington: You can use `generate_signed_transaction` for this fee spend can't you?  Then you can remove this fee parameter from this function that really shouldn't know anything about fees or creating transactions.  Once both the spend and the fee spend are signed, you can aggregate the bundle together and also importantly push both tx records rather than just forgetting about the fee one that you generated in here.

chia/pools/pool_wallet.py
```diff
@@ -778,29 +798,54 @@ async def claim_pool_rewards(self, fee: uint64) -> TransactionRecord:
...
+        current_time = uint64(int(time.time()))
+        # The claim spend, minus the fee amount from the main wallet
         absorb_transaction: TransactionRecord = TransactionRecord(
             confirmed_at_height=uint32(0),
-            created_at_time=uint64(int(time.time())),
+            created_at_time=current_time,
             to_puzzle_hash=current_state.current.target_puzzle_hash,
             amount=uint64(total_amount),
-            fee_amount=fee,
+            fee_amount=uint64(0),  # Notice that fee is zero here - is is accounted in self.standard_wallet
```
Quexington, line 837: You don't have to make this change.  The TX records are mostly just for reporting to the GUI what is true about the transaction.  It is going to be true that this transaction went through with that fee just as it is true that the fee transaction went through with that fee.  You don't need to worry about them adding up in any way.

chia/pools/pool_wallet.py
```diff
@@ -778,29 +798,54 @@ async def claim_pool_rewards(self, fee: uint64) -> TransactionRecord:
...
+                None,
+                None,
+                None,
+                False,
+                {Announcement(current_coin_record.coin.name(), b"$")},
+            )
+
+            full_spend = SpendBundle.aggregate([fee_tx.spend_bundle, claim_spend])
+            # The fee spend only, from the main wallet
+            await self.wallet_state_manager.add_pending_transaction(dataclasses.replace(fee_tx, spend_bundle=None))
```
Quexington, line 827: Don't add this pending transaction until you add the other one.  Something could go wrong in the logic below and now you've just added a pending transaction that will never confirm.

tests/pools/test_pool_rpc.py
```diff
@@ -448,7 +448,7 @@ def wallet_is_synced():
 
     @pytest.mark.asyncio
     @pytest.mark.parametrize("trusted", [True])
-    @pytest.mark.parametrize("fee", [0])
+    @pytest.mark.parametrize("fee", [0, FEE_AMOUNT])
```
Quexington, line 451: I'm not positive that we need the parameterization on the fee amount.  It seems like most fee logic lives under an `if fee > 0` statement which means we're only ever testing a superset of logic when we use a fee.  I don't care that much but it would cut down on test runtime if we were to remove this.

Quexington (changes requested):

chia/pools/pool_wallet.py
```diff
@@ -471,22 +471,32 @@ async def pk_to_sk(pk: G1Element) -> PrivateKey:
...
+            coin_announcements_to_consume=coin_announcements,
+        )
+        return fee_tx
+
+    async def publish_transactions(self, travel_tx: TransactionRecord, fee_tx: Optional[TransactionRecord]):
+        await self.wallet_state_manager.add_pending_transaction(travel_tx)
+        if fee_tx is not None:
+            await self.wallet_state_manager.add_pending_transaction(dataclasses.replace(fee_tx, spend_bundle=None))
+
+    async def generate_travel_transactions(self, fee: uint64) -> Tuple[TransactionRecord, Optional[TransactionRecord]]:
```
aqk, line 499: Originally, I wrote this to return `List[TransactionRecord]`, but I changed it so that we could keep the output of the RPCs backwards compatible.

chia/pools/pool_wallet.py
```diff
@@ -847,8 +840,16 @@ async def claim_pool_rewards(self, fee: uint64) -> TransactionRecord:
...
             name=full_spend.name(),
         )
-        await self.wallet_state_manager.add_pending_transaction(absorb_transaction)
-        return absorb_transaction
+
+        fee_tx = None
+        if fee > 0:
+            absorb_announce = Announcement(current_coin_record.coin.name(), b"$")
+            fee_tx = await self.generate_fee_transaction(fee, coin_announcements=[absorb_announce])
+            full_spend = SpendBundle.aggregate([fee_tx.spend_bundle, claim_spend])
```
Quexington, line 848: It looks like we don't do anything with full_spend here.  I think we need it to be pushed on the absorb_transaction.

Quexington: If I'm correct and the tests still pass we should add some `get_confirmed_balance` checks to make sure our wallet balance drops by the appropriate fee amount.

Quexington (changes requested):

chia/pools/pool_wallet.py
aqk: Nice, thanks

tests/pools/test_pool_rpc.py
aqk: I understand it's that way now, but if this code changes in the future, we want to test both paths.

Our test run-times do suck, but that's a separate issue.

chia/pools/pool_wallet.py
aqk: Cool, I'll update that test

Quexington (dismissed):

PR #9974, opened 2022-01-26
Sync improvements

Yostra:
Make max item count that wallets can subscribe to configurable
Don't disconnect from remote peers until local node is running & synced
Fix the endless sync loop
Add node id to chia show command
Simplify wallet sync. untrusted & trusted mode now use almost the same code path for adding new coin state
Speed up untrusted sync.


tests/wallet/test_wallet.py
```diff
@@ -728,7 +734,7 @@ async def test_wallet_tx_reorg(self, two_wallet_nodes, trusted):
 
     @pytest.mark.parametrize(
         "trusted",
-        [True, False],
+        [False],
```
altendky, line 737: Should we really not be testing this trusted?

altendky (commented):
This is just a quick Python-level review.  I did not check the functionality much at all.  Please coordinate any work on these.  I may try to address some of them.

tests/wallet/test_wallet.py
altendky: https://github.com/Chia-Network/chia-blockchain/pull/10081

PR #9975, opened 2022-01-26
Coin Selection Refactor With CAT Coin Selection Refactor 

jack60612:
- [x]  write new coin selection algorithms 
- [x] decompose coin_select function
- [x] Test and use new function with wallet and cat wallet.
- [x] Add integration test


tests/wallet/cat_wallet/test_cat_wallet.py
```diff
@@ -743,6 +743,105 @@ async def check_wallets(wallet_node):
...
+        else:
+            wallet_node.config["trusted_peers"] = {}
+            wallet_node_2.config["trusted_peers"] = {}
+
+        await server_2.start_client(PeerInfo("localhost", uint16(full_node_server._port)), None)
+        await server_3.start_client(PeerInfo("localhost", uint16(full_node_server._port)), None)
+
+        for i in range(0, num_blocks):
+            await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph))
+
```
Quexington, line 773: I have a PR out to change this boilerplate, it's actually broken right now, if you could update this to work the way it works in that branch that'd be great: https://github.com/Chia-Network/chia-blockchain/pull/9976

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,145 @@
+import random
+from typing import Set, Optional, List, Dict
+
+from chia.types.blockchain_format.coin import Coin
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.ints import uint64
+from chia.wallet.wallet_coin_record import WalletCoinRecord
+
+
+async def select_coins(wallet, amount: uint64, exclude: List[Coin] = None) -> Set[Coin]:
```
mariano54, line 10: It's not good practice to pass the larger higher level object here (wallet).

tests/wallet/cat_wallet/test_cat_wallet.py
```diff
@@ -743,6 +743,105 @@ async def check_wallets(wallet_node):
...
+            await wallet.wallet_state_manager.add_pending_transaction(tx_record)
+            if tx_record.spend_bundle is not None:
+                await time_out_assert(
+                    15, tx_in_pool, True, full_node_api.full_node.mempool_manager, tx_record.spend_bundle.name()
+                )
+
+        for i in range(1, num_blocks):
+            await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph))
+
+        # check that the coin selection only spends the identical coin.
```
Quexington, line 807: Isn't the only coin in cat_wallet_2 a coin of 40000?  The old algorithm would still have used it this way right?

chia/wallet/coin_selection.py
mariano54: Instead, we can pass in the required data:
- spendable cons
- constants
- unconfirmed removals

tests/wallet/test_wallet.py
```diff
@@ -775,3 +775,81 @@ async def test_address_sliding_window(self, wallet_node_100_pk, trusted):
...
+            wallet_node_2.config["trusted_peers"] = {server_1.node_id: server_1.node_id}
+        else:
+            wallet_node.config["trusted_peers"] = {}
+            wallet_node_2.config["trusted_peers"] = {}
+
+        await server_2.start_client(PeerInfo(self_hostname, uint16(server_1._port)), None)
+        await server_3.start_client(PeerInfo(self_hostname, uint16(server_1._port)), None)
+
+        for i in range(0, num_blocks):
+            await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph))
```
Quexington, line 806: Same as above, this boilerplate is actually broken and won't test trusted syncing.

mariano54 (changes requested):

tests/wallet/cat_wallet/test_cat_wallet.py
```diff
@@ -743,6 +743,105 @@ async def check_wallets(wallet_node):
         await time_out_assert(15, cat_wallet.get_confirmed_balance, 35)
         await time_out_assert(15, cat_wallet.get_unconfirmed_balance, 35)
 
+    @pytest.mark.parametrize(
+        "trusted",
+        [True, False],
+    )
+    @pytest.mark.asyncio
+    async def test_cat_coin_selection(self, two_wallet_nodes, trusted):
```
Quexington, line 751: I'm dubious that we need this test since we've changed the CAT select_coins algorithm to use the shared `select_coins` method.  Is there a reason why we should specifically test this for CATs?

jack60612: Nope, ill just remove it.

jack60612:
Closing and opening for CI

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -405,46 +406,25 @@ async def select_coins(self, amount: uint64) -> Set[Coin]:
...
-        self.log.info(f"About to select coins for amount {amount}")
-        spendable: List[WalletCoinRecord] = await self.get_cat_spendable_coins()
-
-        sum = 0
-        used_coins: Set = set()
-
-        # Use older coins first
-        spendable.sort(key=lambda r: r.confirmed_block_height)
-
+        spendable_amount = await self.get_spendable_balance()
```
mariano54, line 409: type annotation: uint64

mariano54 (dismissed):
LGTM when tests pass

mariano54 (dismissed):

mariano54:
Reopened to trigger test runs

jack60612:
ci

mariano54 (dismissed):

Quexington (dismissed):

chia/wallet/wallet.py
```diff
@@ -1,6 +1,6 @@
 import logging
 import time
-from typing import Any, Dict, List, Optional, Set
+from typing import Any, List, Optional, Set, Dict
```
altendky, line 3: isort is going to complain about this now.

```suggestion
from typing import Any, Dict, List, Optional, Set
```

tests/wallet/test_coin_selection.py
```diff
@@ -0,0 +1,61 @@
+from typing import Set
+
+import pytest
+from chia.types.blockchain_format.coin import Coin
+from chia.util.hash import std_hash
+from chia.util.ints import uint64, uint128
+from chia.wallet.coin_selection import check_for_exact_match, find_smallest_coin, knapsack_coin_algorithm
+from chia.consensus.default_constants import DEFAULT_CONSTANTS
```
altendky, line 8: isort again

```suggestion
import pytest

from chia.consensus.default_constants import DEFAULT_CONSTANTS
from chia.types.blockchain_format.coin import Coin
from chia.util.hash import std_hash
from chia.util.ints import uint64, uint128
from chia.wallet.coin_selection import check_for_exact_match, find_smallest_coin, knapsack_coin_algorithm
```

ftruzzi:
Hi! Have you thought about exposing coin selection in the RPC API? I've opened a PR for it, and I can rebase it on this branch if needed.

https://github.com/Chia-Network/chia-blockchain/pull/10495

Thanks

jack60612:
> Hi! Have you thought about exposing coin selection in the RPC API? I've opened a PR for it, and I can rebase it on this branch if needed.
> 
> 
> 
> https://github.com/Chia-Network/chia-blockchain/pull/10495
> 
> 
> 
> Thanks
> 
> 
> 
> 

I would wait until the pr is merged but, feel free.

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,138 @@
...
+        exclude = []
+
+    if amount > spendable_amount:
+        error_msg = (
+            f"Can't select amount higher than our spendable balance.  Amount: {amount}, spendable: {spendable_amount}"
+        )
+        log.warning(error_msg)
+        raise ValueError(error_msg)
+
+    log.info(f"About to select coins for amount {amount}")
```
emlowe, line 33: Maybe these and others should be Debug instead of Info?

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,138 @@
...
+        log.info(f"Selected coins from knapsack algorithm: {best_coin_set}")
+        return best_coin_set
+
+
+# These algorithms were based off of the algorithms in:
+# https://murch.one/wp-content/uploads/2016/11/erhardt2016coinselection.pdf
+
+# we use this to check if one of the coins exactly matches the target.
+def check_for_exact_match(coin_list: List[Coin], target: uint64) -> Optional[Coin]:
+    for coin in coin_list:
```
emlowe, line 90: The list is sorted, so we could do better than iterating the list if we wanted more performance (binary search anyone - unfortunately since sorted descending, I don't think we can use bisect_left directly). This list of spendable coins could be quite large.

emlowe:
I wonder about adding in support for an "ignore dust" feature - so add in the ability to ignore coins below a certain threshold?

chia/wallet/wallet.py
```diff
@@ -241,59 +242,34 @@ def add_condition_to_solution(self, condition: Program, solution: Program) -> Pr
...
+            spendable_amount,
+            self.wallet_state_manager.constants.MAX_COIN_AMOUNT,
+            spendable_coins,
+            unconfirmed_removals,
+            self.log,
+            uint128(target),
+            exclude,
+        )
+        assert coins is not None and len(coins) > 0
+        assert sum(c.amount for c in coins) >= target
```
emlowe, line 271: why are these two asserts here. If these are real errors please Raise something, otherwise just remove them, or if debugging, add debug log entries.
The way I read it is never of these can happen, and if they do it's a serious bug probably worth of a raise of some kind

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,138 @@
...
+    for coin in valid_spendable_coins:
+        if coin.amount < amount:
+            smaller_coin_sum += coin.amount
+            smaller_coins.add(coin)
+    if smaller_coin_sum == amount:
+        log.info(f"Selected all smaller coins because they equate to an exact match of the target.: {smaller_coins}")
+        return smaller_coins
+    elif smaller_coin_sum < amount:
+        smallest_coin: Optional[Coin] = find_smallest_coin(valid_spendable_coins, amount, max_coin_amount)
+        assert smallest_coin is not None
```
emlowe, line 75: This can't happen, so I think this should raise an Exception since it would be a programming error

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,138 @@
...
+    smaller_coins: Set[Coin] = set()
+    for coin in valid_spendable_coins:
+        if coin.amount < amount:
+            smaller_coin_sum += coin.amount
+            smaller_coins.add(coin)
+    if smaller_coin_sum == amount:
+        log.info(f"Selected all smaller coins because they equate to an exact match of the target.: {smaller_coins}")
+        return smaller_coins
+    elif smaller_coin_sum < amount:
+        smallest_coin: Optional[Coin] = find_smallest_coin(valid_spendable_coins, amount, max_coin_amount)
```
emlowe, line 74: You could slice valid_spendable_coins by removing all coins too small since they can't possibly be selected - which I think maybe `valid_spendable_coins[:len(smaller_coins)]` would work.

emlowe: In which case, max_coin_amount would actually be valid_spendable_coins[0].amount, since that is the largest spendable coin and by definition must be > amount. So I think you can rewrite this to not return Optional

tests/wallet/test_wallet.py
```diff
@@ -830,3 +830,90 @@ async def test_wallet_transaction_options(self, two_wallet_nodes, trusted):
 
         await time_out_assert(5, wallet.get_confirmed_balance, funds - AMOUNT_TO_SEND)
         await time_out_assert(5, wallet.get_unconfirmed_balance, funds - AMOUNT_TO_SEND)
+
+    @pytest.mark.parametrize(
+        "trusted",
+        [True, False],
+    )
+    @pytest.mark.asyncio
+    async def test_coin_selection_make_transaction(self, two_wallet_nodes, trusted):
```
emlowe, line 839: I'm not sure we need this test - but we need some tests of select_coins() itself to make sure the overall logic is correct. I see tests for the individual functions, but nothing for select_coins(). Some tests with very large lists might be worthwhile - some people will have 10s of thousands of dust coins, we should check that.

chia/wallet/coin_selection.py
jack60612: a lot of these need to be removed as they cant happen, i forgot to remove them while initially debugging.

chia/wallet/wallet.py
jack60612: yep just needs to be removed.

chia/wallet/coin_selection.py
jack60612: yep

jack60612:
> I wonder about adding in support for an "ignore dust" feature - so add in the ability to ignore coins below a certain threshold?

ill add this in

chia/wallet/coin_selection.py
jack60612: Good point.

chia/wallet/coin_selection.py
jack60612: I am just going to remove this, as it should already be properly sorted.

tests/wallet/test_wallet.py
jack60612: Good point, after i separated the function i forgot to separate this.

chia/wallet/coin_selection.py
mariano54: No, they shouldn't be removed if they can't happen. That is the use cases for asserts

jack60612: Re added

chia/wallet/wallet.py
jack60612: Re added

mariano54 (dismissed):

mariano54 (dismissed):

mariano54:
Need this first:
https://github.com/Chia-Network/chia-blockchain/pull/10967

chia/wallet/coin_selection.py
```diff
@@ -73,23 +74,32 @@ async def select_coins(
...
+    elif smaller_coin_sum > amount:
+        coin_set = knapsack_coin_algorithm(smaller_coins, amount, max_coin_amount)
+        log.debug(f"Selected coins from knapsack algorithm: {coin_set}")
+        if coin_set is None:
+            raise ValueError("Knapsack algorithm failed to find a solution.")
+        if len(coin_set) > max_num_coins:
+            coin = select_smallest_coin_over_target(len(smaller_coins), valid_spendable_coins)
+            if coin is None or coin.amount < amount:
+                raise ValueError(
+                    "Can't make this transaction because the transaction would use over 500 "
```
emlowe, line 93: Suggest more concise string `raise ValueError(f"Transaction of {amount} would use more than 500 coins. Try sending a smaller amount")

chia/wallet/coin_selection.py
```diff
@@ -103,8 +113,23 @@ def check_for_exact_match(coin_list: List[Coin], target: uint64) -> Optional[Coi
     return None
 
 
+# amount of coins smaller than target, followed by a list of all valid spendable coins sorted in descending order.
+def select_smallest_coin_over_target(smaller_coin_amount: int, valid_spendable_coin_list: List[Coin]) -> Coin:
+    if smaller_coin_amount >= len(valid_spendable_coin_list):
+        raise ValueError(
+            "There are no coins greater then the target."
+            " This is caused by having too many dust coins. Try making a smaller transaction to condense the dust."
```
emlowe, line 121: `ValueError(f"Unable to select coins for transaction. Try sending a smaller amount")

chia/wallet/coin_selection.py
```diff
@@ -0,0 +1,160 @@
...
+        if coin_record.coin.amount < min_coin_amount:
+            continue
+        valid_spendable_coins.append(coin_record.coin)
+        sum_spendable_coins += coin_record.coin.amount
+
+    # This happens when we couldn't use one of the coins because it's already used
+    # but unconfirmed, and we are waiting for the change. (unconfirmed_additions)
+    if sum_spendable_coins < amount:
+        raise ValueError(
+            "Can't make this transaction at the moment. "
```
emlowe, line 56: `ValueError("Transaction for {amount} is greater than spendable balance of {sum_spendable_coins}. There may be other transactions pending")`

emlowe:
I made a few suggestions on error strings - I don't think we should mention anything about dust in the error string. (I'd probably prefer returning some error code instead, or a custom exception, but we don't historically do that)

mariano54 (approved):

emlowe:
I was looking at `get_max_send_amount` in both the standard wallet and the cat wallet, and I notice that it doesn't take into account the coin selection algorithm at all. It didn't before (so I think it would give wrong answers often) and it doesn't in this PR either (and so likely still gives wrong answers) - This is trying to actually determine the maximum number of coins that we want to fit in a transaction, but since it doesn't follow the coin selection algorithm it won't pick the correct coins for this calculation.

I'm not sure the best way to fix this function but it seems rather incorrect right now.

emlowe:
> I was looking at `get_max_send_amount` in both the standard wallet and the cat wallet, and I notice that it doesn't take into account the coin selection algorithm at all. EDIT: I was mistaken - it does this `spendable.sort(reverse=True, key=lambda record: record.coin.amount)` , so it does sort by biggest first like the old coin algorithm. However, it certainly doesn't use the new algorithm which will pick very different coins.
> 
> I'm not sure the best way to fix this function but it seems rather incorrect in this PR because it will choose different coins to calculate this max vs what the coin selection algo will do.

Oh, maybe this is ok - this function is actually trying to determine the maximum amount you could send if you choose all the biggest coins with a max number of coins set heuristically based on the standard wallet TX and some arbitrary scaling of max clvm cost. This does mean though that it's unlikely the code that hits coin limits in the new coin selection actually happen in the wild, since this function is check first.

So I think it's fine to merge in this change without further changed here.

emlowe (approved):

PR #9976, opened 2022-01-26
Fix CAT tests and CAT syncing

Quexington:
The CAT tests were not testing trusted sync due to using bytes instead of hex.  This exposed an error in which adding a CAT wallet after you have already received CATs will not find the CATs you have already received.  There's a proposed fix to that in the wallet node in here as well.


Quexington:
Deferring to the changes in https://github.com/Chia-Network/chia-blockchain/pull/9974 and will remake a PR there with anything that's left.

PR #9977, opened 2022-01-26
Fix CAT tests

Quexington:


Quexington:
This was cherry picked to the branch it was targetting

PR #9978, opened 2022-01-27
Atari fix post mega merge

altendky:


emlowe (dismissed):
I approve even in Draft.

chia/data_layer/data_store.py
```diff
@@ -332,9 +333,9 @@ async def get_roots_between(
                 "AND generation >= :generation_begin AND generation < :generation_end ORDER BY generation ASC",
                 {"tree_id": tree_id.hex(), "generation_begin": generation_begin, "generation_end": generation_end},
             )
-            if cursor.rowcount < 1:
-                raise Exception(f"tree id not in store {tree_id.hex()}")
             roots = [Root.from_row(row=row) async for row in cursor]
+            if len(roots) == 0:
```
almogdepaz, line 337: why not do a try catch gere as well if thats the way we are handling that case ?

altendky: How would you write the previous line to result in an exception for 0 elements but not for 1 or more while still capturing them all into `roots`?  But yes, this also left my brain a bit conflicted.  I decided not to spend a lot of time trying to come up with a solution that was good enough to act as a quality reference.  Just made something work.

almogdepaz: i think do it like in get_root and another for loop for parsing into Root, or remove the try/catch there and check the size?
seems  rn you will trow here is there is no result like in the other case

almogdepaz (dismissed):

emlowe (dismissed):
Ok for me - we need to get this into Atari to fix the branch

chia/data_layer/data_store.py
almogdepaz: i think i get the issue, this syntax is terrible sorry

almogdepaz (dismissed):

emlowe (dismissed):
ok

emlowe (approved):
ok

PR #9979, opened 2022-01-27
BlockchainInterface(Protocol)

altendky:
This is a possible alternative to https://github.com/Chia-Network/chia-blockchain/pull/9906.

Draft for:
- [ ] Actually completing the work and making the implementations of `BlockchainInterface` actually properly implement it.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -12,7 +12,7 @@ from clvm.casts import int_from_bytes
 from chia.consensus.block_body_validation import validate_block_body
 from chia.consensus.block_header_validation import validate_unfinished_header_block
 from chia.consensus.block_record import BlockRecord
-from chia.consensus.blockchain_interface import BlockchainInterface
+from chia.consensus.blockchain_interface import check_blockchain_interface
 from chia.consensus.constants import ConsensusConstants
 from chia.consensus.cost_calculator import NPCResult
 from chia.consensus.difficulty_adjustment import get_next_sub_slot_iters_and_difficulty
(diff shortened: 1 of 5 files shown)


chia/consensus/blockchain.py
```diff
@@ -731,6 +732,11 @@ def clean_block_records(self):
     async def get_block_records_in_range(self, start: int, stop: int) -> Dict[bytes32, BlockRecord]:
         return await self.block_store.get_block_records_in_range(start, stop)
 
+    def try_block_record(self, header_hash: bytes32) -> Optional[BlockRecord]:
```
almogdepaz, line 735: why do we need to add this ? its implemented in BlockchainInterface

almogdepaz, line 735: why change this ?
it was previously handled in BlockchainInterface so to not duplicate this code here and in the wallet

altendky: Because this is an exploration of using a `Protocol` instead of inheritance to actually check the interface.  Note that it was already duplicated into one of the blockchain classes and was the only implemented  method on `BlockchainInterface`.  If the copying is problematic we can consider various other options to avoid it.

almogdepaz: this was to avoid replicating the same function all in the interface wallet blockchain and cache

altendky: Yes, and inheritance isn't the only way to share code.  I thought I already added it back the other places.  And yes, that's the quick just hack it in bit.  The more interesting part I think is how the other classes weren't actually implementing the interface (in more ways than we already knew).

PR #9983, opened 2022-01-27
update analyze_blockchain tool to read v2 blockchain databases

arvidn:
and introduce option to run all generators in mempool mode. Also update to use the new clvm_rs API.

I used this to validate the new mempool rules against the existing blockchain

### DIFF
--- a/tools/analyze-chain.py
+++ b/tools/analyze-chain.py
@@ -2,28 +2,29 @@
 
 import sqlite3
 import sys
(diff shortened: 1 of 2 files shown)


richardkiss (approved):
I just looked briefly, but it's obviously not business critical, so I'm fine with a merge.

wjblanke (approved):
aok

PR #9990, opened 2022-01-27
begin/commit/rollback for replace proof

wjblanke:


chia/full_node/full_node.py
```diff
@@ -2152,9 +2152,18 @@ async def _replace_proof(
...
-                await self.block_store.db_wrapper.commit_transaction()
-                replaced = True
+                try:
+                    await self.block_store.db_wrapper.begin_transaction()
+                    await self.block_store.add_full_block(new_block.header_hash, new_block, block_record)
+                    await self.block_store.db_wrapper.commit_transaction()
+                    replaced = True
+                except BaseException as e:
+                    await self.block_store.db_wrapper.rollback_transaction()
+                    log.error(
```
altendky:
```suggestion
                    self.log.error(
```

chia/full_node/full_node.py
```diff
@@ -2152,9 +2152,18 @@ async def _replace_proof(
...
+                        f"_replace_proof error while adding block {block.header_hash} height {block.height},"
+                        f" rolling back: {traceback.format_exc()} {e}"
```
altendky, line 2164: If you intentionally want the 'message' at the end, ok.  I thought we often put it first.

```suggestion
                        f" rolling back: {e} {traceback.format_exc()}"
```

altendky (dismissed):
(you caught the `self.` while I was reviewing...)

I appreciate the concern and agree.  Thanks for finding this.

I'm happy for this to get merged and would also really like to see a context manager for handling transactions.  I would also like to see more db activities in transactions.  I figure many extra explicit transactions where they aren't needed are better than a single one missing where it is needed.  Here's one we've been using in the data layer work.

used like:
https://github.com/Chia-Network/chia-blockchain/blob/d1259a6832585831485ebd281d9573caeeafb98f/chia/data_layer/data_store.py#L280-L282

implementation:
https://github.com/Chia-Network/chia-blockchain/blob/eb02a9a64f43ca59c0512d697dde54586916d839/chia/util/db_wrapper.py#L34-L52

I did notice here that you located more steps inside the `try:` block than I did.  After thinking about it briefly I'm not certain which is better and would welcome some discussion around that (aside from this PR).  Additionally, I made a PoC for a smarter `ContextVar` based locker that doesn't use the manual `lock=` parameter above.  It seems to work but I should get a bit better understanding of the details of `ContextVar`s in asyncio before actually using it.  We might even be able to turn it into a thing to get a separate db connection per task, either brand new or from a pool if needed.

https://replit.com/@altendky/ComplexGiddyWeblogsoftware-2#main.py

chia/full_node/full_node.py
altendky: So there is a related (but not quite right here) context manager now.

https://github.com/Chia-Network/chia-blockchain/blob/cbbdb6510a1f6f2dba096bb499897c124b8ffd94/chia/util/log_exceptions.py#L6-L11

I suggested to Chris that it should be renamed.  We could use that existing name for one that reraises the exception after logging.  Or have a parameter that defaults to `consume=False` or such along with a parameter for an optional custom message like you provide here.  Regardless, getting to standardized tooling used everywhere for this would help with consistency and maybe improve usage.

altendky (approved):
I keep thinking of more nuances, but that can be saved for discussion when we make a context manager for this.

wjblanke:
ya i am not sure where begin transaction should be in the try either. closing since this was merged into mariano's PR

PR #9996, opened 2022-01-28
Add data to wallet RPCs to support metrics (+ update log_exceptions helper)

cmmarslender:
Data required for the initial batch of wallet metrics was mostly already present - just needed a few small additions.

Also updated the `log_exceptions()` util introduced in the last (full node) metrics data PR to _not_ consume the exception by default, per @altendky 's suggestion, so adding him to take a look here as well. In all cases calling log_exceptions() for metrics (only use at the moment), I'm still setting `consume=True` so nothing related to metrics that might throw an exception causes any adverse side effects.


mariano54 (approved):

mariano54:
`_state_changed` definitely can use an upgrade to be something more scalable and less bug prone. It's worth thinking about as you expand to more metrics in the future

PR #9999, opened 2022-01-28
benchmark for get_block_generator()

arvidn:
on Ubuntu, Threadripper, SSD

```
get_block_generator(): 1.696s
```

on Ubuntu, Xeon, HDD:

```
get_block_generator(): 11.960s
```


benchmarks/block_ref.py
```diff
@@ -0,0 +1,90 @@
...
+from chia.consensus.default_constants import DEFAULT_CONSTANTS
+from chia.full_node.block_store import BlockStore
+from chia.full_node.coin_store import CoinStore
+from chia.full_node.hint_store import HintStore
+from chia.types.blockchain_format.program import SerializedProgram
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.db_version import lookup_db_version
+from chia.util.db_wrapper import DBWrapper
+from chia.util.ints import uint32
+from valid_block_heights import transaction_block_heights
```
arvidn, line 21: this list is quite long, and hard coded to our actual chain up until recently. Is there a better way to do this other than including a very large python file?
I don't want to compute it in the benchmark for 2 reasons. 1. It would take a long time. 2. it would bring in blocks into the cache and make the test itself less accurate. (the difference between cold and hot cache on HDD is enormous, 12 seconds -> 3.5 seconds).

chia/consensus/blockchain.py
```diff
@@ -6,7 +6,12 @@
 from concurrent.futures.process import ProcessPoolExecutor
 from enum import Enum
 from pathlib import Path
-from typing import Dict, List, Optional, Set, Tuple, Union
+from typing import Dict, List, Optional, Set, Tuple
+
+try:
+    from typing import Protocol
+except ImportError:
+    from typing_extensions import Protocol  # type: ignore
```
arvidn, line 14: is this according to best practice? mypy still doesn't like the type from `typing_extensions`

chia/consensus/blockchain.py
```diff
@@ -859,8 +865,21 @@ def seen_compact_proofs(self, vdf_info: VDFInfo, height: uint32) -> bool:
...
+        def prev_header_hash(self) -> bytes32:
+            pass
+
+        @property
+        def transactions_generator(self) -> Optional[SerializedProgram]:
+            pass
+
+        @property
+        def transactions_generator_ref_list(self) -> List[uint32]:
+            pass
```
arvidn, line 879: perhaps this is controversial. We really don't need a full block here, just these 3 features of a block. This makes it a lot simpler to test and benchmark this function

benchmarks/block_ref.py
Seifollahy: Hi

chia/consensus/blockchain.py
Seifollahy: Hi

chia/consensus/blockchain.py
mariano54: I need to review this in detail but I rather not add stuff to this file

arvidn: I moved this protocol out into its own file, `chia/types/block_protocol.py`

arvidn:
I'm interested in better ideas than to include a 25000 line python array of block heights with transactions. building that list takes too long to do it on the fly, and it could also affect the benchmark since it would pre-warm the disk cache.
I could store it in binary form to make it smaller.

mariano54:
There is a lot of repeated data, so I think it would compress really well Or just store the deltas between the numbers and encode them in an efficient encoding like huffman or ANS (it's what we do in pospace).

mariano54 (approved):

wjblanke (approved):
aok

PR #10006, opened 2022-01-28
remove the GeneratorArgs type

arvidn:
Instead include the heights and generator programs as separate lists in BlockGenerator. The heights are not necessary when validating blocks, so this makes it easier to omit them in that case. The heights are only used when generating/farming a block.

The heights and the generator programs are always accessed independently, as separate lists, so this also means we don't have to construct those lists on the fly all the time.

The fundamental motivation for this change is to make it easier to move towards `Blockchain.get_block_generator()` looking up the generator programs from the DB, without having to worry about returning the list of heights. The heights are only used when generating a block, so it seems like unnecessary to handle it in the validation path.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -30,6 +30,7 @@ from chia.full_node.coin_store import CoinStore
 from chia.full_node.hint_store import HintStore
 from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions
 from chia.types.blockchain_format.coin import Coin
+from chia.types.blockchain_format.program import SerializedProgram
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.blockchain_format.sub_epoch_summary import SubEpochSummary
 from chia.types.blockchain_format.vdf import VDFInfo
@@ -37,7 +38,7 @@ from chia.types.coin_record import CoinRecord
 from chia.types.condition_opcodes import ConditionOpcode
 from chia.types.end_of_slot_bundle import EndOfSubSlotBundle
 from chia.types.full_block import FullBlock
-from chia.types.generator_types import BlockGenerator, GeneratorArg
+from chia.types.generator_types import BlockGenerator
 from chia.types.header_block import HeaderBlock
 from chia.types.unfinished_block import UnfinishedBlock
 from chia.types.unfinished_header_block import UnfinishedHeaderBlock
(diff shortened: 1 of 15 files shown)


chia/consensus/blockchain.py
```diff
@@ -867,25 +868,28 @@ async def get_block_generator(
...
-                # TODO: address hint error and remove ignore
-                #       error: Argument 1 to "get_full_block" of "Blockchain" has incompatible type "Optional[bytes32]";
-                #       expected "bytes32"  [arg-type]
-                ref_block = await self.get_full_block(header_hash)  # type: ignore[arg-type]
+
+                # if ref_height is invalid, this block should have failed with
+                # FUTURE_GENERATOR_REFS before getting here
+                assert header_hash is not None
+
+                ref_block = await self.block_store.get_full_block(header_hash)
```
arvidn, line 888: I addressed this warning while I was here anyway

chia/types/generator_types.py
```diff
@@ -34,10 +25,7 @@ class CompressorArg:
...
-    generator_args: List[GeneratorArg]
-
-    def block_height_list(self) -> List[uint32]:
-        return [a.block_height for a in self.generator_args]
+    generator_refs: List[SerializedProgram]
 
-    def generator_refs(self) -> List[SerializedProgram]:
-        return [a.generator for a in self.generator_args]
+    # the heights are only used when creating blocks, never when validating
+    block_height_list: List[uint32]
```
arvidn, line 31: this is the main change. Instead of member functions generating those lists, make them members directly. This alters the syntax for accessing them from a function call to a property access, which is the main reason for the other changes in this PR

chia/full_node/mempool_check_conditions.py
```diff
@@ -164,10 +163,7 @@ def get_name_puzzle_conditions(
 def get_puzzle_and_solution_for_coin(generator: BlockGenerator, coin_name: bytes, max_cost: int):
     try:
         block_program = generator.program
-        if not generator.generator_args:
```
arvidn, line 167: this check confused me. `generator_args` is not `Optional`

aqk: I think when this was written, there was not an associated type - I think there was code in the block parsing code that might have converted empty list -> None

aqk (dismissed):
Thanks, this is better

aqk:
`The heights are not necessary when validating blocks`

Is that correct? I think you need the previous block list when validating a block.

arvidn:
> Is that correct? I think you need the previous block list when validating a block.

We need the heights in order to look up the previous generators and pass those in to the current block's generator. But we don't pass in the heights themselves. That's what I mean. And we validate the hash of the ref-list separately, we don't use the `GeneratorArg` for that.

My understanding is that `GeneratorArg` and `BlockGenerator` is used to (1) set up the environment to *run* a block generator and (2) create a new block when farming a block. For (1), the heights are not part of the environment to run the block. In a future patch I'm working on, it will be simpler to not preserve those heights when preparing the environment to *run* the block.

aqk (dismissed):

aqk (approved):

jack60612 (approved):
Looks good

PR #10007, opened 2022-01-28
debugging...  all zero height peers

altendky:


altendky:
> Doesn't this strip the far nicer nomenclature from every other build just because Windows sucks?

the installer stuff i tossed in here was completely exploratory.  i was hoping to see the symptoms on windows etc without much effort.  this PR is actually holding debug info for an entirely unrelated issue.  if i ever decide i want to make a change to the version number handling for windows i'll do my best to remember to get your attention explicitly.  though it seems i don't really need to go out of my way for you to find even mildly hidden changes...  `:]`

PR #10009, opened 2022-01-28
Fix signing with path `m` key

Quexington:
Currently if you try to sign with just the path `m` this code raises an index error.


paninaro (approved):
Thanks for the fix!

PR #10014, opened 2022-01-28
fix: update run_block.py

lzr-cat:
- add asset id
- add parent directory to load transactions_generator_ref_list from a directory


aqk (dismissed):

emlowe (approved):
ok

PR #10015, opened 2022-01-28
Do our best to only show inner puzzle hashes

Quexington:


Yostra (dismissed):
lgtm

Yostra:
look like lint is failing

PR #10023, opened 2022-01-29
tests: Improve `tests_farmer_harvester_rpc.py`

xdustinface:
For me it doesn't look like the plot creation in c90dd3e is needed and it doesn't really test anything related here. It's just one redundant plot creation (the plot in `subdir` isn't even in the test-cache) Let me know if i miss something.

Note: If we merge this we can as well drop https://github.com/Chia-Network/test-cache/tree/main/test-plots/subdir.

### DIFF
--- a/tests/core/test_farmer_harvester_rpc.py
+++ b/tests/core/test_farmer_harvester_rpc.py
@@ -24,202 +24,198 @@ from tests.util.rpc import validate_get_routes
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


xdustinface:
Okay pushed some more commits to make this more like a general improvement of the whole file.

mariano54 (dismissed):

xdustinface:
Rebased after #9922.

mariano54 (approved):

Issue #9954, opened 2022-01-26
Lite Wallet only shows 49 offers[Bug] 


DrKiz:
### What happened?

Lite wallet stops showing offers after 49.  New offers after that point work, but they do not always show up on the list.

Makes it hard to cancel an offer if it doesn't show up, amiright?

An option to delete confirmed or cancelled transactions is required as well as increasing offers past 49 to 100 or unlimited?

### Version

chia wallet beta 1.2.11265

### What platform are you using?

Windows

### What ui mode are you using?

GUI

DrKiz:
Just as a suggestion.  New lite wallet releases should have these features:

1. Show all offers.  Even if its a 1000 pages.
2. Ability to delete or hide old deleted/fullfilled offers as they clutter up the display.
3. Ability to end ALL offers at once as an option.
4. Abitlity to print out or have a file tracking the offers made and fullfilled.

paninaro:
Thanks for the feedback @DrKiz. The GUI (and CLI) have been fixed such that you can show all of the offers with pagination, or all in one go (though I'm not sure I'd recommend that with 1000 offers...)

@sethjenks will be handling the design side of UX improvements. I agree that the GUI should include filtering options. We've also had some discussions about exporting the list of offers.

PR #10025, opened 2022-01-29
Fixed Offer history limit and GUI sorting

paninaro:
trade_store: Migrated the is_my_offer property from the serialized
trade_record to a new `is_my_offer` column in the trade_records table.
Rewrote get_trades_between to move the query/sorting into the database.
This supports the presentation of offers in the GUI (two lists sorted by
relevance).

RPC: get_all_offers now accepts options to restrict the listing to
completed offers, my offers, taken offers. Added get_offers_count call
to return a breakdown of the number of offers (my offers, taken offers,
total).

CLI: Updated `chia wallet get_offers` to include options to exclude
created/taken offers, as well as an option to reverse the listing. The
-ia|--include-all option has been renamed -ic|--include-completed to
make sense in the context of the --exclude-[my-offers|taken-offers]
options. The full listing is now output, fixing the issue with hitting
the prior paging cap of 50 records.

Addresses bug #9954

### DIFF
--- a/chia/wallet/trading/trade_store.py
+++ b/chia/wallet/trading/trade_store.py
@@ -1,6 +1,5 @@
 from time import perf_counter
 from typing import List, Optional, Tuple
-from operator import attrgetter
 
 import aiosqlite
 import logging
@@ -233,7 +232,6 @@ class TradeStore:
         """
         Returns the number of trades in the database broken down by is_my_offer status
         """
-        # select count(*) as TOTAL, sum(case when is_my_offer=1 then 1 else 0 end) as my_offers, sum(case when is_my_offer=0 then 1 else 0 end) as taken_offers from trade_records;
         query = "SELECT COUNT(*) AS total, "
         query += "SUM(CASE WHEN is_my_offer=1 THEN 1 ELSE 0 END) AS my_offers, "
         query += "SUM(CASE WHEN is_my_offer=0 THEN 1 ELSE 0 END) AS taken_offers "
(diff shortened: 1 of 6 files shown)


paninaro:
I'll modify the tests to account for the updated sorting. Beyond that, I'm going to make another change to the `is_my_offer` migration logic. As currently written, DBs that have been migrated (with the added column) will have trouble if downgrading to a prior version when attempting to add a row to `trade_records` due to the missing column.

paninaro:
I updated the trade sorting tests and fixed a minor bug in the process.

The migration issue mentioned previously won't actually cause a problem as we copy over the wallet DB from standalone_wallet, so downgrading (back to the light wallet) doesn't actually involve the light wallet using the migrated DB.

That said, to make future migrations easier, I've updated the INSERT OR REPLACE statement in `trade_store.py` to explicitly list the columns being touched. If a future column addition is made that has a default value, that default value will be applied when an older client runs this code. If the column listing is omitted, the old client will hit an error like the following on an insert: `Error: table trade_records has 7 columns but 6 values were supplied`

chia/wallet/trading/trade_store.py
```diff
@@ -251,38 +327,112 @@ async def get_all_trades(self) -> List[TradeRecord]:
...
+            )
+
+        # Create an ORDER BY clause according to the desired sort type
         if sort_key is None or sort_key == "CONFIRMED_AT_HEIGHT":
-            records = sorted(records, key=attrgetter("confirmed_at_index"), reverse=(not reverse))
+            order_by_clause = (
+                f"ORDER BY confirmed_at_index {'ASC' if reverse else 'DESC'}, "
+                f"trade_id {'DESC' if reverse else 'ASC'} "
+            )
         elif sort_key == "RELEVANCE":
```
Quexington, line 372: The reason these values were put into a class and a separate file was so that we could avoid adding an elif everytime we have something new to sort by.  Maybe extracting this functionality to a separate class or file would be could and would make this now massive function much more readable.

Quexington (approved):
Other than my comment above (which doesn't have to happen right now) this all looks good.

PR #10029, opened 2022-01-29
Detailed notes on Windows installer versioning issues

hoffmang9:
Microsoft Windows requires that any executable have three and only three fields for version number.

Quoting from: https://docs.microsoft.com/en-us/windows/win32/msi/productversion?WT.mc_id=email
The first field is the major version and has a maximum value of 255.
 The second field is the minor version and has a maximum value of 255. 
 The third field is called the build version or the update version and has a maximum value of 65,535.

### DIFF
--- a/build_scripts/installer-version.py
+++ b/build_scripts/installer-version.py
@@ -8,7 +8,7 @@ from setuptools_scm import get_version
 def main():
     windows = len(sys.argv) > 1 and "win" in sys.argv[1]  
     
-    # Special case windows to 0.1.6225
+    # Special case windows to, for example, 1.1.6225
     # Microsoft Windows requires that any executable have three and only three fields
     # for version number.
     # Quoting from: https://docs.microsoft.com/en-us/windows/win32/msi/productversion?WT.mc_id=email
--- a/build_scripts/installer-version.py
+++ b/build_scripts/installer-version.py
@@ -6,7 +6,15 @@ from setuptools_scm import get_version
 
 # example: 1.0b5.dev225
 def main():
-    windows = len(sys.argv) > 1 and "win" in sys.argv[1]  # Special case windows to 0.1.6225
+    windows = len(sys.argv) > 1 and "win" in sys.argv[1]  
+    
+    # Special case windows to 0.1.6225
+    # Microsoft Windows requires that any executable have three and only three fields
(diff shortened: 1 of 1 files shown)


AmineKhaldi (dismissed):
LGTM after PR #10027 lands.

build_scripts/installer-version.py
```diff
@@ -6,7 +6,15 @@
 
 # example: 1.0b5.dev225
 def main():
-    windows = len(sys.argv) > 1 and "win" in sys.argv[1]  # Special case windows to 0.1.6225
+    windows = len(sys.argv) > 1 and "win" in sys.argv[1]  
+    
+    # Special case windows to, for example, 1.1.6225
+    # Microsoft Windows requires that any executable have three and only three fields
+    # for version number.
+    # Quoting from: https://docs.microsoft.com/en-us/windows/win32/msi/productversion?WT.mc_id=email
```
altendky, line 14: I don't see that quote there.  Rather, I see this.

> Note that Windows Installer uses only the first three fields of the product version. If you include a fourth field in your product version, the installer ignores the fourth field.

But it does say that one of the first three has to change for an update so I guess that's the bit where this bubbles up and is important.  Doesn't seem like it needs to drive the odd file name on Windows, but that's separate from this PR.

hoffmang9: Somehow I'm causing confusion. 

Our installers in Windows create a .exe. If you try to run a .exe that has either 1.1.2.1 or 1.1.2dev226 the installer will segfault. 

The windows installer get their versions here as they are translated from setuptools_DCM into mostly semver versions.

altendky: I did not test via the installer as squirrel did not like the non-semver four segment version and I did not take the time to find the rest of this winding path...  but the issue at least does not apply to all executables in all cases as the comment being added to the code states.  Also, semver does have support for pre-release indicators.

image

I do not mean to argue that it is a priority to improve our development version reporting on Windows.  It isn't clear to me what the boundaries of the failure mode are so I'm not sure if or when this three-segment restriction would no longer be applicable.  I would expect a clarification here to express, with whatever degree of detail is remembered, either the specific scenario that was actually observed to be an issue, or, to link to documentation that directly confirms the statement.

PR #10031, opened 2022-01-30
Address test_multiple_writers flakiness

paninaro:
Catch TimeoutError in test_multiple_writers. If raised, we'll skip the test. Increased the timeout to 60 seconds.

Tested that a legitimate error in the test will still trigger a failure.

### DIFF
--- a/tests/core/util/test_config.py
+++ b/tests/core/util/test_config.py
@@ -6,7 +6,7 @@ import yaml
 
 from chia.util.config import create_default_chia_config, initial_config_file, load_config, save_config
 from chia.util.path import mkdir
-from multiprocessing import Pool
+from multiprocessing import Pool, TimeoutError
 from pathlib import Path
 from threading import Thread
 from time import sleep
@@ -210,4 +210,7 @@ class TestConfig:
         # read failures are detected, the failing process will assert.
         with Pool(processes=num_workers) as pool:
             res = pool.starmap_async(run_reader_and_writer_tasks, args)
-            res.get(timeout=10)
+            try:
+                res.get(timeout=60)
+            except TimeoutError:
+                pytest.skip("Timed out waiting for reader/writer processes to complete")


mariano54 (approved):

altendky:
Since we don't closely monitor our skips this is a bit concerning.  It could hit that every time and odds are good that nobody would notice.  What about giving it 5 retries based only on the `TimeoutError`?

paninaro:
> Since we don't closely monitor our skips this is a bit concerning. It could hit that every time and odds are good that nobody would notice. What about giving it 5 retries based only on the `TimeoutError`?

I bumped the timeout to 60 seconds. Retrying 5 times would likely only compound whatever resource constraints the OS is hitting, so I don't see that improving anything. I'm not opposed to rewriting these tests that are timing-sensitive, but the timeout isn't actually relevant to the test. If the test fails for actual reasons relating to the logic being tested, those failures will still trigger a CI failure.

altendky:
Has the test succeeded when the 60 second timeout occurs?  Or has it effectively not done the testing?

mariano54:
It looks like there is another flaky one:
https://github.com/Chia-Network/chia-blockchain/runs/5009690231?check_suite_focus=true


```
self = <tests.core.util.test_file_keyring_synchronization.TestFileKeyringSynchronization object at 0x113f96820>

    @using_temp_file_keyring()
    def test_writer_lock_blocked_by_readers(self):
        """
        When a reader lock is already held, another thread/process should not be able
        to acquire the lock for writing
        """
        lock_path = FileKeyring.lockfile_path_for_file_path(KeyringWrapper.get_shared_instance().keyring.keyring_path)
        lock = fasteners.InterProcessReaderWriterLock(str(lock_path))
    
        # When: a reader lock is already held
        lock.acquire_read_lock()
    
        child_proc_function = dummy_fn_requiring_writer_lock
        timeout = 0.25
        attempts = 4
    
        with Pool(processes=1) as pool:
            # When: a child process attempts to acquire the same lock for writing, failing after 1 second
            res = pool.starmap_async(child_writer_dispatch, [(child_proc_function, lock_path, timeout, attempts)])
    
            # Expect: lock acquisition times out (raises as FileKeyringLockTimeout)
            with pytest.raises(FileKeyringLockTimeout):
>               res.get(timeout=2)

tests/core/util/test_file_keyring_synchronization.py:330: 
```

paninaro:
> 

I'll address the `test_writer_lock_blocked_by_readers` flakiness in a separate PR

PR #10039, opened 2022-01-30
Flaky test in wallet rpc

mariano54:


mariano54:
There is also a flaky test in `test_pool_rpc` which failed here.

Quexington (approved):
Yep makes sense to me.

PR #10042, opened 2022-01-31
add fee config

almogdepaz:
add fee to dl rpc api and client
add fee to cmd's

### DIFF
--- a/chia/rpc/data_layer_rpc_api.py
+++ b/chia/rpc/data_layer_rpc_api.py
@@ -9,6 +9,7 @@ from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.byte_types import hexstr_to_bytes
 
 # todo input assertions for all rpc's
+from chia.util.ints import uint64
 from chia.util.streamable import recurse_jsonify
 
 
@@ -36,11 +37,17 @@ def process_change(change: Dict[str, Any]) -> Dict[str, Any]:
     }
 
 
+def get_fee(config: Dict[str, Any], request: Dict[str, Any]) -> uint64:
+    fee = request.get("fee")
+    if fee is None:
+        return uint64(config["fee"])
+    return uint64(fee)
+
+
 class DataLayerRpcApi:
     # TODO: other RPC APIs do not accept a wallet and the service start does not expect to provide one
     def __init__(self, data_layer: DataLayer):  # , wallet: DataLayerWallet):
         self.service: DataLayer = data_layer
-        # self.data_layer_wallet = wallet
         self.service_name = "chia_data_layer"
 
     def get_routes(self) -> Dict[str, Callable[[Any], Any]]:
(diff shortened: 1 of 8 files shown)


chia/rpc/data_layer_rpc_api.py
```diff
@@ -40,7 +40,10 @@ def process_change(change: Dict[str, Any]) -> Dict[str, Any]:
 def get_fee(config: Dict[str, Any], request: Dict[str, Any]) -> uint64:
     fee = request.get("fee")
     if fee is None:
-        return uint64(config["fee"])
+        config_fee = config.get("fee")
```
emlowe, line 43: config_fee = config.get("fee", 0) will return 0 for you as the default

setup.py
```diff
@@ -26,6 +26,7 @@
     "setproctitle==1.2.2",  # Gives the chia processes readable names
     "sortedcontainers==2.4.0",  # For maintaining sorted mempools
     "websockets==8.1.0",  # For use in wallet RPC and electron UI
+    # TODO: when moving to click 8 remove the pinning of black noted below
```
altendky, line 29: Just to note, this won't actually be added as it is already present in `atari`.  This showing is just an artifact of the branch not having been caught up, which is fine.  Same below.

chia/rpc/data_layer_rpc_client.py
```diff
@@ -15,8 +16,10 @@ async def get_value(self, store_id: bytes32, key: bytes) -> Dict[str, Any]:
         # TODO: better hinting for .fetch() (probably a TypedDict)
         return response  # type: ignore[no-any-return]
 
-    async def update_data_store(self, store_id: bytes32, changelist: Dict[str, str]) -> Dict[str, Any]:
-        response = await self.fetch("batch_update", {"id": store_id.hex(), "changelist": changelist})
+    async def update_data_store(
+        self, store_id: bytes32, changelist: Dict[str, str], fee: Optional[uint64]
```
altendky, line 20: It's not required, but `Optional` often would get a default of `None`.  Should that particular be avoided here?

```suggestion
        self, store_id: bytes32, changelist: Dict[str, str], fee: Optional[uint64] = None
```

chia/rpc/data_layer_rpc_client.py
```diff
@@ -1,12 +1,13 @@
...
 
 from chia.rpc.rpc_client import RpcClient
 from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.ints import uint64
 
 
 class DataLayerRpcClient(RpcClient):
-    async def create_data_store(self) -> Dict[str, Any]:
-        response = await self.fetch("create_data_store", {})
+    async def create_data_store(self, fee: Optional[uint64]) -> Dict[str, Any]:
```
altendky, line 9: Similar to below.

```suggestion
    async def create_data_store(self, fee: Optional[uint64] = None) -> Dict[str, Any]:
```

altendky (approved):
Looks fairly straightforward.  A few comments and suggestions to consider, but not required.  After seeing `bytes32([0] * 32)` here outside the wallet I brought up in keybase the question of where the empty-hash should be defined.  No need for this to wait on that discussion though.

Thanks!

Quexington (approved):

PR #10044, opened 2022-01-31
full_node: remove pending timeout tasks to save memory

mariano54:
From @altendky :

>  if you search for pending_timeouts i think we don't drop references to those until we drop the whole connection?  i think it can go away entirely by using the asyncio.wait_for() option, though i haven't tried to think of corner cases.  anyways, since you were noting memory 'leaks'.  sys.getsizeof(some_random_task_i_made) gives 200 bytes (though who knows what more is hidden behind that in references) so that's ballpark 1MB/day/peer.  400MB for an average of holding a peer for five days with 80 peers

### DIFF
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -362,6 +362,7 @@ class WSChiaConnection:
         await event.wait()
 
         self.pending_requests.pop(message.id)
+        self.pending_timeouts.pop(message.id)
         result: Optional[Message] = None
         if message.id in self.request_results:
             result = self.request_results[message.id]


chia/server/ws_connection.py
```diff
@@ -362,6 +362,7 @@ async def time_out(req_id, req_timeout):
         await event.wait()
 
         self.pending_requests.pop(message.id)
+        self.pending_timeouts.pop(message.id)
```
arvidn, line 365: without fully understanding the details here.. these two calls to `pop()` look like they really belong in a `finally` block

arvidn: also, don't we need to cancel the task and synchronize with it going away before removing the last handle to it?

mariano54: the task itself will finish after the sleep which is 60 seconds, i think that's fine. 
Regarding the pop, yeah I think this code can be improved

arvidn: we have a bunch of error when shutting down about tasks/futures that are not awaited, and attempts to access the database after the connection is closed. I suspect that's because we're not very strict about not "detaching" tasks.

If we shut down before the timeout expires, we'll presumably get another error like that. Perhaps here and now isn't the time to fix that, but I think going forward we should avoid detaching and start cleaning up the ones we have already

PR #10046, opened 2022-01-31
fix datalayer wallet creation

Quexington:


emlowe (dismissed):
Looks reasonable to me

tests/wallet/db_wallet/test_dl_wallet.py
```diff
@@ -239,6 +239,9 @@ async def is_singleton_confirmed(lid: bytes32) -> bool:
 
         new_root = MerkleTree([Program.to("root").get_tree_hash()]).calculate_root()
         dl_tx = await dl_wallet.create_update_state_spend(launcher_id, new_root)
+        with pytest.raises(ValueError) as exception:
+            await dl_wallet.create_update_state_spend(launcher_id, new_root)
+            assert "is currently pending" in str(exception.value)
```
altendky, line 244: I'm pretty sure the assertion does nothing at present since the exception would be raised out of the `await` and flow then jumps to the context manager exit.  The assert would have to be dedented.  But, `match=` should work.

```suggestion
        with pytest.raises(ValueError, match="is currently pending"):
	    await dl_wallet.create_update_state_spend(launcher_id, new_root)
```

altendky (changes requested):

chia/data_layer/data_layer_wallet.py
```diff
@@ -64,18 +63,30 @@ class DataLayerWallet:
...
 
+    @classmethod
+    async def create(
+        cls: Type[_T_DataLayerWallet],
+        wallet_state_manager: Any,
+        wallet: Wallet,
+        wallet_info: WalletInfo,
+        name: Optional[str] = None,
+    ) -> _T_DataLayerWallet:
+        self = cls()
```
altendky, line 74: This should be a dataclass so it has a proper init that can be used instead of just a totally freeform class instance created on the fly.

altendky (approved):

PR #10047, opened 2022-01-31
simplify updating of proofs (when compacting them)

arvidn:
This also saves two database lookups before the update

### DIFF
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -154,7 +154,7 @@ class BlockStore:
             (
                 block_bytes,
                 int(block.is_fully_compactified()),
-                header_hash,
+                self.maybe_to_hex(header_hash),
             ),
         )
 
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -170,7 +170,7 @@ class BlockStore:
             )
 
             await self.db.execute(
-                "INSERT OR REPLACE INTO full_blocks VALUES(?, ?, ?, ?, ?, ?, ?, ?)",
+                "INSERT OR IGNORE INTO full_blocks VALUES(?, ?, ?, ?, ?, ?, ?, ?)",
                 (
                     header_hash,
                     block.prev_header_hash,
@@ -185,7 +185,7 @@ class BlockStore:
 
         else:
             await self.db.execute(
-                "INSERT OR REPLACE INTO full_blocks VALUES(?, ?, ?, ?, ?)",
+                "INSERT OR IGNORE INTO full_blocks VALUES(?, ?, ?, ?, ?)",
                 (
                     header_hash.hex(),
                     block.height,
@@ -196,7 +196,7 @@ class BlockStore:
             )
 
             await self.db.execute(
-                "INSERT OR REPLACE INTO block_records VALUES(?, ?, ?, ?,?, ?, ?)",
+                "INSERT OR IGNORE INTO block_records VALUES(?, ?, ?, ?,?, ?, ?)",
                 (
                     header_hash.hex(),
                     block.prev_header_hash.hex(),
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -137,9 +137,7 @@ class BlockStore:
                 "UPDATE OR FAIL full_blocks SET in_main_chain=1 WHERE header_hash=?", header_hashes
             )
 
-    async def replace_proof(self, header_hash: bytes32, block: FullBlock) -> None:
-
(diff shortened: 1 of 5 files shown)


mariano54 (commented):
Seems fine to me, we would be only calling `add_block` in one place which is nice

arvidn:
in my first revision, I also removed the `OR UPDATE` in the `add_full_block()` function. Thinking that there's never a reason (with this change) to add the same block twice. However, we have tests that do this explicitly, so I'm not sure anymore.

If we don't believe there's any legitimate case where the same block *would* be added twice, it would be nice to have the additional invariant check in the db itself. But that seems secondary to the main part of this change.

mariano54:
Adding a block twice is valid. We don't rollback the adding of blocks if the transaction fails. 
This does not change the state of the blockchain. 
What matters is changing the peak which changes the state of the blockchain.

chia/full_node/block_store.py
```diff
@@ -137,9 +137,28 @@ async def set_in_chain(self, header_hashes: List[Tuple[bytes32]]) -> None:
...
+            block_bytes = bytes(block)
+
+        self.block_cache.put(header_hash, block)
+
+        await self.db.execute(
+            "UPDATE full_blocks SET block=?,is_fully_compactified=? WHERE header_hash=?",
+            (
+                block_bytes,
+                int(block.is_fully_compactified()),
+                header_hash,
```
arvidn, line 157: this hash is a hex string the in the v1 database. Fixing this + the test

mariano54:
Oh yeah now that you mention it, adding the block twice will cause issues, because it will set is_peak to False with this change. 

We do return earlier in `receive_block` if we already have that block though.

arvidn:
> Oh yeah now that you mention it, adding the block twice will cause issues, because it will set is_peak to False with this change.

My understanding is that that's fine. The only remaining place we call this function from now is followed by updating the peak (if everything is added successfully)

arvidn:
I expect we would only end up adding a block twice if we flip-flop between two forks of the chain. Is that right?
would we ever end up replacing the peak?
it makes me thing that maybe it would be more reliable to change the database operation to:

```
INSERT OR IGNORE
```

mariano54:
Flip flopping, or when syncing up. We usually start download blocks from a point in the past, to ensure we download the necessary blocks from peers. Still `receive_block` is under a lock, and it checks for already existing blocks, so we will never try adding it twice. I think it would be safe to IGNORE here or just throw if we try adding twice.

mariano54:
Oh, the check for duplicates is `self.contains_block` which only checks recent block records. Therefore it is possible to theoretically add a block twice. I agree the, ignore would be the best solution.

chia/full_node/block_store.py
```diff
@@ -137,9 +137,28 @@ async def set_in_chain(self, header_hashes: List[Tuple[bytes32]]) -> None:
                 "UPDATE OR FAIL full_blocks SET in_main_chain=1 WHERE header_hash=?", header_hashes
             )
 
-    async def add_full_block(
-        self, header_hash: bytes32, block: FullBlock, block_record: BlockRecord, in_main_chain: bool
-    ) -> None:
+    async def replace_proof(self, header_hash: bytes32, block: FullBlock) -> None:
```
almogdepaz, line 140: can receive only the block without the header hash, would be cleaner i think

chia/full_node/block_store.py
```diff
@@ -151,22 +170,22 @@ async def add_full_block(
...
-                "INSERT OR REPLACE INTO full_blocks VALUES(?, ?, ?, ?, ?, ?, ?, ?)",
+                "INSERT OR IGNORE INTO full_blocks VALUES(?, ?, ?, ?, ?, ?, ?, ?)",
                 (
                     header_hash,
                     block.prev_header_hash,
                     block.height,
                     ses,
                     int(block.is_fully_compactified()),
-                    in_main_chain,  # in_main_chain
+                    False,  # in_main_chain
```
almogdepaz, line 180: looks like we never use this now, maybe a comment if cant remove the column

arvidn: This field is new in the v2 schema and I expect us to rely more on it, as it can simplify a fair number of lookups based on height.

we use the `in_main_chain` when looking for blocks to compactify, and there are a fair number of places where we currently rely on the height-to-hash map in memory because we didn't used to have this field. Some of those places can be relaxed to not need the height-to-hash map in memory.

almogdepaz: very cool, i didn't realize we keep this updated in reconsider_peek

PR #10049, opened 2022-01-31
Simplify FullNode._replace_proof

arvidn:
Currently, `_replace_proof()` asks the `BlockStore` for all blocks at the specific height followed by a linear search for the one with the correct header hash. This patch proposes that we, instead, simply lookup the correct header hash directly (since we need to perform the height-to-hash mapping either way). This simplifies the logic by dropping the loop over the blocks.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -2148,8 +2148,6 @@ class FullNode:
             new_block = None
             if block.header_hash != expected_header_hash:
                 continue
-            block_record = await self.blockchain.get_block_record_from_db(expected_header_hash)
-            assert block_record is not None
 
             if field_vdf == CompressibleVDFField.CC_EOS_VDF:
                 for index, sub_slot in enumerate(block.finished_sub_slots):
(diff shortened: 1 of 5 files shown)


chia/full_node/full_node.py
```diff
@@ -2133,75 +2133,67 @@ async def _can_accept_compact_proof(
             self.log.info(f"Duplicate compact proof. Height: {height}. Header hash: {header_hash}.")
         return is_new_proof
 
+    # returns True if we ended up replacing the proof, and False otherwise
     async def _replace_proof(
         self,
         vdf_info: VDFInfo,
         vdf_proof: VDFProof,
         height: uint32,
```
almogdepaz, line 2141: can pass the header hash from the request into replace proof instead of using hight at all

arvidn: good point. I assumed the reason we made this lookup here was because the network message didn't include the hash, but it appears it does

mariano54 (approved):
Nice simplifications

chia/full_node/full_node.py
```diff
@@ -2133,75 +2133,64 @@ async def _can_accept_compact_proof(
...
+
+        block = await self.block_store.get_full_block(header_hash)
+        if block is None:
+            return False
+
+        new_block = None
+
+        if field_vdf == CompressibleVDFField.CC_EOS_VDF:
+            for index, sub_slot in enumerate(block.finished_sub_slots):
+                if sub_slot.challenge_chain.challenge_chain_end_of_slot_vdf == vdf_info:
```
almogdepaz, line 2153: i think we should error in cases where the vdf_info doesn't match 
since now we pull the block from the db by hash basically what we have here is the same block with different vdf_info

not suggesting we fix that in this pr

arvidn:
one behavior I think this patch changes, is that if we receive a compactified proof for a block that's no longer in the main chain, we used to ignore it. With this change we still update that block.

I can't tell whether that's good or bad.

PR #10056, opened 2022-02-01
Fix run block tool file type and hint

altendky:


altendky:
I think this is what you want...  didn't test it, just figured this was a better way to share the suggestion than english words on keybase.

altendky:
Pretty sure this is outdated now.

PR #10064, opened 2022-02-01
Optimize get ancestors 2

fchirica:
Reopened https://github.com/Chia-Network/chia-blockchain/pull/9748 in order to get the commit signed.


chia/data_layer/data_store.py
```diff
@@ -154,12 +186,35 @@ async def _insert_internal_node(self, left_hash: bytes32, right_hash: bytes32, t
...
-    async def _insert_terminal_node(self, key: bytes, value: bytes, tree_id: bytes32) -> bytes32:
+    async def _insert_ancestor_table(
+        self,
+        left_hash: bytes32,
+        right_hash: bytes32,
+        tree_id: bytes32,
+        generation: int,
+    ) -> None:
+        node_hash = Program.to((left_hash, right_hash)).get_tree_hash(left_hash, right_hash)
+
```
almogdepaz, line 201: we need to handle the case where the value is already in the table in the case of a failed download that left unreferenced nodes in the db

fchirica: fixed, good catch

PR #8965, opened 2021-10-26
Announcement hash should never be None

Quexington:
This is mostly a fix to make the type hints accurately reflect what we are passing to the function `make_solution`.

PR #10066, opened 2022-02-01
Refactor generate_signed_transaction methods and CAT support in tx RPCs

Quexington:
The first few commits in this PR are to refactor mostly the standard wallet's `generate_signed_transaction` method to change two things:
* Every `generate_signed_transaction` method should return a list of transactions (most other wallets need this for fee txs)
* The methods now take `Payment` objects as inputs.  These objects roughly correspond to the `primaries` argument before and allow ease of passing in multiple amounts, puzzle hashes, and memos (especially if you don't want a memo on every transaction).  It also makes much of the logic inside the function simpler.

I also carried over the changes from #8965 since it was relevant to the same function.

There are also a few commits aimed at removing some of the areas where we do an `if/else` branch based on whether the wallet is a CAT wallet or not, simply because the functions previously had different APIs.  This made adding CAT support to `send_transaction` and `create_signed_transaction` much simpler and more sensible.  The APIs should also work for any other wallets that properly implement a `generate_signed_transaction` method.

### DIFF
--- a/chia/cmds/wallet_funcs.py
+++ b/chia/cmds/wallet_funcs.py
@@ -185,14 +185,10 @@ async def send(args: dict, wallet_client: WalletRpcClient, fingerprint: int) -> 
 
     final_fee = uint64(int(fee * units["chia"]))
     final_amount: uint64
-    if typ == WalletType.STANDARD_WALLET:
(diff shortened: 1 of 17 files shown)


Quexington:
Kick CI

PR #10067, opened 2022-02-01
Allow older jobs to run to completion when on main

cmmarslender:
When on main, the concurrency group will now include the SHA, which effectively makes all groups on main unique, so nothing gets cancelled or prevented from running parallel


hoffmang9 (approved):
lgtm - waves by to more runners...

PR #10069, opened 2022-02-02
strict xfail test_create_multiple_pool_wallets() for untrusted

altendky:


mariano54 (approved):

altendky:
Since we've already got https://github.com/Chia-Network/chia-blockchain/pull/10070 presumably it goes in instead of this.

PR #10071, opened 2022-02-02
no need to pass in header hash separate from FullBlock in BlockStore.replace_proof

arvidn:


almogdepaz (approved):

arvidn:
the pool tests seem to fail consistently

mariano54:
> the pool tests seem to fail consistently

This will be fixed in https://github.com/Chia-Network/chia-blockchain/pull/10070.
I hope to merge after tests pass.

chia/full_node/block_store.py
```diff
@@ -137,24 +137,22 @@ async def set_in_chain(self, header_hashes: List[Tuple[bytes32]]) -> None:
...
+    async def replace_proof(self, block: FullBlock) -> None:
 
         block_bytes: bytes
         if self.db_wrapper.db_version == 2:
             block_bytes = self.compress(block)
         else:
             block_bytes = bytes(block)
 
-        self.block_cache.put(header_hash, block)
+        self.block_cache.put(block.header_hash, block)
```
mariano54, line 148: This is actually slower because `.header_hash` computes the hash of the whole object, which can be slow. We did some experiments of creating a cache in streamable in order to cache things like this, but we haven't added that yet.

arvidn: good point. In fact, if I revert this I can also fix the caller to not compute the header hash, since it's already available.
I think, long-term, it would be nice to make `header_hash` a function, rather than an attribute

arvidn: how about https://github.com/Chia-Network/chia-blockchain/pull/10075 instead of this patch?

PR #10076, opened 2022-02-02
add tool to test adding every block from a blockchain database

arvidn:
as a sanity check for a new version of the full node.

I envision this to be a faster alternative to test a sync from scratch. Perhaps not quite as thorough, but I believe it's faster.


tools/test-full-sync.py
```diff
@@ -0,0 +1,77 @@
...
+                pre_validation_results = await bc.pre_validate_blocks_multiprocessing(
+                    block_batch, {}, validate_signatures=False
+                )
+                assert len(pre_validation_results) > 0
+                for pre_valid, block in zip(pre_validation_results, block_batch):
+                    if pre_valid.error is not None:
+                        print(f"block failed pre-validation: {pre_valid.error}")
+                        return
+
+                    added, error_code, fork_height, coin_changes = await bc.receive_block(block, pre_valid, None)
```
mariano54, line 57: I recommend using the outer api: `full_node_api.respond_block(...)`, since it will test a little bit more full node logic, and it will be basically the same speed

mariano54: Oh you can't do it will `full_node_api.respond_block` but you can do it with the internal version `full_node.respond_block`

mariano54 (commented):
I think we should start a full node and use FullNode.respond_block, it's more realistic.

arvidn:
oh, I must have failed to push my commit. I don't know what happened. hold on

mariano54 (approved):

PR #10083, opened 2022-02-02
check_keys: check 50 instead of 500, farm default observer

mariano54:
My daemon was timing out because it was not finding a key, and searching through 500 was way too slow. 

A few changes:
- 500 -> 50 reduces time
- Precompute intermediate paths
- Calculate both observer and non-observer keys
- Change the default to observer

### DIFF
--- a/chia/cmds/init_funcs.py
+++ b/chia/cmds/init_funcs.py
@@ -22,7 +22,6 @@ from chia.util.config import (
     save_config,
     unflatten_properties,
 )
-from chia.util.ints import uint32
 from chia.util.keychain import Keychain
 from chia.util.path import mkdir, path_from_root
 from chia.util.ssl_check import (
@@ -33,7 +32,13 @@ from chia.util.ssl_check import (
     check_and_fix_permissions_for_ssl_file,
     fix_ssl,
 )
-from chia.wallet.derive_keys import master_sk_to_pool_sk, master_sk_to_wallet_sk
+from chia.wallet.derive_keys import (
+    master_sk_to_pool_sk,
+    master_sk_to_wallet_sk_intermediate,
+    master_sk_to_wallet_sk_unhardened_intermediate,
+    _derive_path,
+    _derive_path_unhardened,
+)
 from chia.cmds.configure import configure
 
 private_node_names = {"full_node", "wallet", "farmer", "harvester", "timelord", "daemon"}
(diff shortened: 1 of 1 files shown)


chia/cmds/init_funcs.py
```diff
@@ -74,15 +79,31 @@ def check_keys(new_root: Path, keychain: Optional[Keychain] = None) -> None:
...
+            all_targets.append(
+                encode_puzzle_hash(
+                    create_puzzlehash_for_pk(_derive_path_unhardened(intermediate_o, [i]).get_g1()), prefix
+                )
+            )
             all_targets.append(
-                encode_puzzle_hash(create_puzzlehash_for_pk(master_sk_to_wallet_sk(sk, uint32(i)).get_g1()), prefix)
+                encode_puzzle_hash(create_puzzlehash_for_pk(_derive_path(intermediate_n, [i]).get_g1()), prefix)
             )
             if all_targets[-1] == config["farmer"].get("xch_target_address"):
```
paninaro, line 108: Looks like the farmer and pool rewards addresses are only checked against the non-observer ph. Is that correct?

chia/cmds/init_funcs.py
```diff
@@ -99,7 +120,7 @@ def check_keys(new_root: Path, keychain: Optional[Keychain] = None) -> None:
         updated_target = True
```
paninaro, line 120: Just want to double check that the 2 lines above this are doing the right thing. Makes sense to me that we want to set the reward address to an observer address, which `all_targets[0]` would be.

chia/cmds/init_funcs.py
```diff
@@ -99,7 +120,7 @@ def check_keys(new_root: Path, keychain: Optional[Keychain] = None) -> None:
...
     elif config["farmer"]["xch_target_address"] not in all_targets:
         print(
-            f"WARNING: using a farmer address which we don't have the private"
+            f"WARNING: using a farmer address which we might not have the private"
```
paninaro, line 123: Should `f"WARNING: using a pool address which we don't have the private"` also be updated to use the same language?

chia/cmds/init_funcs.py
mariano54: Oh yeah we have to check both

chia/cmds/init_funcs.py
mariano54: yes

chia/cmds/init_funcs.py
mariano54: Yes thank you

mariano54:
One thing I wanted to ask you is:
I don't see these logs, because we are calling print instead of log.info(...). Do you think this is a problem, and should we fix this?

Also, I don't see much value in this scan of all keys anyway, because people might have more than 100 (or 500) addresses, and it will show the error message.. If people are farming to a cold wallet address, they will have to do the whole scan every time.. maybe we reduce it even more, or remove the check altogether? what do you think @paninaro ?

paninaro:
> One thing I wanted to ask you is: I don't see these logs, because we are calling print instead of log.info(...). Do you think this is a problem, and should we fix this?
> 
> Also, I don't see much value in this scan of all keys anyway, because people might have more than 100 (or 500) addresses, and it will show the error message.. If people are farming to a cold wallet address, they will have to do the whole scan every time.. maybe we reduce it even more, or remove the check altogether? what do you think @paninaro ?

I agree that this isn't helpful when check_keys() is called by the daemon or one of the other services, for the logging reason you mentioned. If we want to continue calling this in the manner it is today, we should probably emit to either the logger if running in a server process, or to stdout if invoked from the CLI. Adding a key is the one area where I can see this still being an appropriate check as it's user-invoked and taking a bit of time isn't too bad.

Seems to me that the goal is to help users understand where the payouts are going, so that they have something actionable if funds don't appear where they expect. For the GUI, we have the "Manage Farming Rewards" dialog, but it's pretty well hidden -- maybe we can surface that somewhere a _bit_ more prominently. For CLI farming, I haven't looked at the farming/pooling rewards code lately, but perhaps an alternative would be to log the payout address there (is it already?) I think this would be my preferred solution -- improving the logging elsewhere or doing something similar to clarify where funds were paid-out.

Another half-baked thought, but perhaps a new `chia diagnostics` or `chia first-aid` command could perform this check as one of several checks (port mapping success? keyring migration? others?). So the user could invoke a check-up if something doesn't appear to be working as expected.

mariano54:
OK I agree this is not so useful at the moment, but it IS useful for people starting up their chia node for the first time (since it sets the address). 

`chia diagnostics` sounds like it would be useful. Also I agree it would be nice to log to file if not using CLI. But I will refrain from making these changes right now. 

The main motivation here was preventing the 30s timeout due to slow keygen, which I think this PR solves.

paninaro (approved):

PR #10085, opened 2022-02-02
Support for performing incremental legacy key migration.

paninaro:
This handles the case where new keys were created by an older client or the light wallet, and those keys then need to be moved to keyring.yaml. `chia keys migrate` has been added to the CLI for manual operation, otherwise the GUI will opportunistically attempt a silent migration whenever the `unlock_keyring` RPC is called.

To avoid unnecessarily hitting the legacy keyring on each launch of the GUI, a file is dropped in `~/.chia_keys/.last_legacy_migration` to help determine whether a migration check should be performed. This check only occurs when attempting to perform a silent migration -- The `chia keys migrate` CLI will always attempt to migrate.

Tests added to cover the new CLI and related functions, as well as the migration code that was originally added to support keyring.yaml migration.

GH Issue: #10063

### DIFF
--- a/chia/util/keychain.py
+++ b/chia/util/keychain.py
@@ -235,10 +235,17 @@ class Keychain:
     list of all keys.
     """
 
-    def __init__(self, user: Optional[str] = None, service: Optional[str] = None):
+    def __init__(self, user: Optional[str] = None, service: Optional[str] = None, force_legacy: bool = False):
         self.user = user if user is not None else default_keychain_user()
         self.service = service if service is not None else default_keychain_service()
-        self.keyring_wrapper = KeyringWrapper.get_shared_instance()
+        if force_legacy:
+            legacy_keyring_wrapper = KeyringWrapper.get_legacy_instance()
+            if legacy_keyring_wrapper is not None:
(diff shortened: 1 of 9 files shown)


chia/util/keychain.py
```diff
@@ -399,6 +406,27 @@ def delete_key_by_fingerprint(self, fingerprint: int):
...
             pkent = self._get_pk_and_entropy(get_private_key_user(self.user, index))
 
+    def delete_keys(self, keys_to_delete: List[Tuple[PrivateKey, bytes]]):
+        """
+        Deletes all keys in the list.
+        """
+        remaining_keys = {str(x[0]) for x in keys_to_delete}
+        index = 0
+        pkent = self._get_pk_and_entropy(get_private_key_user(self.user, index))
+        while index <= MAX_KEYS:
```
emlowe, line 416: if len(remaining_keys)==0 you can also exit loop, correct?

paninaro: Good catch

emlowe (dismissed):
ok for me

emlowe (dismissed):
good for me

PR #10086, opened 2022-02-02
Rename migrated standalone_wallet wallet DB to v2

paninaro:
Updated `initial-config.yaml` so that the wallet DB is named 'v2' instead of 'v1'. standalone_wallet DBs that are migrated to the mainnet directory are renamed to use 'v2' in the name and have the 'sqlite_new' suffix changed to 'sqlite'

e.g.
`~/.chia/standalone_wallet/wallet/db/blockchain_wallet_v1_mainnet_1234567890.sqlite_new`
becomes:
`~/.chia/mainnet/wallet/db/blockchain_wallet_v2_mainnet_1234567890.sqlite`

### DIFF
--- a/chia/util/initial-config.yaml
+++ b/chia/util/initial-config.yaml
@@ -474,7 +474,8 @@ wallet:
     port: 8444
 
   testing: False
-  database_path: wallet/db/blockchain_wallet_v1_CHALLENGE_KEY.sqlite
+  # v2 used by the light wallet sync protocol
+  database_path: wallet/db/blockchain_wallet_v2_CHALLENGE_KEY.sqlite
   # wallet_peers_path is deprecated and has been replaced by wallet_peers_file_path
   wallet_peers_path: wallet/db/wallet_peers.sqlite
   wallet_peers_file_path: wallet/db/wallet_peers.dat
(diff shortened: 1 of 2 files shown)


altendky:
Is the plan to long term have the path in the config file not be what is used?  I guess we already consider it a template so it varies in that sense.  Is `_new` basically v2?  Or just a different name so people can hop back and forth between software versions?

paninaro:
> Is the plan to long term have the path in the config file not be what is used? I guess we already consider it a template so it varies in that sense. Is `_new` basically v2? Or just a different name so people can hop back and forth between software versions?

Probably best answered by @Yostra

Yostra:
it makes sense to use update a path in the config. Old wallet is being obsoleted completely so no plans to support hopping back.

paninaro:
> it makes sense to use update a path in the config. Old wallet is being obsoleted completely so no plans to support hopping back.

I'll update `initial-config.yaml` to reference the new filename. Also, given a brief convo with @Yostra we'll rename the DB template to reference "v2" instead of "v1" and drop "new" from the filename.

emlowe (approved):
Approved

PR #10094, opened 2022-02-03
Ms.wallet fixes

mariano54:
* Configure the correct DNS server for tesntet10
* Set singletons to spent when they are spent
* Don't show "synched" until we finish syncing to the correct height
* Improve code quality slightly


mariano54:
Can you review @Yostra ?

PR #10102, opened 2022-02-03
Fix running blocks with generators that use back-references

aqk:


arvidn:
adam, can you explain what problem this fixes?
There's already a test exercising a block with a back reference, isn't there?

emlowe:
The run_block unit tests are failing.

FYI, you can quickly run them with `pytest tests/tools/test_run_block.py` 
```
>       puzzle_result = puzzle_result.to_program()
E       AttributeError: 'Program' object has no attribute 'to_program'
```

emlowe:
No, unfortunately, there isn't a test with back-references. And it doesn't work currently. We need to add the test, but need to fix the code first.

tools/run_block.py
```diff
@@ -118,11 +110,11 @@ def run_generator(
         _, asset_id, _ = curried_args
         memo = ""
 
-        result = puzzle.run(solution)
+        puzzle_result = puzzle.to_program().run(solution)
```
arvidn, line 113: as far as I can tell, this will de-serialize the program (`puzzle` is a `SerizliedProgram`, right?) and then immediately serialize it again, to run it

tools/run_block.py
```diff
@@ -93,17 +92,10 @@ def run_generator(
...
 
-    if height >= DEFAULT_CONSTANTS.SOFT_FORK_HEIGHT:
-        # conditions must use integers in canonical encoding (i.e. no redundant
-        # leading zeros)
-        # the division operator may not be used with negative operands
-        flags = COND_CANON_INTS | NO_NEG_DIV
-    else:
-        flags = 0
-
-    _, result = block_generator.program.run_with_cost(max_cost, flags, block_generator.generator_refs)
```
arvidn, line 104: as far as I can see, the problem is that `generator_refs` need to be wrapped in a list here

aqk: Yes, thanks!

tools/run_block.py
```diff
@@ -101,15 +102,16 @@ def run_generator(
     else:
         flags = 0
 
-    _, result = block_generator.program.run_with_cost(max_cost, flags, block_generator.generator_refs)
+    args = create_generator_args(block_generator.generator_refs).first()
+    _, block_result = block_generator.program.run_with_cost(max_cost, flags, args)
```
arvidn, line 106: I'm not sure I understand why this works actually. Here's my understanding:

When the full node runs block generators it runs them through our ROM program that takes all the referenced generator programs as byte blobs. it will parse them (in chialisp) and pass them on to the actual generator that's in the block. So, there are two levels of the generator interface.

1. going through the ROM program, the inputs are flat buffers of clvm serialized programs
2. going directly to the generator program as it appears in the block, the inputs are the clvm programs themselves, as clvm tokens (not as a serialized buffer).

In my original patch, I was going for (2), just to keep things simple. One observation is that even though our referenced generators are serialized (in `SerializedProgram`) they are treated as the clvm tokens, not as flat buffers. This is because the interface to `clvm_rs` itself passes all buffers in serialized form. So, `generator_refs` will pass the clvm-tokens to the program. As far as I understand, what `create_generator_args()` does is two things:

1. turns the `SerializedProgram` obejct into clvm atoms containing the serialized form (which is the input the ROM program expects)
2. wraps this list in another list, as this API allows easy extensions in the futre.

Given that this code doesn't use the generator ROM code, I think we want to pass in the `SerializedPrograms` in their existing form, i.e. not turn them into atoms of serialized bytes. The thing that was missing originally was to wrap them in an extra list.

arvidn (dismissed):
as long as you're confident the test actually exercises the block reference, I think this is fine.

If the block that's referencing the other block *doesn't* actually use it for anything, the test wouldn't detect the error I'm suspecting. I suspect that the generator arguments you pass in now are flat atoms of programs in serialized form, whereas it's supposed to be the programs in clvm-token form.

tools/run_block.py
```diff
@@ -101,15 +102,16 @@ def run_generator(
...
 
-    coin_spends = result.first()
+    coin_spends = block_result.first()
 
     cat_list: List[CAT] = []
     for spend in coin_spends.as_iter():
 
         parent, puzzle, amount, solution = spend.as_iter()
-
+        print(f"parent:{parent}\npuzzle:{puzzle}\namount:{amount}\nsolution:{solution}\n")
```
pytlesk4, line 114:
```suggestion
```

emlowe (dismissed):
ok

arvidn (approved):

PR #10103, opened 2022-02-03
Move black from SuperLinter to upload workflow, use 21.12b0 in pre-commit

altendky:
- Match black versions between pre-commit and `setup.py` `dev` extra
- Take direct control of black linting, instead of via Super Linter
- Block PyPI uploads on black linting
- Configure such that `black .` does what we want.

Draft for:
- [x] Consideration of blocking real PyPI uploads triggered by tags.  Must keep associated deps pinned?


pyproject.toml
```diff
@@ -7,3 +7,11 @@ local_scheme = "no-local-version"
 
 [tool.black]
 line-length = 120
+target-version =  ['py37', 'py38', 'py39']
+include = '''
+^/(
+    [^/]*.py
+    | (benchmarks|build_scripts|chia|tests|tools)/.*\.pyi?
```
xdustinface, line 14: Hmm whats the goal here? Can't we get around listing all directories since this requires us to remember to adjust this if we add another directory :)

xdustinface (dismissed):

pyproject.toml
altendky: This is in part to keep from running on random local directories you make.  Long term i want to keep this list small.  In a just me project it would be `chia/` and maybe `build_scripts` and everything else (yes, including tests) would be in `chia/`.  We should be able to roll benchmarks and tools into the project itself at least, maybe with a chiadev command or `chia extras` or...  something.

wjblanke:
wanted to merge but there was a conflict

xdustinface (approved):

PR #10104, opened 2022-02-04
optimize get_block_generator()

arvidn:
There are two parts to this optimization:

1. Perform the database lookup of all blocks simultaneously, instead of one at a time. This allows sqlite to more efficiently schedule the disk accesses.
2. Circumvent the normal `FullBlock` parser, and instead introduce a single-purpose parser that just picks out the generator program

The parser in (2) is not meant as a long term solution, but a stop-gap until we fix the parsing (and probably data structure) for `FullBlock` objects. That's not a trivial amount of work so this will probably be useful for a few versions.

Most of the code is a test that tests the parser on a wide range of `FullBlock` objects. Specifically, all combinations of optional fields are exercised. See `test_full_block_utils.py`.

I've measured the timing of `get_generator()` in four scenarios, on SSD, HDD and v1, v2 databases.

| system  | before | after | after / before |
| --- | --- | --- | --- |
| Ubuntu, v2, SSD  | 1.767s | 0.220s | 12.5% |
| Ubuntu, v2, HDD | 15.319s | 6.155s | 40.18 % |
| Ubuntu, v1, SSD  | 1.808s | 0.387s | 21.4% |
| Ubuntu, v1, HDD | 24.517s | 20.003s | 81.59% |

As one would imagine, the HDD timing is dominated by the disk access.

## only sqlite change

What would happen if we only take the sqlite change?

| system  | before | after | after / before |
| --- | --- | --- | --- |
| Ubuntu, SSD  | 1.767s | 1.558s | 88.17% |
| Ubuntu, HDD | 15.319s | 7.371s | 48.12 % |

### DIFF
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -7,10 +7,13 @@ import zstd
 from chia.consensus.block_record import BlockRecord
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.full_block import FullBlock
+from chia.types.blockchain_format.program import SerializedProgram
 from chia.types.weight_proof import SubEpochChallengeSegment, SubEpochSegments
+from chia.util.errors import Err
 from chia.util.db_wrapper import DBWrapper
 from chia.util.ints import uint32
 from chia.util.lru_cache import LRUCache
+from chia.util.full_block_utils import generator_from_block
 
 log = logging.getLogger(__name__)
 
@@ -293,6 +296,39 @@ class BlockStore:
             for row in await cursor.fetchall():
                 ret.append(self.maybe_decompress(row[0]))
             return ret
+
+    async def get_generators_at(self, heights: List[uint32]) -> List[SerializedProgram]:
+        assert self.db_wrapper.db_version == 2
+
+        if len(heights) == 0:
+            return []
+
+        generators: Dict[uint32, SerializedProgram] = {}
+        heights_db = tuple(heights)
+        formatted_str = (
+            f"SELECT block, height from full_blocks "
+            f'WHERE in_main_chain=1 AND height in ({"?," * (len(heights_db) - 1)}?)'
+        )
+        async with self.db.execute(formatted_str, heights_db) as cursor:
+            async for row in cursor:
+                if self.db_wrapper.db_version == 2:
+                    block_bytes = zstd.decompress(row[0])
+                else:
+                    block_bytes = row[0]
+
+                try:
+                    gen = generator_from_block(block_bytes)
+                except BaseException:
+                    # this is defensive, on the off-chance that
+                    # generator_from_block() fails, fall back to the reliable
+                    # definition of parsing a block
+                    b = FullBlock.from_bytes(block_bytes)
+                    gen = b.transactions_generator
+                if gen is None:
+                    raise ValueError(Err.GENERATOR_REF_HAS_NO_GENERATOR)
+                generators[uint32(row[1])] = gen
+
+        return [generators[h] for h in heights]
 
     async def get_block_records_by_hash(self, header_hashes: List[bytes32]):
         """
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -297,6 +297,33 @@ class BlockStore:
                 ret.append(self.maybe_decompress(row[0]))
             return ret
 
+    async def get_generator(self, header_hash: bytes32) -> Optional[SerializedProgram]:
+
+        cached = self.block_cache.get(header_hash)
+        if cached is not None:
+            log.debug(f"cache hit for block {header_hash.hex()}")
+            return cached.transactions_generator
+
+        formatted_str = "SELECT block, height from full_blocks WHERE header_hash=?"
+        async with self.db.execute(formatted_str, (self.maybe_to_hex(header_hash),)) as cursor:
+            row = await cursor.fetchone()
+            if row is None:
+                return None
+            if self.db_wrapper.db_version == 2:
+                block_bytes = zstd.decompress(row[0])
+            else:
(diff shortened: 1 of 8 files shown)


tests/util/test_full_block_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    generator = SerializedProgram.from_bytes(bytes.fromhex("ff01820539"))
+
+    for foliage in get_foliage():
+        for foliage_transaction_block in get_foliage_transaction_block():
+            height = random.randint(0, 1000000)
+            for reward_chain_block in get_reward_chain_block(height):
+                for transactions_info in get_transactions_info(height):
+                    for challenge_chain_sp_proof in [rand_vdf_proof(), None]:
+                        for reward_chain_sp_proof in [rand_vdf_proof(), None]:
+                            for infused_challenge_chain_ip_proof in [rand_vdf_proof(), None]:
```
aqk, line 186: Is it important to the parser whether the block is partly or fully compactified?

eg for `challenge_chain_sp_proof` and `infused_challenge_chain_ip_proof`

chia/full_node/block_store.py
```diff
@@ -294,6 +297,39 @@ async def get_full_blocks_at(self, heights: List[uint32]) -> List[FullBlock]:
...
+                else:
+                    block_bytes = row[0]
+
+                try:
+                    gen = generator_from_block(block_bytes)
+                except BaseException:
+                    # this is defensive, on the off-chance that
+                    # generator_from_block() fails, fall back to the reliable
+                    # definition of parsing a block
+                    b = FullBlock.from_bytes(block_bytes)
```
aqk, line 325: This is great.

chia/full_node/block_store.py
```diff
@@ -294,6 +297,39 @@ async def get_full_blocks_at(self, heights: List[uint32]) -> List[FullBlock]:
...
+
+        generators: Dict[uint32, SerializedProgram] = {}
+        heights_db = tuple(heights)
+        formatted_str = (
+            f"SELECT block, height from full_blocks "
+            f'WHERE in_main_chain=1 AND height in ({"?," * (len(heights_db) - 1)}?)'
+        )
+        async with self.db.execute(formatted_str, heights_db) as cursor:
+            async for row in cursor:
+                if self.db_wrapper.db_version == 2:
```
mariano54, line 314: guaranteed to be 2, are we doing the if just for consistency?

chia/full_node/block_store.py
mariano54: Maybe put a log message here or we might miss it

tests/util/test_full_block_utils.py
mariano54: no, because the proof itself is just `bytes`, so it's handled correctly regardless of length

mariano54 (commented):
How bad is it if we just do the simultaneous lookup but not the parsing improvements? There is some complexity and risk with doing a lot of this kind of parsing, although it's nice that we have a fallback there just in case.

chia/full_node/block_store.py
arvidn: I would say the parsing is the bulk of the speedup, especially on SSD. I'll run some measurements

arvidn: oh, that response ended up in the wrong thread somehow. Regarding this thread; that's a good point. I wasn't considering that this was a v2-only function to begin with. I'll simplify this

tests/util/test_full_block_utils.py
arvidn: right, I just wanted the test to be somewhat exhaustive. There are variable length fields that aren't exercised at very many different lengths, but at least all combinations of optionals are (most of which are invalid blocks anyway). It takes about 4-5 mintues to run through all of these cases.

arvidn:
I'm working on adding tests for the new function on `BlockStore`. It also occurred to me that the v1 database case isn't covered by this optimization. I'm extending the patch to cover this case too.

I would also like to benchmark this patch against a v1 database, so I'm syncing one up right now.

arvidn:
I have amended the benchmarks to also include a v1 database

aqk:
`get_block_generator` is starting to get deeply nested

aqk (dismissed):

arvidn:
I'm going to add one more commit to make the test take 3:30 min rather than 4:40 min.

aqk (dismissed):

chia/util/full_block_utils.py
```diff
@@ -0,0 +1,208 @@
...
+def skip_uint8(buf: memoryview) -> memoryview:
+    return buf[1:]
+
+
+def skip_bool(buf: memoryview) -> memoryview:
+    assert buf[0] in [0, 1]
+    return buf[1:]
+
+
+# def skip_class_group_element(buf: memoryview) -> memoryview:
```
mariano54, line 56: Why comment this out?

chia/util/full_block_utils.py
```diff
@@ -0,0 +1,208 @@
...
+#    return buf[100:]  # bytes100
+
+
+def skip_vdf_info(buf: memoryview) -> memoryview:
```
mariano54, line 60: and this

mariano54 (approved):

chia/util/full_block_utils.py
arvidn: I made a few small optimizations by not calling some trivial functions, but just skip a fixed number of bytes instead. As you can see by the `skip_vdf_info()` function, it doesn't make those calls that are commented out, it just skips the bytes instead.

python doesn't seem to inline functions, so this abstraction appear to come at a price, and it doesn't seem very valuable.

aqk (approved):

PR #10105, opened 2022-02-04
Correct comment about max future block timestamp

hoffmang9:
The maximum future timestamp of a new block can only be 5 minutes ahead of the previous block.

https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/block_header_validation.py#L814
https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/default_constants.py#L26


mariano54 (approved):
Thanks for reviewing

PR #10113, opened 2022-02-04
Update pyinstaller version

cmmarslender:
PR for testing installers with the latest pyinstaller version.

@hoffmang9 I think there was something in 4.8 for windows you warned to look out for in testing, but I'm not remembering what it was. Do you remember? [pyinstaller changelog](https://pyinstaller.readthedocs.io/en/stable/CHANGES.html#id2)

### DIFF
--- a/build_scripts/build_linux_deb.sh
+++ b/build_scripts/build_linux_deb.sh
@@ -34,7 +34,7 @@ rm -rf dist
 mkdir dist
 
 echo "Create executables with pyinstaller"
-pip install pyinstaller==4.5
+pip install pyinstaller==4.9
 SPEC_FILE=$(python -c 'import chia; print(chia.PYINSTALLER_SPEC_PATH)')
 pyinstaller --log-level=INFO "$SPEC_FILE"
 LAST_EXIT_CODE=$?
--- a/build_scripts/build_linux_rpm.sh
+++ b/build_scripts/build_linux_rpm.sh
@@ -36,7 +36,7 @@ rm -rf dist
 mkdir dist
 
 echo "Create executables with pyinstaller"
-pip install pyinstaller==4.5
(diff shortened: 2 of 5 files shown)


hoffmang9:
Since we're mucking with the final executable's stack size, we need to pay attention to this - https://github.com/pyinstaller/pyinstaller/issues/6459

cmmarslender:
Ok, just did some testing on this. It appears that the stack size we set still applies just fine with the new version of pyinstaller. 

Tested both `1.2.11` and the version from this PR with `dumpbin /headers .\chia.exe` and the results were:

1.2.11:

```
7A1200 size of stack reserve
1000 size of stack commit
100000 size of heap reserve
1000 size of heap commit
```


PR Version:

```
7A1200 size of stack reserve
1000 size of stack commit
100000 size of heap reserve
1000 size of heap commit
```

I also tested making a plot, since the inline comment about the stack size change implies its to enable plotting.. Was able to make a k25 plot with default settings (unsure if the issue applied to all plot sizes or just larger plots?). Plot completed successfully and passed checks with `chia plots check`

hoffmang9:
Can we make a k=32 on Windows with this installer? I expect we're safe but that's the final real test. I'm going to approve under the assumption that will work. We can revert if it does not.

hoffmang9 (approved):
Ok

cmmarslender:
> Can we make a k=32 on Windows with this installer? I expect we're safe but that's the final real test. I'm going to approve under the assumption that will work. We can revert if it does not.

Yeah I can test this on a VM that isn't running on my main computer - will report back once its complete

cmmarslender:
Alrighty. K32 on windows completed and plot passed check with `Proofs 28 / 30, 0.9333`

wjblanke (approved):
aok

PR #10114, opened 2022-02-05
Fix wallet DB migration from the light wallet when initiated by the GUI

paninaro:
When CHIA_ROOT was set (by the GUI), STANDALONE_ROOT_PATH == DEFAULT_ROOT_PATH

STANDALONE_ROOT_PATH now expands to a different env variable to avoid conflicting with DEFAULT_ROOT_PATH when 
CHIA_ROOT is set. When the GUI launches chia services, CHIA_ROOT is set, which was preventing the wallet backend from copying existing DBs from the standalone_wallet.


emlowe (approved):
ok

wjblanke (approved):
aok

PR #10115, opened 2022-02-05
automatic db locking

altendky:
The existing `DBWrapper.locked_transaction()` context manager is used in the data layer store to provide a centralized definition of how locking and transactions should be handled.  The hassle is that you have to explicitly tell it whether to lock or not.  The use case being that sometimes a given function that needs a transaction and locking for integrity might be called directly and sometimes it is called from an outer function that already has to fulfill those requirements itself.  One feature enabled with this change is that there is no need to explicitly tell it to lock or not.  It tracks via [context variables](https://docs.python.org/3.10/library/contextvars.html) whether it needs to acquire the lock and enter a transaction, or if that has already been handled at an outer layer.  It also tracks the task it is used in as new tasks need to acquire the lock regardless of any 'nesting'.  A new task indicates concurrency at which point we are not in control of the interleaving of commands so the two tasks must operate sequentially, one after the other.

This could be expanded to provide separate connections per task as well and move away from the explicit async locking and towards SQLite-level locking via transactions.

Below is a contrived example show how `C.either()` can be called either directly or via `C.directly()`.  The outermost call in any given task will acquire the lock and handle the transaction opening, rollback, and commit.

```python
@dataclass
class C:
    db_wrapper: DBWrapper

    def either(self, addend):
        async with self.db_wrapper.locked_transaction() as connection:
            async with connection.execute("SELECT a_column FROM some_table WHERE other == 3") as cursor:
                row = await cursor.fetchone()
            assert row is not None
            [value] = row
            connection.execute("UPDATE some_table SET a_value = :a_value", {"a_value": value + addend})

    def directly(self):
        async with self.db_wrapper.locked_transaction() as connection:
            async with connection.execute("SELECT a_column FROM some_table WHERE other == 7") as cursor:
                row = await cursor.fetchone()
            assert row is not None
            [value] = row
            await self.either(addend=value)
```

Draft for:
- [ ] self consideration
- [ ] probably come up with more test cases

### DIFF
--- a/chia/data_layer/data_store.py
+++ b/chia/data_layer/data_store.py
@@ -330,9 +330,7 @@ class DataStore:
 
         return Root.from_row(row=row)
 
-    async def get_roots_between(
-        self, tree_id: bytes32, generation_begin: int, generation_end: int
-    ) -> List[Root]:
+    async def get_roots_between(self, tree_id: bytes32, generation_begin: int, generation_end: int) -> List[Root]:
         async with self.db_wrapper.locked_transaction():
             cursor = await self.db.execute(
                 "SELECT * FROM root WHERE tree_id == :tree_id "
@@ -379,9 +377,7 @@ class DataStore:
 
         return ancestors
 
-    async def get_keys_values(
-        self, tree_id: bytes32, root_hash: Optional[bytes32] = None
-    ) -> List[TerminalNode]:
+    async def get_keys_values(self, tree_id: bytes32, root_hash: Optional[bytes32] = None) -> List[TerminalNode]:
         async with self.db_wrapper.locked_transaction():
             if root_hash is None:
                 root = await self.get_tree_root(tree_id=tree_id)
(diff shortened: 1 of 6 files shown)


altendky:
Looks like the macOS issue applies to Windows as well.  https://github.com/omnilib/aiosqlite/pull/162  Closing a URL-defined memory db only clears it as expected in Linux, not in Windows or macOS.

https://www.sqlite.org/inmemorydb.html
> The database is automatically deleted and memory is reclaimed when the last connection to the database closes.

arvidn:
I don't really know what a "context variable" is, but it looks like another name for a global variable (scoped to a module perhaps?).

The approach I had in mind (but haven't explored, so it might be a dead end) is to take advantage of the fact that almost all data in our database (at least for the full node) is immutable once it's written. This means that reads (almost) never need a lock. The cases where you still may need a lock while reading is if you need to make multiple requests and have them be a consistent snapshot of what the current peak is. i.e. no reorgs.

What if we have a read-write lock, and if you read immutable data you don't need a lock at all. We could have two `Connection` objects, one that we use for read-only requests that we initialize with `PRAGMA query_only` that we hand back for read-locks.

But I think having a context manager is a great step in the right direction (and not have any code save off the `Connection` object and accessing it directly). That way we ensure we consistently use either a read- or write-lock when accessing the DB.

Rather than having a recursive mutex, we could just pass in the `Connection` object to the function that needs it. We acquire the lock at one place, and get a `Connection` object, and we use that and pass it on to other functions that need it.

PR #9614, opened 2021-12-18
apply stricter mypy to all new files, and fix a few tidbits

altendky:
This enforces stricter, though not actually all that strict, hinting policies on all new files while leaving existing files exempt.  It also adjusts a couple files that I figured were easier to adjust than to try to handle otherwise.  Note that this will apply to all existing PRs once merged and those that introduce new files will find those files bound by the stricter checking.

Draft for:
- [x] Discussion of type hint policies
- [x] Waiting until January to discuss and deploy
- [x] Standalone wallet being merged into `main`
- [ ] Finish discussion of proper `clvm_rs.run_generator2()` return hint

PR #10129, opened 2022-02-06
just put __init__.py files where we have .py files (plus subdirectories)

altendky:
Could we please add these files so that tools act as expected?  Is there some issue that they cause people given that we have tooling automatically enforcing them?  Long term, we're just not going to have a bunch of different directories that Python code goes in because that means we have to manage all the tooling to pay attention to all those different places.

I'm trying to break down https://github.com/Chia-Network/chia-blockchain/pull/9614 (more mypy) and this seems to make it respect the ignores in the config a bit better.

### DIFF
--- a/benchmarks/block_store.py
+++ b/benchmarks/block_store.py
@@ -29,9 +29,7 @@ random.seed(123456789)
 
 
 def rand_class_group_element() -> ClassgroupElement:
-    # TODO: address hint errors and remove ignores
-    #       error: Argument 1 to "ClassgroupElement" has incompatible type "bytes"; expected "bytes100"  [arg-type]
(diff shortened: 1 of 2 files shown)


jack60612 (approved):

altendky:
Bill, I would like to give Arvid and Mariano a chance to weigh in on this since this is a change from a previous decision they were involved in, IIRC.

mariano54 (approved):
Sure, why not.

arvidn:
I don't even know why python/tooling needs/wants these empty files. So, don't hold anything up because of me :)

arvidn (approved):

PR #10139, opened 2022-02-08
move ddl statements to separate sql directory/files

neurosis69:
I'd like to propose this change to separate the sqlite ddl statements from the other python code.
 
The idea is to

- have a "single point" where the schema definition is stored
- can be reused from other scripts, like db_upgrade_func without having the memory overhead during the upgrade
- easier to read the sql files (formatting) and quicker overview

Maybe a different directory structure would make sense, like one file per table/index and/or separated by version.


neurosis69:
Actually, in regards of the db_upgrade_func scripts, a separation into more files would definitely make sense.
At least one additonal for sub_epoch_segments_v3. Maybe one separate per Index.

altendky:
Sounds good.  Thanks!

We should use [`importlib.resources`](https://docs.python.org/3.10/library/importlib.html#module-importlib.resources) instead of `open(os.path.join(os.path.dirname(__file__), "a/path.sql"), "r")`.  Not the binary part, but, for example...

https://github.com/ericaltendorf/plotman/blob/9ba0c24166647940d7d03c2d587a80f9f62e0ea2/src/plotman/_tests/plotters/test_chianetwork.py#L75-L78
```python
    read_bytes = importlib.resources.read_binary(
        package=plotman._tests.resources,
        resource="chianetwork.marked",
    )
```

We may have to fiddle around with the package data settings in `setup.py` to make it get into the wheel etc.  I can take a look when the code is there if needed.  If I remember correctly it would be `include_package_data=True` (like [this](https://github.com/ericaltendorf/plotman/blob/9ba0c24166647940d7d03c2d587a80f9f62e0ea2/setup.cfg#L34) but in `setup.py` since we are a bit behind) and `MANIFEST.in` like [this](https://github.com/ericaltendorf/plotman/blob/9ba0c24166647940d7d03c2d587a80f9f62e0ea2/MANIFEST.in).  But we don't have a `MANIFEST.in` so...  we'll see.

neurosis69:
finally merged the pkg_resources commits and also split full_blocks and sub_epoch_segments_v3 DDLs into separate files.
I've chosen the name weight_store_tables_v2.sql as I was not really sure what this table is about.

neurosis69:
PR was rejected.

PR #10141, opened 2022-02-08
Crawler RPC

cmmarslender:
This PR adds an RPC server for the crawler (I used port 8561 - 1 higher than harvester RPC, and didn't seem to be in use for anything else here). In order to make this work properly with websockets, it also needed to start the daemon, so the whole thing has been reworked to start via the daemon like everything else, rather than its own process management strategy (This likely should have been done when the crawler/seeder was re-integrated into this repo, but it wasn't done then).

The integration with the daemon results in the following additional changes:
- There is no more `chia_seeder` command - its now just `chia start crawler` or `chia start seeder[-only]`
- All the configure commands that used to exist on the `chia_seeder configure` command have moved to the normal `chia configure` command


hoffmang9 (dismissed):
lgtm

chia/rpc/crawler_rpc_api.py
```diff
@@ -0,0 +1,51 @@
...
+            "/get_peer_counts": self.get_peer_counts,
+        }
+
+    async def _state_changed(self, change: str, change_data: Optional[Dict[str, Any]] = None) -> List[WsRpcMessage]:
+        payloads = []
+
+        if change_data is None:
+            change_data = await self.get_peer_counts({})
+
+        if change == "crawl_batch_completed" or change == "loaded_initial_peers":
```
Nikolaj-K, line 24: also equivalent to
```python
if change in ("crawl_batch_completed", "loaded_initial_peers"):
```

mariano54 (approved):
nice change :)

PR #10142, opened 2022-02-08
first pass at automatic db locking

altendky:


Oceanside22 (commented):
Great read me @ronerony22

PR #10143, opened 2022-02-08
Get root with status

almogdepaz:


chia/data_layer/dl_wallet_store.py
```diff
@@ -154,16 +154,23 @@ async def get_singleton_record(self, coin_id: bytes32) -> Optional[SingletonReco
...
+            # get latest confirmed root
+            cursor = await self.db_connection.execute(
+                "SELECT * from singleton_records WHERE launcher_id=? and confirmed = TRUE "
+                "ORDER BY generation DESC LIMIT 1",
+                (launcher_id,),
+            )
+        else:
+            cursor = await self.db_connection.execute(
+                "SELECT * from singleton_records WHERE launcher_id=? ORDER BY generation DESC LIMIT 1", (launcher_id,)
+            )
```
Nikolaj-K, line 173: If I read the diff correctly, you can cut the lines in half alla
```python
command = " ".join(
    ["SELECT * from singleton_records WHERE launcher_id=?"] + \
    (["and confirmed = TRUE"] if latest_confirmed else []) + \
    ["ORDER BY generation DESC LIMIT 1"]
)
cursor = await self.db_connection.execute(command, (launcher_id,))
```

emlowe:
@almogdepaz Some tests failing:

```
2022-02-08T22:54:53.922 full_node wallet_server           : ERROR    Exception: get_latest_singleton() missing 1 required positional argument: 'latest_confirmed', {'host': '::1', 'port': 50000}. Traceback (most recent call last):
22:54:53 wallet_server: ERROR Exception: get_latest_singleton() missing 1 required positional argument: 'latest_confirmed', {'host': '::1', 'port': 50000}. Traceback (most recent call last):
  File "/home/runner/work/chia-blockchain/chia-blockchain/chia/wallet/wallet_state_manager.py", line 869, in new_coin_state
    pool_state = solution_to_pool_state(launcher_spend)
  File "/home/runner/work/chia-blockchain/chia-blockchain/chia/pools/pool_puzzles.py", line 377, in solution_to_pool_state
    return pool_state_from_extra_data(extra_data)
  File "/home/runner/work/chia-blockchain/chia-blockchain/chia/pools/pool_puzzles.py", line 3[58](https://github.com/Chia-Network/chia-blockchain/runs/5117291036?check_suite_focus=true#step:13:58), in pool_state_from_extra_data
    for key, value in extra_data.as_python():
ValueError: too many values to unpack (expected 2)
```

altendky (dismissed):

chia/data_layer/dl_wallet_store.py
Nikolaj-K: Ye. Without string manipiulation one can still do
```python
if : 
    command = "foo"
else
    command = "bar"
cursor = await self.db_connection.execute(command, (launcher_id,))
```

chia/rpc/data_layer_rpc_api.py
```diff
@@ -159,8 +159,8 @@ async def get_root(self, request: Dict[str, Any]) -> Dict[str, Any]:
         # todo input checks
         if self.service is None:
             raise Exception("Data layer not created")
-        res = await self.service.get_root(store_id)
-        return {"hash": res}
+        hash, status = await self.service.get_root(store_id)
+        return {"hash": hash, "status": status}
```
Nikolaj-K, line 163: I had this thought yesterday, but maybe it's worth asking in light of `too many values to unpack` posted by @emlowe: Is there a convention in the code base w.r.t. getters called with `await` when they return pairs? Since when the result is is always a "maybe type", those returning `None` makes parsing pairs a bit trickier. 
(Given with how `get_root` is used I though it could make that function `return dict(hash=hash, status=status)` directly, and potentially fuse dicts with `update`. But also here I don't know all the conventions used in the codebase.)

altendky: We use tuples regularly.  I would personally rather see custom class instances to avoid the order dependence of tuples and the vagueness of dicts.  And yes, there is `TypedDict` but by the time you define one you've literally written a class and you could have used `@dataclass` instead and had a real class.  Lots of options are discussed at https://www.attrs.org/en/stable/why.html.  (dataclasses is a limited fork of attrs into the stdlib)

But, I'm trying to leave that for a dedicated discussion.  And sure, I often fail to keep quiet about it.  `:|`

chia/data_layer/data_layer.py
```diff
@@ -137,12 +137,12 @@ async def get_ancestors(self, node_hash: bytes32, store_id: bytes32) -> List[Int
             self.log.error("Failed to get ancestors")
         return res
 
-    async def get_root(self, store_id: bytes32) -> Optional[bytes32]:
-        res = await self.data_store.get_tree_root(tree_id=store_id)
-        if res is None:
+    async def get_root(self, store_id: bytes32) -> Tuple[Optional[bytes32], Status]:
```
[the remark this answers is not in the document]
almogdepaz: i want to return None with a confirmed status in case this store was created but root is None

chia/rpc/wallet_rpc_api.py
```diff
@@ -1397,7 +1397,12 @@ async def dl_latest_singleton(self, request) -> Dict:
 
         for _, wallet in self.service.wallet_state_manager.wallets.items():
             if WalletType(wallet.type()) == WalletType.DATA_LAYER:
-                record = await wallet.get_latest_singleton(bytes32.from_hexstr(request["launcher_id"]))
+                latest_confirmed = request.get("latest_confirmed")
+                if latest_confirmed is None:
+                    latest_confirmed = False
```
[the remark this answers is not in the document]
almogdepaz: this is less readable to my eyes

almogdepaz:
@altendky all methods in dl_wallet_store.py use "?" i rather we fix all of them (not in this pr)

chia/rpc/wallet_rpc_client.py
```diff
@@ -489,8 +489,8 @@ async def dl_stop_tracking(self, launcher_id: bytes32) -> None:
         await self.fetch("dl_stop_tracking", request)
         return None
 
-    async def dl_latest_singleton(self, launcher_id: bytes32) -> Optional[SingletonRecord]:
-        request = {"launcher_id": launcher_id.hex()}
+    async def dl_latest_singleton(self, launcher_id: bytes32, latest_confirmed) -> Optional[SingletonRecord]:
```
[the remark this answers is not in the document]
altendky: I guess actually in this case it's a pre-existing/non-atari file that I wasn't adding to the list.  Once we get https://github.com/Chia-Network/chia-blockchain/pull/10133 in we can start addressing a file at a time and then remove it from the less-strict list.

altendky (approved):

PR #10146, opened 2022-02-08
Ms.wallet refactor 

### LINKED ISSUE #10201: [enhancement] improve and optimize light wallet sync performance
Improve the reliability of the light wallet sync that enable faster sync times and better handling of wallets with large number of coins (~ up to 500) and medium # of transactions (~ up to 2000) and ensure accuracy of what is reported in the wallet.

Values are arbitrary and only used to convey approximate ranges.

mariano54:
This is a significant change in how syncing works in the wallet.  It is now organized as a queue where one of four things can get processed at once:
* New puzzle hash subscription
* New coin ID subscription
* new_peak_wallet message
* state_update message

The subscriptions are prioritized to ensure we don't miss data. State updates are cached by header hash (not height) and reorgs are now processed correctly.  Many different bugs and race conditions are fixed by this change. Also, there is a performance improvement in untrusted sync. 

A full node bug in coin store is also fixed (affected wallet protocol), where the min height was not being looked up correctly.

TODO:
- [x] fix flaky tests
- [x] Test on larger wallets, and test switching between trusted/untrusted
- [ ] Fix oom issue (also present in main)
- [ ] Potentially remove unnecessary fork_point argument


chia/full_node/coin_store.py
```diff
@@ -364,8 +364,7 @@ async def get_coin_state_by_ids(
         self,
         include_spent_coins: bool,
         coin_ids: List[bytes32],
-        start_height: uint32 = uint32(0),
-        end_height: uint32 = uint32((2 ** 32) - 1),
+        min_height: uint32 = uint32(0),
```
arvidn, line 367: I notice there's no unit test for this function, in `tests/core/full_node/test_coin_store.py`. It would be good to have one

mariano54: Thank you, yes that would be good considering it's in the coin store.

tests/core/server/test_rate_limits.py
```diff
@@ -23,11 +23,11 @@ async def test_too_many_messages(self):
         # Too many messages
         r = RateLimiter(incoming=True)
         new_tx_message = make_msg(ProtocolMessageTypes.new_transaction, bytes([1] * 40))
-        for i in range(3000):
+        for i in range(4900):
```
Yostra, line 26: Is this related to the rest of the changes in some way?

Yostra (approved):

chia/full_node/full_node.py
```diff
@@ -939,10 +939,10 @@ async def validate_block_batches(batch_queue):
                     await peer.close(600)
                     raise ValueError(f"Failed to validate block batch {start_height} to {end_height}")
                 self.log.info(f"Added blocks {start_height} to {end_height}")
-                await self.send_peak_to_wallets()
                 peak = self.blockchain.get_peak()
                 if len(coin_states) > 0 and fork_height is not None:
                     await self.update_wallets(peak.height, fork_height, peak.header_hash, coin_states)
+                await self.send_peak_to_wallets()
```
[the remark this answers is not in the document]
mariano54: yeah but only in the case where the full node is syncing up, which is not the usual case. It was wrong before, it should be first the new state then the new_peak_wallet

tests/core/server/test_rate_limits.py
mariano54: No it is not, it was just a flaky test that needed a fix.

PR #10153, opened 2022-02-09
Post a state change when an offer is added or soft-cancelled

paninaro:
More offer state changes may be needed in the future, but for now the GUI will be able to refresh wallet balances in response to offer soft-cancellation and addition. Without these updates, the GUI is currently unable to reasonably determine when to refresh wallet balances when an offer is created/soft-cancelled.


wjblanke (approved):
aok

PR #10156, opened 2022-02-09
store db ddl statements in separate directory

altendky:
Draft for:
- [ ] After testing this will be targeted against https://github.com/Chia-Network/chia-blockchain/pull/10139


arvidn:
this collides with https://github.com/Chia-Network/chia-blockchain/pull/10166 which I would prefer to land first.
Also, @mariano54 suggested we implement a relatively aggressive deprecation of v1. This becomes even nicer after we remove all the v1 code.

altendky:
replaced by https://github.com/Chia-Network/chia-blockchain/pull/10224

PR #9050, opened 2021-10-31
Ms.mempool locking

mariano54:
Many improvements with two goals:
- Improve block propagation and validation times, to make sure everyone can farm at all times, and blockchain health does not deteriorate in times of load,
- Improve how many transactions per second each node can accept, especially for low end hardware.

## Improvements
1. The Process pool for validating transactions has been increased from 1 to 2, to better utilize idle cores.
2. Validation of signatures for new transactions now happens in a different process (in the process pool where CLVM was getting validated). This means we can do more TPS and free the main thread for block validation.
3. For unfinished block validation, CLVM and signatures are validated outside the blockchain lock (and clvm in the process pool executor)
4. The mempool validation (which reads coin store) now uses the same lock as block validation (blockchain.lock). This means that block validation gets full priority of using IO and the main thread CPU. 
5. The blockchain lock has been changed to use a LockQueue, where different clients can use different lock priorities. Block validation gets high priority, and mempool low, so that block validation is always prioritized. 
6. A transaction queue is added so that we don't have to keep many long running asyncio tasks from the server, and can keep things in the queue for a while until the node has some free time to validate them. Things are pulled from the cache 200 at a time (with the semaphore) so that we can efficiently validate transactions in different processes. However, note that once a transaction has been selected from the queue, it still might need to wait for the blockchain lock later on. Before, with no queue, transactions would all try to get validated at once, even if the node could not handle them.
7. respond_transction in the full node protocol now returns instantly and puts TX in the queue.
8. send_transaction in the wallet protocol maintains it's behavior, but internally pushes transactions into the queue and waits for a result before returning.
9. CancelledException used to put the node into an invalid state, now it does not; we ensure to call `peak_post_processing` when tasks are cancelled.
10. peak_post_processing has been divided into two methods, `peak_post_processing` and `peak_post_processing_2`. The first one must be called under the blockchain lock, since it modifies full node state. The second one can be called outside the lock, and it deals with sending blocks and messages to other peers and wallets. Before, they were both called inside the lock.
11. When the mempool receives a new block, it must go through the whole mempool and re-check all transactions to see if they are still valid. WIth a new optimization, we now only check to see if the removals were removed in the block, if not, the spendbundle must be valid, so we don't check it.
12. New peaks can also stack up on each other in times of high load, so we limit the number of pending messages, and discard the rest. Also, only 2 run in parallel instead of 8, leading to faster catch-up times in RPI.
[...]

PR #10163, opened 2022-02-09
shutdown the weight proof process pool

altendky:
Sometimes when we shutdown with `chia stop -d all` we get 4 `chia_full_node` processes floating around unclosed.

```console
$ ps aux | grep chia_
altendky  591271 84.4  0.1 547924 124968 pts/2   S    09:58   0:20 chia_full_node
altendky  591284 22.2  0.2 634456 140436 pts/2   S    09:58   0:04 chia_full_node
altendky  591285 22.1  0.2 634200 140180 pts/2   S    09:58   0:04 chia_full_node
altendky  591286 22.5  0.2 634200 140180 pts/2   S    09:58   0:04 chia_full_node
altendky  591345  0.0  0.0   9688  2464 pts/2    S+   09:58   0:00 grep --color=auto chia_
```

Issues:
- [x] WeightProofHandler.validate_weight_proof had a bunch of sync code with no interleaved awaits resulting in a 4 second blocking of all other activity (as timed on my laptop).
  - Added several `await asyncio.sleep(0)` to take turns.  It is called _cooperative_ concurrency after all.
  - If the sync functions are well isolate it may make sense to push them into threads.  This both involves an await that will break the sync nature up at those points and also will revert to regular OS/threads/GIL sharing of CPU time while the sync work is being done.  But, this depends on the functions be thread safe, hopefully just fully isolated functional input/output with no global state dependence or modification.
- [x] The executor was not being closed.
  - Added a context manager to close it.
- [x] Executor subprocesses continue until they are done with their tasks, there is no systematic way to cancel them.
  - In this case the code in the workers had already been added to trigger cancellation via a temp file.  It just needed to be leveraged from this second use of the worker functions.
- [x] The full node sync task was being cancelled late.
  - Moved the cancellation from `FullNode._await_closed()` to `FullNode._close()`
- [x] The full node sync task was not being awaited.
  - Added awaiting of it in `FullNode._await_closed()` along with consuming the cancellation at that point since we are already shutting down there.
- [x] Service start signal handler registration was using the `signal.signal()` function instead of integrating itself with asyncio signal handling via `asyncio.get_running_loop().add_signal_handler()`.  It seems this would end up either overriding other signal handlers or being overridden by them depending on order of execution.  Neither seem good.
  - Switched.
- [ ] Various other shutdown cleanup will be submitted in separate PRs

Draft for:
- [x] Self reflection
  - :mirror: :man_astronaut: 
- [ ] Testing
  - Left it cycling overnight with no failures.  It's not perfect, but a lot better in the test sequence I'm exposing it to.
  - `set -vx; while true; do echo "about to start $(date --iso-8601=n)"; which chia; chia start node; sleep 30; echo "about to stop $(date --iso-8601=n)"; chia stop -d all; sleep 5; ps aux | grep chia_; pkill -9 chia_; done`
- [x] We should explicitly try to shutdown, but can we also adjust a setting so the processes would naturally die as well?
  - Leveraged the existing file-based shutdown approach already implemented for the wallet weight proof pool shtudown.
- [x] Consider an attribute and reusing the same pool across multiple calls
  - Mariano said this was used infrequently enough that just recreating the pool is ok.  Creating it looks to take about 0.1s, that is specifically from right before the context manager to inside of it.  There may be other delays induced later that I didn't isolate.

### DIFF
--- a/chia/full_node/weight_proof.py
+++ b/chia/full_node/weight_proof.py
@@ -600,7 +600,13 @@ class WeightProofHandler:
         peak_height = weight_proof.recent_chain_data[-1].reward_chain_block.height
         log.info(f"validate weight proof peak height {peak_height}")
 
+        # TODO: Consider if this can be spun off to a thread as an alternative to
+        #       sprinkling async sleeps around.
+
+        # timing reference: start
         summaries, sub_epoch_weight_list = _validate_sub_epoch_summaries(self.constants, weight_proof)
+        await asyncio.sleep(0)  # break up otherwise multi-second sync code
+        # timing reference: 1 second
         if summaries is None:
             log.error("weight proof failed sub epoch data validation")
             return False, uint32(0), []
@@ -611,12 +617,17 @@ class WeightProofHandler:
             log.error("failed weight proof sub epoch sample validation")
             return False, uint32(0), []
 
+        # timing reference: 1 second
         with ProcessPoolExecutor(6) as executor:
             try:
+                await asyncio.sleep(0)  # break up otherwise multi-second sync code
+                # timing reference: 1.1 second
                 constants, summary_bytes, wp_segment_bytes, wp_recent_chain_bytes = vars_to_bytes(
                     self.constants, summaries, weight_proof
                 )
+                await asyncio.sleep(0)  # break up otherwise multi-second sync code
 
+                # timing reference: 2 second
                 recent_blocks_validation_task = asyncio.get_running_loop().run_in_executor(
                     executor,
                     _validate_recent_blocks,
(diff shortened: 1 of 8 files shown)


mariano54:
This code was recently changes as part of the new wallet, and it did not get much testing. So it makes sense that there might be a bug here

chia/full_node/weight_proof.py
```diff
@@ -605,43 +605,43 @@ async def validate_weight_proof(self, weight_proof: WeightProof) -> Tuple[bool,
...
-        if not segments_validated:
-            return False, uint32(0), []
-
-        vdf_chunks = chunks(vdfs_to_validate, self._num_processes)
-        vdf_tasks = []
-        for chunk in vdf_chunks:
-            byte_chunks = []
-            for vdf_proof, classgroup, vdf_info in chunk:
-                byte_chunks.append((bytes(vdf_proof), bytes(classgroup), bytes(vdf_info)))
+        with ProcessPoolExecutor(self._num_processes) as executor:
```
arvidn, line 608: do we spin up these processes every time we validate a weight proof? It seems like we should just keep this `ProcessPoolExecutor` around, but maybe we only validate weight proofs very rarely.

Either way, probably not for this PR

arvidn (dismissed):

chia/full_node/weight_proof.py
altendky: Yeah, I asked about this and Mariano said it was fine to just recreate the pool.

altendky:
Thanks for looking.  There's a 'bunch' more coming on this.  It's basically the same situation as with the weight proof process pool in https://github.com/Chia-Network/chia-blockchain/pull/9050.  This will get another temp file to trigger shutdown of the work in the subprocesses since the executor can't handle such cancellation.

mariano54 (approved):

PR #10164, opened 2022-02-09
Add automatic rebasing and conflict handling to DL Wallet

Quexington:
This patch adds the ability for the DL Wallet to detect that a conflicting update happened while it was waiting for its own update.  One half of this is that it can now properly delete its pending state if something like a report spend comes through.  The other half is that it will attempt to automatically try its spend again with the new coin IFF the root never changed in any of the new conflicting history.

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -2,6 +2,7 @@ import logging
 import json
 import time
 import dataclasses
+from operator import attrgetter
 from typing import Any, Optional, Tuple, Set, List, Dict, Type, TypeVar
 
 from blspy import G2Element
@@ -486,7 +487,7 @@ class DataLayerWallet:
             created_at_time=uint64(int(time.time())),
             to_puzzle_hash=next_inner_puzzle.get_tree_hash(),
             amount=uint64(singleton_record.lineage_proof.amount),
-            fee_amount=uint64(0),
+            fee_amount=fee,
             confirmed=False,
             sent=uint32(10),
             spend_bundle=spend_bundle,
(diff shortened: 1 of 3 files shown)


chia/data_layer/data_layer_wallet.py
```diff
@@ -736,6 +737,78 @@ async def singleton_removed(self, parent_spend: CoinSpend, height: uint32) -> No
...
+        for singleton in unconfirmed_singletons:
+            await self.wallet_state_manager.dl_store.delete_singleton_record(singleton.coin_id)
+
+        if not root_changed:
+            # The root never changed so let's attempt a rebase
+            all_txs: List[TransactionRecord] = []
+            try:
+                for singleton in unconfirmed_singletons:
+                    for tx in relevant_dl_txs:
+                        if {singleton.coin_id} & set([c.name() for c in tx.additions]):
```
altendky, line 799:
```suggestion
                        if any(c.name() == singleton.coin_id for c in tx.additions):
```

chia/data_layer/data_layer_wallet.py
```diff
@@ -736,6 +737,78 @@ async def singleton_removed(self, parent_spend: CoinSpend, height: uint32) -> No
...
+                            if tx.spend_bundle is not None:
+                                fee = uint64(tx.spend_bundle.fees())
+                            else:
+                                fee = uint64(0)
+
+                            all_txs.extend(await self.create_update_state_spend(launcher_id, singleton.root, fee))
+            except Exception:
```
altendky, line 806: I don't really like that this cleanup might not happen in case of `asyncio.CancellationError` or `SystemExit`, but I'm not entirely sure at the moment what would be proper.  It could be `BaseException` with a `raise`.  We definitely shouldn't silently consume any exception though.  It should at least get logged.

chia/data_layer/data_layer_wallet.py
```diff
@@ -736,6 +737,78 @@ async def singleton_removed(self, parent_spend: CoinSpend, height: uint32) -> No
...
+                # Something went wrong so let's delete anything pending that was created
+                for singleton in unconfirmed_singletons:
+                    await self.wallet_state_manager.dl_store.delete_singleton_record(singleton.coin_id)
+            for tx in all_txs:
+                await self.wallet_state_manager.add_pending_transaction(tx)
```
altendky, line 811: Should this really be run when there was an exception in the `try` above?  Or maybe the `except:` block needs a `raise`.

chia/data_layer/dl_wallet_store.py
```diff
@@ -170,6 +170,22 @@ async def get_latest_singleton(self, launcher_id: bytes32) -> Optional[Singleton
...
+        """
+        cursor = await self.db_connection.execute(
+            "SELECT * from singleton_records WHERE launcher_id=? AND confirmed=0", (launcher_id,)
+        )
+        rows = await cursor.fetchall()
+        await cursor.close()
+        records = []
+
+        for row in rows:
+            records.append(self._row_to_singleton_record(row))
```
altendky, line 185:
```suggestion
        records = [self._row_to_singleton_record(row) for row in rows]
```

altendky (changes requested):
Overall for this, is removing the transactions entirely just a wallet state cleanup step?  As opposed to any of the steps being something we want to do quickly to make sure that we don't accidentally get something on chain that we don't want.  I guess the singleton spend should be invalid already so it would be cleanup, but could the fees still be 'live' in some cases such that they should get cancelled as soon as possible?

chia/data_layer/data_layer_wallet.py
```diff
@@ -736,6 +737,78 @@ async def singleton_removed(self, parent_spend: CoinSpend, height: uint32) -> No
...
+            root_changed = False
+
+        # Regardless of whether the root changed or not, our old state is bad so let's eliminate it
+        # First let's find all of our txs matching our unconfirmed singletons
+        unconfirmed_ids: Set[bytes32] = {s.lineage_proof.parent_name for s in unconfirmed_singletons}
+        relevant_dl_txs: List[TransactionRecord] = []
+        for id in unconfirmed_ids:
+            tx = await self.wallet_state_manager.tx_store.get_transaction_record(id)
+            if tx is not None:
+                relevant_dl_txs.append(tx)
```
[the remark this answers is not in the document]
Quexington: If it's created by the DL wallet it is yes.  You can see all of the places we build TransactionRecords in the same file.  I don't think the deletion order matters, all of those txs are based on an invalid history now so they all have to go.

chia/data_layer/data_layer_wallet.py
```diff
@@ -736,6 +737,78 @@ async def singleton_removed(self, parent_spend: CoinSpend, height: uint32) -> No
...
+        """
+        unconfirmed_singletons = await self.wallet_state_manager.dl_store.get_unconfirmed_singletons(launcher_id)
+        if len(unconfirmed_singletons) == 0:
+            return
+        unconfirmed_singletons = sorted(unconfirmed_singletons, key=attrgetter("generation"))
+        full_branch: List[SingletonRecord] = await self.wallet_state_manager.dl_store.get_all_singletons_for_launcher(
+            launcher_id,
+            min_generation=unconfirmed_singletons[0].generation,
+        )
+        if len(unconfirmed_singletons) == len(full_branch) and set(unconfirmed_singletons) == set(full_branch):
```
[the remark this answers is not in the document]
Quexington: A bit embarrassing but I don't particularly remember what this is supposed to check but I remember I mostly did not have it and then added it somewhere near the end so I _think_ it's important although I'm unable to fully explain it.

chia/data_layer/data_layer_wallet.py
```diff
@@ -736,6 +737,78 @@ async def singleton_removed(self, parent_spend: CoinSpend, height: uint32) -> No
...
+            return
+
+        # Now we have detected a fork so we should check whether the root changed at all
+        parent_singleton = await self.wallet_state_manager.dl_store.get_singleton_record(
+            unconfirmed_singletons[0].lineage_proof.parent_name
+        )
+        if parent_singleton is None or {parent_singleton.root} != {s.root for s in full_branch if s.confirmed}:
```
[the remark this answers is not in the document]
Quexington: I think this may be a mypy thing because we should always have a parent.  This is also helpful during reorgs I think in which the branch isn't even based on a history that exists anymore and therefore has no parent.

chia/data_layer/data_layer_wallet.py
Quexington: I'll add a log statement :+1:

altendky (approved):

PR #10166, opened 2022-02-09
new DBWrapper supporting concurrent readers

arvidn:
This is derived from work @altendky did, which initially was looking into each task using its own database connection.

This is primarily intended to be a correctness fix. This ensures that only one task at a time is *writing* to the database. Currently, we only acquire the database mutex in some situations, not consistently. This means that one task may have initiated a transaction, while another task is may insert unrelated writes into the transaction (when switched to). Additionally, any other task may insert *reads* into the transactions. This patch also ensures that reads are done independently, and outside of transactions.

In addition to there may be a performance improvements by allowing multiple readers to access the database concurrently. Specifically, lookups like `chia wallet show` or `chia farmer summary` no longer should need to wait for the database mutex to return.

With this patch, all writes are serialized with a mutex, and enforced by protecting the `aiosqlite.Connection` object. It can only be acquired via an API that also:

1. acquires the write mutex
2. initiates a transaction
3. roll backs the transaction on failure
4. commits the transaction when done

It would be possible to support *concurrent* writers, but it would be quite an undertaking. In that case, each write must support failing and retrying. This patch takes a smaller step by implementing:

1. Controlled access to a pool of `Connection` objects.
2. You can only write using the `DBWrapper` context manager, which automatically acquires the write-mutex, starts and commits a transaction (*)
3. You can read using a different context manager, giving you a read-only `Connection` object (attempting to write is an error) and multiple tasks can read concurrently.
4. The nesting behavior is carefully considered. If a task is requesting to read, but already holds the write mutex (and is the current writer) it gets the writer `Connection`, otherwise it wouldn't read uncommitted changes. A task that already has a reader `Connection` assigned to it, will get that same object when requesting to read.

(\*) the fine-print on the transactions is that they technically are sqlite "savepoints", which are like transactions except you're allowed to nest them. Only the outer-most transaction will actually commit, but nested transactions can fail and roll back individually.

# scope

The new class is called `DBWrapper2` to limit the scope of the patch. The wallet also uses `DBWrapper` so I'm hoping to keep this to the full-node only (for now).

# thread pool

Keep in mind that an `aiosqlite.Connection` object is a *thread* that calls the blocking sqlite functions. Therefore, it would be quite expensive to spawn a thread for *every* task (we create tasks for every network request). This patch effectively implements a thread pool of 4 reader threads and 1 writer thread. It would probably make sense to make the number of reader threads configurable. I have not attempted to tune the thread pool size, 4 just seemed like a reasonable number.

# review

The main change is the interface to `DBWrapper` (called `DBWrapper2`), all other changes are consequences of it. It might be easiest to start with understanding the new API. [Here](https://github.com/Chia-Network/chia-blockchain/pull/10166/files?diff=split&w=1#diff-70ec43bc92826e3bab2e1ae6a4c59a21d96a369430ea600f94c903568d47841dR37).

Given that there's now a context manager to access the database, there's a lot of indentation changes. It should be [reviewed ignoring white space](https://github.com/Chia-Network/chia-blockchain/pull/10166/files?diff=split&w=1).

### DIFF
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -1,7 +1,6 @@
 import logging
 from typing import Dict, List, Optional, Tuple, Any
 
-import aiosqlite
 import zstd
 
 from chia.consensus.block_record import BlockRecord
@@ -10,7 +9,7 @@ from chia.types.full_block import FullBlock
 from chia.types.blockchain_format.program import SerializedProgram
 from chia.types.weight_proof import SubEpochChallengeSegment, SubEpochSegments
 from chia.util.errors import Err
-from chia.util.db_wrapper import DBWrapper
+from chia.util.db_wrapper import DBWrapper2
 from chia.util.ints import uint32
 from chia.util.lru_cache import LRUCache
 from chia.util.full_block_utils import generator_from_block
@@ -19,94 +18,92 @@ log = logging.getLogger(__name__)
 
 
 class BlockStore:
-    db: aiosqlite.Connection
     block_cache: LRUCache
-    db_wrapper: DBWrapper
+    db_wrapper: DBWrapper2
     ses_challenge_cache: LRUCache
 
     @classmethod
-    async def create(cls, db_wrapper: DBWrapper):
+    async def create(cls, db_wrapper: DBWrapper2):
         self = cls()
-
         # All full blocks which have been added to the blockchain. Header_hash -> block
         self.db_wrapper = db_wrapper
-        self.db = db_wrapper.db
 
-        if self.db_wrapper.db_version == 2:
+        async with self.db_wrapper.write_db() as conn:
 
-            # TODO: most data in block is duplicated in block_record. The only
-            # reason for this is that our parsing of a FullBlock is so slow,
-            # it's faster to store duplicate data to parse less when we just
-            # need the BlockRecord. Once we fix the parsing (and data structure)
-            # of FullBlock, this can use less space
-            await self.db.execute(
-                "CREATE TABLE IF NOT EXISTS full_blocks("
-                "header_hash blob PRIMARY KEY,"
-                "prev_hash blob,"
-                "height bigint,"
-                "sub_epoch_summary blob,"
-                "is_fully_compactified tinyint,"
-                "in_main_chain tinyint,"
-                "block blob,"
-                "block_record blob)"
-            )
+            if self.db_wrapper.db_version == 2:
 
-            # This is a single-row table containing the hash of the current
-            # peak. The "key" field is there to make update statements simple
-            await self.db.execute("CREATE TABLE IF NOT EXISTS current_peak(key int PRIMARY KEY, hash blob)")
+                # TODO: most data in block is duplicated in block_record. The only
+                # reason for this is that our parsing of a FullBlock is so slow,
+                # it's faster to store duplicate data to parse less when we just
+                # need the BlockRecord. Once we fix the parsing (and data structure)
+                # of FullBlock, this can use less space
+                await conn.execute(
+                    "CREATE TABLE IF NOT EXISTS full_blocks("
+                    "header_hash blob PRIMARY KEY,"
+                    "prev_hash blob,"
+                    "height bigint,"
+                    "sub_epoch_summary blob,"
+                    "is_fully_compactified tinyint,"
+                    "in_main_chain tinyint,"
+                    "block blob,"
+                    "block_record blob)"
+                )
 
-            # If any of these indices are altered, they should also be altered
-            # in the chia/cmds/db_upgrade.py file
-            await self.db.execute("CREATE INDEX IF NOT EXISTS height on full_blocks(height)")
+                # This is a single-row table containing the hash of the current
+                # peak. The "key" field is there to make update statements simple
+                await conn.execute("CREATE TABLE IF NOT EXISTS current_peak(key int PRIMARY KEY, hash blob)")
 
-            # Sub epoch segments for weight proofs
-            await self.db.execute(
-                "CREATE TABLE IF NOT EXISTS sub_epoch_segments_v3("
-                "ses_block_hash blob PRIMARY KEY,"
-                "challenge_segments blob)"
-            )
+                # If any of these indices are altered, they should also be altered
+                # in the chia/cmds/db_upgrade.py file
+                await conn.execute("CREATE INDEX IF NOT EXISTS height on full_blocks(height)")
 
-            # If any of these indices are altered, they should also be altered
-            # in the chia/cmds/db_upgrade.py file
-            await self.db.execute(
-                "CREATE INDEX IF NOT EXISTS is_fully_compactified ON"
-                " full_blocks(is_fully_compactified, in_main_chain) WHERE in_main_chain=1"
-            )
-            await self.db.execute(
-                "CREATE INDEX IF NOT EXISTS main_chain ON full_blocks(height, in_main_chain) WHERE in_main_chain=1"
-            )
+                # Sub epoch segments for weight proofs
+                await conn.execute(
+                    "CREATE TABLE IF NOT EXISTS sub_epoch_segments_v3("
+                    "ses_block_hash blob PRIMARY KEY,"
+                    "challenge_segments blob)"
+                )
 
-        else:
+                # If any of these indices are altered, they should also be altered
+                # in the chia/cmds/db_upgrade.py file
+                await conn.execute(
+                    "CREATE INDEX IF NOT EXISTS is_fully_compactified ON"
+                    " full_blocks(is_fully_compactified, in_main_chain) WHERE in_main_chain=1"
+                )
+                await conn.execute(
+                    "CREATE INDEX IF NOT EXISTS main_chain ON full_blocks(height, in_main_chain) WHERE in_main_chain=1"
+                )
 
-            await self.db.execute(
-                "CREATE TABLE IF NOT EXISTS full_blocks(header_hash text PRIMARY KEY, height bigint,"
-                "  is_block tinyint, is_fully_compactified tinyint, block blob)"
-            )
+            else:
 
-            # Block records
-            await self.db.execute(
-                "CREATE TABLE IF NOT EXISTS block_records(header_hash "
-                "text PRIMARY KEY, prev_hash text, height bigint,"
-                "block blob, sub_epoch_summary blob, is_peak tinyint, is_block tinyint)"
-            )
+                await conn.execute(
+                    "CREATE TABLE IF NOT EXISTS full_blocks(header_hash text PRIMARY KEY, height bigint,"
+                    "  is_block tinyint, is_fully_compactified tinyint, block blob)"
+                )
 
-            # Sub epoch segments for weight proofs
-            await self.db.execute(
-                "CREATE TABLE IF NOT EXISTS sub_epoch_segments_v3(ses_block_hash text PRIMARY KEY,"
-                "challenge_segments blob)"
-            )
+                # Block records
+                await conn.execute(
+                    "CREATE TABLE IF NOT EXISTS block_records(header_hash "
+                    "text PRIMARY KEY, prev_hash text, height bigint,"
+                    "block blob, sub_epoch_summary blob, is_peak tinyint, is_block tinyint)"
+                )
 
-            # Height index so we can look up in order of height for sync purposes
-            await self.db.execute("CREATE INDEX IF NOT EXISTS full_block_height on full_blocks(height)")
-            await self.db.execute(
-                "CREATE INDEX IF NOT EXISTS is_fully_compactified on full_blocks(is_fully_compactified)"
-            )
+                # Sub epoch segments for weight proofs
+                await conn.execute(
+                    "CREATE TABLE IF NOT EXISTS sub_epoch_segments_v3(ses_block_hash text PRIMARY KEY,"
+                    "challenge_segments blob)"
+                )
 
-            await self.db.execute("CREATE INDEX IF NOT EXISTS height on block_records(height)")
+                # Height index so we can look up in order of height for sync purposes
+                await conn.execute("CREATE INDEX IF NOT EXISTS full_block_height on full_blocks(height)")
+                await conn.execute(
+                    "CREATE INDEX IF NOT EXISTS is_fully_compactified on full_blocks(is_fully_compactified)"
+                )
 
-            await self.db.execute("CREATE INDEX IF NOT EXISTS peak on block_records(is_peak)")
+                await conn.execute("CREATE INDEX IF NOT EXISTS height on block_records(height)")
 
-        await self.db.commit()
+                await conn.execute("CREATE INDEX IF NOT EXISTS peak on block_records(is_peak)")
+
         self.block_cache = LRUCache(1000)
         self.ses_challenge_cache = LRUCache(50)
         return self
@@ -134,15 +131,17 @@ class BlockStore:
 
     async def rollback(self, height: int) -> None:
         if self.db_wrapper.db_version == 2:
-            await self.db.execute(
-                "UPDATE OR FAIL full_blocks SET in_main_chain=0 WHERE height>? AND in_main_chain=1", (height,)
-            )
+            async with self.db_wrapper.write_db() as conn:
+                await conn.execute(
+                    "UPDATE OR FAIL full_blocks SET in_main_chain=0 WHERE height>? AND in_main_chain=1", (height,)
+                )
 
     async def set_in_chain(self, header_hashes: List[Tuple[bytes32]]) -> None:
         if self.db_wrapper.db_version == 2:
-            await self.db.executemany(
-                "UPDATE OR FAIL full_blocks SET in_main_chain=1 WHERE header_hash=?", header_hashes
-            )
+            async with self.db_wrapper.write_db() as conn:
+                await conn.executemany(
+                    "UPDATE OR FAIL full_blocks SET in_main_chain=1 WHERE header_hash=?", header_hashes
+                )
 
     async def replace_proof(self, header_hash: bytes32, block: FullBlock) -> None:
 
@@ -156,14 +155,15 @@ class BlockStore:
 
         self.block_cache.put(header_hash, block)
 
-        await self.db.execute(
-            "UPDATE full_blocks SET block=?,is_fully_compactified=? WHERE header_hash=?",
-            (
-                block_bytes,
-                int(block.is_fully_compactified()),
-                self.maybe_to_hex(header_hash),
-            ),
-        )
+        async with self.db_wrapper.write_db() as conn:
+            await conn.execute(
+                "UPDATE full_blocks SET block=?,is_fully_compactified=? WHERE header_hash=?",
+                (
+                    block_bytes,
+                    int(block.is_fully_compactified()),
+                    self.maybe_to_hex(header_hash),
+                ),
+            )
 
     async def add_full_block(self, header_hash: bytes32, block: FullBlock, block_record: BlockRecord) -> None:
         self.block_cache.put(header_hash, block)
@@ -176,56 +176,57 @@ class BlockStore:
                 else bytes(block_record.sub_epoch_summary_included)
             )
 
-            await self.db.execute(
-                "INSERT OR IGNORE INTO full_blocks VALUES(?, ?, ?, ?, ?, ?, ?, ?)",
-                (
-                    header_hash,
-                    block.prev_header_hash,
-                    block.height,
-                    ses,
-                    int(block.is_fully_compactified()),
-                    False,  # in_main_chain
-                    self.compress(block),
-                    bytes(block_record),
-                ),
-            )
+            async with self.db_wrapper.write_db() as conn:
+                await conn.execute(
+                    "INSERT OR IGNORE INTO full_blocks VALUES(?, ?, ?, ?, ?, ?, ?, ?)",
+                    (
+                        header_hash,
+                        block.prev_header_hash,
+                        block.height,
+                        ses,
+                        int(block.is_fully_compactified()),
+                        False,  # in_main_chain
+                        self.compress(block),
+                        bytes(block_record),
+                    ),
+                )
 
         else:
-            await self.db.execute(
-                "INSERT OR IGNORE INTO full_blocks VALUES(?, ?, ?, ?, ?)",
-                (
-                    header_hash.hex(),
-                    block.height,
-                    int(block.is_transaction_block()),
-                    int(block.is_fully_compactified()),
-                    bytes(block),
-                ),
-            )
+            async with self.db_wrapper.write_db() as conn:
+                await conn.execute(
+                    "INSERT OR IGNORE INTO full_blocks VALUES(?, ?, ?, ?, ?)",
+                    (
+                        header_hash.hex(),
+                        block.height,
+                        int(block.is_transaction_block()),
+                        int(block.is_fully_compactified()),
+                        bytes(block),
+                    ),
+                )
 
-            await self.db.execute(
-                "INSERT OR IGNORE INTO block_records VALUES(?, ?, ?, ?,?, ?, ?)",
-                (
-                    header_hash.hex(),
-                    block.prev_header_hash.hex(),
-                    block.height,
-                    bytes(block_record),
-                    None
-                    if block_record.sub_epoch_summary_included is None
-                    else bytes(block_record.sub_epoch_summary_included),
-                    False,
-                    block.is_transaction_block(),
-                ),
-            )
+                await conn.execute(
+                    "INSERT OR IGNORE INTO block_records VALUES(?, ?, ?, ?,?, ?, ?)",
+                    (
+                        header_hash.hex(),
+                        block.prev_header_hash.hex(),
+                        block.height,
+                        bytes(block_record),
+                        None
+                        if block_record.sub_epoch_summary_included is None
+                        else bytes(block_record.sub_epoch_summary_included),
+                        False,
+                        block.is_transaction_block(),
+                    ),
+                )
 
     async def persist_sub_epoch_challenge_segments(
         self, ses_block_hash: bytes32, segments: List[SubEpochChallengeSegment]
     ) -> None:
-        async with self.db_wrapper.lock:
-            await self.db.execute(
+        async with self.db_wrapper.write_db() as conn:
+            await conn.execute(
                 "INSERT OR REPLACE INTO sub_epoch_segments_v3 VALUES(?, ?)",
                 (self.maybe_to_hex(ses_block_hash), bytes(SubEpochSegments(segments))),
             )
-            await self.db.commit()
 
     async def get_sub_epoch_challenge_segments(
         self,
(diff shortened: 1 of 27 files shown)


chia/util/db_wrapper.py
```diff
@@ -29,3 +33,96 @@ async def rollback_transaction(self):
...
+
+        if task in self._in_use:
+            yield self._in_use[task]
+        else:
+            c = await self._read_connections.get()
+            try:
+                # record our connection in this dict to allow nested calls in
+                # the same task to use the same connection
+                self._in_use[task] = c
+                yield c
```
altendky, line 125: Don't we still need a transaction of some form for multiple read statements to be taken from a single snapshot?

chia/util/db_wrapper.py
```diff
@@ -29,3 +33,96 @@ async def rollback_transaction(self):
...
+        # up.
+        task = asyncio.current_task()
+        assert task is not None
+
+        # if this task currently holds the write lock, use the same connection,
+        # so it can read back updates it has made to its transaction, even
+        # though it hasn't been comitted yet
+        if self._current_writer == task:
+            # we allow nesting writers within the same task
+            yield self._write_connection
```
altendky, line 114: This bypasses the read-only nature of the reader connection you would otherwise get.  Can that pragma be applied 'temporarily' while this is yielded here?

altendky (commented):
Not a full review, but a few thoughts here to start.

Should a write manager be allowed to be nested inside a read manager?  Or would that be a useful thing to fail on.

I think there should be some sort of cleansing of the connections between uses.

With the limit on connection counts, have you thought through deadlock hazards?

chia/util/db_wrapper.py
```diff
@@ -29,3 +33,96 @@ async def rollback_transaction(self):
...
+            self._num_read_connections -= 1
+        await self._write_connection.close()
+
+    @contextlib.asynccontextmanager
+    async def write_db(self) -> AsyncIterator[aiosqlite.Connection]:
+        task = asyncio.current_task()
+        assert task is not None
+        if self._current_writer == task:
+            # we allow nesting writers within the same task
+            yield self._write_connection
```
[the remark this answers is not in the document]
arvidn: good point, I will add that + a test

chia/util/db_wrapper.py
```diff
@@ -29,3 +32,96 @@ async def rollback_transaction(self):
...
+        self.db_version = db_version
+        self._num_read_connections = 0
+        self._in_use = {}
+        self._current_writer = None
+
+    async def close(self) -> None:
+        while self._num_read_connections > 0:
+            await (await self._read_connections.get()).close()
+            self._num_read_connections -= 1
+        await self._write_connection.close()
```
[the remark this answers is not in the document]
arvidn: closing the connection while it's being used, I'm pretty sure will result in a `SQLite` exception in the place that's using it. I would think that would suffice, especially if we make tests fail by such exceptions.

The idea here is that the user of `DBWrapper2` is responsible for awaiting all tasks it handed the DBWrapper to before it closes it.

chia/util/db_wrapper.py
```diff
@@ -1,7 +1,13 @@
+from __future__ import annotations
+
 import asyncio
+import contextlib
+from typing import AsyncIterator, Dict, Optional
 
 import aiosqlite
 
+savepoint_name: int = 0
```
arvidn, line 9: it just occurred to me that this might as well be a member of the `DBWrapper2` class

chia/util/db_wrapper.py
```diff
@@ -29,3 +33,96 @@ async def rollback_transaction(self):
...
+        task = asyncio.current_task()
+        assert task is not None
+        if self._current_writer == task:
+            # we allow nesting writers within the same task
+            yield self._write_connection
+            return
+
+        async with self._lock:
+
+            global savepoint_name
```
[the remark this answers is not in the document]
arvidn: I made it a member

chia/util/db_wrapper.py
arvidn: you're right. I'm a bit worried about inserting calls to `pragma query_only=1` and `pragma query_only=0` frequently though.

arvidn: maybe we should have a `test_db` or `strict_db` setting that would do this, as well as setting things like `pragma reverse_unordered_selects`

tests/core/full_node/ram_db.py
```diff
@@ -1,21 +1,24 @@
...
 from chia.full_node.hint_store import HintStore
-from chia.util.db_wrapper import DBWrapper
+from chia.util.db_wrapper import DBWrapper2
 
 
-async def create_ram_blockchain(consensus_constants: ConsensusConstants) -> Tuple[aiosqlite.Connection, Blockchain]:
-    connection = await aiosqlite.connect(":memory:")
-    db_wrapper = DBWrapper(connection)
+async def create_ram_blockchain(consensus_constants: ConsensusConstants) -> Tuple[DBWrapper2, Blockchain]:
+    uri = f"file:db_{random.randint(0, 99999999)}?mode=memory&cache=shared"
```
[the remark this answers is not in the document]
arvidn: this isn't actually creating a file. the `?mode=memory` means it's a memory-only database. The name created here is just to allow multiple connections to the same memory-only DB (otherwise each connection will create its own in-memory database)

arvidn:
> Should a write manager be allowed to be nested inside a read manager? Or would that be a useful thing to fail on.

I agree that it sounds a bit questionable, maybe we should disallow it. I think the current logic works, but it's questionable to allow one tasks to allocate both a reader connection *and* the writer connection. And after the write completes, the reader connection may not see the update (at least not if it's holding on to a cursor whose query was made before the write.

Reading up on this ([here](https://www.sqlite.org/lang_transaction.html)) I think a write transaction probably should be disallowed to be started within a read "transaction". Doing so may fail at the sqlite level, in case there's already a separate writer (but we already avoid this with the writer mutex).

arvidn:
> I think there should be some sort of cleansing of the connections between uses.

I would prefer a test mode that ensures that nobody leaves a connection in a bad state. Do you mean things like resetting possible settings to defaults?

As far as I know, we haven't had any problems with this, and now everyone shares a single connection.

> With the limit on connection counts, have you thought through deadlock hazards?

Not really. Given that there's only a single mutex here, and it's reentrant, all the typical cases are handled. the cases where you may end up in a deadlock is if tasks depend on each other *inside a transaction*, which I suppose could happen. I would hope it would be caught in code review though.

chia/util/db_wrapper.py
arvidn: I suppose it would be nice to offer this guarantee. I'm sure there are places that expects a consistent snapshot across multiple queries, but there may also be places that expect to see transactions immediately as they are committed.

On top of that, I'm not aware of a way to ask for a read-transaction in sqlite. As far as I can tell, a read-transaction is created (and closed) implicitly by a `SELECT` query.

I think, as well as disallowing nesting writes within reads, may require a more careful survey of all the places we use the database, before deciding what to do. (i.e. a follow-up PR)

mypy.ini
```diff
@@ -10,7 +10,7 @@ disallow_untyped_calls = True
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
 check_untyped_defs = True
-disallow_untyped_decorators = True
+disallow_untyped_decorators = False
```
arvidn, line 13: @altendky I couldn't figure out a way around this. It looks like either an issue in mypy or pytest.

Without this I get:

```
tests/core/util/test_db_wrapper.py:51: error: Untyped decorator makes function "test_concurrent_writers" untyped  [misc]
tests/core/util/test_db_wrapper.py:52: error: Untyped decorator makes function "test_concurrent_writers" untyped  [misc]
tests/core/util/test_db_wrapper.py:84: error: Untyped decorator makes function "test_writers_nests" untyped  [misc]
tests/core/util/test_db_wrapper.py:103: error: Untyped decorator makes function "test_partial_failure" untyped  [misc]
tests/core/util/test_db_wrapper.py:126: error: Untyped decorator makes function "test_readers_nests" untyped  [misc]
tests/core/util/test_db_wrapper.py:142: error: Untyped decorator makes function "test_readers_nests_writer" untyped  [misc]
tests/core/util/test_db_wrapper.py:158: error: Untyped decorator makes function "test_concurrent_readers" untyped  [misc]
tests/core/util/test_db_wrapper.py:159: error: Untyped decorator makes function "test_concurrent_readers" untyped  [misc]
tests/core/util/test_db_wrapper.py:188: error: Untyped decorator makes function "test_mixed_readers_writers" untyped  [misc]
tests/core/util/test_db_wrapper.py:189: error: Untyped decorator makes function "test_mixed_readers_writers" untyped  [misc]
Found 10 errors in 1 file (checked 532 source files)
```

tests/core/full_node/ram_db.py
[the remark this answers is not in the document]
altendky: Quite so, thanks.

altendky (commented):
A few tidbits along with a significant deconstruction refactor over at https://github.com/Chia-Network/chia-blockchain/pull/10717 that I would be curious for your feedback on.

chia/util/db_wrapper.py
altendky: https://www.sqlite.org/lang_transaction.html
> If a write statement occurs while a read transaction is active, then the read transaction is upgraded to a write transaction if possible.

I believe you just start a transaction and it is a read transaction until otherwise upgraded.

arvidn: I thought you were asking about seeing a consistent snapshot of the database for the duration of the context manager. i.e. let the user acquire a connection and then read multiple times from the database and have all results be consistent. My understanding of a read-transaction in sqlite, is that a single SELECT query is a read transaction, when it ends, another SELECT query may see an updated state of the tables.

chia/util/db_wrapper.py
```diff
@@ -29,3 +31,107 @@ async def rollback_transaction(self):
...
+
+    def _next_savepoint(self) -> str:
+        name = f"s{self._savepoint_name}"
+        self._savepoint_name += 1
+        return name
+
+    @contextlib.asynccontextmanager
+    async def write_db(self) -> AsyncIterator[aiosqlite.Connection]:
+        task = asyncio.current_task()
+        assert task is not None
```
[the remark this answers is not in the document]
altendky: It provides the information that is present in code in the output.  For example, I believe assertions while running from an installed copy won't show anything more than a file, line number, and the `AssertionError`.  This is similar to the example above from the console where there is no source to be displayed in the traceback.

Anyways, to be explicit, I won't be holding up this PR on this discussion.

chia/util/db_wrapper.py
altendky: That is what I was referring to, yes.  My understanding is that those are implicit read transactions that are created and closed when you haven't explicitly started a transaction prior to the `SELECT`.  Just the same as `INSERT` will create an implicit write transaction.

Yes, I was suggesting that the read connection should have savepoint handling just like the write connection so that when I acquire a connection and make two reads that they will be consistent with each other.  I both see your point about maybe some connections specifically don't want that and also I wonder if that's a questionable practice or at least something that is significantly less common.  I guess we may just have to survey the existing uses.  Maybe there would be one read context manager method for each mode.

As a detail, maybe the read connections wouldn't need the nesting behavior of savepoints, as opposed to just using a single transaction for the outermost request.  Or maybe it is simply enough once the savepoint is a context manager that we can just reuse that.

chia/util/db_wrapper.py
altendky: Aside from the pragma topic, I guess this indicates that we should force the practice of closing _all_ cursors.  Easiest way is to just require use of a context manager consistently for _all_ `.execute()` calls.

1. We could yield a custom 'connection' object that only offers context manager execute mode
2. That custom 'connection' could complain about `pragma` attempts
3. There could be direct `async with db_wrapper.write_execute()` and `read` methods for when you only want to run one

I'll put the pragma topic on my list of followup considerations.

altendky:
Here's a collection of outstanding topics to consider in more detail.  I do not expect them to be addressed prior to merging this PR.  Maybe this list would be good in the OP for quicker reference?  Next, back to reviewing the code directly.

Follow up opportunities:
- [ ] Explicitly handle nesting a write inside a read, maybe fail?
- [ ] Make reads nested in writes actually be protected with `pragma query_only=1`
- [ ] Check for dirty connections being returned and clean connections before reuse
  - Reset read-only and write
  - Abort all savepoints
  - Abort all transactions
  - ?
- [ ] Maybe close all connections on shutdown, even those that are loaned out
- [ ] Savepoints for read transactions
- [ ] Yield a custom connection object with limited functionality to encourage best practices
  - Only offer context managers for execution to guarantee cursor closure
  - Complain about attempts to use pragmas

tests/core/util/test_db_wrapper.py
```diff
@@ -0,0 +1,221 @@
...
+                    async with conn1.execute("SELECT value FROM counter") as cursor:
+                        assert await get_value(cursor) == 1337
+                    raise RuntimeError("failure within a sub-transaction")
+            except BaseException:
+                pass
+            async with conn1.execute("SELECT value FROM counter") as cursor:
+                value = await get_value(cursor)
+
+    # the write of 1337 failed, and was restored to 42
+    assert value == 42
```
altendky, line 123: I'm somewhat tempted to in general collect the values instead of asserting then assert at the end.  For example, with the existing `except BaseException: pass`, any non-syntax coding error could be made inside the `try:` and the test would still pass without having checked what it meant to.  Though, checking coverage in tests would catch some such cases.  None the less, sometimes a useful pattern to make tests less likely to false pass.

tests/core/util/test_db_wrapper.py
```diff
@@ -0,0 +1,221 @@
...
+            if acquire_outside:
+                await exit_stack.enter_async_context(db_wrapper.read_db())
+
+            tasks = []
+            values: List[int] = []
+            for index in range(concurrent_task_count):
+                task = asyncio.create_task(sum_counter(db_wrapper, values))
+                tasks.append(task)
+
+        await asyncio.wait_for(asyncio.gather(*tasks), timeout=None)
```
altendky, line 183: Set a timeout or remove the `.wait_for()`?

altendky (changes requested):
Some more stuff for the follow-up list, and a few minor tweaks for the tests.

chia/util/initial-config.yaml
```diff
@@ -332,6 +332,11 @@ full_node:
   #         the particular system we're running on. Defaults to "full".
   db_sync: "auto"
 
+  # the number of threads used to read from the blockchain database
+  # concurrently. There's always only 1 writer, but the number of readers is
+  # configurable
```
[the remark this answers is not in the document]
altendky: Ah, ok.  I was thinking from the usage direction, not the aiosqlite library implementation detail end.  I think they are both correct.  Perhaps both are worth expressing.

tests/core/util/test_db_wrapper.py
```diff
@@ -0,0 +1,221 @@
...
+        async with contextlib.AsyncExitStack() as exit_stack:
+            if acquire_outside:
+                await exit_stack.enter_async_context(db_wrapper.write_db())
+
+            tasks = []
+            for index in range(concurrent_task_count):
+                task = asyncio.create_task(increment_counter(db_wrapper))
+                tasks.append(task)
+
+        await asyncio.wait_for(asyncio.gather(*tasks), timeout=None)
```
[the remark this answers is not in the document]
altendky: I don't understand what the `wait_for()` call is doing that is useful.  `gather()` could be awaited directly, right?

chia/util/db_wrapper.py
arvidn: I'm not sure creating a savepoint will necessarily make multiple queries operate on the same "snapshot" of the db, until the save point is closed. Does it work like that?

chia/full_node/coin_store.py
```diff
@@ -468,10 +476,12 @@ async def _add_coin_records(self, records: List[CoinRecord]) -> None:
                         record.timestamp,
                     )
                 )
-            await self.coin_record_db.executemany(
-                "INSERT INTO coin_record VALUES(?, ?, ?, ?, ?, ?, ?, ?)",
-                values2,
-            )
+            if values2 != []:
```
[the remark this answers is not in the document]
arvidn: the only reason I preferred this is because it's shorter. I have no strong opinion, I'll change it

chia/full_node/full_node.py
```diff
@@ -758,7 +766,7 @@ def _close(self):
     async def _await_closed(self):
         for task_id, task in list(self.full_node_store.tx_fetch_tasks.items()):
             cancel_task_safe(task, self.log)
-        await self.connection.close()
+        await self.db_wrapper.close()
```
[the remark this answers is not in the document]
arvidn: yeah. This is consistent with the current behavior. Maybe it would make more sense if we would wrap the creation of the write connection too, in a create-function.

tests/core/util/test_db_wrapper.py
```diff
@@ -0,0 +1,221 @@
...
+            async with conn1.execute("SELECT value FROM counter") as cursor:
+                assert await get_value(cursor) == 42
+            try:
+                async with db_wrapper.write_db() as conn2:
+                    await conn2.execute("UPDATE counter SET value = 1337")
+                    async with conn1.execute("SELECT value FROM counter") as cursor:
+                        assert await get_value(cursor) == 1337
+                    raise RuntimeError("failure within a sub-transaction")
+            except BaseException:
+                pass
```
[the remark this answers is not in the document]
arvidn: I can make it catch `RuntimeError` instead, but if I don't catch it, the exception will fall through out of the test, and the test will fail.

tests/core/util/test_db_wrapper.py
arvidn: you're right. I'll fix this

chia/util/db_wrapper.py
```diff
@@ -10,12 +14,10 @@ class DBWrapper:
 
     db: aiosqlite.Connection
     lock: asyncio.Lock
-    db_version: int
 
     def __init__(self, connection: aiosqlite.Connection, db_version: int = 1):
```
mariano54, line 18: remove unused arg

chia/util/db_wrapper.py
```diff
@@ -29,3 +31,108 @@ async def rollback_transaction(self):
...
+        name = f"s{self._savepoint_name}"
+        self._savepoint_name += 1
+        return name
+
+    @contextlib.asynccontextmanager
+    async def write_db(self) -> AsyncIterator[aiosqlite.Connection]:
+        task = asyncio.current_task()
+        assert task is not None
+        if self._current_writer == task:
+            # we allow nesting writers within the same task
```
mariano54, line 78: why do we allow this?

mariano54: (not allowing this could increase safety and code readability)

chia/full_node/block_store.py
```diff
@@ -134,15 +131,17 @@ def maybe_decompress(self, block_bytes: bytes) -> FullBlock:
...
+                await conn.execute(
+                    "UPDATE OR FAIL full_blocks SET in_main_chain=0 WHERE height>? AND in_main_chain=1", (height,)
+                )
 
     async def set_in_chain(self, header_hashes: List[Tuple[bytes32]]) -> None:
         if self.db_wrapper.db_version == 2:
-            await self.db.executemany(
-                "UPDATE OR FAIL full_blocks SET in_main_chain=1 WHERE header_hash=?", header_hashes
-            )
+            async with self.db_wrapper.write_db() as conn:
```
mariano54, line 141: It's very important that this is called inside the transaction along with set_in_chain, etc etc. Is it OK that we are just re-acquiring the lock here? I would think it's safe to actually pass in the DBWrapper2 instead, and disallow nesting.

mariano54 (commented):
It's a great idea to get rid of the begin, commit, rollback pattern in the full node, this makes me feel much better. Are the tests passing, can we get those green as well?

chia/util/db_wrapper.py
arvidn: we do have a few places where we call functions while also holding a write-enabled connection. I can look them up

arvidn:
something happened in `main` recently to make this fail. I need to spend some time understanding why it's not working anymore

chia/full_node/block_store.py
mariano54: Yes, pass the Connection object.
What I am worried about is that this API kind of encourages people to call these methods like 'set_in_chain` and `rollback` without grabbing a Connection, which doesn't seem safe. Wouldn't it be better for this method to require Connection as argument?

arvidn: I agree. It would be nice to not have to rely on nested transactions at all. It would probably create some more clarity in the code as well. Maybe `BlockStore` shouldn't even have a db wrapper member at all.

I'm hesitant to pile that onto this PR though, it's already pretty big.

tests/core/util/test_db_wrapper.py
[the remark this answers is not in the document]
altendky: The `pytest.raises()` context manager consumes the exception.

chia/full_node/block_store.py
altendky: I think there is an opportunity to enhance the clarity of proper usage.  I think this does not cause any significant regression in that regard though.  Add it to the follow-up list I guess?  https://github.com/Chia-Network/chia-blockchain/pull/10166#issuecomment-1070867445

chia/full_node/full_node.py
```diff
@@ -164,23 +163,8 @@ async def _start(self):
         # These many respond_transaction tasks can be active at any point in time
         self.respond_transaction_semaphore = asyncio.Semaphore(200)
         # create the store (db) and full node instance
-        self.connection = await aiosqlite.connect(self.db_path)
-        await self.connection.execute("pragma journal_mode=wal")
-        db_sync = db_synchronous_on(self.config.get("db_sync", "auto"), self.db_path)
-        self.log.info(f"opening blockchain DB: synchronous={db_sync}")
-        await self.connection.execute("pragma synchronous={}".format(db_sync))
-
-        async with self.connection.execute(
```
arvidn, line 173: this was the part that made all the tests fail. I re-jiggered it a bit to do this after creating the DBWrapper and using it.

chia/full_node/full_node.py
```diff
@@ -192,11 +176,37 @@ def sql_trace_callback(req: str):
...
+                await c.set_trace_callback(sql_trace_callback)
+            await self.db_wrapper.add_connection(c)
 
-        db_version: int = await lookup_db_version(self.connection)
+        await (await db_connection.execute("pragma journal_mode=wal")).close()
+        db_sync = db_synchronous_on(self.config.get("db_sync", "auto"), self.db_path)
+        self.log.info(f"opening blockchain DB: synchronous={db_sync}")
+        await (await db_connection.execute("pragma synchronous={}".format(db_sync))).close()
+
+        if db_version != 2:
```
arvidn, line 195: this is the new "version" of setting the db version to 2 (for new files)

arvidn:
another thing for the follow-up list is to convert the wallet to use `DBWrapper2`.

mariano54:
I have no problems with this PR, I'm OK to merge it, but we should run some manual tests with syncing and using the node a bit before releasing. @altendky would you like to approve it?

altendky (dismissed):
While I'm really excited about this I will caveat that this should wait to be merged until we can expect a significant amount of test time.  This is a step towards improved robustness and ease of development, but it's also a big change.

arvidn:
we could also add tests. if you can think of cases that may not be covered

altendky (approved):

tests/core/util/test_db_wrapper.py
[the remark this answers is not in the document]
arvidn: good point. I think I had misunderstood `gather()`. it returns a future, like any other, that can be awaited. the only reason `wait_for()` exists is to be an `await` with a timeout, right?

arvidn:
I just noticed this failure:

```
$ python tools/test_full_sync.py run --db-version 2 ~/.chia/mainnet/db/blockchain_v2_mainnet.sqlite
Chia directory /tmp/tmp2km9et8d
Can't find private CA, creating a new one in /tmp/tmp2km9et8d to generate TLS certificates

height 1540384 16.13 blocks/s   closing full node
  File "/home/arvid/dev/2/chia-blockchain/tools/test_full_sync.py", line 145, in <module>
    main()
  File "/home/arvid/dev/2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/arvid/dev/2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/arvid/dev/2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/arvid/dev/2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/arvid/dev/2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/arvid/dev/2/chia-blockchain/tools/test_full_sync.py", line 125, in run
    asyncio.run(run_sync_test(Path(file), db_version, profile))
  File "/usr/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/arvid/dev/2/chia-blockchain/tools/test_full_sync.py", line 98, in run_sync_test
    success, advanced_peak, fork_height, coin_changes = await full_node.receive_block_batch(
  File "/home/arvid/dev/2/chia-blockchain/chia/full_node/full_node.py", line 1100, in receive_block_batch
    pre_validation_results: List[PreValidationResult] = await self.blockchain.pre_validate_blocks_multiprocessing(
  File "/home/arvid/dev/2/chia-blockchain/chia/consensus/blockchain.py", line 631, in pre_validate_blocks_multiprocessing
    return await pre_validate_blocks_multiprocessing(
  File "/home/arvid/dev/2/chia-blockchain/chia/consensus/multiprocess_validation.py", line 241, in pre_validate_blocks_multiprocessing
    challenge = get_block_challenge(constants, block, BlockCache(recent_blocks), prev_b is None, overflow, False)
  File "/home/arvid/dev/2/chia-blockchain/chia/consensus/get_block_challenge.py", line 91, in get_block_challenge
    curr: BlockRecord = blocks.block_record(header_block.prev_header_hash)
  File "/home/arvid/dev/2/chia-blockchain/chia/util/block_cache.py", line 35, in block_record
    return self._block_records[header_hash]
KeyError: <bytes32: 7c27699b939dbd9b026e872e13650901e38d2e9590934aed67ba516a0f1108a0>
```

I'll rerun it against `main` to ensure it's introduced by this PR or not. It needs fixing, either way, but it might not hold off landing this.

arvidn:
I encounter the exact same error in `main`. So I don't think it should hold up landing this PR.

emlowe:
This may be of interest since it appears to be similar issue?: https://github.com/Chia-Network/chia-blockchain/issues/10883

PR #10176, opened 2022-02-10
Adding encode.io/databases abstraction layer to add support for not only sqlite but also mysql and postgres

irulast-devops:
Refactored and removed transactions
- Transactions with encode.io/databases require a context block or reference to the transaction object being used. Removed the `DBWrapper` global transaction functionality and refactored transactions that used it.
- Redundant transactions surrounding single queries were removed 

Int casting for mysql
- Casting `uint32` to int as `pymysql` produces errors when passed a `int32`, `uint64`, ect...

Changed order of `WHERE IN` sql queries
- Rearranged queries using `WHERE IN` clauses so the parameter placeholder used is the last parameter placeholder because when `sqlalchemy` compiles the query and maps the parameters values to their placeholders it puts the values of a parameter with multiple values last regardless of the order of the placeholders. This causes a mismatch if the multiple-value placeholder is not last.

Increased sleep times
- Increased `asincio.sleep` used in tests to avoid errors

Removed of `set_trace_callback`
- Removed code using `aiosqlite.set_trace_callback` as databases does not support this functionality.

New files

dialect_utils.py
- `dialect_utils` is a module that is used to resolve the differences in syntax between different sql dialects (`mysql`, `postgres`, `sqlite`)

temp_file_db.py
- `encode.io/databases` does not support `sqlite` in-memory databases. A pull request is open with encode.io to add this functionality. In-memory databases are replaced with a tempfile database using logic extracted from `db_connection.py`.

db_factory.py
- Returns a database connection from `encode.io/databases`

tests/db/docker-compose-{database}.yml and README.md
- `docker-compose` files for `mysql` and `postgres` demonstrating full node operation against these databases


aweigold:
Reading through the discussion here.  I'm curious to know what kind of scalability concerns would need to be addressed by moving from a database to a flatfile.

In my mind, there are two main business usages for deployment of a full node, one to help drive local user operations, and one as a component in a more complex architecture that can bring new use cases to the ecosystem (such as a cloud based deployment utilizing the full node in a service oriented and/or microservices architecture).  

For the former, my opinion is that if such optimization is needed, that there may be greater issues with application and/or protocol design that could more readily solve any scalability concerns.  If there aren't scalability concerns that are tangible for local users that can only be addressed through flatfile datastores, then, as a matter of just my humble opinion, solving that problem doesn't tangibly bring much business value.

For the later, concurrent access to shared datastores is more important, as scalability can be addressed horizontally.  A traditional RDBMS already comes with decades of investment into managing fast concurrent access to data in block storage that I don't think could be addressed better by making a new system that will invariably bring headaches to deployment orchestration and operations.  This is also one of the drivers to this PR, as sqlite doesn't lend itself to redundant and concurrent deployment scenarios.  I don't believe replacing sqlite with a flatfile would make things any easier for deployment ops.

AndrewFarley:
> what is the rationale for supporting another database? In order to scale, I envision that we need to stop using a database entirely. The bulk of the blockchain should live in simple flat files. That would be very efficient.
> 
> There's a risk that this patch would make it more difficult to move in that direction.

So, I feel inclined to comment to this PR, especially to respond to this comment.  The bulk of any/every coin node's (not just Chia) engineering seems to be a result of this exact logic. Unfortunately, this is also why most coins are fragile and **extremely** non-redundant and non-scalable.  At every cryptocurrency exchange I've worked at, the only way to gain "redundancy" is for us to spend a ton of unnecessary CPU/RAM resources to run multiple 2-5 completely independent coin nodes, and then try to route customer traffic to these in an round-robin fashion.  And if we need to "scale up" to handle more traffic, we have to run, yet another full node, consuming yet an TON more CPU/RAM/networking resources.  Some coins allow us to specify manual peers so we can peer them to help prevent _some_ waste of networking traffic having to pull down blocks from the internet multiple times, but generally the coin nodes of the world thus far have largely ignored what is considered best-practice when engineering an highly scalable and durable service.

"Efficient" as you are trying to describe it, comes at **GREAT COST** of scalability and redundancy.  Flat-files do not scale beyond one server, ever.  No company out there that hosts anything on the internet will run only one server of any kind and hope everything works out.  Do you think Amazon, or Facebook, or Google, or any-company-ever runs only a single service which you use daily and don't realize when you use your phone or computer on the internet?  Cryptocurrency engineering seems to have ignored this best-practice because they can get away with it, because of the decentralized nature of it.

So, I believe the answer to your question of "what is the rationale" is this:  This PR is trying to pivot the foundation of the coin node to allow for nearly _any_ underlying storage system, and to allow for some future PRs where we support an near-infinite horizontal scale.  Let me expand on what that means below...

With this PR in place, it will be possible (not yet, however) for a company like mine to run one "primary coin" and as many "secondary (readonly) coin" nodes using the same database as I want for some coin.  This database will be stored in an highly-durable/available place (eg: AWS RDS), and almost all client traffic sent to the coin as long as it is deemed read traffic (which is an very high-percentage in my experience) will be shipped to the "secondary coin" nodes, which simply read from the database (and in higher-scale situations, only read from an "secondary" of the database actually).  And only a very small amount of traffic if someone using our services need to write to the "primary coin" node, will have their traffic routed there (and the "primary coin" node will be the only one with credentials to the write-capable user/database).  It will also be possible (with future PRs) to if our "primary coin" crashes, to _instantly_ promote one of the "secondary coins" to an "primary" to continue this service running with zero downtime (with an service that can buffer incoming requests) or with near-zero downtime.

I believe with this PR, and some future follow-up PRs to allow for this coin node to act in an "primary/secondary" fashion, Chia will be (one of) the first forward-thinking coin nodes engineered to support (infinite-ish) horizontal-scalability.

**Reference**: I'm an veteran high-scalability systems architect and also as someone who has been solely responsible for scaling crypto coins at numerous large companies (exchanges).

hoffmang9:
A couple of notes.

First, the reason to head towards flat files is around supporting low end hardware and allowing pruning nodes.

That said, the conversation we have been having internally is very cognizant that being able to support e.g. RDS is a real requirement that we need to address.

There is a history of blockchains having hidden consensus issues that were acquired from their database without expecting it so we are careful here - https://github.com/bitcoin/bips/blob/master/bip-0050.mediawiki

arvidn:
> Reading through the discussion here. I'm curious to know what kind of scalability concerns would need to be addressed by moving from a database to a flatfile.

The dimension I refer to scaling is the number of coin record updates per block. Specifically being able to farm blocks that are close to 100% of what the consensus rules allow and still have the vast majority of nodes be able to apply those updates to their chain in less than 10 seconds. If it takes much more than 10 seconds, you can easily miss opportunities to win the next block. You won't try to farm a block on top of a block you haven't validated yet.

Right now, blocks are generally filled pretty far from capacity, but that will change in the future. Applying block updates will need to get a lot faster.

> In my mind, there are two main business usages for deployment of a full node, one to help drive local user operations, and one as a component in a more complex architecture that can bring new use cases to the ecosystem (such as a cloud based deployment utilizing the full node in a service oriented and/or microservices architecture).

There's also securing the chain by validating and providing proof of space. i.e. farm.

> For the later, concurrent access to shared datastores is more important, as scalability can be addressed horizontally. A traditional RDBMS already comes with decades of investment into managing fast concurrent access to data in block storage that I don't think could be addressed better by making a new system that will invariably bring headaches to deployment orchestration and operations.

Filesystems also have decades of investment. They're actually quite fast these days. they also support concurrent access.

> I don't believe replacing sqlite with a flatfile would make things any easier for deployment ops.

I'm curious to understand why. machines can share drives, right? As long as you have a single writer, you can have any number of readers. Or you could replicate writes to a remote copy as well, to run queries against.

arvidn:
> At every cryptocurrency exchange I've worked at, the only way to gain "redundancy" is for us to spend a ton of unnecessary CPU/RAM resources to run multiple 2-5 completely independent coin nodes

I don't understand what the alternative is. If you need redundancy, you need to run redundant nodes. Also, I'm arguing for making each of those redundant nodes use less CPU, RAM and I/O. So, less waste.

> "Efficient" as you are trying to describe it, comes at GREAT COST of scalability and redundancy. Flat-files do not scale beyond one server, ever.

I don't buy that assertion. At the high frequency trading firm I worked at we used flat files extensively, not just locally, but across networks of machines. It was very efficient and had very low latency and very high throughput. My impression is that other trading firms also employ similar techniques.

> With this PR in place, it will be possible (not yet, however) for a company like mine to run one "primary coin" and as many "secondary (readonly) coin" nodes using the same database as I want for some coin.

That seems like a worthy goal. I don't think this approach is a prerequisite, and I also don't think using a more efficient store for the blockchain is a blocker for that goal. There would need to be a way to replicate updates to the local blockchain to one or more remote machines, which could then be used to query the state of the chain.

mariano54:
First of all, thank you for your interest and enthusiasm, and thank you for making a PR to bring your changes to the chia-blockchain repo.

Developing for a blockchain and cryptocurrency is pretty different from development of high scalability and horizontally scaled systems. The blockchain has had essentially no downtime since it was launched in march, and that is not a coincidence. 

**Horizontal scalability**
First of all, the network has a max capacity of approximately 25 transactions per second, and this will not change in the near or medium term. There is also the requirement that very low power nodes (RPI) should be able to keep up with the network and even sync up (download the entire history). In order for this to be possible, we need to optimize the code and the IO as much as we can. This also means that having super high power machines or even horizontal scalability does not provide much benefit. Of course for redundancy you want to run multiple machines that duplicate effort. Flat files can bring this TPS to even slow SSDs or HDDS as we optimize for our specific use case.

**Consensus bugs**
The most important thing for developing a blockchain is consensus safety. That is, if there are 10 nodes, these 10 nodes should agree on the consensus rules, and should not deviate. The reason for this is that blockchain development has serious consequences for consensus failures. If 30% and 70% of farming nodes disagree on a certain block or tx, then a chain fork can be created. This can mean millions or billions of dollars lost as people transact on a different blockchain and cryptocurrency, and steal from exchanges, and updating these machines is extremely difficult, as there are hundreds of thousands of nodes. This is not the case in traditional development, as broken machines can just be updated.
In order to achieve consensus safety, the system must be absolutely as simple as possible, and this includes limiting configuration as much as possible. More configuration options means more chances of a fork. For example, if we have v1 and v2, and 3 DB options (sqlite, mysql, postgres), we have 6 combinations (v1, sqlite), (v1, mysql), etc.. This increases chances of consensus issues. The ideal situation in my opinion is only 1 data store and version. Changing to flat files means we would not be subject to different DB versions and quirks in DB implementation for performance and critical bugs. We need to understand every single line of code that is being executed by the node. 

**Complexity** 
Another important requirement is keep complexity of the codebase as small and simple as possible. This is not just for consensus safety though, but for scalability of the codebase as the number of contributors increase. As you add more layers of abstraction and more DB types, you know have to understand more combinations of things, and changing the code becomes harder and harder. Supporting users that have issues becomes difficult as you always need to check their DB and DB version. Supporting developers that want to access DB manually, or that want to make changes to schema or SQL becomes harder. Every change takes longer and adds more risk.

My opinion is while these changes are great for certain use cases, they do not align with blockchain development best practices and ultimately would hinder progress and safety of the system.

hoffmang9:
Closing for staleness

PR #10179, opened 2022-02-10
farmer: Move some member instantiations into `Farmer.__init__`

xdustinface:
This fixes logs like below appearing if the GUI accesses the farmer RPC while no keys are loaded:

```
2022-02-10T07:50:47.445 farmer chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia/rpc/rpc_server.py", line 215, in safe_handle
  File "chia/rpc/rpc_server.py", line 206, in ws_api
  File "chia/rpc/farmer_rpc_api.py", line 126, in get_harvesters
  File "chia/farmer/farmer.py", line 705, in get_harvesters
  File "chia/farmer/farmer.py", line 691, in get_cached_harvesters
AttributeError: 'Farmer' object has no attribute 'harvester_cache'
```

### DIFF
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -128,6 +128,17 @@ class Farmer:
         self.started = False
         self.harvester_handshake_task: Optional[asyncio.Task] = None
 
+        # From p2_singleton_puzzle_hash to pool state dict
+        self.pool_state: Dict[bytes32, Dict] = {}
+
+        # From p2_singleton to auth PrivateKey
+        self.authentication_keys: Dict[bytes32, PrivateKey] = {}
+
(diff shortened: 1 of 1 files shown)


arvidn:
I take it `setup_keys()` isn't called multiple times then, right? (otherwise these members wouldn't be reset).

xdustinface:
> I take it `setup_keys()` isn't called multiple times then, right? (otherwise these members wouldn't be reset).

Correct. I mean, it is called multiple times but the part below https://github.com/Chia-Network/chia-blockchain/blob/main/chia/farmer/farmer.py#L152 isn't.

arvidn (approved):

PR #10189, opened 2022-02-11
::1

altendky:
GitHub Actions does support IPv6.  From my glance at https://github.com/actions/virtual-environments/issues/668 it is about _external_ IPv6 connections, not localhost.  Trio and Twisted both test IPv6 as best as both I and they know.  Twisted goes so far as to _disable_ IPv6 entirely for a dedicated IPv4-only test job.  If you want to look at a trivial example of IPv6 seemingly working, see https://github.com/altendky/altendpy/pull/15.

Draft for:
- [ ] deal with localhost
  - Wikipedia says ["The name localhost normally resolves to the IPv4 loopback address 127.0.0.1, and to the IPv6 loopback address ::1."](https://en.wikipedia.org/wiki/Localhost#Loopback) but they cite [RFC 4291](https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.3) which doesn't even mention `localhost` per my search.
  - On my laptop running Ubuntu 20.04, `localhost` does not resolve to an IPv6 unless I manually add it to `/etc/hosts`.  At least according to `nmap -6 -p 0-65535 localhost` which says `Warning: Hostname localhost resolves, but not to any IPv6 address. Try scanning without -6`.
- [ ] making this not a hack
- [ ] considering ux on the parallel `self_hostname` and `prefer_ipv6` configuration options
- [ ] learn what all `prefer_ipv6` affects
- [x] make a common ipv6 aware url builder (or `hyperlink`?)
- [ ] if listening on IPv6 should we allow v4 as well on the same socket?
- [ ] how do we want to setup tests for coverage with `prefer_ipv6` both true and false
- [ ] review all uses of `127.0.0.1` and `::1`

### DIFF
--- a/tests/core/daemon/test_daemon.py
+++ b/tests/core/daemon/test_daemon.py
@@ -58,7 +58,7 @@ class TestDaemon:
             yield get_b_tools
 
     @pytest.mark.asyncio
-    async def test_daemon_simulation(self, simulation, get_daemon, get_b_tools: BlockTools):
+    async def test_daemon_simulation(self, simulation, get_daemon, get_b_tools):
         node1, node2, _, _, _, _, _, _, _, _, server1 = simulation
         await server1.start_client(PeerInfo(self_hostname, uint16(21238)))
 
--- a/tests/core/daemon/test_daemon.py
+++ b/tests/core/daemon/test_daemon.py
@@ -75,8 +75,6 @@ class TestDaemon:
         raw_host = get_b_tools._config["self_hostname"]
         prefer_ipv6 = get_b_tools._config["prefer_ipv6"]
         host = get_host_addr(host=raw_host, prefer_ipv6=prefer_ipv6)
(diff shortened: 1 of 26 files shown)


tests/core/daemon/test_daemon.py
```diff
@@ -57,7 +58,7 @@ async def get_daemon_with_temp_keyring(self, get_b_tools):
             yield get_b_tools
 
     @pytest.mark.asyncio
-    async def test_daemon_simulation(self, simulation, get_daemon, get_b_tools):
+    async def test_daemon_simulation(self, simulation, get_daemon, get_b_tools: BlockTools):
```
Roxksteadye1101, line 61: xhTo

Roxksteadye1101: xhTo

Roxksteadye1101 (commented):
Need further details

PR #10192, opened 2022-02-11
Update version prop in package.json of gui

ChiaMineJP:
Screen Shot 2022-02-11 at 19 51 24 (1)

Fixed an issue where chia version showed in About dialog was out of sync with actual chia version.

### DIFF
--- a/installhelper.py
+++ b/installhelper.py
@@ -33,14 +33,13 @@ def make_semver(version_str: str) -> str:
         prerelease.append("".join(str(x) for x in v._version.pre))
     if v._version.dev:
         prerelease.append("".join(str(x) for x in v._version.dev))
-    prerelease = ".".join(prerelease)
 
     local = v.local
 
     version = "{0}.{1}.{2}".format(major, minor, patch)
 
     if prerelease:
-        version += "-{0}".format(prerelease)
+        version += "-{0}".format(".".join(prerelease))
     if local:
         version += "+{0}".format(local)
 
--- a/installhelper.py
+++ b/installhelper.py
@@ -6,6 +6,7 @@
 #
 import json
 import os
+from os.path import exists
 import subprocess
 from pkg_resources import parse_version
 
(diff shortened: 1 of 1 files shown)


installhelper.py
```diff
@@ -46,20 +47,27 @@ def make_semver(version_str):
...
-def update_version():
-    with open(f"{os.path.dirname(__file__)}/chia-blockchain-gui/package.json") as f:
-        data = json.load(f)
-
+def get_chia_version():
     version: str = "0.0"
     output = subprocess.run(["chia", "version"], capture_output=True)
     if output.returncode == 0:
         version = str(output.stdout.strip(), "utf-8").splitlines()[-1]
+    return make_semver(version)
```
arvidn, line 55: instead of launching a new `chia` process to ask for the version, wouldn't it work to just:
```
from chia import __version__
```
?

It's defined as:

```
from pkg_resources import get_distribution
__version__ = get_distribution("chia-blockchain").version
```

Which looks like it only works if the package is installed. I'm not sure where it gets its version number from, but it looks like there may be a risk of circular dependency, in case `get_distribution()` gets its version from `package.json`

ChiaMineJP: @arvidn Thank you to give me an opportunity to find out how version is determined.

I found the possible scheme of versioning below.
`chia-blockchain/.eggs/setuptools_scm-6.4.2-py3.9.egg/EGG-INFO/PKG-INFO`

```
Default versioning scheme
-------------------------

In the standard configuration ``setuptools_scm`` takes a look at three things:

1. latest tag (with a version number)
2. the distance to this tag (e.g. number of revisions since latest tag)
3. workdir state (e.g. uncommitted changes since latest tag)

and uses roughly the following logic to render the version:

no distance and clean:
    ``{tag}``
distance and clean:
    ``{next_version}.dev{distance}+{scm letter}{revision hash}``
no distance and not clean:
    ``{tag}+dYYYYMMDD``
distance and not clean:
    ``{next_version}.dev{distance}+{scm letter}{revision hash}.dYYYYMMDD``

The next version is calculated by adding ``1`` to the last numeric component of
the tag.
```

It seems version is picked from git tag and not `package.json`. Circular dependency is not going to  happen here.

And your suggestion seems correct and I was nearly adopting your idea but I suddenly asked myself:
`Q. what exact string should be shown on About dialog?`
`A. It should be exactly what "chia version" command shows to console`
What I want to tell here is that even if `__version__ = get_distribution("chia-blockchain").version` returns almost the same version string as `chia version` outputs, it doesn't mean it will return the same format/text as `chia version` in future.
If we use `__version__ = get_distribution("chia-blockchain").version` as a version string in About dialog, it means we should always check that format of `chia version` is not changed for every release.

arvidn: ok, seems reasonable. Do you intend to remove the call to `make_semver()` here then? to preserve the string verbatim?

installhelper.py
```diff
@@ -46,20 +47,27 @@ def make_semver(version_str):
     return version
 
 
-def update_version():
-    with open(f"{os.path.dirname(__file__)}/chia-blockchain-gui/package.json") as f:
-        data = json.load(f)
-
+def get_chia_version():
```
arvidn, line 50: would you mind adding a type hint for the return value here? I would imagine it's: ` -> str`

ChiaMineJP: It should be. Thank you!

installhelper.py
ChiaMineJP: Good point. I've noticed that `make_semver()` works like a function which takes for example `1.2.12.dev271` and returns `1.2.12-dev271`. (`.dev` -> `-dev`)
This clearly does not preserve the string verbatim.
Umm. 
What kind of version string should be on About dialog should be reviewed carefully and maybe discussed in another PR.

I'd like the intention of this PR to be simple.
That is, by merging this PR,  what the previous `installhelper.py` did to `chia-blockchain-gui/package.json` now also applies to `chia-blockchain-gui/packages/gui/package.json`.

emlowe: The problem is that package.json has rather strict requirements on the version format. For example, if you put 1.2.12.dev271 into package.json, this used to fail with NPM throwing some error. So the make_semver() function is to make sure that whatever goes into package.json is NPM compatible

emlowe: Note, the GUI picks up the version from package.json but the GUI is unique in this.

emlowe (approved):
Looks ok.

PR #10193, opened 2022-02-11
Db convert offline mode

arvidn:
I haven't had a chance to benchmark these new options yet. Converting a recent v1 database takes weeks on a slow HDD.

This is from one USB-attached HDD to another USB-attached HDD. The computer is otherwise relatively powerful.

Running:
```
time chia db upgrade --input /media/arvid/plots/blockchain_v1_modified.sqlite --output=/media/arvid/seagate/test-v2.sqlite --offline --vacuum
```

```
opening file for reading: /media/arvid/plots/blockchain_v1_modified.sqlite in offline mode
vacuuming input database
      52469.13 seconds
opening file for writing: /media/arvid/seagate/test-v2.sqlite
initializing v2 version
initializing v2 block store
peak: 56e4b442dba0f58c6489991ea87274507e8cb839cb49f93804e9176c4537cf63 height: 1517080
[1/5] converting full_blocks
      12809.48 seconds
[2/5] converting sub_epoch_segments_v3
      109.66 seconds
[3/5] converting hint_store
      8.71 seconds
[4/5] converting coin_store
      11470.76 seconds
[5/5] build indices
      block store
      coin store
      hint store
      5377.89 seconds


LEAVING PREVIOUS DB FILE UNTOUCHED /media/arvid/plots/blockchain_v1_modified.sqlite


real    1370m47,751s
user    41m35,470s
sys     25m12,156s

```

the baseline to compare against is running without vacuum and not offline (i.e. the input database is read-only).

Running:
```
time chia db upgrade --input /media/arvid/plots/blockchain_v1_mainnet.sqlite --output=/media/arvid/seagate/test-v2.sqlite
```

```
opening file for reading: /media/arvid/plots/blockchain_v1_mainnet.sqlite
opening file for writing: /media/arvid/seagate/test-v2.sqlite
initializing v2 version
initializing v2 block store
peak: 56e4b442dba0f58c6489991ea87274507e8cb839cb49f93804e9176c4537cf63 height: 1517080
[1/5] converting full_blocks
      23123.05 seconds
[2/5] converting sub_epoch_segments_v3
      109.08 seconds
[3/5] converting hint_store
      44.69 seconds
[4/5] converting coin_store
      15011.19 seconds
[5/5] build indices
      block store
      coin store
      hint store
      5589.73 seconds


LEAVING PREVIOUS DB FILE UNTOUCHED /media/arvid/plots/blockchain_v1_mainnet.sqlite


real    731m19,301s
user    46m21,826s
sys     15m57,859s
```

Running in offline mode without vacuuming:

```
time chia db upgrade --input /media/arvid/plots/blockchain_v1_mainnet.sqlite --output=/media/arvid/seagate/test-v2.sqlite --offline
```

```
opening file for reading: /media/arvid/plots/blockchain_v1_mainnet.sqlite in offline mode
opening file for writing: /media/arvid/seagate/test-v2.sqlite
initializing v2 version
initializing v2 block store
peak: 56e4b442dba0f58c6489991ea87274507e8cb839cb49f93804e9176c4537cf63 height: 1517080
[1/5] converting full_blocks
      22958.98 seconds
[2/5] converting sub_epoch_segments_v3
      109.18 seconds
[3/5] converting hint_store
      45.11 seconds
[4/5] converting coin_store
      14024.27 seconds
[5/5] build indices
      block store
      coin store
      hint store
      7433.54 seconds


LEAVING PREVIOUS DB FILE UNTOUCHED /media/arvid/plots/blockchain_v1_mainnet.sqlite


real    742m52,488s
user    46m10,757s
sys     15m47,161s
```

### DIFF
--- a/chia/cmds/db.py
+++ b/chia/cmds/db.py
@@ -26,8 +26,19 @@ def db_cmd() -> None:
     "the database file, you must not have a full node running against it. This may "
     "provide some performance improvements",
 )
+@click.option(
+    "--vacuum",
+    default=False,
+    is_flag=True,
+    help="vacuum the input database in an attempt to improve performance when "
+    "converting. This is only valid when also specifying --offline.",
+)
 @click.pass_context
-def db_upgrade_cmd(ctx: click.Context, no_update_config: bool, offline: bool, **kwargs) -> None:
+def db_upgrade_cmd(ctx: click.Context, no_update_config: bool, offline: bool, vacuum: bool, **kwargs) -> None:
+
+    if vacuum and not offline:
+        print("--vacuum is only valid when also specifying --offline")
+        raise RuntimeError("invalid command line option")
 
     in_db_path = kwargs.get("input")
     out_db_path = kwargs.get("output")
@@ -37,6 +48,7 @@ def db_upgrade_cmd(ctx: click.Context, no_update_config: bool, offline: bool, **
         None if out_db_path is None else Path(out_db_path),
         no_update_config=no_update_config,
         offline=offline,
+        vacuum=vacuum,
     )
 
 
--- a/chia/cmds/db.py
+++ b/chia/cmds/db.py
@@ -18,8 +18,16 @@ def db_cmd() -> None:
     help="don't update config file to point to new database. When specifying a "
     "custom output file, the config will not be updated regardless",
 )
+@click.option(
+    "--offline",
+    default=False,
+    is_flag=True,
+    help="open the input database in offline-mode. no other process may use "
+    "the database file, you must not have a full node running against it. This may "
+    "provide some performance improvements",
+)
 @click.pass_context
-def db_upgrade_cmd(ctx: click.Context, no_update_config: bool, **kwargs) -> None:
+def db_upgrade_cmd(ctx: click.Context, no_update_config: bool, offline: bool, **kwargs) -> None:
 
     in_db_path = kwargs.get("input")
     out_db_path = kwargs.get("output")
@@ -27,7 +35,8 @@ def db_upgrade_cmd(ctx: click.Context, no_update_config: bool, **kwargs) -> None
         Path(ctx.obj["root_path"]),
         None if in_db_path is None else Path(in_db_path),
         None if out_db_path is None else Path(out_db_path),
-        no_update_config,
+        no_update_config=no_update_config,
+        offline=offline,
     )
 
 
--- a/chia/cmds/db_upgrade_func.py
+++ b/chia/cmds/db_upgrade_func.py
@@ -79,8 +79,8 @@ async def convert_v1_to_v2(in_path: Path, out_path: Path, *, offline: bool, vacu
     print(f"opening file for reading: {in_path}{' in offline mode' if offline else ''}")
     async with aiosqlite.connect(in_path) as in_db:
         if offline:
-            await in_db.execute("pragma journal_mode=OFF")
-            await in_db.execute("pragma locking_mode=exclusive")
+            await (await in_db.execute("pragma journal_mode=OFF")).close()
+            await (await in_db.execute("pragma locking_mode=exclusive")).close()
 
         try:
             async with in_db.execute("SELECT * from database_version") as cursor:
--- a/chia/cmds/db_upgrade_func.py
+++ b/chia/cmds/db_upgrade_func.py
@@ -26,6 +26,7 @@ def db_upgrade_func(
     *,
     no_update_config: bool = False,
     offline: bool = False,
+    vacuum: bool = False,
 ):
 
     update_config: bool = in_db_path is None and out_db_path is None and not no_update_config
@@ -48,7 +49,7 @@ def db_upgrade_func(
         out_db_path = path_from_root(root_path, db_path_replaced)
         mkdir(out_db_path.parent)
 
-    asyncio.run(convert_v1_to_v2(in_db_path, out_db_path, offline=offline))
+    asyncio.run(convert_v1_to_v2(in_db_path, out_db_path, offline=offline, vacuum=vacuum))
 
     if update_config:
         print("updating config.yaml")
@@ -67,7 +68,7 @@ HINT_COMMIT_RATE = 2000
 COIN_COMMIT_RATE = 30000
 
 
-async def convert_v1_to_v2(in_path: Path, out_path: Path, *, offline: bool) -> None:
+async def convert_v1_to_v2(in_path: Path, out_path: Path, *, offline: bool, vacuum: bool) -> None:
     import aiosqlite
     from chia.util.db_wrapper import DBWrapper
 
@@ -89,6 +90,13 @@ async def convert_v1_to_v2(in_path: Path, out_path: Path, *, offline: bool) -> N
                     raise RuntimeError("already v2")
         except aiosqlite.OperationalError:
             pass
+
+        if offline and vacuum:
+            print("vacuuming input database")
+            start_time = time()
+            await in_db.execute("vacuum")
+            end_time = time()
+            print(f"\r      {end_time - start_time:.2f} seconds                             ")
 
         store_v1 = await BlockStore.create(DBWrapper(in_db, db_version=1))
 
--- a/chia/cmds/db_upgrade_func.py
+++ b/chia/cmds/db_upgrade_func.py
@@ -23,7 +23,9 @@ def db_upgrade_func(
     root_path: Path,
     in_db_path: Optional[Path] = None,
     out_db_path: Optional[Path] = None,
+    *,
     no_update_config: bool = False,
+    offline: bool = False,
 ):
 
     update_config: bool = in_db_path is None and out_db_path is None and not no_update_config
@@ -46,7 +48,7 @@ def db_upgrade_func(
         out_db_path = path_from_root(root_path, db_path_replaced)
         mkdir(out_db_path.parent)
 
-    asyncio.run(convert_v1_to_v2(in_db_path, out_db_path))
+    asyncio.run(convert_v1_to_v2(in_db_path, out_db_path, offline=offline))
 
     if update_config:
         print("updating config.yaml")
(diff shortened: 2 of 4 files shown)


mariano54:
Why not vaccum by default?

arvidn:
I would expect the typical use case is to convert a database that's being used by a running node (to minimize downtime). and you can't vacuum a database that's being used.

In fact, I just learned that cursor objects still need to be closed explicitly when executing sql commands, even when they don't return data. See this part of the patch:

https://github.com/Chia-Network/chia-blockchain/pull/10193/files#diff-52dd98662e9d0756012c8561eb52eaa8c54f8b1125b59c490218686b2d1d20d6R81

Without closing, the vacuum would fail because of outstanding sql queries. I think we'll need to go through the codebase and fix up these cases. It's probably mostly benign, as I expect them to be closed when the connection objects are garbage collected.

arvidn:
There are some important fixes and optimizations to the test that should land regardless. If these new options are controversial I will pull out the test fixes into a separate PR.

neurosis69:
> but @neurosis69 suggested another mode where you can vacuum into a new file, which would work on a database that's being used (at least that's my understanding).

[Here](https://www.sqlite.org/lang_vacuum.html#vacuuminto) is the relevant documentation regarding vacuum into. 

I'd like to see this `vacuum into` as backup feature in the future.

But if you want to reduce the io done and save time, than you should really try out my approach, instead of using the standard vacuum. 

The performance gains of my approach compared to standard vacuum are:

- dont create a primary keys for the staging table 
- don't copy index db pages as they are recreated later
- full_blocks table only needs to be reorganized on hdds

The staging tables can be seen as the target sqlite file of `vacuum into`.

Here are the timings for my unvacuumed test db with peak 1361584:
(Classic vacuum 14910s vs alternative 7164s)
```
#Runtime of classic vacuum with read from HDD and write to the same HDD, offline
sqlite> vacuum into '/mnt/JoolsDB/test.sqlite';
Run Time: real 14910.431 user 42.907418 sys 104.158201

#Runtime of vacuum-like reorganisation with read from HDD and write to the same HDD, offline
sqlite> attach database '/mnt/JoolsDB/testattach.sqlite' as testdb;
Run Time: real 0.000 user 0.000168 sys 0.000000
sqlite> CREATE TABLE IF NOT EXISTS testdb.temp_full_blocks(header_hash text, height bigint,
is_block tinyint, is_fully_compactified tinyint, block blob);
CREATE TABLE IF NOT EXISTS testdb.temp_block_records(header_hash text, prev_hash text, height bigint,
block blob, sub_epoch_summary blob, is_peak tinyint, is_block tinyint);
CREATE TABLE IF NOT EXISTS testdb.temp_coin_record(coin_name text,confirmed_index bigint,spent_index bigint,spent int,
coinbase int,puzzle_hash text,coin_parent text,amount blob,timestamp bigint);
CREATE TABLE IF NOT EXISTS testdb.temp_sub_epoch_segments_v3(ses_block_hash blob,challenge_segments blob);
Run Time: real 0.743 user 0.000489 sys 0.000599
sqlite> begin transaction;
Run Time: real 0.000 user 0.000010 sys 0.000002
sqlite> insert into temp.temp_full_blocks select * from main.full_blocks;
Run Time: real 0.000 user 0.000038 sys 0.000011
Error: in prepare, no such table: temp.temp_full_blocks (1)
sqlite> insert into testdb.temp_full_blocks select * from main.full_blocks;
Run Time: real 6015.050 user 6.386299 sys 42.225500
sqlite> insert into testdb.temp_block_records select * from main.block_records;
Run Time: real 363.113 user 0.717797 sys 2.819371
sqlite> insert into testdb.temp_coin_record select * from main.coin_record;
Run Time: real 711.108 user 6.094620 sys 9.527292
sqlite> insert into testdb.temp_sub_epoch_segments_v3 select * from main.sub_epoch_segments_v3;
Run Time: real 74.920 user 0.521854 sys 1.424634
sqlite> commit;
Run Time: real 0.408 user 0.001245 sys 0.000511
sqlite> select 6015.050+363.113+711.108+74.920 overall;
+----------+
| overall  |
+----------+
| 7164.191 |
+----------+
```

arvidn:
yeah, I get the impression that in-place `vacuum` on an HDD does not pay off. I will probably remove that feature from this patch.

neurosis69:
btw, please think about changing the [page_size](https://www.sqlite.org/pragma.html#pragma_page_size) as well.

I have seen in the commits, that one of my suggestions, using partial indexes had to be taken back because some of the farmers where using sqlite versions < 3.8. 

When it comes to page_size, until 3.12 the default was 1k, since then it is 4. (and I've heard from a member in my pool that he had a blockchain db with 1k page_size). So if you don't want to increase it, maybe set it at least explicitly to 4k as part of the initialization process, so farmers using old sqlite versions don't have 1k pagesize.

Here are timings for the same inserts from above using same scenario when the attached target db is having 16k page size.

```
sqlite> attach database '/mnt/JoolsDB/testattach_16k.sqlite' as testdb;
Run Time: real 0.001 user 0.000024 sys 0.000070
sqlite> pragma testdb.page_size=16384;
Run Time: real 0.000 user 0.000032 sys 0.000000
sqlite> pragma testdb.page_size;
+-----------+
| page_size |
+-----------+
| 16384     |
+-----------+
Run Time: real 0.000 user 0.000026 sys 0.000000
sqlite> CREATE TABLE IF NOT EXISTS testdb.temp_full_blocks(header_hash text, height bigint,
is_block tinyint, is_fully_compactified tinyint, block blob);
CREATE TABLE IF NOT EXISTS testdb.temp_block_records(header_hash text, prev_hash text, height bigint,
block blob, sub_epoch_summary blob, is_peak tinyint, is_block tinyint);
CREATE TABLE IF NOT EXISTS testdb.temp_coin_record(coin_name text,confirmed_index bigint,spent_index bigint,spent int,
coinbase int,puzzle_hash text,coin_parent text,amount blob,timestamp bigint);
CREATE TABLE IF NOT EXISTS testdb.temp_sub_epoch_segments_v3(ses_block_hash blob,challenge_segments blob);
Run Time: real 0.579 user 0.000835 sys 0.000833
sqlite> begin transaction;
Run Time: real 0.000 user 0.000013 sys 0.000000
sqlite> insert into testdb.temp_full_blocks select * from main.full_blocks;
Run Time: real 5110.779 user 5.192634 sys 39.861083
sqlite> insert into testdb.temp_block_records select * from main.block_records;
Run Time: real 220.492 user 0.517920 sys 2.093178
sqlite> insert into testdb.temp_coin_record select * from main.coin_record;
Run Time: real 203.503 user 4.635895 sys 6.310411
sqlite> insert into testdb.temp_sub_epoch_segments_v3 select * from main.sub_epoch_segments_v3;
Run Time: real 29.416 user 0.332108 sys 0.970081
sqlite> select 5110.779+220.492+203.503+29.416 as overall;
+---------+
| overall |
+---------+
| 5564.19 |
+---------+

```


And I'm still pretty sure that you would see good improvement when using 
1. a separate sqlite file for sub_epoch table
2. a page_size of 64k (or the biggest possible, idk if 64k was introduced later)
I tried to make a PR, but I don't have that much time currently.

arvidn:
> I have seen in the commits, that one of my suggestions, using partial indexes had to be taken back because some of the farmers where using sqlite versions < 3.8.

Actually, we probably landed that change prematurely, because v2 still requires a newer sqlite.

arvidn:
ok. I'll make another PR to set the page size, with timings. probably tomorrow.

neurosis69:
> And I'm still pretty sure that you would see good improvement when using
> 
>     1. a separate sqlite file for sub_epoch table
> 
>     2. a page_size of 64k (or the biggest possible, idk if 64k was introduced later)
>        I tried to make a PR, but I don't have that much time currently.

This was all about the sub_epoch table, if not clear. I'm currently running a full_node with 64k page_size, but I don't think that that would make sense when the sub_epoch table is separated and it would even make the sqlite db bigger. 
I tried to explain the reason for using 64k for sub_epoch and the separation in another thread, but again.
the problem is imo (and how I understand sqlite until now), that after some time, due to the size of the table and its rows, it is scattered all over the sqlite file, slowing down db access.

arvidn:
it doesn't look like *just* opening the input file in offline mode helps either. @neurosis69 does this contradict your results?

I'm increasingly getting the impression that the majority of your speedups come from using temporary files.

neurosis69:
Did you run this on a RPi like low end HW? On my decent plotter HW I didn't have substantial run time differences as well.
The offline part was for sure not that important. Actually, I think I added that at the very end.

Most important what comes into my mind are:

- using vacuum-like source tables (or temp/staging tables), huge impact on coin_record runtime and hdd device
- using recursive CTE for full_blocks to be able to fetch as many as BLOCK_COMMIT_RATE defined rows at once
- using sqlite3 instead of aiosqlite
- using arraysize as defined by _COMMIT_RATE variables on fetchmany

neurosis69:
and of course, really important is how much memory we allow, so in my PR I have two scenarious, one with < 3.5GB mem available and one with more. For the first part, the commit chunks I proposed were already to big, so my RPi one a test when having a full node up and running started heavily swapping, and of course this will slow down the processes.

For simplicity I'd propose only two scenarios, one where <3.5G is available, then these rates are ok (according to my tests):
        BLOCK_COMMIT_RATE = 3000
        COIN_COMMIT_RATE = 12000
        SES_COMMIT_RATE = 5
        HINT_COMMIT_RATE = 50000

And if > 3.5 GB are available, then these rates are a compromise of good performance and memory consumption, imho
        BLOCK_COMMIT_RATE = 80000
        COIN_COMMIT_RATE = 300000
        SES_COMMIT_RATE = 50
        HINT_COMMIT_RATE = 300000

arvidn:
> Did you run this on a RPi like low end HW?

No, just a low-end, USB-attached and USB-powered, 2.5" HDD.

arvidn:
as far as I can tell, `--offline --vacuum` nor `--offline` are improvements (for HDD). The only reason (I can think of) to believe `--offline` mode would be faster is that sqlite's read transaction could be cheaper, since it has exclusive access to the file, it doesn't have to do anything to ensure it has a consistent view.

Perhaps I should test this on an SSD as well, for good measure.

arvidn:
@neurosis69
> using sqlite3 instead of aiosqlite

That seems like a clear win, at least on SSD (testing on HDD now). https://github.com/Chia-Network/chia-blockchain/pull/10259

PR #10207, opened 2022-02-12
Tolerate missing hints

arvidn:
Please review ignoring whitespace.

The problem solved by this patch is that when upgrading from a database version before hints were introduced, there won't be a `hints` table. Currently, the `chia db upgrade` function *assumes* a `hints` table and will fail when performing such upgrade.

This patch also parameterize the test, to run both with and without a hints table. The test as is currently written is quite slow, so this patch also speeds it up by using a pre-built test chain instead of building a new one from scratch every time.

### DIFF
--- a/chia/cmds/db_upgrade_func.py
+++ b/chia/cmds/db_upgrade_func.py
@@ -1,4 +1,5 @@
 from typing import Dict, Optional
+import sqlite3
 from pathlib import Path
 import sys
 from time import time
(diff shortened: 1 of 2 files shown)


tests/core/test_db_conversion.py
```diff
@@ -66,65 +75,81 @@ async def test_blocks(self):
...
                 for block in blocks:
-                    await _validate_and_add_block(bc, block)
+                    # await _validate_and_add_block(bc, block)
+                    results = PreValidationResult(None, uint64(1), None, False)
+                    result, err, _, _ = await bc.receive_block(block, results)
+                    assert err is None
 
-                # now, convert v1 in_file to v2 out_file
-                await convert_v1_to_v2(in_file, out_file)
+            # now, convert v1 in_file to v2 out_file
```
arvidn, line 104: from here and below, one level of indentation was removed to close the database connection to the in-table before running the conversion function (the function being tested). The conversion function opens the database in exclusive mode, so technically, it wasn't exclusive prior to this change.

chia/cmds/db_upgrade_func.py
```diff
@@ -237,18 +238,21 @@ async def convert_v1_to_v2(in_path: Path, out_path: Path) -> None:
...
-                async for row in cursor:
-                    hint_values.append((row[0], row[1]))
-                    commit_in -= 1
-                    if commit_in == 0:
-                        commit_in = HINT_COMMIT_RATE
-                        await out_db.executemany("INSERT OR IGNORE INTO hints VALUES(?, ?)", hint_values)
-                        await out_db.commit()
-                        await out_db.execute("begin transaction")
-                        hint_values = []
+            try:
```
arvidn, line 241: this is the fix. catch the sqlite error and ignore it.

mariano54 (approved):
lgtm

emlowe (approved):
approved

PR #10212, opened 2022-02-12
Fix trusted_peer example in initial-config.yaml

ojura:
The current code will never allow trusting a non-localhost full node peer from the pased in dictionary `dict`.
The node id hex string value will never be `in` a `Dict`, since a `Dict` iterable contains only the keys.
The intention was obviously to check if the trusted node id is `in dict.values()`.

Furthermore, the provided sample config line in `initial-config.yaml` for wallet's `trusted_peers` is misleading.
The dictionary values should be trusted node IDs, and not public certificates, as seen in the relevant code
in `server.py` discussed in the fix above.
We should change this line not to be misleading.


ojura:
I would estimate that this has a pretty large impact.

A bunch of people, myself included, thought that setting wallet's `full_node_peer` to a trusted full node would enable the fast trusted sync. This is NOT the case, its node ID has to be added to `trusted_peers`, and the fix in this PR has to be applied.

Trusted light wallet sync is blazing fast compared to untrusted sync, and so far it is ONLY possible with a localhost full node!

ojura:
Obsoleted by https://github.com/Chia-Network/chia-blockchain/pull/10252

ojura:
#10252 got stale, while these couple of obviously wrong lines should have been fixed a long time ago.

chia/util/initial-config.yaml
```diff
@@ -532,8 +532,9 @@ wallet:
     public_crt:  "config/ssl/wallet/public_wallet.crt"
     public_key:  "config/ssl/wallet/public_wallet.key"
 
+  # node IDs of trusted full node peers
   trusted_peers:
-    trusted_node_1: "config/ssl/full_node/public_full_node.crt"
+    # trusted_node_1: "0000000000000000000000000000000000000000000000000000000000000000"
```
ojura, line 537: The codebase expects a trusted node ID here, not a certificate.

emlowe:
The change to initial-config is good as the current file is rather wrong now. However, we have changed our documentation to say that you should use 
```
trusted_peers:
    0ThisisanexampleNodeID7ff9d60f1c3fa270c213c0ad0cb89c01274634a7c3cb9: Does_not_matter
```
(https://github.com/Chia-Network/chia-blockchain/wiki/FAQ#what-are-trusted-peers-and-how-do-i-add-them)

So this PR would break for anyone who followed that advice. My recommendation is that the code should check both keys and values for maximum backwards compatibility and flexibility.

emlowe:
As a note, it is all rather confusing either way and the other PR was a better change in the right direction, but that got lost and hopelessly left behind. Changing to use the same strange format, but check both sides of the `:` is the simplest improvement without more substantial changes

ojura:
Ohh, it didn't occur to me that you could swap it around and to put the node ID in the YAML key name, so that it works with the current code which checks the keys. I was really misled by the wrong example in initial-config.

In that case, all that remains is to fix the example in initial-config to match the documentation (and revert the change in `server.py`)?

emlowe (approved):
thanks!

wjblanke (approved):
aok

PR #10216, opened 2022-02-12
Fix memory leak in sync_store.py

jack60612:


mariano54 (approved):

mariano54:
This has been manually confirmed to work.

PR #10219, opened 2022-02-13
less except -> false

altendky:
The conversion of all the exceptions of any type to `False` is problematic.  When an exception other than that which you are testing for is raised you have tests passing that haven't actually tested their intended failure and you have no indication of what the new error case is.


aqk (dismissed):
The API of `establish_connection` is a little dodgy. It might be nice to add the return type annotation to `establish_connection` now, if it doesn't cause a linter cascade.

altendky:
It already seems to be hinted to return `bool`.

https://github.com/Chia-Network/chia-blockchain/blob/d49414b582d6196742336b4b1dfc898ae77c6386/tests/core/ssl/test_ssl.py#L24

This matches the only return in the underlying `.perform_handshake()`.

https://github.com/Chia-Network/chia-blockchain/blob/d49414b582d6196742336b4b1dfc898ae77c6386/chia/server/ws_connection.py#L187

But...  maybe you just meant `.perform_handshake()` itself which is un-hinted, which it is.  Both can be hinted `Literal[True]` without any mypy complaints.  Adding that now.  Though then you get to review whether the return value should just be dropped entirely given it can only ever be a single value...  I don't mind some thrash here.  Returning nothing is probably more likely to lead to sensible usage where the dev writing the calling code realizes the return value is meaningless.

Yep, just pushed setting the return type to `None` and catching the (small) cascade of already-pointless assertions.  Though, that does make this a not-just-tests PR.

tests/core/ssl/test_ssl.py
```diff
@@ -37,12 +36,7 @@ async def establish_connection(server: ChiaServer, self_hostname: str, ssl_conte
             100,
             30,
         )
-        handshake = await wsc.perform_handshake(server._network_id, protocol_version, dummy_port, NodeType.FULL_NODE)
-        await session.close()
-        return handshake
-    except Exception:
-        await session.close()
-        return False
```
arvidn, line 45: this interface changed. It used to not throw when failing. I take it we don't rely on that anywhere, yeah?

arvidn (approved):

tests/core/ssl/test_ssl.py
altendky: All the calls on `main` are of the following form.

```python
        connected = await establish_connection(farmer_server, self_hostname, ssl_context)
        assert connected is True
```

Looking below, the `True` assertions are just eliminated as the call will fail due to the raised exception while the `False` assertions have been replaced by `pytest.raises()` expecting a particular exception making the new code actually more picky about asserting the failure.

Assuming I haven't missed something somewhere...

PR #10221, opened 2022-02-13
Remove deprecated initial_num_public_keys_new_wallet from initial-config.yaml

altendky:
https://github.com/Chia-Network/chia-blockchain/blob/231ef6faf20fba7463831a5015edee649a1d5d49/CHANGELOG.md
> Fixed issues where the wallet code would not generate enough addresses when looking for coins, which can result in missed coins due to the address not being checked. Deprecated the config setting initial_num_public_keys_new_wallet. The config setting initial_num_public_keys is now used in all cases.


wjblanke (approved):
aok

PR #10233, opened 2022-02-14
Fix to allow debugging services when launched alongside the GUI

paninaro:
Daemon RPCs `start_service` and `is_running` will now additionally check self.connections to determine if a service is running.

When launching services manually (e.g. when debugging), the GUI would attempt (and fail) to relaunch services that were already running, due to the daemon only reporting the statuses of services started by the daemon.

Tested that `chia start <service` worked properly, as well as launching a full set of services from the GUI for farming. Additionally tested that manually running `chia run_daemon` and `python chia/server/start_wallet.py` worked as expected alongside the GUI.

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -1093,7 +1093,7 @@ class WebSocketServer:
             self.log.info(f"Service {service_command} already registered")
             already_running = True
 
-        if already_running is False and error is None:
+        if not already_running and error is None:
             try:
                 exe_command = service_command
                 if testing is True:
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -1071,6 +1071,7 @@ class WebSocketServer:
         error = None
         success = False
         testing = False
+        already_running = False
         if "testing" in request:
             testing = request["testing"]
 
@@ -1084,9 +1085,15 @@ class WebSocketServer:
                 self.services.pop(service_command)
                 error = None
             else:
-                error = f"Service {service_command} already running"
+                self.log.info(f"Service {service_command} already running")
+                already_running = True
+        elif service_command in self.connections:
+            # If the service was started manually (not launched by the daemon), we should
+            # have a connection to it.
+            self.log.info(f"Service {service_command} already registered")
+            already_running = True
 
-        if error is None:
+        if already_running is False and error is None:
             try:
                 exe_command = service_command
                 if testing is True:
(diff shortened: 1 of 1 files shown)


chia/daemon/server.py
```diff
@@ -1084,9 +1085,15 @@ async def start_service(self, request: Dict[str, Any]):
...
-        if error is None:
+                self.log.info(f"Service {service_command} already running")
+                already_running = True
+        elif service_command in self.connections:
+            # If the service was started manually (not launched by the daemon), we should
+            # have a connection to it.
+            self.log.info(f"Service {service_command} already registered")
+            already_running = True
+
+        if already_running is False and error is None:
```
altendky, line 1096:
```suggestion
        if not already_running and error is None:
```

chia/daemon/server.py
```diff
@@ -1084,9 +1085,15 @@ async def start_service(self, request: Dict[str, Any]):
                 self.services.pop(service_command)
                 error = None
             else:
-                error = f"Service {service_command} already running"
-
-        if error is None:
+                self.log.info(f"Service {service_command} already running")
+                already_running = True
```
altendky, line 1089: Can't the removal of `error =` here affect behavior in existing cases?

chia/daemon/server.py
```diff
@@ -1097,6 +1104,8 @@ async def start_service(self, request: Dict[str, Any]):
             except (subprocess.SubprocessError, IOError):
                 log.exception(f"problem starting {service_command}")
                 error = "start failed"
+        elif already_running:
+            success = True
```
altendky, line 1108: I often like to put the short case first.  Also, in this case, the other condition simplifies to `error is None`.

chia/daemon/server.py
paninaro: I checked and this change doesn't impact any of the callers that I'm aware of (`chia start <service>` and the GUI). If you run `chia start wallet` when the wallet is already running, the behavior is as-expected: "chia_wallet: Already running, use '-r' to restart"

Had I left the error intact, GUI code would need to be changed to handle the case where an error is returned but the service is already running... I felt that would be an uglier solution.

chia/daemon/server.py
paninaro: Sure

altendky (approved):
Thanks!

PR #10244, opened 2022-02-15
Ms.sync cancel

mariano54:
- Allows sync interruption
- Sets the height properly after processing some coin states
- Start from zero always when checking coin id subscriptions
- Fix issue with rate limits from wallet config
- hardcode to 0.6 rate limit for wallet
- Return the correct height in RPC
- Sort the applied coin states by last changed height
- Select a random peer when fetching header blocks, fallback to sync target peer
- Fix many tasks created unnecessarily
- Always sync from zero for a secondary untrusted sync
- Don't re-request the same blocks from two tasks that execute concurrently

### DIFF
--- a/chia/server/start_service.py
+++ b/chia/server/start_service.py
@@ -79,9 +79,9 @@ class Service:
         chia_ca_crt, chia_ca_key = chia_ssl_ca_paths(root_path, self.config)
         inbound_rlp = self.config.get("inbound_rate_limit_percent")
         outbound_rlp = self.config.get("outbound_rate_limit_percent")
-        if NodeType == NodeType.WALLET:
+        if node_type == NodeType.WALLET:
             inbound_rlp = service_config.get("inbound_rate_limit_percent", inbound_rlp)
-            outbound_rlp = service_config.get("outbound_rate_limit_percent", 60)
+            outbound_rlp = 60
 
         assert inbound_rlp and outbound_rlp
         self._server = ChiaServer(
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -610,7 +610,6 @@ class WalletNode:
             try:
                 assert self.validation_semaphore is not None
                 async with self.validation_semaphore:
-                    nonlocal num_concurrent_tasks
                     assert self.wallet_state_manager is not None
                     if header_hash is not None:
                         assert height is not None
@@ -633,6 +632,7 @@ class WalletNode:
                 tb = traceback.format_exc()
                 self.log.error(f"Exception while adding state: {e} {tb}")
             finally:
+                nonlocal num_concurrent_tasks
                 num_concurrent_tasks -= 1
 
         for idx, potential_state in enumerate(items):
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -633,7 +633,7 @@ class WalletNode:
                 self.log.error(f"Exception while adding state: {e} {tb}")
             finally:
                 nonlocal num_concurrent_tasks
-                num_concurrent_tasks -= 1
+                num_concurrent_tasks -= 1  # pylint: disable=E0602
 
         for idx, potential_state in enumerate(items):
             if self.server is None:
@@ -1044,7 +1044,6 @@ class WalletNode:
         # Only use the cache if we are talking about states before the fork point. If we are evaluating something
         # in a reorg, we cannot use the cache, since we don't know if it's actually in the new chain after the reorg.
         if await can_use_peer_request_cache(coin_state, peer_request_cache, fork_height):
-            self.log.info("Cache hit!")
             return True
 
         spent_height = coin_state.spent_height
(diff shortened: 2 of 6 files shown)


chia/wallet/wallet_node.py
```diff
@@ -804,12 +831,16 @@ async def new_peak_wallet(self, new_peak: wallet_protocol.NewPeakWallet, peer: W
                     assert weight_proof is not None
                     old_proof = self.wallet_state_manager.blockchain.synced_weight_proof
                     if syncing:
-                        fork_point: int = max(0, current_height - 32)
+                        # This usually happens the first time we start up the wallet. We roll back slightly to be
+                        # safe, but we don't want to rollback too much (hence 32)
```
Nikolaj-K, line 835: You changed `-32` to `-16`, but the new comment says 32.

mariano54: Thank you i missed that

chia/wallet/wallet_node.py
```diff
@@ -589,41 +602,55 @@ async def receive_state_from_peer(
...
+                try:
+                    if header_hash is not None:
+                        assert height is not None
+                        self.add_state_to_race_cache(header_hash, height, inner_state)
+                        self.log.info(f"Added to race cache: {height}, {inner_state}")
+                    if trusted:
+                        valid = True
+                    else:
+                        valid = await self.validate_received_state_from_peer(inner_state, peer, cache, fork_height)
+                    if valid:
```
Nikolaj-K, line 624: `if trusted or await validate_foo():`
if you want to save the four lines

chia/server/start_service.py
```diff
@@ -79,9 +79,9 @@ def __init__(
         chia_ca_crt, chia_ca_key = chia_ssl_ca_paths(root_path, self.config)
         inbound_rlp = self.config.get("inbound_rate_limit_percent")
         outbound_rlp = self.config.get("outbound_rate_limit_percent")
-        if NodeType == NodeType.WALLET:
+        if node_type == NodeType.WALLET:
```
Yostra, line 82: oh limit was not raised previously at all

chia/wallet/wallet_node.py
Yostra: that's harder to read and reason what's the logic without taking a sec

chia/wallet/wallet_node.py
```diff
@@ -589,41 +602,55 @@ async def receive_state_from_peer(
...
+                            await self.wallet_state_manager.new_coin_state([inner_state], peer, fork_height)
+                        if update_finished_height:
+                            await self.wallet_state_manager.blockchain.set_finished_sync_up_to(
+                                last_change_height_cs(inner_state)
+                            )
+                except Exception as e:
+                    tb = traceback.format_exc()
+                    self.log.error(f"Exception while adding state: {e} {tb}")
+                finally:
+                    num_concurrent_tasks -= 1
```
altendky, line 637: This function can terminate without entering the `try` and thus without decreasing the count.  I think a different mechanism would be better, even short of a worker pool (maybe I'll take a pass at that old PR tomorrow).  I'll try to come up with a suggestion.

chia/server/start_service.py
mariano54: yeah. we are still hitting the limit with 60 sometimes though because the minute barriers can be at different points, so we reset our sent count but the node has not reset theirs yet.

Yostra: thought incoming was at 100% ? or am I missing how it works

chia/wallet/wallet_node.py
mariano54: thanks for point that out. However it can only fail if that task is somehow cancelled right? and we never actually cancel that task here.
Feel free to change this one or make a PR for it, i'm sure you can come up with something more robust

chia/server/start_service.py
mariano54: it's 100% but it resets every minute. The thing is that the point that it resets might not be the same for the client and server

mariano54: and 0.6*2 > 1.0

chia/wallet/wallet_node.py
altendky: There are two asserts at the beginning that are outside of the `try:` along with waiting on the semaphore.  I don't know offhand if the semaphore can fail other than cancellation, but the other two can anyways.  I did just get https://github.com/Chia-Network/chia-blockchain/pull/9122 resurrected without much in the way of conflicts.  I'll need to write up tests and then give it some runtime as well.  Assuming we generally like that direction I'm not super worried about getting this entirely generally proper.  Shifting the `try:` up a few lines seems reasonable though.

chia/wallet/wallet_node.py
```diff
@@ -1016,6 +1045,7 @@ async def validate_received_state_from_peer(
         # Only use the cache if we are talking about states before the fork point. If we are evaluating something
         # in a reorg, we cannot use the cache, since we don't know if it's actually in the new chain after the reorg.
         if await can_use_peer_request_cache(coin_state, peer_request_cache, fork_height):
+            self.log.info("Cache hit!}")
```
altendky, line 1048: Maybe a little more explanation too?

```suggestion
            self.log.info("Cache hit!")
```

mariano54: I'll remove this, it was for debugging

mariano54:
This is ready to merge

Yostra (approved):

PR #10248, opened 2022-02-15
Acquire lock during update RPCs

Quexington:
Previously we didn't stop someone from hitting this endpoint multiple times before the first one had finished and it was leading to selecting the same fee coins for all transactions.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1427,14 +1427,15 @@ class WalletRpcApi:
 
         for _, wallet in self.service.wallet_state_manager.wallets.items():
             if WalletType(wallet.type()) == WalletType.DATA_LAYER:
-                records = await wallet.create_update_state_spend(
-                    bytes32.from_hexstr(request["launcher_id"]),
-                    bytes32.from_hexstr(request["new_root"]),
-                    fee=uint64(request.get("fee", 0)),
-                )
-                for record in records:
-                    await self.service.wallet_state_manager.add_pending_transaction(record)
-                return {"tx_record": records[0].to_json_dict_convenience(self.service.config)}
+                async with self.service.wallet_state_manager.lock:
+                    records = await wallet.create_update_state_spend(
+                        bytes32.from_hexstr(request["launcher_id"]),
+                        bytes32.from_hexstr(request["new_root"]),
+                        fee=uint64(request.get("fee", 0)),
+                    )
+                    for record in records:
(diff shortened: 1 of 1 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1427,14 +1427,15 @@ async def dl_update_root(self, request) -> Dict:
...
             if WalletType(wallet.type()) == WalletType.DATA_LAYER:
-                records = await wallet.create_update_state_spend(
-                    bytes32.from_hexstr(request["launcher_id"]),
-                    bytes32.from_hexstr(request["new_root"]),
-                    fee=uint64(request.get("fee", 0)),
-                )
-                for record in records:
-                    await self.service.wallet_state_manager.add_pending_transaction(record)
-                return {"tx_record": records[0].to_json_dict_convenience(self.service.config)}
+                with self.service.wallet_state_manager.lock:
```
emlowe, line 1430: aync with?

altendky (approved):
I would think this would still need a db transaction with rollback on non-success etc.  But, this is a step forward and there is better tooling for handling it all together that is being worked in another PR.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1445,25 +1446,26 @@ async def dl_update_multiple(self, request) -> Dict:
...
-                return {"tx_records": [rec.to_json_dict_convenience(self.service.config) for rec in modified_txs]}
+                async with self.service.wallet_state_manager.lock:
+                    # TODO: This method should optionally link the singletons with announcements.
+                    #       Otherwise spends are vulnerable to signature subtraction.
+                    tx_records: List[TransactionRecord] = []
+                    for launcher, root in request["updates"].items():
+                        records = await wallet.create_update_state_spend(
+                            bytes32.from_hexstr(launcher), bytes32.from_hexstr(root)
+                        )
+                        tx_records.extend(records)
```
Nikolaj-K, line 1457: Just because I usually see it avoided here, this silently assumes that `records` has to be non-None yeah?

Quexington: It's not so much an assumption as a fact of the code: https://github.com/Chia-Network/chia-blockchain/blob/c231ad449b9be6cf08a6e2548bbe262e08aeb9b1/chia/data_layer/data_layer_wallet.py#L428

_Trust me_ mypy would be letting me know if it could possibly be `None`

PR #10251, opened 2022-02-15
Accommodate multiple coin fees for plotnft

Quexington:


aqk (approved):

jack60612 (approved):

tests/pools/test_pool_rpc.py
```diff
@@ -34,7 +34,7 @@
 
 # TODO: Compare deducted fees in all tests against reported total_fee
 log = logging.getLogger(__name__)
-FEE_AMOUNT = 10
+FEE_AMOUNT = 2000000000000
```
arvidn, line 37: does this change make the test exercise this check?
```
if balance < PoolWallet.MINIMUM_INITIAL_BALANCE + fee:
```

It's hard for me to tell, but I get the impression a lot of the failure cases aren't covered by tests. I don't see any updates to expected failures because the balance is too low for instance.

Quexington: That change is just to make the tests use multiple coin fees.  I agree that the test coverage of the pool tests may be lacking but I think it's out of the scope of this PR.

PR #10252, opened 2022-02-15
Make easier to configure trusted peer on wallet side

Yostra:


emlowe:
I think this does deprecate the ability to have multiple trusted nodes?

Yostra:
> I think this does deprecate the ability to have multiple trusted nodes?

No, there's no support for connecting to multiple trusted peers at the moment.

ojura:
> I think this does deprecate the ability to have multiple trusted nodes?

That would be nice for redundancy, but it's good that we can at least one non-localhost trusted node now. This was actually broken, because the line which checked if the node ID was in the trusted full node dictionary was missing `.values()`.

This makes #10212 obsolete, closing.

mariano54:
Can we also add an ability to set the trusted wallet on the full node side? Otherwise we will get rate limits from the full node..

Yostra:
> Can we also add an ability to set the trusted wallet on the full node side? Otherwise we will get rate limits from the full node..

added

ojura:
@Yostra  @emlowe I closed  https://github.com/Chia-Network/chia-blockchain/pull/10212 because this PR would have made the fix obsolete.

But now, 3 months, the bug that would have been closed with #10212 is still present?

ojura:
> @Yostra @emlowe I closed #10212 because this PR would have made the fix obsolete.
> 
> But now, 3 months later, the bug in `is_trusted_peer` that would have been closed with #10212 is still present on master?

Hi guys, I have reopened #10212 because this PR looks stale. Could you have a look at that one? It fixes an obvious typo in the trusted peer code.

PR #10259, opened 2022-02-16
use synchronous sqlite3 library in db conversion function

arvidn:
This saves time synchronizing with the aiosqlite thread (every `await` means sending a message to the thread and waiting for a response). Since this workload is inherently single threaded, we don't need the additional thread.

| System | before | after | after / before |
| --- | --- | --- | --- |
| Ubuntu, SSD | 1548.5s | 1200.0s | 77,5% |

Before:

```
$ time chia db upgrade --input ~/blockchain_v1_mainnet.sqlite --output test.sqlite
opening file for reading: /home/arvid/blockchain_v1_mainnet.sqlite
opening file for writing: test.sqlite
initializing v2 version
initializing v2 block store
peak: 56e4b442dba0f58c6489991ea87274507e8cb839cb49f93804e9176c4537cf63 height: 1517080
[1/5] converting full_blocks
      588.31 seconds                              
[2/5] converting sub_epoch_segments_v3
      6.56 seconds                             
[3/5] converting hint_store
      7.69 seconds                             
[4/5] converting coin_store
      769.77 seconds                             
[5/5] build indices
      block store
      coin store
      hint store
      175.15 seconds                             


LEAVING PREVIOUS DB FILE UNTOUCHED /home/arvid/blockchain_v1_mainnet.sqlite


real    25m48,532s
user    17m0,143s
sys     4m58,747s
```

after:

```
$ time chia db upgrade --input ~/blockchain_v1_mainnet.sqlite --output test.sqlite
opening file for reading: /home/arvid/blockchain_v1_mainnet.sqlite
opening file for writing: test.sqlite
initializing v2 version
initializing v2 block store
peak: 56e4b442dba0f58c6489991ea87274507e8cb839cb49f93804e9176c4537cf63 height: 1517080
[1/5] converting full_blocks
      380.23 seconds                              
[2/5] converting sub_epoch_segments_v3
      6.39 seconds                             
[3/5] converting hint_store
      4.74 seconds                             
[4/5] converting coin_store
      639.23 seconds                             
[5/5] build indices
      block store
      coin store
      hint store
      168.58 seconds                             


LEAVING PREVIOUS DB FILE UNTOUCHED /home/arvid/blockchain_v1_mainnet.sqlite


real    20m0,032s
user    11m39,920s
sys     3m58,693s
```

### DIFF
--- a/chia/cmds/db_upgrade_func.py
+++ b/chia/cmds/db_upgrade_func.py
@@ -1,17 +1,11 @@
 from typing import Dict, Optional
-import sqlite3
 from pathlib import Path
 import sys
 from time import time
 
-import asyncio
-import zstd
-
 from chia.util.config import load_config, save_config
 from chia.util.path import mkdir, path_from_root
-from chia.full_node.block_store import BlockStore
-from chia.full_node.coin_store import CoinStore
-from chia.full_node.hint_store import HintStore
+from chia.util.ints import uint32
 from chia.types.blockchain_format.sized_bytes import bytes32
 
 
@@ -46,7 +40,7 @@ def db_upgrade_func(
         out_db_path = path_from_root(root_path, db_path_replaced)
         mkdir(out_db_path.parent)
 
-    asyncio.run(convert_v1_to_v2(in_db_path, out_db_path))
+    convert_v1_to_v2(in_db_path, out_db_path)
 
     if update_config:
         print("updating config.yaml")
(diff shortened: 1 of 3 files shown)


chia/cmds/db_upgrade_func.py
```diff
@@ -314,26 +323,39 @@ async def convert_v1_to_v2(in_path: Path, out_path: Path) -> None:
...
             print(f"\r      {end_time - coin_start_time:.2f} seconds                             ")
 
             print("[5/5] build indices")
             index_start_time = time()
             print("      block store")
-            await BlockStore.create(DBWrapper(out_db, db_version=2))
+            out_db.execute("CREATE INDEX height on full_blocks(height)")
+            out_db.execute(
+                "CREATE INDEX is_fully_compactified ON"
+                " full_blocks(is_fully_compactified, in_main_chain) WHERE in_main_chain=1"
```
aqk, line 345: Does `in_main_chain` have a limit, e.g. `(peak-5 blocks)`?

The reason I ask is I'm wondering about the possible danger of new code using this index to mean something slightly different than other "in_main_chain_ equivalent checks elsewhere.

chia/cmds/db_upgrade_func.py
```diff
@@ -314,26 +323,39 @@ async def convert_v1_to_v2(in_path: Path, out_path: Path) -> None:
...
+            out_db.executemany("INSERT INTO coin_record VALUES(?, ?, ?, ?, ?, ?, ?, ?)", coin_values)
+            out_db.commit()
             end_time = time()
             print(f"\r      {end_time - coin_start_time:.2f} seconds                             ")
 
             print("[5/5] build indices")
             index_start_time = time()
             print("      block store")
-            await BlockStore.create(DBWrapper(out_db, db_version=2))
+            out_db.execute("CREATE INDEX height on full_blocks(height)")
```
aqk, line 342: I notice this block is new, but it seems to be all indexes and hints, so no worries.

chia/full_node/block_store.py
```diff
@@ -52,6 +52,8 @@ async def create(cls, db_wrapper: DBWrapper):
             # peak. The "key" field is there to make update statements simple
             await self.db.execute("CREATE TABLE IF NOT EXISTS current_peak(key int PRIMARY KEY, hash blob)")
 
+            # If any of these indices are altered, they should also be altered
+            # in the chia/cmds/db_upgrade.py file
```
aqk, line 56:
```suggestion
            # in the chia/cmds/db_upgrade.py file, and the database version should be incremented
```

aqk (approved):
Thanks!

I learned about `contextlib.closing`.

I see the blocks are already compressed in v2, so that's great, too.

chia/cmds/db_upgrade_func.py
arvidn: `in_main_chain` is a field indicating that the block is part of the current longest chain. whenever we reorg, we also update this field to keep a single trail of blocks from genesis to the peak. One invariant is that every height has exactly one block where `in_main_chain = 1`.

chia/cmds/db_upgrade_func.py
arvidn: right. The existing store classes, `BlockStore`, `CoinStore` and `HintStore` use the `DBWrapper` class to access the database. They specifically use the `aiosqlite` interface to the database, which means we can't use those classes when writing to the DB with the (synchronous) `sqlite3` API.

arvidn: The only reason we created those classes here though, was to create the indices. So, those statements were copy-pasted out of those classes.

chia/full_node/block_store.py
arvidn: technically, it's sometimes possible to slightly alter indices without breaking compatibility with the existing inserts and lookups. That's mainly what I had in mind when writing this comment.

PR #10264, opened 2022-02-16
Remove the "Cleaning up temp keychain in dir" message to reduce noise.

paninaro:


tests/util/keyring.py
```diff
@@ -195,7 +195,6 @@ def cleanup(self):
         if self.delete_on_cleanup:
             self.keychain.keyring_wrapper.keyring.cleanup_keyring_file_watcher()
             temp_dir = self.keychain._temp_dir
-            print(f"Cleaning up temp keychain in dir: {temp_dir}")
             shutil.rmtree(temp_dir)
```
Nikolaj-K, line 198: One could in that case also just pass the member instead of the introduced handle `temp_dir`, which is now not used beyond the line that follows.

paninaro: Updated

arvidn:
this looks reasonable to me. But, it's a bit surprising that the removing of keys (which seems like a test feature) is implemented in the production code. Is this really used by the production code, or is it just used by tests?

paninaro:
> this looks reasonable to me. But, it's a bit surprising that the removing of keys (which seems like a test feature) is implemented in the production code. Is this really used by the production code, or is it just used by tests?

Correct, this is only used by tests. This file implements the TempKeyring() class which is only used by tests to avoid populating ~/.chia_keys/keyring.yaml, or one of the OS-provided key stores.

arvidn:
oh, sorry. I failed to to realize this was in fact part of the test code.

aqk (approved):

altendky (approved):
An info or debug level log entry could be ok as well, presuming this doesn't happen a whole big bunch all at once.

arvidn (approved):

PR #10266, opened 2022-02-16
Fetch data improvements

fchirica:


chia/data_layer/data_layer.py
```diff
@@ -298,7 +306,7 @@ async def subscribe(self, store_id: bytes32, mode: DownloadMode, ip: str, port:
     async def unsubscribe(self, tree_id: bytes32) -> None:
         subscriptions = await self.get_subscriptions()
         if tree_id not in [subscription.tree_id for subscription in subscriptions]:
```
Nikolaj-K, line 308: By changing the bracket, this can return `True` as soon as the validating case is found.

```python
def f(n):
    print(n)
    return n

3 in [f(n) for n in range(10)]  # Prints all ten numbers
3 in (f(n) for n in range(10))  # Prints numbers up to 3
```

chia/data_layer/data_layer.py
```diff
@@ -289,6 +295,8 @@ async def subscribe(self, store_id: bytes32, mode: DownloadMode, ip: str, port:
         subscription = Subscription(store_id, mode, ip, port)
         subscriptions = await self.get_subscriptions()
         if subscription.tree_id in [subscription.tree_id for subscription in subscriptions]:
```
Nikolaj-K, line 297: By changing the bracket, this can return `True` as soon as the validating case is found.

```python
def f(n):
    print(n)
    return n

3 in [f(n) for n in range(10)]  # Prints all ten numbers
3 in (f(n) for n in range(10))  # Prints numbers up to 3
```

chia/data_layer/data_layer.py
```diff
@@ -241,8 +241,12 @@ async def fetch_and_validate(self, subscription: Subscription) -> None:
             f"Target wallet generation: {singleton_record.generation}."
         )
 
-        downloaded = await download_data(self.data_store, subscription, singleton_record.root)
+        try:
+            downloaded = await download_data(self.data_store, subscription, singleton_record.root)
+        except Exception:
```
aqk, line 246: Should probably list the Exceptions here, eg. `RuntimeError`, `HttpWhatever`, ...

almogdepaz (approved):

chia/data_layer/data_layer.py
altendky: Yeah, this will consume cancellation on Python 3.7.  It also totally silently hides typos etc.  This is a very general thing we need to discuss across the code base.

PR #10270, opened 2022-02-16
do not generate __init__.py in .pytest_cache

altendky:


tests/build-init-files.py
```diff
@@ -42,7 +45,7 @@ def command(verbose, root_str):
         path
         for tree_root in tree_roots
         for path in root.joinpath(tree_root).rglob("**/")
-        if "__pycache__" not in path.parts
+        if all(part not in ignores for part in path.parts)
```
Nikolaj-K, line 48: I think you could also do
```python
if not ignores.intersection(path.parts)
```
The `ignores` could also be upper caps.

altendky: I don't even like `CAP_CASE` in `C` so...  `:]`

I considered set operations and went with this since a set isn't really needed and I've encouraged more towards this form elsewhere recently.  Clearly performance doesn't matter here but consistency across the project is useful when trying to set examples.  That said, `.intersection()` would not have to create two extra sets like using `&` would, but it still has one extra.  It would be interesting to see some benchmarks on this.  Also, I prefer `len(a_collection) > 0` over leveraging truthiness of non-boolean stuff.

xdustinface (approved):

PR #10271, opened 2022-02-17
extend pools test workflow timeout to 60 minutes

altendky:
Recent runs on `main` have been taking over 40 minutes.


hoffmang9 (approved):
sigh...

altendky:
We're talking about what we can do to speed up the tests...  but we also need to get coverage in and that doesn't exactly make things faster.

hoffmang9:
Totally. Hence approve with a sigh.

Issue #10057, opened 2022-02-01
Poor quality error message when importing invalid key [Bug] 


danieljperry:
### What happened?

When importing incorrect mnemonics in the GUI, a vague error message is given.
To reproduce, select "IMPORT FROM MNEMONICS (24 WORDS)" in the GUI. Enter something invalid. The error given is 'NoneType' object is not subscriptable:

![error](https://user-images.githubusercontent.com/69756004/151912614-79ea0209-0af3-432a-9a1c-7f1e5cec908d.png)

This is not very useful for end users. Note that the log actually does give a useful error: ValueError: Invalid order of mnemonic words.

### Version

1.2.12 dev 207

### What platform are you using?

Linux

### What ui mode are you using?

GUI

### Relevant log output

```shell
2022-02-01T10:46:33.491 daemon __main__                   : ERROR    Error while handling message: Traceback (most recent call last):
  File "chia/daemon/server.py", line 217, in safe_handle
  File "chia/daemon/server.py", line 315, in handle_message
  File "chia/daemon/keychain_server.py", line 60, in handle_command
  File "chia/daemon/keychain_server.py", line 89, in add_private_key
  File "chia/util/keychain.py", line 126, in wrapper
  File "chia/util/keychain.py", line 279, in add_private_key
  File "chia/util/keychain.py", line 194, in bytes_from_mnemonic
ValueError: Invalid order of mnemonic words
```

PR #10274, opened 2022-02-17
Fix propagation of errors when adding a key with an invalid mnemonic

paninaro:
Adding a key via invalid mnemonic can raise various ValueError exceptions depending on the underlying issue. Most commonly when entered from the GUI, the checksum will fail, raising ValueError("Invalid order of mnemonic words"). Due to some tricky plumbing between Wallet RPC service <-> KeychainProxy <-> Daemon <-> KeychainServer, unhandled exceptions at the KeychainServer layer were not being propagated upstream. The KeychainProxy's call would subsequently time-out and when `handle_error()` was called, an incorrect dereference would lead to the error message "'NoneType' object is not subscriptable" being returned by the Wallet RPC service.

This PR fixes the propagation of errors from the KeychainServer by ensuring that exceptions are caught and that an error message is sent back to the caller.

In the specific case of entering an incorrect mnemonic (bad checksum), the error string "Invalid order of mnemonic words" will be returned.

Tests have been added to cover `add_private_key` RPC use cases.

Issue #10057

### DIFF
--- a/chia/daemon/keychain_proxy.py
+++ b/chia/daemon/keychain_proxy.py
@@ -108,9 +108,11 @@ class KeychainProxy(DaemonProxy):
                 message = error_details.get("message", "")
                 raise MalformedKeychainRequest(message)
             else:
-                err = f"{response['data'].get('command')} failed with error: {error}"
-                self.log.error(f"{err}")
-                raise Exception(f"{err}")
+                # Try to construct a more informative error message including the call that failed
+                if "command" in response["data"]:
+                    err = f"{response['data'].get('command')} failed with error: {error}"
+                    raise Exception(f"{err}")
+                raise Exception(f"{error}")
 
     async def add_private_key(self, mnemonic: str, passphrase: str) -> PrivateKey:
         """
--- a/chia/daemon/keychain_server.py
+++ b/chia/daemon/keychain_server.py
@@ -56,21 +56,25 @@ class KeychainServer:
         return keychain
 
     async def handle_command(self, command, data) -> Dict[str, Any]:
-        if command == "add_private_key":
-            return await self.add_private_key(cast(Dict[str, Any], data))
-        elif command == "check_keys":
-            return await self.check_keys(cast(Dict[str, Any], data))
-        elif command == "delete_all_keys":
-            return await self.delete_all_keys(cast(Dict[str, Any], data))
-        elif command == "delete_key_by_fingerprint":
-            return await self.delete_key_by_fingerprint(cast(Dict[str, Any], data))
-        elif command == "get_all_private_keys":
-            return await self.get_all_private_keys(cast(Dict[str, Any], data))
-        elif command == "get_first_private_key":
-            return await self.get_first_private_key(cast(Dict[str, Any], data))
-        elif command == "get_key_for_fingerprint":
-            return await self.get_key_for_fingerprint(cast(Dict[str, Any], data))
-        return {}
+        try:
+            if command == "add_private_key":
+                return await self.add_private_key(cast(Dict[str, Any], data))
+            elif command == "check_keys":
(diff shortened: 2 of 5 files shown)


chia/daemon/keychain_proxy.py
```diff
@@ -108,9 +108,11 @@ def handle_error(self, response: WsRpcMessage):
                 message = error_details.get("message", "")
                 raise MalformedKeychainRequest(message)
             else:
-                err = f"{response['data'].get('command')} failed with error: {error}"
-                self.log.error(f"{err}")
-                raise Exception(f"{err}")
```
paninaro, line 113: This is where the original `'NoneType' object is not subscriptable` exception was being raised.

chia/daemon/keychain_server.py
```diff
@@ -56,21 +56,25 @@ def get_keychain_for_request(self, request: Dict[str, Any]):
...
+            elif command == "get_all_private_keys":
+                return await self.get_all_private_keys(cast(Dict[str, Any], data))
+            elif command == "get_first_private_key":
+                return await self.get_first_private_key(cast(Dict[str, Any], data))
+            elif command == "get_key_for_fingerprint":
+                return await self.get_key_for_fingerprint(cast(Dict[str, Any], data))
+            return {}
+        except Exception as e:
+            log.exception(e)
+            return {"success": False, "error": str(e), "command": command}
```
paninaro, line 77: "command" is now inserted into the response dict for any exception that isn't handled at a lower level.

chia/daemon/keychain_server.py
```diff
@@ -93,6 +97,12 @@ async def add_private_key(self, request: Dict[str, Any]) -> Dict[str, Any]:
                 "error": KEYCHAIN_ERR_KEYERROR,
                 "error_details": {"message": f"The word '{e.args[0]}' is incorrect.'", "word": e.args[0]},
             }
+        except ValueError as e:
+            log.exception(e)
+            return {
+                "success": False,
+                "error": str(e),
+            }
```
paninaro, line 105: "Invalid order of mnemonic words" is a ValueError raised by `bytes_from_mnemonic()`. To preserve the error message as-is, the ValueError exception case is handled and simply returns the wrapped error message.

chia/daemon/keychain_proxy.py
altendky: `.get()` returns `None` if the key isn't present.  Unrelated to this situation, you can specify an alternate default if you want.

```pycon
Python 3.10.1 (main, Jan 12 2022, 20:28:45) [GCC 9.3.0] on linux
>>> {}.get("something")
>>> print({}.get("something"))
None
```

So `response` must be `None` actually.  But... Above there is `response["data"]` so it can't be `None`.

How as this identified as the source?

paninaro: Sorry about that, not sure why my memory of the root cause pointed at this area. You're correct that this isn't the main issue. I think during the course of making changes I noticed that 'None' was referenced as the command that failed, so perhaps that's where the confusion arose.

paninaro:
While this PR addresses uncaught exceptions from triggering a timeout in `keychain_proxy`, if a timeout is experienced for some other reason, `get_response_for_request()` will still raise a exception due to an unsafe dereference. I'll address this separately as #10288

chia/daemon/keychain_proxy.py
```diff
@@ -108,9 +108,11 @@ def handle_error(self, response: WsRpcMessage):
...
+                # Try to construct a more informative error message including the call that failed
+                if "command" in response["data"]:
+                    err = f"{response['data'].get('command')} failed with error: {error}"
```
altendky, line 113:
```suggestion
                command = response.get("data")
                if command is not None:
                    err = f"{command} failed with error: {error}"
```

altendky (approved):

PR #9130, opened 2021-11-04
change constant

Yostra:

PR #10278, opened 2022-02-17
Change back constant

jack60612:
Reverts #9130

### DIFF
--- a/chia/consensus/default_constants.py
+++ b/chia/consensus/default_constants.py
@@ -38,8 +38,8 @@ testnet_kwargs = {
         "3d8765d3a597ec1d99663f6c9816d915b9f68613ac94009884c4addaefcce6af"
     ),
     "MAX_VDF_WITNESS_SIZE": 64,
-    # Size of mempool = 50x the size of block # temporary change until #9125 gets in
-    "MEMPOOL_BLOCK_BUFFER": 10,
+    # Size of mempool = 50x the size of block
+    "MEMPOOL_BLOCK_BUFFER": 50,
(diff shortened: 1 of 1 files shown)


jack60612:
Thanks @neurosis69

mariano54 (approved):

wjblanke:
this isn't going to affect the majority of nodes already out there. the mempool will still be pruned by peers. why are we increasing this again?

mariano54:
We are setting it back to where it was, and I think it should probably be even higher. We want to support sudden bursts of activity like NFT minting, or any other app that incentivizes users to transact quickly. 

Cardano had serious issues because their mempool was too small and most transactions ended up getting dropped.

mariano54:
@wjblanke bump

mariano54:
Is there any reason not to merge this @wjblanke ?

altendky:
I'm removing the `ready_to_merge` label as this is not getting merged right now.  I think a discussion is needed in other venues to identify the path forward.  In the mean time, the `ready_to_merge` list will be cleaner.

mariano54:
There are significant usability issues from keeping a small mempool, and the people most familiar with this code are in favor of merging it. What is the argument for not merging it @wjblanke @altendky ?

hoffmang9:
I am specifically adding ready_to_merge back. This was a temporary regression that is  being held back from reversion without reason for breaking the original design.

If there is a valid test that needs to be completed, then lay it out, or revert back to the constant as shipped on 3/19/21 and 5/3/21.

PR #10282, opened 2022-02-17
Fix lineage proofs primary key constraint.

AmineKhaldi:


Quexington (approved):
Yep that's a typo alright

PR #10283, opened 2022-02-17
Atari add get_kv_diff endpoint

almogdepaz:


tests/core/data_layer/test_data_rpc.py
```diff
@@ -421,3 +422,91 @@ async def test_get_root_history(one_wallet_node_and_rpc: nodes) -> None:
...
+            {
+                "id": store_id1.hex(),
+                "hash_1": history["root_history"][1]["root_hash"].hex(),
+                "hash_2": history["root_history"][2]["root_hash"].hex(),
+            }
+        )
+        assert len(diff_res["diff"]) == 2
+        diff4 = {"type": "INSERT", "key": key4.hex(), "value": value4.hex()}
+        diff5 = {"type": "INSERT", "key": key5.hex(), "value": value5.hex()}
+        assert diff4 in diff_res["diff"]
```
fchirica, line 511: probably not super important, but I would add a {type: "DELETE"} test here as well.

fchirica (dismissed):

chia/rpc/data_layer_rpc_api.py
```diff
@@ -63,7 +63,8 @@ def get_routes(self) -> Dict[str, Callable[[Any], Any]]:
             "/delete_key": self.delete_key,
             "/insert": self.insert,
             "/subscribe": self.subscribe,
-            "/unsubscribe": self.unsubscribe,
+            "/get_kv_diff": self.get_kv_diff,
+            "/get_root_history": self.get_root_history,
```
altendky, line 67: Should this be removing the `/unsubscribe` endpoint?  I see that `/get_root_history` isn't being added here but the missing endpoint entry is getting filled in.

```suggestion
            "/unsubscribe": self.unsubscribe,
            "/get_kv_diff": self.get_kv_diff,
            "/get_root_history": self.get_root_history,
```

altendky (changes requested):

chia/rpc/data_layer_rpc_api.py
almogdepaz: Thanks for catching, fixed

altendky (approved):
Approving based on @fchirica's previous approval and the correction to the issue I noticed.

PR #10280, opened 2022-02-17
apply isort

altendky:

PR #10284, opened 2022-02-17
Fix fee validation in cat_spend wallet RPC API

AmineKhaldi:


Quexington (approved):
Makes sense to me

Yostra (commented):
What's up with the import changes ?

AmineKhaldi:
> What's up with the import changes ?

They're in a separate commit, required by the pre-commit hook in order to get further into the tests. See PR #10280

PR #10303, opened 2022-02-18
Get local root.

fchirica:


almogdepaz (approved):
lgmt, there is no cli option for get_local_root but not sure if we need it

PR #10307, opened 2022-02-18
use free listen ports when running tests

arvidn:
The purpose of this patch is to make tests more reliable. Currently, they rely on hard coded listen ports to be available on the machine running the tests. This is problematic for two main reasons:

1. You may be unlucky and have some other process already use one of the ports (or there may be a left-over process from a previous test run).
2. It prevents certain tests from being run in parallel, significantly extending the time it takes to run all the tests

## implementation

The main mechanism is implemented in `find_available_listen_port()` ([here](https://github.com/Chia-Network/chia-blockchain/pull/10307/files#diff-4d6d97099999cfdcd1b709572c5de9dc23edf7d77e05edaf6e2bae16930fdd2eR4)) which finds a port that's free. This function is then used in tests to configure them to use those (dynamic) ports rather than hard coded (static) ports.

The implementation of `find_available_listen_port()` is not sophisticated, but it's good enough. If we end up having issues with "address in use" errors, we can make it more sophisticated.

The main challenge has been to understand all the implicit dependencies between default service ports and tests (implicitly) relying on them. Particularly the daemon port (more on that in the code annotation).

### DIFF
--- a/tests/core/daemon/test_daemon.py
+++ b/tests/core/daemon/test_daemon.py
@@ -19,10 +19,6 @@ import pytest_asyncio
 
 
 class TestDaemon:
-    @pytest_asyncio.fixture(scope="function")
-    async def get_daemon(self, get_b_tools):
-        async for _ in setup_daemon(btools=get_b_tools):
-            yield _
 
     # TODO: Ideally, the db_version should be the (parameterized) db_version
     # fixture, to test all versions of the database schema. This doesn't work
@@ -48,7 +44,6 @@ class TestDaemon:
     async def get_b_tools(self, get_temp_keyring):
         local_b_tools = await create_block_tools_async(constants=test_constants_modified, keychain=get_temp_keyring)
         new_config = local_b_tools._config
-        new_config["daemon_port"] = 55401
         local_b_tools.change_config(new_config)
         return local_b_tools
 
@@ -58,9 +53,10 @@ class TestDaemon:
             yield get_b_tools
 
     @pytest.mark.asyncio
-    async def test_daemon_simulation(self, simulation, get_daemon, get_b_tools):
-        node1, node2, _, _, _, _, _, _, _, _, server1 = simulation
-        await server1.start_client(PeerInfo(self_hostname, uint16(21238)))
+    async def test_daemon_simulation(self, simulation, get_b_tools):
+        node1, node2, _, _, _, _, _, _, _, _, server1, daemon1 = simulation
+        node2_port = node2.full_node.config["port"]
+        await server1.start_client(PeerInfo(self_hostname, uint16(node2_port)))
 
         async def num_connections():
             count = len(node2.server.connection_by_type[NodeType.FULL_NODE].items())
@@ -73,7 +69,7 @@ class TestDaemon:
         ssl_context = get_b_tools.get_daemon_ssl_context()
 
         ws = await session.ws_connect(
-            "wss://127.0.0.1:55401",
+            f"wss://127.0.0.1:{daemon1.daemon_port}",
             autoclose=True,
             autoping=True,
             heartbeat=60,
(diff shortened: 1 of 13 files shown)


tests/setup_nodes.py
```diff
@@ -579,4 +633,10 @@ async def num_connections():
             node_api_1.full_node.server,
         )
 
+        if connect_to_daemon:
```
arvidn, line 636: the `connect_to_daemon` option is a bit tricky. this function (`setup_full_system`) did not previously *start* the daemon, but when `connect_to_daemon` was `True`, it would assume there was one running (on the default port) and connect to it.

Since we can't rely on default ports, this function now *starts* the daemon for `b_tools` (not `b_tools_1`) as well as connects to it. that's why the returned tuple has different sizes on these two cases.

tests/util/socket.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def find_available_listen_port(name: str = "free") -> int:
+    s = socket.socket()
+    s.bind(("127.0.0.1", 0))
+    addr = s.getsockname()
+    assert addr[1] > 0
+    s.close()
+    print(f"{name} port: {addr[1]}")
+    return int(addr[1])
```
arvidn, line 11: technically, I believe it would be valid for an operating system to return the same port every time, since the port from the last call isn't necessarily busy by the time we call it again. However, it looks like they behave reasonably by randomizing a port.

If this turns out to be a problem, we can implement the randomization ourselves, but I'd rather keep things simple as long as it works.

Naming ports and printing them helps a lot in interpreting logs, I think it's good to keep it.

tests/core/test_daemon_rpc.py
```diff
@@ -5,11 +5,17 @@
...
+from chia.util.config import save_config
 
 
 class TestDaemonRpc:
     @pytest_asyncio.fixture(scope="function")
     async def get_daemon(self):
+        bt._config["daemon_port"] = find_available_listen_port()
+        # unfortunately, the daemon's WebSocketServer loads the configs from
+        # disk, so the only way to configure its port is to write it to disk
+        save_config(bt.root_path, "config.yaml", bt._config)
```
arvidn, line 18: it's still a problem that `bt` is a global variable, shared between all tests. But that will hopefully be fixed in a separate PR.

aqk: https://github.com/Chia-Network/chia-blockchain/pull/10358

tests/core/daemon/test_daemon.py
```diff
@@ -19,10 +19,6 @@
 
 
 class TestDaemon:
-    @pytest_asyncio.fixture(scope="function")
-    async def get_daemon(self, get_b_tools):
```
arvidn, line 23: the daemon is started by `setup_full_system()` now (called by the `simulation` fixture below)

tests/simulation/test_simulation.py
```diff
@@ -52,10 +54,13 @@ async def simulation(self):
...
-        await server1.start_client(PeerInfo(self_hostname, uint16(21238)))
+
+        node1_port = node1.full_node.config["port"]
+        node2_port = node2.full_node.config["port"]
+        await server1.start_client(PeerInfo(self_hostname, uint16(node2_port)))
         # Use node2 to test node communication, since only node1 extends the chain.
-        await time_out_assert(1500, node_height_at_least, True, node2, 7)
-        await sanitizer_server.start_client(PeerInfo(self_hostname, uint16(21238)))
+        await time_out_assert(600, node_height_at_least, True, node2, 7)
+        await sanitizer_server.start_client(PeerInfo(self_hostname, uint16(node2_port)))
```
mariano54, line 63: We previously had issues with low timeouts here, some CIs can take a really long time. note that this is running the real timelord and farmer all in the same process.

mariano54 (approved):

tests/simulation/test_simulation.py
arvidn: granted, this is unrelated. While testing though, it was frustrating to have this test take 1500 seconds (as well as `test_daemon.py`) before failing. In all my successful local runs, it took 30 seconds to complete. I figured 20x margin would be enough.

Issue #10155, opened 2022-02-09
[Bug] content updates


esaung:
### What happened?

The following content strings need to be updated:
1.) upon quitting the Chia app, we need to update the content to be more generic. current content says, "Are you sure you want to quit? GUI plotting and farming will stop."
<img width="1060" alt="CleanShot 2022-02-21 at 09 56 19@2x" src="https://user-images.githubusercontent.com/663017/154998802-b4c8e70b-329b-4655-a7df-37b788826013.png">

2.) update content on the farmer/wallet selection home screen

3) Update the text in the walled dropdown from "Chia Wallet" to "Chia"

4) The Chia wallet should have "XCH" instead of the Chia logo to be more consistent with the other wallet types. 
<img width="506" alt="CleanShot 2022-02-21 at 10 01 46@2x" src="https://user-images.githubusercontent.com/663017/154999348-7f310ad9-126a-4572-88de-b06d9b16220e.png">

5.) modify the default content from "CAT Wallet" to "Chia Asset Token"
![Screen Shot 2022-02-25 at 3 51 31 PM](https://user-images.githubusercontent.com/93196638/155808076-143f5c1f-2b32-4ccb-a492-09052c4231da.png)

### Version

1.2.12 dev 263

### What platform are you using?

macOS

### What ui mode are you using?

GUI

sethjenks:
2) Update the text in the walled dropdown from "Chia Wallet" to "Chia"
3) The Chia wallet should have "XCH" instead of the Chia logo to be more consistent with the other wallet types.

seeden:
Updated via https://github.com/Chia-Network/chia-blockchain-gui/commit/b9b8c51249ab70b18b93e1383a37a34a71526602

PR #10308, opened 2022-02-18
Allow CAT autodiscovery, refactor CAT default naming

ojura:
Many users in the community want the wallet CAT autodiscovery feature back. 

This way, it has to be explicitly enabled in `config.yaml`.


ojura:
Example of a wallet scratched from sync with the option enabled:

image

chia/wallet/wallet_state_manager.py
```diff
@@ -620,7 +620,9 @@ async def fetch_parent_and_check_for_cat(
                 cat_puzzle = construct_cat_puzzle(CAT_MOD, bytes32(bytes(tail_hash)[1:]), our_inner_puzzle)
                 if cat_puzzle.get_tree_hash() != coin_state.coin.puzzle_hash:
                     return None, None
-                if bytes(tail_hash).hex()[2:] in self.default_cats:
+                if bytes(tail_hash).hex()[2:] in self.default_cats or self.config.get(
```
ojura, line 623: @Quexington please advise if `setdefault` would be more appropriate here

ojura:
There is a clear desire in the community to bring back this feature. Old builds are being circulated because this was removed, which is IMHO alarming:
paste

I think this base layer of implementation should be merged ASAP (of course provided that you're happy with the changes), and then there's time later to implement it in the GUI, add a proper warning when enabling it, etc.

ojura:
I have also done some refactoring of the default CAT wallet name generation.

It is now done in a single place in the entire codebase, and that is the function `CATWallet.default_wallet_name_for_unknown_cat`.

Previously, we had `CAT WALLET` hardcoded in 2 places, and also `CAT Wallet` in the rpc code, and in a test.

chia/wallet/wallet_state_manager.py
Quexington: No I think `get` is correct

Quexington (dismissed):
Looks great to me, thanks for your contribution!

chia/wallet/wallet_state_manager.py
ojura: Okay, keeping get. `setdefault` would add it to the dictionary if it's missing.

I thought it would make sense that new versions of Chia populate the config with their new parameters, so the user is aware that they exist when they scroll through their config.yaml.

ojura:
@Quexington this PR has been approved, but has not been merged yet and became stale against main two times so far because a couple of lines unrelated to the PR (but close to the changed lines) got touched.

Maintaining it is becoming a bit of a burden, could we please expedite the merge?

Quexington (approved):
Refreshing my approval, thanks again!

esaung:
@ojura - Thanks for your contribution. Can you see here - https://github.com/Chia-Network/chia-blockchain/issues/10155 ? we are in the process of making some content changes in the GUI in addition to the work that @ytx1991 made. There's an item that overlaps that I believe @seeden made/making changes for.

see #4 and #5 on the list. Generally we want to refer to any CATs just as CATs instead of calling them CAT wallet. The idea is that we have 1 wallet that will contain many Chia Asset Tokens. We'd want the CLI to report the same way, but we don't need to go far as to change all the references at the code level.

@sethjenks as FYI and to add anything else from a design perspective.

ojura:
@esaung the idea in this PR was that the default name should contain the part of the tail to make it easier for the user to tell them apart after syncing an empty wallet from scratch with autodiscovery enabled. I think "Chia Asset Token XXXXXXXXXXXXXXXX" would be too long. What do you propose?

esaung:
the CAT + Tail ID part is okay. it's the default content underneath where it says "CAT Wallet" - we are changing that to just say Chia Asset Token. I think showing the full TAIL Id is a bit much since it's going to get ellipisified. @sethjenks  - any recommendations here - first 6 digits of the Tail good enough - "CAT 1cc8af..." - we could probably go to 8 digits if we expect any collisions since the user will get the full TAIL ID from the "show details" section.

We are also changing the chia part to show XCH, and take out the "wallet" part of the content. the note is more to make sure the changes are rebased on your PR.

PR #10312, opened 2022-02-19
fix service restart

altendky:
https://github.com/Chia-Network/chia-blockchain/pull/10233

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -1087,7 +1087,7 @@ class WebSocketServer:
             else:
                 self.log.info(f"Service {service_command} already running")
                 already_running = True
-        elif service_command in self.connections:
+        elif len(self.connections.get(service_command, [])) > 0:
             # If the service was started manually (not launched by the daemon), we should
             # have a connection to it.
             self.log.info(f"Service {service_command} already registered")


chia/daemon/server.py
```diff
@@ -1087,7 +1087,7 @@ async def start_service(self, request: Dict[str, Any]):
             else:
                 self.log.info(f"Service {service_command} already running")
                 already_running = True
-        elif service_command in self.connections:
```
arvidn, line 1090: is there a good reason we sometimes don't have this member? is there an important distinction between having the member and having it as an empty list?

arvidn (approved):

chia/daemon/server.py
altendky: I think that we 'always' have `self.connections` as an attribute, are you suggesting that the `self.connections` `dict` could be initialized with all the services and empty lists?  It looks like not-in-the-`dict` vs. empty list is just whether the service has ever been connected to this specific daemon process.

It may be that this new block should have ended up living in `.is_running()` instead of here.

arvidn: oh, I think I misread this. `self.connections` is a Dict mapping "commands" to a list. the current logic doesn't work because we still have the dict entry when the list is empty.

paninaro (approved):

PR #10316, opened 2022-02-19
slight simplification of interpreting the bytes received by the timelord

arvidn:
avoid redundant round-trips to strings.

Another unintuitive thing in the timelord protocol is that the length prefix (4 bytes) is sent in binary, but the proof itself is sent as a hex string (which we then have to turn into binary before interpreting it).

It seems like it would have been simpler to just send the proof itself as binary too.

### DIFF
--- a/chia/timelord/timelord.py
+++ b/chia/timelord/timelord.py
@@ -953,100 +953,94 @@ class Timelord:
                         self.vdf_failures_count += 1
                     break
 
-                msg = ""
-                try:
-                    msg = data.decode()
-                except Exception:
-                    pass
-                if msg == "STOP":
+                if data == b"STOP":
                     log.debug(f"Stopped client running on ip {ip}.")
                     async with self.lock:
                         writer.write(b"ACK")
                         await writer.drain()
                     break
+                try:
+                    # This must be a proof, 4 bytes is length prefix
+                    length = int.from_bytes(data, "big")
+                    proof = await reader.readexactly(length)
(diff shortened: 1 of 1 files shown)


chia/timelord/timelord.py
```diff
@@ -953,100 +953,94 @@ async def _do_process_communication(
...
-                        asyncio.IncompleteReadError,
-                        ConnectionResetError,
-                        Exception,
-                    ) as e:
-                        log.warning(f"{type(e)} {e}")
-                        async with self.lock:
-                            self.vdf_failures.append((chain, proof_label))
-                            self.vdf_failures_count += 1
-                        break
+                try:
```
arvidn, line 962: indentation changed here. please ignore whitespace

chia/timelord/timelord.py
```diff
@@ -953,100 +953,94 @@ async def _do_process_communication(
                         self.vdf_failures_count += 1
                     break
 
-                msg = ""
-                try:
-                    msg = data.decode()
-                except Exception:
-                    pass
-                if msg == "STOP":
```
arvidn, line 961: the only thing we use `msg` for is to compare it to "STOP". Currently we interpret these 4 bytes as utf-8, just to convert it to a string we then compare against "STOP". The vast majority of times, this conversion fails, because the bytes aren't utf-8, that's why we just ignore any exception coming out of the `decode()` function.

There's no real point of converting the bytes to a string, instead, we can just compare it to the *bytes* b"STOP".

fchirica (approved):

PR #10322, opened 2022-02-19
spawn (not fork) for multiprocessing

altendky:
image

Draft for:
- [x] self-testing
- [x] Explore if the title is working and how it relates to what `htop` shows
- [x] Probably just inherit the process name, plus a `_w` suffix or such to indicate it is a worker?  I think the lengths are pretty limited.


mariano54 (commented):
Confirmed to help with wallet memory usage in Linux, significantly. Hopefully will have the same effect on full node.

altendky:
image

Maybe this isn't showing the 'process title', but I guess I should look into this.  Definitely annoying that a `chia_` filter doesn't show the workers.

arvidn:
it would be nice if we could just inherit the proctitle instead. (since we mark test processes with special names now)

altendky:
At first thought anyways, sounds reasonable.  Thanks.

mariano54:
This is running fine for me in full node and wallet. Full node shows one entry per subprocess but wallet only showing one total. In manjaro system monitor.

altendky:
So in some cases we can see more than 15/16 characters.

```console
$ ps
    PID TTY          TIME CMD
3199459 pts/1    00:00:00 bash
3207126 pts/1    00:00:05 chia_daemon
3207260 pts/1    00:00:04 chia_harvester
3207261 pts/1    00:00:05 chia_farmer
3207262 pts/1    00:00:33 chia_full_node
3207263 pts/1    00:00:38 chia_wallet
3207354 pts/1    00:00:00 python
3207404 pts/1    00:00:00 python
3208191 pts/1    00:00:24 chia_wallet_wor
3208192 pts/1    00:00:19 chia_wallet_wor
3208193 pts/1    00:00:19 chia_wallet_wor
3208194 pts/1    00:00:19 chia_wallet_wor
3208462 pts/1    00:00:05 chia_full_node_
3208463 pts/1    00:00:09 chia_full_node_
3208464 pts/1    00:00:06 chia_full_node_
3208465 pts/1    00:00:05 chia_full_node_
3208730 pts/1    00:00:00 ps
```

```console
$ ps aux | grep -E "(chia_|multipro)"
ubuntu   3207126  3.2  1.3 365452 53516 pts/1    Sl   20:45   0:05 chia_daemon
ubuntu   3207260  2.9  1.3 293032 54024 pts/1    Sl   20:45   0:04 chia_harvester
ubuntu   3207261  3.7  1.4 220812 56236 pts/1    Sl   20:45   0:05 chia_farmer
ubuntu   3207262 23.9  4.3 544180 169020 pts/1   Sl   20:45   0:38 chia_full_node
ubuntu   3207263 33.3  6.1 583780 238136 pts/1   Rl   20:45   0:53 chia_wallet
ubuntu   3207354  0.0  0.2  16456 10336 pts/1    S    20:45   0:00 /farm/dev/chia-blockchain/venv/bin/python -c from multiprocessing.resource_tracker import main;main(12)
ubuntu   3207404  0.0  0.2  16456 10200 pts/1    S    20:45   0:00 /farm/dev/chia-blockchain/venv/bin/python -c from multiprocessing.resource_tracker import main;main(14)
ubuntu   3208191 51.8  1.1  54824 43632 pts/1    S    20:46   1:00 chia_wallet_worker
ubuntu   3208192 43.5  0.8  46288 31212 pts/1    S    20:46   0:50 chia_wallet_worker
ubuntu   3208193 43.2  0.7  46272 30908 pts/1    S    20:46   0:50 chia_wallet_worker
ubuntu   3208194 43.0  0.7  46276 31032 pts/1    S    20:46   0:49 chia_wallet_worker
ubuntu   3208906  0.0  0.0   7692   620 pts/0    S+   20:47   0:00 grep --color=auto -E (chia_|multipro)
```

image

chia/__init__.py
```diff
@@ -1,5 +1,14 @@
+import multiprocessing
+
 from pkg_resources import DistributionNotFound, get_distribution, resource_filename
 
+start_method = "spawn"
+try:
+    multiprocessing.set_start_method(start_method)
+except RuntimeError:
+    if multiprocessing.get_start_method(allow_none=True) != start_method:
```
arvidn, line 9: do I understand this correctly; you want to ignore any runtime error from this call, *except* the ones where the "start method" was set. Is that a likely thing to happen?

I think it might warrant a comment if you expect some error here that would still leave us in a state we're happy with.
Without such use case, it would seem simpler to just leave this a plain call, and if it fails, the process dies.

chia/util/setproctitle.py
```diff
@@ -9,3 +9,11 @@
...
     if no_setproctitle is False:
         pysetproctitle.setproctitle(ps_name)
+
+
+def getproctitle() -> str:
+    if no_setproctitle is False:
+        # TODO: add type hints to setproctitle
+        return pysetproctitle.getproctitle()  # type: ignore[no-any-return]
+
+    return ""
```
arvidn, line 19: I suppose in this case, `setproctitle()` isn't likely doing anything anyway, right? (otherwise it might be nice to leave *some* clue in the name, like "chia")

arvidn (dismissed):

mariano54:
Let's merge now and start testing this @wjblanke ?

chia/__init__.py
altendky: Yes, I'll add a comment.

I think the gist was that if it has already been set then you get an error setting regardless of setting it to the same thing or not.  I could check first then set if needed, unless I'm forgetting some nuance.  But I went this route that can be read as "try to set the start method, if there is an error then ignore it if the start method is what we want".

Note that on the test side of things I also set the start method since we have plugins using multiprocessing so even with this being an application we already have "other code" setting the start method before us.

chia/util/setproctitle.py
altendky: Yeah, this is the 'we do not have the ability to set titles' case.  So, this string only matters towards the end of being able to be combined with other strings right now.  If at some point we decided to print our title out the CLI or something...  Well, I might just say that we should have a name we manage and put to the CLI and the title instead of reading out of the title.

mariano54 (approved):

PR #10325, opened 2022-02-20
avoid looping to avoid an unwanted random selection

altendky:


altendky:
https://github.com/Chia-Network/chia-blockchain/pull/10326

PR #10326, opened 2022-02-20
rework _fetch_header_blocks_inner()

altendky:
A follow-up to https://github.com/Chia-Network/chia-blockchain/pull/10324.

Draft for:
- [x] Sleeping on it :zzz: :sleeping_bed:

### DIFF
--- a/chia/wallet/util/wallet_sync_utils.py
+++ b/chia/wallet/util/wallet_sync_utils.py
@@ -290,25 +290,21 @@ async def _fetch_header_blocks_inner(
 ) -> Optional[RespondHeaderBlocks]:
     # We will modify this list, don't modify passed parameters.
     remaining_peers = list(all_peers)
-    response = None
 
     while len(remaining_peers) > 0:
         peer = random.choice(remaining_peers)
 
         response = await peer.request_header_blocks(request)
 
-        if response is None or isinstance(response, RespondHeaderBlocks):
-            # None only occurs for a badly formed request that .request_header_blocks()
-            # rejects locally.  No need to close the peer for this.  All attempts will
-            # fail.  Give up.  If the response is good, return it.
-            break
+        if isinstance(response, RespondHeaderBlocks):
+            return response
 
         # Request to peer failed in some way, close the connection and remove the peer
-        # from our list.
+        # from our local list.
         await peer.close()
         remaining_peers.remove(peer)
 
-    return response
+    return None
 
 
 async def fetch_header_blocks_in_range(
--- a/chia/wallet/util/wallet_sync_utils.py
+++ b/chia/wallet/util/wallet_sync_utils.py
@@ -286,36 +286,38 @@ def last_change_height_cs(cs: CoinState) -> uint32:
 
 
 async def _fetch_header_blocks_inner(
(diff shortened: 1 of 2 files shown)


altendky:
I feel like there's some slight next-layer-up design smell perhaps, but this should address a few oddities in the existing code.

chia/wallet/util/wallet_sync_utils.py
```diff
@@ -286,36 +285,38 @@ def last_change_height_cs(cs: CoinState) -> uint32:
...
-        if isinstance(res, RespondHeaderBlocks):
-            return res
-        else:
-            return None
+    # We will modify this list, don't modify passed parameters.
+    remaining_peers = list(all_peers)
+    response = None
+    peer = None
+
+    if selected_peer_node_id is not None:
```
mariano54, line 295: We don't want to ask the selected peer by default, otherwise we will be requesting from the same peer (sync target peer) over and over, which is not efficient. The selected peer was meant to be the fallback if the random peer failed.

mariano54: If we are contacting all peers, I think we can just remove that argument.

chia/wallet/util/wallet_sync_utils.py
```diff
@@ -286,36 +285,38 @@ def last_change_height_cs(cs: CoinState) -> uint32:
...
+        except StopIteration:
+            peer = None
+
+    while len(remaining_peers) > 0:
+        if peer is None:
+            peer = random.choice(remaining_peers)
+
+        response = await peer.request_header_blocks(request)
+
+        if response is None or isinstance(response, RespondHeaderBlocks):
```
mariano54, line 307: response is None happens when the peer times out, which maybe we should disconnect from? Not clear though

mariano54 (changes requested):

chia/wallet/util/wallet_sync_utils.py
altendky: Cool, that makes the function much simpler.

chia/wallet/util/wallet_sync_utils.py
altendky: Blah, I think I looked at the function definition for the remote side and forgot all the interrim code...  If a timeout should result in a disconnect, should that be done in those middle layers rather than at each request point?

chia/wallet/util/wallet_sync_utils.py
altendky: Updated in response to this thread.

chia/wallet/util/wallet_sync_utils.py
mariano54: It should be done here for now, disconnecting is a serious thing and I don't want to just automatically do it for any time out

mariano54 (dismissed):

chia/wallet/util/wallet_sync_utils.py
altendky: Alrighty, I'll update to disconnecting for anything except a `RespondHeaderBlocks`.

mariano54 (approved):

PR #10330, opened 2022-02-20
improve exception safety in tests

arvidn:


mariano54:
If there is an actual error thrown during the tests, this will execute the finally, and then re-raise the Exception right?

tests/core/daemon/test_daemon.py
```diff
@@ -72,55 +72,56 @@ async def num_connections():
...
+                await ws.send_str(payload)
+
+                await asyncio.sleep(5)
+                blockchain_state_found = False
+                while not message_queue.empty():
+                    message = await message_queue.get()
+                    if message["command"] == "get_blockchain_state":
+                        blockchain_state_found = True
+            finally:
+                read_handler.cancel()
```
altendky, line 123: Now we get into a funny area.  You should await the task as well.  But, you want to ignore the `asyncio.CancelledError` that is going to come out of it.  You fed it in, you know you want to cancel up to this point.  But...  you will be awaiting so this task could be getting cancelled as well.  How do you tell which task was cancelled?  One cancellation you want to ignore, the other you want to propagate.  https://docs.python.org/3.10/library/asyncio-task.html#asyncio.Task.cancel in Python 3.9 they added a `msg=` parameter so we could make distinguishable cancellation requests.  I really need to look into whether we can ease into use of AnyIO piece by piece.

arvidn:
yes. technically the exception is never caught, so it isn't re-raised.

The main risk of doing this is that if the exception is a `KeyboardInterrupt` and the cleanup fails with some other exception, the type of the exception being thrown may change.

tests/core/cmds/test_keys.py
```diff
@@ -69,8 +69,10 @@ class TestKeysCommands:
     @pytest.fixture(scope="function")
     def empty_keyring(self):
         with TempKeyring(user="user-chia-1.8", service="chia-user-chia-1.8") as keychain:
-            yield keychain
-            KeyringWrapper.cleanup_shared_instance()
+            try:
+                yield keychain
+            finally:
+                KeyringWrapper.cleanup_shared_instance()
```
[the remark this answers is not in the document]
arvidn: @paninaro would know I suppose. but this looks like a call that should be made regardless of the test failing or not.

tests/conftest.py
```diff
@@ -24,11 +24,12 @@ async def empty_blockchain(request):
...
-
-    await connection.close()
-    bc1.shut_down()
-    db_path.unlink()
+    try:
+        yield bc1
+    finally:
+        await connection.close()
+        bc1.shut_down()
+        db_path.unlink()
```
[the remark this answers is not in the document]
arvidn: perhaps I've misunderstood the mechanics of these. We seem to use generators as fixtures whenever we need to clean up something at the end. So I was under the impression that the generator is always given an opportunity to do that.

tests/core/daemon/test_daemon.py
arvidn: so.. `await` on a task throws `CancelledError` when it's cancelled? that seems pretty unintuitive. The handle to the task isn't a future, right? or does the return value of the task get passed back though the handle? if so, I suppose it makes sense to receive the exception there too.

One would think there's a way to wait on a future, separate from asking for its value though.

altendky: Yeah, awaiting a task is one way you 'reattach' it and accept it's output, either return or exception.  `x = await af()` is at least pretty similar to `x = await asyncio.create_task(af())` in terms of running the coroutine returned by `af()` and feeding out the return value or exceptions.  You can also `asyncio.gather()`, `asyncio.wait()`, `asyncio.wait_for()`, and probably a few others depending on various needs.

https://docs.python.org/3.10/library/asyncio-task.html#task-object
> A [Future-like](https://docs.python.org/3.10/library/asyncio-future.html#asyncio.Future) object that runs a Python [coroutine](https://docs.python.org/3.10/library/asyncio-task.html#coroutine).

PR #10331, opened 2022-02-20
make blockchain.height_to_hash() return None on failure

arvidn:
rather than assert


arvidn:
this fixes the `full_node_api` to respond with a `RejectHeaderRequest` message, here: https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/full_node_api.py#L1111

mariano54 (approved):

PR #10333, opened 2022-02-20
streamable: Implement `Dict` support

xdustinface:
Enables

```python
@dataclass(frozen=True)
@streamable
class A(Streamable):
    a: Dict[str, uint8]
```
instead of what we currently do in such a case
```python
@dataclass(frozen=True)
@streamable
class A(Streamable):
    a: List[Tuple[str, uint8]]
```

Note: There are some places in the code where we probably can just swap them since the byte representation of those is the same but we need to consider that the JSON representation differs so we need to make sure the places where we swap it only depend on the byte representation.

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -16,7 +16,7 @@ from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.byte_types import hexstr_to_bytes
 from chia.util.hash import std_hash
 from chia.util.ints import int64, int512, uint32, uint64, uint128
-from chia.util.type_checking import is_type_List, is_type_SpecificOptional, is_type_Tuple, strictdataclass
+from chia.util.type_checking import is_type_dict, is_type_List, is_type_SpecificOptional, is_type_Tuple, strictdataclass
 
 if sys.version_info < (3, 8):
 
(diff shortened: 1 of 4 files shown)


mariano54:
One reason for not implementing Dict is that streamable is used as a consensus critical deterministic encoding, used for hashing. Adding a Dict object potentially breaks this pattern by having a serialization which is not deterministic (in which order do we encode the elements?). This can be a problem for example with languages that have Dictionaries where the ordering is not deterministic.

What is the use case for adding Dict?

xdustinface:
> One reason for not implementing Dict is that streamable is used as a consensus critical deterministic encoding, used for hashing. Adding a Dict object potentially breaks this pattern by having a serialization which is not deterministic (in which order do we encode the elements?). This can be a problem for example with languages that have Dictionaries where the ordering is not deterministic.

I see, makes sense. But don't we just defer this issue to the `Dict` <-> `List[Tuple[...]]` conversion then (assumed one would normally implement such structure as a dict internally)?

> What is the use case for adding Dict?

There is not really a use case so far besides replacing `List[Tuple[...]]`. I just worked myself through the "how is all the streamable fun connected" and thought i might as well just implement `Dict` while doing this. I _maybe_ could have a use-case soon for RPC parameter validation but feel free to close this one for now, i can re-open if it will really be required somehow :)

mariano54:
List[Tuple] implies that ordering is important, that's why it's a list, but a dict is confusing and determinism is not obvious unless you go and read the spec of how we expect people to stream these dicts, so it increases streamable complexity. I'll close for now

PR #10335, opened 2022-02-20
tests: Use `pytest.raises` in some places

xdustinface:


xdustinface:
Okay by adding those `raises` it uncovered that:

https://github.com/Chia-Network/chia-blockchain/blob/1073f9d5dd63c866c0b071915d074c3c92bfaf40/tests/core/util/test_type_checking.py#L70-L74

Doesn't make a lot sense (or i just don't get it). It "tests" the same statement as above, which is obvious to succeed and the `assert False` below the statement just triggers the `except AssertionError` which passes over.

I just added [5cf53f3](https://github.com/Chia-Network/chia-blockchain/pull/10335/commits/5cf53f3e791e842aaeed044026c4f8c8af9d334f) instead :)

mariano54 (approved):

arvidn (approved):

PR #10337, opened 2022-02-20
Check if secrets are available for triggering dev docker builds

cmmarslender:


xdustinface (approved):
Thanks! Worked: https://github.com/Chia-Network/chia-blockchain/runs/5265088683.

PR #10338, opened 2022-02-20
test: Skip "empty" directories in `build-init-files.py`

xdustinface:
Empty ones or such with only ignored stuff inside. It's just annoying if you have a branch which adds a new directory with new files and you switch to a branch which doesn't have those new files, running `pre-commit` always adds the init files in the new directory from the other branch because the folder from the other branch stays there and is empty.

### DIFF
--- a/tests/build-init-files.py
+++ b/tests/build-init-files.py
@@ -26,6 +26,10 @@ log_levels = {
 ignores = {"__pycache__", ".pytest_cache"}
 
 
+def ignored_path(path: pathlib.Path):
+    return any(part in ignores for part in path.parts)
+
+
 @click.command()
 @click.option(
     "-r", "--root", "root_str", type=click.Path(dir_okay=True, file_okay=False, resolve_path=True), default="."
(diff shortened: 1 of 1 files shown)


Nikolaj-K:
I'd maybe add a comment that this is what this check does - might be not entirely obvious.
Btw. since it seems you searched for a name for the "path in path", you could also go `any(map(valid_path, path.iterdir()))`

tests/build-init-files.py
```diff
@@ -45,7 +49,7 @@ def command(verbose, root_str):
         path
         for tree_root in tree_roots
         for path in root.joinpath(tree_root).rglob("**/")
-        if all(part not in ignores for part in path.parts)
+        if not ignored_path(path) and not all(map(ignored_path, path.iterdir()))
```
altendky, line 52: So we want to include it if it itself is not an ignored path and also it contains a path which isn't ignored.  `any(not ignored_path(item) for item in path.iterdir())`?  I have a strong preference for comprehensions over `map()` and `filter()`.  But this won't handle two layers of dirs with nothing else in either of them.  I think this in part follows from the fact that pycharm goes and uses the file's directory as cwd when running it, combined with us and the tools we use dumping files into cwd.  Otherwise the directories would go away when we switch branches, I think.  Do we want to take one more stop gap step or consider a new design that is much more aware?  Or maybe look at git instead of just directories?  I'm fine either way.

PR #10341, opened 2022-02-20
Align the wallet node's weight proof timeout with the full node's value.

AmineKhaldi:


chia/server/start_wallet.py
```diff
@@ -29,14 +29,18 @@ def service_kwargs_for_wallet(
 ) -> Dict:
     overrides = config["network_overrides"]["constants"][config["selected_network"]]
     updated_constants = consensus_constants.replace_str_to_bytes(**overrides)
+    full_node_config = load_config(DEFAULT_ROOT_PATH, "config.yaml", "full_node")
```
mariano54, line 32: This should be in the wallet config not full node config

mariano54 (changes requested):

chia/server/start_wallet.py
AmineKhaldi: Hmm I thought of that originally, but then wondered if we're fine with that value existing in two places in config.yaml?

altendky: We generally deal with options in multiple places but wanting an (optional) single point of truth with yaml anchors.  See `self_hostname` and the like.

mariano54 (dismissed):

chia/wallet/wallet_node.py
```diff
@@ -979,7 +979,13 @@ async def fetch_and_validate_the_weight_proof(
         assert self.wallet_state_manager.weight_proof_handler is not None
 
         weight_request = RequestProofOfWeight(peak.height, peak.header_hash)
-        weight_proof_response: RespondProofOfWeight = await peer.request_proof_of_weight(weight_request, timeout=60)
+        wp_timeout = 360
+        if "weight_proof_timeout" in self.config:
+            wp_timeout = self.config["weight_proof_timeout"]
```
altendky, line 984:
```suggestion
wp_timeout = self.config.get("weight_proof_timeout", 300)
```

But, shouldn't this be done at some startup point rather than in the middle of everything?  I thought we processed config in `.__init__()` or such generally.

mariano54 (approved):

altendky (approved):

wjblanke (approved):
aok

PR #10351, opened 2022-02-21
Limit history to local data.

fchirica:
In case of a subscribed store_id, don't retrieve more than we've locally downloaded. Say the chain has hashes ABCDE, and locally we're at C. In this case, return only "ABC" as we can only tree diff A <-> B and B <-> C until we download E.

### DIFF
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -159,7 +159,13 @@ class DataLayer:
         return res.node_hash
 
     async def get_root_history(self, store_id: bytes32) -> List[SingletonRecord]:
-        records = await self.wallet_rpc.dl_history(store_id)
+        try:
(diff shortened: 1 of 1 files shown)


almogdepaz:
not sure why we want to limit this, i think getting all the on-chain roots is useful
maybe we can add a flag for synced/not synced instead ?

chia/data_layer/data_layer.py
```diff
@@ -159,14 +159,22 @@ async def get_local_root(self, store_id: bytes32) -> Optional[bytes32]:
         return res.node_hash
 
     async def get_root_history(self, store_id: bytes32) -> List[SingletonRecord]:
-        records = await self.wallet_rpc.dl_history(store_id)
+        try:
+            # If we're subscribed to the store_id, don't get more than history than we've downloaded.
```
Nikolaj-K, line 163: Is the second `than` a `that`?

PR #10353, opened 2022-02-21
Fix Atari wallet high CPU usage

Quexington:


chia/data_layer/data_layer_wallet.py
```diff
@@ -209,6 +211,7 @@ async def new_launcher_spend_response(self, response: PuzzleSolutionResponse, ac
             CoinSpend(launcher_coin, response.puzzle, response.solution),
             height=response.height,
         )
+        await self.wallet_state_manager.action_store.action_done(action_id)
```
altendky, line 214: I have no idea what is better, but apparently there is `WalletStateManager.set_action_done()` that passes through.

This feels like a thing where a context manager would be appropriate with explicit non-success behavior.  As is I guess we just build up the action table in the db without any culling of no-longer-needed actions, along with completed actions.  It also seems like the action update should be in a transaction with the rest of the db change associated with it?

Presumably this is either just me needing to learn more about this, or if a change is appropriate then at least something to be addressed in separate PR.

chia/data_layer/data_layer_wallet.py
```diff
@@ -166,6 +166,8 @@ async def track_new_launcher_id(  # This is the entry point for non-owned single
         height: Optional[uint32] = None,
         in_transaction: bool = False,
     ) -> None:
+        if await self.wallet_state_manager.dl_store.get_launcher(launcher_id) is not None:
+            return None
```
altendky, line 170: Maybe a log entry saying we are already tracking this?  Info level I suppose.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1344,14 +1344,15 @@ async def create_new_dl(self, request) -> Dict:
                 )
 
         try:
```
altendky, line 1346: Shouldn't there be some transaction handling with commit and rollback to go along with the lock?  I see that is an existing issue and could be addressed separately.

altendky (approved):

PR #10354, opened 2022-02-21
less Optional[] and better shutdown on WalletNode

altendky:


chia/wallet/wallet_node.py
```diff
@@ -120,7 +121,7 @@ def __init__(
         self.proof_hashes: List = []
         self.state_changed_callback = None
         self.wallet_state_manager = None
-        self.new_state_lock = None
+        self.new_state_lock = asyncio.Lock()
```
mariano54, line 124: This needs to happen in _start not in init, init can happen outside the asyncio running loop

chia/wallet/wallet_node.py
```diff
@@ -140,7 +141,7 @@ def __init__(
         self._process_new_subscriptions_task = None
         self._secondary_peer_sync_task = None
         self.node_peaks = {}
-        self.validation_semaphore = None
+        self.validation_semaphore = asyncio.Semaphore(6)
```
mariano54, line 144: same here

chia/wallet/wallet_node.py
```diff
@@ -643,10 +638,23 @@ async def receive_and_validate(inner_state: CoinState, inner_idx: int):
                 return False
             while num_concurrent_tasks >= target_concurrent_tasks:
                 await asyncio.sleep(1)
+                if self._shut_down:
```
mariano54, line 641: this is great, i was thinking we should do this after efishent reported his issue

chia/wallet/wallet_node.py
```diff
@@ -643,10 +638,23 @@ async def receive_and_validate(inner_state: CoinState, inner_idx: int):
...
+                    break
             all_tasks.append(asyncio.create_task(receive_and_validate(potential_state, idx)))
             num_concurrent_tasks += 1
 
-        await asyncio.gather(*all_tasks)
+        if not self._shut_down:
+            await asyncio.gather(*all_tasks)
+        else:
+            for task in all_tasks:
+                task.cancel()
```
mariano54, line 650: isn't this kind of dangerous? Can we wait for the tasks to end, since most of them will be quick anyways?  I don't want half the DB operations to finish for example

altendky: If there is such a hazard then our db locking/transactions are wrong I would think?  For example, I'm doing this because of a pile of errors that lead me to look at this.  https://github.com/Chia-Network/chia-blockchain/runs/5269700934?check_suite_focus=true

No, my changes so far do not fix the flaky test failure, I don't think, but it seemed a useful step to quiet stuff down a bit.

chia/wallet/wallet_node.py
altendky: Thanks for pointing that out.  Why does it matter though?

mariano54: if you construct the object but you are not in an asyncio loop, you will get an error when you try to use it, something related to being in a different asyncio loop.

altendky: Before I only tested that I could construct it outside an `async def`.  Now I've tried acquiring the lock and that seems ok, including running the code below in Python 3.7 locally.  The last test run looked ok outside of some installation issues, so we'll see what it looks like.  Hopefully they exercise this.  But sure, this is a dual purpose PR...  yuck.  I can extract this if it's going to hold things up.

https://replit.com/@altendky/UtterGiantTexts-1
```python
import asyncio


lock = asyncio.Lock()


async def main():
    async with lock:
        print("in the lock")


asyncio.run(main())
```
```
in the lock
```

mariano54: This caused issues in the past, so i am confused why it's working well here..

altendky: Just a simplification, no need to push it in now.  I'll move the other bits forward without this.

chia/wallet/wallet_node.py
altendky: Extracted to https://github.com/Chia-Network/chia-blockchain/pull/10384 so it can move forward.

PR #10355, opened 2022-02-21
streamable: Enable flake8 and pylint

xdustinface:


chia/util/streamable.py
```diff
@@ -295,7 +295,7 @@ def function_to_parse_one_item(cls: Type[cls.__name__], f_type: Type):  # type:
         raise NotImplementedError(f"Type {f_type} does not have parse")
 
     @classmethod
-    def parse(cls: Type[cls.__name__], f: BinaryIO) -> cls.__name__:  # type: ignore
+    def parse(cls: Type[Streamable], f: BinaryIO) -> Streamable:
```
altendky, line 298: Since we inherit, this is not accurate.  The type of `cls` won't be `Streamable` specifically when accessed on an inherited class.  Rather we need something like `_T_Streamable = TypeVar("_T_Streamable", bound="Streamable")` and then:

```suggestion
    def parse(cls: Type[_T_Streamable], f: BinaryIO) -> _T_Streamable:
```

This says that the method will return an instance of whatever `cls` is.

altendky (changes requested):

chia/util/streamable.py
```diff
@@ -156,8 +156,10 @@ def streamable(cls: Any):
 
     1. A tuple of x items is serialized by appending the serialization of each item.
     2. A List is serialized into a 4 byte size prefix (number of items) and the serialization of each item.
-    3. An Optional is serialized into a 1 byte prefix of 0x00 or 0x01, and if it's one, it's followed by the serialization of the item.
-    4. A Custom item is serialized by calling the .parse method, passing in the stream of bytes into it. An example is a CLVM program.
+    3. An Optional is serialized into a 1 byte prefix of 0x00 or 0x01, and if it's one, it's followed by the
+       serialization of the item.
+    4. A Custom item is serialized by calling the .parse method, passing in the stream of bytes into it. An example is
+       a CLVM program.
```
altendky, line 162: Did a linter force this?  I personally write mine to 88 width for easier reading.  `:]`

altendky (approved):

chia/util/streamable.py
xdustinface: jup, flake8 complained.

PR #10358, opened 2022-02-21
Remove block tools and keychain globals

aqk:
This patch removes a number of global variables from the tests. This builds on an earlier patch from @paninaro

### DIFF
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -11,8 +11,7 @@ from keyring.backends.Windows import WinVaultKeyring as WinKeyring
 from keyring.errors import KeyringError, PasswordDeleteError
 from pathlib import Path
 from sys import exit, platform
-from typing import Any, List, Optional, Tuple, Type, Union
-
+from typing import Any, List, Optional, Tuple, Type, Union, TypeVar
 
 # We want to protect the keyring, even if a user-specified master passphrase isn't provided
 #
@@ -82,6 +81,9 @@ def warn_if_macos_errSecInteractionNotAllowed(error: KeyringError) -> bool:
     return False
 
 
+_T_KeyringWrapper = TypeVar("_T_KeyringWrapper", bound="KeyringWrapper")
+
+
 class KeyringWrapper:
     """
     KeyringWrapper provides an abstraction that the Keychain class can use
@@ -104,24 +106,37 @@ class KeyringWrapper:
     cached_passphrase_is_validated: bool = False
     legacy_keyring = None
 
-    def __init__(self, keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH, force_legacy: bool = False):
+    @classmethod
+    def create(
+        cls: Type[_T_KeyringWrapper], keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH, force_legacy: bool = False
+    ) -> Optional[_T_KeyringWrapper]:
         """
-        Initializes the keyring backend based on the OS. For Linux, we previously
-        used CryptFileKeyring. We now use our own FileKeyring backend and migrate
-        the data from the legacy CryptFileKeyring (on write).
+        Create a new KeyringWrapper instance.
         """
-        self.keys_root_path = keys_root_path
+        # instance: KeyringWrapper = KeyringWrapper()
+        instance = cls()
+        instance.keys_root_path = keys_root_path
         if force_legacy:
             legacy_keyring = get_legacy_keyring_instance()
             if check_legacy_keyring_keys_present(legacy_keyring):
-                self.keyring = legacy_keyring
+                instance.keyring = legacy_keyring
             else:
                 return None
         else:
-            self.refresh_keyrings()
+            instance.refresh_keyrings(keys_root_path)
+        return instance
 
-    def refresh_keyrings(self):
-        self.keyring = None
+    def refresh_keyrings(self, keys_root_path: Optional[Path]) -> None:
+        """
+        Initializes or updates the keyring backend based on the OS. For Linux, we
+        previously used CryptFileKeyring. We now use our own FileKeyring backend
+        and migrate the data from the legacy CryptFileKeyring (on write).
+        """
+
+        if keys_root_path is not None:
+            self.keys_root_path = keys_root_path
+
+        self.cleanup_keyring()
         self.keyring = self._configure_backend()
 
         # Configure the legacy keyring if keyring passphrases are supported to support migration (if necessary)
@@ -187,17 +202,29 @@ class KeyringWrapper:
     @staticmethod
     def get_shared_instance(create_if_necessary=True):
         if not KeyringWrapper.__shared_instance and create_if_necessary:
-            KeyringWrapper.__shared_instance = KeyringWrapper(keys_root_path=KeyringWrapper.__keys_root_path)
+            KeyringWrapper.__shared_instance = KeyringWrapper.create(keys_root_path=KeyringWrapper.__keys_root_path)
 
         return KeyringWrapper.__shared_instance
 
     @staticmethod
-    def cleanup_shared_instance():
-        KeyringWrapper.__shared_instance = None
+    def set_shared_instance(instance: "KeyringWrapper"):
+        KeyringWrapper.__shared_instance = instance
 
     @staticmethod
-    def get_legacy_instance() -> Optional["KeyringWrapper"]:
-        return KeyringWrapper(force_legacy=True)
+    def cleanup_shared_instance():
+        if KeyringWrapper.__shared_instance is not None:
+            KeyringWrapper.__shared_instance.cleanup_keyring()
+            KeyringWrapper.__shared_instance = None
+
+    def cleanup_keyring(self):
+        filekeyring: Optional[FileKeyring] = self.keyring if type(self.keyring) == FileKeyring else None
+        if filekeyring is not None:
+            filekeyring.cleanup_keyring_file_watcher()
+        self.keyring = None
+
+    @staticmethod
+    def get_legacy_instance() -> Optional[_T_KeyringWrapper]:
+        return KeyringWrapper.create(force_legacy=True)
 
     def get_keyring(self):
         """
@@ -423,7 +450,8 @@ class KeyringWrapper:
             if user is not None:
                 passphrase = self.get_passphrase(service, user)
 
-            if passphrase is not None:
+            # if passphrase and user:
+            if passphrase:
                 user_passphrase_pairs.append((user, passphrase))
 
             index += 1
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -219,7 +219,7 @@ class KeyringWrapper:
         self.keyring = None
 
     @staticmethod
-    def get_legacy_instance():
+    def get_legacy_instance() -> Optional["KeyringWrapper"]:
         return KeyringWrapper.create(force_legacy=True)
 
     def get_keyring(self):
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -446,8 +446,7 @@ class KeyringWrapper:
             if user is not None:
                 passphrase = self.get_passphrase(service, user)
 
-            # if passphrase and user:
-            if passphrase:
+            if passphrase is not None:
                 user_passphrase_pairs.append((user, passphrase))
 
             index += 1
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -165,7 +165,7 @@ class KeyringWrapper:
 
         return keyring
 
-    def _configure_legacy_backend(self) -> LegacyKeyring:
+    def _configure_legacy_backend(self) -> Optional[LegacyKeyring]:
         # If keyring.yaml isn't found or is empty, check if we're using
         # CryptFileKeyring, Mac Keychain, or Windows Credential Manager
         filekeyring = self.keyring if type(self.keyring) == FileKeyring else None
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -11,7 +11,8 @@ from keyring.backends.Windows import WinVaultKeyring as WinKeyring
 from keyring.errors import KeyringError, PasswordDeleteError
 from pathlib import Path
 from sys import exit, platform
-from typing import Any, List, Optional, Tuple, Type, Union, TypeVar
+from typing import Any, List, Optional, Tuple, Type, Union
+
 
 # We want to protect the keyring, even if a user-specified master passphrase isn't provided
 #
@@ -26,8 +27,6 @@ MASTER_PASSPHRASE_USER_NAME = "Chia Passphrase"
 
 LegacyKeyring = Union[MacKeyring, WinKeyring, CryptFileKeyring]
 OSPassphraseStore = Union[MacKeyring, WinKeyring]
-
-_T_KeyringWrapper = TypeVar("_T_KeyringWrapper", bound="KeyringWrapper")
 
 
 def get_legacy_keyring_instance() -> Optional[LegacyKeyring]:
@@ -105,37 +104,24 @@ class KeyringWrapper:
     cached_passphrase_is_validated: bool = False
     legacy_keyring = None
 
-    @classmethod
-    def create(
-        cls: Type[_T_KeyringWrapper], keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH, force_legacy: bool = False
-    ) -> Optional[_T_KeyringWrapper]:
+    def __init__(self, keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH, force_legacy: bool = False):
         """
-        Create a new KeyringWrapper instance.
+        Initializes the keyring backend based on the OS. For Linux, we previously
+        used CryptFileKeyring. We now use our own FileKeyring backend and migrate
+        the data from the legacy CryptFileKeyring (on write).
         """
-        # instance: KeyringWrapper = KeyringWrapper()
-        instance = cls()
-        instance.keys_root_path = keys_root_path
+        self.keys_root_path = keys_root_path
         if force_legacy:
             legacy_keyring = get_legacy_keyring_instance()
             if check_legacy_keyring_keys_present(legacy_keyring):
-                instance.keyring = legacy_keyring
+                self.keyring = legacy_keyring
             else:
                 return None
         else:
-            instance.refresh_keyrings(keys_root_path)
-        return instance
+            self.refresh_keyrings()
 
-    def refresh_keyrings(self, keys_root_path: Optional[Path]) -> None:
-        """
-        Initializes or updates the keyring backend based on the OS. For Linux, we
-        previously used CryptFileKeyring. We now use our own FileKeyring backend
-        and migrate the data from the legacy CryptFileKeyring (on write).
-        """
-
-        if keys_root_path is not None:
-            self.keys_root_path = keys_root_path
-
-        self.cleanup_keyring()
+    def refresh_keyrings(self):
+        self.keyring = None
         self.keyring = self._configure_backend()
 
         # Configure the legacy keyring if keyring passphrases are supported to support migration (if necessary)
@@ -164,7 +150,7 @@ class KeyringWrapper:
 
         return keyring
 
-    def _configure_legacy_backend(self) -> Optional[LegacyKeyring]:
+    def _configure_legacy_backend(self) -> LegacyKeyring:
         # If keyring.yaml isn't found or is empty, check if we're using
         # CryptFileKeyring, Mac Keychain, or Windows Credential Manager
         filekeyring = self.keyring if type(self.keyring) == FileKeyring else None
@@ -201,25 +187,17 @@ class KeyringWrapper:
     @staticmethod
     def get_shared_instance(create_if_necessary=True):
         if not KeyringWrapper.__shared_instance and create_if_necessary:
-            KeyringWrapper.__shared_instance = KeyringWrapper.create(keys_root_path=KeyringWrapper.__keys_root_path)
+            KeyringWrapper.__shared_instance = KeyringWrapper(keys_root_path=KeyringWrapper.__keys_root_path)
 
         return KeyringWrapper.__shared_instance
 
     @staticmethod
     def cleanup_shared_instance():
-        if KeyringWrapper.__shared_instance is not None:
-            KeyringWrapper.__shared_instance.cleanup_keyring()
-            KeyringWrapper.__shared_instance = None
-
-    def cleanup_keyring(self):
-        filekeyring: Optional[FileKeyring] = self.keyring if type(self.keyring) == FileKeyring else None
-        if filekeyring is not None:
-            filekeyring.cleanup_keyring_file_watcher()
-        self.keyring = None
+        KeyringWrapper.__shared_instance = None
 
     @staticmethod
     def get_legacy_instance() -> Optional["KeyringWrapper"]:
-        return KeyringWrapper.create(force_legacy=True)
+        return KeyringWrapper(force_legacy=True)
 
     def get_keyring(self):
         """
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -27,6 +27,8 @@ MASTER_PASSPHRASE_USER_NAME = "Chia Passphrase"
 LegacyKeyring = Union[MacKeyring, WinKeyring, CryptFileKeyring]
 OSPassphraseStore = Union[MacKeyring, WinKeyring]
 
+_T_KeyringWrapper = TypeVar("_T_KeyringWrapper", bound="KeyringWrapper")
+
 
 def get_legacy_keyring_instance() -> Optional[LegacyKeyring]:
     if platform == "darwin":
@@ -79,9 +81,6 @@ def warn_if_macos_errSecInteractionNotAllowed(error: KeyringError) -> bool:
         )
         return True
     return False
-
-
-_T_KeyringWrapper = TypeVar("_T_KeyringWrapper", bound="KeyringWrapper")
 
 
 class KeyringWrapper:
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -207,10 +207,6 @@ class KeyringWrapper:
         return KeyringWrapper.__shared_instance
 
     @staticmethod
-    def set_shared_instance(instance: "KeyringWrapper"):
-        KeyringWrapper.__shared_instance = instance
-
-    @staticmethod
     def cleanup_shared_instance():
         if KeyringWrapper.__shared_instance is not None:
             KeyringWrapper.__shared_instance.cleanup_keyring()
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -223,7 +223,7 @@ class KeyringWrapper:
         self.keyring = None
 
     @staticmethod
-    def get_legacy_instance() -> Optional[_T_KeyringWrapper]:
+    def get_legacy_instance():
         return KeyringWrapper.create(force_legacy=True)
 
     def get_keyring(self):
--- a/setup.py
+++ b/setup.py
@@ -44,8 +44,7 @@ dev_dependencies = [
     "build",
     "pre-commit",
     "pytest",
-    # >=0.17.0 for the fixture decorator
-    "pytest-asyncio >=0.17.0",
+    "pytest-asyncio>=0.18.1",  # require attribute 'fixture'
     "pytest-monitor; sys_platform == 'linux'",
     "pytest-xdist",
     "twine",
--- a/tests/block_tools.py
+++ b/tests/block_tools.py
@@ -2033,7 +2033,6 @@ async def create_block_tools_async(
     return bt
 
 
-# xxx redirect to the fixture
 def create_block_tools(
     constants: ConsensusConstants = test_constants,
     root_path: Optional[Path] = None,
--- a/tests/block_tools.py
+++ b/tests/block_tools.py
@@ -133,6 +133,7 @@ class BlockTools:
         root_path: Optional[Path] = None,
         const_dict=None,
         keychain: Optional[Keychain] = None,
+        daemon_port=None,
     ):
         self._tempdir = None
         if root_path is None:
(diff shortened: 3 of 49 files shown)


tests/core/daemon/test_daemon.py
```diff
@@ -45,10 +50,10 @@ async def get_b_tools_1(self, get_temp_keyring):
         return await create_block_tools_async(constants=test_constants_modified, keychain=get_temp_keyring)
 
     @pytest_asyncio.fixture(scope="function")
-    async def get_b_tools(self, get_temp_keyring):
+    async def get_b_tools(self, get_temp_keyring):  # xxx
         local_b_tools = await create_block_tools_async(constants=test_constants_modified, keychain=get_temp_keyring)
         new_config = local_b_tools._config
-        new_config["daemon_port"] = 55401
+        new_config["daemon_port"] = 55401  # xxx
```
aqk, line 56: Please take a close look at this line:

`new_config["daemon_port"] = 55401`

tests/core/daemon/test_daemon.py
```diff
@@ -45,10 +50,10 @@ async def get_b_tools_1(self, get_temp_keyring):
         return await create_block_tools_async(constants=test_constants_modified, keychain=get_temp_keyring)
 
     @pytest_asyncio.fixture(scope="function")
-    async def get_b_tools(self, get_temp_keyring):
+    async def get_b_tools(self, get_temp_keyring):  # xxx
```
aqk, line 53: Please take a close look at `get_b_tools`

aqk:
@paninaro could you take a look at `tests/core/cmds/test_keys.py::TestKeysCommands::test_generate_with_new_config` ?

````FAILED                                                                           [1/1]

_________ TestKeysCommands.test_generate_with_new_config ____
tests/core/cmds/test_keys.py:133: in test_generate_with_new_config
    assert len(keychain.get_all_private_keys()) == 1
E   assert 0 == 1
```

tests/core/test_daemon_rpc.py
```diff
@@ -1,20 +1,21 @@
+# flake8: noqa: F821
+# pylint: disable=E0602
```
aqk, line 2: Python fixture auto-imports are not recognized by flake / pylint / black

setup.py
```diff
@@ -43,7 +43,7 @@
 dev_dependencies = [
     "pre-commit",
     "pytest",
-    "pytest-asyncio",
+    "pytest-asyncio>=0.18.1",
```
aqk, line 46: Older `pytest-asyncio` is missing features we use.

tests/block_tools.py
```diff
@@ -2007,25 +2007,37 @@ def create_test_unfinished_block(
...
+# xxx do the callers need different instances?
 async def create_block_tools_async(
     constants: ConsensusConstants = test_constants,
     root_path: Optional[Path] = None,
     const_dict=None,
     keychain: Optional[Keychain] = None,
 ) -> BlockTools:
+    global create_block_tools_async_count
+    create_block_tools_async_count += 1
+    print(f"  create_block_tools_async called {create_block_tools_async_count} times")
```
aqk, line 2023: We can remove these two print stmts before this patch lands

tests/blockchain/test_blockchain_transactions.py
```diff
@@ -14,7 +14,7 @@
 from chia.util.ints import uint64
 from tests.blockchain.blockchain_test_utils import _validate_and_add_block
 from tests.wallet_tools import WalletTool
-from tests.setup_nodes import bt, setup_two_nodes, test_constants
```
aqk, line 17: There are a number of warts & limitations in the Python fixtures implementation -
Removing these explicit references and putting them in `conftest.py` is the accepted way of using fixtures.

aqk:
Important files:

```
tests/setup_nodes.py
tests/block_tools.py
tests/conftest.py
chia/util/keyring_wrapper.py
```

tests/core/daemon/test_daemon.py
arvidn: In my patch (which I'm hoping to land before this) I remove this assignment and make every `create_blockchain_tools` pick an available port and use that.

chia/util/keyring_wrapper.py
```diff
@@ -104,24 +106,37 @@ class KeyringWrapper:
     cached_passphrase_is_validated: bool = False
     legacy_keyring = None
 
-    def __init__(self, keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH, force_legacy: bool = False):
+    @classmethod
+    def create(
+        cls: Type[_T_KeyringWrapper], keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH, force_legacy: bool = False
+    ) -> Optional[_T_KeyringWrapper]:
```
arvidn, line 112: I thought the main reason to have a `create()` function (instead of a constructor) is that it can be `async`. But this isn't. @altendky has also pointed out that with a `create()` function you can mark the class as `@dataclass` (which has some benefits I haven't quite understood yet). But this class isn't a `dataclass` either.

chia/util/keyring_wrapper.py
```diff
@@ -104,24 +106,37 @@ class KeyringWrapper:
...
         """
-        Initializes the keyring backend based on the OS. For Linux, we previously
-        used CryptFileKeyring. We now use our own FileKeyring backend and migrate
-        the data from the legacy CryptFileKeyring (on write).
+        Create a new KeyringWrapper instance.
         """
-        self.keys_root_path = keys_root_path
+        # instance: KeyringWrapper = KeyringWrapper()
+        instance = cls()
```
arvidn, line 117: is `cls` sometimes not `KeyringWrapper`?
If it always is, I think it would be clearer to just construct that type explicitly.

chia/util/keyring_wrapper.py
```diff
@@ -187,17 +202,29 @@ def set_keys_root_path(keys_root_path: Path):
     @staticmethod
     def get_shared_instance(create_if_necessary=True):
         if not KeyringWrapper.__shared_instance and create_if_necessary:
-            KeyringWrapper.__shared_instance = KeyringWrapper(keys_root_path=KeyringWrapper.__keys_root_path)
+            KeyringWrapper.__shared_instance = KeyringWrapper.create(keys_root_path=KeyringWrapper.__keys_root_path)
 
         return KeyringWrapper.__shared_instance
 
+    @staticmethod
+    def set_shared_instance(instance: "KeyringWrapper"):
```
arvidn, line 210: I don't see any calls to this function. It seems like a risky hole in the abstraction too. Can it be removed?

tests/block_tools.py
arvidn: I take it you intend to remove those global counters too, right?

tests/block_tools.py
```diff
@@ -2007,25 +2007,37 @@ def create_test_unfinished_block(
     )
 
 
+create_block_tools_async_count = 0
+create_block_tools_count = 0
+
+
+# xxx do the callers need different instances?
```
arvidn, line 2014: yes, they do. They all need different root directories containing different `config.yaml` files. The daemon's listen port is configured in there, and the only way a test can control which listen port it uses is to write it to the config file.

tests/blockchain/test_blockchain_transactions.py
arvidn: currently `bt` is a global variable that's being imported here. in the new world, `bt` is a fixture, so I don't think you *can* import it (you certainly don't want to)

tests/conftest.py
```diff
@@ -1,18 +1,42 @@
...
+    keyring.cleanup()
+
+
+temp_keyring = TempKeyring(populate=True)
+keychain = temp_keyring.get_keychain()
+atexit.register(cleanup_keyring, temp_keyring)  # Attempt to clean up the temp keychain
+
+
+@pytest.fixture(scope="session", name="bt")
+def bt() -> BlockTools:
```
arvidn, line 22: I think you will need to make this a generator fixture, otherwise you don't get an opportunity to clean it up (which is critical).

Also, one of the problems with the current global variable `bt` is that it's mutable and tests *do* mutate it. Giving this "session" scope, preserves some of those problems (it's still mutable, right?). Would it be feasible to at least make it "module" scoped?

tests/conftest.py
```diff
@@ -1,18 +1,42 @@
...
-#       creation, including the filesystem modification, away from the import but
-#       that seems like a separate step and until then locating the imports in the
-#       fixtures avoids the issue.
+def cleanup_keyring(keyring: TempKeyring):
+    keyring.cleanup()
+
+
+temp_keyring = TempKeyring(populate=True)
+keychain = temp_keyring.get_keychain()
+atexit.register(cleanup_keyring, temp_keyring)  # Attempt to clean up the temp keychain
```
arvidn, line 18: this should also be a fixture, but I consider making `bt` a fixture the top priority. The global keyring can be cleaned up later

tests/conftest.py
```diff
@@ -1,18 +1,42 @@
...
+    # yield _shared_block_tools
+    # _shared_block_tools.cleanup()
+
+
+@pytest.fixture(scope="session")
+def self_hostname(bt):
+    return bt.config["self_hostname"]
+
+
+def setup_shared_block_tools_and_keyring(
```
arvidn, line 34: I don't see this used anywhere. It also looks suspicious, now that `bt` is a fixture, it doesn't look right to also create one like this (that also may not be cleaned up correctly on shutdown).

Is it possible to *set* the keychain on an existing `bt` object? If so, tests that need a specific keyring could still use the `bt` fixture.

tests/connection_utils.py
```diff
@@ -15,21 +15,20 @@
...
 
 log = logging.getLogger(__name__)
 
 
-async def disconnect_all_and_reconnect(server: ChiaServer, reconnect_to: ChiaServer) -> bool:
+async def disconnect_all_and_reconnect(server: ChiaServer, reconnect_to: ChiaServer, self_hostname: str) -> bool:
     cons = list(server.all_connections.values())[:]
     for con in cons:
         await con.close()
     return await server.start_client(PeerInfo(self_hostname, uint16(reconnect_to._port)), None)
```
arvidn, line 27: does the hostname we claim to have matter for any test?
It might not, in which case we might be able to just pick a random string here and be done with it.

tests/core/daemon/test_daemon.py
```diff
@@ -29,9 +29,14 @@ async def get_daemon(self, get_b_tools):
     # because of a hack in shutting down the full node, which means you cannot run
     # more than one simulations per process.
     @pytest_asyncio.fixture(scope="function")
-    async def simulation(self, get_b_tools, get_b_tools_1):
+    async def simulation(self, bt, get_b_tools, get_b_tools_1):
```
arvidn, line 32: I'm pretty sure `get_b_tools` is currently referring to the global `bt` object (and you could replace that with the new fixture. However, `get_b_tools_1` should be a copy of `bt`. i.e. the same kind of fixture but a separate instance.

tests/connection_utils.py
arvidn: I'm experimenting with simplifying this here: https://github.com/Chia-Network/chia-blockchain/pull/10371

tests/core/test_daemon_rpc.py
arvidn: you're removing these, right?

tests/util/blockchain.py
```diff
@@ -35,9 +35,11 @@ async def create_blockchain(constants: ConsensusConstants, db_version: int):
     return bc1, connection, db_path
 
 
+# @pytest.fixture(scope="module")
```
arvidn, line 38: these comments should be removed before we land (and maybe experimented with further in a separate PR)

tests/util/keyring.py
```diff
@@ -165,7 +165,8 @@ def _patch_and_create_keychain(
             add_dummy_key_to_cryptfilekeyring(crypt_file_keyring)
 
         keychain = Keychain(user=user, service=service)
-        keychain.keyring_wrapper = KeyringWrapper(keys_root_path=Path(temp_dir))
+        keychain.keyring_wrapper = KeyringWrapper.create(keys_root_path=Path(temp_dir))
+        # keychain.keyring_wrapper = KeyringWrapper(keys_root_path=Path(temp_dir))
```
arvidn, line 169: and this comment

arvidn (changes requested):
I can't wait for this to land! However, you may want to rebase on top of https://github.com/Chia-Network/chia-blockchain/pull/10307

There's an opportunity for simplifying this patch by also applying https://github.com/Chia-Network/chia-blockchain/pull/10371 first too

tests/conftest.py
aqk: That is a really good idea. I will add a comment.
However, I think the keychain requires that there is only one instance at the moment.

tests/conftest.py
aqk: Thanks, I knew I'd need that once I added the keychain fixture. Also, I looked, but I'm not aware of any BlockTools cleanup functions. Let me look at that again.

chia/util/keyring_wrapper.py
aqk: You know, I had assumed that the reason for the create method was so that we could avoid a global, and instead construct at a later time of our choosing, but I don't know. @altendky @paninaro ?

chia/util/keyring_wrapper.py
aqk: Yeah, I am on the fence - naming the expected type seems better, because it is more constrained.
I'm not sure if the `cls()` helper does some extra Python class magic that I'm not aware of.

chia/util/keyring_wrapper.py
aqk: Oh, if this is added by my patch, this is probably leftover code - I'll remove it.

tests/core/daemon/test_daemon.py
aqk: Yes, that was my impression - `get_b_tools_1` is safe as-is for now, correct?

chia/util/keyring_wrapper.py
paninaro: I honestly don't remember why I had added the create method here... Looking at where it's used and what's checked into main, I don't see any functional difference between `create()` and `__init__()`. It may have been that I was trying to convey that multiple KeyringWrappers shouldn't be created.

tests/block_tools.py
aqk: I'll remove these when they are not being run multiple times

chia/util/keyring_wrapper.py
aqk: removed

tests/connection_utils.py
aqk: Ya, I initially had it that way with the global, too. Thanks!

tests/block_tools.py
```diff
@@ -2031,6 +2052,9 @@ def create_block_tools(
     const_dict=None,
     keychain: Optional[Keychain] = None,
 ) -> BlockTools:
+    global create_block_tools_count
```
mariano54, line 2055: why do we need these globals?

mariano54 (commented):
Seems reasonable to me. I have not reviewed the keychain changes.

tests/block_tools.py
aqk: We don't! I just want to see them go to 1 before we remove these two functions entirely.

chia/util/keyring_wrapper.py
aqk: @altendky do you have an opinion? In this case, I would tend to go with the more restrictive type, a Arvid is suggesting, but the usual thing in Python seems to be `cls()` - perhaps to avoid a class rename breaking code?

tests/block_tools.py
aqk: They allow us to print how many times the BlockTools create functions have been called.
They aren't going to stay.

paninaro (dismissed):
Looks good to me

tests/core/util/test_streamable.py
```diff
@@ -371,5 +365,11 @@ def test_parse_str(self):
             parse_str(io.BytesIO(b"\x00\x00\x02\x01" + b"a" * 512))
 
 
+def test_json(bt):
+    block = bt.create_genesis_block(test_constants, bytes32([0] * 32), uint64(0))
+    dict_block = block.to_json_dict()
+    assert FullBlock.from_json_dict(dict_block) == block
```
altendky, line 371: Did this need to move?

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -137,9 +137,9 @@ def stop_node_cb():
         await time_out_assert(5, wallet.get_confirmed_balance, initial_funds)
         await time_out_assert(5, wallet.get_unconfirmed_balance, initial_funds)
 
-        client = await WalletRpcClient.create(self_hostname, test_rpc_port, bt.root_path, config)
-        client_2 = await WalletRpcClient.create(self_hostname, test_rpc_port_2, bt.root_path, config)
-        client_node = await FullNodeRpcClient.create(self_hostname, test_rpc_port_node, bt.root_path, config)
+        client = await WalletRpcClient.create(hostname, test_rpc_port, bt.root_path, config)
+        client_2 = await WalletRpcClient.create(hostname, test_rpc_port_2, bt.root_path, config)
+        client_node = await FullNodeRpcClient.create(hostname, test_rpc_port_node, bt.root_path, config)
```
altendky, line 142: Do we want to be switching to pulling this out of the config?  If so, should we be using the fixture up above?

altendky (commented):
Thanks for all the work!  Looks like we're close...

chia/util/keyring_wrapper.py
```diff
@@ -187,17 +201,25 @@ def set_keys_root_path(keys_root_path: Path):
...
-        KeyringWrapper.__shared_instance = None
+        if KeyringWrapper.__shared_instance is not None:
+            KeyringWrapper.__shared_instance.cleanup_keyring()
+            KeyringWrapper.__shared_instance = None
+
+    def cleanup_keyring(self):
+        filekeyring: Optional[FileKeyring] = self.keyring if type(self.keyring) == FileKeyring else None
+        if filekeyring is not None:
+            filekeyring.cleanup_keyring_file_watcher()
+        self.keyring = None
```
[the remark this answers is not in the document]
paninaro: I would suggest that we resist the urge to make additional structural changes at this point. Let's let this PR land, get coverage to 100% and then revisit :). If it ain't broke and all...

tests/util/keyring.py
```diff
@@ -165,7 +165,7 @@ def _patch_and_create_keychain(
             add_dummy_key_to_cryptfilekeyring(crypt_file_keyring)
 
         keychain = Keychain(user=user, service=service)
-        keychain.keyring_wrapper = KeyringWrapper(keys_root_path=Path(temp_dir))
+        keychain.keyring_wrapper = KeyringWrapper.create(keys_root_path=Path(temp_dir))
```
arvidn, line 168:
```suggestion
        keychain.keyring_wrapper = KeyringWrapper.create(keys_root_path=Path(temp_dir))
        assert keyring_wrapper is not None
        keychain.keyring_wrapper = keyring_wrapper
```

arvidn: this is to fix the mypy error (untested):
```
tests/util/keyring.py:168: error: Incompatible types in assignment (expression has type "Optional[KeyringWrapper]", variable has type "KeyringWrapper")  [assignment]
```

My understanding is that this is how you unwrap optionals in python

chia/util/keyring_wrapper.py
arvidn: It would seem reasonable to not return an `Optional[]`, but just a `_T_KeyringWrapper` directly. If there's an error, throw an exception. This would fix the mypy error

tests/util/keyring.py
altendky: `.create()` is only optional by mistake I think.  https://github.com/Chia-Network/chia-blockchain/pull/10358#discussion_r820005241  If you are sprinkling `assert`s around for `None`s, maybe you didn't want the thing to be optional.  Also see the keybase discussion about `assert`s.  `:]`

arvidn: I agree that `create()` shouldn't return an `Optional`, but assuming the existing code in this patch is correct, this would be the correct use of assert. The existing code clearly expects it to never return `None` in this case. If it did, it would be a programmer mistake, not an expected/recoverable failure.

aqk: Thanks guys.

tests/core/util/test_streamable.py
aqk: Yes! When executed as a part of the class, it broke because of interactions stemming from `TestStreamable` inheriting from `unittest.TestCase`

tests/wallet/rpc/test_wallet_rpc.py
aqk: `self_hostname` and `bt.config["hostname"]` are certain to be the same if a test imports both `bt` and `self_hostname`

arvidn:
@aqk the submodule update is a mistake I think

tests/core/util/test_streamable.py
altendky: Oh, right...  it grew a fixture dependency.  Maybe we should convert this from unittest to pytest fully as a pre-PR.

arvidn (changes requested):
I think there are a still a lot of remnants left from the first version of this patch that can (and should) be removed.

tests/wallet/rpc/test_wallet_rpc.py
altendky: Which is the preferred method?  It would seem like the fixture would be since that would maintain consistency.  Also, it would leave these lines as they were before.  The other lines should be changed to match, but since this could all not be changed at all in this PR, I'm fine with that route.

tests/core/util/test_streamable.py
altendky: I'm happy to write it.

chia/util/keyring_wrapper.py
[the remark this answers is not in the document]
altendky: Looks like all changes in this file were removed.

tests/core/test_daemon_rpc.py
```diff
@@ -3,15 +3,14 @@
...
 from chia import __version__
 from tests.util.socket import find_available_listen_port
 from chia.util.config import save_config
 
 
 class TestDaemonRpc:
     @pytest_asyncio.fixture(scope="function")
-    async def get_daemon(self):
+    async def get_daemon(self, bt):
         bt._config["daemon_port"] = find_available_listen_port()
```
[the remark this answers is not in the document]
aqk: We may not need it in the sense of "is this necessary". But fixing this would be a further improvement, and out of the (massive) scope of this patch, because `setup_daemon(btools)` in `tests/setup_nodes.py` already too a BlockTools as an argument. So, we either have to understand and change `setup_daemon`, or we have toadd a parameter to `get_daemon` to compensate for the global that has been removed.

tests/core/ssl/test_ssl.py
```diff
@@ -23,7 +21,7 @@
 from tests.util.socket import find_available_listen_port
 
 
-async def establish_connection(server: ChiaServer, ssl_context) -> bool:
+async def establish_connection(server: ChiaServer, self_hostname: str, ssl_context) -> bool:
```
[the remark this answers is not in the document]
aqk: I don't see how we can get away without passing in `self_hostname` to `establish_connection` - it is used twice in the body. `self_hostname` just used to be a global that silently brought state into every test method in a module that imported it.

tests/setup_nodes.py
```diff
@@ -77,14 +88,15 @@ async def setup_daemon(btools):
 async def setup_full_node(
     consensus_constants: ConsensusConstants,
     db_name,
+    self_hostname: str,
     port,
     rpc_port,
-    local_bt,
+    local_bt: BlockTools,
+    connect_to_daemon_port: uint16 = None,
```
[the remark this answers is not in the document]
aqk: I agree that I am setting daemon_port too many times in this patch. Definitely with you there.

However, it gives me way more pain that state is being passed downstream through the config file, so you can get into situations like:

```
def fun_a():
  bt.config["obscure_attribute"] = whatever
  fun_b(bt)

def fun_b(bt):  # e.g. setup_xyz
  fun_c(bt)

def fun_c(bt):
  use_obscure_attribute(bt)
```

So, making a change in fun_a can break the call at fun_c

I suppose you could argue that now we are at least tracking the passage of BlockTools through formal parameters, so I could be convinced to back out the redundant but it still bugs me that state that we NEED is not being pass in the formal parameters.

This is one of the things that makes Blocktools so opaque, hard to understand, and debug.

tests/setup_nodes.py
```diff
@@ -98,6 +110,12 @@ async def setup_full_node(
                 connection.commit()
 
     config = local_bt.config["full_node"]
+
+    connect_to_daemon = False
+    if connect_to_daemon_port is not None:
+        config["daemon_port"] = connect_to_daemon_port
```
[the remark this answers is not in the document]
aqk: Right. This is a broken way of doing things... I didn't introduce doing it this way, but maybe it can be taken out now that your patch is in.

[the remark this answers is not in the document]
aqk: I'll test and see if I can take this out!

tests/block_tools.py
```diff
@@ -133,6 +133,7 @@ def __init__(
         root_path: Optional[Path] = None,
         const_dict=None,
         keychain: Optional[Keychain] = None,
+        daemon_port=None,
```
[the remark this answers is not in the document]
aqk: Yes, now that this is debugged, that can go.

tests/core/daemon/test_daemon.py
```diff
@@ -43,7 +32,10 @@ async def get_b_tools_1(self, get_temp_keyring):
 
     @pytest_asyncio.fixture(scope="function")
     async def get_b_tools(self, get_temp_keyring):
-        local_b_tools = await create_block_tools_async(constants=test_constants_modified, keychain=get_temp_keyring)
+        daemon_port = find_available_listen_port("daemon")
```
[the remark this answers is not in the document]
aqk: Yes. This was before your `find_available_listen_port` patch.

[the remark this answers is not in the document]
aqk: Code leftovers removed

setup.py
```diff
@@ -44,7 +44,7 @@
     "build",
     "pre-commit",
     "pytest",
-    "pytest-asyncio",
+    "pytest-asyncio>=0.18.1",  # require attribute 'fixture'
```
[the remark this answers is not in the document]
aqk: #10560 approved

tests/core/ssl/test_ssl.py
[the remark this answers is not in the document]
arvidn: they way we do in `main` right now. Import it.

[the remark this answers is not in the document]
arvidn: I'm suggesting you revert this part of your patch. to import `self_hostname` in all places it's used, rather than importing it in *some* places and then pass it in as a new parameter

tests/setup_nodes.py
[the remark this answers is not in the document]
arvidn: github shows this code as new in your patch to me.

tests/setup_nodes.py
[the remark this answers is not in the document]
arvidn: in my mind, the daemon port (and most configs actually) should be treated as immutable. If we do that (and are suspicious of any place that alters the config, I think we're good.

With that view, configuring the daemon port in the configs once when the `BlockTools` object is created seems the most reasonable.

tests/conftest.py
```diff
@@ -1,22 +1,44 @@
...
+
+@pytest.fixture(scope="session", name="bt")
+def bt(get_keychain) -> BlockTools:
+    # Note that this causes a lot of CPU and disk traffic - disk, DB, ports, process creation ...
+    _shared_block_tools = create_block_tools(constants=test_constants, keychain=get_keychain)
+    return _shared_block_tools
+
+
+@pytest.fixture(scope="session")
+def self_hostname(bt):
```
arvidn, line 31: oh, I see. You made this a fixture. I think that's a mistake. There's no upside and it adds a lot of extra code.

arvidn: it also means that *any* test that starts a listen port and connects to it is *forced* to create the `BlockTools` fixture. That doesn't make any sense at all.

If you really think it should be a fixture, at least don't make it depend on `bt`. It's just a constant string literal.

tests/wallet/rpc/test_wallet_rpc.py
arvidn: I think it's a mistake to make `self_hostname` a fixture. it restricts access to it to only test functions and other fixtures. Currently we use it in other places too.

It should especially not have to depend on constructing a whole `BlockTools` object.

arvidn: > Which is the preferred method? It would seem like the fixture would be since that would maintain consistency.

in `main` (that this PR is targeting), `self_hostname` is a global string literal in the tests. I don't think there's a reason for this patch to change that.

aqk:
Land after https://github.com/Chia-Network/chia-blockchain/pull/10577

tests/conftest.py
aqk: Well, it's trues that none of the current tests rely on the value of `self_hostname`
It's also true that leaving it the way it is now invades countless test functions with a global.

tests/setup_nodes.py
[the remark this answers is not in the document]
aqk: Removed.

tests/core/test_daemon_rpc.py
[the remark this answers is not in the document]
aqk: Take a look at the change history of `setup_daemon` to see what I mean.

Oh, wait. When you say "we don't need to do this anymore", do you mean:
`    bt._config["daemon_port"] = find_available_listen_port()` ?

I thought you meant adding `bt` as an argument.
Sure, I'll take a look at that config assignment - I don't think we need it.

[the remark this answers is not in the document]
aqk: Okay, take a look at that now.

aqk:
@arvidn Thanks, submodule updated

aqk:
This is now rebased on top of https://github.com/Chia-Network/chia-blockchain/pull/10577

tests/conftest.py
arvidn: a global *constant*, not variable. it's like naming pi, it won't change, it's a constant.

tests/setup_nodes.py
```diff
@@ -33,10 +35,18 @@ def cleanup_keyring(keyring: TempKeyring):
...
+
+
+@pytest.fixture(scope="session")
+def bt() -> BlockTools:
+    _shared_block_tools = create_block_tools(constants=test_constants, keychain=keychain)
+    return _shared_block_tools
+
 
 # if you have a system that has an unusual hostname for localhost and you want
 # to run the tests, change this constant
```
arvidn, line 52: here's the comment describing what `self_hostname` is.

arvidn (approved):

paninaro (approved):

wjblanke (approved):
aok

PR #10359, opened 2022-02-21
chunk coin_store request into smaller sql queries

arvidn:
to not exceed the limit of 999 on old versions of sqlite.

I also took the opportunity to factor out the `chunks()` function from weight proof into its own file and added a simple unit test.

There's indentation change in the `CoinStore` functions, you may want to review those ignoring white space.

### DIFF
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -7,10 +7,13 @@ from chia.types.coin_record import CoinRecord
 from chia.util.db_wrapper import DBWrapper
 from chia.util.ints import uint32, uint64
 from chia.util.lru_cache import LRUCache
+from chia.util.chunks import chunks
 import time
 import logging
 
 log = logging.getLogger(__name__)
+
+MAX_SQLITE_PARAMETERS = 900
 
 
 class CoinStore:
@@ -311,24 +314,25 @@ class CoinStore:
             return []
 
         coins = set()
-        puzzle_hashes_db: Tuple[Any, ...]
-        if self.db_wrapper.db_version == 2:
-            puzzle_hashes_db = tuple(puzzle_hashes)
-        else:
-            puzzle_hashes_db = tuple([ph.hex() for ph in puzzle_hashes])
-        async with self.coin_record_db.execute(
-            f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
-            f"coin_parent, amount, timestamp FROM coin_record INDEXED BY coin_puzzle_hash "
-            f'WHERE puzzle_hash in ({"?," * (len(puzzle_hashes) - 1)}?) '
-            f"AND (confirmed_index>=? OR spent_index>=?)"
-            f"{'' if include_spent_coins else 'AND spent_index=0'}",
-            puzzle_hashes_db + (min_height, min_height),
-        ) as cursor:
+        for puzzles in chunks(puzzle_hashes, MAX_SQLITE_PARAMETERS):
+            puzzle_hashes_db: Tuple[Any, ...]
+            if self.db_wrapper.db_version == 2:
+                puzzle_hashes_db = tuple(puzzles)
+            else:
+                puzzle_hashes_db = tuple([ph.hex() for ph in puzzles])
+            async with self.coin_record_db.execute(
+                f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
+                f"coin_parent, amount, timestamp FROM coin_record INDEXED BY coin_puzzle_hash "
+                f'WHERE puzzle_hash in ({"?," * (len(puzzles) - 1)}?) '
+                f"AND (confirmed_index>=? OR spent_index>=?)"
+                f"{'' if include_spent_coins else 'AND spent_index=0'}",
+                puzzle_hashes_db + (min_height, min_height),
+            ) as cursor:
 
-            for row in await cursor.fetchall():
-                coins.add(self.row_to_coin_state(row))
+                async for row in cursor:
+                    coins.add(self.row_to_coin_state(row))
 
-            return list(coins)
+        return list(coins)
 
     async def get_coin_records_by_parent_ids(
         self,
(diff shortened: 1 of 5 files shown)


chia/full_node/coin_store.py
```diff
@@ -311,24 +314,25 @@ async def get_coin_states_by_puzzle_hashes(
...
-            f"AND (confirmed_index>=? OR spent_index>=?)"
-            f"{'' if include_spent_coins else 'AND spent_index=0'}",
-            puzzle_hashes_db + (min_height, min_height),
-        ) as cursor:
+        for puzzles in chunks(puzzle_hashes, MAX_SQLITE_PARAMETERS):
+            puzzle_hashes_db: Tuple[Any, ...]
+            if self.db_wrapper.db_version == 2:
+                puzzle_hashes_db = tuple(puzzles)
+            else:
+                puzzle_hashes_db = tuple([ph.hex() for ph in puzzles])
```
altendky, line 322: This is a change unrelated to this PR, but maybe catch it while touching it.

```suggestion
                puzzle_hashes_db = tuple(ph.hex() for ph in puzzles)
```

chia/full_node/coin_store.py
```diff
@@ -341,23 +345,24 @@ async def get_coin_records_by_parent_ids(
...
-            for row in await cursor.fetchall():
-                coin = self.row_to_coin(row)
-                coins.add(CoinRecord(coin, row[0], row[1], row[2], row[6]))
-            return list(coins)
+        for ids in chunks(parent_ids, MAX_SQLITE_PARAMETERS):
+            parent_ids_db: Tuple[Any, ...]
+            if self.db_wrapper.db_version == 2:
+                parent_ids_db = tuple(ids)
+            else:
+                parent_ids_db = tuple([pid.hex() for pid in ids])
```
altendky, line 353: As above, this is unrelated to the change.  Ignore if you want it in a separate PR.

```suggestion
                parent_ids_db = tuple(pid.hex() for pid in ids)
```

chia/full_node/coin_store.py
```diff
@@ -369,22 +374,22 @@ async def get_coin_states_by_ids(
...
-
-            for row in await cursor.fetchall():
-                coins.add(self.row_to_coin_state(row))
-            return list(coins)
+        for ids in chunks(coin_ids, MAX_SQLITE_PARAMETERS):
+            coin_ids_db: Tuple[Any, ...]
+            if self.db_wrapper.db_version == 2:
+                coin_ids_db = tuple(ids)
+            else:
+                coin_ids_db = tuple([pid.hex() for pid in ids])
```
altendky, line 382: likewise

```suggestion
                coin_ids_db = tuple(pid.hex() for pid in ids)
```

chia/full_node/coin_store.py
```diff
@@ -311,24 +314,25 @@ async def get_coin_states_by_puzzle_hashes(
...
+            async with self.coin_record_db.execute(
+                f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
+                f"coin_parent, amount, timestamp FROM coin_record INDEXED BY coin_puzzle_hash "
+                f'WHERE puzzle_hash in ({"?," * (len(puzzles) - 1)}?) '
+                f"AND (confirmed_index>=? OR spent_index>=?)"
+                f"{'' if include_spent_coins else 'AND spent_index=0'}",
+                puzzle_hashes_db + (min_height, min_height),
+            ) as cursor:
+
+                async for row in cursor:
```
altendky, line 332: I think I like the middle ground of multiple `.fetchmany()` calls in general.  But, maybe in this form we just want to have the batching be based on the desired fetch batch size rather than the limit of the max query params?  Like maybe we want a batch size of 100 or something and go back to `.fetchall()`?

chia/full_node/coin_store.py
```diff
@@ -7,11 +7,14 @@
 from chia.util.db_wrapper import DBWrapper
 from chia.util.ints import uint32, uint64
 from chia.util.lru_cache import LRUCache
+from chia.util.chunks import chunks
```
altendky, line 10:
```suggestion
from chia.util.chunks import chunks
from chia.util.db_wrapper import DBWrapper
from chia.util.ints import uint32, uint64
from chia.util.lru_cache import LRUCache
```

chia/full_node/weight_proof.py
```diff
@@ -20,6 +20,7 @@
     calculate_sp_iters,
     is_overflow_block,
 )
+from chia.util.chunks import chunks
```
altendky, line 23: Belongs down a bit, GitHub won't let me make a suggestion for this one.

chia/wallet/wallet_node.py
```diff
@@ -6,7 +6,7 @@
 import traceback
 from asyncio import CancelledError
 from pathlib import Path
-from typing import Callable, Dict, List, Optional, Set, Tuple, Any
+from typing import Callable, Dict, List, Optional, Set, Tuple, Any, Iterator
```
altendky, line 9:
```suggestion
from typing import Callable, Dict, Iterator, List, Optional, Set, Tuple, Any
```

chia/wallet/wallet_node.py
```diff
@@ -21,7 +21,7 @@
     KeychainProxy,
     KeyringIsEmpty,
 )
-from chia.full_node.weight_proof import chunks
+from chia.util.chunks import chunks
```
altendky, line 24: Belongs lower down.

tests/util/test_chunks.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_chunks() -> None:
+
+    assert list(chunks([], 0)) == []
+    assert list(chunks(["a"], 0)) == [["a"]]
+    assert list(chunks(["a", "b"], 0)) == [["a"], ["b"]]
+
+    assert list(chunks(["a", "b", "c", "d"], 2)) == [["a", "b"], ["c", "d"]]
+    assert list(chunks(["a", "b", "c", "d"], 200)) == [["a", "b", "c", "d"]]
```
altendky, line 11:
```suggestion
    assert list(chunks(["a", "b", "c", "d"], 4)) == [["a", "b", "c", "d"]]
    assert list(chunks(["a", "b", "c", "d"], 200)) == [["a", "b", "c", "d"]]

    assert list(chunks(["a", "b", "c", "d"], 3)) == [["a", "b", "c"], ["d"]]
```

altendky (changes requested):
Only really needed change is the extra test cases at the end.  The rest are just comments that you may or may not want to include.  The import order will otherwise get addressed when we reduce the isort ignore list size.  The unneeded `[]` are pre-existing.

chia/full_node/weight_proof.py
arvidn: does `isort` not fix this?
If not, what rule are we following?

chia/full_node/coin_store.py
arvidn: I was under the impression list-comprehension was specific to lists, and that you had to include the `[` and `]`. is the expression `f(x) for x in y` producing an iterator?
but then, the expression `[x]` is a list of `x`, even if `x` is an iterator, right?

chia/full_node/coin_store.py
arvidn: I would like to keep it simple until we measure it to be too slow. It probably is a bit slower with an `async for` since it will (probably) need to check the future every turn through the loop, but I wouldn't think it makes a big difference.

arvidn:
I'm hesitant to manually apply `isort` rules. I think it would make more sense to remove some files from the exemption. But maybe not mixed in here.

arvidn:
I added the test cases

chia/full_node/coin_store.py
altendky: > I was under the impression list-comprehension was specific to lists, and that you had to include the `[` and `]`.

There are list comprehensions but also dict and set comprehensions.  Plus generator comprehensions (previously referred to as generator expressions).

> is the expression `f(x) for x in y` producing an iterator?

A small caveat.  `(f(x) for x in y)` is a generator comprehension and creates a generator which does indeed implement the iterator protocol.  If you pass it to a function as the only parameter then you can drop the parentheses for the generator comprehension and just use the existing function call parentheses like in `a_function(f(x) for x in y)`, or `tuple(f(x) for x in y)` as in the code here.

> but then, the expression `[x]` is a list of `x`, even if `x` is an iterator, right?

Yes.  To add to that, if you want to create a list and iterate it to get the elements of the list you would instead `list(the_iterable)`.

chia/full_node/weight_proof.py
altendky: `isort` is configured to ignore presently-not-compliant files.  mypy has a similar setup.  We just have giant lists of exceptions basically so new files will be clean and we can go through existing files one at a time, clean them, then remove them from the exceptions list.

altendky:
> I'm hesitant to manually apply `isort` rules. I think it would make more sense to remove some files from the exemption. But maybe not mixed in here.

I wouldn't do the whole thing.  I would encourage a habit of putting new imports where they belong so there's less delta when we swing through with dedicated `isort` PRs.  Remember that concerns about order being changed in an automated fashion is why we have exceptions to begin with.  Keeping new additions tidy along the way will make the cleanup easier and safer to get merged.

But yes, I'm not pushing for a change here.

altendky (approved):

PR #10361, opened 2022-02-21
Bug fixes.

fchirica:


chia/data_layer/data_layer.py
```diff
@@ -183,7 +184,7 @@ async def _validate_batch(
                 continue
             # Pick the latest root in our data store with the desired hash, before our already validated data.
             root: Optional[Root] = await self.data_store.get_last_tree_root_by_hash(
-                tree_id, record.root, max_generation
+                tree_id, None if record.root == bytes32([0] * 32) else record.root, max_generation
```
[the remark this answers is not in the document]
fchirica: I'm slightly afraid changing data layer representation can break more stuff. (i.e. functions in data layer thinking it's None, while it has changed to bytes32). I've just picked minimal changes to map the 2, but in the future I think it'll be good for them to match.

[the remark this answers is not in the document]
fchirica: also, this is not only in the creation case. You can get  bytes32([0] * 32 when someone deletes all the data, then commits on chain.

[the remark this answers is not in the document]
altendky: There needs to be a definition of how an empty store is hashed since we have to put a hash on chain.  My impression is that @Quexington chose `0x000000...` for the on-chain representation and then that propagated out through the Python side of the wallet.  I think Bram mentioned the hash of nil being an option as well (I don't know what that ends up being).  I'm all for a consistent form in the Python code.  If the form is the on-chain hash then it should be defined in one place and that variable should be referenced (rather than writing `bytes32([0] * 32)` everywhere).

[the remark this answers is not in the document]
almogdepaz: i dont see why this should break stuff feels like it would help avoid issues by us not needing to swap out the nones in favor of the 0x000.. in all kinds of places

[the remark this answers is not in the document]
fchirica: we need to track in all data store code where we compare with None, and replace it with 0x000... it feels like a much bigger change than what I've done, so it has more opportunities to add extra bugs. although I agree in the long run, data store should use 0x0000.. as well

chia/data_layer/data_store.py
```diff
@@ -982,14 +983,18 @@ async def set_validated_wallet_generation(self, tree_id: bytes32, generation: in
...
     ) -> Set[DiffData]:
         async with self.db_wrapper.locked_transaction(lock=lock):
-            old_pairs = set() if hash_1 is None else set(await self.get_keys_values(tree_id, hash_1, lock=False))
-            new_pairs = set() if hash_2 is None else set(await self.get_keys_values(tree_id, hash_2, lock=False))
+            old_pairs = set(await self.get_keys_values(tree_id, hash_1, lock=False))
+            new_pairs = set(await self.get_keys_values(tree_id, hash_2, lock=False))
+            if len(old_pairs) == 0 and hash_1 != bytes32([0] * 32):
+                return set()
+            if len(new_pairs) == 0 and hash_2 != bytes32([0] * 32):
+                return set()
```
emlowe, line 997: Is there a test that exercises this logic? If not can we add one?

chia/data_layer/data_layer.py
```diff
@@ -167,23 +169,25 @@ async def get_root_history(self, store_id: bytes32) -> List[SingletonRecord]:
...
             if prev is None or record.root != prev.root:
                 root_history.append(record)
+                prev = record
         return root_history
 
     async def _validate_batch(
         self,
         tree_id: bytes32,
         to_check: List[SingletonRecord],
+        last_checked_hash: bytes32,
```
altendky, line 179: i feel like the caller should filter this.

chia/data_layer/data_layer.py
```diff
@@ -267,37 +270,51 @@ async def fetch_and_validate(self, subscription: Subscription) -> None:
...
 
         # Light validation: check the new set of operations against the new set of wallet records.
         # If this matches, we know all data will match, as we've previously checked that data matches
         # for `min_generation` data store root and `wallet_current_generation` wallet record.
-        is_valid: bool = await self._validate_batch(tree_id, to_check, min_generation, max_generation)
+        try:
+            is_valid: bool = await self._validate_batch(
+                tree_id, to_check, last_checked_hash, min_generation, max_generation
+            )
+        except Exception as e:
```
altendky, line 294: This consumes cancellation on Python 3.7.

chia/data_layer/data_layer.py
```diff
@@ -267,37 +270,51 @@ async def fetch_and_validate(self, subscription: Subscription) -> None:
...
-                f"Validated chain hash {root.node_hash} in downloaded datastore. "
+                f"Validated chain hash {to_check[0].root} in downloaded datastore. "
                 f"Wallet generation: {to_check[0].generation}"
             )
+            last_checked_hash = to_check[0].root
             to_check.pop(0)
-            is_valid = await self._validate_batch(tree_id, to_check, 0, max_generation)
+            try:
+                is_valid = await self._validate_batch(tree_id, to_check, last_checked_hash, 0, max_generation)
+            except Exception as e:
```
altendky, line 311: This consumes cancellation on Python 3.7.

chia/data_layer/data_store.py
```diff
@@ -376,16 +376,17 @@ async def get_roots_between(
         return roots
 
     async def get_last_tree_root_by_hash(
-        self, tree_id: bytes32, hash: bytes32, max_generation: Optional[int] = None, *, lock: bool = True
+        self, tree_id: bytes32, hash: Optional[bytes32], max_generation: Optional[int] = None, *, lock: bool = True
     ) -> Optional[Root]:
         async with self.db_wrapper.locked_transaction(lock=lock):
             max_generation_str = f"AND generation < {max_generation} " if max_generation is not None else ""
+            node_hash_str = "AND node_hash == :node_hash " if hash is not None else "AND node_hash is NULL "
```
altendky, line 383: I wasn't aware that you can't equality check to `NULL`.  That's a pain.

chia/data_layer/data_store.py
fchirica: added now

altendky, line 997: Should these be using the none bytes constant?

Is this check compensating for the store returning an empty set for roots we won't have?  Shouldn't that be fixed there an an error be the result instead of just propagating the empty set?

chia/data_layer/data_layer.py
```diff
@@ -32,6 +32,7 @@ class DataLayer:
     state_changed_callback: Optional[Callable[..., object]]
     wallet_id: uint64
     initialized: bool
+    none_bytes: bytes32
```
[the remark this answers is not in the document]
fchirica: IMO it shouldn't be in `ConsensusConstants` (the global storage of constants), since it has nothing to do with consensus.

chia/data_layer/data_store.py
fchirica: not sure about SQLite, but for SQL this generally seems to cause problems, as `NULL == NULL` returns false:

https://stackoverflow.com/questions/9581745/sql-is-null-and-null

I think it's safer to keep it in this form regardless.

chia/data_layer/data_store.py
fchirica: this can happen quite frequently when history returns the wallet stored history, yet our local store is behind (waiting for the download to complete). we need a way to signal this to CW - empty set would be consistent with `get_keys_values` returning {} when we're behind in download.

emlowe: An error would be reasonable, however, as Florin says the empty set is consistent with some other places we return empty - like for a hash that is on-chain, but we don't have the data for (where an error would also be reasonable, but currently returns an empty thing)
Do we have a constant for `bytes32([0] * 32)`? That would seem a reasonable thing.

chia/data_layer/data_layer.py
emlowe: Is adding this a reasonable thing to do?
```
    raise

File: chia/data_layer/data_layer.py (hunk 3 above)
Comment (fchirica) [reply chia/data_layer/data_layer.py]: refactored `_validate_batch` to include the first check as well, so we don't have to pass this in.

File: chia/data_layer/data_layer.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (altendky) [reply chia/data_layer/data_layer.py:35]: Sure.  Constants can go anywhere they make sense.

File: chia/data_layer/data_layer.py (hunk 4 above)
Comment (altendky) [reply chia/data_layer/data_layer.py:294]: That's the basic gist, yeah.  Other than just catching exceptions you know what to do with.

File: chia/data_layer/data_store.py (hunk 6 above)
Comment (altendky) [reply chia/data_layer/data_store.py:383]: Yeah, I was going to suggest a change but went and tested it first.  I agree it needs to be this way.

File: chia/data_layer/data_store.py (hunk 2 above)
Comment (altendky) [reply chia/data_layer/data_store.py:997]: I think that at some point we should stop claiming there is nothing for a hash when we do not know if there is or isn't.

**Review by altendky — DISMISSED** [review]
Approved pending CI.

File: chia/data_layer/data_layer.py [hunk 8]
```diff
@@ -167,23 +169,37 @@ async def get_root_history(self, store_id: bytes32) -> List[SingletonRecord]:
...
     async def _validate_batch(
         self,
         tree_id: bytes32,
         to_check: List[SingletonRecord],
         min_generation: int,
-        max_generation: int,
     ) -> bool:
-        last_checked_hash: Optional[bytes32] = None
+        root: Optional[Root] = await self.data_store.get_tree_root(tree_id=tree_id)
+        assert root is not None
```
altendky, line 182:
```suggestion
        root = await self.data_store.get_tree_root(tree_id=tree_id)
```

altendky (changes requested):

chia/data_layer/data_layer.py
altendky: The `Optional` was introduced here by the extra unneeded type hint.

https://github.com/Chia-Network/chia-blockchain/blob/dad8d7a221cf029754e3cd7c6412f85419f55688/chia/data_layer/data_store.py#L335

altendky: Looks like down below there is a reuse of this same variable name that also drives the `Optional`.

```
root = await self.data_store.get_last_tree_root_by_hash(
```

Perhaps these names should be more descriptive.  Or maybe it would be helpful if exceptional cases like requesting a thing we don't have resulted in exceptions.

altendky (approved):
Approving to merge, but please clean up the hints/`assert` bit anyways.

chia/data_layer/data_layer.py
fchirica: so should I rename the 2nd root differently, so it can take an optional, while the first one can keep the Root only type?

altendky: Different names is probably a good idea if the two uses are referring to different logical things.  Also, `Optional` may just be a pain here.  Is getting `None` out of `DataStore.get_last_tree_root_by_hash()` really a 'normal' case?  Or is it exceptional?  The pragmatic question is 'do you want to write all the return value checking at every call site at every layer?'  Or can it just be allowed to propagate up via an exception.

PR #10362, opened 2022-02-21
Wallet record2index change

neurosis69:
There is no need to select the whole record if only the index is used.


neurosis69:
Got obsolete in the meantime.

PR #10364, opened 2022-02-21
farmer: Send a `PUT /farmer` if the signature verification failed

xdustinface:
Due to the bug which was fixed in #9922 some pools obviously still have incorrect authentication public keys from their farmers. This PR is to make sure that the farmers update their authentication public key on the pool if they get an `INVALID_SIGNATURE` response error from a `GET /farmer`.

### DIFF
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -472,13 +472,13 @@ class Farmer:
                 if time.time() >= pool_state["next_farmer_update"]:
                     authentication_token_timeout = pool_state["authentication_token_timeout"]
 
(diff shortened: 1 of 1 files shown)


mariano54 (dismissed):
Thank you! have you tested with a pool to see if it works?

xdustinface:
> Thank you! have you tested with a pool to see if it works?

Nope see Keybase, @wjblanke was going to ask someone to test. But see my last push, it was not correct because it did depend on `farmer_info` to be `not None` in both cases but it's `None` if the signature was invalid.

mariano54 (approved):
Approve, pending test with a bad pool state

PR #10323, opened 2022-02-19
changed column selection: table `transaction_record`

neurosis69:
Follow up to #10300 

Basic comparison, table has only 1 row
```
sqlite> explain select transaction_record from transaction_record;
addr  opcode         p1    p2    p3    p4             p5  comment      
----  -------------  ----  ----  ----  -------------  --  -------------
0     Init           0     7     0                    00  Start at 7   
1     OpenRead       0     10    0     1              00  root=10 iDb=0; transaction_record
2     Rewind         0     6     0                    00               
3       Column         0     0     1                    00  r[1]=transaction_record.transaction_record
4       ResultRow      1     1     0                    00  output=r[1]  
5     Next           0     3     0                    01               
6     Halt           0     0     0                    00               
7     Transaction    0     0     34    0              01  usesStmtJournal=0
8     Goto           0     1     0                    00               
Run Time: real 0.001 user 0.000223 sys 0.000000
```

```
sqlite> explain select * from transaction_record;
addr  opcode         p1    p2    p3    p4             p5  comment      
----  -------------  ----  ----  ----  -------------  --  -------------
0     Init           0     18    0                    00  Start at 18  
1     OpenRead       0     10    0     12             00  root=10 iDb=0; transaction_record
2     Rewind         0     17    0                    00               
3       Column         0     0     1                    00  r[1]=transaction_record.transaction_record
4       Column         0     1     2                    00  r[2]=transaction_record.bundle_id
5       Column         0     2     3                    00  r[3]=transaction_record.confirmed_at_height
6       Column         0     3     4                    00  r[4]=transaction_record.created_at_time
7       Column         0     4     5                    00  r[5]=transaction_record.to_puzzle_hash
8       Column         0     5     6                    00  r[6]=transaction_record.amount
9       Column         0     6     7                    00  r[7]=transaction_record.fee_amount
10      Column         0     7     8                    00  r[8]=transaction_record.confirmed
11      Column         0     8     9                    00  r[9]=transaction_record.sent
12      Column         0     9     10                   00  r[10]=transaction_record.wallet_id
13      Column         0     10    11                   00  r[11]=transaction_record.trade_id
14      Column         0     11    12                   00  r[12]=transaction_record.type
15      ResultRow      1     12    0                    00  output=r[1..12]
16    Next           0     3     0                    01               
17    Halt           0     0     0                    00               
18    Transaction    0     0     34    0              01  usesStmtJournal=0
19    Goto           0     1     0                    00               
[...]

PR #10300, opened 2022-02-18
changed column selection: table `derivation_paths`

neurosis69:
To reduce overhead and readability, in selects use column names instead of *.

Very simple example, selecting the whole row compared with the first column ( table has ~20k rows)
```
sqlite> explain select derivation_index from derivation_paths;
addr  opcode         p1    p2    p3    p4             p5  comment      
----  -------------  ----  ----  ----  -------------  --  -------------
0     Init           0     7     0                    00  Start at 7   
1     OpenRead       1     20    0     k(2,,)         00  root=20 iDb=0; derivation_index_index
2     Rewind         1     6     1     0              00               
3       Column         1     0     1                    00  r[1]=derivation_paths.derivation_index
4       ResultRow      1     1     0                    00  output=r[1]  
5     Next           1     3     0                    01               
6     Halt           0     0     0                    00               
7     Transaction    0     0     34    0              01  usesStmtJournal=0
8     Goto           0     1     0                    00               
Run Time: real 0.000 user 0.000108 sys 0.000114
```

```
sqlite> explain select * from derivation_paths;
addr  opcode         p1    p2    p3    p4             p5  comment      
----  -------------  ----  ----  ----  -------------  --  -------------
0     Init           0     13    0                    00  Start at 13  
1     OpenRead       0     19    0     7              00  root=19 iDb=0; derivation_paths
2     Rewind         0     12    0                    00               
3       Column         0     0     1                    00  r[1]=derivation_paths.derivation_index
4       Column         0     1     2                    00  r[2]=derivation_paths.pubkey
5       Column         0     2     3                    00  r[3]=derivation_paths.puzzle_hash
6       Column         0     3     4                    00  r[4]=derivation_paths.wallet_type
7       Column         0     4     5                    00  r[5]=derivation_paths.wallet_id
8       Column         0     5     6                    00  r[6]=derivation_paths.used
9       Column         0     6     7                    00  r[7]=derivation_paths.hardened
10      ResultRow      1     7     0                    00  output=r[1..7]
11    Next           0     3     0                    01               
12    Halt           0     0     0                    00               
13    Transaction    0     0     34    0              01  usesStmtJournal=0
14    Goto           0     1     0                    00               
Run Time: real 0.001 user 0.000302 sys 0.000000
```

If at all there may be very little performance improvement changing this, but still it brings more readability and less instructions. 
[How to Interpret the Bytecode](https://www.sqlite.org/opcode.html#viewing_the_bytecode)

PR #10368, opened 2022-02-22
changed column selection: table `trade_records`

neurosis69:
Follow up to #10300, #10323

Same reason as for the other PRs, only select the columns needed, saves memory, cpu cycles and execution time.


arvidn (dismissed):
this is a great change! we should never use the `SELECT * ` form, but always be explicit with which columns we request and in which order

emlowe (approved):

PR #10369, opened 2022-02-22
use a proper temporary file for the blockchain database in tests

arvidn:
The problem with the current approach is that if you run tests in parallel, in separate processes, they won't share the global counter variable, and they end up using the same file, which won't work.


arvidn:
The logic to remove the file at the end of the test is already in place in the fixtures that use this function

almogdepaz (approved):

wjblanke (approved):
aok

PR #10371, opened 2022-02-22
simplify the use of localhost in tests

arvidn:
There's no need to jump through these hoops to get to this string

### DIFF
--- a/tests/setup_nodes.py
+++ b/tests/setup_nodes.py
@@ -38,7 +38,9 @@ keychain = temp_keyring.get_keychain()
 atexit.register(cleanup_keyring, temp_keyring)  # Attempt to cleanup the temp keychain
 bt = create_block_tools(constants=test_constants, keychain=keychain)
 
-self_hostname = bt.config["self_hostname"]
+# if you have a system that has an unusual hostname for localhost and you want
+# to run the tests, change this constant
+self_hostname = "localhost"
 
 
 def constants_for_dic(dic):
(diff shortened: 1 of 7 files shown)


tests/setup_nodes.py
```diff
@@ -37,8 +37,6 @@ def cleanup_keyring(keyring: TempKeyring):
 atexit.register(cleanup_keyring, temp_keyring)  # Attempt to cleanup the temp keychain
 bt = create_block_tools(constants=test_constants, keychain=keychain)
 
-self_hostname = bt.config["self_hostname"]
```
arvidn, line 40: a lot of tests import this variable, which really is a constant; "localhost"

mariano54:
Just commenting on the fact that localhost is not mapped to anything on some systems, so this makes testing for these people harder. This was the motivation of a global variable.

arvidn:
I suppose it came up at some point then, someone testing on such a system. But it's quite tricky to figure out what to do if you are on such a system (so I can't imagine it's all that common). You have to change the `initial-config.yml`

Perhaps a better patch would be to just define `self_hostname` as the `"localhost"` string literal instead, rather than depending on `bt`.

arvidn:
I've updated the patch to instead embrace the global variable `self_hostname`, but still disconnect it from depending on the `initial-config.yaml` initializing the global variable `bt`. If the purpose is to make it possible to run tests that don't have the "localhost" name, all that is redundant (and adds quite a lot of complexity)

mariano54:
We still need the config `self_hostname`, because the file `start_service.py` uses it right? I think the idea behind the original way was that there was one unique place to configure hostname for both the tests and the services: in config.yaml. I'm not saying that this is not an improvement though, it's easier to track down if you're just running the tests.

arvidn:
> We still need the config self_hostname, because the file start_service.py uses it right?

Correct. This patch only affects the tests.

> I think the idea behind the original way was that there was one unique place to configure hostname for both the tests and the services: in config.yaml.

The tests won't look at `config.yaml`, they will create their own `config.yaml` in a temporary directory from `initial-config.yaml`. So either way, I would think you would need to configure it in your *installed* (`~/.chia/mainnet/config/config.yaml`) for production, and patch the code `chia/util/inital-config.yaml` for the tests.

mariano54 (approved):

PR #10374, opened 2022-02-22
Add auth key back

mariano54:
This allows downgrading from 1.3 to older versions.

This has been tested by running 1.3 then 1.2.11 and the error is no longer shown.


wjblanke (commented):
aok thanks!

wjblanke (approved):
aok thanks

PR #10376, opened 2022-02-22
asyncify wallet weight proof validation

altendky:
Follows from the testing and sleep additions for the full node weight proof validation in https://github.com/Chia-Network/chia-blockchain/pull/10163.

Draft for:
- [ ] Testing with `while true; do pkill chia_; ps; chia start wallet; sleep 75; ps; chia stop -d all; date --iso-8601=n; sleep 5; ps; pkill chia_; sleep 5; ps; done` and checking for `sending kill signal to chia_wallet` in the logs.


altendky:
This did not fully resolve the shutdown issues but it does seem good and we did identify an additional issue with a long sync deserialization activity.

wjblanke (approved):
aok

PR #9447, opened 2021-12-02
All from get transactions cli

emlowe:
Following on from PR #7979 .

Redoing this PR based on changes to backend queries in 1.3

PR #10380, opened 2022-02-22
Ms.fix get transactions

mariano54:
It was returning max 50, and they were ordered from oldest to newest.
Now allows configuring offset and limit, uses the offset to ask backend, and sorts them in the right order

### DIFF
--- a/chia/cmds/wallet.py
+++ b/chia/cmds/wallet.py
@@ -48,6 +48,15 @@ def get_transaction_cmd(wallet_rpc_port: Optional[int], fingerprint: int, id: in
     show_default=True,
     required=True,
 )
+@click.option(
+    "-l",
+    "--limit",
+    help="Max number of transactions to return",
+    type=int,
+    default=1000000000,
+    show_default=True,
+    required=False,
+)
 @click.option("--verbose", "-v", count=True, type=int)
 @click.option(
     "--paginate/--no-paginate",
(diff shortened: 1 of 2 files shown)


chia/cmds/wallet.py
```diff
@@ -48,6 +48,15 @@ def get_transaction_cmd(wallet_rpc_port: Optional[int], fingerprint: int, id: in
     show_default=True,
     required=True,
 )
+@click.option(
+    "-l",
+    "--limit",
+    help="Max number of transactions to return",
+    type=int,
+    default=1000000000,
```
emlowe, line 56: This is a pretty big change to the default from previous versions - and if I want "all" - well I guess this number is big enough for all.

mariano54: we could do default 50 like before, but i think it's reasonable to default to all. I just downloaded 36k transactions and it took like 20 seconds.

mariano54: the RPC doesn't have a way to request all, we much provide a limit. I guess we could do UINT32 max

emlowe: Yea , in PR #9447, I did add in an all option to the RPC and this then uses the existing get_all_transactions_for_wallet function (and a bunch of other changes that were invalidated by the big merge)

mariano54: Just updated that max. Your PR looks good to, i'm fine with merging that if you want to update. Otherwise I think we should merge this to fix the immediate problem of broken CLI. Many users will exceed the 50 tx limit soon.

emlowe: My PR needs a largish fix-up because it conflicts with changes already made in the merge (especially how the actual queries work). Yea, if the performance is fine, I guess I'm ok with it as just a big number - I agree we need to get this out

emlowe (approved):
aok with room for more improvements over time

Issue #10336, opened 2022-02-20
"Name" parameter missing from call to create_wallet_for_cat in create_new_wallet


joshpainter:
When using create_new_wallet with the ?name=CATNAME parameter, the new CAT wallet is still named "CAT WALLET." I tracked it down to this line of code where it seems the "name" variable up on line 432 is not getting used in the create_wallet_for_cat method call:

https://github.com/Chia-Network/chia-blockchain/blob/45c39f235b5ffa34398449913446656bf3a07e4e/chia/rpc/wallet_rpc_api.py#L448-L450

Because this parameter is null, the default "CAT WALLET" is used when the wallet is created: 

https://github.com/Chia-Network/chia-blockchain/blob/45c39f235b5ffa34398449913446656bf3a07e4e/chia/wallet/cat_wallet/cat_wallet.py#L149-L151

I think the "name" parameter just needs to be added to this method call to fix this?

PR #10381, opened 2022-02-22
Added CAT wallet name parameter

joshpainter:
The `create_wallet_for_cat` method call accepts a `name` parameter for the new CAT being created, but we weren't passing that `name` variable here so all new wallets created via this method end up being named "CAT WALLET" (the default for name parameter in `create_wallet_for_cat`) even if you specified a wallet name here.


joshpainter:
Fixes #10336

mariano54 (approved):

mariano54:
Thank you!

PR #10395, opened 2022-02-23
more test fixes

arvidn:
this is primarily fixing `test_mempool.py` to make the tests truly independent. It turned out that the testcases were using the same coins in their tests, so the order they run in (and whether all run or not) affected the tests.

This patch makes every test farm a fresh block to use their own reward coins for the test.

All this complexity could be avoided if every test had its own simulation, but our simulations can't be instantiated more than once right now.

There's also a few more hard-coded port number addressed.

### DIFF
--- a/tests/wallet/cat_wallet/test_cat_wallet.py
+++ b/tests/wallet/cat_wallet/test_cat_wallet.py
@@ -23,12 +23,6 @@ from tests.setup_nodes import setup_simulators_and_wallets
 from tests.time_out_assert import time_out_assert
 
 
-@pytest.fixture(scope="module")
-def event_loop():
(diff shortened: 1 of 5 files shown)


tests/wallet/cat_wallet/test_cat_wallet.py
```diff
@@ -23,12 +23,6 @@
 from tests.time_out_assert import time_out_assert
 
 
-@pytest.fixture(scope="module")
-def event_loop():
```
arvidn, line 27: some of these are necessary, for now (see comment in `test_mempool.py`). But I think a fair number of them have just been copy-pasted into new tests for no reason

mariano54 (approved):

PR #10396, opened 2022-02-23
Stop assuming `True == 1`

ChiaMineJP:
This might be really dangerous when future version of Python stop recognizing `True` as `1`.

# Changelog
- Changed `record.spent == 1`  to `record.spent` because the type of `record.spent` is `Boolean`, not `int`
- Removed `removal_coin_dict` because it was used just for constructing `removals: List[Coin]` later but `removal_record_dict` alone can do the work.
- Removed `tmp_error` because it was always `None`. Suspecting a debug code left forgotten.

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -486,7 +486,7 @@ class MempoolManager:
         for record in removals.values():
             removal = record.coin
             # 1. Checks if it's been spent already
-            if record.spent == 1:
+            if record.spent:
                 return Err.DOUBLE_SPEND, []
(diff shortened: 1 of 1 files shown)


altendky:
I'm not too worried about Python making `bool` not inherit from `int` or using a value other than `1` for `True`, but...  yeah, don't compare booleans to check their value.  `if the_bool:` is the way to go.  The rest seems unrelated?

mariano54 (approved):

ChiaMineJP:
@altendky Yes, rest is small optimization.

PR #10397, opened 2022-02-23
Add pytest to mypy pre-commit environment

altendky:


arvidn (dismissed):
this fixes the `pre-commit` issue for me!

arvidn:
@altendky actually, we need `pytest_asyncio` as well

arvidn (approved):

PR #10398, opened 2022-02-23
add db validate function to check consistency of blockchain database

arvidn:
this provides the foundation for a more sophisticated consistency check of the database. Currently it just checks the most basic invariants of the block table and (optionally) the encoded blocks as well (but that's slow).

There's currently no checking of coins, hints or sub epoch summaries.

The test is not exhaustive, but covers the most important part (at least in my mind) which is the `in_main_chain` field.

example output:

```
$ chia db validate
opening file for reading: /home/arvid/.chia/mainnet/db/blockchain_v2_mainnet.sqlite
peak hash: 8589c71d90582cc5536c0dacab957a1e40922e4fedb71c9c1f183bd90b2349bc
peak height: 1609723
traversing the full chain
0 orphaned blocks: 141       
141 orphaned blocks


DATABASE IS VALID: /home/arvid/.chia/mainnet/db/blockchain_v2_mainnet.sqlite
```

### DIFF
--- a/chia/cmds/db.py
+++ b/chia/cmds/db.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 import click
 from chia.cmds.db_upgrade_func import db_upgrade_func
+from chia.cmds.db_validate_func import db_validate_func
 
 
 @click.group("db", short_help="Manage the blockchain database")
@@ -8,7 +9,7 @@ def db_cmd() -> None:
     pass
 
 
-@db_cmd.command("upgrade", short_help="EXPERIMENTAL: upgrade a v1 database to v2")
+@db_cmd.command("upgrade", short_help="upgrade a v1 database to v2")
 @click.option("--input", default=None, type=click.Path(), help="specify input database file")
 @click.option("--output", default=None, type=click.Path(), help="specify output database file")
 @click.option(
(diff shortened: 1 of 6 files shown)


mariano54 (approved):

chia/cmds/db.py
```diff
@@ -34,7 +35,22 @@ def db_upgrade_cmd(ctx: click.Context, no_update_config: bool, **kwargs) -> None
...
+@click.option(
+    "--validate-blocks",
+    default=False,
+    is_flag=True,
+    help="validate consistency of properties of the encoded blocks and block records",
+)
+@click.pass_context
+def db_validate_cmd(ctx: click.Context, validate_blocks: bool, **kwargs) -> None:
+    try:
+        in_db_path = kwargs.get("input")
```
ALTracer, line 49: This doesn't match the `--db` argument, so it silently discards the filename given on cmdline.
Discovered when checking manually the incomplete upgrade, and the config was still pointing to the v1 file.
Change to `kwargs.get("db")` or revert the flag to `--input`?

arvidn: thanks for the report! fixed here: https://github.com/Chia-Network/chia-blockchain/pull/10716

PR #10404, opened 2022-02-23
Only show -Trusted when a node is trusted

hoffmang9:
This modifies the CLI back to not showing the status of Trusted unless a node is trusted.

However, this raises a question - I did not think we trusted a node as a node and https://github.com/Chia-Network/chia-blockchain/commit/7fa1861def5797f63edfd2ad54f3efb77b52695b may have pulled wallet code into node showing @mariano54 ?


arvidn (approved):

mariano54:
I don't think we use this feature now but we might use the trusted feature in the future.

mariano54 (approved):

PR #10409, opened 2022-02-24
Check that offer coins exist as well as not spent

Quexington:
I had an incorrect assumption about how `get_coin_states` worked and the `check_offer_validity` would return true even if some of the coins in the offer did not exist.

### DIFF
--- a/chia/wallet/trade_manager.py
+++ b/chia/wallet/trade_manager.py
@@ -353,7 +353,6 @@ class TradeManager:
         coin_states = await self.wallet_state_manager.wallet_node.get_coin_state(
             [c.name() for c in non_ephemeral_removals]
         )
-        assert coin_states is not None
         return len(coin_states) == len(non_ephemeral_removals) and all([cs.spent_height is None for cs in coin_states])
 
     async def calculate_tx_records_for_offer(self, offer: Offer, validate: bool) -> List[TransactionRecord]:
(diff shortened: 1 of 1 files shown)


chia/wallet/trade_manager.py
```diff
@@ -354,7 +354,7 @@ async def check_offer_validity(self, offer: Offer) -> bool:
             [c.name() for c in non_ephemeral_removals]
         )
         assert coin_states is not None
```
emlowe, line 356: Had this discussion with Kyle recently about using `assert`. Not sure why we have this assert - get_coin_state() as defined doesn't return Optional, but if it did, is this any different from an empty list? If this is really an exceptional case, then I would raise something. - so to me this should be either
```
 if coin_states is None:
    raise TypeError("coin_states should not be NoneType")
```
or (assuming an empty list would always be false ?)
```
if not coin_states:
  return False
```

Quexington: I think this is legacy maybe because it can never be none for a hilarious reason: https://github.com/Chia-Network/chia-blockchain/blob/55c4eb3d289bcb99f92da259ed9b1d1d5b8d589c/chia/wallet/wallet_node.py#L1342

Quexington: I can take it out

emlowe: If both sides end up having length == 0  this will return True - is that correct?

emlowe (approved):
looks ok to me as I read it

PR #10415, opened 2022-02-24
Only update height if we have processed everything before that height.

mariano54:
This ensures that if a sync is cancelled (which is normal) we won't set the peak to a height where we have not processed everything before it.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -603,9 +603,10 @@ class WalletNode:
 
         all_tasks: List[asyncio.Task] = []
         target_concurrent_tasks: int = 20
-        num_concurrent_tasks: int = 0
+        concurrent_tasks_cs_heights: List[uint32] = []
 
         async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: int):
+            nonlocal concurrent_tasks_cs_heights
             try:
                 assert self.validation_semaphore is not None
                 async with self.validation_semaphore:
@@ -616,7 +617,7 @@ class WalletNode:
                             self.add_state_to_race_cache(header_hash, height, inner_state)
                             self.log.info(f"Added to race cache: {height}, {inner_state}")
                     if trusted:
-                        valid_states = inner_states
+                        valid_states = inner_states.copy()
                     else:
                         valid_states = [
                             inner_state
@@ -626,21 +627,26 @@ class WalletNode:
                     if len(valid_states) > 0:
                         self.log.info(
                             f"new coin state received ({inner_idx_start}-"
-                            f"{inner_idx_start + len(inner_states) - 1}/ {len(items) + 1})"
+                            f"{inner_idx_start + len(inner_states) - 1}/ {len(items)})"
                         )
                         assert self.new_state_lock is not None
                         async with self.new_state_lock:
                             await self.wallet_state_manager.new_coin_state(valid_states, peer, fork_height)
-                        if update_finished_height:
-                            await self.wallet_state_manager.blockchain.set_finished_sync_up_to(
-                                last_change_height_cs(inner_states[-1])
-                            )
+
+                            if update_finished_height:
+                                if len(concurrent_tasks_cs_heights) == 1:
+                                    # We have processed all past tasks so we can
+                                    synced_up_to = last_change_height_cs(inner_states[-1]) - 1
+                                else:
+                                    # We know we have processed everything before this min height
+                                    synced_up_to = min(concurrent_tasks_cs_heights)
+                                await self.wallet_state_manager.blockchain.set_finished_sync_up_to(synced_up_to)
+
             except Exception as e:
                 tb = traceback.format_exc()
                 self.log.error(f"Exception while adding state: {e} {tb}")
             finally:
-                nonlocal num_concurrent_tasks
-                num_concurrent_tasks -= 1  # pylint: disable=E0602
+                concurrent_tasks_cs_heights.remove(last_change_height_cs(inner_states[0]))  # pylint: disable=E0602
 
         idx = 1
         # Keep chunk size below 1000 just in case, windows has sqlite limits of 999 per query
@@ -653,13 +659,13 @@ class WalletNode:
             if peer.peer_node_id not in self.server.all_connections:
                 self.log.error(f"Disconnected from peer {peer.peer_node_id} host {peer.peer_host}")
                 return False
-            while num_concurrent_tasks >= target_concurrent_tasks:
+            while len(concurrent_tasks_cs_heights) >= target_concurrent_tasks:
                 await asyncio.sleep(0.1)
                 if self._shut_down:
                     self.log.info("Terminating receipt and validation due to shut down request")
                     return False
+            concurrent_tasks_cs_heights.append(last_change_height_cs(states[0]))
             all_tasks.append(asyncio.create_task(receive_and_validate(states, idx)))
-            num_concurrent_tasks += 1
             idx += len(states)
 
         await asyncio.gather(*all_tasks)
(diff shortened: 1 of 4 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -636,14 +637,13 @@ async def new_coin_state(
...
 
-        # Sort by created height, then add the reorg states (created_height is None) to the end
-        created_h_none: List[CoinState] = []
-        for coin_st in coin_states.copy():
-            if coin_st.created_height is None:
-                coin_states.remove(coin_st)
-                created_h_none.append(coin_st)
-        coin_states.sort(key=lambda x: x.created_height, reverse=False)  # type: ignore
-        coin_states.extend(created_h_none)
+        # Input states should already be sorted by cs_height, with reorgs at the beginning
```
mariano54, line 640: Reorgs are ignored anyway, but I think they have to be at the beginning also.

chia/wallet/wallet_node.py
```diff
@@ -603,20 +604,23 @@ async def receive_state_from_peer(
...
                 async with self.validation_semaphore:
-                    assert self.wallet_state_manager is not None
                     if header_hash is not None:
                         assert height is not None
                         for inner_state in inner_states:
                             self.add_state_to_race_cache(header_hash, height, inner_state)
                             self.log.info(f"Added to race cache: {height}, {inner_state}")
                     if trusted:
-                        valid_states = inner_states
+                        valid_states = inner_states.copy()
```
mariano54, line 623: I don't think this is needed anymore, it was before because we were mutating things inside `new_coin_state`

chia/wallet/wallet_node.py
```diff
@@ -626,21 +630,28 @@ async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: i
...
+                                return
                             await self.wallet_state_manager.new_coin_state(valid_states, peer, fork_height)
-                        if update_finished_height:
-                            await self.wallet_state_manager.blockchain.set_finished_sync_up_to(
-                                last_change_height_cs(inner_states[-1])
-                            )
+
+                            if update_finished_height:
+                                if len(concurrent_tasks_cs_heights) == 1:
+                                    # We have processed all past tasks so we can
```
Nikolaj-K, line 643: Is "so we can" the intended end of the sentence?

chia/wallet/wallet_node.py
```diff
@@ -626,21 +630,28 @@ async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: i
...
-                                last_change_height_cs(inner_states[-1])
-                            )
+
+                            if update_finished_height:
+                                if len(concurrent_tasks_cs_heights) == 1:
+                                    # We have processed all past tasks, so we can increase the height safely
+                                    synced_up_to = last_change_height_cs(valid_states[-1]) - 1
+                                else:
+                                    # We know we have processed everything before this min height
+                                    synced_up_to = min(concurrent_tasks_cs_heights)
```
altendky, line 647: Let's say you start up tasks for 100-199, 200-299, and 300-399.  The first one fails out.  The third one gets here.  The minimum is going to be 200.  We have not process 100-199 but this code will set the finished sync up to to 200.  Right?

chia/wallet/wallet_node.py
```diff
@@ -729,15 +740,20 @@ async def state_update_received(self, request: wallet_protocol.CoinStateUpdate,
...
         nodes = self.server.get_full_node_connections()
         if len(nodes) > 0:
             return random.choice(nodes)
         else:
             return None
 
     async def disconnect_and_stop_wpeers(self):
-        # Close connection of non trusted peers
+        if self.server is None:
+            return False
```
altendky, line 754: The existing return hint for this function would have been `None`.  Do we want to change it to `bool` or should this just be `return`?

chia/wallet/wallet_node.py
```diff
@@ -747,6 +763,14 @@ async def disconnect_and_stop_wpeers(self):
             await self.wallet_peers.ensure_is_closed()
             self.wallet_peers = None
 
+    async def check_for_synced_peer(self, header_block: HeaderBlock, request_time: uint64):
```
altendky, line 766: Isn't this `check_for_synced_trusted_peer()`?

```suggestion
    async def check_for_synced_peer(self, header_block: HeaderBlock, request_time: uint64) -> bool:
```

chia/wallet/wallet_state_manager.py
```diff
@@ -636,14 +637,13 @@ async def new_coin_state(
...
+        curr_h = -1
+        for c_state in coin_states:
+            if last_change_height_cs(c_state) < curr_h:
+                raise ValueError("Input coin_states is not sorted properly")
+            curr_h = last_change_height_cs(c_state)
```
altendky, line 645: Function calls aren't the cheapest thing ever for tight loops.  I don't think this particularly reduces legibility so it's probably a good change.

```suggestion
            last_change_height = last_change_height_cs(c_state)
            if last_change_height < curr_h:
                raise ValueError("Input coin_states is not sorted properly")
            curr_h = last_change_height
```

chia/wallet/wallet_node.py
mariano54: If a task fails, then we will not add that state. This is an issue with trusted sync.
WIth untrusted, it's fine, since we always sync from scratch from new peers so we will catch it in the next sync. 

Not sure how to easily solve this problem, maybe another queue for failed CoinStates?

altendky: How about a sorted list. when anything leaves it puts it's group in the list and then checked the already synced to height vs. the first element in the sorted list.  If that group is the next one after the existing height we can move forward.  Loop until either the end of the list or the first non-contiguous item (can use the same check as at first).  Probably better not to remove leading elements so figure out how many are to be removed then `the_list[:] = the_list[x:]` to mutate the existing list in one go.

https://docs.python.org/3/library/bisect.html#bisect.insort_right

mariano54: This is to support handling states that failed to add? I'm not sure about this, there are many reasons why a state might not add. Perhaps it's an invalid state, or there is a bug (or duplicate). I would just skip it or add it to some other queue to reprocess later. I think we should not handle it in this PR.

chia/wallet/wallet_node.py
```diff
@@ -626,21 +628,28 @@ async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: i
...
                             await self.wallet_state_manager.new_coin_state(valid_states, peer, fork_height)
-                        if update_finished_height:
-                            await self.wallet_state_manager.blockchain.set_finished_sync_up_to(
-                                last_change_height_cs(inner_states[-1])
-                            )
+
+                            if update_finished_height:
+                                if len(cs_heights) == 1:
+                                    # We have processed all past tasks, so we can increase the height safely
+                                    synced_up_to = last_change_height_cs(valid_states[-1]) - 1
```
altendky, line 642: If this is the last batch to finish, but not the highest batch, then we will end up setting the synced up to height below where we have actually synced to.  This is safe in the sense that it doesn't make you miss coins, it just makes you reprocess them.

chia/wallet/wallet_node.py
```diff
@@ -626,21 +628,28 @@ async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: i
...
+                                else:
+                                    # We know we have processed everything before this min height
+                                    synced_up_to = min(cs_heights)
```
altendky, line 645: We know that we have initiated a task to attempt to process up to that height.  It may have failed out for any number of reasons prior to even attempting to validate the coin states.

altendky (approved):
More to do, this improves handling of some cases.

PR #10419, opened 2022-02-25
streamable: Cache stream functions

xdustinface:
Applies the same pattern as we have for deserialization to serialization. 

This avoids all those recursive runtime lookups for "how to stream this 
object" which brings a nice speedup:

```
compare: benchmark
to_bytes     | 453.72           | 176.06           | -61.19      

compare: full_block
to_bytes     | 108.72           | 52.32            | -51.87  
```

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -124,6 +124,7 @@ def recurse_jsonify(d):
     return d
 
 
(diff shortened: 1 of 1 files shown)


mariano54 (approved):
great speedup!

wjblanke:
wow

PR #10436, opened 2022-02-25
Stop running DL singleton puzzle in python

Quexington:


altendky (approved):

arvidn:
I'm pretty sure even `Program` actually runs with the rust interpreter, but it will need an additional serialization step first. So this is still an improvement.

PR #10437, opened 2022-02-25
Send tx updates

mariano54:
Questions:
* Need to decide what names to subscribe to
* send tx id or the whole transaction?


freddiecoleman:
The new transaction events currently contain `coin_solutions`. Is it possible for this to have `coin_spends` instead as that is the one I try to use going forwards as it is the non-deprecated one.

I believe that as this has never been sent over the websocket connection before it wouldn't be a breaking change to use `coin_spends` from the start.

mariano54:
Does this PR work for you @freddiecoleman and are you using it?

PR #10446, opened 2022-02-26
New RPC block_spends - Get spends for block using transaction generator

freddiecoleman:
Currently if you want to get all the spends for a block you have two options:

- Call the `get_additions_and_removals` RPC followed by potentially thousands of calls to `get_puzzle_and_solution` (one for each removal)
- Call the `get_block` endpoint and run the transaction generator. This is a bit better but can be slow depending on the environment in which your CLVM implementation is running.

By adding this RPC we can benefit from the speed of the Rust CLVM implementation and it also means less code for anybody who needs this data.


chia/rpc/full_node_rpc_api.py
```diff
@@ -397,6 +404,60 @@ async def get_block_records(self, request: Dict) -> Optional[Dict]:
...
 
+    async def get_block_spends(self, request: Dict) -> Optional[Dict]:
+        if "header_hash" not in request:
+            raise ValueError("No header_hash in request")
+        header_hash = bytes32.from_hexstr(request["header_hash"])
+        full_block: Optional[FullBlock] = await self.service.block_store.get_full_block(header_hash)
+        if full_block is None:
+            raise ValueError(f"Block {header_hash.hex()} not found")
+
+        ref_list = full_block.transactions_generator_ref_list
```
mariano54, line 415: type annotation

chia/rpc/full_node_rpc_api.py
```diff
@@ -397,6 +404,60 @@ async def get_block_records(self, request: Dict) -> Optional[Dict]:
...
+        tx_info = full_block.transactions_info
```
mariano54, line 416: for all of these variables

chia/rpc/full_node_rpc_api.py
```diff
@@ -397,6 +404,60 @@ async def get_block_records(self, request: Dict) -> Optional[Dict]:
...
+        generator_program = full_block.transactions_generator
+        height = full_block.reward_chain_block.height
+        spends: List[CoinSpend] = []
+        if tx_info and generator_program:
```
mariano54, line 420: `if tx_info is not None and generator_program is not None`

chia/rpc/full_node_rpc_api.py
```diff
@@ -397,6 +404,60 @@ async def get_block_records(self, request: Dict) -> Optional[Dict]:
...
+            cost = tx_info.cost
+            generator_args: List[SerializedProgram] = []
+            for height in ref_list:
```
mariano54, line 423: There is a method for this, `self.service.blockchain.get_block_generator`, so you can decrease the amount of code here

tests/core/test_full_node_rpc.py
```diff
@@ -134,6 +134,9 @@ def stop_node_cb():
             additions, removals = await client.get_additions_and_removals(blocks[-1].header_hash)
             assert len(additions) >= 2 and len(removals) == 0
 
+            block_spends = await client.get_block_spends(blocks[-1].header_hash)
```
mariano54, line 137: We should check that they are actually Spends

mariano54 (commented):
Thank you for making this PR! How is the performance?

chia/rpc/full_node_rpc_client.py
```diff
@@ -184,6 +184,13 @@ async def get_block_records(self, start: int, end: int) -> List:
         # TODO: return block records
         return response["block_records"]
 
+    async def get_block_spends(self, header_hash: bytes32) -> List:
+        try:
+            response = await self.fetch("get_block_spends", {"header_hash": header_hash.hex()})
+        except Exception:
```
mariano54, line 190: This should not catch, if it fails, we should return None or allow the caller to handle the exception

chia/rpc/full_node_rpc_client.py
```diff
@@ -184,6 +184,13 @@ async def get_block_records(self, start: int, end: int) -> List:
         # TODO: return block records
         return response["block_records"]
 
+    async def get_block_spends(self, header_hash: bytes32) -> List:
```
mariano54, line 187: If there is a chance of None, we should return Optional[List[CoinSpend]] otherwise List[CoinSpend]

chia/rpc/full_node_rpc_client.py
```diff
@@ -184,6 +184,13 @@ async def get_block_records(self, start: int, end: int) -> List:
...
+        try:
+            response = await self.fetch("get_block_spends", {"header_hash": header_hash.hex()})
+        except Exception:
+            return []
+        return response["block_spends"]
```
mariano54, line 192: Parse the objects.

mariano54 (changes requested):

chia/rpc/full_node_rpc_client.py
freddiecoleman: Opted to return null as some other RPC calls are doing that.

mariano54 (dismissed):
LGTM thanks for the PR!

chia/rpc/full_node_rpc_api.py
```diff
@@ -397,6 +405,48 @@ async def get_block_records(self, request: Dict) -> Optional[Dict]:
...
+            if block_generator is None:
+                return {"block_spends": spends}
+
+            if height >= DEFAULT_CONSTANTS.SOFT_FORK_HEIGHT:
+                # conditions must use integers in canonical encoding (i.e. no redundant
+                # leading zeros)
+                # the division operator may not be used with negative operands
+                flags = COND_CANON_INTS | NO_NEG_DIV
+            else:
+                flags = 0
```
arvidn, line 432: technically, since this doesn't do any validation, you can just assume that the block you find satisfies the consensus rules. So, I don't think you need to specify these flags.

chia/rpc/full_node_rpc_api.py
```diff
@@ -397,6 +405,48 @@ async def get_block_records(self, request: Dict) -> Optional[Dict]:
...
+
+            args = create_generator_args(block_generator.generator_refs).first()
+            _, block_result = block_generator.program.run_with_cost(
```
arvidn, line 435: `program` is a `SerializedProgram`, right?
Getting the generator arguments right here is non trivial. Are you confident the test covers this properly?

chia/rpc/full_node_rpc_api.py
```diff
@@ -397,6 +405,48 @@ async def get_block_records(self, request: Dict) -> Optional[Dict]:
...
 
+    async def get_block_spends(self, request: Dict) -> Optional[Dict]:
+        if "header_hash" not in request:
+            raise ValueError("No header_hash in request")
+        header_hash = bytes32.from_hexstr(request["header_hash"])
+        full_block: Optional[FullBlock] = await self.service.block_store.get_full_block(header_hash)
+        if full_block is None:
+            raise ValueError(f"Block {header_hash.hex()} not found")
+
+        tx_info: Optional[TransactionsInfo] = full_block.transactions_info
```
arvidn, line 416: as far as I can tell, the only reason you pull out `tx_info` is to get to the `cost` field. Since this block can be assumed to have been validated already, I don't think we need to do this. We could just use the `MAX_BLOCK_COST_CLVM` as the upper bound, but we already know the program won't exceed that.

tests/core/test_full_node_rpc.py
```diff
@@ -195,6 +195,10 @@ def stop_node_cb():
             assert len(await client.get_coin_records_by_puzzle_hash(ph, True, 0, blocks[-1].height + 1)) == 2
             assert len(await client.get_coin_records_by_puzzle_hash(ph, True, 0, 1)) == 0
 
+            block_spends = await client.get_block_spends((await full_node_api_1.get_all_full_blocks())[-1].header_hash)
+
+            assert block_spends == spend_bundle.coin_spends
```
arvidn, line 200: it's not obvious that `blocks[-1]` is a very interesting block to test. I would have expected this test for *every* block

arvidn (commented):
I think there are some opportunities for simplifications. I'm also not convinced the test covers an interesting case (just by looking at it)

wjblanke:
would like to get this in, please see Qs. thanks freddy!

freddiecoleman:
Currently think the test fails due to how the tests simulate farming blocks and how block_generator defaults to None here https://github.com/Chia-Network/chia-blockchain/blob/main/tests/block_tools.py#L1892

PR #10447, opened 2022-02-26
Fix balance calculation so it works with future unconfirmed tx

mariano54:
* remove already locked concept, no DB locks are involved
* Improve code quality
* Fix balance calculation, when you have future unconfirmed trades it would result in a negative value.


Quexington (approved):
Yeah this looks like it will do a more reliable job

PR #10454, opened 2022-02-27
Remove sys.exit() from chia daemon /exit endpoint

altendky:
Draft for:
- [x] Basic tests shutdown ok in Linux, need more testing/discussion/etc.


altendky:
I agree that tests passing in CI probably don't exercise that well.  I did do it a number of times locally without any failures to shutdown.  But yes, this is still draft due to needing more thought on how to test it.

Generally you should exit by stopping doing stuff and returning, not `sys.exit()`'ing down in some async callback.  The only question is what corner case was this self-labeled hack covering and is it still present two years later.

arvidn:
@richardkiss do you know what the reason for this `sys.exit()` was?

hoffmang9:
My recall is that @Yostra @richardkiss and I couldn't figure out what was hanging up so we put this in to ship and revert to finding the hang later. As such, if we're now able to shut down without it - that change should be low risk.

altendky:
It was an every time on every platform kind of thing?  As opposed to some super weird corner case?  Obviously not looking for pinky swears here, just collecting what info I can.

hoffmang9:
It was certainly on MacOS x86 and Ubuntu x86. I don't recall about Windows.

altendky:
Kick CI

aqk (approved):

altendky:
> @altendky did you try running a node and farmer with this patch locally? starting and stopping the daemon?

I did at least in Linux and noticed nothing particularly odd.  I think someone checked macOS for me, but it's been awhile.

arvidn (approved):

PR #10460, opened 2022-02-27
Fix `install-gui.sh`

ChiaMineJP:
This PR addersses 3 issues.

## 1. `install-gui.sh` fails on a system with NodeJS < 16 and npm >= 7

The GUI requirements are:
npm>=7 for handling package-lock.json in v2 format
NodeJS>=16 for newer syntax (e.g. nullish coalescing) for some of npm dependencies.

Current `install-gui.sh` only checks `npm>=7` and there is a slight chance that some user have both `npm>=7` and `NodeJS<16` installed on their system.
This PR fixes the failure in such environments.

## 2. `SUBMODULE_BRANCH` not working
When you run `install-gui.sh <ref/tag/branch of chia-blockchain-gui repos>`, it should install the specified commit of `chia-blockchain-gui` but currently it fails because when running `git checkout` on chia-blockchain-gui directory it goes Detached HEAD state thus following `git pull` cannot run.

## 3. Fails running `npm ci` in rare situations
`install-gui.sh` uses `n` as node/npm version switcher because it requires NodeJS>=16 and npm>=7.
However, due to `npm`'s PATH lookup mechanism, there are rare situations where old, incompatible version of `npm` can be used to run `npm install`.
When you run `npm run <script_name>`, npm internally spawns child process with modified `PATH` variable.
Imagine you run `npm run sample` in a situation below:
- Current working directory is `/home/GIT/chia-blockchain/chia-blockchain-gui`
- There is `/home/GIT/chia-blockchain/chia-blockchain-gui/package.json` and its content is:
  ```json
  {
    "name": "sample",
    "scripts": {
      "sample": "npm install"
    }
  }
  ```

When you run `npm run sample`, npm spawns child process to run `npm install`.
To run `npm install` as a separate process, it needs to lookup directories in `PATH` variable to find `npm` executable file.
The problem is, when the original npm spawns child process, it adds lookup directories to "head" of `PATH` like below:
- /home/GIT/chia-blockchain/chia-blockchain-gui/node_modules/.bin
- /home/GIT/chia-blockchain/node_modules/.bin
- /home/GIT/node_modules/.bin
- /home/node_modules/.bin
- /node_modules/.bin

Usually, `npm` executable is stored in `/usr/local/bin` or somewhere globally avaiable. But with the above PATH modification, if there is an old `npm` executable in the newly-added PATH directories, then that old `npm` is used to run `npm install`.
The situation sometimes happens when an user mistakenly ran `npm install npm` in home directory instead of `npm install -g npm` to upgrade `npm` version.

#### Solution for `3.`
By applying this PR, before executing `npm ci` in the `install-gui.sh`:
- it creates `node_modules/.bin` directory under `chia-blockchain` (This is ignored by git. See `.gitignore`)
- it creates a symbolic link of working `npm` to `/.../chia-blockchain/node_modules/.bin`
So even if such `PATH` modification occurs, `/home/GIT/chia-blockchain/node_modules/.bin/npm` is referred and runs without error.

### DIFF
--- a/install-gui.sh
+++ b/install-gui.sh
@@ -92,8 +92,8 @@ patch_inconsistent_npm_issue(){
   if [ ! -d "$node_module_dir" ]; then
     mkdir "$node_module_dir"
   fi
-  if [ ! -d "{$node_module_dir}/.bin" ]; then
-    mkdir "{$node_module_dir}/.bin"
+  if [ ! -d "${node_module_dir}/.bin" ]; then
+    mkdir "${node_module_dir}/.bin"
   fi
   if [ -e "${node_module_dir}/.bin/npm" ]; then
     rm -f "${node_module_dir}/.bin/npm"
--- a/install-gui.sh
+++ b/install-gui.sh
@@ -85,6 +85,22 @@ do_install_npm_locally(){
   fi
 }
 
+# Work around for inconsistent `npm` exec path issue
+# https://github.com/Chia-Network/chia-blockchain/pull/10460#issuecomment-1054492495
+patch_inconsistent_npm_issue(){
+  node_module_dir=$1
+  if [ ! -d "$node_module_dir" ]; then
+    mkdir "$node_module_dir"
+  fi
+  if [ ! -d "{$node_module_dir}/.bin" ]; then
+    mkdir "{$node_module_dir}/.bin"
+  fi
+  if [ -e "${node_module_dir}/.bin/npm" ]; then
+    rm -f "${node_module_dir}/.bin/npm"
+  fi
+  ln -s "$(command -v npm)" "${node_module_dir}/.bin/npm"
+}
+
 # Manage npm and other install requirements on an OS specific basis
 if [ "$(uname)" = "Linux" ]; then
   #LINUX=1
@@ -175,9 +191,15 @@ if [ ! "$CI" ]; then
     echo ""
   fi
 
+  # Work around for inconsistent `npm` exec path issue
+  # https://github.com/Chia-Network/chia-blockchain/pull/10460#issuecomment-1054492495
+  patch_inconsistent_npm_issue "../node_modules"
+
   npm ci
   npm audit fix || true
   npm run build
+
+  # Set modified output of `chia version` to version property of GUI's package.json
   python ../installhelper.py
 else
   echo "Skipping node.js in install.sh on MacOS ci."
--- a/install-gui.sh
+++ b/install-gui.sh
@@ -155,9 +155,8 @@ if [ ! "$CI" ]; then
 
   if [ "$SUBMODULE_BRANCH" ];
   then
-    git fetch
-    git checkout "$SUBMODULE_BRANCH"
-    git pull
+    git fetch --all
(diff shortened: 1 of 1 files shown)


hoffmang9 (dismissed):
lgtm

ChiaMineJP:
I heard @xdustinface encountered installation error on MacOS
```
> root@1.2.10-dev132 bootstrap /xxxxx/chia-blockchain/chia-blockchain-gui
> lerna bootstrap
 
lerna notice cli v4.0.0
lerna info Bootstrapping 6 packages
lerna info Installing external dependencies
lerna ERR! npm install exited 1 in '@chia/api-react'
lerna ERR! npm install stderr:
npm WARN npm npm does not support Node.js v16.14.0
npm WARN npm You should probably upgrade to a newer version of node as we
npm WARN npm can't make any promises that npm will work with this version.
npm WARN npm Supported releases of Node.js are the latest release of 6, 8, 9, 10, 11.
npm WARN npm You can find the latest version at https://nodejs.org/
npm ERR! cb.apply is not a function
``` 

I investigated the cause and found possible path to this error.
The facts:
- The error was resolved by removing `~/node_modules`
- `~/node_modules` looks like npm's global installation folder.
- In `install-gui.sh`, I use `n` as a nodejs version switcher and added `bin` folder which contains `n`'s `npm` to `$PATH` variable. And in the script, `n`'s nodejs/npm should be used entirely to build npm dependencies for the GUI.
- The global installation folder for `n`'s `npm` should be `chia-blockchain/.n/lib/node_modules`
- So in theory, `npm WARN npm npm does not support Node.js v16.14.0` should not happen if `n`'s `npm` is actually used.
- But somehow non-`n` version of `npm` is used in `lerna bootstrap` execution.
- I investigated codes below:
  - https://github.com/lerna/lerna/
  - https://github.com/lerna/lerna/blob/main/commands/bootstrap/index.js
  - https://github.com/lerna/lerna/blob/main/utils/npm-install/npm-install.js
  - https://github.com/lerna/lerna/blob/main/utils/get-npm-exec-opts/get-npm-exec-opts.js
  - https://github.com/lerna/lerna/blob/main/core/child-process/index.js
  - https://github.com/sindresorhus/execa#extendenv
  - https://github.com/sindresorhus/execa/blob/main/index.js
- I found one persuasive cause.
  - `lerna bootstrap` internally calls `npm install`
  - To execute `npm install` , `lerna` tries to spawn child process.
  - The child process inherits parent's `process.env` but not `$PATH` shell variable.
  - `install-gui.sh` just adds `bin` containing `n`'s `npm`/`node` to `$PATH` variable, and it never touches `process.env`.
  - I suspect that users who globally installed old `npm` by `npm install -g npm` may encounter this error, since PATH in `process.env` seems to have precedence over `$PATH` shell variable.

ChiaMineJP:
@hoffmang9
I updated the description of this PR.
https://github.com/Chia-Network/chia-blockchain/pull/10460#issue-1153240817
Added a fix for a rare issue @xdustinface experienced.

hoffmang9 (approved):
lgtm

PR #10470, opened 2022-02-28
full_node: Drop ses and height-hash caches if their genesis is invalid

xdustinface:
The cache should always match the database, i think just checking the genesis on startup and drop/rebuild if it doesn't match should be good enough here. I had a full node being confused and trying to sync from scratch (even though the db was fully synced) because it was on testnet10 while the cache was from mainnet.

### DIFF
--- a/chia/full_node/block_height_map.py
+++ b/chia/full_node/block_height_map.py
@@ -57,6 +57,7 @@ class BlockHeightMap:
         self.__height_to_hash_filename = blockchain_dir / "height-to-hash"
         self.__ses_filename = blockchain_dir / "sub-epoch-summaries"
 
+        genesis_hash: Optional[bytes32] = None
         if db.db_version == 2:
             async with self.db.db.execute("SELECT hash FROM current_peak WHERE key = 0") as cursor:
                 peak_row = await cursor.fetchone()
@@ -70,6 +71,11 @@ class BlockHeightMap:
                 row = await cursor.fetchone()
                 if row is None:
                     return self
+
+            async with db.db.execute("SELECT header_hash FROM full_blocks WHERE height=0") as cursor:
+                genesis_row = await cursor.fetchone()
+                if genesis_row is not None:
+                    genesis_hash = bytes32(genesis_row[0])
         else:
             async with await db.db.execute(
                 "SELECT header_hash,prev_hash,height,sub_epoch_summary from block_records WHERE is_peak=1"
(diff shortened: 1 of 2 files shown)


xdustinface:
@arvidn Any concerns about reduced privacy like in [12d32d7](https://github.com/Chia-Network/chia-blockchain/pull/10470/commits/12d32d76e3954da49f9813c2c111cf67e4b97d5b) to make things better testable?

arvidn:
I think it's problematic to do this. There is already logic to fix a mismatching height-to-hash cache, and this obfuscates that issue. Instead, we should fix the logic that is supposed to detect and re-write the cache.

When [_load_blocks_from()](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/block_height_map.py#L155) is called on startup, it verifies the hashes in the cache against the database. It starts at the peak and moves backwards. Any mismatching hash is updated, and it keeps going until it finds a match (which implies that all hashes below that height are good). There's some special logic to do the same for the sub epoch summaries, which are sparse. There isn't one at every height, so we over shoot a bit to check those too.

It would be better to understand why this logic doesn't fix up the height-to-hash cache, because if it doesn't work for the simple case of switching to testnet, there's a risk there are other cases it also doesn't work for.

arvidn:
> Any concerns about reduced privacy like in [12d32d7](https://github.com/Chia-Network/chia-blockchain/pull/10470/commits/12d32d76e3954da49f9813c2c111cf67e4b97d5b) to make things better testable?

Making tests poke around with implementation details of the code that's supposed to be tested is risky. The main risk is that implementation details change, the test no longer works, but it still passes. So, when relying on implementation details in tests, it's important to have lots of negative tests as well, to carve out testing exactly the behavior that's expected.

Also, even with double underscores it's possible to refer to the field from the outside, you just have to prefix the class name. In a way I think that makes sense, to make it even more obvious for the reader that something fishy is going on.

xdustinface:
> I think it's problematic to do this. There is already logic to fix a mismatching height-to-hash cache, and this obfuscates that issue. Instead, we should fix the logic that is supposed to detect and re-write the cache.

Agree, if the code is anyway supposed to fix the cache its better to fix the cache fixing :) Closing here then. 


> Also, even with double underscores it's possible to refer to the field from the outside, you just have to prefix the class name. In a way I think that makes sense, to make it even more obvious for the reader that something fishy is going on.

Yeah it's possible, but when i look at the tests in that commit it's still rather ugly if you still want to access some internal members for testing (apart from if you like doing that or not). Since its anyway always possible to access those fields its not clear to my why the double underscore helps much here. We could just have a linter restricting access to `_` members if thats a reason why you introduced `__`.

PR #10471, opened 2022-02-28
Singleton 1.1 optimization

Quexington:
This optimization provides about 2 million cost savings on a singleton so about 10%.


richardkiss (dismissed):
I'm a little hazy on the details here, but I think I get it. We only need to verify the parent id and amount, so we skip calculating "my" coin id.

Quexington:
I added another optimization after the original one for this PR, so it will need re-review.

richardkiss (approved):
Very nice, looks good.

PR #10486, opened 2022-03-01
fix updating of sub-epoch-summary map

arvidn:
(part of BlockHeightMap) when the entirey change, including genesis changes


xdustinface (approved):
Works 👍

PR #10487, opened 2022-03-01
tests: Add `_PYTEST_RAISE` to fix exception breakpoints with pytest

xdustinface:
It's currently not possible to have the debuger stop on an uncaught exception when debugging tests. With this patch, adding `_PYTEST_RAISE=1` to the environment variables in the pytest configuration template fixes this.

image

To enable the breakpoints: 

image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import multiprocessing
+import os
 import pytest
 import pytest_asyncio
 import tempfile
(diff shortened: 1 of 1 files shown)


arvidn (approved):
It's quite unclear to me how this causes an exception that's clearly caught/handled to be considered unhandled. I could see the `raise` calls you added here being considered unhandled though, but presumably your breakpoint will stop at the first exception

PR #10495, opened 2022-03-01
add select_coins RPC method

ftruzzi:
Hi!

This PR exposes the `select_coins` method to the RPC client. I think it can be useful for smart coin developers. For example, if you had to contribute to a piggybank coin you would need to select one or more contribution coins in order to build a custom spend bundle to aggregate with the piggybank spend.

Please let me know what you think and if you need any tests.

Thanks :)


Quexington (commented):
Great change! Can you add a test for it somewhere in `tests/wallet/rpc/test_wallet_rpc.py`?

chia/rpc/wallet_rpc_api.py
```diff
@@ -810,6 +811,28 @@ async def delete_unconfirmed_transactions(self, request):
...
+        if await self.service.wallet_state_manager.synced() is False:
+            raise ValueError("Wallet needs to be fully synced before selecting coins")
+
+        amount = request["amount"]
+        if not isinstance(amount, uint64):
+            amount = uint64(amount)
+
+        wallet_id = request["wallet_id"]
+        if not isinstance(wallet_id, uint32):
+            wallet_id = uint32(wallet_id)
```
Quexington, line 826: You can just do `wallet_id = uint32(request["wallet_id"])` here right?

Same with the amount above.

ftruzzi:
Thanks @Quexington! Mind taking another look?

Unfortunately I'm getting errors trying to run the tests on my laptop so I've added only two extremely simple ones, let's see if they pass (do I need run approval on here?)

Quexington:
Got this error in the CI run for the test: `TypeError: select_coins() takes 1 positional argument but 3 were given`

ftruzzi:
@Quexington whoops, looks like I had forgot about the keyword-only arguments, it should be fixed now. I figured I would use kw-only arguments since one would not want to unknowingly swap `amount` with `wallet_id`.

I also let the tests run locally for more time and the errors I was seeing were debug stuff caused by having used `pytest -s` 🥳

Quexington (approved):
Great addition thanks!

ftruzzi:
nice! @Quexington can you also merge it or are we waiting for something else to go in first?

Quexington:
Main is a little locked down right now but I've put in on the radar to merge.

roseiliend:
Hi, @Quexington . if the developers use the coins, the wallet cant not lock the coin instantly, and the new transaction which include the same coins from client may be faild.  How do you think about it?

ftruzzi:
@roseliend I think that's kind of expected? The way I thought about this was to select coins in order to use them in a non-standard way / with custom spend bundles, not with standard transactions.

So you would select coins, craft a spend bundle and push it out shortly after (before another transaction which might want to use the same coins).

And if you want to select coins with standard transactions, your best bet is probably the `create_standard_transaction` RPC method which includes a `coin` argument.

Anyway, a workaround for having multiple "in-progress" transactions could be:

1. You `select_coins` for TX 1
2. While you're building the spend bundle for TX 1, you need to send TX 2 which might end up selecting the same coins.
3. You `select_coins` for TX 2, passing in the coins selected for TX 1 as coins you want to exclude
4. You use the `create_signed_transaction` RPC method with coins selected in step (3).
5. You can safely push out both spend bundles without worrying about coin conflicts.

I would be open to adding an `exclude` argument but the issue is that, as of now, the CAT wallet version of `select_coins` does not support excluding coins.

Might be a discussion worth having in https://github.com/Chia-Network/chia-blockchain/pull/9975?

roseiliend:
@ftruzzi thanks to your reply . add `exclude ` argument may be a good idea

PR #10497, opened 2022-03-01
Fix misleading comment in select_coins

ojura:
We're selecting the largest coins first, not the oldest.


ojura:
Obsoleted by #9975

PR #10498, opened 2022-03-01
when creating a new blockchain database implicitly, make it v2

arvidn:
This patch makes three changes:

1. `chia init` always creates a database file that contains a version number, even with `--v1`
2. When the full node starts up and opens the database file, if it's an empty database (i.e. no `full_blocks` table) it will attempt to set it to be version 2. So, in case the user runs `chia init` and then deletes the database file, the new database file will still be v2.
3. On the off-chance that the user is running `chia init` a second time (after deleting `config.yaml`) with an existing database file, the database file won't be touched.

### DIFF
--- a/chia/cmds/init_funcs.py
+++ b/chia/cmds/init_funcs.py
@@ -1,5 +1,6 @@
 import os
 import shutil
+import sqlite3
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple
 
@@ -23,6 +24,7 @@ from chia.util.config import (
     unflatten_properties,
     get_config_lock,
 )
+from chia.util.db_version import set_db_version
 from chia.util.keychain import Keychain
 from chia.util.path import mkdir, path_from_root
 from chia.util.ssl_check import (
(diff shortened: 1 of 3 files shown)


hoffmang9 (dismissed):
lgtm

emlowe (commented):
If we make this change, isn't the change in chia_init no longer required. By default anytime we create the DB we will use the v2 format.
And this does mean that the filename will possibly be even more confusing since I could do chia init --v1-db to get the v1 DB, then delete my DB file, and restart node, which now results in V2 DB format.... seems like more confusion likely?

arvidn:
yeah, setting v2 in `chia init` would be redundant with this, as far as I can tell.
I don't see how it affects the confusion around the filename though. The default database filename is `_v2_` right now.

Would you expect the version of the database that's created would be affected by what you name the file?

arvidn:
actually, there's an issue with this patch. If you do `chia init --v1-db`, the logic I add in this patch might override the attempt to set it to a v1 database

emlowe (dismissed):
Ok for me - there are still some cases where users specifically asked for v1 and then delete the DB which will then be recreated with v2 - but this is not going to be common as people are quite happy to be on v2. There are probably also some edge cases where people delete config.yaml but do not delete their DB, or copy the DB around.

arvidn:
yeah. deleting config would be fine though, because the default name of the file is `v2`.

arvidn:
something is wrong. the `get_config_lock()` hangs. I'm looking into it

emlowe (approved):
ok

PR #10499, opened 2022-03-01
run tests in parallel in CI

arvidn:


arvidn:
this makes the tests run with `-n auto` by default. On CI, most workflows are run with `-n 3` (which is one more than the 2 cores GHA runners typically have). Some tests that are still flaky under parallel runs are left as `-n 0` (which is to cancel out the default `-n auto`).

mariano54 (commented):
Awesome! is there any benefit to running with -n higher like 10? I'm guessing there are IO and other things which can cause delays in tests

mariano54:
For reference (comparing old to `-n 3`): 
Mac OS blockchain  45 -> 26
Ubuntu blockchain  47 -> 25
Mac OS core: 12 -> 12

The improvements are significant. On the other hand, pooling tests which still don't run in parallel, no improvement (currently at 35 mins). This change is really great and we should merge asap

mariano54 (approved):

Issue #10478, opened 2022-02-28
[Bug] Wallet v2 db forces payout_instructions address to STANDARD_WALLET address


wallentx:
### What happened?

On first run of chia wallet service in beta3, or on sync from 0, the chia wallet service sets the `payout_instructions:` entry in `config.yaml` to be that of a receive address tied to the STANDARD_WALLET.

I have my rewards for pooling configured to send to a cold wallet address. I set this in `config.yaml` at `payout_instructions:`, and `chia plotnft show` verifies my desired address at `Payout instructions (pool will pay you with this):`.

To replicate this, simply `chia stop all -d` and delete the contents of `~/.chia/mainnet/wallet/db/` and then `chia start wallet`. This will replace my `payout_instructions:` entry in `config.yaml`. This behavior also happens if updating from a working `1.2.11` -> `beta3`

### Version

1.2.12.dev376

### What platform are you using?

Linux

### What ui mode are you using?

CLI

paninaro:
```-> main()
  /Users/jeffc/src/macos/chia-blockchain/chia/server/start_wallet.py(94)main()
-> return run_service(**kwargs)
  /Users/jeffc/src/macos/chia-blockchain/chia/server/start_service.py(278)run_service()
-> return asyncio.run(async_run_service(*args, **kwargs))
  /usr/local/Cellar/python@3.9/3.9.7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py(44)run()
-> return loop.run_until_complete(main)
  /usr/local/Cellar/python@3.9/3.9.7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py(629)run_until_complete()
-> self.run_forever()
  /usr/local/Cellar/python@3.9/3.9.7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py(596)run_forever()
-> self._run_once()
  /usr/local/Cellar/python@3.9/3.9.7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py(1890)_run_once()
-> handle._run()
  /usr/local/Cellar/python@3.9/3.9.7/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/events.py(80)_run()
-> self._context.run(self._callback, *self._args)
[...]

PR #10507, opened 2022-03-01
Preserve existing pool payout_instructions when creating a PoolWallet

paninaro:
When syncing a wallet db from scratch, if a pool nft launcher is found, we were previously overwriting the pool config's `payout_instructions` with a new puzzle hash. This change will preserve the existing `payout_instruction` puzzle hash, if found in the config.

Issue #10478

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -237,7 +237,7 @@ class PoolWallet:
         payout_instructions: str = existing_config.payout_instructions if existing_config is not None else ""
 
         if len(payout_instructions) == 0:
-            self.log.info(f"Missing payout_instructions for {current_state.launcher_id}. Generating new puzzle hash...")
+            self.log.info("New config entry, generating new puzzle hash...")
             payout_instructions = (await self.standard_wallet.get_new_puzzlehash(in_transaction=True)).hex()
             self.log.info(f"Generated payout_instructions puzzle hash: {payout_instructions}")
 
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -229,16 +229,17 @@ class PoolWallet:
     async def get_tip(self) -> Tuple[uint32, CoinSpend]:
         return self.wallet_state_manager.pool_store.get_spends_for_wallet(self.wallet_id)[-1]
 
-    async def update_pool_config(self, make_new_authentication_key: bool):
+    async def update_pool_config(self):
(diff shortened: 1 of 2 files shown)


paninaro:
If this approach looks good, I'll add a couple of tests to cover the change.

chia/pools/pool_wallet.py
```diff
@@ -229,16 +229,17 @@ async def get_unconfirmed_transactions(self) -> List[TransactionRecord]:
...
         pool_config_dict: Dict[bytes32, PoolWalletConfig] = {c.launcher_id: c for c in pool_config_list}
         existing_config: Optional[PoolWalletConfig] = pool_config_dict.get(current_state.launcher_id, None)
+        payout_instructions: str = existing_config.payout_instructions if existing_config is not None else ""
 
-        if make_new_authentication_key or existing_config is None:
-            payout_instructions: str = (await self.standard_wallet.get_new_puzzlehash(in_transaction=True)).hex()
-        else:
-            payout_instructions = existing_config.payout_instructions
+        if len(payout_instructions) == 0:
+            self.log.info(f"Missing payout_instructions for {current_state.launcher_id}. Generating new puzzle hash...")
```
mariano54, line 240: This sounds like an error but it's the normal scenario when syncing up on a new machine. Can we change it to: "New config entry, generating...."

chia/pools/pool_wallet.py
```diff
@@ -229,16 +229,17 @@ async def get_unconfirmed_transactions(self) -> List[TransactionRecord]:
...
+        payout_instructions: str = existing_config.payout_instructions if existing_config is not None else ""
 
-        if make_new_authentication_key or existing_config is None:
-            payout_instructions: str = (await self.standard_wallet.get_new_puzzlehash(in_transaction=True)).hex()
-        else:
-            payout_instructions = existing_config.payout_instructions
+        if len(payout_instructions) == 0:
+            self.log.info("New config entry, generating new puzzle hash...")
+            payout_instructions = (await self.standard_wallet.get_new_puzzlehash(in_transaction=True)).hex()
+            self.log.info(f"Generated payout_instructions puzzle hash: {payout_instructions}")
```
emlowe, line 242: I'd probably prefer this all on one line in the log

paninaro: image

paninaro: As you wish

mariano54 (approved):

PR #10508, opened 2022-03-01
Atari merge main ca84fa9f8d332687534399fc824d6017effaed2e

altendky:
The only interesting conflict was on `WalletNode.complete_blocks()` which `atari` had significantly modified but was deleted in `main`.  The deletion was accepted and the changes thrown out.

Follow up steps:
- Remove unused import of `DataLayerWallet` for flake8
- Use `self.wallet_state_manager.wallet_node.get_coin_state()` (add `.wallet_node`)


Quexington (approved):
If the tests pass then I see nothing out of place :+1:

PR #10516, opened 2022-03-02
Ak.test daemon simulation

aqk:
Test branch to run CI tests

### DIFF
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -11,8 +11,7 @@ from keyring.backends.Windows import WinVaultKeyring as WinKeyring
 from keyring.errors import KeyringError, PasswordDeleteError
 from pathlib import Path
 from sys import exit, platform
-from typing import Any, List, Optional, Tuple, Type, Union
-
+from typing import Any, List, Optional, Tuple, Type, Union, TypeVar
 
 # We want to protect the keyring, even if a user-specified master passphrase isn't provided
 #
@@ -82,6 +81,9 @@ def warn_if_macos_errSecInteractionNotAllowed(error: KeyringError) -> bool:
     return False
 
 
+_T_KeyringWrapper = TypeVar("_T_KeyringWrapper", bound="KeyringWrapper")
+
+
 class KeyringWrapper:
     """
     KeyringWrapper provides an abstraction that the Keychain class can use
@@ -104,24 +106,37 @@ class KeyringWrapper:
     cached_passphrase_is_validated: bool = False
     legacy_keyring = None
 
-    def __init__(self, keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH, force_legacy: bool = False):
+    @classmethod
+    def create(
+        cls: Type[_T_KeyringWrapper], keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH, force_legacy: bool = False
+    ) -> Optional[_T_KeyringWrapper]:
         """
-        Initializes the keyring backend based on the OS. For Linux, we previously
-        used CryptFileKeyring. We now use our own FileKeyring backend and migrate
-        the data from the legacy CryptFileKeyring (on write).
+        Create a new KeyringWrapper instance.
         """
-        self.keys_root_path = keys_root_path
+        # instance: KeyringWrapper = KeyringWrapper()
+        instance = cls()
+        instance.keys_root_path = keys_root_path
         if force_legacy:
             legacy_keyring = get_legacy_keyring_instance()
             if check_legacy_keyring_keys_present(legacy_keyring):
-                self.keyring = legacy_keyring
+                instance.keyring = legacy_keyring
             else:
                 return None
         else:
-            self.refresh_keyrings()
+            instance.refresh_keyrings(keys_root_path)
+        return instance
 
-    def refresh_keyrings(self):
-        self.keyring = None
+    def refresh_keyrings(self, keys_root_path: Optional[Path]) -> None:
+        """
+        Initializes or updates the keyring backend based on the OS. For Linux, we
+        previously used CryptFileKeyring. We now use our own FileKeyring backend
+        and migrate the data from the legacy CryptFileKeyring (on write).
+        """
+
+        if keys_root_path is not None:
+            self.keys_root_path = keys_root_path
+
+        self.cleanup_keyring()
         self.keyring = self._configure_backend()
 
         # Configure the legacy keyring if keyring passphrases are supported to support migration (if necessary)
@@ -187,17 +202,29 @@ class KeyringWrapper:
     @staticmethod
     def get_shared_instance(create_if_necessary=True):
         if not KeyringWrapper.__shared_instance and create_if_necessary:
-            KeyringWrapper.__shared_instance = KeyringWrapper(keys_root_path=KeyringWrapper.__keys_root_path)
+            KeyringWrapper.__shared_instance = KeyringWrapper.create(keys_root_path=KeyringWrapper.__keys_root_path)
 
         return KeyringWrapper.__shared_instance
 
     @staticmethod
-    def cleanup_shared_instance():
-        KeyringWrapper.__shared_instance = None
+    def set_shared_instance(instance: "KeyringWrapper"):
+        KeyringWrapper.__shared_instance = instance
 
     @staticmethod
-    def get_legacy_instance() -> Optional["KeyringWrapper"]:
-        return KeyringWrapper(force_legacy=True)
+    def cleanup_shared_instance():
+        if KeyringWrapper.__shared_instance is not None:
+            KeyringWrapper.__shared_instance.cleanup_keyring()
+            KeyringWrapper.__shared_instance = None
+
+    def cleanup_keyring(self):
+        filekeyring: Optional[FileKeyring] = self.keyring if type(self.keyring) == FileKeyring else None
+        if filekeyring is not None:
+            filekeyring.cleanup_keyring_file_watcher()
+        self.keyring = None
+
+    @staticmethod
+    def get_legacy_instance() -> Optional[_T_KeyringWrapper]:
+        return KeyringWrapper.create(force_legacy=True)
 
     def get_keyring(self):
         """
@@ -423,7 +450,8 @@ class KeyringWrapper:
             if user is not None:
                 passphrase = self.get_passphrase(service, user)
 
-            if passphrase is not None:
+            # if passphrase and user:
+            if passphrase:
                 user_passphrase_pairs.append((user, passphrase))
 
             index += 1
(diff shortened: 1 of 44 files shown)


tests/setup_nodes.py
```diff
@@ -583,7 +600,7 @@ async def setup_full_system(
     shared_b_tools: BlockTools,
     b_tools: BlockTools = None,
     b_tools_1: BlockTools = None,
-    connect_to_daemon=False,
```
aqk: This was a bool that meant "read the daemon_port from the config for this BlockTools, and attempt to connect the services to it"

tests/core/daemon/test_daemon.py
```diff
@@ -69,16 +82,17 @@ async def num_connections():
 
         await time_out_assert_custom_interval(60, 1, num_connections, 1)
 
-        await time_out_assert(1500, node_height_at_least, True, node2, 1)
+        # await time_out_assert(60, node_height_at_least, True, node2, 1)
```
aqk, line 85: @mariano54 Is this time out to wait until the system is up, or to check that we are advancing the chain?

tests/core/daemon/test_daemon.py
```diff
@@ -57,8 +52,26 @@ async def get_daemon_with_temp_keyring(self, get_b_tools):
         async for _ in setup_daemon(btools=get_b_tools):
             yield get_b_tools
 
+    # TODO: Ideally, the db_version should be the (parameterized) db_version
+    # fixture, to test all versions of the database schema. This doesn't work
+    # because of a hack in shutting down the full node, which means you cannot run
+    # more than one simulations per process.
+    @pytest_asyncio.fixture(scope="function")
```
aqk, line 59: This is just moved below the fixtures it uses.

tests/core/daemon/test_daemon.py
```diff
@@ -57,8 +52,26 @@ async def get_daemon_with_temp_keyring(self, get_b_tools):
...
+            b_tools=get_b_tools,
+            b_tools_1=get_b_tools_1,
+            connect_to_daemon_port=get_b_tools._config["daemon_port"],
+            db_version=1,
+        ):
+            yield _
+
     @pytest.mark.asyncio
-    async def test_daemon_simulation(self, simulation, get_daemon, get_b_tools, self_hostname):
+    async def test_daemon_simulation(self, simulation, bt, get_b_tools, get_b_tools_1):
```
aqk, line 74: `test_daemon_simulation`: what is the goal of this test? What minimum configuration of nodes / processes is required? (Are we simulating 1 node running all services, and one node running ... ?)

tests/setup_nodes.py
```diff
@@ -650,19 +667,20 @@ async def setup_full_system(
...
                 db_version=db_version,
             ),
             setup_vdf_client(shared_b_tools, shared_b_tools.config["self_hostname"], vdf2_port),
             setup_timelord(timelord1_port, 1000, timelord1_rpc_port, vdf2_port, True, consensus_constants, b_tools_1),
         ]
 
-        if connect_to_daemon:
+        if connect_to_daemon_port:
             node_iters.append(setup_daemon(btools=b_tools))
+            daemon1 = await node_iters[9].__anext__()
```
aqk, line 683: This moves the daemon to be created first

tests/core/daemon/test_daemon.py
mariano54: it waits for advancing the chain, checking that node2 has at least 2 blocks (height 1).

tests/core/daemon/test_daemon.py
mariano54: Not sure, I didn't write this one

tests/setup_nodes.py
```diff
@@ -636,11 +653,11 @@ async def setup_full_system(
                 node1_port,
                 rpc1_port,
                 b_tools,
+                connect_to_daemon_port,
```
arvidn, line 656: it's not clear to me why you need to pass in this port here. You're already passing in `b_tools`, which knows which port the daemon runs on. You could just ask it.

Unless there's some shenanigans where you're actually connecting to some *other* BlockTool's daemon.. but in that case, I would think there would be some comments in the code making the reader aware of that.

tests/setup_nodes.py
arvidn: I wouldn't expect the *order* they are created to matter. Assuming they've all been configured with an available listen port already.

tests/setup_nodes.py
aqk: That's true. I shouldn't need to both pass the port AND have it in the config. I did this to make it explicit to the caller that it needs to be set

arvidn: one of my changes (that weren't in when you started this patch, but has landed since) made all BlockTools instances always pick a free port and configure the daemon with it. So, in `main` it should always be set.

chia/util/keyring_wrapper.py
```diff
@@ -187,17 +202,25 @@ def set_keys_root_path(keys_root_path: Path):
...
+    def cleanup_keyring(self):
+        filekeyring: Optional[FileKeyring] = self.keyring if type(self.keyring) == FileKeyring else None
+        if filekeyring is not None:
+            filekeyring.cleanup_keyring_file_watcher()
+        self.keyring = None
 
     @staticmethod
-    def get_legacy_instance() -> Optional["KeyringWrapper"]:
-        return KeyringWrapper(force_legacy=True)
+    def get_legacy_instance():
```
paninaro, line 222: Can we add back the return type for this method? Maybe this is another case where TypeVar is appropriate. `get_shared_instance()` should probably specify the return type as well (though I understand if we want to skip that for now to land this PR).

tests/core/full_node/test_mempool.py
```diff
@@ -1406,7 +1422,7 @@ async def test_stealing_fee(self, two_nodes):
         )
 
         full_node_1, full_node_2, server_1, server_2 = two_nodes
-        peer = await connect_and_get_peer(server_1, server_2)
+        peer = await connect_and_get_peer(server_1, server_2, bt.config["self_hostname"])
```
paninaro, line 1425: Should this use the `self_hostname` fixture?

paninaro: Looks like it would apply to 9 occurrences in this file.

tests/core/full_node/test_mempool.py
```diff
@@ -1449,15 +1465,33 @@ async def test_stealing_fee(self, two_nodes):
         assert status == MempoolInclusionStatus.FAILED
 
     @pytest.mark.asyncio
-    async def test_double_spend_same_bundle(self, two_nodes):
+    async def test_double_spend_same_bundle(self, bt, two_nodes, wallet_a):
```
paninaro, line 1468: This test looks substantially different than the prior implementation. Just curious what led to the changes.

tests/setup_nodes.py
```diff
@@ -33,10 +36,16 @@ def cleanup_keyring(keyring: TempKeyring):
     keyring.cleanup()
 
 
-temp_keyring = TempKeyring()
+temp_keyring = TempKeyring(populate=True)
 keychain = temp_keyring.get_keychain()
 atexit.register(cleanup_keyring, temp_keyring)  # Attempt to cleanup the temp keychain
```
paninaro, line 41: Shouldn't be necessary anymore, but I understand if you don't want to make that change with this PR

tests/setup_nodes.py
```diff
@@ -98,6 +110,14 @@ async def setup_full_node(
                 connection.commit()
 
     config = local_bt.config["full_node"]
+
+    connect_to_daemon = False
+    if connect_to_daemon_port:
```
paninaro, line 115: Minor nit:

```suggestion
    if connect_to_daemon_port is not None:
```

tests/setup_nodes.py
```diff
@@ -476,15 +515,18 @@ async def setup_simulators_and_wallets(
             port = find_available_listen_port(f"node{index}")
             rpc_port = find_available_listen_port(f"node{index} rpc")
             db_name = f"blockchain_test_{port}.db"
+            # TODO convert this bt_tools variable to a passed-in variable
```
paninaro, line 518: Does this still apply?

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -136,9 +136,9 @@ def stop_node_cb():
         await time_out_assert(5, wallet.get_confirmed_balance, initial_funds)
         await time_out_assert(5, wallet.get_unconfirmed_balance, initial_funds)
 
-        client = await WalletRpcClient.create(self_hostname, test_rpc_port, bt.root_path, config)
-        client_2 = await WalletRpcClient.create(self_hostname, test_rpc_port_2, bt.root_path, config)
-        client_node = await FullNodeRpcClient.create(self_hostname, test_rpc_port_node, bt.root_path, config)
+        client = await WalletRpcClient.create(hostname, test_rpc_port, bt.root_path, config)
+        client_2 = await WalletRpcClient.create(hostname, test_rpc_port_2, bt.root_path, config)
+        client_node = await FullNodeRpcClient.create(hostname, test_rpc_port_node, bt.root_path, config)
```
paninaro, line 141: Minor: seems like these changes could be reverted to just use the `self_hostname` fixture that's passed in. But I also see that `hostname` (which should be the same as `self_hostname`) is used above, so :shrug:

tests/setup_nodes.py
aqk: No, just hoped to speed connection up. Not sure what the poll interval is

chia/util/keyring_wrapper.py
aqk: Yes, this looks like a merge mistake

tests/core/full_node/test_mempool.py
aqk: I'll take a look. Mindnumbing number of changes, plus conflicts with an earlier rebase, thanks.

tests/core/full_node/test_mempool.py
aqk: I just did this for convenience becase we already have bt. `self_hostname` depends on `bt`, so it is guaranteed to be the same

tests/setup_nodes.py
aqk: will remove - thanks!

tests/wallet/rpc/test_wallet_rpc.py
aqk: Yeah, those two are equivalent

tests/core/full_node/test_mempool.py
aqk: Thank you. This looks like a rebase snafu. `test_agg_sig_condition` as well. Looking into it

chia/util/keyring_wrapper.py
altendky: You can't `TypeVar` without a parameter to relate to the return.  Will have to switch to a `@classmethod` and use `cls`.

PR #10522, opened 2022-03-02
Only rewrite config when there is a difference

mariano54:
This should fix the infinite loop of write, read, write, as well as the constant writes once per second.

### DIFF
--- a/chia/pools/pool_config.py
+++ b/chia/pools/pool_config.py
@@ -73,7 +73,7 @@ def add_auth_key(root_path: Path, config_entry: PoolWalletConfig, auth_key: G1El
                 ):
                     auth_key_hex = bytes(auth_key).hex()
                     if pool_config_dict.get("authentication_public_key", "") != auth_key_hex:
-                        pool_config_dict["authentication_public_key"] = bytes(auth_key).hex()
+                        pool_config_dict["authentication_public_key"] = auth_key_hex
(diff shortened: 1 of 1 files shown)


chia/pools/pool_config.py
```diff
@@ -63,18 +63,23 @@ def load_pool_config(root_path: Path) -> List[PoolWalletConfig]:
...
     if pool_list is not None:
         for pool_config_dict in pool_list:
             try:
                 if (
                     G1Element.from_bytes(hexstr_to_bytes(pool_config_dict["owner_public_key"]))
                     == config_entry.owner_public_key
                 ):
-                    pool_config_dict["authentication_public_key"] = bytes(auth_key).hex()
+                    auth_key_hex = bytes(auth_key).hex()
+                    if pool_config_dict.get("authentication_public_key", "") != auth_key_hex:
```
altendky, line 75: `None`, the default if the key is missing, will be unequal to any string so that's probably good enough.

```suggestion
                    if pool_config_dict.get("authentication_public_key") != auth_key_hex:
```

altendky (commented):
I'll try this locally and confirm.

altendky (approved):
Symptom gone in local testing, reads like it is correct.

wjblanke (approved):
aok

Issue #10502, opened 2022-03-01
[Bug]  Incorrect balance when switching to trusted sync


mariano54:
### What happened?

@altendky shared issues when syncing to untrusted and then switching over to trusted mid sync.

### Version

1.2.12.376

### What platform are you using?

Linux

### What ui mode are you using?

CLI

mariano54:
Actually it's the opposite, start out trusted then reconnect

Issue #10449, opened 2022-02-26
[Bug] Reprocess Failed States in Trusted Sync


mariano54:
### What happened?

In the new wallet sync, if a coin state fails to be added, it will never be retried. Please see discussion here:
https://github.com/Chia-Network/chia-blockchain/pull/10415

### Version

1.3

### What platform are you using?

Linux

### What ui mode are you using?

CLI

PR #10523, opened 2022-03-02
Abort trusted sync if a state update fails

mariano54:
This solves:
https://github.com/Chia-Network/chia-blockchain/issues/10502
https://github.com/Chia-Network/chia-blockchain/issues/10449


wjblanke (approved):
aok

PR #10527, opened 2022-03-02
Update some defaults and add upnp

emlowe:
Update the following defaults for CW purposes:

- set datalayer fee to 1 billion mojos (`fee:1000000000`)
- set the port to 8575 (`host_port: 8575`)
- run the DL server by default (`run_server: True`)
 
Also, use our existing UPnP class to try to create a hole in the router for the DL service


emlowe:
closing for CI

fchirica (approved):
lgtm

PR #10528, opened 2022-03-02
make multiprocessing start method configurable

altendky:


chia/full_node/full_node.py
```diff
@@ -185,10 +190,25 @@ def sql_trace_callback(req: str):
         self.log.info("Initializing blockchain from disk")
         start_time = time.time()
         reserved_cores = self.config.get("reserved_cores", 0)
+        multiprocessing_start_method = process_config_start_method(
+            method=self.config.get("multiprocessing_start_method"),
```
mariano54, line 194: There is no default value here, should we set one?

chia/util/config.py
```diff
@@ -162,3 +164,34 @@ def traverse_dict(d: Dict, key_path: str) -> Any:
...
+    "forkserver": "forkserver",
+    "spawn": "spawn",
+}
+
+
+def process_config_start_method(
+    method: object,
+    log=logging.Logger,
+) -> Union[None, Literal["fork"], Literal["fork_server"], Literal["spawn"]]:
+    sentinel = object()
```
mariano54, line 182: This is confusing.

chia/util/config.py
```diff
@@ -162,3 +164,34 @@ def traverse_dict(d: Dict, key_path: str) -> Any:
...
+    None: None,
+    "default": None,
+    "fork": "fork",
+    "forkserver": "forkserver",
+    "spawn": "spawn",
+}
+
+
+def process_config_start_method(
+    method: object,
```
mariano54, line 179: This is also confusing, why not just pass the config object?

chia/util/config.py
```diff
@@ -162,3 +164,34 @@ def traverse_dict(d: Dict, key_path: str) -> Any:
...
+    log=logging.Logger,
+) -> Union[None, Literal["fork"], Literal["fork_server"], Literal["spawn"]]:
+    sentinel = object()
+    # TODO: I am aware that some objects may not be in the dict, that is why I am
+    #       providing .get() with a default.
+    processed_method = start_methods.get(method, sentinel)  # type: ignore[call-overload]
+
+    if processed_method is sentinel:
```
mariano54, line 187: Why not just use None? it would make the code more readable.

chia/util/config.py
altendky: It's an alternative to `None` since `None` is presently a valid value.

chia/util/config.py
altendky: Because `None` is a valid result and thus not distinguishable from invalid input.

chia/util/config.py
altendky: Passing in a thousand items when one is all that is needed is often not great.  I'll flip this around to your suggestion here.  The config needs a redesign anyways.

chia/full_node/full_node.py
altendky: The default here is `None`, albeit implicitly, just to make the config less picky and allow bare `multiprocessing_start_method:` to be valid.

altendky (commented):
I have justifications for each individual issue you raise, but overall I think I can take some of the suggestions and adjust this to be inline with your expectations.  Thanks.

mariano54 (approved):
Thank you, it is much more readable like this :)

PR #10529, opened 2022-03-02
wallet: Reduce log level for `Pulled from queue` message

xdustinface:
It prints messages per connection for a new peak, `DEBUG` is probably good enough?


hoffmang9 (approved):
lgtm and thanks

mariano54 (approved):

PR #10532, opened 2022-03-03
Wallet consistancy

Yostra:
Added db transaction when adding new coin state, this will prevent state corruption if forever reason Wallet shuts off during sync or exception is thrown in new_coin_state.
 Added `-1` when saving synced up to height, as same height can span multiple chunks 
 cleanup of some unused functions.


chia/pools/pool_wallet.py
```diff
@@ -350,12 +350,16 @@ async def create(
             if spend.coin.name() == launcher_coin_id:
                 launcher_spend = spend
         assert launcher_spend is not None
-        await self.wallet_state_manager.pool_store.add_spend(self.wallet_id, launcher_spend, block_height)
+        await self.wallet_state_manager.pool_store.add_spend(
+            self.wallet_id, launcher_spend, block_height, in_transaction
```
mariano54, line 354: PoolWallet.create is called only from within `new_coin_state` and it's using `in_transction=False` so that seems wrong

chia/wallet/trading/trade_store.py
```diff
@@ -205,31 +205,6 @@ async def increment_sent(
         await self.add_trade_record(tx, False)
         return True
 
-    async def set_not_sent(self, id: bytes32):
```
mariano54, line 208: Unused function, removing this makes sense

chia/wallet/wallet_blockchain.py
```diff
@@ -186,7 +186,7 @@ async def get_peak_block(self) -> Optional[HeaderBlock]:
 
     async def set_finished_sync_up_to(self, height: int):
         if height > await self.get_finished_sync_up_to():
-            await self._basic_store.set_object("FINISHED_SYNC_UP_TO", uint32(height))
+            await self._basic_store.set_object("FINISHED_SYNC_UP_TO", uint32(height), True)
```
mariano54, line 189: This is not always True, there are places where we call `set_finished_sync_up_to` outside of a transaction

chia/wallet/wallet_node.py
```diff
@@ -632,17 +633,28 @@ async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: i
...
-                                if len(cs_heights) == 1:
-                                    # We have processed all past tasks, so we can increase the height safely
-                                    synced_up_to = last_change_height_cs(valid_states[-1]) - 1
-                                else:
-                                    # We know we have processed everything before this min height
-                                    synced_up_to = min(cs_heights)
-                                await self.wallet_state_manager.blockchain.set_finished_sync_up_to(synced_up_to)
-
+                            try:
+                                await self.wallet_state_manager.db_wrapper.commit_transaction()
```
mariano54, line 637: do we need a commit before?

chia/pools/pool_wallet.py
mariano54: NVm i saw that it's been changed to True

chia/wallet/wallet_node.py
```diff
@@ -661,16 +673,25 @@ async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: i
...
-                    await self.wallet_state_manager.new_coin_state(states, peer, fork_height)
-                    await self.wallet_state_manager.blockchain.set_finished_sync_up_to(
-                        last_change_height_cs(states[-1]) - 1
-                    )
-                except Exception as e:
-                    tb = traceback.format_exc()
-                    self.log.error(f"Error adding states.. {e} {tb}")
-                    return False
+                async with self.new_state_lock:
+                    async with self.wallet_state_manager.db_wrapper.lock:
```
mariano54, line 677: Why do we use the DB_wrapper lock for trusted but not for untrusted?

chia/wallet/wallet_node.py
```diff
@@ -661,16 +673,25 @@ async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: i
...
-                    self.log.info(f"new coin state received ({idx}-" f"{idx + len(states) - 1}/ {len(items)})")
-                    await self.wallet_state_manager.new_coin_state(states, peer, fork_height)
-                    await self.wallet_state_manager.blockchain.set_finished_sync_up_to(
-                        last_change_height_cs(states[-1]) - 1
-                    )
-                except Exception as e:
-                    tb = traceback.format_exc()
-                    self.log.error(f"Error adding states.. {e} {tb}")
-                    return False
+                async with self.new_state_lock:
```
mariano54, line 676: Why do we need new_state lock? 
Also, we might already be in wallet_state_manage.lock, so it seems like a lot of locking

Yostra: I'm changing this to use db lock, we want to avoid anything else writing to the db during this transaction. (calling commit from elsewhere)

chia/wallet/wallet_node.py
Yostra: changing to use for both

mariano54 (dismissed):

wjblanke (approved):
aok

PR #10538, opened 2022-03-03
Add incoming tx records when cancelling an offer

Quexington:


paninaro (dismissed):
Tested and works for me.

paninaro (approved):
Tested latest changes with both a single asset offer, as well as a multi-asset offer. Inbound txn is created for each offered asset.

PR #10539, opened 2022-03-03
streamable: Enable `isort` + more `mypy`

xdustinface:
Currently based on #10509


tests/core/util/test_streamable.py
```diff
@@ -84,74 +92,83 @@ class TestClassPlain(Streamable):
...
+def test_not_lists() -> None:
     assert not is_type_List(Dict)
 
 
-def test_basic_optional():
-    assert is_type_SpecificOptional(Optional[int])
-    assert is_type_SpecificOptional(Optional[Optional[int]])
+def test_basic_optional() -> None:
+    # Below mypy failure: Argument 1 to "is_type_SpecificOptional" has incompatible type "object"; expected "Type[Any]"
+    assert is_type_SpecificOptional(Optional[int])  # type: ignore[arg-type]
```
altendky, line 113: I'm not quite sure but it may be that these functions should be hinted to accept `object` instead.  Hinting acceptance of 'a hint' is I think a thing that is not well defined at present.  In the end, _anything_ can be put in an annotation.

tests/core/util/test_streamable.py
```diff
@@ -160,11 +177,11 @@ class TestClass(Streamable):
         c: Optional[Optional[uint8]]
         d: Optional[Optional[uint8]]
 
-    good = TestClass(12, None, 13, None)
+    good = TestClass(12, None, 13, None)  # type: ignore[arg-type] # we want to test invalid here, hence the ignore.
```
altendky, line 180:
```suggestion
    # we want to test invalid here, hence the ignore.
    good = TestClass(12, None, 13, None)  # type: ignore[arg-type]
```

tests/core/util/test_streamable.py
```diff
@@ -391,15 +411,15 @@ def test_parse_bool():
...
+def test_uint32() -> None:
     assert parse_uint32(io.BytesIO(b"\x00\x00\x00\x00")) == 0
     assert parse_uint32(io.BytesIO(b"\x00\x00\x00\x01")) == 1
     assert parse_uint32(io.BytesIO(b"\x00\x00\x00\x01"), "little") == 16777216
     assert parse_uint32(io.BytesIO(b"\x01\x00\x00\x00")) == 16777216
     assert parse_uint32(io.BytesIO(b"\x01\x00\x00\x00"), "little") == 1
     assert parse_uint32(io.BytesIO(b"\xff\xff\xff\xff"), "little") == 4294967295
 
-    def test_write(value, byteorder):
+    def test_write(value: int, byteorder: Union[Literal["little"], Literal["big"]]) -> None:
```
altendky, line 422:
```suggestion
    def test_write(value: int, byteorder: Literal["little", "big"]) -> None:
```

tests/core/util/test_streamable.py
```diff
@@ -580,11 +602,11 @@ class DataclassMissing(Streamable):
...
 
-def test_streamable_inheritance_missing():
+def test_streamable_inheritance_missing() -> None:
 
     with raises(DefinitionError):
 
         @streamable
         @dataclass(frozen=True)
-        class StreamableInheritanceMissing:
+        class StreamableInheritanceMissing:  # type: ignore[type-var] # we want to test invalid here, hence the ignore.
```
altendky, line 611: Comment above the class?

altendky (commented):
See how many places you can use `object` instead of `Any`.  `Any` lets you both put anything in there as well as do anything with what is in there.  `object` only does the former while restricting the usage to what `object` provides (and everything inherits from `object`).  It might point out some opportunities to further detail hints as well.

Other comments are just a quick pass here since there are soooo many `Any`.  Hopefully they can be peared down a bit.

chia/util/streamable.py
```diff
@@ -408,20 +411,20 @@ def __post_init__(self):
...
+        inner_type: Type[Any]
         if f_type is bool:
             return parse_bool
         if is_type_SpecificOptional(f_type):
             inner_type = get_args(f_type)[0]
             parse_inner_type_f = cls.function_to_parse_one_item(inner_type)
             return lambda f: parse_optional(f, parse_inner_type_f)
         if hasattr(f_type, "parse"):
-            return f_type.parse
+            return lambda f: f_type.parse(f)
```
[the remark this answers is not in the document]
xdustinface: Since you _think_ and not **think** i would just leave it as it :D

chia/util/streamable.py
```diff
@@ -116,7 +117,7 @@ def dataclass_from_dict(klass, d):
         return klass(d)
 
 
-def recurse_jsonify(d):
+def recurse_jsonify(d: Any) -> Any:
```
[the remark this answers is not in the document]
xdustinface: Hmm, see [00dc188](https://github.com/Chia-Network/chia-blockchain/pull/10539/commits/00dc188cf9b91cf4dffd80faaf9e574628181569). Its sad that `black` adds two lines between the overloads and at the end we still need an `type: ignore[no-untyped-def]` for the base definition :/

tests/core/util/test_streamable.py
xdustinface: See [1bbeacb](https://github.com/Chia-Network/chia-blockchain/pull/10539/commits/1bbeacb42c13ce52511ed7d1043b32475d409016)

tests/core/util/test_streamable.py
xdustinface: See [21df10b](https://github.com/Chia-Network/chia-blockchain/pull/10539/commits/21df10bf03a1db07bb8eaf9a8860a362369f87d6)

tests/core/util/test_streamable.py
xdustinface: See [21df10b](https://github.com/Chia-Network/chia-blockchain/pull/10539/commits/21df10bf03a1db07bb8eaf9a8860a362369f87d6)

xdustinface:
> See how many places you can use `object` instead of `Any`

@altendky [bb691d6](https://github.com/Chia-Network/chia-blockchain/pull/10539/commits/bb691d602385fda09dcc5f84dcf9efb54e79a99f) and [0f30693](https://github.com/Chia-Network/chia-blockchain/pull/10539/commits/0f3069302c22804fd634117d81a6b6127e5bb652) is what i could figure in reasonable time.. Touching others ends up up with mypy errors in the rest of the codebase and i don't feel like thats the point here then. Feel free to provide some commits with changes if you have something specific in mind though to eliminate more of the `Any`.

chia/util/streamable.py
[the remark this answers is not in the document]
altendky: Sorry...  programmatic obfuscation to get around a hint error ain't gonna fly.  Hopefully I got this right.

```suggestion
            # Ignoring for now as the proper solution isn't obvious
            return f_type.parse  # type: ignore[no-any-return]
```

chia/util/streamable.py
```diff
@@ -116,7 +134,22 @@ def dataclass_from_dict(klass, d):
...
 
-def recurse_jsonify(d):
+@overload
+def recurse_jsonify(d: List[Any]) -> List[Any]:
+    ...
+
+
+@overload
+def recurse_jsonify(d: Tuple[Any, ...]) -> Tuple[Any, ...]:
+    ...
```
altendky, line 144: Either a list or a tuple in results in a list out.  I feel like we ought to be able to do something about the `Any` here, but I think the `Tuple` -> `List` conversion will end up making a mess.  Though I'm not entirely sure why it is needed since in the end `json` will turn tuples into json arrays anyways.  And this all seems like a reimplementation of [JSONEncoder](https://docs.python.org/3.10/library/json.html#json.JSONEncoder).  Oh well, we'll get the easy thing anyways.

```suggestion
@overload
def recurse_jsonify(d: Union[List[Any], Tuple[Any, ...]) -> List[Any]:
    ...
```

chia/util/streamable.py
```diff
@@ -116,7 +134,22 @@ def dataclass_from_dict(klass, d):
...
+
+
+@overload
+def recurse_jsonify(d: Dict[str, Any]) -> Dict[str, Any]:
+    ...
+
+
+def recurse_jsonify(d):  # type: ignore[no-untyped-def]
```
altendky, line 152: Others agree about the excessive whitespace.  https://github.com/psf/black/issues/1797

Without saying this is a beautiful situation...  here's a thing that I would start with and see if mypy is ok with it.

```suggestion
def recurse_jsonify(d: Union[List[Any], Tuple[Any, ...], Dict[str, Any]]) -> Union[List[Any], Dict[str, Any]:
```

chia/util/streamable.py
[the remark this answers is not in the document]
altendky: Is that better?  `:]`

altendky (changes requested):

chia/util/streamable.py
xdustinface: Added 6c18bfef66c2c8114f006a0ee0f1d2c3b3994b00, the suggestion was missing a `]`

chia/util/streamable.py
xdustinface: Also missing a `]` :) See 5e2ba5f6060161e04cb73365e242e63b774452d5.

altendky (approved):

PR #10541, opened 2022-03-03
switch back to official dnspython for v2.2.0

altendky:


altendky:
v2.1.0 hangs with the following code while v2.2.0 prints the output.

```python
import asyncio
import dns.asyncresolver

r = dns.asyncresolver.Resolver()
s = "dns-introducer-testnet10.chia.net"


async def test():

    try:
        result = await r.resolve(s, rdtype="A", lifetime=30)
        print(f"Recived {len(result)} peers from DNS serrver, using rdtype=A")
        print(f"{result}")

    except Exception as e:
        print(f"querying {s} failed: {e}")


asyncio.run(test())
```
```
Recived 27 peers from DNS serrver, using rdtype=A
<dns.resolver.Answer object at 0x000001D27FF997C0>
```

emlowe (approved):

PR #10543, opened 2022-03-03
Detect hints correctly in the TX

### LINKED ISSUE #10537: [Bug] missing CAT tokens
### What happened?

CAT tokens that are on the Verified list (Eg. CH21) that have a balance are not showing up automatically after importing from mnemonics. As reported by @cmmarslender 

### Version

1.2.12 dev389

Linux

CLI

### Relevant log output

_No response_

mariano54:
Fixes: https://github.com/Chia-Network/chia-blockchain/issues/10537

### DIFF
--- a/chia/wallet/util/compute_hints.py
+++ b/chia/wallet/util/compute_hints.py
@@ -1,34 +1,22 @@
 from typing import List
 
-from blspy import G2Element
-
+from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.condition_opcodes import ConditionOpcode
 from chia.types.blockchain_format.program import INFINITE_COST
 from chia.types.coin_spend import CoinSpend
-from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions
-from chia.consensus.default_constants import DEFAULT_CONSTANTS
-from chia.types.spend_bundle import SpendBundle
-from chia.full_node.bundle_tools import simple_solution_generator
 
 
-def compute_coin_hints(cs: CoinSpend) -> List[bytes]:
+def compute_coin_hints(cs: CoinSpend) -> List[bytes32]:
+    _, result_program = cs.puzzle_reveal.run_with_cost(INFINITE_COST, cs.solution)
 
-    bundle = SpendBundle([cs], G2Element())
-    generator = simple_solution_generator(bundle)
-
-    npc_result = get_name_puzzle_conditions(
-        generator,
-        INFINITE_COST,
-        cost_per_byte=DEFAULT_CONSTANTS.COST_PER_BYTE,
-        mempool_mode=False,
-        height=DEFAULT_CONSTANTS.SOFT_FORK_HEIGHT,
-    )
-    h_list = []
-    for npc in npc_result.npc_list:
-        for opcode, conditions in npc.conditions:
-            if opcode == ConditionOpcode.CREATE_COIN:
-                for condition in conditions:
-                    if len(condition.vars) > 2 and condition.vars[2] != b"":
-                        h_list.append(condition.vars[2])
-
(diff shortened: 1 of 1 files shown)


chia/wallet/util/compute_hints.py
```diff
@@ -1,34 +1,22 @@
...
+def compute_coin_hints(cs: CoinSpend) -> List[bytes32]:
+    _, result_program = cs.puzzle_reveal.run_with_cost(INFINITE_COST, cs.solution)
 
+    h_list: List[bytes32] = []
+    for condition_data in result_program.as_python():
+        condition = condition_data[0]
+        args = condition_data[1:]
+        if condition == ConditionOpcode.CREATE_COIN and len(args) > 2:
+            if isinstance(args[2], bytes):
+                h_list.append(bytes32(args[2]))
```
Quexington, line 18: I don't think this is a valid definition of a hint.  I think it has to be the first item in a list, it cannot be an atom.  @arvidn should double check me.

mariano54: I had this as a fallback since sometimes the first item is returned flattened, while sometimes it's returned as a list, from rust CLVM

chia/wallet/util/compute_hints.py
```diff
@@ -1,34 +1,22 @@
...
+            elif isinstance(args[2], list):
+                if isinstance(args[2][0], bytes):
+                    h_list.append(bytes32(args[2][0]))
```
Quexington, line 21: maybe try/catch this whole block instead of all of the if/isinstance's.  Maybe that's just stylistic so maybe it's not very important.

chia/wallet/util/compute_hints.py
```diff
@@ -1,34 +1,22 @@
...
-    for npc in npc_result.npc_list:
-        for opcode, conditions in npc.conditions:
-            if opcode == ConditionOpcode.CREATE_COIN:
-                for condition in conditions:
-                    if len(condition.vars) > 2 and condition.vars[2] != b"":
-                        h_list.append(condition.vars[2])
+def compute_coin_hints(cs: CoinSpend) -> List[bytes32]:
+    _, result_program = cs.puzzle_reveal.run_with_cost(INFINITE_COST, cs.solution)
 
+    h_list: List[bytes32] = []
```
Quexington, line 12: Should this be a dictionary of which coins have which hints?

chia/wallet/util/compute_hints.py
mariano54: The behaviour is different between `run_chia_program` and `run_generator2`, maybe names should be changed to be more descriptive?

chia/wallet/util/compute_hints.py
mariano54: we only care about the hint values in this case.

chia/wallet/util/compute_hints.py
mariano54: yeah I rather not throw, because it's possible for the coin spend to have a different format and therefore yeah I rather not throw, because it's possible for the coin spend to have a different format and just return [] here right?

chia/wallet/util/compute_hints.py
mariano54: OK fixed in latest commit

arvidn:
we really should have a test case demonstrating this problem

arvidn:
@mariano54 this looks correct. But what was the problem with the existing version?

mariano54:
Good point about the test case. The issue was that `get_name_puzzle_conditions` was failing due to an announcement assertion (we only run one spend here and don't include all removals). And therefore we did not get any npcs in the npc list.

Quexington (approved):
Seems like a valid change to me.

PR #10545, opened 2022-03-04
Add fee to changing pool, and more PlotNFT syncing fixes

mariano54:
* There were two separate transactions being created instead of one, and therefore no fee was getting adding to the singleton spend. Also we remove warning logs for the second TX.
* Make sure to download all coin updates (don't filter)
* Make sure to actually subscribe to coin updates (the boolean was set to False)
* Give a more clear error if the wallet is not a plot nft
* Give a more clear error when the height has not reached the height required to submit the 2nd pool change tx
* add support for get_transaction and get_transactions which were failing previously


chia/pools/pool_wallet.py
```diff
@@ -495,11 +494,6 @@ async def generate_fee_transaction(self, fee: uint64, coin_announcements=None) -
         )
         return fee_tx
 
-    async def publish_transactions(self, travel_tx: TransactionRecord, fee_tx: Optional[TransactionRecord]):
-        await self.wallet_state_manager.add_pending_transaction(travel_tx)
-        if fee_tx is not None:
-            await self.wallet_state_manager.add_pending_transaction(dataclasses.replace(fee_tx, spend_bundle=None))
-
```
Quexington, line 502: I think you should leave this pattern in.  You fixed the problem where the spend bundles were not being aggregated before being pushed, but as you identified in your comment, this means that the standard wallet has no transaction record for the fee leaving.  If you keep this method, that problem should be solved.  This is a pattern we already use for CATs, Offers, and other things still in development.

Quexington (changes requested):

chia/pools/pool_wallet.py
mariano54: We broadcast all unconfirmed TXs to the full node though, so that is incompatible with this pattern. Is there a way to create an unconfirmed TX but not broadcast it?

mariano54: Wouldn't it be simpler to create the TX when the first on gets confirmed instead?

Quexington: We don't broadcast anything to a full node if the SpendBundle slot is None, that's why `dataclasses.replace(fee_tx, spend_bundle=None)` is in there

aqk: The way that the transaction book-keeping really works was very very not-obvious to me - I'd really appreciate a comment in the above function.

chia/pools/pool_wallet.py
```diff
@@ -572,14 +566,14 @@ async def generate_travel_transactions(self, fee: uint64) -> Tuple[TransactionRe
...
-            fee_tx = await self.generate_fee_transaction(fee)
-
         signed_spend_bundle = await self.sign(outgoing_coin_spend)
         assert signed_spend_bundle.removals()[0].puzzle_hash == singleton.puzzle_hash
         assert signed_spend_bundle.removals()[0].name() == singleton.name()
         assert signed_spend_bundle is not None
+        fee_tx = None
+        if fee > 0:
+            fee_tx = await self.generate_fee_transaction(fee)
+            signed_spend_bundle = SpendBundle.aggregate([signed_spend_bundle, fee_tx.spend_bundle])
```
aqk, line 577: Thank you. I had this aggregate in an earlier iteration, and I missed it while making other changes.

chia/pools/pool_wallet.py
```diff
@@ -718,7 +712,10 @@ async def join_pool(
         if current_state.current.state == LEAVING_POOL:
             history: List[Tuple[uint32, CoinSpend]] = await self.get_spend_history()
             last_height: uint32 = history[-1][0]
-            if self.wallet_state_manager.get_peak().height <= last_height + current_state.current.relative_lock_height:
+            if (
+                self.wallet_state_manager.blockchain.get_peak_height()
```
aqk, line 719: `self.wallet_state_manager.get_peak().height` vs. `self.wallet_state_manager.blockchain.get_peak_height()`

This one might be subtle - is `blockchain.get_peak_height()` more vetted than `get_peak_height()` ?

chia/wallet/wallet_node.py
```diff
@@ -536,14 +536,14 @@ def is_new_state_update(cs: CoinState) -> bool:
...
         # sync from 0 every time.
-        continue_while = False
+        continue_while = True
         all_coin_ids: List[bytes32] = await self.get_coin_ids_to_subscribe(0)
         already_checked_coin_ids: Set[bytes32] = set()
         while continue_while:
             one_k_chunks = chunks(all_coin_ids, 1000)
             for chunk in one_k_chunks:
                 c_update_res: List[CoinState] = await subscribe_to_coin_updates(chunk, full_node, 0)
-                c_update_res = list(filter(is_new_state_update, c_update_res))
```
aqk, line 546: That's a pretty major change here - totally fine, as long as the downstream code `self.receive_state_from_peer` is prepared for it.

BTW, is
    `ph_update_res = list(filter(is_new_state_update, ph_update_res))`
above still valid as well?

aqk (changes requested):
Everything looks great to me, except two things:

1. Are you sure `receive_state_from_peer` is robust against the change to `c_update_res = list(filter(is_new_state_update, c_update_res))` ?
2. Is it better to leave in / modify `publish_transactions` ?

chia/pools/pool_wallet.py
aqk: I mean, is it possible for them to be different blocks?

Not a review issue, just wondering.

mariano54: WSM no longer has a `get_peak` method so that code wasn't working before this PR. (I think this was broken in the big merge)

chia/wallet/wallet_node.py
mariano54: Yeah we want to filter PH stuff but not coin stuff, the reason is that there are many many more PH updates usually, coin subscriptions are only really needed for a few things like plotnfts, so it's fine to not filter them. Worst case we reprocess some stuff. 

Before this PR, this code wasn't even being ran due to the `continue_while` value being False.

chia/wallet/wallet_state_manager.py
```diff
@@ -874,6 +874,7 @@ async def new_coin_state(
                         "pool_wallet",
                     )
                     coin_added = launcher_spend.additions()[0]
+                    assert len(launcher_spend.additions()) == 1
```
mariano54, line 877: Not needed but just a sanity check, according to Quex this is always true.

chia/pools/pool_wallet.py
aqk: cool, thanks

mariano54:
> Everything looks great to me, except two things:
> 
>     1. Are you sure `receive_state_from_peer` is robust against the change to `c_update_res = list(filter(is_new_state_update, c_update_res))` ?
> 
>     2. Is it better to leave in / modify `publish_transactions` ?

Thanks, I can add that function back and put a comment.

chia/wallet/wallet_state_manager.py
Quexington: Can we make this not run `additions()` twice?  It runs CLVM which is often slow.

Quexington (approved):
lgtm

PR #10547, opened 2022-03-04
correct multiprocessing start method logging, add python_default

altendky:


chia/util/config.py
```diff
@@ -177,23 +180,24 @@ def traverse_dict(d: Dict, key_path: str) -> Any:
 def process_config_start_method(
     config: Dict[str, Any],
     log=logging.Logger,
-) -> Optional[Literal["fork", "forkserver", "spawn"]]:
-    from_config = config.get("multiprocessing_start_method")
+) -> method_values:
+    from_config: object = config.get("multiprocessing_start_method")
```
mariano54, line 184: Why not do 

```
from_config = config.get("multiprocessing_start_method", "default")
```

mariano54 (commented):
LGTM but I would remove the TODO, there is nothing to do here.

chia/util/config.py
altendky: If we allow `multiprocessing_start_method:` or ``multiprocessing_start_method: null` then we have to handle `None` anyways.  We could drop handling that though, sure.  It's probably more useful with lists like the pools.

mariano54 (approved):

PR #10551, opened 2022-03-04
Resubmit peak to timelord for failure.

fchirica:


chia/full_node/full_node.py
```diff
@@ -1526,6 +1527,8 @@ async def respond_block(
                     ):
                         self.full_node_store.previous_generator = None
 
+                if raise_if_not_added and added != ReceiveBlockResult.NEW_PEAK:
+                    raise RuntimeError(f"Expected block to be added, received {added}")
```
emlowe, line 1531: This does override the later check for invalid_block which raises ConsensusError

fchirica: would it be better to raise it in every if brach, before the `return None`? I went for this just to introduce minimal change.

mariano54:
This is also potentially problematic: If the full node still has a local invalid unfinished block, and the timelord will just retry and reinfuse the same bad thing, it will just be an infinite loop of bad infusion. I think we need to fully reset the timelord here by clearing it's unfinished block list (maybe an extra param in the timelord protocol)?

mariano54 (changes requested):
This will also raise for orphan blocks, which I don't think we want. What if instead we return the ReceiveBlockStatus and check for NEW_PEAK or ADDED_AS_ORPHAN, if not, then we reset timelord

chia/full_node/full_node.py
emlowe: I don't know all the nuances, but this code already has a failsafe raise at the end, so I think it makes sense to only add this to the DISCONNECTED_BLOCK section.

ALREADY_HAVE_BLOCK: (should this raise?)
INVALID_BLOCK: already raise ConsensusError
DISCONNECTED_BLOCK: `add raise`
NEW_PEAK: no raise
ADDED_AS_ORPHAN: no raise
Anything else: already raise RuntimeError

mariano54: I don't think we should raise there, it's possible a peer gives us a block that is disconnected, we don't want to disconnect. 
We should just return as I mentioned in the other comment

mariano54:
> This will also raise for orphan blocks, which I don't think we want. What if instead we return the ReceiveBlockStatus and check for NEW_PEAK or ADDED_AS_ORPHAN, if not, then we reset timelord

Oh I don't think we can return since we already return Optional[Message]. Maybe add in an arguemnt that is: raise_on_disconnected. 

I agree with earle, we shouldn't change how INVALID_BLOCK gets handled, and we don't want to raise usually on a disconnected block, since that can disconnect honest peers.

mariano54 (approved):

chia/full_node/full_node.py
fchirica: fixed, I didn't initially realised it's just 1 case we're interested in.

PR #10554, opened 2022-03-04
Enable clvm_tools_rs by default

prozacchiwawa:


prozacchiwawa:
Fairly certain this is ready.

setup.py
```diff
@@ -8,8 +8,9 @@
     "chiabip158==1.1",  # bip158-style wallet filters
     "chiapos==1.0.9",  # proof of space
     "clvm==0.9.7",
+    "clvm_tools==0.4.3",  # Currying, Program.to, other conveniences
```
aqk, line 11: Thanks for comment

chia/wallet/puzzles/load_clvm.py
```diff
@@ -1,53 +1,106 @@
...
-            compile_clvm_rs(str(full_path), str(output), treated_include_paths)
 
-            if os.environ["CLVM_TOOLS_RS"] == "check":
-                orig = str(output) + ".orig"
-                compile_clvm_py(full_path, orig, search_paths=search_paths)
-                orig256 = sha256file(orig)
-                rs256 = sha256file(output)
+def compile_clvm(full_path, output, search_paths=[]):
+    # Ensure that we're the only one recompiling full_path
+    lock_filename = f"{full_path}.lock"
```
aqk, line 89: Oh, yay

chia/wallet/puzzles/load_clvm.py
```diff
@@ -1,53 +1,106 @@
...
 
-        from clvm_tools_rs import compile_clvm as compile_clvm_rs
+        orig = "%s.orig" % output
+
+        compile_clvm_py(full_path, orig, search_paths=search_paths)
+        orig256 = sha256file(orig)
+        rs256 = sha256file(output)
+
+        if orig256 != rs256:
+            print("Compiled %s: %s vs %s\n" % (full_path, orig256, rs256))
```
aqk, line 78: You should identify which is which in the diagnostic output

chia/wallet/puzzles/load_clvm.py
```diff
@@ -1,53 +1,106 @@
...
 import pathlib
 
 import pkg_resources
-from clvm_tools.clvmc import compile_clvm as compile_clvm_py
 from chia.types.blockchain_format.program import Program, SerializedProgram
+from clvm_tools_rs import compile_clvm as compile_clvm_rust
+
+
+# Cribbed mostly from chia/daemon/server.py
+def create_exclusive_lock(lockfile) -> Optional[TextIO]:
```
aqk, line 15: I'd vote to put `create_exclusive_lock` in a new file e.g. chia/util/lock.py

aqk: I know it's short - just to prevent multiple versions arising

chia/wallet/puzzles/load_clvm.py
```diff
@@ -1,53 +1,106 @@
...
+        module_object = importlib.import_module(p)
+        return os.path.dirname(inspect.getfile(module_object))
+
+
+# Handle optional use of python clvm_tools if available and requested
+if "CLVM_TOOLS" in os.environ:
     try:
+        from clvm_tools.clvmc import compile_clvm as compile_clvm_py_candidate
 
+        compile_clvm_py = compile_clvm_py_candidate
```
aqk, line 48: If it is possible, it would be nice to import the version of the python clvm module. Strictly nice-to-have.

aqk (commented):
This looks good. I think one thing we should discuss in the clvm channel is: should this fail if we can't load both compilers to compare the output?

chia/wallet/puzzles/load_clvm.py
prozacchiwawa: makes sense.

chia/wallet/puzzles/load_clvm.py
prozacchiwawa: Do i understand right that you're asking to also import python clvm alongside python clvm_tools?

chia/wallet/puzzles/load_clvm.py
```diff
@@ -1,53 +1,106 @@
...
+
+
+# Cribbed mostly from chia/daemon/server.py
+def create_exclusive_lock(lockfile) -> Optional[TextIO]:
+    """
+    Open a lockfile exclusively.
+    """
+
+    try:
+        fd = os.open(lockfile, os.O_CREAT | os.O_EXCL | os.O_RDWR)
```
arvidn, line 21: this looks very posix-y. does it work on windows?

chia/wallet/puzzles/load_clvm.py
```diff
@@ -1,53 +1,106 @@
...
+        orig = "%s.orig" % output
+
+        compile_clvm_py(full_path, orig, search_paths=search_paths)
+        orig256 = sha256file(orig)
+        rs256 = sha256file(output)
+
+        if orig256 != rs256:
+            print("Compiled %s: %s vs %s\n" % (full_path, orig256, rs256))
+            print("Aborting compilation due to mismatch with rust")
+            assert orig256 == rs256
```
arvidn, line 80: I think it would be good to raise an exception explicitly here, rather than the assert. I believe asserts primarily are meant to document invariants. If this happens, we would want to exit, even if we're running in "release" mode.

(our codebase doesn't follow this rule, and sometimes use assert as input validation, but I think we should move away from that use of asserts)

chia/wallet/puzzles/load_clvm.py
```diff
@@ -1,53 +1,106 @@
...
-                    print("Aborting compilation due to mismatch with rust")
-                    assert orig256 == rs256
+    lock_file = None
+    while True:
+        lock_file = create_exclusive_lock(lock_filename)
+        if lock_file is not None:
+            break
 
-        compile_clvm = rust_compile_clvm
+        time.sleep(0.1)
```
arvidn, line 97: perhaps not for this PR. but this looks like a perfect candidate for a context manager, that loops trying to acquire the lock, and then releases it when leaving the context manager.

setup.py
arvidn: `Program.to()` comes from `chia-blockchain`. in `chia/types/blockchain_format/program.py`. It relies on `SExp` from `clvm`. Currying might come from `clvm_tools` though

arvidn (commented):
looks good!

chia/wallet/puzzles/load_clvm.py
prozacchiwawa: i like how it turned out

chia/wallet/puzzles/load_clvm.py
prozacchiwawa: I will test it.

chia/wallet/puzzles/load_clvm.py
prozacchiwawa: It's not meant precisely as input validation and is intended to be transient while people settle in (basically to be able to ask: is this really the same as it was).  Assert fit my mental model because it really does indicate that the world has gone bad; nothing else can be expected to work right.  I will consider how to do this better.

chia/wallet/puzzles/load_clvm.py
prozacchiwawa: Makes sense.  It was a somewhat desperate move as I didn't want to leave everyone waiting while a concurrency problem caused grief.  Amine helped me come to the decision to do a heavy handed solution but I think I can probably eliminate it with more open time for investigation.

chia/wallet/puzzles/load_clvm.py
arvidn: yeah, that's a good point. This isn't really code that runs in production

arvidn:
I think this looks good, but the LGTM issues should probably be addressed before landing

chia/wallet/puzzles/load_clvm.py
aqk: I agree not to do it in this PR

arvidn (approved):

chia/wallet/puzzles/load_clvm.py
aqk: No, I just mean to be able to say "hey, I tried clvm_tools version x.y.z" in a diagnostic.
As even though the `clvm_tools` dependency on `clvm` is `>=`, having the version in a failure diagnostic will put them on the right track if they need to update.

PR #10558, opened 2022-03-04
less optional around ssl

altendky:


emlowe (approved):
Good for me - these aren't Optional in the logical sense, if we don't have an SSL context, we don't connect

PR #10559, opened 2022-03-05
DID changes all in one 

ytx1991:
This is a draft version for clarify the requirement. I think it could be better if we can make it more generalized, I mean cover the other modes. Another question is if the injection attack is possible in the Chialisp? Since we are not signing the solution.


chia/rpc/wallet_rpc_api.py
```diff
@@ -512,14 +516,22 @@ async def create_new_wallet(self, request: Dict):
...
+
                 async with self.service.wallet_state_manager.lock:
                     did_wallet: DIDWallet = await DIDWallet.create_new_did_wallet(
                         wallet_state_manager,
                         main_wallet,
                         uint64(request["amount"]),
                         backup_dids,
                         uint64(num_needed),
+                        metadata,
+                        request["wallet_name"] if "wallet_name" in request else "DID Wallet",
```
paninaro, line 532:
```suggestion
                        request.get("wallet_name", "DID Wallet"),
```

chia/rpc/wallet_rpc_api.py
```diff
@@ -1089,6 +1101,21 @@ async def did_update_recovery_ids(self, request):
...
+        wallet_id = int(request["wallet_id"])
+        wallet: DIDWallet = self.service.wallet_state_manager.wallets[wallet_id]
+        metadata: Dict[str, str] = {}
+        if "metadata" in request:
+            if type(request["metadata"]) is dict:
+                metadata = request["metadata"]
+        async with self.service.wallet_state_manager.lock:
+            update_success = await wallet.update_metadata(metadata)
+            # Update coin with new ID info
+            spend_bundle = await wallet.create_update_spend()
```
paninaro, line 1114:
```suggestion
            spend_bundle = await wallet.create_update_spend() if update_success else None
```

paninaro: Avoids unnecessarily calling create_update_spend() if the call to `update_metadata` failed.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1111,10 +1138,20 @@ async def did_get_recovery_list(self, request):
...
             "wallet_id": wallet_id,
-            "recover_list": recover_hex_list,
+            "recovery_list": recover_hex_list,
             "num_required": wallet.did_info.num_of_backup_ids_needed,
         }
 
+    async def did_get_metadata(self, request):
+        wallet_id = int(request["wallet_id"])
+        wallet: DIDWallet = self.service.wallet_state_manager.wallets[wallet_id]
+        metadata = json.loads(wallet.did_info.metadata)
```
paninaro, line 1148: Is there a basic schema/shape that we're expecting the metadata to conform to? Array of simple string --> string key value pairs, for instance? I'd be careful with accepting whatever JSON data the user provides.

paninaro: I would also recommend wrapping the JSON parsing in a try:except: so that we can return an error stating that the JSON was invalid. Otherwise, the user is going to see whatever exception was thrown by the JSON parser.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1230,7 +1287,10 @@ async def nft_mint_nft(self, request):
                 ("h", request["hash"]),
             ]
         )
-        await nft_wallet.generate_new_nft(metadata, request["artist_percentage"], address)
+        if "amount" in request:
+            await nft_wallet.generate_new_nft(metadata, request["artist_percentage"], address, request["amount"])
+        else:
+            await nft_wallet.generate_new_nft(metadata, request["artist_percentage"], address)
```
paninaro, line 1293:
```suggestion
        await nft_wallet.generate_new_nft(metadata, request["artist_percentage"], address, request.get("amount"))
```

chia/wallet/did_wallet/did_info.py
```diff
@@ -21,3 +21,4 @@ class DIDInfo(Streamable):
     temp_puzhash: Optional[bytes32]
     temp_pubkey: Optional[bytes]
     sent_recovery_transaction: bool
+    metadata: str  # User defined metadata
```
paninaro, line 24:
```suggestion
    metadata: str  # User defined metadata stored as JSON
```

paninaro: I'm assuming the metadata is a JSON blob based on the RPC

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -141,18 +149,19 @@ async def create_new_did_wallet_from_recovery(
...
         self.standard_wallet = wallet
         self.log = logging.getLogger(name if name else __name__)
+        self.log.info("Creating DID wallet from recovery file ...")
         self.wallet_state_manager = wallet_state_manager
-        self.did_info = DIDInfo(None, [], uint64(0), [], None, None, None, None, False)
+        self.did_info = DIDInfo(None, [], uint64(0), [], None, None, None, None, False, "")
         info_as_string = json.dumps(self.did_info.to_json_dict())
         self.wallet_info = await wallet_state_manager.user_store.create_wallet(
-            "DID Wallet", WalletType.DISTRIBUTED_ID.value, info_as_string
+            name, WalletType.DISTRIBUTED_ID.value, info_as_string, in_transaction=True
```
paninaro, line 164: Is the `in_transaction=True` correct here? Where is the DB transaction being created? From what I can tell, `create_new_did_wallet_from_recovery` is only called by `wallet_rpc_api.py` which doesn't seem to create a DB transaction prior to calling this function. I see the wallet_state_manager's lock being acquired, but that's not the same lock as used by the wallet_user_store's DBWrapper.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -163,6 +172,70 @@ async def create_new_did_wallet_from_recovery(
...
+            None,
+            None,
+            None,
+            False,
+            json.dumps(did_wallet_puzzles.program_to_metadata(metadata)),
+        )
+        info_as_string = json.dumps(self.did_info.to_json_dict())
+
+        self.wallet_info = await wallet_state_manager.user_store.create_wallet(
+            name, WalletType.DISTRIBUTED_ID.value, info_as_string, in_transaction=True
```
paninaro, line 227: Passing `in_transaction=True` makes sense to me here because `create_new_did_wallet_from_coin_spend()` is called in in the context of an already-opened DB transaction (in `wallet_node.py` `receive_and_validate()`.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -365,20 +441,32 @@ async def load_backup(self, filename: str):
                 None,
                 None,
                 False,
+                metadata,
             )
             await self.save_info(did_info, False)
             await self.wallet_state_manager.update_wallet_puzzle_hashes(self.wallet_info.id)
+            await self.load_parent(did_info)
+        except Exception as e:
+            raise e
```
paninaro, line 450: Remove this? Or is there some intent to handle the exception?

chia/wallet/did_wallet/did_wallet_puzzles.py
```diff
@@ -82,24 +74,68 @@ def get_innerpuzzle_from_puzzle(puzzle: Program) -> Optional[Program]:
...
+
+def program_to_metadata(program: Program) -> Dict:
+    """
+    Convert a program to a metadata dict
+    :param program: Chialisp program contains the metadata
+    :return: Metadata dict
+    """
+    metadata = {}
+    for key, val in program.as_iter():
+        metadata[str(key.as_python(), "ascii")] = str(val.as_python(), "ascii")
```
paninaro, line 140: Is the ascii really the right encoding to use? I would have thought that the default of UTF-8 would be more appropriate.

ytx1991: It depends if we want to support utf-8 chars. Anyway, I changed it and covered this in the test.

paninaro (approved):

PR #10561, opened 2022-03-05
streamable: Cache convert functions from `dataclass_from_dict`

xdustinface:
Applies the pattern from #10419 to `dataclass_from_dict` i.e. it caches the conversion functions on startup which leads to nice improvements of the `from_json` benchmark (Calling `Class.from_json_dict`). Not so much for `full_block`, i guess because the subgroup checks of the BLS deserialisation still dominate. 

```
compare: benchmark, old: 4bd5c53f4-dirty, new: 51f97c79c-dirty
from_json    | 1091.09          | 568.4            | -47.9       

compare: full_block, old: 4bd5c53f4-dirty, new: 51f97c79c-dirty
from_json    | 3090.49          | 2739.37          | -11.36     
```

It also adds some more tests where `dataclass_from_dict` gets used with non-streamable dataclasses because there is a special handling of them now in there where the cache gets created on the fly (we can't do it on startup for non-streamable dataclasses because of the missing decorator call).

https://github.com/Chia-Network/chia-blockchain/blob/5b1b4ddd9d3b72f66ef57e42528c1f9674b869d3/chia/util/streamable.py#L124-L135
 
Based on #10539

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -81,7 +81,7 @@ ConvertFunctionType = Callable[[object], object]
 FIELDS_FOR_STREAMABLE_CLASS: Dict[Type[object], Dict[str, Type[object]]] = {}
 STREAM_FUNCTIONS_FOR_STREAMABLE_CLASS: Dict[Type[object], List[StreamFunctionType]] = {}
 PARSE_FUNCTIONS_FOR_STREAMABLE_CLASS: Dict[Type[object], List[ParseFunctionType]] = {}
-CONVERT_FUNCTIONS_FOR_STREAMABLE_CLASS: Dict[Type[Any], List[ConvertFunctionType]] = {}
+CONVERT_FUNCTIONS_FOR_STREAMABLE_CLASS: Dict[Type[object], List[ConvertFunctionType]] = {}
 
 
 def is_type_List(f_type: object) -> bool:
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -142,6 +142,10 @@ def convert_primitive(f_type: Type[Any], item: Any) -> Any:
 
 
 def dataclass_from_dict(klass: Type[Any], item: Any) -> Any:
+    """
+    Converts a dictionary based on a dataclass, into an instance of that dataclass.
+    Recursively goes through lists, optionals, and dictionaries.
+    """
     if type(item) == klass:
         return item
     obj = object.__new__(klass)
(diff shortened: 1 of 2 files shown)


mariano54:
@arvidn is making changes to the same code, so it's probably good to coordinate

xdustinface:
@mariano54 Do you mean https://github.com/Chia-Network/chia-blockchain/pull/11537? If so, thats `to_json` but this PR here is about `from_json`. Or do you have other stuff in progress about `dataclass_from_dict` @arvidn ?

mariano54:
> @mariano54 Do you mean #11537? If so, thats `to_json` but this PR here is about `from_json`. Or do you have other stuff in progress about `dataclass_from_dict` @arvidn ?

Yes that is the one I was referring to.

chia/util/streamable.py
```diff
@@ -97,45 +99,105 @@ def is_type_Tuple(f_type: object) -> bool:
...
+def convert_primitive(f_type: Type[Any], item: Any) -> Any:
+    if type(item) == f_type:
+        return item
+    return f_type(item)
+
+
+def dataclass_from_dict(klass: Type[Any], item: Any) -> Any:
+    if type(item) == klass:
+        return item
+    obj = object.__new__(klass)
```
altendky, line 147: I think this blocks `.__init__()` and thus also `.__post_init__()` which I think we at least need to think through.  I personally want us doing this less, not more.

chia/util/streamable.py
```diff
@@ -97,45 +99,105 @@ def is_type_Tuple(f_type: object) -> bool:
...
+    for field, convert_func in zip(fields, convert_funcs):
+        object.__setattr__(obj, field, convert_func(item[field]))
+    return obj
+
+
+def function_to_convert_one_item(f_type: Type[Any]) -> ConvertFunctionType:
     """
     Converts a dictionary based on a dataclass, into an instance of that dataclass.
     Recursively goes through lists, optionals, and dictionaries.
     """
```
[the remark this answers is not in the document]
xdustinface: See 439ed9c0026d9fa8f1053c886fa2cfe6400e3652

chia/util/streamable.py
xdustinface: Hm yeah so if iirc i did this intentionally to avoid calling `__post_init__` in this case for performance reasons and because it would be redundant in this case since the data we would pass with 

```
return klass(**{field: convert_func(item[field]) for field, convert_func in zip(fields, convert_funcs)})
```

is already validated. Whats the exact problem you see with this approach?

altendky (approved):

PR #10566, opened 2022-03-05
Check prefix on send_transaction before sending

cmmarslender:
Checks the target address prefix on send_transaction compared to the expected value in the config and raises a ValueError if the prefix doesn't match what is expected.

Checked the initial config, and prefixes have been in there since at least 1.0.0: [1.0.0 Initial Config](https://github.com/Chia-Network/chia-blockchain/blob/91f038e2193755e2a6ca22e2160e2c8f547c23fe/src/util/initial-config.yaml)

**_I had zero balance, so I couldn't actually send anything, but the error varies, and is inline with expectations with this code change_**

### Mainnet

```
chia wallet send -f 2241697793 -i 1 -a 1 -m 0 -t xfx1ll4q092ulfqdzq7vnksdtht5t4h99zzn52lcvzs2mvmhprpv3jdqy2twhh
Submitting transaction...
Exception from 'wallet' {'error': 'Unexpected Address Prefix', 'success': False}
```

```
chia wallet send -f 2241697793 -i 1 -a 1 -m 0 -t xch1ll4q092ulfqdzq7vnksdtht5t4h99zzn52lcvzs2mvmhprpv3jdqafv254
Submitting transaction...
Exception from 'wallet' {'error': "Can't send more than 0 in a single transaction", 'success': False}
```

### Testnet

```
chia wallet send -f 2241697793 -i 1 -a 1 -m 0 -t xfx1ll4q092ulfqdzq7vnksdtht5t4h99zzn52lcvzs2mvmhprpv3jdqy2twhh
Submitting transaction...
Exception from 'wallet' {'error': 'Unexpected Address Prefix', 'success': False}
```

```
chia wallet send -f 2241697793 -i 1 -a 1 -m 0 -t txch1ll4q092ulfqdzq7vnksdtht5t4h99zzn52lcvzs2mvmhprpv3jdqswtu4x
Submitting transaction...
Exception from 'wallet' {'error': "Can't send more than 0 in a single transaction", 'success': False}
```

## UI Test

Screen Shot 2022-03-05 at 3 57 17 PM

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -756,7 +756,12 @@ class WalletRpcApi:
         if not isinstance(request["amount"], int) or not isinstance(request["fee"], int):
             raise ValueError("An integer amount or fee is required (too many decimals)")
         amount: uint64 = uint64(request["amount"])
-        puzzle_hash: bytes32 = decode_puzzle_hash(request["address"])
+        address = request["address"]
+        selected_network = self.service.config["selected_network"]
+        expected_prefix = self.service.config["network_overrides"]["config"][selected_network]["address_prefix"]
+        if address[0 : len(expected_prefix)] != expected_prefix:
+            raise ValueError("Unexpected Address Prefix")
+        puzzle_hash: bytes32 = decode_puzzle_hash(address)
 
         memos: List[bytes] = []
         if "memos" in request:
--- a/tests/wallet/rpc/test_wallet_rpc.py
+++ b/tests/wallet/rpc/test_wallet_rpc.py
@@ -142,7 +142,7 @@ class TestWalletRpc:
         client_node = await FullNodeRpcClient.create(self_hostname, test_rpc_port_node, bt.root_path, config)
         try:
             await time_out_assert(5, client.get_synced)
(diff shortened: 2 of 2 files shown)


hoffmang9 (approved):
lgtm pending CI passing

chia/rpc/wallet_rpc_api.py
```diff
@@ -756,7 +756,12 @@ async def send_transaction(self, request):
         if not isinstance(request["amount"], int) or not isinstance(request["fee"], int):
             raise ValueError("An integer amount or fee is required (too many decimals)")
         amount: uint64 = uint64(request["amount"])
-        puzzle_hash: bytes32 = decode_puzzle_hash(request["address"])
+        address = request["address"]
+        selected_network = self.service.config["selected_network"]
+        expected_prefix = self.service.config["network_overrides"]["config"][selected_network]["address_prefix"]
```
mariano54, line 761: Can we have a method to get the prefix or store it in the WSM object?

mariano54:
This is fine, but it only covers one case. There are many other methods here which take in addresses which we should also update. It might be best to just change the method `decode_puzzle_hash` to take in prefix?

PR #10568, opened 2022-03-06
Add configurable request size limit for data layer service

altendky:


emlowe (dismissed):
Looks fine, but I don't think chia-blockchain-gui and mozille-ca submodules should be included in here?

altendky:
*#@%&$@*!!!

emlowe (approved):

PR #10570, opened 2022-03-07
update dropped_tx metrics

0xJohnZW:
the origin type of dropped_tx is set, and there is no code to decrease the size of it

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -134,7 +134,7 @@ class FullNode:
         self.multiprocessing_context = None
 
         # Used for metrics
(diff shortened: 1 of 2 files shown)


emlowe:
Not sure this is proper as it changes rather significantly what `dropped_tx` is - however, it is true that dropped_tx would only grow as the node runs. The transaction queue is rarely full, however, so this should be infrequent that we add to this set. It doesn't seem like `dropped_tx` is used anywhere else though. Perhaps @mariano54 can comment.

arvidn (approved):
it also seems there is no code that ever inspects this variable. we just increment it but never read it..

PR #10571, opened 2022-03-07
fix [Bug] #10569

### LINKED ISSUE #10569: [Bug] Wallet RPC get_offers_count return an error when there are no trades
### What happened?

Calling the wallet rpc method `get_offers_count` in a wallet without trades (it is still syncing) returned this error `int() argument must be a string, a bytes-like object or a number, not 'NoneType'`

(Where I indicate CLI below, I mean from my own code)

The error comes from (in main as of right now) _trade_store.py_ line 223.

        return int(row[0]), int(row[1]), int(row[2])

this previous check does not appear effective:

```
        if row is None:
            return 0, 0, 0
```

Also, there is no call to this rpc endpoint anywhere in the codebase, including unit tests. 

### Version

1.2.12.dev413

Linux

CLI

### Relevant log output

```shell
2022-03-06T20:25:59.545 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "/home/don/src/github/Chia-Network/chia-blockchain/chia/rpc/rpc_server.py", line 215, in safe_handle
    response = await self.ws_api(message)
  File "/home/don/src/github/Chia-Network/chia-blockchain/chia/rpc/rpc_server.py", line 206, in ws_api
    return await f(data)
  File "/home/don/src/github/Chia-Network/chia-blockchain/chia/rpc/wallet_rpc_api.py", line 995, in get_offers_count
    (total, my_offers_count, taken_offers_count) = await trade_mgr.trade_store.get_trades_count()
  File "/home/don/src/github/Chia-Network/chia-blockchain/chia/wallet/trading/trade_store.py", line 223, in get_trades_count
    return int(row[0]), int(row[1]), int(row[2])
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

dkackman:
This fixes https://github.com/Chia-Network/chia-blockchain/issues/10569 by checking that none of the fields returned from the query are `None` before casting them to `int`.

Includes an unrelated Black reformat. 

I am not sure how to recreate the db condition in the test environment, so there is no unit test included. Happy to do so but would need some guidance. Otherwise it looks, via code inspection, to be a pretty safe change.

### DIFF
--- a/chia/wallet/trading/trade_store.py
+++ b/chia/wallet/trading/trade_store.py
@@ -32,10 +32,7 @@ async def migrate_is_my_offer(log: logging.Logger, db_connection: aiosqlite.Conn
         updates.append((is_my_offer, row[1]))
 
     try:
-        await db_connection.executemany(
-            "UPDATE trade_records SET is_my_offer=? WHERE trade_id=?",
-            updates,
-        )
+        await db_connection.executemany("UPDATE trade_records SET is_my_offer=? WHERE trade_id=?", updates)
     except (aiosqlite.OperationalError, aiosqlite.IntegrityError):
         log.exception("Failed to migrate is_my_offer property in trade_records")
         raise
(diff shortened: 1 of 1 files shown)


wjblanke (approved):
aok

wjblanke:
thanks!

PR #10573, opened 2022-03-07
streamable|pools: Fix `Optional` parsing in `dataclass_from_dict`

xdustinface:
35b645b9cf97b6430d3f6bedb28dcf83894e067e Fixes a bug which leads to `dataclass_from_dict` setting values to `None` even though they are actual values like `""` or `False` or `[]` because the current code asks for `if not d` instead of `if d is None`. You can see the tests added by 39bd473bbab61e46af1eda5d81e947e736eda38b failing without 35b645b9cf97b6430d3f6bedb28dcf83894e067e.

Note that i had to adjust the pool wallet / pool tests here 6822e27ff10291d74d3f463a1f05c1cee6a09984. In general i think we should have a URL validator or so instead of just checking for `""`/`None` in all those places but thats a separate issue.

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -127,7 +127,7 @@ class PoolWallet:
     @classmethod
     def _verify_self_pooled(cls, state) -> Optional[str]:
         err = ""
-        if state.pool_url != "":
+        if state.pool_url not in [None, ""]:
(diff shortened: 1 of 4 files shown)


mariano54 (dismissed):

xdustinface:
Rebased after #10577

mariano54 (approved):

PR #10574, opened 2022-03-07
Re create process pool executor for each weight proof

mariano54:


mariano54:
I will take a different approach

PR #10611, opened 2022-03-09
Remove websockets dependency & do some refactoring

jack60612:
replaces #6734 

Also a mild clean up


mariano54:
Has the Unclosed client session error been fixed in RPC server?

jack60612:
> Has the Unclosed client session error been fixed in RPC server?

it has now been fixed

chia/rpc/rpc_server.py
```diff
@@ -259,30 +258,30 @@ async def connection(self, ws):
...
-                        heartbeat=60,
-                        ssl_context=self.ssl_context,
-                        max_msg_size=100 * 1024 * 1024,
-                    ) as ws:
-                        self.websocket = ws
-                        await self.connection(ws)
-                    self.websocket = None
+                self.client_session = aiohttp.ClientSession()
+                self.websocket = await self.client_session.ws_connect(
+                    f"wss://{self_hostname}:{daemon_port}",
```
emlowe, line 266: Wonder why you changed this code as I think the "with" pattern is preferred in general as cleanup is handled for you

chia/daemon/client.py
```diff
@@ -46,7 +61,9 @@ async def _get(self, request: WsRpcMessage) -> WsRpcMessage:
         request_id = request["request_id"]
         self._request_dict[request_id] = asyncio.Event()
         string = dict_to_json_str(request)
-        asyncio.create_task(self.websocket.send(string))
+        if self.websocket is None:
```
emlowe, line 64: Won't the `asyncio.create_task(self.websocket.send(string))` throw anyway in this case?

chia/daemon/server.py
```diff
@@ -198,52 +205,48 @@ def cancel_task_safe(self, task: Optional[asyncio.Task]):
...
-                self.log.info(f"ConnectionClosedOk. Closing websocket with {service_name} {e}")
-            elif isinstance(e, WebSocketException):
-                self.log.info(f"Websocket exception. Closing websocket with {service_name} {e} {tb}")
-            else:
-                self.log.error(f"Unexpected exception in websocket: {e} {tb}")
-        finally:
-            self.remove_connection(websocket)
-            await websocket.close()
+    async def incoming_connection(self, request):
+        ws: WebSocketResponse = web.WebSocketResponse(max_msg_size=self.daemon_max_message_size, heartbeat=30)
```
emlowe, line 215: Does this `heartbeat=30` override the `heartbeat=60` specified in either "connect_to_daemon" or the `daemon/client.py`

chia/daemon/client.py
mariano54: why would it throw?

chia/daemon/server.py
mariano54: This is replacing the `ping` task that ws being manually executed every 30 seconds, do it does not change the timeout from the daemon side.
The RPC service side is still at 60. I suppose it's fine to have different ping times for each side of the socket.

chia/rpc/rpc_server.py
mariano54: We were having some issues with `with`, I believe it might be related to saving the object to the class in with, and reusing it later.

mariano54: This also potentially fixes the annoying error logging when shutting down the wallet.

chia/daemon/client.py
```diff
@@ -18,26 +18,41 @@ def __init__(self, uri: str, ssl_context: Optional[ssl.SSLContext]):
...
-        self.websocket = await websockets.connect(self._uri, max_size=None, ssl=self.ssl_context)
+        try:
+            self.client_session = aiohttp.ClientSession()
+            self.websocket = await self.client_session.ws_connect(
+                self._uri,
+                autoclose=True,
+                autoping=True,
+                heartbeat=60,
+                ssl_context=self.ssl_context,
+                max_msg_size=100 * 1024 * 1024,
```
mariano54, line 37: We should probably hardcode this setting like at the top of the file with a constant.

chia/daemon/server.py
mariano54: Maybe we should hardcode these constants at the top of the file as well (heartbeat, keepalive timeout etc)

mariano54 (changes requested):
I tried stopping and restarting the wallet, but it won't let me restart because it thinks the wallet is still running. To replicate:

```
chia start wallet
chia stop wallet
chia start wallet
```

jack60612:
> I tried stopping and restarting the wallet, but it won't let me restart because it thinks the wallet is still running. To replicate:
> 
> 
> 
> ```
> 
> chia start wallet
> 
> chia stop wallet
> 
> chia start wallet
> 
> ```

I tried that before pushing, strange

chia/rpc/rpc_server.py
jack60612: Yeah it seems to fix the not closed error, and I think it looks better for this use case

chia/daemon/client.py
emlowe: if `self.websocket is none` - then I would expect `asyncio.create_task(self.websocket.send(string))` to throw so one approach is to not bother with the "if" check and just let it throw. Which I assume the previous code does.

chia/daemon/server.py
emlowe: Oh right, I didn't notice that ping task - nice to get rid of that.

chia/daemon/client.py
jack60612: I'll just do that.

jack60612: correction, one of the linters wants it.

jack60612:
100% working.

jack60612:
All 100% working on my end, should be ready.

chia/daemon/server.py
```diff
@@ -656,18 +634,18 @@ def plot_queue_to_payload(self, plot_queue_item, send_full_log: bool) -> Dict[st
             item["log"] = plot_queue_item.get("log")
         return item
 
-    def prepare_plot_state_message(self, state: PlotEvent, id):
+    def prepare_plot_state_message(self, state: PlotEvent, plot_task_id):
```
emlowe, line 637: This seems unrelated - we should remove from this PR?

chia/daemon/server.py
```diff
@@ -745,10 +725,7 @@ def _common_plotting_command_args(self, request: Any, ignoreCount: bool) -> List
         p = request.get("p")  # Pool pubkey
         c = request.get("c")  # Pool contract address
 
-        command_args: List[str] = []
-        command_args.append(f"-n{n}")
-        command_args.append(f"-d{d}")
-        command_args.append(f"-r{r}")
+        command_args: List[str] = [f"-n{n}", f"-d{d}", f"-r{r}"]
```
emlowe, line 728: more unrelated changes

chia/daemon/server.py
jack60612: Me and mariano did those, having a variable named id can screw up python stuff and can cause errors.

emlowe: Please do it in another PR; it may be a good idea, it's just wholly unrelated to this PR

jack60612: > Please do it in another PR; it may be a good idea, it's just wholly unrelated to this PR

ok, time for pain.

hoffmang9: > ok, time for pain.

Builds character.

jack60612: done

chia/daemon/server.py
jack60612: fixed

mariano54 (commented):
In general I think this is a good change. maybe @emlowe or @Yostra wants to review first before merging. There are not a lot of automated tests of daemon so this requires some manual testing

jack60612:
> It's working for me, but I'm still getting `Unclosed client session` when doing:
> 
> `chia stop wallet`.
> 
> 
> 
> 

I'll check that again, thought I solved that

chia/daemon/server.py
```diff
@@ -1148,22 +1126,18 @@ async def is_running(self, request: Dict[str, Any]) -> Dict[str, Any]:
...
         for k in self.services.keys():
             jobs.append(kill_service(self.root_path, self.services, k))
         if jobs:
             await asyncio.wait(jobs)
         self.services.clear()
+        if self.websocket_runner is not None:
+            await self.websocket_runner.cleanup()
+        self.shutdown_event.set()
 
-        # TODO: fix this hack
```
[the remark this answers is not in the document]
jack60612: yeah basically, the new WebSocket library does it differently so this old way does not work.

chia/daemon/server.py
```diff
@@ -228,28 +230,27 @@ async def safe_handle(self, websocket: WebSocketServerProtocol, path: str):
...
-        finally:
-            self.remove_connection(websocket)
-            await websocket.close()
+                service_name = "Unknown"
+                if ws in self.remote_address_map:
+                    service_name = self.remote_address_map[ws]
+                if msg.type == WSMsgType.CLOSE:
+                    self.log.info(f"ConnectionClosed. Closing websocket with {service_name}")
+                elif msg.type == WSMsgType.ERROR:
+                    print(f"Websocket exception. Closing websocket with {service_name}. {ws.exception()}")
```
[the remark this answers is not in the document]
jack60612: oops

chia/daemon/server.py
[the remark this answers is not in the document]
jack60612: i made a whole new way which fixes the errors we were encountering before.

chia/daemon/server.py
```diff
@@ -184,16 +183,19 @@ def master_close_cb():
...
+        app = web.Application(client_max_size=self.daemon_max_message_size)
+        app.add_routes([web.get("/", self.incoming_connection)])
+        self.websocket_runner = web.AppRunner(app, access_log=None, logger=self.log, keepalive_timeout=300)
+        await self.websocket_runner.setup()
+
+        site = web.TCPSite(
+            self.websocket_runner,
+            host=self.self_hostname,
+            port=self.daemon_port,
+            shutdown_timeout=3,
```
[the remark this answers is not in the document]
jack60612: hmm, looking at the original pr.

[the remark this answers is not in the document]
jack60612: its the amount of time that the server will wait for all connections to stop peacefully during shutdown

mariano54:
can we update / rebase again @jack60612 ?

jack60612:
> can we update / rebase again @jack60612 ?

Done

mariano54 (approved):

PR #10613, opened 2022-03-09
Ak.do spend

aqk:
Merge after https://github.com/Chia-Network/chia-blockchain/pull/10612

### DIFF
--- a/tests/wallet/cat_wallet/test_cat_lifecycle.py
+++ b/tests/wallet/cat_wallet/test_cat_lifecycle.py
@@ -44,64 +44,64 @@ async def setup_sim():
     return sim, sim_client
 
 
+async def do_spend(
+    sim: SpendSim,
+    sim_client: SimClient,
+    tail: Program,
+    coins: List[Coin],
+    lineage_proofs: List[Program],
+    inner_solutions: List[Program],
+    expected_result: Tuple[MempoolInclusionStatus, Err],
+    reveal_limitations_program: bool = True,
+    signatures: List[G2Element] = [],
+    extra_deltas: Optional[List[int]] = None,
+    additional_spends: List[SpendBundle] = [],
+    limitations_solutions: Optional[List[Program]] = None,
+) -> int:
+    if limitations_solutions is None:
+        limitations_solutions = [Program.to([])] * len(coins)
+    if extra_deltas is None:
+        extra_deltas = [0] * len(coins)
+
+    spendable_cat_list: List[SpendableCAT] = []
+    for coin, innersol, proof, limitations_solution, extra_delta in zip(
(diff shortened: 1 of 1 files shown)


tests/core/cmds/test_keys.py
```diff
@@ -65,50 +65,55 @@ def delete_password(self, service, username):
...
+
+    KeyringWrapper.cleanup_shared_instance()
+    KeyringWrapper.set_keys_root_path(tmp_path)
+    KeyringWrapper.get_shared_instance().legacy_keyring = DummyLegacyKeyring()
+    yield
+    KeyringWrapper.cleanup_shared_instance()
+    KeyringWrapper.set_keys_root_path(DEFAULT_KEYS_ROOT_PATH)
+
+
+class TestKeysCommands:
```
arvidn, line 116: technically, I don't think you needed to add back the `class`. the tests could be free functions

tests/wallet/cat_wallet/test_cat_lifecycle.py
```diff
@@ -36,71 +36,72 @@
...
-        await sim.farm_block()
+    )
+    assert result == expected_result
+    cost = cost_of_spend_bundle(spend_bundle)
+    await sim.farm_block()
+    return cost
+
+
+class TestCATLifecycle:
+    cost: Dict[str, int] = {}
```
arvidn, line 104: this looks scary. Do tests share state here? does order matter? what happens if you run different tests in different processes?

arvidn (dismissed):
I definitely think this is an improvement. I think we could take it even further (later) to remove all of those `Test*` classes. If there are test classes that actually have state, it seems a bit suspicious and should probably be reviewed

tests/core/cmds/test_keys.py
aqk: Yeah, trying to do as little as possible in each of these

tests/wallet/cat_wallet/test_cat_lifecycle.py
aqk: I had to carefully read this to make sure - `cost` is never used to produce test output - it just accumulates the results.

By the way, the very last test that displays the results .. I'm not sure that pytest guarantees the order the tests are run in, so that may not be correct (it doesn't compare, it just prints the stored results)

arvidn: I see. we should probably just make the tests themselves print the result. or assert that it is what we expect it to be

arvidn (approved):

tests/wallet/cat_wallet/test_cat_lifecycle.py
aqk: Ya, I'll Ping Matt Hauff @quexington

PR #10616, opened 2022-03-09
Ak.setup node and wallet

aqk:


aqk:
This code was merged in #10614

PR #10617, opened 2022-03-09
Ak.345 nodes

aqk:


tests/conftest.py
```diff
@@ -128,3 +130,27 @@ def default_10000_blocks_compact(bt):
 def tmp_dir():
     with tempfile.TemporaryDirectory() as folder:
         yield Path(folder)
+
+
+@pytest_asyncio.fixture(scope="function")
+async def wallet_node(self_hostname):
+    async for _ in setup_node_and_wallet(test_constants, self_hostname):
+        yield _
```
altendky, line 138: Is this supposed to be added here?

tests/core/test_crawler_rpc.py
```diff
@@ -1,26 +1,12 @@
-import atexit
```
altendky, line 1: This is covered in https://github.com/Chia-Network/chia-blockchain/pull/10614.

tests/conftest.py
aqk: Not, I think that is a merge mistake on my part, thanks!

tests/core/test_crawler_rpc.py
aqk: This is on top of #10614. Let me see if I can separate them.

PR #10618, opened 2022-03-09
If a v2 DB already exists

cmmarslender:
Check if a database file already exists in the `db` folder and don't re-run the database commands, as they will fail in this case.

This is causing some of our automation to fail, since we download a backup of the database on new nodes. We can't download after init, since the automation will think the database already exists, since `init` creates the DB file.


arvidn:
I made the same change here: https://github.com/Chia-Network/chia-blockchain/pull/10498
With that change this check has to happen in two places.

cmmarslender:
Ok thanks. i'll close this PR in favor of yours

PR #10621, opened 2022-03-09
improve listen port selection for tests

arvidn:
a problem with using random.randint() to pick a port is that all processes (running in parallel) are seeded the same, and so pick the same ports at the same time, causing conflicts. This uses proper entropy instead.

### DIFF
--- a/tests/util/socket.py
+++ b/tests/util/socket.py
@@ -1,15 +1,18 @@
 import random
+import secrets
 import socket
 from typing import Set
 
 recent_ports: Set[int] = set()
+prng = random.Random()
+prng.seed(secrets.randbits(32))
 
 
 def find_available_listen_port(name: str = "free") -> int:
     global recent_ports
 
     while True:
-        port = random.randint(2000, 65535)
+        port = prng.randint(2000, 65535)
         if port in recent_ports:
             continue
 
@@ -17,6 +20,7 @@ def find_available_listen_port(name: str = "free") -> int:
             try:
                 s.bind(("127.0.0.1", port))
             except OSError:
+                recent_ports.add(port)
                 continue
 
         recent_ports.add(port)


altendky:
I don't think we need cryptographic grade randomness for port selection?  Could just seed with the time.  Not that I think it super matters.  Might be worth knowing exactly what stage the processes are apparently forked at though.

altendky:
I guess using a local `random.Random` and seeding with the time would play nicely with my desired optional deterministic selection.

arvidn:
seeding with time would not work, because the processes are started at the same time

arvidn:
seeding with the process ID would work. What I worry about then is that other aspects of tests may benefit from the random generator being deterministic.

arvidn:
> You can create an instance of random.Random for individual use and seed it instead of the global instance accessed through the random.seed() module functions

Yeah, that's probably better, at least with the feature of picking ports deterministically. It's trivial to change along with that feature though, and I'm not sure it's worth the extra CI cycle to change now.

tests/util/socket.py
```diff
@@ -1,22 +1,26 @@
...
+prng = random.Random()
+prng.seed(secrets.randbits(32))
 
 
 def find_available_listen_port(name: str = "free") -> int:
     global recent_ports
 
     while True:
-        port = random.randint(2000, 65535)
+        port = prng.randint(2000, 65535)
```
aqk, line 15: 2000 - conservative

aqk (approved):

arvidn:
I actually think the first version of this patch was simpler and more reliable. as kyle pointed out, in case pytest-xdist initializes globals first and then forks, keeping any global state is risky. But, this appears to work, and I don't think it's a problem worth spending more time on.

PR #10630, opened 2022-03-09
Fix timelord closing.

fchirica:
Attempt to fix the infinite loop `Trying to stop {chain} before its initialization.` if we try to close a chain that's not in `self.allow_iters`.

### DIFF
--- a/chia/timelord/timelord.py
+++ b/chia/timelord/timelord.py
@@ -183,25 +183,22 @@ class Timelord:
 
     async def _stop_chain(self, chain: Chain):
         try:
-            while chain not in self.allows_iters:
-                self.lock.release()
-                await asyncio.sleep(0.05)
-                log.error(f"Trying to stop {chain} before its initialization.")
-                await self.lock.acquire()
-                if chain not in self.chain_type_to_stream:
-                    log.warning(f"Trying to stop a crashed chain: {chain}.")
-                    return None
-            stop_ip, _, stop_writer = self.chain_type_to_stream[chain]
-            stop_writer.write(b"010")
-            await stop_writer.drain()
+            _, _, stop_writer = self.chain_type_to_stream[chain]
             if chain in self.allows_iters:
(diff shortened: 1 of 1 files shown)


hoffmang9 (approved):
lgtm

hoffmang9:
This still needs a look @mariano54

chia/timelord/timelord.py
```diff
@@ -183,25 +183,22 @@ async def _handle_client(self, reader: asyncio.StreamReader, writer: asyncio.Str
 
     async def _stop_chain(self, chain: Chain):
         try:
-            while chain not in self.allows_iters:
```
mariano54, line 186: Why are we removing the while loop?

mariano54 (commented):
I don't really understand this change & this code TBH. We should make sure that when we deploy, we have some old timelords running as well.

chia/timelord/timelord.py
```diff
@@ -183,25 +183,22 @@ async def _handle_client(self, reader: asyncio.StreamReader, writer: asyncio.Str
...
                 self.allows_iters.remove(chain)
+            else:
+                log.error(f"Trying to stop {chain} before its initialization.")
+                stop_writer.close()
+                await stop_writer.wait_closed()
             if chain not in self.unspawned_chains:
                 self.unspawned_chains.append(chain)
-            if chain in self.chain_type_to_stream:
-                del self.chain_type_to_stream[chain]
+            del self.chain_type_to_stream[chain]
```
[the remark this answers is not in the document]
fchirica: it should always be included at this point - in case it's not we have no writer to close, it'll just log the error (but I don't think it ever gets to this point).

chia/timelord/timelord.py
fchirica: to avoid the infinite loop that happens in rare cases (written in the description).

mariano54 (approved):

PR #10631, opened 2022-03-09
Safe config file reading and writing

### LINKED ISSUE #10585: [Bug] Pool NFT in GUI shows "Self Pooling" and joined to na1.pool.space after update.  Both are wrong
### What happened?

Updated to 1.3.0, noticed my my pool info was incorrect.

Checked in CLI and the pool NFT shows as self pooling but not pointed to space pool.

Checked the config.yaml and indeed it showed pointed to spacepool as well but this is different from my backup config.yaml (made before the update) which showed the correct information for flexpool

Removed all pool info from the config.yaml, deleted all of my wallet DBs and resynced.  After resync, the pool showed correctly in GUI and CLI

### Version

1.3.0

Windows

GUI

### Relevant log output

_No response_

mariano54:
Config writes need to be atomic.
On windows, `os.replace`, which is atomic, sometimes throws a `PermissionError`. Then we fall back to `shutil.move` which copies the file on error. However copy is not atomix, and thus there can be conflicts if another file is reading the config. 

I believe that we might have encountered this issue on Windows on 1.3, since farmer and wallet both write to the config file. 
This PR does a retry if read fails. It also uses a new tmp directory for every write, just in case. 

This fix does seem to help the users that had plotNFT syncing issues.

* load_config by default uses the Lock, to keep the PR simple, since we use load_config a lot. However, if you want to also save after, you should use the lock around `load_config` and `save_config`, and pass in `acquire_lock=False` to `load_config`.
* test_config.py was not throwing on exceptions, the exceptions were being eaten by the threads.
* NOTE: a lot of the diff changes are just tab under the lock context, so view with Hide whitespace changes toggled.

Next steps:
- Make a config locking PR - https://github.com/Chia-Network/chia-blockchain/pull/10680
- Make a keyring PR

### DIFF
--- a/chia/cmds/chia.py
+++ b/chia/cmds/chia.py
@@ -16,6 +16,7 @@ from chia.cmds.wallet import wallet_cmd
 from chia.cmds.plotnft import plotnft_cmd
 from chia.cmds.plotters import plotters_cmd
 from chia.cmds.db import db_cmd
+from chia.util.config import config_path_for_filename
 from chia.util.default_root import DEFAULT_KEYS_ROOT_PATH, DEFAULT_ROOT_PATH
 from chia.util.keychain import (
     Keychain,
@@ -91,7 +92,8 @@ def cli(
         except Exception as e:
             print(f"Failed to read passphrase: {e}")
 
-    check_ssl(Path(root_path))
+    if config_path_for_filename(Path(root_path), "config.yaml").exists():
+        check_ssl(Path(root_path))
 
 
 if not supports_keyring_passphrase():
--- a/chia/util/config.py
+++ b/chia/util/config.py
@@ -51,7 +51,7 @@ def get_config_lock(root_path: Path, filename: Union[str, Path]) -> BaseFileLock
     lock_path: Path = config_path_for_filename(root_path, filename).with_suffix(".lock")
     if not lock_path.exists():
         lock_path.touch(exist_ok=True)
-    return FileLock(str(lock_path.absolute()))
+    return FileLock(lock_path)  # type: ignore[arg-type]
 
 
 def save_config(root_path: Path, filename: Union[str, Path], config_data: Any):
--- a/chia/util/config.py
+++ b/chia/util/config.py
@@ -49,7 +49,7 @@ def config_path_for_filename(root_path: Path, filename: Union[str, Path]) -> Pat
 def save_config(root_path: Path, filename: Union[str, Path], config_data: Any):
     path: Path = config_path_for_filename(root_path, filename)
     with tempfile.TemporaryDirectory(dir=path.parent) as tmp_dir:
-        tmp_path: Path = Path(tmp_dir) / (filename + str(os.getpid()))
+        tmp_path: Path = Path(tmp_dir) / (Path(filename).with_suffix("." + str(os.getpid())))
         with open(tmp_path, "w") as f:
             yaml.safe_dump(config_data, f)
         try:
--- a/chia/util/config.py
+++ b/chia/util/config.py
@@ -14,6 +14,7 @@ import yaml
 from typing_extensions import Literal
 
 from chia.util.path import mkdir
+from filelock import Timeout, FileLock
 
 PEER_DB_PATH_KEY_DEPRECATED = "peer_db_path"  # replaced by "peers_file_path"
 WALLET_PEERS_PATH_KEY_DEPRECATED = "wallet_peers_path"  # replaced by "wallet_peers_file_path"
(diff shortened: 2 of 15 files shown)


chia/util/config.py
```diff
@@ -66,10 +72,20 @@ def load_config(
...
+            r = yaml.safe_load(open(path, "r"))
+            if r is None:
+                raise RuntimeError(f"Yaml load on {path} returned None")
+            if sub_config is not None:
+                r = r.get(sub_config)
+            return r
+        except Exception as e:
+            tb = traceback.format_exc()
+            log.error(f"Error loading file: {tb} {e} Retrying {i}")
+            time.sleep(i * 0.1)
```
paninaro, line 87: Is the increased back-off intentional, or was it meant to be a constant delay?

chia/wallet/util/wallet_sync_utils.py
```diff
@@ -41,7 +41,7 @@ async def fetch_last_tx_from_peer(height: uint32, peer: WSChiaConnection) -> Opt
         if response is not None and isinstance(response, RespondBlockHeader):
             if response.header_block.is_transaction_block:
                 return response.header_block
-        else:
+        elif request_height < height:
```
paninaro, line 44: Is this change supposed to be part of this PR?

chia/util/config.py
```diff
@@ -43,13 +48,14 @@ def config_path_for_filename(root_path: Path, filename: Union[str, Path]) -> Pat
...
-        os.replace(str(tmp_path), path)
-    except PermissionError:
-        shutil.move(str(tmp_path), str(path))
+    with tempfile.TemporaryDirectory(dir=path.parent) as tmp_dir:
+        tmp_path: Path = Path(tmp_dir) / (Path(filename).with_suffix("." + str(os.getpid())))
+        with open(tmp_path, "w") as f:
+            yaml.safe_dump(config_data, f)
+        try:
+            os.replace(str(tmp_path), path)
+        except PermissionError:
```
paninaro, line 57: Interestingly, the docs for `tempFile.TemporaryDirectory` mention PermissionError being raised on Windows if the file is already open:

"The directory can be explicitly cleaned up by calling the cleanup() method. If ignore_cleanup_errors is true, any unhandled exceptions during explicit or implicit cleanup (such as a [PermissionError](https://docs.python.org/3/library/exceptions.html#PermissionError) removing open files on Windows) will be ignored, and the remaining removable items deleted on a “best-effort” basis. Otherwise, errors will be raised in whatever context cleanup occurs (the cleanup() call, exiting the context manager, when the object is garbage-collected or during interpreter shutdown)."

I wonder if PermissionError is being seen more commonly now that the config is being (opened and) read more frequently. If this change doesn't completely fix the issue, maybe we should consider the same type of retry loop that you've implemented in `load_config`.

paninaro (dismissed):

chia/util/config.py
mariano54: it's intentional.

mariano54: it's intentional.

chia/wallet/util/wallet_sync_utils.py
mariano54: Oh i have another pr for that, i will revert, thanks

mariano54: Oh i have another pr for that, i will revert, thanks

paninaro (dismissed):

altendky (changes requested):
I think this probably just has to become an async activity and really should change from just write atomicity to read-through-write atomicity, maybe via a lock file.

chia/util/config.py
```diff
@@ -43,13 +49,15 @@ def config_path_for_filename(root_path: Path, filename: Union[str, Path]) -> Pat
...
 def save_config(root_path: Path, filename: Union[str, Path], config_data: Any):
     path: Path = config_path_for_filename(root_path, filename)
-    tmp_path: Path = path.with_suffix("." + str(os.getpid()))
-    with open(tmp_path, "w") as f:
-        yaml.safe_dump(config_data, f)
-    try:
-        os.replace(str(tmp_path), path)
-    except PermissionError:
-        shutil.move(str(tmp_path), str(path))
+    with FileLock(path.with_suffix(".lock")):
```
emlowe, line 52: I like the idea of using filelock - I believe the default is this will never timeout - are we ok with that?

mariano54: I think that should be fine

paninaro:
Since we don't have CI running tests on Windows, we should have a manual run to verify that the FileLock module works as expected... I seem to recall the fasteners module not working quite as expected on Windows.

mariano54:
q_size should be replaced with .empty()
There is still a test failing in keyring can you check @paninaro ? It might be related to not having read locks, and every lock being equivalent with the new library. I switched libraries because the old one didn't support locking within the same process (only cross process)

chia/cmds/chia.py
```diff
@@ -91,7 +92,8 @@ def cli(
         except Exception as e:
             print(f"Failed to read passphrase: {e}")
 
-    check_ssl(Path(root_path))
+    if config_path_for_filename(Path(root_path), "config.yaml").exists():
```
paninaro, line 95: Without this check, `chia version` (or similar commands) will fail on a fresh install (no `~/.chia` dir)

chia/util/config.py
```diff
@@ -41,35 +47,72 @@ def config_path_for_filename(root_path: Path, filename: Union[str, Path]) -> Pat
...
 
+def get_config_lock(root_path: Path, filename: Union[str, Path]) -> BaseFileLock:
+    lock_path: Path = config_path_for_filename(root_path, filename).with_suffix(".lock")
+    if not lock_path.exists():
+        lock_path.touch(exist_ok=True)
+    return FileLock(lock_path)
+
+
 def save_config(root_path: Path, filename: Union[str, Path], config_data: Any):
+    # This must be called under an acquired config lock
```
paninaro, line 58: Can we require that save_config() has a 'lock' object of some sort passed-in to prove that the lock was acquired?

PR #10633, opened 2022-03-09
Increases the probability of connecting to local trusted node

mariano54:
Especially when restarting services. There was a significant chance of connecting to an untrusted node and syncing to them first.

### DIFF
--- a/chia/wallet/util/wallet_sync_utils.py
+++ b/chia/wallet/util/wallet_sync_utils.py
@@ -41,7 +41,8 @@ async def fetch_last_tx_from_peer(height: uint32, peer: WSChiaConnection) -> Opt
         if response is not None and isinstance(response, RespondBlockHeader):
             if response.header_block.is_transaction_block:
                 return response.header_block
-        else:
+        elif request_height < height:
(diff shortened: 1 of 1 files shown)


chia/wallet/util/wallet_sync_utils.py
```diff
@@ -41,7 +41,8 @@ async def fetch_last_tx_from_peer(height: uint32, peer: WSChiaConnection) -> Opt
         if response is not None and isinstance(response, RespondBlockHeader):
             if response.header_block.is_transaction_block:
                 return response.header_block
-        else:
+        elif request_height < height:
```
aqk, line 44: Do you think we could get stuck in an infinite loop here? (Well, "infinite" for as long as the other peer is behind our current height)

mariano54: No because we do decrement the request height for every iteration of the loop, so as soon as we get either:
1. a tx block
2. a negative response

we break

aqk: Ah, yes, we fall through out of the if block in that case, thanks.

aqk (approved):

PR #10636, opened 2022-03-10
Better management of KeyringWrapper's keys_root_path when using TempKeyring for tests

paninaro:


tests/util/keyring.py
```diff
@@ -181,10 +182,19 @@ def _patch_and_create_keychain(
 
     def __enter__(self):
         assert not self.cleaned_up
-        return self.get_keychain()
+        if KeyringWrapper.get_shared_instance(create_if_necessary=False) is not None:
+            self.old_keys_root_path = KeyringWrapper.get_shared_instance().keys_root_path
```
aqk, line 186: Since this is an object variable, you could end up with a situation where we restore an earlier directory, when a new one has been set in the meantime.

eg.
`Set dir A
Create dir B
Create dir C
destroy dir B
Create dir D
`
Not that I'm encouraging building a proper stack for the tests :)

paninaro: That's a fair point. I can add an assert to at least notify us when we restore the former keys_root_path _IF_ the current keys_root_path isn't the value that we set.

That said, it shouldn't be the case that a new shared instance has been set in the middle of the test, now that we avoid setting the shared instance in our tests (from what I've seen). But yes, an assert will help validate that claim.

aqk (dismissed):

aqk (approved):

Issue #10628, opened 2022-03-09
[Bug] chia keys migrate crash


Ealrann:
### What happened?

The command `chia keys migrate` always crash. It happens on two different computers. Just do a new install to reproduce...

I guess it's no big deal, keys doesn't need to be migrated, but I didn't expect user command crash this way.

Step to reproduce:
1. Clone the repo
2. sh install.sh
3. . ./activate
4. chia init
5. chia keys generate
6. chia keys migrate

### Version

1.3.0

### What platform are you using?

Linux

### What ui mode are you using?

CLI

### Relevant log output

```shell
(venv) [chia@Field ~]$ chia keys migrate
Traceback (most recent call last):
  File "/home/chia/git/chia-blockchain/venv/bin/chia", line 33, in <module>
    sys.exit(load_entry_point('chia-blockchain', 'console_scripts', 'chia')())
  File "/home/chia/git/chia-blockchain/chia/cmds/chia.py", line 148, in main
    cli()  # pylint: disable=no-value-for-parameter
  File "/home/chia/git/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/chia/git/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/chia/git/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/chia/git/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/chia/git/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/chia/git/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/chia/git/chia-blockchain/venv/lib/python3.9/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/chia/git/chia-blockchain/chia/cmds/keys.py", line 144, in migrate_cmd
    migrate_keys()
[...]

PR #10655, opened 2022-03-10
Fix exception when `chia keys migrate` is run without needing migration

paninaro:
An exception was being raised when `chia keys migrate` was invoked without any keys needing migration. The code assumed that returning None from `__init__()` would result in the caller being able to test if object creation succeeded, but this didn't work as expected. Instead of returning `None`, forcing `KeyringWrapper` creation with a legacy keyring will raise `KeyringNotSet` so that the caller can handle the case where legacy keys are not present.

Issue #10628

### DIFF
--- a/chia/util/keychain.py
+++ b/chia/util/keychain.py
@@ -242,10 +242,15 @@ class Keychain:
     def __init__(self, user: Optional[str] = None, service: Optional[str] = None, force_legacy: bool = False):
         self.user = user if user is not None else default_keychain_user()
         self.service = service if service is not None else default_keychain_service()
-        if force_legacy:
-            self.keyring_wrapper = KeyringWrapper.get_legacy_instance()
-        else:
-            self.keyring_wrapper = KeyringWrapper.get_shared_instance()
+
+        keyring_wrapper: Optional[KeyringWrapper] = (
+            KeyringWrapper.get_legacy_instance() if force_legacy else KeyringWrapper.get_shared_instance()
+        )
+
+        if keyring_wrapper is None:
+            raise KeyringNotSet(f"KeyringWrapper not set: force_legacy={force_legacy}")
+
+        self.keyring_wrapper = keyring_wrapper
 
     @unlocks_keyring(use_passphrase_cache=True)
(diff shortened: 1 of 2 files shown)


chia/util/keychain.py
```diff
@@ -238,14 +242,15 @@ class Keychain:
...
-                return None
-        else:
-            self.keyring_wrapper = KeyringWrapper.get_shared_instance()
+
+        keyring_wrapper: Optional[KeyringWrapper] = (
+            KeyringWrapper.get_legacy_instance() if force_legacy else KeyringWrapper.get_shared_instance()
+        )
+
+        if keyring_wrapper is None:
+            raise KeyringNotSet(f"KeyringWrapper not set: force_legacy={force_legacy}")
```
paninaro, line 251: This case isn't actually hit today, but I included it to help surface potential issues with future changes.

chia/util/keychain.py
```diff
@@ -43,6 +43,10 @@ class KeyringMaxUnlockAttempts(Exception):
     pass
 
 
+class KeyringNotSet(Exception):
+    pass
```
paninaro, line 47: New exception type

chia/util/keychain.py
paninaro: The prior code assumed that returning `None` would help the caller detect an error condition. Since that doesn't work, raising an exception seems cleaner. The object isn't usable without a valid keyring_wrapper property.

chia/util/keychain.py
```diff
@@ -561,8 +566,10 @@ def migrate_legacy_keyring(
 
     @staticmethod
     def get_keys_needing_migration() -> Tuple[List[Tuple[PrivateKey, bytes]], Optional["Keychain"]]:
-        legacy_keyring: Optional[Keychain] = Keychain(force_legacy=True)
-        if legacy_keyring is None:
+        try:
+            legacy_keyring: Keychain = Keychain(force_legacy=True)
+        except KeyringNotSet:
+            # No legacy keyring available, so no keys need to be migrated
```
paninaro, line 572: Catch the exception instead of attempting to check for `None` (which doesn't work)

emlowe (approved):
looks ok to me - is there a test that can be added for this case?

PR #10676, opened 2022-03-11
Ak.setup two nodes

aqk:


paninaro (dismissed):
Looks good to me. There are of course lots of opportunities to de-dupe fixtures that are redefined in multiple places, but as a stepping stone to moving the fixtures out of class scope, this looks good (and CI seems to be green).

PR #10679, opened 2022-03-11
WIP: add cat_spend_multi RPC method

ftruzzi:
This would allow for one-click CAT airdrops by aggregating all spends in a single transaction.

Any initial feedback? Thanks!


Quexington:
I appreciate the change, but I actually would like to move _away_ from dedicated CAT RPCs rather than adding new ones.  I have a PR (that's been sitting around for a while due to the recent sensitivity of main) that would add cat functionality to the existing `send_transaction_multi` API https://github.com/Chia-Network/chia-blockchain/pull/10066

PR #10680, opened 2022-03-11
Add configuration locking

altendky:
Extracted from https://github.com/Chia-Network/chia-blockchain/pull/10631

### DIFF
--- a/chia/cmds/chia.py
+++ b/chia/cmds/chia.py
@@ -16,6 +16,7 @@ from chia.cmds.wallet import wallet_cmd
 from chia.cmds.plotnft import plotnft_cmd
 from chia.cmds.plotters import plotters_cmd
 from chia.cmds.db import db_cmd
+from chia.util.config import config_path_for_filename
 from chia.util.default_root import DEFAULT_KEYS_ROOT_PATH, DEFAULT_ROOT_PATH
 from chia.util.keychain import (
     Keychain,
@@ -91,7 +92,8 @@ def cli(
         except Exception as e:
             print(f"Failed to read passphrase: {e}")
 
-    check_ssl(Path(root_path))
+    if config_path_for_filename(Path(root_path), "config.yaml").exists():
+        check_ssl(Path(root_path))
 
 
 if not supports_keyring_passphrase():
--- a/chia/cmds/chia.py
+++ b/chia/cmds/chia.py
@@ -92,8 +92,7 @@ def cli(
         except Exception as e:
             print(f"Failed to read passphrase: {e}")
 
-    if config_path_for_filename(Path(root_path), "config.yaml").exists():
-        check_ssl(Path(root_path))
+    check_ssl(Path(root_path))
 
 
 if not supports_keyring_passphrase():
--- a/chia/cmds/chia.py
+++ b/chia/cmds/chia.py
@@ -16,7 +16,6 @@ from chia.cmds.wallet import wallet_cmd
 from chia.cmds.plotnft import plotnft_cmd
 from chia.cmds.plotters import plotters_cmd
 from chia.cmds.db import db_cmd
-from chia.util.config import config_path_for_filename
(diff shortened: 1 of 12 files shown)


chia/cmds/chia.py
```diff
@@ -91,7 +92,8 @@ def cli(
         except Exception as e:
             print(f"Failed to read passphrase: {e}")
 
-    check_ssl(Path(root_path))
+    if config_path_for_filename(Path(root_path), "config.yaml").exists():
```
paninaro, line 95: Without this check, `chia version` (or similar commands) will fail on a fresh install (no `~/.chia dir`)

chia/farmer/farmer.py
```diff
@@ -581,34 +581,36 @@ async def get_reward_targets(self, search_for_private_key: bool) -> Dict:
...
-                        if hexstr_to_bytes(list_element["launcher_id"]) == bytes(launcher_id):
-                            list_element["payout_instructions"] = payout_instructions
-                        new_list.append(list_element)
-
-                config["pool"]["pool_list"] = new_list
-                save_config(self._root_path, "config.yaml", config)
-                # Force a GET /farmer which triggers the PUT /farmer if it detects the changed instructions
-                pool_state_dict["next_farmer_update"] = 0
-                return
+                with get_config_lock(self._root_path, "config.yaml"):
```
paninaro, line 599: Seems like the lock should be acquired before the for loop on 597

chia/cmds/init_funcs.py
```diff
@@ -193,17 +196,19 @@ def migrate_from(
...
+        flattened_keys = unflatten_properties({k: "" for k in do_not_migrate_settings})
+        dict_add_new_default(config, default_config, flattened_keys)
 
-    create_all_ssl(new_root)
+        save_config(new_root, "config.yaml", config)
 
-    return 1
+        create_all_ssl(new_root)
+
+        return 1
```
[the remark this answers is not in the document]
altendky: Pending a double check it doesn't touch the config anywhere, agreed.

chia/farmer/farmer.py
altendky: Took a little inspection, but the body only runs once.  Good as is.

chia/util/config.py
```diff
@@ -41,35 +47,72 @@ def config_path_for_filename(root_path: Path, filename: Union[str, Path]) -> Pat
...
-    if sub_config is not None:
-        r = r.get(sub_config)
-    return r
+
+    config_lock: Optional[BaseFileLock] = None
+    if acquire_lock:
+        config_lock = get_config_lock(root_path, filename)
+        acquired = config_lock.acquire()
+        if not acquired:
+            raise RuntimeError("Not able to acquire lock in load_config")
```
altendky, line 86: I think this is a good case for an exit stack context manager, if the conditional acquisition is needed.  But, filelock specifically says it allows nesting which both means that this is unneeded and that the lock can't be used with async in the body.

paninaro (dismissed):
lgtm

chia/util/config.py
altendky: Silly mistake.  A _single_ lock object can be nested, separate lock objects on the same file cannot.

tests/core/util/test_config.py
```diff
@@ -20,47 +32,117 @@
...
+            else:
+                path: Path = config_path_for_filename(root_path, filename="config.yaml")
+                # print("Trying to get lock write")
+                with get_config_lock(root_path, "config.yaml"):
+                    # print("Write start")
+                    with tempfile.TemporaryDirectory(dir=path.parent) as tmp_dir:
+                        tmp_path: Path = Path(tmp_dir) / Path("config.yaml")
+                        with open(tmp_path, "w") as f:
+                            yaml.safe_dump(config, f)
+                        shutil.copy2(str(tmp_path), str(path))
```
altendky, line 70: This seems pretty similar to, but not quite the same as `save_config()`.  Should it be the same?  Should `save_config()` just be called here?

PR #9206, opened 2021-11-09
Improve message counts in test_too_many_messages()

altendky:
The existing numbers were insufficient to guarantee exceeding the limit in the case where the test spans a reset boundary for the rate limiting.  It would be nice to make the test agnostic to this alignment, but this should at least avoid (rare) flakes.

https://github.com/Chia-Network/chia-blockchain/runs/4154993776?check_suite_focus=true
```
=================================== FAILURES ===================================
____________________ TestRateLimits.test_too_many_messages _____________________

self = <test_rate_limits.TestRateLimits object at 0x7f1df617b7c0>

    @pytest.mark.asyncio
    async def test_too_many_messages(self):
        # Too many messages
        r = RateLimiter(incoming=True)
        new_tx_message = make_msg(ProtocolMessageTypes.new_transaction, bytes([1] * 40))
        for i in range(3000):
            assert r.process_msg_and_check(new_tx_message)
    
        saw_disconnect = False
        for i in range(3000):
            response = r.process_msg_and_check(new_tx_message)
            if not response:
                saw_disconnect = True
>       assert saw_disconnect
E       assert False

tests/core/server/test_rate_limits.py:34: AssertionError
```

Draft for:
- [ ] Addressing comments
- [ ] Considering https://github.com/Chia-Network/chia-blockchain/pull/8861#issuecomment-974702335

PR #10681, opened 2022-03-12
Add Windows CI

### LINKED ISSUE #11434: [Bug] Unexplored Windows CI failures
### What happened?

There are failures in the Windows CI PR that have not been attributed to things we don't care about.  Might be worth making sure we know what they are and whether they are relevant to 1.3.5.

https://github.com/Chia-Network/chia-blockchain/pull/10681

### Version

main

Windows

CLI

### Relevant log output

_No response_

altendky:
Draft for:
- [x] Self review and tidying
- [x] Presumably various issues to address
- [x] Check the Python version used in envs, seems to be 3.10, probably due to versionless `py` call in `Install.ps1`.
- [x] Revisit `--numprocesses 0` in case xdist is usable on Windows.
- [x] Windows server test failures.
  - https://github.com/aio-libs/aiohttp/issues/2934
- [x] Break out PRs:
  - https://github.com/Chia-Network/chia-blockchain/pull/10940
  - https://github.com/Chia-Network/chia-blockchain/pull/11117
- [ ] Maybe get timelord going...

Items to note.

The build workflow script currently sets `-n 0` for all windows tests and ignores the config `Parallel` setting. Windows tests do work locally quite well with multiple subprocesses, however, the GH runners stumble with 4 processes in some tests. Subsequent PRs can explore expanding the config to have a dict with per OS settings.

The IPv6 test is skipped (also skipped on macOS).


.github/workflows/build-test-windows-blockchain.yml
```diff
@@ -0,0 +1,100 @@
...
+
+    - name: Checkout test blocks and plots
+      uses: actions/checkout@v2
+      with:
+        repository: 'Chia-Network/test-cache'
+        path: '.chia'
+        ref: '0.28.0'
+        fetch-depth: 1
+
+    - name: Link home directory
```
arvidn, line 73: this step has unixy stuff in it. Is it necessary to do on windows?

.github/workflows/build-test-windows-blockchain.yml
```diff
@@ -0,0 +1,100 @@
...
+    - name: Run install script
+      env:
+        INSTALL_PYTHON_VERSION: ${{ matrix.python-version }}
+      run: |
+        ./Install.ps1
+
+    - name: Install timelord
+      run: |
+        . ./activate
+        sh install-timelord.sh
```
arvidn, line 89: I would expect this to be problematic on windows too.

.github/workflows/build-test-windows-blockchain.yml
altendky: I started to poke at alternatives over in https://github.com/Chia-Network/chia-blockchain/pull/10682.

tests/pools/test_pool_config.py
```diff
@@ -1,15 +1,16 @@
...
 from chia.pools.pool_config import PoolWalletConfig
-from chia.util.config import load_config, lock_config, save_config, create_default_chia_config
+from chia.util.config import create_default_chia_config, load_config, lock_config, save_config
 
 
 def test_pool_config():
-    test_root = Path("/tmp")
-    test_path = Path("/tmp/config")
+    test_root = Path(gettempdir())
+    test_path = test_root / "config"
```
altendky, line 13: This is a step forward, but tests should be getting a temporary directory from the `tmp_path` fixture.  This will provide a per-test temporary directory where all of those directories in a given run will be inside the pytest temporary directory for that run which are all inside the overall pytest temporary directory.  We definitely should never have put a hard coded file name directly in `/tmp`, nor even the system temp directory.

tests/core/server/test_dos.py
```diff
@@ -201,9 +201,15 @@ async def test_spam_tx(self, setup_two_nodes_fixture, self_hostname):
...
-        await asyncio.sleep(1)
+        try:
+            for i in range(6000):
+                await ws_con._send_message(new_tx_message)
+            await asyncio.sleep(1)
+        except ConnectionResetError:
+            # based on timing the connection may get closed by rate limits during the loop
+            # and this is actually want we want so catch this exception and ignore it
+            # seen primarily on my Windows box
+            pass
```
altendky, line 212: Typo in `want we want`.  The sleep could go outside as well, I think?

Why do we not get the exception everywhere else?  This seems like a good thing to understand a bit better.

altendky (commented):
Thanks for the work and for finding some of the issues here.

tests/core/server/test_dos.py
emlowe: The rate limits are x messages within x timeframe - we actually want it to fire sometime here in the loop. I kinda would have expected this exception in more places actually.

tests/core/test_farmer_harvester_rpc.py
```diff
@@ -435,7 +435,7 @@ def test_plot_matches_filter(filter_item: FilterItem, match: bool):
         ),
         (FarmerRpcClient.get_harvester_plots_invalid, [], None, True, 13),
         (FarmerRpcClient.get_harvester_plots_invalid, ["invalid_0"], None, False, 6),
-        (FarmerRpcClient.get_harvester_plots_invalid, ["inval", "lid_1/"], None, False, 2),
+        (FarmerRpcClient.get_harvester_plots_invalid, ["inval", "lid_1"], None, False, 2),
```
altendky, line 438: @xdustinface, is the need to remove the `/` here a sign of some lack of normalization of the paths or something?  @emlowe, could you comment on how this was failing?  Was it `/` vs. `\`?  or just no trailing slash to match at all?

emlowe: Yea, it's a string \ vs / - I dunno if there is a better way to build the test, but removing the / was what Kevin suggested

altendky: I'm definitely tempted to normalize them for comparison, though I haven't thought through corner cases.  `/` is generally valid all over Windows.  But, that would be a feature addition so not really required here.  But, might be worth xfailing this on Windows?  This suggestion needs at least an `import sys` added so apply this manually.

pytest.param(1, 3, marks=pytest.mark.xfail(reason="some bug")),
```suggestion
        pytest.param(
            FarmerRpcClient.get_harvester_plots_invalid,
            ["inval", "lid_1/"],
            None,
            False,
            2,
            marks=pytest.mark.xfail(condition=sys.platform == "win32", reason="not presently supported on Windows"),
        ),
        pytest.param(
            FarmerRpcClient.get_harvester_plots_invalid,
            ["inval", "lid_1\\"],
            None,
            False,
            2,
            marks=pytest.mark.skip(condition=sys.platform != "win32", reason="only for Windows"),
        ),
```

xdustinface: @altendky It's good enough to just drop the `/`. It's just filtering for the path there, not really important to test that path filtering for windows is with `\` and for others with `/` imo. I initially just added it to avoid flakiness with the temp directory names because i thought it might contain `lid_1` but well let's see how that goes?

altendky (commented):
I don't like all the `time_out_assert()` time increases, but maybe we just accept that and focus instead on https://github.com/Chia-Network/chia-blockchain/pull/11535 and then applying it more.

tests/wallet/test_wallet_retry.py
```diff
@@ -42,6 +44,10 @@ def evict_from_pool(node: FullNodeAPI, sb: SpendBundle) -> None:
 
 
 @pytest.mark.asyncio
+@pytest.mark.skipif(
+    condition=("GITHUB_ACTIONS" in os.environ) and (sys.platform in {"doodad"}),
+    reason="testing stuff",
+)
```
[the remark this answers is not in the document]
emlowe: oh, ya, I forgot to remove that !

tests/core/server/test_dos.py
mariano54: The connection does close, that is checked in line 217. So it must be one of two scenarios:
1. On other platforms, there is a delay when closing connection
2. On other platforms, you can still call ws_con.send_message on a closed WS.

I'm not sure which is the right one

altendky: https://github.com/Chia-Network/chia-blockchain/pull/11612

emlowe: I was thinking of the tests could be rewritten a little bit - we can use the code in rate_limits.py to use the exact number of messages that should send it over the limit - rather than this rather obscure 6000. I believe the frequency limit for NewTransaction is 5000. This is also testing the inbound limit, so the other end needs to close the connection - I suspect Windows ends up closing the connection much faster than Unix - so my guess is it's (1). That said, I'm not sure how much it matters to track it down exactly.

mariano54:
I think we should consider using -n 4 and just commenting out the tests which don't work (or attempt to fix them).
I rather have no windows tests and 20min CI, than windows tests and 40min CI.

tests/core/server/test_dos.py
mariano54: I think with kyle's fix we are good, there is no need to spend more time on this

altendky: the tests are flaky anyways.  unless mariano ended up handling this.  i had started working on that in https://github.com/Chia-Network/chia-blockchain/pull/9206 a long time ago.  but that's just general, not about windows.

tests/build-workflows.py
```diff
@@ -146,7 +146,13 @@ def dir_path(string):
             logging.info(f"Skipping {dir}: no tests collected")
             continue
         conf = update_config(module_dict(testconfig), dir_config(dir))
+        if conf["install_timelord"] and os == "windows":
+            # TODO: enable timelord for windows
+            continue
         replacements = generate_replacements(conf, dir)
+        if os == "windows":
+            # TODO: force windows to sequential only
```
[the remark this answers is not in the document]
altendky:
```suggestion
            # TODO: review for potential stable level of parallel testing processes
```

1-justinstarr (commented):
for merge into win ci

tests/core/server/test_dos.py
altendky: Undoing changes here to allow https://github.com/Chia-Network/chia-blockchain/pull/11612 to fix it when it gets merged.

```suggestion
        for i in range(6000):
            await ws_con._send_message(new_tx_message)
        await asyncio.sleep(1)
```

emlowe:
Ok various experiments leave me to conclude that using 2 xdist processes (`-n 2`) results in the best performance on the Windows runners. The longest test is the core-full-node tests which take 25-30 mins. Using more than 2 processes seems to increase the run time and increase flakes (due to timeouts being hit).
To further improve run times, we would likely need to break up the core-full-node tests into multiple jobs

tests/build-workflows.py
```diff
@@ -75,9 +75,15 @@ def generate_replacements(conf, dir):
         "TEST_FILES": "",
         "TEST_NAME": "",
         "PYTEST_PARALLEL_ARGS": "",
+        "RUN": "true",
     }
 
-    xdist_numprocesses = {False: 0, True: 4}.get(conf["parallel"], conf["parallel"])
+    if os == "windows":
+        # TODO: review for potential stable level of parallel testing processes
```
altendky, line 82:
```suggestion
    # TODO: design a configurable system for this
    if os == "windows":
```

tests/testconfig.py
```diff
@@ -1,12 +1,12 @@
...
+from typing import List, Union
 
-if TYPE_CHECKING:
-    from typing_extensions import Literal
+from typing_extensions import Literal
 
+Oses = Literal["macos", "ubuntu", "windows"]
 # Github actions template config.
-oses = ["ubuntu", "macos"]
+oses: List[Oses] = ["ubuntu", "macos", "windows"]
```
altendky, line 9: Really wish this was `linux`, but i'll let this comment satisfy my desires for now.

```suggestion
oses: List[Oses] = ["macos", "ubuntu", "windows"]
```

emlowe (dismissed):
I approve

tests/pools/test_pool_rpc.py
```diff
@@ -269,8 +269,8 @@ async def test_create_new_pool_wallet_farm_to_pool(self, one_wallet_node_and_rpc
 
         await farm_blocks(full_node_api, our_ph, 6)
         assert full_node_api.full_node.mempool_manager.get_spendbundle(creation_tx.name) is None
-
         await time_out_assert(20, wallet_is_synced, True, wallet_node_0, full_node_api)
+
```
xdustinface, line 273: nit: unrelated?

tests/runner_templates/checkout-test-plots.include.yml
```diff
@@ -7,9 +7,16 @@
           ${{ github.workspace }}/.chia/test-plots
         key: 0.29.0
 
-    - name: Checkout test blocks and plots
-      if: steps.test-blocks-plots.outputs.cache-hit != 'true'
+    - name: Checkout test blocks and plots (Linux, macOS)
+      if: steps.test-blocks-plots.outputs.cache-hit != 'true' && (startswith(matrix.os, 'ubuntu') || startswith(matrix.os, 'macos'))
```
xdustinface, line 11: Wouldn't it be possible to merge those into one step and have different run sections depending on the OS? Doesn't matter much just thought it might be cleaner in the action overview page :)

xdustinface (dismissed):
Looks good, two nits to maybe address, or not :)

tests/pools/test_pool_rpc.py
altendky: see 9fc881f0ae5787c030be1b92421a4c503a5ed114

tests/runner_templates/checkout-test-plots.include.yml
xdustinface: Got it. I just thought there might be some way to do something like 

```
- name: Checkout test blocks and plots
   if: macos/linux
   run: ....
   if windows:
   run: ....
```

But i don't know much about github actions :)

xdustinface (changes requested):
The LGTM alert indeed was introduced here because you removed the usage of `Path` in `tests/pools/test_pool_config.py` and it seems like i made the Windows CI unhappy again with some path related issues via #11625, sorry :) Seems like its related to the path string matching in `with pytest.raises(ValueError, match=f"Path doesn't exist: {test_path}")`?

xdustinface (approved):

wjblanke (approved):
aok

PR #10686, opened 2022-03-12
logging and proctitle in tests

arvidn:
when running multiple services in the same process (in tests), don't initialize logging for all of them, and don't set the proctitle of the test

### DIFF
--- a/chia/server/start_service.py
+++ b/chia/server/start_service.py
@@ -43,6 +43,7 @@ class Service:
         advertised_port: int,
         service_name: str,
         network_id: str,
+        *,
         upnp_ports: List[int] = [],
         server_listen_ports: List[int] = [],
         connect_peers: List[PeerInfo] = [],
@@ -51,7 +52,7 @@ class Service:
         rpc_info: Optional[Tuple[type, int]] = None,
         parse_cli_args=True,
         connect_to_daemon=True,
-        handle_signals=True,
+        running_new_process=True,
         service_name_prefix="",
     ) -> None:
         self.root_path = root_path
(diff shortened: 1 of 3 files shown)


altendky:
I like what's here but I think we need to set the tests to configure logging as well?  I thought I did a hacky thing in this direction at some point and ended up getting no logging until i made the tests initialize it.  Assuming I am remembering correctly, I'm not sure offhand what's best to handle this.  Could be a session scope auto-use fixture.  Could add a pytest plugin file and use one of those hooks.

arvidn:
I see. I would have thought logging would happen by default

altendky:
I might be wrong here.  Though even if there is a default handler we may want to get our regular formatting for familiarity.

arvidn:
I confirmed that tests still produce logs, and that there are no duplicates. So I think this is a clear improvement over status quo.

chia/server/start_service.py
```diff
@@ -51,7 +52,7 @@ def __init__(
         rpc_info: Optional[Tuple[type, int]] = None,
         parse_cli_args=True,
         connect_to_daemon=True,
-        handle_signals=True,
+        running_new_process=True,
```
altendky, line 55: I might go with `manage_process_state` or something perhaps?  I also somewhat wonder if this should just always be handled outside of the service.  But, I'm excited to have this improvement.

altendky (dismissed):

chia/server/start_service.py
arvidn: yeah, I agree that it probably belongs closer to the code spawning new processes, rather than here

altendky (approved):

PR #10695, opened 2022-03-12
cmds: Fix trusted peer hint in `chia wallet show`

xdustinface:
`config` is the root config here.


altendky (approved):
This is a good correction, though I'll suggest that this 'is trusted' logic probably ought to be handled on the api side.  Also note that the bug around this only applies to non-`127.0.0.1` trusted nodes since the CLI has that address hard coded.  This should probably also change.

https://github.com/Chia-Network/chia-blockchain/blob/c63324abe9e6734c4e5382c7da463e21573c9919/chia/util/network.py#L78-L86

PR #10698, opened 2022-03-12
Switch to integrated lock_and_load_config() context manager

altendky:
Draft for:
- [x] Includes https://github.com/Chia-Network/chia-blockchain/pull/10696, merge it first
- [x] Includes https://github.com/Chia-Network/chia-blockchain/pull/10694, merge it first


chia/util/config.py
```diff
@@ -48,10 +48,21 @@ def config_path_for_filename(root_path: Path, filename: Union[str, Path]) -> Pat
     return root_path / "config" / filename
 
 
-def get_config_lock(root_path: Path, filename: Union[str, Path]) -> BaseFileLock:
+@contextlib.contextmanager
+def lock_config(root_path: Path, filename: Union[str, Path]) -> Iterator[None]:
```
altendky, line 52: I believe the rename here was kind of a refactor aid to make sure I visited all usages.  I think I might like it better, but I can change it back as well.

chia/util/config.py
```diff
@@ -48,10 +48,24 @@ def config_path_for_filename(root_path: Path, filename: Union[str, Path]) -> Pat
...
+    # TODO: This is presently used in some tests to lock the saving of the
+    #       configuration file without having loaded it right there.  This usage
+    #       should probably be removed and this function made private.
     config_path = config_path_for_filename(root_path, filename)
     lock_path: Path = config_path.with_name(config_path.name + ".lock")
-    return FileLock(lock_path)
+    with FileLock(lock_path):
+        yield
```
paninaro, line 59: By yielding at this point, I'm assuming that an uncaught exception will cause the lock to be released. That definitely sounds like an important improvement.

chia/util/config.py
paninaro: +1 for the `lock_config` name

paninaro (dismissed):

chia/util/config.py
altendky: Well, what we returned was used as a context manager so that was covered anyways.  The effect of this is to keep the `FileLock` instance 'private' instead of giving it out to the caller.  This probably doesn't matter, especially in combination with the comment above suggesting this should be a private function, but...  I dunno.  I'm trying to remember why I made this particular change and I don't recall a strong reason.  I'm guessing it was related to some intermediate point in the refactor.  I think it's slightly better, but can also undo it if preferred.

chia/cmds/init_funcs.py
```diff
@@ -480,14 +477,14 @@ def chia_init(
 
             save_config(root_path, "config.yaml", config)
 
-        else:
-            config = load_config(root_path, "config.yaml", acquire_lock=False)["full_node"]
-            db_path_replaced = config["database_path"].replace("CHALLENGE", config["selected_network"])
-            db_path = path_from_root(root_path, db_path_replaced)
-            mkdir(db_path.parent)
+    else:
+        config = load_config(root_path, "config.yaml")["full_node"]
```
mariano54, line 481: Loading config not under lock here

chia/util/config.py
```diff
@@ -48,10 +48,24 @@ def config_path_for_filename(root_path: Path, filename: Union[str, Path]) -> Pat
...
+
+
+@contextlib.contextmanager
+def lock_and_load_config(root_path: Path, filename: Union[str, Path]) -> Iterator[Dict[str, Any]]:
+    with lock_config(root_path=root_path, filename=filename):
+        config = _load_config_maybe_locked(root_path=root_path, filename=filename, acquire_lock=False)
+        yield config
+        # TODO: would need to be fancier to integrate this since it is conditional
+        # save_config(root_path=root_path, filename=filename, config_data=config)
```
mariano54, line 68: Why is this comment here?

chia/cmds/init_funcs.py
altendky: The `load_config()` function does acquire the lock inside.  The intent was to work towards all 'external' functions just getting the lock.  Saving is the remaining one that does not.

chia/util/config.py
altendky: Perhaps my use of TODOs is leftover from developing alone and I should move on.  The use is that as people look at the code it is an opportunity to express an added feature of interest that they could be aware of as well as why it is not yet implemented.  Removing.

chia/cmds/init_funcs.py
mariano54: OK got it

mariano54 (approved):

PR #10700, opened 2022-03-12
coin spend improvements - DRAFT

jack60612:
I tried to do as much as i could.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -7,7 +7,7 @@ from concurrent.futures.process import ProcessPoolExecutor
 from enum import Enum
 from multiprocessing.context import BaseContext
 from pathlib import Path
-from typing import Dict, List, Optional, Set, Tuple, Callable
+from typing import Callable, Dict, List, Optional, Set, Tuple
 
 from clvm.casts import int_from_bytes
 
@@ -435,36 +435,12 @@ class Blockchain(BlockchainInterface):
                     for record in added_rec:
                         assert record
                         lastest_coin_state[record.name] = record
-                        if record.name in self.coin_subscriptions:
-                            subscribed_peers = self.coin_subscriptions[record.name]
-                            for peer in subscribed_peers:
-                                if peer not in changes_for_peers:
-                                    changes_for_peers[peer] = set()
-                                changes_for_peers[peer].add(record.coin_state)
-
-                        if record.coin.puzzle_hash in self.ph_subscriptions:
-                            subscribed_peers = self.ph_subscriptions[record.coin.puzzle_hash]
-                            for peer in subscribed_peers:
-                                if peer not in changes_for_peers:
-                                    changes_for_peers[peer] = set()
-                                changes_for_peers[peer].add(record.coin_state)
+                        self.check_subscriptions(record, changes_for_peers)
 
                     for record in removed_rec:
                         assert record
                         lastest_coin_state[record.name] = record
-                        if record.name in self.coin_subscriptions:
-                            subscribed_peers = self.coin_subscriptions[record.name]
-                            for peer in subscribed_peers:
-                                if peer not in changes_for_peers:
-                                    changes_for_peers[peer] = set()
-                                changes_for_peers[peer].add(record.coin_state)
-
-                        if record.coin.puzzle_hash in self.ph_subscriptions:
-                            subscribed_peers = self.ph_subscriptions[record.coin.puzzle_hash]
-                            for peer in subscribed_peers:
-                                if peer not in changes_for_peers:
-                                    changes_for_peers[peer] = set()
-                                changes_for_peers[peer].add(record.coin_state)
+                        self.check_subscriptions(record, changes_for_peers)
 
                     if npc_res is not None:
                         hint_list: List[Tuple[bytes32, bytes]] = self.get_hint_list(npc_res)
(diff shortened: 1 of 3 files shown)


daverof:
While you're here, can you fix the typo in `lastest_coin_state` ? (6 instances)

arvidn:
I suggest keeping this PR contained to solving one problem. Fixing the typo should be a separate PR (which presumably is less controversial and will just make it straight through review)

arvidn:
@jack60612 can you add a description to this PR? to explain what you're doing and why?

chia/consensus/blockchain.py
```diff
@@ -218,18 +222,18 @@ async def receive_block(
...
-            return ReceiveBlockResult.ALREADY_HAVE_BLOCK, None, None, ([], {})
+            return ReceiveBlockResult.ALREADY_HAVE_BLOCK, None, None, ([], {}), {}
 
         if not self.contains_block(block.prev_header_hash) and not genesis:
-            return (ReceiveBlockResult.DISCONNECTED_BLOCK, Err.INVALID_PREV_BLOCK_HASH, None, ([], {}))
+            return ReceiveBlockResult.DISCONNECTED_BLOCK, Err.INVALID_PREV_BLOCK_HASH, None, ([], {}), {}
 
         if not genesis and (self.block_record(block.prev_header_hash).height + 1) != block.height:
-            return ReceiveBlockResult.INVALID_BLOCK, Err.INVALID_HEIGHT, None, ([], {})
+            return ReceiveBlockResult.INVALID_BLOCK, Err.INVALID_HEIGHT, None, ([], {}), {}
```
arvidn, line 231: I thought this return type was beyond unwieldy to begin with. I really don't think we should make this more complicated. If you need to return more stuff, I think we better give the return type a name

jack60612:
> @jack60612 can you add a description to this PR? to explain what you're doing and why?

yep, its mostly for https://trello.com/c/CTtzHXhW/1453-performance-improvement-of-wallet-subscriptions-of-coin-spends . the pr is not ready in any sense and is still a work in progress.

chia/consensus/blockchain.py
```diff
@@ -420,13 +431,41 @@ async def _reconsider_peak(
...
                     record: Optional[CoinRecord]
                     for record in added_rec:
                         assert record
                         lastest_coin_state[record.name] = record
+                        if record.name in coin_subscriptions:
+                            subscribed_peers = coin_subscriptions[record.name]
+                            for peer in subscribed_peers:
+                                if peer not in changes_for_peer:
+                                    changes_for_peer[peer] = set()
+                                changes_for_peer[peer].add(record.coin_state)
```
arvidn, line 444: I think it would make sense to factor out the subscription management to a separate function. It would give it a name, and deduplicate some of this code, and make the main purpose of the logic here less diluted

chia/consensus/blockchain.py
```diff
@@ -447,10 +498,11 @@ async def _reconsider_peak(
                 block_record.height,
                 records_to_add,
                 (list(lastest_coin_state.values()), hint_coin_state),
+                changes_for_peer,
```
arvidn, line 501: I'm pretty sure you no longer need to return the `lastest_coin_state` and `hint_coin_state`. The only reason (afaik) for this to be returned was for checking subscriptions later

arvidn (commented):
we need tests for this too. there are none right now

chia/consensus/blockchain.py
jack60612: Gotcha

chia/consensus/blockchain.py
jack60612: It is still being used by various other functions.

chia/consensus/blockchain.py
```diff
@@ -436,12 +450,29 @@ async def _reconsider_peak(
                             key = hint
                             if key not in hint_coin_state:
                                 hint_coin_state[key] = {}
-                            hint_coin_state[key][coin_id] = lastest_coin_state[coin_id]
+                            coin_record = lastest_coin_state[coin_id]
```
mariano54, line 453: I'm skeptical of adding any more code to the `reconsider_peak` method. In fact. all of this code for subscriptions should be called in peak_post_processing, *after* receive_block returns, otherwise it becomes critical for consensus

chia/consensus/blockchain.py
```diff
@@ -98,6 +99,10 @@ class Blockchain(BlockchainInterface):
     compact_proof_lock: asyncio.Lock
     hint_store: HintStore
 
+    # Subscriptions for light clients
+    coin_subscriptions: Dict[bytes32, Set[bytes32]]  # Puzzle Hash : Set[Peer ID]
```
mariano54, line 103: This is not blockchain related, it's full node related, so it should be in the FullNodeStore

chia/consensus/blockchain.py
```diff
@@ -474,6 +505,21 @@ async def get_tx_removals_and_additions(
         else:
             return [], [], None
 
+    def check_subscriptions(self, record: CoinRecord, changes_for_peers: Dict[bytes32, Set[CoinState]]) -> None:
```
mariano54, line 508: This also does not belong in this class

mariano54 (changes requested):

arvidn:
I think this improvement (at least as I envision it) is a bit complex. The key objectives (in my mind) are:

1. simplify the code
2. Speed up the code (where we avoid allocating memory or returning entries that are not relevant to us)
3. compartmentalize the consensus code from the coin, puzzle and hint subscription logic

Perhaps a better approach than what I have suggested earlier would be to:

instead of collecting all coin updates and return them from the blockchain, take an optional callback function that's called for every coin update. The caller (i.e. the full node) can then pass in a callback function that filters the coin updates based on its subscriptions. Such approach would:

1. separate the subscription logic from the blockchain class into the full node class
2. provide performance improvements because we wouldn't have to build a list of irrelevant coin updates

We would still somehow have to pass out the `CoinRecord`s we look up in block_body_validation() though, since a coin update needs that, and we should avoid looking them up twice from the database.

mariano54:
I agree this logic needs to be outside of the blockchain. A callback could work, but if it's inside `_reconsider_peak` then we still have the same issue of making it consensus critical code. This should be executed after the state transition of the blockchain (reconsider peak). 

One way we can do this is to move all of the code that handles hints outside of blockchain.py and into full node. This would mean that if something fails in hint handling the blockchain is still intact and can move forward. It also means that it might be possible to not have consensus on the hints (one node might have different hints than another). I think this is an acceptable compromise to keep consensus code smaller and safer.

arvidn:
the problem of not performing *any* hints handling within consensus critical code is that it must then be allocated to be dealt with later. That's an eager memory allocation that will in the vast majority of cases be unnecessary. Maybe we can live with that for now.

PR #10717, opened 2022-03-13
new DBWrapper supporting concurrent readers

altendky:


chia/util/db_wrapper.py
```diff
@@ -70,39 +70,52 @@ def _next_savepoint(self) -> str:
...
+        except:  # noqa E722
+            await connection.execute(f"ROLLBACK TO {name}")
+            raise
+        finally:
+            # rollback to a savepoint doesn't cancel the transaction, it
+            # just rolls back the state. We need to cancel it regardless
+            await connection.execute(f"RELEASE {name}")
+
+    @contextlib.contextmanager
+    def _set_current_writer(self, writer: asyncio.Task) -> Iterator[None]:
```
arvidn, line 90: I think this context manager may be taking it a little bit too far (in terms of lines of code and ease-of-reading to utility ratio)

chia/util/db_wrapper.py
```diff
@@ -70,39 +70,52 @@ def _next_savepoint(self) -> str:
         self._savepoint_name += 1
         return name
 
+    @contextlib.asynccontextmanager
+    async def _savepoint(self, connection) -> AsyncIterator[None]:
+        name = self._next_savepoint()
+        await connection.execute(f"SAVEPOINT {name}")
+        try:
+            # TODO: maybe yield out something to make it possible to cancel the
+            #       savepoint other than raising an exception?
```
arvidn, line 79: in my experience, aborting without collaborating with the holder of the resource is really questionable and unreliable (mostly really complex). If we cancel a task, and get a `TaskCancelled` exception, I would expect it all to just work out anyway.

That's a long way of saying; "I don't think we need this comment here".

arvidn (commented):
I this generally makes sense. It's more lines of code though, and the original version (I would say) didn't really suffer much from complexity.

chia/util/db_wrapper.py
altendky: I didn't mean to replace the existing behavior, rather that augmenting it might be useful.  basically something to allow:

```python
async with wrapper.write_db() as connection:
    connection.execute(something)
    if condition():
        connection.undo_my_changes()
```

Instead of forcing raising of an exception through the context manager and subsequent handling of that except just to achieve an undo:

```python
try:
    async with wrapper.write_db() as connection:
        connection.execute(something)
        if condition():
            raise CustomErrorJustToCancelTransaction()
finally CustomErrorJustToCancelTransaction:
    pass
```

The comment isn't a big deal I suppose.  It seems like there may be cases where it would be a nice addition.  We'll see if they come up or not.  I guess maybe it would just be a rollback to the relevant savepoint.

chia/util/db_wrapper.py
altendky: Yeah, this is on the edge for me too.  But while reading the existing code I kept going back to the question of whether the multiple activities in a `finally:` block were ok or not.  Could have added another `try:`/`finally:` for them, or just drop this here.  Either way.  I guess at the present implementation here it wouldn't be "another `try`", it would just be the only `try`, so that doesn't really bug me.

altendky:
I agree that the existing code certainly wasn't unreadable.  It did though require you to think about multiple things at a time that were supposed to make sure they were being properly cleaned up while also tracking which of the paths were being taken for nesting or not.  This separates the cleanup guarantees from what each path does.

https://github.com/Chia-Network/chia-blockchain/blob/56980b0c73387ec4f7169dd9c356c4fb37f7421b/chia/util/db_wrapper.py#L102-L110

There are only two duplicated lines to compare when checking what is the same and different between nested or not vs. eight.

https://github.com/Chia-Network/chia-blockchain/blob/552cc5c6aa382dfafec6a8c346d44e301c836c23/chia/util/db_wrapper.py#L77-L105

PR #10718, opened 2022-03-13
Add healthcheck endpoint to rpc services

Starttoaster:
Some deployment environments (for example, within the chia-docker container) attempt to self-correct when services end abruptly. Generally, for that to happen, some kind of "health check" is required. This adds a minimal endpoint that just returns a short success message and an http 200 if the RPC service is running. The idea is that this endpoint will be queried at intervals and should ideally do a minimal amount of work to succeed.

The "/healthz" route was made popular by convention set by google, and is now used widely for many other applications that provide dedicated healthcheck endpoints. The idea behind the "z" suffix is to avoid naming collisions.

### DIFF
--- a/chia/rpc/rpc_client.py
+++ b/chia/rpc/rpc_client.py
@@ -66,6 +66,9 @@ class RpcClient:
 
     async def stop_node(self) -> Dict:
         return await self.fetch("stop_node", {})
+    
+    async def healthz(self) -> Dict:
+        return await self.fetch("healthz", {})
 
     def close(self):
         self.closing_task = asyncio.create_task(self.session.close())
--- a/chia/rpc/rpc_client.py
+++ b/chia/rpc/rpc_client.py
@@ -66,7 +66,7 @@ class RpcClient:
 
     async def stop_node(self) -> Dict:
         return await self.fetch("stop_node", {})
-    
+
     async def healthz(self) -> Dict:
         return await self.fetch("healthz", {})
 
--- a/chia/rpc/rpc_server.py
+++ b/chia/rpc/rpc_server.py
@@ -87,6 +87,7 @@ class RpcServer:
             "/close_connection": self.close_connection,
             "/stop_node": self.stop_node,
             "/get_routes": self._get_routes,
+            "/healthz": self.healthz,
         }
 
     async def _get_routes(self, request: Dict) -> Dict:
@@ -182,6 +183,11 @@ class RpcServer:
         if self.stop_cb is not None:
             self.stop_cb()
         return {}
+
+    async def healthz(self, request: Dict) -> Dict:
+        return {
+            "success": "true",
+        }
 
     async def ws_api(self, message):
         """


freddiecoleman:
It would be nice if it did some actual checks such as checking if there are any connections to peers.

stefan-lange:
This is great! So I can remove my mix of 'nc' and 'pgrep'  checks

Starttoaster:
> It would be nice if it did some actual checks such as checking if there are any connections to peers.

In my opinion, that could be helpful as a separate endpoint, but shouldn't be combined with this one. The point of this endpoint is to be queried often in intervals, and to do the least amount of work possible, so as to not take resources away from what your farm should be doing (farming.) In kubernetes, for example, this would be called by liveness/readiness probes by default every 10 seconds, but at most every 1 second.

The inspiration behind this endpoint was observing that at one point a chia farmer deployed to kubernetes had suddenly stopped running. The RPC service halted abruptly, and kubernetes did not notice anything was wrong, so it did nothing. This endpoint, while minimal, would have prevented that with the correct liveness/readiness probe configuration. And kubernetes would have then flagged that pod as unhealthy and rescheduled it (essentially turning it off and back on again.)

If this gets added to Chia RPCs, the healthcheck script within the chia-docker container here, would be updated to use this endpoint instead of the "/get_routes" one. https://github.com/Chia-Network/chia-docker/blob/main/docker-healthcheck.sh#L63

cmmarslender (approved):
lgtm. Tested node, farmer, harvester, and wallet locally w/ curl and all returned the expected response

emlowe:
This seems a reasonable thing - my counter-argument is that /get_routes is already pretty lightweight. I'm not familiar enough with this concept of "z-pages" to say whether this is a useful quasi-standard to implement

Starttoaster:
> my counter-argument is that /get_routes is already pretty lightweight. 

@emlowe That's agree-able. The main counter-counter argument I have off the top of my head is that it's not intuitive. If we were to document an endpoint specifically for checking that the service is available `/get_routes` would not immediately make sense to me. In writing that script, I only used it after reading the currently available RPCs and determining it's probably the lightest-weight endpoint we currently have.

> I'm not familiar enough with this concept of "z-pages" to say whether this is a useful quasi-standard to implement

I'm not married to sticking to the "z-pages" standard if there's a holdup specifically over that. I can rename it to be "/health", "/live", "/healthcheck", etc etc, if there's a preference here. This is just the standard I've most often followed when writing services with uptime in mind, but it's also pretty common to just see "/health". I think the most important thing is that whatever endpoint is used it is rational that it would be used for the purpose it is being used for, which is presumably why it's exceedingly common to have a dedicated API endpoint for it.

emlowe:
This is a good point - one would have to study the code to understand to use /get_routes as a health check - in fact, I didn't know this myself until checking the code.

Ok, I'm good with adding it as /healthz

emlowe (approved):
I think this is a reasonable thing to add as a more obvious healthcheck endpoint

PR #10721, opened 2022-03-14
added height check

wjblanke:


arvidn:
we really should have a unit test exercising the `cs_height` > `height`.

wjblanke:
It looks like this is double checked in can_use_peer_request_cache() which seems to be the only place that uses this cache

wjblanke:
if coin_state.created_height is not None and coin_state.created_height > fork_height:
        return False
    if coin_state.spent_height is not None and coin_state.spent_height > fork_height:
        return False

mariano54:
We should still fix this I think

PR #10727, opened 2022-03-15
Atari merge main 8a5c12541809526ef55995674595ed6485c6c330

altendky:
Merge required a few manual selections of order where both sides appended to a list.  Also removed the data layer specific section of the mypy config as the strictness used there is now covered by the general section of the config.


emlowe (approved):
Looks good - but did you mean to include the chia-blockchain-gui submodules changes

altendky:
The intent would be to track the GUI pin that `main` had at the commit being merged here.  I think my checking shows that to be what is happening here.

PR #10729, opened 2022-03-15
harvester: Reuse legacy refresh interval if new params aren't available

xdustinface:
Currently the default parameters are used if `plots_refresh_parameter` is missing in the config even if `plot_loading_frequency_seconds` is still provided.


emlowe (approved):
lgtm

PR #10737, opened 2022-03-15
enable stricter mempool rule

arvidn:
enable the `STRICT_ARGS_COUNT` mempool rule and fix the mempool tests accordingly.

This rule forbids conditions from having too many arguments. Under normal consensus rules, additional arguments are ignored, so this only affects the mempool.

### DIFF
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -1,6 +1,6 @@
 import logging
 from typing import Dict, Optional
-from chia_rs import MEMPOOL_MODE, COND_CANON_INTS, NO_NEG_DIV, STRICT_ARGS_COUNT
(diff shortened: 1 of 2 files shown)


chia/full_node/mempool_check_conditions.py
```diff
@@ -42,8 +42,7 @@ def get_name_puzzle_conditions(
     assert (MEMPOOL_MODE & NO_NEG_DIV) != 0
 
     if mempool_mode:
-        # Don't apply the strict args count rule yet
-        flags = MEMPOOL_MODE & (~STRICT_ARGS_COUNT)
```
arvidn, line 46: this is the change to production code, the rest are tests being updated

mariano54 (approved):

PR #10744, opened 2022-03-15
single test workflow

altendky:


.github/workflows/build-test.yml
```diff
@@ -0,0 +1,180 @@
...
+    name: ${{ matrix.configuration.name }} ${{ matrix.os.emoji }} ${{ matrix.python.name }}
+    needs: configure
+    runs-on: ${{ matrix.os.runs-on }}
+    timeout-minutes: ${{ matrix.configuration.job_timeout }}
+    strategy:
+      fail-fast: false
+      matrix:
+        configuration: ${{ fromJson(needs.configure.outputs.configuration) }}
+        os:
+          - emoji: 🍎
```
hoffmang9, line 49: This is smart.

wallentx: Woah, that's :banana:s

.github/workflows/test.yml
```diff
@@ -0,0 +1,198 @@
+name: test
```
justinengland, line 1: typically this should be more descriptive than "test" this is I believe the friendly name that will show up in the gui, something like "omnitest-workflow" that is more descriptive would be great.

.github/workflows/test.yml
```diff
@@ -0,0 +1,198 @@
...
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - '**'
+  pull_request:
+    branches:
+      - '**'
```
justinengland, line 11: any of these we'd want to split to say run a portion on prs to main only?

.github/workflows/test.yml
```diff
@@ -0,0 +1,198 @@
...
+            runs-on: macos-latest
+          - emoji: 🐧
+            matrix: ubuntu
+            name: Ubuntu
+            runs-on: ubuntu-latest
+#          - emoji: 🪟
+#            matrix: windows
+#            name: Windows
+#            runs-on: windows-latest
+        python:
```
justinengland, line 66: do python version of fun chibi emojis too? that would be great from a visual scanning perspective if they did.

tests/build-job-matrix.py
```diff
@@ -0,0 +1,120 @@
+import argparse
```
justinengland, line 1: would like to see this added to the internal actions repo as its own job maybe. not sure. since it is a specific python job for this task. Leaving this comment to spur discussion.

justinengland (commented):
I will let Chris do his own review. I don't see any structural concerns here really. How does this run vs the separate jobs workflow? Any gains from this refactor other than the move to a single file being easier to read and work with?

.github/workflows/test.yml
altendky: Could go with something below for the minor version if you think not-just-numerals looks better.  I think this might deal with the width, but I'm not sure it actually makes it better.

⑦ ⑧ ⑨ ⑩
❼ ❽ ❾ ❿

- 🍎❽
- 🍎❾
- 🍎❿

- 🍎❸❽
- 🍎❸❾
- 🍎❸❿

- 🍎3.8
- 🍎3.9
- 🍎3.10

I guess the full-width numerals are bigger and more visible?

- 🍎３.８
- 🍎３.９
- 🍎３.１０

.github/workflows/test.yml
altendky: I hope we don't have to go to partial CI, personally.  But I guess maybe.  Could be fed into the matrix generation if it made sense to handle it there.

.github/workflows/test.yml
altendky: There are various views so yeah, there's probably somewhere this would be annoying.  On the flip side, I was getting tired of the job lists looking like `MacOS blockchain Tests / MacOS blockchain Tests (3.8, macOS-latest) (pull_request)` vs. `test / 🍎 clvm - 3.8 (pull_request) `.

What particular place do you think `test` is insufficient so I can see what we need to clarify it against?

.github/workflows/build-test.yml
altendky: So a downside here is that it'll be annoying to search for the workflow you want.  Which I have been doing a lot.  `:|`

emlowe:
I believe the primary benefit here is code coverage, using a single workflow makes it much easier to produce a combined code coverage report.

.github/workflows/test.yml
cmmarslender: I'd prefer non-emoji numbers. Easier to read. Normal spacing seems fine to me

.github/workflows/test.yml
cmmarslender: Since the final job name ends up being something like `test / 🍎 clvm - 3.8 (pull_request)` - I think this looks ok to me. Easier to read at a glance and doesn't get cut off with `...` most places

tests/build-job-matrix.py
cmmarslender: I think it might be kind of tricky to make this an action as a python script.. Actions I think typically only support javascript (or else, more basic yaml). Might be nice to make this generic enough to be an action and we can reuse it elsewhere down the road, but I think I'd be ok if this goes in chia-blockchain for now and we can abstract/move it when we need it somewhere else? I think it's basically just replacing a similar script that used to live here anyways.

.github/workflows/test.yml
justinengland: normal numbers works for me as well. resolving.

justinengland (dismissed):

emlowe (dismissed):

AmineKhaldi (dismissed):
Looks great!

AmineKhaldi (dismissed):

altendky:
Lives on in https://github.com/Chia-Network/chia-blockchain/pull/11315

PR #10749, opened 2022-03-15
Do not use fixed ports for data layer tests

altendky:


chia/server/start_data_layer.py
```diff
@@ -22,7 +22,11 @@
 
 # TODO: Review need for config and if retained then hint it properly.
 def service_kwargs_for_data_layer(
-    root_path: pathlib.Path, config: Dict[str, Any], wallet_rpc_port: Optional[uint16] = None
+    root_path: pathlib.Path,
+    config: Dict[str, Any],
+    port: Optional[int] = None,
+    rpc_port: Optional[int] = None,
+    wallet_rpc_port: Optional[int] = None,
```
arvidn, line 29: the fact that you pass both the `config` and a series of optional ports seem to suggest (to me at least) that the picking of ports should happen one layer up, and set them in the `config` object. I suppose there's a good reason to do it at this level.

This is also "production code" which ideally would have minimal impact by test logic.

tests/core/data_layer/test_data_rpc.py
```diff
@@ -22,16 +22,22 @@
...
 from tests.wallet.rl_wallet.test_rl_rpc import is_transaction_confirmed
 
 pytestmark = pytest.mark.data_layer
 nodes = Tuple[WalletNode, FullNodeSimulator]
 
 
-async def init_data_layer(root_path: Path) -> AsyncIterator[DataLayer]:
-    test_rpc_port = uint16(21529)
+async def init_data_layer(root_path: Path, wallet_rpc_port: int) -> AsyncIterator[DataLayer]:
     config = bt.config
```
arvidn, line 33: I guess I'm still confused about this. what's `bt` here? it's not the fixture, right?

I would have expected it to be more reasonable to configure the service here, by updating `config`. If `bt` is the fixture, it will have a root path configured already too.

tests/core/data_layer/test_data_rpc.py
```diff
@@ -22,16 +22,22 @@
...
-    kwargs = service_kwargs_for_data_layer(root_path, config, test_rpc_port)
+    kwargs = service_kwargs_for_data_layer(
+        root_path=root_path,
```
arvidn, line 35: afaict, the `root_path` is just pulled out of `bt` already (line 79), so passing in `bt` might make more sense (which would allow it to be a fixture)

altendky: `service_kwargs_for_data_layer()` is production code so passing it a blocktools seems unlikely to be proper.  Also, `root_path` and `config` are the standard parameters for other `service_kwargs_for_*()`.

altendky (commented):
Thanks for the help tidying this up.

arvidn (approved):

altendky:
The M1 brokenness seems unlikely to be related to this.  Rather there's some global state that seems to have changed and broken all runs, possibly by uncovering an underlying issue in main.  We'll keep working forward towards being caught up with main.

PR #10673, opened 2022-03-11
Improve the CI runs w.r.t. timelord installation

AmineKhaldi:
Superficial analysis showed that only two test groups require (for now) installing the timelord.

This change aims to save us hours of CI running time by simply running the install timelord script only for those test groups, with everything else having it omitted.

Dedicated to @hoffmang9

PR #10752, opened 2022-03-15
stop using deadsnakes, unless we need it

altendky:
Draft for:
- [x] Check again after https://github.com/Chia-Network/chia-blockchain/pull/10673 lands to see if we can maybe isolate this to the few workflows that actually need the timelord.


AmineKhaldi (approved):
Nice!

PR #10754, opened 2022-03-15
run benchmarks separately

arvidn:
This patch separates tests that check the runtime of some operation from other tests. Such tests are marked with `@pytest.mark.benchmark` to indicate this.

All normal github workflows are updated to *not* run benchmark tests, and a new workflow is added that *only* runs benchmark tests (on dedicated hardware). With this added performance reliability, the current time limits can also be tightened.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [ 3.9 ]
 
     steps:
     - name: Checkout Code
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,7 +1,4 @@
-#
-# THIS FILE IS GENERATED. SEE https://github.com/Chia-Network/chia-blockchain/tree/main/tests#readme
-#
-name: Ubuntu clvm Test
+name: Benchmarks
 
 on:
   push:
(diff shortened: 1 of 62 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,69 @@
...
+jobs:
+  build:
+    name: Benchmarks
+    runs-on: benchmark
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      max-parallel: 4
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
```
arvidn, line 27: maybe this is unnecessary. we wouldn't really expect the python version to matter much

hoffmang9: Historically we've been bitten on the move to the next version. I lean that we drop to just 3.9 and then add 3.10 once we're ready and keep a pattern of newest and one behind.

tests/core/full_node/test_mempool.py
```diff
@@ -2183,6 +2183,7 @@ class TestMaliciousGenerators:
             ConditionOpcode.ASSERT_SECONDS_RELATIVE,
         ],
     )
+    @pytest.mark.benchmark
```
aqk, line 2186: I see. You had the funny choice of whether to mark the benchmarks, or mark the "not benchmarks"

tests/core/full_node/test_performance.py
```diff
@@ -201,8 +206,10 @@ async def test_full_block_performance(self, bt, wallet_nodes, self_hostname):
         # No transactions generator, the full node already cached it from the unfinished block
         block_small = dataclasses.replace(block, transactions_generator=None)
         res = await full_node_1.full_node.respond_block(fnp.RespondBlock(block_small))
+        end = time.time()
         log.warning(f"Res: {res}")
-        log.warning(f"Time for full block: {time.time() - start}")
+        log.warning(f"Time for full block: {end - start:f}")
```
aqk, line 211: Could bundle this log.warning and assert into a small function

aqk (dismissed):

tests/core/full_node/test_mempool.py
arvidn: I think it's important that tests not be considered benchmarks by default

aqk (approved):

PR #10757, opened 2022-03-16
wallet: Fix `STANDARD_WALLET` creation for `wallet_info.id != 1`

xdustinface:
Not sure about the workflow to create a standard wallet with an id != 1 but i somehow ended up with one with ID 29 and when i wanted to query it it failed due to `wallet_state_manager` being the `config` dict.


emlowe (approved):
Clearly needed - thanks

wjblanke:
this happens with reorgs

PR #10759, opened 2022-03-16
add locking around reorg_rollback to avoid cannot start a transaction…

wjblanke:
… within a transaction


wjblanke:
these should be in atomic sqlite transactions too with rollback

PR #10760, opened 2022-03-16
Adding chia peer command

### LINKED ISSUE #12298: CLI naming
### What happened?

I propose renaming `chia show` to `chia node` since half the commands are not about showing information but about interacting with the full node.
Or only the commands related to doing actions on the full node could be moved to `chia node`, while other commands that are about showing information like `chia netspace` could be moved to a `chia show` subcommand.

Second thing, the 2 letter short options are a footgun for people used to CLI tools. The UNIX convention is for short options to be only 1 letter long so that they can be combined in one argument, but for example `chia show` has 3 different options `-bh`, `-b` and `-h`.

This might be low priority but the CLI is often the first point of contact for beginners, so reducing confusion is a good thing. I was personally confused by it. 

### Version

1.4.0

Linux

CLI

### Relevant log output

_No response_

wallentx:
Addition of chia peer command for (wallet, farmer, full node and harvester) & slight refactor of chia show.


chia/cmds/peer.py
```diff
@@ -0,0 +1,157 @@
+from typing import Optional, Dict
+
+import click
+
+from chia.util.network import is_trusted_inner
+
+
+async def print_connections(client, time, NodeType, trusted_peers: Dict):
```
altendky, line 8: If you need the `time` module, go ahead and import it.

```suggestion
async def print_connections(client, NodeType, trusted_peers: Dict) -> None:
```

chia/cmds/peer.py
```diff
@@ -0,0 +1,157 @@
...
+    from chia.util.ints import uint16
+
+    try:
+        config = load_config(DEFAULT_ROOT_PATH, "config.yaml")
+        self_hostname = config["self_hostname"]
+        rpc_port = config["full_node"]["rpc_port"]
+        client = await FullNodeRpcClient.create(self_hostname, uint16(rpc_port), DEFAULT_ROOT_PATH, config)
+
+        if show_connections:
+            trusted_peers: Dict = config["full_node"].get("trusted_peers", {})
```
altendky, line 78: I think this is what comes out of the config?

```suggestion
            trusted_peers: Dict[str, str] = config["full_node"].get("trusted_peers", {})
```

jack60612:
Testing locally now

jack60612:
Local CLI tests successful!!

jack60612:
> This pull request **introduces 8 alerts** when merging [d398434](https://github.com/Chia-Network/chia-blockchain/commit/d3984347973b3989a1827e735bbe59b189bdd7d3) into [6923ad8](https://github.com/Chia-Network/chia-blockchain/commit/6923ad8e12a376e7f5fa428a0cea0de70ce53e7f) - [view on LGTM.com](https://lgtm.com/projects/g/Chia-Network/chia-blockchain/rev/pr-d254e07bc975b818c0b621089bdf3fb6b4bf2367)
> 
> **new alerts:**
> 
> * 8 for Module-level cyclic import

False alarm

emlowe:
This break backwards compatibility right - `chia show -c` doesn't work anymore?  Should we maintain the `chia show` options? At the very least I think `chia show -c` should print that it is deprecated and use `chia peer -c`?

jack60612:
> This break backwards compatibility right - `chia show -c` doesn't work anymore? Should we maintain the `chia show` options? At the very least I think `chia show -c` should print that it is deprecated and use `chia peer -c`?

i can add that. It does break backwards compatibility

chia/cmds/show.py
```diff
@@ -229,19 +229,24 @@ async def show_async(
 )
 @click.option("-s", "--state", help="Show the current state of the blockchain", is_flag=True, type=bool, default=False)
 @click.option(
-    "-bh", "--block-header-hash-by-height", help="Look up a block header hash by block height", type=str, default=""
+    "-h", "--block-header-hash-by-height", help="Look up a block header hash by block height", type=str, default=""
```
emlowe, line 232: breaking this backwards compat should be avoided (I understand the multi-letter options are a problem, but we shouldn't just break things)

emlowe:
You also need to fix `chia show -a` and `chia show -r` and anythign else removed from `chia show`

For example, I know the `--wallet-rpc-port` is unused, but by removing it you break scripts that were using it accidentally - it would be better to leave it in and make the help say it's unused and deprecated. Oh, click has a deprecated flag https://click.palletsprojects.com/en/8.1.x/api/?highlight=deprecated#click.Command

jack60612:
> You also need to fix `chia show -a` and `chia show -r` and anythign else removed from `chia show`
> 
> For example, I know the `--wallet-rpc-port` is unused, but by removing it you break scripts that were using it accidentally - it would be better to leave it in and make the help say it's unused and deprecated. Oh, click has a deprecated flag https://click.palletsprojects.com/en/8.1.x/api/?highlight=deprecated#click.Command

ok that makes sense.

jack60612:
@emlowe is it good now?

jack60612:
all fixed

chia/cmds/show.py
```diff
@@ -326,16 +217,6 @@ async def show_async(
...
-@click.option(
-    "-p",
-    "--rpc-port",
-    help=(
-        "Set the port where the Full Node is hosting the RPC interface. "
-        "See the rpc_port under full_node in config.yaml"
-    ),
-    type=int,
-    default=None,
-)
```
emlowe, line 338: This needs to be left in - it's at least used later on.

jack60612: This was accidently removed, i'm going to re review this whole pr before making any more changes

jack60612: Sorry about the review spam, should be good now

mariano54 (changes requested):
I think the `chia peer` command should support any service (farmer, harvester, wallet etc), since they all support the same RPC calls.

This would make it more generic and useful for all types of situations. `chia show` was meant to be just for the node, but if we are refactoring this RPC, I think we should have a higher standard and make it generalizeable.

jack60612:
both the chia peer and the chia show command were successfully tested locally.

chia/cmds/cmds_util.py
```diff
@@ -10,3 +26,48 @@ def transaction_submitted_msg(tx: TransactionRecord) -> str:
...
+    try:
+        await rpc_client.healthz()
+    except ClientConnectorError:
+        print(f"Connection error. Check if {node_type.replace('_', ' ')} rpc is running at {rpc_port}")
+        print(f"This is normal if {node_type.replace('_', ' ')} is still starting up")
+        return False
+    return True
+
+
+async def execute_with_any_node(
```
mariano54, line 43: I would rename to `execute_with_any_service`. Node means fulls node in chia vocabulary.

mariano54 (dismissed):
LGTM apart from small nit for naming

chia/cmds/cmds_util.py
jack60612: Renamed!

mariano54 (dismissed):

mariano54 (changes requested):
Need to fix the LGTM issues (cyclic import)

jack60612:
> Need to fix the LGTM issues (cyclic import)

they are wrong, i dont know what they are yapping about.

emlowe:
Yea, LGTM has some issues - Kyle would actually like us to stop using it as it's not adding much value

mariano54:
We can't merge this PR if it adds 40 LGTM issues. 
Or should we just start ignoring it completely?

jack60612:
closing this pr in favor of #12532

PR #10763, opened 2022-03-16
cmds: Implement `chia rpc` command

xdustinface:
Adds the new command `chia rpc` to have an easy way to access all RPC endpoints via CLI. It enables direct access to all endpoints of the selected service with the JSON request provided as string argument like in 

```
chia rpc wallet get_wallet_balance '{"wallet_id": 12}'
``` 

or with the JSON request provided via a JSON file like in 

```
chia rpc wallet get_wallet_balance -j ~/request.json 
```

### Help
```
chia rpc
Usage: chia rpc [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --help  Show this message and exit.

Commands:
  farmer     RPC client for the farmer RPC API
  full_node  RPC client for the full_node RPC API
  harvester  RPC client for the harvester RPC API
  routes     Print all routes of a service
  status     Print the status of all available RPC services
  wallet     RPC client for the wallet RPC API
```

## Examples

```
chia rpc wallet get_wallets
{
    "success": true,
    "wallets": [
        {
            "data": "",
            "id": 1,
            "name": "Chia Wallet",
            "type": 0
        },
        {
            "data": "",
            "id": 12,
            "name": "Pool wallet",
            "type": 9
        }
    ]
}
```

and 

```
chia rpc wallet get_wallet_balance '{"wallet_id": 12}'
{
    "success": true,
    "wallet_balance": {
        "confirmed_wallet_balance": 0,
        "fingerprint": 172057028,
        "max_send_amount": 0,
        "pending_change": 0,
        "pending_coin_removal_count": 0,
        "spendable_balance": 0,
        "unconfirmed_wallet_balance": 0,
        "unspent_coin_count": 0,
        "wallet_id": 12
    }
}
```

### DIFF
--- a/chia/cmds/rpc.py
+++ b/chia/cmds/rpc.py
@@ -1,0 +1,140 @@
+import asyncio
+import json
+import sys
+from typing import Any, Dict, List, Optional, TextIO
+
+import click
+from aiohttp import ClientResponseError
+
+from chia.util.config import load_config
+from chia.util.default_root import DEFAULT_ROOT_PATH
+from chia.util.ints import uint16
+
+services: List[str] = ["farmer", "full_node", "harvester", "wallet"]
+
+
+async def call_endpoint(
+    service: str, host: str, port: uint16, endpoint: str, request: Dict[str, Any], config: Dict[str, Any]
+) -> Dict[str, Any]:
+    from chia.rpc.rpc_client import RpcClient
+
+    try:
+        client = await RpcClient.create(host, port, DEFAULT_ROOT_PATH, config)
+    except Exception as e:
+        raise Exception(f"Failed to create RPC client {service}: {e}")
+    result: Dict[str, Any]
+    try:
+        result = await client.fetch(endpoint, request)
+    except ClientResponseError as e:
+        if e.code == 404:
+            raise Exception(f"Invalid endpoint for {service}: {endpoint}")
+        raise e
+    except Exception as e:
+        raise Exception(f"Request failed: {e}")
+    finally:
+        client.close()
+        await client.await_closed()
+    return result
+
+
+def print_result(json_dict: Dict[str, Any]) -> None:
+    print(json.dumps(json_dict, indent=4, sort_keys=True))
+
+
+def get_routes(service: str, config: Dict[str, Any]) -> Dict[str, Any]:
+    return asyncio.run(
+        call_endpoint(service, config["self_hostname"], uint16(config[service]["rpc_port"]), "get_routes", {}, config)
+    )
+
+
+@click.group("rpc", short_help="RPC Client")
+def rpc_cmd() -> None:
+    pass
+
+
+@rpc_cmd.command("endpoints", help="Print all endpoints of a service")
+@click.argument("service", type=click.Choice(services))
+def endpoints_cmd(service: str) -> None:
+    config = load_config(DEFAULT_ROOT_PATH, "config.yaml")
+    try:
+        routes = get_routes(service, config)
+        for route in routes["routes"]:
+            print(route[1:])
+    except Exception as e:
+        print(e)
+
+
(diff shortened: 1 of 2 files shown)


hoffmang9:
I have long wanted exactly this. Very cool!

xdustinface:
> I have long wanted exactly this

Me too! 😄

chia/cmds/rpc.py
```diff
@@ -0,0 +1,140 @@
...
+        try:
+            if not get_routes(service, config)["success"]:
+                raise Exception()
+        except Exception:
+            status = "INACTIVE"
+        print_row(service, status)
+
+
+def create_commands() -> None:
+    for service in services:
```
ChiaMineJP, line 89: Dynamically executing `def` like this is not working because the final `def func_name` will always override the previous `def func_name`. (Maybe it works if function names are different.)
You can try ` chia rpc full_node get_blockchain_state` to check this.
This works if  
`services: List[str] = ["farmer", "full_node", "harvester", "wallet"]`  
is updated to  
`services: List[str] = ["farmer", "harvester", "wallet", "full_node"]`

chia/cmds/rpc.py
```diff
@@ -0,0 +1,140 @@
...
+    except Exception as e:
+        raise Exception(f"Request failed: {e}")
+    finally:
+        client.close()
+        await client.await_closed()
+    return result
+
+
+def print_result(json_dict: Dict[str, Any]) -> None:
+    print(json.dumps(json_dict, indent=4, sort_keys=True))
```
ChiaMineJP, line 41: This might lead to religeous war, but I'd suggest `indent=2` for it does not consume console width.

chia/cmds/rpc.py
xdustinface: Not sure what you mean here? `chia rpc full_node get_blockchain_state` works fine for me.

chia/cmds/rpc.py
```diff
@@ -0,0 +1,139 @@
...
+
+import click
+from aiohttp import ClientResponseError
+
+from chia.rpc.rpc_client import RpcClient
+from chia.util.config import load_config
+from chia.util.default_root import DEFAULT_ROOT_PATH
+from chia.util.ints import uint16
+
+services: List[str] = ["farmer", "full_node", "harvester", "wallet"]
```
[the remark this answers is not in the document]
xdustinface: I considered that both initially, but for `crawler` it doesn't work currently unless i add some ugly hack to pull the port because its inside the `seeder` config. For `timelord` i wasn't very convinced to add it because it doesn't provide any useful endpoint besides the standard ones to handle connections and stop it but i take a look at it later again, im afk now for few hours.

chia/cmds/rpc.py
ChiaMineJP: After I force-pulled your latest commit and error has gone...
What I meant here is repeating `def func_name` with the same `func_name` discards function definition except for the last one.
But you seem like adding `service` variable into the decorators so it eventually defines unique func name per a `service` so error went away.
This has been resolved.

chia/cmds/rpc.py
ChiaMineJP: Fewer indent reduces amount of movement of eyeball and it saves my eye health.
But I won't justify my indent preference any more...

image

image

ChiaMineJP (dismissed):

chia/cmds/rpc.py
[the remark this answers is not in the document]
xdustinface: @ChiaMineJP See f59bb3d281e731e86edaa5e75d58693e6005420f, i added both :)

ChiaMineJP (approved):
I tested full_node, farmer, harvester, wallet with randomly chosen endpoints and it worked fine.

PR #10770, opened 2022-03-16
Remove accidental parameters from calls to setup_simulators_and_walle…

aqk:
…ts and prevent future mistakes

### DIFF
--- a/tests/setup_nodes.py
+++ b/tests/setup_nodes.py
@@ -474,6 +474,7 @@ async def setup_simulators_and_wallets(
     simulator_count: int,
     wallet_count: int,
     dic: Dict,
+    *,
     starting_height=None,
     key_seed=None,
     initial_num_public_keys=5,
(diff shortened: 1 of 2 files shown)


altendky:
Looks like CI is maybe working again.  Closing and reopening to hopefully trigger a run.

arvidn:
is `True` really implicitly converted to an `int` in python?

arvidn (approved):

arvidn (dismissed):

aqk:
Thanks - I'm glad I split this out.

In an earlier patch, I was going to preserve the `starting_height=1`, but after re-reading the named arguments a few more times, and reading some of the tests that use that fixture, I'm pretty sure this wasn't intended.

altendky:
> is `True` really implicitly converted to an `int` in python?

It isn't "converted", rather it _is_ an `int` of value 1.  `bool` inherits from `int`.  For better or worse...

```pycon
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0] on linux
>>> type(True)
<class 'bool'>
>>> bool.__mro__
(<class 'bool'>, <class 'int'>, <class 'object'>)
>>> True == 1
True
```

altendky (approved):
From what I've heard the M1 failure is already being discussed separately.

PR #10775, opened 2022-03-16
Atari merge main 05f96670181654415b1ad219d38463228f30b0ce

altendky:
Other than some order clarification merge resolutions, see below.

```patch
diff --git a/chia/wallet/dlo_wallet/dlo_wallet.py b/chia/wallet/dlo_wallet/dlo_wallet.py
index d1da0ac63..20a694580 100644
--- a/chia/wallet/dlo_wallet/dlo_wallet.py
+++ b/chia/wallet/dlo_wallet/dlo_wallet.py
@@ -254,9 +254,8 @@ class DLOWallet:
         return uint64(amount)
 
     async def get_unconfirmed_balance(self, record_list: Optional[Set[WalletCoinRecord]] = None) -> uint64:
-        confirmed = await self.get_confirmed_balance(record_list)
         # TODO: should the uint128 be changed?
-        return await self.wallet_state_manager._get_unconfirmed_balance(self.id(), uint128(confirmed))  # type: ignore[return-value]  # noqa: E501
+        return await self.wallet_state_manager.get_unconfirmed_balance(self.id(), record_list)  # type: ignore[return-value]  # noqa: E501
 
     async def get_spendable_balance(self, unspent_records: Optional[Set[WalletCoinRecord]] = None) -> uint64:
         spendable_am = await self.wallet_state_manager.get_confirmed_spendable_balance_for_wallet(
```


emlowe (approved):
All looks correct to me

PR #10780, opened 2022-03-16
add optional persistence to SpendSim

Quexington:
This file is purely a test file and these changes should be backwards compatible with any tests that use it already.  It just adds the option to use a different database path than `:memory:`.


emlowe (approved):
seems ok to me

PR #10788, opened 2022-03-17
Bump clvm_tools_rs to fix a problem running as daemon

prozacchiwawa:
caused by old log message that is now eliminated


hoffmang9 (approved):
lgtm

PR #10790, opened 2022-03-18
wallet: Drop `puzzles/genesis_checkers.py` and related puzzles

xdustinface:
Its all duplicated code and puzzles as far as i can tell, see `chia/wallet/puzzles/tails.py`.


Quexington (approved):
Yeah this was a bad merge product, tails.py and the filenames with underscores are what we should use :+1:

PR #10791, opened 2022-03-18
wallet: Fix `CATLineageStore` creation in `create_new_cat_wallet`

xdustinface:
We currently create the `CATLineageStore` before the new asset id exists which leads to all CATs created by `create_new_cat_wallet` using the same lineage store table: `lineage_proofs_0000000000000000000000000000000000000000`.

The test added in ee02c5cf5baea787d6f75c012de4b44a95a5f34f fails without d295e49ca68aa2260b0eb4115ddac8eb456aed88

Note that i fully fixed it in `puzzles/tails.py` only, not in `puzzles/genesis_checker.py` because of #10790.

### DIFF
--- a/chia/wallet/puzzles/tails.py
+++ b/chia/wallet/puzzles/tails.py
@@ -5,7 +5,6 @@ from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.spend_bundle import SpendBundle
 from chia.util.ints import uint64
 from chia.util.byte_types import hexstr_to_bytes
-from chia.wallet.lineage_proof import LineageProof
 from chia.wallet.puzzles.load_clvm import load_clvm
 from chia.wallet.cat_wallet.cat_utils import (
     CAT_MOD,
@@ -38,7 +37,7 @@ class LimitationsProgram:
     @classmethod
     async def generate_issuance_bundle(
         cls, wallet, cat_tail_info: Dict, amount: uint64
-    ) -> Tuple[TransactionRecord, SpendBundle]:
+    ) -> Tuple[TransactionRecord, SpendBundle, bytes32]:
         raise NotImplementedError("Need to implement 'generate_issuance_bundle' on limitations programs")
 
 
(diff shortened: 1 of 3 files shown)


chia/wallet/puzzles/tails.py
```diff
@@ -108,7 +108,7 @@ async def generate_issuance_bundle(cls, wallet, _: Dict, amount: uint64) -> Tupl
                 False,
             )
 
-        return tx_record, SpendBundle.aggregate([tx_record.spend_bundle, signed_eve_spend])
+        return tx_record, SpendBundle.aggregate([tx_record.spend_bundle, signed_eve_spend]), origin_id
```
Quexington, line 111: Origin ID is not the thing we should be returning here (although it might work anyway), I think we want to return the `tail.get_tree_hash()` instead.

Quexington (commented):
This function is not super useful anymore, it's not very critical in wallet operations, it just lets you do a one time issuance of a CAT so I see this change as low risk.  We should probably make a bigger change to do all of the lineage proof is one table maybe as suffix-ing the table name seems a little hacky.  This certainly works for now though.

chia/wallet/puzzles/tails.py
xdustinface: > I think we want to return the `tail.get_tree_hash()` instead.

@Quexington I don't see how that would help? The other way to fix this would be to create the lineage storage inside of `generate_issuance_bundle`, see [0d4c8a0](https://github.com/Chia-Network/chia-blockchain/pull/10791/commits/0d4c8a00a692e351474dbdb9bb065841d6ed3328)

xdustinface:
> This function is not super useful anymore, it's not very critical in wallet operations, it just lets you do a one time issuance of a CAT so I see this change as low risk. 

> We should probably make a bigger change to do all of the lineage proof is one table maybe as suffix-ing the table name seems a little hacky. 

Agree the way its done right now seems hacky but thats not the point here. Wether if its useful or not.. its there and accessible so it should work properly until it gets refactored i would say :)

Quexington (approved):

PR #10793, opened 2022-03-18
produce a correct error message when referencing a block that doesn't exist

arvidn:
When a block reference points to a height that doesn't exist in the blockchain, rather than failing with `KeyError`, fail with `ValueError` and a proper consensus error code.


arvidn:
I think `KeyError` is a better way to fail, to make it more clear that this is a bug, and not an invalid block

PR #10799, opened 2022-03-18
atomic rollback for wallet

wjblanke:


chia/wallet/wallet_node.py
```diff
@@ -590,14 +608,10 @@ async def receive_state_from_peer(
         if self.validation_semaphore is None:
             self.validation_semaphore = asyncio.Semaphore(6)
 
-        # If there is a fork, we need to ensure that we roll back in trusted mode to properly handle reorgs
-        if trusted and fork_height is not None and height is not None and fork_height != height - 1:
-            await self.wallet_state_manager.reorg_rollback(fork_height)
-            await self.wallet_state_manager.blockchain.set_finished_sync_up_to(fork_height)
+        # If there is a fork, we need to ensure that we roll back to properly handle reorgs
+        if fork_height is not None and height is not None and fork_height != height - 1:
```
mariano54, line 612: Rollback is handled in wallet_short_sync_backtrack for untrusted peers, so we don't need to do it here. 
Also it's not safe to rollback, an untrusted peer can give us a super old fork point and make our TX dissapear. 
Probably worth adding a comment about that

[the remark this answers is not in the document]
mariano54: wallet_short_sync_backtrack can safely rollback because we validated the weight for the new peak so we know the peer is telling the truth about the reorg.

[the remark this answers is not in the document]
mariano54: wallet_short_sync_backtrack can safely rollback because we validated the weight for the new peak so we know the peer is telling the truth about the reorg.

chia/wallet/wallet_node.py
```diff
@@ -590,14 +608,10 @@ async def receive_state_from_peer(
...
+            await self.perform_atomic_rollback(fork_height)
         cache: PeerRequestCache = self.get_cache_for_peer(peer)
-        if fork_height is not None:
```
mariano54, line 598: Add this back

mariano54, line 598: Add this back

mariano54 (changes requested):
These seem like good changes except for untrusted rollback

chia/wallet/wallet_node.py
wjblanke: This is done in rollback_request_caches in perform_atomic_rollback for all caches making this superfluous. Is that the wrong thing to do?

mariano54: oh okay it's already done... well we also want to clear the cache for untrusted, and we can do this safely because each peer gets their own cache.

wjblanke: def rollback_request_caches(self, reorg_height: int):
        # Everything after reorg_height should be removed from the cache
        for cache in self.untrusted_caches.values():
            cache.clear_after_height(reorg_height)

I think this clears everybody, including the particular peer cache above. Is there some other cache I need to handle?

mariano54 (dismissed):

PR #10802, opened 2022-03-19
stop helping mkdir() do what it already does

altendky:
I don't really like that we've done this necessarily, but that `mkdir()` isn't python's `mkdir()`.  It is `from chia.util.path import mkdir` https://github.com/Chia-Network/chia-blockchain/blob/657f3c9018c85e65267429b254590b0dd3fafc1c/chia/util/path.py#L18-L23 which uses `pathlib.Path.mkdir(parents=True, exist_ok=True)`.

### DIFF
--- a/tests/util/blockchain.py
+++ b/tests/util/blockchain.py
@@ -1,6 +1,5 @@
 import os
 import pickle
-from os import path
 from pathlib import Path
 from typing import List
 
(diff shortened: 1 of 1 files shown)


xdustinface (approved):

wjblanke:
https://www.python-engineer.com/posts/create-nested-directory/

I think we need mkdirs if we do this?

aqk (approved):

aqk:
It's a bit hidden that this calls `chia.util.path.mkdir` == `pathlib.Path..mkdir(parents=True, ...)`

Probably the original writer may have confused this with `os.mkdir` as well.

Personally I would rename it.

altendky:
I vote we just delete our version entirely at some point. Using the pathlib form directly is legible and not painful.

PR #10828, opened 2022-03-21
Hardcoded SSL test certs/keys

paninaro:
Modified `create_all_ssl()` to optionally take cert/key data to bypass generating new certs/keys. `BlockTools` is the only place that calls `create_all_ssl()` with the optional cert/key data to speed-up test infra setup.

`tests/util/ssl_certs.py` imports several sets of certs/keys and provides functions (`get_next_private_ca_cert_and_key()` and `get_next_nodes_certs_and_keys()`) to cycle through those certs/keys during test setup.

Added `tests/util/gen_ssl_certs.py` as a standalone tool for generating sets of certs/keys that can be included in tests. The files `ssl_certs_{n}.py` have been generated with this tool.


aqk:
Awesome. I haven't looked closely yet, but I expect the problems to crop up in two places:

Places where we connect multiple nodes (e.g. `two_nodes` fixtures, etc)

And maybe `test_ssl.py`

chia/cmds/init_funcs.py
```diff
@@ -236,32 +241,53 @@ def create_all_ssl(root_path: Path):
...
+    ssl_dir: Path,
+    ca_crt: bytes,
+    ca_key: bytes,
+    *,
+    nodes: List[str],
+    private: bool,
+    overwrite: bool = True,
+    node_certs_and_keys: Optional[Dict[str, Dict]] = None,
+):
+    for node_name in nodes:
```
arvidn, line 290: previously we would use `private` to pick which list we would iterate, but now we pass in that list instead. The need for the `private` parameter is not diminished to just affecting the prefix (and it seems like it might be clearer to just pass in the prefix instead, and have it default to "public".

The new API of this function allows you to make the prefix "private" but still use the public nodes, and vice versa. If we at least passed in the prefix instead of the `bool`, I think it would be a bit clearer at the call site if we mix and match public and private.

chia/cmds/init_funcs.py
```diff
@@ -270,6 +296,14 @@ def generate_ssl_for_nodes(ssl_dir: Path, ca_crt: bytes, ca_key: bytes, private:
             prefix = "public"
         key_path = node_dir / f"{prefix}_{node_name}.key"
         crt_path = node_dir / f"{prefix}_{node_name}.crt"
+        if node_certs_and_keys is not None:
```
arvidn, line 299: if feels a bit like this function really are two functions glued together. I suppose `crt_path` and `key_path` are shared between the two parts

tests/util/ssl_certs.py
```diff
@@ -0,0 +1,57 @@
...
+    SSL_TEST_PRIVATE_CA_CERTS_AND_KEYS[idx]
+    for idx in itertools.cycle([*range(len(SSL_TEST_PRIVATE_CA_CERTS_AND_KEYS))])
+)
+
+
+def get_next_private_ca_cert_and_key() -> Tuple[bytes, bytes]:
+    return next(ssl_test_private_ca_certs_and_keys_gen)  # type: ignore[no-any-return]
+
+
+ssl_test_certs_and_keys_gen = (
```
arvidn, line 51: more global variables feels risky. but I can't think of a way for it to cause trouble

arvidn (dismissed):

tests/util/ssl_certs.py
paninaro: Yeah I couldn't easily make these fixtures without updating a bunch of other code (all the places where BlockTools are constructed). I'm open to revisiting this when the rest of the test code settles a bit.

chia/cmds/init_funcs.py
paninaro: Updated. The `private` param is now named `prefix` and takes a string value.

arvidn (approved):

PR #10829, opened 2022-03-22
Handle cases where one node doesn't have the coin we are looking for

emlowe:
Handle the case where one node doesn't have the coin we are looking for and continue to check other nodes for it.

Possibly needs code in the case of zero nodes have this coin, but leaving that edge case for another day.

Maybe needs a test?


mariano54 (approved):
Good change thank you. Maybe we want to put this logic into the `get_coin_state` as well, or pass it as an option. 

We are also calling `get_coin_state` in the offers code... and using a random peer.

PR #10837, opened 2022-03-22
Disable the pytest-monitor plugin in CI if not checking results

altendky:
pytest-monitor uses multiprocessing and has caused multiple confusing
issues.  Perhaps it can be adjusted to not use multiprocessing, but
for now lets just isolate the oddities to where we actually use it.

### DIFF
--- a/.github/workflows/build-test-ubuntu-blockchain.yml
+++ b/.github/workflows/build-test-ubuntu-blockchain.yml
@@ -88,7 +88,7 @@ jobs:
     - name: Test blockchain code with pytest
       run: |
         . ./activate
-        ./venv/bin/py.test tests/blockchain/test_*.py -s -v --durations 0 -m "not benchmark"
+        ./venv/bin/py.test tests/blockchain/test_*.py -s -v --durations 0 -m "not benchmark" -p no:monitor
 
 
 #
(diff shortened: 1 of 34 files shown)


arvidn:
this disables it for all tests, right?

altendky:
It is intended to leave it active for the jobs where we report the collected information.  My understanding is that this is enabled in two directories and is achieved as below.

https://github.com/Chia-Network/chia-blockchain/blob/24bee7d612962cf17097a8391a6e4bc7d02f99f5/tests/core/full_node/config.py#L3-L8

arvidn:
but I don't see any changes to a `config.py` that controls `DISABLE_PYTEST_MONITOR`. perhaps I'm misunderstanding something. I would have expected this to not be set in the generated workflows that do run with monitoring

altendky:
I made it choose based on the presence of `CHECK_RESOURCE_USAGE` in the config instead of adding a new adjacent variable.  But yes, this is a bit muddy.  Note that the `tests/build-workflows.py` generation script has been modified.  I can try to think it through again.  While I do think the idea of tracking resource usage is good, my only experiences with pytest-monitor have been multiple days of debugging issues that it has triggered (even if the causes are down in `multiprocessing` and such).

Perhaps I should change `CHECK_RESOURCE_USAGE` to a boolean and move the value there to another `.yml` like is used for the timelord install step.  Would that be more legible?

And yes, `-p no:monitor` should only be set for workflows that do not have `CHECK_RESOURCE_USAGE` set.  If that isn't what is happening then that is a mistake.  There was only one workflow with that but then I split it into two directories so now there are two.  I don't know if it makes sense on the new one or not.  I don't know if our limits we check against are 'up to date' either.

altendky:
Now `check_resource_usage` works a lot more like `install_timelord`.  Though it does also control the pytest option.

arvidn (approved):

PR #10842, opened 2022-03-22
remove chia.util.path.mkdir()

altendky:
Draft for:
- [x] Self consideration of this and the whole `chia.util.path` file.


aqk (approved):
Thank you. This is much less confusing now!

wjblanke (approved):
aok

Issue #4869, opened 2021-05-13
[BUG] No CLI option to generate a new (additional) receive address (possible in GUI only)


Crumbles-Pieces:
I see nowhere in the command line interface to create new wallet receive addresses. It's possible in the GUI though. I'm referring to the infinite receiving addresses that link back to the same wallet.

Plopsi:
`chia wallet get_address`

Crumbles-Pieces:
> `chia wallet get_address`

that just shows me my first wallet address

Plopsi:
whoops.. you are right, this is definitly a bug. change the headline with a starting [BUG] to mark it

sargonas:
Is `chia wallet get_address` not working as intended? My understanding is this should (and does) generate a receiving wallet address for you on the CLI?

Crumbles-Pieces:
> Is `chia wallet get_address` not working as intended? My understanding is this should (and does) generate a receiving wallet address for you on the CLI?

I thought `chia wallet get_address` would create a new receiving address for a wallet. However, when I use that command it just prints a used address from one of my most recent transactions.

Plopsi:
yes, but it generate a new receiving address anytime called, but currently just returns the current one, so always the same initial address

sargonas:
Ahh i see the problem, you are right, just tested it myself, it just repeats the first address each time. There is no CLI to generate a new address.. I just dug through the CLI code in chia/cmd and looking through it all, I'm pretty confident the ability to do so doesn't exist in there anywhere. I'll convert this to a proper bug and flag it.

altendky:
Since I just looked this up separately...

The `False` here:
https://github.com/Chia-Network/chia-blockchain/blob/66c2399ad5cedafed8fa8af7b61d1eda8082859a/chia/cmds/wallet_funcs.py#L111

Could be filled from a CLI option to let the user request a new address via the `new_address` option here:
https://github.com/Chia-Network/chia-blockchain/blob/66c2399ad5cedafed8fa8af7b61d1eda8082859a/chia/rpc/wallet_rpc_client.py#L121

At least, presuming the rest is plumbed up behind the RPC client there.

altendky:
As was done over in https://github.com/Chia-Network/chia-blockchain/pull/4808.

wrosko:
Can we get an update on this?

altendky:
1.3 got a new sub-command for working with keys.

```console
$ chia keys derive wallet-address --help
[...]

PR #10861, opened 2022-03-23
Added `-n`/`--new-address` option to `chia wallet get_address`

paninaro:
`chia wallet address` defaults to returning the wallet address computed from the most recently derived wallet key. This change introduces a `-n` or `--new-address` option to return a new wallet address using a newly derived wallet key.

Issue #4869

### DIFF
--- a/chia/cmds/wallet.py
+++ b/chia/cmds/wallet.py
@@ -167,7 +167,9 @@ def show_cmd(wallet_rpc_port: Optional[int], fingerprint: int, wallet_type: Opti
 )
 @click.option("-i", "--id", help="Id of the wallet to use", type=int, default=1, show_default=True, required=True)
 @click.option("-f", "--fingerprint", help="Set the fingerprint to specify which wallet to use", type=int)
-@click.option("-n", "--new-address", help="Create a new wallet receive address", is_flag=True, default=False, show_default=True)
+@click.option(
+    "-n", "--new-address", help="Create a new wallet receive address", is_flag=True, default=False, show_default=True
+)
 def get_address_cmd(wallet_rpc_port: Optional[int], id, fingerprint: int, new_address: bool) -> None:
     extra_params = {"id": id, "new_address": new_address}
     import asyncio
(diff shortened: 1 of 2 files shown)


chia/cmds/wallet_funcs.py
```diff
@@ -216,7 +216,8 @@ async def send(args: dict, wallet_client: WalletRpcClient, fingerprint: int) ->
 
 async def get_address(args: dict, wallet_client: WalletRpcClient, fingerprint: int) -> None:
     wallet_id = args["id"]
-    res = await wallet_client.get_next_address(wallet_id, False)
+    new_address: bool = args.get("new_address", False)
```
altendky, line 219: Is the `"new_address"` key going to be any less likely to be present than `"id"`?

Though, the whole `execute_with_wallet()` could just be a context manager probably which would allow elimination of this indirection of parameters through a dict entirely.  But that's entirely unrelated to this PR...

mariano54 (dismissed):

mariano54:
Actually maybe it should default to True? It would increase privacy somewhat.

chia/cmds/wallet.py
```diff
@@ -167,8 +167,11 @@ def show_cmd(wallet_rpc_port: Optional[int], fingerprint: int, wallet_type: Opti
 )
 @click.option("-i", "--id", help="Id of the wallet to use", type=int, default=1, show_default=True, required=True)
 @click.option("-f", "--fingerprint", help="Set the fingerprint to specify which wallet to use", type=int)
-def get_address_cmd(wallet_rpc_port: Optional[int], id, fingerprint: int) -> None:
-    extra_params = {"id": id}
+@click.option(
+    "-n", "--new-address", help="Create a new wallet receive address", is_flag=True, default=False, show_default=True
```
altendky, line 171: I like to make sure that either value can be selected for a flag.  This is valuable for scripting and backwards compatibility if we were to shift to a different default in the future.

I can't swear to the handling of this with both short and long names, but at least the `/` in the long name works.

```suggestion
    "-n/-l", "--new-address/--latest-address", help="Create a new wallet receive address", show_default=True
```

altendky (changes requested):
Thanks for taking this on.  It's an old request.  `:]`

chia/cmds/wallet.py
paninaro: Added the complementing option with an updated help string

chia/cmds/wallet_funcs.py
paninaro: I can imagine `get_address` being called elsewhere. I agree that `execute_with_wallet` could be updated outside of this PR.

altendky: I hope we don't leak the `args: dict` interface out for other use.  But yeah, that can be dealt with independently if any of us choose to.

altendky (approved):

PR #10864, opened 2022-03-24
Check for requesting items when creating an offer

Quexington:


arvidn:
I really think you should add a test to cover this case. And maybe other cases that aren't tested right now too

paninaro (approved):
lgtm

PR #10872, opened 2022-03-24
reduce indentation

arvidn:
in a few functions in blockchain.py by negating early-exit checks and loop continues.

Please review with whitespace changes disabled (and maybe enabled too, to ensure I'm in fact removing the right number of indentation levels).

https://github.com/Chia-Network/chia-blockchain/pull/10872/files?diff=split&w=1

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -359,120 +359,121 @@ class Blockchain(BlockchainInterface):
             return None, None, [], ([], {})
 
         assert peak is not None
-        if block_record.weight > peak.weight:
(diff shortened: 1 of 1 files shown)


altendky (approved):
I think this makes the situation better.  Without getting into the details of how this works, it feels like this code is 'too complicated' and could use some breaking up into smaller functions.

mariano54 (approved):
Good change. The main problem I see here is that hint handling needs to be moved outside of this consensus critical code. (not related to this PR)

wjblanke (approved):
aok

PR #10874, opened 2022-03-24
Atari merge main cf54aae2b79abbdb0682f0640335d1bb24bd6257

altendky:


altendky:
The data layer wallet seems to add some extra layers that acquire locks/transactions and then call existing general wallet functions that were otherwise called without a lock.  This causes some inclusion of changes to 'regular' wallet code.  I would be more concerned about this if we weren't working on https://github.com/Chia-Network/chia-blockchain/pull/10166.  I expect to have to refactor to use that lock/transaction manager to get to correct locking.

altendky:
The M1 macOS installer started failing before due to a long-standing issue error in the GUI repo that was seemingly masked by something on our runner.  The Intel macOS installer has started failing as well.  We did not do any GUI work (yet) for the data layer so I hesitate to bother to pull a branch there to keep those builds happy while doing catch up merges.  Once caught up to `main` I expect these builds to just start working again.  If not...  then yeah, we'll have to figure it out.

Quexington (approved):
Sure, seems fine to me.

PR #8862, opened 2021-10-18
run_generator2 rust call and compact conditions data structure

arvidn:
This patch makes `chia-blockchain` use the new interface to the rust implementation of `run_generator()`.

The main improvement is to use a fixed struct containing all conditions, rather than a list of "arbitrary" conditions (where duplicates and redundant conditions are allowed).

This data structure simplifies the handling and checking of the conditions, and making it more straight-forward. It also reflects the internal data structure in the rust implementation.

It removes boiler plate code and opens the door to further performance improvements (by using the rust type as a streamable type for instance).

For example, this simpler condition summary object has a single field for the sum of all `RESERVE_FEE` conditions. Similarly for `ASSERT_HEIGHT_ABSOLUTE` also only requires a single field (the most strict one). Making this data structure compact on the rust side improves performance (by copying and allocating less memory) and simplifies checking the conditions, since we can have a single check, rather than a linear scan over all conditions.

PR #10877, opened 2022-03-24
Mempool style checking of offer files

Quexington:
This PR is pretty big but there's not a lot of new stuff here in reality.  The mempool code that checks spend bundles has been brought over to the wallet folder and tweaked to work with the wallet rather than the coin store.  We now check offers using this, rather than just checking that the relevant coins are unspent.


Quexington:
We should probably not merge this until #8862 is merged and I've had a chance to rebase

PR #10878, opened 2022-03-24
use rust clvm in Program.run()

arvidn:


richardkiss (commented):
Seems reasonable. We're getting closer to getting rid of the python `clvm` dependency, which is a decent goal.

Quexington (approved):
Seems good to me, this will make `SpendBundle.debug()` so much faster.

PR #10880, opened 2022-03-24
bump clvm_tools dependency

arvidn:
to make every chia-blockchain installation get the new brun that reports cost accurately


richardkiss (approved):

arvidn:
@richardkiss I suppose we should add type annotations to `clvm_tools`, cut a new release, and then bump this version again

PR #10888, opened 2022-03-25
fixup and enable condition checking tests

arvidn:
for the edge cases of absolute timestamp and absolute height conditions in mempool_manager

### DIFF
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -1256,7 +1256,7 @@ class FullNodeAPI:
         return msg
 
     @api_request
-    async def send_transaction(self, request: wallet_protocol.SendTransaction, test=False) -> Optional[Message]:
+    async def send_transaction(self, request: wallet_protocol.SendTransaction, *, test=False) -> Optional[Message]:
         spend_name = request.transaction.name()
         await self.full_node.transaction_queue.put(
             (0, TransactionQueueEntry(request.transaction, None, spend_name, None, test))
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -1256,11 +1256,10 @@ class FullNodeAPI:
         return msg
 
     @api_request
-    async def send_transaction(self, request: wallet_protocol.SendTransaction) -> Optional[Message]:
+    async def send_transaction(self, request: wallet_protocol.SendTransaction, test=False) -> Optional[Message]:
         spend_name = request.transaction.name()
-
         await self.full_node.transaction_queue.put(
-            (0, TransactionQueueEntry(request.transaction, None, spend_name, None, False))
+            (0, TransactionQueueEntry(request.transaction, None, spend_name, None, test))
         )
         # Waits for the transaction to go into the mempool, times out after 45 seconds.
         status, error = None, None
--- a/tests/core/full_node/test_mempool.py
+++ b/tests/core/full_node/test_mempool.py
@@ -79,7 +79,7 @@ def generate_test_spend_bundle(
     return transaction
 
 
-@pytest_asyncio.fixture(scope="module")
+@pytest_asyncio.fixture(scope="function")
 async def two_nodes_mempool(bt, wallet_a):
     async_gen = setup_simulators_and_wallets(2, 1, {})
     nodes, _ = await async_gen.__anext__()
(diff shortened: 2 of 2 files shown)


tests/core/full_node/test_mempool.py
```diff
@@ -277,17 +280,17 @@ async def test_basic_mempool_manager(self, bt, two_nodes_mempool, wallet_a, self
...
-            # tests use the same ones. See comment at the two_nodes_mempool fixture
-            # (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 2, MempoolInclusionStatus.SUCCESS),
-            # (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 3, MempoolInclusionStatus.SUCCESS),
-            # (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 4, MempoolInclusionStatus.PENDING),
-            # (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 5, MempoolInclusionStatus.PENDING),
+            # run the test on. The fixture (two_nodes_mempool) creates 3 blocks,
+            # then condition_tester2 creates another 3 blocks
+            (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 4, MempoolInclusionStatus.SUCCESS),
+            (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 5, MempoolInclusionStatus.SUCCESS),
+            (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 6, MempoolInclusionStatus.PENDING),
```
aqk, line 287: @mariano54  Is `PENDING` the correct return for `ASSERT_HEIGHT_ABSOLUTE 6` when chain height is 6?

arvidn: note that this is testing the mempool. When inserting a spend bundle in the mempool, conditions that *will hold* in the future are still saved, in a separate *pending* queue. That's what the `PENDING` return means.
Testing the consensus (which is done in a different test) these same tests are already enabled.

arvidn:
The reason I have to pass in `test=True` in so many places is because the full node uses wall clock time. I think that's problematic and we should move in a direction of abstracting that away instead. In such case, the clock node clock could be simulated just like the chain timestamps are.

https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/full_node.py#L652

tests/core/full_node/test_mempool.py
aqk: In a funny coincidence, I just made a change that involved `MempoolInclusionStatus`, so I now understand this section better.

chia/full_node/full_node_api.py
```diff
@@ -1256,11 +1256,10 @@ async def request_removals(self, request: wallet_protocol.RequestRemovals) -> Op
         return msg
 
     @api_request
-    async def send_transaction(self, request: wallet_protocol.SendTransaction) -> Optional[Message]:
+    async def send_transaction(self, request: wallet_protocol.SendTransaction, test=False) -> Optional[Message]:
```
aqk, line 1259:
```suggestion
    async def send_transaction(self, request: wallet_protocol.SendTransaction, *, test=False) -> Optional[Message]:
```

aqk: It's nice to force these named arguments to be named, especially boolean arguments.

tests/core/full_node/test_mempool.py
```diff
@@ -277,17 +282,17 @@ async def test_basic_mempool_manager(self, bt, two_nodes_mempool, wallet_a, self
...
+            (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 7, MempoolInclusionStatus.PENDING),
             # genesis timestamp is 10000 and each block is 10 seconds
-            # (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, 10029, MempoolInclusionStatus.SUCCESS),
-            # (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, 10030, MempoolInclusionStatus.SUCCESS),
-            # (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, 10031, MempoolInclusionStatus.FAILED),
-            # (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, 10032, MempoolInclusionStatus.FAILED),
+            (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, 10049, MempoolInclusionStatus.SUCCESS),
```
aqk, line 292: It's such a big jump from 7 to 10049 - why is the calculation so different from the 10029 it replaces?
(I understand why the low numbers changed).

Oh, right, these are seconds, and we are using 10-second intervals to quantize these steps. OK.

Sometimes it's nice to build up the tested value as an explanation (eg. `10000 + 49`), but I don't think that's needed here.

aqk (dismissed):

tests/core/full_node/test_mempool.py
arvidn: the low numbers (4,5,6 and 7) are block *heights*. The large numbers (starting at 10000) are *seconds*. note that those tests are for different conditions

aqk (approved):

PR #10891, opened 2022-03-25
Atari merge main bf2976a2a9dfdd1b71db48030ef79a4a571feae0 (1.3.1)

altendky:
The initial merge went "smoothly" but left behind various errors, several indentation related.  See the commits at the end.  Concern about other not-yet-caught issues is understandable and perhaps this should be broken into multiple catch up merge PRs.

Draft for:
- [x] Consideration of the proper response to the removal of `WalletActionStore.action_done()` in https://github.com/Chia-Network/chia-blockchain/pull/10594.
  - @Quexington suggested that all uses of the actions should be removed in time.  If this works as is, we can continue.  The call in question was added as part of the fix for the high cpu usage/infinite loop issue.  If it is not present now we can continue as is.


emlowe (approved):
manual tests pass - syncing wallet, tracking DL changes, fetching updates.
Approved

PR #10892, opened 2022-03-25
move pytest.ini to the root directory

altendky:


aqk (dismissed):

arvidn:
how does pytest know where the tests are? is that included in the ini file?

altendky:
Earle got us all confused by running just `pytest` and not getting the expected config, yeah.  I forget the details of how it finds tests but it seemed to find them just fine for him.  It looks like we could set `testpaths = tests/` perhaps to explicitly state where to look (https://docs.pytest.org/en/7.1.x/reference/customize.html).

emlowe:
This does make running `pytest` and `pytest tests` the same. The one difference might be if you `cd tests && pytest` - that may be impacted, I'd have to check

arvidn:
my concern is that if pytest need to go search for tests, it might slow it down. If we could say, in `pytest.ini` where the tests are, that might improve things

altendky:
Here are the docs I should have found before...  i'll add that.

https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=testpaths#confval-testpaths
> `testpaths`
> Sets list of directories that should be searched for tests when no specific directories, files or test ids are given in the command line when executing pytest from the [rootdir](https://docs.pytest.org/en/7.1.x/reference/customize.html#rootdir) directory. Useful when all project tests are in a known location to speed up test collection and to avoid picking up undesired tests by accident.

arvidn (approved):

PR #10893, opened 2022-03-26
Type check WalletStateManager

aqk:
Setting type hints for WalletStateManager instances uncovered some other issues. Please see those PRs here:

* Add get_name to WalletProtocol https://github.com/Chia-Network/chia-blockchain/pull/13995
* DID Wallet https://github.com/Chia-Network/chia-blockchain/pull/10932
* CAT Wallet https://github.com/Chia-Network/chia-blockchain/pull/10934
* RL Wallet https://github.com/Chia-Network/chia-blockchain/pull/10935
* Use `uint128` for wallet balances https://github.com/Chia-Network/chia-blockchain/pull/10936


chia/pools/pool_wallet.py
```diff
@@ -320,7 +324,7 @@ async def rewind(self, block_height: int, in_transaction: bool) -> bool:
 
     @staticmethod
     async def create(
-        wallet_state_manager: Any,
+        wallet_state_manager: "WalletStateManager",
```
aqk, line 327: Note: technically, you only need
    `from __future__ import annotations`
OR
    the type name as a string

However, this plays nicer with some tools that perform type checks on the fly, as you are editing the file.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -16,6 +17,9 @@
 from chia.util.ints import uint64, uint32, uint8
 from chia.wallet.util.transaction_type import TransactionType
 
+if TYPE_CHECKING:
+    from chia.wallet.wallet_state_manager import WalletStateManager
+
```
altendky, line 22: I don't super care here, but I would probably default to putting this at the end of the import list.  I don't know if isort cares.

chia/wallet/wallet_node.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import annotations
```
altendky, line 1: ?

chia/wallet/wallet_node.py
```diff
@@ -6,7 +7,7 @@
 import traceback
 from asyncio import CancelledError
 from pathlib import Path
-from typing import Callable, Dict, List, Optional, Set, Tuple, Any, Iterator
+from typing import Callable, Dict, List, Optional, Set, Tuple, Iterator, Any
```
altendky, line 10: ?

chia/wallet/wallet_state_manager.py
```diff
@@ -15,7 +17,10 @@
 from chia.consensus.coinbase import pool_parent_id, farmer_parent_id
 from chia.consensus.constants import ConsensusConstants
 from chia.pools.pool_puzzles import SINGLETON_LAUNCHER_HASH, solution_to_pool_state
-from chia.pools.pool_wallet import PoolWallet
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from chia.pools.pool_wallet import PoolWallet
```
altendky, line 23: I probably wouldn't worry about the conditional import unless it is needed to avoid issues.  Was there a circular import hiding here without breaking stuff?

The typing import should go above in it's proper section though, if kept.

It looks like this applies to the rest of the files below.

chia/wallet/did_wallet/did_wallet.py
aqk: Yeah, was a toss-up of keeping it in alphabetical order, or separating it out

chia/wallet/wallet_node.py
aqk: removed

chia/wallet/wallet_node.py
aqk: I know exactly what happened there, thanks.

chia/wallet/wallet_state_manager.py
aqk: Oh, this is leftovers - this shouldn't be here.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -1136,9 +1141,9 @@ async def get_innerpuz_for_new_innerhash(self, pubkey: G1Element):
         )
 
     async def inner_puzzle_for_did_puzzle(self, did_hash: bytes32) -> Program:
-        record: DerivationRecord = await self.wallet_state_manager.puzzle_store.get_derivation_record_for_puzzle_hash(
-            did_hash
-        )
+        record: Optional[
+            DerivationRecord
+        ] = await self.wallet_state_manager.puzzle_store.get_derivation_record_for_puzzle_hash(did_hash)
```
altendky, line 1146: Just don't hint the variable?

chia/wallet/trade_manager.py
```diff
@@ -230,12 +234,10 @@ async def cancel_pending_offer_safely(
             if wallet is None:
                 continue
 
-            if wallet.type() == WalletType.NFT:
-                new_ph = await wallet.wallet_state_manager.main_wallet.get_new_puzzlehash()
-            else:
-                new_ph = await wallet.get_new_puzzlehash()
+            # All the wallets currently return the standard wallet puz hash
```
altendky, line 237: Looks like DataLayer still has a custom implementation.  Might do the same thing, didn't check yet.

https://github.com/Chia-Network/chia-blockchain/blob/060cf7a90bc6f293368df5647fe10b721ba8238e/chia/data_layer/data_layer_wallet.py#L1076-L1077

chia/wallet/trade_manager.py
```diff
@@ -254,10 +256,11 @@ async def cancel_pending_offer_safely(
...
                 # ATTENTION: new_wallets
-                txs = await wallet.generate_signed_transaction(
-                    [coin.amount], [new_ph], fee=fee_to_pay, coins={coin}, ignore_max_send_amount=True
-                )
-                all_txs.extend(txs)
+                if isinstance(wallet, (CATWallet, NFTWallet, DataLayerWallet)):
+                    txs = await wallet.generate_signed_transaction(
+                        [uint64(coin.amount)], [new_ph], fee=fee_to_pay, coins={coin}, ignore_max_send_amount=True
+                    )
+                    all_txs.extend(txs)
```
altendky, line 263: Should this be an `elif ...:`?  Maybe with an `else:` still to explicitly catch not-yet-considered wallets?  Seems like an exception causing failure would be better than a comment.

chia/wallet/trade_manager.py
```diff
@@ -335,13 +336,14 @@ async def cancel_pending_offers(
...
-                            bundles.append(tx.spend_bundle)
-                            all_txs.append(dataclasses.replace(tx, spend_bundle=None))
+                    if isinstance(wallet, (CATWallet, NFTWallet, DataLayerWallet)):
+                        txs = await wallet.generate_signed_transaction(
+                            [uint64(coin.amount)], [new_ph], fee=fee_to_pay, coins={coin}, ignore_max_send_amount=True
+                        )
+                        for tx in txs:
+                            if tx is not None and tx.spend_bundle is not None:
+                                bundles.append(tx.spend_bundle)
+                                all_txs.append(dataclasses.replace(tx, spend_bundle=None))
```
altendky, line 346: As above.

chia/wallet/trade_manager.py
```diff
@@ -560,14 +567,15 @@ async def _create_offer_for_ids(
...
                     # ATTENTION: new_wallets
-                    txs = await wallet.generate_signed_transaction(
-                        [abs(offer_dict[id])],
-                        [Offer.ph()],
-                        fee=fee_left_to_pay,
-                        coins=set(selected_coins),
-                        puzzle_announcements_to_consume=announcements_to_assert,
-                    )
-                    all_transactions.extend(txs)
+                    if isinstance(wallet, (CATWallet, DataLayerWallet)):
```
altendky, line 570: As above.

tests/pools/test_pool_wallet.py
```diff
@@ -97,7 +97,7 @@ async def mock_get_current_state(self: Any) -> Any:
 
     # Create an empty PoolWallet and populate only the required fields
     wallet = PoolWallet(
-        wallet_state_manager=MockWalletStateManager(),
+        wallet_state_manager=MockWalletStateManager(),  # type: ignore
```
altendky, line 100: When ignoring errors, ignore a specific error.  `# type: ignore[param-type]` or whatever it is.  Also, a comment is generally warranted.  But...  I think in this case it's obvious and a cast would be the most specific thing to do.

```suggestion
        wallet_state_manager=cast(WalletStateManager, MockWalletStateManager()),
```

tests/pools/test_pool_wallet.py
```diff
@@ -180,7 +180,7 @@ async def mock_get_current_state(self: Any) -> Any:
...
```
altendky, line 183:
```suggestion
        wallet_state_manager=cast(WalletStateManager, MockWalletStateManager()),
```

altendky (changes requested):

chia/wallet/trade_manager.py
emlowe: Yea, I'll need to study this a little more because I'm not entirely sure what would/should happen with not-yet-considered wallets. It should probably raise somewhere , I'm just not sure where

chia/wallet/did_wallet/did_wallet.py
emlowe: Good idea

chia/wallet/trade_manager.py
emlowe: yea, still looking into this - but quex thinks it's the same - but needs confirmation

altendky: If confirmed the same, will it be removed?

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -59,6 +59,7 @@
 if TYPE_CHECKING:
     from chia.wallet.wallet_state_manager import WalletStateManager
 
+
```
altendky, line 62: denoise

```suggestion
```

chia/wallet/trade_manager.py
```diff
@@ -488,11 +490,15 @@ async def _create_offer_for_ids(
...
-                        raise ValueError(f"Cannot offer coins from wallet id {wallet.id()}")
-                    coins_to_offer[id] = await wallet.get_coins_to_offer(
-                        asset_id, uint64(abs(amount)), min_coin_amount, max_coin_amount
-                    )
+
+                    if isinstance(wallet, (CATWallet, Wallet)):
+                        coins_to_offer[id] = list(
+                            await wallet.get_coins_to_offer(uint64(abs(amount)), min_coin_amount, max_coin_amount)
+                        )
+                    elif asset_id is not None and isinstance(wallet, (NFTWallet, DataLayerWallet)):
```
altendky, line 498: Maybe the `asset_id` check should be done inside the `if` and provide its own error message?  If we just don't have an asset id, that doesn't mean an NFT wallet is incapable of offers.  Also, I thought asset ID was a CAT term and this is for not-CAT wallets.

chia/wallet/trade_manager.py
```diff
@@ -626,7 +634,9 @@ async def calculate_tx_records_for_offer(self, offer: Offer, validate: bool) ->
                 wallet_id, _ = wallet_info
                 if addition.parent_coin_info in settlement_coin_ids:
                     wallet = self.wallet_state_manager.wallets[wallet_id]
-                    to_puzzle_hash = await wallet.convert_puzzle_hash(addition.puzzle_hash)  # ATTENTION: new wallets
+                    to_puzzle_hash = await self.wallet_state_manager.convert_puzzle_hash(
+                        wallet_id, addition.puzzle_hash
+                    )  # Use WSM call here rather than adding to wallet
```
altendky, line 639: ?

```suggestion
                    )
```

chia/wallet/wallet_protocol.py
```diff
@@ -64,6 +65,9 @@ def require_derivation_paths(self) -> bool:
     def get_name(self) -> str:
         ...
 
+    async def get_puzzle_info(self, nft_id: bytes32) -> PuzzleInfo:
```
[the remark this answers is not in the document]
altendky: Or maybe it isn't given only NFT and DataLayer wallets support this.  Though, with only two supporting it...  Doesn't seem like a protocol candidate to me.

PR #10906, opened 2022-03-27
await the db commit in the async version of set_db_version

aqk:


chia/util/db_version.py
```diff
@@ -19,7 +19,7 @@ async def lookup_db_version(db: aiosqlite.Connection) -> int:
 async def set_db_version_async(db: aiosqlite.Connection, version: int) -> None:
     await db.execute("CREATE TABLE database_version(version int)")
     await db.execute("INSERT INTO database_version VALUES (?)", (version,))
-    db.commit()
+    await db.commit()
```
aqk, line 22: `db.commit` and `set_db_version_async` are both async.

arvidn (approved):

PR #10907, opened 2022-03-27
fix type annotations for get_block_generator()

arvidn:


chia/consensus/multiprocess_validation.py
```diff
@@ -314,8 +317,8 @@ async def pre_validate_blocks_multiprocessing(
             # We ONLY add blocks which are in the past, based on header hashes (which are validated later) to the
             # prev blocks dict. This is important since these blocks are assumed to be valid and are used as previous
             # generator references
-            prev_blocks_dict: Dict[uint32, Union[FullBlock, HeaderBlock]] = {}
```
arvidn, line 317: it looks like someone typoed `uint32` instead of `bytes32`. I don't know why mypy accepted this.

chia/consensus/block_body_validation.py
```diff
@@ -45,7 +46,7 @@ async def validate_block_body(
     height: uint32,
     npc_result: Optional[NPCResult],
     fork_point_with_peak: Optional[uint32],
-    get_block_generator: Callable,
+    get_block_generator: Callable[[BlockInfo], Awaitable[Optional[BlockGenerator]]],
```
arvidn, line 49: even though the function object we pass in here takes an extra (optional) parameter, we don't mention it in the type since we rely on it being optional within this function. Mypy seems fine converting the true function signature to this one

chia/consensus/multiprocess_validation.py
```diff
@@ -169,11 +170,13 @@ async def pre_validate_blocks_multiprocessing(
     constants: ConsensusConstants,
     constants_json: Dict,
     block_records: BlockchainInterface,
-    blocks: Sequence[Union[FullBlock, HeaderBlock]],
+    blocks: Sequence[FullBlock],
     pool: ProcessPoolExecutor,
     check_filter: bool,
     npc_results: Dict[uint32, NPCResult],
-    get_block_generator: Optional[Callable],
+    get_block_generator: Optional[
```
arvidn, line 177: this should probably not be optional actually

arvidn: updated

chia/consensus/block_body_validation.py
altendky: If you are interested, you can hint argument names as well, though it is a bit more verbose.

```python
class SomeCallable(Protocol):
    def __call__(a_parameter: BlockInfo) -> Awaitable[Optional[BlockGenerator]]:
        pass
```

chia/consensus/multiprocess_validation.py
altendky: Yep, and `FullBlock.header_hash` and `HeaderBlock.header_hash` are both `@property` that have no return hints thus leaving them as `Any` and allowing this error to go uncaught by mypy.  Exactly why stricter settings are needed everywhere to avoid unhinted stuff and then later to entirely ban the explicit use of `Any` as well.

altendky (approved):
Reads ok.  Would be nice to expose the issues so that mypy complains and then fix them.

chia/consensus/block_body_validation.py
aqk: I think you need `mypy --no-implicit-optional` to enforce that.

aqk (approved):

PR #10908, opened 2022-03-27
rollback height map

arvidn:
When performing a reorg, we roll back existing blocks to an earlier state. This change makes us *also* roll back the height-to-hash map.

Before the height-to-hash refactor, we did also not roll this back, and I preserved this behaviour. However, there's some really subtle (and way too complex) logic in `get_block_generator()` ([here](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/consensus/blockchain.py#L889)) that determines whether or not we are currently in a reorg based on whether a height-to-hash entry exists. This can (in rare instances) lead to thinking we are *not*, when we in fact are.

This addresses the "in-main-chain" issue gene experienced last week.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -380,6 +380,7 @@ class Blockchain(BlockchainInterface):
         # Rollback sub_epoch_summaries
         self.__height_map.rollback(fork_height)
         await self.block_store.rollback(fork_height)
+        self._peak_height = None if fork_height < 0 else uint32(fork_height)
 
         # Collect all blocks from fork point to new peak
         blocks_to_add: List[Tuple[FullBlock, BlockRecord]] = []
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -391,7 +391,16 @@ class Blockchain(BlockchainInterface):
             cm.callback(restore_height_map)
 
             await self.block_store.rollback(fork_height)
+
+            original_peak = self._peak_height
+
+            def restore_peak() -> None:
+                self._peak_height = original_peak
+
             self._peak_height = None if fork_height < 0 else uint32(fork_height)
+
+            # in case of failure during reorg, restore the _peak_height
+            cm.callback(restore_peak)
 
             # Collect all blocks from fork point to new peak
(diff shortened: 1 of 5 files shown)


chia/consensus/blockchain.py
```diff
@@ -380,6 +380,7 @@ async def _reconsider_peak(
         # Rollback sub_epoch_summaries
         self.__height_map.rollback(fork_height)
         await self.block_store.rollback(fork_height)
+        self._peak_height = None if fork_height < 0 else uint32(fork_height)
```
arvidn, line 383: one possible issue with this is that we don't restore anything in case of a failure. but we don't restore the `__height_map` either.

mariano54: yeah that is a potential issue. In the case of reverting the TX, everything that happened needs to be rolled back

mariano54: Wouldn't this put the blockchain into an inconsistent state? the DB is not committed yet, so we still have all the coins and blocks there (not rolled back yet)

mariano54: Also shouldn't line 389 be:
`while fork_height >= 0` ?

arvidn: > Wouldn't this put the blockchain into an inconsistent state? the DB is not committed yet, so we still have all the coins and blocks there (not rolled back yet)

Yes, it would. In fact, it *is* because the call to `rollback()` is never undone right now.

mariano54 (changes requested):
The desired behavior for receive_block is either it adds the block completely and finishes the reorg or does not add anything and keeps the peak identical. A key requirement of the blockchain is that it can only move forward in weight (you cannot end up with less weight than before). It looks like this is not the case, so it's a bug, but we should not move forward with trying to fix and settle the node into an intermediate state IMO.

arvidn:
@mariano54 I added two more commits that implements the restore behaviour, for when a reorg fails.

The most intrusive aspect is that I introduced an `ExitStack` context manager in the `_reconsider_peak()` function, meaning everything in there was indented one more level (please review with whitespace disabled).

The `ExitStack` is used to add the restore callback functions to, and then pop them at the end if we're successful.

PR #10909, opened 2022-03-27
fix type annotations for FullBlock.header_hash and prev_header_hash

arvidn:
and fix up the fallout

### DIFF
--- a/benchmarks/block_store.py
+++ b/benchmarks/block_store.py
@@ -288,10 +288,10 @@ async def run_add_block_benchmark(version: int):
             print("profiling get_full_blocks_at")
 
         start = monotonic()
-        for h in range(1, block_height):
-            blocks = await block_store.get_full_blocks_at([h])
+        for hi in range(1, block_height):
+            blocks = await block_store.get_full_blocks_at([hi])
             assert len(blocks) == 1
-            assert blocks[0].height == h
+            assert blocks[0].height == hi
 
         stop = monotonic()
         total_time += stop - start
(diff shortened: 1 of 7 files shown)


benchmarks/block_store.py
```diff
@@ -288,10 +288,10 @@ async def run_add_block_benchmark(version: int):
             print("profiling get_full_blocks_at")
 
         start = monotonic()
-        for h in range(1, block_height):
-            blocks = await block_store.get_full_blocks_at([h])
+        for hi in range(1, block_height):
+            blocks = await block_store.get_full_blocks_at([hi])
```
altendky, line 292: This expects a `uint32` but `hi` would be an `int`.  Anyways, this is progress and further fixes can be taken separately.  I would note that i would rather see `height` and `header_hash` above or such than `h` and `hi`.

This is not complained about since `BlockStore.create()` has no return type hint and thus `block_store` is `Any`.

No request for change here as smaller PRs are useful and there's no real expectation that this chain will end.

benchmarks/block_store.py
```diff
@@ -288,10 +288,10 @@ async def run_add_block_benchmark(version: int):
...
             assert len(blocks) == 1
-            assert blocks[0].height == h
+            assert blocks[0].height == hi
```
altendky, line 294: `blocks` is `Any` so even with strict equality checks this would allow the `int` to be compared with the ``uint32`.

altendky (approved):
Thanks!

benchmarks/block_store.py
arvidn: the reason I had to rename this variable is that it's also used as `bytes32` in a few places.

altendky: Yeah, I saw that.  That's what the other comment was about, offering more descriptive names.  The comment here was just pointing out a further hinting 'issue'.  Sort of.

PR #10915, opened 2022-03-27
Add support for excluding coins in `create_signed_transaction` wallet rpc

felixbrucker:
Currently it is not possible to exclude specific coins in the `create_signed_transaction` wallet rpc api, only to select specific coins (allow vs deny list).

This PR adds a new option `exclude_coins` to the `create_signed_transaction` wallet rpc api, which gets passed down into the `select_coins()` method, which already supports exclusion of coins.

This PR is missing tests because i don't understand the current testing methods and their dependencies, it seems all tests use data of previous tests and are not contained in single test methods with each their own setup and teardown. If possible i'd want to add tests ofc.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1699,6 +1699,10 @@ class WalletRpcApi:
         if "coins" in request and len(request["coins"]) > 0:
             coins = set([Coin.from_json_dict(coin_json) for coin_json in request["coins"]])
 
+        exclude_coins = None
+        if "exclude_coins" in request and len(request["exclude_coins"]) > 0:
+            exclude_coins = set([Coin.from_json_dict(coin_json) for coin_json in request["exclude_coins"]])
+
         coin_announcements: Optional[Set[Announcement]] = None
         if (
             "coin_announcements" in request
@@ -1740,6 +1744,7 @@ class WalletRpcApi:
                     bytes32(puzzle_hash_0),
                     fee,
                     coins=coins,
+                    exclude_coins=exclude_coins,
                     ignore_max_send_amount=True,
                     primaries=additional_outputs,
                     memos=memos_0,
@@ -1753,6 +1758,7 @@ class WalletRpcApi:
                 bytes32(puzzle_hash_0),
                 fee,
                 coins=coins,
+                exclude_coins=exclude_coins,
                 ignore_max_send_amount=True,
                 primaries=additional_outputs,
                 memos=memos_0,
--- a/chia/rpc/wallet_rpc_client.py
+++ b/chia/rpc/wallet_rpc_client.py
@@ -208,6 +208,7 @@ class WalletRpcClient(RpcClient):
         coin_announcements: Optional[List[Announcement]] = None,
         puzzle_announcements: Optional[List[Announcement]] = None,
         min_coin_amount: uint64 = uint64(0),
+        exclude_coins: Optional[List[Coin]] = None,
     ) -> TransactionRecord:
         # Converts bytes to hex for puzzle hashes
         additions_hex = []
@@ -245,6 +246,10 @@ class WalletRpcClient(RpcClient):
         if coins is not None and len(coins) > 0:
             coins_json = [c.to_json_dict() for c in coins]
             request["coins"] = coins_json
+
+        if exclude_coins is not None and len(exclude_coins) > 0:
+            exclude_coins_json = [exclude_coin.to_json_dict() for exclude_coin in exclude_coins]
+            request["exclude_coins"] = exclude_coins_json
 
         response: Dict = await self.fetch("create_signed_transaction", request)
         return TransactionRecord.from_json_dict_convenience(response["signed_tx"])
--- a/chia/wallet/wallet.py
+++ b/chia/wallet/wallet.py
@@ -291,6 +291,7 @@ class Wallet:
         memos: Optional[List[bytes]] = None,
         negative_change_allowed: bool = False,
         min_coin_amount: Optional[uint64] = None,
+        exclude_coins: Optional[Set[Coin]] = None,
     ) -> List[CoinSpend]:
         """
         Generates a unsigned transaction in form of List(Puzzle, Solutions)
@@ -312,7 +313,14 @@ class Wallet:
                 raise ValueError(f"Can't send more than {max_send} in a single transaction")
             self.log.debug("Got back max send amount: %s", max_send)
         if coins is None:
-            coins = await self.select_coins(uint64(total_amount), min_coin_amount=min_coin_amount)
+            exclude_coins_list: Optional[List[Coin]] = None
+            if exclude_coins is not None:
+                exclude_coins_list = list(exclude_coins)
+            coins = await self.select_coins(
+                uint64(total_amount), min_coin_amount=min_coin_amount, exclude=exclude_coins_list
+            )
+        elif exclude_coins is not None:
+            raise ValueError("Can't exclude coins when also specifically including coins")
         assert len(coins) > 0
         self.log.info(f"coins is not None {coins}")
         spend_value = sum([coin.amount for coin in coins])
@@ -418,6 +426,7 @@ class Wallet:
         memos: Optional[List[bytes]] = None,
         negative_change_allowed: bool = False,
         min_coin_amount: Optional[uint64] = None,
+        exclude_coins: Optional[Set[Coin]] = None,
     ) -> TransactionRecord:
         """
         Use this to generate transaction.
@@ -443,6 +452,7 @@ class Wallet:
             memos,
             negative_change_allowed,
             min_coin_amount,
+            exclude_coins=exclude_coins,
         )
         assert len(transaction) > 0
         self.log.info("About to sign a transaction: %s", transaction)
--- a/tests/wallet/rpc/test_wallet_rpc.py
+++ b/tests/wallet/rpc/test_wallet_rpc.py
@@ -416,6 +416,39 @@ async def test_create_signed_transaction_with_puzzle_announcement(wallet_rpc_env
 
 
 @pytest.mark.asyncio
+async def test_create_signed_transaction_with_exclude_coins(wallet_rpc_environment: WalletRpcTestEnvironment) -> None:
+    env: WalletRpcTestEnvironment = wallet_rpc_environment
+    wallet_1: Wallet = env.wallet_1.wallet
+    wallet_1_rpc: WalletRpcClient = env.wallet_1.rpc_client
+    full_node_api: FullNodeSimulator = env.full_node.api
+    full_node_rpc: FullNodeRpcClient = env.full_node.rpc_client
+    await generate_funds(full_node_api, env.wallet_1)
+
+    async def it_does_not_include_the_excluded_coins() -> None:
+        selected_coins = await wallet_1_rpc.select_coins(amount=250000000000, wallet_id=1)
+        assert len(selected_coins) == 1
+        outputs = await create_tx_outputs(wallet_1, [(uint64(250000000000), None)])
+
+        tx = await wallet_1_rpc.create_signed_transaction(outputs, exclude_coins=selected_coins)
+
+        assert len(tx.removals) == 1
+        assert tx.removals[0] != selected_coins[0]
+        assert tx.removals[0].amount == uint64(1750000000000)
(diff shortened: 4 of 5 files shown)


chia/wallet/wallet.py
```diff
@@ -332,7 +333,10 @@ async def _generate_unsigned_transaction(
                 raise ValueError(f"Can't send more than {max_send} in a single transaction")
 
         if coins is None:
-            coins = await self.select_coins(total_amount)
+            exclude_coins_list: Optional[List[Coin]] = None
```
aqk, line 336: We'll need a test to exercise these cases.

aqk (dismissed):

aqk:
This is looking great. Two things I'd like to see before we land it:

* Update docs about intended behaviour. Right now, this patch only uses the `exclude_coins` if `coins` is `None` (but not `[]`)

* Add tests to exercise at least these cases: exclude_coins in [None, [], some] x coins in [None, [], some]


@mariano54

chia/wallet/wallet.py
```diff
@@ -332,7 +333,10 @@ async def _generate_unsigned_transaction(
...
+            if exclude_coins is not None:
```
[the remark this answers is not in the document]
mariano54: The first line of the function should be something like:
```
if exclude_coins is: None: exclude_coins = []
if coins is None: coins = []
```
That way, there is no ambiguity about things being None or empty. 

I think it's reasonable to be able to use both, but this should be tested explicitly, and if there is a conflict, it should throw an error. Ignoring what the user passed in is a bad idea and will lead to bugs and complaints.

mariano54 (dismissed):

chia/wallet/wallet.py
felixbrucker: Right. I'd love to add some tests, but as i stated above im a bit lost on where tests start and end, currently they seem to be a little messy. Any pointers are appreciated 👍

chia/wallet/wallet.py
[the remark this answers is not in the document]
felixbrucker: My reason for only using `exclude_coins` when selecting coins is that when a user wants to select specific coins via `coins`, he can already filter those client side, as the inputs for both are the same: `Coin`s. That being said, we can ofc also do a server side filter of `coins` by `exclude_coins` if both are set.

When also removing the optional part of both options we will have to throw early in case the server side filter leaves no coins in the array, because otherwise we'd fallback to auto selecting, because now we only check for empty.

> and if there is a conflict, it should throw an error

what do you mean by that, i mean how would they conflict?

felixbrucker:
Still waiting for answers..

emlowe:
I would suggest adding a test in `test_coin_selection.py` that exercises the exclude option of `select_coins()` (this should really already be there, but it's not)
You could also add a test to `test_wallet_rpc.py` if you look at `test_create_signed_transaction` you can probably replicate much of that test and modify as needed

felixbrucker:
@emlowe Thanks, i added some tests in these files 👍

wallentx:
Hey @felixbrucker, sorry about this. We've been pretty busy the past few weeks, and we haven't done a great job of budgeting the needed time to tending to community PRs. There's definitely been some interest in this particular PR, and we would love to work with you to get this in if you are willing to reopen it.

chia/wallet/wallet.py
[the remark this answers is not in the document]
emlowe: I believe the suggestion is talking about what happens when a non-empty set of `coins` and a non-empty set of `exclude_coins` are both passed in. This PR ignores in this case the `exclude_coins` which I think is what Mariano is talking about. But if you don't ignore it, then what happens if the same coin is in both `coins` and `exclude_coins` - to which the suggestion was to raise an exception. Somewhat edge case, also very easy to check in python since they are both Sets `bool(coins & excluded_coins)`

jack60612:
similar changes in this pr too https://github.com/Chia-Network/chia-blockchain/pull/12360, if you want a point of reference

chia/rpc/wallet_rpc_api.py
```diff
@@ -1696,6 +1696,10 @@ async def create_signed_transaction(self, request, hold_lock=True) -> EndpointRe
         if "coins" in request and len(request["coins"]) > 0:
             coins = set([Coin.from_json_dict(coin_json) for coin_json in request["coins"]])
 
+        exclude_coins = None
+        if "exclude_coins" in request and len(request["exclude_coins"]) > 0:
+            exclude_coins = set([Coin.from_json_dict(coin_json) for coin_json in request["exclude_coins"]])
```
jack60612, line 1701: exclude_coins is a list, not a set.

chia/rpc/wallet_rpc_client.py
```diff
@@ -246,6 +247,10 @@ async def create_signed_transaction(
             coins_json = [c.to_json_dict() for c in coins]
             request["coins"] = coins_json
 
+        if exclude_coins is not None and len(exclude_coins) > 0:
+            exclude_coins_json = [exclude_coin.to_json_dict() for exclude_coin in exclude_coins]
+            request["exclude_coins"] = exclude_coins_json
```
jack60612, line 252: why are we doing this?

chia/wallet/wallet.py
```diff
@@ -291,6 +291,7 @@ async def _generate_unsigned_transaction(
         memos: Optional[List[bytes]] = None,
         negative_change_allowed: bool = False,
         min_coin_amount: Optional[uint128] = None,
+        exclude_coins: Optional[Set[Coin]] = None,
```
jack60612, line 294: why is this not just `Optional[List[Coin]]`? solves a lot of problems

chia/wallet/wallet.py
```diff
@@ -418,6 +424,7 @@ async def generate_signed_transaction(
...
```
jack60612, line 427: List please

chia/wallet/wallet.py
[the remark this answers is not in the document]
jack60612: I say after making it a list, we do if `coins is not None and excluded_coins is not None:` make a copy of coins & a new list and loop through the list only including the non excluded coins. ex:
`old_coins = coins`
`coins: List[Coin] = [] for coin in old_coins:`
       `    if coin is not in excluded_coins:`
        `coins.append(coin)`

jack60612 (changes requested):
Made a couple of comments to simplify the logic and make it more efficient. I put another pr that did something similar below if you need any code to look at.

chia/wallet/wallet.py
felixbrucker: This is a set to be consistent with `coins` from above

jack60612: Why dont we just make everything a list?

felixbrucker: Good question, i found it a bit weird to constantly switch between set and list (both have their up and downsides).
I dont really want to make such "huge" changes, those are probably better suited for a refactoring pr unrelated to this issue that uses set or list consistently across the codebase for coin sets/lists

chia/wallet/wallet.py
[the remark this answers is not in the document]
felixbrucker: @emlowe Is the expected behavior of the method that when passing both in we do a filter of the `coins` using `exclude_coins` first thing? in that case we wouldnt need to throw as well. If no, then what is the expected behavior in this scenario?

chia/wallet/wallet.py
jack60612: > Good question, i found it a bit weird to constantly switch between set and list (both have their up and downsides). I dont really want to make such "huge" changes, those are probably better suited for a refactoring pr unrelated to this issue that uses set or list consistently across the codebase for coin sets/lists

I meant just for exclude_coins, i think the normal coins should stay a set for efficiency reasons

felixbrucker: Having different types feels weird to me

jack60612: Unless you think exclude_coins will end up quite large, which then a set makes sense.

felixbrucker: i mean `coins` and `exclude_coins` are basically the same thing

jack60612: > Having different types feels weird to me

in my mind longer sets are better then longer lists & sets are a bit faster to loop through.

jack60612: > i mean `coins` and `exclude_coins` are basically the same thing

true, but generally exclude_coins will be smaller than coins.
Please don't switch between sets and lists in this pr though, just use lists and this can be a discussion for later

felixbrucker: I kept the api consistent, `exclude_coins` should be the same type as `coins`, anything else is unexpected.

jack60612: > I kept the api consistent, `exclude_coins` should be the same type as `coins`, anything else is unexpected.

fair, ill refactor it later if need be.

chia/wallet/wallet.py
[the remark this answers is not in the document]
felixbrucker: The following piece of code could be added below the coin selection condition:
```
elif exclude_coins is not None:
    if bool(coins & exclude_coins):
        raise ValueError(f"Can't exclude coins specifically included")
    coins = coins - exclude_coins
```
Depending on what the expected behavior of the method is when given both coins and exclude_coins we can also remove the throw before the filter

chia/rpc/wallet_rpc_client.py
felixbrucker: I used the same approach as done just a few lines above for `coins`

chia/rpc/wallet_rpc_api.py
felixbrucker: `exclude_coins` gets passed into `generate_signed_transaction()` which accepts a set, or do you mean something else?

jack60612: sorry i mixed up the api and the client when reviewing

chia/wallet/wallet.py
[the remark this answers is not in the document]
jack60612: I would remove the throw but put a warning saying that this is not recommended.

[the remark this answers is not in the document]
felixbrucker: > I would remove the throw but put a warning saying that this is not recommended.

But if its the expected behavior why warn?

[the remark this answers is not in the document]
jack60612: > > I would remove the throw but put a warning saying that this is not recommended.
> 
> But if its the expected behavior why warn?

fair point, i think it should error out then.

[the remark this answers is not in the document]
felixbrucker: Yes, i agree. So the expected behavior is that when passing both and they intersect we throw. One might even throw directly, even if they dont intersect, because there is no valid case where they are ever used. What do you think?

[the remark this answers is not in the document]
felixbrucker: Updated the PR to reflect that :+1:

jack60612 (dismissed):
LGTM on test pass
coin_selection tests seem fine to me

felixbrucker:
General note on tests: i noticed there are a few flaky (unrelated) tests, however i cant restart failed tests on gha. All recent master commits have failing tests as well.

jack60612:
> General note on tests: i noticed there are a few flaky (unrelated) tests, however i cant restart failed tests on gha. All recent master commits have a few failing tests as well.

no worries, flake is flake.

felixbrucker:
@jack60612 3 unrelated tests are failing

jack60612:
> This pull request has conflicts, please resolve those before we can evaluate the pull request.

sorry :)

felixbrucker:
@jack60612 i noticed we now have inconsistent naming: `exclude` is used throughout the codebase, i added `exclude_coins` to be consistent to that. Now with the latest PRs `excluded_coins` was added as well, so now we have an inconsistent mix.
For users and devs it would probably be best if naming were consistent

felixbrucker:
Ah, ofc main is failing with 70+ test fails :sweat_smile: 

someone please ping me when main is safe to rebase onto again

jack60612:
> Ah, ofc main is failing with 70+ test fails :sweat_smile: 
> 
> 
> 
> someone please ping me when main is safe to rebase onto again

Try it now

jack60612 (approved):
lgtm

emlowe (approved):

wjblanke (approved):
aok

PR #10919, opened 2022-03-28
single thread executor

arvidn:
this adds an option to run the full node in single-thread mode. This is primarily useful when profiling, since we only profile the main process.

### DIFF
--- a/chia/util/inline_executor.py
+++ b/chia/util/inline_executor.py
@@ -1,0 +1,24 @@
+from __future__ import annotations
+
+from concurrent.futures import Executor, Future
+from typing import Callable, TypeVar
+
+_T = TypeVar("_T")
+
+
+class InlineExecutor(Executor):
+    _closing: bool = False
+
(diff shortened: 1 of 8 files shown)


chia/util/inline_executor.py
```diff
@@ -0,0 +1,24 @@
...
+            raise RuntimeError("executor shutting down")
+
+        f: Future[_T] = Future()
+        try:
+            f.set_result(fn(*args, **kwargs))
+        except BaseException as e:  # lgtm[py/catch-base-exception]
+            f.set_exception(e)
+        return f
+
+    def close(self) -> None:
```
aqk, line 23: Is the default `shutdown` method compatible with this self._close state variable?

https://docs.python.org/3/library/concurrent.futures.html#executor-objects

aqk (approved):
I think this is a very safe change, because it is not default, and is behind a command line flag.

PR #10924, opened 2022-03-28
Fix trailing bytes shown in CAT asset ID row when using `chia wallet show`

paninaro:
A wallet RPC change is needed to properly separate out the asset ID from the TAIL program returned by `get_all_wallet_info_entries()`. For now, this change will just fix the display of the asset ID when `chia wallet show` is used.


paninaro:
Thanks to @joshpainter for noticing the issue!

altendky (dismissed):

altendky (approved):

PR #10926, opened 2022-03-28
ignore lack of hinting on clvm_tools.binutils.assemble()

altendky:


paninaro (approved):
I'm not understanding why other similar uses aren't causing mypy to complain (I guess they aren't in a typed context?). This looks benign.

aqk (approved):

aqk:
I'm very familiar with the clvm / clvm_tools code. This is OK - our tools changed, not this code.

Also, this is a good fix - we want to change that particular script as little as possible.

chia/wallet/puzzles/prefarm/make_prefarm_ph.py
```diff
@@ -26,7 +26,9 @@ def make_puzzle(amount: int) -> int:
     puzzle = f"(q . ((51 0x{ph1.hex()} {amount}) (51 0x{ph2.hex()} {amount})))"
     # print(puzzle)
 
-    puzzle_prog = Program.to(binutils.assemble(puzzle))
+    # TODO: properly type hint clvm_tools
```
aqk, line 29: > # TODO: properly type hint clvm_tools

Agreed

PR #10929, opened 2022-03-29
Change client connection errors from info to error

chiameh:
When clients cannot connect to the node, log these as errors, not info.


chiameh:
I propose this log statement be changed to error, as this may hide critical connection failures from logs. See this log where a harvester cannot connect to the node and the connection failures are logged as INFO. Without changing the default log level, nothing was logged indicating why the harvester was not connecting.

 debug log

Jacek-ghub:
I strongly concur with this change.  

With respect of harvester / farmer connections, when harvester cannot connect, that is a catastrophic error from harvester point of view (not from code point of view).

Also, from farmer / harvester point of view, seeing harvesters trying to connect with wrong certificates may indicate a potential attack on the farmer.

Therefore, the log level change should be done on both harvester and farmer side.

aqk:
@mariano54 @arvidn I think logging connection failures as errors makes sense for protocol connections on localhost, or `Trusted` connections (connections that a user may have initiated). But not for everyday connections from light wallet to node peer or node to node peer.

hoffmang9:
Unfortunately, all connections - including all peer to peer gossip transactions - flow through this logic. As such all 225,000 farmers will now have logs full of disconnects that are totally normal but marked error.

To correctly fix this, there is going to have to be an ERROR log statement localized to the specific issue in e.g. harvester.

Jacek-ghub:
I think that we are all on the same page (only local / relevant connections).  However, I would not mind demoting dis-/connections on the remote side to debug level, as those have mostly value to engineers debugging protocol, and are worthless to those 225k farmers (we kind of had similar discussion some time ago, when those peer related disconnections were causing logs to blow up - first dust storm time if I remember correctly)

chiameh:
Oops, I didn't see how global that connection code was. I thought it was just being used for internal peer connections.

Yes to make more sense this will need to differentiate between Trusted and/or local connections. Hopefully someone more familiar with the Chia code than I am can suggest a reasonable way to do that. I'll continue learning Chia code as I can.

Thanks for considering.

aqk:
Thank you for submitting your idea. We've certainly run into a similar issue before. The protocol code is very generic, and that's great. But it doesn't currently include the idea of "services inside 'my' group" vs. outside. (Although we have the idea of trusted, and the ssl certs, so we could certainly add that idea.

Jacek-ghub:
Therefore, maybe you could go with what @hoffmang9 suggested - add a localized ERROR level logs in the harvester and farmer.  Also, maybe you could consider moving the protocol log line to DEBUG to remove it from average logs.

During November dust storm, at least on my setup (INFO level), the code was pumping log lines every millisecond or so, basically killing the box, making tracking anything in those logs impossible.  In my case, those 20MB logs were rotating every 5 minutes or so (yup, tail had hard time to keep up).

chiameh:
In the screenshot, the log ident shows `harvester harvester_server`. Would it make sense to log harvester (and other?) connection errors as errors but leave them as info for the node? If so then maybe we can change the log level using that in some way.

chiameh:
Hi @aqk , Thank you for your feedback on the issue. How about something like this?

    log_function = self.log.error if self.log.name in [ 'harvester_server', 'farmer_server' ] else self.log.info
    log_function(f"{e}")

I tested this in on my node and it works. Connection errors to other nodes from the full node are INFO, while harvester and farmer connection errors are logged as ERROR.

I am not sure if there other logger names that should be included in this list, or if it should also include "harvester" and "farmer" in addition to the _server names. I only saw harvester_server messages, but I was only looking at the full node logs.

Thank you!

emlowe:
The suggested change in the last comment is better. The PR should be updated with this suggested code.
```
log_function = self.log.error if self.log.name in [ 'harvester_server', 'farmer_server' ] else self.log.info
log_function(f"{e}")
```

chiameh:
Hi thanks for reviewing. I created a new PR #12821 with the updated code.

PR #10931, opened 2022-03-29
Narrow checked type with an extra assert

aqk:


tests/core/full_node/stores/test_full_node_store.py
```diff
@@ -426,10 +426,8 @@ async def test_basic_store(self, empty_blockchain, normalized_to_identity: bool
         )
 
         # Get signage point by hash
-        # TODO: address hint error and remove ignore
-        #       error: Argument 1 to "get_signage_point" of "FullNodeStore" has incompatible type "Optional[bytes32]";
-        #       expected "bytes32"  [arg-type]
-        assert store.get_signage_point(saved_sp_hash) is not None  # type: ignore[arg-type]
+        assert saved_sp_hash is not None
```
altendky, line 429: I think this belongs right after the loop where this is assigned.  It seems that this will only happen if either the outer or inner loop do not have any items to iterate.

emlowe:
Closing PR - this change was included in #10962

PR #10932, opened 2022-03-29
Type check DID Wallet

aqk:
This patch contains 3 types of changes:

Use wider uint128 for values representing a wallet balance
Check for possible None from user_store.create_wallet
A look at the type of the second argument of self.wallet_state_manager.get_unconfirmed_balance

### DIFF
--- a/chia/wallet/did_wallet/did_wallet.py
+++ b/chia/wallet/did_wallet/did_wallet.py
@@ -69,11 +69,12 @@ class DIDWallet:
             raise ValueError("Cannot require more IDs than are known.")
         self.did_info = DIDInfo(None, backups_ids, num_of_backup_ids_needed, [], None, None, None, None, False)
         info_as_string = json.dumps(self.did_info.to_json_dict())
-        self.wallet_info = await wallet_state_manager.user_store.create_wallet(
+        new_wallet_info = await wallet_state_manager.user_store.create_wallet(
             "DID Wallet", WalletType.DISTRIBUTED_ID.value, info_as_string
         )
-        if self.wallet_info is None:
+        if new_wallet_info is None:
             raise ValueError("Internal Error")
+        self.wallet_info = new_wallet_info
         self.wallet_id = self.wallet_info.id
         std_wallet_id = self.standard_wallet.wallet_id
         bal = await wallet_state_manager.get_confirmed_balance_for_wallet(std_wallet_id)
@@ -151,15 +152,18 @@ class DIDWallet:
         self.wallet_state_manager = wallet_state_manager
         self.did_info = DIDInfo(None, [], uint64(0), [], None, None, None, None, False)
         info_as_string = json.dumps(self.did_info.to_json_dict())
-        self.wallet_info = await wallet_state_manager.user_store.create_wallet(
+
+        new_wallet_info = await wallet_state_manager.user_store.create_wallet(
             "DID Wallet", WalletType.DISTRIBUTED_ID.value, info_as_string
         )
+        if new_wallet_info is None:
+            raise ValueError("Internal Error")
+        self.wallet_info = new_wallet_info
+
         await self.wallet_state_manager.add_new_wallet(self, self.wallet_info.id)
         # load backup will also set our DIDInfo
         await self.load_backup(filename)
 
-        if self.wallet_info is None:
-            raise ValueError("Internal Error")
         self.wallet_id = self.wallet_info.id
         return self
 
@@ -223,8 +227,7 @@ class DIDWallet:
         return uint64(addition_amount)
 
     async def get_unconfirmed_balance(self, record_list=None) -> uint128:
-        confirmed = await self.get_confirmed_balance(record_list)
-        return await self.wallet_state_manager._get_unconfirmed_balance(self.id(), confirmed)
+        return await self.wallet_state_manager.get_unconfirmed_balance(self.id())
 
     async def select_coins(self, amount, exclude: List[Coin] = None) -> Optional[Set[Coin]]:
         """Returns a set of coins that can be used for generating a new transaction."""
@@ -560,6 +563,8 @@ class DIDWallet:
         )
         pubkey = did_wallet_puzzles.get_pubkey_from_innerpuz(innerpuz)
         index = await self.wallet_state_manager.puzzle_store.index_for_pubkey(pubkey)
+        if index is None:
+            raise RuntimeError(f"Could not find index for pubkey {pubkey}")
         private = master_sk_to_wallet_sk_unhardened(self.wallet_state_manager.private_key, index)
         signature = AugSchemeMPL.sign(private, message)
         # assert signature.validate([signature.PkMessagePair(pubkey, message)])
@@ -627,6 +632,8 @@ class DIDWallet:
         )
         pubkey = did_wallet_puzzles.get_pubkey_from_innerpuz(innerpuz)
         index = await self.wallet_state_manager.puzzle_store.index_for_pubkey(pubkey)
+        if index is None:
+            raise RuntimeError(f"Could not find index for pubkey {pubkey}")
         private = master_sk_to_wallet_sk_unhardened(self.wallet_state_manager.private_key, index)
         signature = AugSchemeMPL.sign(private, message)
         # assert signature.validate([signature.PkMessagePair(pubkey, message)])
@@ -699,6 +706,8 @@ class DIDWallet:
         message = to_sign + coin.name() + self.wallet_state_manager.constants.AGG_SIG_ME_ADDITIONAL_DATA
         pubkey = did_wallet_puzzles.get_pubkey_from_innerpuz(innerpuz)
         index = await self.wallet_state_manager.puzzle_store.index_for_pubkey(pubkey)
+        if index is None:
+            raise RuntimeError(f"Could not find index for pubkey {pubkey}")
         private = master_sk_to_wallet_sk_unhardened(self.wallet_state_manager.private_key, index)
         signature = AugSchemeMPL.sign(private, message)
         # assert signature.validate([signature.PkMessagePair(pubkey, message)])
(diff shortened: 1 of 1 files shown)


chia/wallet/did_wallet/did_wallet.py
```diff
@@ -222,9 +226,11 @@ async def get_pending_change_balance(self) -> uint64:
...
         return uint64(addition_amount)
 
-    async def get_unconfirmed_balance(self, record_list=None) -> uint64:
-        confirmed = await self.get_confirmed_balance(record_list)
-        return await self.wallet_state_manager._get_unconfirmed_balance(self.id(), confirmed)
+    async def get_unconfirmed_balance(self, record_list=None) -> uint128:
+        # confirmed = await self.get_confirmed_balance(record_list)
+        # This seems like a bug - did we intend to also pass in other unspent CoinRecords?
+        # return await self.wallet_state_manager.get_unconfirmed_balance(self.id(), confirmed)
+        return await self.wallet_state_manager.get_unconfirmed_balance(self.id())
```
aqk, line 233: @matt-o-how I'm wondering if the `confirmed` parameter is leftover code from an earlier time.

If not, we'll have to change this line to collect a set of `CoinRecord`s, instead of a count of mojos.

aqk:
This patch contains 3 types of changes:

* Use wider uint128 for values representing a wallet balance
* Check for possible `None` from `user_store.create_wallet`
* A look at the type of the second argument of `self.wallet_state_manager.get_unconfirmed_balance`

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -151,15 +152,18 @@ async def create_new_did_wallet_from_recovery(
         self.wallet_state_manager = wallet_state_manager
         self.did_info = DIDInfo(None, [], uint64(0), [], None, None, None, None, False)
         info_as_string = json.dumps(self.did_info.to_json_dict())
-        self.wallet_info = await wallet_state_manager.user_store.create_wallet(
+
+        new_wallet_info = await wallet_state_manager.user_store.create_wallet(
```
paninaro, line 156:
```suggestion
        new_wallet_info: Optional[WalletInfo] = await wallet_state_manager.user_store.create_wallet(
```

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -189,18 +193,18 @@ def type(cls) -> uint8:
     def id(self):
         return self.wallet_info.id
 
-    async def get_confirmed_balance(self, record_list=None) -> uint64:
+    async def get_confirmed_balance(self, record_list=None) -> uint128:
```
paninaro, line 196: Rather than guessing what size integer should be used for which balance/weight/iters/whatever, we should really have some type definitions to make this easier to reason about. For instance, a WalletBalance type, or maybe just MojoAmount or something. Not suggesting a change here, just dropping a thought.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -189,18 +193,18 @@ def type(cls) -> uint8:
...
+        amount: uint128 = uint128(0)
         for record in record_list:
             parent = self.get_parent_for_coin(record.coin)
             if parent is not None:
-                amount = uint64(amount + record.coin.amount)
+                amount = uint128(amount + record.coin.amount)
 
         self.log.info(f"Confirmed balance for did wallet is {amount}")
-        return uint64(amount)
+        return uint128(amount)
```
paninaro, line 207: nit: I'd probably drop the uint128(), but I understand if you'd prefer to keep it for clarity.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -189,18 +193,18 @@ def type(cls) -> uint8:
...
 
     async def get_pending_change_balance(self) -> uint64:
```
paninaro, line 209: Should this also be uint128 to match the others?

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -560,6 +566,8 @@ async def create_message_spend(self, messages: List[Tuple[int, bytes]], new_inne
         )
         pubkey = did_wallet_puzzles.get_pubkey_from_innerpuz(innerpuz)
         index = await self.wallet_state_manager.puzzle_store.index_for_pubkey(pubkey)
```
paninaro, line 568:
```suggestion
        index: Optional[uint32] = await self.wallet_state_manager.puzzle_store.index_for_pubkey(pubkey)
```

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -627,6 +635,8 @@ async def create_exit_spend(self, puzhash: bytes32):
...
```
paninaro, line 637:
```suggestion
        index: Optional[uint32] = await self.wallet_state_manager.puzzle_store.index_for_pubkey(pubkey)
```

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -699,6 +709,8 @@ async def create_attestment(
         message = to_sign + coin.name() + self.wallet_state_manager.constants.AGG_SIG_ME_ADDITIONAL_DATA
         pubkey = did_wallet_puzzles.get_pubkey_from_innerpuz(innerpuz)
         index = await self.wallet_state_manager.puzzle_store.index_for_pubkey(pubkey)
```
paninaro, line 711:
```suggestion
        index: Optional[uint32] = await self.wallet_state_manager.puzzle_store.index_for_pubkey(pubkey)
```

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -1012,10 +1026,7 @@ async def generate_eve_spend(self, coin: Coin, full_puzzle: Program, innerpuz: P
         spend_bundle = SpendBundle(list_of_solutions, aggsig)
         return spend_bundle
 
-    async def get_frozen_amount(self) -> uint64:
-        return await self.wallet_state_manager.get_frozen_balance(self.wallet_info.id)
-
-    async def get_spendable_balance(self, unspent_records=None) -> uint64:
+    async def get_spendable_balance(self, unspent_records=None) -> uint128:
         spendable_am = await self.wallet_state_manager.get_confirmed_spendable_balance_for_wallet(
```
paninaro, line 1030:
```suggestion
        spendable_am: uint128 = await self.wallet_state_manager.get_confirmed_spendable_balance_for_wallet(
```

chia/wallet/did_wallet/did_wallet.py
aqk: Yes! We should name this important type.

chia/wallet/did_wallet/did_wallet.py
aqk: Thanks, I was considering dropping it, but kept it for clarity

ytx1991:
FYI, the DID in main is far behind the working branch.

chia/wallet/did_wallet/did_wallet.py
aqk: This seems like a bug - did we intend to also pass in other unspent CoinRecords, or did we intend to move to just getting the total mojos?

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -69,11 +69,12 @@ async def create_new_did_wallet(
             raise ValueError("Cannot require more IDs than are known.")
         self.did_info = DIDInfo(None, backups_ids, num_of_backup_ids_needed, [], None, None, None, None, False)
         info_as_string = json.dumps(self.did_info.to_json_dict())
-        self.wallet_info = await wallet_state_manager.user_store.create_wallet(
+        new_wallet_info = await wallet_state_manager.user_store.create_wallet(
```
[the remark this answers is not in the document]
altendky: Why do you prefer this?

chia/wallet/did_wallet/did_wallet.py
altendky: I believe that this is not about clarity but rather correctness.  Our fixed width integer types need some work.  1) they are really expensive to create since they serialized and deserialize on creation and 2) they don't really retain their types very well.

```pycon
>>> from chia.util.ints import uint128
>>> x = uint128(0)
>>> y = x + x
>>> type(y)
<class 'int'>
>>> x += uint128(1)
>>> type(x)
<class 'int'>
```

altendky (commented):
Made some comments to discuss, but we don't need to hold up the PR on them.  Aside from the question for Matt, the rest looks mergeable to me.

chia/wallet/did_wallet/did_wallet.py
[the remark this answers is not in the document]
paninaro: Because the thing being returned could be None, there are now explicit checks testing for None, and the instance variable that the value is being assigned to is of type WalletInfo. When I loaded up the PR in VSCode, the type hinting was showing the type as any, so that didn't seem especially helpful. Maybe it's just a tooling issue, but I generally prefer having clear typing info available. That's why I prefer the explicit type annotations.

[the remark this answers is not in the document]
altendky: The other PR that was referenced would fix (at least part of) the `Any` hinting by fixing the cause of that lack of information.  Explicit is nice, but it can also introduce inaccurately wide hints that cause extra work.  I've seen cases where we hinted like `x: Optional[int] = f()` with `def f() -> int:` resulting in various handling of `None` when there's no `None` to be had.  It also means that changing or correcting hints involves every call site instead of just changing the function definition and then letting the new information propagate.

Anyways, there is interest in the explicitness from more than just you so just sharing opinions here for now.

Related to the existing optionality, a create method returning `None` is a bit odd so I made https://github.com/Chia-Network/chia-blockchain/pull/10957.  The call sites either didn't handle the case or raised an exception anyways.

emlowe:
The DID wallet code has been completely changed since this PR. I think at this point it's just easier to close this and revisit possible changes again in a new PR

PR #10934, opened 2022-03-29
Type check CAT Wallet

aqk:


Quexington (approved):
Yeah this all makes sense to me.

PR #10936, opened 2022-03-29
Use uint128 for wallet balances

aqk:


Quexington (approved):
Yeah, makes sense, I always wondered why these were out of sync with each other.

PR #10944, opened 2022-03-29
restrict click to < 8.1 for black (#10923)

altendky:
https://github.com/pallets/click/issues/2225

Doing this instead of updating since updating black will change several
files due to some formatting change.  I would like to take that on
separately from unbreaking CI.


emlowe (approved):
ok

PR #10945, opened 2022-03-29
atari - more in_transaction handling

altendky:


emlowe (approved):
Testing passes with this. I believe it's correct

PR #10946, opened 2022-03-29
use .add_interested_puzzle_hashes() in DLOWallet

altendky:


Quexington (approved):
This code is not in use so :+1:

PR #10947, opened 2022-03-29
Also throw DB error on double spending a coin

mariano54:
This is not necessary since validation captures all double spends, but it's an extra layer of safety.

I'm not sure if `AND spent_index=0` decreases performance. Thoughts @arvidn ?

### DIFF
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -1,4 +1,7 @@
 from typing import List, Optional, Set, Dict, Any, Tuple
+
+from aiosqlite import Cursor
+
 from chia.protocols.wallet_protocol import CoinState
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.sized_bytes import bytes32
@@ -515,16 +518,27 @@ class CoinStore:
         for coin_name in coin_names:
             r = self.coin_record_cache.get(coin_name)
             if r is not None:
+                if r.spent_block_index != uint32(0):
+                    raise ValueError(f"Coin already spent in cache: {coin_name}")
+
                 self.coin_record_cache.put(
                     r.name, CoinRecord(r.coin, r.confirmed_block_index, index, r.coinbase, r.timestamp)
                 )
             updates.append((index, self.maybe_to_hex(coin_name)))
 
-        if updates != []:
-            async with self.db_wrapper.write_db() as conn:
-                if self.db_wrapper.db_version == 2:
-                    await conn.executemany("UPDATE OR FAIL coin_record SET spent_index=? WHERE coin_name=?", updates)
-                else:
-                    await conn.executemany(
-                        "UPDATE OR FAIL coin_record SET spent=1,spent_index=? WHERE coin_name=?", updates
-                    )
+        assert len(updates) == len(coin_names)
+        async with self.db_wrapper.write_db() as conn:
+            if self.db_wrapper.db_version == 2:
+                ret: Cursor = await conn.executemany(
+                    "UPDATE OR FAIL coin_record SET spent_index=? WHERE coin_name=? AND spent_index=0", updates
+                )
+
+            else:
+                ret = await conn.executemany(
+                    "UPDATE OR FAIL coin_record SET spent=1,spent_index=? WHERE coin_name=? AND spent_index=0",
+                    updates,
+                )
+            if ret.rowcount != len(coin_names):
+                raise ValueError(
+                    f"Invalid operation to set spent, total updates {ret.rowcount} expected {len(coin_names)}"
+                )
--- a/tests/core/full_node/stores/test_coin_store.py
+++ b/tests/core/full_node/stores/test_coin_store.py
@@ -1,4 +1,5 @@
 import logging
+import sqlite3
 from typing import List, Optional, Set, Tuple
 
 import pytest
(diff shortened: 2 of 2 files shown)


tests/core/full_node/stores/test_coin_store.py
```diff
@@ -161,22 +161,40 @@ async def test_set_spent(self, cache_size: uint32, db_version, bt):
...
+                        coins = block.get_included_reward_coins()
+                        records = [await coin_store.get_coin_record(coin.name()) for coin in coins]
 
                     await coin_store._set_spent([r.name for r in records], block.height)
 
+                    if len(records) > 0:
+                        for r in records:
+                            assert (await coin_store.get_coin_record(r.name)) is not None
+
+                        if cache_size > 0:
```
arvidn, line 184: I don't understand why you skip this test when cache_size == 0

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -161,22 +161,40 @@ async def test_set_spent(self, cache_size: uint32, db_version, bt):
...
+                            # Check that we can't spend a coin twice in cache
+                            with pytest.raises(ValueError) as exec_info:
+                                await coin_store._set_spent([r2.name for r2 in records], block.height)
+                            assert "Coin already spent" in exec_info.value.args[0]
+
+                            for r3 in records:
+                                coin_store.coin_record_cache.remove(r3.name)
+
+                        # Check that we can't spend a coin twice in DB
```
arvidn, line 193: I think this comment should say: "Check that we can't spend coins missing in DB"

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -161,22 +161,40 @@ async def test_set_spent(self, cache_size: uint32, db_version, bt):
...
+                            assert (await coin_store.get_coin_record(r.name)) is not None
+
+                        if cache_size > 0:
+                            # Check that we can't spend a coin twice in cache
+                            with pytest.raises(ValueError) as exec_info:
+                                await coin_store._set_spent([r2.name for r2 in records], block.height)
+                            assert "Coin already spent" in exec_info.value.args[0]
+
+                            for r3 in records:
+                                coin_store.coin_record_cache.remove(r3.name)
```
arvidn, line 191: is there a reason why you alternate between `r2` and `r3` for the iterator variable name? I was a bit confused at first, assuming those names were relevant. I think you should use the same name in all loops, and probably not have a number at all

chia/full_node/coin_store.py
```diff
@@ -515,16 +518,27 @@ async def _set_spent(self, coin_names: List[bytes32], index: uint32):
...
-                    await conn.executemany("UPDATE OR FAIL coin_record SET spent_index=? WHERE coin_name=?", updates)
-                else:
-                    await conn.executemany(
-                        "UPDATE OR FAIL coin_record SET spent=1,spent_index=? WHERE coin_name=?", updates
-                    )
+        assert len(updates) == len(coin_names)
+        async with self.db_wrapper.write_db() as conn:
+            if self.db_wrapper.db_version == 2:
+                ret: Cursor = await conn.executemany(
+                    "UPDATE OR FAIL coin_record SET spent_index=? WHERE coin_name=? AND spent_index=0", updates
```
arvidn, line 533: there is an (implicit) index for `coin_name` (since it's the primary key). We also have an index for `spent_index`. If sqlite decides to use the latter index, this operation will be very slow (because all unspent coins have `spent_index=0`, so you get a linear search among those).

I think it would be good to explicitly specify `INDEX BY` the correct index here, just to make sure sqlite and future readers of this code understands what the intention is. However, I don't know how to name the implicit primary key index.

arvidn (changes requested):

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -161,22 +161,40 @@ async def test_set_spent(self, cache_size: uint32, db_version, bt):
...
 
+                    if len(records) > 0:
+                        for r in records:
+                            assert (await coin_store.get_coin_record(r.name)) is not None
+
+                        if cache_size > 0:
+                            # Check that we can't spend a coin twice in cache
+                            with pytest.raises(ValueError) as exec_info:
+                                await coin_store._set_spent([r2.name for r2 in records], block.height)
+                            assert "Coin already spent" in exec_info.value.args[0]
```
arvidn, line 188: there's a fancier way to check the message of the exception.
```
with pytest.raises(ValueError, match='Coin already spent'):
    await coin_store._set_spent([r2.name for r2 in records], block.height)
```

tests/core/full_node/stores/test_coin_store.py
mariano54: it's not missing, it's there but already spent

tests/core/full_node/stores/test_coin_store.py
mariano54: We are testing that we will get an error in the cache if we try to spend a coin which is already spent in the cache

mariano54: If cache size is zero there will be no error in cache because the cache will always be empty. (There will be an error in the DB though)

tests/core/full_node/stores/test_coin_store.py
mariano54: To avoid overlap in names, if you have:
```
for r in records:
    for r in records:
        ...
    print(r)
```
We would get the inner loop's r right?

chia/full_node/coin_store.py
mariano54: We had this problem before but I don't remember the conclusion. Would `index by coin_name` solve the performance issue?

tests/core/full_node/stores/test_coin_store.py
mariano54: thanks

tests/core/full_node/stores/test_coin_store.py
arvidn: oh, I thought you removed the coins on line 191..

tests/core/full_node/stores/test_coin_store.py
arvidn: I see, so the number is the nesting level of `r`? in that case, all `r3` could be renamed `r2` I believe

chia/full_node/coin_store.py
arvidn: something *like* that, yes. but there is no index called `coin_name`. That index is created implicitly

arvidn:
I think this looks good, except I'm worried about which index sqlite will use (especially which index it will use in the future after this file has changed more)

arvidn:
actually, `PRIMARY KEY` implies `UNIQUE`, so sqlite *should* prefer that index.

arvidn (dismissed):

arvidn (approved):

mariano54:
Confirmed to not have a performance impact on the benchmarks

PR #10952, opened 2022-03-29
Quick fix for improper v2 DB initialization when targeting testnet

### LINKED ISSUE #10951: [Bug] Targeting testnet leads to DB corruption
### What happened?

This happens on Windows, MacOS, possibly all other OSs.

Targeting testnet results in a corrupted DB after the first initial successful run: `full_blocks` missing in the `in_main_chain` column, because DB gets created as a v1 DB

How to replicate:

Have a working Chia set up, but no testnet DBs present
`chia stop all -d` (or close the GUI)
Nuke your config.yaml
`chia init`
`chia configure -t t`
Start a chia full_node (either CLI or GUI in Farmer Mode), and everything will run seemingly fine
`chia stop all -d` (or close the GUI)
Try to start the full_node again (either CLI or GUI in Farmer Mode), and the full node will not start successfully

### Version

1.3.2.dev72

Windows

CLI

### Relevant log output

```shell
Daemon not started yet
Starting daemon
chia_harvester: started
chia_farmer: started
chia_full_node: started
chia_wallet: started
PS C:\Users\William\AppData\Local\chia-blockchain\app-1.3.272\resources\app.asar.unpacked\daemon> Traceback (most recent call last):
  File "Lib\site-packages\chia\server\start_full_node.py", line 65, in <module>
  File "Lib\site-packages\chia\server\start_full_node.py", line 60, in main
  File "chia\server\start_service.py", line 286, in run_service
  File "asyncio\runners.py", line 44, in run
  File "asyncio\base_events.py", line 642, in run_until_complete
  File "chia\server\start_service.py", line 280, in async_run_service
  File "chia\server\start_service.py", line 190, in run
  File "chia\server\start_service.py", line 152, in start
  File "chia\full_node\full_node.py", line 210, in _start
  File "chia\full_node\block_store.py", line 73, in create
  File "aiosqlite\core.py", line 184, in execute
  File "aiosqlite\core.py", line 129, in _execute
  File "aiosqlite\core.py", line 102, in run
sqlite3.OperationalError: no such column: in_main_chain
[12528] Failed to execute script 'start_full_node' due to unhandled exception!

-----------

sqlite> .schema
CREATE TABLE database_version(version int);
CREATE TABLE full_blocks(header_hash text PRIMARY KEY, height bigint,  is_block tinyint, is_fully_compactified tinyint, block blob);
CREATE TABLE block_records(header_hash text PRIMARY KEY, prev_hash text, height bigint,block blob, sub_epoch_summary blob, is_peak tinyint, is_block tinyint);
CREATE TABLE sub_epoch_segments_v3(ses_block_hash text PRIMARY KEY,challenge_segments blob);
CREATE INDEX full_block_height on full_blocks(height);
CREATE INDEX is_fully_compactified on full_blocks(is_fully_compactified);
CREATE INDEX height on block_records(height);
CREATE INDEX peak on block_records(is_peak);
CREATE TABLE hints(id INTEGER PRIMARY KEY AUTOINCREMENT, coin_id blob, hint blob);
CREATE TABLE sqlite_sequence(name,seq);
CREATE INDEX hint_index on hints(hint);
CREATE TABLE coin_record(coin_name text PRIMARY KEY, confirmed_index bigint, spent_index bigint, spent int, coinbase int, puzzle_hash text, coin_parent text, amount blob, timestamp bigint);
CREATE INDEX coin_confirmed_index on coin_record(confirmed_index);
CREATE INDEX coin_spent_index on coin_record(spent_index);
CREATE INDEX coin_puzzle_hash on coin_record(puzzle_hash);
CREATE INDEX coin_parent_index on coin_record(coin_parent);
```

paninaro:
This may not be the best way to handle this, but it works as a quick fix.

Issue #10951

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -202,6 +202,7 @@ class FullNode:
                             # this is a new DB file. Make it v2
                             async with self.db_wrapper.write_db() as w_conn:
                                 await set_db_version_async(w_conn, 2)
+                                self.db_wrapper.db_version = 2
                         except sqlite3.OperationalError:
                             # it could be a database created with "chia init", which is
                             # empty except it has the database_version table


chia/full_node/full_node.py
```diff
@@ -202,6 +202,7 @@ def sql_trace_callback(req: str):
                             # this is a new DB file. Make it v2
                             async with self.db_wrapper.write_db() as w_conn:
                                 await set_db_version_async(w_conn, 2)
+                                self.db_wrapper.db_version = 2
```
paninaro, line 205: It might be safer to re-create `self.db_wrapper` instead of poking at its internals. Open to suggestions.

arvidn: oh, I see. I think this is reasonable.
hopefully we can remove this code in a year or so, when we only support v2

arvidn (approved):

arvidn:
I think this would only happen if you delete your db (without deleting settings for instance) and then restart the node.
`chia init` creates the db with the correct version, I'm pretty sure.

paninaro:
> I think this would only happen if you delete your db (without deleting settings for instance) and then restart the node. `chia init` creates the db with the correct version, I'm pretty sure.

I did try deleting my mainnet DB to see if it would be re-created with the mismatching DB version info, but it worked properly. It seemed to go through a different code path than the testnet case. I didn't dig too deeply however.

arvidn:
interesting

PR #10954, opened 2022-03-29
Fix incorrect return in "rollback_to_block"

jack60612:
This might be causing the incorrect wallet balance and spend issues we are experiencing 

- [x] Fix Bug 
- [x] Add tests for the coin store and block reversion

### DIFF
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -424,6 +424,7 @@ class CoinStore:
             self.coin_record_cache.remove(coin_name)
 
         coin_changes: Dict[bytes32, CoinRecord] = {}
+        # Add coins that are confirmed in the reverted blocks to the list of updated coins.
         async with self.db_wrapper.write_db() as conn:
             async with conn.execute(
                 "SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
@@ -435,12 +436,13 @@ class CoinStore:
                     record = CoinRecord(coin, uint32(0), row[1], row[2], uint64(0))
                     coin_changes[record.name] = record
 
-            # Delete from storage
+            # Delete reverted blocks from storage
             await conn.execute("DELETE FROM coin_record WHERE confirmed_index>?", (block_index,))
 
+            # Add coins that are confirmed in the reverted blocks to the list of changed coins.
             async with conn.execute(
                 "SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
-                "coin_parent, amount, timestamp FROM coin_record WHERE confirmed_index>?",
+                "coin_parent, amount, timestamp FROM coin_record WHERE spent_index>?",
                 (block_index,),
             ) as cursor:
                 for row in await cursor.fetchall():
--- a/tests/core/full_node/stores/test_coin_store.py
+++ b/tests/core/full_node/stores/test_coin_store.py
@@ -222,8 +222,10 @@ class TestCoinStoreWithBlocks:
             coin_store = await CoinStore.create(db_wrapper, cache_size=uint32(cache_size))
 
             selected_coin: Optional[CoinRecord] = None
+            all_coins: List[Coin] = []
 
             for block in blocks:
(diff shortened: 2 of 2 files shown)


mariano54:
We were not properly returning the coin changes when reverting a spend in the coin store. This does not affect the full node, but it would affect wallets whose transactions got reorged.

arvidn:
@jack60612 Does the new test fail without the fix?

chia/full_node/coin_store.py
```diff
@@ -435,12 +436,13 @@ async def rollback_to_block(self, block_index: int) -> List[CoinRecord]:
                     record = CoinRecord(coin, uint32(0), row[1], row[2], uint64(0))
                     coin_changes[record.name] = record
 
-            # Delete from storage
+            # Delete reverted blocks from storage
             await conn.execute("DELETE FROM coin_record WHERE confirmed_index>?", (block_index,))
```
arvidn, line 440: shouldn't coins that disappear in the reorg also be reported as "changed" to the wallet?

chia/full_node/coin_store.py
```diff
@@ -435,12 +436,13 @@ async def rollback_to_block(self, block_index: int) -> List[CoinRecord]:
...
 
+            # Add coins that are confirmed in the reverted blocks to the list of changed coins.
             async with conn.execute(
                 "SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
-                "coin_parent, amount, timestamp FROM coin_record WHERE confirmed_index>?",
+                "coin_parent, amount, timestamp FROM coin_record WHERE spent_index>?",
```
arvidn, line 445: this is the main bugfix, right? previously we would only consider coins that appeared after the fork-point as "changed", and now we also include coins that were spent after the fork point.

However, the fact that we delete all the new coins *first* would suggest that we're still failing to report those as changed.

arvidn (commented):
it's not obvious to me that this is correct. It's definitely *more* correct than existing logic, but I would expect us to also report back coins that disappear because of the reorg to the wallet. Which means we have to collect them before deleting them.

Also, I think it's generally a good idea to write *new* tests to test specific features, rather than extending existing ones to test more things.

chia/full_node/coin_store.py
mariano54: The reverted coins are being reported in line 429, I think. If a coin is both created and spent after the fork, it will only be reported as non-existing, not as unspent.

chia/full_node/coin_store.py
mariano54: Yeah those are collected starting in line 429.

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -239,19 +239,44 @@ async def test_rollback(self, cache_size: uint32, db_version, bt):
...
+                    if spend_selected_coin:
+                        break
+
+            assert selected_coin is not None
+            reorg_index = selected_coin.confirmed_block_index
+
+            # The reorg will revert the creation and spend of many coins. It will also revert the spend (but not the
+            # creation) of the selected coin.
+            changed_records = await coin_store.rollback_to_block(reorg_index)
+            assert selected_coin in changed_records
```
arvidn, line 279: maybe I'm misunderstanding this test; but it seems to me we would need to ensure `changed_records` contains exactly the coins we expect to have been reverted by the rollback

mariano54: We could do that but it would make the test a little more fragile, ie if we changed the blocks then the result will be different, also it's more work.

However it's probably a good idea to do that to make the test more useful

jack60612: Just added more checks

jack60612:
> @jack60612 Does the new test fail without the fix?

Just tested it and it does fail.

arvidn (approved):

mariano54 (approved):

PR #10956, opened 2022-03-30
atari - remove some code...

altendky:
Draft for:
- [x] Check with @Quexington to see if this should be removed.  It got removed in one of the exploratory PRs but I forget exactly why and don't recall quex's reference branch.


Quexington (approved):
Yeah get this crap outta here.

PR #10957, opened 2022-03-30
Make WalletUserStore.create_wallet() raise on failure, and return non-optional

altendky:


chia/wallet/wallet_user_store.py
```diff
@@ -65,12 +65,15 @@ async def create_wallet(
                 (id, name, wallet_type, data),
             )
             await cursor.close()
+            wallet = await self.get_last_wallet()
+            if wallet is None:
+                raise ValueError("Failed to get the just-created wallet")
```
altendky, line 70: I went with `ValueError` because that was the specific error used in other handling of this situation here.  Not sure it's the best choice but aiming towards retaining behavior.

mariano54: yeah that makes sense to me

chia/wallet/wallet_user_store.py
```diff
@@ -55,7 +55,7 @@ async def _clear_database(self):
 
     async def create_wallet(
         self, name: str, wallet_type: int, data: str, id: Optional[int] = None, in_transaction=False
-    ) -> Optional[WalletInfo]:
+    ) -> WalletInfo:
```
aqk, line 58: Sweet! I was going to attempt this after getting https://github.com/Chia-Network/chia-blockchain/pull/10893 in.

Feel free to go ahead of me. I can resolve the conflicts in my branch.

chia/wallet/wallet_user_store.py
aqk, line 70: If any of the previous SQL commands fail, will they also throw `ValueError` ?

aqk (approved):

chia/wallet/wallet_user_store.py
altendky: I don't know offhand and it probably depends on how they fail.

altendky:
I dunno, being reported as having a failing job in the pr list.

emlowe (approved):
makes sense to me

PR #10959, opened 2022-03-30
limit test output on CI

arvidn:
by dropping -s and -v. Also, only print the 10 slowest tests, instead of all.

The relevant changes are down here: https://github.com/Chia-Network/chia-blockchain/pull/10959/files#diff-a1ba19578b59bb0e3f08301748324ea5d678bacef69f461df8479f94efd88395


emlowe (dismissed):
100% in support of reducing overall output

mariano54:
Are we confident that failing tests will still output everything?

wjblanke:
conflicts

emlowe (approved):
approved

PR #10962, opened 2022-03-30
Fix remaining linting issues

mariano54:
Nothing very risky here, they were straightforward. There was one I was not sure about, which is in DID, so I didn't do one.

### DIFF
--- a/benchmarks/utils.py
+++ b/benchmarks/utils.py
@@ -1,9 +1,11 @@
+from secrets import token_bytes
+
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
 from chia.util.ints import uint64, uint32, uint8
 from chia.consensus.coinbase import create_farmer_coin, create_pool_coin
 from chia.types.blockchain_format.classgroup import ClassgroupElement
 from chia.types.blockchain_format.coin import Coin
-from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.types.blockchain_format.sized_bytes import bytes32, bytes100
 from chia.types.blockchain_format.vdf import VDFInfo, VDFProof
 from chia.types.blockchain_format.foliage import Foliage, FoliageBlockData, FoliageTransactionBlock, TransactionsInfo
 from chia.types.blockchain_format.pool_target import PoolTarget
@@ -49,16 +51,11 @@ def rewards(height: uint32) -> Tuple[Coin, Coin]:
 
 
 def rand_bytes(num) -> bytes:
-    ret = bytearray(num)
-    for i in range(num):
-        ret[i] = random.getrandbits(8)
-    return bytes(ret)
+    return token_bytes(num)
 
 
 def rand_hash() -> bytes32:
-    # TODO: address hint errors and remove ignores
-    #       error: Incompatible return value type (got "bytes", expected "bytes32")  [return-value]
-    return rand_bytes(32)  # type: ignore[return-value]
+    return bytes32(rand_bytes(32))
 
 
 def rand_g1() -> G1Element:
@@ -72,9 +69,7 @@ def rand_g2() -> G2Element:
 
 
 def rand_class_group_element() -> ClassgroupElement:
-    # TODO: address hint errors and remove ignores
-    #       error: Argument 1 to "ClassgroupElement" has incompatible type "bytes"; expected "bytes100"  [arg-type]
-    return ClassgroupElement(rand_bytes(100))  # type: ignore[arg-type]
+    return ClassgroupElement(bytes100(rand_bytes(100)))
 
 
 def rand_vdf() -> VDFInfo:
--- a/benchmarks/utils.py
+++ b/benchmarks/utils.py
@@ -1,5 +1,3 @@
-from secrets import token_bytes
-
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
 from chia.util.ints import uint64, uint32, uint8
 from chia.consensus.coinbase import create_farmer_coin, create_pool_coin
@@ -51,7 +49,10 @@ def rewards(height: uint32) -> Tuple[Coin, Coin]:
 
 
 def rand_bytes(num) -> bytes:
-    return token_bytes(num)
+    ret = bytearray(num)
+    for i in range(num):
+        ret[i] = random.getrandbits(8)
+    return bytes(ret)
 
 
 def rand_hash() -> bytes32:
--- a/chia/consensus/block_body_validation.py
+++ b/chia/consensus/block_body_validation.py
@@ -245,18 +245,12 @@ async def validate_block_body(
         return root_error, None
 
     # 12. The additions and removals must result in the correct filter
-    byte_array_tx: List[bytes32] = []
+    byte_array_tx: List[bytearray] = []
 
     for coin in additions + coinbase_additions:
-        # TODO: address hint error and remove ignore
-        #       error: Argument 1 to "append" of "list" has incompatible type "bytearray"; expected "bytes32"
-        #       [arg-type]
-        byte_array_tx.append(bytearray(coin.puzzle_hash))  # type: ignore[arg-type]
+        byte_array_tx.append(bytearray(coin.puzzle_hash))
     for coin_name in removals:
-        # TODO: address hint error and remove ignore
-        #       error: Argument 1 to "append" of "list" has incompatible type "bytearray"; expected "bytes32"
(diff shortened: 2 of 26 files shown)


arvidn:
please let this land first: https://github.com/Chia-Network/chia-blockchain/pull/10943

tests/util/network_protocol_data.py
```diff
@@ -66,10 +66,8 @@
     ),
 )
 
-# TODO: address hint error and remove ignore
-#       error: Argument 1 to "PoolTarget" has incompatible type "bytes"; expected "bytes32"  [arg-type]
 pool_target = PoolTarget(
-    bytes.fromhex("d23da14695a188ae5708dd152263c4db883eb27edeb936178d4d988b8f3ce5fc"),  # type: ignore[arg-type]
+    bytes32.from_hexstr("d23da14695a188ae5708dd152263c4db883eb27edeb936178d4d988b8f3ce5fc"),
```
altendky, line 70: I got `bytes32.fromhex()` to work in terms of hints, fwiw.  This is of course fine as well.

altendky (commented):
Thanks for the cleanup!

benchmarks/utils.py
```diff
@@ -49,16 +51,11 @@ def rewards(height: uint32) -> Tuple[Coin, Coin]:
 
 
 def rand_bytes(num) -> bytes:
-    ret = bytearray(num)
-    for i in range(num):
-        ret[i] = random.getrandbits(8)
-    return bytes(ret)
+    return token_bytes(num)
```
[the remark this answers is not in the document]
arvidn: I used `random` specifically to ensure each run is identical in order to have more apples-to-apples comparisons between runs

chia/consensus/blockchain.py
```diff
@@ -809,23 +803,20 @@ async def get_block_records_at(self, heights: List[uint32], batch_size=900) -> L
...
         """
         records: List[BlockRecord] = []
-        hashes = []
+        hashes: List[bytes32] = []
         assert batch_size < 999  # sqlite in python 3.7 has a limit on 999 variables in queries
         for height in heights:
-            hashes.append(self.height_to_hash(height))
+            header_hash: Optional[bytes32] = self.height_to_hash(height)
+            if header_hash is None:
+                raise ValueError(f"Do not have block at height {height}")
```
altendky, line 811: Would this have resulted in an exception before? Or just not gotten a record from the db?  I looked down to the query but haven't traced back up the call stack to see the effect of an incomplete result.

chia/consensus/blockchain.py
```diff
@@ -761,10 +757,8 @@ async def get_header_blocks_in_range(
         hashes = []
         for height in range(start, stop + 1):
             if self.contains_height(uint32(height)):
-                # TODO: address hint error and remove ignore
-                #       error: Incompatible types in assignment (expression has type "Optional[bytes32]", variable has
-                #       type "bytes32")  [assignment]
-                header_hash: bytes32 = self.height_to_hash(uint32(height))  # type: ignore[assignment]
+                header_hash: Optional[bytes32] = self.height_to_hash(uint32(height))
+                assert header_hash is not None
```
altendky, line 761: Would this have resulted in an exception before? Or just not gotten a record from the db?  I looked down to the query but haven't traced back up the call stack to see the effect of an incomplete result.

chia/full_node/full_node_api.py
```diff
@@ -1331,18 +1317,17 @@ async def request_header_blocks(self, request: wallet_protocol.RequestHeaderBloc
...
-        header_hashes = []
+        header_hashes: List[bytes32] = []
         for i in range(request.start_height, request.end_height + 1):
             if not self.full_node.blockchain.contains_height(uint32(i)):
                 reject = RejectHeaderBlocks(request.start_height, request.end_height)
                 msg = make_msg(ProtocolMessageTypes.reject_header_blocks, reject)
                 return msg
-            header_hashes.append(self.full_node.blockchain.height_to_hash(uint32(i)))
+            header_hash: Optional[bytes32] = self.full_node.blockchain.height_to_hash(uint32(i))
+            assert header_hash is not None
```
altendky, line 1327: I believe that previously this would have just returned no results instead of raising.

altendky (commented):
A few more thoughts from a second pass.  I'll try to finish the second pass in the morning.  (for my reference, `chia/server/server.py` is next...)

altendky (commented):
Alrighty, I just finished my second pass over the production code without any more comments.  I would suggest we get other people to review the consensus and timelord changes as well.  Moving where exceptions are raised and what exceptions are raised can have unexpected results, even though early failure is generally better.

benchmarks/utils.py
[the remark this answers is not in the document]
altendky: Right, the seeding is done elsewhere.  What with async and using the global `Random` instance I'm a bit skeptical about exact determinism at this point, but maybe, and I like that direction for sure.

chia/consensus/block_body_validation.py
```diff
@@ -245,18 +245,12 @@ async def validate_block_body(
         return root_error, None
 
     # 12. The additions and removals must result in the correct filter
-    byte_array_tx: List[bytes32] = []
+    byte_array_tx: List[bytearray] = []
```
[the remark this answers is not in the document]
mariano54: We have always used it with bytearray so I don't see a reason to change that in this PR

chia/consensus/blockchain.py
mariano54: Probably an exception in DB, which would  make the whole API call or weight proof creation process fail

benchmarks/utils.py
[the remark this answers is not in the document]
arvidn: @altendky the benchmarks in this directory don't run as part of `pytest`. They do seed the random number generator with a constant.

chia/consensus/block_body_validation.py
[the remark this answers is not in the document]
altendky: Yeah, I'll take this discussion elsewhere and experiment a bit more than what's shown in #9468.

chia/consensus/blockchain.py
altendky: My comment was that the db looked like it would just not match anything thus continuing on instead of raising an exception.

chia/full_node/full_node_api.py
altendky: As noted above, may as well just get the hash and check `None`ness instead of pre-checking and post-asserting.

chia/consensus/blockchain.py
altendky: To be explored separately.

chia/full_node/full_node_api.py
altendky: To be explored separately.

altendky (approved):
I think it would be good for someone more familiar than myself to review the specifically the consensus and timelord changes.

chia/consensus/blockchain.py
altendky: https://github.com/Chia-Network/chia-blockchain/pull/11007

chia/consensus/blockchain.py
```diff
@@ -678,11 +674,11 @@ def block_record(self, header_hash: bytes32) -> BlockRecord:
         return self.__block_records[header_hash]
 
     def height_to_block_record(self, height: uint32) -> BlockRecord:
-        header_hash = self.height_to_hash(height)
-        # TODO: address hint error and remove ignore
-        #       error: Argument 1 to "block_record" of "Blockchain" has incompatible type "Optional[bytes32]"; expected
-        #       "bytes32"  [arg-type]
-        return self.block_record(header_hash)  # type: ignore[arg-type]
+        # Precondition: height is in the blockchain
+        header_hash: Optional[bytes32] = self.height_to_hash(height)
```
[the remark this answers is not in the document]
altendky: Looks like about 33 failure-of-some-form-or-api-return-failures vs. 8 keep-running-in-some-way, as a quick rough estimation searching for all uses of `.height_to_hash()` across the multiple classes that have it (all under `BlockchainInterface`'s guidance I think).  Aside from that I just kind of like to make sure that the keep-running cases have been thought through well and explicitly addressed rather than just happening accidentally because a dev forgot to be really careful about the `None` case.  Also, for APIs, centralizing failure handling seems nice, though there's some complexity with 'just fail' vs. 'fail and ban for awhile' and such.

PR #10964, opened 2022-03-30
assert about averages in test_mempool_update_performance()

altendky:
This is a small improvement.  Future checks could allow for a limited number of outliers that need not count towards the average etc.

### DIFF
--- a/tests/core/full_node/test_mempool_performance.py
+++ b/tests/core/full_node/test_mempool_performance.py
@@ -40,7 +40,7 @@ class Statistics:
     maximum: float
     mean: float
     percentile_90: float
-    data: Tuple[float]
+    data: Tuple[float, ...]
 
     @classmethod
     def create(cls, data: Iterable[float]):
--- a/tests/core/full_node/test_mempool_performance.py
+++ b/tests/core/full_node/test_mempool_performance.py
@@ -1,7 +1,8 @@
 # flake8: noqa: F811, F401
-
+from dataclasses import dataclass, field
 import logging
-from statistics import mean
+from statistics import mean, quantiles
+from typing import Iterable, List, Tuple
 import time
 
 import pytest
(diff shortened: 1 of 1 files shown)


tests/core/full_node/test_mempool_performance.py
```diff
@@ -69,12 +70,22 @@ async def test_mempool_update_performance(self, bt, wallet_nodes_mempool_perf, d
...
+                slow_durations.append(duration)
                 assert duration < 0.1
             else:
+                fast_durations.append(duration)
                 assert duration < 0.0003
+
+        fast_mean = mean(fast_durations)
+        slow_mean = mean(slow_durations)
+
+        assert fast_mean < 0.0002 and slow_mean < 0.1
```
mariano54, line 91: 0.0002 is really fast, i would change it to something like 0.001, that is fast enough anyway

altendky: Chatting with Arvid about this as well.  I was just putting it back where it was before the bump to 0.0003.  I was hoping that the averaging would suffice for the moment.

arvidn:
I think this patch makes sense. I don't have a strong opinion about the specific threshold, but I think it would be a slight improvement to look at 90th percentile instead of mean. (I'm guessing the python stats package has an easy way to do that)

tests/core/full_node/test_mempool_performance.py
```diff
@@ -69,12 +92,24 @@ async def test_mempool_update_performance(self, bt, wallet_nodes_mempool_perf, d
...
-                assert duration < 0.0003
+                fast_durations.append(duration)
+
+        fast_statistics = Statistics.create(data=fast_durations)
+        slow_statistics = Statistics.create(data=slow_durations)
+
+        message = f"fast: {fast_statistics}\n\nslow: {slow_statistics}"
+        print(f"\n\n{message}")
+        assert (
+            fast_statistics.percentile_90 < 0.00005 and slow_statistics.percentile_90 < 0.1
```
altendky, line 114: 90th percentile on 3 samples isn't clearly a good idea.  Maybe it's ok though since it seems to keep 2 and ignore 1?

tests/core/full_node/test_mempool_performance.py
```diff
@@ -69,12 +92,24 @@ async def test_mempool_update_performance(self, bt, wallet_nodes_mempool_perf, d
         blocks = bt.get_consecutive_blocks(3, blocks)
         await full_node_api_1.full_node.respond_block(full_node_protocol.RespondBlock(blocks[-3]))
 
+        fast_durations: List[float] = []
+        slow_durations: List[float] = []
+
         for idx, block in enumerate(blocks):
             start_t_2 = time.time()
```
altendky, line 99: Do we specifically want to be measuring wall-clock time as opposed to maybe process time or thread time?  I haven't dug into the code.

arvidn (approved):

altendky:
Of course...

wjblanke:
conflict unfortunately

mariano54 (changes requested):
This pr makes this check less useful, because it's not testing the block with the TX in it (due to percentiles). Discussed already with Kyle.

PR #10965, opened 2022-03-30
Ms.flaky gen speed

mariano54:


aqk (approved):

tests/core/test_cost_calculation.py
```diff
@@ -207,7 +207,7 @@ async def test_tx_generator_speed(self, softfork_height):
         assert len(npc_result.npc_list) == LARGE_BLOCK_COIN_CONSUMED_COUNT
         log.info(f"Time spent: {duration}")
 
-        assert duration < 1
+        assert duration < 2
```
aqk, line 210: btw, I was talking with Arvid - I think that these wall-clock time checks are great for certain benchmarks, for for really consensus-critical ones like this, I think we need to do something a little more controlled. Maybe on a machine with a standardized CPU?

aqk (approved):

PR #10966, opened 2022-03-30
Ms.parallel pool t

mariano54:
This used to be the slowest test.

### DIFF
--- a/tests/build-workflows.py
+++ b/tests/build-workflows.py
@@ -83,7 +83,10 @@ def generate_replacements(conf, dir):
         ] = "# Omitted checking out blocks and plots repo Chia-Network/test-cache"
     if not conf["install_timelord"]:
         replacements["INSTALL_TIMELORD"] = "# Omitted installing Timelord"
-    replacements["PYTEST_PARALLEL_ARGS"] = " -n 4" if conf["parallel"] else " -n 0"
+    if conf.get("custom_parallel_n", None):
+        replacements["PYTEST_PARALLEL_ARGS"] = f" -n {conf['custom_parallel_n']}"
+    else:
+        replacements["PYTEST_PARALLEL_ARGS"] = " -n 4" if conf["parallel"] else " -n 0"
     if conf["job_timeout"]:
         replacements["JOB_TIMEOUT"] = str(conf["job_timeout"])
     replacements["TEST_DIR"] = "/".join([*dir.relative_to(root_path.parent).parts, "test_*.py"])
--- a/tests/pools/test_pool_rpc.py
+++ b/tests/pools/test_pool_rpc.py
@@ -372,7 +372,7 @@ class TestPoolWalletRpc:
         assert len(await wallet_node_0.wallet_state_manager.tx_store.get_unconfirmed_for_wallet(3)) == 0
         # Doing a reorg reverts and removes the pool wallets
         await full_node_api.reorg_from_index_to_new_index(ReorgProtocol(uint32(0), uint32(20), our_ph_2))
-        await asyncio.sleep(5)
+        await time_out_assert(30, wallet_is_synced, True, wallet_node_0, full_node_api)
         summaries_response = await client.get_wallets()
         assert len(summaries_response) == 1
 
--- a/tests/pools/test_pool_rpc.py
+++ b/tests/pools/test_pool_rpc.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import tempfile
 from dataclasses import dataclass
 from pathlib import Path
 from shutil import rmtree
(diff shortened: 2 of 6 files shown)


tests/pools/test_pool_rpc.py
```diff
@@ -400,7 +398,7 @@ async def test_create_multiple_pool_wallets(self, one_wallet_node_and_rpc, fee,
 
         # Test creation of many pool wallets. Use untrusted since that is the more complicated protocol, but don't
         # run this code more than once, since it's slow.
-        if fee == 0 and not trusted:
+        if not trusted:
```
arvidn, line 401: this change seems unrelated

tests/pools/test_pool_rpc.py
```diff
@@ -165,9 +163,9 @@ async def setup(two_wallet_nodes, bt, self_hostname):
 
 class TestPoolWalletRpc:
     @pytest.mark.asyncio
-    @pytest.mark.parametrize("trusted", [True, False])
-    @pytest.mark.parametrize("fee", [0, FEE_AMOUNT])
-    async def test_create_new_pool_wallet_self_farm(self, one_wallet_node_and_rpc, fee, trusted, self_hostname):
+    @pytest.mark.parametrize("trusted_and_fee", [(True, FEE_AMOUNT), (False, 0)])
```
arvidn, line 166: how can we know that these are unrelated? and did you pick this combination arbitrarily?

tests/pools/test_pool_rpc.py
```diff
@@ -693,10 +691,14 @@ async def test_absorb_pooling(self, one_wallet_node_and_rpc, fee, trusted, bt, s
             assert (250000000000 + fee) in [tx.additions[0].amount for tx in tx1]
 
     @pytest.mark.asyncio
-    @pytest.mark.parametrize("trusted", [True])
-    @pytest.mark.parametrize("fee", [0])
-    async def test_self_pooling_to_pooling(self, setup, fee, trusted, self_hostname):
-        """This tests self-pooling -> pooling"""
+    @pytest.mark.parametrize("trusted_and_fee", [(True, 0), (False, 0)])
```
arvidn, line 694: `fee` is always 0 here...

tests/pools/test_pool_rpc.py
```diff
@@ -793,8 +795,8 @@ async def tx_is_in_mempool(wid, tx: TransactionRecord):
                 fetched: Optional[TransactionRecord] = await client.get_transaction(wid, tx.name)
                 return fetched is not None and fetched.is_in_mempool()
 
-            await time_out_assert(5, tx_is_in_mempool, True, wallet_id, join_pool_tx)
-            await time_out_assert(5, tx_is_in_mempool, True, wallet_id_2, join_pool_tx_2)
+            await time_out_assert(10, tx_is_in_mempool, True, wallet_id, join_pool_tx)
+            await time_out_assert(10, tx_is_in_mempool, True, wallet_id_2, join_pool_tx_2)
```
arvidn, line 799: was this flaky? I would expect speeding up the tests would mean shorter timeouts

arvidn (commented):
the tests that take this long, do they create plots?
https://github.com/Chia-Network/chia-blockchain/runs/5763094009?check_suite_focus=true#step:9:82

tests/pools/test_pool_rpc.py
mariano54: I don't think they were flaky, but when running with -n 4 I saw this fail, so i increased it. I think creating plots was stalling the cores and therefore there were timeouts in the node.

mariano54:
> the tests that take this long, do they create plots? https://github.com/Chia-Network/chia-blockchain/runs/5763094009?check_suite_focus=true#step:9:82

Yes but that's not the reason for the slowdown. That test in particular creates many blocks and wallets so it's very slow.

tests/pools/test_pool_rpc.py
arvidn: I see. in fact, in the previous version we only test this with `trusted=True` and `fee=0`. You added one more case.
But `fee` should probably not be a parameter at all to this test then

tests/pools/test_pool_rpc.py
arvidn: I see. so, technically, this is a separate, independent, test parameter.

arvidn (approved):

tests/pools/test_pool_rpc.py
mariano54: yeah good point, I guess i can parametrize that differently

tests/build-workflows.py
```diff
@@ -83,7 +83,10 @@ def generate_replacements(conf, dir):
         ] = "# Omitted checking out blocks and plots repo Chia-Network/test-cache"
     if not conf["install_timelord"]:
         replacements["INSTALL_TIMELORD"] = "# Omitted installing Timelord"
-    replacements["PYTEST_PARALLEL_ARGS"] = " -n 4" if conf["parallel"] else " -n 0"
+    if conf.get("custom_parallel_n", None):
```
altendky, line 86: Instead of having it be conditionally present we can set defaults in `testconfig.py`.

https://github.com/Chia-Network/chia-blockchain/blob/8833cc351c49a7e9340626c7c9e4715a0ecd3626/tests/testconfig.py#L1-L12

mariano54: I didn't see that file, thank you

PR #10968, opened 2022-03-31
Simplify Blockchain.get_header_blocks_in_range()

altendky:
1. Avoid a copy
2. Avoid reiteration
3. Avoid continuing iteration when all remaining heights will be ignored
4. Avoid removal from a list which is O(n)
   - First it searches to find the item, then it has to shift the rest of the items after that down an index.

I have not profiled nor benchmarked anything.  This is an off the cuff optimization opportunity that I figured I would just write.  It may or may not be worth changing.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -765,7 +765,7 @@ class Blockchain(BlockchainInterface):
         if peak_height is None:
             return {}
 
-        for height in range(start, min(stop, peak_height + 1)):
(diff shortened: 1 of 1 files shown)


mariano54 (commented):
We only call with maybe up to 1000 items ATM so I think it's only worth changing if we add a test to ensure the results are correct. AFAIK this function is not directly tested aside from integration tests of weight proof and wallet. It might be a slight improvement in the weight proof case. In the wallet case we only call it with 32.

PR #10974, opened 2022-03-31
fix test_full_sync.py to only feed the blocks in the main chain

arvidn:
to the node. This avoids an exception thrown by `receive_block()`, and the test doesn't have to handle it.

Also, there was a missing `mkdir()` for the v1 database case in `chia init`.

### DIFF
--- a/chia/cmds/init_funcs.py
+++ b/chia/cmds/init_funcs.py
@@ -502,6 +502,7 @@ def chia_init(
             db_path_replaced = new_db_path.replace("CHALLENGE", config["selected_network"])
             db_path = path_from_root(root_path, db_path_replaced)
 
+            mkdir(db_path.parent)
             with sqlite3.connect(db_path) as connection:
                 set_db_version(connection, 1)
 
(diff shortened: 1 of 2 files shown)


chia/cmds/init_funcs.py
```diff
@@ -502,6 +502,7 @@ def chia_init(
             db_path_replaced = new_db_path.replace("CHALLENGE", config["selected_network"])
             db_path = path_from_root(root_path, db_path_replaced)
 
+            mkdir(db_path.parent)
```
aqk, line 505: Nothing to do with this patch, but it weirds me out that we have a utility function with the same name as `os.mkdir`, but different behaviour.

aqk (approved):

chia/cmds/init_funcs.py
mariano54: Agreed, that is not good

tools/test_full_sync.py
```diff
@@ -132,6 +134,9 @@ def main() -> None:
     help="run node in a single process, to include validation in profiles",
 )
 def run(file: Path, db_version: int, profile: bool, single_thread: bool) -> None:
+    """
```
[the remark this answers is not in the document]
arvidn: Right. I was confused when I first wrote this program exactly which database the version flag referred to (and I think it was just unused). Requiring the input database be v2 simplifies things a bit.

PR #10983, opened 2022-03-31
fix performance tests

arvidn:
There were two performance tests that hadn't been marked as such. I also bumped the time limit on the mempool performance test since it seems to have a lot of jitter in CI.


altendky (approved):
I reran the benchmark workflow for a total of 10 runs and all passed.  Seems more consistently passing than before.

PR #10984, opened 2022-03-31
Improve handling of unknown pending balances (likely change from addi…

### LINKED ISSUE #10882: [Bug] cancel_offer function from chia/cmds/wallet_funcs.py fails
### What happened?

While running `chia wallet cancel_offer -id <id_string> -f <fingerprint> -m <fee>`, the cancel_offer function from chia/cmds/wallet_funcs.py fails while attempting to print the offer details (trade record), printing the following message: `Exception from 'wallet' non-hexadecimal number found in fromhex() arg at position 0`.

Commenting out line 434, `await print_trade_record(trade_record, wallet_client, summaries=True)` is sufficient to make the aforementioned function work. Sorry that I don't have time to track it down any further.

### Version

1.3.1

Linux

CLI

### Relevant log output

```shell
N/A
```

paninaro:
…ng a maker fee).

Minor improvement for fingerprint selection -- enter/return selects the logged-in fingerprint.

### DIFF
--- a/chia/cmds/wallet_funcs.py
+++ b/chia/cmds/wallet_funcs.py
@@ -355,11 +355,7 @@ async def print_trade_record(record, wallet_client: WalletRpcClient, summaries: 
         print("Summary:")
         offer = Offer.from_bytes(record.offer)
         offered, requested = offer.summary()
-        outbound_balances: Dict[str, int] = {
-            asset_id: v
-            for asset_id, v in offer.get_pending_amounts().items()
-            if asset_id in offered or asset_id == "unknown"
-        }
+        outbound_balances: Dict[str, int] = offer.get_pending_amounts()
         fees: Decimal = Decimal(offer.bundle.fees())
         cat_name_resolver = wallet_client.cat_asset_id_to_name
         print("  OFFERED:")
--- a/chia/cmds/wallet_funcs.py
+++ b/chia/cmds/wallet_funcs.py
@@ -4,7 +4,7 @@ import sys
 import time
 from datetime import datetime
 from decimal import Decimal
-from typing import Any, Callable, Coroutine, Dict, List, Optional, Tuple
+from typing import Any, Awaitable, Callable, Dict, List, Optional, Tuple
 
 import aiohttp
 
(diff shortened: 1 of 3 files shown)


chia/cmds/wallet_funcs.py
```diff
@@ -308,21 +310,37 @@ def timestamp_to_time(timestamp):
     return datetime.fromtimestamp(timestamp).strftime("%Y-%m-%d %H:%M:%S")
 
 
-async def print_offer_summary(wallet_client: WalletRpcClient, sum_dict: dict):
+async def print_offer_summary(cat_name_resolver: CATNameResolver, sum_dict: Dict[str, int], has_fee: bool = False):
```
paninaro, line 313: passing in `cat_name_resolver` instead of `wallet_client` makes it easier to write a unit test to test this function.

chia/cmds/wallet_funcs.py
```diff
@@ -543,15 +569,21 @@ async def get_wallet(wallet_client: WalletRpcClient, fingerprint: int = None) ->
...
-            if not val.isdigit():
+            elif val == "" and logged_in_fingerprint is not None:
+                fingerprint = logged_in_fingerprint
+                break
+            elif not val.isdigit():
                 val = None
             else:
                 index = int(val) - 1
-                if index >= len(fingerprints):
+                if index < 0 or index >= len(fingerprints):
```
paninaro, line 586: This fixed an issue I found where you could enter `0` when selecting a key which would lead to `index == -1`

chia/cmds/wallet_funcs.py
```diff
@@ -308,21 +310,37 @@ def timestamp_to_time(timestamp):
...
+        mojo_amount: int = int(Decimal(amount))
+        name: str = "XCH"
+        if asset_id != "xch":
             name = asset_id
-            unit = units["cat"]
-            if result is not None:
-                wid = str(result[0])
-                name = result[1]
-        print(f"    - {name} (Wallet ID: {wid}): {Decimal(int(amount)) / unit} ({int(Decimal(amount))} mojos)")
+            if asset_id == "unknown":
```
paninaro, line 322: This is the new bit of handling for "unknown" outbound balances. We believe the amount is likely to be the returned change associated with the included fee coin(s), but we aren't 100% sure, hence the description set below.

Quexington (changes requested):
These are great changes and well tested, but can we also remove the line from `Offer.get_pending_balances` that sets the fee as a pending balance because that line is just super broken.

paninaro:
> These are great changes and well tested, but can we also remove the line from `Offer.get_pending_balances` that sets the fee as a pending balance because that line is just super broken.

Removed the fee calculation from `get_pending_balances`. That allowed me to remove another bit of code that was filtering out the incorrect XCH pending balance.

Quexington (approved):
lgtm

PR #10978, opened 2022-03-31
Ms.plot load perf2

mariano54:
Alternate PR to https://github.com/Chia-Network/chia-blockchain/pull/10847

PR #10847, opened 2022-03-22
2.7 seconds -> 0.45 seconds

mariano54:

PR #10986, opened 2022-04-01
Faster full node tests

mariano54:
22 -> 12 mins approximately

For optimal speed, we still need to merge:
https://github.com/Chia-Network/chia-blockchain/pull/10978
or 
https://github.com/Chia-Network/chia-blockchain/pull/10847

There is one real code change here, and that is for the full node to loop with fine grained sleep when someone is sending a transaction, instead of the slow 100ms sleep from before. This will also improve UX of sending TXs.

### DIFF
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -1264,8 +1264,9 @@ class FullNodeAPI:
         )
         # Waits for the transaction to go into the mempool, times out after 45 seconds.
         status, error = None, None
-        for i in range(450):
-            await asyncio.sleep(0.1)
+        sleep_time = 0.01
+        for i in range(int(45 / sleep_time)):
+            await asyncio.sleep(sleep_time)
             for potential_name, potential_status, potential_error in self.full_node.transaction_responses:
                 if spend_name == potential_name:
                     status = potential_status
--- a/tests/core/full_node/test_full_node.py
+++ b/tests/core/full_node/test_full_node.py
@@ -1226,9 +1226,10 @@ class TestFullNodeProtocol:
         block_2 = recursive_replace(block_2, "foliage.foliage_transaction_block_signature", new_fbh_sig)
         block_2 = recursive_replace(block_2, "transactions_generator", None)
 
-        await full_node_2.full_node.respond_block(fnp.RespondBlock(block_2), dummy_peer)
+        rb_task = asyncio.create_task(full_node_2.full_node.respond_block(fnp.RespondBlock(block_2), dummy_peer))
 
         await time_out_assert(10, time_out_messages(incoming_queue, "request_block", 1))
+        rb_task.cancel()
 
     @pytest.mark.asyncio
     async def test_request_unfinished_block(self, wallet_nodes, bt, self_hostname):
--- a/tests/core/full_node/test_full_node.py
+++ b/tests/core/full_node/test_full_node.py
@@ -97,7 +97,7 @@ async def get_block_path(full_node: FullNodeAPI):
 
 class TestFullNodeBlockCompression:
     @pytest.mark.asyncio
-    @pytest.mark.parametrize("tx_size", [10000, 3000000000000])
+    @pytest.mark.parametrize("tx_size", [3000000000000])
     async def test_block_compression(self, setup_two_nodes_and_wallet, empty_blockchain, tx_size, bt, self_hostname):
         nodes, wallets = setup_two_nodes_and_wallet
         server_1 = nodes[0].full_node.server
@@ -153,7 +153,6 @@ class TestFullNodeBlockCompression:
             return tx.confirmed
 
         await time_out_assert(30, check_transaction_confirmed, True, tr)
-        await asyncio.sleep(2)
 
         # Confirm generator is not compressed
         program: Optional[SerializedProgram] = (await full_node_1.get_all_full_blocks())[-1].transactions_generator
@@ -182,7 +181,6 @@ class TestFullNodeBlockCompression:
         await time_out_assert(30, wallet_is_synced, True, wallet_node_1, full_node_1)
 
         await time_out_assert(10, check_transaction_confirmed, True, tr)
-        await asyncio.sleep(2)
 
         # Confirm generator is compressed
         program: Optional[SerializedProgram] = (await full_node_1.get_all_full_blocks())[-1].transactions_generator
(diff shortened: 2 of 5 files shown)


chia/full_node/full_node_api.py
```diff
@@ -1264,8 +1264,9 @@ async def send_transaction(self, request: wallet_protocol.SendTransaction, *, te
         )
         # Waits for the transaction to go into the mempool, times out after 45 seconds.
         status, error = None, None
-        for i in range(450):
-            await asyncio.sleep(0.1)
+        sleep_time = 0.01
+        for i in range(int(45 / sleep_time)):
+            await asyncio.sleep(sleep_time)
```
arvidn, line 1269: oh. in an ideal world we wouldn't just blindly sleep here, but we would have a queue for responses that we could wait on.

tests/core/full_node/test_mempool.py
```diff
@@ -282,7 +253,7 @@ async def test_basic_mempool_manager(self, bt, two_nodes_mempool, wallet_a, self
             (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, 0, MempoolInclusionStatus.PENDING),
             (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, 1, MempoolInclusionStatus.PENDING),
             # the absolute height and seconds tests require fresh full nodes to
-            # run the test on. The fixture (two_nodes_mempool) creates 3 blocks,
+            # run the test on. The fixture (one_node_one_block) creates 3 blocks,
```
arvidn, line 256: the name of this fixture suggests that it only creates one block (compared to 3 block by `two_nodes_mempool`). If that's right, I would expect the heights below need updating as well.

tests/core/full_node/test_mempool.py
```diff
@@ -311,8 +282,10 @@ def test_fun(coin_1: Coin, coin_2: Coin) -> SpendBundle:
             bundle = SpendBundle.aggregate([tx1, tx2])
             return bundle
 
-        full_node_1, _, server_1, _ = two_nodes_mempool
-        blocks, bundle, status, err = await self.condition_tester2(bt, two_nodes_mempool, wallet_a, test_fun)
+        full_node_1, server_1 = one_node_one_block
+        _ = await next_block(full_node_1, wallet_a, bt)
+        _ = await next_block(full_node_1, wallet_a, bt)
```
arvidn, line 287: oh,. I see. you just make two extra blocks here to make the numbers the same as before. you would get less code and faster tests by just decrementing the expected block heights by 2 instead

tests/core/full_node/test_mempool.py
```diff
@@ -565,7 +538,12 @@ async def condition_tester(
             farmer_reward_puzzle_hash=reward_ph,
             pool_reward_puzzle_hash=reward_ph,
         )
-        peer = await connect_and_get_peer(server_1, server_2, bt.config["self_hostname"])
+        _, dummy_node_id = await add_dummy_connection(server_1, bt.config["self_hostname"], 100)
+        dummy_peer = None
+        for node_id, wsc in server_1.all_connections.items():
+            if node_id == dummy_node_id:
+                dummy_peer = wsc
+                break
```
arvidn, line 546: it looks like `connect_and_get_peer()` was useful. We do this at least twice. Why remove it?

tests/core/full_node/test_mempool.py
```diff
@@ -772,15 +758,18 @@ async def test_correct_my_id(self, bt, two_nodes_mempool, wallet_a):
...
 
     @pytest.mark.asyncio
-    async def test_my_id_garbage(self, bt, two_nodes_mempool, wallet_a):
+    async def test_my_id_garbage(self, bt, one_node_one_block, wallet_a):
 
-        full_node_1, full_node_2, server_1, server_2 = two_nodes_mempool
+        full_node_1, server_1 = one_node_one_block
+
+        _ = await next_block(full_node_1, wallet_a, bt)
+        _ = await next_block(full_node_1, wallet_a, bt)
```
arvidn, line 766: you add these extra two blocks in a bunch of places. I *think* we could just change what the test expects instead, but I might be forgetting something

tests/core/full_node/test_full_node.py
```diff
@@ -806,87 +795,90 @@ async def test_new_transaction_and_mempool(self, wallet_nodes, bt, self_hostname
...
-                receiver_puzzlehash = wallet_receiver.get_new_puzzlehash()
-                puzzle_hashes.append(receiver_puzzlehash)
-                output = ConditionWithArgs(ConditionOpcode.CREATE_COIN, [receiver_puzzlehash, int_to_bytes(100000000)])
-
-                conditions_dict[ConditionOpcode.CREATE_COIN].append(output)
-
-            spend_bundle = wallet_a.generate_signed_transaction(
-                100,
-                puzzle_hashes[0],
-                get_future_reward_coins(blocks[1 + i])[0],
```
arvidn, line 826: running this test in different blocks isn't relevant I take it?

tests/core/full_node/test_mempool.py
mariano54: It wasn't passing though

chia/full_node/full_node_api.py
mariano54: Yes true, but this is out of scope of this PR>=

tests/core/full_node/test_mempool.py
mariano54: It only creates one block, i'll update the comment

tests/core/full_node/test_mempool.py
mariano54: The problem is that there are many many tests, so the ones that require the 3 blocks are probably less than 15% here.

tests/core/full_node/test_mempool.py
mariano54: connect and get peer is for two nodes to connect to each other. But we don't actually need 2 nodes for this test. Creating unnecessary node processes slows things down due to the setup time. Here a dummy peer is sufficient.

tests/core/full_node/test_full_node.py
mariano54: Yeah that is not important for this test. The point of this section is to fill the mempool. I decreased the size of the mempool so now we don't need to many TXs.

tests/core/full_node/test_mempool.py
arvidn: I see. `connect_and_get_peer()` did more than this block

mariano54: yeah but the main problem is the fixture which used to get 2 nodes, now it only gets one.

arvidn (approved):

PR #10988, opened 2022-04-01
Check for vulnerable openssl

hoffmang9:
This checks for OpenSSL versions in the 1.1.1x class less than 1.1.1n and warns via the install script if the openssl version picked up by `import ssl` appears to be vulnerable to CVE-2022-0778


emlowe (approved):
looks ok to me

skweee:
Thanks for providing this so quickly!

PR #11000, opened 2022-04-02
Add __init__.pys

hoffmang9:
pre-commit seems pretty certain that we need these.

Open to other views.


hoffmang9:
Artifact of my clone. Closing.

altendky:
All these directories are gone.  Presumably there are files floating around in there causing git to not remove the directories.  The not-as-smart-as-we-would-like custom `__init__.py` adder script is falling for the bait and causing confusion.

PR #11001, opened 2022-04-02
Compile python 3.9.11 which is aware of the openssl issue

hoffmang9:
This moves the various yum installer based RH linux flavors to Python 3.9.11 - which is the first version of Python that is aware of CVE-2022-0778. That's not strictly required as the underlying OSes are what are relied upon for OpenSSL on Linux, but it seems like a wise idea to have a more modern Python built.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -96,11 +96,11 @@ install_python3_and_sqlite3_from_source_with_yum() {
   sudo make install | stdbuf -o0 cut -b1-"$(tput cols)" | sed -u 'i\\o033[2K' | stdbuf -o0 tr '\n' '\r'; echo
   # yum install python3 brings Python3.6 which is not supported by chia
   cd ..
-  echo "wget https://www.python.org/ftp/python/3.9.9/Python-3.9.9.tgz"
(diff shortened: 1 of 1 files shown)


ChiaMineJP:
I'm just reviewing/testing in my Linux virtual machines.

ChiaMineJP:
@hoffmang9 
I may misunderstand the interpretation of  the result below, but it seems even if upgrading to `3.9.11`, `import ssl` returns openssl installed on system.
(I used EC2 instance: Amazon Linux 2 AMI (HVM) - Kernel 4.14, SSD Volume Type)

image

image

ChiaMineJP (approved):
As I wrote in previous comment, we need to upgrade openssl installed on system, but that fact doesn't block upgrading python to `3.9.11`.

hoffmang9:
As I mentioned, openssl comes from the OS and not from Python on Linux. However, now that we're requiring at least Python 3.9.11 on Windows it seemed like a good idea to catch up the bug fix version of Python to the same version.

PR #11003, opened 2022-04-02
install.sh is not upgrading OpenSSL on MacOS

hoffmang9:
Unfortunately, git install users on MacOS are not getting the `brew install openssl` correctly.

Install.sh will now exit if it can not find `brew`

This may be worth back porting to latest.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -167,10 +167,12 @@ if [ "$(uname)" = "Linux" ]; then
       sudo yum install -y python39 openssl
     fi
   fi
-elif [ "$(uname)" = "Darwin" ] && ! type brew >/dev/null 2>&1; then
-  echo "Installation currently requires brew on MacOS - https://brew.sh/"
-  exit 1
 elif [ "$(uname)" = "Darwin" ]; then
+  echo "Installing on macOS."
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -169,6 +169,9 @@ if [ "$(uname)" = "Linux" ]; then
   fi
 elif [ "$(uname)" = "Darwin" ] && ! type brew >/dev/null 2>&1; then
   echo "Installation currently requires brew on MacOS - https://brew.sh/"
+  exit 1
+elif [ "$(uname)" = "Darwin" ]; then
+  echo "Installing OpenSSL"
   brew install openssl
```
altendky, line 175: Add the 'installing on' notification regardless, and group the macOS code all together in one `if`.  I don't know that I have this exactly right, coded off the cuff and untested.

```suggestion
elif [ "$(uname)" = "Darwin" ]; then
  echo "Installing on macOS."
  if ! type brew >/dev/null 2>&1; then
    echo "Installation currently requires brew on macOS - https://brew.sh/"
    exit 1
  fi
  echo "Installing OpenSSL"
  brew install openssl
```

PR #11010, opened 2022-04-03
Non Hobo patch the winstaller for CVE-2022-0778 (#10995)

hoffmang9:
libssl.dll calls libcrypto.dll where the vulnerability actually is so use a patched version of Python 3.9 instead of the quick and dirty .dll fix.


wjblanke (approved):
aok

hoffmang9:
Errored

PR #11011, opened 2022-04-03
Gh 1.3.3v2

hoffmang9:
Correctly patch the Windows Installer as libssl.dll depends on libcrypto.dll and the earlier hobo patch did not catch that correctly.

Back port an openssl edge cases on MacOS installed via git.


wjblanke (dismissed):
aok

wjblanke (approved):
aok

PR #11022, opened 2022-04-03
Update peers.dat filename to handle different nets

RightSexyOrc:
Propose including the net name, e.g. mainnet / testnet10 into the peers.dat filename, to peers_net.dat.

Lack of this has cost me hours of energy trying to switch from mainnet and sync up testnet10 and then go back.

Please test, as I haven't tested this.

Thank yee

### DIFF
--- a/chia/util/initial-config.yaml
+++ b/chia/util/initial-config.yaml
@@ -341,7 +341,7 @@ full_node:
   database_path: db/blockchain_v2_CHALLENGE.sqlite
   # peer_db_path is deprecated and has been replaced by peers_file_path
   peer_db_path: db/peer_table_node.sqlite
-  peers_file_path: db/peers.dat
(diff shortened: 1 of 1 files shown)


mariano54 (approved):

paninaro:
Thanks for the suggestion -- it's a good idea. We'll need to update `peer_store_resolver.py` and the associated tests to make this work.

paninaro (changes requested):
`peer_store_resolver.py` will need to be updated to handle this change.

paninaro:
I made some changes in a branch: [paninaro.wip_peer_store_resolver_with_challenge](https://github.com/Chia-Network/chia-blockchain/tree/paninaro.wip_peer_store_resolver_with_challenge)

Still needs a bit of testing to check upgrades, downgrades, and fresh install scenarios.

PR #10967, opened 2022-03-30
Add proper dust handling to new coinselection 

jack60612:
^

- [x] Add Tests
- [x] Set a limit on the number of spends 
- [x] Fix failing test

PR #11038, opened 2022-04-04
Fix the case of claiming a large number of coins

### LINKED ISSUE #10640: [Bug] Unable to claim self-pooling rewards in CLI or GUI after upgrade to 1.3
### What happened?

I am unable to claim self-pooling rewards after upgrading to 1.3 from 1.2.11.  I understand that there is a known issue with the CLI, but the GUI appears to fail as well.  Specifically I am take the following steps:

1. Go to Pool tab
2. Click "Claim Rewards"
3. In the dialog screen, the text shows "You will receive 1.75 XCH to" and does NOT show any receiving address.  There is also no way to enter an address into this area.
4. For the sake of trying to see if this would work anyways, I confirmed the transaction with 100k mojo to give it some priority.
5. Transaction with 100k mojo fee shows as pending in my wallet, but nothing happens.  I assume this is because it is an invalid transaction (due to missing address) however I don't know how to try to confirm this.

It is also worth noting that if I go to "Manage Farming Rewards" from the Farm tab, I receive the warning "No private keys for one or both addresses. Safe only if you are sending rewards to another wallet."  and both the Farmer Reward Address and the Pool Reward Address text boxes are blank.  However, the xch_target_address field for the farmer and the pool are filled in properly in the config.yaml file.

I have tried upgrading the DB, and I have also tried fully deleting the wallet db's.

I'm not sure where the root cause is on this one, but something definitely appears to have not made the transition from 1.2.11 to 1.3 properly
chia_claim_rewards_address_missing
.

### Version

1.3

Windows

GUI

### Relevant log output

_No response_

aqk:


chia/util/initial-config.yaml
```diff
@@ -214,6 +214,7 @@ pool:
   logging: *logging
   network_overrides: *network_overrides
   selected_network: *selected_network
+  default_max_spends_in_tx: 100
```
aqk, line 217: I put this here to ask if you think we should make this configurable, or leave it hard-coded to 100

mariano54:
How was 100 chosen? I would go a little higher since the max standard spends is about 2000 per block (soft limited at around 1000). 

Maybe 300?

chia/rpc/wallet_rpc_api.py
```diff
@@ -1397,13 +1397,14 @@ async def pw_absorb_rewards(self, request) -> Dict:
         if await self.service.wallet_state_manager.synced() is False:
             raise ValueError("Wallet needs to be fully synced before collecting rewards")
         fee = uint64(request.get("fee", 0))
+        max_spends_in_tx = request.get("max_spends_in_tx", None)
```
aqk, line 1400: I would like opinions on the name `max_spends_in_tx`.

Should we mention we are in a claim? It's obvious from the context of the API call name `pw_absorb_rewards`.
We may also want to use this parameter or something similar in other calls.

chia/pools/pool_wallet.py
```diff
@@ -807,13 +816,16 @@ async def claim_pool_rewards(self, fee: uint64) -> Tuple[TransactionRecord, Opti
...
 
-        current_coin_record = None
+        first_coin_record = None
         for coin_record in unspent_coin_records:
             if coin_record.coin not in coin_to_height_farmed:
                 continue
-            current_coin_record = coin_record
-            if len(all_spends) >= 100:
-                # Limit the total number of spends, so it fits into the block
+            if first_coin_record is None:
```
aqk, line 823: This is the bug - we were choosing the last coin we considered, but not *necessarily* the last included in the spend.

Fixed by always choosing the first.

aqk:
@mariano54 I think you put 100 in as a placeholder ;)

aqk:
Okay, 300 sounds good to me.

mariano54 (commented):
This PR: https://github.com/Chia-Network/chia-blockchain/pull/10967/files  should proabably use the constant as well. It's not a responsibility of 11038 to handle that part though

aqk:
@mariano54 What do you think - constant in the config file, or just in the source?
It will only affect very large farmers.

mariano54:
Either is acceptable. I would slightly prefer the config, since that allows users in extreme situations to be able to bypass the limit if it's necessary for them.

aqk:
Okay - I'll put that in another PR. I have another one I want to get to today,

aqk:
Rebased to main because of the benchmarks workflow

chia/util/initial-config.yaml
arvidn: it looks like this isn't used at all right now, and it's hard coded to 300.
I don't have a strong opinion, I can't judge if configuring the default is ever important (given you can specify an exact limit in the RPC).

tests/pools/test_pool_rpc.py
```diff
@@ -843,8 +934,6 @@ async def test_leave_pool(self, setup, trusted_and_fee, self_hostname):
             PeerInfo(self_hostname, uint16(full_node_api.full_node.server._port)), None
         )
 
-        WAIT_SECS = 200
```
arvidn, line 846: I see, we used a constant like this already. I think we should still leave it at 10 seconds though

tests/pools/test_pool_rpc.py
```diff
@@ -38,6 +39,7 @@
 # TODO: Compare deducted fees in all tests against reported total_fee
 log = logging.getLogger(__name__)
 FEE_AMOUNT = 2000000000000
+WAIT_SECS = 200  # A high value for WAIT_SECS is useful when paused in the debugger
```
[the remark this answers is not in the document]
aqk: I agree, of course. This is an alternative to having:
```
await time_out_assert(10, ...
await time_out_assert(5, ...
```

One thing to note: this is the maximum time it will wait. If the function returns true earlier, `time_out_assert` will return immediately.

I too look forward to when we don' have these anymore, but that's a ways off.
Also, much better than `sleep`

chia/util/initial-config.yaml
aqk: Thanks, will remove it from the config - that will go in in a separate PR if we do it.

chia/pools/pool_wallet.py
aqk: Thanks - I'll add a comment that explains this!

wallentx:
This fixes https://github.com/Chia-Network/chia-blockchain/issues/10640

arvidn:
> Either is acceptable. I would slightly prefer the config, since that allows users in extreme situations to be able to bypass the limit if it's necessary for them.

It's not a limit though, is it? It's just a default in case you don't specify it in the RPC

emlowe:
I've been doing some testing of this PR today. I had a test wallet with > 1000 txch in claimable pool rewards. I was testing with a fee of 0.001 txch

If I try to claim the max 300 that is currently in this PR, it would consistently not work. I would get `ALREADY_INCLUDING_TRANSACTION` errors, and the tx would just keep stuck for hours and never really finish.

Using the RPC I could adjust the max coin amount, some interesting items here from the testing - note in each case here, this was reproduced several times, these were not one-off cases, it was consistent:

20 - no errors , tx confirmed quickly
30 - get 1 (one) ALREADY_INCLUDING_TRANSACTION error - tx get confirmed anyway after a few mins
40 - get 2 (two) ALREADY_INCLUDING_TRANSACTION errors- tx gets confirmed anyway after a few mins
100 - get 2 (two) ALREADY_INCLUDING_TRANSACTION errors- tx gets confirmed anyway after a few mins
200 - get 2 (two) ALREADY_INCLUDING_TRANSACTION errors- tx gets confirmed anyway after a few mins
300 - get 2 (two) ALREADY_INCLUDING_TRANSACTION errors - tx appears stuck pending after several hours

With a "normal" amount of coins in the self claim pool (say 2-3 wins) - the CLI (`chia plotnft claim` and the GUI) work just fine with fee and without any errors.

This PR seems to resolve some issues, but some unusual edge cases clearly remain. In the interest of moving the current picture forward I suggest we merge this despite the current issues. This at least allows my test wallet to claim some things (using the RPC) and before it was totally unclaimable.

chia/pools/pool_wallet.py
aqk: @arvidn The reason the last coin considered might not be included is because of the line 
    `if len(all_spends) >= max_spends_in_tx`
That stops the current coin in the loop being included in `absorb_spend`, but the previous code was incorrectly using the last value of `current_coin` as if it were.

emlowe (approved):
Approved - still has some edge cases, but overall is an improvement

PR #11044, opened 2022-04-04
Move from multiprocessing to multithreading

xdustinface:
**Do not merge**: This is only meant for testing so far.

Swaps the most `ProcessPoolExecutors` to `ThreadPoolExecutors` and drops (de)serialization which was required for the multiprocessing. There are more ways to improve this and the related parts in timelord code are still TBD but its a start at least :)

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -643,7 +643,7 @@ class Blockchain(BlockchainInterface):
         task = asyncio.get_running_loop().run_in_executor(
             self.pool,
             _run_generator,
-            self.constants_json,
+            self.constants,
             unfinished_block,
             generator,
             height,
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -3,8 +3,7 @@ import dataclasses
 import logging
 import multiprocessing
 import traceback
-from concurrent.futures import Executor
-from concurrent.futures.process import ProcessPoolExecutor
+from concurrent.futures import Executor, ThreadPoolExecutor
 from enum import Enum
 from multiprocessing.context import BaseContext
 from pathlib import Path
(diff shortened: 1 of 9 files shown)


arvidn:
I think we'll have to do this, but in my (informal) testing just now, this brings the sync rate (with vdf validation disabled) from ~160 blocks/s to ~87 blocks/s. This is on my 8 core M1. It seems only a single core is pegged (~98% CPU)

arvidn:
in fact, I don't see any threads being spawned in `htop` (MacOS).

xdustinface:
> in fact, I don't see any threads being spawned in `htop` (MacOS).

Do you maybe have the `single_threaded` option set?

arvidn:
> > in fact, I don't see any threads being spawned in `htop` (MacOS).
> 
> Do you maybe have the `single_threaded` option set?

No, I don't. I was mistaken about not seeing the threads. I do see them spawned, they just don't seem to do any work.

But I'm also still testing with VDF validation disabled (to make it run faster)

xdustinface:
> > > in fact, I don't see any threads being spawned in `htop` (MacOS).
> > 
> > 
> > Do you maybe have the `single_threaded` option set?
> 
> I do see them spawned, they just don't seem to do any work.

Okay, yeah well i also didn't investigate much here but i assume its somehow related to now a lot pure python code which was running in multiple processes earlier is now single threaded due to python being python and for the most time the threads are waiting until some BLS/VDF operation comes up where it then will release the GIL and switch to the next one.

mariano54:
> > > > in fact, I don't see any threads being spawned in `htop` (MacOS).
> > > 
> > > 
> > > Do you maybe have the `single_threaded` option set?
> > 
> > 
> > I do see them spawned, they just don't seem to do any work.
> 
> Okay, yeah well i also didn't investigate much here but i assume its somehow related to now a lot pure python code which was running in multiple processes earlier is now single threaded due to python being python and for the most time the threads are waiting until some BLS/VDF operation comes up where it then will release the GIL and switch to the next one.

Yeah that makes sense.. It might be useful to profile this python code and see what are the slowest pieces of it. one thing that comes to mind is the MerkleSet operations which are all in python. The slowest part was the serialization which we now no longer use. I can't really think of any big operations that happen in python.

arvidn:
the merkle set is slow and shows up in profiles, but it's already happening in the main thread. I don't think it would contribute reduced parallelism. The issue with this patch is that python code that run in the worker threads is slow (and prevents parallelization)

PR #11047, opened 2022-04-04
Force apt to install the things we asked it to

aqk:


.github/workflows/benchmarks.yml
```diff
@@ -51,11 +51,13 @@ jobs:
           ${{ runner.os }}-pip-
 
     - name: Install ubuntu dependencies
+      env:
+        DEBIAN_FRONTEND=noninteractive
```
xdustinface, line 55:
```suggestion
        DEBIAN_FRONTEND: noninteractive
```

Not sure why this is needed? But i think this should be like ^

xdustinface: Oh, just saw the chat in IPS channel. Sounds like we could just drop it anyway?

aqk: Thanks - is this how we set env variables?

xdustinface: Yeah, the Benchmark workflow didn't even run before. Obviously because of wrong syntax.

xdustinface (approved):
Seems to work now and fixes the issue 👍

aqk:
Thanks!

wjblanke (approved):
aok

PR #11050, opened 2022-04-05
github: Drop unused `BUILD_VDF_CLIENT` variables

xdustinface:
From my understanding this is only used by `chiavdf` source builds which happen only if `install-timelord.sh` gets called but it doesn't in the addressed cases.


hoffmang9 (approved):
Cruft from when everything built timelord

PR #11051, opened 2022-04-05
Ms.fast test blockchain

mariano54:
Total time: 33m -> 13m. This includes installing and the whole CI process.

A few things contributed to the speed gains:
1. Changed test plots from k22 to k20. This saves time downloading the plots in many of the CI workflows. Also it saves time creating blocks for any tests that make blocks.
2. test_long_reorg now uses block cache and doesn't regenerate hundreds of blocks.
3. Test compact blockchain uses 2k blocks instead of 10k blocks. Testing with so many was unnecessary.   
4. A few other tests were optimized
5. Load fewer blocks if we can, for example don't load 10k blocks if you only need 1k, parsing them all is extremely slow.

Also I fixed a bug in the weight proof validation. An edge case was hit with this new list of blocks.

### DIFF
--- a/chia/full_node/weight_proof.py
+++ b/chia/full_node/weight_proof.py
@@ -1278,7 +1278,7 @@ def validate_recent_blocks(
         ses = False
         height = block.height
         for sub_slot in block.finished_sub_slots:
-            prev_challenge = challenge
+            prev_challenge = sub_slot.challenge_chain.challenge_chain_end_of_slot_vdf.challenge
             challenge = sub_slot.challenge_chain.get_hash()
             deficit = sub_slot.reward_chain.deficit
             if sub_slot.challenge_chain.subepoch_summary_hash is not None:
--- a/chia/full_node/weight_proof.py
+++ b/chia/full_node/weight_proof.py
@@ -1504,6 +1504,10 @@ def __get_rc_sub_slot(
 
     assert segment.rc_slot_end_info is not None
     if idx != 0:
+        # this is not the first slot, ses details should not be included
+        ses_hash = None
+        new_ssi = None
+        new_diff = None
         cc_vdf_info = VDFInfo(sub_slot.cc_slot_end_info.challenge, curr_ssi, sub_slot.cc_slot_end_info.output)
         if sub_slot.icc_slot_end_info is not None:
             icc_slot_end_info = VDFInfo(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -101,7 +101,7 @@ def softfork_height(request):
     return request.param
 
 
-block_format_version = "rc4"
+block_format_version = "rc5"
 
 
 @pytest.fixture(scope="session")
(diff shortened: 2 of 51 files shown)


tests/conftest.py
```diff
@@ -97,44 +97,85 @@ def softfork_height(request):
     return request.param
 
 
-block_format_version = "rc4"
+block_format_version = "rc5"
```
arvidn, line 100: this variable name is not intuitive to me. this sounds like a leftover from before the plot format was nailed down, a release candidate. Right now it just seems to be a version number in the filename.

to me it seems like this variable should be renamed to *just* `block_version` or maybe even `chain_version` and its value should be just `v1`, `v2` etc.

mariano54: yeah it is just a block version, I'll change that.

arvidn:
is this an expected flakiness? https://github.com/Chia-Network/chia-blockchain/runs/5876940304?check_suite_focus=true#step:9:83

```
_________________ TestPerformance.test_full_block_performance __________________
[77](https://github.com/Chia-Network/chia-blockchain/runs/5876940304?check_suite_focus=true#step:9:77)
tests/core/full_node/test_performance.py:174: in test_full_block_performance
[78](https://github.com/Chia-Network/chia-blockchain/runs/5876940304?check_suite_focus=true#step:9:78)
    res = await full_node_1.respond_unfinished_block(fnp.RespondUnfinishedBlock(unfinished), fake_peer)
[79](https://github.com/Chia-Network/chia-blockchain/runs/5876940304?check_suite_focus=true#step:9:79)
chia/full_node/full_node_api.py:466: in respond_unfinished_block
[80](https://github.com/Chia-Network/chia-blockchain/runs/5876940304?check_suite_focus=true#step:9:80)
    await self.full_node.respond_unfinished_block(
[81](https://github.com/Chia-Network/chia-blockchain/runs/5876940304?check_suite_focus=true#step:9:81)
chia/full_node/full_node.py:1757: in respond_unfinished_block
[82](https://github.com/Chia-Network/chia-blockchain/runs/5876940304?check_suite_focus=true#step:9:82)
    raise ConsensusError(Err(validate_result.error))
[83](https://github.com/Chia-Network/chia-blockchain/runs/5876940304?check_suite_focus=true#step:9:83)
E   chia.util.errors.ConsensusError: Error code: INVALID_CC_CHALLENGE
```

tests/core/full_node/test_full_node.py
```diff
@@ -1541,7 +1541,7 @@ async def test_compact_protocol(self, setup_two_nodes_fixture, bt):
         )
 
         # Note: the below numbers depend on the block cache, so might need to be updated
-        assert cc_eos_count == 4 and icc_eos_count == 3
```
arvidn, line 1544: were these `4` -> `3` a consequence of changing k22 -> k20?

chia/full_node/weight_proof.py
```diff
@@ -1504,6 +1504,10 @@ def __get_rc_sub_slot(
 
     assert segment.rc_slot_end_info is not None
     if idx != 0:
+        # this is not the first slot, ses details should not be included
+        ses_hash = None
+        new_ssi = None
+        new_diff = None
```
arvidn, line 1510: My understanding is that this was a (somewhat) unrelated bug fix, which I really think we ought to have a test covering as well. Do we?
I don't think a test need to hold up this PR though

arvidn (changes requested):
I think this looks good. I'm pretty sure you did not mean to update the `mozilla-ca` submodule though. That should be restored before we land this.

chia/full_node/weight_proof.py
mariano54: Yeah it's a bug fix. @almogdepaz knows more about it, but it happens when there is an empty slot at the start of an epoch

arvidn (approved):

PR #11053, opened 2022-04-05
Remove dead snakes usage from benchmark tests

altendky:


hoffmang9 (dismissed):
lgtm

altendky:
This keeps runningthe benchmarks on `beast` which has consistently timed out for recent historical runs triggered by others, while `sumo` runs fine.  There's no rush on this so let's do it right and wait for really successful CI.

hoffmang9 (approved):
lgtm

PR #11057, opened 2022-04-05
CHIA-597: Use `poetry` for package management

altendky:
Devs should continue to use the `install.sh` / `Install.ps1` scripts for convenience. These scripts do the following:

Download and install `poetry` into `.penv` 
Setup a chia-blockchain specific `.venv`
Create a symlink between `venv` and `.venv`

You can then use activate in the same way as before

This approach keeps the poetry deployment in `.penv` separate from the chia-blockchain install in `.venv`. If you want to run the `poetry` command you will need to run it from the `.penv` directory.

You can install `poetry` globally if you want, but then you may need to pay attention to the versioning in poetry specifically (running a different version from the one installed in `install.sh` may have unexpected results). We are still exploring options on better ways to handle the `poetry` install.

Draft for:
- [x] Some `TODO`s
- [x] Lots of consideration
- [x] Not being a preliminary exploration
- [x] Switch the artifact check to use poetry
- [x] miniupnpc extra
- [x] ipython
- [x] making lgtm ignore .venv
- [x] figure out m1 trying to build `chiavdf` from source https://github.com/Chia-Network/chia-blockchain/runs/8059378328?check_suite_focus=true
- [x] handle automatic versioning
- [x] catch up pre-version checking
- [x] review details of automatic versioning


emlowe:
close and reopen

emlowe:
@SocketSecurity ignore pypi/pygments@2.18.0

emlowe:
close and reopn for macos installer CI

.github/workflows/upload-pypi-source.yml
```diff
@@ -194,6 +199,7 @@ jobs:
 
       - name: Build source distribution
         run: |
+          # TODO: let poetry do this?
```
altendky, line 202: Using `-m build` is correct, though we should not pass `--sdist` in general and let it build both a wheel and sdist and publish both.  That's for another time though.

```suggestion
```

emlowe:
Manual testing of using `-i` to the install script to make a non-editable install was complete on Windows and MacOS - option is working as expected - using `-i` makes a non-editable install in that changes made to source files do not change the runtime behaviour

emlowe:
@SocketSecurity ignore pypi/dill@0.3.8

.github/workflows/reflow-version.yml
```diff
@@ -40,9 +40,12 @@ jobs:
...
         id: version-number
+        # TODO figure out better way to handle versioning
         run: |
           python3 -m venv ../venv
           . ../venv/bin/activate
-          pip install setuptools_scm
-          echo "chia-installer-version=$(python3 ./build_scripts/installer-version.py)" >> "$GITHUB_OUTPUT"
+          python -m pip install --upgrade pip
+          pip install poetry
+          pip install "poetry-dynamic-versioning[plugin]"
```
altendky, line 49: One pip call when possible.  Well, except for the pip upgrade call before...  so...  two.

```suggestion
          pip install poetry "poetry-dynamic-versioning[plugin]"
```

.github/workflows/test-single.yml
```diff
@@ -228,7 +228,7 @@ jobs:
           echo '==== collecting installed tests ===='
           if pytest --collect-only -qq --import-mode=append --pyargs chia._tests > installed_tests_raw
           then
-            cat installed_tests_raw | sed -n 's;^venv/.*/\(chia/_tests/.*\);\1;p' | sort > installed_tests
+            cat installed_tests_raw | sed -n 's;^.venv/.*/\(chia/_tests/.*\);\1;p' | sort > installed_tests
```
altendky, line 231: I think?

```suggestion
            cat installed_tests_raw | sed -n 's;^\.venv/.*/\(chia/_tests/.*\);\1;p' | sort > installed_tests
```

emlowe:
@SocketSecurity ignore pypi/dill@0.3.7
@SocketSecurity ignore pypi/clvm-rs@0.2.5

arvidn:
when testing this on Mac (M1) I get this warning:
```
Warning: The file chosen for install of twine 5.1.0 (twine-5.1.0-py3-none-any.whl) is yanked. Reason for being yanked: https://github.com/pypa/twine/issues/1125
```

I suppose this is related to https://github.com/Chia-Network/chia-blockchain/pull/18253

arvidn (commented):
looks good. It would be nice to have some documented guidance on how treat the lockfile, and PRs making changes to it. I assume it would be accepted with caution, as we want to lock down our supply-chain risk.

arvidn (dismissed):

Starttoaster:
Waiting on coverage-diff override

emlowe:
> ul

emlowe:
Sorry Clicked the wrong thing and ended up closing and reopening the PR

emlowe:
> looks good. It would be nice to have some documented guidance on how treat the lockfile, and PRs making changes to it. I assume it would be accepted with caution, as we want to lock down our supply-chain risk.

Yes, generally you will want to use poetry to make any changes, eg:

`.penv/bin/poetry update urllib3`

However, I'm also hopeful that now all the dependencies are listed in the lock file, we can relay on `dependabot` creating PRs to update anything

emlowe:
Well, dependabot seems to have a flow in that it updates the `lock` files but not the `toml` (https://github.com/dependabot/dependabot-core/issues/8603) 

So that might take some manual effort per the reporter until that gets resolved....

arvidn (approved):

PR #11058, opened 2022-04-05
Use Install.ps1 in build_windows.ps1

altendky:
Draft for:
- [x] Will need to add `-d` support to `Install.ps1` https://github.com/Chia-Network/chia-blockchain/pull/11062
- [x] Move misc dependencies to the dev extra rather than mixed in with various scripts


build_scripts/build_windows.ps1
```diff
@@ -26,12 +26,9 @@ else
 Write-Output "   ---"
 Write-Output "Create venv - python3.9 is required in PATH"
 Write-Output "   ---"
-python -m venv venv
+.\Install.ps1 -d
```
cmmarslender, line 29: In the interest in keeping this more similar to the build scripts for the other platforms, should we move this to the workflow file? (I think that is where install.sh is called for the other platforms)

altendky: I don't super care where that happens.  It is already present over in https://github.com/Chia-Network/chia-blockchain/pull/11077 so it will happen eventually.

cmmarslender: Can wait for the other changes then, this one looks good for now

cmmarslender (approved):

PR #11060, opened 2022-04-05
Run benchmarks workflow in a container

altendky:


cmmarslender (changes requested):

.github/workflows/benchmarks.yml
```diff
@@ -20,45 +20,21 @@ jobs:
...
+      env:
+        DEBIAN_FRONTEND: noninteractive
+      run: |
+        apt-get --yes update
+        apt-get install --yes git lsb-release sudo
+
     - name: Checkout Code
       uses: actions/checkout@v3
       with:
         fetch-depth: 0
```
wallentx, line 37: - name: Checkout test blocks and plots
      uses: actions/checkout@v3
      with:
        repository: 'Chia-Network/test-cache'
        path: '.chia'
        ref: '0.28.0'
        fetch-depth: 1

PR #11062, opened 2022-04-05
add -d for Install.ps1

altendky:


Quexington (approved):
I haven't tested but this makes sense to me.

PR #11065, opened 2022-04-05
Speedup preparing test blocks and plots

AmineKhaldi:
Achieved by opting for a release download instead of a shallow git clone, and also by putting a caching layer on top of that.


mariano54 (approved):
Great. Combined with the other PR it will be even faster

PR #11069, opened 2022-04-05
Improve detection of disconnected websocket between services

aqk:
When running `chia wallet show`, I got the error
`No keys loaded. Run 'chia keys generate' or import a key`

The problem is that the keychain_proxy is not able to  get the RPC call `get_key_for_fingerprint` because the websocket is not connected.

### DIFF
--- a/chia/daemon/client.py
+++ b/chia/daemon/client.py
@@ -67,7 +67,7 @@ class DaemonProxy:
         request_id = request["request_id"]
         self._request_dict[request_id] = asyncio.Event()
         string = dict_to_json_str(request)
-        if self.websocket is None:
+        if self.websocket is None or self.websocket.closed:
             raise Exception("Websocket is not connected")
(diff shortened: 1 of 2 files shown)


chia/daemon/client.py
```diff
@@ -67,7 +67,7 @@ async def _get(self, request: WsRpcMessage) -> WsRpcMessage:
         request_id = request["request_id"]
         self._request_dict[request_id] = asyncio.Event()
         string = dict_to_json_str(request)
-        if self.websocket is None:
+        if self.websocket is None or self.websocket.closed:
```
aqk, line 70: Note that I've only tested this path, in `client.py`

jack60612 (approved):
this makes sense to me

pnoch:
Very nice, this also fixes the chia-dashboard-satellite issue.

wjblanke (approved):
aok

PR #11070, opened 2022-04-06
can we get by without dead snakes?

altendky:


hoffmang9:
I wonder if we could get by with libpython3.8-dev instead? https://packages.ubuntu.com/bionic-updates/libpython3.8-dev

altendky:
Hmm, so it's actually failing right now only on 3.7 and actually because the timelord script is trying to install the dev apt package...  which we may just not need.

I did find a stale complaint about `Python.h` being missing, but let's see what would happen if the timelord script just pretended everything was ok for the moment.  https://github.com/actions/setup-python/issues/221

altendky:
And if you are suggesting using 3.8 dev for 3.7, I think I'll be looking for other options first.  `:]`

altendky:
https://github.com/Chia-Network/chia-blockchain/runs/5843561053?check_suite_focus=true

Ok, so either we pre-detect if we already have Python dev stuff or we make an explicit CLI parameter or environment variable.  I'm thinking CLI parameter.

AmineKhaldi (approved):

PR #11075, opened 2022-04-06
Set keychain_proxy to None in await_closed() to support reinitialization

paninaro:
When `_await_closed()` is called in the wallet (or farmer), `self.keychain_proxy` was being closed, but not set to `None`. In the wallet case, using `chia wallet show` to select a key and "log in" has the effect of closing the current wallet node (destroying the keychain proxy) and then creating a new wallet node. Because the old keychain_proxy instance was not reset to None, the new wallet node wasn't able to communicate with the keychain.

`_await_closed()` now takes an optional `shutting_down` param to control whether the keychain proxy is closed. When the process is terminating, `shutting_down` will be set to `True`, but when `wallet_rpc_api` makes calls to `_await_closed` it will pass `shutting_down=False` to keep the keychain proxy alive.

### DIFF
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -140,7 +140,7 @@ class Farmer:
         self.harvester_cache: Dict[str, Dict[str, HarvesterCacheEntry]] = {}
 
     async def ensure_keychain_proxy(self) -> KeychainProxy:
-        if not self.keychain_proxy:
+        if self.keychain_proxy is None:
             if self.local_keychain:
                 self.keychain_proxy = wrap_local_keychain(self.local_keychain, log=self.log)
             else:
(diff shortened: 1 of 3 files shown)


chia/farmer/farmer.py
```diff
@@ -212,13 +212,15 @@ async def start_task():
     def _close(self):
         self._shut_down = True
 
-    async def _await_closed(self):
+    async def _await_closed(self, shutting_down: bool = True):
```
paninaro, line 215: This isn't strictly necessary, as the farmer doesn't seem to have the same stop/restart pattern that's employed by the wallet. Still, I made the same `shutting_down` change here to keep the code in sync (though it would be nice to have a non-duplicated code solution...)

jack60612: maybe at some point we will make some base class, but i think we can add that to the list of growing refactors.

jack60612 (approved):
lgtm if tests pass.

PR #11090, opened 2022-04-07
Expose transaction sorting options to CLI

emlowe:
In 1.3 the GUI and back-end were updated with some new sorting options for transactions. Specifically one can sort by height (`order by confirmed_at_height {ASC}`) or "relevance" (`order by confirmed {ASC}, confirmed_at_height {DESC}, created_at_time {DESC}`)

The GUI uses relevance by default.

This PR exposes these options to the CLI for `get_transactions` with `--sort-by-height` and `--sort-by-relevance` with relevance as the default. This should make get_transactions return the same list in the same order as the GUI displays them

For completeness, also expose the `--reverse` option which reverses the order of the sort

### DIFF
--- a/chia/cmds/wallet.py
+++ b/chia/cmds/wallet.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, Optional, Tuple
 import click
 
 from chia.wallet.util.wallet_types import WalletType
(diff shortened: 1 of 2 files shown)


Quexington (approved):
:+1:

PR #11094, opened 2022-04-08
transition to using chia_rs module

arvidn:
we have separated core CLVM functionality (executing clvm programs) from the higher level Chia functions (parsing and partially validate conditions, with more to come) into two separate projects. We now have `clvm_rs` and `chia_rs`. The latter exports all functions from both libraries.

This patch changes our dependency from `clvm_rs` to `chia_rs`. One note worthy change is that the `MEMPOOL_MODE` in the new version is stricter, where it requires the number of arguments to conditions to match what we use. This rule is being disabled by this patch in order to preserve the current behavior. In order to enable it, we'll also need to update tests, so I'm putting that in a separate PR.

### DIFF
--- a/tests/wallet/test_singleton.py
+++ b/tests/wallet/test_singleton.py
@@ -51,7 +51,7 @@ def test_only_odd_coins():
     try:
         cost, result = SINGLETON_MOD.run_with_cost(INFINITE_COST, solution)
     except Exception as e:
-        assert e.args == ("clvm raise",)
+        assert e.args == ("clvm raise", "80")
     else:
         assert False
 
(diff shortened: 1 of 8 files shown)


tests/wallet/test_singleton.py
```diff
@@ -84,7 +84,7 @@ def test_only_one_odd_coin_created():
     try:
         cost, result = SINGLETON_MOD.run_with_cost(INFINITE_COST, solution)
     except Exception as e:
-        assert e.args == ("clvm raise",)
+        assert e.args == ("clvm raise", "80")
```
arvidn, line 87: this changed because `chia_rs` throws `ValueError` whereas `clvm_rs` throws `EvalError` (imported from the `clvm` python package). Richard made this change in order to no longer depend on the `clvm` python package (I believe)

trepca: what's behind `80`? is that the value like `(x "80")`?

prozacchiwawa (approved):
pulled this branch, ran tests.
overall changes look good, peeked at chia_rs, which is nice.

tests/wallet/test_singleton.py
arvidn: `80` is the serialized form of an empty list or NIL, in clvm

PR #11096, opened 2022-04-08
consistently name installer github actions artifact zips

altendky:
Broken out from https://github.com/Chia-Network/chia-blockchain/pull/11077.


cmmarslender (approved):
Looks good to me. These aren't used downstream, so consistency makes sense and should have little impact on release automation.

PR #11103, opened 2022-04-08
Optimize test_wallet_sync

mariano54:
Running every test twice with two modes is wasteful, we can share a lot of the slow stuff

### DIFF
--- a/tests/wallet/sync/test_wallet_sync.py
+++ b/tests/wallet/sync/test_wallet_sync.py
@@ -198,8 +198,7 @@ class TestWalletSync:
         wallets[1][0].config["trusted_peers"] = {}
 
         phs = []
-        for wallet_pair in wallets:
-            wallet_node, wallet_server = wallet_pair
+        for wallet_node, wallet_server in wallets:
             wallet = wallet_node.wallet_state_manager.main_wallet
             phs.append(await wallet.get_new_puzzlehash())
             await wallet_server.start_client(PeerInfo(self_hostname, uint16(full_node_server._port)), None)
--- a/tests/wallet/sync/test_wallet_sync.py
+++ b/tests/wallet/sync/test_wallet_sync.py
@@ -27,27 +27,20 @@ log = getLogger(__name__)
 
 
 class TestWalletSync:
-    @pytest.mark.parametrize(
-        "trusted",
-        [True, False],
-    )
     @pytest.mark.asyncio
-    async def test_basic_sync_wallet(self, bt, wallet_node, default_400_blocks, trusted, self_hostname):
-
-        full_node_api, wallet_node, full_node_server, wallet_server = wallet_node
+    async def test_basic_sync_wallet(self, bt, two_wallet_nodes, default_400_blocks, self_hostname):
+        full_nodes, wallets = two_wallet_nodes
+        full_node_api = full_nodes[0]
+        full_node_server = full_node_api.full_node.server
 
         for block in default_400_blocks:
             await full_node_api.full_node.respond_block(full_node_protocol.RespondBlock(block))
 
-        if trusted:
-            wallet_node.config["trusted_peers"] = {full_node_server.node_id.hex(): full_node_server.node_id.hex()}
-        else:
-            wallet_node.config["trusted_peers"] = {}
-        await wallet_server.start_client(PeerInfo(self_hostname, uint16(full_node_server._port)), None)
+        for wallet_node, wallet_server in wallets:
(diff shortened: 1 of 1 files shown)


tests/wallet/sync/test_wallet_sync.py
```diff
@@ -27,209 +27,207 @@ def wallet_height_at_least(wallet_node, h):
...
-            wallet_node.config["trusted_peers"] = {fn_server.node_id.hex(): fn_server.node_id.hex()}
-        else:
-            wallet_node.config["trusted_peers"] = {}
+        # Untrusted node sync
+        wallets[1][0].config["trusted_peers"] = {}
 
-        await server_2.start_client(PeerInfo(self_hostname, uint16(fn_server._port)), None)
+        phs = []
+        for wallet_pair in wallets:
+            wallet_node, wallet_server = wallet_pair
```
altendky, line 202:
```suggestion
        for wallet_node, wallet_server in wallets:
```

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -27,209 +27,207 @@ def wallet_height_at_least(wallet_node, h):
...
+            wallet = wallet_node.wallet_state_manager.main_wallet
+            phs.append(await wallet.get_new_puzzlehash())
+            await wallet_server.start_client(PeerInfo(self_hostname, uint16(full_node_server._port)), None)
 
         # Insert 400 blocks
         for block in default_400_blocks:
             await full_node_api.full_node.respond_block(full_node_protocol.RespondBlock(block))
 
         # Farm few more with reward
+        for i in range(0, num_blocks - 1):
```
altendky, line 212: Why `- 1`?

altendky (commented):
Thanks for the extra work to retain coverage of the two sync modes while reducing runtime.

tests/wallet/sync/test_wallet_sync.py
mariano54: The test was testing only one wallet, but farming rewards take 1 block to get confirmed. So here the first wallet needs less rewards because all of them will get confirmed, in the second one, the last one will not get confirmed

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -27,209 +27,207 @@ def wallet_height_at_least(wallet_node, h):
...
+            wallet = wallet_node.wallet_state_manager.main_wallet
+            both_phs.append(await wallet.get_new_puzzlehash())
 
-        if trusted:
-            wallet_node.config["trusted_peers"] = {full_node_server.node_id.hex(): full_node_server.node_id.hex()}
-        else:
-            wallet_node.config["trusted_peers"] = {}
+        for i in range(20):
+            # Tests a reorg with the wallet
+            ph = both_phs[i % 2]
```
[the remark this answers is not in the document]
mariano54: That you, but that is much more complicated imo

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -27,209 +27,207 @@ def wallet_height_at_least(wallet_node, h):
...
+        full_node_api = full_nodes[0]
+        full_node_server = full_node_api.full_node.server
 
         for block in default_400_blocks:
             await full_node_api.full_node.respond_block(full_node_protocol.RespondBlock(block))
 
-        if trusted:
-            wallet_node.config["trusted_peers"] = {full_node_server.node_id.hex(): full_node_server.node_id.hex()}
-        else:
-            wallet_node.config["trusted_peers"] = {}
```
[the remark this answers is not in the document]
mariano54: thank you! that's a good catch

tests/wallet/sync/test_wallet_sync.py
altendky: Thanks for the explanation.  I think I might generally prefer to make sure all the rewards get processed rather than finishing the test with unconfirmed transactions and off-by-one compensations.

But, you can go ahead and resolve this conversation once you've read my comment.  No action requested.

tests/wallet/sync/test_wallet_sync.py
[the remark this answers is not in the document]
altendky: It is true that modulus is a common programming idiom.  How about this?  We really don't care about the range nor the modulus of the iteration.  

```suggestion
        for ph in both_phs * 10:
            # Tests a reorg with the wallet
```

I don't mean to badger about this but I realized this much simpler form than my last suggestion.  Feel free to mark this resolved.

altendky (approved):

PR #11119, opened 2022-04-10
remove setting of SCM_VERSION in install-version.py

altendky:
This seems unused internally and I am not aware of any scenario where it would get inherited back out to other processes.

Extracted from https://github.com/Chia-Network/chia-blockchain/pull/11077.

### DIFF
--- a/build_scripts/installer-version.py
+++ b/build_scripts/installer-version.py
@@ -1,4 +1,3 @@
-import os
 import sys
 
 from setuptools_scm import get_version
@@ -10,7 +9,6 @@ def main():
 
     scm_full_version = get_version(root="..", relative_to=__file__)
     # scm_full_version = "1.0.5.dev22"
-    os.environ["SCM_VERSION"] = scm_full_version
 
     left_full_version = scm_full_version.split("+")
     version = left_full_version[0].split(".")


hoffmang9:
This is there to override a version in a CI failure by simply adding it to a workflow to override. Fine to remove it but important to record what it was doing.

altendky:
I don't quite follow.  How would this environment variable have been exposed to anything else?  Did `installer-version.py` used to subprocess to other stuff so that they would inherit this variable?

hoffmang9:
One could modify eg build-MacOS-installer.yml to add an env variable to the build step that would obtain installers with version 1.3.4 and not 1.3.5.dev01 in an instance where there was an installer build failure

altendky:
Sure, but this is setting an env var, not reading it, with no apparent path out of this process which doesn't use it from what I can tell.  It does not subprocess nor would the variable be inherited back up to the parent process since that isn't how environment variables work.  This script is also mostly 'not importable' since it has a `-` in the name (you can, but only via contortions).  I don't see a practical way for this to be leveraged as a way to get access to the version.  Am I missing a path for the data to get used?

This is entirely a side note and not needed for anything else I'm working on so it isn't important for it to change.  I'm just trying to make sure I understand how leaving the code in place could be leveraged by anything.

hoffmang9:
Yep, looking again it looks like it's some sort of now lost vestigial thing.

xdustinface (approved):

wjblanke:
i think this was the byproduct of debugging SCM long long ago

PR #11120, opened 2022-04-10
have pyinstaller check platlib for dll's, not ROOT

altendky:
I believe we mean to find the `site-packages/*.dll` files not those a couple directories up from the `chia` package.  This differs, for example, when the `chia` source is found such as when using an `--editable` install.

Extracted from https://github.com/Chia-Network/chia-blockchain/pull/11077.

### DIFF
--- a/chia/pyinstaller.spec
+++ b/chia/pyinstaller.spec
@@ -2,6 +2,7 @@
 import importlib
 import pathlib
 import platform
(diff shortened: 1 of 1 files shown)


cmmarslender (approved):

wjblanke (approved):
aok

PR #11131, opened 2022-04-11
replace some asyncio.get_event_loop().run_until_complete() with asyncio.run()

altendky:
Extracted from https://github.com/Chia-Network/chia-blockchain/pull/9930.

### DIFF
--- a/tests/core/util/test_config.py
+++ b/tests/core/util/test_config.py
@@ -138,7 +138,7 @@ def run_reader_and_writer_tasks(root_path: Path, default_config: Dict):
     Subprocess entry point. This function spins-off threads to perform read/write tasks
     concurrently, possibly leading to synchronization issues accessing config data.
     """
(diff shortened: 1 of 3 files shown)


xdustinface (changes requested):
Any reason to not adjust the ones in `block_tools.py`?

altendky:
There was some funny business there where the fact we were doing silly stuff actually caused problems when doing it a bit less silly.  I'm punting on that for the sake of getting 3.10 done.  There's a warning ignore added related to those so they are 'on a list' as needing to be addressed.   Sort of.

xdustinface (approved):
> There was some funny business there where the fact we were doing silly stuff actually caused problems when doing it a bit less silly.

Hm, ok but i don't really understand whats different in there and what kind of "funny business" would lead to an issue since it works just fine (at least locally).

wjblanke (approved):
aok

PR #11132, opened 2022-04-11
clean up some extra type ignores in the daemon 

jack60612:


jack60612:
seemed to work locally, i guess it does not here??

PR #11133, opened 2022-04-11
Print MempoolInclusionStatus as string

aqk:


chia/types/mempool_submission_status.py
```diff
@@ -0,0 +1,28 @@
...
+from typing import Dict, Optional, Union
+
+from chia.types.mempool_inclusion_status import MempoolInclusionStatus
+from chia.util.ints import uint8
+from chia.util.streamable import Streamable, streamable
+
+
+@streamable
+@dataclass(frozen=True)
+class MempoolSubmissionStatus(Streamable):
```
aqk, line 11: Note that I'm not converting the `sent_to` members in `TradeRecord` and `TransactionRecord` to use this struct, because that would change the DB, the Trade protocol, and the blockchain protocol.

chia/cmds/cmds_util.py
```diff
@@ -0,0 +1,12 @@
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.types.mempool_submission_status import MempoolSubmissionStatus
+from chia.wallet.transaction_record import TransactionRecord
+
+
+def transaction_submitted_msg(tx: TransactionRecord) -> str:
```
aqk, line 6: I prefer this type of function to return string, rather than doing the printing itself.

aqk:
Before:
```
Transaction submitted to nodes: [('8af8436ad87...', 1, None)]
Do chia wallet get_transaction -f 2536661959 -tx 0x785d2ced78dc... to get status
```

After:
```
Transaction submitted to nodes: [{'peer_id': '8af8436ad87...', 'inclusion_status': 'SUCCESS', 'error_msg': None}]
Run 'chia wallet get_transaction -f 2536661959 -tx 0x785d2ced78dc...' to get status
```

aqk:
This patch changes the output of the command line only. It does not change the types of the values returned in the `get_transaction` RPC.

emlowe (approved):
looks good

PR #11136, opened 2022-04-11
Add get_unconfirmed_transactions wallet RPC

aqk:
The immediate use of this RPC is to help end users understand the current state of their "stuck" transactions. For example, is it still being considered by the wallet for re-submission to other nodes, or has it been dropped?

There are so many ways to implement this - I started out making this a filter on the `get_transactions_between` call to the DB. But looking a little closer, we use a cache above the DB, and `get_unconfirmed_transactions` is what wallets actually call to determine if they have pending transactions, so I thought it would be better to wire this command to that method.


emlowe (dismissed):
Looks fine to me

xdustinface (changes requested):
I agree having a separate endpoint for it makes sense. See the suggestions and the following two points:

- `get_wallet_info` can also be applied in `get_transaction`
- There should also be tests for this new RPC endpoint

PR #11137, opened 2022-04-12
Merge commit with main (9ff3fc993f69b5c59ecf14470284a5b9e799e6cf)

paninaro:


ytx1991:
Are you going to fix the rest failing tests? I'm ok to merge with the failed DID tests since they will be modify anyway.

paninaro:
> 



> Are you going to fix the rest failing tests? I'm ok to merge with the failed DID tests since they will be modified anyway.

Possibly... I was able to fix some of the test failures, but the issue I'm looking at now may be the result of some wallet backend code that either didn't merge properly, or assumptions have changed regarding how the wallet should work.

But I agree that depending on what the remaining failures are, we might be able to merge without fixing every test.

ytx1991 (approved):
Sync off line

PR #11139, opened 2022-04-12
make NFTWallet a dataclass

altendky:


paninaro (approved):
The type checking changes are not easily grokked, but I understand that it improves the ability to annotate types in certain scenarios.

PR #11140, opened 2022-04-12
Derivation from just a master public key

Yostra:


mariano54:
What's the purpose of this change?

wjblanke (approved):
be sure to try show all keys

xdustinface:
> To be able to derive wallet public keys from just root public key (no sk).

@Yostra But whats the point about adding this in `cmds/key_funcs.py`, where it now calculated a redundant address just to assert its the same and then not even use it? This should rather be properly tested in some test.

mariano54:
This will fail for non observer derivation
This change should be reverted unless I'm missing something

mariano54:
@wjblanke

PR #11146, opened 2022-04-12
add tool to generate a blockchain with full blocks, as a benchmark

arvidn:
The bulk of this patch is a tool that can generate a block chain of full blocks. Such chain can be used as a benchmark and performance test to ensure we can handle a chain consisting of only blocks with the maximum number of transactions. i.e. the highest load on nodes.

Generating this chain is currently quite slow, partly because we haven't optimized some of the operations that are very heavily used in it. In order to make it possible to use the test k20 plots and a smaller initial difficulty, this chain does not use the mainnet constants, but uses its own test constants (in `tools/test_constants.py`).

The second, smaller, part of this patch is to make it possible to use this chain in the `test_full_sync.py` test program. This is where the full node can be profiled and benchmarked. There's a new option to launch the test full node with the appropriate test constants for this purpose.

### DIFF
--- a/tools/generate_chain.py
+++ b/tools/generate_chain.py
@@ -1,0 +1,142 @@
+import cProfile
+import random
+import sqlite3
+import time
+from contextlib import closing, contextmanager
+from pathlib import Path
+from typing import Iterator, List
+
+import zstd
+
+from chia.types.blockchain_format.coin import Coin
+from chia.types.spend_bundle import SpendBundle
+from chia.util.chia_logging import initialize_logging
+from chia.util.ints import uint64
+from chia.util.path import mkdir
+from tests.block_tools import create_block_tools
+from tests.util.keyring import TempKeyring
+from tools.test_constants import test_constants
+
+
+@contextmanager
+def enable_profiler(profile: bool) -> Iterator[None]:
+    if not profile:
+        yield
+        return
+
+    with cProfile.Profile() as pr:
+        yield
+
+    pr.create_stats()
+    pr.dump_stats("generate-chain.profile")
+
+
+root_path = Path("./test-chain").resolve()
+mkdir(root_path)
+with TempKeyring() as keychain:
+
+    bt = create_block_tools(constants=test_constants, root_path=root_path, keychain=keychain)
+    initialize_logging(
+        "generate_chain", {"log_level": "DEBUG", "log_stdout": False, "log_syslog": False}, root_path=root_path
+    )
+
+    with closing(sqlite3.connect("stress-test-blockchain.sqlite")) as db:
+
+        print("initializing v2 block store")
+        db.execute(
+            "CREATE TABLE full_blocks("
+            "header_hash blob PRIMARY KEY,"
+            "prev_hash blob,"
+            "height bigint,"
+            "in_main_chain tinyint,"
+            "block blob)"
+        )
+
+        wallet = bt.get_farmer_wallet_tool()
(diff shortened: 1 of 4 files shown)


tools/generate_chain.py
```diff
@@ -0,0 +1,142 @@
...
+root_path = Path("./test-chain").resolve()
+mkdir(root_path)
+with TempKeyring() as keychain:
+
+    bt = create_block_tools(constants=test_constants, root_path=root_path, keychain=keychain)
+    initialize_logging(
+        "generate_chain", {"log_level": "DEBUG", "log_stdout": False, "log_syslog": False}, root_path=root_path
+    )
+
+    with closing(sqlite3.connect("stress-test-blockchain.sqlite")) as db:
```
trepca, line 43: why not `:memory:`? disk reads/writes can produce inconsistent profiling results

tools/generate_chain.py
```diff
@@ -0,0 +1,142 @@
...
+                    if coin.puzzle_hash == coinbase_puzzlehash:
+                        unspent_coins.append(coin)
+                unspent_coins.extend(new_coins)
+
+                if b.transactions_info:
+                    fill_rate = b.transactions_info.cost / test_constants.MAX_BLOCK_COST_CLVM
+                else:
+                    fill_rate = 0
+
+                end_time = time.monotonic()
```
trepca, line 125: you might want to use `time.perf_counter` (or `time.process_time`) as it has a much higher tick rate and is meant for benchmarking https://peps.python.org/pep-0418/

tools/generate_chain.py
```diff
@@ -0,0 +1,142 @@
...
+            time_per_block=30,
+        )
+
+        unspent_coins: List[Coin] = []
+
+        for b in blocks:
+            for coin in b.get_included_reward_coins():
+                if coin.puzzle_hash == coinbase_puzzlehash:
+                    unspent_coins.append(coin)
+            db.execute(
```
trepca, line 73: it'd be faster to use `executemany` here

tools/generate_chain.py
trepca: ok, given this is meant to generate a db, then I'd suggest 
```
PRAGMA journal_mode=WAL;
PRAGMA synchronous = 0;
PRAGMA journal_mode=MEMORY;
```
to speed things up

arvidn: The whole point of this tool is to produce this database file, it has to be saved to disk. It currently takes ~50 seconds to produce one block. The time spent writing that block to disk is negligible

tools/generate_chain.py
arvidn: this is mostly not used for benchmarking, but to let the user estimate how many hours it will take to produce the output chain. This time is in the order of tens of seconds, the clock precision is not critical.

tools/generate_chain.py
arvidn: probably, but it wouldn't affect the runtime of the program (because the disk write is negligible), and it was more important to me to have the block written to disk as soon as possible after it was done, so I could run a test against it. to ensure the blocks validate successfully.

tools/generate_chain.py
arvidn: I used those pragmas initially, as well as `locking_mode=exclusive`. But being able to read the database from a separate process while this process is still writing to it is a more important feature

trepca (approved):
didn't manage to test it locally yet, but lgtm

wjblanke (approved):
aok

PR #11147, opened 2022-04-13
Fix farmer rewards checks and add observer tests

emlowe:
The GUI farmer rewards check was not checking for observer keys and the code was duplicated from farmer and wallet, which is one reason this was missed.

Consolidated some code into new function called by both the wallet and farmer and added tests for observer keys.

Also added the ability to pass in the number of derivations to the RPC call - primarily this was so the tests can pass in much smaller numbers than the default (500) and so the tests can run much faster. However, this flexibility seems worth while regardless


xdustinface (changes requested):
I will review tomorrow but one thing i noticed as i skimmed it: I think you should drop all those unrelated `isort` changes.. imo we shouldn't just mix those into other PRs to have clear scopes why things have changed.

chia/farmer/farmer.py
```diff
@@ -500,7 +499,7 @@ async def update_pool_farmer_info() -> Tuple[Optional[GetFarmerResponse], Option
                         farmer_info, error_code = await update_pool_farmer_info()
                         if error_code == PoolErrorCode.FARMER_NOT_KNOWN:
                             # Make the farmer known on the pool with a POST /farmer
-                            owner_sk_and_index: Optional[PrivateKey, uint32] = find_owner_sk(
+                            owner_sk_and_index: Optional[Tuple[G1Element, uint32]] = find_owner_sk(
```
paninaro, line 502: Seems odd to me that `find_owner_sk` is returning a `G1Element` instead of a `PrivateKey`. I would fix the return type of `find_owner_sk` instead of changing the type of `owner_sk_and_index`.

chia/farmer/farmer.py
```diff
@@ -524,7 +523,7 @@ async def update_pool_farmer_info() -> Tuple[Optional[GetFarmerResponse], Option
                             and pool_config.payout_instructions.lower() != farmer_info.payout_instructions.lower()
                         )
                         if payout_instructions_update_required or error_code == PoolErrorCode.INVALID_SIGNATURE:
-                            owner_sk_and_index: Optional[PrivateKey, uint32] = find_owner_sk(
+                            owner_sk_and_index: Optional[Tuple[G1Element, uint32]] = find_owner_sk(
```
paninaro, line 526: Same comment here re: `find_owner_sk` return type.

chia/farmer/farmer.py
```diff
@@ -547,25 +546,30 @@ def get_public_keys(self):
...
+    async def get_reward_targets(self, search_for_private_key: bool, max_ph_to_search: int = 500) -> Dict:
         if search_for_private_key:
             all_sks = await self.get_all_private_keys()
-            stop_searching_for_farmer, stop_searching_for_pool = False, False
-            for i in range(500):
-                if stop_searching_for_farmer and stop_searching_for_pool and i > 0:
+            have_farmer_sk, have_pool_sk = False, False
+            search_addresses: List[bytes32] = [self.farmer_target, self.pool_target]
+            for sk, _ in all_sks:
+                found_dict: Dict[bytes32, bool] = match_address_to_sk(sk, search_addresses, max_ph_to_search)
```
paninaro, line 555: minor: A set would be a bit tidier to work with, but the dictionary approach works as well.

chia/wallet/derive_keys.py
```diff
@@ -110,3 +112,26 @@ def find_authentication_sk(all_sks: List[PrivateKey], owner_pk: G1Element) -> Op
                 # NOTE: ONLY use 0 for authentication key index to ensure compatibility
                 return master_sk_to_pooling_authentication_sk(sk, uint32(pool_wallet_index), uint32(0))
     return None
+
+
+def match_address_to_sk(
+    sk: PrivateKey, addresses_to_search: List[bytes32], max_ph_to_search: int = 500
+) -> Dict[bytes32, bool]:
```
paninaro, line 119: minor: As mentioned earlier, a set would be a bit more natural to work with, but not critical.

chia/wallet/derive_keys.py
```diff
@@ -110,3 +112,26 @@ def find_authentication_sk(all_sks: List[PrivateKey], owner_pk: G1Element) -> Op
...
+    if sk is None or not addresses_to_search:
+        return {}
+
+    found_dict: Dict[bytes32, bool] = {}
+    for i in range(max_ph_to_search):
+        phs = [
+            create_puzzlehash_for_pk(master_sk_to_wallet_sk(sk, uint32(i)).get_g1()),
+            create_puzzlehash_for_pk(master_sk_to_wallet_sk_unhardened(sk, uint32(i)).get_g1()),
+        ]
+        for address in addresses_to_search:
```
paninaro, line 130: I'd make a copy of addresses_to_search and remove entries as they're found.

emlowe: I thought about this, but at the moment, it's only used for a List of 2 and for small lists, I thought that overhead wouldn't be worth it. Although there is some problem here if the list contains any duplicated addresses, this will loop around more than required as the length check doesn't work in this case, so I'll need to fix that.

chia/farmer/farmer.py
emlowe: yea, good idea

chia/farmer/farmer.py
emlowe: I wasn't sure which direction to go here - I also found it odd, but that is long-standing code, so I wasn't sure.

emlowe: Fixed, turns out all callers were expecting find_owner_sk to return PrivateKey - so fixed that

paninaro (dismissed):
lgtm

PR #11148, opened 2022-04-13
tests: Improve `test_wallet_rpc.py`

xdustinface:
This is preparation for a follow up PR which will split up the single test into smaller tests.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -209,7 +209,6 @@ extend_skip=
     tests/wallet/cat_wallet/test_offer_lifecycle.py
     tests/wallet/did_wallet/test_did.py
     tests/wallet/did_wallet/test_did_rpc.py
-    tests/wallet/rpc/test_wallet_rpc.py
     tests/wallet/simple_sync/test_simple_sync_protocol.py
     tests/wallet/test_singleton_lifecycle_fast.py
     tests/wallet/test_singleton_lifecycle.py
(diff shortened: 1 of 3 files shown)


mariano54 (changes requested):
Tests are much faster, which is good, but is that because we removed parametrization? I think it's important for this test to test trusted and untrusted.

This is a great change though, it was really needed.

xdustinface:
> Tests are much faster, which is good, but is that because we removed parametrization? 

Nope with parameterization we just would have two runs with the same time. Its mostly because it always generated 5 blocks just to confirm a transaction with a 0.5s sleep between the blocks and there were also just some other `sleep` calls in there which i replaced with waits.

> I think it's important for this test to test trusted and untrusted.

Last time we talked about this on zoom you agreed its ok to not test both 😄 But i can look into it tomorrow or so how to make parameterize it again since its tricker now with the fixture i think.

mariano54:
Yeah my bad, I think I misunderstood. I think it's fine not to test both if the test is not closely related to the wallet. But in this case it's testing important wallet features that might not be properly tested elsewhere so we do need to test untrusted

xdustinface:
@mariano54 Rebased + See last two commits, one recovers trusted/untrusted sync and the other does some basic splitting of the tests.

mariano54 (approved):

PR #11149, opened 2022-04-13
wallet: Some rollback fixes

xdustinface:
Note that im not super familiar yet with all the surrounding code/workflows so, review well :)

- 31a86d462e0f1e745a529927c0a67079187f9b27 Pretty sure this is needs to be applied but not sure if those calls even raise in practice
- b3763cb880e38badb449209af7b18f10bbcbde84 and b16a2c5f75d855d43b39ec8ceea0a0c89bc4640f Not sure if they are really needed or if the block records and wallet would also just recover in some other way if they are removed but the DB transaction rolled back


mariano54 (approved):
LGTM. The commits are needed, failing and rolling back but not having the block records can be bad

PR #11154, opened 2022-04-13
optimize wallet tool

arvidn:
by not caching the puzzle_hash -> derivation index, but caching puzzle_hash -> secret key (which is the lookup we're actually interested in). This avoids duplicating the actual derivation.

Searching for the derivation index isn't the only expensive step in finding the private key, the call to `master_sk_to_wallet_sk()` is also expensive. This saves that call.

### DIFF
--- a/tests/wallet_tools.py
+++ b/tests/wallet_tools.py
@@ -30,6 +30,7 @@ assert len(DEFAULT_SEED) == 32
 class WalletTool:
     next_address = 0
     pubkey_num_lookup: Dict[bytes, uint32] = {}
+    puzzle_pk_cache: Dict[bytes32, PrivateKey] = {}
 
     def __init__(self, constants: ConsensusConstants, sk: Optional[PrivateKey] = None):
         self.constants = constants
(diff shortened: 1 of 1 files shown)


tests/wallet_tools.py
```diff
@@ -30,6 +30,7 @@
 class WalletTool:
     next_address = 0
     pubkey_num_lookup: Dict[bytes, uint32] = {}
+    puzzle_pk_cache: Dict[bytes32, PrivateKey] = {}
```
aqk, line 33:
```suggestion
    puzzle_sk_cache: Dict[bytes32, PrivateKey] = {}
```

By cryptography programmer conventions, this variable should be named `puzzle_sk_cache`

tests/wallet_tools.py
```diff
@@ -48,29 +49,27 @@ def get_next_address_index(self) -> uint32:
...
-            child = self.puzzle_pk_cache[puzzle_hash]
-            private = master_sk_to_wallet_sk(self.private_key, uint32(child))
-            #  pubkey = private.get_g1()
-            return private
-        else:
-            for child in range(self.next_address):
-                pubkey = master_sk_to_wallet_sk(self.private_key, uint32(child)).get_g1()
-                if puzzle_hash == puzzle_for_pk(bytes(pubkey)).get_tree_hash():
-                    return master_sk_to_wallet_sk(self.private_key, uint32(child))
+        sk = self.puzzle_pk_cache.get(puzzle_hash)
```
aqk, line 52:
```suggestion
        sk = self.puzzle_sk_cache.get(puzzle_hash)
```

tests/wallet_tools.py
```diff
@@ -48,29 +49,27 @@ def get_next_address_index(self) -> uint32:
...
         next_address_index: uint32 = self.get_next_address_index()
-        pubkey: G1Element = master_sk_to_wallet_sk(self.private_key, next_address_index).get_g1()
+        sk: PrivateKey = master_sk_to_wallet_sk(self.private_key, next_address_index)
+        pubkey: G1Element = sk.get_g1()
         self.pubkey_num_lookup[bytes(pubkey)] = next_address_index
 
         puzzle: Program = puzzle_for_pk(pubkey)
 
-        self.puzzle_pk_cache[puzzle.get_tree_hash()] = next_address_index
+        self.puzzle_pk_cache[puzzle.get_tree_hash()] = sk
```
aqk, line 72:
```suggestion
        self.puzzle_sk_cache[puzzle.get_tree_hash()] = sk
```

aqk (approved):

tests/wallet_tools.py
arvidn: I wanted to keep my changes focused on just this issue of mine. Also, by the same standards, the `PrivateKey` type should be called `SecretKey` :)

tests/wallet_tools.py
```diff
@@ -48,29 +49,27 @@ def get_next_address_index(self) -> uint32:
...
-                pubkey = master_sk_to_wallet_sk(self.private_key, uint32(child)).get_g1()
-                if puzzle_hash == puzzle_for_pk(bytes(pubkey)).get_tree_hash():
-                    return master_sk_to_wallet_sk(self.private_key, uint32(child))
+        sk = self.puzzle_pk_cache.get(puzzle_hash)
+        if sk:
+            return sk
+        for child in range(self.next_address):
+            pubkey = master_sk_to_wallet_sk(self.private_key, uint32(child)).get_g1()
+            if puzzle_hash == puzzle_for_pk(bytes(pubkey)).get_tree_hash():
+                return master_sk_to_wallet_sk(self.private_key, uint32(child))
```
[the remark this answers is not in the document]
arvidn: I don't think that would help, and maybe make things worse. I was considering removing entries from the cache here. Typically we call this function to get the private key to spend a coin. once spent, we are very unlikely to call this for the same puzzle hash again

wjblanke (approved):
aok

PR #11155, opened 2022-04-13
wallet sync speed up by avoiding serialisation on the node

trepca:
- adds capability to the full node api to expose "block headers api", avoiding transactions filter as new wallet sync doesn't require it anymore


chia/full_node/block_store.py
```diff
@@ -400,6 +406,37 @@ async def get_block_records_by_hash(self, header_hashes: List[bytes32]):
             ret.append(all_blocks[hh])
         return ret
 
+    async def get_block_blobs_by_hash(self, header_hashes: List[bytes32]) -> List[bytes]:
```
arvidn, line 409: The current naming scheme is this:

| return type | single hash | multiple hashes |
| --- | --- | --- |
| `FullBlock` | `get_full_block()` | `get_blocks_by_hash()` |
| `bytes` | `get_full_block_bytes()` | *this is your new function* |

I think the naming would be more consistent if this function would be called `get_block_bytes_by_hash()`.

chia/server/rate_limits.py
```diff
@@ -92,6 +92,9 @@ class RLSettings:
     ProtocolMessageTypes.request_header_blocks: RLSettings(500, 100),
     ProtocolMessageTypes.reject_header_blocks: RLSettings(100, 100),
     ProtocolMessageTypes.respond_header_blocks: RLSettings(500, 2 * 1024 * 1024, 100 * 1024 * 1024),
+    ProtocolMessageTypes.request_header_block_blobs: RLSettings(500, 100),
+    ProtocolMessageTypes.reject_header_block_blobs: RLSettings(100, 100),
+    ProtocolMessageTypes.respond_header_block_blobs: RLSettings(500, 2 * 1024 * 1024, 100 * 1024 * 1024),
```
arvidn, line 97: I don't think we need new APIs for this. The existing APIs always serialize everything already. The only difference you make here is that you serialize the block as a blob of bytes (which in-turn is a serialized FullBlock). It would seem reasonable to just include the serialized blob verbatim into the existing response instead.

chia/full_node/block_store.py
arvidn: Also, this function needs test coverage in the block store test

arvidn (commented):
I think a lot of this code could be simplified away by not introducing a new API.
If we add new API calls. we need new tests for those as well.

chia/server/rate_limits.py
mariano54: You do need a new request because the current one expects a filter in the response (for old protocol)

chia/protocols/wallet_protocol.py
```diff
@@ -140,6 +140,28 @@ class RequestHeaderBlocks(Streamable):
...
+    start_height: uint32
+    end_height: uint32
+
+
+@streamable
+@dataclass(frozen=True)
+class RespondHeaderBlockBlobs(Streamable):
+    start_height: uint32
+    end_height: uint32
+    header_block_blobs: List[bytes]
```
mariano54, line 155: Everything in streamable is already streamed as bytes so we don't need to use bytes I  the protocol message, we can use header blocks. As long as we return the bytes verbatim from DB without calling HeaderBlock constructors

chia/full_node/full_node_api.py
```diff
@@ -1314,13 +1314,34 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
...
+        header_hashes: List[bytes32] = []
+        for i in range(request.start_height, request.end_height + 1):
+            header_hash: Optional[bytes32] = height_to_hash(uint32(i))
+            if header_hash is None:
+                reject = RejectHeaderBlocks(request.start_height, request.end_height)
+                msg = make_msg(ProtocolMessageTypes.reject_header_blocks, reject)
+                return msg
+            header_hashes.append(header_hash)
 
+        blocks: List[bytes] = await self.full_node.block_store.get_block_blobs_by_hash(header_hashes)
```
mariano54, line 1330: We seem to be returning full blocks, but we need header blocks right?

chia/full_node/full_node_api.py
```diff
@@ -1314,13 +1314,34 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
         return response_msg
 
     @api_request
-    async def request_header_blocks(self, request: wallet_protocol.RequestHeaderBlocks) -> Optional[Message]:
+    async def request_header_block_blobs(self, request: wallet_protocol.RequestHeaderBlockBlobs) -> Optional[Message]:
```
mariano54, line 1317: We should add an extra param in the request called return_filter. The old protocol still uses the filter so it would be nice to still support returning it, even if we don't use it in current wallet

chia/protocols/wallet_protocol.py
mariano54: The response object can be the same

chia/protocols/shared_protocol.py
```diff
@@ -17,6 +17,7 @@
 # These are passed in as uint16 into the Handshake
 class Capability(IntEnum):
     BASE = 1  # Base capability just means it supports the chia protocol at mainnet
+    BLOB_API = 2
```
mariano54, line 20: I would call this something different since we don't need to use blobs directly to obtain the speedup

chia/full_node/full_node_api.py
mariano54: We are returning a large amount of data, and forcing the wallet to parse it all which is very slow

mariano54 (changes requested):

chia/full_node/full_node_api.py
trepca: but the old protocol would just use the old method no?

mariano54: Yeah but in the future we can depreciate the old one, but the filter might still be useful for other developers

mariano54: I'm not opposed to keeping both if it's a clean solution though

chia/server/rate_limits.py
arvidn: an alternative would be to accept an optional flag to indicate that you don't need the filter in the response. But maybe that's more complicated (although, it would make it backwards compatible)

chia/protocols/shared_protocol.py
```diff
@@ -17,6 +17,7 @@
 # These are passed in as uint16 into the Handshake
 class Capability(IntEnum):
     BASE = 1  # Base capability just means it supports the chia protocol at mainnet
+    BLOCK_HEADERS = 2
```
trepca, line 20: we have some name clashing here, we had `RequestHeaderBlocks` (old api call) and `RespondBlockHeader`, and new api call is `RequestBlockHeaders` 

naming is hard 🤷

chia/server/rate_limits.py
```diff
@@ -92,6 +92,9 @@ class RLSettings:
     ProtocolMessageTypes.request_header_blocks: RLSettings(500, 100),
     ProtocolMessageTypes.reject_header_blocks: RLSettings(100, 100),
     ProtocolMessageTypes.respond_header_blocks: RLSettings(500, 2 * 1024 * 1024, 100 * 1024 * 1024),
+    ProtocolMessageTypes.request_block_headers: RLSettings(500, 100),
+    ProtocolMessageTypes.reject_block_headers: RLSettings(100, 100),
+    ProtocolMessageTypes.respond_block_headers: RLSettings(500, 2 * 1024 * 1024, 100 * 1024 * 1024),
```
trepca, line 97: this probably needs to be adjusted

tests/core/full_node/stores/test_block_store.py
```diff
@@ -264,3 +266,40 @@ def generator(i: int) -> SerializedProgram:
             assert await store.get_generator(blocks[4].header_hash) == new_blocks[4].transactions_generator
             assert await store.get_generator(blocks[6].header_hash) == new_blocks[6].transactions_generator
             assert await store.get_generator(blocks[7].header_hash) == new_blocks[7].transactions_generator
+
+
+class TestBlockStoreAPI:
```
arvidn, line 271: I don't think you should put these tests inside a `class`. I think it's been a mistake that we have done this with existing tests

tests/core/full_node/stores/test_block_store.py
```diff
@@ -264,3 +266,40 @@ def generator(i: int) -> SerializedProgram:
...
+            full_blocks_by_hash = await store.get_blocks_by_hash(hashes)
+            assert full_blocks_by_hash == blocks
+
+            full_block_bytes_by_hash = await store.get_block_bytes_by_hash(hashes)
+
+            assert [FullBlock.from_bytes(x) for x in full_block_bytes_by_hash] == blocks
+
+            assert not await store.get_block_bytes_by_hash([])
+            with pytest.raises(ValueError):
+                await store.get_block_bytes_by_hash([bytes32(b"yolo")])
```
arvidn, line 305: this is testing the `bytes32` constructor, not `BlockStore`. you should make the has a valid 32 byte blob

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -26,6 +30,44 @@ def wallet_height_at_least(wallet_node, h):
 log = getLogger(__name__)
 
 
+class TestWalletProtocol:
```
arvidn, line 33: I don't think you should put this test in a `class` either.

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -26,6 +30,44 @@ def wallet_height_at_least(wallet_node, h):
...
+        new_blocks = bt.get_consecutive_blocks(
+            num_blocks, block_list_input=default_1000_blocks, pool_reward_puzzle_hash=ph
+        )
+        for i in range(0, len(new_blocks)):
+            await full_node_api.full_node.respond_block(full_node_protocol.RespondBlock(new_blocks[i]))
+
+        msg = await full_node_api.request_block_headers(wallet_protocol.RequestBlockHeaders(110, 115, True))
+        res_block_headers = RespondBlockHeaders.from_bytes(msg.data)
+        bh = res_block_headers.header_blocks
+        assert len(bh) == 6
```
arvidn, line 68: there's no tests that trigger the reject message. I think there should be. Especially to ensure we disallow requesting too many blocks, but we should also ensure we correctly reject other invalid requests

chia/full_node/full_node.py
```diff
@@ -128,7 +129,17 @@ def __init__(
...
         self.compact_vdf_requests: Set[bytes32] = set()
         self.log = logging.getLogger(name if name else __name__)
-
+        self.capabilities = capabilities
+        if config.get("disable_capabilities"):
+            try:
+                disable_capabilities_values = config["disable_capabilities"]
+                if Capability.BASE.value in disable_capabilities_values:
+                    disable_capabilities_values.remove(Capability.BASE.value)
+                self.capabilities = [
```
arvidn, line 138: I would have thought this would be a bitfield. is this a list of strings?

chia/server/rate_limits.py
mariano54: Not sure how to add an optional argument, this is not really possible with streamable protocol right now.

chia/full_node/block_store.py
```diff
@@ -400,6 +406,37 @@ async def get_block_records_by_hash(self, header_hashes: List[bytes32]):
...
+        if len(header_hashes) == 0:
+            return []
+
+        header_hashes_db: Tuple[Any, ...]
+        if self.db_wrapper.db_version == 2:
+            header_hashes_db = tuple(header_hashes)
+        else:
+            header_hashes_db = tuple([hh.hex() for hh in header_hashes])
+        formatted_str = (
+            f'SELECT header_hash, block from full_blocks WHERE header_hash in ({"?," * (len(header_hashes_db) - 1)}?)'
```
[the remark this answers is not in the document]
trepca: added limit and a test

chia/full_node/full_node.py
trepca: agree, but it's unfortunately defined like this in Handshake protocol: `capabilities: List[Tuple[uint16, str]]`

chia/full_node/full_node.py
```diff
@@ -128,7 +129,17 @@ def __init__(
         self.uncompact_task = None
         self.compact_vdf_requests: Set[bytes32] = set()
         self.log = logging.getLogger(name if name else __name__)
-
+        self.capabilities = capabilities
+        if config.get("disable_capabilities"):
```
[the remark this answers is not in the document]
trepca: this is supposed to allow user to disable capabilities if user doesn't want to use them (and also easier to make it testable), but I think it's making less and less sense to me to use capabilities vs protocol
especially as in our case we're upgrading the protocol and not "adding features"
@arvidn what do you think?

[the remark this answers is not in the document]
arvidn: we would still need some mechanism to communicate which protocol messages we support though.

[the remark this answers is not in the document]
arvidn: but you mean we could bump the protocol version number instead?

chia/util/full_block_utils.py
```diff
@@ -207,3 +212,57 @@ def generator_from_block(buf: memoryview) -> Optional[SerializedProgram]:
...
+
+    transactions_info: Optional[TransactionsInfo] = None
+    transactions_info_optional: bytes = b""
+    # this is the transactions_info optional
+    if buf2[0] == 0:
+        transactions_info_optional = bytes([0])
+    else:
+        buf3 = buf2[1:]
+        transactions_info_optional = bytes([1])
+        transactions_info = TransactionsInfo.parse(io.BytesIO(buf3))
```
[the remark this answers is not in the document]
trepca: great catch, updated

chia/full_node/full_node.py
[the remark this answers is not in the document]
trepca: yes, bump protocol
or can continue with this, and add comments in config etc

chia/full_node/full_node_api.py
```diff
@@ -1314,13 +1316,40 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
         return response_msg
 
     @api_request
-    async def request_header_blocks(self, request: wallet_protocol.RequestHeaderBlocks) -> Optional[Message]:
+    async def request_block_headers(self, request: wallet_protocol.RequestBlockHeaders) -> Optional[Message]:
         if request.end_height < request.start_height or request.end_height - request.start_height > 32:
             return None
```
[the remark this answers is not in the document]
trepca: not sure, reject msg would make more sense, but old method does this and wallet code is expecting None too

given it's a new method, I can change it to reject, but this PR is getting quite large already, maybe as a separate PR better

chia/full_node/full_node_api.py
```diff
@@ -1314,13 +1316,40 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
...
+        header_hashes: List[bytes32] = []
+        for i in range(request.start_height, request.end_height + 1):
+            header_hash: Optional[bytes32] = height_to_hash(uint32(i))
+            if header_hash is None:
+                reject = RejectBlockHeaders(request.start_height, request.end_height)
+                msg = make_msg(ProtocolMessageTypes.reject_header_blocks, reject)
+                return msg
+            header_hashes.append(header_hash)
+
+        blocks_bytes: List[bytes] = await self.full_node.block_store.get_block_bytes_by_hash(header_hashes)
```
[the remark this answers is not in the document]
trepca: true, can we do this as a separate PR?

tests/wallet/sync/test_wallet_sync.py
trepca: added

chia/full_node/full_node_api.py
```diff
@@ -1314,13 +1316,40 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
...
+        respond_header_blocks_manually_streamed: bytes = (
+            bytes(uint32(request.start_height))
+            + bytes(uint32(request.end_height))
+            + len(header_blocks_bytes).to_bytes(4, "big", signed=False)
+        )
+        respond_header_blocks_manually_streamed += b"".join(header_blocks_bytes)
+        return make_msg(ProtocolMessageTypes.respond_block_headers, respond_header_blocks_manually_streamed)
+
+    @api_request
+    async def request_header_blocks(self, request: wallet_protocol.RequestHeaderBlocks) -> Optional[Message]:
```
arvidn, line 1346: Are we documenting this protocol, @danieljperry? (this is a new, more efficient, RPC in the light-wallet <-> full-node protocol)

chia/server/rate_limits.py
arvidn: I take it this means you can download at most 100 MiB of blocks from a peer, is that right?
@mariano54

chia/full_node/full_node_api.py
```diff
@@ -1314,13 +1316,40 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
...
+            header_hash: Optional[bytes32] = height_to_hash(uint32(i))
+            if header_hash is None:
+                reject = RejectBlockHeaders(request.start_height, request.end_height)
+                msg = make_msg(ProtocolMessageTypes.reject_block_headers, reject)
+                return msg
+            header_hashes.append(header_hash)
+
+        blocks_bytes: List[bytes] = await self.full_node.block_store.get_block_bytes_by_hash(header_hashes)
+        header_blocks_bytes: List[bytes] = [
+            header_block_from_block(memoryview(b), request.return_filter) for b in blocks_bytes
```
arvidn, line 1334: is there a use case for the `return_filter` parameter? If we can't think of one, we probably shouldn't include it, and simplify the parser

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -27,6 +31,67 @@ def wallet_height_at_least(wallet_node, h):
...
+
+    @pytest.mark.asyncio
+    async def test_request_block_headers_rejected(self, bt, wallet_node, default_1000_blocks):
```
arvidn, line 71: you can use `pytest` parameterization to boil down this test to its essence, and have it run over all parameters

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -27,6 +31,67 @@ def wallet_height_at_least(wallet_node, h):
...
+        assert [x.transactions_filter for x in bh] == [b"\x00"] * 6
+
+        num_blocks = 20
+        new_blocks = bt.get_consecutive_blocks(
+            num_blocks, block_list_input=default_1000_blocks, pool_reward_puzzle_hash=ph
+        )
+        for i in range(0, len(new_blocks)):
+            await full_node_api.full_node.respond_block(full_node_protocol.RespondBlock(new_blocks[i]))
+
+        msg = await full_node_api.request_block_headers(wallet_protocol.RequestBlockHeaders(110, 115, True))
```
arvidn, line 65: as far as I can tell, this is the only place you test passing in `True` for `return_filter`. It doesn't looks like the test ensures the filter is correct though

chia/full_node/full_node_api.py
danieljperry: @arvidn Not yet, thanks for the heads-up.

chia/server/rate_limits.py
mariano54: Correct

chia/full_node/full_node.py
```diff
@@ -128,7 +129,17 @@ def __init__(
         self.uncompact_task = None
         self.compact_vdf_requests: Set[bytes32] = set()
         self.log = logging.getLogger(name if name else __name__)
-
+        self.capabilities = capabilities
+        disable_capabilities_values = config.get("disable_capabilities")
```
mariano54, line 133: type annotation

chia/full_node/full_node.py
```diff
@@ -128,7 +129,17 @@ def __init__(
...
+        if disable_capabilities_values and isinstance(disable_capabilities_values, list):
```
mariano54, line 134: if disable_capabilities is not None

chia/full_node/full_node.py
mariano54: What is the point of `disable_capabilities`?

chia/full_node/full_node_api.py
```diff
@@ -1314,13 +1316,40 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
         return response_msg
 
     @api_request
-    async def request_header_blocks(self, request: wallet_protocol.RequestHeaderBlocks) -> Optional[Message]:
+    async def request_block_headers(self, request: wallet_protocol.RequestBlockHeaders) -> Optional[Message]:
         if request.end_height < request.start_height or request.end_height - request.start_height > 32:
```
mariano54, line 1320: Can we increase the max num blocks?

chia/full_node/full_node_api.py
```diff
@@ -1314,13 +1316,40 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
...
+
+    @api_request
+    async def request_header_blocks(self, request: wallet_protocol.RequestHeaderBlocks) -> Optional[Message]:
+        if request.end_height < request.start_height or request.end_height - request.start_height > 32:
+            return None
+        height_to_hash = self.full_node.blockchain.height_to_hash
         header_hashes: List[bytes32] = []
         for i in range(request.start_height, request.end_height + 1):
-            header_hash: Optional[bytes32] = self.full_node.blockchain.height_to_hash(uint32(i))
+            header_hash: Optional[bytes32] = height_to_hash(uint32(i))
```
mariano54, line 1352: why create a function variable instead of just calling it?

chia/full_node/full_node_api.py
```diff
@@ -1330,8 +1359,11 @@ async def request_header_blocks(self, request: wallet_protocol.RequestHeaderBloc
         blocks: List[FullBlock] = await self.full_node.block_store.get_blocks_by_hash(header_hashes)
         header_blocks = []
         for block in blocks:
-            added_coins_records = await self.full_node.coin_store.get_coins_added_at_height(block.height)
-            removed_coins_records = await self.full_node.coin_store.get_coins_removed_at_height(block.height)
+            added_coins_records_coroutine = self.full_node.coin_store.get_coins_added_at_height(block.height)
+            removed_coins_records_coroutine = self.full_node.coin_store.get_coins_removed_at_height(block.height)
+            added_coins_records, removed_coins_records = await asyncio.gather(
+                added_coins_records_coroutine, removed_coins_records_coroutine
+            )
```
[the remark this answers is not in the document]
mariano54: doesn't hurt to have it though right? But I agree it would be better to aggregate

chia/protocols/shared_protocol.py
```diff
@@ -28,3 +29,6 @@ class Handshake(Streamable):
     server_port: uint16
     node_type: uint8
     capabilities: List[Tuple[uint16, str]]
+
+
+capabilities = [(uint16(Capability.BASE.value), "1"), (uint16(Capability.BLOCK_HEADERS.value), "2")]
```
mariano54, line 34: The intent here was that "1" would be enabled, so each capability would have "1" as the value. But we have not standardized or documented any of this, this is the first use of capabilities.

chia/server/rate_limits.py
mariano54: We should increase the number of request per minute, wallets can make many of these requests

chia/wallet/wallet_node.py
```diff
@@ -1239,9 +1238,12 @@ async def validate_received_state_from_peer(
                 # Peer is telling us that coin that was previously known to be spent is not spent anymore
                 # Check old state
 
-                request = RequestHeaderBlocks(current.spent_block_height, current.spent_block_height)
-                res = await peer.request_header_blocks(request)
-                spent_state_block = res.header_blocks[0]
+                spent_state_blocks = await request_header_blocks(
+                    peer, current.spent_block_height, current.spent_block_height
+                )
+                if not spent_state_blocks:
```
mariano54, line 1244: if spent_state_blocks is None

chia/wallet/wallet_node.py
```diff
@@ -1239,9 +1238,12 @@ async def validate_received_state_from_peer(
                 # Peer is telling us that coin that was previously known to be spent is not spent anymore
                 # Check old state
 
-                request = RequestHeaderBlocks(current.spent_block_height, current.spent_block_height)
-                res = await peer.request_header_blocks(request)
-                spent_state_block = res.header_blocks[0]
+                spent_state_blocks = await request_header_blocks(
```
mariano54, line 1241: type annotation

chia/full_node/full_node_api.py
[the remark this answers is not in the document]
trepca: true, will add support

chia/full_node/full_node_api.py
trepca: increased to 128

chia/full_node/full_node_api.py
trepca: because in python every `.` means a key lookup every time it's called, this creates a pointer to function avoiding extra lookups. (it does add up)

chia/full_node/full_node_api.py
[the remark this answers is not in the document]
trepca: ok, so what should we do? should I revert for now?

chia/protocols/shared_protocol.py
trepca: ok, added some docs and made it so that "1" is considered when checking what features are enabled

tests/util/test_full_block_utils.py
```diff
@@ -243,3 +246,10 @@ async def test_parser(self):
             assert gen == block.transactions_generator
             # this doubles the run-time of this test, with questionable utility
             # assert gen == FullBlock.from_bytes(block_bytes).transactions_generator
+
+    @pytest.mark.asyncio
+    async def test_header_block(self):
+        for block in get_full_blocks():
+            hb: HeaderBlock = get_block_header(block, [], [])
+            hb_bytes = header_block_from_block(memoryview(bytes(block)))
+            assert HeaderBlock.from_bytes(hb_bytes) == hb
```
arvidn, line 255: I suppose this is a simple way of writing this test. but I would expect it to take a lot longer to run than it would need to. iirc, looping over all `get_full_blocks()` takes about 4 minutes. The `HeaderBlock` is smaller and I would expect there to be fewer variations of it to generate in the test.

tests/util/test_full_block_utils.py
```diff
@@ -204,9 +207,9 @@ def get_full_blocks():
 
     for foliage in get_foliage():
         for foliage_transaction_block in get_foliage_transaction_block():
-            height = random.randint(0, 1000000)
+            height = random.randint(0, 100000)
```
arvidn, line 210: why did you make this change?

tests/util/test_full_block_utils.py
```diff
@@ -129,20 +131,21 @@ def get_foliage_transaction_block():
     )
 
 
-def get_transactions_info(height):
-    yield None
+def get_transactions_info(height, foliage_transaction_block):
```
arvidn, line 134: I realize I failed to add type annotations to this file when I wrote it, but I think we should work towards adding that everywhere, starting with new code we add.

is `foliage_transaction_block` a `bool`? because it doesn't look like what's passed in to this call is.

tests/util/test_full_block_utils.py
```diff
@@ -129,20 +131,21 @@ def get_foliage_transaction_block():
...
+        yield TransactionsInfo(
+            hsh(),  # generator_root
+            hsh(),  # generator_refs_root
+            g2(),  # aggregated_signature
+            fees,
+            uint64(random.randint(0, 12000000000)),  # cost
+            reward_claims_incorporated,
+        )
+    else:
+        yield None
```
arvidn, line 148: you should check this condition and return `None` as early as possible in this function, to avoid paying the cost of generating all those other values. Alternatively you could move those into the if-block as well. But an early exit saves on indentation levels

tests/wallet/sync/test_wallet_sync.py
arvidn: I think this would still make sense. you can parameterize it on tuples of (start_height, end_height, expected_response). At least for the tests of `request_block_headers()`

chia/protocols/shared_protocol.py
```diff
@@ -28,3 +31,7 @@ class Handshake(Streamable):
     server_port: uint16
     node_type: uint8
     capabilities: List[Tuple[uint16, str]]
+
+
+# "1" means capability is enabled
+capabilities = [(uint16(Capability.BASE.value), "1"), (uint16(Capability.BLOCK_HEADERS.value), "1")]
```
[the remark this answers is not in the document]
mariano54: There might be custom node types in the future that don't support the whole 1.0 Chia protocol, and maybe they just support one method, for example.

tests/util/test_full_block_utils.py
mariano54: It takes about 90 seconds on my machine (I guess do to the serialization improvements from G1), I think it's not so bad.

tests/wallet/sync/test_wallet_sync.py
mariano54: I tried it, it works, but it makes the test slower here, since we are adding 150 blocks to the chain. It might not be worth the performance hit

tests/util/test_full_block_utils.py
mariano54: i'll do this

arvidn:
`mozilla-ca` submodule needs to be reverted

chia/full_node/full_node.py
```diff
@@ -128,7 +129,22 @@ def __init__(
...
+            try:
+                if Capability.BASE.value in disable_capabilities_values:
+                    # BASE capability cannot be removed
+                    disable_capabilities_values.remove(Capability.BASE.value)
+
+                updated_capabilities = []
+                for capability in self.capabilities:
+                    if not Capability(int(capability[0])).name in disable_capabilities_values:
```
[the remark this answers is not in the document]
trepca: yes, there is a test testing that, I do agree it would be strange for someone to try to disable base capability

tests/util/test_full_block_utils.py
trepca: shouldn't be added, this was left from testing, will remove

trepca:
see https://github.com/Chia-Network/chia-blockchain/pull/11636

chia/full_node/block_store.py
```diff
@@ -400,6 +406,39 @@ async def get_block_records_by_hash(self, header_hashes: List[bytes32]):
             ret.append(all_blocks[hh])
         return ret
 
+    async def get_block_bytes_by_hash(self, header_hashes: List[bytes32]) -> List[bytes]:
```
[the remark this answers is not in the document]
mariano54: This has been done below, by sebastian

PR #11161, opened 2022-04-13
make sure that we do not use any unnecessary dust 

jack60612:


jack60612:
test failure is unrelated

tests/wallet/test_coin_selection.py
```diff
@@ -108,7 +108,28 @@ async def test_coin_selection_with_dust(self, a_hash: bytes32) -> None:
...
+                spendable_amount,
+                DEFAULT_CONSTANTS.MAX_COIN_AMOUNT,
+                coin_list,
+                {},
+                logging.getLogger("test"),
+                uint128(target_amount),
+            )
+            assert dusty_result is not None
+            assert sum([coin.amount for coin in dusty_result]) >= target_amount
+            assert target_amount == len(dusty_result) * 2000  # make sure that we do not use any dust coins.
```
mariano54, line 131: This doesn't prove that, I think we need to do:

```
for coin in dusty_result:
    assert coin.amount > 1
```

tests/wallet/test_coin_selection.py
```diff
@@ -108,7 +108,28 @@ async def test_coin_selection_with_dust(self, a_hash: bytes32) -> None:
...
+            assert len(result) == 1  # only one coin should be selected
+
+        for i in range(100):
+            coin_list.append(
+                WalletCoinRecord(
+                    Coin(a_hash, std_hash(i), uint64(2000)), uint32(1), uint32(1), False, True, WalletType(0), 1
+                )
+            )
+        spendable_amount = uint128(spendable_amount + 2000 * 100)
+        for target_amount in [50000, 10000]:  # select the first 100 values
```
mariano54, line 120: Try some other targets that don't make us choose exact amount as well

tests/wallet/test_coin_selection.py
jack60612: Will do, i was trying to avoid that

tests/wallet/test_coin_selection.py
jack60612: gotcha.

jack60612:
oops

PR #11165, opened 2022-04-13
Use get latest release endpoint for plotters, so that we ignore any p…

cmmarslender:
…re-releases that could be returned by listReleases


hoffmang9 (approved):
lgtm

PR #11181, opened 2022-04-15
nft_transfer_nft: Introduce some error handling and make new_did_inner_hash and trade_price params optional for now

AmineKhaldi:
* Introduce some error handling.
* The `new_did_inner_hash` and `trade_price` params should be optional for now.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1306,7 +1306,9 @@ class WalletRpcApi:
         return {"wallet_id": wallet_id, "success": True, "nfts": nft_uri_pairs}
 
     async def nft_transfer_nft(self, request):
+        assert self.service.wallet_state_manager is not None
         wallet_id = int(request["wallet_id"])
+        trade_price = request.get("trade_price", 0)
         nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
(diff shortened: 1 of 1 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1244,21 +1244,28 @@ async def nft_get_current_nfts(self, request):
...
+        try:
+            sb = await nft_wallet.transfer_nft(
+                request["nft_coin_info"],
+                bytes.fromhex(request["new_did"]),
+                new_did_inner_hash,
+                trade_price,
+            )
+            return {"wallet_id": wallet_id, "success": True, "spend_bundle": sb}
+        except Exception as e:
+            log.error(f"Failed to transfer NFT: {e}")
```
trepca, line 1267: might be useful to use `log.exception` here to show full stack trace in logs for easier debugging

paninaro:
Thanks @AmineKhaldi, this is looking good. 

Just to plant the idea in your head, at some point I'd like to see us (the broader us) use error codes in addition to the error strings to allow callers to better handle errors. Currently the GUI has a few places where it does string comparison to figure out how to handle specific errors, which is gross. Also, we can better handle string localization of errors if we have a mapping of error codes to localized strings.

trepca (approved):

PR #11185, opened 2022-04-15
fix block_tools feature when specifying a list of block references

arvidn:
Also add feature `keep_going_until_tx_block`. This is different from `guarantee_transaction_block` in that you still get the non-transaction blocks in between. Which allows you to create a chain as it would normally, but still control what transactions go in the transaction blocks.

### DIFF
--- a/tests/block_tools.py
+++ b/tests/block_tools.py
@@ -419,6 +419,7 @@ class BlockTools:
         self,
         num_blocks: int,
         block_list_input: List[FullBlock] = None,
+        *,
         farmer_reward_puzzle_hash: Optional[bytes32] = None,
         pool_reward_puzzle_hash: Optional[bytes32] = None,
         transaction_data: Optional[SpendBundle] = None,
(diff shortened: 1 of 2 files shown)


arvidn:
one thing I was disappointed to learn is that there's a very large block of code that appears to be duplicated in the `get_consecutive_blocks()` function. The feature I added a while back, to allow specifying the block ref list explicitly had to be applied to the second block of code.

tests/block_tools.py
```diff
@@ -850,12 +855,15 @@ def get_consecutive_blocks(
...
-                                block_generator = best_solution_generator_from_template(
+                            if type(previous_generator) is CompressorArg:
+                                block_generator: Optional[BlockGenerator] = best_solution_generator_from_template(
                                     previous_generator, transaction_data
                                 )
                             else:
                                 block_generator = simple_solution_generator(transaction_data)
+                                if type(previous_generator) is list:
+                                    block_generator = BlockGenerator(block_generator.program, [], previous_generator)
+
```
arvidn, line 866: this is the second place this check had to be added to. This is to fix the existing feature where `previous_generator` is allowed to be a list of heights.

tests/block_tools.py
```diff
@@ -650,7 +656,7 @@ def get_consecutive_blocks(
                         latest_block = blocks[full_block.header_hash]
                         finished_sub_slots_at_ip = []
                         num_blocks -= 1
-                        if num_blocks == 0:
+                        if num_blocks <= 0 and not keep_going_until_tx_block:
```
aqk, line 659: This return condition would be a bit easier to read as something like `if done:` with done being set above, but I see `keep_going_until_tx_block` is used elsewhere. No action needed.

aqk (approved):

PR #11187, opened 2022-04-15
Cmj.bladebit

ChiaMineJP:
# Summary
This PR is made to prepare new `bladebit` and its new options.
Will poke around `chia plotters install bladebit` command.

# Important changes
## Windows/Ubuntu users who install GUI app from installer are now able to integrate `bladebit` installed from git source code.
Currently the chia app always searches for bladebit bundled in chia app folder(s.t. %LocalAppData%\chia-blockchain\app-1.x.x\resources\app.asar.unpacked\daemon\bladebit\bladebit.exe) and never looks into `$CHIA_ROOT/plotters/bladebit`.  
This makes it difficult to use non-bundled version of `bladebit` with CLI/GUI.
This PR will make the app first search `$CHIA_ROOT/plotters/bladebit` and if it doesn't exist then it goes for bundled `bladebit`.

## Enabled bladebit update
Currently if bladebit is already installed on `$CHIA_ROOT/plotters/bladebit`, it won't update/override existing bladebit.  
But after applying this PR, you can speficy `-o/--override` option to update/override installed bladebit.  
example: `chia plotters install bladebit -o`

## Enabled to specify git branch/tag/commit on installing bladebit from GitHub
You can speficy a commit of https://github.com/Chia-Network/bladebit by adding `-c/--commit` option like this:  
`chia plotters install bladebit --commit a65f438f1e3acefca287dbc4bc14b2b141db6ee6`

## Added `chia plotters version` command
```shell
$ chia plotters version
chiapos: 1.0.9
bladebit: 2.0.0.dev
```

## Added `chia plotters install bladebit2` command
If you already have `bladebit` installed by `chia plotters install bladebit` command, you can override it with:
`chia plotters install bladebit2 -o`

## Added `chia plotters bladebit2` command family
You can set options for `bladebit2` now.
`chia plotters bladebit2 [-r|--threads <n>] [-n|--count <number-of-plots-to-create>] [-f|--farmerkey <farmer-public-key>] [-p|--pool-key <pool-public-key>] [-c|--contract <pool-contract-address>] [-w|--warmstart] [-i|--id <plot-id>] [-v|--verbose] [-m|--nonuma] [-u|--buckets <n>] [-2 <t2 dir>] [--no-cpu-affinity] [--cache <n>] [--f1-threads <n>] [--fp-threads <n>] [--c-threads <n>] [--p2-threads <n>] [--p3-threads <n>] [-a|--alternate] -t <t1 dir>
 -d <out_dir>`

# Confirmed to work
|OS|arch|Result|
|---|---|---|
|Windows 10|x86_64|OK|
|Ubuntu 22.04|x86_64|OK|
|Amazon Linux 2|x86_64|No (cmake version available via yum is too old)|
|RHEL 8.5|x86_64|OK|
|Fedora 35|x86_64|OK|
|macOS M1|arm64|OK|

# How to try new bladebit
Currently new bladebit(version 2) will at least run on Windows, macOS, Ubuntu.
Additionally, you need NodeJS>=16 when you run the GUI.
The below is command line snippets for challengers.
```bash
### You can skip the lines below if you already have the git repo ###
cd /tmp # Any directory you like
mkdir bb2-test && cd bb2-test
git clone https://github.com/Chia-Network/chia-blockchain
cd chia-blockchain
### You can skip until here ###

### For macOS/Ubuntu users ###
git checkout cmj.bladebit
sh install.sh
. ./activate
sh install-gui.sh origin/cmj.bladebit 
chia plotters install bladebit2 # This installs bladebit2 to $CHIA_ROOT/plotters/bladebit
chia plotters version # Make sure bb2 is installed
bash start-gui.sh

### For Windows users ###
git checkout cmj.bladebit
install.ps1
. venv\Scripts\Activate.ps1
# This only "git clone" bladebit2 to $CHIA_ROOT/plotters/bladebit (Usually <UserDir>\.chia\mainnet\plotters)
chia plotters install bladebit2
# Launch VisualStudio >= 2019 on $CHIA_ROOT/plotters/bladebit folder
# On the Developer PowerShell console in Visual Studio:
#
# PS> pwd     # Check whether you are in ...\plotters\bladebit folder.
# PS> mkdir build
# PS> cd build
# PS> cmake ..
# PS> cmake --build . --target bladebit --config Release
#
chia plotters version # Make sure bb2 is installed
cd chia-blockchain-gui
npm ci
npm run build
npx lerna run dev:fast --scope=@chia/gui
```

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -788,6 +788,7 @@ class WebSocketServer:
 
         return command_args
 
+    # @TODO Do something for bladebit version 2.
     def _bladebit_plotting_command_args(self, request: Any, ignoreCount: bool) -> List[str]:
         w = request.get("w", False)  # Warm start
         m = request.get("m", False)  # Disable NUMA
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -788,6 +788,7 @@ class WebSocketServer:
     def _bladebit_plotting_command_args(self, request: Any, ignoreCount: bool) -> List[str]:
         w = request.get("w", False)  # Warm start
         m = request.get("m", False)  # Disable NUMA
+        no_cpu_affinity = request.get("no_cpu_affinity", False)
 
         command_args: List[str] = []
 
@@ -795,6 +796,8 @@ class WebSocketServer:
             command_args.append("-w")
         if m is True:
             command_args.append("-m")
+        if no_cpu_affinity is True:
+            command_args.append("--no-cpu-affinity")
 
         return command_args
 
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -815,6 +815,7 @@ class WebSocketServer:
         c_threads = request.get("c_threads")
         p2_threads = request.get("p2_threads")
         p3_threads = request.get("p3_threads")
+        alternate = request.get("alternate", False)
 
         command_args: List[str] = []
 
@@ -848,6 +849,8 @@ class WebSocketServer:
         if p3_threads:
             command_args.append("--p3-threads")
             command_args.append(str(p3_threads))
+        if alternate:
+            command_args.append("--alternate")
 
         return command_args
 
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -807,14 +807,14 @@ class WebSocketServer:
         no_cpu_affinity = request.get("no_cpu_affinity", False)
         # memo = request["memo"]
         t1 = request["t"]  # Temp directory
-        t2 = request["t2"]  # Temp2 directory
-        u = request["u"]  # Buckets
-        cache = request["cache"]
-        f1_threads = request["f1_threads"]
-        fp_threads = request["fp_threads"]
-        c_threads = request["c_threads"]
-        p2_threads = request["p2_threads"]
-        p3_threads = request["p3_threads"]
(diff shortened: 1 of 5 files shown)


ChiaMineJP:
## This PR does not assure you to make an installer which bundles bladebit2
In order to make installers which contains production ready bladebit2 binary,
bladebit 2.0.0 (without alpha/beta suffix) must be released on GitHub Repository

harold-b (commented):
CLI mapping looks good. The only thing I might add is that perhaps we'd like to add support for the -`-no-t1/t2-direct` switches since these are needed if anyone wants to use a ramdisk for their temp directories.

paninaro:
@ChiaMineJP I appreciate all the hard work that went into making this as simple as possible for installing BB2.

Going forward, I think we need to seriously consider an alternative way for our source-based users to get BB2 and madmax plotters -- ideally through prebuilt packages. That should greatly simplify the chia plotters code and avoid complicated maintenance. In general I'd like to avoid scripting the build process since there are so many places where the process can breakdown.

Also we should add some tests to cover the affected code. I'll add a separate task to track that work.

paninaro (approved):
Approving for inclusion in the beta program. I'd like to revisit the source-based install options prior to release.

harold-b (approved):

wallentx (approved):

PR #11196, opened 2022-04-16
Change did_innerpuz and fix wallets for new design

matt-o-how:
Removes unnecessary spend cases, unnecessary solution params, and unnecessary curry params, and tweaks DID and NFT wallets and wallet_state_manager


matt-o-how:
Basically this patch changes the Chialisp to offload a bunch of work to the wallet - saving a bunch of cost and simplifying the code massively

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -690,6 +710,7 @@ async def create_message_spend(
             ]
         )
         list_of_coinspends = [CoinSpend(coin, full_puzzle, fullsol)]
+        # breakpoint()
```
Quexington, line 713: Probably remove this

chia/wallet/puzzles/did_innerpuz.clvm
```diff
@@ -9,16 +9,15 @@
...
   NUM_VERIFICATIONS_REQUIRED  ; how many of the above list are required for a recovery
   SINGLETON_STRUCT  ; my singleton_struct, formerly a Truth - ((SINGLETON_MOD_HASH, (LAUNCHER_ID, LAUNCHER_PUZZLE_HASH)))
-  DID_MOD_HASH  ; DID inner puzzle treehash
   METADATA ; Customized metadata, e.g KYC info
-  mode  ; this indicates which spend mode we want. 0. Recovery mode 1. Send message mode 2. Transfer mode
-  new_amount  ; DIDs can receive payments so when we recreate ourselves sometimes we want to change our amount
-  p2_solution ; Solution of the inner P2 puzzle, only required in the create message mode and transfer mode.
-  new_inner_puzhash  ; In message mode, this should be the DID puzzle hash. In the transfer case, this should be the recipient p2 puzzle hash. In recovery mode, this will be the new wallet DID puzzle hash
+  mode  ; this indicates which spend mode we want. 0. Recovery mode 1. Run P2_PUZZLE with p2_solution
+  new_amount_or_p2_solution  ; In mode 0, maybe we want to change our amount
```
Quexington, line 14: How is new_amount secured in this scenario?  It seems like a malicious node would always change it to zero.

chia/wallet/puzzles/did_innerpuz.clvm
```diff
@@ -132,28 +113,16 @@
...
-      ; Send normal messages
-      (c (list CREATE_COIN new_inner_puzhash new_amount (list new_inner_puzhash)) (c (list AGG_SIG_ME (get_pubkey P2_PUZZLE) (sha256tree1 (list new_inner_puzhash new_amount))) (a P2_PUZZLE p2_solution)))
-      ; Transfer DID
-      (if (= (sha256tree1 recovery_list_reveal) RECOVERY_DID_LIST_HASH)
-         (c (list CREATE_COIN (calculate_did_puzzle_hash new_inner_puzhash RECOVERY_DID_LIST_HASH NUM_VERIFICATIONS_REQUIRED SINGLETON_STRUCT DID_MOD_HASH METADATA) new_amount (list new_inner_puzhash)) (c (list AGG_SIG_ME (get_pubkey P2_PUZZLE) (sha256tree1 (list new_inner_puzhash new_amount recovery_list_reveal))) (a P2_PUZZLE p2_solution)))
-         (x)
-      )
-    )
+    ; mode 1 - run P2_PUZZLE
+    (a P2_PUZZLE new_amount_or_p2_solution)
```
Quexington, line 121: P2_PUZZLE is a weird name for what is just an inner puzzle.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -605,10 +608,20 @@ async def transfer_did(self, puzzle_hash: bytes32, fee: uint64) -> TransactionRe
...
         coin = coins.pop()
 
+        new_did_puz = did_wallet_puzzles.create_innerpuz(
+            new_puzhash,
+            0,
+            0,
+            self.did_info.origin_coin.name(),
+            self.did_info.metadata,
+        )
+        p2_solution = p2_solution = self.standard_wallet.make_solution(
```
ytx1991, line 618: I think we need to reveal the recovery list in the transfer case. How doesn't the recipient know the recovery list in this approach?

chia/wallet/puzzles/did_innerpuz.clvm
ytx1991, line 121: Same as above, I think we need to reveal the recovery list in the transfer case. Better to add a hash check at higher layer, in order to saving cost and prevent the sender revealing a wrong recovery list.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -655,7 +668,10 @@ async def transfer_did(self, puzzle_hash: bytes32, fee: uint64) -> TransactionRe
 
     # The message spend can tests\wallet\rpc\test_wallet_rpc.py send messages and also change your innerpuz
     async def create_message_spend(
-        self, message_puz: Program, message_sol: Program = Program.to([]), new_innerpuzhash: Optional[bytes32] = None
+        self,
+        coin_announcements: Optional[Set[bytes]] = None,
```
ytx1991, line 672: Are we only have these 2 types of condition?

chia/wallet/did_wallet/did_wallet.py
ytx1991: We also need to modify the create_new_did_wallet_from_coin_spend(), since the curried parameters and solution are changed.

ytx1991 (commented):
Have some concerns about the transfer case.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -605,10 +608,20 @@ async def transfer_did(self, puzzle_hash: bytes32, fee: uint64) -> TransactionRe
         assert coins is not None
         coin = coins.pop()
 
+        new_did_puz = did_wallet_puzzles.create_innerpuz(
+            new_puzhash,
+            0,
+            0,
+            self.did_info.origin_coin.name(),
+            self.did_info.metadata,
```
ytx1991, line 616: You need to call did_wallet_puzzle.message_to_program() here

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -605,10 +608,20 @@ async def transfer_did(self, puzzle_hash: bytes32, fee: uint64) -> TransactionRe
         assert coins is not None
         coin = coins.pop()
 
+        new_did_puz = did_wallet_puzzles.create_innerpuz(
+            new_puzhash,
+            0,
```
ytx1991, line 613: I had a discussion about if we should transfer the recovery list to the new owners with Cameron. In most cases it should be some one transfer his DID to a new wallet, he may not want to set the recovery list again after the transfer.

chia/wallet/puzzles/did_innerpuz.clvm
matt-o-how: Ok yeah, let's just lock this down to our actual amount and do an assert_my_amount

chia/wallet/puzzles/did_innerpuz.clvm
matt-o-how: Changed to INNER_PUZZLE

chia/wallet/did_wallet/did_wallet.py
matt-o-how: We can add more later if we need to

Quexington (approved):

PR #11197, opened 2022-04-16
Offer and trade manager generalization

Quexington:
The idea here is to open the door for assets that are not CATs to be used in trade offers.  There's no change to the offer file format, but if someone is running software before a new asset type is added they cannot accept offers with that new type of asset.  New versions of software will be able to accept offers from old versions just fine.

There's also added `# ATTENTION: new wallets` comments to highlight functions/sections of code that need to be changed when adding a new type of asset for the trade manager to handle.


matt-o-how:
Looks sensible to me - are we going to leave this PR open and then add the new wallet stuff before merging? Or merge and then make another PR?

Quexington:
I think this should go into main so it can get some testing with the current offers and then modifications for new wallets will come in with those new wallets

chia/rpc/wallet_rpc_api.py
```diff
@@ -924,6 +925,17 @@ async def create_offer_for_ids(self, request):
...
         fee: uint64 = uint64(request.get("fee", 0))
         validate_only: bool = request.get("validate_only", False)
+        type_dict_str: Optional[Dict[str, str]] = request.get("type_dict", None)
+
+        # This type_dict construction is to maintain backward compatibility where everything is assumed to be a CAT
+        type_dict: Dict[str, str] = {}
+        if type_dict_str is None:
+            for key in offer:
+                if len(key) >= 32:
+                    type_dict[bytes32.from_hexstr(key)] = AssetType.CAT
```
paninaro, line 935: `bytes32`'s initializer requires exactly 32 bytes, otherwise it will raise a `ValueError`. So I think the len check should be `if len(key) == 64:`

chia/rpc/wallet_rpc_api.py
```diff
@@ -924,6 +925,17 @@ async def create_offer_for_ids(self, request):
         offer: Dict[str, int] = request["offer"]
         fee: uint64 = uint64(request.get("fee", 0))
         validate_only: bool = request.get("validate_only", False)
+        type_dict_str: Optional[Dict[str, str]] = request.get("type_dict", None)
+
+        # This type_dict construction is to maintain backward compatibility where everything is assumed to be a CAT
+        type_dict: Dict[str, str] = {}
```
paninaro, line 931: The `type_dict` type definition differs from what TradeManager's `create_offer_for_ids` expects: `Dict[bytes32, AssetType]`

chia/wallet/trading/offer.py
```diff
@@ -114,22 +121,18 @@ def get_offered_coins(self) -> Dict[Optional[bytes32], List[Coin]]:
...
+                offer_ph: bytes32 = construct_puzzle(self.type_dict[asset_id], asset_id, OFFER_MOD).get_tree_hash()
             else:
-                tail_hash = None
                 offer_ph = OFFER_MOD.get_tree_hash()
 
             # Check if the puzzle_hash matches the hypothetical `settlement_payments` puzzle hash
             if addition.puzzle_hash == offer_ph:
-                if tail_hash in offered_coins:
-                    offered_coins[tail_hash].append(addition)
+                if asset_id in offered_coins:
```
paninaro, line 132: From the code right above this, it looks like `asset_id` could be None. Is it safe to enter this block of code if `asset_id` is None (doesn't look like it...)

Should the `if addition.puzzle_hash == offer_ph` check above also check that `asset_id is not None`?

Quexington: No, offered coins is `Dict[Optional[bytes32], List[Coin]`.  `asset_id == None` is representing standard XCH

paninaro (approved):
lgtm

matt-o-how (approved):
looks good

PR #11203, opened 2022-04-18
Fix filename of latest intel dev installer

cmmarslender:
Update the latest dev installer to match the name the website is expecting


hoffmang9 (approved):
aok

PR #11208, opened 2022-04-18
daemon|util: Don't remove from list while iterating

xdustinface:
See commit messages. Initially i discovered the issue in the daemon, then i reviewed the rest of the codebase to see if its done like this in other cases, while doing that i found the one in `check_fork_next_block`.

Also see the below script missing to print `3` is the root issue this PR fixes.

```
items = [1,2,3,4]

for item in items:
    print(item)
    try:
        if item == 2:
            raise Exception
    except Exception:
        items.remove(item)
```

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -613,7 +613,7 @@ class WebSocketServer:
 
         response = create_payload("keyring_status_changed", keyring_status, "daemon", destination)
 
-        for websocket in websockets:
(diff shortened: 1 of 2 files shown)


emlowe:
Nice catch Kevin. Yea, removing items from the same list you are iterating results in skipping some list items.

paninaro (approved):
lgtm

wjblanke:
nice

PR #11226, opened 2022-04-19
Remove 3.9.11 for Windows Installer

hoffmang9:
And replace with just 3.9.x


hoffmang9:
Python 3.9.11 was pinned to confirm an OpenSSL fixed python version. Python 3.9.12 is now the standard on the GH Windows runner so we can stay with the current released 3.9.x moving forward.

cmmarslender (approved):

wjblanke (approved):
aok

PR #11237, opened 2022-04-20
Initial iteration of the chia wallet did create command

AmineKhaldi:


paninaro:
@esaung and I discussed the DID CLI a bit and agreed that it makes sense to keep the commands under the `chia wallet` umbrella, as you already have. We can revisit the question of whether to promote `did` to a top-level CLI command at a later date once we have more data, but for now it makes sense to keep wallet-centric commands grouped under `chia wallet <cmd>`

paninaro:
@esaung and I discussed the DID CLI a bit and agreed that it makes sense to keep the commands under the `chia wallet` umbrella, as you already have. We can revisit the question of whether to promote `did` to a top-level CLI command at a later date once we have more data, but for now it makes sense to keep wallet-centric commands grouped under `chia wallet <cmd>`

chia/cmds/wallet.py
```diff
@@ -360,3 +360,29 @@ def cancel_offer_cmd(wallet_rpc_port: Optional[int], fingerprint: int, id: str,
...
+@click.option(
+    "-wp",
+    "--wallet-rpc-port",
+    help="Set the port where the Wallet is hosting the RPC interface. See the rpc_port under wallet in config.yaml",
+    type=int,
+    default=None,
+)
+@click.option("-f", "--fingerprint", help="Set the fingerprint to specify which wallet to use", type=int)
+@click.option("-n", "--name", help="Set the DID wallet name", type=str)
+@click.option("-a", "--amount", help="Set the DID amount in mojos", type=int)
```
paninaro, line 380:
```suggestion
@click.option("-a", "--amount", help="Set the DID amount in mojos. Value must be an odd number.", type=int)
```

chia/cmds/wallet_funcs.py
```diff
@@ -622,3 +622,18 @@ async def execute_with_wallet(
             print(f"Exception from 'wallet' {e}")
     wallet_client.close()
     await wallet_client.await_closed()
+
+
+async def create_did_wallet(args: Dict, wallet_client: WalletRpcClient, fingerprint: int) -> None:
+    name = args["name"]
+    amount = args["amount"]
+    if amount is None:
+        amount = 1
```
paninaro, line 631: If we set a default value of 1 in the corresponding `@click.option`, we can drop this check. We might want to replace this with a check that `amount %2 == 1`.

chia/cmds/wallet_funcs.py
```diff
@@ -622,3 +622,18 @@ async def execute_with_wallet(
...
+    try:
+        response = await wallet_client.create_new_did_wallet(amount, name)
+        wallet_id = response["wallet_id"]
+        my_did = response["my_did"]
+        print(f"Successfully created a DID wallet with name {name} and id {wallet_id} on key {fingerprint}")
+        print(f"Successfully created a DID {my_did} in the newly created DID wallet")
```
paninaro, line 637: I could be mistaken, but I believe the `create_new_did_wallet()` call doesn't wait for the transaction to become confirmed on-chain. I think we'll probably want to handle this in a manner similar to `create_new_pool_wallet()`. Take a look at the `create()` command in `plotnft_funcs.py`: https://github.com/Chia-Network/chia-blockchain/blob/73e1fabb5dca522c27c16592bdc07c327c99abf6/chia/cmds/plotnft_funcs.py#L104

chia/cmds/wallet.py
```diff
@@ -360,3 +360,29 @@ def cancel_offer_cmd(wallet_rpc_port: Optional[int], fingerprint: int, id: str,
...
+
+@did_cmd.command("create", short_help="Create DID wallet")
+@click.option(
+    "-wp",
+    "--wallet-rpc-port",
+    help="Set the port where the Wallet is hosting the RPC interface. See the rpc_port under wallet in config.yaml",
+    type=int,
+    default=None,
+)
+@click.option("-f", "--fingerprint", help="Set the fingerprint to specify which wallet to use", type=int)
```
paninaro, line 378: We'll need to add an option to set a fee. See `chia plotnft create --help` for an example (using `-m/--fee`). I'm not sure if creating a DID will require two spends as it does for plotnft creation. Feel free to revisit the fee issue in a separate PR.

@ytx1991 has `create_did_wallet()` been updated to take a fee parameter, and if so, is that fee applied twice as it is for the plotnft case?

ytx1991: It wasn't but I changed it after saw this.

chia/cmds/wallet_funcs.py
AmineKhaldi: Agreed. Regarding the odd number check, we seem to be doing that in `DIDWallet::create_new_did_wallet()` so we'll get an informative exception when that's the case.

chia/cmds/wallet_funcs.py
AmineKhaldi: Should we revisit that when we merge a recent main that has the up to date version of this code?

paninaro (approved):
I had to apply the latest `main_dids` to test this branch out, but I was able to create a DID wallet from the CLI!

Issue #11238, opened 2022-04-20
[Bug] Install.sh won't install on Arch Linux


SemnoSao:
### What happened?

Clean clone of repo on Manjaro Linux.

```$ sh install.sh
Installing on Arch Linux.
warning: git-2.35.1-1 is up to date -- skipping
warning: openssl-1.1.1.n-1 is up to date -- skipping
 there is nothing to do
```

the next command on the instruction from installing from source is `. ./activate` that gives the following output:

`bash: ./activate: No such file or directory`

I suspect the problem is the install.sh is just not doing what it should 🧐

### Version

1.3.4

### What platform are you using?

Linux

### What ui mode are you using?

CLI

LuaKT:
Chia currently only supports python 3.9 and below. If you have an up-to-date Arch install you likely have python 3.10 so you need to install 3.9 from AUR.

There is also a bug in the install script where it doesn't correctly display the wrong version error.

SemnoSao:
Thankss! 
installing python39 from AUR solved the problem.

PR #11239, opened 2022-04-20
Write python version error to stderr

LuaKT:
Hi,
Currently if an incompatible Python version is found it reports an error and writes it to `stdout`. The issue is `stdout` is captured so it never reaches the user. This PR instead sends the error message to `stderr`.

Fixes the confusion like in #11238.

Thanks


wjblanke (approved):
aok

PR #11242, opened 2022-04-20
improve generate_chain

arvidn:
`./tools/generate_chain.py` is made a proper command line tool using `click`. The test constants are updated to be more similar to mainnet.
`./tools/test_full_sync.py` gets a anew feature to ingest blocks via the normal, "stay in sync" code path. this allows performance testing of both syncing and staying in sync.
This patch also adds another profiling tool, `cpu_utilization.py`, which measures CPU usage of a process as well as its child processes. This helps spot inefficiencies in our parallelization.

The changes to `generate_chain.py` is best reviewed ignoring whitespace, since a lot of code changed indentation level.

### DIFF
--- a/tools/generate_chain.py
+++ b/tools/generate_chain.py
@@ -1,17 +1,19 @@
 import cProfile
 import random
 import sqlite3
+import sys
 import time
 from contextlib import closing, contextmanager
 from pathlib import Path
-from typing import Iterator, List
+from typing import Iterator, List, Optional
 
+import click
 import zstd
 
 from chia.types.blockchain_format.coin import Coin
 from chia.types.spend_bundle import SpendBundle
 from chia.util.chia_logging import initialize_logging
-from chia.util.ints import uint64
+from chia.util.ints import uint32, uint64
 from chia.util.path import mkdir
 from tests.block_tools import create_block_tools
 from tests.util.keyring import TempKeyring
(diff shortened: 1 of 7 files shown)


tools/generate_chain.py
```diff
@@ -28,107 +30,92 @@ def enable_profiler(profile: bool) -> Iterator[None]:
...
+@click.option("--length", type=int, default=None, required=False, help="the number of blocks to generate")
+@click.option(
+    "--fill-rate",
+    type=int,
+    default=100,
+    required=False,
+    help="the transaction fill rate of blocks. Specified in percent of max block cost",
+)
+@click.option("--profile", is_flag=True, required=False, default=False, help="dump CPU profile at the end")
+@click.option(
```
trepca, line 46: minor thing, but why not have this one as a flag too? `--no-block-refs` ?

tools/generate_chain.py
```diff
@@ -1,25 +1,27 @@
 import cProfile
```
trepca, line 1: another minor thing, but some commands seem to have executable permissions while others like `generate_chain.py` don't

trepca (approved):
lgtm

`generate_chain` doesn't seem to clean the db after a run, but I assume that's intentional

tools/generate_chain.py
arvidn: it probably should have been

arvidn:
I don't know which database you're referring to that's not cleaned (or what you mean by "cleaned").
`generate_chain` produces a database as its output.

tools/generate_chain.py
arvidn: I run all the python programs with an explicit call to `python`. setting the executable bit would also have required a `#!` comment at the top I believe

PR #11249, opened 2022-04-21
Pass Recovery Info in the Transfer case

ytx1991:


paninaro (approved):
Looks fine to me, but @matt-o-how should review as well.

chia/wallet/puzzles/did_innerpuz.clvm
```diff
@@ -118,10 +118,17 @@
   ;MAIN
   (if mode
     ; mode 1 - run INNER_PUZZLE
-    (a INNER_PUZZLE my_amount_or_inner_solution)
+    (if (= mode 2)
+       ; If mode is 2, it means this is a transfer spend.
+       (if (= (sha256tree1 recovery_list_reveal) RECOVERY_DID_LIST_HASH)
```
matt-o-how, line 123: Why do we want to force this?
If the user wants to reveal it when they transfer the wallet then they can.

There are situations where they might not want to reveal the list.

ytx1991: I want to add a check in case of user reveals wrong recovery list. It was (if recovery_list_reveal), but seems it require use set 0 for recovery_list_reveal

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -203,16 +204,18 @@ async def create_new_did_wallet_from_coin_spend(
...
-        # inner_solution: Program = full_solution.rest().rest().first()
+        _, recovery_list_hash, num_verification, _, metadata = args
+        full_solution: Program = Program.from_bytes(bytes(coin_spend.solution))
+        inner_solution: Program = full_solution.rest().rest().first()
         recovery_list: List[bytes] = []
-        # for did in list(inner_solution.rest().rest().rest().rest().rest().rest().as_iter()):
-        #     recovery_list.append(did.as_python()[0])
+        backup_required: int = num_verification.as_int()
+        if recovery_list_hash != Program.to([]).get_tree_hash():
+            for did in list(inner_solution.rest().rest().rest().rest().rest().as_iter()):
```
matt-o-how, line 213: What does the wallet do if the reveal is not present?

chia/wallet/puzzles/did_innerpuz.clvm
```diff
@@ -121,7 +121,7 @@
     (a INNER_PUZZLE my_amount_or_inner_solution)
 
     ; mode 0 - recovery
-    (if (= (sha256tree1 recovery_list_reveal) RECOVERY_DID_LIST_HASH)
+    (if (all (= (sha256tree1 recovery_list_reveal) RECOVERY_DID_LIST_HASH) (> NUM_VERIFICATIONS_REQUIRED 0))
```
matt-o-how, line 124: `(> NUM_VERIFICATIONS_REQUIRED 0)` can probably be just optimised to `NUM_VERIFICATIONS_REQUIRED` - which checks that it is not `0` - though this would potentially allow negative verifications required.

Up to you whether you want to change this.

chia/wallet/did_wallet/did_wallet.py
ytx1991: It will reset recovery list to [] and the required number to 0

chia/wallet/puzzles/did_innerpuz.clvm
ytx1991: If someone set a negative number mistakenly, I think others shouldn't have permission to spend it.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -1266,3 +1249,63 @@ async def save_info(self, did_info: DIDInfo, in_transaction: bool):
...
+        for wallet in self.wallet_state_manager.wallets.values():
+            if (
+                wallet.type() == WalletType.DISTRIBUTED_ID
+                and self.did_info.origin_coin.name() == wallet.did_info.origin_coin.name()
+            ):
+                self.log.warning(f"DID {self.did_info.origin_coin} already existed, ignore the wallet creation.")
+                raise ValueError("Wallet already exists")
+
+    @staticmethod
+    def get_did_from_file(filename: str) -> DIDInfo:
```
Quexington, line 1280: I don't think that we should be reading/writing files in the wallet.  We have this ability to run remote machines and connect GUIs to them from a computer that is not actually running the wallet.  The wallet and all other services should always ship a blob across the RPC and then let the UX write it to a file if it wants.  Similarly it should take a blob from the RPC rather than reading it from a file.

Quexington (changes requested):

chia/wallet/did_wallet/did_wallet.py
ytx1991: So I think this should also be true for creating the attestments?

Quexington (approved):
lgtm

PR #11258, opened 2022-04-21
postinst/prerm scripts with the UI .deb

cmmarslender:
Adds a post install and pre remove script that creates/cleans up symlinks so that `chia` is in path by default (for CLI usage) and `/opt/chia` in both the UI and CLI version points to the same set of files

`/usr/bin/chia` -> `/usr/lib/chia-blockchain/resources/app.asar.unpacked/daemon/chia`
`/opt/chia` -> `/usr/lib/chia-blockchain/resources/app.asar.unpacked/daemon`


hoffmang9 (approved):
lgtm

justinengland (approved):
lgtm

PR #11259, opened 2022-04-21
Check observer addresses in farm rewards dialog

### LINKED ISSUE #11036: Farming rewards dialog incorrectly claims there is no private key for address
### What happened?

image

### Version

1.3.3

Windows

GUI

### Relevant log output

_No response_

emlowe:
The GUI farmer rewards check was not checking for observer keys and the code was duplicated from farmer and wallet, which is one reason this was missed.

Consolidated some code into new function called by both the wallet and farmer and added tests for observer keys.

Also added the ability to pass in the number of derivations to the RPC call - primarily this was so the tests can pass in much smaller numbers than the default (500) and so the tests can run much faster. However, this flexibility seems worth while regardless

(new PR because git hates me, but it is mutual)

Fixes #11036

### DIFF
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -4,13 +4,12 @@ import logging
 import time
 import traceback
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Tuple
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple
 
 import aiohttp
 from blspy import AugSchemeMPL, G1Element, G2Element, PrivateKey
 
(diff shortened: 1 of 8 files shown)


paninaro (approved):
lgtm

mariano54 (commented):
Thanks for doing this, it's a good cleanup.
The performance of this can also be optimized significantly, by doing the same thing as init funcs and keeping the intermediate values:
https://github.com/Chia-Network/chia-blockchain/blob/main/chia/cmds/init_funcs.py#L96

`master_sk_to_wallet_sk` repeats a lot of work (3/4 of the derivations are being done over and over)

PR #11261, opened 2022-04-22
make BlockTools generate more accurate block timestamps

arvidn:
when generating blocks, we truncate the time-per-block to whole seconds before assigning the timestamp to the next block. When the time-per-block has a fractional part (say, 18.75) the blocks will consistently come 0.75 seconds early.

This patch rounds up or down with a probability proportional to the size of the fractional part. This means blocks will come at the right time on average

### DIFF
--- a/tests/block_tools.py
+++ b/tests/block_tools.py
@@ -439,12 +439,14 @@ class BlockTools:
         previous_generator: Optional[Union[CompressorArg, List[uint32]]] = None,
         genesis_timestamp: Optional[uint64] = None,
         force_plot_id: Optional[bytes32] = None,
+        use_timestamp_residual: bool = False,
     ) -> List[FullBlock]:
         assert num_blocks > 0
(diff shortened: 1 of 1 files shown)


mariano54 (commented):
This will change the blocks that are created, and therefore it will not be compatible with the test-cache. We can update test cache.
Also we need to run the tests without the cache to see if they still pass

mariano54 (changes requested):
See above

arvidn:
ah, I see. is it the `tests/blockchain/` tests I should run?
Do we regenerate the cache periodically?

mariano54:
All the tests need to be run.
There is no automatic way to update cache right now

arvidn:
I made the residual deterministic, and optional. This means tests will keep behaving the same, and this new behavior is opt-in.

mariano54 (approved):
LGTM as long as cached blocks don't change

PR #11264, opened 2022-04-22
add tool to measure CPU utilization of a process tree

arvidn:
and plot it over time. This helps identify inefficiencies in our services. It generates output that looks like this
cpu
:

### DIFF
--- a/tools/cpu_utilization.py
+++ b/tools/cpu_utilization.py
@@ -21,7 +21,13 @@ class Counters:
     default="cpu-usage.log",
     help="the file to print CPU usage stats to",
 )
-def main(pid: int, output: str) -> None:
+@click.option(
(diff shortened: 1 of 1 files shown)


trepca:
I ran this on Mac and had issues with thread tracking, as it requires to be run as superuser and it also crashed the process I was profiling. Once I disabled thread tracking it worked well. I'd make tracking threads as an option, not default. 

Another thing that's probably Mac specific is that it's not showing CPU % correctly. This is how it looks for a process that was doing 100%. I have M1 Pro with 10 cores. 
cpu

arvidn:
disabling thread collection by default makes this work fine for me on the M1

trepca (approved):

PR #11268, opened 2022-04-22
Needed for DID CLI: Add /did_set_wallet_name, /did_get_wallet_name to RPC and set_name(), get_name() to DIDWallet

AmineKhaldi:
Needed for `chia did wallet name` and `chia did wallet set_name` commands:
* `RPC`: Add `/did_set_wallet_name` and `/did_get_wallet_name`.
* `DIDWallet`: `Add set_name()` and `get_name()`.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -107,6 +107,8 @@ class WalletRpcApi:
             "/cancel_offer": self.cancel_offer,
             "/get_cat_list": self.get_cat_list,
             # DID Wallet
+            "/did_set_wallet_name": self.did_set_wallet_name,
+            "/did_get_wallet_name": self.did_get_wallet_name,
             "/did_update_recovery_ids": self.did_update_recovery_ids,
             "/did_update_metadata": self.did_update_metadata,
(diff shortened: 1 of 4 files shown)


paninaro:
I think we can drop `did_get_wallet_name` and `get_name()`. The wallet name should be exposed through the `get_wallets` RPC via `get_all_wallet_info_entries()` in `wallet_user_store.py`.

AmineKhaldi:
> 

Hmm, I saw we have the same notion (but for cats right now) in the `make_offer` command, and I'm not sure how would that logic be generalized and whether it would leverage this function to get the DID name similarly. What do you see?

paninaro:
> > 
> 
> Hmm, I saw we have the same notion (but for cats right now) in the `make_offer` command, and I'm not sure how would that logic be generalized and whether it would leverage this function to get the DID name similarly. What do you see?

I see. Good point about `make_offer` using it. Ok I suppose if there's a caller (or will be a caller) then we should keep it. I still think we don't need a dedicated CLI command for getting the name, but otherwise looks good.

paninaro (approved):

PR #11271, opened 2022-04-22
Additional metrics

cmmarslender:
Adds additional `state_changed` events for the farmer and harvester. Also adds `validation_time` and `pre_validation_time` to the existing `block` event on the full node.

Finally, there were a few instances of `def _set_state_changed_callback` being defined in `{service}_api` files, but this is never actually used. The only place `_set_state_changed_callback` is called is in the [rpc_server](https://github.com/Chia-Network/chia-blockchain/blob/25ab0c90cb34cd048463082801c3cc26bfac389a/chia/rpc/rpc_server.py#L318). It actually calls ` rpc_server.rpc_api.service._set_state_changed_callback(...)`, calling the function on the service, not the API itself, so the function definition on the api and service was unnecessary and confusing, when trying to reason about the code.

### DIFF
--- a/chia/farmer/farmer_api.py
+++ b/chia/farmer/farmer_api.py
@@ -1,6 +1,6 @@
 import json
 import time
-from typing import Any, Callable, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import aiohttp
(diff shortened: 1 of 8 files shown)


mariano54 (approved):

wjblanke (approved):
aok

PR #11272, opened 2022-04-22
Remove multidict from setup.py for python 3.10

emlowe:
Removed multidict from setup.py and allow it to get pulled in from aiohttp. This will pull in the current version which is 6.0.2.

Tested with Avast One - no issues

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup
 
 dependencies = [
-    "multidict==5.1.0",  # Avoid 5.2.0 due to Avast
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,7 +1,7 @@
 from setuptools import setup
 
 dependencies = [
-    "multidict==5.1.0",  # Avoid 5.2.0 due to Avast
+    "multidict==6.0.2",  # Avoid 5.2.0 due to Avast
```
altendky, line 4: We don't use this directly so we can just drop it.  `aiohttp` will bring it in for us.

```suggestion
```

altendky (approved):

PR #11277, opened 2022-04-23
also m1 and arm64 for wheel checks

altendky:
Draft for:
- [x] Correct Python versions on m1
- [x] ~~pycryptodome wheels for m1~~ (dealing with that separately, excepting for now)

### DIFF
--- a/.github/workflows/check_wheel_availability.yaml
+++ b/.github/workflows/check_wheel_availability.yaml
@@ -44,4 +44,4 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Check Wheel Availability
-      run: python build_scripts/check_dependency_artifacts.py
+      run: python3 build_scripts/check_dependency_artifacts.py
--- a/.github/workflows/check_wheel_availability.yaml
+++ b/.github/workflows/check_wheel_availability.yaml
@@ -17,31 +17,40 @@ concurrency:
 
 jobs:
   check_dependency_artifacts:
-    name: Check Dependency Artifacts - ${{ matrix.os.name }} ${{ matrix.python-version }}
-    runs-on: ${{ matrix.os.runs-on }}
-    container: ${{ format(matrix.os.container_format, matrix.python-version) }}
+    name: Check Dependency Artifacts - ${{ matrix.os.name }} ${{ matrix.arch.name }} ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/check_wheel_availability.yaml
```diff
@@ -19,20 +19,32 @@ jobs:
...
       fail-fast: false
       matrix:
         python-version: ['3.7', '3.8', '3.9']
-        os: [macOS-latest, ubuntu-latest]
+        os: [macos-latest, m1, ubuntu-latest, arm64]
+        exclude:
+          - os: macos-latest
+            python-version: 3.7
+          - os: m1
+            python-version: 3.7
```
hoffmang9, line 32: Shouldn't this be 3.8?

hoffmang9: I should read "exclude" when it's exclude and not include...

altendky: This is an exclude of 3.7 from the matrix.  In the 3.10 PR this should remain but also 3.8 should be added to the exclusions. 
 Admittedly exclude vs. include is a bit of an option now with just two platforms but when we add Windows using the exclude option will make more sense.  With a bit more information in the matrix this could be a single exclude rather  than two.

.github/workflows/check_wheel_availability.yaml
```diff
@@ -19,20 +19,32 @@ jobs:
   check_dependency_artifacts:
     name: Check Dependency Artifacts
     runs-on: ${{ matrix.os }}
+    container: ${{ matrix.os == 'arm64' && format('docker://python:{0}', matrix.python-version) || '' }}
     strategy:
       fail-fast: false
       matrix:
         python-version: ['3.7', '3.8', '3.9']
-        os: [macOS-latest, ubuntu-latest]
+        os: [macos-latest, m1, ubuntu-latest, arm64]
```
cmmarslender, line 27:
```suggestion
        os: [macos-latest, m1, ubuntu-latest, [Linux, ARM64]]
```
So that this doesn't end up landing on the native m1 runners

cmmarslender (changes requested):

cmmarslender (dismissed):
lgtm

cmmarslender (approved):
lgtm

PR #11279, opened 2022-04-23
Treat all Ubuntu the same in install.sh with python3-*

altendky:


hoffmang9:
There was an Ubuntu/Debian flavor that needed that other math minimum too. It may have moved forward since but I don't remember which distro it was.

altendky:
See https://github.com/Chia-Network/chia-blockchain/pull/11288 for an alternative that simplifies the code but is intended to not change behavior at all.

altendky:
> There was an Ubuntu/Debian flavor that needed that other math minimum too. It may have moved forward since but I don't remember which distro it was.

Linux Mint 21 provides Python 3.10 for `python3` which would be trouble here until we get 3.10 support done.  Adding Linux Mint to the installer test matrix to catch things like this.  https://github.com/Chia-Network/chia-blockchain/pull/11295.

PR #11307, opened 2022-04-25
stop installing distutils

altendky:
Draft for:
- [ ] Just exploring now
- [ ] Is this better than https://github.com/Chia-Network/chia-blockchain/pull/11306?

PR #11306, opened 2022-04-25
request corresponding distutils in install.sh

altendky:
Draft for:
- [ ] Just exploring now
- [ ] Is this better than https://github.com/Chia-Network/chia-blockchain/pull/11307?

PR #11308, opened 2022-04-25
Only Ubuntu 18 needs distutils

altendky:
Following the experiments in https://github.com/Chia-Network/chia-blockchain/pull/11306 and https://github.com/Chia-Network/chia-blockchain/pull/11307 it seems that there's some oddity in using the secondary Python version in Ubuntu 18.04 that still requires distutils to create a venv.  The other Ubuntu versions seem to be ok.


cmmarslender (dismissed):

wjblanke (approved):
aok

PR #11318, opened 2022-04-26
Optimize BLS verification when public key is repeated

mariano54:
Many times, one address pay out to many other addresses, so the same public key is used repeatedly (pool payouts, dust etc).
Before we were performing the subgroup check many times. 

In this change we optimize with a cache to only convert from bytes -> G1Element once per pk. It obtains a 10% validation speedup  with 1000 public keys.

### DIFF
--- a/chia/util/cached_bls.py
+++ b/chia/util/cached_bls.py
@@ -23,17 +23,19 @@ def get_pairings(cache: LRUCache, pks: List[bytes48], msgs: Sequence[bytes], for
             if missing_count > len(pks) // 2:
                 return []
         pairings.append(pairing)
+
+    # G1Element.from_bytes can be expensive due to subgroup check, so we avoid recomputing it with this cache
     pk_bytes_to_g1: Dict[bytes48, G1Element] = {}
     for i, pairing in enumerate(pairings):
         if pairing is None:
             aug_msg = pks[i] + msgs[i]
             aug_hash: G2Element = AugSchemeMPL.g2_from_message(aug_msg)
 
-            if pks[i] in pk_bytes_to_g1:
-                pk_parsed: G1Element = pk_bytes_to_g1[pks[i]]
-            else:
+            pk_parsed: Optional[G1Element] = pk_bytes_to_g1.get(pks[i])
+            if pk_parsed is None:
                 pk_parsed = G1Element.from_bytes(pks[i])
                 pk_bytes_to_g1[pks[i]] = pk_parsed
+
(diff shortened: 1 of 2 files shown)


chia/util/cached_bls.py
```diff
@@ -23,12 +23,18 @@ def get_pairings(cache: LRUCache, pks: List[bytes48], msgs: Sequence[bytes], for
...
         pairings.append(pairing)
-
+    pk_bytes_to_g1: Dict[bytes48, G1Element] = {}
     for i, pairing in enumerate(pairings):
         if pairing is None:
             aug_msg = pks[i] + msgs[i]
             aug_hash: G2Element = AugSchemeMPL.g2_from_message(aug_msg)
-            pairing = G1Element.from_bytes(pks[i]).pair(aug_hash)
+
+            if pks[i] in pk_bytes_to_g1:
```
arvidn, line 32: this seems like something that might happen often enough to change this to a single lookup, using `pk_bytes_to_g1.get()`

arvidn (commented):
looks good. Is this something you expect to happen in the wild, frequent enough to outweigh the cost of the cache?

arvidn:
also, I don't think it's obvious to a casual reader that the `from_bytes()` call is expensive, and performs the subgroup check. I think that's worth pointing out in a comment, to justify the cache.

chia/util/cached_bls.py
mariano54: Does the performance differ at all? Is this just a style comment?

mariano54:
> also, I don't think it's obvious to a casual reader that the `from_bytes()` call is expensive, and performs the subgroup check. I think that's worth pointing out in a comment, to justify the cache.

Makes sense. Although I do plan on changing this in the future

chia/util/cached_bls.py
arvidn: I would not assume that python can merge the two lookups into one. but maybe I'm just pessemistic.
The point is that if `get()` returns `not None` you can use that value rather than looking it up again.

Right now you do two lookups, first using `in` and then using `[]`.

arvidn: I also didn't realize at first this is just a cache for this one call, meaning it won't have more than 2000 elements in it, so maybe it won't ever matter.

mariano54:
It is something that I expect to happen very often. For example the dust storm, in this case it's one address sending to many

arvidn (approved):

PR #11319, opened 2022-04-26
Atari startup fix.

fchirica:


chia/data_layer/data_layer.py
```diff
@@ -352,6 +349,20 @@ async def get_kv_diff(self, tree_id: bytes32, hash_1: bytes32, hash_2: bytes32)
...
+        success = False
+        while not self._shut_down and not success:
+            success = False
+            async with self.subscription_lock:
+                try:
+                    subscriptions = await self.data_store.get_subscriptions()
+                    for subscription in subscriptions:
+                        await self.wallet_rpc.dl_track_new(subscription.tree_id)
+                    success = True
+                except Exception as e:
```
emlowe, line 360: maybe be more specific and just catch aiohttp.ClientError ?

chia/data_layer/data_layer.py
```diff
@@ -71,9 +71,6 @@ async def _start(self) -> bool:
         self.subscription_lock: asyncio.Lock = asyncio.Lock()
         if self.config.get("run_server", False):
             await self.data_layer_server.start()
-        subscriptions = await self.get_subscriptions()
-        for subscription in subscriptions:
-            await self.wallet_rpc.dl_track_new(subscription.tree_id)
         self.periodically_fetch_data_task: asyncio.Task[Any] = asyncio.create_task(self.periodically_fetch_data())
```
emlowe, line 74: Do we need to do anything on shutdown if this task is still pending (ie start data but never start wallet and then stop data)

fchirica: no, it'll just stay in the first loop until shutdown (as it never received `success=True`), then ignore the second loop (as it received `self._shut_down = True`). also the task is previously cancelled in `_await_closed` on shutdown.

emlowe (dismissed):

chia/data_layer/data_layer.py
```diff
@@ -352,6 +349,19 @@ async def get_kv_diff(self, tree_id: bytes32, hash_1: bytes32, hash_2: bytes32)
...
+            async with self.subscription_lock:
+                try:
+                    subscriptions = await self.data_store.get_subscriptions()
+                    for subscription in subscriptions:
+                        await self.wallet_rpc.dl_track_new(subscription.tree_id)
+                    success = True
+                except aiohttp.client_exceptions.ClientConnectorError:
+                    self.log.error("Cannot connect to the wallet. Retrying in 3s.")
+            if not success:
+                await asyncio.sleep(3)
```
altendky, line 363: Three seconds is a pretty long time to delay shutdown, I think we can check more often.  Given that the connection failure is happening regularly on startup I think this should be a warning instead of an error.

```suggestion
        while not self._shut_down:
            async with self.subscription_lock:
                try:
                    subscriptions = await self.data_store.get_subscriptions()
                    for subscription in subscriptions:
                        await self.wallet_rpc.dl_track_new(subscription.tree_id)
                    break
                except aiohttp.client_exceptions.ClientConnectorError:
                    pass

            self.log.warning("Cannot connect to the wallet. Retrying in 3s.")

            delay_until = time.monotonic() + 3
            while time.monotonic() < delay_until:
                if self._shut_down:
                    break
                await asyncio.sleep(0.1)
```

altendky (changes requested):

altendky (approved):

PR #11324, opened 2022-04-26
Fix race condition, blockchain can change between two calls to get_peak

mariano54:
The prev block was inconsistent with the peak used to build the bundle from the mempool. If the peak changes in between those two calls (line 743 and 784), then we can create an invalid block, leading to failed validation and fallback to empty block.

This is pretty hard to test because we would need to add a block right in between those two points. I have provided this fix to some large farmers to test it. It's not clear if it really fixes is, but it's a zero risk change.

### DIFF
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -781,7 +781,7 @@ class FullNodeAPI:
             def get_pool_sig(_1, _2) -> Optional[G2Element]:
                 return request.pool_signature
 
-            prev_b: Optional[BlockRecord] = self.full_node.blockchain.get_peak()
+            prev_b: Optional[BlockRecord] = peak
 
             # Finds the previous block from the signage point, ensuring that the reward chain VDF is correct
             if prev_b is not None:


arvidn:
do I understand the situation correctly?
this would happen if:

1. some other farmer (possibly with a lower quality proof-of-space) submits a block to the network
2. We receive (and accept) it while we're building our block
3. our block has a higher quality proof-of-space, and will replace the one we just received.
4. because of this race, we'll fail to build our block and submit an empty block, which still wins because of our proof-of-space

wjblanke:
Mariano are you still pondering this?

arvidn:
I didn't mean to hold this up, I was just curious

mariano54:
> do I understand the situation correctly? this would happen if:
> 
>     1. some other farmer (possibly with a lower quality proof-of-space) submits a block to the network
> 
>     2. We receive (and accept) it while we're building our block
> 
>     3. our block has a higher quality proof-of-space, and will replace the one we just received.
> 
>     4. because of this race, we'll fail to build our block and submit an empty block, which still wins because of our proof-of-space

It's slightly different. It's not a competition, both blocks can make it into the blockchain at the same time, and the second one will not be a TX block so the transactions from it will get dropped.

If they both specify height 6 as the prev, then the foreign block will get added as 7, and our block will get added as 8 (the prev_hash pointer is not signed so it can be replaced). 

The issue is if height is 6, we choose mempool TXs as if our block is 7, but then set the prev block as 7 in our block, but we have already chosen all the transactions from the mempool as if 7 didn't exist (since height was 6 right before) . Then that block will not validate because it's transactions are inconsistent with it's prev height.

mariano54:
> Mariano are you still pondering this?

This is completely safe to merge. I'm not sure if it fixes anything, since the line that was replaced, was not async. So I'm not sure there is much time for a race condition to happen here. Right after the `async with lock` we make this call. Is there enough time for another task to grab the lock, and do everything required when adding the block? probably not

mariano54:
Still I recommend merging this

arvidn (approved):

PR #11328, opened 2022-04-27
lookup coins in large batches

arvidn:
rather than one at a time, in block_body_validation. This gives a modest speedup during sync.

7.09% -> 2.24%

EDIT: although, this profile only measures CPU time, not iowait. It's not clear that this saves any wall-clock time.

benchmarked on raspberry PI:

with this patch:
slow-batch-00448

prior to this patch:
slow-batch-00448

### DIFF
--- a/chia/consensus/block_body_validation.py
+++ b/chia/consensus/block_body_validation.py
@@ -382,7 +382,7 @@ async def validate_block_body(
     if len(unspent_records) != len(removals_from_db):
         # some coins could not be found in the DB. We need to find out which
         # ones and look for them in additions_since_fork
-        found = set([u.name for u in unspent_records])
+        found: Set[bytes32] = set([u.name for u in unspent_records])
         for rem in removals_from_db:
             if rem in found:
                 continue
--- a/chia/consensus/block_body_validation.py
+++ b/chia/consensus/block_body_validation.py
@@ -340,6 +340,9 @@ async def validate_block_body(
             assert curr is not None
 
     removal_coin_records: Dict[bytes32, CoinRecord] = {}
+    # the removed coins we need to look up from the DB
+    # i.e. all non-ephemeral coins
+    removals_from_db: List[bytes32] = []
     for rem in removals:
         if rem in additions_dic:
             # Ephemeral coin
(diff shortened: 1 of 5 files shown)


chia/full_node/coin_store.py
```diff
@@ -179,6 +179,36 @@ async def get_coin_record(self, coin_name: bytes32) -> Optional[CoinRecord]:
...
+                names_db: Tuple[Any, ...]
+                if self.db_wrapper.db_version == 2:
+                    names_db = tuple(names_chunk)
+                else:
+                    names_db = tuple([n.hex() for n in names_chunk])
+                cursors.append(
+                    conn.execute(
+                        f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
+                        f"coin_parent, amount, timestamp FROM coin_record "
+                        f'WHERE coin_name in ({"?," * (len(names_db) - 1)}?) ',
```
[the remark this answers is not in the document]
arvidn: the -1 is to compensate for the last `?` at the end, that doesn't have a `,` after it. I agree that `join()` is a more elegant solution to that

arvidn:
I also just realized that the new function I added, `get_coin_records()` doesn't interact with the cache. I noticed in the profile that there was a new call to `get_coin_record()` showing up because of that. I'm fixing that now as well.

chia/full_node/coin_store.py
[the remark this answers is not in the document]
arvidn: how about:

```
",".join(["?"] * len(names_db))
```
?

chia/consensus/block_body_validation.py
```diff
@@ -353,35 +356,53 @@ async def validate_block_body(
...
                 return Err.DOUBLE_SPEND_IN_FORK, None
+            removals_from_db.append(rem)
+
+    unspent_records = await coin_store.get_coin_records(removals_from_db)
+
+    # some coin spends we need to ensure exist in the fork branch. Both coins we
+    # can't find in the DB, but also coins that were spent after the fork point
+    look_in_fork: List[bytes32] = []
+    for unspent in unspent_records:
+        if unspent.confirmed_block_index <= fork_h:
```
arvidn, line 372: I'm a little bit tempted to move this condition into the database query. I think that might simplify things a little bit

chia/consensus/blockchain.py
```diff
@@ -410,9 +410,7 @@ async def _reconsider_peak(
                 tx_additions,
                 tx_removals,
             )
-            removed_rec: List[Optional[CoinRecord]] = [
-                await self.coin_store.get_coin_record(name) for name in tx_removals
-            ]
+            removed_rec: List[CoinRecord] = await self.coin_store.get_coin_records(tx_removals)
```
arvidn, line 413: there's an issue here. The assert on line 422 need to be removed and instead an assert here needs to be:

```
assert len(removed_rec) == len(tx_removals)
```

chia/full_node/coin_store.py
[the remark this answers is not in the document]
richardkiss: Oh I see, I didn't notice the `?` after the `}`.

I guess I actually meant `", ".join("?" for _ in range(len(names_db)))` (ie. I forgot `range`).

Your way works too of course. I've never seen it written that way. Your way knows the string length in advance, which means python could conceivably avoid doing `realloc` (although I don't know for a fact it does do that) at the cost of allocating memory for the list. My way likely has to allocate memory for the iterator, since its current state has to be saved somewhere. In the grand scheme of things, it's all just nit-picky and pretty irrelevant since it's not a very tight loop.

chia/full_node/coin_store.py
```diff
@@ -179,6 +179,50 @@ async def get_coin_record(self, coin_name: bytes32) -> Optional[CoinRecord]:
...
+                    conn.execute(
+                        f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
+                        f"coin_parent, amount, timestamp FROM coin_record "
+                        f'WHERE coin_name in ({",".join(["?"] * len(names_db))}) ',
+                        names_db,
+                    )
+                )
+
+            for cur_handle in cursors:
+                cursor = await cur_handle
```
mariano54, line 217: The intent is to do them all at a time, not concurrently right?

chia/consensus/block_body_validation.py
```diff
@@ -353,35 +356,53 @@ async def validate_block_body(
...
+                # Check for coins spent in an ancestor block
+                return Err.DOUBLE_SPEND, None
+            removal_coin_records[unspent.name] = unspent
+        else:
+            look_in_fork.append(unspent.name)
+
+    if len(unspent_records) != len(removals_from_db):
+        # some coins could not be found in the DB. We need to find out which
+        # ones and look for them in additions_since_fork
+        found = set([u.name for u in unspent_records])
```
mariano54, line 385: type annotation

chia/consensus/block_body_validation.py
```diff
@@ -353,35 +356,53 @@ async def validate_block_body(
...
+        for rem in removals_from_db:
```
mariano54, line 386: do we have test coverage of all these cases? I'm scared one of these cases might be missing in tests. IIRC i tested all of the cases before, but we might want to check coverage here just in case

mariano54 (commented):
Seems fine to me, but it is an important piece of code, so I would check test coverage

chia/full_node/coin_store.py
arvidn: I was assuming that once we issued the `execute()` call, the queries would be issued. fetching the responses is fine to do serially. it's all serially executed in the aiosqlite thread anyway.

Maybe I'm misunderstanding you. The list of coins is chunked up into the largest sqlite batch. But there may be more than one batch, so those are issued before waiting for the results.

mariano54: I think the coroutine does not start until you call await. If you want to start it sooner, you can do `asyncio.create_task(cur_handle)`.

arvidn: aiosqlite submits the queries to a thread. each `Connection` object is a thread and `execute()` submits the query to the queue of queries handled by the thread. It's a bit unclear why `execute()` itself is awaitable though. I would think the only true waiting to happen would be when asking for the results. Perhaps the queue has a size limit, in which case just submitting a job might block.

yeah, maybe it would be better to await the calls to `execute()` right away, but to collect all the cursors to let those work in parallel.

mariano54: Execute is a coroutine. I believe coroutiens are not started until you await them

chia/consensus/block_body_validation.py
arvidn: all the cases in my patch are exercised by the tests. We can't necessarily conclude that all cases that we need to handle are covered by the tests though.

I count these different cases:

1. We spend a coin that was confirmed before or at the fork height (this is the normal, non-reorg case too)
2. We spend an ephemeral coin
3. We attempt to spend a coin that was spent in the fork we're considering (i.e. after the fork height)
4. We attempt to spend a coin that was spent before the fork height (i.e. a normal double spend, not reorg case)
5. We spend a coin that was confirmed in the fork we're considering (i.e. after the fork height. It won't be in the coin_store)
6. We attempt to spend a coin that doesn't exist in the main chain nor in the fork (this is `UNKNOWN_UNSPENT`)

case (2) is very well covered by the ephemeral coin test cases I added a while back.

We have these tests:
* [test_conditions](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L1758) (covers 1, 2)
* [double_spend_in_reorg](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L2653) (covers 1, 2, 3, 6)
* [minting_coin](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L2745) (covers 1)
* [reorg_flip_flip](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L3251) (covers 5)
* [chain_failed_rollback](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L3193) (covers 1)
* [reorg_stale_fork_height](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L3193) (covers 1, 5)
* [reorg_new_ref](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L3066) (covers 1)
* [get_header_blocks_in_range_tx_filter](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L3019) (covers 1)
* [reorg_transaction](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L2974) (covers 1)
* [double_spent_in_coin_store](https://github.com/Chia-Network/chia-blockchain/blob/main/tests/blockchain/test_blockchain.py#L2619) (covers 1, 4)

chia/full_node/coin_store.py
arvidn: yeah, I'm a bit disappointed looking closer at this. I expected `execute()` to submit the query and return a future, but it actually submits the job and *waits* to hear back from the other thread. There doesn't seem to be a way to submit multiple jobs back-to-back without synchronizing with the thread in between (other than `executemany()`, but that can't be used for `SELECT`).

mariano54 (approved):

PR #11329, opened 2022-04-27
add daemon entry point

altendky:


paninaro (dismissed):

arvidn:
I think @richard should see this

altendky:
Bill asked if the arg is needed at all on `main()`.  I'll review if we use it anywhere.  Probably not.

richardkiss (dismissed):
What's the motivation for this patch?

altendky:
Consistency which was exposed to be lacking when I made a set of systemd files for each of our services.  Sorry, I thought I had made a brief explanation in the OP, but clearly I did not.  `:|`

https://github.com/altendky/farm/blob/e21a86c4e1a101f3899ac2214c7f88df21eae157/systemd/chia_daemon.service#L8
```ini
ExecStart=/farm/chia-blockchain/venv/bin/python -m chia.daemon.server
```

https://github.com/altendky/farm/blob/e21a86c4e1a101f3899ac2214c7f88df21eae157/systemd/chia_node.service#L8
```ini
ExecStart=/farm/chia-blockchain/venv/bin/chia_full_node
```

richardkiss:
Ah, cool. When I created the daemon, the intent was just to have a long-running process to manage the other processes around (with the option to possibly even collapse them all into one process, although that never materialized). If that were still the case, `systemd` could completely replace it.

However, somewhere along the way, the daemon became responsible for forwarding messages for some reason unknown to me using a mechanism also unknown to me. So without reviewing that decision, it probably cannot be dispensed with even if systemd is around, which is rather unfortunate. It might be worth better understanding this though and revisiting, which could allow us to delegate start-up and monitoring to systemd (or similar `init`-type things).

altendky:
systemd seems to be working ok with the direct service startup.  Though, it is pretty confusing for people to get started since they do stuff like have systemd execute `chia start wallet` which creates a process that may or may not start another process for the daemon and then asks the daemon to start the service.  Both that extra layer and the conditionality of it make it hard/impossible to launch and monitor properly that way.  But yes, I would expect for 'serious' systems that regular system service handling would be the 'correct' approach.  I know there are some caveats around this and key security.

paninaro (approved):
lgtm

PR #11334, opened 2022-04-27
Initial iteration of the chia wallet did set_name command

AmineKhaldi:


paninaro:
I did a quick test, merging this branch with main_dids (and resolving conflicts). I was able to rename an existing DID wallet, but I found that I could also rename other non-DID wallets. We should add a check somewhere to ensure that the wallet type is a DID wallet before proceeding.

AmineKhaldi:
> I did a quick test, merging this branch with main_dids (and resolving conflicts). I was able to rename an existing DID wallet, but I found that I could also rename other non-DID wallets. We should add a check somewhere to ensure that the wallet type is a DID wallet before proceeding.

I created PR #11391 for this, thanks.

paninaro (approved):
LGTM

PR #11338, opened 2022-04-27
Fix issue with missing coins

mariano54:
For very large wallets, `register_interest_in_puzzle_hash` can take a long time, due to a large amount of data being downloaded. If it timed out, it would return [], and syncing would continue. It would quickly succeed syncing, and then set the height in long_sync, to a really high height. Subsequent syncs would start syncing from that very high height.

Eventually, a secondary sync might happen and the missing coins should show up anyway, without this change. If you have so many coins that it always times out, you would be in trouble. The  timeout was also changed from 60 -> 300 to reduce the chance of this happening.


wjblanke (dismissed):
aok

wjblanke (approved):
aok

PR #11342, opened 2022-04-28
farmer|rpc: Some changes to `get_harvesters{_summary}` RPC endpoints

xdustinface:
This is part of the farm GUI enhancements and changes the output of a harvester entry in the `get_harvesters` farmer RPC as well as in the output of the `get_harvesters_summary` endpoint introduced in #11245.

Changes:
- Adds the new field `syncing` which contains either `null` if there is no sync in progress or the following object during a sync event
```
{
  "initial": True if the this is the first sync from the harvester to the farmer and False if its just an repeated/diff sync
  "plot_files_processed": Number of plots already processed
  "plot_files_total": Number of all plot files which will be processed in this sync event
}
```
- Adds the new filed `total_plot_size` which contains the sum of file sizes for all valid plots of the harvester.
- Always include the `last_sync_time` field and set it to `null` if the receiver is not yet synced. The field was prior to this patch just not included if the receiver wasn't synced.

Based on #11267

### DIFF
--- a/chia/plot_sync/receiver.py
+++ b/chia/plot_sync/receiver.py
@@ -40,7 +40,7 @@ class Sync:
     plots_processed: uint32 = uint32(0)
     plots_total: uint32 = uint32(0)
     delta: Delta = field(default_factory=Delta)
-    time_done: float = 0
+    time_done: Optional[float] = None
 
     def bump_next_message_id(self) -> None:
         self.next_message_id = uint64(self.next_message_id + 1)
@@ -289,7 +289,7 @@ class Receiver:
         await self._process(self._sync_done, ProtocolMessageTypes.plot_sync_done, data)
 
     def to_dict(self, counts_only: bool = False) -> Dict[str, Any]:
-        result: Dict[str, Any] = {
+        return {
             "connection": {
                 "node_id": self._connection.peer_node_id,
                 "host": self._connection.peer_host,
@@ -299,7 +299,5 @@ class Receiver:
             "failed_to_open_filenames": get_list_or_len(self._invalid, counts_only),
             "no_key_filenames": get_list_or_len(self._keys_missing, counts_only),
             "duplicates": get_list_or_len(self._duplicates, counts_only),
+            "last_sync_time": self._last_sync.time_done,
         }
-        if self._last_sync.time_done != 0:
-            result["last_sync_time"] = self._last_sync.time_done
-        return result
--- a/chia/plot_sync/receiver.py
+++ b/chia/plot_sync/receiver.py
@@ -302,6 +302,13 @@ class Receiver:
         await self._process(self._sync_done, ProtocolMessageTypes.plot_sync_done, data)
 
     def to_dict(self, counts_only: bool = False) -> Dict[str, Any]:
+        syncing = None
+        if self._current_sync.in_progress():
+            syncing = {
+                "initial": self.initial_sync(),
+                "plot_files_processed": self._current_sync.plots_processed,
+                "plot_files_total": self._current_sync.plots_total,
+            }
         return {
             "connection": {
                 "node_id": self._connection.peer_node_id,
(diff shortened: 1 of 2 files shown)


chia/plot_sync/receiver.py
```diff
@@ -299,7 +312,13 @@ def to_dict(self, counts_only: bool = False) -> Dict[str, Any]:
...
             "no_key_filenames": get_list_or_len(self._keys_missing, counts_only),
             "duplicates": get_list_or_len(self._duplicates, counts_only),
+            "total_plot_size": self._total_plot_size,
+            "syncing": {
+                "initial": self.initial_sync(),
+                "plot_files_processed": self.current_sync().plots_processed,
+                "plot_files_total": self.current_sync().plots_total,
+            }
+            if self._current_sync.in_progress()
+            else None,
```
altendky, line 322: This reads like three parameters.  I would calculate this outside and add the value in.

tests/plot_sync/test_receiver.py
```diff
@@ -247,10 +251,24 @@ async def test_to_dict(counts_only: bool) -> None:
...
+        if state != State.idle and state != State.removed and state != State.done:
+            expected_plot_files_processed += len(sync_steps[state].args[0])
+
+        sync_data = receiver.to_dict()["syncing"]
+        if state == State.done:
+            assert sync_data is None
+        else:
+            assert sync_data["initial"]
+            assert sync_data["plot_files_processed"] == expected_plot_files_processed
+            assert sync_data["plot_files_total"] == expected_plot_files_total
```
altendky, line 270: One assertion with all the information when there's an error, verifies that `"initial"` is `True` not `[1, 2, 3]` or `"abc"`, checks there aren't extra keys.

```suggestion
        if state == State.done:
            expected_sync_data = None
        else:
            expected_sync_data = {
                "initial": True,
                "plot_files_processed": expected_plot_files_processed,
                "plot_files_total": expected_plot_files_total,
            }

        assert sync_data == expected_sync_data
```

tests/plot_sync/test_receiver.py
```diff
@@ -259,7 +277,23 @@ async def test_to_dict(counts_only: bool) -> None:
...
+            PlotSyncIdentifier(uint64(time.time()), uint64(receiver.last_sync().sync_id + 1), uint64(0)),
+            False,
+            receiver.last_sync().sync_id,
+            uint32(1),
+        )
+    )
+    sync_data = receiver.to_dict()["syncing"]
+    assert not sync_data["initial"]
+    assert sync_data["plot_files_processed"] == 0
+    assert sync_data["plot_files_total"] == 1
```
altendky, line 296: See above.

chia/plot_sync/receiver.py
```diff
@@ -88,6 +94,9 @@ def current_sync(self) -> Sync:
     def last_sync(self) -> Sync:
         return self._last_sync
 
+    def initial_sync(self) -> bool:
+        return self._last_sync.sync_id == 0
```
altendky, line 98: Already brought up in chat, but for the record here, could we just have `self._last_sync` be `None` instead of having to check an attribute for a special value?

chia/plot_sync/receiver.py
```diff
@@ -299,7 +312,13 @@ def to_dict(self, counts_only: bool = False) -> Dict[str, Any]:
             "failed_to_open_filenames": get_list_or_len(self._invalid, counts_only),
             "no_key_filenames": get_list_or_len(self._keys_missing, counts_only),
             "duplicates": get_list_or_len(self._duplicates, counts_only),
+            "total_plot_size": self._total_plot_size,
+            "syncing": {
+                "initial": self.initial_sync(),
+                "plot_files_processed": self.current_sync().plots_processed,
+                "plot_files_total": self.current_sync().plots_total,
```
altendky, line 319: This is `self`, you don't need the accessors?

altendky (dismissed):
Suggestions can be addressed separately.

tests/plot_sync/test_receiver.py
xdustinface: Thanks, pretty sure you complained about something similar already at some point but i forgot haha. Applied both, see d0c961e43f954c63785acf6afc89f431ade3f6f4

chia/plot_sync/receiver.py
xdustinface: See 9baa49f68a3343451345a6fb5fbe9a4d2777137a

chia/plot_sync/receiver.py
xdustinface: Like discussed in chat im not sure yet if its possible but i have it on my list and will revisit it later.

chia/plot_sync/receiver.py
xdustinface: See af8c3352b6d4e1d6f499b110530da77802fcd2d6

altendky (approved):

PR #11343, opened 2022-04-28
cmds: Adjust stop daemon output

xdustinface:
Stopping the daemon without this patch prints:

```
daemon: {'ack': True, 'command': 'exit', 'data': {'success': True}, 
'destination': 'client', 'origin': 'daemon', 'request_id': 'some 
request_id hash here'}
```

Stopping the daemon with this patch prints: 

```
Daemon stopped
```

and if it fails:

```
Stop daemon failed {'ack': True, 'command': 'exit', 'data': {'success': 
True}, 
'destination': 'client', 'origin': 'daemon', 'request_id': 'some 
request_id hash here'}
```

### DIFF
--- a/chia/cmds/stop.py
+++ b/chia/cmds/stop.py
@@ -17,7 +17,7 @@ async def async_stop(root_path: Path, group: str, stop_daemon: bool) -> int:
     if stop_daemon:
         r = await daemon.exit()
         await daemon.close()
-        if "data" in r and "success" in r["data"] and r["data"]["success"]:
+        if r.get("data", {}).get("success", False):
             print("Daemon stopped")
         else:
             print(f"Stop daemon failed {r}")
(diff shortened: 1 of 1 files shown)


chia/cmds/stop.py
```diff
@@ -17,7 +17,10 @@ async def async_stop(root_path: Path, group: str, stop_daemon: bool) -> int:
     if stop_daemon:
         r = await daemon.exit()
         await daemon.close()
-        print(f"daemon: {r}")
+        if "data" in r and "success" in r["data"] and r["data"]["success"]:
```
altendky, line 20: Maybe.

```suggestion
        if r.get("data", {}).get("success", False):
```

xdustinface: Hm this doesn't fail if `success` is not in `r["data"]` right?

altendky (approved):

PR #11351, opened 2022-04-28
Add nft_get_nft_info API

ytx1991:


paninaro:
Can you add one or more tests to cover the changes?

chia/rpc/wallet_rpc_api.py
```diff
@@ -124,6 +124,7 @@ def get_routes(self) -> Dict[str, Callable]:
             # NFT Wallet
             "/nft_mint_nft": self.nft_mint_nft,
             "/nft_get_current_nfts": self.nft_get_current_nfts,
```
paninaro, line 126: We don't appear to be using `nft_get_current_nfts` anywhere, and your new API appears to supercede it. Can you delete `nft_get_current_nfts`?

chia/rpc/wallet_rpc_api.py
```diff
@@ -124,6 +124,7 @@ def get_routes(self) -> Dict[str, Callable]:
...
+            "/nft_get_nft_info": self.nft_get_nft_info,
```
paninaro, line 127: Can we rename this to `nft_get_nfts` to keep it short and simple?

chia/wallet/did_wallet/did_wallet_puzzles.py
```diff
@@ -21,49 +21,85 @@
...
+    """
+    backup_ids_hash = Program(Program.to(recovery_list)).get_tree_hash()
+    singleton_struct = Program.to((SINGLETON_MOD_HASH, (launcher_id, LAUNCHER_PUZZLE_HASH)))
     return DID_INNERPUZ_MOD.curry(p2_puzzle, backup_ids_hash, num_of_backup_ids_needed, singleton_struct, metadata)
 
 
 def get_inner_puzhash_by_p2(
     p2_puzhash: bytes32,
-    identities: List[bytes],
+    recovery_list: List[bytes],
```
paninaro, line 45: is the `recovery_list` a list of `bytes32`?

chia/wallet/nft_wallet/nft_info.py
```diff
@@ -0,0 +1,62 @@
...
+
+@streamable
+@dataclass(frozen=True)
+class NFTInfo(Streamable):
+    """
+    NFT Info
+    """
+
+    launcher_id: str
+    """
```
paninaro, line 16: I appreciate the comments that you've added for each field, but it's hard to read the code in this format. Is it formatted this way to help some IDE with inline code documentation?

chia/wallet/nft_wallet/nft_info.py
```diff
@@ -0,0 +1,62 @@
...
+    Hash of the license
+    """
+    version: str
+    """
+    Current NFT version
+    """
+    edition_count: uint64
+    """
+    Times of modification
```
paninaro, line 61: edition_count == Times of modification?

chia/wallet/nft_wallet/nft_info.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    Current NFT coin ID
+    """
+    did_owner: str
+    """
+    Owner DID
+    """
+    royalty: uint64
+    """
+    Percentage of the transaction fee paid to the author
```
paninaro, line 29: I might mention that the royalty is represented as the percentage * 1000, so 1500 == 1.5%

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -207,6 +209,53 @@ def get_percentage_from_puzzle(puzzle: Program) -> Optional[uint64]:
...
+        nft_transfer_program_hash,
+        transfer_program_curry_params,
+        metadata,
+    ) = curried_args
+    (
+        royalty_address,
+        trade_price_percentage,
+        settlement_mod_hash,
+        cat_mod_hash,
+    ) = transfer_program_curry_params.as_iter()
```
paninaro, line 234: `transfer_program_curry_params` could be None if `match_nft_puzzle() returns (False, iter(())).

Some sort of error handling would be nice, even if it's just adding an `assert transfer_program_curry_params is not None`. That would at least help point toward the failure in `match_nft_puzzle`.

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -207,6 +209,53 @@ def get_percentage_from_puzzle(puzzle: Program) -> Optional[uint64]:
...
 
+def get_nft_info_from_puzzle(puzzle: Program, nft_coin: Coin) -> NFTInfo:
+    """
+    Extract NFT info from a full puzzle
+    :param puzzle: NFT full puzzle
+    :param nft_coin: NFT coin
+    :return: NFTInfo
+    """
+    # TODO Update this method after the NFT code finalized
+    curried_args = match_nft_puzzle(puzzle)[1]
```
paninaro, line 220: Should probably check that the boolean returned by match_nft_puzzle(puzzle)[0] is actually True before attempting to pull values out of the returned iterator. 

```suggestion
    match = match_nft_puzzle(puzzle)
    if not match[0]:
        raise ValueError("Unable to match NFT puzzle")
    curried_args = match_nft_puzzle(puzzle)[1]
```

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -207,6 +209,53 @@ def get_percentage_from_puzzle(puzzle: Program) -> Optional[uint64]:
...
+    ) = curried_args
+    (
+        royalty_address,
+        trade_price_percentage,
+        settlement_mod_hash,
+        cat_mod_hash,
+    ) = transfer_program_curry_params.as_iter()
+    for kv_pair in metadata.as_iter():
+        if kv_pair.first().as_atom() == b"u":
+            data_uris = []
```
paninaro, line 237: I'd move this line right above the for loop to ensure that it's both declared and initialized to an empty list. If the "u" key isn't found in the metadata, NFTInfo would be constructed with data_uris == None

paninaro, line 237: I'd move this line right above the for loop to ensure that it's both declared and initialized to an empty list. If the "u" key isn't found in the metadata, `NFTInfo` would be constructed with `data_uris` == None

paninaro (changes requested):
Some minor changes and questions, but overall looks good. I'd like to see some tests included though.

Looking forward to integrating this with the UI.

chia/wallet/nft_wallet/nft_puzzles.py
altendky: This is why returning tuples and returning statuses are bad.  The code outside is illegible (what is element 1?) and errors are easily not propagated (the status ceases to be available after indexing the result).  These are general issues we have so not really picking out this code in particular beyond being the example at hand now.

Also, instead of adding a `raise` here, and at every other call, we could just raise inside `match_nft_puzzle()`.  And when you do need to return multiple values, write a dataclass so the values are accessible by name and clearly defined.

ytx1991:
> 

Forgot check in the tests

chia/wallet/nft_wallet/nft_info.py
ytx1991: Python is not my first language, I googled how to add comment in Python. This format is a standard comment style and compatible with document auto-generation tools. Although we don't have plan to generate code document now, but I'd like to keep it in good shape to reduce the future works

chia/wallet/nft_wallet/nft_info.py
ytx1991: TBH, I don't what it is. My understanding it is NFT be edited how many times.

paninaro: Ah ok. I believe it represents how many NFTs were minted in a particular series or batch. It certainly isn't the number of times the NFT was "edited".

chia/wallet/nft_wallet/nft_info.py
paninaro: Can you make these 1-line comments instead of 3-lines then?

e.g.
"""Launcher coin ID"""

chia/wallet/did_wallet/did_wallet_puzzles.py
ytx1991: I think it is, currently. Matt defined this as bytes, not sure if it can be something else in the future.

chia/wallet/nft_wallet/nft_info.py
altendky: Here's an example of some source and generated docs.  I don't recall all the details probably and I should reread this but...  https://peps.python.org/pep-0257/

https://github.com/altendky/qtrio/blob/43c7ff24c0be2f7a3df86eef3c6cc5dae2f7ffd3/qtrio/_core.py#L167-L184
https://qtrio.readthedocs.io/en/v0.5.0/core.html#qtrio.Emission

chia/wallet/nft_wallet/nft_puzzles.py
ytx1991: I agree with you. I saw this pattern in CAT wallet and many other places so I thought it's a common thing in python.

chia/wallet/nft_wallet/nft_info.py
altendky: Also, if the comment such as `Current NFT coin ID` is more descriptive than the attribute name `nft_coin_id` then we should drop the comment and name it `current_nft_coin_id` I think.

chia/wallet/nft_wallet/nft_puzzles.py
ytx1991: For the raise I think not all the use cases want to handle the exception. In some cases they just want a simple if check and move on instead of failing the whole function.

paninaro: Well if the match fails, an exception will be raised regardless with the current code. My point was to at least make it a bit more obvious what the failure was (that the match failed) rather than getting an exception like `ValueError: not enough values to unpack (expected 6, got 0)`

paninaro: I agree with @altendky that it's probably better to fix `match_nft_puzzle`.

altendky: They can catch the exception and handle it.  But sure, there's some balance in deciding the meaning of the return of the function.  It may be that it should just return a list.  If nothing is found the list is empty.  I would have to learn more about the `match_nft_puzzle()` function to weigh in in more detail about that.

I don't think this is particularly common in Python, but it certainly is common here.  It's one of the points I think we need to discuss and come to a conclusion about our approach.

chia/wallet/nft_wallet/nft_info.py
```diff
@@ -0,0 +1,47 @@
...
+from typing import List
+
+from chia.util.ints import uint64
+from chia.util.streamable import Streamable, streamable
+
+
+@streamable
+@dataclass(frozen=True)
+class NFTInfo(Streamable):
+    """NFT Info"""
```
trepca, line 11: can you add some additional info where this is used and why

chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -0,0 +1,97 @@
+from chia.types.blockchain_format.program import Program
+from chia.wallet.puzzles.load_clvm import load_clvm
+
+SINGLETON_TOP_LAYER_MOD = load_clvm("singleton_top_layer_v1_1.clvm")
+NFT_MOD = load_clvm("nft_innerpuz.clvm")
+
+
+class UncurriedNFT:
+    """Uncurry a NFT puzzle"""
```
trepca, line 9: add some more details in docstrings as before pls

trepca (changes requested):

chia/wallet/nft_wallet/nft_info.py
altendky: Another point about trivial docstrings, if we do get to doc generation or even just checking for docstrings then trivial ones will effectively silence any warnings about missing docstrings.

chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -0,0 +1,98 @@
...
+SINGLETON_TOP_LAYER_MOD = load_clvm("singleton_top_layer_v1_1.clvm")
+NFT_MOD = load_clvm("nft_innerpuz.clvm")
+
+
+class UncurriedNFT:
+    """
+    A simple solution for uncurry NFT puzzle.
+    Initial the class with a full NFT puzzle, it will do a deep uncurry.
+    This is the only place you need to change after modified the Chialisp curried parameters.
+    """
```
altendky, line 13: https://peps.python.org/pep-0257/

I like semantic newlines.  I don't offhand know if there are any issues with them in docstrings or how others feel about them.

It looks like there are a few other places to match this format.  Thanks, by the way, for caring about docs.  `:]`

```suggestion
    """A simple solution for uncurry NFT puzzle.

    Initial the class with a full NFT puzzle, it will do a deep uncurry.
    This is the only place you need to change after modified the Chialisp curried parameters.
    """
```

chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -0,0 +1,98 @@
...
+
+    metadata: Program
+    """
+    NFT metadata
+    [("u", data_uris), ("h", data_hash)]
+    """
+    data_uris: Program
+    data_hash: Program
+
+    def __init__(self, puzzle: Program, raise_exception: bool = False):
```
altendky, line 54: There has not been a team consensus on this but I strongly encourage that every class be a `@dataclass` and not use `.__post_init__()` to add back complexity to the constructor.  Rather, have a helper function or `@classmethod` that does the fancy calculations needed to figure out the attributes of the to-be-created class.

chia/wallet/nft_wallet/uncurry_nft.py
ytx1991: I think a new line is more friendly with doc generators and human reading

paninaro (approved):
lgtm

chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -0,0 +1,108 @@
...
+
+    metadata: Program = field(init=False)
+    """
+    NFT metadata
+    [("u", data_uris), ("h", data_hash)]
+    """
+    data_uris: Program = field(init=False)
+    data_hash: Program = field(init=False)
+
+    @staticmethod
```
altendky, line 59: `@classmethod` not `@staticmethod`, use the `cls` parameter, and generally figure out what a class is going to be before building it instead of building entirely the wrong instance and then modifying everything.

chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -0,0 +1,108 @@
...
+
+@dataclass
+class UncurriedNFT:
+    """
+    A simple solution for uncurry NFT puzzle.
+    Initial the class with a full NFT puzzle, it will do a deep uncurry.
+    This is the only place you need to change after modified the Chialisp curried parameters.
+    """
+
+    matched: bool = field(init=False)
```
altendky, line 20: `init=False` is not used much.  An instance of the class should mostly be able to be constructed by a single constructor call instead of a totally generic constructor call and then 20 attribute assignments.  In some cases, once you reach that structure you can make the class instances frozen/immutable.

altendky (commented):
I'll try to find time to correct this soon as an example like I did over in https://github.com/Chia-Network/chia-blockchain/pull/11139.

chia/wallet/nft_wallet/uncurry_nft.py
altendky: Note that this ends up leaving the class with maybe-not-even-present attributes.

```python-console
Python 3.8.10 (default, Mar 15 2022, 12:22:08) 
[GCC 9.4.0] on linux
>>> import dataclasses
>>> @dataclasses.dataclass
... class C:
...     x: int = dataclasses.field(init=False)
... 
>>> c = C()
>>> c
  File "/usr/lib/python3.8/dataclasses.py", line 368, in wrapper
    result = user_function(self)
  File "<string>", line 3, in __repr__
AttributeError: 'C' object has no attribute 'x'
```

chia/wallet/nft_wallet/uncurry_nft.py
ytx1991: Good to know, if you will work on this I will leave it to you.

altendky: Being worked over in https://github.com/Chia-Network/chia-blockchain/pull/11379.  I'll ping you for a review when it is ready.

PR #11353, opened 2022-04-28
Wallet Allow Unsynced

cmmarslender:
Allows syncing from an unsynced node, if this configuration flag is set. Useful in cases like syncing an offline cold wallet. With this enabled, and chia installed from the cli deb, or some other method, a backup DB can be provided to the node via USB or some other method, this configuration option enabled, and the wallet will sync up to the point of the DB provided.

### DIFF
--- a/chia/util/initial-config.yaml
+++ b/chia/util/initial-config.yaml
@@ -501,6 +501,9 @@ wallet:
   multiprocessing_start_method: default
 
   testing: False
+  # Allow unsynced full node will allow syncing the wallet from an unsynced full node
+  # Useful in edge cases like syncing an offline wallet from a backup/snapshot of a full node DB w/ an offline node
+  allow_unsynced_full_node: False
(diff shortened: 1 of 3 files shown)


chia/util/initial-config.yaml
```diff
@@ -501,6 +501,9 @@ wallet:
   multiprocessing_start_method: default
 
   testing: False
+  # Allow unsynced full node will allow syncing the wallet from an unsynced full node
+  # Useful in edge cases like syncing an offline wallet from a backup/snapshot of a full node DB w/ an offline node
+  allow_unsynced_full_node: False
```
mariano54, line 506: Default should be False

mariano54 (changes requested):

chia/util/initial-config.yaml
mariano54: NVM it's False already.

PR #11356, opened 2022-04-28
readd ping_task as keepalive for daemon

wjblanke:


wjblanke:
Screen Shot 2022-04-28 at 12 22 13 PM

i got rid of this part, not sure if it is still needed

chia/daemon/server.py
```diff
@@ -270,6 +272,28 @@ def remove_connection(self, websocket: WebSocketResponse):
...
+                for socket in sockets:
+                    try:
+                        self.log.info(f"About to ping: {service_name}")
+                        await socket.ping()
+                    except asyncio.CancelledError:
+                        self.log.info("Ping task received Cancel")
+                        restart = False
+                        break
+                    except Exception as e:
+                        self.log.info(f"Ping error: {e}")
```
paninaro, line 290: self.log.exception(f"Ping error")

chia/daemon/server.py
```diff
@@ -270,6 +272,28 @@ def remove_connection(self, websocket: WebSocketResponse):
...
+        await asyncio.sleep(30)
+        for remote_address, service_name in self.remote_address_map.items():
+            if service_name in self.connections:
+                sockets = self.connections[service_name]
+                for socket in sockets:
+                    try:
+                        self.log.info(f"About to ping: {service_name}")
+                        await socket.ping()
+                    except asyncio.CancelledError:
+                        self.log.info("Ping task received Cancel")
```
paninaro, line 286: self.log.warning  maybe?

chia/daemon/server.py
mariano54: log.error?

PR #10273, opened 2022-02-17
fix typo and index issues in wallet database

arvidn:
This PR addresses the most serious issues raised by @neurosis69 here: https://github.com/Chia-Network/chia-blockchain/issues/10276

it:

* fixed the `PRIMARY_KEY` typo
* introduce a pre-commit check for duplicate index and table names
* deduplicates index names in wallet database ('name', 'wallet_type' and 'wallet_id')

This is the output from the checking script:

```
$ python tests/check_sql_statements.py 
duplicate INDEX "name"
    key_val_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS name on key_val_store(key)")
    wallet_action_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS name on action_queue(name)")
duplicate INDEX "wallet_type"
    wallet_action_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS wallet_type on action_queue(wallet_type)")
    wallet_coin_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS wallet_type on coin_record(wallet_type)")
duplicate INDEX "wallet_id"
    wallet_action_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS wallet_id on action_queue(wallet_id)")
    wallet_coin_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS wallet_id on coin_record(wallet_id)")
duplicate INDEX "wallet_type"
    wallet_coin_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS wallet_type on coin_record(wallet_type)")
    wallet_puzzle_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS wallet_type on derivation_paths(wallet_type)")
duplicate INDEX "wallet_id"
    wallet_coin_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS wallet_id on coin_record(wallet_id)")
    wallet_puzzle_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS wallet_id on derivation_paths(wallet_id)")
duplicate INDEX "wallet_id"
    wallet_puzzle_store.py:        await self.db_connection.execute("CREATE INDEX IF NOT EXISTS wallet_id on derivation_paths(wallet_id)")
[...]

PR #11358, opened 2022-04-28
Add wallet_id as derivation_paths primary key with puzzle_hash

altendky:
In https://github.com/Chia-Network/chia-blockchain/pull/10273 (released in 1.3.4) the `derivation_paths` table was given a primary key by correcting the typo `PRIMARY_KEY`.  Up until then when `WalletPuzzleStore.add_derivation_paths()` executed `INSERT OR REPLACE INTO derivation_paths` every new row was considered unique since there was no primary key to match.  With the correction, any addition would replace instead of inserting if the puzzle hash matched.  This means that adding the same puzzle hash for new wallet id would replace the existing entry.  This does not seem like the intended behavior and caused trouble in the datalayer work.

Draft for:
- [x] General discussion
- [x] Consideration of implications to 1.3.4
  - It seems that the issue is not exercised by current main or released code.
- [x] Should this be included in 1.3.5
  - No need to rush this out.
- [ ] How to handle migration
  - Properly...
  - There is no change to other queries so no need to track versioning for basic functionality.  This change can be put in place now such that it affects new DBs.  In the future we may choose to force an upgrade on this.

### DIFF
--- a/chia/wallet/wallet_puzzle_store.py
+++ b/chia/wallet/wallet_puzzle_store.py
@@ -40,11 +40,12 @@ class WalletPuzzleStore:
                 "CREATE TABLE IF NOT EXISTS derivation_paths("
                 "derivation_index int,"
                 " pubkey text,"
-                " puzzle_hash text PRIMARY KEY,"
+                " puzzle_hash text,"
                 " wallet_type int,"
                 " wallet_id int,"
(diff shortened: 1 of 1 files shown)


arvidn (approved):

altendky:
implicitly merged by way of atari

PR #11362, opened 2022-04-29
remove unused functions on MerkleSet

arvidn:
They are also not tested, so they don't deserve to exist.

### DIFF
--- a/chia/util/merkle_set.py
+++ b/chia/util/merkle_set.py
@@ -100,10 +100,6 @@ class Node(metaclass=ABCMeta):
         pass
 
     @abstractmethod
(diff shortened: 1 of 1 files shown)


chia/util/merkle_set.py
```diff
@@ -349,18 +316,10 @@ class SetError(Exception):
     pass
 
 
-def confirm_included(root: bytes32, val: bytes, proof: bytes32) -> bool:
-    return confirm_not_included_already_hashed(root, sha256(val).digest(), proof)
```
arvidn, line 353: it also looks like this function never worked right

mariano54 (approved):
Yes they don't deserve it

PR #11363, opened 2022-04-29
Add fees support to DID wallet creation RPC client

AmineKhaldi:
Needed to enable setting fees in the DID wallet creation command.


paninaro (approved):
Looks good to me. I see that `create_new_wallet`'s DID path already handles the fee.

PR #11247, opened 2022-04-21
util: Implement `Paginator` class as interface to access a list by pages

xdustinface:
Preparation for a follow up PR to provide paginated plot RPC endpoints. Im fine waiting with the merge until that other PR is ready, just opening this to already get it reviewed.

PR #11365, opened 2022-04-29
farmer|rpc|tests: Implement paginated harvester plot endpoints

xdustinface:
This PR introduces four new farmer RPC endpoints which provide paginated results as well as a `filter`, `sort_key` and `reverse` option: 

- `get_harvester_plots_valid` -> To query the valid plot which are used for the farming
- `get_harvester_plots_invalid` -> To query invalid plots (the ones the harvester couldn't load for whatever reason)
- `get_harvester_plots_keys_missing` -> To query the plots where the farmer is missing the related keys
- `get_harvester_plots_duplicates` -> To query the plots which are duplicated on the harvester

### Request data
The request data for the valid plots endpoint is slightly different because it returns all attributes of a plot, the other endpoints only return the paths of the plots.

#### Valid plots endpoint

```
@dataclasses.dataclass
class PlotInfoRequestData:
    node_id: bytes32
    page: int
    page_size: int
    filter: List[FilterItem] = dataclasses.field(default_factory=list)
    sort_key: str = "filename"
    reverse: bool = False
```

- `node_id` is the node id of the harvester to query plots from 
- `page` the page to query
- `page_size` the number of element per size
- `filter` is a list of key/value pairs which allows to filter any attribute of the plot. (see class `KeyValue` and `Plot` below) The filters will be applied in the same order they are in the list. 
- `sort_key` allows to sort the plot data by any of the keys available in the plot data. See class `Plot` below.
- `reverse` allows to revert the order of the sorting

##### Filter class
```
@dataclasses.dataclass
class FilterItem:
    key: str
    value: str
```

##### Plot class

```
@streamable
@dataclass(frozen=True)
class Plot(Streamable):
    filename: str
    size: uint8
    plot_id: bytes32
    pool_public_key: Optional[G1Element]
    pool_contract_puzzle_hash: Optional[bytes32]
    plot_public_key: G1Element
    file_size: uint64
    time_modified: uint64
```

#### Other plot endpoints 

```
@dataclasses.dataclass
class PlotPathRequestData:
    node_id: bytes32
    page: int
    page_size: int
    filter: List[str] = dataclasses.field(default_factory=list)
    reverse: bool = False
```

- `node_id` is the node id of the harvester to query plots from 
- `page` the page to query
- `page_size` the number of element per size
- `filter` is a list of strings which allows to filter for the path. The filters will be applied in the same order they are in the list. 
- `reverse` allows to revert the order of the sorting

### Valid plots example request

```
chia rpc farmer get_harvester_plots_valid '{"node_id": "0xb429aab1fe8686c6163ae442fa45ddcdef301d751adfa6855c
54f2d4d6a1a747", "page": 2, "page_size": 1, "filter": [{"key": "filename", "value": "27-19-4"}], "sort_key": "time_modified", "reverse": 1}'
{
    "page": 2,
    "page_count": 5,
    "node_id": "b429aab1fe8686c6163ae442fa45ddcdef301d751adfa6855c54f2d4d6a1a747",
    "plots": [
        {
            "file_size": 671906245,
            "filename": "/Users/dustinface/chia/plots/testnet10/plot-k25-2022-02-27-19-44-748597cd5d19e21515df1873267ea6ba8e66526f6bd9aa4e15f33c5955bb27bd.plot",
            "plot_id": "0x748597cd5d19e21515df1873267ea6ba8e66526f6bd9aa4e15f33c5955bb27bd",
            "plot_public_key": "0x928df0d5eaab39a3530c1e1925730b0e1b21e485ddc457d9de6ef4f403e7710fe897bc40f0306cc1ba29d423f9bfdc6f",
            "pool_contract_puzzle_hash": "0x9af69ba2e65b27eef27ae5e5293899a7e4d93bf781f6a8c13db1af06af905cfa",
            "pool_public_key": null,
            "size": 25,
            "time_modified": 1645987573
        }
    ],
    "success": true,
    "total_count": 5
}
```

Based on #11247 and #11342

### DIFF
--- a/chia/rpc/farmer_rpc_api.py
+++ b/chia/rpc/farmer_rpc_api.py
@@ -15,7 +15,7 @@ from chia.util.ws_message import WsRpcMessage, create_payload_dict
 
 
 class PaginatedRequestData(Protocol):
-    peer_id: bytes32
+    node_id: bytes32
     page: int
     page_size: int
 
@@ -28,7 +28,7 @@ class FilterItem:
 
 @dataclasses.dataclass
 class PlotInfoRequestData:
-    peer_id: bytes32
+    node_id: bytes32
     page: int
     page_size: int
     filter: List[FilterItem] = dataclasses.field(default_factory=list)
@@ -38,7 +38,7 @@ class PlotInfoRequestData:
 
 @dataclasses.dataclass
 class PlotPathRequestData:
-    peer_id: bytes32
+    node_id: bytes32
     page: int
     page_size: int
     filter: List[str] = dataclasses.field(default_factory=list)
@@ -48,7 +48,7 @@ class PlotPathRequestData:
 def paginated_plot_request(source: List[Any], request: PaginatedRequestData) -> Dict[str, object]:
     paginator: Paginator = Paginator(source, request.page_size)
     return {
-        "peer_id": request.peer_id.hex(),
+        "node_id": request.node_id.hex(),
         "page": request.page,
         "page_count": paginator.page_count(),
         "total_count": len(source),
@@ -224,7 +224,7 @@ class FarmerRpcApi:
     async def get_harvester_plots_valid(self, request_dict: Dict[str, object]) -> Dict[str, object]:
         # TODO: Consider having a extra List[PlotInfo] in Receiver to avoid rebuilding the list for each call
         request = dataclass_from_dict(PlotInfoRequestData, request_dict)
-        plot_list = list(self.service.get_receiver(request.peer_id).plots().values())
+        plot_list = list(self.service.get_receiver(request.node_id).plots().values())
         # Apply filter
         plot_list = [
             plot for plot in plot_list if all(plot_matches_filter(plot, filter_item) for filter_item in request.filter)
@@ -241,7 +241,7 @@ class FarmerRpcApi:
         self, source_func: Callable[[Receiver], List[str]], request_dict: Dict[str, object]
     ) -> Dict[str, object]:
         request: PlotPathRequestData = dataclass_from_dict(PlotPathRequestData, request_dict)
-        receiver = self.service.get_receiver(request.peer_id)
+        receiver = self.service.get_receiver(request.node_id)
         source = source_func(receiver)
         request = dataclass_from_dict(PlotPathRequestData, request_dict)
         # Apply filter
--- a/chia/rpc/farmer_rpc_api.py
+++ b/chia/rpc/farmer_rpc_api.py
@@ -242,8 +242,7 @@ class FarmerRpcApi:
         source = source_func(receiver)
         request = dataclass_from_dict(PlotPathRequestData, request_dict)
         # Apply filter
-        for filter_item in request.filter:
-            source = [plot for plot in source if filter_item in plot]
+        source = [plot for plot in source if all(filter_item in plot for filter_item in request.filter)]
         # Apply reverse
         source = sorted(source, reverse=request.reverse)
         return paginated_plot_request(source, request)
--- a/chia/rpc/farmer_rpc_api.py
+++ b/chia/rpc/farmer_rpc_api.py
@@ -1,6 +1,6 @@
 import dataclasses
 import operator
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional
 
 from typing_extensions import Protocol
 
@@ -51,14 +51,12 @@ def paginated_plot_request(source: List[Any], request: PaginatedRequestData) -> 
     }
 
 
-def is_filter_match(plot: Union[Plot, Dict[str, Any]], filter_item: KeyValue) -> bool:
-    if isinstance(plot, Plot):
-        plot_attribute = getattr(plot, filter_item.key)
+def plot_matches_filter(plot: Plot, filter_item: KeyValue) -> bool:
+    plot_attribute = getattr(plot, filter_item.key)
+    if filter_item.value is None:
+        return plot_attribute is None
     else:
-        plot_attribute = plot[filter_item.key]
-    none_match = filter_item.value is None and plot_attribute is None
-    attribute_match = filter_item.value is not None and filter_item.value in str(plot_attribute)
-    return none_match or attribute_match
+        return filter_item.value in str(plot_attribute)
 
 
 class FarmerRpcApi:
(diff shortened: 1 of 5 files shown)


xdustinface:
Looking at the failing tests (on CI) on monday and i will still add some failure case tests but the implementation is ready for review.

chia/rpc/farmer_rpc_api.py
```diff
@@ -150,6 +198,64 @@ async def get_harvesters(self, _: Dict):
...
+            return {
+                "peer_id": request.peer_id.hex(),
+                "page": request.page,
+                "page_count": paginator.page_count(),
+                "total_count": len(source),
+                "plots": paginator.get_page(request.page),
+            }
+        except Exception as e:
+            self.service.log.error(f"paginated_plot_request: failed with {e}")
+            return {"error": str(e)}
```
altendky, line 213: Handled by the API framework.

chia/rpc/farmer_rpc_api.py
```diff
@@ -150,6 +198,64 @@ async def get_harvesters(self, _: Dict):
...
+        try:
+            request: PlotPathRequestData = dataclass_from_dict(PlotPathRequestData, request_dict)
+            receiver = self.service.get_receiver(request.peer_id)
+            source = source_func(receiver)
+            request = dataclass_from_dict(PlotPathRequestData, request_dict)
+            # Apply filter
+            for filter_item in request.filter:
+                source = [plot for plot in source if filter_item in plot]
+            # Apply reverse
+            source = sorted(source, reverse=request.reverse)
```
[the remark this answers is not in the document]
xdustinface: Nope, it's only a list of paths in this case.

chia/rpc/farmer_rpc_api.py
```diff
@@ -159,6 +218,45 @@ async def get_harvesters(self, _: Dict):
...
+    def paginated_plot_path_request(
+        self, source_func: Callable[[Receiver], List[str]], request_dict: Dict[str, object]
+    ) -> Dict[str, object]:
+        request: PlotPathRequestData = dataclass_from_dict(PlotPathRequestData, request_dict)
+        receiver = self.service.get_receiver(request.peer_id)
+        source = source_func(receiver)
+        request = dataclass_from_dict(PlotPathRequestData, request_dict)
+        # Apply filter
+        for filter_item in request.filter:
+            source = [plot for plot in source if filter_item in plot]
```
altendky, line 246: To explain, my point, that may not be the important one, is that this form avoids creating several possibly large intermediate lists.

```suggestion
        source = [plot for plot in source if all(filter_item in plot for filter_item in request.filter)]
```

chia/util/misc.py
```diff
@@ -75,3 +78,13 @@ def prompt_yes_no(prompt: str = "(y/n) ") -> bool:
...
+
+
+def dataclass_to_json_dict(instance: Any) -> Dict[str, Any]:
+    return recurse_jsonify(dataclasses.asdict(instance))
+
+
+@dataclasses.dataclass
+class KeyValue:
+    key: str
+    value: Optional[str]
```
altendky, line 90: This is too generic a name for the specific types, I think.  Where you hinted this you would use `KeyValue[str, Optional[str]]`.

```suggestion
@dataclasses.dataclass
class KeyValue(Generic[K, V]):
    key: K
    value: V
```

https://mypy-play.net/?mypy=latest&python=3.10&gist=dde9ea576b05527136c4c74b11a9cde1
```python
import dataclasses
import typing


K = typing.TypeVar("K")
V = typing.TypeVar("V")


@dataclasses.dataclass
class KeyValue(typing.Generic[K, V]):
    key: K
    value: V


kva = KeyValue(key="abc", value=123)
reveal_type(kva)

kvb = KeyValue[str, typing.Optional[int]](key="abc", value=123)
reveal_type(kvb)

kvc: KeyValue[bool, float]
kvc = KeyValue(key=True, value=4.2)
reveal_type(kvc)
```

```
main.py:16: note: Revealed type is "__main__.KeyValue[builtins.str, builtins.int]"
main.py:19: note: Revealed type is "__main__.KeyValue[builtins.str, Union[builtins.int, None]]"
main.py:23: note: Revealed type is "__main__.KeyValue[builtins.bool, builtins.float]"
Success: no issues found in 1 source file
```

chia/rpc/farmer_rpc_api.py
```diff
@@ -1,11 +1,66 @@
...
+
+
+def is_filter_match(plot: Union[Plot, Dict[str, Any]], filter_item: KeyValue) -> bool:
+    if isinstance(plot, Plot):
+        plot_attribute = getattr(plot, filter_item.key)
+    else:
+        plot_attribute = plot[filter_item.key]
+    none_match = filter_item.value is None and plot_attribute is None
+    attribute_match = filter_item.value is not None and filter_item.value in str(plot_attribute)
+    return none_match or attribute_match
```
altendky, line 61: I feel like the two types are a bit disparate to be handled in the same class.  `isinstance()` kind of smells anyways.  How about something like `def plot_dict_matches_filter():` to handle dicts and `def plot_matches_filter():` to handle `Plot`s by converting them to a dict and calling `plot_dict_matches_filter()`?  Or just those two functions where both call a third function with the common portion?

tests/core/test_farmer_harvester_rpc.py
```diff
@@ -363,3 +387,158 @@ async def remove_all_and_validate() -> bool:
...
+            plots = [plot for plot in plots if is_filter_match(plot, filter_item)]
+        plots.sort(key=operator.itemgetter(sort_key, "plot_id"), reverse=reverse)
+    else:
+        for filter_item in filtering:
+            plots = [plot for plot in plots if filter_item in plot]
+        plots.sort(reverse=reverse)
+
+    total_count = len(plots)
+    assert total_count > 0
+    assert len(plots) == total_count == expected_plot_count
```
altendky, line 520:
```suggestion
    total_count = len(plots)
    assert total_count == expected_plot_count
```

chia/rpc/farmer_rpc_api.py
altendky, line 61:
```suggestion
    if filter_item.value is None:
        return plot_attribute is None
    else:
        return filter_item.value in str(plot_attribute)
```

altendky (changes requested):

chia/util/misc.py
xdustinface: See [bc2c789](https://github.com/Chia-Network/chia-blockchain/pull/11365/commits/bc2c7895e723240fbad0d568b7c5ea771b5b1bcd), i moved it into `farmer_rpc_api.py` and named it less generic since your approach doesn't work with `dataclass_from_dict`.

chia/rpc/farmer_rpc_api.py
xdustinface: I just made it only work with `Plot` and now it converts to `Plot` in tests before calling it. See da6f90b9a08d6fb8570acad95caed03ba9bf8bd9

xdustinface: Also handled in da6f90b9a08d6fb8570acad95caed03ba9bf8bd9

altendky (approved):

PR #11367, opened 2022-04-29
farmer|gui: Enable paginated plot loading and improved state reporting

xdustinface:
This PR pins the new GUI and changes the event which gets sent from the farmer to the GUI which was previously called `new_plots` with the payload of  `get_harvesters` with the plots of all harvesters.

Now with this PR the event gets renamed to `harvester_update` and it:

- Only sends a summary (basically one entry of `get_harvester_summay`) for the harvester which actually sent an update
- Is used to notify for each loaded batch during initial loading of an harvester

There is also a new state changed event called `harvester_removed` which gets triggered by the farmer if the harvester was disconnected and has the payload `{"node_id": <node_id of the harvester which was removed>}`

Based on #11342 and required #11365

### DIFF
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -256,11 +256,14 @@ class Farmer:
         self.state_changed("close_connection", {})
         if connection.connection_type is NodeType.HARVESTER:
             del self.plot_sync_receivers[connection.peer_node_id]
+            self.state_changed("harvester_removed", {"node_id": connection.peer_node_id})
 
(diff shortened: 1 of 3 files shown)


xdustinface:
[70679c1](https://github.com/Chia-Network/chia-blockchain/pull/11367/commits/70679c1d6ec578760c8071a4547c8dc91e3f6c95) is ready for review but GUI will probably receive another update in a bit with small style changes.

altendky (dismissed):

emlowe (approved):
approved

PR #11371, opened 2022-04-29
Keep daemon websocket alive during keyring unlock

paninaro:
When prompting to unlock the keyring during daemon launch, it's possible that the daemon websocket connection will  timeout before the user has entered their passphrase. ~We can't quickly detect that the connection has timed-out, so instead we now reconnect after collecting the passphrase, unlock the keyring, and then close the old connection.~ We now collect the passphrase on a separate thread to keep the websocket connection alive and happy.

Testing notes:

Previously, running `chia start farmer`, waiting 60+ seconds at the passphrase prompt, and then entering the passphrase would lead to a 30 second delay before an exception was raised indicating that the daemon connection had been closed.

With the patch, running `chia start farmer`, waiting 60+ seconds at passphrase prompt, and then entering the passphrase --> services started as expected

Tested on macOS, Windows, and Linux

### DIFF
--- a/chia/cmds/start_funcs.py
+++ b/chia/cmds/start_funcs.py
@@ -1,4 +1,5 @@
 import asyncio
+from concurrent.futures import ThreadPoolExecutor
 import os
 import subprocess
 import sys
@@ -37,7 +38,8 @@ async def create_start_daemon_connection(root_path: Path) -> Optional[DaemonProx
         if await connection.is_keyring_locked():
             passphrase = Keychain.get_cached_master_passphrase()
             if not Keychain.master_passphrase_is_valid(passphrase):
-                passphrase = await asyncio.get_running_loop().run_in_executor(None, get_current_passphrase)
+                with ThreadPoolExecutor(max_workers=1, thread_name_prefix="get_current_passphrase") as executor:
+                    passphrase = await asyncio.get_running_loop().run_in_executor(executor, get_current_passphrase)
 
         if passphrase:
             print("Unlocking daemon keyring")
--- a/chia/cmds/start_funcs.py
+++ b/chia/cmds/start_funcs.py
@@ -45,6 +45,7 @@ async def create_start_daemon_connection(root_path: Path) -> Optional[DaemonProx
                     # create a new connection, unlock the keyring, and then close the old connection.
                     old_connection = connection
                     connection = await connect_to_daemon_and_validate(root_path)
+                    assert connection is not None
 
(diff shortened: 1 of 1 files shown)


paninaro:
@altendky suggested another approach of using `asyncio.loop.run_in_executor` to make the passphrase input function async. I've tested this approach (wrapping the call to `get_current_passphrase()` on macOS and Windows and it does appear to work, with the (probably low) risk of introducing multithreading issues. The execution path under `get_current_pasphrase()` looks clean to me currently, but who knows what future changes might introduce.

paninaro:
> @altendky suggested another approach of using `asyncio.loop.run_in_executor` to make the passphrase input function async compatible. I've tested this approach (wrapping the call to `get_current_passphrase()`) on macOS and Windows and it does appear to work, with the (probably low) risk of introducing multithreading issues. The execution path under `get_current_pasphrase()` looks clean to me currently, but who knows what future changes might introduce.

That change looks like:

```
    if connection:
        passphrase = None
        if await connection.is_keyring_locked():
            passphrase = Keychain.get_cached_master_passphrase()
            if not Keychain.master_passphrase_is_valid(passphrase):
 -->            passphrase = await asyncio.get_running_loop().run_in_executor(None, get_current_passphrase)

        if passphrase:
            print("Unlocking daemon keyring")
            await connection.unlock_keyring(passphrase)
```

chia/cmds/start_funcs.py
```diff
@@ -37,7 +37,7 @@ async def create_start_daemon_connection(root_path: Path) -> Optional[DaemonProx
         if await connection.is_keyring_locked():
             passphrase = Keychain.get_cached_master_passphrase()
             if not Keychain.master_passphrase_is_valid(passphrase):
-                passphrase = get_current_passphrase()
+                passphrase = await asyncio.get_running_loop().run_in_executor(None, get_current_passphrase)
```
trepca, line 40: default executor will create `num_cpus * 5` worker threads so I'd define a custom executor:
```
with ThreadPoolExecutor(1, "getpass") as executor:
        return await asyncio.get_event_loop().run_in_executor(executor, get_current_passphrase)

```

altendky: I think the question would be whether we want to have a general pool for other potential uses like this or not.

paninaro: Updated to use a ThreadPoolExecutor with 1 worker thread.

trepca (approved):

PR #11385, opened 2022-05-01
Remove outdated 3.8 upgrade comment

altendky:


hoffmang9:
Is there still a comment somewhere to deal with the windows issue of needing hand generated miniupnpc wheel on each version upgrade?

altendky:
No, but there is a workflow that checks for having wheels available that runs on all platforms and versions (as long as we keep that list updated).  It allows a couple pure-Python exceptions to the wheel requirement at the moment.

https://github.com/Chia-Network/chia-blockchain/blob/795f1995976fee6da630bef5140976d21e77fa38/build_scripts/check_dependency_artifacts.py
https://github.com/Chia-Network/chia-blockchain/runs/6247010061?check_suite_focus=true

On this topic, now that 3.10 is out I should add a 3.11 PR to track progress...  forgot about that.  `:]`

altendky:
As to keeping the workflow matrix updated, I've been wanting to make a checker so we can define our matrix in one place and verify that all all workflows that should end up covering that.  But, not yet.

hoffmang9:
All sounds fine - it would just be a very easy mistake to ship a version of windows installer without uPnP and knock like 50% of our open nodes out of availability...

altendky:
Ugh, after working on Windows CI for so long I forget it still is just a PR...  Hopefully quickly adding it to just the wheels check workflow in https://github.com/Chia-Network/chia-blockchain/pull/11388.

altendky:
@cmmarslender added the 3.7 and 3.8 Windows miniupnpc wheels that that workflow turned up as missing.  But, there are some differences between what you built, what Chris built, and the `.egg`s that miniupnpc published that we want to look into.

cmmarslender:
> @cmmarslender added the 3.7 and 3.8 Windows miniupnpc wheels that that workflow turned up as missing. But, there are some differences between what you built, what Chris built, and the `.egg`s that miniupnpc published that we want to look into.

All the wheels have the .dll bundled inside correctly now.. Tested the 3.10 w/ a upnp enabled router, and it forwarded the port successfully on this test.

cmmarslender (approved):

PR #11389, opened 2022-05-01
make sure we set the syncto height correctly when we roll back instea…

wjblanke:
…d of noop


chia/wallet/wallet_blockchain.py
```diff
@@ -184,8 +184,8 @@ async def get_peak_block(self) -> Optional[HeaderBlock]:
             return self._peak
         return await self._basic_store.get_object("PEAK_BLOCK", HeaderBlock)
 
-    async def set_finished_sync_up_to(self, height: int, in_transaction=False):
-        if height > await self.get_finished_sync_up_to():
+    async def set_finished_sync_up_to(self, height: int, in_transaction=False, in_rollback=False):
```
mariano54, line 187: I would rename this argument to: `allow_rollback`

mariano54 (approved):

PR #11391, opened 2022-05-02
DID set wallet name RPC API: Ensure that the wallet type is a DID wallet before proceeding

AmineKhaldi:


paninaro (approved):
Thanks for adding this check. LGTM.

PR #11396, opened 2022-05-02
more flake8

altendky:


xdustinface (approved):
Ok, but why doesn't this just use the config?

altendky:
I think it didn't offer a way to do this in the config.  If you can find one I would much prefer that.  Even if we just got to specifying `.`.

wjblanke (approved):
aok

PR #11402, opened 2022-05-02
Create SECURITY.md

justinengland:


SECURITY.md
```diff
@@ -0,0 +1,19 @@
...
+| 1.2.x   | :x:                |
+| 1.3.x   | :white_check_mark: |
+
+## Reporting a Vulnerability
+
+Please report security concerns to security@chia.net.
+
+If your security issue is established to be valid, we will reach out immediately to establish
+communication channels and compensate the issue reporter for responsibly reporting security bugs via
+our bug bounty program.
```
Starttoaster, line 19: Might be nice to link to info on the BBP? Like https://www.chia.net/2021/10/21/bugcrowd-bounty-launch.en.html

SECURITY.md
```diff
@@ -0,0 +1,19 @@
+# Security Policy
+
+## Supported Versions
```
TheLastCicada, line 3: What are the odds that this actually gets updated regularly like it should?  Suppose it could be part of @wallentx 's process.....

SECURITY.md
justinengland: we cannot link to bug bounty now because its not open to the public. This is an ongoing thing @wriches is sorting out with them.

SECURITY.md
justinengland: we should add automation to update this. IMO

PR #11409, opened 2022-05-03
Implement nft transfer program

ytx1991:
Just implemented the transfer program in chialisp. Not sure what is the meaning of "minter can make royalty etc rules here". I assume the minter can assign any royalty logic here, so I introduced an additional layer (royalty program). If this is unnecessary, let me know. I haven't implement the royalty program yet since there still have some things are uncertain.


chia/wallet/puzzles/nft_ownership_transfer_program.clvm
```diff
@@ -1,4 +1,47 @@
-(mod ARGS
+(mod
+    (
+        ; NFT Singleton ID
+        NFT_ID
```
matt-o-how, line 4: NFT_ID is the same thing as SINGLETON_LAUNCHER ID

chia/wallet/puzzles/nft_ownership_transfer_program.clvm
```diff
@@ -1,4 +1,47 @@
...
+                                        (sha256tree SINGLETON_STRUCT)
+       )
+    )
+
+    ; main
+    (list
+        (c (list CHANGE_OWNERSHIP new_owner)
+            (c (list ASSERT_PUZZLE_ANNOUNCEMENT (sha256 (calculate_full_puzzle_hash SINGLETON_STRUCT current_did_puzhash) (sha256tree royalty_prog_solution) new_owner))
+                (c (list ASSERT_PUZZLE_ANNOUNCEMENT (sha256 (calculate_full_puzzle_hash (c (f SINGLETON_STRUCT) (c new_did (r (r SINGLETON_STRUCT)))) new_did_puzhash) (sha256tree royalty_prog_solution)))
+                    (a royalty_prog_reveal royalty_prog_solution)
```
matt-o-how, line 42: instead of running a royalty_generating program, we can just have it exist in this layer - no need to delegate it further.
You can adapt the old royalty generating code from the old transfer program for this.

PR #11411, opened 2022-05-03
fix offer to_valid_spend

0xJohnZW:
when aggreate mulit offers which offer the same CATs, we need to combine all the offer coins.


Quexington:
Hey, thanks for the contribution!  This is certainly a real error and a great fix.

There's an unfortunate complexity because the offer drivers have been rewritten in a branch that is currently headed towards main which would break these fixes immediately.  To avoid that headache of a merge conflict, I've taken your test and copied it over to that branch, then re-implemented your fixes on top of the changes, adapted to how the new offer driver will work.  See #11554.

Since offer aggregation is not technically a chia supported feature yet, I'm inclined to close this PR and leave this broken until that branch makes its way into a release.

PR #11421, opened 2022-05-03
optimize hash_coin_list()

arvidn:
by sorting pure coin_ids, rather than Coins that need to compute its name multiple times during sorting, and also by sorting the actual binary coin names, not hexadecimal strings

### DIFF
--- a/chia/consensus/block_creation.py
+++ b/chia/consensus/block_creation.py
@@ -15,7 +15,7 @@ from chia.consensus.constants import ConsensusConstants
 from chia.consensus.cost_calculator import NPCResult
 from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions
 from chia.full_node.signage_point import SignagePoint
-from chia.types.blockchain_format.coin import Coin, hash_coin_list
+from chia.types.blockchain_format.coin import Coin, hash_coin_ids
 from chia.types.blockchain_format.foliage import Foliage, FoliageBlockData, FoliageTransactionBlock, TransactionsInfo
 from chia.types.blockchain_format.pool_target import PoolTarget
 from chia.types.blockchain_format.proof_of_space import ProofOfSpace
@@ -191,8 +191,9 @@ def create_foliage(
             tx_additions.append(coin)
             byte_array_tx.append(bytearray(coin.puzzle_hash))
         for coin in removals:
-            tx_removals.append(coin.name())
-            byte_array_tx.append(bytearray(coin.name()))
+            cname = coin.name()
+            tx_removals.append(cname)
+            byte_array_tx.append(bytearray(cname))
 
         bip158: PyBIP158 = PyBIP158(byte_array_tx)
         encoded = bytes(bip158.GetEncoded())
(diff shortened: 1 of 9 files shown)


chia/types/blockchain_format/coin.py
```diff
@@ -53,11 +53,11 @@ def __bytes__(self) -> bytes:  # pylint: disable=E0308
         assert False
 
 
-def hash_coin_list(coin_list: List[Coin]) -> bytes32:
-    coin_list.sort(key=lambda x: x.name_str, reverse=True)
```
arvidn, line 57: this is the main change in this PR. note how this list sorts `Coins`, and each time the sorting algorithm is comparing two coins, we compute the coin_id (which is not trivial) and then turn it into a *hex string*, to compare.

This patch converts the list of coins to a list of `bytes32` objects instead, and we sort them directly. No need to recompute them every time we need to compare them. There's also no point in generating a string, and comparing strings.

chia/consensus/block_creation.py
```diff
@@ -191,8 +191,9 @@ def create_foliage(
             tx_additions.append(coin)
             byte_array_tx.append(bytearray(coin.puzzle_hash))
         for coin in removals:
-            tx_removals.append(coin.name())
-            byte_array_tx.append(bytearray(coin.name()))
+            cname = coin.name()
```
arvidn, line 194: this is to save one call to `name()`

chia/types/blockchain_format/coin.py
arvidn: this new version of this function does not convert the name to a hex string before sorting. In order to ensure that there's no unexpected reason the hex string would sort different than the binary name, there's a test here: https://github.com/Chia-Network/chia-blockchain/pull/11421/files#diff-9e2f2e07ef5b18a6367ed8a05b8c369930af75267ce08c0229d0ed6da9bad7acR27

altendky: https://docs.python.org/3/library/stdtypes.html#list.sort
> The key corresponding to each item in the list is calculated once and then used for the entire sorting process.

But in general I'm all for using the proper form of the data (bytes) instead of a serialization or human representation form (hexadecimal string).

arvidn: interesting. I would expect the common case is that you pick one field in your type in the lambda. But that suggests that the sort function still copies that field just to avoid invoking the lambda multiple times.

chia/consensus/block_creation.py
altendky: I wouldn't have asked a comment to be added but if it is worth an explanation here in the context of an optimization PR then maybe it is worth a comment in the code.

Also, `coin_name` isn't that hard to write in three places.  `:]`

tests/core/test_coins.py
```diff
@@ -0,0 +1,31 @@
...
+
+    for i in permutations([A, B, C, D, E, F]):
+        assert hash_coin_ids(list(i)) == expected
+
+
+def test_sorting() -> None:
+    for _ in range(5000):
+        h1 = rand_hash()
+        h2 = rand_hash()
+        assert (h1 < h2) == (h1.hex() < h2.hex())
```
altendky, line 31: I hesitate to bring in randomness without the reproducibility etc that [hypothesis](https://hypothesis.works/) would give.  I've only used it a little bit.  I'm also on the fence as to whether this is a good test to keep around.  It seems more like an initial exploration and ongoing example since it does really test our code at all, other than kind of `rand_hash()`.

altendky (approved):
Just a few thoughts to consider.

chia/consensus/block_creation.py
arvidn: the main reason I pointed it out here is because this change is (somewhat) unrelated to the main change. but it's a bit along the same theme

chia/types/blockchain_format/coin.py
```diff
@@ -53,11 +53,11 @@ def __bytes__(self) -> bytes:  # pylint: disable=E0308
...
+def hash_coin_ids(coin_ids: List[bytes32]) -> bytes32:
+    coin_ids.sort(reverse=True)
     buffer = bytearray()
 
-    for coin in coin_list:
-        buffer.extend(coin.name())
+    for name in coin_ids:
+        buffer.extend(name)
 
     return std_hash(buffer)
```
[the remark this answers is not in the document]
arvidn: oh. it seems like a mistake to me to always construct a new `bytes` object from the input. but also, does a list of `bytes` even support being converted to `bytes` and are they just concatenated? I wouldn't have thought a list would be allowed as input.

tests/core/test_coins.py
arvidn: I agree, it probably doesn't make sense to keep this around. I wanted to demonstrate this property in the PR somehow, and putting this in with a green CI seemed like a reasonable way to do that.

chia/types/blockchain_format/coin.py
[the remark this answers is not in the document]
altendky: Bother, I missed a layer for some of my comments.  Sorry.  Ok, so `std_hash(b for coin_id in coin_ids for b in coin_id)` to flatten via a generator.  Or `std_hash(itertools.chain(*coin_ids))`.  And I stand by my claim that this may or may not be better.  `:]`

PR #11424, opened 2022-05-04
more merkle tree tests

arvidn:
This adds a test case that causes empty nodes to be included in the merkle tree, as well as a few negative test cases for invalid input

### DIFF
--- a/tests/core/test_merkle_set.py
+++ b/tests/core/test_merkle_set.py
@@ -40,13 +40,40 @@ class TestMerkleSet:
         assert merkle_set.get_root() == merkle_set_reverse.get_root()
 
 
-prefix = bytes([0] * 30)
+def hashdown(buf: bytes) -> bytes32:
+    return bytes32(sha256(bytes([0] * 30) + buf).digest())
+
+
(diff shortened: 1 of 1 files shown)


tests/core/test_merkle_set.py
```diff
@@ -40,14 +40,41 @@ async def test_basics(self, bt):
...
+def hashdown(buf: bytes) -> bytes32:
+    return bytes32(sha256(bytes([0] * 30) + buf).digest())
+
+
+@pytest.mark.asyncio
+async def test_merkle_set_invalid_hash_size():
+    merkle_set = MerkleSet()
+
+    # this is too large
+    with pytest.raises(AssertionError):
```
trepca, line 52: shouldn't these be ValueErrors?

trepca (approved):

tests/core/test_merkle_set.py
arvidn: I think it depends on whether this is considered a programmer error or something that's expected to happen in production. I'm pretty sure it's the former

PR #11429, opened 2022-05-04
simplify SizedBytes and StructStream

altendky:
Draft for:
- [x] Make sure the limits around the integer size checks are well covered by tests.
  - Arvid: maybe one test for the bounds checks could be to try to round-trip the value through struct.pack() and ensure it fails/passes for exactly the same cases the new bounds-checking fails/pass for.
- [x] Make sure we test `SizedBytes.parse()` failure
- [x] Consider `int.to_bytes()` instead of `struct.pack()` (and the other way)
- [x] Specify integers by bit size and signedness not `PACK`
  - Well, parsed the name, but hey
- [x] ~~Consider making `StructStream.to_bytes()` take no parameters and pad it in another way in the couple places that use it~~


chia/util/byte_types.py
```diff
@@ -20,41 +19,28 @@ class SizedBytes(bytes):
...
     @classmethod
     def from_hexstr(cls: Type[_T_SizedBytes], input_str: str) -> _T_SizedBytes:
         if input_str.startswith("0x") or input_str.startswith("0X"):
             return cls.fromhex(input_str[2:])
         return cls.fromhex(input_str)
 
-    def __bytes__(self) -> bytes:
-        f = io.BytesIO()
-        self.stream(f)
-        return bytes(f.getvalue())
```
arvidn, line 56: heh, yeah this seemed quite redundant :)

arvidn (commented):
looks good to me. I get a bit nervous about the bounds checking of the integer types. I think we need really convincing tests for those, to make sure they are right

arvidn:
maybe one test for the bounds checks could be to try to round-trip the value through `struct.pack()` and ensure it fails/passes for exactly the same cases the new bounds-checking fails/pass for.

chia/util/struct_stream.py
```diff
@@ -13,33 +12,43 @@ def __trunc__(self) -> int:
...
-            if value != v1:
-                raise ValueError(f"Value {value} does not fit into {cls.__name__}")
-        except Exception:
-            bits = struct.calcsize(cls.PACK) * 8
-            raise ValueError(
-                f"Value {value} of size {value.bit_length()} does not fit into " f"{cls.__name__} of size {bits}"
-            )
-        return int.__new__(cls, value)
+    def __init__(self, value: Union[str, bytes, SupportsInt, SupportsIndex, SupportsTrunc]) -> None:
+        if not (self.MINIMUM <= self < self.MAXIMUM_EXCLUSIVE):
```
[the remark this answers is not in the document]
altendky: I did quick checks that getting it wrong failed, but I did not review the test cases.  I agree it is worth paying careful attention here.  I'll add that to a 'draft for' list in the OP.

chia/util/byte_types.py
```diff
@@ -20,41 +19,28 @@ class SizedBytes(bytes):
...
-            raise ValueError("bad %s initializer %s" % (cls.__name__, v))
-        return bytes.__new__(cls, v)
+    def __init__(self, v) -> None:
+        if len(self) != self._size:
+            raise ValueError("bad %s initializer %s" % (type(self).__name__, v))
 
     @classmethod
     def parse(cls: Type[_T_SizedBytes], f: BinaryIO) -> _T_SizedBytes:
         b = f.read(cls._size)
-        assert len(b) == cls._size
```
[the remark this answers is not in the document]
altendky: Added to the draft list to review that.

chia/util/struct_stream.py
```diff
@@ -13,33 +12,67 @@ def __trunc__(self) -> int:
         ...
 
 
+packing_strings: Dict[Tuple[int, bool], str] = {
+    (1, False): "!B",
```
mariano54, line 16: Do we need to use PACK? It seems very confusing to me. What's wrong with `int.to_bytes`?

chia/util/struct_stream.py
```diff
@@ -13,48 +11,88 @@ def __trunc__(self) -> int:
...
-        result = cls.parse(f)
-        assert f.read() == b""
-        return result
+    def from_bytes(cls: Type[_T_StructStream], blob: bytes) -> _T_StructStream:  # type: ignore[override]
+        if len(blob) != cls.SIZE:
+            raise ValueError(f"{cls.__name__}.from_bytes() requires {cls.SIZE} bytes but got: {len(blob)}")
+        return cls(int.from_bytes(blob, "big", signed=cls.SIGNED))
+
+    def to_bytes(  # type: ignore[override]
+        self, length: Optional[int] = None, byteorder: Literal["little", "big"] = "big", signed: bool = None
```
arvidn, line 89: is `byteorder` a new feature? It would seem reasonable to me to hard code it to always be "big". Also, I don't understand why `signed` is a parameter here, shouldn't it always be `self.SIGNED`?

tests/core/util/test_streamable.py
```diff
@@ -315,7 +315,7 @@ class TestClassList(Streamable):
         a: List[uint8]
 
     # Does not have the required elements
-    with raises(AssertionError):
+    with raises(ValueError):
```
arvidn, line 318: it seems correct that an invalid length is a `ValueError` rather than an `AssertionError`.

chia/util/struct_stream.py
```diff
@@ -13,48 +11,88 @@ def __trunc__(self) -> int:
...
+    (2, True): "!h",
+    (4, False): "!L",
+    (4, True): "!l",
+    (8, False): "!Q",
+    (8, True): "!q",
+    (16, False): None,
+}
+
+
+def parse_metadata_from_name(cls: Type[_T_StructStream]) -> Type[_T_StructStream]:
```
arvidn, line 27: what do you think about adding a test for this function?

arvidn (commented):
still looking good. I would welcome spitting it into smaller PRs, but I don't think it's critical.

chia/util/struct_stream.py
altendky: I exactly agreed.  I used this as a boundary for limiting expansion of this PR, but that decision can be reconsidered.  Previously `StructStream.to_bytes()` was just inherited from `int.to_bytes()` which of course provides the general control over the serialization such as `byteorder` and `length` and `signedness`.  I'm tempted to just put the `super().to_bytes()` call into `.__bytes__()` instead and leave this definition out so it is unchanged.  Not because the code calling this is obviously good, but just as a boundary for this PR, still.

Anyways, we are calling `.to_bytes()` in a few places.

(length and byte order)
https://github.com/Chia-Network/chia-blockchain/blob/8486311afe4858d5645653d4cbd905c8fcded000/chia/util/streamable.py#L204-L205

(length and byte order)
https://github.com/Chia-Network/chia-blockchain/blob/8486311afe4858d5645653d4cbd905c8fcded000/chia/consensus/coinbase.py#L13-L18

(length and byte order and signedness)
https://github.com/Chia-Network/clvm/blob/fc73cd9dc2fc30a1fd461d0f05af9f9679e042c8/clvm/casts.py#L8-L17
```python
def int_to_bytes(v):
    byte_count = (v.bit_length() + 8) >> 3
    if v == 0:
        return b""
    r = v.to_bytes(byte_count, "big", signed=True)
    # make sure the string returned is minimal
    # ie. no leading 00 or ff bytes that are unnecessary
    while len(r) > 1 and r[0] == (0xFF if r[1] & 0x80 else 0):
        r = r[1:]
    return r
```

`Streamable.write_uint32()` and `Streamable.parse_uint32()` seem to be kind of pointless now.  I'm guessing they had some use historically when maybe we didn't have `StructStream` or something.

If you want me to expand this PR down these paths I can take a look at them.

chia/util/struct_stream.py
altendky: If nothing else, the error cases should be exercised to make sure they work.  At which point I'll probably add some other tests too.

chia/util/struct_stream.py
arvidn: I think I'm confused because I don't see the definition of this `to_bytes()` on the left hand side. Would it make sense, for now, to assert that the parameters that are passed in here matches our own?

mypy.ini
```diff
@@ -17,7 +17,7 @@ no_implicit_reexport = True
...
..._protocol_files,tests.util.db_connection,tests.util.generator_tools_testing,tests.util.keyring,tests.util.key_tool,tests.util.misc,tests.util.network,tests.util.rpc,tests.util.test_full_block_utils,tests.util.test_lock_queue,tests.util.test_network_protocol_files,tests.wallet.cat_wallet.test_cat_lifecycle,tests.wallet.cat_wallet.test_cat_wallet,tests.wallet.cat_wallet.test_offer_lifecycle,tests.wallet.cat_wallet.test_trades,tests.wallet.did_wallet.test_did,tests.wallet.did_wallet.test_did_rpc,tests.wallet.rl_wallet.test_rl_rpc,tests.wallet.rl_wallet.test_rl_wallet,tests.wallet.rpc.test_wallet_rpc,tests.wallet.simple_sync.test_simple_sync_protocol,tests.wallet.sync.test_wallet_sync,tests.wallet.test_bech32m,tests.wallet.test_chialisp,tests.wallet.test_puzzle_store,tests.wallet.test_singleton,tests.wallet.test_singleton_lifecycle,tests.wallet.test_singleton_lifecycle_fast,tests.wallet.test_taproot,tests.wallet.test_wallet,tests.wallet.test_wallet_blockchain,tests.wallet.test_wallet_interested_store,tests.wallet.test_wallet_key_val_store,tests.wallet.test_wallet_user_store,tests.wallet_tools,tests.weight_proof.test_weight_proof,tools.analyze-chain,tools.run_block,tools.test_full_sync]
```
arvidn, line 20: the github UI doesn't really highlight was was removed here, but I assume it was struct stream and sized bytes, yeah?

arvidn (dismissed):

chia/util/struct_stream.py
```diff
@@ -13,48 +11,70 @@ def __trunc__(self) -> int:
         ...
 
 
+def parse_metadata_from_name(cls: Type[_T_StructStream]) -> Type[_T_StructStream]:
+    # TODO: turn this around to calculate the PACK from the size and signedness
+
+    name_signedness, _, name_bit_size = cls.__name__.partition("int")
```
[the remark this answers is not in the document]
altendky: No, that's why it bothers to return the delimiter.  I catch that case in the general regeneration of the 'proper' name below which is compared with the actual.  I believe cases without `int` in them are reasonably tested.  Let me know if you see a hole in the testing.

mypy.ini
altendky: I made sure that the six files being touched here were removed.  Just double checking now I see `chia.util.byte_types`, `chia.util.ints`, `chia.util.struct_stream`, and `tests.util.test_struct_stream`.  `chia-tests.core.util.test_streamable` appears to have already not been in the list.

chia/util/struct_stream.py
```diff
@@ -13,48 +11,70 @@ def __trunc__(self) -> int:
...
-            v1 = struct.unpack(cls.PACK, struct.pack(cls.PACK, value))[0]
-            if value != v1:
-                raise ValueError(f"Value {value} does not fit into {cls.__name__}")
-        except Exception:
-            bits = struct.calcsize(cls.PACK) * 8
-            raise ValueError(
-                f"Value {value} of size {value.bit_length()} does not fit into " f"{cls.__name__} of size {bits}"
-            )
-        return int.__new__(cls, value)
+    def __init__(self, value: Union[str, bytes, SupportsInt, SupportsIndex, SupportsTrunc]) -> None:
```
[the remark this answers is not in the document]
altendky: I don't see the other comment here, just from chat, but I assume it is in reference to `SizedBytes.__init__()` and the `v` parameter.  For both `bytes` and `int` they handle the value prior to calling `.__init__()`.  Even in a pure Python class, as opposed to one written in C, the `.__new__()` method is called when you `TheClass()`.  It receives the parameters, instantiates the object, and then calls `TheClass.__init__()` passing the new instance and usually also the parameters.  In this case by the time we have gotten to `.__init__()` the instance already exists including having handled the parameters.  So, we have nothing left to do with the parameter.

But...  my understanding is that type checking (mypy, IDE, etc) will still look to the `.__init__()` for the hinting.  Also, `.__new__()` is passing the value on to `.__init__()`.  So, we need to keep this unused parameter around.

chia/util/byte_types.py
```diff
@@ -20,51 +19,31 @@ class SizedBytes(bytes):
 
     _size = 0
 
-    @staticmethod
-    def __new__(cls: Type[_T_SizedBytes], v) -> _T_SizedBytes:
-        v = bytes(v)
-        if not isinstance(v, bytes) or len(v) != cls._size:
-            raise ValueError("bad %s initializer %s" % (cls.__name__, v))
-        return bytes.__new__(cls, v)
+    def __init__(self, v: Union[Iterable[int]]) -> None:
```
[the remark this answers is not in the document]
altendky: Ah, here's the comment...  outdated already so I missed it when looking at the comments in the diff.  Explanation at https://github.com/Chia-Network/chia-blockchain/pull/11429#discussion_r869555746.  And yes, this was weird and non-obvious to me as well.  I'll at least add summary comments.

arvidn (dismissed):

mariano54 (commented):
Very nice changes

arvidn (approved):

PR #11432, opened 2022-05-04
Replace MerkleSet with the rust implementation

arvidn:
of computing the merkle root. This saves about 20% of the CPU work in the main thread when validating blocks.

The new implementation of the merkle set root function is found here:

   https://github.com/Chia-Network/chia_rs/blob/main/src/merkle_set.rs

### DIFF
--- a/chia/consensus/block_creation.py
+++ b/chia/consensus/block_creation.py
@@ -5,6 +5,7 @@ from typing import Callable, Dict, List, Optional, Tuple
 
 import blspy
 from blspy import G1Element, G2Element
+from chia_rs import compute_merkle_set_root
 from chiabip158 import PyBIP158
 
 from chia.consensus.block_record import BlockRecord
@@ -28,7 +29,6 @@ from chia.types.generator_types import BlockGenerator
 from chia.types.unfinished_block import UnfinishedBlock
 from chia.util.hash import std_hash
 from chia.util.ints import uint8, uint32, uint64, uint128
-from chia.util.merkle_set import MerkleSet
 from chia.util.prev_transaction_block import get_prev_transaction_block
 from chia.util.recursive_replace import recursive_replace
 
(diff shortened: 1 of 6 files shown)


chia/consensus/block_creation.py
```diff
@@ -216,11 +211,11 @@ def create_foliage(
 
         # Addition Merkle set contains puzzlehash and hash of all coins with that puzzlehash
         for puzzle, coin_ids in puzzlehash_coin_map.items():
-            addition_merkle_set.add_already_hashed(puzzle)
-            addition_merkle_set.add_already_hashed(hash_coin_ids(coin_ids))
+            additions_names.append(puzzle)
```
mariano54, line 214: the puzzle_hash is not an addition_name, but it is part of the addition merkle set

chia/consensus/block_creation.py
```diff
@@ -5,6 +5,7 @@
 
 import blspy
 from blspy import G1Element, G2Element
+from chia_rs import compute_merkle_root
```
mariano54, line 8: might want to rename to `compute_merkle_set_root`. Merkle root implies merkle tree which is the main data structure most people use. But this is not a merkle tree

chia/consensus/block_creation.py
arvidn: yeah. what's a better name for this list? `additions_merkle_items`?

mariano54: yeah that sounds good

mariano54 (approved):
As long as is syncs with the blockchain, I think it's a safe enough change

arvidn:
the long-sync test passed:

```
$ python ./tools/test_full_sync.py run ~/.chia/mainnet/db/blockchain_v2_mainnet.sqlite
PID: 3610048
Chia directory /tmp/tmpzqwe4k1l
Can't find private CA, creating a new one in /tmp/tmpzqwe4k1l to generate TLS certificates

height 100000 0.01 s/block
height 200000 0.01 s/block
height 300000 0.02 s/block
height 400000 0.01 s/block
height 500000 0.04 s/block
height 600000 0.03 s/block
height 700000 0.03 s/block
height 800000 0.05 s/block
height 900000 0.04 s/block
height 1000000 0.06 s/block
height 1100000 0.02 s/block
height 1200000 0.03 s/block
height 1300000 0.05 s/block
height 1400000 0.02 s/block
height 1500000 0.02 s/block
height 1600000 0.11 s/block
height 1700000 0.09 s/block
height 1800000 0.12 s/block
height 1900000 0.09 s/block
height 1937600 0.08 s/block   closing full node
```

PR #11433, opened 2022-05-04
New nfts driver

trepca:


trepca:
> when the time comes, this won't work as the -24 won't hit the blockchain as it is an invalid condition - I suggest running the innerpuz and innersolution and scanning for -24 in the results of that instead.

wouldn't full solution still contain those conditions as it won't get run to get them excluded? (I extract the conditions from the inner_solution in the solution)

trepca:
> when the time comes, this won't work as the -24 won't hit the blockchain as it is an invalid condition - I suggest running the innerpuz and innersolution and scanning for -24 in the results of that instead.

wouldn't full solution still contain those conditions as it won't get run to get them excluded? (I extract the conditions from the inner_solution in the solution)

trepca:
> if this changes to `(c CREATE_COIN (c (nft_state_layer_puzzle_hash NFT_STATE_LAYER_MOD_HASH METADATA METADATA_UPDATER_PUZZLE_HASH (f (r (f conditions)))) (c my_amount (r (r (f conditions))))))` then any future appendages to create_coin are supported too
> 
> good spot with the missing hint though

good point, fixed

chia/rpc/wallet_rpc_api.py
```diff
@@ -1289,91 +1306,56 @@ async def did_transfer_did(self, request):
...
         nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
-        address = request["artist_address"]
+        address = request.get("artist_address")
         if isinstance(address, str):
-            address = decode_puzzle_hash(address)
+            puzzle_hash = decode_puzzle_hash(address)
+        elif address is None:
+            puzzle_hash = await nft_wallet.standard_wallet.get_new_puzzlehash()
+        else:
+            puzzle_hash = address
```
paninaro, line 1320: Is it possible to hit this condition? I would think that the address could only be either a string or None to successfully be marshalled through the RPC interface. I would raise an exception in this case.

trepca: previous RPC was sending over bytes to pass puzzle hash and string to pass address, I added the None condition but it doesn't _need_ to be there, it's just the way I was writing unit tests for RPC

chia/rpc/wallet_rpc_api.py
```diff
@@ -1289,91 +1306,56 @@ async def did_transfer_did(self, request):
...
+        wallet_id = uint32(request["wallet_id"])
+        assert self.service.wallet_state_manager
         nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
         nfts = nft_wallet.get_current_nfts()
-        nft_info_list = []
+        nft_uri_pairs = []
         for nft in nfts:
-            nft_info_list.append(get_nft_info_from_puzzle(nft.full_puzzle, nft.coin))
-        return {"wallet_id": wallet_id, "success": True, "nft_list": nft_info_list}
+            uri = get_uri_list_from_puzzle(nft.full_puzzle)
```
paninaro, line 1337: This will break the GUI integration, as the frontend is expecting an array of NFTInfo objects, but we can fix it in a follow-up PR.

chia/wallet/wallet_state_manager.py
```diff
@@ -360,11 +367,13 @@ async def get_unused_derivation_record(
...
+                self.log.debug("No unused paths, generate more ")
                 await self.create_more_puzzle_hashes()
+                # Now we must have unused public keys
+                unused = await self.puzzle_store.get_unused_derivation_path()
+                assert unused is not None
 
-            # Now we must have unused public keys
-            unused = await self.puzzle_store.get_unused_derivation_path()
-            assert unused is not None
+            self.log.debug("Fetching derivation record for: %s %s %s", unused, wallet_id, hardened)
```
paninaro, line 376: nit: we don't seem to be entirely consistent with format strings in the codebase, but generally f-strings are preferred.

chia/rpc/wallet_rpc_api.py
paninaro: I see. So it sounds like there are some RPC callers (maybe only in the tests) that don't go through JSON.

paninaro (approved):
lgtm with the caveat that `nft_get_nfts` won't work in the GUI until the RPC has been updated to return NFTInfo objects.

chia/wallet/wallet_state_manager.py
trepca: f-strings are good, but the problem with using f-string formatting in logging is those variables get evaluated even if logging level is above. See [this](https://github.com/Chia-Network/chia-blockchain/pull/11184).

PR #11440, opened 2022-05-04
Main dids data hash

ytx1991:
The issue is because the data hash is expected to be a hex, but it is a string. Since JSON doesn't support hex number, use string instead


tests/wallet/did_wallet/test_nft_wallet.py
```diff
@@ -110,7 +110,7 @@ async def test_nft_wallet_trade_chia_and_cat(self, three_wallet_nodes, trusted):
         metadata = Program.to(
             [
                 ("u", ["https://www.chia.net/img/branding/chia-logo.svg"]),
-                ("h", 0xD4584AD463139FA8C0D9F68F4B59F185),
+                ("h", "0xD4584AD463139FA8C0D9F68F4B59F185"),
```
arvidn, line 113: I don't think it's a good idea to encode data as an ascii string.

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -142,7 +142,7 @@ def get_nft_info_from_puzzle(puzzle: Program, nft_coin: Coin) -> NFTInfo:
         uncurried_nft.owner_did.as_python(),
         uint64(uncurried_nft.trade_price_percentage.as_int()),
         data_uris,
-        uncurried_nft.data_hash.as_python().hex(),
+        str(uncurried_nft.data_hash.as_python(), "utf-8"),
```
arvidn, line 145: if the data is not valid utf-8, this would throw. There's no test covering this case. (but I think this change is poorly justified to begin with)

arvidn (changes requested):
> The issue is because the data hash is expected to be a hex, but it is a string. Since JSON doesn't support hex number, use string instead

hex *is* a string representation of binary data. The code you're pointing at is encoding binary data as hex in order to ship it as a json string. Can you elaborate on what the problem is?

Are you experiencing some error that you believe is caused by this message encoding?

paninaro:
> > The issue is because the data hash is expected to be a hex, but it is a string. Since JSON doesn't support hex number, use string instead
> 
> hex _is_ a string representation of binary data. The code you're pointing at is encoding binary data as hex in order to ship it as a json string. Can you elaborate on what the problem is?
> 
> Are you experiencing some error that you believe is caused by this message encoding?

The issue I'm seeing is in the JSON representation of the hash, so I would expect the fix to be at the RPC level. I previously used the `nft_mint_nft` RPC to create an NFT with the data hash specified as the string "c520c99955118784734af730590cf74e9c8961b9036b772052597fdb5206ee40" in the JSON payload. Expecting the hash to be provided as a string in the JSON seems correct to me, so I would expect the RPC to convert that string to a byte array (interpreted as hex bytes) if that's what the CLVM is expecting.

Similarly, when calling the `nft_get_nfts` RPC, the corresponding data hash was returned as the string "63353230633939393535313138373834373334616637333035393063663734653963383936316239303336623737323035323539376664623532303665653430". It may be that this code is already doing the right thing, and that the conversion only needs to occur in the `nft_mint_nft` RPC, but I haven't done a full accounting.

paninaro:
Maybe it's as simple as making this change:

image

ytx1991:
There are two problems.
1) Like Jeff explained, this is a string in the JSON payload, but I agree it can be covert to bytes at RPC level. 
2) If the fields in the metadata can be any type then people who want to uncurry it needs to know this info, since different types have different uncurry ways. 

From my POV,  it's better to use a general type for all fields in stead of arbitrary types. If we want to support multiples type in the metadata, I suggest to add a type info field to indicate the way of uncurry and casting.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -199,14 +200,28 @@ async def puzzle_solution_received(self, coin_spend: CoinSpend, in_transaction:
...
-        matched, singleton_curried_args, curried_args = nft_puzzles.match_nft_puzzle(puzzle)
-        if matched:
-            (_, metadata, metadata_updater_puzzle_hash, inner_puzzle) = curried_args
-            params = singleton_curried_args.first()
-            self.log.info(f"found the info for NFT coin {coin_name} {inner_puzzle} {params}")
-            singleton_id = bytes32(params.rest().first().atom)
+        try:
+            uncurried_nft = UncurriedNFT.uncurry(puzzle)
+        except Exception:
+            # The parent is not an NFT which means we need to scrub all of its children from our DB
```
arvidn, line 206: I don't see a test covering this case

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -212,7 +212,7 @@ async def test_nft_wallet_rpc_creation_and_list(two_wallet_nodes: Any, trusted:
         {
             "wallet_id": nft_wallet_0_id,
             "artist_address": ph,
-            "hash": 0xD4584AD463139FA8C0D9F68F4B59F184,
+            "hash": "0xD4584AD463139FA8C0D9F68F4B59F184",
```
arvidn, line 215: I don't understand your rationale for changing this from a (compact and efficient) binary representation to a (slow and inflated) ascii string representation.

Even if there's a really good reason to represent this hash in ascii-form, surely you wouldn't need the `0x` prefix.

arvidn: I may be misunderstanding something actually, I was thinking that this was part of the smart contract, but this is the API to our RPC client, right? So this isn't really in a place where we have the ambition to be efficient, since we already use JSON at this level. And presumably the client class will turn this hash into a string when encoding it as JSON anyway.

However, I think there's a consistency argument to be made. The "artist_address" is a puzzle hash, fundamentally the same type as this "hash". They should probably use the same representation

arvidn:
please make sure to have unit tests covering the new cases you're adding, and ideally covering other cases that lack tests as well. If the code needs to be refactored in order to be tested, I think that's fine.

tests/wallet/nft_wallet/test_nft_wallet.py
paninaro: That's correct, the tests are bypassing the normal JSON serialization/deserialization that would be applied in `RpcServer`, so the onus is on the test writer to ensure that the requests/responses are JSON compatible. In this case, the hash(es) being supplied as raw bytes is incorrect, as that isn't directly possible with JSON. Ideally we should come up with a test fixture (or something) that can wrap an RPC API object instance and enforce JSON compatibility.

I agree that we should be consistent, using strings to pass around hashes (puzzle hashes, data hashes, etc.)

chia/wallet/wallet_state_manager.py
```diff
@@ -569,10 +569,12 @@ async def determine_coin_type(
         # Check if the coin is a NFT
         #                                                        hint
         # First spend where 1 mojo coin -> Singleton launcher -> NFT -> NFT
-        nft_matched, _, nft_curried_args = match_nft_puzzle(Program.from_bytes(bytes(coin_spend.puzzle_reveal)))
-        self.log.debug("Matching NFT: %s", nft_matched)
-        if nft_matched:
-            return await self.handle_nft(coin_spend, iter(nft_curried_args))
+        try:
+            uncurried_nft = UncurriedNFT.uncurry(Program.from_bytes(bytes(coin_spend.puzzle_reveal)))
+        except Exception:
```
trepca, line 574: lets be a bit more specific here, as can easily lead to very hard to debug bugs

trepca (changes requested):

chia/wallet/nft_wallet/nft_wallet.py
ytx1991: After a deep look, I think this case is impossible. Since we already checked and successfully uncurried NFT before this point. I will just remove this exception section

chia/rpc/wallet_rpc_api.py
```diff
@@ -1321,22 +1321,21 @@ async def nft_mint_nft(self, request) -> Dict:
...
         nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
         nfts = nft_wallet.get_current_nfts()
-        nft_uri_pairs = []
+        nft_info_list = []
         for nft in nfts:
-            uri = get_uri_list_from_puzzle(nft.full_puzzle)
-            nft_uri_pairs.append((nft, uri))
-        return {"wallet_id": wallet_id, "success": True, "nft_list": nft_uri_pairs}
+            nft_info_list.append(get_nft_info_from_puzzle(nft.full_puzzle, nft.coin))
+        return {"wallet_id": wallet_id, "success": True, "nft_list": nft_info_list}
```
trepca, line 1338: @paninaro do you need "coin_id" for current instance of the NFT here?

PR #11445, opened 2022-05-05
flag lgtm issues in test and benchmark code

arvidn:
It's important to not have bugs in our tests, otherwise it may not be testing what we think it is.

the lgtm config is a bit unintuitive (I think). One prerequisite to understand this change is that lgtm has a heuristic to classify files as benchmarks and tests. Furthermore, *by default* it hides alerts found in those kinds of files.

What this patch does is that it *excludes* the specified directories from being classified as tests or benchmarks respectively. By not being classified, the alerts in those files are shown.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,3 +1,9 @@
+path_classifiers:
+  test:
+    - exclude: /
+  benchmark:
+    - exclude: /
+
 extraction:
   javascript:
     index:


altendky (commented):
How about we just exclude all of our sources from both?  Such as this list we use for black.  Instead of trying to guess which sources it is likely to identify as what type of code.

https://github.com/Chia-Network/chia-blockchain/blob/1b2d119cc340b34d9dc5864389e37e0ad49e1050/pyproject.toml#L15

lgtm.yml
```diff
@@ -1,3 +1,10 @@
+path_classifiers:
+  test:
+    - exclude: tests
+    - exclude: tools
+  benchmark:
+    - exclude: benchmarks
```
altendky, line 6: I see `test` at https://lgtm.com/help/lgtm/file-classification#built-in-tags but not `benchmark`.  Where is `benchmark` documented?

arvidn: I just see the tag in the file browser: https://lgtm.com/projects/g/Chia-Network/chia-blockchain/latest/files/benchmarks/?sort=name&dir=ASC&mode=heatmap

altendky:
I guess this is a server-side-only tool that probably doesn't make a venv in the project directory so yeah, in this case excluding the root ought to be ok.

altendky (approved):

PR #11447, opened 2022-05-05
Serializing NFT puzzle

trepca:


chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -56,6 +56,31 @@ class NFTCoinInfo(Streamable):
...
+                    # hash value
+                    value = values[0].hex()
+                else:
+                    value = [x.decode("utf-8") for x in values]
+                metadata_dict[key.decode("utf-8")] = value
+            return {
+                "nft_id": singleton_id,
+                "metadata": metadata_dict,
+                "metadata_updater_puzzle_hash": metadata_updater_puzzle_hash.atom.hex(),
+                "inner_puzzle_hash": inner_puzzle.get_tree_hash().hex(),
```
paninaro, line 80: Can we add a few of the fields from NFTInfo? These are the ones I think we'll need in the GUI:

`nft_coin_id` (`nft_transfer_nft` expects an `nft_coin_id` param in the request... or is that the launcher_id?)
`royalty`

There's also `version` which might be useful to differentiate between NFT0 and NFT1. And `edition_count`, though I think that field needs more thought -- is it the number of NFTs minted in a series, or is it the serial number within a series?

trepca:
Sure. Also, nft_id == singleton_id == launcher_id

chia/wallet/nft_wallet/nft_wallet.py
paninaro: https://developers.chia.net/t/nft-on-chain-metadata/466

Mentions that there should be both `edition_number` _and_ `edition_count` (where edition_count is optional).

ytx1991:
I'm doing the same thing. I think we should modify the NFTInfo to make it compatible with the new design. We should encapsulate these fields and uncurry logic in one place. Currently this rest() first() usages for uncurry are very hard to maintain and easy to cause bugs. I have created a class UncurriedNFT for handling all uncurry things.

trepca:
@ytx1991 will take over this

PR #11463, opened 2022-05-06
Uses the new `from_bytes_unchecked` method in blspy, to improve perfo…

mariano54:
…rmance. Based on this change: https://github.com/Chia-Network/bls-signatures/pull/328. 
FullBlock.from_bytes improved from 2600 µs to 954 µs.

TODO: 
- [x] Update blspy version (windows fix) once released
- [x] Consider changing from_json as well

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -239,10 +239,13 @@ def parse_tuple(f: BinaryIO, list_parse_inner_type_f: List[ParseFunctionType]) -
     return tuple(full_list)
 
 
-def parse_size_hints(f: BinaryIO, f_type: Type[Any], bytes_to_read: int) -> Any:
+def parse_size_hints(f: BinaryIO, f_type: Type[Any], bytes_to_read: int, unchecked: bool) -> Any:
     bytes_read = f.read(bytes_to_read)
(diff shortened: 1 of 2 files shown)


arvidn:
yeah, I can run a test

arvidn:
My sync test (with full blocks) does not measure any improvement with this patch. But looking closely, I don't think that's surprising. This patch would improve performance of deserializing an object with many BLS elements in it. But I don't think that's very common. Even a `FullBlock` doesn't have very many of those. Most of them come out of the generator program, and they are constructed and used basically at the same time (so the sepatation of performing the subgroup check and checking signature won't matter in that context).

It seems reasonable to me to make this change, but it might be a good idea to have some idea of what circumstance it matters in.

mariano54:
This patch would have an effect if we are passing around public keys and parsing them multiple times, before validation. After this patch, each public key will only get checked for validity once, even if it is parsed multiple times. It might be that we optimized everything so much that this is not necessary (since we keep pks around as `bytes48` instead of G1Element). However it's still a good change to have, since we can now use G1Elements with less fear of performance impact

arvidn:
I agree

arvidn (dismissed):

arvidn (approved):

PR #11468, opened 2022-05-09
harvester: Introduce `recursive_plot_scan`

xdustinface:
This adds the new config option `harvester.recursive_plot_scan` which if enabled (set to `True`) tells the harvester to scan for plots recursively in the directories in `harvester.plot_directories`.

### DIFF
--- a/chia/plotting/util.py
+++ b/chia/plotting/util.py
@@ -121,20 +121,9 @@ def get_filenames(directory: Path, recursive: bool) -> List[Path]:
         return []
     all_files: List[Path] = []
     try:
-        if recursive:
-            all_files = [
-                child
-                for child in directory.rglob("*.plot")
-                if child.is_file() and child.suffix == ".plot" and not child.name.startswith("._")
-            ]
-        else:
-            for child in directory.iterdir():
-                if not child.is_dir():
-                    # If it is a file ending in .plot, add it - work around MacOS ._ files
-                    if child.suffix == ".plot" and not child.name.startswith("._"):
-                        all_files.append(child)
-                else:
-                    log.debug(f"Not checking subdirectory {child}, subdirectories not added by default")
+        glob_function = directory.rglob if recursive else directory.glob
+        all_files = [child for child in glob_function("*.plot") if child.is_file() and not child.name.startswith("._")]
+        log.debug(f"get_filenames: {len(all_files)} files found in {directory}, recursive: {recursive}")
     except Exception as e:
         log.warning(f"Error reading directory {directory} {e}")
     return all_files
--- a/chia/plotting/util.py
+++ b/chia/plotting/util.py
@@ -70,9 +70,11 @@ def get_plot_directories(root_path: Path, config: Dict = None) -> List[str]:
 def get_plot_filenames(root_path: Path) -> Dict[Path, List[Path]]:
     # Returns a map from directory to a list of all plots in the directory
     all_files: Dict[Path, List[Path]] = {}
-    for directory_name in get_plot_directories(root_path):
+    config = load_config(root_path, "config.yaml")
+    recursive_scan: bool = config["harvester"].get("recursive_plot_scan", False)
+    for directory_name in get_plot_directories(root_path, config):
         directory = Path(directory_name).resolve()
-        all_files[directory] = get_filenames(directory)
+        all_files[directory] = get_filenames(directory, recursive_scan)
     return all_files
 
 
@@ -109,7 +111,7 @@ def remove_plot(path: Path):
         path.unlink()
 
 
-def get_filenames(directory: Path) -> List[Path]:
+def get_filenames(directory: Path, recursive: bool) -> List[Path]:
     try:
         if not directory.exists():
             log.warning(f"Directory: {directory} does not exist.")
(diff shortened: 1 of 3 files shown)


chia/plotting/util.py
```diff
@@ -119,13 +121,20 @@ def get_filenames(directory: Path) -> List[Path]:
...
-                # If it is a file ending in .plot, add it - work around MacOS ._ files
-                if child.suffix == ".plot" and not child.name.startswith("._"):
-                    all_files.append(child)
-            else:
-                log.debug(f"Not checking subdirectory {child}, subdirectories not added by default")
+        if recursive:
+            all_files = [
+                child
+                for child in directory.rglob("*.plot")
+                if child.is_file() and child.suffix == ".plot" and not child.name.startswith("._")
```
altendky, line 128:
```suggestion
                if child.is_file() and not child.name.startswith("._")
```

chia/util/initial-config.yaml
```diff
@@ -195,6 +195,7 @@ harvester:
 
   # Plots are searched for in the following directories
   plot_directories: []
+  recursive_plot_scan: False # If True the harvester scans plots recursively in the provided directories.
```
altendky, line 198: I wonder if this is better or if we should allow `/**` on the end of directories for recursion selection or something.

chia/plotting/util.py
```diff
@@ -119,13 +121,20 @@ def get_filenames(directory: Path) -> List[Path]:
...
+            ]
+        else:
+            for child in directory.iterdir():
+                if not child.is_dir():
+                    # If it is a file ending in .plot, add it - work around MacOS ._ files
+                    if child.suffix == ".plot" and not child.name.startswith("._"):
+                        all_files.append(child)
+                else:
+                    log.debug(f"Not checking subdirectory {child}, subdirectories not added by default")
```
altendky, line 137: I guess this logging is the only reason we need two separate mechanisms here, otherwise it is just `.rglob()` vs. `.glob()`.  `:|`  Though this message should probably reference the new feature now.  And maybe it should log that it is recursing when doing so.  It is a debug message after all.

```suggestion
                    log.debug(f"Subdirectory not searched for plots, see setting harvester / recursive_plot_scan: {child}")
```

Note that bringing these two paths back to one might be done via https://docs.python.org/3.10/library/os.html#os.walk.  It has a really weird but useful interface.  As you iterate it you can modify the results such as removing directories from its list to tell it to not traverse them.

Or, just use this function recursively on recursive subdirectory configs.

tests/plotting/test_plot_manager.py
```diff
@@ -606,3 +606,41 @@ def raising_callback(event: PlotRefreshEvents, _: PlotRefreshResult):
...
+    # Set the recursive scan flag in the config
+    with lock_and_load_config(env.root_path, "config.yaml") as config:
+        config["harvester"]["recursive_plot_scan"] = True
+        save_config(env.root_path, "config.yaml", config)
+
+    # With the flag enabled it should load all expected plots
+    expected_result.loaded = expected_plot_list  # type: ignore[assignment]
+    expected_result.removed = []
+    expected_result.processed = len(expected_plot_list)
+    expected_result.remaining = 0
```
altendky, line 639: Instead of maximal mutation and reuse, just create the instance you want?  It feels like this class could be frozen as well, though I didn't careful inspect all uses.  If frozen, tuples instead of lists would be nice.

tests/plotting/test_plot_manager.py
```diff
@@ -606,3 +606,41 @@ def raising_callback(event: PlotRefreshEvents, _: PlotRefreshResult):
...
+    await env.refresh_tester.run(expected_result)
+
+    # Adding the subdirectories also should not lead to some failure or duplicated loading
```
altendky, line 642: The plots are loaded into a dict based on the configuration path entry.  Do we care which one the plots show up under?  I feel like they should, maybe, rather be in the most specific path they match but I'm not sure it really matters in terms of actual usage.

altendky (changes requested):
I think there should be one block of code handling both the recursive and non-recursive plot search cases.  There are a few possible paths for that that are likely ok.

tests/plotting/test_plot_manager.py
xdustinface: Doesn't matter, see https://github.com/Chia-Network/chia-blockchain/blob/main/chia/plotting/manager.py#L140-L141.

chia/plotting/util.py
xdustinface: Makes sense, see bf0cd5b5cebb3874eac2689cb913d0d30e045e29. I merged the lookups and changed the log a bit. I guess since its `DEBUG` anyway its not super important to keep this "warning".

tests/plotting/test_plot_manager.py
xdustinface: I dropped the two excessive assignments. Now this way you at least see what should change from the last test run. Its done like this in all other tests too so if you want to have it frozen which seems unrelated here, you can PR it and change it everywhere :)

chia/plotting/util.py
xdustinface: See bf0cd5b5cebb3874eac2689cb913d0d30e045e29.

chia/util/initial-config.yaml
altendky: And less flexible, yep.  Probably usually not an issue.

tests/plotting/test_plot_manager.py
altendky: Agreed, any change to being frozen should be separate.

altendky (approved):

antonmakarenko:
Did it make into release 1.3.5? I tested this on config change and my farm cannot find any plots

antonmakarenko:
Did it make into release 1.3.5? I tested this on config change and my farm cannot find any plots

emlowe:
No, it is not in 1.3.5

PR #11469, opened 2022-05-09
added state change events in nft wallet

trepca:


paninaro (approved):
Applied locally and confirmed that these state changes update the GUI appropriately. A newly minted NFT appears in the NFT gallery automatically, and transferring an NFT causes the NFT to disappear (after the txn is confirmed).

PR #11480, opened 2022-05-10
add long_lived/** and release/** to (almost) all workflow push trigge…

altendky:
…rs (#11439)

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -3,7 +3,9 @@ name: Benchmarks
 on:
   push:
     branches:
+      - 'long_lived/**'
       - main
+      - 'release/**'
     tags:
       - '**'
   pull_request:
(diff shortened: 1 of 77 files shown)


arvidn (approved):
are we not building everything on `releases/` branches?

arvidn:
I suppose the installers are built on push to tags, that's why we don't build those on `releases/*` branches. is that right?

altendky:
Note that this is a cherry pick from the merge of https://github.com/Chia-Network/chia-blockchain/pull/11439 onto `atari`.  I think just publishing should be gated on being a tag, and maybe also gated on being on the `main` branch.  All installer building and tests and checks and benchmarks and...  are intended to run for all PRs, `main`, `long_lived/**`, and `release/**`.

I think your comments end up being a request to cherry pick 3abcbf1df as well which was applied to `main` prior to the commit this is cherry picking.  I'm a bit tempted to just leave this as is and get the other stuff in the next catch up merge since it isn't relevant to `atari`.  But, we'll see when I think it through.

PR #11486, opened 2022-05-11
Fix Newer block issue

mariano54:
In blockchain.py, the order of making updates is:
1. add block record to DB
2. add block record to memory
3. set peak height
4. commit DB transaction

The problem that we saw, might have been related to making a DB request for the peak block between 3 and 4. If instead of fetching from the DB, we fetch the block record from memory, we are guaranteed that it will be in memory

### DIFF
--- a/chia/rpc/full_node_rpc_api.py
+++ b/chia/rpc/full_node_rpc_api.py
@@ -468,9 +468,16 @@ class FullNodeRpcApi:
         newer_block_bytes = bytes32.from_hexstr(newer_block_hex)
         older_block_bytes = bytes32.from_hexstr(older_block_hex)
 
-        newer_block: BlockRecord = self.service.blockchain.block_record(newer_block_bytes)
-        older_block: BlockRecord = self.service.blockchain.block_record(older_block_bytes)
-
+        newer_block = await self.service.block_store.get_block_record(newer_block_bytes)
+        if newer_block is None:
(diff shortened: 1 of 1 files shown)


mariano54:
Actually this is not correct, we only keep recent block records in memory. Making another fix shortly

hoffmang9:
Just started running this - will update if I see the intermittent issue.

chia/rpc/full_node_rpc_api.py
```diff
@@ -470,7 +470,11 @@ async def get_network_space(self, request: Dict) -> Optional[Dict]:
 
         newer_block = await self.service.block_store.get_block_record(newer_block_bytes)
         if newer_block is None:
-            raise ValueError(f"Newer block {newer_block_hex} not found")
+            # It's possible that the peak block has not yet been committed to the DB, so as a fallback, check memory
+            try:
+                newer_block = self.service.blockchain.block_record(newer_block_bytes)
```
arvidn, line 475: @mariano54 is this a more expensive call? does it need to lock the blockchain mutex? I assume there's some reason to not reach for this call in the first place

arvidn (approved):
do you have a theory of why this started happening?

mariano54:
> do you have a theory of why this started happening?

Probably related to one of the many changes we made to the `receive_block` method, for example reordering of statements, DBWrapper, etc

PR #11493, opened 2022-05-11
Add the ability to set fees from CLI for minting NFTs, transferring NFTs and adding URIs to NFTs

AmineKhaldi:


chia/cmds/wallet.py
```diff
@@ -490,13 +490,23 @@ def nft_wallet_create_cmd(wallet_rpc_port: Optional[int], fingerprint: int) -> N
 @click.option("-aa", "--artist-address", help="Artist's backpayment address", type=str, required=True)
 @click.option("-nh", "--hash", help="NFT content hash", type=str, required=True)
 @click.option("-u", "--uris", help="Comma separated list of URIs", type=str, required=True)
+@click.option(
+    "-m",
+    "--fee",
+    help="Set the fees per transaction, in XCH.",
+    type=str,
```
trepca, line 497: I think this should be float?

trepca (changes requested):

chia/cmds/wallet.py
trepca: inner functions for send transaction later on transforms the fee into Decimal actually 
https://github.com/Chia-Network/chia-blockchain/blob/f735c8d7de637e5768b1a246beb9ed9ada25b5e5/chia/cmds/wallet_funcs.py#L190

AmineKhaldi: It's validated by the `validate_fee` callback.

trepca (approved):

PR #11503, opened 2022-05-12
upgrade chia-rs to streamable support

arvidn:
Update `chia_rs` to the latest release which includes Streamable serialization and deserialization support for `PySpend` and `PySpendBundleConditions` classes, which are the output from running a generator program and parsing and validating the conditions. (those classes have changed names to remove the `Py` prefix and gained type-hints).

In addition to supporting serialization and deserialization, it also adds support to `to_json_dict()`, which we use to print these types (mostly in tests, but also in error logging).

The reason to serialize and deserialize these in rust instead of python is that they are expensive enough to show up in profiles of the main process. The main thread is especially sensitive to using CPU since it's doing all serial work and holds up the parallel block validation in the worker processes.

This change only makes a small impact on wall-clock time of the benchmarks. This is because the main process is dominated by disk I/O. Improving the disk performance is a bigger long-term project that will have to be addressed separately. This patch is primarily concerned with reducing the CPU usage.

Wall clock time:
| test | before | after | after/before |
| --- | --- | --- | --- |
| M1, SSD, sync | 114.04s | 109.73s | 0.96 |

CPU time (`receive_block_batch()`):
| test | before | after | after/before |
| --- | --- | --- | --- |
| M1, SSD, sync | 21.37% | 17.11% | 0.80 |

# Profiles

(`pre_validate_blocks_multiprocessing()` is where we deserialize the generator results from the worker processes)

before:
slow-batch-00448

after:

slow-batch-00448

### DIFF
--- a/chia/types/spend_bundle_conditions.py
+++ b/chia/types/spend_bundle_conditions.py
@@ -1,30 +1,3 @@
-from dataclasses import dataclass
-from typing import List, Optional, Tuple
+from chia_rs import Spend, SpendBundleConditions
 
-from chia.types.blockchain_format.sized_bytes import bytes32, bytes48
-from chia.util.ints import uint32, uint64
-from chia.util.streamable import Streamable, streamable
-
-
-# the Spend and SpendBundleConditions classes are mirrors of native types, returned by
-# run_generator
-@streamable
-@dataclass(frozen=True)
-class Spend(Streamable):
-    coin_id: bytes32
-    puzzle_hash: bytes32
-    height_relative: Optional[uint32]
-    seconds_relative: uint64
-    create_coin: List[Tuple[bytes32, uint64, Optional[bytes]]]
-    agg_sig_me: List[Tuple[bytes48, bytes]]
-
-
-@streamable
-@dataclass(frozen=True)
-class SpendBundleConditions(Streamable):
-    spends: List[Spend]
-    reserve_fee: uint64
-    height_absolute: uint32
-    seconds_absolute: uint64
-    agg_sig_unsafe: List[Tuple[bytes48, bytes]]
-    cost: uint64
+__all__ = ["Spend", "SpendBundleConditions"]
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import io
+import os
 import pprint
 from enum import Enum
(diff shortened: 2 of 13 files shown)


chia/util/streamable.py
```diff
@@ -66,6 +67,7 @@ class DefinitionError(StreamableError):
     "G2Element",
     "Program",
     "SerializedProgram",
+    "PySpendBundleConditions",
```
arvidn, line 70: I believe this is wrong. It was an earlier attempt I made

chia/types/spend_bundle_conditions.py
```diff
@@ -1,30 +1,3 @@
-from dataclasses import dataclass
-from typing import List, Optional, Tuple
+from chia_rs import Spend, SpendBundleConditions
```
arvidn, line 1: the type stubs for these classes are here: https://github.com/Chia-Network/chia_rs/blob/main/wheel/chia_rs.pyi

arvidn:
I ran through a successful sync-from-scratch:

```
$ python ./tools/test_full_sync.py run --profile ~/chia-backup/db/blockchain_v2_mainnet.sqlite
PID: 2009491
Chia directory /tmp/tmp854fo5t9
Can't find private CA, creating a new one in /tmp/tmp854fo5t9 to generate TLS certificates

height 100000 0.01 s/block
height 200000 0.01 s/block
height 300000 0.02 s/block
height 400000 0.02 s/block
height 500000 0.04 s/block
height 600000 0.02 s/block
height 700000 0.03 s/block
height 800000 0.03 s/block
height 900000 0.04 s/block
height 1000000 0.05 s/block
height 1100000 0.02 s/block
height 1200000 0.03 s/block
height 1300000 0.04 s/block
height 1400000 0.02 s/block
height 1500000 0.02 s/block
height 1600000 0.12 s/block
height 1700000 0.09 s/block
height 1772864 0.02 s/block   closing full node
```

richardkiss (approved):
This seems reasonable.

Note that `maturin` now supports mixed-mode wheels that have python alongside the rust-based shared object that's loaded. You might consider investigating that, as it might allow you to push a lot of these wrapping diffs into python in the `chia_rs` wheel (you'd have to push the `bytes32` definition into `chia_rs`, but that's might not be completely terrible).

wjblanke (approved):
aok

PR #11506, opened 2022-05-12
Switch data layer to using the new DBWrapper v2

altendky:
To be dealt with separately
- [ ] Review for closure of cursors, probably use context managers
- [ ] Review for minimal possible code in the context managers
- [ ] Review of data download's need for a context manager or not
  - `# TODO: should this all be in a db wrapper context manager?`


chia/data_layer/data_store.py
```diff
@@ -775,12 +787,11 @@ async def delete(
         hint_keys_values: Optional[Dict[bytes, bytes]] = None,
         use_optimized: bool = True,
         *,
-        lock: bool = True,
         status: Status = Status.PENDING,
     ) -> None:
-        async with self.db_wrapper.locked_transaction(lock=lock):
+        async with self.db_wrapper.write_db():
```
arvidn, line 792: There are several examples of where you just grab the write-lock for the DB, without (seemingly) writing to it. If this is intentional and correct, I really think it warrants comments

chia/data_layer/data_store.py
```diff
@@ -631,17 +650,15 @@ async def autoinsert(
         tree_id: bytes32,
         hint_keys_values: Optional[Dict[bytes, bytes]] = None,
         use_optimized: bool = True,
-        *,
-        lock: bool = True,
     ) -> bytes32:
-        async with self.db_wrapper.locked_transaction(lock=lock):
-            was_empty = await self.table_is_empty(tree_id=tree_id, lock=False)
+        async with self.db_wrapper.write_db():
```
arvidn, line 654: if you don't save the connection object here, why are you acquiring a write lock for the DB?

chia/data_layer/data_store.py
```diff
@@ -930,49 +929,49 @@ async def get_proof_of_inclusion_by_key(
...
-        *,
-        lock: bool = True,
     ) -> ProofOfInclusion:
         """Collect the information for a proof of inclusion of a key and its value in
         the Merkle tree.
         """
-        async with self.db_wrapper.locked_transaction(lock=lock):
-            node = await self.get_node_by_key(key=key, tree_id=tree_id, lock=False)
-            return await self.get_proof_of_inclusion_by_hash(node_hash=node.hash, tree_id=tree_id, lock=False)
+        async with self.db_wrapper.read_db():
```
arvidn, line 936: it seems redundant to ask for a read connection object and then not use it. Is there some subtle reason to do this at all?

chia/data_layer/data_store.py
```diff
@@ -981,32 +980,30 @@ async def get_single_operation(
         root_hash_begin: Optional[bytes32],
         root_hash_end: Optional[bytes32],
         root_status: Status,
-        *,
-        lock: bool = False,
     ) -> Union[InsertionData, DeletionData]:
-        async with self.db_wrapper.locked_transaction(lock=lock):
+        async with self.db_wrapper.read_db():
```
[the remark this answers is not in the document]
altendky: That is exactly the point, yes.  Some of the methods might be called directly or as a subcall from another method.  Just getting the connection where we need to describe a transaction seemed to be the flexible use of the allowed nesting.  I agree there's an amount of nuance here.  I would suggest that we either just have a group discussion to explain this, or provide another method that makes the point explicit and yields probably just `None`.  Remembering to put and maintain comments everywhere seems pretty error prone.

altendky:
Will recreate later.

PR #11517, opened 2022-05-13
log error for failed service start

altendky:
This was initially introduced in the data layer work.  It is a very small step that just makes sure an exception is logged quickly when there is a failure.  Without this the exception is not output until the service is attempted to be shut down and even then it only goes to stdout.  A more complete improvement of the startup process and error handling would be welcome, but I believe this makes a useful improvement to developer awareness of the issue with minimal effort.


chia/rpc/rpc_server.py
```diff
@@ -314,23 +314,28 @@ async def start_rpc_server(
...
+        async def cleanup():
+            await rpc_server.stop()
+            await runner.cleanup()
+            if connect_to_daemon:
+                await daemon_connection
+
+        return cleanup
+    except Exception:
+        tb = traceback.format_exc()
+        log.error(f"Starting RPC server failed. Exception {tb}.")
```
arvidn, line 340: I just learned about ```log.exception("Starting RPC server failed")``` today. I think that would be appropriate here

arvidn: it will print the stack trace of the current exception for you. You just add an additional message

arvidn (dismissed):

altendky:
I want to play around with that and do bulk changeover for it if we do.

arvidn (approved):

PR #11524, opened 2022-05-14
Add RPC API for get any NFT

ytx1991:
The input coin id should be the only valid unspent NFT coin id. The API will fetch the puzzle reveal of its parent coin. For check the history of the NFT, set latest = False. The reason we do in this way is because it is more efficient. Since there is only one parent for a coin but one coin can have multiple children.

### DIFF
--- a/chia/rpc/full_node_rpc_api.py
+++ b/chia/rpc/full_node_rpc_api.py
@@ -17,6 +17,8 @@ from chia.util.byte_types import hexstr_to_bytes
 from chia.util.ints import uint32, uint64, uint128
 from chia.util.log_exceptions import log_exceptions
 from chia.util.ws_message import WsRpcMessage, create_payload_dict
+from chia.wallet.nft_wallet.nft_info import NFTInfo
+from chia.wallet.nft_wallet.nft_puzzles import get_nft_info_from_puzzle
 
 
 def coin_record_dict_backwards_compat(coin_record: Dict[str, Any]):
@@ -57,6 +59,7 @@ class FullNodeRpcApi:
             "/get_coin_records_by_hint": self.get_coin_records_by_hint,
             "/push_tx": self.push_tx,
             "/get_puzzle_and_solution": self.get_puzzle_and_solution,
+            "/get_nft_info": self.get_nft_info,
             # Mempool
             "/get_all_mempool_tx_ids": self.get_all_mempool_tx_ids,
             "/get_all_mempool_items": self.get_all_mempool_items,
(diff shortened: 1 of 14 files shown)


chia/rpc/full_node_rpc_api.py
```diff
@@ -17,6 +17,8 @@
 from chia.util.ints import uint32, uint64, uint128
 from chia.util.log_exceptions import log_exceptions
 from chia.util.ws_message import WsRpcMessage, create_payload_dict
+from chia.wallet.nft_wallet.nft_info import NFTInfo
+from chia.wallet.nft_wallet.nft_puzzles import get_nft_info_from_puzzle
```
paninaro, line 21: I'm not sure `full_node_rpc_api.py` is the right place for this RPC, given that it doesn't currently import anything from `chia.wallet`. There may be a reason for keeping the full node and wallet RPC implementations separate. @mariano54 do you have any objections to adding a dependency on wallet code here?

I can see how this RPC sort of fits with the other similar `get_coin_records_*` RPCs, so I understand the logic behind including it here. Was there some other reason for not including this RPC in `wallet_rpc_api.py`?

mariano54: There should be no wallet specific code in the full node. This doesn't belong here

mariano54: There are many many reasons for this but TLDR is modularity

mariano54 (changes requested):
No wallet code in full node

chia/rpc/wallet_rpc_client.py
```diff
@@ -586,11 +586,11 @@ async def add_uri_to_nft(self, wallet_id, nft_coin_id, uri, fee):
         response = await self.fetch("nft_add_uri", request)
         return response
 
-    async def transfer_nft(self, wallet_id, nft_coin_id, artist_address, fee):
+    async def transfer_nft(self, wallet_id, nft_coin_id, target_address, fee):
```
paninaro, line 589: This change doesn't seem relevant to the RPC addition. Was this included on purpose?

paninaro (changes requested):
Looks good, but the RPC should be moved into the wallet.

paninaro:
> I think this is a general API, it doesn't need any wallet information. Just like other get coin information API. Moving this to the wallet rpc will couple the full node RPC and wallet RPC, since the information is not in the wallet database.

The `check_offer_validity` RPC may be a good example to look at as I think it's conceptually similar to this RPC. That RPC calls into `trade_manager.py` and requests `get_coin_state` from the wallet node. That method abstracts-away the full node dependency.

mariano54:
There is no concept of an "nft" in the full node. The full node only knows about coins, it does not inspect individual puzzles to see what is inside of them. This might be added into the chia dev tools or some other package that calls the full node RPC.

chia/rpc/wallet_rpc_client.py
ytx1991: I think the old name of the variable is confusing, transfer to the artist doesn't looks right to me. The front-end may need some change but I think it's worth to change.

ytx1991:
> > I think this is a general API, it doesn't need any wallet information. Just like other get coin information API. Moving this to the wallet rpc will couple the full node RPC and wallet RPC, since the information is not in the wallet database.
> 
> The `check_offer_validity` RPC may be a good example to look at as I think it's conceptually similar to this RPC. That RPC calls into `trade_manager.py` and requests `get_coin_state` from the wallet node. That method abstracts-away the full node dependency.

I tried this but it doesn't work, it only work for NFTs created by this node.

paninaro:
> I tried this but it doesn't work, it only work for NFTs created by this node.

I'm not sure why that would be the case. When checking the validity of an offer that someone else created, I'm pretty sure it goes through the same code path to fetch the coin states for the offered coins. @Quexington do you know if `get_coin_state` can be used to fetch coin info for an arbitrary coin like an NFT?

paninaro:
As mentioned in chat, there's an offer creation scenario where we only have the `launcher_id` and not the current NFT coin ID.

PR #11527, opened 2022-05-14
add test for streamable -> json conversion

arvidn:
I am about to propose a refactor to `recurse_jsonify()`, so I would like to have tests first

### DIFF
--- a/tests/core/util/test_jsonify.py
+++ b/tests/core/util/test_jsonify.py
@@ -21,42 +21,7 @@ def test_primitives() -> None:
         None,
         "foobar",
         b"\0\1\0\1",
-        bytes32(
-            [
-                0,
-                1,
-                2,
-                3,
-                4,
-                5,
-                6,
-                7,
-                8,
-                9,
-                10,
-                11,
-                12,
-                13,
-                14,
-                15,
-                16,
-                17,
-                18,
-                19,
-                20,
-                21,
-                22,
-                23,
-                24,
-                25,
-                26,
-                27,
-                28,
-                29,
-                30,
-                31,
-            ]
-        ),
+        bytes32(range(32)),
     )
 
     assert t.to_json_dict() == {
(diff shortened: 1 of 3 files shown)


trepca (dismissed):
nesting would be nice to have too (A.b => B.c => C)

arvidn:
this test covers one level on nesting. I wouldn't expect testing three layers would add that much coverage:
https://github.com/Chia-Network/chia-blockchain/pull/11527/files#diff-8e5b980067a2cb9e5545b0d7933776df9dd1412fdabe53bc8b0c075065f49689R93

trepca:
> this test covers one level on nesting. I wouldn't expect testing three layers would add that much coverage: https://github.com/Chia-Network/chia-blockchain/pull/11527/files#diff-8e5b980067a2cb9e5545b0d7933776df9dd1412fdabe53bc8b0c075065f49689R93

I meant nested with a custom class, but I guess custom classes are not much different from Tuple

arvidn:
There the `Inner` class which is nested inside the `NestedTest` class

tests/core/util/test_jsonify.py
```diff
@@ -0,0 +1,113 @@
...
+                24,
+                25,
+                26,
+                27,
+                28,
+                29,
+                30,
+                31,
+            ]
+        ),
```
altendky, line 59: This way I don't have to read every line to know what is happening.

```suggestion
        bytes32(range(32)),
```

altendky:
I think adding a direct nesting would be good as well.  Maybe also nesting through a list.  It may or may not be 'relevant' with the present implementation, but it seems reasonable that there might be differences in some people's implementations that may arise in the future, if not now.

tests/core/util/test_jsonify.py
arvidn: heh, yeah. it didn't look like this before `black` btw :)

altendky (approved):
I appreciate the separate test cases, as opposed to just mixing all the nesting types in one test.

PR #11532, opened 2022-05-14
Work on test_wallet_tx_retry() reliability

altendky:
Draft for:
- [ ] Being more than just diagnostic (hopefully the sleep isn't 'the fix')


mariano54:
Test is fixed here: 
https://github.com/Chia-Network/chia-blockchain/pull/11576

altendky:
I was about to ask about just going with this, but a real fix is even better.  Thanks!

PR #11535, opened 2022-05-15
Block and transaction processing helpers

altendky:
Draft for:
- [x] Discussion about the APIs
- [x] Some manner of timeouts
- [x] ~Consider where `wait_for_coins_in_wallet()` and any other helper functions in `full_node_simulator.py` belong.  https://github.com/Chia-Network/chia-blockchain/pull/11535/files#r879554651~ this can just be a follow up API design PR.
- [x] Review added `TODO:`s


chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +173,219 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+            height = uint32(height_before + count)
+            rewards += calculate_pool_reward(height) + calculate_base_farmer_reward(height)
+
+            if rewards >= amount:
+                await self.farm_blocks(count=count, wallet=wallet)
+                return rewards
+
+        raise Exception("internal error")
+
+    async def wait_transaction_records_entered_mempool(self, records: Collection[TransactionRecord]) -> None:
```
mariano54, line 273: Maybe add a timeout to prevent 60min long CI runs

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +173,219 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+                if tx is not None:
+                    found.add(spend_bundle_name)
+            ids_to_check = ids_to_check.difference(found)
+
+            if len(ids_to_check) == 0:
+                return
+
+            await asyncio.sleep(0.050)
+
+    async def process_transaction_records(self, records: Collection[TransactionRecord]) -> None:
```
mariano54, line 300: it doesn't look like we are actually adding any TX records to the mempool here

mariano54: I think it makes more sense to add them to the mempool first

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +173,219 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
 
         for block in more_blocks:
             await self.full_node.respond_block(RespondBlock(block))
+
+    async def process_blocks(self, count: int, farm_to: bytes32 = bytes32([0] * 32)) -> int:
```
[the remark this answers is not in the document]
mariano54: I suggest we rename this to `farm_new_transaction_blocks` and have a corresponding `farm_new_blocks`. Or something like that

mariano54 (dismissed):
I like the helper methods, i think devs will benefit and be able to write tests more easily

chia/simulator/full_node_simulator.py
```diff
@@ -1,13 +1,41 @@
...
+    while True:
+        spendable_wallet_coin_records = await wallet.wallet_state_manager.get_spendable_coins_for_wallet(
+            wallet_id=wallet.id()
+        )
+        spendable_wallet_coins = {record.coin for record in spendable_wallet_coin_records}
+
+        if coins.issubset(spendable_wallet_coins):
+            return
+
+        await asyncio.sleep(0.050)
```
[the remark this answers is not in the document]
mariano54: should be done in a few ms in the normal case

tests/simulation/test_simulation.py
```diff
@@ -52,6 +63,12 @@ async def simulation(bt):
         yield _
 
 
+@pytest_asyncio.fixture(scope="function")
```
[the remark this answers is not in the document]
altendky: That sounds right.  This was originally coded last fall and I think this is a thing I didn't get caught up with latest practices.  Thanks.

tests/simulation/test_simulation.py
```diff
@@ -109,3 +126,203 @@ async def has_compact(node1, node2):
         await server3.start_client(PeerInfo(self_hostname, uint16(node1_port)))
         await server3.start_client(PeerInfo(self_hostname, uint16(node2_port)))
         await time_out_assert(600, node_height_at_least, True, node3, peak_height)
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize(argnames="count", argvalues=[0, 1, 2, 5, 10])
```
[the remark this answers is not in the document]
altendky: Mmm, so these are tests of the full node simulator rather than tests that the full node simulator is used for.  Yeah, I agree that's a good distinction and will look into creating another file.

chia/simulator/full_node_simulator.py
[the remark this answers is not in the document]
altendky: Thanks for the reference time.  I'll look at implementing the backoff.

chia/simulator/full_node_simulator.py
```diff
@@ -1,13 +1,41 @@
...
 from chia.types.full_block import FullBlock
 from chia.util.api_decorators import api_request
-from chia.util.ints import uint8
+from chia.util.ints import uint8, uint32, uint64
+from chia.wallet.transaction_record import TransactionRecord
+from chia.wallet.util.wallet_types import AmountWithPuzzlehash
+from chia.wallet.wallet import Wallet
+
+
+async def wait_for_coins_in_wallet(coins: Set[Coin], wallet: Wallet):
```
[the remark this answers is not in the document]
altendky: I'm a bit conflicted because this function will be useful in the same places where the full node wrapper is, yes?  I don't mean this is the right place, but having one thing in the chia package and another in the tests package doesn't obviously make sense to me yet.

chia/simulator/full_node_simulator.py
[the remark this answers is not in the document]
altendky: The loops usually came in pairs.  One for the farming and another for calculating the rewards.  This handles both those pieces.  But, I should revisit this from the API perspective and will think through your comments on that.

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +173,219 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+
+        if count == 0:
+            return rewards
+
+        for _ in range(count):
+            block: FullBlock = await self.farm_new_transaction_block(FarmNewBlockProtocol(farm_to))
+            height = uint32(block.height)
+            rewards += calculate_pool_reward(height) + calculate_base_farmer_reward(height)
+
+        while True:
```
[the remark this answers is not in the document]
altendky: I coded this so long ago that I don't recall if I got here by pure speculation or because I hit some funny case and needed it.  `:[`  If I can't find a reason for it, it can go away.

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +173,219 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+                raise RuntimeError("Peak height still None after processing at least one block")
+
+            if peak_height >= height:
+                break
+
+            await asyncio.sleep(0.050)
+
+        return rewards
+
+    async def farm_blocks(self, count: int, wallet: Wallet) -> int:
```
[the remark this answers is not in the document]
altendky: I'm not clear if you are pointing out a problem, or simply that not-connected wallets won't hear about anything.  I agree that the wallet won't realize it has funds if it isn't connected.

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +173,219 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+        height_before = self.full_node.blockchain.get_peak_height()
+        if height_before is None:
+            height_before = uint32(0)
+
+        for count in itertools.count(1):
+            height = uint32(height_before + count)
+            rewards += calculate_pool_reward(height) + calculate_base_farmer_reward(height)
+
+            if rewards >= amount:
+                await self.farm_blocks(count=count, wallet=wallet)
```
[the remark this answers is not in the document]
altendky: One of the properties of this function is that the rewards will be present in the wallet prior to returning.  How will that happen without any transaction blocks?  Though I could imagine using non-transaction blocks up until the last block.  Or am I still missing something?

chia/simulator/full_node_simulator.py
altendky: Consideration of timeouts is indeed on the draft reason list in the OP.  https://github.com/Chia-Network/chia-blockchain/pull/11535#issue-1236288585

chia/simulator/full_node_simulator.py
altendky: This was intended to make sure that already submitted transactions get processed.  For example, request that the wallet submit the tx, then call this to wait until the tx has been confirmed in a block.  Does it need a new name to help express that? Or do you think that's a bad API to offer?

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +173,219 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+                    found.add(coin)
+
+            coins_to_wait_for = coins_to_wait_for.difference(found)
+
+            if len(coins_to_wait_for) == 0:
+                return
+
+    async def create_coins_with_amounts(
+        self,
+        amounts: List[int],
```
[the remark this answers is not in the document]
altendky: I'm not completely convinced that using the serialized form throughout our codebase is best, but I agree that is what we do and what should be used here.

altendky (commented):
Thanks for the time and feedback.

I'm not sure if it would be worthwhile to consolidate to just an API list to discuss at that level, or if it wouldn't be useful without the code to say what the thing does.  Do you have any preference?

chia/simulator/full_node_simulator.py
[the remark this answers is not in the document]
mariano54: yeah the last one can be a transaction block, but the others don't need to be

chia/simulator/full_node_simulator.py
mariano54: Is there a reason to separate the adding to the mempool and then the farming / waiting?
Worst case, if it's already in the mempool, adding it will have no effect. So it seems like we can avoid some boilerplate by also offering to add to mempool

mariano54:
"I'm not sure if it would be worthwhile to consolidate to just an API list to discuss at that level, or if it wouldn't be useful without the code to say what the thing does. Do you have any preference?"

No preference really

chia/simulator/full_node_simulator.py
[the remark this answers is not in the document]
altendky: Addressed in https://github.com/Chia-Network/chia-blockchain/commit/4980ba605114d672660ed96aca337b17ddfb9ed6.

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +173,219 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+            for spend_bundle_name in ids_to_check:
+                tx = self.full_node.mempool_manager.get_spendbundle(spend_bundle_name)
+                if tx is not None:
+                    found.add(spend_bundle_name)
+            ids_to_check = ids_to_check.difference(found)
+
+            if len(ids_to_check) == 0:
+                return
+
+            await asyncio.sleep(0.050)
```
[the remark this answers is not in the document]
altendky: Addressed in https://github.com/Chia-Network/chia-blockchain/commit/4980ba605114d672660ed96aca337b17ddfb9ed6.

chia/simulator/full_node_simulator.py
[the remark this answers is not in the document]
altendky: Addressed in https://github.com/Chia-Network/chia-blockchain/commit/176b451074e00e0b6932ab36c159b2ed45235bcc.

chia/simulator/full_node_simulator.py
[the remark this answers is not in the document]
altendky: Addressed in https://github.com/Chia-Network/chia-blockchain/commit/30083cd4a30b944ce8e3a665deb5486b6639ebac and https://github.com/Chia-Network/chia-blockchain/commit/6e7e551d1c92f324d2999e743b88a86655d492fe.

chia/simulator/full_node_simulator.py
altendky: I'm not sure that `push_transaction_and_process(wallet=wallet, node=node, transactions=[tx])` is significantly better than two calls.  For some reason this activity feels like 'a helper' but pushing a transaction sounds like 'normal simple usage of a wallet'.  Burying the latter in the former feels like it might be obfuscation rather than abstraction.  `*shrug*`

Are you expecting that a combined helper would process it through the normal wallet calls or just direct to the node mempool?

mariano54: It should be `push_transaction_and_process(self, records)`

No need to pass in wallet or the node. Just call self.send_transaction before doing the other things

chia/simulator/full_node_simulator.py
```diff
@@ -44,7 +91,7 @@ async def get_all_full_blocks(self) -> List[FullBlock]:
         return blocks
 
     @api_request
-    async def farm_new_transaction_block(self, request: FarmNewBlockProtocol):
+    async def farm_new_transaction_block(self, request: FarmNewBlockProtocol) -> FullBlock:
```
mariano54, line 94: These api requests are not designed to be called manually, they are meant to be called through the chia protocol (through server.py). So it's kind of weird that we are returning stuff from them.

When an API method does return something, it's usually a Message which then gets sent back over the wire. We are breaking that patter here.

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +193,250 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+            if guarantee_transaction_blocks:
+                block = await self.farm_new_transaction_block(FarmNewBlockProtocol(farm_to))
+            else:
+                block = await self.farm_new_block(FarmNewBlockProtocol(farm_to))
+            height = uint32(block.height)
+            rewards += calculate_pool_reward(height) + calculate_base_farmer_reward(height)
+
+        peak_height = self.full_node.blockchain.get_peak_height()
+
+        if peak_height is None:
```
mariano54, line 229: this is impossible

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +193,250 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+            raise RuntimeError("Peak height still None after processing at least one block")
+
+        if peak_height < height:
```
mariano54, line 232: this is also impossible

chia/simulator/full_node_simulator.py
mariano54: I usually just use assert in those cases because it keeps the code cleaner and makes it clear that it is not a needed check

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +193,250 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+            return 0
+
+        target_puzzlehash = await wallet.get_new_puzzlehash()
+        rewards = 0
+
+        block_reward_coins = set()
+        expected_reward_coin_count = 2 * count
+
+        # TODO: why 2 and not 1?
+        for index in range(count + 2):
```
mariano54, line 259: It's 2 because each transaction will include all rewards up to and including the previous transaction block,

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +193,250 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+            raise RuntimeError("Not all reward coins identified")
+
+        if len(block_reward_coins) != expected_reward_coin_count:
+            raise RuntimeError(f"Expected {expected_reward_coin_count} reward coins, got: {len(block_reward_coins)}")
+
+        await wait_for_coins_in_wallet(coins=block_reward_coins, wallet=wallet)
+
+        return rewards
+
+    async def farm_rewards_to_wallet(self, amount: int, wallet: Wallet) -> int:
```
mariano54, line 288: This function seems redundant and almost the same as the previous one

chia/simulator/full_node_simulator.py
altendky: Maybe this should have been designed differently to avoid needing to handle impossible cases.  But, we use assert all over in places I consider very much not appropriate with results like error messages that say nothing about what happened such as `{"error": "", "success": False}`.  There are times I perpetuate that but I chose not to here.

Maybe we could make `.get_peak_height()` just raise an exception itself instead of `None` checking every single call where getting `None` is 'impossible'?

chia/simulator/full_node_simulator.py
altendky: `for index in range(count):` would iterate `count` times with all cases satisfying `index < count` and thus farming a block (maybe transaction, maybe not) to the wallet.  With `+ 1` it would iterate an extra time which should result in a final transaction block being processed (without the reward going to the wallt) that should include all the rewards from the previous `count` blocks.  `+ 2` processes a second transaction block.  Why do we need two transaction blocks at the end?

Though, I should probably write this a bit differently.  Reworded as below in https://github.com/Chia-Network/chia-blockchain/commit/980418d48de08c174c2831f1eb37ceb568475915..

```python
        # TODO: why two final transaction blocks and not just one?
        for to_wallet in [*([True] * count), False, False]:
            if to_wallet:
```

chia/simulator/full_node_simulator.py
altendky: It is part of trying to work away from tests being coded based around inner workings of the blockchain.  A lot of tests are not about farming yet they have farming loops where they depend on the reward size instead of just trying to make sure they have some amount of xch/coins available so they can run.  I don't think this is the final form but I do think a majority of tests probably shouldn't have any code about farming rewards or the number of blocks needed to get enough coins.

chia/simulator/full_node_simulator.py
altendky: Disappeared in https://github.com/Chia-Network/chia-blockchain/commit/93c525cb999fa89d6d06d4fea78c9e81345912ca.

tests/simulation/test_simulator.py
```diff
@@ -0,0 +1,266 @@
...
+            puzzle_hash=await wallet_node.wallet_state_manager.main_wallet.get_new_puzzlehash(),
+            coins={coin},
+        )
+        await wallet.push_transaction(tx)
+
+        await full_node_api.wait_transaction_records_entered_mempool(records=[tx])
+        assert tx.spend_bundle is not None
+        assert full_node_api.full_node.mempool_manager.get_spendbundle(tx.spend_bundle.name()) is not None
+        # TODO: this fails but it seems like it shouldn't when above passes
+        # assert tx.is_in_mempool()
```
altendky, line 167: Is there an explanation for this still failing when uncommented?

tests/simulation/test_simulator.py
```diff
@@ -0,0 +1,266 @@
...
+        tx = await wallet.generate_signed_transaction(
+            amount=uint64(tx_amount),
+            puzzle_hash=await wallet_node.wallet_state_manager.main_wallet.get_new_puzzlehash(),
+            coins={coin},
+        )
+        await wallet.push_transaction(tx)
+
+        await full_node_api.process_transaction_records(records=[tx])
+        # TODO: is this the proper check?
+        assert full_node_api.full_node.coin_store.get_coin_record(coin.name()) is not None
```
altendky, line 201: Is this the proper check?

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +203,288 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+
+            if count == 0:
+                return rewards
+
+            for _ in range(count):
+                if guarantee_transaction_blocks:
+                    await self.farm_new_transaction_block(FarmNewBlockProtocol(farm_to))
+                else:
+                    await self.farm_new_block(FarmNewBlockProtocol(farm_to))
+                # hopefully there is no race between the
```
mariano54, line 241: comment

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +203,288 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+                        count=1,
+                        farm_to=target_puzzlehash,
+                        guarantee_transaction_blocks=False,
+                        timeout=None,
+                    )
+                else:
+                    await self.farm_blocks_to_puzzlehash(count=1, guarantee_transaction_blocks=True, timeout=None)
+
+                peak_height = self.full_node.blockchain.get_peak_height()
+                if peak_height is None:
```
mariano54, line 295: not possible

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +203,288 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+
+        Returns:
+            The total number of reward mojos farmed to the requested wallet.
+        """
+        rewards = 0
+
+        if amount == 0:
+            return rewards
+
+        height_before = self.full_node.blockchain.get_peak_height()
```
mariano54, line 339: can we use type annotations? For example: `Optional[uint32]`

chia/simulator/full_node_simulator.py
```diff
@@ -144,3 +203,288 @@ async def reorg_from_index_to_new_index(self, request: ReorgProtocol):
...
+            coin_store = self.full_node.coin_store
+
+            await self.wait_transaction_records_entered_mempool(records=records, timeout=None)
+
+            while True:
+                await self.farm_blocks_to_puzzlehash(count=1, timeout=None)
+
+                found: Set[Coin] = set()
+                for coin in coins_to_wait_for:
+                    # TODO: is this the proper check?
```
mariano54, line 419: check for removals and additions?

chia/simulator/full_node_simulator.py
altendky: I'll add it but I dislike this since it can broaden the hinting unnecessarily which can cause extra handling code to be introduced downstream.  I've added this to a list to discuss later in terms of preferred coding practices.

chia/simulator/full_node_simulator.py
altendky: Adding the assertion here as you want.  I still think that the assertion is an indication of an opportunity for improvement elsewhere.  It will take a bit of work to unravel this one though.

chia/simulator/full_node_simulator.py
altendky: removing

tests/simulation/test_simulator.py
altendky: This is cached data.

```suggestion
```

setup.py
```diff
@@ -44,6 +44,7 @@
 ]
 
 dev_dependencies = [
+    "anyio",
```
almogdepaz, line 47: anyone else needs to get involved when adding new dep ?

chia/simulator/full_node_simulator.py
```diff
@@ -299,30 +317,34 @@ async def process_blocks(self, count: int, farm_to: bytes32 = bytes32([0] * 32))
...
+                assert height is not None
 
-            await asyncio.sleep(0.050)
+                rewards += calculate_pool_reward(height) + calculate_base_farmer_reward(height)
 
-        return rewards
+            return rewards
 
-    async def farm_blocks(self, count: int, wallet: Wallet) -> int:
+    async def farm_blocks_to_wallet(
```
almogdepaz, line 342: what is the diff between this and farm_blocks_to_puzzlehash except using wallet instead of specific puzzhash ?

tests/wallet/test_wallet.py
```diff
@@ -688,58 +596,55 @@ async def test_wallet_tx_reorg(
...
         peak_height = peak.height
-        print(peak_height)
+        assert peak_height is not None
 
+        target_height_after_reorg = peak_height + 3
         # Perform a reorg, which will revert the transaction in the full node and wallet, and cause wallet to resubmit
         await full_node_api.reorg_from_index_to_new_index(
-            ReorgProtocol(uint32(peak_height - 3), uint32(peak_height + 3), bytes32(32 * b"0"), None)
+            ReorgProtocol(uint32(reorg_height - 1), uint32(target_height_after_reorg), bytes32(32 * b"0"), None)
         )
```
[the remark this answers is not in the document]
altendky: Added in https://github.com/Chia-Network/chia-blockchain/pull/11535/commits/f6972f60d6045a2654bec88392c02adf98319ffe.

chia/simulator/full_node_simulator.py
almogdepaz: would it be better if we move the wait_for_coins_in_wallet to farm_blocks_to_puzzlehash and just avoid explicitly waiting for wallet in the tests ? or is there value in farm_blocks_to_puzzlehash without waiting for the wallet ?

(not saying this is relevant to this pr)

tests/wallet/test_wallet.py
```diff
@@ -307,61 +261,48 @@ async def test_wallet_make_transaction_hop(
...
+            uint64(tx_amount), await wallet_0.get_new_puzzlehash(), uint64(0)
+        )
         await wallet_1.push_transaction(tx)
-        await time_out_assert(20, full_node_0.mempool_manager.get_spendbundle, tx.spend_bundle, tx.name)
+        await full_node_api_0.wait_transaction_records_entered_mempool(records=[tx])
 
-        for i in range(0, 4):
-            await full_node_api_0.farm_new_transaction_block(FarmNewBlockProtocol(bytes32(32 * b"0")))
+        await full_node_api_0.farm_blocks_to_puzzlehash(count=4, guarantee_transaction_blocks=True)
+        funds += 5
```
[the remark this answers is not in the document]
almogdepaz: i mean we used to have new_funds - 5 in the assertions below im trying to understand if we still assert the same amount as we used to

tests/wallet/test_wallet.py
```diff
@@ -792,23 +697,38 @@ async def test_address_sliding_window(
...
 
-        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(bytes32(32 * b"0")))
-        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hashes[0]))
-        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hashes[210]))
-        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hashes[114]))
-        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(bytes32(32 * b"0")))
+        funds = 0
+        gapped_funds = 0
+
+        funds += await full_node_api.farm_blocks_to_puzzlehash(count=1, farm_to=puzzle_hashes[0])
```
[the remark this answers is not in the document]
altendky: A farm to nowhere can be irrelevant.  But, I'll look harder here.

chia/simulator/full_node_simulator.py
altendky: If you are farming to a puzzle hash you may not even have a wallet.  A wallet could be made an optional parameter and if present we could verify the puzzle hash belongs to it and also wait for the wallet to sync etc.  I think mostly I wanted to shift away from fiddling with puzzle hashes.  In most of the tests I don't think we care about the puzzle hash.

Big picture, I think that the majority of times we farm we shouldn't even be doing that anyways.  We farm to a wallet (or non-zero puzzlehash) generally to get coins/funds into it so we can run the test.  something like `gimme_these_coins(wallet=wallet, coins=[1, 27, 13, 47])` would be a lot more representative in a test.  I imagine that could be implemented by sending from the prefarm coins and not "farming" at all.  Just processing blocks to complete transactions.

tests/wallet/test_wallet.py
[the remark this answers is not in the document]
altendky: The first farm to `bytes32(32 * b"0")` doesn't seem to me to have any relevance.  I don't see any transactions before it that should be processed.  It isn't putting funds/coins into a wallet we control.  _Seems_ ok to drop.

setup.py
justinengland: approving

justinengland (dismissed):

chia/simulator/full_node_simulator.py
almogdepaz: i agree we should keep it separate

chia/simulator/full_node_simulator.py
altendky: Pre-existing `TODO:`, can deal with separately at some point.

tests/wallet/test_wallet.py
```diff
@@ -912,11 +822,11 @@ async def test_wallet_transaction_options(
         await time_out_assert(20, wallet.get_unconfirmed_balance, funds - AMOUNT_TO_SEND)
         await time_out_assert(20, full_node_api.full_node.mempool_manager.get_spendbundle, tx.spend_bundle, tx.name)
 
-        for i in range(0, num_blocks):
-            await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(bytes32([0] * 32)))
+        await full_node_api.farm_blocks_to_puzzlehash(count=num_blocks, guarantee_transaction_blocks=True)
+        funds -= AMOUNT_TO_SEND
```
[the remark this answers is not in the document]
altendky: 5b717cbabf1d2d914d07cdd6d3f857cbb65a3604

almogdepaz (dismissed):

almogdepaz (approved):

emlowe (approved):

PR #11537, opened 2022-05-15
make recurse_jsonify() work directly on types

arvidn:
circumventing the dataclasses.asdict() step. This enables simpler integration of non dataclasses into the Streamable and JSON protocols. Specifically, it opens the door for allowing native rust types to participate in `Streamable` (which includes json conversion).

As a bonus, this change avoids creating an intermediate dictionary, before converting it into JSON form.

A dictionary in JSON form is one that only contains `Dict`, `List`, `str` and `int`. `bytes32` objects are turned into hexadecimal strings.

This patch has a counterpart in `chia_rs`: https://github.com/Chia-Network/chia_rs/pull/18

**EDIT:**

I added a benchmark to ensure this doesn't make it more expensive to generate JSON messages. The benchmark calls `to_json_dict()` on `FullNode` objects of varying kinds (32768 combinations) timing those calls.

| --- | --- | --- |
| 336.15s | 101.90s | 0.303 |

So, a material speed-up

### DIFF
--- a/chia/util/misc.py
+++ b/chia/util/misc.py
@@ -1,4 +1,3 @@
-import dataclasses
 from typing import Any, Dict, Sequence, Union
 
 from chia.util.streamable import recurse_jsonify
@@ -81,4 +80,5 @@ def get_list_or_len(list_in: Sequence[object], length: bool) -> Union[int, Seque
 
 
 def dataclass_to_json_dict(instance: Any) -> Dict[str, Any]:
-    return recurse_jsonify(dataclasses.asdict(instance))
+    ret: Dict[str, Any] = recurse_jsonify(instance)
+    return ret
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -5,21 +5,7 @@ import io
 import pprint
 import sys
 from enum import Enum
-from typing import (
-    Any,
-    BinaryIO,
-    Callable,
-    Dict,
-    Iterator,
-    List,
-    Optional,
-    Tuple,
-    Type,
-    TypeVar,
-    Union,
-    get_type_hints,
-    overload,
-)
+from typing import Any, BinaryIO, Callable, Dict, Iterator, List, Optional, Tuple, Type, TypeVar, Union, get_type_hints
 
 from blspy import G1Element, G2Element, PrivateKey
 from typing_extensions import Literal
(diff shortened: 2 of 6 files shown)


chia/util/streamable.py
```diff
@@ -134,54 +120,38 @@ def dataclass_from_dict(klass: Type[Any], d: Any) -> Any:
...
     """
     Makes bytes objects and unhashable types into strings with 0x, and makes large ints into
     strings.
     """
-    if isinstance(d, list) or isinstance(d, tuple):
+    if dataclasses.is_dataclass(d):
+        new_dict = {}
+        for field in dataclasses.fields(d):
+            new_dict[field.name] = recurse_jsonify(getattr(d, field.name))
+        return new_dict
```
arvidn, line 132: this block,  effectively, does the equivalent of `dataclasses.asdict()`, by traversing the members recursively

chia/util/streamable.py
```diff
@@ -134,54 +120,38 @@ def dataclass_from_dict(klass: Type[Any], d: Any) -> Any:
...
-            if isinstance(value, int) and type(value) in big_ints:
-                d[key] = int(value)
-    return d
+            new_list.append(recurse_jsonify(item))
+        return new_list
+
+    elif isinstance(d, dict):
+        new_dict = {}
+        for name, val in d.items():
+            new_dict[str(name)] = recurse_jsonify(val)
```
altendky, line 143: Do we have names that aren't strings?

chia/util/streamable.py
```diff
@@ -134,54 +120,38 @@ def dataclass_from_dict(klass: Type[Any], d: Any) -> Any:
...
-    ...
-
-
-@overload
-def recurse_jsonify(d: Dict[str, Any]) -> Dict[str, Any]:
-    ...
-
-
-def recurse_jsonify(d: Union[List[Any], Tuple[Any, ...], Dict[str, Any]]) -> Union[List[Any], Dict[str, Any]]:
+def recurse_jsonify(d: Any) -> Any:
```
altendky, line 123: Could we maybe keep the hinting and just add a `Any` -> `Dict[str, Any]` to the end?  I think that would just be a catch all if the existing overloads don't match.

arvidn: I might need some help with the hinting. The return type isn't necessarily a `Dict[]`, it's something we might call "JSONCompatible" i.e. one of `Dict[str, JSONCompatible]`, `List[JSONCompatible]`, 'int', 'str'

arvidn: at least the simple type definition:
```
JsonCompat = Union[Dict[str, "JsonCompat"], List["JsonCompat"], str, int, None]
```
makes mypy fail with:
``` 
error: Cannot resolve name "JsonCompat" (possible cyclic definition)  [misc]
```

chia/util/streamable.py
altendky: If they aren't strings, is it reasonable to convert them all?  Is this an important change here?

chia/util/streamable.py
```diff
@@ -134,54 +120,38 @@ def dataclass_from_dict(klass: Type[Any], d: Any) -> Any:
...
     """
     Makes bytes objects and unhashable types into strings with 0x, and makes large ints into
     strings.
     """
-    if isinstance(d, list) or isinstance(d, tuple):
+    if dataclasses.is_dataclass(d):
```
altendky, line 128: Should this work for any dataclass?  Or should this be a streamable thing.  If streamable we can use a streamable hint in a new overload above and a streamable function to get the attributes.

arvidn: in a way this is not a publicly facing function, and we control all call sites; so we can make it do whatever we want really. For example, the `Streamable` serialization format doesn't support enums, so we could probably remove that here (but it would take some work to ensure we don't rely on it in some way that isn't covered by a test).

The important think I'm trying to make this function do is to support types that do *not* derive from `Streamable`,  but implement a protocol that can more easily be implemented, by a rust native type for instance. I'm not sure if it's possible with any python bindings for a native type to derive from a pure python type. It's not supported by pyo3 that that we use in rust, and I think in other binding libraries it would be very complicated.

The hook I'm envisioning is that types that aren't a dataclass (nor one of the built-in Dict, List, int, str, None) can implement a `to_json_dict()` to do the same thing.

In other words, I'm trying to separate this from the `Streamable` class.

xdustinface:
@arvidn Not sure if we need another benchmark here. Have you seen `benchmarks/streamable.py`? 

You can generate a record on `main` with `python streamable.py -o main.json` optionally add `-m to_json` and then compare it like below

```
python streamable.py -c main.json -m to_json

benchmarks: to_json, data: BenchmarkClass runs: 100, ms/run: 50, commit_hash: eea7e0764-dirty
mode       | µs/iteration | stdev µs/iteration % | avg iterations/run | stdev iterations/run %
to_json    | 294.22       |                 5.34 |                169 |                    3.3

benchmarks: to_json, data: FullBlock runs: 100, ms/run: 50, commit_hash: eea7e0764-dirty
mode       | µs/iteration | stdev µs/iteration % | avg iterations/run | stdev iterations/run %
to_json    | 219.71       |                 8.97 |                226 |                   5.42

compare: benchmark, old: 99275e6e3-dirty, new: eea7e0764-dirty
to_json      | 827.98           | 294.22           | -64.46      

compare: full_block, old: 99275e6e3-dirty, new: eea7e0764-dirty
to_json      | 445.7            | 219.71           | -50.7    
```

altendky (commented):
I think not `str()`ing might be important.  The hinting I can leave for a separate PR if you don't like it.

chia/util/streamable.py
arvidn: I see your point. I was actually convinced that `str` conversion existed before my patch. But I can't find where I got that from now. I'll remove it.

chia/util/misc.py
```diff
@@ -81,4 +80,5 @@ def get_list_or_len(list_in: Sequence[object], length: bool) -> Union[int, Seque
 
 
 def dataclass_to_json_dict(instance: Any) -> Dict[str, Any]:
-    return recurse_jsonify(dataclasses.asdict(instance))
+    ret: Dict[str, Any] = recurse_jsonify(instance)
+    return ret
```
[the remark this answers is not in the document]
arvidn: yes

arvidn:
> @arvidn Not sure if we need another benchmark here. Have you seen benchmarks/streamable.py?

I have seen it. I assumed it was benchmarking `Streamable` serialization, not json serialization.

altendky (approved):
I like the idea of the change and the implementation looks ok as far as I can tell.  But, I will include a caveat here that I am generally uncomfortable with the form of our serialization and specifically here that there may be some hidden corners.  While they were probably not actually problematic to change, I did catch to such non-obvious corners in this PR (those changes have been removed).

PR #11546, opened 2022-05-17
Optimize code to not perform useless subgroup checks

mariano54:
The new method is safe to call if we are in any of the following circumstances:
1. We call AugSchemeMPL.aggregate_verify with the public keys
2. We derived the public key from a private key (even if it was done at some point in the past)

The previous change made performance worse due to double subgroup checks. This was the commit that introduced the issue:
https://github.com/Chia-Network/chia-blockchain/commit/56e6e68eb3021799d8561192de402a67749dc853

full sync:

| --- | --- | --- |
| 129.59s | 117.03s | 0.903 |

### DIFF
--- a/chia/cmds/keys_funcs.py
+++ b/chia/cmds/keys_funcs.py
@@ -178,8 +178,8 @@ def sign(message: str, private_key: PrivateKey, hd_path: str, as_bytes: bool):
 
 def verify(message: str, public_key: str, signature: str):
     messageBytes = bytes(message, "utf-8")
-    public_key = G1Element.from_bytes(bytes.fromhex(public_key))
-    signature = G2Element.from_bytes(bytes.fromhex(signature))
+    public_key = G1Element.from_bytes_unchecked(bytes.fromhex(public_key))
+    signature = G2Element.from_bytes_unchecked(bytes.fromhex(signature))
     print(AugSchemeMPL.verify(public_key, messageBytes, signature))
 
 
--- a/chia/cmds/keys_funcs.py
+++ b/chia/cmds/keys_funcs.py
@@ -178,8 +178,8 @@ def sign(message: str, private_key: PrivateKey, hd_path: str, as_bytes: bool):
 
 def verify(message: str, public_key: str, signature: str):
     messageBytes = bytes(message, "utf-8")
-    public_key = G1Element.from_bytes_unchecked(bytes.fromhex(public_key))
-    signature = G2Element.from_bytes_unchecked(bytes.fromhex(signature))
+    public_key = G1Element.from_bytes(bytes.fromhex(public_key))
+    signature = G2Element.from_bytes(bytes.fromhex(signature))
     print(AugSchemeMPL.verify(public_key, messageBytes, signature))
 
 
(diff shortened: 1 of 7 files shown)


chia/cmds/keys_funcs.py
```diff
@@ -178,8 +178,8 @@ def sign(message: str, private_key: PrivateKey, hd_path: str, as_bytes: bool):
 
 def verify(message: str, public_key: str, signature: str):
     messageBytes = bytes(message, "utf-8")
-    public_key = G1Element.from_bytes(bytes.fromhex(public_key))
-    signature = G2Element.from_bytes(bytes.fromhex(signature))
```
arvidn, line 182: how can you be sure that we can trust this key and signature?

chia/consensus/multiprocess_validation.py
```diff
@@ -122,7 +122,8 @@ def batch_pre_validate_blocks(
                         if npc_result is not None and block.transactions_info is not None:
                             assert npc_result.conds
                             pairs_pks, pairs_msgs = pkm_pairs(npc_result.conds, constants.AGG_SIG_ME_ADDITIONAL_DATA)
-                            pks_objects: List[G1Element] = [G1Element.from_bytes(pk) for pk in pairs_pks]
+                            # Using AugSchemeMPL.aggregate_verify, so it's safe to use from_bytes_unchecked
+                            pks_objects: List[G1Element] = [G1Element.from_bytes_unchecked(pk) for pk in pairs_pks]
```
arvidn, line 126: this looks reasonable, because your previous patch added subgroup checks to `aggregate_verify()`, right?

chia/util/keychain.py
```diff
@@ -264,7 +264,7 @@ def _get_pk_and_entropy(self, user: str) -> Optional[Tuple[G1Element, bytes]]:
             return None
         str_bytes = bytes.fromhex(read_str)
         return (
-            G1Element.from_bytes(str_bytes[: G1Element.SIZE]),
+            G1Element.from_bytes_unchecked(str_bytes[: G1Element.SIZE]),
```
arvidn, line 267: do we call this function often enough for this optimization to matter?

chia/wallet/sign_coin_spends.py
```diff
@@ -46,7 +46,7 @@ async def sign_coin_spends(
 
         # Create signature
         for pk_bytes, msg in pkm_pairs_for_conditions_dict(conditions_dict, coin_spend.coin.name(), additional_data):
-            pk = blspy.G1Element.from_bytes(pk_bytes)
+            pk = blspy.G1Element.from_bytes_unchecked(pk_bytes)
```
arvidn, line 49: it looks like this is coming from a generator program. is it always  a program that our wallet just created?

chia/cmds/keys_funcs.py
mariano54: We now perform the subgroup checks inside the verify method

chia/consensus/multiprocess_validation.py
mariano54: exactly

chia/util/keychain.py
mariano54: the wallet might call this many times, but it probably does not matter

chia/wallet/sign_coin_spends.py
mariano54: we control the private key so we know that the pk was generated properly

arvidn (approved):

PR #11549, opened 2022-05-17
Optimize std_hash in coin.py

mariano54:
constructing a bytes object from a bytes object is not necessary. In these cases we are passing in bytes so we don't need to convert. These are also the places where `std_hash` is called the most.

Also, using hashlib instead of calling into C++ is more efficient.

This makes `std_hash` around 25% faster

test_full_sync
| --- | --- | --- |
| 119s | 116s |  0.975 |

Another thing to improve would be not checking the size when constructing a bytes32 here. It's very slow for some reason (calling len(bytes)). However I'm not sure how to add an optional arg to the constructor of bytes32

### DIFF
--- a/chia/types/blockchain_format/coin.py
+++ b/chia/types/blockchain_format/coin.py
@@ -28,7 +28,9 @@ class Coin(Streamable):
         # significant bit is set, to encode it as a positive number. This
         # despite "amount" being unsigned. This way, a CLVM program can generate
         # these hashes easily.
-        return std_hash(self.parent_coin_info + self.puzzle_hash + int_to_bytes(self.amount))
+        return std_hash(
+            self.parent_coin_info + self.puzzle_hash + int_to_bytes(self.amount), skip_bytes_conversion=True
+        )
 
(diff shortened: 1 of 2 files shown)


arvidn:
I'm experimenting with a more radincal performance improvement to coin. I'm tempted to suggest holding this off a bit. It's a pretty small speed-up too (maybe even within the error margins).

I suspect `std_hash` releases the GIL, and that's why it's so much slower than the python-native hash function.

mariano54:
I ran multiple times so I'm confident in the speedup. Also seems like a harmless change.  If this is addressed in your other changes, we can close

mariano54:
@arvidn do you plan on making those Coin changes? If so, let's close. 
If not, I think we should merge this one and you can build on top when you come back to it

arvidn:
yeah, it probably makes sense to merge this anyway. It's a step in the right direction. My coin patch depends on https://github.com/Chia-Network/chia-blockchain/pull/11503 so it will probably be at least a week before it can land.

arvidn (approved):

PR #11557, opened 2022-05-18
Bring DID/NFT effort into main

AmineKhaldi:


chia/wallet/puzzles/nft_ownership_layer.clvm.hex
```diff
@@ -0,0 +1 @@
+expected defun, defun-inline, defmacro, or defconstant
```
prozacchiwawa, line 1: Looks like this file isn't present (was removed or missing)?  Probably should synchronize the hex and sha256.

xdustinface: Looks like @trepca removed the content in https://github.com/Chia-Network/chia-blockchain/commit/4df1f4658d9d29704c7e876aa8913fe5863344ac

prozacchiwawa (dismissed):
noticed one thing, otherwise seems ok

mariano54 (changes requested):
I might have reviewed the wrong PR. Not sure what the differences are, but seems to be mostly the same. I left some comments in the other one. https://github.com/Chia-Network/chia-blockchain/pull/11583

chia/wallet/puzzles/nft_ownership_layer.clvm.hex
trepca: @xdustinface indeed, I think it slipped while resolving conflicts, will fix, thanks

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -0,0 +1,564 @@
...
+    def get_current_nfts(self) -> List[NFTCoinInfo]:
+        return self.nft_wallet_info.my_nft_coins
+
+    async def save_info(self, nft_info: NFTWalletInfo, in_transaction: bool) -> None:
+        self.nft_wallet_info = nft_info
+        current_info = self.wallet_info
+        data_str = json.dumps(nft_info.to_json_dict())
+        wallet_info = WalletInfo(current_info.id, current_info.name, current_info.type, data_str)
+        self.wallet_info = wallet_info
+        await self.wallet_state_manager.user_store.update_wallet(wallet_info, in_transaction)
```
xdustinface, line 564: Actually we should have the wallet versioned in general but since we don't have that at the moment and it will be a bigger change i think we should at least for now wrap `NFTWalletInfo` to give it a version since its stored in the DB and is likely to be changed? See https://github.com/Chia-Network/chia-blockchain/pull/11587 which could be used.

xdustinface (changes requested):

jspro25:
recieved nft?

tests/wallet/did_wallet/test_did_rpc.py
```diff
@@ -1,121 +1,57 @@
-import logging
+import asyncio
```
xdustinface, line 1: So we have this file now to test some DID RPCs and we have a section in `test_wallet_rpc.py`, see https://github.com/Chia-Network/chia-blockchain/blob/a9b1ba8575edcf8eefe15751c24d5f77c6c14b56/tests/wallet/rpc/test_wallet_rpc.py#L758

I think this should be consolidated.

altendky (commented):
Just a few things I noticed when i barely started to read tihs.

tests/wallet/did_wallet/test_did_rpc.py
trepca: @ytx1991 can you look into this?

ghost (commented):
ready to add calls i believe

PR #11558, opened 2022-05-18
tests: Some adjustments in `test_offer_endpoints`

xdustinface:
This is a follow up after #11552 which basically used the initial parts of `test_cat_endpoints` to make `test_offer_endpoints` working. Now this PR drops a lot of the stuff we can assume to be not needed in this test because it's tested in `test_cat_endpoints` and it adds some general improvements to it.

### DIFF
--- a/tests/wallet/rpc/test_wallet_rpc.py
+++ b/tests/wallet/rpc/test_wallet_rpc.py
@@ -628,91 +628,69 @@ async def test_offer_endpoints(wallet_rpc_environment: WalletRpcTestEnvironment)
     env: WalletRpcTestEnvironment = wallet_rpc_environment
 
     wallet_node: WalletNode = env.wallet_1.node
-
-    client: WalletRpcClient = env.wallet_1.rpc_client
-    client_2: WalletRpcClient = env.wallet_2.rpc_client
-
+    wallet_1_rpc: WalletRpcClient = env.wallet_1.rpc_client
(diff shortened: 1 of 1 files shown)


tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -628,91 +628,69 @@ async def test_offer_endpoints(wallet_rpc_environment: WalletRpcTestEnvironment)
...
-    cat_0_id = res["wallet_id"]
-    asset_id = bytes32.fromhex(res["asset_id"])
-    assert len(asset_id) > 0
-
-    # TODO: Investigate why farming only one block here makes it flaky
+    # Creates a CAT wallet with 20 mojos
+    res = await wallet_1_rpc.create_new_cat_and_wallet(uint64(20))
+    cat_wallet_id = res["wallet_id"]
+    cat_asset_id = bytes32.fromhex(res["asset_id"])
+    # TODO: Investigate why farming only two blocks here makes it flaky
```
mariano54, line 642: Because we are creating a tx (minting CAT) but not waiting for the tx to enter the mempool

mariano54 (approved):

tests/wallet/rpc/test_wallet_rpc.py
xdustinface: Will address this once we have a central place for our helpers like in #11535.

PR #11565, opened 2022-05-19
Quexington's Offer generalization bringup

AmineKhaldi:
This is part of PR #11557 to be reviewed separately for convenience.


AmineKhaldi:
@Quexington @mariano54 this is an effort to bring-in parts of nft work that can be reviewed/merged separately. Your reviews are appreciated.

Quexington (commented):
I wrote this code so I'm not sure I should be the one to give it the thumbs up.  This is basically a refactor of the offer code so if the existing tests are working then that gives me pretty high confidence.

chia/rpc/wallet_rpc_client.py
```diff
@@ -420,13 +420,24 @@ async def cat_spend(
 
     # Offers
     async def create_offer_for_ids(
-        self, offer_dict: Dict[uint32, int], fee=uint64(0), validate_only: bool = False
+        self,
+        offer_dict: Dict[uint32, int],
```
Quexington, line 424: To fix the mypy error this should be Dict[Union[uint32, str], int]

AmineKhaldi:
Closing to reopen, the checks seem stuck.

Quexington (commented):
:+1: to Amine's merge skills, but someone else should approve this before it gets merged into main

xdustinface (changes requested):
@Quexington Left some comments, some of them should be addressed here, others maybe not (yet?). Overall it would be nice to see more tests for all the new introduced code and logic!

The merge seems fine, i got the same result when picking the PRs into `main` after resolving the conflicts.

chia/wallet/cat_wallet/cat_outer_puzzle.py
```diff
@@ -0,0 +1,92 @@
...
+    construct_cat_puzzle,
+    match_cat_puzzle,
+    unsigned_spend_bundle_for_spendable_cats,
+)
+from chia.wallet.lineage_proof import LineageProof
+from chia.wallet.puzzle_drivers import PuzzleInfo, Solver
+
+
+@dataclass(frozen=True)
+class CATOuterPuzzle:
```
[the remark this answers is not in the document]
Quexington: This is covered by the tests that already exist for offers.  This code was scattered throughout the offer driver before and has just been extracted to this file now.

chia/wallet/trade_manager.py
```diff
@@ -26,6 +26,42 @@
...
+                    "parent_spend": bytes
+                    "siblings": List[bytes]  # other coins of the same type being offered
+                    "sibling_spends": List[bytes]  # The parent spends for the siblings
+                    "sibling_puzzles": List[Program]  # The inner puzzles of the siblings (always OFFER_MOD)
+                    "sibling_solutions": List[Program]  # The inner solution of the siblings
+                }
+            )
+
+    Wallet:
+      - Segments in this code that call general wallet methods are highlighted by comments: # ATTENTION: new wallets
```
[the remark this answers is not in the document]
Quexington: This is a good idea, I knew something like that must be possible but I didn't know how to do it.

chia/wallet/trading/offer.py
```diff
@@ -242,15 +253,20 @@ def aggregate(cls, offers: List["Offer"]) -> "Offer":
...
                 else:
-                    total_requested_payments[tail] = payments
+                    total_requested_payments[asset_id] = payments
+
+            for key, value in offer.driver_dict.items():
+                if key in total_driver_dict and total_driver_dict[key] != value:
+                    raise ValueError(f"The offers to aggregate disagree on the drivers for {key.hex()}")
 
             total_bundle = SpendBundle.aggregate([total_bundle, offer.bundle])
+            total_driver_dict = {**total_driver_dict, **offer.driver_dict}
```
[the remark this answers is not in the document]
xdustinface: It's just a (at least imo) simpler statement and you update the dict instead of creating a new one each time. Like you would use `dict["new_key"] = new_value` instead of `dict = {**dict, "new_key": new_value}`.

chia/wallet/cat_wallet/cat_outer_puzzle.py
```diff
@@ -0,0 +1,92 @@
...
+                "tail": "0x" + tail_hash.as_python().hex(),
+            }
+            next_constructor = self._match(inner_puzzle)
+            if next_constructor is not None:
+                constructor_dict["also"] = next_constructor.info
+            return PuzzleInfo(constructor_dict)
+        else:
+            return None
+
+    def asset_id(self, constructor: PuzzleInfo) -> Optional[bytes32]:
```
mariano54, line 42: this is not Optional right?

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -790,3 +809,19 @@ async def save_info(self, cat_info: CATInfo, in_transaction):
...
+        )
+
+    def get_puzzle_info(self, asset_id: bytes32) -> PuzzleInfo:
+        return PuzzleInfo({"type": AssetType.CAT.value, "tail": "0x" + self.get_asset_id()})
+
+    async def get_coins_to_offer(self, asset_id: Optional[bytes32], amount: uint64) -> Set[Coin]:
+        balance = await self.get_confirmed_balance()
+        if balance < amount:
+            raise Exception(f"insufficient funds in wallet {self.id()}")
+        return await self.select_coins(amount)
```
mariano54, line 827: This will also raise if there are enough confirmed balance, but not enough spendable balance. Is that OK?

chia/wallet/wallet.py
```diff
@@ -526,3 +526,11 @@ async def create_spend_bundle_relative_chia(self, chia_amount: int, exclude: Lis
...
         )
         return spend_bundle
+
+    async def get_coins_to_offer(self, asset_id: Optional[bytes32], amount: uint64) -> Set[Coin]:
+        if asset_id is not None:
+            raise ValueError(f"The standard wallet cannot offer coins with asset id {asset_id}")
+        balance = await self.get_confirmed_balance()
+        if balance < amount:
+            raise Exception(f"insufficient funds in wallet {self.id()}")
+        return await self.select_coins(amount)
```
mariano54, line 536: same here, this can raise if there is not enough spendable funds

mariano54 (commented):
These changes seem good, but I am not deeply familiar with offer code so I can't provide much useful feeback.
Looks like no slow tests are added. We should ensure all tests pass and without flakiness.

chia/wallet/trade_manager.py
```diff
@@ -246,9 +283,15 @@ async def save_trade(self, trade: TradeRecord):
...
-        self, offer: Dict[Union[int, bytes32], int], fee: uint64 = uint64(0), validate_only: bool = False
+        self,
+        offer: Dict[Union[int, bytes32], int],
+        driver_dict: Optional[Dict[bytes32, PuzzleInfo]] = None,
+        fee: uint64 = uint64(0),
+        validate_only: bool = False,
     ) -> Tuple[bool, Optional[TradeRecord], Optional[str]]:
-        success, created_offer, error = await self._create_offer_for_ids(offer, fee=fee)
+        if driver_dict is None:
+            driver_dict = {}
```
xdustinface, line 293:
```suggestion
```

Its only passed into `_ create_offer_for_ids` here where it gets created if its `None`.

xdustinface (changes requested):
@Quexington After the "avoid mutation of the default `driver_dict` value" changes the test `test_trades.py` is failing here https://github.com/Chia-Network/chia-blockchain/blob/71a8ac2e698726d4b7d43832be382c161c3ecf8c/tests/wallet/cat_wallet/test_trades.py#L199

I debuged into it a bit and it fails here 

https://github.com/Chia-Network/chia-blockchain/blob/22919901f1a1556194c74773488e1dd15e4abe3b/chia/wallet/trading/offer.py#L124-L131

because the `asset_id` calculated with the `puzzle_driver` is not in `driver_dict`. I assume its related to the fact that we before had the drivers of the previous offer creations in `driver_dict` always because it always used the mutated default value. And the code doesn't seem to generate the correct one by itself now. Can you look into it?

chia/wallet/cat_wallet/cat_outer_puzzle.py
Quexington: For this class it's not, but in general it's possible to have one of these classes that does return None.  These should probably be implemented as `Protocol`s but I didn't know about those until Kevin's comments on this PR.

chia/wallet/cat_wallet/cat_wallet.py
Quexington: Yeah we don't want to offer coins that aren't spendable

chia/wallet/wallet.py
Quexington: ditto to above

tests/clvm/test_puzzle_drivers.py
```diff
@@ -0,0 +1,37 @@
...
+    test_also: Dict[str, Any] = {"type": "TEST", "string": "hello"}
+    test_driver["also"] = test_also
+
+    with pytest.raises(ValueError, match="A type is required"):
+        PuzzleInfo(test_driver)
+    solver = Solver(test_driver)
+
+    test_driver["type"] = "TEST"
+    puzzle_info = PuzzleInfo(test_driver)
+
```
xdustinface, line 27:
```suggestion
assert puzzle_info.type() == "TEST"
```

xdustinface (changes requested):
@Quexington So you fixed the tests by just providing the infos for all CATs in the `driver_dict`. Correct me if im wrong but from my understanding it should work even with no `driver_dict` provided given the fact that the wallet has `get_asset_id` and `get_puzzle_info` implemented?

Also im not sure but it looks to me like we have not all of the cases in `create_offer_for_ids` tested. Like for offering/requesting both with wallet_id/asset_id, with and without `driver_dict` and also test with an invalid drivers which differs from the the `get_puzzle_info` result. 

Ah and btw, `pre-commit` is still unhappy about workflows and some `isort` fun :)

xdustinface:
@Quexington The issue is the indentation of https://github.com/Chia-Network/chia-blockchain/blob/79ddd63ac830fe49b79726d6dfa898953156a37e/chia/wallet/trade_manager.py#L372-L382 It needs to be shifted one to the left so that it adds the driver info for each entry of the offer. Right now it only adds it for the last one. So you can revert your "just provide all drivers" changes to the test but like i said it would be nice to see all those cases tested in here :)

xdustinface:
@Quexington Can you explain why we even need to allow providing the `driver_dic` in `_create_offer_for_ids` and also in the RPC interface since from my understanding the driver infos can be fetched in both cases like:

- Offer creation: By calling the wallet functions
- Offer taking: By puzzle matching from the spend bundle

Do i miss something here?

xdustinface (changes requested):
@Quexington See the comments about `__eq__`.
@AmineKhaldi There is a conflict, see https://github.com/AmineKhaldi/chia-blockchain/pull/1

chia/wallet/puzzle_drivers.py
```diff
@@ -0,0 +1,92 @@
...
+        return decode_info_value(PuzzleInfo, value)
+
+    def __eq__(self, other: object) -> bool:
+        for key, value in self.info.items():
+            try:
+                if self[key] != other[key]:  # type: ignore
+                    return False
+            except Exception:
+                return False
+        return True
```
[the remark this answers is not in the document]
xdustinface: Oh i see, can you then you also add tests for this to test the key and value error cases? I got no test failure when removing your `__eq__` implementations.

xdustinface (approved):

emlowe:
I have no problems with this, just double-checked with Amine that this is what he wants to do - and Amine is good. So I'm therefore also good.

wjblanke (approved):
aok

PR #11569, opened 2022-05-19
Remove is not None and length assertion in select_coins()

altendky:
The `coins is not None` check seems unneeded since `select_coins()` is hinted to return `Set[Coin]`.  The length requirement needlessly restricts from being able to request a zero amount.  Either situation would either trigger an exception in the next assert, or trigger the assert itself for non-zero amounts requested.

Draft for:
- [x] Checking for other cases, such as the mentioned CAT wallet
  - DID wallet has this in `main` as well but DID and NFT wallets in the branch do not need changed in regards to this topic so we'll leave the `main` DID wallet untouched.

### DIFF
--- a/chia/wallet/cat_wallet/cat_wallet.py
+++ b/chia/wallet/cat_wallet/cat_wallet.py
@@ -451,7 +451,6 @@ class CATWallet:
             exclude,
             min_coin_amount,
         )
-        assert coins is not None and len(coins) > 0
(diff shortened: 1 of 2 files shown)


jack60612 (dismissed):
Lgtm

jack60612:
Remove in the cat wallet too

jack60612 (approved):

PR #11573, opened 2022-05-19
Increase some of the timeouts to reduce flakiness

mariano54:


mariano54:
replaced by https://github.com/Chia-Network/chia-blockchain/pull/11576

PR #11576, opened 2022-05-19
Fix several flaky tests

mariano54:
Fixes several tests that failed frequently:
* test_new_transaction_and_mempool
* test_wallet_tx_retry
* test_self_pooling_to_pooling
* test_wallet_make_transaction_hop (hopefully? not guaranteed)
* test_periodic_reset

Also bumps the timeout on mempool benchmark from 0.2 to 0.3


arvidn (approved):

emlowe:
I think Bill would like to understand why the run time increases - however, increasing the time seems reasonable for now until we can generate some graphs or understand the general variance in the run time.

mariano54:
let's merge @wjblanke

PR #11578, opened 2022-05-20
Bind port 0 to fix race condition when grabbing available ports

mariano54:
Removes the vast majority of calls to `find_available_listen_port`, and defaults setup_node functions to starting on port 0. The server and rpc_server now allows using port 0, and sets the correct variables to the newly aquicred port.

For some reason the timelord performance gets really bad on CI if we set port 0 for the timelord server (which uses different server code, not aiohttp). So I left those with the old code for now. More work can be done to improve setup_full_system and make it nicer. 

- [x] all CI tests passing
- [x] cleanup the ugly code in setup_full_system
- [x] figure out if we can pass 0 to aiohttp somehow


chia/server/server.py
```diff
@@ -263,13 +263,17 @@ async def start_server(self, on_connect: Callable = None):
...
         self.site = web.TCPSite(
             self.runner,
-            port=self._port,
+            host="0.0.0.0",
+            port=int(self._port),
             shutdown_timeout=3,
             ssl_context=ssl_context,
         )
         await self.site.start()
+        self._port = self.runner.addresses[0][1]
```
arvidn, line 276: what's at `addresses[1]` and at `addresses[0][0]`?

chia/rpc/rpc_server.py
```diff
@@ -324,13 +325,15 @@ async def start_rpc_server(
         daemon_connection = asyncio.create_task(rpc_server.connect_to_daemon(self_hostname, daemon_port))
     runner = web.AppRunner(app, access_log=None)
     await runner.setup()
+
```
arvidn, line 328: I was going to suggest that asking for the port later (like `get_port()`) might be a nicer model. But I realize we drop the task handle on the floor here and just detach it here.. so I suppose that isn't an option until this is fixed

tests/setup_services.py
```diff
@@ -327,10 +327,14 @@ def stop():
 
 
 async def setup_timelord(
-    port, full_node_port, rpc_port, vdf_port, sanitizer, consensus_constants: ConsensusConstants, b_tools: BlockTools
+    full_node_port,
+    sanitizer,
+    consensus_constants: ConsensusConstants,
+    b_tools: BlockTools,
+    vdf_port: uint16 = uint16(0),
+    rpc_port: uint16 = uint16(0),
```
arvidn, line 335: maybe we could even remove these port parameters here and above. Any service where we can set the port to 0, we should just hard code that in the tests.

arvidn (dismissed):
my main comment is that I would encourage you to delete as much as possible of the left-over from when we *had* to specify a port in the tests. I have a feeling the tests could still be cut shorter by hard coding listen port 0 everywhere where it works.

chia/server/server.py
mariano54: addresses that are listening for this running.
[0][0] is host, [0][1] is port

tests/setup_services.py
mariano54: hmm but in the test we would then have to write to the config, which seems like a weird thing to do in the test no?

mariano54: I managed to remove a lot of them

arvidn (approved):

chia/rpc/rpc_server.py
altendky: We can't just directly start using this, but maybe look at https://trio.readthedocs.io/en/stable/reference-core.html#trio.Nursery.start would be useful as a reference for launching concurrent tasks and getting some info out of them after they 'start up'.

PR #11585, opened 2022-05-20
Extend on chain metadata update/API

ytx1991:


chia/rpc/wallet_rpc_api.py
```diff
@@ -1441,12 +1447,17 @@ async def nft_get_info(self, request: Dict) -> Optional[Dict]:
     async def nft_add_uri(self, request) -> Dict:
         assert self.service.wallet_state_manager is not None
         wallet_id = uint32(request["wallet_id"])
-        uri = request["uri"]
+        # Note metadata updater can only add one uri for one field at once
```
paninaro, line 1450: So this means that if I try to add a meta_uri and license_uri in the same call, the update will fail?

If that's the case, we should raise a ValueError indicating that only a single URI can be added per call.

paninaro: Or is the limitation once per spend?

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -83,21 +83,29 @@ def get_nft_info_from_puzzle(puzzle: Program, nft_coin: Coin) -> NFTInfo:
...
+        str(uncurried_nft.data_hash).upper(),
+        meta_uris,
+        str(uncurried_nft.meta_hash).upper(),
+        license_uris,
+        str(uncurried_nft.license_hash).upper(),
+        uint64(uncurried_nft.series_total.as_int()),
+        uint64(uncurried_nft.series_total.as_int()),
+        LAUNCHER_PUZZLE_HASH.hex().upper(),
+        str(uncurried_nft.metdata_updater_hash).upper(),
+        str(uncurried_nft.metadata).upper(),
```
paninaro, line 108: No uppercase conversion on this because it will cause problems with the case-sensitive comparison when creating an offer with the metadata. Also, the URLs in the metadata should not be modified.

```suggestion
        str(uncurried_nft.metadata),
```

chia/rpc/wallet_rpc_api.py
ytx1991: You can add one uri for data-uri, metadata-uri and license-uri independently in spend, check the unit test.

chia/wallet/nft_wallet/nft_puzzles.py
ytx1991: Ok, but the url will not be changed by this, since their are in hex.

paninaro: Oh I see

paninaro: Ok disregard if the data is still a hex string at this point.

chia/rpc/wallet_rpc_api.py
paninaro: The "for one field at once" part of the comment confused me. Maybe the comment can clarify that. Something like:

# Note metadata updater can only add a single URI of each type per spend. Adding multiple URIs of the same type will require multiple spends.

paninaro (approved):
Looks good to me, but I think someone else needs to sign off on the Chialisp changes.

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -83,21 +83,29 @@ def get_nft_info_from_puzzle(puzzle: Program, nft_coin: Coin) -> NFTInfo:
...
-        uncurried_nft.data_hash.as_python().hex().upper(),
-        [],
-        "",
-        [],
-        "",
-        "NFT0",
-        uint64(1),
-        uint64(1),
+        str(uncurried_nft.data_hash).upper(),
+        meta_uris,
```
trepca, line 100: @paninaro for `*_uris` fields, given that they should all be the same and only added if url is obsolete or wrong, shouldn't we be returning only the first and most recent one vs all?
returning old urls might cause security problems if NFT metadata gets compromised and NFT owner wants to effectively remove the old one

trepca (changes requested):

chia/wallet/nft_wallet/nft_puzzles.py
paninaro: But the metadata hash guards against that case, as modifying any of the URLs will invalidate the hash verification performed by the frontend.

I would ask that we return the first 3 or so in case the most recent addition(s) aren't available due to transient network/service issues. But sure, we probably don't need to return the entire list if it contains many entries.

chia/wallet/nft_wallet/nft_puzzles.py
mariano54: we should not be using hex strings in most places in the backed. You should probably just use bytes32. When you return something in the rpc_api, the middleware will convert everything to hexstr. Using strings internally will cause more bugs and hidden inconsistencies.

ytx1991: 1. This is only for the frontend, so bytes have to be serialized as hex in json.
2. I'm okay with lower case as long as we commit use it everywhere. The reason I use upper is because the upper case alignment is better than lower case. In lower case, even the hex length are same, they may look different.
For example,
dddddddddddddddddddd
ffffffffffffffffffff

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -83,21 +83,29 @@ def get_nft_info_from_puzzle(puzzle: Program, nft_coin: Coin) -> NFTInfo:
...
+        str(uncurried_nft.meta_hash).upper(),
+        license_uris,
+        str(uncurried_nft.license_hash).upper(),
+        uint64(uncurried_nft.series_total.as_int()),
+        uint64(uncurried_nft.series_total.as_int()),
+        LAUNCHER_PUZZLE_HASH.hex().upper(),
+        str(uncurried_nft.metdata_updater_hash).upper(),
```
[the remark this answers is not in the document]
ytx1991: good catch

chia/wallet/nft_wallet/nft_puzzles.py
mariano54: You don't need to serialize manually in the backend. Just retun bytes objects and the RPC server will translate everything to hex. It's bad practice to use hexstrings in the backend

chia/wallet/nft_wallet/nft_info.py
```diff
@@ -40,11 +40,17 @@ class NFTInfo(Streamable):
...
-
-    edition_count: uint64
+    series_total: uint64
     """How many NFTs in the current series"""
 
-    edition_number: uint64
+    series_number: uint64
     """Number of the current NFT in the series"""
+
+    launcher_puzhash: str
```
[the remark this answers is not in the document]
ytx1991: It's required by Jeff. If we don't pass this to the frontend then we need to hardcode this in the frontend, which is not good for code maintaining

chia/wallet/puzzles/nft_metadata_updater_default.clvm
```diff
@@ -4,18 +4,65 @@
...
+        ()
+      )
+      METADATA
+    )
+  )
 
   ; main
   ; returns ((new_metadata new_metadata_updater_puzhash) conditions)
-  (list (list (if solution (add_url CURRENT_METADATA solution) CURRENT_METADATA) METADATA_UPDATER_PUZZLE_HASH) 0)
+  ; Supported fields
```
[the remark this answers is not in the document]
ytx1991: It's a trade-off between functions, cost, and code readability.

tests/wallet/nft_wallet/test_nft_clvm.py
```diff
@@ -58,11 +59,19 @@ def test_update_metadata() -> None:
...
     condition_list = [make_create_coin_condition(destination.get_tree_hash(), my_amount, [])]
-    condition_list.append([-24, NFT_METADATA_UPDATER, "https://www.chia.net/img/branding/chia-logo-2.svg"])
+    condition_list.append(
+        [-24, NFT_METADATA_UPDATER, [("u", "https://url1"), ("mu", "https://url2"), ("lu", "https://url3")]]
+    )
+
     metadata = [
         ("u", ["https://www.chia.net/img/branding/chia-logo.svg"]),
         ("h", 0xD4584AD463139FA8C0D9F68F4B59F185),
+        ("mu", []),
```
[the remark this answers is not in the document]
ytx1991: I haven't try that, but I think it's a part of standard. These fields must be a list of string.

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -83,21 +83,29 @@ def get_nft_info_from_puzzle(puzzle: Program, nft_coin: Coin) -> NFTInfo:
     data_uris = []
     for uri in uncurried_nft.data_uris.as_python():
         data_uris.append(str(uri, "utf-8"))
+    meta_uris = []
```
mariano54, line 86: we should use type hinting here:
meta_uris: List[str] = []

chia/wallet/nft_wallet/nft_puzzles.py
mariano54: lower case vs upper case shoud not be an arguement, because we do not use hex in the backend.

chia/wallet/nft_wallet/nft_puzzles.py
ytx1991: good catch

chia/wallet/nft_wallet/nft_puzzles.py
ytx1991: We have another data structure (UncurriedNFT) for backend usage. This NFTInfo is only used for frontend.

chia/wallet/nft_wallet/nft_info.py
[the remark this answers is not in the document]
trepca: this can be a constant on the backend is what I meant

chia/wallet/nft_wallet/nft_puzzles.py
trepca: non-JSON types are automatically serialized into json types when RPC serializes the data returned, hence no need to to do it in the data class

ytx1991: I tried to switch to bytes/bytes32, the issue is the fields are serialized with "0x" prefix. The frontend may need to change. @paninaro I'm okay with any approach, what do you think?

paninaro: Sure, we can handle a "0x" prefix in the GUI

chia/wallet/puzzles/nft_metadata_updater_default.clvm
```diff
@@ -1,15 +1,29 @@
 (mod (CURRENT_METADATA METADATA_UPDATER_PUZZLE_HASH solution)
```
trepca, line 1: it's good progress, but still quite complex and not sure why we need to support multiple field updating at once. 
Here's code that's much simpler and only does one iteration over METADATA: 
```suggestion
(mod (CURRENT_METADATA METADATA_UPDATER_PUZZLE_HASH (key new_url))

  ; METADATA and METADATA_UPDATER_PUZZLE_HASH are passed in as truths from the layer above

  ; This program returns ((new_metadata new_metadata_updater_puzhash) conditions)

  ; once we find the key we don't need to continue looping

  (defun add_url (METADATA key new_url)
    (if METADATA
      (if (= (f (f METADATA)) key)
        (c (c key (c new_url (r (f METADATA)))) (r METADATA))
        (c (f METADATA) (add_url (r METADATA) key new_url))
      )
      ()
    )
  )

  ; main
  ; returns ((new_metadata new_metadata_updater_puzhash) conditions)
  (list
    (list
        (if (all key new_url)
            (if (any (= key "mu") (= key "lu") (= key "u"))
                (add_url CURRENT_METADATA key new_url)
                CURRENT_METADATA
            )
            CURRENT_METADATA
         )
        METADATA_UPDATER_PUZZLE_HASH)
    0
  )
)
```

ytx1991: For the users, they may not want to spend the NFT multiple times if they want to update multiple fields, kind of expensive.

trepca: You're assuming that most common scenario is that all 3 links will become obsolete at once and need to be changed, and I'd say it's much more likely only one of them will become invalid and require another one. Because in current form it's much more expensive adding a single url, right?

PR #11589, opened 2022-05-21
add tests.util.misc.assert_maximum_duration for benchmark assertions

altendky:
Draft for:
- [x] Discussion around garbage collecting's effect on the timing 'accuracy'
- [x] Add back some logging/printing
- [x] Double check all the times
- [x] Review options related to explicit garbage collection
- [ ] Review duration limits against present times on the actual benchmark runners
- [ ] ~~How do each of garbage collection and perf counter contribute to the improved consistency relative to the irrelevant but relevant commits~~
- [x] Separate measurement from assertion (maybe)
- [x] Check and account for induced overhead.
- [x] Think through the api around `assert_maximum_duration()`
- [x] Review type error ignores and at least provide an explanatory comment if they are deemed appropriate

### DIFF
--- a/tests/core/full_node/test_mempool.py
+++ b/tests/core/full_node/test_mempool.py
@@ -2238,10 +2238,10 @@ class TestMaliciousGenerators:
         ],
     )
     @pytest.mark.benchmark
-    def test_duplicate_large_integer_ladder(self, opcode, softfork_height):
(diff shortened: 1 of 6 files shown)


arvidn (dismissed):

arvidn (approved):
there are a lot of options and features that we aren't using in here. like GC mode and custom clock.

PR #11591, opened 2022-05-21
increment the dirty counter when setting height-to-hash map entries

arvidn:
make sure we use `__set_hash()` when updating the height-to-hash map. That's where we increment the `__dirty` counter, which in turn triggers flushing of this map to disk.

Currently, we don't end up flushing this to disk during sync.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -296,8 +296,8 @@ class Blockchain(BlockchainInterface):
                 )
                 raise
 
-            # This is done outside the try-except in case it fails, since we do not want to revert anything if it does
-            await self.__height_map.maybe_flush()
+        # This is done outside the try-except in case it fails, since we do not want to revert anything if it does
+        await self.__height_map.maybe_flush()
 
         if state_change_summary is not None:
             # new coin records added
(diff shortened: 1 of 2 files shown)


chia/consensus/blockchain.py
```diff
@@ -296,8 +296,8 @@ async def receive_block(
                 )
                 raise
 
-            # This is done outside the try-except in case it fails, since we do not want to revert anything if it does
-            await self.__height_map.maybe_flush()
```
arvidn, line 300: I moved this out of the `write_db()` block. There's no point in keeping the DB write lock while flushing the height map.

chia/full_node/block_height_map.py
```diff
@@ -130,9 +130,7 @@ async def create(cls, blockchain_dir: Path, db: DBWrapper2) -> "BlockHeightMap":
     def update_height(self, height: uint32, header_hash: bytes32, ses: Optional[SubEpochSummary]):
         # we're only updating the last hash. If we've reorged, we already rolled
         # back, making this the new peak
-        idx = height * 32
-        assert idx <= len(self.__height_to_hash)
-        self.__height_to_hash[idx : idx + 32] = header_hash
+        self.__set_hash(height, header_hash)
```
arvidn, line 133: see __set_hash() here: https://github.com/Chia-Network/chia-blockchain/pull/11591/files#diff-c0265f35094c3ceaa2cba46d0d4679d53b44fffe34996615de8cdb877d4cef21L211

mariano54: should we add back the assert there?

arvidn: good point. I was thinking the assert was part of `__set_hash()`, which maybe it should be

mariano54 (approved):

PR #11596, opened 2022-05-21
allow bigger chunks of bind variables per SQL statement 

neurosis69:
I'd suggest to use parameter limits according to the SQLite version the full node is running with.
This is achieved using a dbwrapper constant and replacing the currently used constant  `MAX_SQLITE_PARAMETERS`.

To match the parameter name used  [in sqlite itself](https://www.sqlite.org/lang_expr.html#varparam), I also renamed the constant to `SQLITE_MAX_VARIABLE_NUMBER`.

### Performance impact
We should expect a slight improvement, also less parsing and syscalls.

This was tested using my ansible roles and the standard `long sync` operation on full node from main branch, starting from height 1537152 to 1556960. This should cover a short enough period with dust and regular tx in between.

System|Testcase|Runtime|%
---|---|---|---
RPi4B-8G|main|4h03m48s|100%
RPi4B-8G|PR|3h52m07s|95,21%

Test Limitation: Each sync test was only done 1x, variations can be high on such tests.

### Limits In SQLite:
SQLite versions < 3.32.0 allow up to 999 bind variables per single SQL statement
with versions >= 3.32.0 up to 32766 bind variables are possible
[Manual](https://www.sqlite.org/limits.html#max_variable_number)

### Identified enforced limits in chia

- minimal SQLite version for chia is 3.8
- in `full_node/coin_store.py` a constant `MAX_SQLITE_PARAMETERS = 900` is set and used to split parameter in chunks
- in `consensus/blockchain.py` 
  - the function `get_block_records_at` is using a default `batch_size: int = 900` and
  later in the function an `assert batch_size < 999`
  - `get_block_records_at` is calling the function `get_block_records_by_hash` in `full_node/block_store.py`
  - `get_block_records_by_hash` is not using `chia.util.chunks` right now, so not as easy to change

### DIFF
--- a/chia/util/db_wrapper.py
+++ b/chia/util/db_wrapper.py
@@ -7,6 +7,11 @@ from typing import AsyncIterator, Dict, Optional
 import aiosqlite
 
 
+if aiosqlite.sqlite_version_info < (3, 32, 0):
+    SQLITE_MAX_VARIABLE_NUMBER = 900
+else:
+    SQLITE_MAX_VARIABLE_NUMBER = 32700
+
 class DBWrapper:
     """
     This object handles HeaderBlocks and Blocks stored in DB used by wallet.
@@ -42,7 +47,6 @@ class DBWrapper2:
     _in_use: Dict[asyncio.Task, aiosqlite.Connection]
     _current_writer: Optional[asyncio.Task]
     _savepoint_name: int
-    SQLITE_MAX_VARIABLE_NUMBER: int
 
     async def add_connection(self, c: aiosqlite.Connection) -> None:
         # this guarantees that reader connections can only be used for reading
@@ -60,10 +64,6 @@ class DBWrapper2:
         self._in_use = {}
         self._current_writer = None
         self._savepoint_name = 0
-        if aiosqlite.sqlite_version_info < (3, 32, 0):
-            self.SQLITE_MAX_VARIABLE_NUMBER = 999
-        else:
-            self.SQLITE_MAX_VARIABLE_NUMBER = 32766
 
 
     async def close(self) -> None:
--- a/chia/util/db_wrapper.py
+++ b/chia/util/db_wrapper.py
@@ -42,6 +42,7 @@ class DBWrapper2:
     _in_use: Dict[asyncio.Task, aiosqlite.Connection]
     _current_writer: Optional[asyncio.Task]
     _savepoint_name: int
+    SQLITE_MAX_VARIABLE_NUMBER: int
 
     async def add_connection(self, c: aiosqlite.Connection) -> None:
         # this guarantees that reader connections can only be used for reading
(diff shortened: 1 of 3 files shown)


mariano54:
from the manual:
the maximum value of a host parameter number is SQLITE_MAX_VARIABLE_NUMBER, which defaults to 999 for SQLite versions prior to 3.32.0 (2020-05-22) or 32766 for SQLite versions after 3.32.0. 
Which means it's just a default, we can set it to whatever we want right?

neurosis69:
> Which means it's just a default, we can set it to whatever we want right? EDIT: or maybe we can only lower but not increase it?

We would need to recompile sqlite to raise the limits, if I understand it correctly.

When logging the sql commands during sync, then I only noticed a need for more bind variables per SQL statement for coin record inserts/removals. This may be different for wallet syncs though, didn't dive into that too much yet.

But I am currently also testing several other sqlite compile options which could have a positive impact with now drawback (hopefully). So I can just add a higher MAX_VARIABLE_NUMBER limit and do a testrun as well.

Using the following options during my tests, most of them taken from the [sqlite compile recommendations](https://www.sqlite.org/compile.html#recommended_compile_time_options).
```
 CFLAGS="-O2 -DSQLITE_ENABLE_STAT4=1 -DSQLITE_DEFAULT_MEMSTATUS -DSQLITE_LIKE_DOESNT_MATCH_BLOBS -DSQLITE_MAX_EXPR_DEPTH=0 -DSQLITE_OMIT_PROGRESS_CALLBACK -DSQLITE_OMIT_SHARED_CACHE -DSQLITE_USE_ALLOCA -DSQLITE_DIRECT_OVERFLOW_READ"
```

chia/util/db_wrapper.py
```diff
@@ -7,6 +7,11 @@
 import aiosqlite
 
 
+if aiosqlite.sqlite_version_info < (3, 32, 0):
```
arvidn, line 10: I think ensuring that this comparison is correct is key to the correctness of this change

arvidn:
right. the "default" mentioned in the manual is the default *limit* when building sqlite. We can't choose that limit, we have to live with it.

@neurosis69 what version of sqlite did you run that test with?

I think this change makes sense. I'm interested in running this on a more deterministic benchmark though.

arvidn:
@neurosis69 would you mind running `pre-commit run --all` to fix up the patch?

chia/util/db_wrapper.py
neurosis69: That's what I tested:

```python
>>> import aiosqlite
>>> aiosqlite.sqlite_version_info
(3, 34, 1)
>>> if aiosqlite.sqlite_version_info < (3, 32, 0):
...   SQLITE_MAX_VARIABLE_NUMBER = 999
... else:
...   SQLITE_MAX_VARIABLE_NUMBER = 32766
...
>>> print(SQLITE_MAX_VARIABLE_NUMBER)
32766
>>> if aiosqlite.sqlite_version_info < (3, 4, 0):
...   SQLITE_MAX_VARIABLE_NUMBER = 999
... else:
...   SQLITE_MAX_VARIABLE_NUMBER = 32766
...
>>> print(SQLITE_MAX_VARIABLE_NUMBER)
32766
>>> if aiosqlite.sqlite_version_info > (3, 32, 0):
...   SQLITE_MAX_VARIABLE_NUMBER = 999
... else:
...   SQLITE_MAX_VARIABLE_NUMBER = 32766
...
>>> print(SQLITE_MAX_VARIABLE_NUMBER)
999
```

neurosis69:
> @neurosis69 what version of sqlite did you run that test with?

Tested with 3.34.1 and 3.38.5.

arvidn (approved):

PR #11597, opened 2022-05-21
Fix a few additional flaky tests

mariano54:
The problem was that block tools was being created outside of an async context, and running the async functions using `get_running_loop` can potentially use the wrong loop.

Fixes the following flaky test files:
* test_full_node_store
* test_singleton_lifecycle

### DIFF
--- a/tests/core/full_node/stores/test_full_node_store.py
+++ b/tests/core/full_node/stores/test_full_node_store.py
@@ -39,9 +39,8 @@ def cleanup_keyring(keyring: TempKeyring):
 
 @pytest_asyncio.fixture(scope="function")
 async def custom_block_tools() -> BlockTools:
-    temp_keyring = TempKeyring()
-    keychain = temp_keyring.get_keychain()
-    atexit.register(cleanup_keyring, temp_keyring)  # Attempt to cleanup the temp keychain
+    with TempKeyring() as keychain:
+        yield await create_block_tools_async(constants=test_constants, keychain=keychain)
 
     return await create_block_tools_async(constants=test_constants, keychain=keychain)
 
--- a/tests/core/full_node/stores/test_full_node_store.py
+++ b/tests/core/full_node/stores/test_full_node_store.py
@@ -32,10 +32,6 @@ test_constants = test_constants_original.replace(**{"DISCRIMINANT_SIZE_BITS": 32
 log = logging.getLogger(__name__)
 
 
-def cleanup_keyring(keyring: TempKeyring):
-    keyring.cleanup()
-
-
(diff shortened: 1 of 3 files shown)


tests/core/full_node/stores/test_full_node_store.py
```diff
@@ -29,18 +29,21 @@
...
-keychain = temp_keyring.get_keychain()
-atexit.register(cleanup_keyring, temp_keyring)  # Attempt to cleanup the temp keychain
-test_constants = test_constants_original.replace(**{"DISCRIMINANT_SIZE_BITS": 32, "SUB_SLOT_ITERS_STARTING": 2 ** 12})
-bt = create_block_tools(constants=test_constants, keychain=keychain)
-
+@pytest_asyncio.fixture(scope="function")
+async def custom_block_tools() -> BlockTools:
+    temp_keyring = TempKeyring()
+    keychain = temp_keyring.get_keychain()
+    atexit.register(cleanup_keyring, temp_keyring)  # Attempt to cleanup the temp keychain
```
paninaro, line 44: Thanks for cleaning this up. I'm pretty sure that we can use the more recent pattern of:

```suggestion
    with TempKeyring() as keychain:
        yield await create_block_tools_async(constants=test_constants, keychain=keychain)
```

There's some cleanup tasks that the TempKeychain performs during __exit__.

tests/core/full_node/stores/test_full_node_store.py
```diff
@@ -29,18 +28,18 @@
...
+log = logging.getLogger(__name__)
 
 
-temp_keyring = TempKeyring()
-keychain = temp_keyring.get_keychain()
-atexit.register(cleanup_keyring, temp_keyring)  # Attempt to cleanup the temp keychain
-test_constants = test_constants_original.replace(**{"DISCRIMINANT_SIZE_BITS": 32, "SUB_SLOT_ITERS_STARTING": 2 ** 12})
-bt = create_block_tools(constants=test_constants, keychain=keychain)
+def cleanup_keyring(keyring: TempKeyring):
+    keyring.cleanup()
```
paninaro, line 36: Should be able to remove this now.

tests/core/full_node/test_conditions.py
```diff
@@ -36,12 +34,6 @@ def cleanup_keyring(keyring: TempKeyring):
     keyring.cleanup()
```
paninaro, line 34: Should be able to remove this one too.

paninaro (dismissed):
lgtm. The cleanup_keyring() methods aren't used anymore, so those could be removed.

paninaro (approved):

PR #11601, opened 2022-05-23
Show test output in benchmarks workflow

altendky:


altendky:
We ought to really collect the results in a structure form etc, but at least until then this gives us _some_ ability to check the numbers via CI.

aqk (approved):

emlowe (approved):
Seems good so we can start to have a collect these details

wjblanke (approved):
aok

PR #11605, opened 2022-05-23
Allow switching keys during sync

mariano54:


trepca (approved):
Not aware what the issue was with switching keys, but makes sense to create a separate task for such a long lasting operation.

wjblanke (approved):
aok

PR #11606, opened 2022-05-23
Huge speedup in trusted wallet sync

mariano54:
This provides about a 50x speedup in trusted sync, when connected to a local node, for situations when there are a lot of TXs.

Previous time to sync 1000 tx:  90 seconds
New time: 2 seconds

This assumes the coins are all received in the same block. If the coins are received approximately 1 per block, the performance will not be this fast, due to fetching the header block to receive the timestamp


chia/wallet/wallet_state_manager.py
```diff
@@ -637,13 +638,18 @@ async def new_coin_state(
...
+                    wallet_id, type=TransactionType.OUTGOING_TRADE
+                )
+                all_incoming_trades: List[TransactionRecord] = await self.tx_store.get_all_transactions_for_wallet(
+                    wallet_id, type=TransactionType.INCOMING_TRADE
+                )
+                all_outgoing_std_tx: List[TransactionRecord] = await self.tx_store.get_all_transactions_for_wallet(
+                    wallet_id, type=TransactionType.OUTGOING_TX
+                )
+                all_txs_per_wallet[wallet_id] = all_outgoing_std_tx + all_outgoing_trades + all_incoming_trades
+            all_outgoing_tx: List[TransactionRecord] = all_txs_per_wallet[wallet_id]
```
arvidn, line 652: `all_outgoing` used to only contain outgoing transactions, but if I'm reading this right, it will now also contain incoming trades and outgoing trades.

Was this change deliberate? if so, should the variable still be called `all_outgoing_tx`?

arvidn:
Can you explain where the speed-up happens?

mariano54:
> Can you explain where the speed-up happens?

There are two types of transactions: the ones that we make (and therefore start as unconfirmed) and the ones that we receive (and therefore are confirmed as soon as we see them). That is what we refer to as `outgoing_tx` here. Creating a trade or accepting a trade are also outgoing since we are part of those TX.  


Let's say we have 50k transactions. This code was calling coin.name() for every single addition in every single transaction we receive. During wallet sync, we don't have any outgoing transactions (or have very few). Every transaction is incoming. 

Therefore, changing this list to not include every single thing means the CPU operations change from 50k to negligible

chia/wallet/wallet_state_manager.py
```diff
@@ -637,13 +638,18 @@ async def new_coin_state(
...
+            if wallet_id not in all_txs_per_wallet:
+                all_outgoing_trades: List[TransactionRecord] = await self.tx_store.get_all_transactions_for_wallet(
+                    wallet_id, type=TransactionType.OUTGOING_TRADE
+                )
+                all_incoming_trades: List[TransactionRecord] = await self.tx_store.get_all_transactions_for_wallet(
+                    wallet_id, type=TransactionType.INCOMING_TRADE
+                )
+                all_outgoing_std_tx: List[TransactionRecord] = await self.tx_store.get_all_transactions_for_wallet(
+                    wallet_id, type=TransactionType.OUTGOING_TX
+                )
```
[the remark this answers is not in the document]
mariano54: True, but we have a cache here, and I didn't want to change too much. I don't see this as a bottleneck atm.

chia/wallet/wallet_state_manager.py
```diff
@@ -739,7 +747,7 @@ async def new_coin_state(
                         # Reorg rollback adds reorged transactions so it's possible there is tx_record already
                         # Even though we are just adding coin record to the db (after reorg)
                         tx_records: List[TransactionRecord] = []
-                        for out_tx_record in all_outgoing:
+                        for out_tx_record in all_outgoing_tx:
                             for rem_coin in out_tx_record.removals:
                                 if rem_coin.name() == coin_state.coin.name():
```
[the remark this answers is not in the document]
mariano54: good idea yeah

chia/wallet/wallet_state_manager.py
```diff
@@ -965,15 +976,15 @@ async def coin_added(
...
+                name=coin_name,
                 memos=[],
             )
             await self.tx_store.add_transaction_record(tx_record, True)
         else:
             records: List[TransactionRecord] = []
             for record in all_outgoing_transaction_records:
                 for add_coin in record.additions:
-                    if add_coin.name() == coin.name():
+                    if add_coin.name() == coin_name:
```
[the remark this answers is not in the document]
mariano54: yes

chia/wallet/wallet_state_manager.py
arvidn: I see, so this list (`all_outgoing`) is a lot bigger prior to your patch. Those 3 DB lookups are small and that's the bulk of the gain.

chia/wallet/wallet_state_manager.py
```diff
@@ -965,15 +962,15 @@ async def coin_added(
...
             )
             await self.tx_store.add_transaction_record(tx_record, True)
         else:
             records: List[TransactionRecord] = []
-            for record in all_outgoing_transaction_records:
+            for record in all_unconfirmed_transaction_records:
                 for add_coin in record.additions:
-                    if add_coin.name() == coin.name():
+                    if add_coin == coin:
                         records.append(record)
```
arvidn, line 974: one more question about this pattern. We have this block, essentially, in 4 places. What I find a bit odd is that this code seems to expect that `coin` (or `coin_state.coin` in other instances) can occur multiple times in `record.additions` / `out_tx_record.removals`.

I would be surprised if they can. Once we find the coin, why do we keep searching for it?

This boils down to that these loops, at least the inner loop, should probably be changed to an actual search call. e.g.
```
for record in all_outgoing_transaction_records:
    if coin in record.additions:
        records.append(record)
        break
```

chia/wallet/wallet_state_manager.py
```diff
@@ -600,17 +600,17 @@ async def new_coin_state(
...
 
         for coin_state_idx, coin_state in enumerate(coin_states):
+            coin_name: bytes32 = coin_state.coin.name()
             wallet_info: Optional[Tuple[uint32, WalletType]] = await self.get_wallet_id_for_puzzle_hash(
                 coin_state.coin.puzzle_hash
             )
-            local_record: Optional[WalletCoinRecord] = await self.coin_store.get_coin_record(coin_state.coin.name())
-            self.log.debug(f"{coin_state.coin.name()}: {coin_state}")
+            local_record: Optional[WalletCoinRecord] = await self.coin_store.get_coin_record(coin_name)
+            self.log.debug(f"{coin_name}: {coin_state}")
```
trepca, line 613: this serializes `coin_state` even if log_level > DEBUG, unless you use logging string formatting

trepca (approved):

chia/wallet/wallet_state_manager.py
```diff
@@ -802,8 +802,16 @@ async def new_coin_state(
...
             elif coin_state.created_height is not None and coin_state.spent_height is None:
                 await self.coin_added(
-                    coin_state.coin, coin_state.created_height, all_unconfirmed, wallet_id, wallet_type, coin_name
+                    coin_state.coin,
+                    coin_state.created_height,
+                    all_unconfirmed,
+                    wallet_id,
+                    wallet_type,
+                    coin_name,
+                    skip_lookup=local_record is None,
```
arvidn, line 811: it's not obvious to me why it's safe to skip this check here

mariano54: If local record is none, there is no reason to do the same exact lookup again. It will just be none, and we will continue

wjblanke (approved):
aok

PR #11607, opened 2022-05-23
Fix autoinsert xfails.

fchirica:
Changed the autoinsert tests to assert the average node height is around `log(num_nodes)` and the maximum doesn't exceed the average by too far. As we've changed the autoinsert algorithm, the previous behaviour no longer applies.

### DIFF
--- a/tests/core/data_layer/test_data_store.py
+++ b/tests/core/data_layer/test_data_store.py
@@ -1,5 +1,5 @@
 import itertools
-import math
+import statistics
 import logging
 from typing import Awaitable, Callable, Dict, List, Optional, Tuple, Set
 from random import Random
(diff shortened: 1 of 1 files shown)


emlowe (dismissed):

tests/core/data_layer/test_data_store.py
```diff
@@ -400,91 +401,59 @@ async def test_inserting_duplicate_key_fails(
...
+        node_hash = await data_store.autoinsert(key, value, tree_id, hint_keys_values)
+        hashes.append(node_hash)
 
-    assert result == expected
+    avg_height = 0
+    for node_hash in hashes:
+        ancestors = await data_store.get_ancestors_optimized(node_hash, tree_id)
+        assert len(ancestors) <= 14
+        avg_height += len(ancestors)
+    assert math.ceil(avg_height // 2000) == 11
```
altendky, line 422: `//` already guarantees an integer.  With the `math.ceil()` i'm guess you meant `/` here?  Though that will probably end up `== 12` instead.

I think the above note is an error that should be fixed.  Below is a non-mandatory alternative to consider.  `:]`

I really like fewer assertions and to do them on data that will print interesting diagnostics on failure.  In this case this will notify you both of all the problematic node hashes and their heights.  I'm not certain that the second assertion is exactly right, but I think it matches the intent?  Or maybe it's off by one.

```python
heights = {node_hash: await data_store.get_ancestors_optimized(node_hash, tree_id) for node_hash in hashes}
too_tall = {hash: height for hash.height in heights.items() if height > 14}
assert too_tall == {}
assert 10 <= statistics.mean(heights.values()) <= 11
```

altendky (changes requested):

altendky (approved):

PR #11608, opened 2022-05-23
test installer on fedora:36, drop 33

altendky:


TheLastCicada (approved):
Looks good to me - straightforward.   We should regularly do this to keep up with fedora releases

PR #11618, opened 2022-05-24
tests: Skip `plot_sync_callback` if `delta` is `None`

xdustinface:
This reduces a lot pointless error logs when running plot sync tests.


altendky:
Consider the hinting error in https://github.com/Chia-Network/chia-blockchain/pull/11620 as there seems to be more than just an optionality issue, maybe.

xdustinface:
Can be merged via https://github.com/Chia-Network/chia-blockchain/pull/11620

PR #11622, opened 2022-05-24
[NO REVIEWS] release/1.4.0 dogfooding builds

AmineKhaldi:


AmineKhaldi:
This served us well, but we've been dogfooding from the release branch for a while so this is no longer needed.

PR #11629, opened 2022-05-24
fix jsonify bool

arvidn:
without this `bool` is converted to an `int`. I would have expected it to throw an exception.

### DIFF
--- a/tests/core/util/test_jsonify.py
+++ b/tests/core/util/test_jsonify.py
@@ -1,9 +1,13 @@
 from dataclasses import dataclass
-from typing import List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.ints import uint32
 from chia.util.streamable import Streamable, recurse_jsonify, streamable
+
+
+def dict_with_types(d: Dict[str, Any]) -> Dict[str, Any]:
+    return {k: (v, type(v)) for k, v in d.items()}
 
 
 def test_primitives() -> None:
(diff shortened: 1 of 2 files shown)


tests/core/util/test_jsonify.py
```diff
@@ -15,21 +15,42 @@ class PrimitivesTest(Streamable):
...
     )
 
-    assert t.to_json_dict() == {
+    assert t1.to_json_dict() == {
         "a": 123,
         "b": None,
         "c": "foobar",
         "d": "0x00010001",
         "e": "0x000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f",
+        "f": True,
```
altendky, line 35: This doesn't test what you want it to.  It also passes with `1` here.  `bool` actually inherits from `int`, for better or `*cough*` worse.  `True` has a value of 1 and `False` has a value of 0.

The following snippets might be a reasonable choice to address this.  I like this form of creating the dictionaries so that pytest's failure output can present the most info.

```python3
def dict_with_types(d):
    return {k: (v, type(v)) for k, v in d.items()}
```

And then assert such as this.

```python3
assert dict_with_types(t1.to_json_dict()) == dict_with_types(e1)
```

altendky (changes requested):
This does leave a little difference from before https://github.com/Chia-Network/chia-blockchain/pull/11537 in that the 8/16/32 bit fixed width integer types are now converted to regular `int`s here.  It isn't obvious what trouble that would cause.

Separately, `big_ints` could have been removed before.  I have offered that in https://github.com/Chia-Network/chia-blockchain/pull/11630.  The effect this had was not to provide special handling of the larger integers that may not be accurately represented in a double, but rather to just `int()` the big ones and do nothing to the little ones.  It's not clear that this distinction is useful.  Though, certainly the idea of differentiating large integers makes sense for JSON interfaces that do not support higher range integers.

altendky (approved):
Thanks.

PR #8324, opened 2021-09-03
add debug option to log all SQL commands

arvidn:

PR #11635, opened 2022-05-25
add debug option to log all SQL commands for wallet db

neurosis69:
Add debug option for wallet db as well, implementation like in PR #8324


chia/wallet/wallet_state_manager.py
```diff
@@ -139,6 +141,18 @@ async def create(
...
 
+        if self.config.get("log_sqlite_cmds", False):
+            sql_log_path = path_from_root(self.root_path, "log/wallet_sql.log")
+            self.log.info(f"logging SQL commands to {sql_log_path}")
+
+            def sql_trace_callback(req: str):
+                timestamp = datetime.now().strftime("%H:%M:%S.%f")
+                log = open(sql_log_path, "a")
+                log.write(timestamp + " " + req + "\n")
+                log.close()
```
arvidn, line 152: I recognize this is copy-pasted from what I did for the wallet. However, today I would have written this as:

```
with open(sql_lot_path, "a") as log:
   log.write(timestamp + " " + req + "\n")
```

arvidn (dismissed):
this is good. I think it would be a slight improvement to use a context manager for the file

chia/wallet/wallet_state_manager.py
neurosis69: Yes, copy-paste is my well known superpower ;)
I'll change my PR according to your suggestion.

neurosis69:
I am also having a hard time finding the correct log statement if using active trace over days.

Therefore I'd like to change the following:

`timestamp = datetime.now().strftime("%H:%M:%S.%f")`
to
`timestamp = datetime.now().isoformat(timespec='milliseconds')`

```python
>>> datetime.now().strftime("%H:%M:%S.%f")
'10:04:28.939514'
>>> datetime.now().isoformat(timespec='milliseconds')
'2022-05-27T10:04:35.041'
```

This would be the same format as used in debug.log.

arvidn (dismissed):

arvidn:
the reason I used this time format in the full node was to match `strace`

arvidn (approved):

wjblanke (approved):
aok

PR #11636, opened 2022-05-25
Request header blocks, and new rate limits

mariano54:
Builds on the changes in https://github.com/Chia-Network/chia-blockchain/pull/11155
Adds a few new mechanisms to allow changes to the network protocol, as well as the first actual changes to the protocol.
* Adds a capability `BLOCK_HEADERS` to signal that the node supports the new `request_block_headers` api.
* Adds a capability `RATE_LIMITS_V2` to signal that the node supports the new, more permissive rate limits. This allows wallets to make more requests per second, since the backend has been optimized (in the other PRs).
* The lowest compatible rate limit that both peers support is used
* The new api `request_block_headers` is much much faster than the old one. This enables wallets to sync up to 500 tps in trusted mode and 200 tps in untrusted, under optimal assumptions (close peer, fast CPU, etc).

- [x] pull out wallet optimizations
- [x] put in trepca's latest changes
- [x] address arvid's comments
- [x] Figure out how to increase rate limits
- [x] CI passing
- [x] Ensure everything is still fast after Ci pass
- [x] Ensure everything still works after pulling out wallet opts
- [x] self review
- [x] rate limits diffs


chia/wallet/util/wallet_sync_utils.py
```diff
@@ -327,13 +361,12 @@ async def fetch_header_blocks_in_range(
...
+                res_h_blocks: Optional[Union[RespondBlockHeaders, RespondHeaderBlocks]] = res_h_blocks_task.result()
             else:
                 res_h_blocks = await res_h_blocks_task
         else:
             log.debug(f"Fetching: {start}-{end}")
-            request_header_blocks = RequestHeaderBlocks(request_start, request_end)
-            res_h_blocks_task = asyncio.create_task(_fetch_header_blocks_inner(all_peers, request_header_blocks))
+            res_h_blocks_task = asyncio.create_task(_fetch_header_blocks_inner(all_peers, request_start, request_end))
             peer_request_cache.add_to_block_requests(request_start, request_end, res_h_blocks_task)
             res_h_blocks = await res_h_blocks_task
```
arvidn, line 371: it looks odd to `await` the request task inside the loop. Why do we create a task for each request unless we want to request all block ranges in parallel? (which I would expect us to want)

chia/wallet/wallet_node.py
```diff
@@ -672,7 +672,7 @@ async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: i
                     ]
                     if len(valid_states) > 0:
                         async with self.wallet_state_manager.db_wrapper.lock:
-                            self.log.info(
+                            self.log.warning(
```
arvidn, line 675: it doesn't look right to make this a warning. there's nothing wrong with this, right?

tests/core/server/test_rate_limits.py
```diff
@@ -1,28 +1,46 @@
...
 
 class TestRateLimits:
+    @pytest.mark.asyncio
+    async def test_get_rate_limits_to_use(self):
+        assert get_rate_limits_to_use(rl_v2, rl_v2) != get_rate_limits_to_use(rl_v2, rl_v1)
+        assert get_rate_limits_to_use(rl_v1, rl_v1) == get_rate_limits_to_use(rl_v2, rl_v1)
+        assert get_rate_limits_to_use(rl_v1, rl_v1) == get_rate_limits_to_use(rl_v1, rl_v2)
+
     @pytest.mark.asyncio
     async def test_too_many_messages(self):
```
arvidn, line 34: it might be appropriate to have a fixture for rate-limit-version, that's parameterized on 1 and 2. similar to `db_version`

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -58,6 +153,18 @@ async def test_basic_sync_wallet(self, bt, two_wallet_nodes, default_400_blocks,
...
+        [
+            dict(
+                disable_capabilities=[Capability.BLOCK_HEADERS],
+            ),
+            dict(
+                disable_capabilities=[Capability.BASE],
+            ),
+        ],
+        indirect=True,
+    )
```
arvidn, line 167: I would have expected this parameterization to be applied to the `wallet_node` or possibly `two_wallet_nodes` fixtures. That way all tests would be run against both sets of capabilities.

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -23,6 +27,97 @@ def wallet_height_at_least(wallet_node, h):
...
+            await full_node_api.full_node.respond_block(full_node_protocol.RespondBlock(block))
+
+        msg = await full_node_api.request_block_headers(
+            wallet_protocol.RequestBlockHeaders(uint32(80), uint32(99), False)
+        )
+        assert msg.type == ProtocolMessageTypes.respond_block_headers.value
+        msg = await full_node_api.request_block_headers(
+            wallet_protocol.RequestBlockHeaders(uint32(10), uint32(8), False)
+        )
+        assert msg is None
```
arvidn, line 99: is there a good reason for returning `None` instead of a rejection here?

arvidn (changes requested):

tests/wallet/sync/test_wallet_sync.py
trepca: was trying to make it backwards compatible, but rejection would make more sense in this case, not sure why other calls return None. Rejection seems to be reserved for missing hashes only.

tests/wallet/sync/test_wallet_sync.py
arvidn: whatever level those fixtures are constructed at. my point is that if those fixtures are parameterized, all tests using them will also be parameterized, which means we cover both new and old clients with the existing tests

arvidn:
The new message types added by this PR need to be included in the protocol message regression test. see `tests/util/test_network_protocol_files.py`.

chia/full_node/block_store.py
```diff
@@ -490,6 +531,25 @@ async def get_block_records_in_range(
 
         return ret
 
+    async def get_block_bytes_in_range(
```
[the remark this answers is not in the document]
mariano54: Yeah I was thinking the same thing. But there is another method which behaves in the same way. Maybe we should change both of them to throw an error, I'll try that

chia/full_node/full_node_api.py
```diff
@@ -1327,8 +1375,11 @@ async def request_header_blocks(self, request: wallet_protocol.RequestHeaderBloc
         blocks: List[FullBlock] = await self.full_node.block_store.get_blocks_by_hash(header_hashes)
         header_blocks = []
         for block in blocks:
-            added_coins_records = await self.full_node.coin_store.get_coins_added_at_height(block.height)
-            removed_coins_records = await self.full_node.coin_store.get_coins_removed_at_height(block.height)
+            added_coins_records_coroutine = self.full_node.coin_store.get_coins_added_at_height(block.height)
+            removed_coins_records_coroutine = self.full_node.coin_store.get_coins_removed_at_height(block.height)
+            added_coins_records, removed_coins_records = await asyncio.gather(
+                added_coins_records_coroutine, removed_coins_records_coroutine
+            )
```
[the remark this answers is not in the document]
mariano54: are you suggesting we pipeline all blocks (optimize further) or remove this optimization?

chia/protocols/shared_protocol.py
```diff
@@ -17,6 +17,13 @@
 # These are passed in as uint16 into the Handshake
 class Capability(IntEnum):
     BASE = 1  # Base capability just means it supports the chia protocol at mainnet
+    # introduces RequestBlockHeaders, which is a faster API for fetching header blocks
+    # !! the old API is *RequestHeaderBlock* !!
+    BLOCK_HEADERS = 2
+    # Specifies support for v1 and v2 versions of rate limits. Peers will ues the lowest shared capability:
+    # if peer A support v3 and peer B supports v2, they should send:
+    # (BASE, RATE_LIMITS_V2, RATE_LIMITS_V3), and (BASE, RATE_LIMITS_V2) respectively. They will use the V2 limits.
+    RATE_LIMITS_V2 = 3
```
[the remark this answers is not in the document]
mariano54: What is the benefit of splitting them, that the history is cleaner

chia/server/rate_limits.py
```diff
@@ -2,124 +2,17 @@
...
-
-# All non-transaction apis also have an aggregate limit
-NON_TX_FREQ = 1000
-NON_TX_MAX_TOTAL_SIZE = 100 * 1024 * 1024
-
-
-# The three values in the tuple correspond to the three limits above
-# The third is optional
-
-rate_limits_tx = {
```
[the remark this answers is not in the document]
mariano54: Yes

chia/wallet/util/wallet_sync_utils.py
mariano54: We create a task because we might have two async io threads that are requesting the same info at the same time, and we want to cache the task so they can both obtain the result when it arrives.

tests/core/server/test_rate_limits.py
mariano54: The rate limits are different though so many tests would fail and require more significant changes. We do have an integration test at the bottom of the file though.

chia/full_node/full_node_api.py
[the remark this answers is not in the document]
arvidn: I'm weakly suggesting removing this optimization. My concern is that (as far as I understand) an `async` function isn't *guaranteed* to return a future, which I believe is required for `gather()`. It's possible I've misunderstood something, but my impression was that you had to `create_task()` to ensure you can wait on it as a future.

So, my ignorance of asyncio makes me concerned about this change, but if this is obviously correct to someone who's more experienced, that's fine. Feel free to ignore my concern.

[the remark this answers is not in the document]
mariano54: asyncio.gather takes in awaitables, which are either tasks or coroutines. If it's a coroutine, it immediately schedules it as a task:

```
awaitable asyncio.gather(*aws, return_exceptions=False)

    Run [awaitable objects](https://docs.python.org/3/library/asyncio-task.html#asyncio-awaitables) in the aws sequence concurrently.

    If any awaitable in aws is a coroutine, it is automatically scheduled as a Task.
    ```

**Comment by 1-justinstarr** [comment]
Seems like it does not hurt to have in their and at least make sure tasks and out there waiting. I would only suggest adding some sort of timer to ensure the task does not sit out there too long by itself if ever. Also, it would need a back end counter that would count after it was pulled off a task. Assign new function requires finding and debugging any issue preventing progress. Thanks guys!

Justin Starr

**Comment by arvidn** [comment]
@mariano54 if you rebase this on top of `main` now, you'll have to add the new messages to `tests/util/build_network_protocol_files.py`, to the visitor function. Then you have to run that script.

That will update the test to include the new messages. make sure no other messages were changed.

File: chia/wallet/util/wallet_sync_utils.py [hunk 10]
```diff
@@ -33,19 +37,18 @@
 
 
 async def fetch_last_tx_from_peer(height: uint32, peer: WSChiaConnection) -> Optional[HeaderBlock]:
-    request_height: int = height
+    request_height: uint32 = height
```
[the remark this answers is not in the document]
mariano54: Fixed

chia/wallet/util/wallet_sync_utils.py
```diff
@@ -33,19 +37,18 @@
...
     while True:
         if request_height == -1:
```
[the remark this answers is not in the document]
mariano54: Yes this needs a test, i will revert for now. Tests can be added in future PRs, large parts of wallet are untested

tests/core/full_node/stores/test_block_store.py
```diff
@@ -264,3 +266,42 @@ def generator(i: int) -> SerializedProgram:
             assert await store.get_generator(blocks[4].header_hash) == new_blocks[4].transactions_generator
             assert await store.get_generator(blocks[6].header_hash) == new_blocks[6].transactions_generator
             assert await store.get_generator(blocks[7].header_hash) == new_blocks[7].transactions_generator
+
+
+@pytest.mark.asyncio
+async def test_get_blocks_by_hash(tmp_dir, bt, db_version):
```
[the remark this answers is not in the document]
mariano54: done

tests/wallet/sync/test_wallet_sync.py
mariano54: I think it's overkill to run all the tests again for this change. however I do see an argument  for doing it for the wallet sync test, i'll do that

tests/core/server/test_rate_limits.py
```diff
@@ -332,3 +332,26 @@ async def test_different_versions(self, node_with_params, node_with_params_b, se
         test_different_versions_results.append(total_tx_msg_count)
         if len(test_different_versions_results) >= 4:
             assert len(set(test_different_versions_results)) >= 2
+
+    @pytest.mark.asyncio
```
arvidn, line 336: it doesn't look like this needs to be marked `asyncio`, nor be a method. but I suppose that's a nit-pick

chia/server/rate_limit_numbers.py
```diff
@@ -0,0 +1,178 @@
...
+            ProtocolMessageTypes.reject_header_request: RLSettings(500, 100),
+            ProtocolMessageTypes.request_removals: RLSettings(5000, 50 * 1024, 10 * 1024 * 1024),
+            ProtocolMessageTypes.respond_removals: RLSettings(5000, 1024 * 1024, 10 * 1024 * 1024),
+            ProtocolMessageTypes.reject_removals_request: RLSettings(500, 100),
+            ProtocolMessageTypes.request_additions: RLSettings(50000, 100 * 1024 * 1024),
+            ProtocolMessageTypes.respond_additions: RLSettings(50000, 100 * 1024 * 1024),
+            ProtocolMessageTypes.reject_additions_request: RLSettings(500, 100),
+            ProtocolMessageTypes.reject_header_blocks: RLSettings(1000, 100),
+            ProtocolMessageTypes.respond_header_blocks: RLSettings(5000, 2 * 1024 * 1024),
+            ProtocolMessageTypes.request_block_headers: RLSettings(5000, 100),
```
arvidn, line 163: if someone claims to support this (or the other two) new message, but without claiming support for rate limits v2, I suppose there would be a problem, right? (since the rate limit for this message is only specified in v2)

arvidn (dismissed):

chia/server/rate_limit_numbers.py
mariano54: yeah.. i guess that would be an incompatible combination. Rate limits V2 is a prerequisite for block headers

arvidn (dismissed):

chia/server/rate_limit_numbers.py
```diff
@@ -0,0 +1,184 @@
...
+    max_total_size: Optional[int] = None  # Max cumulative size of all requests in that period
+
+
+def get_rate_limits_to_use(our_capabilities: List[Capability], peer_capabilities: List[Capability]) -> Dict[str, Any]:
+    # This will use the newest possible rate limits that both peers support. At this time there are only two
+    # options, v1 and v2.
+
+    if Capability.RATE_LIMITS_V2 in our_capabilities and Capability.RATE_LIMITS_V2 in peer_capabilities:
+        # Use V2 rate limits
+        if (1, 2) in compose_rate_limits_cache:
```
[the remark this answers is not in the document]
mariano54: yeah that's better!

arvidn (dismissed):
is `rate_limits` immutable?

mariano54:
> is `rate_limits` immutable?

It is not immutable.

arvidn (approved):

trepca (approved):

PR #11647, opened 2022-05-26
extend benchmark to make the db-sync mode and node-profiler configurable

arvidn:
example output from the node profiler:

image

### DIFF
--- a/tools/run_benchmark.sh
+++ b/tools/run_benchmark.sh
@@ -9,6 +9,8 @@ run_benchmark() {
    mv test-full-sync.log cpu.png cpu-usage.log plot-cpu.gnuplot "$2"
    python ./tools/test_full_sync.py analyze
    mv slow-batch-*.profile slow-batch-*.png "$2"
+   python ./chia/util/profiler.py profile-node >"$2/node-profile.txt"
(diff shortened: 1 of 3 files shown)


1-justinstarr (commented):
running

tools/run_benchmark.sh
```diff
@@ -2,13 +2,15 @@
 
 run_benchmark() {
    # shellcheck disable=SC2086
-   python ./tools/test_full_sync.py run $3 --profile --test-constants "$1" &
+   python ./tools/test_full_sync.py run $3 --profile --node-profiler --test-constants "$1" &
```
arvidn, line 5: it just occurred to me that combining `--profile` and `--node-profiler` here will cause two `cProfile` instances to be active at the same time. Is that supported by python? I don't recall anything stating it isn't..

mariano54, line 5: Is there any reduced performance from this?

mariano54 (dismissed):

trepca (dismissed):

trepca (approved):
lgtm

wjblanke (approved):
aok

PR #11654, opened 2022-05-26
set overflow to the last SES and not add them all

emlowe:
Set the overflow to the last SES overflow instead of adding all the overflow together
Simplify the loop slightly by not using enumerate since it wasn't being used anyway

Should help with syncing by rolling back to a more recent block rather than one that is considerable older


wjblanke (dismissed):
aok

emlowe:
Closing PR as was merged into 1.4 and will make it into main through catch-up merges

PR #11609, opened 2022-05-23
Consolidate test_did_rpc.py into test_wallet_rpc.py, make the tests use the wallet RPC client and fix name annotation in create_new_did_wallet()

AmineKhaldi:

PR #11657, opened 2022-05-26
Fix for merging NFT0 to main

ytx1991:
WIP


AmineKhaldi:
Please note that PR #11609 already exists for some of what you're doing here. I'd suggest reviewing and landing that instead, to avoid conflicting it as well as its main version.

AmineKhaldi (changes requested):
Please note that this PR is meant to address nft0 feedback. As a result, we would like to see isolated, well contained commits that we can bring back to nft0 and elsewhere. This PR contains other people's PRs, PRs that already landed into nft1_soon_tm and changes that are completely unrelated to addressing nft0 reviews and therefore is not usable in its current state.
Thank you for your understanding.

trepca:
@ytx1991 what's the status here? are you planning to fix the diff?

PR #11663, opened 2022-05-27
Catchup/long lived atari from release 1.4.0 46ebed37d8a8cc86ba8c6db511fd936fd53f700d fix merge history

altendky:
This is a no-op 'ours' merge to add back the parent and history tracking lost when https://github.com/Chia-Network/chia-blockchain/pull/11648 was squash rebased instead of merged into `long_lived/atari`.

```
$ git merge -X ours 46ebed37d8a8cc86ba8c6db511fd936fd53f700d
```


arvidn (approved):

arvidn:
this is correct because the resulting files are identical to what we already have in `long_lived/atari`.

PR #11669, opened 2022-05-27
Ms.optimize request additions

mariano54:
These changes have a massive effect on wallet syncing performance.
1. Only return the relevant coins in `respond_additions`. This was a bug, we were sending back every single coin instead of the ones the client asked for. (EDIT: this was found to actually be working properly before)
2. Maintain a least recently used (LRU) cache for addition coins at each height. This is particularly useful when a client has many coins at the same height, and has a negligible impact on memory.

### DIFF
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -1149,6 +1149,7 @@ class FullNodeAPI:
                 coins_map.append((puzzle_hash, coins))
             response = wallet_protocol.RespondAdditions(block.height, block.header_hash, coins_map, None)
         else:
+            coins_map_filtered = [(ph, c) for (ph, c) in coins_map if ph in request.puzzle_hashes]
             # Create addition Merkle set
             addition_merkle_set = MerkleSet()
             # Addition Merkle set contains puzzlehash and hash of all coins with that puzzlehash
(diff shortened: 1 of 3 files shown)


chia/full_node/full_node_api.py
```diff
@@ -1147,16 +1132,16 @@ async def request_additions(self, request: wallet_protocol.RequestAdditions) ->
         if request.puzzle_hashes is None:
             for puzzle_hash, coins in puzzlehash_coins_map.items():
                 coins_map.append((puzzle_hash, coins))
-            response = wallet_protocol.RespondAdditions(block.height, block.header_hash, coins_map, None)
+            response = wallet_protocol.RespondAdditions(request.height, header_hash, coins_map, None)
         else:
+            coins_map_filtered = [(ph, c) for (ph, c) in coins_map if ph in request.puzzle_hashes]
```
trepca, line 1137: isn't `coins_map` always empty here?

trepca: this should probably be called before `response` is generated?

mariano54: uhh yeah you're right... good catch. Should have added a test here.

trepca (approved):
lgtm

wjblanke (approved):
aok, note that coin issue is not in this pr

PR #11670, opened 2022-05-27
Don't re-hash the same objects

mariano54:


chia/wallet/puzzles/cat_loader.py
```diff
@@ -2,3 +2,5 @@
 
 CAT_MOD = load_clvm("cat.clvm", package_or_requirement=__name__)
 LOCK_INNER_PUZZLE = load_clvm("lock.inner.puzzle.clvm", package_or_requirement=__name__)
+
+CAT_MOD_HASH = CAT_MOD.get_tree_hash()
```
arvidn, line 6: you don't seem to be using this anywhere. is that deliberate?

chia/wallet/cat_wallet/cat_utils.py
```diff
@@ -41,11 +41,15 @@ def match_cat_puzzle(puzzle: Program) -> Tuple[bool, Iterator[Program]]:
         return False, iter(())
 
 
-def construct_cat_puzzle(mod_code: Program, limitations_program_hash: bytes32, inner_puzzle: Program) -> Program:
+def construct_cat_puzzle(
```
arvidn, line 44: aren't there any call sites to this function that needs updating too?

mariano54: yes but they are not so performance critical so i chose to make the patch smallerinstead

chia/wallet/puzzles/cat_loader.py
mariano54: oh that was a bad cherry pick

mariano54: Please re-review

arvidn (dismissed):

wjblanke (dismissed):
aok

arvidn (approved):

wjblanke (approved):
aok

PR #11671, opened 2022-05-27
Ms.optimize wallet sync

mariano54:
Various additions to the cache, and moving around some pieces of validation to reduce CPU usage significantly

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -51,7 +51,6 @@ from chia.util.config import WALLET_PEERS_PATH_KEY_DEPRECATED
 from chia.util.default_root import STANDALONE_ROOT_PATH
 from chia.util.ints import uint32, uint64
 from chia.util.keychain import Keychain, KeyringIsLocked
-from chia.util.network import is_localhost
 from chia.util.path import mkdir, path_from_root
 from chia.util.profiler import profile_task
 from chia.wallet.transaction_record import TransactionRecord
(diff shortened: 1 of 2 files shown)


chia/wallet/wallet_node.py
```diff
@@ -1377,52 +1379,77 @@ async def validate_block_inclusion(
...
-                if len(en_block.finished_sub_slots) > 0:
-                    #  What to do here
-                    reversed_slots = en_block.finished_sub_slots.copy()
-                    reversed_slots.reverse()
-                    for slot_idx, slot in enumerate(reversed_slots[:-1]):
-                        hash_val = reversed_slots[slot_idx + 1].reward_chain.get_hash()
-                        if not hash_val == slot.reward_chain.end_of_slot_vdf.challenge:
-                            self.log.error("Failed validation 6")
+                    if len(en_block.finished_sub_slots) > 0:
+                        #  What to do here
```
trepca, line 1426: confusing comment

chia/wallet/wallet_node.py
```diff
@@ -1377,52 +1379,77 @@ async def validate_block_inclusion(
...
+                        reversed_slots = en_block.finished_sub_slots.copy()
+                        reversed_slots.reverse()
+                        for slot_idx, slot in enumerate(reversed_slots[:-1]):
```
trepca, line 1429: FYI, not sure how big this list can be, but to avoid reversing the whole list, you could use a range 
```python
l = ["a", "b", "c", "d", "e"]
for idx in range(len(l)-2, -1, -1):
    slot = l[idx]
    prev_slot = l[idx+1]
    print(f"comparing {slot} with {prev_slot}")
```

trepca (dismissed):

chia/wallet/wallet_node.py
mariano54: 99.9% + of the time it will just be length 1.

wjblanke (dismissed):
aok

wjblanke (approved):
aok

PR #11673, opened 2022-05-27
Fix nft_get_info

ytx1991:
Fix the bug while calling nft_get_info

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1439,10 +1439,13 @@ class WalletRpcApi:
             inner_solution: Program = Program.from_bytes(bytes(coin_spend.solution)).rest().rest().first().first()
             full_puzzle: Program = Program.from_bytes(bytes(coin_spend.puzzle_reveal))
             update_condition = None
-            for condition in inner_solution.rest().first().rest().as_iter():
-                if condition.first().as_int() == -24:
-                    update_condition = condition
-                    break
+            try:
+                for condition in inner_solution.rest().first().rest().as_iter():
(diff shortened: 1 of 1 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1439,10 +1439,13 @@ async def nft_get_info(self, request: Dict) -> Optional[Dict]:
...
-                if condition.first().as_int() == -24:
-                    update_condition = condition
-                    break
+            try:
+                for condition in inner_solution.rest().first().rest().as_iter():
+                    if condition.first().as_int() == -24:
+                        update_condition = condition
+                        break
+            except Exception:
+                log.info(f"Inner solution is not a metadata updater solution: {inner_solution}")
```
paninaro, line 1448: So when an NFT is transferred via an offer, the inner solution is different, that's the source of the trouble? Is there another way to detect this rather than catching and ignoring the exception? @Quexington and @geoffwalmsley may have some ideas here maybe?

Quexington: I think this is the correct fix.  Maybe specifically handling a lack of that condition would be better than just catching blanket exceptions, but either way gets the job done.  This is just looking for IF the metadata changed, and if it didn't, move along, which is the correct set of actions during an offer.

Quexington: However, looking for conditions in the solution is always wrong and only works right now because we only have one type of custody puzzle.  You should instead get the conditions via `inner_puzzle.run(inner_solution)`.  This is fine for now, but if we ever start supporting multisig as a valid inner puzzle, for example, all of this breaks.

PR #11714, opened 2022-05-31
Set the overflow to the last SES overflow

emlowe:
Set the overflow to the last SES overflow instead of adding all the overflow together
Simplify the loop slightly by not using enumerate since it wasn't being used anyway

Should help with syncing by rolling back to a more recent block rather than one that is considerable older


wjblanke (approved):
aok

PR #11723, opened 2022-06-01
add checkpoint feature to test_full_sync

arvidn:
This patch extends the `test_full_sync.py` test/benchmark with a new feature allowing a benchmark run to start from a height > 0. This makes it more practical to benchmark with the actual mainnet database.

first, create a checkpoint:

```
python ./tools/test_full_sync.py create-checkpoint --height 1500000 ~/.chia/mainnet/db/blockchain_v2_mainnet.sqlite my-checkppoint
```

Then run tests:

```
python ./tools/test_full_sync.py run --start-at-checkpoint my-checkpoint ~/.chia/mainnet/db/blockchain_v2_mainnet.sqlite
```

The test run will start at height 1500000.

Creating the checkpoint could be a lot faster by simply pruning the input database, but the code to do so would be more complicated and error prone. The current approach is slow but simple, with the idea that you'll only create the checkpoint once or twice anyway.

### DIFF
--- a/tests/tools/test_full_sync.py
+++ b/tests/tools/test_full_sync.py
@@ -23,5 +23,6 @@ def test_full_sync_test(keep_up: bool):
             keep_up=keep_up,
             db_sync="off",
             node_profiler=False,
+            start_at_checkpoint=None,
         )
     )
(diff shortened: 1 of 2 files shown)


mariano54 (approved):

wjblanke (approved):
aok

PR #11731, opened 2022-06-01
Fee estimation interface

aqk:
The purpose of this PR is to provide a versatile interface for fee estimation for the following uses:
* Chia UI users
* 3rd party wallets
* Large farmers
* command line use

Design considerations

Note the ability to return multiple estimates in one request.
Fee estimates are denominated in seconds (instead of blocks)
Note that we do not store any state information, so estimators must re-run their inputs when the client starts.

The API is different from the Bitcoin RPC in these major respects:
* Transaction cost is a required input (explicit or implicit)
* The fee for a transaction of a certain cost is returned, not a fee rate
* There are no [hard-coded settings](https://github.com/bitcoin/bitcoin/blob/master/src/rpc/fees.cpp)

We also intend to return max fee, min fee, and fee rate with each call.
Jeff had the idea to use a state update to notify the UI when the fee estimates change. The UI can then store a historical
Jeff also had the idea of displaying the current fee rate in the UI.
I like this idea, because it can help the user form an intuition about the rates passively.

One thing I would like input on:

What is the default maximum time window we should attempt to return an estimate for? 24 hours?

While researching blockchain fee estimation, I found out that big users like Pool Operators and exchanges do not use the bitcoin fee estimator. It does not take into consideration daily and weekly cycles that might allow those users to process many transactions at a time when fees are at their daily or weekly minimum.

I think that particular issue is out of scope for this estimator, but the current design makes it MUCH easier for those users to plug in custom code by implementing the `FeeEstimatorInterface`. On a related note, for ease of future expansion, I am mildly advocating for an RPC parameter `estimation_method` that could allow us to select other algorithms.

Note that we require a cost or a SpendBundle to calculate our fee estimates, instead of passing. This decision is a little more involved than it looks. Yes, it makes it easier for the caller. But the more important reason is that there can be some subtlety on the edge cases when the mempool is ALMOST full.

Further, the Fee Rate can be retrieved by passing a cost of 1.

I like the idea of introducing a new type to represent FeeRate, as bitcoin does [here](https://github.com/bitcoin/bitcoin/blob/master/src/policy/feerate.cpp). Let me know what you think.

### DIFF
--- a/chia/policy/fee_estimator.py
+++ b/chia/policy/fee_estimator.py
@@ -1,6 +1,7 @@
-import abc
+from typing_extensions import Protocol
 
 from chia.policy.fee_estimation import FeeBlockInfo, FeeMempoolInfo
+from chia.util.ints import uint64
 
 
 class FeeEstimatorConfig:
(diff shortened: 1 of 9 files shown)


chia/policy/fee_estimator.py
```diff
@@ -0,0 +1,41 @@
+import abc
```
aqk, line 1: Hey @altendky is this the best way to enforce an interface in Pyton?

altendky: I like `Protocol`.  https://docs.python.org/3.10/library/typing.html#typing.Protocol  Though we get it via `from typing_extensions import Protocol` since it was introduced in Python 3.8.

PR #11730, opened 2022-06-01
streamable: Introduce `Streamable.__post_init__` processing cache

xdustinface:
Speeds up normal constructor calls:

```
compare: benchmark, old: aa95b1e39, new: 23cae4bcc
mode         | µs/iteration old | µs/iteration new | diff %      
creation     | 39.76            | 9.97             | -74.92      

compare: full_block, old: aa95b1e39, new: 23cae4bcc
mode         | µs/iteration old | µs/iteration new | diff %      
creation     | 53.07            | 10.26            | -80.66     
```

Based on: 
- #10652

PR #11732, opened 2022-06-01
streamable: Fix default value assignments for `dataclass_from_dict`

### LINKED ISSUE #11709: [Bug]  Chia Farmer not showing plots
### What happened?

Good morning, I installed beta versions 1.3.6174.0, 1.3.6175.0, and 1.3.6179.0 on a test node. None show when selecting the PLOT tab and then the up caret, the plots

### Version

1.3.6179.0

Windows

GUI

### Relevant log output

_No response_

xdustinface:
This fixes default value assignments after https://github.com/Chia-Network/chia-blockchain/pull/10561. See the new test in c539809db49e1d0d781b3abda0fc70ede50ca2ef failing without a85be6d72f9b6e9ccb9d462b1bb2c7a590dc6418. But this also leads to worse performance due to `__post_init__` being called which at least gets mitigated by https://github.com/Chia-Network/chia-blockchain/pull/11730.

### DIFF
--- a/tests/core/util/test_streamable.py
+++ b/tests/core/util/test_streamable.py
@@ -128,15 +128,15 @@ def test_pure_dataclasses_in_dataclass_from_dict() -> None:
     "test_class, input_dict, error",
     [
         [TestDataclassFromDict1, {"a": "asdf", "b": "2", "c": G1Element()}, ValueError],
-        [TestDataclassFromDict1, {"a": 1, "b": "2"}, KeyError],
+        [TestDataclassFromDict1, {"a": 1, "b": "2"}, TypeError],
         [TestDataclassFromDict1, {"a": 1, "b": "2", "c": "asd"}, ValueError],
         [TestDataclassFromDict1, {"a": 1, "b": "2", "c": "00" * G1Element.SIZE}, ValueError],
         [TestDataclassFromDict1, {"a": [], "b": "2", "c": G1Element()}, TypeError],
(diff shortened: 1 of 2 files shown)


tests/core/util/test_streamable.py
```diff
@@ -145,6 +145,28 @@ def test_dataclass_from_dict_failures(test_class: Type[Any], input_dict: Dict[st
...
+    [
+        [{}, {"a": 1, "b": "default", "c": []}],
+        [{"a": 2}, {"a": 2, "b": "default", "c": []}],
+        [{"b": "not_default"}, {"a": 1, "b": "not_default", "c": []}],
+        [{"c": [1, 2]}, {"a": 1, "b": "default", "c": [1, 2]}],
+        [{"a": 2, "b": "not_default", "c": [1, 2]}, {"a": 2, "b": "not_default", "c": [1, 2]}],
+    ],
+)
+def test_from_json_dict_default_values(input_dict: Dict[str, object], output_dict: Dict[str, object]) -> None:
+    assert TestFromJsonDictDefaultValues.from_json_dict(input_dict).to_json_dict() == output_dict
```
arvidn, line 167: as kyle pointed out to me earlier, it's risky to compare dictionaries directly like this. The problematic case then was that `int` can be compared to `bool`. I would think it would be safer to convert both dictionaries to strings  and compare those.

arvidn (dismissed):

arvidn (approved):

wjblanke (approved):
aok

PR #11733, opened 2022-06-01
dont request ses

almogdepaz:
during wallet sync when validating block inclusion  we currently request the needed sub epoch data from the peer
this happens after we already downloaded and validated a Weight Proof containing a list of SubEpochData from genesis

this change uses the SubEpochData list inside the already validated WP and removes the redundant request


mariano54 (approved):

emlowe (approved):
Can we add a little more description to this PR - mostly so we can generate release notes as the time comes.

wallentx:
Verified that post-merge PR title changes do reflect on the generated changelog. Merging.

PR #11735, opened 2022-06-01
introduce DBWrapper2.write_transaction() and .read_transaction()

altendky:
Draft for:
- [ ] Revisiting
- [ ] Unbreaking CI


chia/full_node/full_node.py
```diff
@@ -2243,7 +2243,7 @@ async def _replace_proof(
                 new_block = dataclasses.replace(block, challenge_chain_ip_proof=vdf_proof)
         if new_block is None:
             return False
-        async with self.db_wrapper.write_db():
+        async with self.db_wrapper.write_transaction():
```
altendky, line 2246: This one seems unnecessary.  If it is needed, shouldn't it be handled down inside `.replace_proof()` instead?

arvidn: I agree. this looks unnecessary

arvidn (dismissed):

altendky:
https://github.com/Chia-Network/chia-blockchain/pull/13468

PR #11745, opened 2022-06-02
streamable: Introduce `streamable.Field`

xdustinface:
Intention is mostly preparation for a follow up where i want to check the default values of the fields.

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -48,19 +48,23 @@ StreamFunctionType = Callable[[object, BinaryIO], None]
 ConvertFunctionType = Callable[[object], object]
 
 
+@dataclasses.dataclass(frozen=True)
+class Field:
+    name: str
+    type: Type[object]
+
+
 # Caches to store the fields and (de)serialization methods for all available streamable classes.
-FIELDS_FOR_STREAMABLE_CLASS: Dict[Type[object], Tuple[dataclasses.Field[object], ...]] = {}
+FIELDS_FOR_STREAMABLE_CLASS: Dict[Type[object], Tuple[Field, ...]] = {}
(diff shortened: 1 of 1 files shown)


chia/util/streamable.py
```diff
@@ -49,12 +49,21 @@ class DefinitionError(StreamableError):
...
 STREAM_FUNCTIONS_FOR_STREAMABLE_CLASS: Dict[Type[object], List[StreamFunctionType]] = {}
 PARSE_FUNCTIONS_FOR_STREAMABLE_CLASS: Dict[Type[object], List[ParseFunctionType]] = {}
 CONVERT_FUNCTIONS_FOR_STREAMABLE_CLASS: Dict[Type[object], List[ConvertFunctionType]] = {}
 
 
+def create_fields_cache(cls: Type[object]) -> Tuple[dataclasses.Field[object], ...]:
+    hints = get_type_hints(cls)
+    fields = dataclasses.fields(cls)
+    for field in fields:
+        field.type = hints.get(field.name, None)
```
altendky, line 62: I hesitate at least a bit to mutate the actual `Field` object.  It may be 'public', but that doesn't mean it makes sense to write to it.  I should have thought about this as soon as I saw you were hinting holding a `Field` object, but I somewhat intentionally did not do that when I did the fix for 3.10.  I think there's some use to having our own structure so that we isolate the interactions with the dataclass data to the analysis phase here.

I think this ends up not making a huge difference to this PR since any class we would write ourselves would presumably have `.type` and `.name` as well.

xdustinface: Makes sense, see d567c446c303df63e2805eedb745919030facacd.

altendky (approved):

wjblanke (approved):
aok

PR #11746, opened 2022-06-02
update unknown condition test

arvidn:
We recently landed https://github.com/Chia-Network/chia_rs/pull/4, which made condition opcode 1 defined as always being true.
Since this condition code is about to become well defined, use some other opcode to test unknown conditions.

### DIFF
--- a/tests/core/full_node/test_mempool.py
+++ b/tests/core/full_node/test_mempool.py
@@ -2102,7 +2102,7 @@ class TestGeneratorConditions:
         ],
     )
     def test_unknown_condition(self, mempool: bool, height: uint32):
-        for c in ['(2 100 "foo" "bar")', "(100)", "(3 1) (2 2) (3 3)", '("foobar")']:
+        for c in ['(2 100 "foo" "bar")', "(100)", "(4 1) (2 2) (3 3)", '("foobar")']:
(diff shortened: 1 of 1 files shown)


tests/core/full_node/test_mempool.py
```diff
@@ -2102,7 +2102,7 @@ def test_create_coin_with_hint(self, softfork_height):
         ],
     )
     def test_unknown_condition(self, mempool: bool, height: uint32):
-        for c in ['(1 100 "foo" "bar")', "(100)", "(1 1) (2 2) (3 3)", '("foobar")']:
+        for c in ['(2 100 "foo" "bar")', "(100)", "(3 1) (2 2) (3 3)", '("foobar")']:
```
richardkiss, line 2105: You might change the second `1` to a `4` since you already have a `3`. But it almost certainly doesn't matter.

richardkiss, line 2105:
```suggestion
        for c in ['(2 100 "foo" "bar")', "(100)", "(3 1) (2 2) (3 3)", '("foobar")']:
        for c in ['(2 100 "foo" "bar")', "(100)", "(4 1) (2 2) (3 3)", '("foobar")']:
```

richardkiss (approved):

wjblanke (approved):
aok

PR #11749, opened 2022-06-02
make the protocol regression test also ensure all messages are covered

arvidn:
Whenever we add a new message type, we also need to add that type to the protocol regression test. that test ensures the encoding of message types remain compatible over time.


mariano54 (approved):

fchirica (approved):

wjblanke (approved):
aok

PR #11751, opened 2022-06-02
generalize script to generate protocol message test data

arvidn:
This change generalizes `build_network_protocol_files.py` to enable using it to build more tests. Specifically, I'm interested in adding a regression test for the JSON encoding and decoding  of message types. This patch lays the foundation for that.

The updated script produces an identical test file for the existing protocol message test.


fchirica (approved):

wjblanke (approved):
aok

PR #11758, opened 2022-06-02
consensus: Make `ConsensusConstants` streamable + use `byte` conversions

xdustinface:
This is preparation to restrict `dataclass_from_dict` to streamable classes only. 6f374fb2828f410abd8357ed49166455d47bed69 is not absolutely required for that but since its possible after 176482917b3c6b9823ded303aa62d6c2698d9c6e i think it makes sense to be consistent with the other things passed around and the naming of `vars_to_bytes` and `bytes_to_vars`.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -8,7 +8,7 @@ from concurrent.futures.process import ProcessPoolExecutor
 from enum import Enum
 from multiprocessing.context import BaseContext
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import Dict, List, Optional, Set, Tuple
 
 from chia.consensus.block_body_validation import validate_block_body
 from chia.consensus.block_header_validation import validate_unfinished_header_block
@@ -48,7 +48,6 @@ from chia.util.generator_tools import get_block_header, tx_removals_and_addition
 from chia.util.inline_executor import InlineExecutor
 from chia.util.ints import uint16, uint32, uint64, uint128
 from chia.util.setproctitle import getproctitle, setproctitle
-from chia.util.streamable import recurse_jsonify
 
 log = logging.getLogger(__name__)
 
@@ -78,7 +77,7 @@ class StateChangeSummary:
 
 class Blockchain(BlockchainInterface):
     constants: ConsensusConstants
-    constants_json: Dict[str, Any]
+    constants_bytes: bytes
 
     # peak of the blockchain
     _peak_height: Optional[uint32]
@@ -144,7 +143,7 @@ class Blockchain(BlockchainInterface):
         self.constants = consensus_constants
         self.coin_store = coin_store
         self.block_store = block_store
-        self.constants_json = recurse_jsonify(self.constants)
+        self.constants_bytes = bytes(self.constants)
         self._shut_down = False
         await self._load_chain_from_store(blockchain_dir)
         self._seen_compact_proofs = set()
@@ -608,7 +607,7 @@ class Blockchain(BlockchainInterface):
     ) -> List[PreValidationResult]:
         return await pre_validate_blocks_multiprocessing(
             self.constants,
-            self.constants_json,
+            self.constants_bytes,
             self,
             blocks,
             self.pool,
@@ -624,7 +623,7 @@ class Blockchain(BlockchainInterface):
         task = asyncio.get_running_loop().run_in_executor(
             self.pool,
             _run_generator,
-            self.constants_json,
+            self.constants_bytes,
             unfinished_block,
             bytes(generator),
             height,
--- a/chia/consensus/constants.py
+++ b/chia/consensus/constants.py
@@ -4,13 +4,15 @@ from typing import Any
 
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.byte_types import hexstr_to_bytes
-from chia.util.ints import uint8, uint32, uint64, uint128
+from chia.util.ints import uint8, uint16, uint32, uint64, uint128
+from chia.util.streamable import Streamable, streamable
 
 log = logging.getLogger(__name__)
 
 
+@streamable
 @dataclasses.dataclass(frozen=True)
-class ConsensusConstants:
+class ConsensusConstants(Streamable):
     SLOT_BLOCKS_TARGET: uint32  # How many blocks to target per sub-slot
     MIN_BLOCKS_PER_CHALLENGE_BLOCK: uint8  # How many blocks must be created per slot (to make challenge sb)
     # Max number of blocks that can be infused into a sub-slot.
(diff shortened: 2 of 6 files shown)


AmineKhaldi (dismissed):

chia/consensus/blockchain.py
```diff
@@ -144,7 +143,7 @@ async def create(
         self.constants = consensus_constants
         self.coin_store = coin_store
         self.block_store = block_store
-        self.constants_json = recurse_jsonify(self.constants)
+        self.constants_bytes = bytes(self.constants)
```
xdustinface, line 146: @arvidn Answering your comment https://github.com/Chia-Network/chia-blockchain/pull/11763#discussion_r888873940 here since its related to this PR:

Retaining the extra field here avoids consensus params conversions for each call into multiprocessing.

chia/consensus/constants.py
```diff
@@ -26,37 +28,37 @@ class ConsensusConstants:
...
-    SIGNIFICANT_BITS: int  # The number of bits to look at in difficulty and min iters. The rest are zeroed
-    DISCRIMINANT_SIZE_BITS: int  # Max is 1024 (based on ClassGroupElement int size)
-    NUMBER_ZERO_BITS_PLOT_FILTER: int  # H(plot id + challenge hash + signage point) must start with these many zeroes
-    MIN_PLOT_SIZE: int
-    MAX_PLOT_SIZE: int
-    SUB_SLOT_TIME_TARGET: int  # The target number of seconds per sub-slot
-    NUM_SP_INTERVALS_EXTRA: int  # The difference between signage point and infusion point (plus required_iters)
-    MAX_FUTURE_TIME: int  # The next block can have a timestamp of at most these many seconds more
-    NUMBER_OF_TIMESTAMPS: int  # Than the average of the last NUMBER_OF_TIMESTAMPS blocks
+    SIGNIFICANT_BITS: uint8  # The number of bits to look at in difficulty and min iters. The rest are zeroed
```
xdustinface, line 31: @arvidn Answering your comment https://github.com/Chia-Network/chia-blockchain/pull/11763#discussion_r888880357 here since its related to this PR:

`int` is not streamable no. Whats the risk? If you have concerns about the size of one let me know but i think they are all sufficiently scaled.

arvidn: the risk is that `int` has very different behavior compared to our fixed size integer types. for example `bytes(int)` vs. `bytes(uint64)` or `int.to_bytes()` vs. `uint64.to_bytes()`.

I would worry about some existing use of these constants silently changing behavior. It might require an audit of the places where we use them.

arvidn:
I agree with moving away from passing the constants in JSON form, that seems unnecessarily verbose. However, it seems plain pickling would be appropriate here too.

In fact, I would expect the end-game of this would be to set up the constants as a global variable when initializing each process in the process pool. Since they don't ever change, we don't really need to pass them along for every job we submit to the pool.

chia/consensus/blockchain.py
xdustinface: Conversion goes on here https://github.com/Chia-Network/chia-blockchain/pull/11758#issuecomment-1145913196

chia/consensus/constants.py
mariano54: It's a good point, perhaps there is a place where -1 is used somewhere, but I don't know of any from the top of my head. These are all positive constants.  Also, there are so many places where these are used, auditing would be very difficult.

chia/consensus/constants.py
```diff
@@ -26,37 +28,37 @@ class ConsensusConstants:
...
     GENESIS_PRE_FARM_POOL_PUZZLE_HASH: bytes32  # The block at height must pay out to this pool puzzle hash
     GENESIS_PRE_FARM_FARMER_PUZZLE_HASH: bytes32  # The block at height must pay out to this farmer puzzle hash
-    MAX_VDF_WITNESS_SIZE: int  # The maximum number of classgroup elements within an n-wesolowski proof
+    MAX_VDF_WITNESS_SIZE: uint16  # The maximum number of classgroup elements within an n-wesolowski proof
     # Size of mempool = 10x the size of block
-    MEMPOOL_BLOCK_BUFFER: int
+    MEMPOOL_BLOCK_BUFFER: uint32
     # Max coin amount uint(1 << 64). This allows coin amounts to fit in 64 bits. This is around 18M chia.
-    MAX_COIN_AMOUNT: int
+    MAX_COIN_AMOUNT: uint64
```
mariano54, line 51: Maybe this should be uin128 to allow for higher coin amounts for custom chains.

chia/consensus/constants.py
```diff
@@ -26,37 +28,37 @@ class ConsensusConstants:
...
-    MAX_FUTURE_TIME: int  # The next block can have a timestamp of at most these many seconds more
-    NUMBER_OF_TIMESTAMPS: int  # Than the average of the last NUMBER_OF_TIMESTAMPS blocks
+    SIGNIFICANT_BITS: uint8  # The number of bits to look at in difficulty and min iters. The rest are zeroed
+    DISCRIMINANT_SIZE_BITS: uint16  # Max is 1024 (based on ClassGroupElement int size)
+    NUMBER_ZERO_BITS_PLOT_FILTER: uint16  # H(plot id + challenge hash + signage point) must start with these many zeroes # noqa: E501
+    MIN_PLOT_SIZE: uint16
+    MAX_PLOT_SIZE: uint16
+    SUB_SLOT_TIME_TARGET: uint32  # The target number of seconds per sub-slot
+    NUM_SP_INTERVALS_EXTRA: uint16  # The difference between signage point and infusion point (plus required_iters)
+    MAX_FUTURE_TIME: uint32  # The next block can have a timestamp of at most these many seconds more
```
mariano54, line 38: timestamps are uint64 in chia

mariano54 (changes requested):
I don't think it's a performance issue to use JSON since it's so small. However it is probably better to do it this way,  since we use streamable in other places as well (or pickling). So I agree with this PR.

chia/consensus/constants.py
xdustinface: Hm i guess you mean forks? They should just adjust it if required, right? We assigned `uint64` here already. See https://github.com/Chia-Network/chia-blockchain/blob/00c80ae6078f2c97765fe52b2a79aef5225d76d9/chia/consensus/default_constants.py#L44

chia/consensus/constants.py
xdustinface:
```
The next block can have a timestamp of at most these many seconds more
```

This is not a timestamp, its just an offset.

xdustinface:
Okay i reverted https://github.com/Chia-Network/chia-blockchain/commit/a86e85afca1302ec08810764925b0965bcc8634f since it even if it makes sense is unrelated to this PR. It's back to what i initially wanted to do here: Make it streamable and use byte conversions.

arvidn (commented):
I see this as unnecessary risk (by changing the types of constants to fixed width ints) and unnecessary complexity. It doesn't really matter how complex or how risky. I don't see any gains with moving in this direction.

I would rather move towards simplifying this code by removing the `constants_json` member and not replace it by anything. Just pass the `constants` member directly to the pool executor. There's precedence in every other parameter we submit to the pool using python's built-in pickle to ship the values over to the subprocess.

If the argument in favour of having two copies of the constants (`constants` and `constants_bytes`) is to improve performance by pre-computing the serialization; I can't imagine it matters that much (right now). But more importantly, the way to fix that is to copy the constants once to each subprocess and never copy them over again. They are constants after all.

arvidn:
I propose we do this instead: https://github.com/Chia-Network/chia-blockchain/pull/11848

xdustinface:
I think there is value to have well defined consensus constants with fixed size ints for all of them not just a handful being sized and others are just what they are but i also agree that doing this now afterwards brings some risk with it and at the end: What i wanted is to get rid of the `dataclass_from_dict` conversions of `ConsensusConstants` so im also fine with https://github.com/Chia-Network/chia-blockchain/pull/11848.

PR #11759, opened 2022-06-02
harvester: Make `PlotsRefreshParameter` streamable + use `from_json_dict`

xdustinface:
Also preparation for to restrict `dataclass_from_dict` to streamable classes.

### DIFF
--- a/tests/block_tools.py
+++ b/tests/block_tools.py
@@ -182,7 +182,6 @@ class BlockTools:
             updated_constants = updated_constants.replace(**const_dict)
         self.constants = updated_constants
 
-        self.refresh_parameter: PlotsRefreshParameter = PlotsRefreshParameter(batch_size=2)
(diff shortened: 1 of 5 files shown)


tests/block_tools.py
```diff
@@ -182,7 +182,6 @@ def __init__(
             updated_constants = updated_constants.replace(**const_dict)
         self.constants = updated_constants
 
-        self.refresh_parameter: PlotsRefreshParameter = PlotsRefreshParameter(batch_size=2)
```
xdustinface, line 185: Dropping `BlockTools.refresh_parameter` and using the ones from `BlockTools.plot_manager` below is required now since we create new params with `dataclasses.replace`.

trepca (approved):

wjblanke (approved):
aok

PR #11762, opened 2022-06-02
farmer: Make `Plot{Info|Path}RequestData` streamable + use `from_json_dict`

xdustinface:
Another preparation to restrict `dataclass_from_dict` to streamable classes. 

Based on:
- #11755

### DIFF
--- a/chia/rpc/farmer_rpc_api.py
+++ b/chia/rpc/farmer_rpc_api.py
@@ -1,6 +1,8 @@
 import dataclasses
 import operator
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional
+
+from typing_extensions import Protocol
 
 from chia.farmer.farmer import Farmer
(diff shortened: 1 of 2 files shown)


chia/rpc/farmer_rpc_api.py
```diff
@@ -1,50 +1,49 @@
...
     node_id: bytes32
-    page: int
-    page_size: int
+    page: uint32
+    page_size: uint32
     filter: List[str] = dataclasses.field(default_factory=list)
     reverse: bool = False
 
 
+PaginatedRequestData = Union[PlotInfoRequestData, PlotPathRequestData]
```
xdustinface, line 44: I had to drop the `Protocol` and use this `Union` instead because `mypy` doesn't seem to like it combined with frozen dataclasses. Im happy about suggestions though how we could still use the `Protocol`

AmineKhaldi (approved):

wjblanke (approved):
aok

PR #11764, opened 2022-06-02
Handle ephemeral ports and dual stack (ipv4 & ipv6)

### LINKED ISSUE #11628: [Bug] IPv6 RPC connections broken
### What happened?

https://github.com/Chia-Network/chia-blockchain/pull/11578 broke IPv6 RPC connections by hard coding `0.0.0.0` for the host at https://github.com/Chia-Network/chia-blockchain/pull/11578/files#diff-926449f427065cf100c0af5785ac61529c768ab4cccbef25ccb86e82074b753fR270.  Initial exploration started in https://github.com/Chia-Network/chia-blockchain/pull/11623.  Actual fix being worked in https://github.com/Chia-Network/chia-blockchain/pull/11764.

### Version

main: https://github.com/Chia-Network/chia-blockchain/commit/817baa3096ee7cb4f820bf04599a2d231faeff94

Linux

CLI

### Relevant log output

```shell
2022-05-24T14:53:31.336 full_node chia.rpc.util           : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia\rpc\util.py", line 16, in inner
  File "chia\rpc\rpc_server.py", line 172, in open_connection
ValueError: Start client failed, or server is not set
```
```

### PR DESCRIPTION
**Pull request by emlowe**
Handle cases of ephemeral ports while using `host=""` and a dual-stack system where the server will listen to multiple ports and the code needs to make sure it is consistent on which port it is using

When using port 0 as an ephemeral port like used in the tests - the system will pick a port to use, However, in a dual-stack (ipv4 & ipv6) system, 2 ports will be picked, one listening on the IPv4 stack and a different port for IPv6. The tests only need one, but they need to connect to the correct one - if the service advertises the IPv6 port, but the caller uses IPv4 to connect, it will fail to connect.

This is only a problem in the test code that uses 0 port - when you request a specific port (like in normal usage via config.yaml), the system will listen to that port on both IPv4 and IPv6 - so this is limited to cases when using port 0 which today is limited to CI.

Code is written to handle cases where `prefer_ipv6` is true, which should exercise the IPv6 stack, however, this is not tested in CI and so this is untested.

Fixes #11628

### DIFF
--- a/chia/server/server.py
+++ b/chia/server/server.py
@@ -28,10 +28,9 @@ from chia.server.ssl_context import private_ssl_paths, public_ssl_paths
 from chia.server.ws_connection import WSChiaConnection
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.peer_info import PeerInfo
-from chia.util.config import load_config
 from chia.util.errors import Err, ProtocolError
 from chia.util.ints import uint16
-from chia.util.network import is_in_network, is_localhost
+from chia.util.network import is_in_network, is_localhost, select_port
 from chia.util.ssl_check import verify_ssl_certs_and_keys
 
 max_message_size = 50 * 1024 * 1024  # 50MB
@@ -279,17 +278,7 @@ class ChiaServer:
         # prefer_ipv6 is set in which case we use the IPv6 port
         #
         if self._port == 0:
-            global_config = load_config(self.root_path, "config.yaml")
-            prefer_ipv6 = global_config.get("prefer_ipv6", False)
-            self._port = self.runner.addresses[0][1]  # sets the default to handle some edge cases
-            for x in self.runner.addresses:
-                ip_addy = ip_address(x[0])
-                if ip_addy.version == 6 and prefer_ipv6:
-                    self._port = x[1]
-                    break
-                elif ip_addy.version == 4 and not prefer_ipv6:
-                    self._port = x[1]
-                    break
+            self._port = select_port(self.root_path, self.runner.addresses)
 
         self.log.info(f"Started listening on port: {self._port}")
 
--- a/chia/server/server.py
+++ b/chia/server/server.py
@@ -1,4 +1,5 @@
 import asyncio
+import ipaddress
 import logging
 import ssl
 import time
@@ -28,6 +29,7 @@ from chia.server.ssl_context import private_ssl_paths, public_ssl_paths
 from chia.server.ws_connection import WSChiaConnection
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.peer_info import PeerInfo
+from chia.util.config import load_config
 from chia.util.errors import Err, ProtocolError
 from chia.util.ints import uint16
 from chia.util.network import is_in_network, is_localhost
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: chia/server/server.py [hunk 1]
```diff
@@ -267,13 +268,29 @@ async def start_server(self, on_connect: Callable = None):
...
         await self.site.start()
-        self._port = self.runner.addresses[0][1]
+        #
+        # On a dual-stack system, we want to get the (first) IPv4 port unless
+        # prefer_ipv6 is set in which case we use the IPv6 port
+        #
+        if self._port == 0:
+            global_config = load_config(self.root_path, "config.yaml")
+            prefer_ipv6 = global_config.get("prefer_ipv6", False)
+            self._port = self.runner.addresses[0][1]  # sets the default to handle some edge cases
```
altendky, line 284: Explain the corner cases?  Maybe do or maybe don't use it, but be aware of the search-and-find-or-default form of a for loop.  This way you don't pre-assign (maybe) wrong values.

```python
for x in y:
    if match(x):
        value = x
        break
else:
    # when we leave the loop without a break
    value = whatever
```

chia/server/server.py
```diff
@@ -267,13 +268,29 @@ async def start_server(self, on_connect: Callable = None):
...
+            for x in self.runner.addresses:
+                ip_addy = ip_address(x[0])
```
altendky, line 286:
```suggestion
                address = ip_address(raw_address)
```

altendky (commented):
Do I understand correctly that the basic issue is that sometimes we would get the IPv6 port back instead of the IPv4 that the surrounding code was assuming?

Duplicated code should go in a function.

It seems like this code should return all of the data and leave it to the caller to pick out what they want.  Probably via a form of the function suggested above.

chia/server/server.py
emlowe: Used this for/else syntax and added a utility function

chia/server/server.py
emlowe: thanks, used this to make the code more readable

arvidn (dismissed):
this looks good and is probably the simplest patch to fix the problem.

However, it would seem like a better approach to have the production code log all ports it's listening on, and have the *test* code fish out the right one from the logs.

All this picking the address logic, in production code, is only there to have a single log line for which port we're listening on.

chia/server/server.py
```diff
@@ -267,13 +267,19 @@ async def start_server(self, on_connect: Callable = None):
         # this port from the socket itself and update self._port.
         self.site = web.TCPSite(
             self.runner,
-            host="0.0.0.0",
+            host="",  # should listen to both IPv4 and IPv6 on a dual-stack system
```
altendky, line 270: @arvidn, this line is the fix for the production issue whereby you can not make IPv6 connections.  Fixing that causes tests to fail so the rest is needed to make the tests ok.

altendky (dismissed):
In some discussion I did suggest what I think is similar in that all listened to ports should be returned and then whatever is receiving that can decide what to do with it.  For example, the tests could pick whichever port they feel like to use.  I would _not_ suggest this data path be 'the logs'.

I already discussed the suggestion with Earle, but formalizing here.  Otherwise I'm good with this for now.  Other improvements can be handled separately against `main`.

chia/server/server.py
arvidn: right, I forgot about this when I made my comments. I just remembered the comment..

wjblanke (dismissed):
aok

chia/server/server.py
emlowe: yea, this one line change is enough for production, because the rest of the problems are actually limited to CI (we don't support using port=0 in production) - so the rest of the code is just a workaround around the fact that `port=0` ends up listening to multiple ports in CI when used in combination with `host=""`

chia/util/network.py
```diff
@@ -84,3 +86,22 @@ def is_trusted_inner(peer_host: str, peer_node_id: bytes32, trusted_peers: Dict,
...
 
     return True
+
+
+def select_port(root_path: Path, addresses: List[Any]) -> uint16:
+    global_config = load_config(root_path, "config.yaml")
+    prefer_ipv6 = global_config.get("prefer_ipv6", False)
+    selected_port: uint16
+    for x in addresses:
+        address_string, port = x[0], x[1]
```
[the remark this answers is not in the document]
emlowe: done

PsyDafke:
lgtm

altendky (approved):

PR #11522, opened 2022-05-13
RpcApiProtocol and RpcServiceProtocol

altendky:
Draft for:
- [x] Cleanup `TODO`s
- [x] Reconsider `custom_get_connections()` name
- [x] On hold for some other adjustments to get more application of and checking of this hinting effort
  - [x] https://github.com/Chia-Network/chia-blockchain/pull/12172
  - [x] https://github.com/Chia-Network/chia-blockchain/pull/12204
  - [x] https://github.com/Chia-Network/chia-blockchain/pull/12153
  - Other granular follow-ups

PR #11765, opened 2022-06-02
Confirm equality of client and server RPC routes

altendky:
Draft for:
- [x] Should follow https://github.com/Chia-Network/chia-blockchain/pull/11522 at this point


tests/util/rpc.py
```diff
@@ -2,17 +2,24 @@
...
     assert len(routes_api) > 0
-    for route in routes_api + routes_server:
-        assert route in routes_client
+
+    all_server_routes = {*routes_api, *routes_server}
+    all_client_routes = {*routes_client}
+    only_server = sorted(all_server_routes - all_client_routes)
+    only_client = sorted(all_client_routes - all_server_routes)
+
+    assert [only_server, only_client] == [[], []]
```
xdustinface, line 25:
```suggestion
    assert {*routes_client} == {*routes_api, *routes_server}
```

Any reason to not just do the above?

xdustinface (changes requested):

tests/util/rpc.py
xdustinface: or maybe rather:

```suggestion
    assert sorted(routes_client) == sorted(routes_api + routes_server)
```

xdustinface (approved):

emlowe (approved):

PR #11769, opened 2022-06-03
Typo fix

BrandtH22:


trepca:
thanks, but that puzzle is obsolete now and has been deleted/overwritten in newer commits

PR #11770, opened 2022-06-03
Protocol tests json

arvidn:
the main motivation for this patch is to add regression tests for the JSON dictionary representations for all network protocols, to ensure they don't accidentally change to become incompatible.

We already have this test for the streamable binary representation of messages (`test_network_protocol_files.py`).

In order to create this test, I extended the visitor in `build_network_protocol_files.py` to not only pass the network message object, but also the variable name. this allows that script to not just generate the test data, but also the test itself. This makes it much more convenient (and likely) to add new protocol messages to this test when we add them to the protocol.

With this new capability, I also made the existing test be generated by the build script.


tests/util/build_network_protocol_files.py
```diff
@@ -155,13 +155,106 @@ def visitor(obj: Any) -> None:
...
+
+    visit_all_messages(visitor)
+
+    return result
+
+
 if __name__ == "__main__":
-    filename = get_network_protocol_filename()
-    data = get_protocol_bytes()
-    if os.path.exists(filename):
```
arvidn, line 161: opening the file with `"w"` already truncates any existing file

tests/util/test_network_protocol_files.py
```diff
@@ -29,670 +17,461 @@ def parse_blob(input_bytes):
...
-    ret: List[str] = []
-    mod_name = mod.__name__
-    for sym in dir(mod):
-        obj = getattr(mod, sym)
-        if hasattr(obj, "__module__") and obj.__module__ == mod_name:
-            ret.append(sym)
-    return set(ret)
-
-
-def test_missing_messages_state_machine() -> None:
```
arvidn, line 535: I moved these tests into a separate file, since they are not generated

fchirica (dismissed):
LGTM

fchirica (approved):

wjblanke (approved):
aok

PR #11763, opened 2022-06-02
streamable: Turn `dataclass_from_dict` into `streamable_from_dict`

xdustinface:
This PR changes the behaviour of the original `dataclass_from_dict` to only work with valid streamable compatible classes and renames it `streamable_from_dict` to make sure we use the same types and pattern wherever we convert dicts to classes.

Based on: 
- #10652
- #11807
- #11848
- #11851 
- #11759  
- #11760
- #11762

PR #11772, opened 2022-06-03
types: `SpendBundle.from_json_dict` -> `SpendBundle.create_from_dict`

xdustinface:
Im open for other names (initially i had `from_json_dict_custom`) but at the end this is to avoid replacing the original `SpendBundle.from_json_dict` as preparation for #11763.

### DIFF
--- a/chia/rpc/full_node_rpc_api.py
+++ b/chia/rpc/full_node_rpc_api.py
@@ -624,7 +624,7 @@ class FullNodeRpcApi:
         if "spend_bundle" not in request:
             raise ValueError("Spend bundle not in request")
 
-        spend_bundle = SpendBundle.from_json_dict(request["spend_bundle"])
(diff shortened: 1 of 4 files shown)


arvidn:
I think the new name needs to indicate what its purpose is. why do we have a custom `from_json_dict()` for spendbundle? Presumably it's compatible with the generic one, to some extent (otherwise, why would it shadow the other one?)

Is there a way in python to prevent shadowing `classmethod` functions like this? I think we need to ensure things like this are prevented going forward.

xdustinface:
Note that i just realised this may also be called indirectly via `Streamable` classes which integrate `SpendBundle` which would make things only partially compatible after this patch. There is for example `MempoolItem` where it's used but obviously not with JSON so far.

xdustinface:
This should be deprecated and removed properly together with `to_json_dict` rather.

PR #11775, opened 2022-06-03
Add royalty percentage flag to the NFT minting command

AmineKhaldi:


chia/cmds/wallet.py
```diff
@@ -506,6 +506,9 @@ def nft_wallet_create_cmd(wallet_rpc_port: Optional[int], fingerprint: int) -> N
     show_default=True,
     callback=validate_fee,
 )
+@click.option(
+    "-rp", "--royalty-percentage", help="NFT royalty percentage in basis points", type=int, default=0, show_default=True
```
paninaro, line 510: I'd recommend adding an example in the help text, as it may still not be clear to the user what basis points are.

chia/cmds/wallet.py
```diff
@@ -506,6 +506,14 @@ def nft_wallet_create_cmd(wallet_rpc_port: Optional[int], fingerprint: int) -> N
     show_default=True,
     callback=validate_fee,
 )
+@click.option(
+    "-rp",
+    "--royalty-percentage",
+    help="NFT royalty percentage in basis points. Example: 1750 would represent 1.75%",
```
arvidn, line 512: perhaps this is a bit of a nit-pick; but "percent" means hundreds, and this is specified in ten thousandths. using the term "percentage" is odd in my mind. it seems "percentage" here is used to mean "fraction" or "factor".

trepca: we have updated the naming, thanks

trepca (approved):

PR #11778, opened 2022-06-03
add final coverage handling

altendky:
Notes:
- Windows coverage is being left unprocessed for now related to https://github.com/nedbat/coveragepy/issues/991.

Draft for:
- [x] Extract the test cleanup to another PR that addresses similar code more completely.
  - It was decided that improved flake reduction can be handled separately after this is merged.


tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -572,9 +572,10 @@ async def test_cat_endpoints(wallet_rpc_environment: WalletRpcTestEnvironment):
     assert name == next(iter(DEFAULT_CATS.items()))[1]["name"]
 
     # TODO: Investigate why farming only one block here makes it flaky
-    await farm_transaction_block(full_node_api, wallet_node)
-    await farm_transaction_block(full_node_api, wallet_node)
+    for _ in range(5):
```
almogdepaz, line 575: does it not work with the previous 2 block setup ?
seems strange that it will change for the worst

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -868,16 +873,17 @@ async def test_nft_endpoints(wallet_rpc_environment: WalletRpcTestEnvironment):
     )
     assert res["success"]
 
-    for _ in range(3):
-        await farm_transaction_block(full_node_api, wallet_1_node)
+    spend_bundle = SpendBundle.from_json_dict(json_dict=res["spend_bundle"])
+
+    await farm_transaction(full_node_api, wallet_1_node, spend_bundle)
```
almogdepaz, line 878: why did you change from farm_transaction_block to farm_transaction ?
if it is more consistent why not the other tests as well ?

altendky: I was just hacking around at broken stuff, not trying to correct everything.  I think the intent here is to process the transaction and then check some stuff after that.  Instead of guessing the minimum number of blocks required to always make that happen on all systems, it seemed more robust to use the function that would do we what want.  Get the transaction processed.  (though 'farm' is an odd word for that)

All of this not-really-coverage stuff should probably be taken as 'inspiration' for a separate pr to make improvements in these and similar areas instead of just being a side note here.

almogdepaz: i would leave it then and lets do a pr where we overhaul all these flakey tests instead

.github/workflows/test.yml
```diff
@@ -125,7 +125,7 @@ jobs:
         run: |
           # skip windows until https://github.com/nedbat/coveragepy/issues/991 is resolved
           rm coverage-data/*.tests_windows_* coverage-data/.*.tests_windows_*
-          coverage combine --data-file coverage-reports/.coverage coverage-data/
+          coverage combine--rcfile=.coveragerc --data-file coverage-reports/.coverage coverage-data/
```
emlowe, line 128: Is there a space missing here?

emlowe (approved):

PR #11786, opened 2022-06-04
black the generated protocol messages code

altendky:


tests/util/build_network_protocol_files.py
```diff
@@ -1,6 +1,9 @@
 # flake8: noqa
 
 import os
+import subprocess
+import sys
```
arvidn, line 5: I think you should remove the `sys` import (as highlighted by lgtm)

arvidn (dismissed):
looks good, apart from the new lgtm issue

tests/util/build_network_protocol_files.py
altendky: I guess I'll have to dig into why nothing else complained about this, but yes, that is a remnant from when I tried to use `-m black`.

```suggestion
```

altendky: Oh, the entire file got blanket `noqa`'ed.  Let's not do that either.

arvidn (approved):

arvidn:
thanks for fixing this!

wjblanke (approved):
aok

PR #11787, opened 2022-06-04
add `chia db backup --backup-file <backup_file_destination>`

neurosis69:
Add a simple CLI command to `vacuum` the sqlite blockchain db to a given target destination.
I've chosen the command `chia db backup` as I think it is more intuitive compared to `chia db vacuum`.

If `--backup-file` is not defined, then the backup will be created at the `database_path` specified in `config.yaml` using a `vacuumed_` file prefix.

**Example**
```shell
(venv) chia@ZenBook1 ~/chia-blockchain (db_backup)$ chia db backup 
reading from blockchain database: /home/chia/.chia/mainnet/db/blockchain_v2_testnet10.sqlite
writing to backup file: /home/chia/.chia/mainnet/db/vacuumed_blockchain_v2_testnet10.sqlite


Database backup finished : /home/chia/.chia/mainnet/db/vacuumed_blockchain_v2_testnet10.sqlite

(venv) chia@ZenBook1 ~/chia-blockchain (db_backup)$ ls -la /home/chia/.chia/mainnet/db/blockchain_v2_testnet10.sqlite
-rw-r--r-- 1 chia chia 26141528064 Jun  3 00:21 /home/chia/.chia/mainnet/db/blockchain_v2_testnet10.sqlite
(venv) chia@ZenBook1 ~/chia-blockchain (db_backup)$ ls -la /home/chia/.chia/mainnet/db/vacuumed_blockchain_v2_testnet10.sqlite
-rw-r--r-- 1 chia chia 24912998400 Jun  5 00:41 /home/chia/.chia/mainnet/db/vacuumed_blockchain_v2_testnet10.sqlite

```

I used a similar structure like the implementation for  `chia db validate`.

The sqlite feature in this PR is [`VACUUM INTO`](https://www.sqlite.org/lang_vacuum.html#vacuuminto), which can be used to create an online copy of a sqlite database.

**--no-indexes option**
Option for farmers, willing to exchange the reduced disk space for the backup with the increased node startup time for the index creation. It would make sense to make the node startup phase more verbose, like logging info about index recreation in INFO mode.

system|network|backup</br>type|backup</br>time|size|create indexes </br> during startup
---|---|---|---|---|---
i5-1135G7, SSD|testnet10|vacuum|4m47,046s|24G|0s
i5-1135G7, SSD|testnet10|--no_indexes|4m29,853s|19G|143s
i7-11700k,NVME-SSD|mainnet|vacuum|2m58.913s|74G|0s
i7-11700k,NVME-SSD|mainnet|--no_indexes|2m13.498s|53G|375s


arvidn (dismissed):
I think this looks good. Although, I think it would be good to have this code exercised by a test. there's a similar test at `tests/core/test_db_conversion.py`

wjblanke (dismissed):
aok

emlowe:
This can be done live, right? That is with a running node?

neurosis69:
> This can be done live, right? That is with a running node?

Yes, the sqlite command `VACUUM INTO` is used to create a consistent online copy of the database.

`--no-indexes` uses a similar approach like in my [alternative upgrade](https://github.com/Chia-Network/chia-blockchain/pull/10116#issuecomment-1036221729) PR, which was suggested by [Mr. Richard Hipp](https://sqlite.org/forum/forumpost/61ced9482a).

neurosis69:
Just noticed that the `detach database` is missing when using the
 `--no_indexes` option - will add that later.

neurosis69:
`VACUUM INTO` was introduced with sqlite version 3.27.
In the [install.sh](https://github.com/Chia-Network/chia-blockchain/blob/00c80ae6078f2c97765fe52b2a79aef5225d76d9/install.sh#L253) it is mentioned, that chia supports sqlite starting with vesion 3.8, so this could be an issue when using that command. Although the `--no_indexes` branch should work with versions between 3.8 and 3.27.

neurosis69:
There are a couple of possibilities to work around that issue.

some of them are
- remove the `--no_indexes` option beeing selectable and make it as only backup method available. 
  `chia db backup` would then create a copy of the blockchain db with no indexes by default.
- enforce the `chia db backup` command using the `--no_indexes` branch for versions < 3.27.0
- Just let the `VACUUM INTO` command fail if sqlite version is < 3.27 

I`d tend to make it simple and don't use `VACUUM INTO`, instead use the manual copy approach.
Then we could offer the user to decide if we create the indexes as part of the backup or as part of the node startup.

Any suggestions?

arvidn:
I think it's fine to not support 3.8, as long as the command fails with a reasonable error message without leaving any garbage files behind.

neurosis69:
> I think it's fine to not support 3.8, as long as the command fails with a reasonable error message without leaving any garbage files behind.

Changed it to abort if sqlite version is < 3.27.0 with an error message suggesting to use the backup method without indexes.

arvidn (approved):

neurosis69:
> Although, I think it would be good to have this code exercised by a test. there's a similar test at `tests/core/test_db_conversion.py`

Tried to create a test like suggested, created a draft PR in #11842 
But I wasn't able to handle hints properly, so I spared them out for now and also I was not sure what else would be necessary for such tests.

wjblanke (approved):
aok

PR #11795, opened 2022-06-05
More generated stuff work

altendky:
Draft for:
- [x] Based on https://github.com/Chia-Network/chia-blockchain/pull/11786
- [ ] Some continued work including review and verification of consistent coverage
- [ ] Can we tie back to the actual protocol definitions as a reference list to make sure we aren't missing anything?
- [ ] Address `TODO:`s
- [ ] Move `all_protocols.py` to tests


altendky:
Here is an attempt to document this change.

There were previously four generated files.  Binary seralized form of the example message instances, JSON of the basic types form of the example message instances, and a generated test file for each.  The bytes and JSON serialization have been merged into a single JSON formatted data file while the tests are just not generated anymore as they can be parametrized against the data instead and are in a single file.

- `protocol_messages_bytes-v1.0` -> `protocol_messages_bytes-v1.0.json`
- `protocol_messages_json.py` -> `protocol_messages_bytes-v1.0.json`
- `test_network_protocol_files.py` -> `test_network_protocol_json_bytes.py`
- `test_network_protocol_json.py` -> `test_network_protocol_json_bytes.py`

The Python-code-defined message instances in `network_protocol_data.py` have been reorganized into a dictionary to make them processable by code, such as to create fixtures that non-repetitive hand-coded tests can be parametrized by.  The ability to programmatically process this data also allows for removal of the `visitor()` functions that were hard coded forms of lists.

The generation script `build_network_protocol_files.py` now generates a single file `protocol_messages_bytes-v1.0.json` instead of four.  It no longer has the data redefined in the form of the `visitor()` code.  It no longer deals with any Python code generation so the black formatting calls are removed.

The tests are now parametrized against the defined examples.  This avoids the generation step and provides more granular checks with a result for each aspect of each message being checked.  Here's a list of the previous tests and an attempt to map them to the new tests.

- `test_network_protocol_files.py`
  - For each message:
    - Deserialization from bytes via `.from_bytes()` is equal to Python coded example data
      - `test_network_protocol_json_bytes.py::test_protocol_from_bytes_matches_instance`
    - Converting both the reference and deserialized instance back to bytes results in equal bytes.
      - It seems like this should have been comparing the reference bytes, not a reserialization.
      - `test_network_protocol_json_bytes.py::test_protocol_to_bytes_matches`
- `test_network_protocol_json.py`
  - For each message:
    - The `str()` of the reference JSON and the reference instance converted via `.to_json_dict()` are equal
      - `test_network_protocol_json_bytes.py::test_protocol_json_to_dict_str_matches`
    - The deserialization of the reference JSON via `.from_json_dict()` results in an instance equal to the reference
      - `test_network_protocol_json_bytes.py::test_protocol_json_from_json_instance_matches`

More cross checking tests etc have also been added.  More more are needed to actually check against the enumeration of message types, the message classes, and the defined endpoints.

PR #11803, opened 2022-06-06
Start wallet when starting data

emlowe:
Configure `chia start data` to also start the wallet, as data_layer needs the wallet to be running.


altendky (approved):
This reads ok.  I presume you tested it out both starting and stopping?  If yes, merge whenever you want.

PR #11807, opened 2022-06-06
Rust coin

arvidn:
This patch replaces the python implementation of `Coin` with the Rust native `Coin` from `chia_rs`.

The main benefits of the rust type are:

1. It has a fixed size, compact representation in memory
2. Its implementation of `name()` does not allocate memory and is very efficient

# Type annotations

The `chia_rs` module has type stubs ([here](https://github.com/Chia-Network/chia_rs/blob/main/wheel/chia_rs.pyi)). The python `Coin` has two `bytes32` and one `uint64` field. Since the python bindings makes it difficult to actually instantiate those types, the rust `Coin` technically has two `bytes` and one `int` field. The `bytes` fields are still annotated as `bytes32`, because those types are essentially compatible. They are guaranteed to be 32 bytes long.

However, the `amount` field is annotated as its actual `int` type. This is because `uint64` is not compatible with `int`. e.g. `int.to_bytes()` and `uint64.to_bytes()` have different behaviors. Also `bytes(int)` and `bytes(uint64)`. Using `int` helps mypy pin point the places where the `amount` field needs to be cast to a `uint64`.

Casting `amount` into a `uint64` is the majority of this patch.

The main difference between a `bytes` object that's guaranteed to be 32 bytes and `bytes32` is when formatting to string. Our `bytes32` type prints as hex, whereas `bytes` needs to be explicitly converted with `.hex()`.

# benchmarks

Running the `test_full_sync.py` benchmark gives the following results:

| test | before | after | after / before |
| --- | --- | --- | --- |
| wall-clock | 167.33 | 156.49 | 0.935 |
| CPU | 24.31% | 19.42% | 0.799 |

The CPU timings are a bit difficult to interpret. The CPU measurement is over the same batch of blocks (starting with 448), but since the patched version is faster, the CPU usage is over a slightly shorter time window. It's mostly a metric of the CPU usage to iowait ratio of the main thread.

# profiles

before:
slow-batch-00448

after:
slow-batch-00448


chia/wallet/did_wallet/did_wallet.py
```diff
@@ -1319,7 +1319,7 @@ def deserialize_backup_data(backup_data: str) -> DIDInfo:
         :return: DIDInfo
         """
         details = backup_data.split(":")
-        origin = Coin(bytes32(bytes.fromhex(details[0])), bytes32(bytes.fromhex(details[1])), uint64(int(details[2])))
+        origin = Coin(bytes32.fromhex(details[0]), bytes32.fromhex(details[1]), uint64(int(details[2])))
```
Quexington, line 1322: I think this should be `from_hexstr` and not `fromhex`.  Pretty sure this code is not in use so it doesn't matter but if I try to from `bytes32.fromhex("abcdef")` I get `ValueError: bad bytes32 initializer b'\xab\xcd\xef'`.

Quexington (dismissed):
I left a comment but pretty sure its on unused code so I think this is fine as far as I can see.  I also cannot vouch for the rust side of things working.

mariano54:
What if we override the .amount property to automatically convert to uint64, would something like this be possible?

chia/wallet/did_wallet/did_wallet.py
arvidn: you mean because the strings start with `0x`? if it isn't used, we really should remove it though

arvidn:
> What if we override the .amount property to automatically convert to uint64, would something like this be possible?

You mean derive from the rust version of `Coin` and add an `amount` property?

arvidn:
I've run a sync-from-scratch test (not all the way to the current peak though):

```
$ python ./tools/test_full_sync.py run --profile ~/chia-backup/db/blockchain_v2_mainnet.sqlite
PID: 1068064
Chia directory /tmp/tmp__3ek5fr
Can't find private CA, creating a new one in /tmp/tmp__3ek5fr to generate TLS certificates

height 100000 0.01 s/block
height 200000 0.01 s/block
height 300000 0.02 s/block
height 400000 0.02 s/block
height 500000 0.03 s/block
height 600000 0.02 s/block
height 700000 0.04 s/block
height 800000 0.04 s/block
height 900000 0.04 s/block
height 1000000 0.06 s/block
height 1100000 0.03 s/block
height 1200000 0.03 s/block
height 1300000 0.05 s/block
height 1400000 0.03 s/block
height 1500000 0.03 s/block
height 1600000 0.09 s/block
height 1700000 0.09 s/block
height 1772864 0.02 s/block   closing full node
```

chia/wallet/did_wallet/did_wallet.py
arvidn: I believe the error you got was because the hex string wasn't 32 bytes long. My understanding of the difference between `fromhex()` and `from_hexstr()` is that the latter expects a `0x` prefix

Quexington: Oh yeah whoops you're right :)

arvidn:
I experimented with wrapping the `Coin` type in python, to provide a `@property` for the `amount` field, that returns a `uint64`.
I think doing so is more risky than what's in this PR.

The current approach leverages `mypy` to detect the places where the `int` amount needs to be cast to `uint64`. All places in this PR that performs the cast were detected by mypy, except the ones in `CoinStorage` where we convert the amount to bytes, by calling `bytes(c.amount)`. Those were discovered by critical failure of the tests.

# Coin wrapper experiment

The approach I took to test the wrapping + property approach was to rename the rust type to `RustCoin`. and its member to `_amount`. Then add the following python code:

```
class Coin(RustCoin):
    @property
    def amount(self) -> uint64:
    return uint64(self._amount)
```

The reason to rename the rust type and its member was that none of my attempts to shadow the `amount` member with an `amount` property failed. It did not seem to work to call `super().amount` for example.

This change caused test failures originating from the streamable type checks. It turned out that the `reward_claims_incorporated` field in our precomputed blocks ended up having the native `Coin` type, rather than the python one. I don't understand how that happened. mypy did not detect any type errors and the reward claims list is clearly declared with the correct `Coin` type.

We store those blocks in test files and load them in by calling `FullBlock.from_bytes()`. When we deserialize `Coin`, we call the `parse_rust()` function on the underlying `Coin` type, which returns the native type, not our wrapped one. We would need to extend the wrapper to also wrap all objects we create from parsing streamable bytes.

It also occurred to me that renaming the underlying `amount` member also breaks the JSON generation and parsing. We are supposed to look for a JSON field called "amount", but with this change we'll look for "_amount" instead. The same problem appears in this case, `from_json_dict()` will return the native `Coin` type, not our wrapper.

Addressing that further increases complexity of this wrapper, and I don't think it would pull its weight anymore.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -441,8 +441,8 @@ def create_backup(self) -> str:
         """
         assert self.did_info.current_inner is not None
         assert self.did_info.origin_coin is not None
-        output_str = f"{self.did_info.origin_coin.parent_coin_info}:"
-        output_str += f"{self.did_info.origin_coin.puzzle_hash}:"
+        output_str = f"{self.did_info.origin_coin.parent_coin_info.hex()}:"
+        output_str += f"{self.did_info.origin_coin.puzzle_hash.hex()}:"
```
xdustinface, line 445: Why is this required here?

arvidn: our fixed-size byte types print as hex, whereas the built-in prints as binary.

xdustinface: I should have read the initial description it seems :) Hmm.. so we have `byte` objects but it's hinted as `bytes32`. Shouldn't we also just hint `bytes` instead like with `int`/`uint64` since its confusing when you look at the definition of `Coin` and it appears to be `bytes32` what it isn't?

arvidn: the reason I hinted the puzzle hash and parent id as `bytes32` despite actually being `bytes` (but guaranteed to be 32 bytes) was that there would have been a *lot* of places requiring a cast in this patch. In almost all of those cases the cast wouldn't technically be necessary, because our `bytes32` type is so similar to `bytes`

mariano54 (approved):

mariano54:
This will help with wallet performance too.

trepca (approved):
lgtm

PR #11812, opened 2022-06-07
Fix DID related NFT bugs

ytx1991:
One bug wip

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -135,6 +135,7 @@ class WalletRpcApi:
             # NFT Wallet
             "/nft_mint_nft": self.nft_mint_nft,
             "/nft_get_nfts": self.nft_get_nfts,
+            "/nft_get_by_did": self.nft_get_by_did,
             "/nft_get_info": self.nft_get_info,
             "/nft_transfer_nft": self.nft_transfer_nft,
             "/nft_add_uri": self.nft_add_uri,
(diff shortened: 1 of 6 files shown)


tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -605,9 +608,16 @@ async def test_nft_with_did_wallet_creation(two_wallet_nodes: Any, trusted: Any)
...
     sb = resp["spend_bundle"]
-
+    memos = compute_memos(sb)
+    assert memos
+    puzhashes = []
+    for x in memos.values():
+        puzhashes.extend(list(x))
+    assert len(puzhashes) > 0
+    for puzhash in puzhashes:
+        assert wallet_0.wallet_state_manager.puzzle_store.get_derivation_record_for_puzzle_hash(puzhash)
```
trepca, line 618: this works when non-synthetic pubkey is used too, so we need a different way to verify minted NFT will resync correctly

trepca (changes requested):

chia/rpc/wallet_rpc_api.py
```diff
@@ -1388,6 +1389,16 @@ async def nft_get_nfts(self, request) -> Dict:
...
         return {"wallet_id": wallet_id, "success": True, "nft_list": nft_info_list}
 
+    async def nft_get_by_did(self, request) -> Dict:
+        did_id: Optional[bytes32] = None
+        if "did_id" in request:
+            did_id = bytes32.from_hexstr(request["did_id"])
+        assert self.service.wallet_state_manager is not None
+        for wallet in self.service.wallet_state_manager.wallets.values():
+            if isinstance(wallet, NFTWallet) and wallet.get_did() == did_id:
+                return {"wallet_id": wallet.wallet_id, "wallet_info": wallet.wallet_info, "success": True}
```
paninaro, line 1399: I don't think we need to include `wallet_info`. The client doesn't do anything with this data today, and likely won't in the future.

ytx1991: I thought you may want to check if a NFT already in a wallet or something else. If you don't need this I can remove it.

paninaro: I appreciate that. For the intended use case, I can call `nft_get_nfts` for the returned `wallet_id`. I think that fits well with what the client needs.

`wallet_info` is really an internal implementation detail that shouldn't be exposed to the RPC callers.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -596,6 +597,7 @@ async def update_metadata(
         inner_solution = Program.to([solution_for_conditions(condition_list), 1])
         nft_tx_record = await self._make_nft_transaction(nft_coin_info, inner_solution, [puzzle_hash], fee)
         await self.standard_wallet.push_transaction(nft_tx_record)
+        await self.update_coin_status(nft_coin_info.coin.name(), True)
```
paninaro, line 600: Thanks for fixing this too.

paninaro (approved):
lgtm

trepca (approved):

PR #11816, opened 2022-06-07
Hint and lightly refactor chia.server.upnp.UPnP

altendky:
- `self.queue` is no longer a class attribute (effectively global), it is now an instance attribute (not global)
- `self.upnp` is now assigned initially as `None` then reassigned an actual instance in the thread, was previously just not present until the thread assignment
- Stricter mypy rules are now applied to `chia.server.upnp`
- The thread run target is now coded as a method instead of a nested function with a closure on `self`.


arvidn (approved):

wjblanke (approved):
aok

PR #11818, opened 2022-06-07
El.delete unknown key

emlowe:
Changes to essentially ignore deletes for keys that are not present. Those actions are ignored and do not raise exceptions. 

This allows for easier upsert semantics, as you can {delete,insert} without needing to worry about the key's existence beforehand

Also added code to handle cases where the changelist does not actually modify the tree - for example the changelist is empty, or has only delete actions for unknown keys. In this case, the update will raise an exception. This includes any changes that result in the same tree hash as before.

Previously one could provide an empty changelist and this would actually result in the singleton getting "updated" on chain (eg, `chia data update_data_store --changelist []` would actually create an on-chain transaction)


chia/data_layer/data_layer.py
```diff
@@ -117,11 +118,12 @@ async def batch_update(
                 key = change["key"]
                 await self.data_store.delete(key, tree_id, hint_keys_values)
 
-        await self.data_store.get_tree_root(tree_id)
-        root = await self.data_store.get_tree_root(tree_id)
+        new_root = await self.data_store.get_tree_root(tree_id)
+        if old_root == new_root:  # does this comparison work?
```
altendky, line 122: I think this was supposed to change to checking the hash?  We wouldn't want status or generation to affect this check.

```suggestion
        if old_root.node_hash == new_root.node_hash:
```

chia/data_layer/data_store.py
```diff
@@ -783,7 +783,8 @@ async def delete(
                 node = await self.get_node_by_key(key=key, tree_id=tree_id, lock=False)
             else:
                 if bytes(key) not in hint_keys_values:
-                    raise Exception(f"Key not found: {key.hex()}")
+                    log.debug(f"Request to delete an unknown key ignored: {key.hex()}")
+                    return
```
altendky, line 787: For now I think it doesn't matter here but note that changing from raising an exception to returning results in the db context manager committing the transaction instead of rolling it back.  This might be one of the cases where it would be nice to decouple that and be able to tell the transaction to roll back and then we return.

@arvidn, this isn't actually using the v2 wrapper at this point, but the idea is the same.  Our intent here is to _not_ have changed anything and it would be nice to say that to the db wrapper explicitly rather than just being really careful to have not put any db write code ahead of this.

tests/core/data_layer/test_data_rpc.py
```diff
@@ -132,6 +141,56 @@ async def test_create_insert_get(one_wallet_node_and_rpc: nodes_with_port, bt: B
...
+    assert wallet_node.server
+    await wallet_node.server.start_client(PeerInfo("localhost", uint16(full_node_api.server._port)), None)
+    assert wallet_node.wallet_state_manager is not None
+    ph = await wallet_node.wallet_state_manager.main_wallet.get_new_puzzlehash()
+    for i in range(0, num_blocks):
+        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph))
+        await asyncio.sleep(0.5)
+    funds = sum(
+        [calculate_pool_reward(uint32(i)) + calculate_base_farmer_reward(uint32(i)) for i in range(1, num_blocks)]
+    )
```
altendky, line 163: This can stay for now since I'm still working on getting the test helpers into main then back here to atari in their new form...  but this is definitely not going to be the form going forward.

altendky (changes requested):

chia/data_layer/data_layer.py
emlowe: changed and updated

tests/core/data_layer/test_data_rpc.py
```diff
@@ -132,6 +141,56 @@ async def test_create_insert_get(one_wallet_node_and_rpc: nodes_with_port, bt: B
...
+        data_rpc_api = DataLayerRpcApi(data_layer)
+        key = b"a"
+        value = b"\x00\x01"
+        changelist: List[Dict[str, str]] = [
+            {"action": "delete", "key": key.hex()},
+            {"action": "insert", "key": key.hex(), "value": value.hex()},
+        ]
+        res = await data_rpc_api.create_data_store({})
+        assert res is not None
+        store_id = bytes32(hexstr_to_bytes(res["id"]))
```
altendky, line 177:
```suggestion
        store_id = bytes32.from_hexstr(res["id"])
```

altendky (changes requested):
just a little tidbit

altendky (approved):

PR #11823, opened 2022-06-07
write blockchain db full node startup progress to debug.log

neurosis69:
This PR depends on #11787 and is a very simple approach to reflect the DB related full node startup progress.
In INFO logging or higher, we will write info about the Index that is beeing recreated.

In the past, when a new release introduced index changes, it was already not really transparent to the user, 
why the following node restart was stuck that long in between db connect and blockchain initialization. 

When using a db backup without indexes, then this will also be the case and at least the `coin_record` indexes 
are really big and will take some time when beeing recreated.


emlowe (approved):
Seems ok to me - only real downside is everytime node starts it will print out an additional 15 lines of info (ie, it will say "Creating xyz", but it's not really creating it)

PR #11826, opened 2022-06-08
Wp validate from fork

almogdepaz:
when validating a weight proof only validate sub epoch that are higher then the fork point

fixes a bug in the fork point calculation
fixes bug where we insert zero height fork to the WP cache


mariano54 (dismissed):
great change!

almogdepaz:
> great change!

had to push force to revert unwanted changes to chia-blockchain-gui

mariano54 (dismissed):

almogdepaz:
@emlowe approvals are stale but we can merge this

emlowe (approved):

PR #11834, opened 2022-06-08
Create stores for NFT

ytx1991:


chia/wallet/wallet_nft_store.py
```diff
@@ -0,0 +1,135 @@
...
+                "CREATE TABLE IF NOT EXISTS users_nfts("
+                " nft_id text PRIMARY KEY,"
+                " nft_coin_id text,"
+                " wallet_id int,"
+                " did_id text,"
+                " coin text,"
+                " lineage_proof text,"
+                " mint_height bigint,"
+                " status text,"
+                " full_puzzle text)"
```
trepca, line 41: why are you using `text` and not `blob` to store bytes directly?

trepca:
please add tests that test the store directly

chia/rpc/wallet_rpc_api.py
```diff
@@ -433,13 +433,17 @@ async def farm_block(self, request):
...
         wallet_type: Optional[WalletType] = None
         if "type" in request:
             wallet_type = WalletType(request["type"])
 
         wallets: List[WalletInfo] = await self.service.wallet_state_manager.get_all_wallet_info_entries(wallet_type)
-
+        if not include_data:
+            result: List[WalletInfo] = []
+            for wallet in wallets:
+                result.append(WalletInfo(wallet.id, wallet.name, wallet.type, ""))
```
trepca, line 445: do we still need this if we're moving coins to a separate store?

ytx1991: No harm to have

ytx1991:
Tested in tn10, need a resync

chia/wallet/wallet_nft_store.py
```diff
@@ -0,0 +1,140 @@
...
+            await cursor.close()
+        finally:
+            if not in_transaction:
+                await self.db_connection.commit()
+                self.db_wrapper.lock.release()
+
+    async def get_nft_list(
+        self, wallet_id: Optional[uint32] = None, did_id: Optional[bytes32] = None
+    ) -> List[NFTCoinInfo]:
+        sql: str = "SELECT * from users_nfts"
```
neurosis69, line 97: please don't just use `SELECT * from ...`, instead select only the columns needed.
[example of trade_store](https://github.com/Chia-Network/chia-blockchain/pull/10368#pullrequestreview-986535381)

chia/wallet/wallet_nft_store.py
```diff
@@ -0,0 +1,140 @@
...
+                    None if row[5] is None else LineageProof.from_json_dict(json.loads(row[5])),
+                    Program.from_bytes(row[8]),
+                    uint32(row[6]),
+                    row[7] == IN_TRANSACTION_STATUS,
+                )
+            )
+        return result
+
+    async def get_nft_by_id(self, nft_id: bytes32) -> Optional[NFTCoinInfo]:
+        cursor = await self.db_connection.execute("SELECT * from users_nfts WHERE nft_id=?", (nft_id.hex(),))
```
neurosis69, line 123: same like above, please be explicit with the columns needed

chia/wallet/wallet_nft_store.py
```diff
@@ -0,0 +1,140 @@
...
+            await self.db_wrapper.lock.acquire()
+        try:
+            cursor = await self.db_connection.execute(
+                "INSERT or REPLACE INTO users_nfts VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)",
+                (
+                    nft_coin_info.nft_id.hex(),
+                    nft_coin_info.coin.name().hex(),
+                    int(wallet_id),
+                    did_id.hex() if did_id else None,
+                    json.dumps(nft_coin_info.coin.to_json_dict()),
```
trepca, line 79: why not store all fields you're transforming into json or hex string into bytes? It saves space and avoids serialisation. Like you did for the`full_puzzle` field?

ytx1991: For lineage_proof and full_puzzle it make sense, but for other fields, they should be queryable. I think it's not worth to create another class just have lineage_proof and full_puzzle.

trepca: ok, fair point

trepca (dismissed):

mariano54:
Does the nft properly handle rollback and reorgs?
If I transfer an nft and that gets confirmed, and then that gets reverted, I will not see that nft in wallet right? @trepca

trepca:
@mariano54 good point, we rely on wallet state manager calling our wallet from `new_coin_state`. I assumed that it called when blocks are reverted on reorgs?

trepca:
@ytx1991 @mariano54's point still stands but it's generic to how we're handling new coin states and lets solve it in a separate PR

trepca (approved):

PR #11835, opened 2022-06-08
Correct UncurriedNFT

altendky:
Draft for:
- [ ] This is only present in release code so the PR has to target that, but it only matters if we start inheriting from this class which I'm guessing we won't for release so...  this can wait until after release.


altendky:
recreated in https://github.com/Chia-Network/chia-blockchain/pull/12239

PR #11837, opened 2022-06-08
Refactor away from post init for UPnP

altendky:
Draft for:
- [ ] Consideration of comments


xdustinface (dismissed):
Looks good but maybe make sure `thread` is `None` when `setup` is called?

emlowe (approved):

PR #11842, opened 2022-06-08
tests for `chia db backup`

neurosis69:
first test version, created similar to `tests_db_conversion.py` and `tests_db_validation.py` like suggested [here](https://github.com/Chia-Network/chia-blockchain/pull/11787#pullrequestreview-996290922)

wasn't sure how to handle hints, so I spared them out for now.


altendky:
isort in precommit will just make changes for you.  You can add and commit them.

For the hints, test functions just hint `-> None` as the return.  For the parameter you can start with searching for existing uses to figure out the proper hint.  Or look at the fixture to figure out what it provides.

https://github.com/Chia-Network/chia-blockchain/blob/157f462598c7eef6e33f7aa5e4edc8c578e01da5/tests/core/test_db_validation.py#L156

Then you just get down to the coin hints, as opposed to Python type hints, which I don't know either.  `:|`

PR #11848, opened 2022-06-09
remove the constants_json field

arvidn:
This simplifies the code by removing the duplicate constants field from `Blockchain`. It also unifies all our uses of the constants by consistently using our type, `ConsensusConstants`, without ever converting it to and from JSON.


xdustinface (approved):

wallentx:
@emlowe gave an audible "I approve" during a meeting

PR #11849, opened 2022-06-09
Initial iteration of the chia wallet nft set_did command

AmineKhaldi:


chia/cmds/wallet_funcs.py
```diff
@@ -799,3 +799,16 @@ async def list_nfts(args: Dict, wallet_client: WalletRpcClient, fingerprint: int
...
+
+async def set_nft_did(args: Dict, wallet_client: WalletRpcClient, fingerprint: int) -> None:
+    wallet_id = args["wallet_id"]
+    did_id = args["did_id"]
+    nft_coin_id = args["nft_coin_id"]
+    fee = args["fee"]
+    try:
+        response = await wallet_client.set_nft_did(wallet_id, did_id, nft_coin_id, fee)
+        spend_bundle = response["spend_bundle"]
+        print(f"DID set on NFT successfully with spend bundle: {spend_bundle}")
```
trepca, line 812: Technically we pushed that spend bundle to mempool and it's going to get processed at some point, but it's NFT hasn't be updated yet at this point. 
We could provide the user with command to check status of the transaction here? Similar to how `send_transaction` does it.

trepca, line 812: Technically we pushed that spend bundle to mempool and it's going to get processed at some point, but it's NFT hasn't be updated yet at this point. 
We could provide the user with command to check status of the transaction here? Similar to how `send_transaction` does it.

AmineKhaldi: That would require altering the API as it currently only returns the spend bundle. Should we change the wording here to indicate this spend bundle was sent? or?

trepca: ok, lets change the wording for now

AmineKhaldi: Sent a different wording.

trepca (approved):

emlowe (approved):

PR #11853, opened 2022-06-09
streamable: Move some class methods out of `Streamable`

xdustinface:

PR #11856, opened 2022-06-09
streamable: Expand `Field` and introduce `Streamable._streamable_fields`

xdustinface:
257d53c74aed9652153b8693bf802632985c4972 adds all functions we currently cache in the `*_FUNCTIONS_FOR_STREAMABLE_CLASS` globals into `streamable.Field` and fcc2caf4f13b0a5dcc32622cecaa5606879649c9 makes the fields related to each class members of the class and drops the global `FIELDS_FOR_STREAMABLE_CLASS`.

Based on:
- #11853 
- #11730 
- #11763

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -80,8 +80,7 @@ def create_fields(cls: Type[object]) -> StreamableFields:
     hints = get_type_hints(cls)
     fields = []
     for field in dataclasses.fields(cls):
-        hint = hints.get(field.name, None)
-        assert hint is not None
+        hint = hints[field.name]
         fields.append(
             Field(
                 name=field.name,
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -9,6 +9,7 @@ from typing import (
     Any,
     BinaryIO,
     Callable,
+    ClassVar,
     Collection,
     Dict,
     List,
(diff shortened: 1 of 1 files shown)


chia/util/streamable.py
```diff
@@ -67,28 +67,35 @@ class Field:
...
     hints = get_type_hints(cls)
-    fields = tuple(
-        Field(
-            name=field.name,
-            type=hints.get(field.name, None),
-            has_default=field.default is not dataclasses.MISSING or field.default_factory is not dataclasses.MISSING,
+    fields = []
+    for field in dataclasses.fields(cls):
+        hint = hints.get(field.name, None)
+        assert hint is not None
```
altendky, line 84: ?  At which point it can be inlined in the `Field` call like everything else and the list can be defined via comprehension.  Alternatively, a real message could be provided (with a 'regular' exception?) that enhances debuggability by reporting the the `cls` and the field name with some words.

```suggestion
        hint = hints[field.name]
```

xdustinface: Added but inlining it requires `hints[field.name]` in multiple places since `hint` is used also for the functions below.

altendky (approved):

wjblanke (approved):
aok

PR #11865, opened 2022-06-10
Initial iteration of updating RPCs and CLI to use bech32m identifiers for NFTs and DIDs

AmineKhaldi:


chia/rpc/wallet_rpc_api.py
```diff
@@ -538,7 +539,7 @@ async def create_new_wallet(self, request: Dict):
                         uint64(request.get("fee", 0)),
                     )
```
paninaro, line 541: GH isn't letting me add a comment in some earlier code, but line 522 where we collect the `backup_dids` probably needs some thought re: bech32m conversion. Might be worth applying the conversion there now to avoid having some parts of the RPC using hex strings and others using bech32m.

AmineKhaldi: Done.

trepca (dismissed):

paninaro:
I think we should update NFT wallet creation when a DID is specified:

wallet_rpc_api.py:581: (creating an NFT wallet with a DID ID)
`                    did_id = bytes32.from_hexstr(request["did_id"])`

chia/rpc/wallet_rpc_api.py
```diff
@@ -1186,13 +1187,13 @@ async def did_get_recovery_list(self, request):
         wallet_id = int(request["wallet_id"])
         wallet: DIDWallet = self.service.wallet_state_manager.wallets[wallet_id]
         recovery_list = wallet.did_info.backup_ids
-        recover_hex_list = []
+        recover_dids_list = []
         for _ in recovery_list:
```
paninaro, line 1191: nit: Can we use a variable name here? I expect that the original code was ignoring the value, but since we're actually passing it into a function, probably best to give it a name.

paninaro:
`did_get_information_needed_for_recovery` and `did_get_current_coin_info` could use benefit from encoding my_did.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1409,7 +1413,8 @@ async def nft_transfer_nft(self, request):
             return dict(success=False, error="target_address parameter missing")
         nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
         try:
-            nft_coin_info = nft_wallet.get_nft_coin_by_id(bytes32.from_hexstr(request["nft_coin_id"]))
+            nft_coin_id = decode_puzzle_hash(request["nft_coin_id"])
```
paninaro, line 1416: Actually, can we revert this one? Since we're still showing the NFT coin ID as hex (which I think is fine), I think it makes sense to keep this as a hex string.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1421,7 +1426,7 @@ async def nft_get_info(self, request: Dict) -> Optional[Dict]:
         assert self.service.wallet_state_manager is not None
         if "coin_id" not in request:
             return {"success": False, "error": "Coin ID is required."}
-        coin_id = bytes32.from_hexstr(request["coin_id"])
+        coin_id = decode_puzzle_hash(request["coin_id"])
```
paninaro, line 1429: This one is sort of interesting because the coin ID can either by the launcher ID or some descendent coin ID. At the risk of making things complicated, do you think we could check if the value has the `nft1` prefix, and if so, decode the puzzle hash, otherwise treat as a hex string?

If that's too complicated, I would go with just a hex string instead of bech32m here.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1512,7 +1517,8 @@ async def nft_add_uri(self, request) -> Dict:
         try:
             uri = request["uri"]
             key = request["key"]
-            nft_coin_info = nft_wallet.get_nft_coin_by_id(bytes32.from_hexstr(request["nft_coin_id"]))
+            nft_coin_id = decode_puzzle_hash(request["nft_coin_id"])
```
paninaro, line 1520: As above, I would recommend treating this value as being either a bech32m string or a hex string. And if that's too complicated, I think a hex string is the better option.

paninaro (approved):
Looks good to me. Thanks!

trepca (approved):

PR #11866, opened 2022-06-10
streamable: `_T_Streamable` -> `T_Streamable`

xdustinface:
It's used outside `chia.util.streamable`.


altendky (approved):

altendky (changes requested):
On second thought, I think the hint being used as it is is totally wrong.  Which is aside from this change, but still.  I'll come back to say more in a few minutes.

PR #11874, opened 2022-06-11
fix other case of self.did_id assignment after making it an @property

altendky:


AmineKhaldi (approved):

wallentx:
@trepca approved via convo

PR #11876, opened 2022-06-11
server: Hint `data` parameter of `make_msg`

xdustinface:


altendky (approved):
Imperfect, but useful.  Hinting 'can be accepted by `bytes()`' is non-trivial and this seems like a useful step.

emlowe (approved):

PR #11880, opened 2022-06-11
Add an all job for the test matrix

altendky:


.github/workflows/test.yml
```diff
@@ -75,3 +75,13 @@ jobs:
...
+
+  all:
+    uses: ./.github/workflows/all.yml
+    if: always()
+    needs:
+      - macos
+      - ubuntu
+      - windows
+    with:
+      needs: ${{ toJSON(needs) }}
```
webknjaz, line 87: Hey, I noticed you're reinventing the wheel here :)

I've got this https://github.com/re-actors/alls-green implemented a while back. Inspired by PyCA and some other research.

```suggestion
  all:  # This job does nothing and is only used for the branch protection
    if: always()

    needs:
      - macos
      - ubuntu
      - windows

    runs-on: ubuntu-latest

    steps:
      - name: Decide whether the needed jobs succeeded or failed
        uses: re-actors/alls-green@release/v1
        with:
          jobs: ${{ toJSON(needs) }}
```

altendky: I don't know about reinventing...  https://github.com/pytest-dev/pytest-twisted/commit/4c75379161ef94dc6b9d0bcf266b169e00e72e86  `:]`  but yes, thanks for the heads up on your action.  I noticed it over in the towncrier comments as well.  I'm glad to see this practice has some traction.

webknjaz: Yeah, we've faced this problem in aiohttp where automerge would merge broken PRs and I've accidentally seen the solution in PyCA/cryptography, then gone ahead and generalized it a bit :)

webknjaz: As for "one check for branch protection" idea, it's ancient. Coming from CIs like Zuul and Bors.

altendky:
closing in favor of https://github.com/Chia-Network/chia-blockchain/pull/20026 for logistical reasons.  thanks :ukraine:

PR #11884, opened 2022-06-12
Decrease flakiness and improve lint command

mariano54:


CONTRIBUTING.md
```diff
@@ -53,7 +53,7 @@ to configure how the tests are run. For example, for more logging: change the lo
 ```bash
 sh install.sh -d
 . ./activate
-black . && isort benchmarks build_scripts chia tests tools *.py && mypy && flake8 benchmarks build_scripts chia tests tools *.py
+black bechmarks build_scripts chia tests tools *.py && isort benchmarks build_scripts chia tests tools *.py && mypy && flake8 benchmarks build_scripts chia tests tools *.py
```
altendky, line 56: Is there a reason this isn't addressed by the black config?  The others are only specified on the CLI because I wasn't able to figure out how to embed the settings in their config files.

https://github.com/Chia-Network/chia-blockchain/blob/15342c7a2db2989e666e4e877686ba3c4f5b1c51/pyproject.toml#L12-L17

mariano54: No idea, maybe it's related to my shell doing something weird. But the old does not work and the new works

altendky: Could we debug that instead of changing it?  For example, it fails without the `.`.  What does it do for you?

```console
$ venv/bin/black .
All done! ✨ 🍰 ✨
625 files left unchanged.
$ venv/bin/black
No Path provided. Nothing to do 😴
```

mariano54:
```
Using configuration from /home/mariano54/Chia/chia-blockchain/pyproject.toml.
venv/bin/python3 ignored: is a symbolic link that points outside /home/mariano54/Chia/chia-blockchain
venv/bin/python3.10 ignored: is a symbolic link that points outside /home/mariano54/Chia/chia-blockchain
venv/bin/python ignored: is a symbolic link that points outside /home/mariano54/Chia/chia-blockchain
^C
```

mariano54: It's looking at folders that it shouldn't be looking at like venv and node_modules

PR #11886, opened 2022-06-12
Add more optimizations

mariano54:
* Adds various small caches
* `RequestPuzzleSolution` is parsed as a `SerializedProgram` by default instead of `Program`

Trusted sync time for the toilet wallet (100k tx):
Before: 290 seconds
After: 185 seconds

Untrusted sync:
Before: 28 minutes
After: 13 minutes

### DIFF
--- a/chia/protocols/wallet_protocol.py
+++ b/chia/protocols/wallet_protocol.py
@@ -2,7 +2,7 @@ from dataclasses import dataclass
 from typing import List, Optional, Tuple
 
 from chia.types.blockchain_format.coin import Coin
-from chia.types.blockchain_format.program import Program, SerializedProgram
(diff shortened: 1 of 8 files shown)


Quexington (approved):
This seems fine from what I can see.  I _think_ the change to the `PuzzleSolutionResponse` is backwards compatible right? Since `SerializedProgram` and `Program` serialize to bytes the same?

emlowe (approved):

mariano54:
> This seems fine from what I can see. I _think_ the change to the `PuzzleSolutionResponse` is backwards compatible right? Since `SerializedProgram` and `Program` serialize to bytes the same?

Yes exactly

PR #11887, opened 2022-06-12
New NFT chialisp

Quexington:
This contains some optimizations/changes to how both NFT1s and NFT0s work.  Here's a summary of the changes:
```
STATE LAYER
- The state layer is now implemented as a single loop.  It strips out the odd create coin and -24, then puts the wrapped coin condition as well as any conditions out of the metadata updater at the end.
- There's also no longer any need to pass in the amount.  Previously, the amount of the NFT could not change which I'm pretty sure is not what we want.
- The state layer was also used to remove ALL magic conditions, which I think is not what we want.  This means you couldn't melt the singleton and it made the ownership puzzle broken because it relied on the layer above to filter out the negative condition.

OWNERSHIP LAYER
- I optimized the ownership layer similarly to the state layer.  It's one loop that sees the odd coin and the magic condition, then punts their morph/concatenation to the end of the loop.
- You'll also notice that there's an extra spot in the loop where we CREATE_PUZZLE_ANNOUNCEMENT.  This announcement is <prefix> + (sha256tree transfer_program_solution).  The reason for the prefix is that we will block any 40 byte announcements coming out of the inner puzzle with the same prefix.  Basically, we want to create an announcement of the TP solution that we can trust, but I didn't want to take the CAT approach of morphing all of the inner puzzle's conditions.
- We also no longer rely on the state layer to filter out our magic condition, we do it ourselves.
- Finally, the transfer program can no longer morph the conditions.  I changed this partly because of implementation simplicity and partly because it's more consistent with our other designs like TAILs and MUs.  We stopped needing to morph the conditions in the TP when we removed the concept of magic agg sig, but we kept the condition morphing.

TRANSFER PROGRAM
- We no longer need to enforce anything about the create coin, it's allowed to be an arbitrary puzzle.  This simplified the module a lot, and now we just assert a DID announcement if there's a new owner and assert the royalty announcements if there's anything in the trade prices list.
- The XCH royalties work like the CAT royalties now (they go through OFFER_MOD).  This made things simpler on the offer end and I think it makes more sense regardless.
```


chia/wallet/puzzles/nft_state_layer.clvm
```diff
@@ -28,62 +34,55 @@
...
-      (metadata_updater_loader
-       METADATA_UPDATER_PUZZLE_HASH METADATA (a INNER_PUZZLE solution)
-      )
-      my_amount
-      0
-    )
+  (wrap_odd_create_coins
+    NFT_STATE_LAYER_MOD_HASH
+    (a INNER_PUZZLE inner_solution)
+    ()
```
ytx1991, line 84: Why hardcode () here?

chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -72,9 +70,6 @@ class UncurriedNFT:
     supports_did: bool
     """If the inner puzzle support the DID"""
 
-    owner_pubkey: Optional[G1Element]
```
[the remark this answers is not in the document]
Quexington: There is still a p2_puzzle hash, we just don't care about the pubkey in it anymore. It works like every other wallet now.

chia/wallet/puzzles/nft_state_layer.clvm
Quexington: We start the loop with this set to `()` and we expect it to get set during the loop.  We check at the end of the recursion whether or not it's still `()`, and if it is, we raise.

trepca (approved):

PR #11889, opened 2022-06-13
rpc: Introduce `rpc_server.Endpoint` hint alias

xdustinface:


altendky (dismissed):

altendky (approved):
I'll follow up with something like an `EndpointResult` hint to apply everywhere so we can have more confidence the `Endpoint` hint is actually being follow.

emlowe (approved):
seems ok - a brief description of the rationale would be useful in the PR.

PR #11894, opened 2022-06-13
WIP spam filter

mariano54:
* Removes a few unused config values
* Adds two new config values for filtering XCH spam. This enables syncing much much faster for anyone that has dust in their wallet. All the TXs are still downloaded, but some are filtered out and not processed

- [ ] tests for coin store
- [ ] tests for spam filter
- [ ] decide defaults

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -225,8 +225,6 @@ class WalletNode:
 
         assert self.wallet_state_manager is not None
 
-        self.config["starting_height"] = 0
-
         if self.wallet_peers is None:
             self.initialize_wallet_peers()
 
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -657,7 +657,12 @@ class WalletNode:
         concurrent_tasks_cs_heights: List[uint32] = []
 
         # Ensure the list is sorted
-        items = sorted(items_input, key=last_change_height_cs)
+
+        items_input_2 = list(sorted(items_input, key=last_change_height_cs))
+        self.log.info(f"Filtering spam, total items: {len(items_input_2)}")
+        items = [item for item in items_input_2 if (await self.wallet_state_manager.is_not_standard_spam(item))]
+        # items = list(filter(self.wallet_state_manager.is_not_standard_spam, items))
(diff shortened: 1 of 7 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -585,6 +585,42 @@ async def determine_coin_type(
...
+                and cs.coin.amount < xch_spam_amount
+                and (cs.coin.puzzle_hash in is_standard_wallet_phs or await self.is_standard_wallet_tx(cs))
+            ):
+                is_standard_wallet_phs.add(cs.coin.puzzle_hash)
+                if small_unspent_count < spam_filter_after_n_txs:
+                    self.log.info(f"Not filteed 1 {small_unspent_count}")
+                    filtered_cs.append(cs)
+                small_unspent_count += 1
+            else:
+                self.log.info(f"Not filteed 2 {small_unspent_count}")
```
hoffmang9, line 612: Is the "filteed" spelling intentional?

mariano54: just debug stuff i will remove

chia/wallet/wallet_node.py
```diff
@@ -657,7 +655,10 @@ async def receive_state_from_peer(
         concurrent_tasks_cs_heights: List[uint32] = []
 
         # Ensure the list is sorted
-        items = sorted(items_input, key=last_change_height_cs)
+
+        self.log.info(f"Filtering spam, total items: {len(items_input)}")
+        items = await self.wallet_state_manager.filter_spam(list(sorted(items_input, key=last_change_height_cs)))
```
trepca, line 660: could this be done on the full_node side? ie wallet node provides a min xch amount per coin?

mariano54: yes it could, but we would lose a little bit of the power. For example we can't specify that we only filter if we have more than 200 transctions. We also cannot filter based on CAT asset id, which we plan to do later.

On the other hand, it would help a lot for wallets with many TXs as having any more than 100k creates a massive message which can exceed max message sizes

trepca: maybe both then :)

PR #11895, opened 2022-06-13
Rename the outdated singleton that atari uses

Quexington:


altendky (approved):
Test catch-up merge done in https://github.com/Chia-Network/chia-blockchain/pull/11896.  Earle tested and said that with a db delete and resync it seems to work fine.  Thanks again quex.

PR #11903, opened 2022-06-14
another assert to make sure wallet is correct

wjblanke:


altendky:
I would expect this to target `release/1.4.0`.  It is 'just tests', but in so much as it does make this more reliable then I would think we would want that in the release ci runs.

AmineKhaldi:
> I would expect this to target `release/1.4.0`. It is 'just tests', but in so much as it does make this more reliable then I would think we would want that in the release ci runs.

Bill is okay with that. I'll retarget, close and reopen.

altendky:
Yeah, gotta rebase or make a new PR.  Otherwise you are bringing main into release.

altendky:
Just made it draft to avoid accidental merging until the above comment is addressed.

AmineKhaldi:
> Just made it draft to avoid accidental merging until the above comment is addressed.

Thanks, should be good to go.

altendky (dismissed):

altendky (approved):
_This_ got conflicted twice today?  Wow...

Issue #11897, opened 2022-06-13
[Bug] Error message every time `chia plotnft show` is ran


wallentx:
### What happened?

On mainnet, from current HEAD of `release/1.4.0` (b8e34c856)

Every time I run `chia plotnft show`, the plotnft info shows as expected up until it is supposed to show the pool payout address, where it spits out:
```
Exception from 'wallet' <bytes32: 84d52a439ed7d11b73054198add4a2b34c34b56c03abf67198d5f3d70753ee5b>
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f39a1645a80>
Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7f39a1698c40>, 563381.436435318)]']
connector: <aiohttp.connector.TCPConnector object at 0x7f39a1644e20>
```
This happens on a wallet that has a plotnft.

### Version

b8e34c856

### What platform are you using?

Linux

### What ui mode are you using?

CLI

### Relevant log output

```shell
No relevant debug log output
```

emlowe:
`plotnft show` does need the farmer to be running - see #9919. So this might be related

PR #11911, opened 2022-06-14
cmds: Only access existing pool states in `pprint_pool_wallet_state`

xdustinface:
Fixes issues like #11897 where the command fails if the state for the wallet with `launcher_id` gets printed but the `launcher_id` isn't available in the farmer's pool state list (i.e. its not in the config).

Note: This moves the `Payout instructions` output above `Relative lock height` to avoid needing another `pool_state_dict is not None` :)

### DIFF
--- a/chia/cmds/plotnft_funcs.py
+++ b/chia/cmds/plotnft_funcs.py
@@ -8,7 +8,7 @@ import json
 import time
 
 from pprint import pprint
-from typing import List, Dict, Optional, Callable
+from typing import Any, List, Dict, Optional, Callable
 
 from chia.cmds.units import units
 from chia.cmds.wallet_funcs import print_balance, wallet_coin_unit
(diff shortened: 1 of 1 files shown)


emlowe:
This is interesting - the wallet knows about this launcher_id - but it's not in config.yaml?

wallentx:
> This is interesting - the wallet knows about this launcher_id - but it's not in config.yaml?

I had it pulled out of my config.yaml to remove confusion since its an unusable plotnft for me.
(which resulted in confusion for Chia)

wallentx:
Checked this out and verified this fixed it.

altendky (approved):
reads well, thanks

emlowe:
I'm somewhat conflicted whether this is a must-have for 1.4. AFAIK it's not a regression in 1.4 and it doesn't seem that critical.

emlowe (approved):
Update: this is a regression in 1.4 - so on that basis, I approve

PR #11919, opened 2022-06-15
Merge release/1.4.0's 73ecb10 into main

AmineKhaldi:


altendky (approved):
While the `uint64()` added in [`5be6db6` (#11919)](https://github.com/Chia-Network/chia-blockchain/pull/11919/commits/5be6db6971d13c65132b7aab29f1e5899cea8cb0) looks to be required now, at least for hinting, it doesn't seem like it should be needed long term.  `chia_rs` does indicate that the type should be `u64` https://github.com/Chia-Network/chia_rs/blame/59f6709693b7b68ff2d5cdf2fe8711f0f537e33b/wheel/src/coin.rs#L25 but the generated hints just say `int`.  I haven't checked the actual runtime type of either the amount nor the others that are supposed to be `bytes32`.

```python
class Coin:
    parent_coin_info: bytes32
    puzzle_hash: bytes32
    amount: int
```

PR #11927, opened 2022-06-16
rpc: Be specific about `*args` in `RpcServer` state changed methods

xdustinface:


chia/rpc/wallet_rpc_api.py
```diff
@@ -153,31 +153,19 @@ def get_routes(self) -> Dict[str, Callable]:
...
 
-    async def _state_changed(self, *args) -> List[WsRpcMessage]:
+    async def _state_changed(self, change: str, change_data: Dict[str, Any]) -> List[WsRpcMessage]:
         """
         Called by the WalletNode or WalletStateManager when something has changed in the wallet. This
         gives us an opportunity to send notifications to all connected clients via WebSocket.
         """
         payloads = []
-        if args[0] is not None and args[0] == "sync_changed":
+        if change in {"sync_changed", "coin_added"}:
```
altendky, line 162: I think that handling `"coin_added"` here maybe misses it having the `"state"` key in the `change_data` dict?

chia/rpc/wallet_rpc_api.py
```diff
@@ -153,31 +153,19 @@ def get_routes(self) -> Dict[str, Callable]:
...
-            data["wallet_id"] = args[1]
-        if args[2] is not None:
-            data["additional_data"] = args[2]
-
-        payloads.append(create_payload_dict("state_changed", data, self.service_name, "wallet_ui"))
-
-        if args[0] == "coin_added":
-            payloads.append(create_payload_dict(args[0], data, self.service_name, "metrics"))
+        if "wallet_id" in change_data or "additional_data" in change_data:
+            change_data["state"] = change
```
altendky, line 167: Mutating inputs isn't usually great.

altendky (commented):
Can we use `object` instead of `Any` if we don't replace the dict with a class or a `TypedDict`?

Maybe separately, but is it plausible to use `Literal` here instead of `str` for `change`?  Or maybe an enumeration of strings.

xdustinface:
> Can we use `object` instead of `Any` if we don't replace the dict with a class or a `TypedDict`?
> 
> Maybe separately, but is it plausible to use `Literal` here instead of `str` for `change`? Or maybe an enumeration of strings.

I just adjusted it to match all the other state changed methods we have all over the place. They are `Dict[str, Any]` (if they are even hinted). You can in your new RPC API Protocol just hint how you want it to be and then adjust all where it complains, right?

See the [here](https://github.com/Chia-Network/chia-blockchain/compare/55e4fbedddf1cf16cee4be981bd5d0d2fe7a4000..3a28953ca5e69e97bb3066fab293c89b24dba9e1), this should address the two issues you pointed out above :)

altendky (approved):
i think...  i look forward to this being refactored again.

emlowe (approved):
an improvement - it would seem a class for "change_data" would make some sense

PR #11942, opened 2022-06-16
Merge release/1.4.0's 45dee3e into main

AmineKhaldi:


altendky (approved):
Caveat about this shouldn't be needed has already been mentioned previously and ticketed.  But, seems appropriate for the present state of main.  Will add this to the ticket so we have a list.

PR #11944, opened 2022-06-16
Pre-launch bug fix N in 1

ytx1991:
Add: set_nft_status API

Fix:
Add supports_did check for set_did API
Refine the did_id inputs
Fix NFT transaction status
Rewrite ownership logic in the wallet state manager


paninaro:
What's the intended use case for this RPC? I'm not aware of the requirements.

paninaro:
> > What's the intended use case for this RPC? I'm not aware of the requirements.
> 
> For reset the NFT transaction status after use cancelled the offer.

I think it would be best if the backend handled this without requiring the frontend (GUI and CLI) to call an additional RPC.

ytx1991:
> > > What's the intended use case for this RPC? I'm not aware of the requirements.
> > 
> > 
> > For reset the NFT transaction status after use cancelled the offer.
> 
> I think it would be best if the backend handled this without requiring the frontend (GUI and CLI) to call an additional RPC.

That's the idea case. But in case of that doesn't work as expected, we still need an API to reset the status manually.

trepca:
what are some reasons that it wouldn't work as expected?

ytx1991:
> what are some reasons that it wouldn't work as expected?
I assume offer cancelling is a one-time event. If it failed for some reasons, we cannot redo it.

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -1261,6 +1262,8 @@ async def test_nft_set_did(two_wallet_nodes: Any, trusted: Any) -> None:
         time_left -= 0.5
     nft_wallet_1_id = coins_response.get("wallet_id")
     assert nft_wallet_1_id
+    assert len(wallet_node_0.wallet_state_manager.wallets[nft_wallet_0_id].my_nft_coins) == 0
```
trepca, line 1202: is this checking that old wallet doesn't have coins anymore?

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -1226,7 +1227,7 @@ async def test_nft_set_did(two_wallet_nodes: Any, trusted: Any) -> None:
         await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph))
 
     # Check DID NFT
-    time_left = 5.0
+    time_left = 10.0
     coins_response = {}
     while time_left > 0:
```
trepca: can you migrate these tests to new utility functions `make_new_block_with` and `wait_rpc_state_condition`

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -1251,7 +1252,7 @@ async def test_nft_set_did(two_wallet_nodes: Any, trusted: Any) -> None:
     for i in range(1, num_blocks):
```
trepca: no need to create multiple blocks

trepca: also, you need to fetch spend bundle from `nft_set_nft_did` and wait for it to get into mempool before continuing to create blocks, this applies to other calls that push spend bundles

chia/wallet/wallet_state_manager.py
```diff
@@ -721,67 +727,74 @@ async def handle_nft(
...
+                    uncurried_nft.singleton_launcher_id.hex(),
+                    old_did_id,
+                )
+                nft_wallet: NFTWallet = self.wallets[wallet_info.id]
+                await nft_wallet.remove_coin(coin_spend.coin, in_transaction=True)
+            if nft_wallet_info.did_id == new_did_id:
+                self.log.info(
+                    "Adding new NFT, NFT_ID:%s, DID_ID:%s",
+                    uncurried_nft.singleton_launcher_id.hex(),
+                    new_did_id,
```
trepca, line 781: it might be good to check if we already have this coin in the store and remove it to be safe? or does add_coin function replace the coin if it shares the same id?

tests/wallet/nft_wallet/test_nft_wallet.py
ytx1991: yes

chia/rpc/wallet_rpc_api.py
```diff
@@ -1406,10 +1406,14 @@ async def nft_set_nft_did(self, request):
             assert self.service.wallet_state_manager is not None
             wallet_id = uint32(request["wallet_id"])
             nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
-            did_id: Optional[bytes32] = None
-            if "did_id" in request:
-                did_id = decode_puzzle_hash(request["did_id"])
+            did_id = request.get("did_id", "")
```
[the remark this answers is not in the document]
ytx1991: The problem is if we map None to unassigned NFT, then user have to always input the DID_ID

chia/wallet/wallet_state_manager.py
ytx1991: It will replace

chia/rpc/wallet_rpc_api.py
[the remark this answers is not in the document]
ytx1991: Discussed in Keybase

trepca (dismissed):

wjblanke (dismissed):
aok

paninaro (approved):
Approving based on the latest commit and prior approvals from @trepca and @wjblanke

trepca (approved):

PR #11947, opened 2022-06-17
NFT wallet test updates

trepca:


tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -33,13 +33,43 @@ async def tx_in_pool(mempool: MempoolManager, tx_id: bytes32) -> bool:
...
+
+
+async def wait_rpc_state_condition(timeout, coroutine, params, condition_func) -> Dict:
+    while timeout > 0:
+        resp = await coroutine(*params)
+        if condition_func(resp):
+            assert isinstance(resp, dict)
+            return resp
+        await asyncio.sleep(0.5)
+        timeout -= 0.5
```
altendky, line 51: This drifts.  It should be of the form of identifying the termination time (use `time.monotonic()` not `time.time()`) and just looping until then.  No timeout decrement.

It's not ready yet, but looking to the future there will be https://github.com/Chia-Network/chia-blockchain/pull/11535/files#diff-ed857c49736dea1b775dddce47f5ec9989dd7525d3558c656db792a3c42c9830R22-R37 and probably some layers around it.

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -33,13 +33,43 @@ async def tx_in_pool(mempool: MempoolManager, tx_id: bytes32) -> bool:
...
 
+def assert_rpc_response(response, condition_func=None) -> Dict:
+    assert isinstance(response, dict)
+    assert response.get("success")
+    if condition_func:
+        assert condition_func(response)
+    return response
+
+
+async def wait_rpc_state_condition(timeout, coroutine, params, condition_func) -> Dict:
```
altendky, line 44: I want less callbacks.  But, I don't have the alternative tooling in place yet.  Going forward, try to find ways to avoid callbacks.  Python has a lot of tools.

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -33,13 +33,43 @@ async def tx_in_pool(mempool: MempoolManager, tx_id: bytes32) -> bool:
     return True
 
 
+def assert_rpc_response(response, condition_func=None) -> Dict:
+    assert isinstance(response, dict)
```
altendky, line 37: Generally hinting should be the first line of defense against this.  Some work is being done over in https://github.com/Chia-Network/chia-blockchain/pull/11889 to centralize the requirements and force them to be propagated, at least for fully hinted code which this apparently is not.  `:[`

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -321,16 +352,7 @@ async def test_nft_wallet_rpc_creation_and_list(two_wallet_nodes: Any, trusted:
...
-    while time_left > 0:
-        coins_response = await api_0.nft_get_nfts(dict(wallet_id=nft_wallet_0_id))
-        if len(coins_response.get("nft_list", [])) > 0:
-            break
-        await asyncio.sleep(0.5)
-        time_left -= 0.5
-    else:
-        raise AssertionError("NFT not minted")
-    await time_out_assert(15, wallet_1.get_pending_change_balance, 0)
+    await wait_rpc_state_condition(5, api_0.nft_get_nfts, [dict(wallet_id=nft_wallet_0_id)], lambda x: x["nft_list"])
```
altendky, line 355: `dict(x=3)` makes it harder to search for key accesses than `{"x": 3}`

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -33,13 +33,43 @@ async def tx_in_pool(mempool: MempoolManager, tx_id: bytes32) -> bool:
     return True
 
 
+def assert_rpc_response(response, condition_func=None) -> Dict:
+    assert isinstance(response, dict)
+    assert response.get("success")
```
altendky, line 38: Does not-success not result in an exception from the client?

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -345,27 +367,18 @@ async def test_nft_wallet_rpc_creation_and_list(two_wallet_nodes: Any, trusted:
...
-            assert len(uris) == 2
-            assert "https://chialisp.com/img/logo.svg" in uris
-            assert bytes32.fromhex(coins[1].to_json_dict()["nft_coin_id"][2:]) in [x.name() for x in sb.additions()]
-        except AssertionError:
-            if time_left < 1:
-                raise
-        await asyncio.sleep(0.5)
-        time_left -= 0.5
+    coins_response = await wait_rpc_state_condition(
+        5, api_0.nft_get_nfts, [{"wallet_id": nft_wallet_0_id}], lambda x: x["success"] and len(x["nft_list"]) == 2
```
altendky, line 371: Could we pass by name so we can read what each thing is here?  That would also end up with each parameter on a separate line which will be more readable as well.

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -345,27 +367,18 @@ async def test_nft_wallet_rpc_creation_and_list(two_wallet_nodes: Any, trusted:
...
-            assert len(coins) == 2
-            uris = []
-            for coin in coins:
-                assert not coin.supports_did
-                uris.append(coin.data_uris[0])
-                assert coin.mint_height > 0
-            assert len(uris) == 2
-            assert "https://chialisp.com/img/logo.svg" in uris
-            assert bytes32.fromhex(coins[1].to_json_dict()["nft_coin_id"][2:]) in [x.name() for x in sb.additions()]
-        except AssertionError:
```
altendky, line 364: Thank you thank you thank you for getting rid of this.

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -587,8 +585,7 @@ async def test_nft_with_did_wallet_creation(two_wallet_nodes: Any, trusted: Any)
     spend_bundle = spend_bundle_list[0].spend_bundle
     await time_out_assert_not_none(5, full_node_api.full_node.mempool_manager.get_spendbundle, spend_bundle.name())
 
-    for _ in range(1, num_blocks):
-        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph))
+    await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph))
```
altendky, line 588: Unrelated to the change, do you really need the reward here?  If not then not receiving it makes the accounting for the expected balance that much simpler.

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -673,21 +666,13 @@ async def test_nft_with_did_wallet_creation(two_wallet_nodes: Any, trusted: Any)
...
-    while time_left > 0:
-        coins_response = await api_0.nft_get_nfts(dict(wallet_id=nft_wallet_0_id))
-        if coins_response.get("nft_list"):
-            break
-        await asyncio.sleep(0.5)
-        time_left -= 0.5
-    assert coins_response["nft_list"], isinstance(coins_response, dict)
-    assert coins_response.get("success")
+    coins_response = await wait_rpc_state_condition(
+        5, api_0.nft_get_nfts, [dict(wallet_id=nft_wallet_0_id)], lambda x: x["nft_list"]
```
altendky, line 674: I'm guessing we mean something more like `len(x["nft_list"]) > 0`?  We probably don't want to succeed in cases where `x["nft_list"]` is `object()` or `True` or `7`.

altendky (commented):
Everything new should be hinted.  The whole file should be required to be hinted but...  that's a separate issue.

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -33,13 +33,43 @@ async def tx_in_pool(mempool: MempoolManager, tx_id: bytes32) -> bool:
     return True
 
 
+def assert_rpc_response(response, condition_func=None) -> Dict:
+    assert isinstance(response, dict)
+    assert response.get("success")
+    if condition_func:
+        assert condition_func(response)
```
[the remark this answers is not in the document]
trepca: the value is to wait for a condition function to be true or timeout. Not sure what you mean by separate asserts

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -33,13 +33,43 @@ async def tx_in_pool(mempool: MempoolManager, tx_id: bytes32) -> bool:
...
+    return response
```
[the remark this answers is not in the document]
trepca: to extract data from a response that passes specific conditions (or add extra custom asserts in the test)

tests/wallet/nft_wallet/test_nft_wallet.py
trepca: true, thanks

tests/wallet/nft_wallet/test_nft_wallet.py
trepca: pass by name?

altendky (approved):
thanks!

PR #11981, opened 2022-06-20
not optional - `WalletNode.wallet_state_manager`, `.server`, `.keychain_proxy`

altendky:
This makes the attributes accessible via properties that raise if the attribute is `None` instead of making every single call site decide and code how to handle `None`ness.  This is a stop gap until the class and it's creation are refactored to simply have the attributes be known prior to creating the instance.

Draft for:
- [x] Check exception types between raise sites and catch sites


Quexington (dismissed):
Makes sense to me.

mariano54 (dismissed):

mariano54 (approved):

trepca (approved):

PR #11995, opened 2022-06-21
Process coverage in Windows

altendky:


wjblanke (approved):
aok

altendky:
It runs, and doesn't fail...  but does list duplicates dependent on slash direction of the original platform on which the coverage data was recorded and is also flagrantly missing coverage of some clearly run files.

PR #12032, opened 2022-06-23
Fix pending roots.

fchirica:


chia/data_layer/data_store.py
```diff
@@ -98,7 +98,6 @@ async def create(cls, db_wrapper: DBWrapper) -> "DataStore":
                     tree_id TEXT NOT NULL,
                     generation INTEGER NOT NULL,
                     PRIMARY KEY(hash, tree_id, generation),
-                    FOREIGN KEY(tree_id, generation) REFERENCES root(tree_id, generation),
```
fchirica, line 101: I removed this since the primary key changed, and we'd need a `status` field here as well, which would always have to be equal to `Status.COMMITTED` (since `PENDING` roots never write to ancestor table anymore). I didn't find any way to reference constants as foreign keys (for example `(2, tree_id, generation) REFERENCES root(status, tree_id, generation)`), so our only option would be adding status column and always setting it to `2` constant.

tests/core/data_layer/test_data_store.py
```diff
@@ -1165,3 +1203,35 @@ async def test_data_server_files(data_store: DataStore, tree_id: bytes32, test_d
...
+        tree_id=tree_id,
+        status=Status.PENDING,
+    )
+    pending_root = await data_store.get_pending_root(tree_id=tree_id)
+    assert pending_root is not None
+    assert pending_root.generation == 2 and pending_root.status == Status.PENDING
+
+    await data_store.clear_pending_roots(tree_id=tree_id)
+    pending_root = await data_store.get_pending_root(tree_id=tree_id)
+    assert pending_root is None
```
altendky, line 1237: Could assert we're back at the expected not-pending root instead of having accidentally confirmed it or such.

altendky (dismissed):
Merging now for wider spread testing etc.  Comments will be revisited later.

altendky (approved):

PR #12037, opened 2022-06-23
fix for derivation path generation hogging CPU for NFT wallet 

trepca:
NFT wallets  were returning same puzzles for derivation path index and were cancelling each other, so only last wallet had values in the db, others were being regenerated over and over again 

NFT wallet doesn't require derivation paths so added support in wallet state manager to avoid it


Quexington (approved):
Seems a little weird to remove those functions but it looks like they were broken anyway.

paninaro (approved):
lgtm

emlowe (approved):

PR #12052, opened 2022-06-24
remove unused WalletCoinStore.get_unspent_coins_at_height

arvidn:
a coin where `spent=0` is not necessarily "unspent at height" if `confirmed_block_height` > height. i.e. if it didn't exist yet at the specified height, it should not be included in the returned set.

The function does not appear to be used anywhere, so I removed it.

### DIFF
--- a/chia/wallet/wallet_coin_store.py
+++ b/chia/wallet/wallet_coin_store.py
@@ -206,9 +206,8 @@ class WalletCoinStore:
         else:
             all_unspent = set()
             for name, coin_record in self.coin_record_cache.items():
-                if (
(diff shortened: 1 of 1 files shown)


mariano54 (commented):
How could we have a coin that is non existing? Wouldnt rollback delete it?

arvidn:
> How could we have a coin that is non existing? Wouldnt rollback delete it?

By calling the function with a height that's a lot older than some of the coins in the wallet

mariano54 (approved):

wjblanke (approved):
aok

Issue #11955, opened 2022-06-17
Decreasing coin add log line from info to debug breaks third party monitors such as Chiadog.


guydavis:
Please see this [line](https://github.com/Chia-Network/chia-blockchain/blob/c6f60e791b1d32fac47a0e6d124f10b459aca92e/chia/wallet/wallet_state_manager.py#L1159) of logging.  Here's the [commit](https://github.com/Chia-Network/chia-blockchain/pull/11886/files#diff-a367f650aed3b60eba2c253ce15852630c1baf5e9891223d1557ac9cb3333d20R1118) that made this change.

![20220617_105120](https://user-images.githubusercontent.com/12936/174343514-aaa964f3-0c55-467f-bbbc-a79ad1ea50d9.png)

Log monitors such as [Chiadog](https://github.com/martomi/chiadog) look for this log line in the Chia log at `info` level.  By moving this message to `debug` level, Chiadog no longer can report on added coins as in previous versions of Chia. 

Would it be possible to get this log line back to `info` level again?  Thanks!

mariano54:
Sure, can you make a PR for it?

PR #12056, opened 2022-06-24
Log added coins at info for monitors like Chiadog.

guydavis:
Addresses #11955

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1199,7 +1199,7 @@ class WalletStateManager:
             if existing is not None:
                 return None
 
-        self.log.debug(
+        self.log.info(
             "Adding record to state manager coin: %s at %s wallet_id: %s and type: %s",
             coin,
             height,


hoffmang9:
This may have serious performance impacts for dusted wallets.

mariano54:
> This may have serious performance impacts for dusted wallets.

Good point I had forgotten about this. IIRC it would add around 10% to the syncing time

Chardmw:
> > This may have serious performance impacts for dusted wallets.
> 
> Good point I had forgotten about this. IIRC it would add around 10% to the syncing time

Can it be set to change the "added coins" log level to info only when the node is in sync?

wjblanke (dismissed):
aok

wallentx (commented):
test

chia/wallet/wallet_state_manager.py
```diff
@@ -1199,7 +1199,7 @@ async def coin_added(
             if existing is not None:
                 return None
 
-        self.log.debug(
+        self.log.info(
```
wallentx, line 1202: Conversation for performance review

wallentx: Just trying to trigger the GitHub Conversation feature

trepca:
A more long term solution might be to add support in Chia Dog to parse `DEBUG` log level entries too. That way it doesn't affect it if this repo changes levels / visibility of logging.

PR #12060, opened 2022-06-24
Go to untrusted peers for coin state if trusted isn't synced

Quexington:
Currently, if you have a trusted node, but it is not synced, then `get_coin_state` still tries to go to the local node rather than its synced untrusted peers.  This small fix should change that behavior.


mariano54:
Good catch! Does the same thing apply for request puzzle solution?

We should get this into main too

mariano54 (approved):

trepca (approved):

PR #12062, opened 2022-06-24
New RPC block_spends - Get spends for block using transaction generator

freddiecoleman:
Currently if you want to get all the spends for a block you have two options:

- Call the `get_additions_and_removals` RPC followed by potentially thousands of calls to `get_puzzle_and_solution` (one for each removal)
- Call the `get_block` endpoint and run the transaction generator. This is a bit better but can be slow depending on the environment in which your CLVM implementation is running.

By adding this RPC we can benefit from the speed of the Rust CLVM implementation and it also means less code for anybody who needs this data.


chia/rpc/full_node_rpc_client.py
```diff
@@ -200,6 +200,16 @@ async def get_block_records(self, start: int, end: int) -> List:
...
         return response["block_records"]
 
+    async def get_block_spends(self, header_hash: bytes32) -> Optional[List[CoinSpend]]:
+        try:
+            response = await self.fetch("get_block_spends", {"header_hash": header_hash.hex()})
+            block_spends = []
+            for block_spend in response["block_spends"]:
+                block_spends.append(CoinSpend.from_json_dict(block_spend))
+            return block_spends
+        except Exception:
```
trepca, line 210: how does the caller know what went wrong here? we don't log any exceptions, just catch them and return None

jack60612: Its what every other api call does.

mariano54: I don't think we have to catch anything. The RPC server will catch it and return the error string to the caller. In this case we don't know what failed

mariano54 (commented):
Has the issue with the empty puzzle been solved?

jack60612:
> Has the issue with the empty puzzle been solved?

No, we are still waiting

wjblanke (dismissed):
aok

jack60612 (dismissed):
Still not working, clvm compression issues still exist. DO NOT MERGE

chia/rpc/full_node_rpc_api.py
```diff
@@ -399,6 +404,32 @@ async def get_block_records(self, request: Dict) -> EndpointResult:
...
+        if full_block is None or full_block.transactions_generator is None:
+            raise ValueError(f"Block {header_hash.hex()} not found or invalid block generator")
+
+        spends: List[CoinSpend] = []
+        block_generator = await self.service.blockchain.get_block_generator(full_block)
+        if block_generator is None:
+            return {"block_spends": spends}
+
+        args = create_generator_args(block_generator.generator_refs).first()
+        _, block_result = block_generator.program.run_with_cost(self.service.constants.MAX_BLOCK_COST_CLVM, 0, args)
```
arvidn, line 421: this has to be run through the generator "rom" program. see here: https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/mempool_check_conditions.py#L55

jack60612: Thanks

jack60612: Made a custom program as that program did not output what i needed, please take a look,

jack60612 (commented):
.

jack60612:
This is done so please review

mariano54 (approved):

trepca (approved):

PR #12063, opened 2022-06-24
Convert DID Wallet to use the new coin selection algorithm that the normal wallet and the CAT wallet already use

jack60612:
Also make a simple type change in wallet.py


altendky (commented):
The hinting/type corrections seem good, though they are an example of how it is annoying to have the limits applied at every internal step rather than just at boundaries (db, mempool, offer files, etc).  The rest ~reds~ reads sensibly to me but I'm not going to approve as I think that someone involved with this wallet should review it.  Sorry for the wait.

jack60612:
> The hinting/type corrections seem good, though they are an example of how it is annoying to have the limits applied at every internal step rather than just at boundaries (db, mempool, offer files, etc).  The rest reds sensibly to me but I'm not going to approve as I think that someone involved with this wallet should review it.  Sorry for the wait.

No worries, thanks for your review.

wjblanke (approved):
aok

PR #12072, opened 2022-06-25
Replace existing simulator config & Fix simulator

jack60612:
Remove old simulator config from config file, and add new config to the file while adding backwards compatibility support and allowing simulator to be more easily customized
Changes were also made to tests that might decrease flakiness 
Selected_network (Implemented)
Introducer_peer (implemented)
Key_fingerprint: (Implemented)
Auto_farm: on/off (Implemented) & working well
Plot_directory (implemented)
Farming_address ( Implemented)

Tests were added


chia/util/initial-config.yaml
```diff
@@ -550,3 +546,27 @@ wallet:
 
   # Interval to resend unconfirmed transactions, even if previously accepted into Mempool
   tx_resend_timeout_secs: 1800
+
+simulator:
+  # The full node simulator (if run) will run on this port
+  port: 8444
+
```
mariano54, line 553: If the config variables are already in the full node section, there is no need to repeat them here

mariano54 (changes requested):

chia/util/initial-config.yaml
jack60612: These will be changed by the wizard later, and they allow the existing simulator to continue working

jack60612: The existing simulator which is designed to work with the gui expects the simulator to be at the default ports, so we will have the wizard change the ports to the testnet ports once it is run.

jack60612:
I will remove draft once i have a chance to test it

chia/util/initial-config.yaml
mariano54: can we have the GUI read the config instead? I think the way it works with existing GUI simulator is kind of bad because you can only run one at a time

jack60612:
All tested in working, this is part 1 of the new simulator

chia/util/initial-config.yaml
jack60612: I removed the ports from it, should be good now

chia/simulator/SimulatorFullNodeRpcApi.py
```diff
@@ -9,12 +9,30 @@
...
     def get_routes(self) -> Dict[str, Endpoint]:
         routes = super().get_routes()
-        routes["/farm_tx_block"] = self.farm_tx_block
+        routes["/farm_block"] = self.farm_block
+        routes["/auto_farm"] = self.auto_farm
         return routes
 
-    async def farm_tx_block(self, _request: Dict[str, object]) -> Dict[str, object]:
+    async def farm_block(self, _request: Dict[str, object]) -> Dict[str, object]:
         request_address = str(_request["address"])
```
mariano54, line 17: I would instead provide an optional parameter: `guarantee_tx_block` which defaults to False. Also, FullBlock can be a tx block. We can never guarantee a non-tx block.  So if `guarantee_tx_block` is true, then we force a tx block to be created. Otherwise, we just create a block but we don't know whether it is tx or not.

chia/util/initial-config.yaml
```diff
@@ -550,3 +546,27 @@ wallet:
 
   # Interval to resend unconfirmed transactions, even if previously accepted into Mempool
   tx_resend_timeout_secs: 1800
+
+simulator:
+
+  selected_network: testnet0
+
+  # The Introducer is set to connect locally because the simulator is not on Mainnet
```
mariano54, line 554: these variables are already present in full node config. We should remove it from here.

chia/util/initial-config.yaml
```diff
@@ -550,3 +546,27 @@ wallet:
 
   # Interval to resend unconfirmed transactions, even if previously accepted into Mempool
   tx_resend_timeout_secs: 1800
+
+simulator:
+
+  selected_network: testnet0
```
mariano54, line 552: this variable is already present for full node, so remove

chia/simulator/full_node_simulator.py
```diff
@@ -1,26 +1,35 @@
...
-        self.config = full_node.config
+        self.config = config
         self.time_per_block = None
-        if "simulation" in self.config and self.config["simulation"] is True:
+        self.full_node.simulator_transaction_callback = self.autofarm_transaction
+        if "simulation" in self.config["full_node"] and self.config["full_node"]["simulation"] is True:
             self.use_current_time = True
         else:
             self.use_current_time = False
+        if "auto_farm" in self.config["simulator"] and self.config["simulator"]["auto_farm"] is True:
```
mariano54, line 29: you can do: 
```
auto_farm = self.config["simulator"].get("auto_farm", False)
```

mariano54: that's a default value in case it's not present. Same for config["simulator"]

chia/simulator/full_node_simulator.py
```diff
@@ -43,6 +52,26 @@ async def get_all_full_blocks(self) -> List[FullBlock]:
         blocks.reverse()
         return blocks
 
+    async def autofarm_transaction(self, spend_name: bytes32) -> None:
+        if self.auto_farm:
+            self.log.info(f"Autofarm triggered by tx-id: {spend_name.hex()}")
+            new_block = FarmNewBlockProtocol(self.bt.farmer_ph)
+            await self.farm_new_transaction_block(new_block)
+
+    async def update_autofarm_config(self, enable_autofarm: Optional[bool] = None) -> bool:
```
mariano54, line 61: we should not pass in Optional, it's confusing. Better to pass in a bool, and check if None before calling this method

mariano54 (changes requested):
Can do the following to reduce plotting time:
* reduce k size
* reduce plot filter
* reduce number of plots to 2 or 3

The BlockTools changes are quite extensive...  but I guess they are necessary to use real plots and keys

tests/block_tools.py
```diff
@@ -1394,8 +1421,10 @@ def get_challenges(
     return cc_challenge, rc_challenge
 
 
-def get_plot_dir() -> Path:
-    cache_path = DEFAULT_ROOT_PATH.parent.joinpath("test-plots")
+def get_plot_dir(plot_dir: Optional[str] = None) -> Path:
```
mariano54, line 1424: Should be Plot_dir: str="test-plots"

tests/block_tools.py
```diff
@@ -1405,8 +1434,8 @@ def get_plot_dir() -> Path:
     return cache_path
 
 
-def get_plot_tmp_dir():
-    return get_plot_dir() / "tmp"
+def get_plot_tmp_dir(plot_dir: Optional[str] = None):
```
mariano54, line 1437: same here

tests/block_tools.py
```diff
@@ -153,36 +155,49 @@ def __init__(
...
+                "wallet",
+                "introducer",
+                "timelord",
+                "pool",
+                "simulator",
+            ]:
+                self._config[service]["selected_network"] = "testnet0"
+
+            # some tests start the daemon, make sure it's on a free port
+            self._config["daemon_port"] = find_available_listen_port("BlockTools daemon")
```
mariano54, line 187: I think these configs can be passed in using `config_overrides` and this would make the code cleaner

tests/block_tools.py
```diff
@@ -153,36 +155,49 @@ def __init__(
...
 
-        # some tests start the daemon, make sure it's on a free port
-        self._config["daemon_port"] = find_available_listen_port("BlockTools daemon")
         self._config = override_config(self._config, config_overrides)
-
-        with lock_config(self.root_path, "config.yaml"):
-            save_config(self.root_path, "config.yaml", self._config)
+        if automated_testing:
```
mariano54, line 190: shouldn't we also override this for simulator?

jack60612: if we eliminate the extra cmds for the simulator, which we are, then yes

jack60612: so yes

tests/block_tools.py
jack60612: Ill move those too

chia/simulator/full_node_simulator.py
jack60612: Yeah, i forgot about that. Thanks!

chia/util/initial-config.yaml
jack60612: Im removing selected_network and introducer, is that good?

tests/block_tools.py
jack60612: What if plot_dir is None, this wont work

tests/block_tools.py
jack60612: This can not be passed in through config_overrides, due to some of the changes being made on the base level of the config file

jack60612:
> Can do the following to reduce plotting time:
> 
> * reduce k size
> * reduce plot filter
> * reduce number of plots to 2 or 3
> 
> The BlockTools changes are quite extensive... but I guess they are necessary to use real plots and keys

Blockfilter is already set to 1 by the test constants, the k size has been reduced to 19 and only 3 plots are made. This takes about 5-6 seconds and is not really noticeable

chia/util/initial-config.yaml
```diff
@@ -550,3 +546,18 @@ wallet:
...
   # Interval to resend unconfirmed transactions, even if previously accepted into Mempool
   tx_resend_timeout_secs: 1800
+
+simulator:
+
+  # Should the simulator farm a block whenever a transaction is in mempool
+  auto_farm: True
+
+  # The key used by the simulator to farm blocks and send transactions
+  key_fingerprint:
```
mariano54, line 556: Does this actually work, with no value?

tests/block_tools.py
mariano54: Why can root level changes not work?

tests/block_tools.py
```diff
@@ -257,25 +283,35 @@ def change_config(self, new_config: Dict):
...
+            self._config = add_plot_directory(self.root_path, str(path))
+        except ValueError:
+            if not self.automated_testing:  # ignore if this is user run
+                pass
+            else:
+                raise
 
-    async def setup_plots(self):
+    async def setup_plots(self, plots: Optional[int] = None, plot_size: int = 20):
+        if plots is not None:
```
mariano54, line 295: Use the default argument value instead:

`ansyc def setup_plots(self, num_og_plots: int=15, num_pool_plots: int=5 ...)`

mariano54:
Also, we might want to run a test that runs the simulator and requests to create a few blocks, to ensure that it doesn't break. We have no tests for these new API calls for example for autofarming.

chia/util/initial-config.yaml
jack60612: Yes it does, that's why I do the check for none in start_simulator

tests/block_tools.py
jack60612: I did that to try to avoid using another variable, however I'll just use self.automated_testing

jack60612: Done, thanks!

tests/block_tools.py
mariano54: It can't be none, we need test plots right?

jack60612: I meant what if the variable was left as default, which is none

mariano54: The default should be test-plots not None

jack60612: Ok so change it upstream too!

mariano54: Wait actually we already pass in test plots into init, so we can just save it as an instance variable and remove this one right?

jack60612: > Wait actually we already pass in plot dir into init, so we can just save it as an instance variable and remove this one right?

oh yeah lol, thats even better

jack60612: Oh wait, its not part of the block_tools class

chia/simulator/start_simulator.py
```diff
@@ -22,54 +24,102 @@
...
+        server_listen_ports=[config[SERVICE_NAME]["port"]],
         on_connect_callback=node.on_connect,
-        rpc_info=(SimulatorFullNodeRpcApi, config["rpc_port"]),
+        rpc_info=(SimulatorFullNodeRpcApi, config[SERVICE_NAME]["rpc_port"]),
         network_id=network_id,
     )
     return kwargs
 
 
+def create_block_tools_simulator(
```
mariano54, line 54: Having a new function here just adds a bunch of lines that we don't need. We can just create the BlockTools directly within main

tests/block_tools.py
```diff
@@ -257,25 +287,33 @@ def change_config(self, new_config: Dict):
             save_config(self.root_path, "config.yaml", self._config)
 
     def add_plot_directory(self, path: Path) -> None:
-        self._config = add_plot_directory(self.root_path, str(path))
+        try:
+            self._config = add_plot_directory(self.root_path, str(path))
```
mariano54, line 291: Why do we need to catch the value error here?

tests/block_tools.py
```diff
@@ -1394,8 +1433,8 @@ def get_challenges(
     return cc_challenge, rc_challenge
 
 
-def get_plot_dir() -> Path:
-    cache_path = DEFAULT_ROOT_PATH.parent.joinpath("test-plots")
+def get_plot_dir(plot_dir_name: str = "test-plots") -> Path:
```
mariano54, line 1436: We are hardcoding "test-plots" in multiple places, so I recommend not having a default value here

tests/block_tools.py
```diff
@@ -1405,8 +1444,8 @@ def get_plot_dir() -> Path:
     return cache_path
 
 
-def get_plot_tmp_dir():
-    return get_plot_dir() / "tmp"
+def get_plot_tmp_dir(plot_dir_name: str = "test-plots") -> Path:
```
mariano54, line 1447: same here

tests/block_tools.py
```diff
@@ -2057,9 +2096,8 @@ def create_block_tools(
     print(f"  create_block_tools called {create_block_tools_count} times")
     bt = BlockTools(constants, root_path, const_dict, keychain, config_overrides=config_overrides)
 
-    asyncio.get_event_loop().run_until_complete(bt.setup_keys())
-    asyncio.get_event_loop().run_until_complete(bt.setup_plots())
-
+    asyncio.run(bt.setup_keys())
```
mariano54, line 2099: let's revert this, I remember that this caused an issue before, but not exactly sure why.

tests/block_tools.py
jack60612: It will throw an error if the plot directory is already in the config, let me see if i can solve that better

mariano54:
Tests are failing for some reason

jack60612:
> Tests are failing for some reason

i think it was caused by removing the "test-plots" default value.

jack60612:
I added an auto_farming test

jack60612:
ci will fail for a little bit

jack60612:
ill change the user back once the changes are merged

jack60612:
OOPS

chia/simulator/SimulatorFullNodeRpcClient.py
```diff
@@ -0,0 +1,21 @@
...
+class SimulatorFullNodeRpcClient(FullNodeRpcClient):
+    async def farm_block(self, target_ph: bytes32, number_of_blocks: int = 1) -> None:
+        address = encode_puzzle_hash(target_ph, "txch")
+        await self.fetch("farm_block", {"address": address, "blocks": number_of_blocks})
+
+    async def farm_transaction_block(self, target_ph: bytes32, number_of_blocks: int = 1) -> None:
+        address = encode_puzzle_hash(target_ph, "txch")
+        await self.fetch("farm_block", {"address": address, "blocks": number_of_blocks, "guarantee_tx_block": True})
+
+    async def enable_auto_farming(self, enable_auto_farming: bool) -> bool:
```
mariano54, line 15: We should probable rename to `set_auto_farming`, since if we turn it off, the word enable doesn't make sense.

chia/simulator/SimulatorFullNodeRpcClient.py
```diff
@@ -0,0 +1,21 @@
...
+        result = (await self.fetch("auto_farm", {"auto_farm": enable_auto_farming}))["auto_farm_enabled"]
+        assert result == enable_auto_farming
+        return bool(result)
+
+    async def get_auto_farming_status(self) -> bool:
```
mariano54, line 20: If so, this one can be `get_auto_farming`.

chia/simulator/SimulatorFullNodeRpcClient.py
```diff
@@ -0,0 +1,21 @@
+from chia.rpc.full_node_rpc_client import FullNodeRpcClient
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.bech32m import encode_puzzle_hash
+
+
+class SimulatorFullNodeRpcClient(FullNodeRpcClient):
+    async def farm_block(self, target_ph: bytes32, number_of_blocks: int = 1) -> None:
```
mariano54, line 7: can we combine both of these into one RPC api?

chia/simulator/SimulatorFullNodeRpcApi.py
```diff
@@ -9,12 +9,28 @@
...
-        await self.service.server.api.farm_new_transaction_block(req)
+        if guarantee_tx_block:
+            for i in range(blocks):  # these can only be tx blocks
+                await self.service.server.api.farm_new_transaction_block(req)
+        else:
+            for i in range(blocks):  # these can either be full blocks or tx blocks
+                await self.service.server.api.farm_new_block(req)
         return {}
+
+    async def auto_farm(self, _request: Dict[str, object]) -> EndpointResult:
```
jack60612, line 30: separate this

tests/simulation/test_start_simulator.py
```diff
@@ -0,0 +1,183 @@
...
+        # connect to rpc
+        rpc_port = config["full_node"]["rpc_port"]
+        simulator_rpc_client = await SimulatorFullNodeRpcClient.create(
+            config["self_hostname"], uint16(rpc_port), root_path, config
+        )
+        # test auto_farm logic
+        assert await simulator_rpc_client.get_auto_farming()
+        await time_out_assert(10, simulator_rpc_client.set_auto_farming, False, False)
+        await simulator.autofarm_transaction(dummy_hash)  # this should do nothing
+        await asyncio.sleep(3)  # wait for block to be processed
```
mariano54, line 166: Sleep is bad and causes flaky tests. You can do a timeout assert to wait for a specific block height

jack60612: It's not supposed to advance, the sleepyness is only for sanity checking. Ex: fail condition

mariano54: got it thanks

mariano54 (dismissed):

mariano54 (approved):

wjblanke (approved):
aok

PR #12084, opened 2022-06-25
Deduplicate pre-commit python lints from the super linter

AmineKhaldi:
This takes this workflow's linting time from ~19 minutes to ~1

### DIFF
--- a/.github/workflows/super-linter.yml
+++ b/.github/workflows/super-linter.yml
@@ -64,9 +64,6 @@ jobs:
           DEFAULT_BRANCH: main
           LINTER_RULES_PATH: .
           MARKDOWN_CONFIG_FILE: .markdown-lint.yml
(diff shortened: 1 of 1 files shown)


trepca (approved):

altendky (changes requested):
I don't think we run pylint anywhere else so this is a removal and not a deduplication it seems.  When run elsewhere pylint ends up complaining about some stuff.

altendky:
Trying to add pylint to the regular lint run over in https://github.com/Chia-Network/chia-blockchain/pull/12100 so that this can in fact be a dedupe PR.

emlowe (approved):

altendky (approved):
Added pylint in https://github.com/Chia-Network/chia-blockchain/pull/12100 so now this is truly a dedupe/tidy pr.  Thanks!

PR #12087, opened 2022-06-25
Fix sql error when only config file is deleted

jack60612:
If the config file is deleted and chia init is then ran, it will throw an error because chia init tries to write the db version to an existing database.


altendky:
This leaves a race condition hazard between the check and write.  Can we just ignore the particular exception?  Though I'm not sure about the additional logic around here.

jack60612:
> This leaves a race condition hazard between the check and write. Can we just ignore the particular exception? Though I'm not sure about the additional logic around here.

Ill just ignore it, thanks!

arvidn (approved):

wjblanke (approved):
aok

PR #12118, opened 2022-06-27
Refine wallet full node peer selection

Quexington:
There are a few methods in the wallet node that are utilities for fetching information as best as possible.  These used to require a peer get specified to ask for this information, but this led to a lot of peer logic in the wallet code.  This PR refines the selection process of peers for these requests to prefer trusted & synced > untrusted & synced > trusted & unsynced > untrusted & unsynced.

PR #12122, opened 2022-06-27
Nft_get_info flakiness

Quexington:
This is a hacky fix for nft_get_info that would also be fixed in https://github.com/Chia-Network/chia-blockchain/pull/12118 but that's a larger PR and this may be easier to swallow.


ytx1991 (approved):
lgtm

trepca:
we're merging the other PR

PR #12124, opened 2022-06-27
Fix an accidental regression in the wallet node

Quexington:


wjblanke (approved):
aok

paninaro (approved):
Tested, working for me

PR #12125, opened 2022-06-27
cmds: Can't shadow `info` which is `NFTInfo` in the first place

xdustinface:
`mypy` complains [here](https://github.com/xdustinface/chia-blockchain/blob/7ae530a43088e58c234e1d8757020a1b7f627da0/chia/cmds/wallet_funcs.py#L300) if `NFTInfo.from_json_dict` gets hinted properly like in #11763.


Quexington (approved):
Sure

emlowe (approved):

PR #12004, opened 2022-06-21
Add CLI support for NFT offers

Quexington:

PR #12126, opened 2022-06-27
cmds: `NFTInfo.royalty_puzzle_hash` is `Optional` but not `None` here

xdustinface:
`mypy` complains [here](https://github.com/xdustinface/chia-blockchain/blob/dc805f1e6314a34201ac9270c65660373b86a5bc/chia/cmds/wallet_funcs.py#L323) if `NFTInfo.from_json_dict` gets hinted properly like in #11763.

It should be fine to just assert this here, see https://github.com/Chia-Network/chia-blockchain/pull/12004#discussion_r907604272


Quexington (approved):
Sure

emlowe (approved):

PR #12130, opened 2022-06-28
Wallet coin store cache

arvidn:
This patch removes the `coin_record_cache` and `unspent_coin_wallet_cache` fields from `WalletCoinStore`. Keeping all coins ever received and spent in memory does not scale.

Removing this cache offers some substantial simplifications to the `WalletCoinStore` class. In particular, being able to remove the `rebuild_wallet_cache()` function simplifies a future transition to using `DBWrapper2` in the wallet (which is a goal this patch serves to move towards).

The other benefit of this patch is that it results in using a lot less memory (for large, long-lived or dusted wallets).

The downside, of course, is that we'll have to hit the DB a lot more often instead. Since the DB also has a RAM cache, the main additional cost appears to be in constructing python objects, parsing serialized objects and the aiosqlite thread rount-trip.

To mitigate this, there's a second commit that optimizes some of the aiosqlite interactions, by saving round-trips.

# benchmarks

To benchmark, I synced the toilet wallet on testnet:

`main`:

```
chia.wallet.wallet_node    : INFO     Successfully subscribed and updated 103906 coin ids
chia.wallet.wallet_node    : INFO     Sync (trusted: True) duration was: 174.8691442012787
```

this patch:

```
chia.wallet.wallet_node    : INFO     Successfully subscribed and updated 103906 coin ids
chia.wallet.wallet_node    : INFO     Sync (trusted: True) duration was: 159.03764295578003
```

The benchmark looks better than it is in reality. `main` would probably be sped up by the aiosqlite optimization as well, but I did not run that test. I think it's fair to say that the magnitude of the slow-down caused by removing the cache is small relative to the optimization opportunities in the database interaction.

# profiles

the wallet sync has two phases. The first subscribes to all coins and puzzle hashes, the second phase receives, parses and ingests all updates.

## main

phase 1:
main-phase1

phase 2:
main-phase2

## this patch

phase 1:
patched-phase1

phase 2:
patched-phase2

As these profiles show, Phase 2 in particular, has a lot more DB interactions show up with this patch. I expect a fair amount of that to be shaved off in optimizations to streaming in the future.

### DIFF
--- a/tests/wallet/test_wallet_coin_store.py
+++ b/tests/wallet/test_wallet_coin_store.py
@@ -40,7 +40,7 @@ record_4 = WalletCoinRecord(
 record_5 = WalletCoinRecord(
     coin_5,
     uint32(5),
-    uint32(15),
+    uint32(0),
     False,
     False,
     WalletType.STANDARD_WALLET,
@@ -58,7 +58,7 @@ record_6 = WalletCoinRecord(
 record_7 = WalletCoinRecord(
     coin_7,
     uint32(5),
-    uint32(15),
+    uint32(0),
     False,
     False,
     WalletType.POOLING_WALLET,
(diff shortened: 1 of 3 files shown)


tests/wallet/test_wallet_coin_store.py
```diff
@@ -58,7 +58,7 @@
 record_7 = WalletCoinRecord(
     coin_7,
     uint32(5),
-    uint32(15),
```
arvidn, line 61: these two coins did not make sense. They were confirmed *after* being spent.

mariano54 (dismissed):
It's nice to see all this code getting removed!

trepca (dismissed):

trepca (approved):

PR #12134, opened 2022-06-28
remove wallet transaction store cache

arvidn:
This is part of an effort to simplify the store-classes in the wallet and transition to `DBWrapper2`.

Currently, the DB transaction handling (the error handling in particular) is quite complex. We store an in-memory copy of the database which is re-created from the DB on certain errors. This patch remove the in-memory copy to just rely on the sqlite database.

There's also a second commit to optimize the aiosqlite interaction a bit.

# benchmarks

main:

```
chia.wallet.wallet_node    : INFO     Successfully subscribed and updated 103906 coin ids
chia.wallet.wallet_node    : INFO     Sync (trusted: True) duration was: 159.03764295578003
```

this patch:

```
chia.wallet.wallet_node    : INFO     Successfully subscribed and updated 103906 coin ids
chia.wallet.wallet_node    : INFO     Sync (trusted: True) duration was: 158.9499671459198
```


mariano54:
Can you do a toilet trusted sync and report the speed difference?

mariano54 (approved):

emlowe (approved):
Nice.

PR #12135, opened 2022-06-28
farmer: Handle `KeychainProxyConnectionFailure` in `Farmer.setup_keys`

### LINKED ISSUE #11380: [Bug] Harvester doesn't load plots if farmer starts before daemon RPC is ready (edge case)
### What happened?

_Before I get into the details, I want to make a note that I don't believe this will happen if starting chia services the "common" way via the daemon (`chia start farmer`, etc) because of the fact that the daemon RPC has to be already up when starting that way._ This does apply to a handful of us that are using systemd or similar tools to manage the services separately, vs starting all of them through the daemon.

It appears that if the farmer starts up before the daemon RPC is fully up and ready to accept connections, the harvester never attempts to look up plots. I haven't really been able to figure out why this happens, but starting both daemon + + farmer + harvester at the same time with systemd without any logic to delay dependent services until daemon was fully up nearly always resulted in the harvester thinking it had zero plots to farm, unless I got lucky on timing.

Adding in a check to systemd to wait for the port was "up" before marking daemon as ready solved this problem, but it seems like an edge case that would be nice to solve in the harvester/farmer code eventually as well.

## Steps to Reproduce

You can actually fairly reliably reproduce this by starting each service in a particular order:

1. `start_harvester` and `start_farmer` (don't think the order of starting these matter)
2. `start_daemon` after farmer is up

Observe in the logs that plots are never loaded.

If you then stop and start the farmer, leaving the daemon/harvester up, you'll see that the plots do get loaded.

## My guess on the issue

I think the issue stems from here https://github.com/Chia-Network/chia-blockchain/blob/main/chia/farmer/farmer.py#L181

If the daemon isn't yet up on that, then we Raise here: https://github.com/Chia-Network/chia-blockchain/blob/795f1995976fee6da630bef5140976d21e77fa38/chia/farmer/farmer.py#L128 - I do see that exception in my logs

I don't see any other place this might be retried, so my assumption is farmer then never ends up with keys, and harvester as a result never loads any plots

### Version

1.3.4

Linux

CLI

### Relevant log output

_No response_

xdustinface:
Fixes https://github.com/Chia-Network/chia-blockchain/issues/11380

### DIFF
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -134,7 +134,11 @@ class Farmer:
 
     async def setup_keys(self) -> bool:
         no_keys_error_str = "No keys exist. Please run 'chia keys generate' or open the UI."
-        self.all_root_sks: List[PrivateKey] = [sk for sk, _ in await self.get_all_private_keys()]
+        try:
+            self.all_root_sks: List[PrivateKey] = [sk for sk, _ in await self.get_all_private_keys()]
+        except KeychainProxyConnectionFailure:
+            return False
+
         self._private_keys = [master_sk_to_farmer_sk(sk) for sk in self.all_root_sks] + [
             master_sk_to_pool_sk(sk) for sk in self.all_root_sks
         ]


altendky (approved):
Tested using `while true; sudo systemctl stop chia_\*; sudo systemctl start chia_{farmer,harvester}; sleep 30; sudo systemctl start chia_daemon; sleep 30; venv/bin/chia farm summary; end` (in fish).  5x without this fix and all times the farm size was zero.  5x with this fix and all times the farm reported correct plot counts for both my local and remote harvester.  Thanks!

emlowe (approved):

PR #12153, opened 2022-06-29
rework service creation to be hint checkable

altendky:


chia/server/start_introducer.py
```diff
@@ -14,37 +16,54 @@
...
+    return 0
+
+
+def main() -> int:
+    return async_run(async_main())
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
```
xdustinface, line 69:
```suggestion
    sys.exit(async_run(main()))
```

Why not drop the `def main() -> int` above, rename `async_main` to `main` and do the above in all files?

chia/server/start_service.py
```diff
@@ -271,20 +273,7 @@ async def wait_closed(self) -> None:
...
-        logging_config=load_config(kwargs["root_path"], "config.yaml", kwargs["service_name"])["logging"],
-        root_path=kwargs["root_path"],
-    )
-    service = Service(*args, **kwargs)
-    await service.setup_process_global_state()
-    return await service.run()
-
-
-def run_service(*args, **kwargs) -> None:
+def async_run(coro: Coroutine[object, object, T]) -> T:
```
xdustinface, line 276:
```suggestion
def async_run(coro: Awaitable[int]) -> int:
```

Why not just ^? i.e. why is the `Coroutine` hinting required here and why the `TypeVar`?

chia/server/start_wallet.py
```diff
@@ -55,7 +58,14 @@ def service_kwargs_for_wallet(
...
         node.full_node_peer = None
     network_id = service_config["selected_network"]
-    kwargs = dict(
+    rpc_port = service_config.get("rpc_port")
+    rpc_info: RpcInfo = None
+    if rpc_port is not None:
+        rpc_info = (WalletRpcApi, service_config["rpc_port"])
+
+    return Service(
+        advertised_port=service_config["port"],
```
xdustinface, line 67:
```suggestion
        advertised_port=service_config["port"],
```

nit: You added it at the end in an other place.

chia/server/start_wallet.py
```diff
@@ -66,26 +76,16 @@ def service_kwargs_for_wallet(
...
-    return kwargs
 
 
-def main() -> None:
+async def async_main() -> int:
     # TODO: refactor to avoid the double load
     config = load_config(DEFAULT_ROOT_PATH, "config.yaml")
     service_config = load_config_cli(DEFAULT_ROOT_PATH, "config.yaml", SERVICE_NAME)
     config[SERVICE_NAME] = service_config
-
```
xdustinface, line 88: nit: unrelated

chia/simulator/start_simulator.py
```diff
@@ -1,18 +1,18 @@
...
+from chia.server.start_service import Service, async_run
 from chia.simulator.SimulatorFullNodeRpcApi import SimulatorFullNodeRpcApi
 from chia.util.config import load_config, load_config_cli
 from chia.util.default_root import DEFAULT_ROOT_PATH
 from chia.util.path import path_from_root
-from tests.block_tools import BlockTools, create_block_tools, test_constants
+from tests.block_tools import BlockTools, create_block_tools_async, test_constants
+from chia.util.ints import uint16
 from tests.util.keyring import TempKeyring
-
```
xdustinface, line 15: nit: unrelated

chia/simulator/start_simulator.py
```diff
@@ -46,13 +52,14 @@ def service_kwargs_for_full_node_simulator(root_path: Path, config: Dict, bt: Bl
         service_name=SERVICE_NAME,
         server_listen_ports=[service_config["port"]],
         on_connect_callback=node.on_connect,
-        rpc_info=(SimulatorFullNodeRpcApi, service_config["rpc_port"]),
```
xdustinface, line 49: I assume this just has been moved to match the other files?

xdustinface (changes requested):
Makes sense and looks mostly good but i left some comments.

chia/server/start_service.py
altendky: So that would be ok now, but only because typeshed has fixed the hints since the last released version in mypy, or something.  But they have fixed it and `Awaitable` does not satisfy `Coroutine`.  `asyncio.run()` is about processing a coroutine, not awaiting an awaitable.  I think...

https://github.com/python/typeshed/commit/eab82c838a395ef0323d930b8047d037e18832a5#diff-816a4437e484a5998b1afef4f0e0e18aca85e6399b877f3fd883e30e0b4e2010

chia/server/start_wallet.py
altendky: nit: Who cares since I pass by name instead of position.  `:]`  But seriously, yep, will change.  de7535c9e026526cb13e856c0102ed13c5778186

chia/simulator/start_simulator.py
altendky: Sure, let's go with that.  I don't recall but it makes sense.

chia/simulator/start_simulator.py
altendky: e67efd704fde6432cc6351690bb353dc76beef23

chia/server/start_wallet.py
altendky: e67efd704fde6432cc6351690bb353dc76beef23

chia/server/start_service.py
```diff
@@ -33,6 +31,10 @@
 # signal handlers. We need to ignore signals in the sub processes.
 main_pid: Optional[int] = None
 
+T = TypeVar("T")
+
+RpcInfo = Optional[Tuple[type, int]]
```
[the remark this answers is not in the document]
altendky: Yeah, that does seem nicer, thanks.  8473198a4aee477d733d6a6e79ea4616e8d2a296

altendky (commented):
Thanks!

chia/server/start_introducer.py
xdustinface: >  then the entry points reminded me that we need a sync callable 

Not sure what you mean here, don't we have our sync callable with `async_run` which gets used in the entry point and all the other fun we want to do can just go into the async `main` then or why do we need that one to be sync too?  

```python
async def main() -> int:
    return 0


if __name__ == "__main__":
    sys.exit(async_run(main()))
```

chia/server/start_service.py
xdustinface: Okay i see, and the `TypeVar` is there to make it some generic helper? Just wondered because with the service start fun we will probably not return anything else than `int` right?

chia/seeder/start_crawler.py
```diff
@@ -20,11 +22,13 @@
...
-    root_path: pathlib.Path, config: Dict, consensus_constants: ConsensusConstants
-) -> Dict:
+def create_full_node_crawler_service(
+    root_path: pathlib.Path,
+    config: Dict,
+    consensus_constants: ConsensusConstants,
+    connect_to_daemon: bool = True,
+) -> Service:
     service_config = config[SERVICE_NAME]
-
```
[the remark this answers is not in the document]
altendky: 541f077701722be11c1b7ac2d7542a81dec21237

xdustinface (approved):

emlowe (approved):

PR #12183, opened 2022-07-01
El.dependabot/minimist 1.2.6

emlowe:
Combine the minimist 1.2.5->1.2.6 update into a single PR


altendky (approved):
Reads ok to a not-js dev, but I don't know when these files are 'usually' updated etc.

PR #12204, opened 2022-07-01
Simplify service start configuration loading

altendky:
Draft for:
- [x] Use `config`/`service_config` naming instead of `full_config`/`config`.  Leaving as is until 'the end' since I think the diff is easier to read.
- [x] Local testing
  - [x] Launched each service directly via their entry point file on my test system and observed it seeming to launch without failure.  Did not 'use' the services here.
  - [x] Launched each service directly via my systemd setup on my actual node and observed the whole system seeming to work as expected.

### DIFF
--- a/chia/seeder/start_crawler.py
+++ b/chia/seeder/start_crawler.py
@@ -23,32 +23,32 @@ log = logging.getLogger(__name__)
 def service_kwargs_for_full_node_crawler(
     root_path: pathlib.Path, full_config: Dict, consensus_constants: ConsensusConstants
 ) -> Dict:
-    config = full_config[SERVICE_NAME]
+    service_config = full_config[SERVICE_NAME]
 
     crawler = Crawler(
-        config,
+        service_config,
         root_path=root_path,
         consensus_constants=consensus_constants,
     )
     api = CrawlerAPI(crawler)
 
-    network_id = config["selected_network"]
+    network_id = service_config["selected_network"]
     kwargs = dict(
         root_path=root_path,
         config=full_config,
         node=api.crawler,
         peer_api=api,
         node_type=NodeType.FULL_NODE,
-        advertised_port=config["port"],
+        advertised_port=service_config["port"],
         service_name=SERVICE_NAME,
         upnp_ports=[],
-        server_listen_ports=[config["port"]],
+        server_listen_ports=[service_config["port"]],
         on_connect_callback=crawler.on_connect,
         network_id=network_id,
     )
 
-    if config.get("crawler", {}).get("start_rpc_server", True):
-        kwargs["rpc_info"] = (CrawlerRpcApi, config.get("crawler", {}).get("rpc_port", 8561))
+    if service_config.get("crawler", {}).get("start_rpc_server", True):
+        kwargs["rpc_info"] = (CrawlerRpcApi, service_config.get("crawler", {}).get("rpc_port", 8561))
 
     return kwargs
 
@@ -56,9 +56,9 @@ def service_kwargs_for_full_node_crawler(
 def main():
     # TODO: refactor to avoid the double load
     full_config = load_config(DEFAULT_ROOT_PATH, "config.yaml")
-    config = load_config_cli(DEFAULT_ROOT_PATH, "config.yaml", SERVICE_NAME)
-    full_config[SERVICE_NAME] = config
-    overrides = config["network_overrides"]["constants"][config["selected_network"]]
+    service_config = load_config_cli(DEFAULT_ROOT_PATH, "config.yaml", SERVICE_NAME)
+    full_config[SERVICE_NAME] = service_config
+    overrides = service_config["network_overrides"]["constants"][service_config["selected_network"]]
     updated_constants = DEFAULT_CONSTANTS.replace_str_to_bytes(**overrides)
     kwargs = service_kwargs_for_full_node_crawler(DEFAULT_ROOT_PATH, full_config, updated_constants)
     return run_service(**kwargs)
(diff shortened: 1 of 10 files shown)


chia/seeder/start_crawler.py
```diff
@@ -21,8 +21,10 @@
 
 
 def service_kwargs_for_full_node_crawler(
-    root_path: pathlib.Path, config: Dict, consensus_constants: ConsensusConstants
+    root_path: pathlib.Path, full_config: Dict, consensus_constants: ConsensusConstants
 ) -> Dict:
+    config = full_config[SERVICE_NAME]
```
xdustinface, line 26:
```suggestion
    service_config = config[SERVICE_NAME]
```

chia/seeder/start_crawler.py
```diff
@@ -51,10 +54,13 @@ def service_kwargs_for_full_node_crawler(
 
 
 def main():
+    # TODO: refactor to avoid the double load
+    full_config = load_config(DEFAULT_ROOT_PATH, "config.yaml")
     config = load_config_cli(DEFAULT_ROOT_PATH, "config.yaml", "seeder")
+    full_config[SERVICE_NAME] = config
```
xdustinface, line 60:
```suggestion
    # TODO: refactor to avoid the double load
    config = load_config(DEFAULT_ROOT_PATH, "config.yaml")
    service_config = load_config_cli(DEFAULT_ROOT_PATH, "config.yaml", "seeder")
    config[SERVICE_NAME] = service_config
```

xdustinface (commented):
Makes sense but maybe do the naming rather like proposed below? Of course adjust all other places too, i just highlighted those two to show what i mean :)

chia/seeder/start_crawler.py
altendky: Added as a `Draft for:` reason.

chia/seeder/start_crawler.py
altendky: Added as a `Draft for:` reason.

mariano54 (dismissed):
Why are we doing a double load? It seems simple to just do:
```python
service_config = config.get("full_node", {})
```

altendky:
I am assuming the question is about the several cases of code like this.

```python
    # TODO: refactor to avoid the double load
    config = load_config(DEFAULT_ROOT_PATH, "config.yaml")
    service_config = load_config_cli(DEFAULT_ROOT_PATH, "config.yaml", SERVICE_NAME)
```

My intent was to retain previous the behavior where the cli parameters are applied to the service config.  I think the application of the cli parameters should be adjusted to be applicable to the loaded dict instead of exclusively being available in combination with loading the config but I decided to leave that for another potential follow up PR.  I'm already a couple layers deep in granular cleanup here relative to my overall goal.

There were already multiple loads being done of the config.  1) the service config at the location mentioned above, 2) the whole config in `Service.__init__()`, and 3) the service config again in `Service.__init__()`.  So at least this shifts the loading to a single location and reduces one of the duplicate loads.  For what it's worth, https://github.com/Chia-Network/chia-blockchain/pull/11522 is kind of the 'top level' of this 'group' of work (for me).  There will be a couple more cleanups at this level before I get back to that one.

mariano54:
Are we using any CLI parameters anyway? This might be outdated code that is no longer used.
For example if you do `chia start farmer --rpc-port=1234` will it actually pass the CLI params through?

altendky:
I hadn't looked but I had assumed this was for manual testing and was only used when running the `chia_farmer` etc entry points directly.  But, it looks like it is your code.  https://github.com/Chia-Network/chia-blockchain/blame/95803f09feafd8e71ff5c7dc3f9563593dde519b/chia/util/config.py#L136-L158  If you want it gone then we can check around and make sure nobody else seems to be using it.  Maybe removing that could be a separate PR though?  In so much as it could fit here, it could trivially be another PR to separate the refactor from a feature removal.

chia/seeder/start_crawler.py
```diff
@@ -23,36 +23,42 @@
...
 
     return kwargs
 
 
 def main():
-    config = load_config_cli(DEFAULT_ROOT_PATH, "config.yaml", SERVICE_NAME)
-    overrides = config["network_overrides"]["constants"][config["selected_network"]]
+    # TODO: refactor to avoid the double load
+    config = load_config(DEFAULT_ROOT_PATH, "config.yaml")
+    service_config = load_config_cli(DEFAULT_ROOT_PATH, "config.yaml", SERVICE_NAME)
```
xdustinface, line 59:
```suggestion
    config = load_config_cli(DEFAULT_ROOT_PATH, "config.yaml")
    service_config = config[SERVICE_NAME]
```

Couldn't we just do ^ and then get rid of `load_config_cli` in a separete PR if its outdated and replace it here with `load_config`?

xdustinface (approved):
Looks good but see the one "maybe" :)

altendky:
As explained, this PR is changing from _3_ config loads to 2.

chia/seeder/start_crawler.py
altendky: This would change the behavior and the cli interface.  It would require the service name be included in the cli arguments and would apply the override behavior to the entire configuration.

mariano54:
Thanks for explaining. I was concerned about doing more loads but that is not the case. We can clean up the old code in another PR then.

mariano54:
Can we merge this soon? It conflicts with https://github.com/Chia-Network/chia-blockchain/pull/12072 which we want to merge soon also.

wjblanke (approved):
aok

PR #12209, opened 2022-07-02
use a variable in the workflow for blocks and plots version

altendky:


jack60612:
https://github.com/Chia-Network/chia-blockchain/blob/db04a90eefbcbadb605b74475a2a3be95e4bbfa5/.github/workflows/benchmarks.yml

altendky:
We don't really presently have tooling to share across workflows, but we can at least use the same form.  Thanks for catching this.  https://github.com/Chia-Network/chia-blockchain/pull/12086 is one exploration I did into reusable bits.  We could also potentially make a local action inside this repo that could be used in all the cases with a hard coded or default value for the version.

jack60612 (approved):
lgtm

wjblanke (approved):
aok

wallentx (approved):
👍

PR #12228, opened 2022-07-04
Remove unused WalletActionStore

nuyawktuah:
`WalletActionStore` seems unused, though I may be missing something. This pr removes it and some associated methods.


olivernyc:
Hmm, not sure what to make of the failing tests. I don't see how they are related to the changes. Is it possible they are just flaky?

mariano54:
Test looks green to me. 
The action store was used for CATs, but looks like we are not using it anymore. It should be fine to remove.

olivernyc:
Oh weird, there were 3 failures earlier but looks like they passed on the second attempt. Here's one https://github.com/Chia-Network/chia-blockchain/runs/7183574313?check_suite_focus=true

altendky:
I thought at some point @Quexington caught me referencing the action store and stopped me, but I do see references in the data layer wallet.  There's no urgency to this removal, right?  I just got data layer caught up with 1.4.0 and will be starting on catching up with `main` this week after a little testing of the existing merge work.  Maybe we plan to revisit after that?  Unless quex has something else to say about it.

@olivernyc, yeah, we do have some flaky tests we still need to clean up.

For one concrete case at least...
https://github.com/Chia-Network/chia-blockchain/blob/ab9a04b84bba96157bed61b63013dd3d10fa9840/chia/data_layer/data_layer_wallet.py#L216-L228

mariano54:
It's not high priority. We can close this and come back to it once datalayer doesn't use it anymore.

mariano54:
Closing as it conflicts with data layer

PR #12234, opened 2022-07-05
Wallet dbwrapper 2

arvidn:
This switches the wallet from using `DBWrapper` to `DBWrapper2`. The main benefits are:

* `DBWrapper2` uses nested transactions and the outer-most always commits. There is no longer any need for functions to be told (and pass on) whether they are in a transaction or not, every level just starts a new one, that may be nested. For this reason, all `in_transaction` parameters have been removed.
* `DBWrapper2` supports concurrent readers, but still serializes all writes over a single `Connection` object. This can improve latency for read-only tasks. It also required some minor changes to tests and the `wallet_node` to properly synchronize tasks that read with tasks that write in some cases.
* `DBWrapper2` has a context manager-based API, which means transactions will always either committed or aborted (on exceptions). There's no longer any manual committing and rolling-back necessary. This simplifies a number of functions that previously had to explicitly handle exceptions.

### DIFF
--- a/chia/util/initial-config.yaml
+++ b/chia/util/initial-config.yaml
@@ -483,6 +483,11 @@ wallet:
   # see description for full_node.db_sync
   db_sync: auto
 
+  # the number of threads used to read from the wallet database
+  # concurrently. There's always only 1 writer, but the number of readers is
+  # configurable
+  db_readers: 2
+
   connect_to_unknown_peers: True
 
   # The minimum height that we care about for our transactions. Set to zero
--- a/chia/wallet/cat_wallet/lineage_store.py
+++ b/chia/wallet/cat_wallet/lineage_store.py
@@ -1,10 +1,8 @@
 import logging
 from typing import Dict, Optional
 
-import aiosqlite
-
 from chia.types.blockchain_format.sized_bytes import bytes32
-from chia.util.db_wrapper import DBWrapper
+from chia.util.db_wrapper import DBWrapper2
 from chia.wallet.lineage_proof import LineageProof
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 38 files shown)


chia/util/initial-config.yaml
```diff
@@ -483,6 +483,11 @@ wallet:
   # see description for full_node.db_sync
   db_sync: auto
 
+  # the number of threads used to read from the wallet database
+  # concurrently. There's always only 1 writer, but the number of readers is
+  # configurable
+  db_readers: 2
```
arvidn, line 489: I don't think the wallet has as much pressure from tasks wanting to read from its DB, so two reader threads seems sufficient.

chia/wallet/cat_wallet/lineage_store.py
```diff
@@ -17,84 +15,61 @@ class CATLineageStore:
...
-        try:
-            await self.db_connection.execute(
+        async with self.db_wrapper.write_db() as conn:
+            await conn.execute(
                 (f"CREATE TABLE IF NOT EXISTS {self.table_name}(" " coin_id text PRIMARY KEY," " lineage blob)")
             )
-        finally:
-            if not in_transaction:
-                await self.db_connection.commit()
-                self.db_wrapper.lock.release()
```
arvidn, line 39: this pattern of manually commit and acquire the db mutex is removed everywhere. `DBWrapper2` handles committing and locking of the DB under the hood.

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -813,6 +813,11 @@ async def test_did_endpoints(wallet_rpc_environment: WalletRpcTestEnvironment):
     for _ in range(3):
         await farm_transaction_block(full_node_api, wallet_1_node)
 
+    async def num_wallets() -> int:
+        return len(await wallet_2_node.wallet_state_manager.get_all_wallet_info_entries())
+
+    await time_out_assert(30, num_wallets, 2)
```
arvidn, line 819: this fixed an issue caused by supporting concurrent readers. Without this, the transaction won't have been committed by the time we call `get_all_wallet_info_entries()`, and we won't see the new wallet.

chia/wallet/wallet_node.py
```diff
@@ -498,27 +498,30 @@ async def on_connect(self, peer: WSChiaConnection):
...
             else:
                 await self.wallet_state_manager.blockchain.clean_block_records()
 
                 for wallet_id in removed_wallet_ids:
                     self.wallet_state_manager.wallets.pop(wallet_id)
 
+        # this has to be called *after* the transaction commits, otherwise it
+        # won't see the changes (since we spawn a new task to handle potential
+        # resends)
+        self._pending_tx_handler()
```
arvidn, line 523: this change is contained in the last commit. If fixes the issue described in this comment. Note taht `wallet_state_manager` does not control when its changes are committed to the database, since we're wrapping it in an outer transaction here.

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -138,7 +138,11 @@ async def test_nft_wallet_creation_automatically(two_wallet_nodes: Any, trusted:
     )
     for i in range(1, num_blocks):
         await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph1))
-    await time_out_assert(15, len, 2, wallet_node_1.wallet_state_manager.wallets)
+
+    async def num_wallets() -> int:
+        return len(await wallet_node_1.wallet_state_manager.get_all_wallet_info_entries())
+
+    await time_out_assert(15, num_wallets, 2)
```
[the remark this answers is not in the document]
trepca: nice, thanks

arvidn:
In its current form, this patch slows down syncing.

main:
```
INFO     Successfully subscribed and updated 103906 coin ids
INFO     Sync (trusted: True) duration was: 164.8119740486145
```

this patch:
```
INFO     Successfully subscribed and updated 103906 coin ids
INFO     Sync (trusted: True) duration was: 214.31865620613098
```

arvidn:
The reason this patch causes a slow down is the additional cost of starting and releasing nested transaction. in `main` we don't nest transactions, we just have one at the top level.

If I patch `DBWrapper2` to do the same thing, I get this result:

```
INFO     Successfully subscribed and updated 103906 coin ids
INFO     Sync (trusted: True) duration was: 166.06344199180603
```

I will propose a separate patch for this. It requires some design discussion since such API can more easily be misused.

mariano54 (approved):
This is a huge improvement to how the wallet deals with DB operations. I think it will help a lot in the future. 

We do need to test this patch manually to make sure the wallet is behaving properly, since it's a large change. Also, we might want to run the CI a few times to ensure that there are no flaky tests yet.

wjblanke (approved):
aok

PR #12237, opened 2022-07-05
[PoC] Singleton spend aggregation using mempool's replace bundle feature 

trepca:
This is a proof of concept for singleton spend aggregation based on replace bundle feature in mempool that allows bundles to be replaced given the new bundle contains at least all spends in previous bundle.
This is utilised by `aggregator` puzzle that's creating a `lock` coin for each solution and asserting that it's being spent in the same bundle. Lock coin ensures order of solutions is kept and that they're run. 

Appreciate any feedback on how this concept can break, thanks!


mariano54:
This is a very interesting program, and some chialisp devs might benefit from using this.

That said, I don't think it belongs in chia-blockchain, since it's just an example chialisp application, and doesn't have any actual plaform changes. 

Maybe we should put this into an examples repo.

chia/wallet/puzzles/aggregator.clvm
```diff
@@ -0,0 +1,133 @@
...
+                (x "bad inner puzzle")
+            )
+            ; return back create coin that will be used to curry new puzzle hash for aggregator
+            (c create_coin_condition conditions)
+        )
+    )
+
+    (defun morph-conditions (conditions singleton_amount odd_create_coin morphed_conditions)
+        (if conditions
+            (if (all (= (f (f conditions)) CREATE_COIN)); (= (f (r (r (f conditions)))) singleton_amount) (not odd_create_coin))
```
Quexington, line 87: For conditions like announcements that only have one parameter, this will raise a `path into atom` error because of the eager evaluation (announcments don't have a `(f (r (r`)

chia/wallet/puzzles/aggregator.clvm
```diff
@@ -0,0 +1,133 @@
...
+                ; it's a create odd coin, add it to main params
+                (morph-conditions
+                    (r conditions)
+                    singleton_amount
+                    (f conditions)
+                    morphed_conditions
+                )
+;                (if (all (= (f conditions) CREATE_PUZZLE_ANNOUNCEMENT) (= (substr (f (r (conditions))) 0 3) POSITION_ANN_PREFIX))
```
Quexington, line 95: `POSITION_ANN_PREFIX` is not defined

chia/wallet/puzzles/aggregator.clvm
```diff
@@ -0,0 +1,133 @@
...
+                )
+                ;(list (morph-conditions (a (f (f solutions)) (f (r (f solutions)))) AMOUNT create_coin_condition conditions)); solutions conditions)
+                (x "bad inner puzzle")
+            )
+            ; return back create coin that will be used to curry new puzzle hash for aggregator
+            (c create_coin_condition conditions)
+        )
+    )
+
+    (defun morph-conditions (conditions singleton_amount odd_create_coin morphed_conditions)
```
Quexington, line 85: This function doesn't seem like it's morphing anything.  It seems like it's checking for one odd create coin which, if this is inside a singleton is already covered.  I think it can just be simplified to checking for the malicious announcements and raising if it sees one.

Quexington: Oh I see, we need it to continously process the new puzzlehash, nvm

chia/wallet/puzzles/aggregator.clvm
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+    (defun get-new-puzzle-hash (mod_hash mod_lock_hash inner_agg_puzzle_hash)
+        (puzzle-hash-of-curried-function
+            mod_hash (sha256tree1 inner_agg_puzzle_hash) (sha256tree1 mod_lock_hash) (sha256tree1 mod_hash)
+        )
+    )
+
+    (defun ensure-order (MOD_HASH MOD_LOCK_HASH solutions position)
+        (if (all solutions (> 10000 position))
```
Quexington, line 29: What is the point of this hardcoded `10000`?

chia/wallet/puzzles/aggregator.clvm
```diff
@@ -0,0 +1,133 @@
...
+        )
+    )
+    (merge-lists
+        (recreate
+            (aggregate-solutions
+                INNER_AGG_PUZZLE_HASH amount solutions () ())
+            MOD_HASH
+            MOD_LOCK_HASH
+        )
+        (ensure-order MOD_HASH MOD_LOCK_HASH solutions 1)
```
Quexington, line 131: I'm not sure I follow exactly what `ensure-order` does.  If I understand correctly, the idea here is to do multiple spends of the same singleton all together in one spend.  There's no intermediary steps on chain before all of the states get processed so why are we creating all of these coins and consuming them?  The solution value `solutions` is malleable by the mempool and that solution value is the determination of the order so the order can be manipulated no matter what, right?

Quexington:
This is a pretty cool idea, but I'm not sure it solves a lot of the problems we face when trying to aggregate singleton spends:

1) This implementation runs a series of _inner_ puzzles only.  Any checking that might be done by layers of the singleton higher than the inner puzzles can be skipped and as long as the final state is compatible with those outer layers.

2) The intermediate states are not available for coins that depend on this singleton.  If you want the singleton to be an oracle, a lot of the coins that are going to be using it are going to depend an announcement from its puzzle hash because they want to depend on a certain state.  If spending the singleton updates that spend, the first and last ones may be fine, but other coins in the sequence will have no chance to assert their own state announcement.

3) Presumably, if you are a wallet/entity that is smart enough to look into the mempool and see this aggregator puzzle and update it accordingly, could you not similarly just spend the latest version of the singleton and save all of this overhead? This puzzle does not solve the tough question of how a mempool can handle a bunch of races to update a single spend.

trepca:
> This is a pretty cool idea, but I'm not sure it solves a lot of the problems we face when trying to aggregate singleton spends:
> 
> 
> 
> 1) This implementation runs a series of _inner_ puzzles only.  Any checking that might be done by layers of the singleton higher than the inner puzzles can be skipped and as long as the final state is compatible with those outer layers.
> 
Not sure what you mean here, that's true if we don't have aggregation either. 

> 
> 
> 2) The intermediate states are not available for coins that depend on this singleton.  If you want the singleton to be an oracle, a lot of the coins that are going to be using it are going to depend an announcement from its puzzle hash because they want to depend on a certain state.  If spending the singleton updates that spend, the first and last ones may be fine, but other coins in the sequence will have no chance to assert their own state announcement.
> 

But this is easily solved by using coin announcements. Other coins attaching to the singleton will have to figure out it's coin id anyway to attach the solution.

> 
> 
> 3) Presumably, if you are a wallet/entity that is smart enough to look into the mempool and see this aggregator puzzle and update it accordingly, could you not similarly just spend the latest version of the singleton and save all of this overhead? This puzzle does not solve the tough question of how a mempool can handle a bunch of races to update a single spend. 

It pushes the race condition to the client as mempool will return double spend.

chia/wallet/puzzles/aggregator.clvm
trepca: It's using replace bundle feature in mempool that requires at least all spends to be in the new bundle as old one. 
So aggregator creates lock coins for each solution with solution index in the list. This coin needs to be spend when solution is added as aggregator asserts announcement from it and lock coin asserts announcement from aggregator. 
The part I haven't added yet is that to replace the bundle it also needs to have higher fee per cost than previous one. This requires that each new solution spends a coin to pay for transaction where it could also assert that lock coin has been spent. This makes it impossible for farmer to mess with the bundle and solutions. 
It's possible to also add agg sig me in the lock coin, maybe that's better.

PR #12242, opened 2022-07-05
Fix Coin selection bug

jack60612:
Fix coin selection case where there is dust but we don't have one individual coin above target, but instead multiple non dusty coins which equate or are more then the target

### DIFF
--- a/chia/wallet/coin_selection.py
+++ b/chia/wallet/coin_selection.py
@@ -7,6 +7,8 @@ from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.ints import uint64, uint128
 from chia.wallet.wallet_coin_record import WalletCoinRecord
 
+max_num_coins = 500
+
 
 async def select_coins(
     spendable_amount: uint128,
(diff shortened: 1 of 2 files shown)


mariano54 (changes requested):
There is a misunderstanding here as to what the actual bug is. The bug is not in the algorithm. Knapsack works perfectly in this case, because it requires the input to be sorted. The big coins will be checked first and will be selected.

This proposed change does not fix it and adds complexity by falling back to a different algorithm, so we can close this.

EDIT: I misunderstood the changes here, apologies. There were actually two bugs. One of them was correctly identified in this PR. The other was a bug in `test_smallest_coin_over_amount`. My Pr here fixes both:
https://github.com/Chia-Network/chia-blockchain/pull/12261

altendky (commented):
I made a first pass at understanding this and didn't find any bugs.  If others that are more familiar don't provide a review I can take a second pass through for a more careful review.

In the mean time, here are a few side notes.

jack60612:
Replaced by https://github.com/Chia-Network/chia-blockchain/pull/12261

PR #12252, opened 2022-07-06
Removing `logging.basicConfig()` from `ChiaServer.__init__()`

altendky:
https://docs.python.org/3.10/library/logging.html#logging.basicConfig
> This function does nothing if the root logger already has handlers configured, unless the keyword argument force is set to True.

Previously this was being run before our logging config but https://github.com/Chia-Network/chia-blockchain/pull/12172 reordered this to come first.  A fix will be made to relocate the logging config prior to the `ChiaServer` instantiation but this shouldn't be called regardless.

### DIFF
--- a/chia/server/server.py
+++ b/chia/server/server.py
@@ -118,7 +118,6 @@ class ChiaServer:
         name: str = None,
     ):
         # Keeps track of all connections to and from this node.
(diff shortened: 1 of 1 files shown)


jack60612 (approved):
well that does not belong there, lgtm

wjblanke (approved):
aok

PR #12261, opened 2022-07-06
Ms.fix coin selection

### LINKED ISSUE #12226: [Bug] Repeated timeout when trying to send XCH
### What happened?

Attempting to send ~2 XCH from one wallet to another on Win10 Chia 1.4 and getting repeated timeouts on the Tx. 
EDIT: Amount accuracy 

Never had trouble with this on past versions. Any suggestions? Thanks.

### Version

1.4

Windows

GUI

```shell
2022-07-02T16:xxxxxx wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia\rpc\rpc_server.py", line 229, in safe_handle
  File "chia\rpc\rpc_server.py", line 220, in ws_api
  File "chia\rpc\wallet_rpc_api.py", line 838, in send_transaction
  File "chia\wallet\wallet.py", line 432, in generate_signed_transaction
  File "chia\wallet\wallet.py", line 315, in _generate_unsigned_transaction
  File "chia\wallet\wallet.py", line 267, in select_coins
  File "chia\wallet\coin_selection.py", line 90, in select_coins
  File "chia\wallet\coin_selection.py", line 119, in select_smallest_coin_over_target
ValueError: Unable to select coins for this transaction. Try sending a smaller amount
```

mariano54:
Two bugs :
1. `test_smallest_coin_over_amount` did not work properly when all coins where smaller than the amount.
2. In the cases where knapsack found a solution but it was too large, we fell back to selecting the "largest coin over the amount" as in issue 1, but instead we should be adding up the largest coins that are < the target amount first, because we might not have any large coins.

It also fixes a performance issue with knapsack, where it keeps searching even if we are selecting more coins than we can actually select.
Old performance with 200k coins:  60 seconds. New: 0.78 seconds.

A bunch of tests are added to test all the functionality.

### DIFF
--- a/chia/wallet/coin_selection.py
+++ b/chia/wallet/coin_selection.py
@@ -100,11 +100,11 @@ async def select_coins(
         return coin_set
     else:
         # if smaller_coin_sum == amount and len(smaller_coins) >= max_num_coins.
-        coin = select_smallest_coin_over_target(amount, valid_spendable_coins)
-        if coin is None:
+        potential_large_coin: Optional[Coin] = select_smallest_coin_over_target(amount, valid_spendable_coins)
+        if potential_large_coin is None:
             raise ValueError("Too many coins are required to make this transaction")
-        log.debug(f"Resorted to selecting smallest coin over target due to dust.: {coin}")
-        return {coin}
+        log.debug(f"Resorted to selecting smallest coin over target due to dust.: {potential_large_coin}")
+        return {potential_large_coin}
 
 
 # These algorithms were based off of the algorithms in:
--- a/chia/wallet/coin_selection.py
+++ b/chia/wallet/coin_selection.py
@@ -85,7 +85,7 @@ async def select_coins(
     elif smaller_coin_sum > amount:
         coin_set: Optional[Set[Coin]] = knapsack_coin_algorithm(smaller_coins, amount, max_coin_amount, max_num_coins)
         log.debug(f"Selected coins from knapsack algorithm: {coin_set}")
-        if coin_set is None or len(coin_set) > max_num_coins:
+        if coin_set is None:
             coin_set = sum_largest_coins(amount, smaller_coins)
             if coin_set is None or len(coin_set) > max_num_coins:
                 greater_coin = select_smallest_coin_over_target(amount, valid_spendable_coins)
@@ -119,7 +119,6 @@ def check_for_exact_match(coin_list: List[Coin], target: uint64) -> Optional[Coi
 # amount of coins smaller than target, followed by a list of all valid spendable coins.
 # Coins must be sorted in descending amount order.
 def select_smallest_coin_over_target(target: uint128, sorted_coin_list: List[Coin]) -> Optional[Coin]:
-    # sorted_coin_list: List[Coin] = list(sorted(coin_list, key=lambda c: c.amount))
     if sorted_coin_list[0].amount < target:
         return None
     for coin in reversed(sorted_coin_list):
--- a/chia/wallet/coin_selection.py
+++ b/chia/wallet/coin_selection.py
@@ -115,7 +115,7 @@ def check_for_exact_match(coin_list: List[Coin], target: uint64) -> Optional[Coi
 
 # amount of coins smaller than target, followed by a list of all valid spendable coins sorted in descending order.
 def select_smallest_coin_over_target(smaller_coin_amount: int, valid_spendable_coin_list: List[Coin]) -> Coin:
-    if smaller_coin_amount >= len(valid_spendable_coin_list):
+    if smaller_coin_amount >= sum([c.amount for c in valid_spendable_coin_list]):
         raise ValueError("Unable to select coins for this transaction. Try sending a smaller amount")
     if smaller_coin_amount > 0:  # in case we only have bigger coins.
         greater_coins = valid_spendable_coin_list[:-smaller_coin_amount]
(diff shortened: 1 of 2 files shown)


emlowe:
`valid_spendable_coins` is already sorted in descending order when passed into `select_smallest_coin_over_target` so I don't see why you are sorting it again. Same with `sum_largest_coins` - the `smaller_coin` list is already sorted in the order you need. The prior implementation logic was actually correct but with confusing parameter names.

jack60612:
> `valid_spendable_coins` is already sorted in descending order when passed into `select_smallest_coin_over_target` so I don't see why you are sorting it again. Same with `sum_largest_coins` - the `smaller_coin` list is already sorted in the order you need. The prior implementation logic was actually correct but with confusing parameter names.

There is a bug, but it is really hard to see, so mariano simplified the logic and fixed the bug.
The increase in time is under a couple of milliseconds and it is worth the really easy to read logic in my opinion

jack60612 (changes requested):
Just some minor test changes

jack60612:
re running ci

mariano54:
> `valid_spendable_coins` is already sorted in descending order when passed into `select_smallest_coin_over_target` so I don't see why you are sorting it again. Same with `sum_largest_coins` - the `smaller_coin` list is already sorted in the order you need. The prior implementation logic was actually correct but with confusing parameter names.

There was a bug when `smaller_coin_amount == len(valid_spendable_coin_list)`. I refactored that method so it's clear what it does now.

The reason I sorted within the method is that it makes the API simpler: it now no longer depends on a specific sorting. I thought that this API is harder to mess up for callers. For reference, sorting 200,000 coins in a worst case ordering takes about 0.4 seconds. This is probably OK when making a transaction, but I can see an argument either way.

tests/wallet/test_coin_selection.py
```diff
@@ -132,6 +139,66 @@ async def test_coin_selection_with_dust(self, a_hash: bytes32) -> None:
...
+                uint128(target_amount),
+            )
+            assert dusty_below_target is not None
+            assert sum([coin.amount for coin in dusty_below_target]) >= target_amount
+            for coin in dusty_below_target:
+                assert coin.amount == 5000
+            assert len(dusty_below_target) <= 500
+
+    @pytest.mark.asyncio
+    async def test_dust_and_one_large_coin(self, a_hash: bytes32) -> None:
```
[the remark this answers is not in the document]
mariano54: Are you sure that it's exactly the same? I'm not sure about that, I rather keep both for safety

mariano54:
@jack60612 @emlowe , take a look at the new changes.
Knapsack perf is fixed, and I reverted the sorting for more performance

chia/wallet/coin_selection.py
```diff
@@ -83,11 +83,9 @@ async def select_coins(
         log.debug(f"Selected closest greater coin: {smallest_coin.name()}")
         return {smallest_coin}
     elif smaller_coin_sum > amount:
-        coin_set: Optional[Set[Coin]] = knapsack_coin_algorithm(smaller_coins, amount, max_coin_amount)
+        coin_set: Optional[Set[Coin]] = knapsack_coin_algorithm(smaller_coins, amount, max_coin_amount, max_num_coins)
         log.debug(f"Selected coins from knapsack algorithm: {coin_set}")
-        if coin_set is None:
-            raise ValueError("Knapsack algorithm failed to find a solution.")
-        if len(coin_set) > max_num_coins:
+        if coin_set is None or len(coin_set) > max_num_coins:
```
jack60612, line 88: `len(coin_set) > max_num_coins` cant happen

chia/wallet/coin_selection.py
```diff
@@ -119,19 +117,22 @@ def check_for_exact_match(coin_list: List[Coin], target: uint64) -> Optional[Coi
 
 
 # amount of coins smaller than target, followed by a list of all valid spendable coins.
-def select_smallest_coin_over_target(target: uint128, coin_list: List[Coin]) -> Optional[Coin]:
-    sorted_coin_list: List[Coin] = list(sorted(coin_list, key=lambda c: c.amount))
-    if sorted_coin_list[-1].amount < target:
+# Coins must be sorted in descending amount order.
+def select_smallest_coin_over_target(target: uint128, sorted_coin_list: List[Coin]) -> Optional[Coin]:
+    # sorted_coin_list: List[Coin] = list(sorted(coin_list, key=lambda c: c.amount))
```
jack60612, line 122: remove comment

jack60612 (dismissed):
lgtm

jack60612 (dismissed):
lgtm

jack60612:
once he fixes the lint issues i still approve it, just merge it even if i dont have the chance to re approve

jack60612 (approved):
lgtm

chia/wallet/coin_selection.py
```diff
@@ -78,28 +78,31 @@ async def select_coins(
...
+                coin_set = {greater_coin}
         return coin_set
     else:
         # if smaller_coin_sum == amount and len(smaller_coins) >= max_num_coins.
-        coin = select_smallest_coin_over_target(len(smaller_coins), valid_spendable_coins)
-        log.debug(f"Resorted to selecting smallest coin over target due to dust.: {coin}")
-        return {coin}
+        potential_large_coin: Optional[Coin] = select_smallest_coin_over_target(amount, valid_spendable_coins)
+        if potential_large_coin is None:
+            raise ValueError("Too many coins are required to make this transaction")
```
emlowe, line 103: In the interest of consistency - should this be the same error text as above

` f"Transaction of {amount} mojo would use more than {max_num_coins} coins. Try sending a smaller amount"`

emlowe (approved):
Approved - minor comment on error string, but not gating

PR #12275, opened 2022-07-06
add new_transaction() to DBWrapper2

arvidn:
which creates a nested transaction, and make `write_db()` just add writes to the existing transaction.

This improves performance, getting it back in line with pre-DBWrapper2. The full node performance is not particularly affected by this, but the wallet is.

This PR https://github.com/Chia-Network/chia-blockchain/pull/12234 makes wallet sync take 22% longer by switching to DBWrapper2. With this patch, that performance penalty goes away.

I suspect the main cost of the nested transactions are the round-trips to the sqlite thread, and not the actual DB operations themselves. But aiosqlite does not have an API that supports pipelining commands to its thread.

### DIFF
--- a/chia/clvm/spend_sim.py
+++ b/chia/clvm/spend_sim.py
@@ -113,7 +113,7 @@ class SpendSim:
         self.defaults = defaults
 
         # Load the next data if there is any
-        async with self.db_wrapper.write_db() as conn:
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
             await conn.execute("CREATE TABLE IF NOT EXISTS block_data(data blob PRIMARY_KEY)")
             cursor = await conn.execute("SELECT * from block_data")
             row = await cursor.fetchone()
@@ -134,7 +134,7 @@ class SpendSim:
             return self
 
     async def close(self) -> None:
-        async with self.db_wrapper.write_db() as conn:
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
             c = await conn.execute("DELETE FROM block_data")
             await c.close()
             c = await conn.execute(
@@ -159,7 +159,7 @@ class SpendSim:
 
     async def all_non_reward_coins(self) -> List[Coin]:
         coins = set()
-        async with self.mempool_manager.coin_store.db_wrapper.read_db() as conn:
+        async with self.mempool_manager.coin_store.db_wrapper.reader_no_transaction() as conn:
             cursor = await conn.execute(
                 "SELECT * from coin_record WHERE coinbase=0 AND spent=0 ",
             )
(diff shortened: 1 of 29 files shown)


arvidn:
@altendky this is what I have in mind. I'm not sure it makes sense to actually land this into the PR (since it's already quite large), but perhaps it could be rebased and landed into main immediately after.

mariano54:
How should these methods be used, and what happens if they are used incorrectly? I'm afraid it might cause some bugs.

Is it worth making the api more complex for only the 22% perfdcrmance speedup?

arvidn:
> What happens if these api methods are used incorrectly? I'm afraid it might cause some bugs.

The risk is that some piece of code expects nested-transaction behavior without getting it. Perhaps a better way to think about it is, you start a transaction, something throws an exception while you're working on it, but you handle it and go along with committing your transaction. Whatever threw the exception may have violated the invariant of the DB by having its first write go through, but threw an exception before its second write was made.

In that scenario, you still have an all-or-nothing behavior, but only at the top level. If the top level handles an exception, it is responsible to ensure the DB invariant is still intact. That responsibility is error prone. Partly because we don't exercise all failures in tests.

In my mind, the rule of thumb should probably be: "never catch exceptions when you're in a DB transaction".
As far as I can tell, our current code (and prior to DBWrapper2 in the full node), we only caught exceptions to manually roll back the DB.

> Is it worth making the api more complex for only the 22% performance speedup?

Since the PR that switches to DBWrapper2, another point of view is that we don't simplify the code quite as much, without introducing a performance regression.

I think 22% is quite significant. I would not be surprised if we can cut the toilet wallet sync time in half, if not more.
We spend about 40% of "phase 1" sleeping, and 80% of "phase 2" (but phase 1 is longer than 2). On top of the sleeping, we also spend CPU on things that should be a lot cheaper. Like we use python clvm instead of clvm_rs. But that's a bit of a longer tail.

altendky:
> In my mind, the rule of thumb should probably be: "never catch exceptions when you're in a DB transaction".

Hopefully we can be a bit more nuanced with 'catch specific exceptions with limited possible causes that you really know how to _handle_, not just silence'.  Where us raising custom exceptions instead of generic indistinguishable exceptions would make this more possible.  Since I know some people are not familiar with the ability to define your own exceptions in Python, here's an example of a custom exception hierarchy.  https://github.com/altendky/qtrio/blob/43c7ff24c0be2f7a3df86eef3c6cc5dae2f7ffd3/qtrio/_exceptions.py

mariano54 (dismissed):

altendky (dismissed):

arvidn:
I had to rebase to resolve conflicts

altendky (approved):

wjblanke (approved):
aok

PR #12281, opened 2022-07-07
avoid twin error reporting

wjblanke:


mariano54:
Should be replaced with this one:
https://github.com/Chia-Network/chia-blockchain/pull/12282

PR #12282, opened 2022-07-07
Tx submission idempotance, and prioritize wallet

mariano54:
1. We check for inclusion in mempool before adding to the queue. even if queue is full, `send_transaction` will return success if it's in mempool. Also return SUCCESS from full_node.respond_transaction in case it gets into the mempool while we are waiting in the queue.
2. Increase the priority of wallet transactions vs full node broadcasted transactions, so we don't have to wait in line as a wallet user. We do this by increasing the full node transactions priority to 1 (wallet is 0). Lower numbers go first.


wjblanke (approved):
aok

emlowe (commented):
looks ok to me, with the caveat that without really knowing how the code works, it's difficult to analyze

PR #12285, opened 2022-07-07
Increment macOS Intel installer build timeout to 55

altendky:
There are plenty of runs in the high 30s and even low 40s lately.


wjblanke (approved):
aok

PR #12290, opened 2022-07-07
remove extra bt from as many tests as possible ( speed up tests)

jack60612:
max speed, 25% faster now

### DIFF
--- a/chia/server/start_service.py
+++ b/chia/server/start_service.py
@@ -10,6 +10,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple
 from chia.daemon.server import singleton, service_launch_lock_path
 from chia.server.ssl_context import chia_ssl_ca_paths, private_ssl_ca_paths
 from ..protocols.shared_protocol import capabilities
+from chia.util.config import load_config
 
 try:
     import uvloop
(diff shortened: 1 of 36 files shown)


jack60612:
Appears to be faster

mariano54:
Can you provide some concrete numbers on the speed difference?

jack60612:
> Can you provide some concrete numbers on the speed difference?

is there an easy way to benchmark it?

jack60612:
> > Can you provide some concrete numbers on the speed difference?
> 
> is there an easy way to benchmark it? ( I am going to just time running all the tests without the change, then with)

ah, ill profile it

mariano54:
You can compare the changed test files in this PR vs main for example, and post the difference

jack60612:
This makes all of the tests run on my mac in 31 minutes, while main takes 42 minutes. that is a 10 MINUTE DECREASE!

mariano54 (approved):

mariano54:
Let's merge @wallentx

wallentx:
> Let's merge @wallentx

@Chia-Network/required-reviewers have been busy the past few days, so I'm trusting your suggestion (and I want to speed up tests)

PR #12294, opened 2022-07-08
Remove kwargs from `Service.start()`

altendky:


xdustinface (changes requested):
There is still the related parameter which the comment refers too in `WalletNode._start` https://github.com/Chia-Network/chia-blockchain/blob/280f462071e5fe1b7883cefac73712789e22b664/chia/wallet/wallet_node.py#L196

Which gets used in the wallet RPC here:

https://github.com/Chia-Network/chia-blockchain/blob/280f462071e5fe1b7883cefac73712789e22b664/chia/rpc/wallet_rpc_api.py#L209 

https://github.com/Chia-Network/chia-blockchain/blob/280f462071e5fe1b7883cefac73712789e22b664/chia/rpc/wallet_rpc_api.py#L287

https://github.com/Chia-Network/chia-blockchain/blob/280f462071e5fe1b7883cefac73712789e22b664/chia/rpc/wallet_rpc_api.py#L351


But seems like there is still no usage of the `kwargs` in `Service.start` so this PR could still be good to go in 🤔

altendky:
Apparently my searching failed, thanks.

altendky:
Yeah, I think the callee needs to keep it for the linked cases but the caller never exposes the fingerprint control.

xdustinface (approved):

emlowe (approved):

PR #12287, opened 2022-07-07
run pre-commit across platforms in ci

altendky:
why?  because we have developers working on multiple platforms.

Draft for:
- [x] would be nicer with https://github.com/Chia-Network/chia-blockchain/pull/12075

PR #12299, opened 2022-07-08
more mypy platforms in ci

altendky:
The existing setup only ran mypy on Intel Linux with Python 3.8.  This now runs mypy on the full matrix of operating systems, architectures, and Python versions.  The stdlib, and other library, interfaces can vary against all these parameters so this provides some amount of checking against those variations.

Draft for:
- [x] Windows type hinting issues are being addressed in https://github.com/Chia-Network/chia-blockchain/pull/12287


wallentx (dismissed):
test

emlowe (approved):
I like it

PR #12310, opened 2022-07-11
wallet code reuse

almogdepaz:
move WalletWeighProofHandler to wallet_node (remove from wallet_state_manager and wallet_blockchain)

remove code duplication from WalletWeighProofHandler

avoid double validation of wp in wallet long sync

remove _ses_requests cache from peer_request_cache (this code is not used since we don't use that request anymore)


chia/wallet/wallet_blockchain.py
```diff
@@ -57,30 +54,20 @@ async def create(
...
-        else:
-            self._sub_slot_iters = constants.SUB_SLOT_ITERS_STARTING
-            self._difficulty = constants.DIFFICULTY_STARTING
+        self._sub_slot_iters = constants.SUB_SLOT_ITERS_STARTING
+        self._difficulty = constants.DIFFICULTY_STARTING
 
         return self
 
-    async def new_weight_proof(self, weight_proof: WeightProof, records: Optional[List[BlockRecord]] = None) -> None:
+    async def new_valid_weight_proof(self, weight_proof: WeightProof, records: List[BlockRecord]) -> None:
```
almogdepaz, line 62: we where passing already validated WP's here since we already validate in "fetch_and_validate"

chia/wallet/wallet_blockchain.py
```diff
@@ -57,30 +54,20 @@ async def create(
         self._latest_timestamp = uint64(0)
         self._height_to_hash = {}
         self._block_records = {}
-        if self.synced_weight_proof is not None:
-            await self.new_weight_proof(self.synced_weight_proof)
```
mariano54, line 61: we need to initialize the blockchain when starting up, to set the block records and peak, right?

chia/wallet/wallet_node.py
```diff
@@ -257,7 +261,11 @@ async def _start(
         )
 
         assert self._wallet_state_manager is not None
-
+        if self._wallet_state_manager.blockchain.synced_weight_proof is not None:
+            weight_proof = self._wallet_state_manager.blockchain.synced_weight_proof
+            success, _, records = await self._weight_proof_handler.validate_weight_proof(weight_proof, True)
```
mariano54, line 266: why do we validate it if it's already in synced_weight_proof, don't we already know that it's valid?

chia/wallet/wallet_node.py
```diff
@@ -689,9 +698,6 @@ async def receive_state_from_peer(
     ) -> bool:
         # Adds the state to the wallet state manager. If the peer is trusted, we do not validate. If the peer is
         # untrusted we do, but we might not add the state, since we need to receive the new_peak message as well.
-
```
mariano54, line 692: why remove?

mariano54 (commented):
Mostly looks good, just a couple of comments

chia/wallet/wallet_node.py
almogdepaz: most of the validation is skipped we do this to get the records from the proof (we don't cache currently)
this happens only on a restart and replaces the initialization we used to do in wallet_blockchain.py

chia/wallet/wallet_blockchain.py
almogdepaz: yes, moved this to wallet_node so we can organise the code in a better way and not have references to the weight_proof_handler in multiple places, i think this closer to what we do in the full node.

chia/wallet/wallet_node.py
almogdepaz: this is a lint issue state_manager should never be None after startup
this is since we only initialize it in _start. felt like this should throw and not return without any error.
i changed it to an assert

mariano54 (approved):

wallentx:
Thanks for the discussion resolution

almogdepaz:
testing testing

PR #12314, opened 2022-07-11
Add /subscriptions RPC and add data_layer to chia rpc

emlowe:
Added `/subscriptions` RPC call to data_layer
Added data_layer as included server for `chia rpc`
Changed return from `get_owned_stores` to return `store_ids` instead of `launcher_ids`

Can now use the following CLI commands to get your own stores and your subscriptions

`chia rpc data_layer get_owned_stores`
`chia rpc data_layer subscriptions`


altendky (dismissed):

fchirica (approved):
LGTM

PR #12315, opened 2022-07-11
Add check to verify wallet is tracking DL in batch_update

emlowe:
Add check to make sure the wallet is tracking the DL in batch_update.

The wallet can still throw various exceptions that we should catch and rollback update - but this likely catches most of the common errors before we actually update the DL tree.

### DIFF
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -115,7 +115,7 @@ class DataLayer:
 
         # check before any DL changes that this singleton is currently owned by this wallet
         singleton_records: List[SingletonRecord] = await self.get_owned_stores()
-        if not [singleton for singleton in singleton_records if tree_id == singleton.launcher_id]:
+        if not any(tree_id == singleton.launcher_id for singleton in singleton_records):
             raise ValueError(f"Singleton with launcher ID {tree_id} is not owned by DL Wallet")
 
         t1 = time.monotonic()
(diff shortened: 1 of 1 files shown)


chia/data_layer/data_layer.py
```diff
@@ -112,6 +112,12 @@ async def batch_update(
         changelist: List[Dict[str, Any]],
         fee: uint64,
     ) -> TransactionRecord:
+
+        # check before any DL changes that this singleton is currently owned by this wallet
```
fchirica, line 116: I'm not sure if this checks if we're _subscribed_ to the singleton, or if we _own_ it. If we want owned only singletons, this _might_ give us subscribed singletons as well (I use the same method in `fetch_and_validate`, which is for subscriptions, not only for owned singletons).

emlowe: I believe it will give subscribed singletons as well - but I was trying to catch the most obvious case with the simplest check - I suppose I could use the new get_owned call to return the list of owned singletons. Maybe that is better

fchirica (dismissed):

chia/data_layer/data_layer.py
```diff
@@ -112,6 +112,12 @@ async def batch_update(
...
+        singleton_records: List[SingletonRecord] = await self.get_owned_stores()
+        if not [singleton for singleton in singleton_records if tree_id == singleton.launcher_id]:
```
altendky, line 118:
```suggestion
        if not any(tree_id == singleton.launcher_id for singleton in singleton_records):
```

altendky (approved):

PR #12316, opened 2022-07-11
streamable: More custom exceptions

xdustinface:


chia/util/streamable.py
```diff
@@ -44,6 +44,34 @@ def __init__(self, message: str, cls: Type[object]):
         )
 
 
+class ParameterMissingError(StreamableError):
+    def __init__(self, cls: type, missing: List[str]):
+        super().__init__(
+            f"{len(missing)} field{'s' if len(missing) > 1 else ''} missing for {cls.__name__}: {', '.join(missing)}"
```
altendky, line 50: I'm on the fence between english correctness and searchability.  Like being able to search a log for `fields missing for` without having to go all regex on it with `fields? missing for`.  But, in most cases you could just search for `ParameterMissingError`, at least when we don't strip it out in our logging...  so not totally sure I want to bank on that.  Anyways, rambling done.  Up to you.

chia/util/streamable.py
```diff
@@ -123,26 +151,26 @@ def convert_optional(convert_func: ConvertFunctionType, item: Any) -> Any:
 
 
 def convert_tuple(convert_funcs: List[ConvertFunctionType], items: Collection[Any]) -> Tuple[Any, ...]:
-    if len(items) != len(convert_funcs):
-        raise ValueError(f"Invalid size. Expected: {len(convert_funcs)}, got: {len(items)}")
     if not isinstance(items, (list, tuple)):
-        raise TypeError(f"expected: tuple or list, actual: {type(items).__name__}")
+        raise InvalidTypeError(tuple, type(items))
```
altendky, line 155: You know I like my passing by name...  Enough to mention it, not enough to raise a concern over it here.  Though the arguments could probably just be `expected` and `actual` since based on the exception type the `_type` or `_size` etc suffixes can be deduced.

chia/util/streamable.py
```diff
@@ -123,26 +151,26 @@ def convert_optional(convert_func: ConvertFunctionType, item: Any) -> Any:
...
 
 def convert_hex_string(item: str) -> bytes:
     if not isinstance(item, str):
-        raise TypeError(f"expected: hex-string, actual: {type(item).__name__}")
+        raise InvalidTypeError(str, type(item))
     try:
         return hexstr_to_bytes(item)
     except Exception as e:
-        raise TypeError(f"Can't convert the string {item!r} to bytes: {e}") from e
+        raise ConversionError(item, bytes, str(e)) from e
```
altendky, line 173: Maybe just pass in `e` itself?  Then you can use `traceback` to include the original exception type which is often nice to see along with the message.  Maybe something like `''.join(traceback.format_exception_only(type(x), value=x)).strip()`.

altendky (changes requested):
I think I'd drop the trailing `.`s since the exception messages aren't sentences in these cases.

As always, sorry for my latency...  `:[`

xdustinface:
@altendky Thanks, rebased and addressed your feedback :)

chia/util/streamable.py
```diff
@@ -44,6 +45,35 @@ def __init__(self, message: str, cls: Type[object]):
...
+
+class InvalidSizeError(StreamableError):
+    def __init__(self, expected: int, actual: int):
+        super().__init__(
+            f"Invalid size: Expected {expected}, Actual: {actual}",
+        )
+
+
+class ConversionError(StreamableError):
+    def __init__(self, value: Any, to_type: type, exception: Exception):
```
altendky, line 70: Does this need to be `Any` vs. `object`?

xdustinface: 6aced9cc55c5e22d20356566b1871f1cee5d10c7, nope :)

altendky (dismissed):

altendky (approved):

mariano54 (approved):

PR #12317, opened 2022-07-11
streamable: Raise `NotImplementedError` in `recurse_jsonify`

xdustinface:
This is to match the not implemented errors in `function_to_parse_one_item` and `function_to_stream_one_item`.

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -302,7 +302,7 @@ def recurse_jsonify(d: Any) -> Any:
     elif hasattr(d, "to_json_dict"):
         ret: Union[List[Any], Dict[str, Any], str, None, int] = d.to_json_dict()
         return ret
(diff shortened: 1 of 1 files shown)


altendky (changes requested):
I think this isn't the intended usage of this exception in any of the cases.

https://docs.python.org/3.10/library/exceptions.html#NotImplementedError
> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

Without encouraging more use of built-in exceptions, of those I think that `TypeError` would be the fit here since what we passed in has a type we don't handle.

xdustinface:
Closing here, just added [71cf7d5](https://github.com/Chia-Network/chia-blockchain/pull/12316/commits/71cf7d5bfbaf8dcd9bc2032fa707c5d75602cd4e) to #12316

PR #12322, opened 2022-07-12
Switch DL to NFT Chialisp

Quexington:
This is a large PR in which we delete a lot of the old DL Chialisp and DL offers code in order to replace the DL with the Chialisp we use for NFTs.  It also removes the ability for a stranger to spend the DL in order to announce its root.

I tried to break up the commits so that they're easy to review in sequence rather than all together at once.

This depends on https://github.com/Chia-Network/chia-blockchain/pull/12185


geoffwalmsley (commented):
General question: should we be handling cases where multiple updates of the DL occur in the same block?

geoffwalmsley (dismissed):
LGTM

emlowe (approved):

PR #12332, opened 2022-07-12
DL generate signed transaction

Quexington:
This implements the method that every wallet has `generate_signed_transaction` for the DL wallet. a few changes were made to the main method with which to update the root, but otherwise it's pretty non invasive to general wallet operations.


Quexington:
Depends on https://github.com/Chia-Network/chia-blockchain/pull/12322

chia/data_layer/data_layer_wallet.py
```diff
@@ -521,6 +542,49 @@ async def create_update_state_spend(
...
+        new_root_hash: Optional[bytes32] = None,
+        sign: bool = True,  # This only prevent signing of THIS wallet's part of the tx (fee will still be signed)
+    ) -> List[TransactionRecord]:
+        # Figure out the launcher ID
+        if len(coins) == 0:
+            if launcher_id is None:
+                raise ValueError("Not enough info to know which DL coin to send")
+        else:
+            if len(coins) != 1:
+                raise ValueError("The wallet can only send one DL coin at a time")
```
geoffwalmsley, line 566: How come we're limited to only sending one DL coin at a time? I'm not sure of the possible use cases - would we ever have a scenario of wanting to do bulk updates on multiple DLs?

geoffwalmsley (approved):
LGTM

chia/data_layer/data_layer_wallet.py
Quexington: You're right that we could do that, however we'd have to work out some complex-ish logic for making sure the coins are dependent on each other and don't all try to spend the same fee coin etc.  Since that didn't already exist and wasn't a current requirement, I decided to just restrict the user to one DL transaction at a time.

PR #12335, opened 2022-07-12
Wallet execute fetchall

arvidn:
this saves round-trips to the DB thread, saves some line of code and provides a small speedup compared to `main`.
Another part of this patch is to avoid building sql queries via string manipulation, and uses arguments instead.

sqlite will cache compiled queries, so the fewer query strings we use, the better the cache performs.

main:
```
INFO     Successfully subscribed and updated 103906 coin ids
INFO     Sync (trusted: True) duration was: 196.18760204315186
```

this patch:
```
INFO     Successfully subscribed and updated 103906 coin ids
INFO     Sync (trusted: True) duration was: 180.83409690856934
```

### DIFF
--- a/chia/util/db_wrapper.py
+++ b/chia/util/db_wrapper.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
-from typing import AsyncIterator, Dict, Optional
+import sqlite3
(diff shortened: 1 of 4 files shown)


mariano54 (approved):

wjblanke (approved):
aok

PR #12341, opened 2022-07-13
Puzzle store from row

arvidn:
minor simplification


altendky (approved):
I don't like the existing pattern of having a query in one place that hopefully keeps the selected columns properly matched to the indexing off in another function...  but I think this refactor is still correct.  Even if we can't use dict rows like in the data layer code, we could at least have an enumeration of the columns and reference by that rather than integer literals.

trepca (approved):

PR #12352, opened 2022-07-13
Typecheck ints used as port values.

aqk:
Keep values representing IP ports as uint16 within our own code.

Note that `YAML` and `web.TCPSite` do not support uint16 for port values, so we use `int` in those cases.


AmineKhaldi (dismissed):

AmineKhaldi (dismissed):

tests/util/socket.py
```diff
@@ -2,14 +2,16 @@
 import socket
 from typing import Set
 
+from chia.util.ints import uint16
+
 recent_ports: Set[int] = set()
```
altendky, line 7:
```suggestion
recent_ports: Set[uint16] = set()
```

tests/connection_utils.py
```diff
@@ -34,6 +34,7 @@ async def disconnect_all_and_reconnect(server: ChiaServer, reconnect_to: ChiaSer
 async def add_dummy_connection(
     server: ChiaServer, self_hostname: str, dummy_port: int, type: NodeType = NodeType.FULL_NODE
 ) -> Tuple[asyncio.Queue, bytes32]:
+    dummy_port = uint16(dummy_port)
```
altendky, line 37: It seems like it would be more consistent to push this out to the call sites?

altendky (commented):
Was enough isort-preference applied that the files can be removed from the isort exception list?

emlowe:
Rebased and revived PR - closing conversations.

emlowe:
Close and reopen for CI

emlowe:
Managed to get this up and running - but starting to think maybe it's not worth the hassle of the code and all the casts to make yaml happy.

emlowe:
Closing old draft PR

PR #12388, opened 2022-07-14
cmds: Move config loading into start/stop/passphrase commands

xdustinface:
I want to access the config in start/stop to print a warning if the `beta` section is in the config. The alternative to this PR would be to double load the config in start/stop, i think avoid loading twice makes sense but if there are objections about all this passing the config around let me know.

### DIFF
--- a/chia/cmds/passphrase_funcs.py
+++ b/chia/cmds/passphrase_funcs.py
@@ -4,6 +4,7 @@ import os
 import sys
 
 from chia.daemon.client import acquire_connection_to_daemon
+from chia.util.config import load_config
 from chia.util.keychain import Keychain, obtain_current_passphrase, supports_os_passphrase_storage
 from chia.util.keyring_wrapper import DEFAULT_PASSPHRASE_IF_NO_MASTER_PASSPHRASE
 from chia.util.misc import prompt_yes_no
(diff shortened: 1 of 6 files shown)


chia/cmds/passphrase_funcs.py
```diff
@@ -325,7 +326,7 @@ async def async_update_daemon_migration_completed_if_running() -> None:
         print("Missing root_path in context. Unable to notify daemon")
         return None
 
-    async with acquire_connection_to_daemon(root_path, quiet=True) as daemon:
+    async with acquire_connection_to_daemon(root_path, load_config(root_path, "config.yaml"), quiet=True) as daemon:
```
altendky, line 329: Offhand, it looks a bit odd that the config is loaded in this function but passed into the one in the diff segment above.  Maybe it should be passed in here as well?

altendky (commented):
Yeah, I dunno.  Configuration is one of those tough basically global state things that it's still useful to pass around but then how do you manage modifications to it and reloading and...  Maybe someday we end up with a configuration 'manager' object we can pass and that helps deal with all of that.

This does touch a lot of CLI stuff where we don't have tests and maybe don't have strict mypy, I'm not sure offhand.  So I guess that making sure it all still works is where my hang up on approval is.  Have you tested launching all of the relevant code?

xdustinface:
> This does touch a lot of CLI stuff where we don't have tests and maybe don't have strict mypy, I'm not sure offhand. So I guess that making sure it all still works is where my hang up on approval is. Have you tested launching all of the relevant code?

Yes the related commands still to work fine and i reviewed the places where those methods are called few times.

chia/cmds/passphrase_funcs.py
xdustinface: You mean pass it into `async_update_daemon_migration_completed_if_running`? Why?

altendky: We discussed this on a call.  While the theory stands, there's too much else around this that needs adjusted.  We'll deal with config loading consistency later.

altendky (approved):

wjblanke (approved):
aok

PR #12366, opened 2022-07-14
util: Improve `prompt_yes_no`

xdustinface:
Move the `(y/n)` addition into the function and drop it as default parameter.

PR #12389, opened 2022-07-14
cmds: Implement `chia beta`

xdustinface:
This PR introduces the undocumented CLI command `chia beta` with the following possible commands:

- `chia beta status` to show the current configuration status
- `chia beta configure` to change beta related settings
- `chia beta enable` to enable the beta test mode (this will put the `beta` section with the new log root path into the config. The default is the home directory, should be work for linux/mac/windows but i only tested on mac so far.)
- `chia beta disable` to disable the beta test mode
- `chia beta prepare_submission` to pre-process and zip the logfiles

Enabling the beta test mode has the following impact: 

```
beta:
  path: <the path choosen in `chia beta enable`>
```

will be added to the config which then leads to a different logging initialisation on the start of the daemon or any service. The logging root path will be set to the value of the config entry `beta.path` and the actual logging config used looks like below:

```
def get_beta_logging_config() -> Dict[str, Any]:
    return {
        "log_filename": f"{chia_full_version_str()}/chia-blockchain/beta.log",
        "log_level": "DEBUG",
        "log_stdout": False,
        "log_maxfilesrotation": 100,
        "log_maxbytesrotation": 50 * 1024 * 1024,
        "log_use_gzip": True,
    }
```

- `chia_full_version_str()` translates for example to `1.4.0`
- `log_maxfilesrotation` is bumped to 100 here to make sure we don't miss something which though means that we will potentially end up with ~5GiB log data per version so the user should probably clean those up after submission :) We could alternatively drop the files in `chia beta prepare_submission` once they are zipped.  
- `log_use_gzip` is introduced by this PR (set to `False` by default) and set to `True` in beta mode.

Enabling beta test mode also leads to the plotting logs being written to files in the `beta.path` directory.

The file structure in `beta.path` will look like:
image

Based on:
- #12366 
- #12388

### DIFF
--- a/chia/cmds/beta_funcs.py
+++ b/chia/cmds/beta_funcs.py
@@ -1,72 +1,79 @@
 import os
 import sys
-import zipfile
-from datetime import datetime
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional
 
 from chia.util.chia_logging import get_beta_logging_config
-from chia.util.config import lock_and_load_config, save_config
-from chia.util.misc import format_bytes, prompt_yes_no
+from chia.util.errors import InvalidPathError
+from chia.util.misc import format_bytes, prompt_yes_no, validate_directory_writable
 
 
-def warn_if_beta(config: Dict[str, Any]) -> None:
-    if "beta" in config:
+def default_beta_root_path() -> Path:
+    return Path(os.path.expanduser(os.getenv("CHIA_BETA_ROOT", "~/chia-beta-test"))).resolve()
+
+
+def warn_if_beta_enabled(config: Dict[str, Any]) -> None:
+    if config.get("beta", {}).get("enabled", False):
         print("\nWARNING: beta test mode is enabled. Run `chia beta disable` if this is unintentional.\n")
 
 
-def configure_beta_test_mode(root_path: Path, enable: bool) -> None:
-    with lock_and_load_config(root_path, "config.yaml") as config:
-        changed = False
-        result = "NOT ACTIVE"
-        if enable:
-            if "beta" in config:
-                sys.exit("beta test mode is already enabled. Disable it first with `chia beta disable`.")
-            logging_config = get_beta_logging_config()
-            # The `/ 5` is just a rough estimation for `gzip` being used by the log rotation in beta mode. It was like
-            # 7-10x compressed in example tests with 2MB files.
-            min_space = format_bytes(
-                int(logging_config["log_maxfilesrotation"] * logging_config["log_maxbytesrotation"] / 5)
-            )
-            print(
-                f"\nWARNING: Enabling the beta test mode increases disk writes and may lead to {min_space} of "
-                "extra logfiles getting stored on your disk. This should only be done if you are part of the beta test "
-                "program at: https://chia.net/beta-test\n"
-            )
-            if prompt_yes_no("Do you really want to enable the beta test mode?"):
-                default_path = os.path.expanduser("~/chia-beta-test")
-                beta_path: Optional[str] = None
-                while beta_path is None:
-                    user_input = input(
-                        "\nFill in a custom directory for the beta test logs or press enter to use the default "
-                        f"[{default_path}]:"
-                    )
-                    if user_input:
-                        if not Path(user_input).is_dir():
-                            print(f"Directory {user_input!r} doesn't exist.")
-                            continue
-                        beta_path = user_input
-                    else:
-                        beta_path = default_path
+def prompt_beta_warning() -> bool:
+    logging_config = get_beta_logging_config()
+    # The `/ 5` is just a rough estimation for `gzip` being used by the log rotation in beta mode. It was like
+    # 7-10x compressed in example tests with 2MB files.
+    min_space = format_bytes(int(logging_config["log_maxfilesrotation"] * logging_config["log_maxbytesrotation"] / 5))
+    return prompt_yes_no(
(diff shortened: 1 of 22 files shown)


altendky (commented):
I took a high level look and had a few vague thoughts.

- I think it might be nicer to have an enable flag in the beta config instead of deleting it to disable it.  Presently the config is just a path so it's not a huge deal, but even that, let alone more stuff, would get annoying to have to delete and reconfigure.
- Instead of disabling the user's existing logging, perhaps the beta logging should just be a separate log handler?  Handlers are the basic mechanism for controlling log output so a second one for the second location and configuration sounds reasonable, at least in passing.
- Should the subcommands be hidden?  Or just the group.  I haven't used this feature so don't offhand know for sure how it works, but I would imagine that hiding the group would make `chia --help` not show it but it seems like it would be nice to have `chia beta --help` list out the subcommands.
- I think that having the actual command line (or a json dump of it maybe) in the log would be nice.  It makes it easy to copy and paste to recreate that run externally and it makes it flagrantly obvious what was passed instead of hoping the plotter outputs that info for you even when it fails.

chia/cmds/beta_funcs.py
```diff
@@ -0,0 +1,130 @@
...
+from pathlib import Path
+from typing import Any, Callable, Dict, List, Optional
+
+from chia.util.chia_logging import get_beta_logging_config
+from chia.util.config import lock_and_load_config, save_config
+from chia.util.misc import format_bytes, prompt_yes_no
+
+
+def warn_if_beta(config: Dict[str, Any]) -> None:
+    if "beta" in config:
```
almogdepaz, line 14: i think i agree with @altendky here,  whats the benefit of using  this as apposed to an enable/disable flag ?

if your worried about updating the conf then a missing beta can mean False and if it is in the config we can check enabled/disabled

chia/cmds/beta_funcs.py
```diff
@@ -0,0 +1,130 @@
...
+from datetime import datetime
+from pathlib import Path
+from typing import Any, Callable, Dict, List, Optional
+
+from chia.util.chia_logging import get_beta_logging_config
+from chia.util.config import lock_and_load_config, save_config
+from chia.util.misc import format_bytes, prompt_yes_no
+
+
+def warn_if_beta(config: Dict[str, Any]) -> None:
```
almogdepaz, line 13: i think i agree with @altendky about this, we can treat beta not in the config as False and if it is check if Enabled/Disabled

chia/util/chia_logging.py
```diff
@@ -1,12 +1,25 @@
...
 from logging.handlers import SysLogHandler
 
+from chia.cmds.init_funcs import chia_full_version_str
 from chia.util.path import path_from_root
+from chia.util.default_root import DEFAULT_ROOT_PATH
+
+
+def get_beta_logging_config() -> Dict[str, Any]:
+    return {
+        "log_filename": f"{chia_full_version_str()}/chia-blockchain/beta.log",
```
almogdepaz, line 16: filename = version/chia-blockchain/beta.log ?
if i got it correct chia-blockchain seems redundant here

xdustinface: This is just to have separate folders for normal logs vs plotting, see the structure here:

almogdepaz:
lgtm other then the comment on beta enabled/disabled vs in config not in config 

i dont understand the benefit of multiple loggers, they would output the same thing if i understand what @altendky is suggesting and seems like this will require more code to support this feature

xdustinface:
> lgtm other then the comment on beta enabled/disabled vs in config not in config

Im currently working on some refactoring of the code for the enabled/disabled fun.

> i dont understand the benefit of multiple loggers, they would output the same thing if i understand what @altendky is suggesting and seems like this will require more code to support this feature

One reason i can think about to keep the standard logging going is for example if people use tools like chiadog where the logs are parsed to monitor the status of the farm it would need to change that config too when beta mode is enabled.. so, i will also take a look at that after the config stuff :)

xdustinface:
@altendky @almogdepaz 
> * I think it might be nicer to have an enable flag in the beta config instead of deleting it to disable it.

See bf37bff

> * Instead of disabling the user's existing logging, perhaps the beta logging should just be a separate log handler? 

See e3b012aa1372dd6bbf49a38d4cce236759989cbe

> * Should the subcommands be hidden? 

See ac0e0f0305ce03e0af78f79fb72d5288d531ec5b

> * I think that having the actual command line (or a json dump of it maybe) in the log would be nice.

See 40a5cc0f3a20ac71b73afce19d186042ca604f26


I also added:

- Two new subcommands `beta status` (e0139884e6671e401ec8992490aa063f981d4e0d) `beta configure` (e5cf203d652c073254220bc7582c6687f002ebfa)
- Tests for all beta CLI commands in b3fa8f9504385ece1179aeaf6a08a7d9253602c7

altendky:
Space joining the plotting command hazards paths with spaces presenting as separate tokens if you don't quote properly.  That's what made me mention just a JSON dump of the list as an option that wouldn't have that issue.  There are bash quoting tools in ...  `shutil` maybe?  But that's bash and...  I dunno.  I guess JSON is less copy/pasteable when trying recreate but it's also zero doubt about what each token is.

almogdepaz (dismissed):

almogdepaz (approved):
ignoring lgtm warnings since it has been buggy

mariano54 (approved):

PR #12392, opened 2022-07-14
block sending an amount of 0

jack60612:


jack60612:
https://trello.com/c/C13Dohrx/1433-keybase-chat-chianetworkpublicgeneral-210991-bug-report-you-can-send-0-mojos-with-override-option

altendky:
Does this check belong in `check_unusual_transaction()`?  I'm not sure.  It is unusual, but maybe it is also totally invalid?  Does nothing else downstream complain?  If I just don't have the context then someone else can certainly fill in.

jack60612:
> Does this check belong in `check_unusual_transaction()`? I'm not sure. It is unusual, but maybe it is also totally invalid? Does nothing else downstream complain? If I just don't have the context then someone else can certainly fill in.

i don't think it does, sending a transaction with no value makes no sense, and should be avoided if possible because of possible edge cases.

mariano54:
> > Does this check belong in `check_unusual_transaction()`? I'm not sure. It is unusual, but maybe it is also totally invalid? Does nothing else downstream complain? If I just don't have the context then someone else can certainly fill in.
> 
> i don't think it does, sending a transaction with no value makes no sense, and should be avoided if possible because of possible edge cases.

Good point. For regular users of CLI, they don't need to send value 0. Users of the RPC might need to though, so this should be kept at the UI layer (CLI).

jack60612:
> > > Does this check belong in `check_unusual_transaction()`? I'm not sure. It is unusual, but maybe it is also totally invalid? Does nothing else downstream complain? If I just don't have the context then someone else can certainly fill in.
> > 
> > 
> > i don't think it does, sending a transaction with no value makes no sense, and should be avoided if possible because of possible edge cases.
> 
> Good point. For regular users of CLI, they don't need to send value 0. Users of the RPC might need to though, so this should be kept at the UI layer (CLI).

agreed and thats what i did.

jack60612:
> So we want to have to reimplement this in each UI layer? Our CLI and GUI and any other general client that community devs write? I guess I care enough to comment, but not enough to raise a concern. Don't wait up on me on this point.
> 
> What does a 0 amount send mean that it would be useful somewhere?

I think that it's just not a good look for the cli as people have been thinking this is a bug, to my knowledge the gui and all other community apps don't allow sending a transaction of 0. We should leave it enabled on the rpc's but disable it on the cli in my opinion, it just causes confusion.

altendky:
It's not really clear to me why there should be a difference between the CLI and RPC.  In both cases it seems like it would be uncommon.  Both cases could have the same override ability for uncommon cases.  Also, even if the RPC and CLI should be different that doesn't make this not belong in `check_unusual_transaction()` since that's only used in the CLI per my quick search.  Drifting off topic, the unusual check is pretty lax.  `:]`  The fee has to be as much as the amount being sent.

jack60612:
> It's not really clear to me why there should be a difference between the CLI and RPC. In both cases it seems like it would be uncommon. Both cases could have the same override ability for uncommon cases. Also, even if the RPC and CLI should be different that doesn't make this not belong in `check_unusual_transaction()` since that's only used in the CLI per my quick search. Drifting off topic, the unusual check is pretty lax. `:]` The fee has to be as much as the amount being sent.

someone was complaining about it and it was on our trello board, if you think its a non issue ill close this pr

altendky:
I've only questioned 1) should it be impossible, 2) should the RPC be different, and 3) should we group these things in the uncommon check.  Making it an 'are you sure?' scenario seems perfectly sensible to me.  But, I don't yet know what the true use case is for a 0 amount send so I haven't really formed an actual opinion on 1).  For 2), the RPC is already different.  And with 3), I think none of the comments so far actually distinguish between the options.

jack60612:
> I've only questioned 1) should it be impossible, 2) should the RPC be different, and 3) should we group these things in the uncommon check. Making it an 'are you sure?' scenario seems perfectly sensible to me. But, I don't yet know what the true use case is for a 0 amount send so I haven't really formed an actual opinion on 1). For 2), the RPC is already different. And with 3), I think none of the comments so far actually distinguish between the options.

basically this can only happen after overriding, so should i just ask for a confirmation before sending?

altendky:
Hmm, I guess the interesting manner of checking for an unusual transaction does end up very implicitly catching a 0 amount.  If that is part of the intent I'd suggest it be more explicit about it.  Also, I'm not sure that a 10 mojo fee is unusual for a 10 mojo send.  Anyways, in so much as it seems that 0 amount transactions have some use (still curious what) I don't see why we need to block people from making them after the provide an override for the unusualness of the send.

I still don't need to block this if Mariano wants to approve.  But me learning what happens with a 0 amount tx would probably be good before chatting much more.  Do we create a 0 amount coin for the recipient and send the entirety of whatever coin we were using to send back to ourselves as a new change coin?  Or does the command just succeed having done nothing at all?  Or do we output a cryptic error?

jack60612:
> Hmm, I guess the interesting manner of checking for an unusual transaction does end up very implicitly catching a 0 amount. If that is part of the intent I'd suggest it be more explicit about it. Also, I'm not sure that a 10 mojo fee is unusual for a 10 mojo send. Anyways, in so much as it seems that 0 amount transactions have some use (still curious what) I don't see why we need to block people from making them after the provide an override for the unusualness of the send.
> 
> I still don't need to block this if Mariano wants to approve. But me learning what happens with a 0 amount tx would probably be good before chatting much more. Do we create a 0 amount coin for the recipient and send the entirety of whatever coin we were using to send back to ourselves as a new change coin? Or does the command just succeed having done nothing at all? Or do we output a cryptic error?

ill try it and let you know

jack60612:
> Hmm, I guess the interesting manner of checking for an unusual transaction does end up very implicitly catching a 0 amount. If that is part of the intent I'd suggest it be more explicit about it. Also, I'm not sure that a 10 mojo fee is unusual for a 10 mojo send. Anyways, in so much as it seems that 0 amount transactions have some use (still curious what) I don't see why we need to block people from making them after the provide an override for the unusualness of the send.
> 
> I still don't need to block this if Mariano wants to approve. But me learning what happens with a 0 amount tx would probably be good before chatting much more. Do we create a 0 amount coin for the recipient and send the entirety of whatever coin we were using to send back to ourselves as a new change coin? Or does the command just succeed having done nothing at all? Or do we output a cryptic error?

it outputs a cryptic error `Exception from 'wallet' {'error': '', 'success': False}`

altendky:
Mm, an assertion error exposed to the users as an empty string.

mariano54 (approved):

trepca (approved):

PR #12398, opened 2022-07-15
tests: Drop redundant `expected_state` parameter of `run_sync_step`

xdustinface:
The expected state is already present in `sync_step`.


altendky (commented):
What was the point of this code?

altendky (approved):

mariano54 (approved):

PR #12412, opened 2022-07-15
make DBWrapper2 test more reliable

arvidn:


tests/core/util/test_db_wrapper.py
```diff
@@ -216,6 +214,7 @@ async def test_mixed_readers_writers(acquire_outside: bool) -> None:
                     tasks.append(task)
```
aqk, line 214: One small quibble / note:

This loop creates two tasks when index == 100

aqk (approved):
LGTM

emlowe (approved):

PR #12427, opened 2022-07-16
Add bulk cancel API

ytx1991:
A new bulk API for cancel NFT offers based on the wallet id.
If wallet_id = None then cancel all NFT offers.


trepca:
@ytx1991 lint not happy, please check

chia/rpc/wallet_rpc_api.py
```diff
@@ -1066,13 +1067,62 @@ async def cancel_offer(self, request: Dict) -> EndpointResult:
...
-
+        txs = None
         async with self.service.wallet_state_manager.lock:
             if secure:
-                await wsm.trade_manager.cancel_pending_offer_safely(bytes32(trade_id), fee=fee)
+              txs = await wsm.trade_manager.cancel_pending_offer_safely(bytes32(trade_id), fee=fee)
             else:
                 await wsm.trade_manager.cancel_pending_offer(bytes32(trade_id))
-        return {}
+        return {"success": True, "TXs": txs}
```
Quexington, line 1076: Not a necessary change, but we don't need `"success": True`, the RPC API class takes care of that automatically for us.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1066,13 +1067,62 @@ async def cancel_offer(self, request: Dict) -> EndpointResult:
...
+        cancel_all = request.get("cancel_all", False)
+        if cancel_all:
+            asset_id = None
+        else:
+            asset_id = request.get("asset_id", "xch")
+
+        start: int = 0
+        end: int = start + batch_size
+        trade_mgr = self.service.wallet_state_manager.trade_manager
+        log.info(f"Start cancelling offers for asset_id {asset_id} ...")
```
Quexington, line 1091: This will log will output `"Start cancelling offers for asset_id None ..." in the `cancel_all` case.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1066,13 +1067,62 @@ async def cancel_offer(self, request: Dict) -> EndpointResult:
...
+                    records.append(trade)
+                    continue
+                if trade.offer and trade.offer != b'':
+                    summary = Offer.from_bytes(trade.offer).summary()
+                    if asset_id == list(summary[0].keys())[0]:
+                        records.append(trade)
+                        continue
+                    if asset_id == list(summary[1].keys())[0]:
+                        records.append(trade)
+                        continue
```
Quexington, line 1115: The `summary` method is really just a display method, you shouldn't use it for business logic.  `if bytes32.from_hexstr(asset_id) in offer.driver_dict` is a better check (The asset_id for XCH in that case is None).

chia/rpc/wallet_rpc_api.py
```diff
@@ -1066,13 +1067,62 @@ async def cancel_offer(self, request: Dict) -> EndpointResult:
...
+
+            async with self.service.wallet_state_manager.lock:
+                await trade_mgr.cancel_pending_offers(records, fee, secure)
```
Quexington, line 1118: It's not clear to the user of this API that you're paying this fee every batch.  Maybe that's desired but it's definitely not clear in the API.  The parameter should at least be `fee_per_bundle` or something.

chia/wallet/trade_manager.py
```diff
@@ -289,6 +289,96 @@ async def cancel_pending_offer_safely(
...
+        # Aggregate spend bundles to the first tx
+        if len(all_txs) > 0:
+            all_txs[0] = dataclasses.replace(all_txs[0], spend_bundle=SpendBundle.aggregate(bundles))
+        if secure:
+            for tx in all_txs:
+                await self.wallet_state_manager.add_pending_transaction(
+                    tx_record=dataclasses.replace(tx, fee_amount=fee)
+                )
+        else:
+            self.wallet_state_manager.state_changed("offer_cancelled")
```
Quexington, line 377: Shouldn't this state be sent, regardless of `secure`?

Quexington (changes requested):

chia/wallet/trade_manager.py
ytx1991: In the secure mode it should send a notification after the transaction sent, I assume. I just copy the logic from the single cancelling API. It only sends notification for secure = false.

Quexington (dismissed):
Looks good to me :+1:

trepca (dismissed):

mariano54:
Can you change the NFT tests to not create so many blocks? This causes delay in CI, and the NFT tests are approaching the slowest of all. Please see here for how to make them faster:

https://github.com/Chia-Network/chia-blockchain/pull/12688

wjblanke (approved):
aok

PR #12433, opened 2022-07-16
Return before fetching weight proof if secondary sync task is running

nuyawktuah:
I noticed that we often download multiple weight proofs from peers with which we are currently performing a secondary sync. You can verify this with `cat ~/.chia/mainnet/log/debug.log | grep 'request_proof_of_weight to peer'` and looking at the timestamps. This seems like a waste of resources as we will return early a [few lines later](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/wallet_node.py#L1008) without handling the new peak message.

This pr adds a check to see if we are currently performing a secondary sync before fetching and validating the weight proof. From what I can tell, sync behavior should remain unchanged, apart from skipping the [call](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/wallet_node.py#L990) to `wallet_state_manager.blockchain.new_weight_proof`, which seems safe to me.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -966,7 +966,7 @@ class WalletNode:
                     syncing = True
                     self.wallet_state_manager.set_sync_mode(True)
 
-                if syncing and not (self._secondary_peer_sync_task is None or self._secondary_peer_sync_task.done()):
+                if not syncing and not (self._secondary_peer_sync_task is None or self._secondary_peer_sync_task.done()):
                     self.log.info("Will not do secondary sync, there is already another sync task running.")
                     return
 
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -965,6 +965,11 @@ class WalletNode:
                 if far_behind or len(self.synced_peers) == 0:
                     syncing = True
                     self.wallet_state_manager.set_sync_mode(True)
+
+                if syncing and not (self._secondary_peer_sync_task is None or self._secondary_peer_sync_task.done()):
+                    self.log.info("Will not do secondary sync, there is already another sync task running.")
+                    return
+
                 try:
                     (
                         valid_weight_proof,
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -966,7 +966,7 @@ class WalletNode:
                     syncing = True
                     self.wallet_state_manager.set_sync_mode(True)
 
-                if not syncing and not (self._secondary_peer_sync_task is None or self._secondary_peer_sync_task.done()):
+                if not (self._secondary_peer_sync_task is None or self._secondary_peer_sync_task.done()):
                     self.log.info("Will not do secondary sync, there is already another sync task running.")
                     return
 
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -966,7 +966,9 @@ class WalletNode:
                     syncing = True
                     self.wallet_state_manager.set_sync_mode(True)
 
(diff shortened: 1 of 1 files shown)


chia/wallet/wallet_node.py
```diff
@@ -965,6 +965,11 @@ async def new_peak_wallet(self, new_peak: wallet_protocol.NewPeakWallet, peer: W
                 if far_behind or len(self.synced_peers) == 0:
                     syncing = True
                     self.wallet_state_manager.set_sync_mode(True)
+
+                if not syncing and not (self._secondary_peer_sync_task is None or self._secondary_peer_sync_task.done()):
```
almogdepaz, line 969: this should be 
"if  (self._secondary_peer_sync_task is None or self._secondary_peer_sync_task.done()):"

"syncing" is True if we are far behind/first time syncing to this peer
in both cases we should validate the WP

almogdepaz (changes requested):

chia/wallet/wallet_node.py
olivernyc: I removed the `syncing` check, but on second thought I'm not sure that's correct? We only want to skip the weight proof validation if we're going down the secondary sync path (when `syncing` is true). If we're going to do a "primary" sync, we don't want to skip it.

olivernyc: Also I don’t think your description of `syncing` is correct either.. it’s true if we are far behind or we have no synced peers. My understanding is that we want to do a quick sync from the fork point with the first peer and then kick off a sync from 0 in the background for subsequent peers while we continue handling peaks. Regardless, the naming is pretty confusing and should probably be called something like `sync_from_fork_point`.

almogdepaz: we always want to validate a weight proof prior to syncing to a peer 
if we are far behind or this is the first time syncing to this peer. i agree the naming here needs some work

almogdepaz (dismissed):

chia/wallet/wallet_node.py
olivernyc: Right, and we always will, even with the `syncing` condition. The purpose of this check is simply to avoid the case where we validate a weight proof and then return without syncing because we are already performing a secondary sync. We need the `syncing` condition here to prevent the edge case where there is a secondary sync running but we are going down the primary sync path for this peer (probably rare but I think it's still worth guarding against).

olivernyc: No worries, appreciate you taking a look! I've gone ahead and added back the syncing check. I believe the changes are safe now.

I was actually thinking of moving the weight proof request for secondary syncs into the background task so we can continue handling messages while fetching and validating the proof. I think this would fix the race condition you are referring to and also prevent a potential stall of up to six minutes (the default wp request timeout). But I decided against it because it would be a bigger refactor and I wasn't sure if there would be side effects. Something to consider for the future maybe.

almogdepaz: i was looking at this code some more and since we actually only call new_peak_wallet in _process_new_subscriptions and that allows only one concurrent call to new_peak_wallet, its looks like your suggestion will do the trick.

im working on refactoring new_peak_wallet, ill include your change

olivernyc: Awesome, looking forward to seeing the refactor

mariano54 (approved):

wjblanke (approved):
aok

wallentx:
Unresolved conversations, and unsigned commits are blocking this from merge

mariano54:
This has been already included in https://github.com/Chia-Network/chia-blockchain/pull/12877. Thanks @olivernyc !

PR #12446, opened 2022-07-17
stop consuming the exception

altendky:
Can monitor with:
```bash
while true; do date --iso-8601=s; yes q | venv/bin/chia wallet show; echo; echo sleeping...; sleep 30; done
```

Draft for:
- [ ] This is just debugging


altendky:
Replaced by a production ready alternative.  https://github.com/Chia-Network/chia-blockchain/pull/12459

PR #12447, opened 2022-07-17
define `FullNode` attributes in `.__init__()`

altendky:
Draft for:
- [x] Hinting is unhappy now that it is aware of all the optional (or previously maybe not present) stuff


xdustinface (changes requested):
There are a bunch of redundant hints in `__init__` now, shouldn't they be dropped?

chia/full_node/full_node.py
```diff
@@ -127,57 +153,147 @@ def __init__(
...
+        return self._compact_vdf_sem
+
     def _set_state_changed_callback(self, callback: Callable[..., Any]) -> None:
         self.state_changed_callback = callback
 
     async def _start(self) -> None:
-        self.timelord_lock = asyncio.Lock()
-        self.compact_vdf_sem = asyncio.Semaphore(4)
+        self._timelord_lock = asyncio.Lock()
+        self._compact_vdf_sem = asyncio.Semaphore(4)
```
xdustinface, line 289: Why not just move this into `__init__`, make them non-optional and get rid of the related properties? Same question for all the other non-db related ones below.

altendky: Some asyncio stuff expects to be created inside an event loop.  I don't know how required it is vs. just a thing it warns when you don't do.  Mostly though I was targeting handling that separately, maybe.

xdustinface: Here is a similar case where it seems to work, at least with `Lock`, https://github.com/Chia-Network/chia-blockchain/blob/main/chia/data_layer/data_layer.py#L86, couldn't find any for `Semaphore` though.. are there any warnings? Because if not, this `Optional` + `@property` thing is for no reason :/

xdustinface (dismissed):

xdustinface (approved):

emlowe (approved):

PR #12462, opened 2022-07-18
Set a minimum for initial_num_public_keys and change the default, to …

mariano54:
…avoid issues with not enough addresses.

It takes around 8 seconds on my laptop to generate 750 addresses.
Down to 3 seconds with the latest changes in main: https://github.com/Chia-Network/chia-blockchain/pull/12467

Has been tested in full node to not have any performance impact, for untrusted peers connected to nodes.
This change will reduce the number of issues people have with missing balances.


emlowe:
So we are ok with saying any value less than 750 is not allowed?

mariano54:
> So we are ok with saying any value less than 750 is not allowed?

Value can be lower or higher. We used to do 500. 
It will help a lot in terms of support request of missing coins etc.

paninaro (commented):
lgtm

paninaro (approved):

paninaro (approved):
lgtm

emlowe (approved):
Seems a good time to introduce this change

PR #12463, opened 2022-07-18
Sleep to allow neworking layer to execute

mariano54:
To replicate the issue:
Set `initial_num_public_keys` to something that takes > 30 seconds, like 10000.

Then start the wallet. When it's done generating the PHs, do `chia wallet show`. It results in `No keys loaded` issue.

This fixes it.

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -342,6 +342,8 @@ class WalletStateManager:
                     self.log.debug(
                         f"Puzzle at index {index} wallet ID {wallet_id} puzzle hash {puzzlehash_unhardened.hex()}"
                     )
+                    # We await sleep here to allow an asyncio context switch (since the other parts of this loop do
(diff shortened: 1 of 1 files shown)


paninaro (dismissed):
Looks good to me. I would suggest adding a comment on the line so that future readers understand why the sleep is there (though I suppose they could blame the line).

emlowe:
This is good for this particular case, but I imagine there are many scenarios where similar things can happen, so a more robust overall solution should still be considered

paninaro (approved):

mariano54:
> This is good for this particular case, but I imagine there are many scenarios where similar things can happen, so a more robust overall solution should still be considered

Good point. What I think we need to fix is that the wallet does not try reconnecting to the daemon. If it's interrupted, it should reconnect right? This is probably why so many people have zombies.

wjblanke (approved):
aok

PR #12467, opened 2022-07-18
optimize ph generation (~65% faster)

mariano54:
Time for generation 1000 puzzle hashes:
before:  9-10 seconds
after: 3-3.5 seconds

These timings are for XCH puzzles. For CAT or other wallets, the improvements might be less than 65%.

This required fixing some call sites where we passed in bytes. It was working because we double called bytes.
This is a low risk change.

### DIFF
--- a/chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
+++ b/chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
@@ -73,8 +73,6 @@ DEFAULT_HIDDEN_PUZZLE_HASH = DEFAULT_HIDDEN_PUZZLE.get_tree_hash()  # this puzzl
 
 MOD = load_clvm("p2_delegated_puzzle_or_hidden_puzzle.clvm")
 
-SYNTHETIC_MOD = load_clvm("calculate_synthetic_public_key.clvm")
-
 PublicKeyProgram = Union[bytes, Program]
 
 GROUP_ORDER = 0x73EDA753299D7D483339D80809A1D80553BDA402FFFE5BFEFFFFFFFF00000001
--- a/chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
+++ b/chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
@@ -89,7 +89,7 @@ def calculate_synthetic_offset(public_key: G1Element, hidden_puzzle_hash: bytes3
 
 def calculate_synthetic_public_key(public_key: G1Element, hidden_puzzle_hash: bytes32) -> G1Element:
     r = SYNTHETIC_MOD.run([bytes(public_key), hidden_puzzle_hash])
-    return G1Element.from_bytes(r.as_atom())
+    return G1Element.from_bytes_unchecked(r.as_atom())
 
 
 def calculate_synthetic_secret_key(secret_key: PrivateKey, hidden_puzzle_hash: bytes32) -> PrivateKey:
--- a/chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
+++ b/chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
@@ -88,8 +88,10 @@ def calculate_synthetic_offset(public_key: G1Element, hidden_puzzle_hash: bytes3
 
 
 def calculate_synthetic_public_key(public_key: G1Element, hidden_puzzle_hash: bytes32) -> G1Element:
-    r = SYNTHETIC_MOD.run([bytes(public_key), hidden_puzzle_hash])
-    return G1Element.from_bytes_unchecked(r.as_atom())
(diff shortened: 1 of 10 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -41,7 +41,14 @@
...
 from chia.wallet.cat_wallet.cat_wallet import CATWallet
 from chia.wallet.derivation_record import DerivationRecord
-from chia.wallet.derive_keys import master_sk_to_wallet_sk, master_sk_to_wallet_sk_unhardened
+from chia.wallet.derive_keys import (
+    master_sk_to_wallet_sk,
+    master_sk_to_wallet_sk_unhardened,
+    master_sk_to_wallet_sk_intermediate,
+    _derive_path,
+    master_sk_to_wallet_sk_unhardened_intermediate,
+    _derive_path_unhardened,
```
altendky, line 50: Uses a couple private functions outside of their module.

tests/clvm/test_singletons.py
```diff
@@ -87,7 +87,7 @@ async def test_singleton_top_layer(self, version):
             # START TESTS
             # Generate starting info
             key_lookup = KeyTool()
-            pk: G1Element = public_key_for_index(1, key_lookup)
+            pk: G1Element = G1Element.from_bytes(public_key_for_index(1, key_lookup))
```
altendky, line 90: I would ask that the function be hinted so it will get checked to catch stuff like this...  but that exposes 72 complaints...  Another time I guess.

tests/core/test_cost_calculation.py
```diff
@@ -131,7 +132,7 @@ async def test_mempool_mode(self, softfork_height, bt):
         pk = bytes.fromhex(
             "88bc9360319e7c54ab42e19e974288a2d7a817976f7633f4b43f36ce72074e59c4ab8ddac362202f3e366f0aebbb6280"
         )
-        puzzle = p2_delegated_puzzle_or_hidden_puzzle.puzzle_for_pk(pk)
+        puzzle = p2_delegated_puzzle_or_hidden_puzzle.puzzle_for_pk(G1Element.from_bytes(pk))
```
altendky, line 135: We could hint to accept both.  We just `bytes()` down at the bottom which would end up doing nothing.

altendky (commented):
I'm mostly ok with the Python.  I don't see any errors.  I can't review the adjustments to the key related calculations though.

mariano54:
> I'm mostly ok with the Python. I don't see any errors. I can't review the adjustments to the key related calculations though.

It's the same calculation except it's using BLS instead of going into CLVM and calling the BLS operations there. We have all those methods so it makes no sense to do it in CLVM.

tests/clvm/test_singletons.py
mariano54: you mean G1Element.from_bytes? We should def. hint the bls stuff

altendky: No, the test function.  Sorry.  But yes, everything should get hinted.  `:]`

emlowe (approved):

chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
```diff
@@ -73,8 +73,6 @@
 
 MOD = load_clvm("p2_delegated_puzzle_or_hidden_puzzle.clvm")
 
-SYNTHETIC_MOD = load_clvm("calculate_synthetic_public_key.clvm")
```
arvidn, line 76: do we have any clvm code that depends on this now? or is this part of a library of useful chialisp functions? `calculate_synthetic_public_key.clvm`

chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
```diff
@@ -88,8 +86,10 @@ def calculate_synthetic_offset(public_key: G1Element, hidden_puzzle_hash: bytes3
 
 
 def calculate_synthetic_public_key(public_key: G1Element, hidden_puzzle_hash: bytes32) -> G1Element:
-    r = SYNTHETIC_MOD.run([bytes(public_key), hidden_puzzle_hash])
-    return G1Element.from_bytes(r.as_atom())
+    synthetic_offset: PrivateKey = PrivateKey.from_bytes(
+        calculate_synthetic_offset(public_key, hidden_puzzle_hash).to_bytes(32, "big")
```
arvidn, line 90: this is a bit confusing (at least to a layman). We create a private key and call it "offset".

chia/wallet/wallet_state_manager.py
arvidn: we should probably make these public

arvidn (approved):

chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
mariano54: Dont think so, it can be removed. Not sure why we did it in clvm in the first place

PR #12475, opened 2022-07-19
Add RPCs for getting/extending the current derivation path index

paninaro:
This PR adds the RPCs:
`get_current_derivation_index` - Returns `index` referencing the current derivation index (for the XCH wallet, but should apply across all wallets). The value of the returned index may be null if the DB doesn't have a current derivation path.

`extend_derivation_index` - Takes an `index` parameter from the request and generates additional puzzle hashes up to and including that index. If the `index` value is < the current derivation index, no action is taken. Returns `index` in the same manner as `get_current_derivation_index`

Adds a `state_changed` message:
`new_derivation_index` - Posted when the wallet state manager generates new puzzle hashes. The `additional_data` object in the payload will include `index` as above.

CLI options have been added to use the new RPCs:

`chia wallet get_derivation_index`
`chia wallet update_derivation_index -i <new index>`

### DIFF
--- a/chia/cmds/wallet.py
+++ b/chia/cmds/wallet.py
@@ -254,6 +254,45 @@ def delete_unconfirmed_transactions_cmd(wallet_rpc_port: Optional[int], id, fing
     asyncio.run(execute_with_wallet(wallet_rpc_port, fingerprint, extra_params, delete_unconfirmed_transactions))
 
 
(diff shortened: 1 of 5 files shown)


mariano54 (approved):
This looks right.it would be nice to have tests though

wjblanke (approved):
aok

PR #12479, opened 2022-07-19
keychain: Drop `supports_keyring_passphrase`

xdustinface:
Drops the not longer needed option to disable the passphrase feature via the env variable `CHIA_PASSPHRASE_SUPPORT`.

### DIFF
--- a/chia/cmds/passphrase_funcs.py
+++ b/chia/cmds/passphrase_funcs.py
@@ -14,8 +14,6 @@ from io import TextIOWrapper
 from pathlib import Path
 from typing import Any, Dict, Optional, Tuple
 
-# Click drops leading dashes, and converts remaining dashes to underscores. e.g. --set-passphrase -> 'set_passphrase'
-PASSPHRASE_CLI_OPTION_NAMES = ["keys_root_path", "set_passphrase", "passphrase_file", "current_passphrase_file"]
 
 SAVE_MASTER_PASSPHRASE_WARNING = (
     colorama.Fore.YELLOW
--- a/chia/cmds/passphrase_funcs.py
+++ b/chia/cmds/passphrase_funcs.py
@@ -27,17 +27,6 @@ SAVE_MASTER_PASSPHRASE_WARNING = (
 )
 
 
-def remove_passphrase_options_from_cmd(cmd) -> None:
-    """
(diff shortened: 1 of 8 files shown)


chia/daemon/server.py
```diff
@@ -385,7 +383,6 @@ async def keyring_status(self) -> Dict[str, Any]:
         response: Dict[str, Any] = {
             "success": True,
             "is_keyring_locked": locked,
-            "passphrase_support_enabled": passphrase_support_enabled,
```
xdustinface, line 388: I guess this needs to be addressed in the GUI?

paninaro: Correct. I'll create a PR for the GUI to remove this dependency.

paninaro: https://github.com/Chia-Network/chia-blockchain-gui/pull/842

chia/cmds/passphrase_funcs.py
```diff
@@ -27,17 +27,6 @@
 )
```
paninaro, line 29: PASSPHRASE_CLI_OPTION_NAMES can be removed as well

paninaro:
Thanks for doing this. Looks good to me with the removal of `PASSPHRASE_CLI_OPTION_NAMES`

xdustinface:
@paninaro Dropped `PASSPHRASE_CLI_OPTION_NAMES`! But i didn't yet update the GUI because it looks like the NFT stuff isn't in `main` for `chia-blockchain-gui` but your PR was into `main`. I guess it breaks the GUI if we now merge this without your PR so we should wait until NFT gets merged into `main`?

paninaro:
> @paninaro Dropped `PASSPHRASE_CLI_OPTION_NAMES`! But i didn't yet update the GUI because it looks like the NFT stuff isn't in `main` for `chia-blockchain-gui` but your PR was into `main`. I guess it breaks the GUI if we now merge this without your PR so we should wait until NFT gets merged into `main`?

We'll update main on chia-blockchain-gui soon. It should be fine to check this change in now, it just means that the passphrase options won't show up in the GUI settings until we merge and update the submodule pin.

paninaro (approved):
lgtm

emlowe (approved):

PR #12491, opened 2022-07-19
daemon: Implement system metrics logger based on beta config

xdustinface:
Adds the class `BetaMetricsLogger` which gets instantiated in the daemon to periodically log some system metrics while in beta test mode.

The metrics logged in here now are a first draft from my side, feel free to provide suggestions about what we should log additionally or maybe remove from the current logs :) 

Based on:
- #12389

### DIFF
--- a/chia/cmds/beta.py
+++ b/chia/cmds/beta.py
@@ -12,9 +12,12 @@ from chia.cmds.beta_funcs import (
     prepare_plotting_log,
     prompt_beta_warning,
     prompt_for_beta_path,
+    prompt_for_metrics_log_interval,
     update_beta_config,
     validate_beta_path,
+    validate_metrics_log_interval,
 )
+from chia.util.beta_metrics import metrics_log_interval_default
 from chia.util.config import lock_and_load_config, save_config
 
 
@@ -29,8 +32,9 @@ def beta_cmd() -> None:
 
 @beta_cmd.command("configure", help="Configure the beta test mode parameters")
 @click.option("-p", "--path", help="The beta mode root path", type=str, required=False)
+@click.option("-i", "--interval", help="System metrics will be logged based on this interval", type=int, required=False)
 @click.pass_context
-def configure(ctx: click.Context, path: Optional[str]) -> None:
+def configure(ctx: click.Context, path: Optional[str], interval: Optional[int]) -> None:
     root_path = ctx.obj["root_path"]
     with lock_and_load_config(root_path, "config.yaml") as config:
         if "beta" not in config:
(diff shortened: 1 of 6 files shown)


chia/cmds/beta.py
```diff
@@ -29,8 +32,9 @@ def beta_cmd() -> None:
 
 @beta_cmd.command("configure", help="Configure the beta test mode parameters")
 @click.option("-p", "--path", help="The beta mode root path", type=str, required=False)
+@click.option("-i", "--interval", help="System metrics will be logged based on this interval", type=int, required=False)
 @click.pass_context
```
almogdepaz, line 36: what do you  think about making the interval be the default 
if not specified and losing the prompt ?

almogdepaz:
@xdustinface what do you think of doing default intervals instead 
2 intervals one for the  disc and ports and another for cpu etc 
losing the prompt and just let users add args if they want to set the intervals themselves ?
i think it will give better granularity and to me sounds simpler

xdustinface:
> @xdustinface what do you think of doing default intervals instead

There is already the default of 5s https://github.com/Chia-Network/chia-blockchain/pull/12491/files#diff-59f6723b04e9ed9e93f73da061ebdffd0e8dec3f0338756ecd86709c2987a312R15? For `chia beta enable` there is no prompt for it.

It's just for `chia beta configure` where it prompts for it, which is probably something most users anyway won't use but in case they do without params the prompts walk them through the possible settings.. any objections about that? 

> 2 intervals one for the disc and ports and another for cpu etc

Hm, so now we log the fast stuff in a range from 1s to 30s and the slow stuff in a range from 10s to 300s based on 1 interval setting. Of course i could add another interval but im not sure if we really need it or if there is much value in setting them separately because the idea was just to log some parts less frequent than the others. What do you think?

almogdepaz (approved):

almogdepaz:
> > @xdustinface what do you think of doing default intervals instead
> 
> There is already the default of 5s https://github.com/Chia-Network/chia-blockchain/pull/12491/files#diff-59f6723b04e9ed9e93f73da061ebdffd0e8dec3f0338756ecd86709c2987a312R15? For `chia beta enable` there is no prompt for it.
> 
> It's just for `chia beta configure` where it prompts for it, which is probably something most users anyway won't use but in case they do without params the prompts walk them through the possible settings.. any objections about that?
> 
> > 2 intervals one for the disc and ports and another for cpu etc
> 
> Hm, so now we log the fast stuff in a range from 1s to 30s and the slow stuff in a range from 10s to 300s based on 1 interval setting. Of course i could add another interval but im not sure if we really need it or if there is much value in setting them separately because the idea was just to log some parts less frequent than the others. What do you think?

thanks for clarifying that, i dont think its a big deal, depends on usage really and if people will want that customization

mariano54 (approved):

PR #12493, opened 2022-07-19
Support extend_derivation_index specifying the actual last index value

paninaro:
Previously, when `extend_derivation_index` was called, the index value was being passed to `create_more_puzzle_hashes`, which would add `initial_num_public_keys` to the index value. This has caused some confusion as the value provided by the user gets bumped up to a larger value than expected.

This change introduces an optional `num_additional_phs` param to `create_more_puzzle_hashes` that overrides the default behavior of adding `initial_num_public_keys` to the last index. `extend_derivation_index` specifies a value of 0 to prevent generating additional puzzle hashes beyond the specified index.

This also makes a small change to account for `range` not including the `last_index`. This was a bit confusing because the logged message made it seem that the range was inclusive.

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -272,7 +272,11 @@ class WalletStateManager:
         return pubkey, private
 
     async def create_more_puzzle_hashes(
-        self, from_zero: bool = False, in_transaction=False, up_to_index: Optional[uint32] = None
+        self,
+        from_zero: bool = False,
+        in_transaction=False,
+        up_to_index: Optional[uint32] = None,
+        num_additional_phs: Optional[int] = None,
     ):
         """
         For all wallets in the user store, generates the first few puzzle hashes so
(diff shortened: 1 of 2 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -281,7 +285,7 @@ async def create_more_puzzle_hashes(
         targets = list(self.wallets.keys())
         self.log.debug("Target wallets to generate puzzle hashes for: %s", repr(targets))
         unused: Optional[uint32] = (
-            up_to_index if up_to_index is not None else await self.puzzle_store.get_unused_derivation_path()
+            uint32(up_to_index + 1) if up_to_index is not None else await self.puzzle_store.get_unused_derivation_path()
```
paninaro, line 288: The +1 is to account for range() not including the stop value (`last_index` in this case)

wjblanke (approved):
aok

PR #12513, opened 2022-07-20
When extending the derivation index, make sure we don't mark previously

paninaro:
unused indices as used. This helps minimize gaps in the address space.


wjblanke (approved):
aok

PR #12517, opened 2022-07-21
util: Fix passphrase hint

xdustinface:
- [1114a55](https://github.com/Chia-Network/chia-blockchain/pull/12517/commits/1114a55b1b6b358b6c2fae0a77d5a8f3b4fba6e1) Fix for: Currently that the hint gets dropped if the keychain gets re-written after the hint has been set becaue the `passphrase_hint` isn't included in "to write payload" which gets created
https://github.com/Chia-Network/chia-blockchain/blob/11ddc8c27ef1410dcafb1e8d885fc549b6bbc18c/chia/util/file_keyring.py#L421-L426
It works the first time because it there gets injected here, just once: 
https://github.com/Chia-Network/chia-blockchain/blob/11ddc8c27ef1410dcafb1e8d885fc549b6bbc18c/chia/util/file_keyring.py#L428-L430
- [f3eb90e](https://github.com/Chia-Network/chia-blockchain/pull/12517/commits/f3eb90e4b08d48b60cb01f3cd0273b8f4d42b44b) Fix for: "hint removal" also doesn't work currently on main. The related test only worked because there is a write between the removal and the assertion.

See 5182ffde39089d45d042dd56f767ca33d4d55158 which is failing without 1114a55b1b6b358b6c2fae0a77d5a8f3b4fba6e1 and `test_passphrase_hint_removal` fails with 1114a55b1b6b358b6c2fae0a77d5a8f3b4fba6e1 only and passes with f3eb90e4b08d48b60cb01f3cd0273b8f4d42b44b on top.

### DIFF
--- a/chia/util/file_keyring.py
+++ b/chia/util/file_keyring.py
@@ -423,6 +423,7 @@ class FileKeyring(FileSystemEventHandler):
             "salt": self.salt.hex(),
             "nonce": nonce.hex(),
             "data": base64.b64encode(encrypted_inner_payload).decode("utf-8"),
+            "passphrase_hint": self.outer_payload_cache.get("passphrase_hint", None),
         }
 
         # Merge in other properties like "passphrase_hint"
--- a/chia/util/file_keyring.py
+++ b/chia/util/file_keyring.py
@@ -469,8 +469,4 @@ class FileKeyring(FileSystemEventHandler):
         Store the new passphrase hint in the staging dict (outer_payload_properties_for_next_write) to
         be written-out on the next write to the keyring.
         """
-        assert self.outer_payload_properties_for_next_write is not None
-        if passphrase_hint is not None and len(passphrase_hint) > 0:
-            self.outer_payload_properties_for_next_write["passphrase_hint"] = passphrase_hint
-        elif "passphrase_hint" in self.outer_payload_properties_for_next_write:
-            del self.outer_payload_properties_for_next_write["passphrase_hint"]
+        self.outer_payload_properties_for_next_write["passphrase_hint"] = passphrase_hint
--- a/tests/core/util/test_keyring_wrapper.py
+++ b/tests/core/util/test_keyring_wrapper.py
@@ -427,6 +427,12 @@ class TestKeyringWrapper:
         # Expect: to retrieve the passphrase hint that was just set
         assert KeyringWrapper.get_shared_instance().get_master_passphrase_hint() == "rhymes with bassphrase"
 
+        # When: writing the keyring again
+        KeyringWrapper.get_shared_instance().keyring.write_keyring()
+
+        # Expect: the hint is still set
+        assert KeyringWrapper.get_shared_instance().get_master_passphrase_hint() == "rhymes with bassphrase"
+
     @using_temp_file_keyring()
     def test_passphrase_hint_removal(self):
         """


chia/util/file_keyring.py
```diff
@@ -423,6 +423,7 @@ def write_keyring(self, fresh_salt: bool = False):
             "salt": self.salt.hex(),
             "nonce": nonce.hex(),
             "data": base64.b64encode(encrypted_inner_payload).decode("utf-8"),
+            "passphrase_hint": self.outer_payload_cache.get("passphrase_hint", None),
```
paninaro, line 426: This is fine, but I have another suggestion which would help us start preserving future keys/values that aren't currently present/understood:

```suggestion
        outer_payload = self.outer_payload_cache.copy()

        outer_payload.update({
            "version": 1,
            "salt": self.salt.hex(),
            "nonce": nonce.hex(),
            "data": base64.b64encode(encrypted_inner_payload).decode("utf-8"),
        })
```

paninaro (changes requested):
Thanks for finding and fixing this. Can you double check that calling `set_passphrase_hint` with None clears the passphrase hint from the cached dict and the keyring file? I see a test that checks if it's None after clearing, but I'm not sure if it gets written to the keyring file.

chia/util/file_keyring.py
xdustinface: Not sure if this is something we should go for in general because thats why we have versioning? And at least not in this PR because we anyway exit on load if the version is unrecognized https://github.com/Chia-Network/chia-blockchain/blob/1910c3891c425a4a50c83f310b134fe528867f49/chia/util/file_keyring.py#L350-L355

chia/util/file_keyring.py
```diff
@@ -468,8 +469,4 @@ def set_passphrase_hint(self, passphrase_hint: Optional[str]) -> None:
         Store the new passphrase hint in the staging dict (outer_payload_properties_for_next_write) to
         be written-out on the next write to the keyring.
         """
-        assert self.outer_payload_properties_for_next_write is not None
-        if passphrase_hint is not None and len(passphrase_hint) > 0:
-            self.outer_payload_properties_for_next_write["passphrase_hint"] = passphrase_hint
-        elif "passphrase_hint" in self.outer_payload_properties_for_next_write:
-            del self.outer_payload_properties_for_next_write["passphrase_hint"]
+        self.outer_payload_properties_for_next_write["passphrase_hint"] = passphrase_hint
```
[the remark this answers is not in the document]
xdustinface: Correct this in combination with 1114a55b1b6b358b6c2fae0a77d5a8f3b4fba6e1 sets the `passphrase_hint` to `None` in the file if there is no hint set. The `del` expression did only remove from `outer_payload_properties_for_next_write` not from `outer_payload_cache` thats why the removal was broken also. Sure we could add another `if "passphrase_hint" in self.outer_payload_cache: del ...` but then, if the keyring gets reloaded between the call of `set_passphrase_hint(None)` and the next write it just gets reloaded into the cache  and we end up not removing it. We could also drop it before writing based on it being `None` in `outer_payload_properties_for_next_write` but i anyway have another PR coming which leads to all fields always being present in the file so i think we can just leave it like this unless there are any other objections.

xdustinface:
> Can you double check that calling `set_passphrase_hint` with None clears the passphrase hint from the cached dict and the keyring file? I see a test that checks if it's None after clearing, but I'm not sure if it gets written to the keyring file.

Like explained here https://github.com/Chia-Network/chia-blockchain/pull/12517#discussion_r926549088 the removal works yeah but it's not actually being removed from the file, it's just set to `None`, which is basically the same?

chia/util/file_keyring.py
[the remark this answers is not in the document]
paninaro: Ok that sounds fine. I was thinking it would be nice to prune `passphrase_hint` from keyring.yaml if the hint was removed, but setting it to None should be fine.

chia/util/file_keyring.py
paninaro: Yeah, basically I was thinking it could be useful to add more flexibility around unknown keys by leaving them intact. This could give us the flexibility to make minor adjustments/additions to keyring.yaml _without_ bumping the version.

It would be nice to not have to worry about keyring migration until it's absolutely necessary, so little changes that give us some wiggle room are worth considering.

paninaro (approved):
lgtm

emlowe (approved):

PR #12521, opened 2022-07-21
Move simulator test requirements over to new module 

jack60612:
This simulator requires this.
This also includes some minor simulator changes 
This moves blocktools to the chia/simulator dir


arvidn:
this is a very large change with a very minimalist explanation.

It looks like the bulk of the changes are linting fixes. I think those changes should not be mixed in with any other changes, but be their own PR.

The other change is moving `block_tools.py` and `wallet_tools.py` into `simulator`. Why do you do that? Are those only used in simulations? I think that should also be done in a separate PR.

That way it makes it very easy to review the actual changes you made to the simulator. Those are buried in noise right now

jack60612:
> this is a very large change with a very minimalist explanation.
> 
> It looks like the bulk of the changes are linting fixes. I think those changes should not be mixed in with any other changes, but be their own PR.
> 
> The other change is moving `block_tools.py` and `wallet_tools.py` into `simulator`. Why do you do that? Are those only used in simulations? I think that should also be done in a separate PR.
> 
> That way it makes it very easy to review the actual changes you made to the simulator. Those are buried in noise right now

Messaged you on keybase, but lint changes were necessary to pass ci.

mariano54 (approved):

trepca (approved):

PR #12523, opened 2022-07-21
Improve CI performance

ChiaMineJP:
Proof of concept branch to improve CI performance.

### DIFF
--- a/.github/workflows/build-linux-installer-deb.yml
+++ b/.github/workflows/build-linux-installer-deb.yml
@@ -168,7 +168,7 @@ jobs:
       id: cache-gui
       with:
         path: ./chia-blockchain-gui
-        key: ${{ runner.os }}-chia-blockchain-gui-${{ steps.gui-ref.outputs.GUI_REF }}
+        key: ${{ runner.os }}-deb-chia-blockchain-gui-${{ steps.gui-ref.outputs.GUI_REF }}
 
     - if: steps.cache-gui.outputs.cache-hit != 'true'
       name: Build GUI
--- a/.github/workflows/build-linux-installer-deb.yml
+++ b/.github/workflows/build-linux-installer-deb.yml
@@ -154,14 +154,36 @@ jobs:
       run: |
         sudo apt-get install -y jq
 
+    - name: Prepare GUI cache
+      id: gui-ref
+      run: |
+        gui_ref=$(git submodule status chia-blockchain-gui | sed -e 's/^ //g' -e 's/ chia-blockchain-gui.*$//g')
+        echo "${gui_ref}"
+        echo "::set-output name=GUI_REF::${gui_ref}"
+        echo "rm -rf ./chia-blockchain-gui"
+        rm -rf ./chia-blockchain-gui
+
+    - name: Cache GUI
+      uses: actions/cache@v3
+      id: cache-gui
+      with:
+        path: ./chia-blockchain-gui
+        key: ${{ runner.os }}-chia-blockchain-gui-${{ steps.gui-ref.outputs.GUI_REF }}
+
+    - if: steps.cache-gui.outputs.cache-hit != 'true'
+      name: Build GUI
+      continue-on-error: false
+      run: |
+        cd ./build_scripts
+        bash build_linux_deb-1-gui.sh
+
     - name: Build .deb package
       env:
         CHIA_INSTALLER_VERSION: ${{ steps.version_number.outputs.CHIA_INSTALLER_VERSION }}
       run: |
         ldd --version
-        git -C ./chia-blockchain-gui status
         cd ./build_scripts
-        sh build_linux_deb.sh amd64
+        sh build_linux_deb-2-installer.sh amd64
 
     - name: Upload Linux artifacts
       uses: actions/upload-artifact@v3
@@ -241,3 +263,7 @@ jobs:
       if: startsWith(github.ref, 'refs/tags/')
       run: |
         curl -s -XPOST -H "Authorization: Bearer ${{ secrets.GLUE_ACCESS_TOKEN }}" --data '{"chia_ref": "${{ steps.tag-name.outputs.TAG_NAME }}"}' ${{ secrets.GLUE_API_URL }}/api/v1/${{ steps.tag-name.outputs.REPO_NAME }}/${{ steps.tag-name.outputs.TAG_NAME }}/success/build-linux-deb
+
+    - name: Remove working files to exclude from cache
+      run: |
+        rm -rf ./chia-blockchain-gui/packages/gui/daemon
--- a/build_scripts/build_windows-2-installer.ps1
+++ b/build_scripts/build_windows-2-installer.ps1
@@ -4,7 +4,7 @@ $ErrorActionPreference = "Stop"
 
 mkdir build_scripts\win_build
 
-git status
+git submodule
 
 if (-not (Test-Path env:CHIA_INSTALLER_VERSION)) {
   $env:CHIA_INSTALLER_VERSION = '0.0.0'
@@ -30,35 +30,17 @@ Write-Output "   ---"
 Set-Location -Path ".\npm_windows" -PassThru
 npm ci
 $Env:Path = $(npm bin) + ";" + $Env:Path
-Set-Location -Path "..\" -PassThru
 
-Set-Location -Path "..\chia-blockchain-gui" -PassThru
+Set-Location -Path "..\..\chia-blockchain-gui" -PassThru
 # We need the code sign cert in the gui subdirectory so we can actually sign the UI package
 If ($env:HAS_SECRET) {
-    Copy-Item "win_code_sign_cert.p12" -Destination "packages\gui\"
+    Copy-Item "..\win_code_sign_cert.p12" -Destination "packages\gui\"
 }
-
-git status
 
 Write-Output "   ---"
 Write-Output "Prepare Electron packager"
 Write-Output "   ---"
 $Env:NODE_OPTIONS = "--max-old-space-size=3000"
-
-lerna clean -y
-npm ci
-# Audit fix does not currently work with Lerna. See https://github.com/lerna/lerna/issues/1663
-# npm audit fix
-
-git status
-
-Write-Output "   ---"
-Write-Output "Electron package Windows Installer"
-Write-Output "   ---"
-npm run build
-If ($LastExitCode -gt 0){
-    Throw "npm run build failed!"
-}
 
 # Change to the GUI directory
 Set-Location -Path "packages\gui" -PassThru
@@ -85,15 +67,20 @@ Write-Output "   ---"
 
 Write-Output "   ---"
 Write-Output "electron-packager"
-electron-packager . Chia --asar.unpack="**\daemon\**" --overwrite --icon=.\src\assets\img\chia.ico --app-version=$packageVersion
+electron-packager . Chia --asar.unpack="**\daemon\**" `
+--overwrite --icon=.\src\assets\img\chia.ico --app-version=$packageVersion `
+--no-prune --no-deref-symlinks `
+--ignore="/node_modules/(?!ws(/|$))(?!@electron(/|$))" --ignore="^/src$" --ignore="^/public$"
+# Note: `node_modules/ws` and `node_modules/@electron/remote` are dynamic dependencies
+# which GUI calls by `window.require('...')` at runtime.
+# So `ws` and `@electron/remote` cannot be ignored at this time.
+dir Chia-win32-x64\resources
 Write-Output "   ---"
 
 Write-Output "   ---"
 Write-Output "node winstaller.js"
 node winstaller.js
 Write-Output "   ---"
-
-git status
 
 If ($env:HAS_SECRET) {
    Write-Output "   ---"
@@ -104,8 +91,6 @@ If ($env:HAS_SECRET) {
    }   Else    {
    Write-Output "Skipping timestamp and verify signatures - no authorization to install certificates"
 }
-
-git status
 
 Write-Output "   ---"
 Write-Output "Moving final installers to expected location"
--- a/build_scripts/build_windows-2-installer.ps1
+++ b/build_scripts/build_windows-2-installer.ps1
@@ -74,7 +74,7 @@ electron-packager . Chia --asar.unpack="**\daemon\**" `
 # Note: `node_modules/ws` and `node_modules/@electron/remote` are dynamic dependencies
 # which GUI calls by `window.require('...')` at runtime.
 # So `ws` and `@electron/remote` cannot be ignored at this time.
-dir Chia-win32-x64\resources
+Get-ChildItem Chia-win32-x64\resources
 Write-Output "   ---"
 
 Write-Output "   ---"
--- a/build_scripts/build_windows-2-installer.ps1
+++ b/build_scripts/build_windows-2-installer.ps1
@@ -4,6 +4,7 @@ $ErrorActionPreference = "Stop"
 
 mkdir build_scripts\win_build
 
+git status
 git submodule
 
 if (-not (Test-Path env:CHIA_INSTALLER_VERSION)) {
--- a/.github/workflows/build-linux-arm64-installer.yml
+++ b/.github/workflows/build-linux-arm64-installer.yml
@@ -114,14 +114,36 @@ jobs:
 
     - uses: chia-network/actions/activate-venv@main
 
+    - name: Prepare GUI cache
+      id: gui-ref
+      run: |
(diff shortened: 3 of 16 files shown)


ChiaMineJP:
## Actions which took the longest time for each CI jobs. (n=1)
|Job|Action|Time(sec)|
|----|-------|------:|
|Linux amd64 DEB Installer (3.8, ubuntu-18.04)|Build .deb package|927|
|Linux arm64 DEB Installer (3.8)|Build arm64 .deb package|884|
|Linux amd64 RPM Installer (3.9)|Build .rpm package|991|
|MacOS Intel Installer (3.9, macos-10.15)|Build MacOS DMG|1,863|
|MacOS arm64 installer (3.9)|Build MacOS DMG|737|
|Windows 10 Installer (3.9)|Build Windows installer with build_scripts\build_windows.ps1|2,028|

## Windows installer - 3.9 (Total 2,028s)
- `pyinstaller` took 07m02s
- `npm ci` on `chia-blockchain-gui`  took 03m18s
- `npm build` on `chia-blockchain-gui`  took 10m04s
- `electron-packager` took 09m51s

## MacOS Intel installer - 3.9 (Total 1,863s)
- `pyinstaller` took 04m16s
- `npm ci` on `chia-blockchain-gui`  took 02m26s
- `npm build` on `chia-blockchain-gui`  took 09m32s
- `electron-packager` took 06m08s
- `notalize-cli` took 06m20s

## Linux amd64 RPM installer - 3.9 (Total 991s)
- `pyinstaller` took 03m26s
- `npm ci` on `chia-blockchain-gui`  took 01m26s
- `npm build` on `chia-blockchain-gui`  took 06m39s
- `electron-packager` took 02m42s
- `electron-installer-redhat` took 01m34s

## Linux amd64 DEB installer - 3.8, ubntu-18.04 (Total 927s)
- `pyinstaller` took 03m04s
- `npm ci` on `chia-blockchain-gui`  took 01m08s
- `npm build` on `chia-blockchain-gui`  took 05m50s
- `electron-packager` took 02m38s
- `electron-installer-debia` took 02m06s

## Linux arm64 DEB installer - 3.8 (Total 884s)
- `pyinstaller` took 03m00s
- `npm ci` on `chia-blockchain-gui`  took 01m04s
- `npm build` on `chia-blockchain-gui`  took 06m34s
- `electron-packager` took 02m09s
- `electron-installer-debia` took 01m22s

## MacOS arm64 installer - 3.9 (Total 737s)
- `pyinstaller` took 01m32s
- `npm ci` on `chia-blockchain-gui`  took 00m34s
- `npm build` on `chia-blockchain-gui`  took 02m29s
- `electron-packager` took 01m28s
- `notalize-cli` took 05m08s

ChiaMineJP:
# Ranking of heavy CI jobs
Reference PR: https://github.com/Chia-Network/chia-blockchain/pull/12550
(Jobs which took more than or equal to 900s)

|Rank|Job name|Time(sec)|Time(min)|
|-----:|----------|----------:|----------:|
|1|📦🚀 Build Installer - Windows 10 / Windows 10 Installer (3.9)|2,520 sec|42 min|
|2|📦🚀 Build Installer - MacOS Intel / MacOS Intel Installer (3.9, macos-10.15)|2,040 sec|34 min|
|3|📦🚀 Build Installer - Linux RPM AMD64 / Linux amd64 RPM Installer (3.9)|1,440 sec|24 min|
|4|🧪 test / windows / 🪟 blockchain - 3.8|1,320 sec|22 min|
|5|🧪 test / windows / 🪟 blockchain - 3.10|1,320 sec|22 min|
|6|📦🚀 Build Installer - Linux DEB AMD64 / Linux amd64 DEB Installer (3.8, ubuntu-18.04)|1,260 sec|21 min|
|7|🧪 test / windows / 🪟 wallet.nft_wallet - 3.8|1,260 sec|21 min|
|8|🧪 test / macos / 🍎 wallet.nft_wallet - 3.9|1,200 sec|20 min|
|9|🧪 test / macos / 🍎 wallet.nft_wallet - 3.10|1,200 sec|20 min|
|10|🧪 test / windows / 🪟 blockchain - 3.9|1,200 sec|20 min|
|11|🧪 test / windows / 🪟 wallet.sync - 3.8|1,200 sec|20 min|
|12|🧪 test / windows / 🪟 core.full_node.full_sync - 3.10|1,140 sec|19 min|
|13|🧪 test / windows / 🪟 core.full_node.full_sync - 3.8|1,080 sec|18 min|
|14|🧪 test / windows / 🪟 core.full_node.full_sync - 3.9|1,080 sec|18 min|
|15|🧪 test / windows / 🪟 wallet.nft_wallet - 3.7|1,080 sec|18 min|
|16|🧪 test / windows / 🪟 wallet.sync - 3.7|1,080 sec|18 min|
|17|📦🚀 Build Installer - Linux DEB ARM64 / Linux arm64 DEB Installer (3.8)|1,020 sec|17 min|
|18|🧪 test / windows / 🪟 wallet.cat_wallet - 3.7|1,020 sec|17 min|
|19|🧪 test / windows / 🪟 wallet.nft_wallet - 3.9|1,020 sec|17 min|
|20|🧪 test / windows / 🪟 wallet.nft_wallet - 3.10|1,020 sec|17 min|
|21|🧪 test / macos / 🍎 wallet.sync - 3.9|960 sec|16 min|
|22|🧪 test / macos / 🍎 wallet.cat_wallet - 3.10|900 sec|15 min|
|23|🧪 test / windows / 🪟 blockchain - 3.7|900 sec|15 min|
|24|🧪 test / windows / 🪟 core.full_node.full_sync - 3.7|900 sec|15 min|
|25|🧪 test / windows / 🪟 wallet.cat_wallet - 3.9|900 sec|15 min|
|26|🧪 test / windows / 🪟 wallet.rpc - 3.8|900 sec|15 min|
|27|🧪 test / windows / 🪟 wallet.sync - 3.9|900 sec|15 min|
|28|LGTM analysis: Python|900 sec|15 min|

ChiaMineJP:
# Core problems found
- `chia-blockchain-gui` rebuilds every time even if it does not get updated.
- More than 80% of asar(app source archive) file generated by `electron-packager` is not used.

## `chia-blockchain-gui` rebuilds every time even if it does not get updated.
`chia-blockchain` and `chia-blockchain-gui` are basically separate modules.
Although they are not yet completely independent from each other right now, I've found that rebuilding `chia-blockchain-gui` is not necessary if a submodule ref to `chia-blockchain-gui` specified in `chia-blockchain` does not change.
What makes it difficult to isolate both modules are:
- The `version` property of `packages/gui/package.json` in `chia-blockchain-gui` needs to be in sync with `CHIA_INSTALLER_VERSION`.  
  The version string is used just to display info. Also I checked whether the version string is stamped/embedded to compiled JavaScript output but I confirmed it is not.  
  This means branding `CHIA_INSTALLER_VERSION` to `package.json` can be decoupled from `npm run build` process. It is OK to set `version` after `npm run build` completes or restored from cache.
- Artifact of `chia-blockchain` compiled by `pyinstaller` needs to be copied to `packages/gui/`  
  This is one of the primary intersection of `chia-blockchain` and `chia-blockchain-gui`.  
  The `daemon` folder built by `pyinstaller` will be copied into `packages/gui/` and later the folder is used for packaging.  
  I've confirmed that the `daemon` folder copied into `packages/gui/` folder is not a prerequisite for
  `npm ci` nor `npm run build`. So copying the folder can be safely decoupled to npm build process.

## More than 80% of asar(app source archive) file generated by `electron-packager` is not used.
I extracted whole asar file and found that `app.asar` contains `node_modules` and other unused source files.  
So excluding `packages/gui/{src, node_modules}/**` from asar file may greatly reduce the execution time and package size of `electron-packager` command.

wallentx:
😍

altendky:
Thanks for digging into this!

ChiaMineJP:
# The results of optimizations

#### Build Installer - Windows 10 / Windows 10 Installer (3.9)
- Reduced 27 minutes (59%) in job execution time
  - 46mins => 19mins
- Reduced 69MB (28%) in installer size
  - 248MB => 178MB

#### Build Installer - MacOS Intel / MacOS Intel Installer (3.9, macos-10.15)
- Reduced 18 minutes (49%) in job execution time
  - 37mins => 19mins
- Reduced 72MB (30%) in installer size
  - 239MB => 167MB

#### Build Installer - Linux RPM AMD64 / Linux amd64 RPM Installer (3.9)
- Reduced 10 minutes (50%) in job execution time
  - 20mins => 10mins
- .rpm installer does not contain GUI so installer size gains no improvement.

#### Build Installer - Linux DEB AMD64 / Linux amd64 DEB Installer (3.8, ubuntu-18.04)
- Reduced 9 minutes (47%) in job execution time
  - 19mins => 10mins
- .deb installer does not contain GUI so installer size gains no improvement.

#### Build Installer - Linux DEB ARM64 / Linux arm64 DEB Installer (3.8)
- Reduced 10 minutes (59%) in job execution time
  - 17mins => 7mins
- .deb installer does not contain GUI so installer size gains no improvement.

#### Build Installer - MacOS arm64 / MacOS arm64 installer (3.9)
- Reduced 4 minutes (27%) in job execution time
  - 15mins => 11mins
- Reduced 72MB (30%) in installer size
  - 243MB => 171MB

ChiaMineJP:
# Important notice for GUI developers
CC: @seeden @paninaro (Reviewers for GUI)

## TLDR;
Please consider to avoid using dynamic (runtime) package loading via `window.require('...')` on `chia-blockchain-gui`, or modify `build_xxxx-2-installer.sh/ps1` so that `electron-packager` command can include runtime dependencies to `app.asar`. (Dynamic loading seems to be required to use nodejs feature but then you should `import` in electron code part and call them via IPC from renderer. This ensures that relevant libraries in `node_modules` will be compiled into `.../gui/build/electron/`)
```bash
electron-packager . Chia --asar.unpack="**/daemon/**" --platform=darwin \
--icon=src/assets/img/Chia.icns --overwrite --app-bundle-id=net.chia.blockchain \
--appVersion=$CHIA_INSTALLER_VERSION \
--no-prune --no-deref-symlinks \
--ignore="^/src$" --ignore="^/public$" \
--ignore="/node_modules/(?!ws(/|$))(?!@electron(/|$))"
```
For example, If you want to add `some-package` for dynamic loading, please modify `--ignore` part like this:
```bash
--ignore="/node_modules/(?!ws(/|$))(?!@electron(/|$))(?!some-package(/|$))"
```

## Details
In order to reduce file size of `app.asar`, almost all files in `chia-blockchain-gui/packages/gui/node_modules` will be excluded from `app.asar` with a few exceptions in this PR (https://github.com/Chia-Network/chia-blockchain/pull/12584).
Since almost all packages in the `node_modules` are usually compiled into `.../gui/build/{electron, renderer}`, copying `.../gui/node_modules/**` into `app.asar` along with `.../gui/build/**` makes the installer unnecessarily fat. (Around 4 times larger in size)
So I decided to exclude `.../gui/node_modules` from the final output.

But I found that few npm packages are dynamically loaded by `window.require('...')` at runtime of GUI.

- .../gui/src/hooks/useSelectDirectory.tsx  
```typescript
    if (isElectron()) {
      const {dialog} = window.require('@electron/remote');
      // @ts-ignore
      const result = await dialog.showOpenDialog({
        properties: ['openDirectory', 'showHiddenFiles'],
        ...defaultOptions,
        ...options,
      });
      const filePath = result.filePaths[0];

      return filePath;
    }
```
- .../gui/src/components/app/AppProviders.tsx
```typescript
export default function App(props: AppProps) {
  const { children, outlet } = props;
  const [isReady, setIsReady] = useState<boolean>(false);
  const { isDarkMode } = useDarkMode();

  const theme = isDarkMode ? dark : light;

  async function init() {
    const config = await waitForConfig();
    const { cert, key, url } = config;
    const WS = window.require('ws');

    store.dispatch(api.initializeConfig({
      url,
      cert,
      key,
      webSocket: WS,
    }));
```

Those packages loaded at runtime are not to be compiled into `.../gui/build/{electron, renderer}`, but loaded from `node_modules/` compressed in a `app.asar` file.
So at this time, `ws` and `@electron` npm packages cannot be added to `--ignore` option to `electron-packager` command described above.

After a few investigation, I found a way to add allow-list to include them to `app.asar` as described in TLDR part.
So, please consider modifying CI build scripts when you want to add dynamic loading via `window.require(...)`.

.github/workflows/build-linux-installer-deb.yml
```diff
@@ -154,14 +154,36 @@ jobs:
...
+        echo "::set-output name=GUI_REF::${gui_ref}"
+        echo "rm -rf ./chia-blockchain-gui"
+        rm -rf ./chia-blockchain-gui
+
+    - name: Cache GUI
+      uses: actions/cache@v3
+      id: cache-gui
+      with:
+        path: ./chia-blockchain-gui
+        key: ${{ runner.os }}-chia-blockchain-gui-${{ steps.gui-ref.outputs.GUI_REF }}
```
cmmarslender, line 171: I think this will share the same runner.os as the rpm build (linux). Should these be separated into variants with `-rpm` and `-deb` ?

build_scripts/build_windows-2-installer.ps1
```diff
@@ -4,7 +4,7 @@ $ErrorActionPreference = "Stop"
 
 mkdir build_scripts\win_build
 
-git status
+git submodule
```
cmmarslender, line 7: Thinking this was maybe an unintentional change?

.github/workflows/build-linux-installer-deb.yml
ChiaMineJP: I compared diff of GUI build output between fedora/ubuntu and found that checksum of `chia-blockchain-gui/packages/gui/build/renderer/loadable-stats.json` is different.
This is because `loadable-stats.json` contains full path to a js file and this is rather install path dependant than rpm/deb.
But yes, your suggestion makes sense. Adding `-rpm` and `-deb` to cache key indirectly resolves this issue.

build_scripts/build_windows-2-installer.ps1
ChiaMineJP: Adding `git submodule` was intentional, but deleting `git status` was unintentional...
I'll fix this.

paninaro:
Thanks for the in-depth investigation @ChiaMineJP! I'm especially happy to see the reduction in the overall app size.

I think this approach of only rebuilding if the chia-blockchain-gui submodule is changed (or if the cached version isn't available) makes sense. There are some places where the JS code will look at package.json, but not for the version string, so I think we're safe to update the version in package.json _after_ running `npm build`.

Regarding the window.require usage, it looks like we can safely replace those with standard imports. I have a PR open with this change: https://github.com/Chia-Network/chia-blockchain-gui/pull/866

@seeden if you're ok with this approach, I think we can restore @wallentx's sanity by taking these changes :)

altendky:
Presumably as a separate PR, but could we move the build into the GUI repo?  If we can cache it then it would seem like we could.  It seems like that would introduce a more explicit layer than a CI cache which would enhance control of rebuilding when we decide a dependency update deserves inclusion.  In the mean time there should be an explicit discussion about this and how this change in build behavior will affect the set of dependencies we get and whether that change matters.

All that said...  :tada:  thanks for improving the developer experience!

ChiaMineJP:
Before mergint this PR to `main`, if the submodule ref to `chia-blockchain-gui` points to its current main, the whole `chia-blockchain-gui/packages/gui/node_modules` can be excluded from `app.asar`. (Additinal reduction in `app.asar` file size can be expected)
Currently `ws` and `@electron` are included to `app.asar` because they were loaded via `window.require()` instead of `import`.

seeden:
@ChiaMineJP, very nice investigation and improvements. I have a few comments:
Our electron application has two parts: main and renderer.
Renderer is building and combining everything from src and node_modules together, and it is safe to remove node_modules for this part because it is a pure "web" application without any binary files.
Main electron process is more problematic because we are excluding all node_modules packages, and therefore, you needed to add "ws" and "@electron." This can be fixed when we remove **exclude: node_modules** from **packages/gui/webpack.electron.babel.ts**. After this change, you can altogether remove node_modules as you want to do.

What I see as a problem in the future is binary content. Currently, we are not using any package that uses compiled binary files internally. But it can be changed in the future, and the whole build process will be broken because webpack will not import these files. For example, @paninaro aro wanted to use FFmpeg. One solution can be to do it during the webpack build process and copy necessary binaries to /build directory. We can do it inside the webpack config.

ChiaMineJP:
@seeden Copying binary dependencies to `/build` directory is good. My current PR just excludes `/node_modules`, `/src` and `/public` so `/build` will be included automatically.

mariano54:
I'm so excited to get this merged. Thank you @ChiaMineJP !

cmmarslender (approved):
Changes look like they should be good now. Tested m1 mac and UI all seems to be working.

wjblanke (approved):
aok

wallentx (approved):
LGTM

PR #12531, opened 2022-07-21
util: Enable `isort` and more `mypy` in `file_keyring.py`

xdustinface:

PR #12541, opened 2022-07-22
util: Unify lockfile handling 

xdustinface:
This patch:
- Implements `chia.util.lock.Lockfile`, a wrapper around `FileLock`
- Unifies all places where we create lock files to use `chia.util.lock.Lockfile`
- Changes the daemon lockfile from `start-daemon.launching` to `daemon.lock` like we have it with all `<service>.lock` in the `run` folder
- Moves the CLVM compile lockfiles into a temp directory
- Drops the `fastener` dependency

Based on:
- #12531


arvidn:
> Changes the daemon lockfile from start-daemon.launching to daemon.lock like we have it with all <service>.lock in the run folder

do you think we should do anything about backwards compatibility when you have an previous-version deamon running?

arvidn:
it looks like we only use the daemon lockfile to serialize opening of the listen port. and then we use the listen port as the "lock" of ensuring we're just running a single daemon I assume

xdustinface:
Yeah so you get an 

```
OSError: [Errno 48] error while attempting to bind on address ('::1', 55401, 0, 0): address already in use
```

if you run `chia run_daemon` after the upgrade with the daemon still running since in that case it succeeds to acquire the new lockfile (what i assume not many will do and if so they can just stop the running daemon) but its all working fine if you just use the usual start commands since it tries to connect to the daemon then in the first place and only tries to start it if the connection fails.

xdustinface:
This is ready now. The `poll_interval=0.2` in `file_keyring.py` is because of the tests with 20 processes in `test_file_keyring_synchronization.py` failed on CI with the default poll interval of 50ms. Other option would be to refactor the tests i guess but the 200ms poll interval should be fine too?

chia/daemon/server.py
```diff
@@ -1181,15 +1176,15 @@ def daemon_launch_lock_path(root_path: Path) -> Path:
...
 
 
 def service_launch_lock_path(root_path: Path, service: str) -> Path:
     """
-    A path to a file that is lock when a service is running.
+    A path that is locked when a service is running.
     """
     service_name = service.replace(" ", "-").replace("/", "-")
-    return root_path / "run" / f"{service_name}.lock"
+    return Path(root_path / "run" / service_name)
```
arvidn, line 1187: passing this to the `Path()` constructor here looks redundant. `root_path` is already a `Path` object, right? that's why `/` works as path concatenation. Presumably the resulting value from `/` is also a `Path`, right?

chia/daemon/server.py
```diff
@@ -1388,30 +1359,30 @@ async def async_run_daemon(root_path: Path, wait_for_unlock: bool = False) -> in
...
+                root_path,
+                ca_crt_path,
+                ca_key_path,
+                crt_path,
+                key_path,
+                shutdown_event,
+                run_check_keys_on_unlock=wait_for_unlock,
+            )
+            await ws_server.start()
+            await shutdown_event.wait()
```
arvidn, line 1378: if I'm reading this correctly, the call to `wait()` here will suspend this co-routine until we're shutting down, right?
If so, this lock-file won't be released when we're done launching, but when we shut down.

it doesn't look right.

tests/core/util/test_file_keyring_synchronization.py
```diff
@@ -108,27 +108,26 @@ def child_writer_dispatch_with_readiness_check(
...
+                result = func()
+                return result
+        except Exception as e:
+            log.warning(
+                f"[pid:{os.getpid()}] caught exception in child_writer_dispatch_with_readiness_check: "
+                f"type: {type(e)}, {e}"
+            )
+            continue
+        finally:
+            if attempts == 0:
```
arvidn, line 125: this makes the condition in `while attempts > 0` unnecessary (and a bit misleading). This is where this loop exits, or at `return result`.

tests/core/util/test_file_keyring_synchronization.py
```diff
@@ -108,27 +108,26 @@ def child_writer_dispatch_with_readiness_check(
...
+                raise LockfileError()
+            # Write out a file indicating this process has completed its work
+            finished_file_path: Path = finished_dir / f"{os.getpid()}.finished"
+            with open(finished_file_path, "w") as f:
+                f.write(f"{os.getpid()}\n")
```
arvidn, line 130: this looks a bit odd. You do this every time we fail to lock the file.
The `finally` block is executed right after the `continue` on line 123, right?

arvidn (changes requested):
this is definitely an improvement. I think you should write a test for the `LockFile` to ensure it works on all platforms we test, and also that it actually behaves the way we expect. for example with regards to writing the complete file.

It doesn't look like it's correct, and if so those cases should be exercised by the test.

chia/daemon/server.py
xdustinface: It was the same before, wasn't it? I mean, `lockfile` still lived when the `wait()` was called or do i miss something?

arvidn: actually, it's unclear to me what `singleton` was supposed to do. It doesn't look like it's a lock file at all on win32 and cygwin. It deletes the file if it's there, and then creates a new one.

so, you might be right. But it seems to me that the *intention* was to only create a critical section for the startup, before opening the listen socket.

However, it's also not clear to me why the socket itself couldn't be used as the system-wide lock.

tests/core/util/test_file_keyring_synchronization.py
xdustinface: Thanks, addressed this and the below!

chia/daemon/server.py
xdustinface: > It doesn't look like it's a lock file at all on win32 and cygwin. It deletes the file if it's there, and then creates a new one.

Hm, maybe it raises the `IOError` in case the file is already locked.. should get my windows back up at some point..

> it's also not clear to me why the socket itself couldn't be used as the system-wide lock.

Hm i also wondered about that already but maybe to allow better error reporting since we know if the file is locked the daemon is running. But we don't know just from the fact that the port is bound if it's the daemon or something else unless we query something from it or so? Or is there another way to get that from the socket? Anyway, thats rather something for another PR i guess.

chia/daemon/server.py
```diff
@@ -1181,15 +1176,15 @@ def daemon_launch_lock_path(root_path: Path) -> Path:
     A path to a file that is lock when a daemon is launching but not yet started.
     This prevents multiple instances from launching.
     """
-    return root_path / "run" / "start-daemon.launching"
```
arvidn, line 1184: the name `.launching` here suggests that this lock file was intended to only be used while starting the daemon

chia/util/lock.py
```diff
@@ -1,44 +1,47 @@
...
+    pass
 
-        f.write("lock")
-    except IOError:
-        return None
 
-    return f
+@final
+@dataclass(frozen=True)
+class Lockfile:
```
arvidn, line 18: I think this code would be a lot simpler (and possibly more robust) if you write it as a single function with the `@contextmanager` attribute.

Then you could also use the `FileLock.acquire()` as a context manager as well.

Something like this:

```
@contextlib.contextmanager
def Lockfile(path: Path, timeout: float = -1, poll_interval: float = 0.05) -> None:
    fl = FileLock(path.with_name(path.name + ".lock"))
    try:
        using fl.aquire(timeout=timeout, poll_interval=poll_interval):
            yield
    except Timeout as e:
        raise LockfileError(e) from e
```

do you need the manual `acquire()` and `release()` calls also?

xdustinface: > I think this code would be a lot simpler (and possibly more robust) if you write it as a single function with the `@contextmanager` attribute.

Thats what i had initially.

> Then you could also use the `FileLock.acquire()` as a context manager as well.

You can also just use the `FileLock` as context manager like:

```
with FileLock(path.with_name(path.name + ".lock", timeout=timeout, poll_interval=poll_interval):
    yield
```

> do you need the manual `acquire()` and `release()` calls also?

But thats ^ why i added the wrapping class because we `release` inside the context in some tests and also to make sure we have the `FileLock` creation wrapper and the exception thrown for timeouts unified and imported from one file. However, im open for suggestions to get around this with a single function `contextmanager`.

chia/daemon/server.py
xdustinface: Yeah sure but whats the point? And whats wrong with keeping the lock while it's running? @richardkiss can you explain the intention since you implemented this here https://github.com/Chia-Network/chia-blockchain/pull/217?

chia/daemon/server.py
richardkiss: I wrote this a long time ago, so I don't remember everything. But I believe there were cases where listening on an already-listened socket did *not* fail, so it was not an effective locking mechanism. It also evolved from using unix sockets where supported (ie. everything but Windows) to a tcp socket.

chia/daemon/server.py
richardkiss: I think maybe the idea was, using unix sockets, you could delete the unix socket and the daemon would notice and gracefully shut down. (Obviously that doesn't work anymore since you can't delete a tcp socket.)

arvidn:
using tcp sockets as a mutex to only have one process running is a bit complicated on windows where you can't (by default) bind a new socket to the same port as a socket that was recently closed, and is in `TIMED_WAIT` mode. This made a lot of people set the `REUSEADDR` flag on windows, which means you can no longer use it as a mutex. Starting in vista (or something like that) microsoft added another flag to mean "make REUSEADDR have the right semantics".

Anyway, given that you renamed the daemon file lock to not say "launching" and it's now definitely used as the mutex for running the daemon, for its entire lifetime, everything seems consistent. So as long as this works well, I think it's fine.

chia/util/lock.py
arvidn: it would be nice to be able to remove any uses that releases and re-acquires the lock while inside the context manager. Especially if it's only done in tests.

arvidn (approved):

wjblanke (approved):
aok

PR #12549, opened 2022-07-23
Last new simulator rpc's 

jack60612:
Pretty self explanatory pr, tests coming soon.
Does not explicitly require #12521 but it would be nice.
- [x] Coin store addition, get_all_coins & get_all_puzzle_hashes RPC
- [x] Chain Reversion RPC (complete block deletion  & reorg RPC's)
- [x] Unit & Integration Tests (all 4 rpc's.)


jack60612:
Tests were added, PR is now ready for review.

jack60612:
Slight Sql change made, still ready for review.

trepca (commented):
why are some file names in camel case? (this is not Java :) )

jack60612:
> why are some file names in camel case? (this is not Java :) )

The other protocols we used for the simulator did the same, so I was just following what we already did.

chia/full_node/coin_store.py
```diff
@@ -242,6 +242,22 @@ async def get_coins_removed_at_height(self, height: uint32) -> List[CoinRecord]:
                         coins.append(coin_record)
                 return coins
 
+    async def get_all_coins(self, include_spent_coins: bool) -> List[CoinRecord]:
+        # this is used by the simulator, using this normally would be a joke.
```
trepca, line 246: can we elaborate this a bit like `WARNING: this should only be used for testing or simulation, running it on a synced testnet or mainnet node will most likely result in OOM`

chia/simulator/full_node_simulator.py
```diff
@@ -69,6 +70,45 @@ async def update_autofarm_config(self, enable_autofarm: bool) -> bool:
                 await self.farm_new_transaction_block(FarmNewBlockProtocol(self.bt.farmer_ph))
             return self.auto_farm
 
+    @api_request
+    async def get_all_coins(self, request: GetAllCoinsProtocol) -> List[CoinRecord]:
+        return await self.full_node.coin_store.get_all_coins(request.include_spent_coins)
+
+    async def revert_block_height(self, new_height: uint32) -> None:
+        # this completely deletes blocks from the blockchain.
```
trepca, line 78: best if we use [docstring](https://peps.python.org/pep-0257/) for this

chia/full_node/coin_store.py
jack60612: yeah that would probably be better :)

trepca (approved):

jack60612:
🎺

PR #12550, opened 2022-07-23
Support for remembering the last used wallet key

paninaro:
This change allows the WalletNode to manage a file tracking the most recently used master key used for wallet syncing. During initialization, the WalletNode will attempt to sync using the key that was used during the previous launch unless a new key is specified. When switching keys for wallet sync, the fingerprint of the desired key is passed in, whereas at launch, no fingerprint is specified.

WalletNode writes the selected key's fingerprint into a file at the path CHIA_ROOT/wallet/db/last_used_fingerprint. As this is a quality of life improvement feature, if there's any failure to read or write to the file, the previous default behavior is preserved. That behavior is to begin syncing with the first key in the keyring.

Performed some minor cleanup around "logged_in*" member management. Also introduced a new exception type to identify the case where a key is requested by fingerprint, and that key cannot be found.

Tests have been included/updated.


paninaro:
The newly added tests have a pytest fixture dependency on `root_path_populated_with_config`. This fixture is being moved into `conftests.py` in PR: https://github.com/Chia-Network/chia-blockchain/pull/12520

chia/wallet/wallet_node.py
```diff
@@ -201,9 +219,9 @@ async def _start(
         self._new_peak_queue = NewPeakQueue(inner_queue=asyncio.PriorityQueue())
 
         self.synced_peers = set()
-        private_key = await self.get_key_for_fingerprint(fingerprint)
+        private_key = await self.get_private_key(fingerprint or self.get_last_used_fingerprint(), True)
```
xdustinface, line 222:
```suggestion
        if fingerprint is None:
            fingerprint = self.get_last_used_fingerprint()
        private_key = await self.get_private_key(fingerprint)
        # If the key for the fingerprint we provided doesn't exist just try to get the first key
        if private_key is None and fingerprint is not None:
            private_key = await self.get_private_key(None)
```

You can just drop the `use_first_key_if_not_found` parameter and all the related code if you use the above unless i miss something.

chia/wallet/wallet_node.py
```diff
@@ -175,20 +176,37 @@ def rollback_request_caches(self, reorg_height: int):
         for cache in self.untrusted_caches.values():
             cache.clear_after_height(reorg_height)
 
-    async def get_key_for_fingerprint(self, fingerprint: Optional[int]) -> Optional[PrivateKey]:
+    async def get_private_key(
```
xdustinface, line 179: nit but the renaming seems to be unrelated to the PR's topic.

xdustinface (changes requested):
Makes sense! But i left some comments.

chia/wallet/wallet_node.py
paninaro: I could do this, but it moves the complexity to the caller instead of making it an implementation detail of `get_private_key()`. The behavior is the same, yes, but I don't see a reason to make `_start()` more complicated than it already is. Seems like just a stylistic difference of opinion.

chia/wallet/wallet_node.py
paninaro: I renamed it because the exisiting behavior and intended use case is broader than simply getting a key by fingerprint. If there were other callers, perhaps that would have influenced taking this in a different direction, but as currently used, it's really about attempting to fetch a specific master key for wallet syncing, and failing that, (optionally) returning the first private key

tests/wallet/test_wallet_node.py
```diff
@@ -0,0 +1,297 @@
...
+# Can remove once root_path_populated_with_config is moved to conftests.py
+def stub() -> Any:
+    return root_path_populated_with_config
+
+
+@pytest.mark.asyncio
+async def test_get_private_key(root_path_populated_with_config: Path, get_temp_keyring: Keychain) -> None:
+    root_path: Path = root_path_populated_with_config
+    keychain: Keychain = get_temp_keyring
+    config: Dict[str, Any] = load_config(root_path, "config.yaml")["wallet"]
```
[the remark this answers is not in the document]
paninaro: I'll change this, but what's the rationale for using this other form when the function explicitly supports returning a `sub_config`? This is another case where we're moving an implementation detail to the caller instead of having it remain internal to the function (granted, it's just simple dictionary traversal in this case). A quick scan of the code shows that we mostly using the form in your suggestion, but that seems like an odd choice.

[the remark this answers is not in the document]
xdustinface: I don't care much i just saw this, remembered we are usually using this parameter instead so i suggested it :)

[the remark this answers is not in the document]
paninaro: lol, I just realized that I was arguing the case _for_ the change you suggested :)

chia/wallet/wallet_node.py
xdustinface: Hm, so right now you are adding complexity to `get_private_key` by adding the parameter `use_first_key_if_not_found` which kind of re-implements an already existing behaviour of `get_private_key`
inside the function itself only because its needed in one place. Maybe it's just me but i rather would avoid this. 

Would you now also add another parameter `alternative_fingerprint` to try fingerprint `B` in case fetching `A` fails if you need this in some place? This would blow the function up more and more. 

But ok, if you really rather want to stick with `use_first_key_if_not_found` then maybe at least drop `retry_with_first_key` and just do the fun inside the `except` block based on `use_first_key_if_not_found` or was there a reason for moving it down?

tests/wallet/test_wallet_node.py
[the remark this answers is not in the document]
altendky: Either is valid and I don't mean to try to sway the path, but my understanding is that we don't really have any concept of a subconfig other than that indirection of the indexing.  And I don't even think that indexing should be indexing at all, it should be attribute access on a class that has well-defined our config form.  `:]`  So my take is that using the parameter is just a confusing thing that misleads into thinking there's more to a 'subconfig' than there is.

But really, the parameter exists and is used elsewhere and I don't really care which form is used here.

chia/wallet/wallet_node.py
```diff
@@ -178,17 +179,41 @@ def rollback_request_caches(self, reorg_height: int):
...
+
+    async def get_private_key(self, fingerprint: Optional[int]) -> Optional[PrivateKey]:
+        """
+        Attempt to get the private key for the given fingerprint. If the fingerprint is None,
+        get_key_for_fingerprint() will return the first private key. Similarly, if a key isn't
+        returned for the provided fingerprint, the first key will be returned.
+        """
+        key: Optional[PrivateKey] = await self.get_key_for_fingerprint(fingerprint)
+
+        if key is None:
```
xdustinface, line 208:
```suggestion
        if key is None and fingerprint is not None:
```

Otherwise you double try in the `fingerprint=None` case because the above `get_key_for_fingerprint(None)` already failed.

chia/wallet/wallet_node.py
```diff
@@ -178,17 +179,41 @@ def rollback_request_caches(self, reorg_height: int):
...
+            try:
+                keychain_proxy = await self.ensure_keychain_proxy()
+                key = await keychain_proxy.get_first_private_key()
+            except Exception:
+                self.log.exception("Failed to get first private key")
```
xdustinface, line 213:
```suggestion
            key = await self.get_key_for_fingerprint(None)
```

Why re-implement this?

xdustinface (commented):
-  Now with the split function you should revert the renaming in `test_wallet_rpc.py` because its behaviour change and now anyway unrelated.
- I added two more comments about the new `get_private_key`.
- I still don't really understand why we not just put this logic into `WalletNode._start` since its from my understanding the only place we need this unless you have some other uses cases in mind.

xdustinface (dismissed):
Looking forward to use it 👍

xdustinface (approved):

emlowe (approved):

PR #12586, opened 2022-07-25
Rework wallet new peak 

almogdepaz:
separate trusted and non-trusted  sync

separate non-trusted long and short sync

remove redundant code paths

validate peak header block - currently we request the block but dont validate the response this has unexpected results

don't request WP when not needed (untrusted short sync)

dont allow WP requests during secondary sync


chia/wallet/wallet_node.py
```diff
@@ -906,173 +909,181 @@ async def new_peak_wallet(self, new_peak: wallet_protocol.NewPeakWallet, peer: W
...
+        ):
+            self.log.warning(f"bad header block response from Peer {peer.get_peer_info()}.")
+            await peer.close(120)
+            return
+
         latest_timestamp: Optional[uint64] = await self.is_peer_synced(peer, header_block, request_time)
         if latest_timestamp is None:
             if trusted:
                 self.log.debug(f"Trusted peer {peer.get_peer_info()} is not synced.")
-                return
```
almogdepaz, line 913: this was redundant, we return in the end of this clause

chia/wallet/wallet_node.py
```diff
@@ -906,173 +909,181 @@ async def new_peak_wallet(self, new_peak: wallet_protocol.NewPeakWallet, peer: W
...
-                        else:
-                            self.log.info("Will not do secondary sync, there is already another sync task running.")
-                            return
-                    self.log.info(f"New peak wallet.. {new_peak.height} {peer.get_peer_info()} 12")
-                    if (
-                        self.wallet_state_manager.blockchain.synced_weight_proof is None
-                        or weight_proof.recent_chain_data[-1].weight
-                        > self.wallet_state_manager.blockchain.synced_weight_proof.recent_chain_data[-1].weight
-                    ):
-                        await self.wallet_state_manager.blockchain.new_weight_proof(weight_proof, block_records)
```
almogdepaz, line 1015: we already set the WP as the synced WP in line 990, we never enter this path

chia/wallet/wallet_node.py
```diff
@@ -297,9 +305,10 @@ def _close(self):
 
     async def _await_closed(self, shutting_down: bool = True):
         self.log.info("self._await_closed")
-
         if self._server is not None:
             await self.server.close_all_connections()
+        if self._weight_proof_handler is not None:
+            self._weight_proof_handler.cancel_weight_proof_tasks()
```
arvidn, line 311: this doesn't look right. It looks like this should be in `_close()` and here you should `await` all tasks having been terminated.

chia/wallet/wallet_node.py
```diff
@@ -957,175 +960,180 @@ async def new_peak_wallet(self, new_peak: wallet_protocol.NewPeakWallet, peer: W
...
+            tb = traceback.format_exc()
+            self.log.error(f"Error syncing to {peer.get_peer_info()} {e} {tb}")
+            tb = traceback.format_exc()
+            self.log.error(f"Error syncing to {peer.get_peer_info()} {tb}")
+            await peer.close()
+            if syncing:
+                self.wallet_state_manager.set_sync_mode(False)
+        return
+
+    async def long_sync_from_untrusted(self, syncing: bool, new_peak_hb: HeaderBlock, peer: WSChiaConnection):
```
arvidn, line 1061: I take it the code was moved verbatim into these functions, right? i.e. this is a pure refactor, no change in logic.

chia/wallet/wallet_weight_proof_handler.py
```diff
@@ -67,100 +55,30 @@ def cancel_weight_proof_tasks(self):
...
+                validate_from,
+            )
         )
         self._weight_proof_tasks.append(task)
-        valid, summaries, block_records = await task
+        valid, block_records = await task
         self._weight_proof_tasks.remove(task)
         return valid, summaries, block_records
 
-    async def _validate_weight_proof_inner(
```
arvidn, line 79: so this was basically completely duplicated?

arvidn (commented):
since this is a pretty big patch, please just add commits when updating it. avoid rebasing or squashing. That can be done at the very end, after it's approved.

chia/wallet/wallet_weight_proof_handler.py
almogdepaz: yes. i didn't change any logic regarding the WP validation,  just removed redundant code duplication 

all the changes related to the weight_proof are in https://github.com/Chia-Network/chia-blockchain/pull/12310 
i needed that so i can organise the code better but i think its better to review that there (Mariano also reviewed it)

chia/full_node/weight_proof.py
```diff
@@ -576,7 +576,8 @@ def validate_weight_proof_single_proc(self, weight_proof: WeightProof) -> Tuple[
         if not _validate_sub_epoch_segments(self.constants, rng, wp_segment_bytes, summary_bytes):
             return False, uint32(0)
         log.info("validate weight proof recent blocks")
-        if not _validate_recent_blocks(self.constants, wp_recent_chain_bytes, summary_bytes):
+        success, _ = validate_recent_blocks(self.constants, wp_recent_chain_bytes, summary_bytes)
```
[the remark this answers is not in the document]
almogdepaz: it used to return a bool here, this method was duplicated in the wallet code where the only difference was we return the records as well, i unified the code.

chia/full_node/weight_proof.py
```diff
@@ -1734,3 +1652,76 @@ def _validate_vdf_batch(
...
     return True
+
+
+async def validate_weight_proof_inner(
+    constants,
+    executor,
+    shutdown_file_name,
+    num_processes,
+    weight_proof: WeightProof,
+    summaries: List[SubEpochSummary],
```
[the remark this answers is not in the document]
almogdepaz: serialize and deserialize is done inside validate_weight_proof_inner 
before and after we call validate_recent_blocks
which is the method that the executor runs

chia/full_node/weight_proof.py
```diff
@@ -1734,3 +1652,76 @@ def _validate_vdf_batch(
...
+    validate_from: int,
+) -> Tuple[bool, List[BlockRecord]]:
+    assert len(weight_proof.sub_epochs) > 0
+    if len(weight_proof.sub_epochs) == 0:
+        return False, []
+
+    peak_height = weight_proof.recent_chain_data[-1].reward_chain_block.height
+    log.info(f"validate weight proof peak height {peak_height}")
+    seed = summaries[-2].get_hash()
+    rng = random.Random(seed)
```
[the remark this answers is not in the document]
almogdepaz: we depend on both peers using the same method so we depend on pythons implementation rn i raised this issue at the time (pre mainnet) and was told it is OK since we dont plan on multi client and we can change it without forking

chia/full_node/weight_proof.py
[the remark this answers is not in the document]
arvidn: ah, right. the source of my confusion was the fact that you passed `executor` to the new task. It made me think it was also scheduling the task on that executor.

chia/full_node/weight_proof.py
```diff
@@ -1734,3 +1652,76 @@ def _validate_vdf_batch(
...
+    if not validate_sub_epoch_sampling(rng, sub_epoch_weight_list, weight_proof):
+        log.error("failed weight proof sub epoch sample validation")
+        return False, []
+
+    summary_bytes, wp_segment_bytes, wp_recent_chain_bytes = vars_to_bytes(summaries, weight_proof)
+    recent_blocks_validation_task = executor.submit(
```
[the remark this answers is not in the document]
almogdepaz: this code is not running in the executor

chia/full_node/weight_proof.py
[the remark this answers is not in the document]
arvidn: I just wouldn't expect python to guarantee stability on this

chia/full_node/weight_proof.py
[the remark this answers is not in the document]
arvidn: I retract this comment. I had misunderstood and thought `validate_inner` was running on the executor as well

chia/wallet/wallet_node.py
almogdepaz: i removed some unreachable code and avoid a WP request where it not needed
avoid a block header request for the peak in the case where we don't use it and added validation that the peak matches the block we requested (also added a test for that) which resulted in unexpected behaviour by the node


the intention was for the logic to stay the same, having said that this code was not easy to refactor.

chia/wallet/wallet_node.py
almogdepaz: this was moved from wallet_state_manager where it was in the await_closed
i moved it now but looking at the code am not sure whats the purpose of the separation between await_closed and close

arvidn (dismissed):

mariano54:
I'll fix the conflicts

almogdepaz:
> I'll fix the conflicts

@arvidn @trepca conflicts resolved, there are no changes this just got forgotten

almogdepaz:
@arvidn @trepca i revived the pr, conflicts resolved

chia/wallet/wallet_node.py
```diff
@@ -1028,180 +1027,181 @@ async def new_peak_wallet(self, new_peak: wallet_protocol.NewPeakWallet, peer: W
...
+
+        if await self.check_for_synced_trusted_peer(new_peak_hb, request_time):
+            self.log.info("Cancelling untrusted sync, we are connected to a trusted peer")
+            return
+
+        try:
+            await self.long_sync_from_untrusted(syncing, new_peak_hb, peer)
+        except Exception as e:
+            tb = traceback.format_exc()
+            self.log.error(f"Error syncing to {peer.get_peer_info()} {e} {tb}")
```
trepca, line 1121: using `log.exception(f"Error syncing to {peer.get_peer_info()}")` will automatically add the stacktrace

chia/wallet/wallet_node.py
```diff
@@ -1028,180 +1027,181 @@ async def new_peak_wallet(self, new_peak: wallet_protocol.NewPeakWallet, peer: W
...
+            tb = traceback.format_exc()
+            self.log.error(f"Error syncing to {peer.get_peer_info()} {tb}")
```
trepca, line 1123: why another `log.error` here, but without exception this time?

trepca (commented):
@almogdepaz this is great, thanks. I just have a few comments

chia/wallet/wallet_node.py
almogdepaz: bad merge, thanks

almogdepaz:
> @almogdepaz this is great, thanks. I just have a few comments

fixed, thanks

trepca (approved):

PR #12589, opened 2022-07-25
Slight change to get_block_spends

jack60612:
We should not return none when the block is not a transaction block, we should only throw when the block is invalid


freddiecoleman (approved):
I have tested this locally and it works as expected.

PR #12590, opened 2022-07-25
remove redundant tuple conversions

arvidn:
this patch extends the block_store unit test as well as removes some redundant conversions from lists to tuples in the block_store sqlite calls. aiosqlite only requires parameters to be iterable, not specifically tuples.

It gives a small improvement in the `benchmarks/block_store.py` benchmark.

main:
```
all tests completed in 95.1360s
```

this patch:

```
all tests completed in 88.1565s
```

I removed the (unnecessary) class in the block_store test, which reduced the indentation by one level. The first commit is best viewed ignoring whitespace.

### DIFF
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -381,7 +381,7 @@ class BlockStore:
 
         return [generators[h] for h in heights]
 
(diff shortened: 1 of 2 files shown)


mariano54 (approved):
LGTM

emlowe (approved):

PR #12603, opened 2022-07-26
Atari get keys.

fchirica:


chia/rpc/data_layer_rpc_api.py
```diff
@@ -97,6 +98,13 @@ async def get_value(self, request: Dict[str, Any]) -> Dict[str, Any]:
             hex = value.hex()
         return {"value": hex}
 
+    async def get_keys(self, request: Dict[str, Any]) -> Dict[str, Any]:
+        store_id = bytes32(hexstr_to_bytes(request["id"]))
+        if self.service is None:
+            raise Exception("Data layer not created")
+        keys = await self.service.get_keys(store_id)
+        return {"keys": keys}
```
altendky, line 106: I always forget, are we supposed to explicitly turn the bytes to hex?

```suggestion
        return {"keys": [f"0x{key.hex()}" for key in keys]}
```

I'll note that I really dislike the `0x` prefix.  It always reads to me as indicating a hex-encoded integer.  Is there some other case where people prefix arbitrary blobs with `0x`?  Certainly this is unrelated to this PR and won't hold it up.

tests/core/data_layer/test_data_rpc.py
```diff
@@ -308,6 +309,9 @@ async def test_keys_values_ancestors(one_wallet_node_and_rpc: nodes_with_port, b
         assert dic["0x" + key3.hex()] == "0x" + value3.hex()
         assert dic["0x" + key4.hex()] == "0x" + value4.hex()
         assert dic["0x" + key5.hex()] == "0x" + value5.hex()
+        assert len(keys["keys"]) == len(dic)
+        for key in keys["keys"]:
+            assert "0x" + key.hex() in dic
```
altendky, line 314: Related to above, shouldn't `key` be JSON serializable and not bytes here?  maybe...

altendky (commented):
Is there a reason to not allow specification of the root hash of interest?

I _think_ there will be changes related to my comments, but not all require changes.

altendky (approved):

Issue #12268, opened 2022-07-06
[Bug]1.4.1293 The wallet of light money cannot be synchronized.


Concord-us:
### What happened?

[debug.log](https://github.com/Chia-Network/chia-blockchain/files/9056554/debug.log)
Look down on the log for details

### Version

1.4.1293

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output

```shell
Detailed log has been uploaded to the attachment
```

HVNC1:
что вы сделали с бумажником.почините его

emlowe:
closing issue

PR #12604, opened 2022-07-26
handle errors during shutdown

almogdepaz:
dont throw Errors on shutdown/restart while syncing 

check that service name is included in register_service 
(this error appeared in the logs here https://github.com/Chia-Network/chia-blockchain/issues/12268 )


chia/wallet/wallet_node.py
```diff
@@ -1392,6 +1395,8 @@ async def validate_block_inclusion(
             start, end, peer_request_cache, all_peers
         )
         if blocks is None:
+            if self._shut_down:
+                self.log.debug(f"shutdown, dont fetch blocks {start} {end}")
             self.log.error(f"Error fetching blocks {start} {end}")
```
jack60612, line 1400: we should have an else here, just to make it clear.
instead of the current message, we should probably put `"Shutting down, block fetching from: {start}  to {end} canceled."` just to make it easier for users to understand

chia/wallet/wallet_node.py
```diff
@@ -710,7 +710,10 @@ async def receive_and_validate(inner_states: List[CoinState], inner_idx_start: i
 
             except Exception as e:
                 tb = traceback.format_exc()
-                self.log.error(f"Exception while adding state: {e} {tb}")
+                if self._shut_down:
+                    self.log.debug(f"shutdown while adding state : {e} {tb}")
```
jack60612, line 714: shutdown to Shutting down

chia/daemon/server.py
```diff
@@ -1155,7 +1155,10 @@ async def exit(self) -> None:
 
     async def register_service(self, websocket: WebSocketResponse, request: Dict[str, Any]) -> Dict[str, Any]:
         self.log.info(f"Register service {request}")
-        service = request["service"]
+        service = request.get("service")
+        if service is None:
+            self.log.error("cant register, missing service name")
```
jack60612, line 1160: `"Service Name missing from request to 'register_service'"` would provide a little more information

jack60612 (changes requested):
Mostly good, just 1 or 2 grammar knitpicks.

jack60612 (approved):
lgtm, was OOO

emlowe (approved):

PR #12606, opened 2022-07-26
Release/1.5.0

wallentx:


chia/daemon/server.py
```diff
@@ -244,7 +244,6 @@ async def incoming_connection(self, request):
                             self.log.error(f"Unexpected exception trying to send to websocket: {e} {tb}")
                             self.remove_connection(socket)
                             await socket.close()
-                            break
```
hoffmang9, line 247: Is this a regression?

wjblanke: ya from the websocket move. this fixes it so less likely to close everything

cmmarslender: related PR https://github.com/Chia-Network/chia-blockchain/pull/12241

altendky (changes requested):
So it's on the record, this is being replaced by CI efforts on https://github.com/Chia-Network/chia-blockchain/tree/long_lived/post_1.5.0 from which a will be processed into `main`.  This PR is not expected to be used anymore as far as I know.

PR #12629, opened 2022-07-27
catchup: into main from long_lived/post_1.5.0  @ 916ccee549b1594349c19fcbb0d0d210698c899b

altendky:
Source hash: 916ccee549b1594349c19fcbb0d0d210698c899b
Remaining commits: 0

Quick follow:
- [x] fixup the test to use `with pytest.raises():` https://github.com/Chia-Network/chia-blockchain/pull/12629/files#diff-57280248a92746bc3b13a961f30435ab1fe38be2d62fb67b142202d1aa160dc4 `test_cat_outer_puzzle()`
- [x] fix the new xfail tests
  - `test_block_retired_cat_with_memo()`
  - `test_block_retired_cat_no_memo()`
  - `test_block_cat()`

Draft for:
- [x] Review for unneeded changes
- [x] Undo GUI submodule change
- [x] Discuss and decide about handling of unsigned commits


emlowe (dismissed):
Looks ok to me

emlowe (approved):

emlowe (changes requested):
Need to deal with unsigned commits in some way - discussion ongoing

tests/wallet/cat_wallet/test_cat_outer_puzzle.py
```diff
@@ -62,5 +62,10 @@ def test_cat_outer_puzzle() -> None:
         ACS,
         inner_solution,
     )
-    double_cat_puzzle.run(solution)
+    try:
+        double_cat_puzzle.run(solution)
```
altendky, line 66: Why should this be failing now?

Since I'm commenting here, this should use `with pytest.raises():` to require an exception get raised.  But, that will be taken care of in a followup as noted in the OP.

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -622,7 +622,7 @@ async def generate_unsigned_spendbundle(
         for coin in cat_coins:
             if first:
                 first = False
-                announcement = Announcement(coin.name(), std_hash(b"".join([c.name() for c in cat_coins])), b"\xca")
+                announcement = Announcement(coin.name(), std_hash(b"".join([c.name() for c in cat_coins])))
```
altendky, line 625: It looks like this comment is now out of date and should be updated?  If so I'll add this to the post-merge quick follow list.

https://github.com/Chia-Network/chia-blockchain/blob/catchup/main_from_long_lived_post_1.5.0_916ccee549b1594349c19fcbb0d0d210698c899b/chia/types/announcement.py#L8-L12

tests/wallet/cat_wallet/test_cat_outer_puzzle.py
Quexington: We can actually remove this `double_cat_puzzle` stuff from the test entirely.  I was just using it as a way to test the recursiveness of the puzzle solving driver, but now that CATs don't work inside of each other, it's a less than ideal choice.

Quexington (approved):

tests/wallet/cat_wallet/test_cat_outer_puzzle.py
altendky: Thanks for the info.  Are you saying that at this point this entire test is just verifying that we no longer have this feature?  Is it important to not have the feature?  Or did we just not make it work with CAT2.

Quexington: No the test is still validating that the cat driver works, I just wanted to check that it correctly passed some information to another layer, and since I already had the cat puzzle instantiated, I was using that.  The fix is to use a different puzzle's driver but for now it's okay to just delete all the double CAT stuff.  It was not important to have the CAT within a CAT feature.

Quexington: Actually after thinking on it 5 more seconds, leaving the raise catching is fine just with `pytest.raises(ValueError, match="clvm raise")` instead and that still tests what was intentioned

matt-o-how (approved):
cat 2 commit looks like my code

emlowe (approved):

emlowe:
Leaving final decision to Bill on merging.

emlowe:
Bill and Justin have approved this.

PR #12665, opened 2022-07-29
Nft reorg

ytx1991:
Issue: https://github.com/Chia-Network/issue-tracker/issues/3
Solution: Add two new columns (removed_height, latest_height) to the NFT tables. 
removed_height: Save the block height in the removed_height, the row will be deleted after certain blocks. If reorg happened the removed_height will be set to null.
latest_height: the confirmed block height of latest NFT coin. If reorg happened the row will be removed.


mariano54:
Awesome, this needs to be fixed :)
Do you mind adding a test that checks that NFT dissapears from user's wallet, like this one?
https://github.com/Chia-Network/chia-blockchain/pull/12230/files

ytx1991:
> Awesome, this needs to be fixed :) Do you mind adding a test that checks that NFT dissapears from user's wallet, like this one? https://github.com/Chia-Network/chia-blockchain/pull/12230/files

Sure, will do

mariano54:
Also, have you considered what happens in all reorg cases?
For example:

- Received an NFT reorg
- Sent an NFT reorg
- Accepted offer for NFT reorg
- Accepted offer for XCH reorg

etc.

ytx1991:
> Also, have you considered what happens in all reorg cases? For example:
> 
> * Received an NFT reorg
> * Sent an NFT reorg
> * Accepted offer for NFT reorg
> * Accepted offer for XCH reorg
> 
> etc.

I assume there isn't a big difference between offer and normal transfer. My local run of tests have a quite different results comparing with the CI. Seems the reorged txs will be added to blocked automatically. I'm not sure if this is related to the mempool.

chia/wallet/wallet_nft_store.py
```diff
@@ -110,5 +139,25 @@ async def get_nft_by_id(self, nft_id: bytes32) -> Optional[NFTCoinInfo]:
...
+    async def rollback_to_block(self, height: int) -> None:
+        """
+        Rolls back the blockchain to block_index. All coins confirmed after this point are removed.
+        All coins spent after this point are set to unspent. Can be -1 (rollback all)
+        """
+        if height < 0:
+            return
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
+            # Remove reorged NFTs
+            await (await conn.execute("DELETE FROM users_nfts WHERE latest_height>?", (height,))).close()
```
trepca, line 155: shouldn't you also remove all the tx records for this block height?

ytx1991: It's handled in the tx store. It doens't delete the tx. It just revert the spent and confirmed flags.

chia/wallet/wallet_nft_store.py
```diff
@@ -110,5 +139,25 @@ async def get_nft_by_id(self, nft_id: bytes32) -> Optional[NFTCoinInfo]:
...
+            uint32(row[6]),
             row[4] == IN_TRANSACTION_STATUS,
         )
+
+    async def rollback_to_block(self, height: int) -> None:
+        """
+        Rolls back the blockchain to block_index. All coins confirmed after this point are removed.
+        All coins spent after this point are set to unspent. Can be -1 (rollback all)
+        """
+        if height < 0:
```
trepca, line 151: the function doc says height can be -1 to rollback all but we return here without removing any nfts?

tests/wallet/test_nft_store.py
```diff
@@ -33,5 +35,18 @@ async def test_nft_store(self) -> None:
...
+                a_bytes32_1,
+                Coin(a_bytes32, a_bytes32, uint64(1)),
+                LineageProof(a_bytes32, a_bytes32, uint64(1)),
+                puzzle,
+                uint32(1),
+                uint32(12),
+            )
+            await db.save_nft(uint32(1), a_bytes32_1, nft1)
+            assert nft1 == (await db.get_nft_list(wallet_id=uint32(1)))[0]
+            await db.rollback_to_block(10)
```
arvidn, line 51: I would think we want to have more than a single test-case for this function. Ideally the tests would be split up into separate functions too.

chia/wallet/nft_wallet/nft_info.py
```diff
@@ -84,6 +84,7 @@ class NFTCoinInfo(Streamable):
     lineage_proof: Optional[LineageProof]
     full_puzzle: Program
     mint_height: uint32
+    latest_height: uint32 = uint32(0)
```
arvidn, line 87: what does this field mean? (a comment in the code may be warranted)

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -471,6 +470,10 @@ async def update_metadata(
     def get_current_nfts(self) -> List[NFTCoinInfo]:
         return self.my_nft_coins
 
+    async def load_current_nft(self) -> List[NFTCoinInfo]:
+        self.my_nft_coins = await self.wallet_state_manager.nft_store.get_nft_list(wallet_id=self.wallet_id)
```
arvidn, line 474: do we store these both in RAM in `self.my_nft_coins` as well as in the `nft_store`?

arvidn:
@ytx1991 would you mind adding a description to this PR? e.g. what the problem is and how you're solving it.

chia/wallet/nft_wallet/nft_wallet.py
ytx1991: It was only in my_nft_coins as a part of the wallet metadata. We splitted the nft and wallet data later. I didn't touch this because of it can improve the performance a little bit. I'm fine to remove this in another PR.

mariano54 (approved):

trepca (approved):

PR #12666, opened 2022-07-29
show which processes get closed on daemon termination  

jack60612:
added to logs and RPC endpoint

### DIFF
--- a/chia/cmds/stop.py
+++ b/chia/cmds/stop.py
@@ -1,3 +1,4 @@
+import asyncio
 import sys
 from pathlib import Path
 from typing import Any, Dict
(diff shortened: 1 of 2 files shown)


chia/cmds/stop.py
```diff
@@ -20,6 +21,8 @@ async def async_stop(root_path: Path, config: Dict[str, Any], group: str, stop_d
         r = await daemon.exit()
         await daemon.close()
         if r.get("data", {}).get("success", False):
+            [print(f"{service}: Stopped") for service in r["data"]["services_stopped"]]
```
mariano54, line 24: it will throw an error if services_stopped is not present

jack60612: would it not always be there?

jack60612: cant be too safe, ill add a check

jack60612: safe then sorry, good point

jack60612:
squashing and rebasing for ci

paninaro (approved):
lgtm

PR #12669, opened 2022-07-29
util: Some cleanups in `file_keyring.py`

xdustinface:
See commit messages.

### DIFF
--- a/chia/util/file_keyring.py
+++ b/chia/util/file_keyring.py
@@ -79,6 +79,10 @@ def default_outer_payload() -> Dict[str, Any]:
     return {"version": 1}
 
 
+def default_file_keyring_data() -> Dict[str, Any]:
+    return {"keys": {}}
+
+
 def keyring_path_from_root(keys_root_path: Path) -> Path:
     """
     Returns the path to keyring.yaml
@@ -101,7 +105,7 @@ def loads_keyring(method: FileKeyringUnlockingCallable) -> FileKeyringUnlockingC
 
         # Check the outer payload for 'data', and check if we have a decrypted cache (payload_cache)
         with self.load_keyring_lock:
-            if (self.has_content() and not self.cached_data_dict) or self.needs_load_keyring:
+            if self.needs_load_keyring:
                 self.load_keyring()
         return method(self, *args, **kwargs)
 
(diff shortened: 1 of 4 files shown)


chia/util/file_keyring.py
```diff
@@ -197,7 +199,7 @@ class FileKeyring(FileSystemEventHandler):  # type: ignore[misc] # Class cannot
     needs_load_keyring: bool = False
     salt: Optional[bytes] = None  # PBKDF2 param
     # Cache of the decrypted YAML contained in outer_payload_cache['data']
-    payload_cache: Dict[str, Any] = field(default_factory=dict)
+    cached_data_dict: Dict[str, Any] = field(default_factory=default_file_keyring_data)
```
paninaro, line 202: As I recall, there was some subtle logic to help determine if keyring migration was necessary, and it was based on whether the keyring had "content" which basically meant that the data field was present.

I do expect that there will be some user keyring.yaml files that just contain "version: 1". By changing the default value here, are there side-effects that we need to worry about? If we're getting rid of the migration checks, then maybe this is fine.

chia/util/file_keyring.py
```diff
@@ -286,18 +285,12 @@ def get_password(self, service: str, user: str) -> Optional[str]:
 
     @loads_keyring
     def _inner_set_password(self, service: str, user: str, passphrase: str) -> None:
-        keys = self.ensure_cached_keys_dict()
-        # Convert the passphrase to a string (if necessary)
-        passphrase = bytes(passphrase).hex() if type(passphrase) == bytes else str(passphrase)  # type: ignore
```
paninaro, line 291: Why is this check being removed? It was necessary for some code path (maybe a legacy keyring migration case).

xdustinface: Hm, because we hint `passphrase` as `str` and ideally we shouldn't end up with `bytes` there anyway. I checked the related places and the only way i could think of how we would end with `bytes` there is when we set the key here https://github.com/Chia-Network/chia-blockchain/blob/4c87c99d1563d6d118ef6af8cef90b4f19b0d8b4/chia/util/keyring_wrapper.py#L423 which we got from the legacy keyring earlier here https://github.com/Chia-Network/chia-blockchain/blob/4c87c99d1563d6d118ef6af8cef90b4f19b0d8b4/chia/util/keyring_wrapper.py#L413 right? So 3757cfadb179ec9a265cc5d20e67ee3361cfb7d8 should be fine i guess and the conversion is in the legacy area now.

paninaro: Ok if the check is done upstream then I agree this is fine.

paninaro (approved):
lgtm

paninaro (dismissed):
lgtm

xdustinface:
@paninaro Rebased and added 67410e823a4a71b541416444050169f12abfbcb6 on top.

jack60612:
ci

paninaro (approved):
lgtm

PR #12672, opened 2022-07-29
util: Introduce `FileKeyringContent`

xdustinface:
Factors out all the file handling related stuff from the `FileKeyring` class into `FileKeyringContent` which is a `dataclass` representing the file structure.

Based on:
- #12541 
- #12669

### DIFF
--- a/chia/util/file_keyring.py
+++ b/chia/util/file_keyring.py
@@ -6,7 +6,7 @@ import os
 import shutil
 import sys
 import threading
-from dataclasses import dataclass, field
+from dataclasses import asdict, dataclass, field
 from hashlib import pbkdf2_hmac
 from pathlib import Path
 from secrets import token_bytes
@@ -20,6 +20,7 @@ from watchdog.observers import Observer
 
 from chia.util.default_root import DEFAULT_KEYS_ROOT_PATH
 from chia.util.lock import Lockfile
+from chia.util.streamable import convert_byte_type
 
 SALT_BYTES = 16  # PBKDF2 param
 NONCE_BYTES = 12  # ChaCha20Poly1305 nonce is 12-bytes
@@ -74,10 +75,6 @@ def decrypt_data(input_data: bytes, key: bytes, nonce: bytes) -> bytes:
     return output[len(CHECKBYTES_VALUE) :]
 
 
-def default_outer_payload() -> Dict[str, Any]:
-    return {"version": 1}
-
-
 def default_file_keyring_data() -> Dict[str, Any]:
     return {"keys": {}}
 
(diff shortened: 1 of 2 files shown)


chia/util/file_keyring.py
```diff
@@ -243,237 +254,146 @@ def on_modified(self, event: Union[FileSystemEvent, DirModifiedEvent]) -> None:
         self.check_if_keyring_file_modified()
 
     def check_if_keyring_file_modified(self) -> None:
-        if self.keyring_path.exists():
```
paninaro, line 246: Unfortunately I don't remember why this check was here. It may have been to prevent hitting some exception beyond `FileNotFoundError`, possibly specific to a single platform. Anyway, hopefully the tests will sufficiently cover any corner cases that this code may have handled.

chia/util/file_keyring.py
```diff
@@ -112,69 +109,124 @@ def inner(self: FileKeyring, *args: object, **kwargs: object) -> Optional[str]:
     return inner
 
 
+class FileKeyringVersionError(Exception):
+    def __int__(self, actual_version: int) -> None:
```
paninaro, line 113: Is this supposed to be `__init__`?

chia/util/file_keyring.py
```diff
@@ -112,69 +109,124 @@ def inner(self: FileKeyring, *args: object, **kwargs: object) -> Optional[str]:
...
-    pick an appropriate keyring store backend based on the OS.
 
-    The keyring file format uses YAML with a few top-level keys:
+    # The version of the whole keyring file structure
+    version: int = 1
+    # Random salt used as a PBKDF2 parameter. Updated when the master passphrase changes
+    salt: bytes = field(default_factory=generate_salt)
+    # Random nonce used as a ChaCha20Poly1305 parameter. Updated on each write to the file.
+    nonce: bytes = field(default_factory=generate_nonce)
+    # Encrypted and base64 decoded keyring data.
```
paninaro, line 135:
```suggestion
    # Encrypted and base64 encoded keyring data.
```

paninaro: I'd mention that the data is prepended with the checkbytes value, even though the checkbytes are added/removed elsewhere.

chia/util/file_keyring.py
```diff
@@ -112,69 +109,124 @@ def inner(self: FileKeyring, *args: object, **kwargs: object) -> Optional[str]:
...
+    @classmethod
+    def create_from_path(cls, path: Path) -> FileKeyringContent:
+        loaded_dict = dict(yaml.safe_load(open(path, "r")))
+        version = int(loaded_dict["version"])
+
+        if version > MAX_SUPPORTED_VERSION:
+            raise FileKeyringVersionError(version)
+
+        loaded_dict["salt"] = convert_byte_type(bytes, loaded_dict["salt"])
+        loaded_dict["nonce"] = convert_byte_type(bytes, loaded_dict["nonce"])
```
paninaro, line 151: Some keyring.yaml files won't have the salt and nonce set yet if migration hasn't occurred yet. The previous code used .get in `load_outer_payload()` and `load_keyring()` to avoid hitting an exception when attempting to read the (possibly non-existent) salt/nonce values.

chia/util/file_keyring.py
xdustinface: This belongs into #12669

chia/util/file_keyring.py
xdustinface: Yes, thanks, adjusted!

chia/util/file_keyring.py
xdustinface: Good point, i moved the `convert_byte_type` calls into `FileKeyring.__post_init__`.

paninaro (approved):

PR #12680, opened 2022-07-29
DL third party on chain mirrors

Quexington:
The idea here is that any individual can create a coin that hints to a particular datalayer singleton and provides some data which the wallets tracking that datalayer singleton will sync.


geoffwalmsley (approved):
LGTM

geoffwalmsley (approved):
LGTM

PR #12682, opened 2022-07-29
Add small coin selection improvements

jack60612:
- Add support for a target of 0
- Change type of max_coin_amount to uint64
- Validate if coins are below max_coin_amount for future RPC use.

I would go commit by commit for easier reviewing.

### DIFF
--- a/chia/wallet/coin_selection.py
+++ b/chia/wallet/coin_selection.py
@@ -102,7 +102,7 @@ async def select_coins(
                 coin_set = {greater_coin}
         return coin_set
     else:
-        # if smaller_coin_sum == amount and len(smaller_coins) >= max_num_coins, or amount == 0.
+        # if smaller_coin_sum == amount and (len(smaller_coins) >= max_num_coins or amount == 0)
         potential_large_coin: Optional[Coin] = select_smallest_coin_over_target(amount, valid_spendable_coins)
         if potential_large_coin is None:
             raise ValueError("Too many coins are required to make this transaction")
--- a/chia/wallet/coin_selection.py
+++ b/chia/wallet/coin_selection.py
@@ -44,7 +44,7 @@ async def select_coins(
             continue
         if coin_record.coin in exclude:
             continue
-        if coin_record.coin.amount < min_coin_amount:
+        if coin_record.coin.amount < min_coin_amount or coin_record.coin.amount > max_coin_amount:
             continue
         valid_spendable_coins.append(coin_record.coin)
         sum_spendable_coins += coin_record.coin.amount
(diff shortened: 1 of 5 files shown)


chia/wallet/coin_selection.py
```diff
@@ -56,6 +56,11 @@ async def select_coins(
             f"Transaction for {amount} is greater than spendable balance of {sum_spendable_coins}. "
             "There may be other transactions pending or our minimum coin amount is too high."
         )
+    if amount == 0 and sum_spendable_coins == 0:
+        raise ValueError(
+            "No coins available to spend, you can not create a coin with an amount of 0,"
+            " without already having coins."
+        )
```
Quexington, line 63: I see the point of this raise, but is there any scenario where we would want to call `select_coins` and have it return an empty list?  It wouldn't hurt to have this in here, but I don't see the point.

chia/wallet/coin_selection.py
```diff
@@ -97,7 +102,7 @@ async def select_coins(
                 coin_set = {greater_coin}
         return coin_set
     else:
-        # if smaller_coin_sum == amount and len(smaller_coins) >= max_num_coins.
+        # if smaller_coin_sum == amount and len(smaller_coins) >= max_num_coins, or amount == 0.
```
Quexington, line 105: This comment seems inconsistent with the line it's trying to reference.  I can't tell if it's incorrect, but it's different.

chia/wallet/coin_selection.py
jack60612: No, as we assert it later, however i want a clean and easy to understand error

chia/wallet/coin_selection.py
jack60612: yeah ill make it clearer

Quexington (approved):
Seems fine to me, but someone closer to this code might also review it.

chia/wallet/coin_selection.py
```diff
@@ -97,7 +102,7 @@ async def select_coins(
                 coin_set = {greater_coin}
         return coin_set
     else:
-        # if smaller_coin_sum == amount and len(smaller_coins) >= max_num_coins.
+        # if smaller_coin_sum == amount and (len(smaller_coins) >= max_num_coins or amount == 0)
```
trepca, line 105: what is the reason for leaving this comment in?

jack60612: This improves readability, and makes it easier to understand, Mariano suggested the comment so I think it's a good idea.

trepca (approved):

PR #12683, opened 2022-07-29
New test files with CAT2 coins

emlowe:
Update run_block tests with new test blocks with CAT2 coins.
Remove xfail


jack60612 (dismissed):
LGTM, sorry for not doing earlier!

altendky (changes requested):
It looks like  `149988.json` is no longer used.

It looks like the block height is present inside the file.  Perhaps another unchanging name would be better and would allow future diffs to be relevant diffs?

jack60612:
> It looks like `149988.json` is no longer used.
> 
> It looks like the block height is present inside the file. Perhaps another unchanging name would be better and would allow future diffs to be relevant diffs?

yeah thats what i recommended in an earlier pr.

emlowe:
The program can also follow chains of blocks via `transactions_generator_ref_list` which lists additional blocks which are then found by using the `height as filename` scheme (eg. 466212 references 442734)

Notes, any diffs would really be very readable any way since this is using the `transactions_generator` which is just a long string of hex bytes.

altendky (approved):
Yeah, the file naming wasn't a big deal.  I just look for opportunities for diffs to have context.  Anyways, thanks for the explanation.

jack60612 (approved):
lgtm

wjblanke (approved):
thanks for these tests!

PR #12686, opened 2022-07-29
Catchup/main from long lived post 1.5.0 916ccee549b1594349c19fcbb0d0d210698c899b (repeat for history tracking)

altendky:
Repeat of https://github.com/Chia-Network/chia-blockchain/pull/12629 which was squash merged.  If this gets a regular merge then the commits will show up in the history of `main`, including the commit tagged as 1.5.0.

Draft for:
- [ ] Verifying that this is in reality a no-op change in terms of code.


altendky:
doesn't look how i want it to.

PR #12688, opened 2022-07-30
Fix more flaky tests, and optimize `test_pool_rpc` and NFT tests

mariano54:
1st commit fixes flaky test:
https://github.com/Chia-Network/chia-blockchain/runs/7587020815?check_suite_focus=true

2nd commit hopefully fixes flaky test:
https://github.com/Chia-Network/chia-blockchain/runs/7587995530?check_suite_focus=true

3rd commit attempts to fix flaky pool test:
https://github.com/Chia-Network/chia-blockchain/runs/7586924493?check_suite_focus=true

Also, improves test_pool_rpc by completely removing all sleep calls, hopefully increasing stability.
Tests are also sped up by these changes.

NFT test stability and performance is improved by removing scattered sleeps. 
NFT tests also now test some untrusted wallets usage.


tests/conftest.py
```diff
@@ -42,6 +42,16 @@
...
+async def wallet_is_synced(wallet_node: WalletNode, full_node_api) -> bool:
+    wallet_height = await wallet_node.wallet_state_manager.blockchain.get_finished_sync_up_to()
+    full_node_height = full_node_api.full_node.blockchain.get_peak_height()
+    return wallet_height == full_node_height
+
+
+async def wallets_are_synced(wns: List[WalletNode], full_node_api) -> bool:
+    return all([await wallet_is_synced(wn, full_node_api) for wn in wns])
+
+
```
altendky, line 54: I think this belongs in a test util file.  In my experience `conftest.py` files are just there for pytest fixtures. So far that applies here as well.

tests/conftest.py
```diff
@@ -42,6 +42,16 @@
...
 
 
+async def wallet_is_synced(wallet_node: WalletNode, full_node_api) -> bool:
+    wallet_height = await wallet_node.wallet_state_manager.blockchain.get_finished_sync_up_to()
+    full_node_height = full_node_api.full_node.blockchain.get_peak_height()
+    return wallet_height == full_node_height
+
+
+async def wallets_are_synced(wns: List[WalletNode], full_node_api) -> bool:
+    return all([await wallet_is_synced(wn, full_node_api) for wn in wns])
```
altendky, line 52:
```suggestion
    return all(await wallet_is_synced(wn, full_node_api) for wn in wns)
```

altendky (commented):
I still need to look over the rewritten NFT test files but figured I'd get these notes in now anyways.

altendky (commented):
Alrighty, all files looked at.

tests/conftest.py
altendky: I think this should be changed since it is structural regression.

tests/conftest.py
altendky: May as well change this since it's trivial and a button click away from being committed.

tests/wallet/nft_wallet/test_nft_1_offers.py
```diff
@@ -640,17 +645,19 @@ async def test_nft_offer_sell_nft_for_cat(two_wallet_nodes: Any, trusted: Any) -
     cats_to_mint = 100000
     cats_to_trade = uint64(10000)
     async with wallet_node_maker.wallet_state_manager.lock:
+        full_node_api.full_node.log.warning(
+            f"Mempool size: {len(full_node_api.full_node.mempool_manager.mempool.spends)}"
+        )
```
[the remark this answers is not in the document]
altendky: I won't block approval on this.

chia/timelord/timelord_launcher.py
```diff
@@ -16,12 +16,11 @@
 
 active_processes: List = []
 stopped = False
-lock = asyncio.Lock()
```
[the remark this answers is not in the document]
altendky: Up to you on this one.

altendky (changes requested):
I've documented each comment as just a comment vs. a change being requested.

chia/timelord/timelord_launcher.py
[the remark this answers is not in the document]
mariano54: Yeah this is causing trouble only sometimes. Creating an async lock outside of any async function causes issues, because it might be running on a different event loop. Each lock belongs to one event loop. This has caused issues in the past in various places. In this case, it causes a flaky test.

We might be able to remove the lock, but i'm not 100% sure right now. If we do, I would prefer to do it in a separate PR.

tests/conftest.py
mariano54: Doesn't work for me, `E   TypeError: 'async_generator' object is not iterable`.

I guess all doesn't work with async iters

tests/conftest.py
mariano54: yep makes sense

tests/conftest.py
altendky: I played briefly locally and don't have a solution so we can move on for now.  I'm _pretty_ sure that's not an async generator though...  that'd involve `async for` afaik.

altendky (approved):
Thanks for all of this, I'm looking forward to fewer reruns...

wallentx (approved):
LGTM
*There's an unsigned commit in here though

mariano54:
Replaced by https://github.com/Chia-Network/chia-blockchain/pull/12761 with signed commits

PR #12691, opened 2022-07-30
rename testnet constants to default constants, because they are the base for all networks

jack60612:


mariano54 (dismissed):
nice

emlowe (dismissed):
I'm a little confused - this is not mainnet genesis - it does not match mainnet genesis listed in initial-config.yaml - so why is this called mainnet?

jack60612:
> I'm a little confused - this is not mainnet genesis - it does not match mainnet genesis listed in initial-config.yaml - so why is this called mainnet?

These are the base constants for all networks, which then get overridden depending on the network. 
another possible name is base constants but mainnet sounded more accurate.

emlowe:
I think this name is more confusing. It seems it's neither mainnet nor testnet. It's default - so I would call it `default_kwargs`

mariano54:
These are the mainnet constants, except for the initial challenge (since when we released 1.0 we didn't have the challenge yet). That said, default_kwargs is fine too

wjblanke (approved):
aok

wallentx (approved):
:shipit:

PR #12702, opened 2022-07-31
Few additional sets of test SSL certs/keys for CI

paninaro:
Possible solution to CI issues involving nodes using the same SSL certs, resulting in node id conflicts.

Example error message: "Trying to connect to a peer ({'host': 'localhost', 'port': 41131}) with the same peer_id"

When using `get_next_private_ca_cert_and_key()` or `get_next_nodes_certs_and_keys()`, a wrapper object is returned which internally tracks when the SSL test collateral is in use (simple constructor/destructor semantics).


mariano54 (commented):
It's worth a try, thanks for addressing this.

jack60612 (commented):
All of the code is way cleaner + tests pass, and this code is fairly simple. However, I don't understand how this will solve the problem you mentioned above. does the state of which ssl certs are in use somehow get passed in somewhere, or am I missing something?

mariano54:
Lint failure needs to be fixed

mariano54 (dismissed):
Pending lint fix before merge

paninaro:
> Pending lint fix before merge

Should be fixed by https://github.com/Chia-Network/chia-blockchain/pull/12704.

jack60612:
> > Pending lint fix before merge
> 
> Should be fixed by #12704.

try catching up with main

paninaro:
> All of the code is way cleaner + tests pass, and this code is fairly simple. However, I don't understand how this will solve the problem you mentioned above. does the state of which ssl certs are in use somehow get passed in somewhere, or am I missing something?

It may not fix the CI failures, but if the assumption is that some tests are using all of the test certs/keys, then adding a few more may help. The other changes have been added to provide some logging if we detect that a set of certs/keys is being reused while still in use by some part of the test.

jack60612 (dismissed):
lgtm

mariano54 (dismissed):

paninaro:
Rebased on top of main. Changes previously made to `test_start_simulator.py` have been moved to `simulator_test_tools.py`

jack60612 (approved):
lgtm, hope it helps

mariano54 (approved):

wjblanke (approved):
aok

PR #12711, opened 2022-07-31
Replace NFT_HRP and DID_HRP with AddressType

paninaro:


AmineKhaldi (approved):
Aside from the linting issue that we're already resolving with the pending PR #12704 this looks good to me. Thanks!

emlowe (approved):

PR #12618, opened 2022-07-27
Allow Identification of CAT1 coins in offers.

jack60612:
- [x] 1.5 gets merged into main

This adds a function to identify cat1 coins, and a shell driver that throws a value error making the offer invalid

PR #12730, opened 2022-08-01
Block summaries of CAT1 offers in the RPC and CLI

Quexington:
This is a less invasive alternative to https://github.com/Chia-Network/chia-blockchain/pull/12618


jack60612 (dismissed):
Works great, tested on CLI

emlowe:
We need some test code that exercises this. I trust the existing CI runs would test that this doesn't break anything - but I think we need a CAT1 test to make sure this does what we want

emlowe (approved):
Looks good

jack60612:
Missed the notification, reviewing now

jack60612 (approved):
Lgtm

PR #12736, opened 2022-08-01
Mint from DID

geoffwalmsley:
This PR adds the DID wallet functionality needed for the standalone nft bulk minting tool. It includes: 
- a function and tests for the DID wallet to mint NFTs directly 
- a small chialisp puzzle that makes an m-of-n announcement to provide a unique launcher for each bulk-minted nft
- RPC endpoints for bulk mint spend bundles. (to be used by the main bulk minting tool
- ~~Extends the wallet rpc push_tx endpoint to accept TransactionRecord objects. This is needed to track transaction confirmation without a full node connection~~


chia/rpc/wallet_rpc_api.py
```diff
@@ -1354,6 +1355,95 @@ async def did_transfer_did(self, request) -> EndpointResult:
             "transaction_id": txs.name,
         }
 
+    async def did_mint_nfts(self, request) -> EndpointResult:
+        if await self.service.wallet_state_manager.synced() is False:
+            raise ValueError("Wallet needs to be fully synced.")
+        wallet_id = uint32(request["wallet_id"])
+        did_wallet: DIDWallet = self.service.wallet_state_manager.wallets[wallet_id]
```
ytx1991, line 1362: Better to check the wallet type is DIDWallet

chia/rpc/wallet_rpc_api.py
```diff
@@ -1354,6 +1355,95 @@ async def did_transfer_did(self, request) -> EndpointResult:
...
+            if not isinstance(meta.get("meta_uris", []), list):
+                return {"success": False, "error": "Metadata URIs must be a list"}
+            if not isinstance(meta.get("license_uris", []), list):
+                return {"success": False, "error": "License URIs must be a list"}
+            nft_metadata = [
+                ("u", meta["uris"]),
+                ("h", hexstr_to_bytes(meta["hash"])),
+                ("mu", meta.get("meta_uris", [])),
+                ("lu", meta.get("license_uris", [])),
+                ("sn", uint64(meta.get("series_number", 1))),
```
ytx1991, line 1390: I think we are using term edition_xxx

chia/rpc/wallet_rpc_api.py
```diff
@@ -1354,6 +1355,95 @@ async def did_transfer_did(self, request) -> EndpointResult:
...
+        else:
+            royalty_puzhash = royalty_address
+        royalty_percentage = request.get("royalty_percentage", None)
+        if royalty_percentage is None:
+            royalty_percentage = uint16(0)
+        else:
+            royalty_percentage = uint16(int(royalty_percentage))
+        metadata_list = []
+        for meta in request["metadata_list"]:
+            if "uris" not in meta.keys():
```
ytx1991, line 1377: Should we also check if these lists are empty?

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -1352,3 +1356,209 @@ def deserialize_backup_data(backup_data: str) -> DIDInfo:
             metadata,
         )
         return did_info
+
+    async def mint_nfts(
```
ytx1991, line 1360: Why not put this in the NFT wallet, like the bulk minting tool? It's kind of weird to put this in the DID wallet.

chia/rpc/wallet_rpc_api.py
trepca: good catch, it's `edition_number` and `edition_total`

tests/wallet/nft_wallet/test_nft_mint_from_did.py
```diff
@@ -0,0 +1,500 @@
...
+    await time_out_assert(10, wallet_0.get_confirmed_balance, funds)
+
+    did_wallet: DIDWallet = await DIDWallet.create_new_did_wallet(
+        wallet_node_0.wallet_state_manager, wallet_0, uint64(1)
+    )
+    spend_bundle_list = await wallet_node_0.wallet_state_manager.tx_store.get_unconfirmed_for_wallet(did_wallet.id())
+    spend_bundle = spend_bundle_list[0].spend_bundle
+    await time_out_assert_not_none(5, full_node_api.full_node.mempool_manager.get_spendbundle, spend_bundle.name())
+
+    for _ in range(1, num_blocks):
```
trepca, line 78: do we need to create that many blocks?

tests/wallet/nft_wallet/test_nft_mint_from_did.py
```diff
@@ -0,0 +1,500 @@
...
+    target_list = [ph_taker for x in range(n)]
+
+    sb = await nft_wallet_maker.mint_from_did(
+        metadata_list, target_list=target_list, starting_num=1, max_num=n, fee=fee
+    )
+
+    await api_0.push_tx({"spend_bundle": bytes(sb).hex()})
+
+    await time_out_assert_not_none(5, full_node_api.full_node.mempool_manager.get_spendbundle, sb.name())
+    for _ in range(1, num_blocks):
```
trepca, line 118: one block should suffice to process the transaction

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +932,210 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+        ]
+
+        n = len(metadata_list)
+        amount = uint64(1)
+        zero_coin_spends = []
+        launcher_spends = []
+        launcher_ids = []
+        eve_spends = []
+        p2_inner_puzzle = await self.standard_wallet.get_new_puzzle()
+        ending_num = starting_num + n
```
morrillup, line 1011: Not sure if this is correct. Wouldn't it be `ending_num = starting_num + max_num`? 
In the next line, `range` is used in a way that does not include `ending_num`. Not sure if that is intended either.

tests/wallet/nft_wallet/test_nft_mint_from_did.py
```diff
@@ -0,0 +1,494 @@
...
+
+        for i in range(0, n, chunk):
+            resp: Dict[str, Any] = await client.nft_mint_from_did(
+                wallet_id=nft_wallet_maker["wallet_id"],
+                metadata_list=metadata_list[i : i + chunk],
+                target_list=target_list[i : i + chunk],
+                royalty_percentage=royalty_percentage,
+                royalty_address=royalty_address,
+                starting_num=i + 1,
+                max_num=n,
```
morrillup, line 274: I don't get why `max_num=n`. Shouldn't it be `max_num=chunk`? Because `nft_mint_from_did` gets a sliced metadata_list (of length `chunk`) a few lines above, not the complete one. Also, why is `starting_num` not 1? All confusing to me.
This test only checks for correct amount of minted NFTs. Would be nice if it would also confirm they have in fact the correct metadata (of course don't use the same metadata for each as it is now).

tests/wallet/nft_wallet/test_nft_mint_from_did.py
```diff
@@ -0,0 +1,494 @@
...
+
+    sample = {
+        "hash": bytes32(token_bytes(32)).hex(),
+        "uris": ["https://data.com/1234"],
+        "meta_hash": bytes32(token_bytes(32)).hex(),
+        "meta_uris": ["https://meatadata.com/1234"],
+        "license_hash": bytes32(token_bytes(32)).hex(),
+        "license_uris": ["https://license.com/1234"],
+        "series_numer": 1,
+        "series_total": 1,
```
morrillup, line 245: I believe these should be `edition_number` and `edition_total`? Would be nice if the test checked the contents of minted NFTs.

chia/wallet/nft_wallet/nft_wallet.py
geoffwalmsley: max _num is equivalent to edition_total, so it gets used in the ephemeral puzzles. We have to define ending num as the starting_num + n because n is the chunk size of the current metadata_list for which we're creating spend bundles. It's admittedly a bit confusing, but is necessary for creating multiple spend bundles where we need to infer the edition number from a given starting index, and hold the edition_total as a constant unrelated to the metadata_list we're processing.

tests/wallet/nft_wallet/test_nft_mint_from_did.py
geoffwalmsley: Have updated test metadata to vary urls and edition numbers, and added a section in the test to confirm the NFT data and edition numbers match the input data.

tests/wallet/nft_wallet/test_nft_mint_from_did.py
geoffwalmsley: max_num is always fixed as the total number of NFTs to be minted, equivalent to edition_total.  Admittedly the indexing is a bit confusing,  but basically we give the function a starting num and a max num, and within the minting function we then calculate the other values needed so we can create the ephemeral puzzles with unique m-of-n values. 
Have updated the test (see comment below)

chia/wallet/nft_wallet/nft_wallet.py
geoffwalmsley: max_num is equivalent to edition_total, so for the given metadata_list (which is a chunk from a larger set of metadata being bulk minted) we just want to loop over the range from the starting number (edition_number) increasing up to the length of the chunk of metadata.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +932,210 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+            new_innerpuzhash = innerpuz.get_tree_hash()
+
+        # create primaries
+        primaries = [
+            AmountWithPuzzlehash(
+                {"puzzlehash": new_innerpuzhash, "amount": uint64(did_coin.amount), "memos": [new_innerpuzhash]}
+            )
+        ]
+
+        n = len(metadata_list)
```
matt-o-how, line 1004: There are no comments in the huge code block after this point, could you add some?

chia/wallet/nft_wallet/nft_wallet.py
matt-o-how, line 1011: what's the difference between `max_num` and `ending_num`?

geoffwalmsley: max_num is equivalent to edition_total, ie the total number of NFTs that are going to be minted in this run. Then ending_num is the the highest value for edition_number that will be used in this chunk of NFTs getting put in a spend bundle. So we pass a metadata_list, with a starting_num, and we use this as the m value in the ephemeral puzzles from starting_num up to ending_num. So ending_num is the max m value we'll use in this run, and max_num is always fixed as the edition_total (the n in our m-of-n ephemeral puzzle)...bit confusing, sorry. Will add comments as requested above also.

tests/wallet/nft_wallet/test_nft_mint_from_did.py
morrillup: Are you aware of the recent definition clarifications of `series` and `edition` as outlined in the [CHIP 005](https://github.com/Chia-Network/chips/blob/nft1/CHIPs/chip-0005.md)? I believe the overwhelming majority of NFTs (including Chia Friends) intend(ed) to have `edition_number` and `edition_total` set to **1** for all NFTs. I would be surprised if "max_num is always fixed as the total number of NFTs to be minted, equivalent to edition_total" as this means single mints as before.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +932,210 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+                launcher_coin.name(),
+                b"",
+                p2_inner_puzzle,
+                metadata["royalty_pc"],
+                royalty_puzzle_hash=metadata["royalty_ph"],
+            )
+            eve_fullpuz = nft_puzzles.create_full_puzzle(
+                launcher_coin.name(), metadata["program"], NFT_METADATA_UPDATER.get_tree_hash(), inner_puzzle
+            )
+            announcement_set: Set[Announcement] = set()
```
matt-o-how, line 1038: This is dangerous!
You're clearing the set inside the loop, and you aren't asserting these announcements from a signed spend.

If these announcements aren't asserted from the signed spend that creates them then anybody can change what gets created

tests/wallet/nft_wallet/test_nft_mint_from_did.py
morrillup: Actually I believe this minting method should not care at all what the user supplies as `edition_number` and `edition_total`. It should just pass it on, which it does (it's contained of the metadata list given to this method), so that's fine. The above comment is just for clarification.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,325 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+        amount = uint64(1)
+        intermediate_coin_spends = []
+        launcher_spends = []
+        launcher_ids = []
+        eve_spends = []
+        transfer_spends = []
+        p2_inner_puzzle = await self.standard_wallet.get_new_puzzle()
+
+        # Loop to create each intermediate coin, launcher, eve and (optional) transfer spends
+        for edition_number in range(edition_number_start, edition_number_end):
```
morrillup, line 1043: Naming this variable `edition_number` (and naming related method parameters `edition_number_start`, `edition_total`) is misleading as the actual edition number minted into the nft will be taken from `metadata_list` and may differ. And it rightly differs as explained in the next paragraph. I recommend to rename it (and related method parameters) to `mint_number`, `mint_list_offset` or similar.

Let me clarify again that many (certainly almost all existing) NFT collections will have `edition_number` and `edition_total` set to **1** for **each** NFT (Definition of "edition" as in [CHIP0005](https://github.com/Chia-Network/chips/blob/nft1/CHIPs/chip-0005.md). For example, Chia Friends would have edition_number set to 1 for each NFT as it does not contain "Identical copies of the same NFT" (CHIP 0005 definition of "edition".) I would be surprised if collections such as Chia Friends would not be able to benefit from a Bulk Mint tool (had to effectively single mint because their `edition_total` is 1.)

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,323 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+        total_amount = len(metadata_list) + fee
+        if xch_coins is None:
+            xch_coins = await self.standard_wallet.select_coins(uint64(total_amount))
+        assert len(xch_coins) > 0
+
+        # set the chunk size for the spend bundle we're going to create
+        chunk_size = len(metadata_list)
+
+        # Because bulk minting may not mint all the NFTs in one bundle, we
+        # calculate the edition numbers that will be used in the intermediate
```
trepca, line 1025: `edition numbers` -> `mint numbers` in the comment

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,323 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+            )
+
+            # Create the eve transaction setting the DID owner, and applying
+            # the announcements from announcement_set to match the launcher
+            # coin annnouncement
+            eve_txs = await self.generate_signed_transaction(
+                [uint64(eve_coin.amount)],
+                [p2_inner_puzzle.get_tree_hash()],
+                nft_coin=nft_coin,
+                new_owner=did_wallet.did_info.origin_coin.name(),
```
trepca, line 1123: since we set minter/creator via DID launcher, it might not make sense to also set owner in the beginning as this saves us a DID spend per batch, did the requirement specify first owner to be set?

geoffwalmsley: It wasn't specified in the requirement, but my thinking was that it's more complete to include the owner in the eve spend. So if an explorer is looking for launched NFTs it can get the did from the eve spend and then confirm it is correct by going back through the launchers to the DID. It may be redundant, although I don't think it's an extra spend since it just gets included in the eve spend

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,323 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+            # Create  the puzzle, solution and coin spend for the intermediate launcher
+            intermediate_launcher_puz = did_wallet_puzzles.DID_NFT_LAUNCHER_MOD.curry(
+                did_wallet_puzzles.LAUNCHER_PUZZLE_HASH, mint_number, mint_total
+            )
+            primaries.append(
+                AmountWithPuzzlehash(
+                    {
+                        "puzzlehash": intermediate_launcher_puz.get_tree_hash(),
+                        "amount": uint64(0),
+                        "memos": [intermediate_launcher_puz],
```
Quexington, line 1053: You forgot `.get_tree_hash()` here but I also don't know why you would need this memo sine you're spending this right now anyways.  Probably a good idea to assign an `intermediate_launcher_puz_hash` before calling `get_tree_hash` a bunch of times.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,323 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+                    }
+                )
+            )
+            intermediate_launcher_sol = Program.to([did_wallet_puzzles.LAUNCHER_PUZZLE_HASH, mint_number, chunk_size])
```
Quexington, line 1057: Don't we not need this solution since all of the solution values are curried in?

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,323 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+            intermediate_launcher_coin_spend = CoinSpend(
+                intermediate_launcher_coin, intermediate_launcher_puz, intermediate_launcher_sol
+            )
+            intermediate_coin_spends.append(intermediate_launcher_coin_spend)
+
+            # create an ASSERT_COIN_ANNOUNCEMENT for the DID spend. The
+            # intermediate launcher coin issues a CREATE_COIN_ANNOUNCEMENT of
+            # the mint_number and mint_total for the launcher coin it creates
+            intermediate_announcement_message = std_hash(int_to_bytes(mint_number) + int_to_bytes(mint_total))
+            did_announcements.add(std_hash(intermediate_launcher_coin.name() + intermediate_announcement_message))
```
Quexington, line 1068: I don't think we need any data, just the coin ID implicit in `ASSERT_COIN_ANNOUNCEMENT` is fine.  It doesn't matter so much because it doesn't really result in any extra data on chain but it does make `DID_NFT_LAUNCHER_MOD` a bit smaller

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,323 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+            # coin annnouncement
+            eve_txs = await self.generate_signed_transaction(
+                [uint64(eve_coin.amount)],
+                [p2_inner_puzzle.get_tree_hash()],
+                nft_coin=nft_coin,
+                new_owner=b"",
+                new_did_inner_hash=b"",
+                additional_bundles=[],
+                memos=[[p2_inner_puzzle.get_tree_hash()]],
+                coin_announcements_to_consume=announcement_set,
```
Quexington, line 1127: Like @matt-o-how said, these announcements should be consumed by the spend that creates the launchers, not by any spend afterwards, because launchers are entirely malleable and don't have to create any of their children if they are manipulated.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,323 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+                full_puzzle=eve_fullpuz,
+                mint_height=uint32(0),
+            )
+
+            # Create the eve transaction setting the DID owner, and applying
+            # the announcements from announcement_set to match the launcher
+            # coin annnouncement
+            eve_txs = await self.generate_signed_transaction(
+                [uint64(eve_coin.amount)],
+                [p2_inner_puzzle.get_tree_hash()],
```
Quexington, line 1121: Sine we have to do the eve spend anyways, why don't we do it directly to the `target` specified.  No reason to create a second transaction after this one.  If there is no target specified, we can use our wallet's puzzle hash.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,323 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+        xch_primaries = [
+            AmountWithPuzzlehash({"puzzlehash": xch_change_ph, "amount": change, "memos": [xch_change_ph]})
+        ]
+        for xch_coin in xch_coins:
+            puzzle: Program = await self.standard_wallet.puzzle_for_puzzle_hash(xch_coin.puzzle_hash)
+            solution: Program = self.standard_wallet.make_solution(
+                primaries=xch_primaries,
+                fee=fee,
+            )
+            xch_spends.append(CoinSpend(xch_coin, puzzle, solution))
```
Quexington, line 1210: This loop seems like it shouldn't work which means you need to have a test where you select more than one xch_coin to do this operation.  In this loop, you're forcing every xch_coin to create every output and also it will pay the full fee for each.  You should probably be using the standard wallet's `generate_signed_transaction` it allows you to specify what coins you want to spend.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,323 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+        :param xch_coins: [Optional] For use with bulk minting to provide the coin used for funding the minting spend.
+        This coin can be one that will be created in the future
+        :param xch_change_ph: [Optional] For use with bulk minting, so we can specify the puzzle hash that the change
+        from the funding transaction goes to.
+        :param new_innerpuzhash: [Optional] The new inner puzzle hash for the DID once it is spent. For bulk minting we
+        generally don't provide this as the default behaviour is to re-use the existing inner puzzle hash
+        :param did_coin: [Optional] The did coin to use for minting. Required for bulk minting when the DID coin will
+        be created in the future
+        :param did_lineage_parent: [Optional]  The  parent coin to use for the lineage proof in the DID spend. Needed
+        for bulk minting when the coin will be created in the future
```
Quexington, line 977: When will the DID coin be created in the future?  You can't mint any of these without the DID to create them.

Quexington (changes requested):
Also, as a general comment, I don't see any support for bulk minting DIDless NFTs with this RPC.  Maybe it already exists, I'm not sure, but it's certainly a valid use case.

chia/wallet/nft_wallet/nft_wallet.py
geoffwalmsley: The values are there to make sure that each intermediate launcher coin is unique. If we curry them in and don't use them for anything I'm not sure whether the compiled clvm just drops the values so we end up with non-unique launchers. Have left as is for now, happy to remove if this would work and be preferred though

chia/wallet/nft_wallet/nft_wallet.py
geoffwalmsley: when we're creating a set of spend bundles offline for bulk minting, the DID gets spent once for each batch of minted NFTs. When we create the next spend bundle we have to pass in the right parent info, otherwise the built-in function to construct the lineage fails

chia/wallet/nft_wallet/nft_wallet.py
geoffwalmsley: We're passing in a specific xch coin that might not exist when we're creating the spend bundle (since it will be the change coin that will get created when an earlier spend bundle gets submitted. So the generate_signed_tx method isn't available and hence the "manual" solution. In general when bulk minting you'd set the fee to zero when creating spend bundles, and then add on a fee only when you're actually submitting the spends

chia/wallet/nft_wallet/nft_wallet.py
Quexington: Currying changes the puzzle hash whether you use the values or not, but like I said, it doesn't really matter so I'm happy to leave it as it is.

chia/wallet/nft_wallet/nft_wallet.py
Quexington: I think generate_signed_transaction will let you spend coins even that don't exist yet: https://github.com/Chia-Network/chia-blockchain/blob/7c9fb654a3563e69c54848d137859856c8566587/chia/wallet/wallet.py#L291.  But if for some reason that doesn't work, you still need to fix this loop because it will fail (thankfully) if more than one xch_coins are passed/selected.

Quexington: Actually it won't necessarily fail now that I think about it so the user might end up just spending len(xch_coins)x more money than expected

tests/wallet/nft_wallet/test_nft_mint_from_did.py
```diff
@@ -0,0 +1,626 @@
...
+            full_node_api.full_node.server.node_id.hex(): full_node_api.full_node.server.node_id.hex()
+        }
+    else:
+        wallet_node_0.config["trusted_peers"] = {}
+        wallet_node_1.config["trusted_peers"] = {}
+
+    await server_0.start_client(PeerInfo("localhost", uint16(full_node_server._port)), None)
+    await server_1.start_client(PeerInfo("localhost", uint16(full_node_server._port)), None)
+
+    for _ in range(1, num_blocks):
```
mariano54, line 60: we shouldn't make unnecessary blocks, it slows down tests

tests/wallet/nft_wallet/test_nft_mint_from_did.py
```diff
@@ -0,0 +1,626 @@
...
+        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph_maker))
+    await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph_token))
+
+    funds = sum(
+        [calculate_pool_reward(uint32(i)) + calculate_base_farmer_reward(uint32(i)) for i in range(1, num_blocks)]
+    )
+
+    await time_out_assert(10, wallet_0.get_unconfirmed_balance, funds)
```
mariano54, line 68: time outs to 30 to reduce flakiness, for anything related to wallet balance or TX or syncing

mariano54 (dismissed):

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,296 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+
+        # Create the DID spend using the announcements collected when making the intermediate launcher coins
+        did_p2_solution = self.standard_wallet.make_solution(
+            primaries=primaries,
+            puzzle_announcements=set(launcher_ids),
+            coin_announcements=did_coin_announcement,
+            coin_announcements_to_assert=did_announcements,
+            puzzle_announcements_to_assert=puzzle_assertions,
+        )
+        did_inner_sol: Program = Program.to([1, did_p2_solution])
```
Quexington, line 1192: This must be working, but for future reference I think this should be `(1, did_p2_solution)` which is the difference between `(q . foo)` and `(q foo)`

Quexington (dismissed):

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -930,3 +934,294 @@ async def set_nft_did(self, nft_coin_info: NFTCoinInfo, did_id: bytes, fee: uint
...
+            did_announcements.add(std_hash(intermediate_launcher_coin.name() + intermediate_announcement_message))
+
+            # Create the launcher coin, and add its id to a list to be asserted in the DID spend
+            launcher_coin = Coin(intermediate_launcher_coin.name(), did_wallet_puzzles.LAUNCHER_PUZZLE_HASH, amount)
+            launcher_ids.append(launcher_coin.name())
+
+            # Grab the metadata from metadata_list. The index for metadata_list
+            # needs to be offset by mint_number_start, and since
+            # mint_number starts at 1 not 0, we also subtract 1.
+            metadata = metadata_list[mint_number - mint_number_start - 1]
```
greimela, line 1078: I think this calculation is not correct.
Yes, `mint_number_start` does start at 1, but so does `mint_number`.
So for the first element (`mint_number = 1`, `mint_number_start = 1`), `mint_number - mint_number_start - 1` results in `-1`, not `0`.

trepca (dismissed):

chia/rpc/wallet_rpc_api.py
```diff
@@ -1853,6 +1854,114 @@ async def nft_calculate_royalties(self, request) -> EndpointResult:
...
+            }
+            metadata_list.append(metadata_dict)
+        target_address_list = request.get("target_list", None)
+        target_list = []
+        if target_address_list:
+            for target in target_address_list:
+                target_list.append(decode_puzzle_hash(target))
+        mint_number_start = request.get("mint_number_start", 1)
+        mint_total = request.get("mint_total", None)
+        xch_coin_list = request.get("xch_coins", None)
```
greimela, line 1913: This default case results in an error in the next line, when it tries to iterate over `None`.

```suggestion
        xch_coin_list = request.get("xch_coins", [])
```

geoffwalmsley: Thanks,  good catch

trepca (approved):

PR #12742, opened 2022-08-02
parse the puzzle reveal once, instead of 3 times

arvidn:


chia/wallet/wallet_state_manager.py
```diff
@@ -634,24 +634,25 @@ async def determine_coin_type(
...
         # First spend where 1 mojo coin -> Singleton launcher -> NFT -> NFT
         try:
-            uncurried_nft = UncurriedNFT.uncurry(Program.from_bytes(bytes(coin_spend.puzzle_reveal)))
+            uncurried_nft = UncurriedNFT.uncurry(puzzle)
+            return await self.handle_nft(coin_spend, uncurried_nft)
         except Exception:
             # This is not a NFT coin, skip NFT handling
             pass
-        else:
-            return await self.handle_nft(coin_spend, uncurried_nft)
```
arvidn, line 651: I moved this up into the `try`-block, as I think it simplifies the code (and saves one line)

mariano54: yep, else is confusing, it's not executed in exceptions, but it's not clear

mariano54 (approved):

jack60612 (approved):
less lines, but how much faster is it?

trepca (approved):

arvidn:
> but how much faster is it?

I don't have a good measurement of the speed-up. It's just a thing that shows up in my profiles. Ideally we would have `uncurry()` that would work on `SerializedProgram` directly, so we wouldn't even have to parse it into a python structure

PR #12748, opened 2022-08-02
util: Drop partially implemented BIP39 passphrase support

xdustinface:
It's not fully implemented + we already encrypt the keyring with a separate mechanism.

Right now it's possible to add keys with a BIP39 passphrase via the daemon API but there is no way to use them since we only use the default password (no password, `""`) for decryption so i assume its safe to entirely delete this from the code and (also from the daemon API which breaks compatibility but i think its fine in this case since most likely no one uses/used it).

Note: This adjusts the seeds in the BIP39 test vector file to contain the seed without a passphrase, i.e. without `TREZOR`.

### DIFF
--- a/chia/cmds/keys_funcs.py
+++ b/chia/cmds/keys_funcs.py
@@ -56,8 +56,7 @@ def add_private_key_seed(mnemonic: str):
     """
 
     try:
-        passphrase = ""
-        sk = Keychain().add_private_key(mnemonic, passphrase)
+        sk = Keychain().add_private_key(mnemonic)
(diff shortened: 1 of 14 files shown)


mariano54 (dismissed):

wjblanke:
Would like jeff to review

paninaro (dismissed):
lgtm

wjblanke (dismissed):
approved but should still wait post branch

mariano54 (dismissed):

mariano54 (approved):

PR #12749, opened 2022-08-02
Remove test pypi upload step

pmaslana:


cmmarslender (approved):
Approved as the short term fix to all the failures with test pypi - we have a longer term fix in the backlog to get this working again

PR #12750, opened 2022-08-02
util: Simplify key deletion in `Keychain`

xdustinface:

PR #12747, opened 2022-08-02
util: Raise in `Keychain.add_private_key` if the key already exists

xdustinface:
Based on:
-  #12756

PR #12752, opened 2022-08-02
util: Introduce `KeyData` and `_get_pk_and_entropy` -> `_get_key_data`

xdustinface:
Introduces the dataclass `KeyData` to cache all related key data calculated by the mnemonic in one structure. Preparation for key names + enables some simplifications.

Based on: 
- #12748 
- #12747 
- #12750
- #12751

### DIFF
--- a/chia/util/errors.py
+++ b/chia/util/errors.py
@@ -184,9 +184,9 @@ class KeychainException(Exception):
     pass
 
 
-class KeychainKeyDataMissmatch(KeychainException):
+class KeychainKeyDataMismatch(KeychainException):
     def __init__(self, data_type: str):
-        super().__init__(f"KeyData missmatch for: {data_type}")
+        super().__init__(f"KeyData mismatch for: {data_type}")
 
 
 class KeychainIsLocked(KeychainException):
--- a/chia/util/errors.py
+++ b/chia/util/errors.py
@@ -184,7 +184,16 @@ class KeychainException(Exception):
     pass
 
 
+class KeychainKeyDataMissmatch(KeychainException):
+    def __init__(self, data_type: str):
+        super().__init__(f"KeyData missmatch for: {data_type}")
+
+
 class KeychainIsLocked(KeychainException):
+    pass
+
+
+class KeychainSecretsMissing(KeychainException):
     pass
 
 
@@ -235,6 +244,11 @@ class KeychainLockTimeout(KeychainException):
 class KeychainProxyConnectionTimeout(KeychainException):
     def __init__(self) -> None:
         super().__init__("Could not reconnect to keychain service in 30 seconds.")
+
+
+class KeychainUserNotFound(KeychainException):
+    def __init__(self, service: str, user: str) -> None:
+        super().__init__(f"user {user!r} not found for service {service!r}")
 
 
 class KeychainFingerprintError(KeychainException):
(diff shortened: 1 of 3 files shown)


chia/util/errors.py
```diff
@@ -184,10 +184,19 @@ class KeychainException(Exception):
     pass
 
 
+class KeychainKeyDataMissmatch(KeychainException):
```
paninaro, line 187:
```suggestion
class KeychainKeyDataMismatch(KeychainException):
```

chia/util/errors.py
```diff
@@ -184,10 +184,19 @@ class KeychainException(Exception):
...
+    def __init__(self, data_type: str):
+        super().__init__(f"KeyData missmatch for: {data_type}")
```
paninaro, line 189:
```suggestion
        super().__init__(f"KeyData mismatch for: {data_type}")
```

chia/util/keychain.py
```diff
@@ -132,6 +143,104 @@ def get_private_key_user(user: str, index: int) -> str:
...
+@streamable
+@dataclass(frozen=True)
+class KeyDataSecrets(Streamable):
+    mnemonic: List[str]
+    entropy: bytes
+    private_key: PrivateKey
+
+    def __post_init__(self) -> None:
+        # This is redundant if `from_*` methods are used but its to make sure there can't be an `KeyDataSecrets`
+        # instance with an attribute missmatch for calculated cached values. Should be ok since we don't handle a lot of
```
paninaro, line 156: Missmatch --> mismatch

```suggestion
        # instance with an attribute mismatch for calculated cached values. Should be ok since we don't handle a lot of
```

chia/util/keychain.py
```diff
@@ -132,6 +143,104 @@ def get_private_key_user(user: str, index: int) -> str:
...
+        # keys here.
+        try:
+            bytes_from_mnemonic(self.mnemonic_str())
```
paninaro, line 159: nit: assign to a local variable since this function is called 3 times?

chia/util/keychain.py
```diff
@@ -132,6 +143,104 @@ def get_private_key_user(user: str, index: int) -> str:
...
+
+    @property
+    def entropy(self) -> bytes:
+        if self.secrets is None:
+            raise KeychainSecretsMissing()
+        return self.secrets.entropy
+
+    @property
+    def private_key(self) -> PrivateKey:
+        if self.secrets is None:
```
paninaro, line 239: Can this be moved into `__post_init__()` to avoid the duplicated check for each property/instance-method?

chia/util/keychain.py
```diff
@@ -282,15 +378,16 @@ def delete_key_by_fingerprint(self, fingerprint: int) -> int:
         """
         removed = 0
         for index in range(MAX_KEYS + 1):
-            pkent = self._get_pk_and_entropy(get_private_key_user(self.user, index))
-            if pkent is not None:
-                pk, ent = pkent
-                if pk.get_fingerprint() == fingerprint:
+            try:
+                key_data = self._get_key_data(index)
```
paninaro, line 382: Since neither the private key nor the entropy are used, use includes_secrets=False?

paninaro (changes requested):
Typos and a few optional suggestions.

chia/util/keychain.py
xdustinface: Moving this check into `__post_init__` would make impossible to have an `KeyData` instance without secrets and would anyway also still require some assert then or so to make `mypy` happy.

chia/util/keychain.py
xdustinface: See [adec931](https://github.com/Chia-Network/chia-blockchain/pull/12752/commits/adec93142ac3c4907250f6773d32996766d918cd)

paninaro (approved):
lgtm

PR #12760, opened 2022-08-03
remove seen_bundle_hashes

arvidn:
rather than having a cache of recently seen SpendBundles for the mempool, just use the mempool itself to know whether we've seen a spend bundle or not.

It appears we currently keep the `seen` dictionary in sync with what's in the mempool anyway.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -2064,29 +2064,21 @@ class FullNode:
             return MempoolInclusionStatus.FAILED, Err.NO_TRANSACTIONS_WHILE_SYNCING
 
         if self.mempool_manager.get_spendbundle(spend_name) is not None:
-            self.mempool_manager.remove_seen(spend_name)
             return MempoolInclusionStatus.SUCCESS, None
         if self.mempool_manager.seen(spend_name):
             return MempoolInclusionStatus.FAILED, Err.ALREADY_INCLUDING_TRANSACTION
-        self.mempool_manager.add_and_maybe_pop_seen(spend_name)
         self.log.debug(f"Processing transaction: {spend_name}")
         # Ignore if syncing
         if self.sync_store.get_sync_mode():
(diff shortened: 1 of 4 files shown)


mariano54 (changes requested):
This will have massive performance impact. If you receive a TX from 50 peers at the same time, the `respond_transaction` function will try to process them all at the same time, and the mempool will not be updated yet. The point of the seen dict is that as soon as one asyncio thread sees the TX, the other ones can just return early and not re-process the same tx

arvidn:
I see, so this is to prevent initiating an insertion into the mempool. makes sense.
The line of reasoning I was going down was that this `seen_bundle_hashes` uses an unexpected data structure. it's a dictionary where the value is set to the same as the key. It also has a limit of 10000 hashes. The way it's used right now (afaict) better mean the max mempool size better not fit more than this many transactions.

Perhaps it would make more sense to maintain a set of all spend bundle's that's currently in the process of being validated and added, and when we receive an incoming spend bundle, we check it both against the mempool *and* the ones currently being added.

However, one aspect I haven't understood yet is failing transactions. When we fail to add a transaction to the mempool, we don't seem to leave it in the "seen" set. Afaict, this means we can receive the same (failing) transaction over and over, and initiate adding it every time. Is that right?

mariano54:
Yeah your proposal to only store the ones being validated makes a lot of sense. We can also increase the max size probably.

Yes, a failing transaction is not kept in seen set. I don't remember exactly why, but one reason is that a failing transaction can be successful later. For example, if the receiving node did not yet process the block, they don't have the same state and therefore coins that are required for validating the TX.

it's fine if they keep resubmitting the failed TX. they will be rate limited eventually, and it doesn't affect the whole network, only that one peer getting spammed. We might want to add disconnects after a certain amount of invalid cost is sent from a certain peer.

PR #12762, opened 2022-08-03
minor fixes

arvidn:
remove a redundant enumerate()
remove redundant bytes32 copies
print a warning when inserting a mempool item takes a long time

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -469,10 +469,10 @@ class MempoolManager:
 
         new_item = MempoolItem(new_spend, uint64(fees), npc_result, cost, spend_name, additions, removals, program)
         self.mempool.add_to_pool(new_item)
-        now = time.time()
+        duration = time.time() - start_time
         log.log(
-            logging.DEBUG,
-            f"add_spendbundle {spend_name} took {now - start_time:0.2f} seconds. "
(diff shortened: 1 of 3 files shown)


chia/full_node/mempool_manager.py
```diff
@@ -469,10 +469,10 @@ async def add_spendbundle(
 
         new_item = MempoolItem(new_spend, uint64(fees), npc_result, cost, spend_name, additions, removals, program)
         self.mempool.add_to_pool(new_item)
-        now = time.time()
+        duration = time.time() - start_time
         log.log(
-            logging.DEBUG,
-            f"add_spendbundle {spend_name} took {now - start_time:0.2f} seconds. "
+            logging.DEBUG if duration < 2 else logging.WARNING,
```
arvidn, line 474: is 2 seconds for a mempool insertion a reasonable cut-off?

trepca: FWIW, my full node on a 8 years old mac never shows this going above ~0.3 secs. What's interesting is that higher durations don't seem to be related to higher cost of bundle.

arvidn: yeah, I suspect at those times it's going to be a lot of regular computer noise. scheduling and memory cache.

mariano54 (approved):

PR #12764, opened 2022-08-03
avoid redundant printing of puzzle program in NFT uncurry attempt

arvidn:
The caller of `UncurriedNft.uncurry()` doesn't care about the exception's message, it just interprets it as the puzzle not being an NFT. Formatting the message, especially ones that contain the whole puzzle program is expensive. Since this is the common path through `uncurry()`, it spends the majority of its time serializing and formatting the puzzle program into a string, which is then discarded.

### DIFF
--- a/chia/wallet/nft_wallet/uncurry_nft.py
+++ b/chia/wallet/nft_wallet/uncurry_nft.py
@@ -85,7 +85,7 @@ class UncurriedNFT:
     trade_price_percentage: Optional[uint16]
 
     @classmethod
-    def uncurry(cls: Type[_T_UncurriedNFT], puzzle: Program) -> _T_UncurriedNFT:
+    def uncurry(cls: Type[_T_UncurriedNFT], puzzle: Program) -> Optional[_T_UncurriedNFT]:
         """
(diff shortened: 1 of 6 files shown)


chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -94,18 +94,21 @@ def uncurry(cls: Type[_T_UncurriedNFT], puzzle: Program) -> _T_UncurriedNFT:
         """
         mod, curried_args = puzzle.uncurry()
         if mod != SINGLETON_TOP_LAYER_MOD:
-            raise ValueError(f"Cannot uncurry NFT puzzle, failed on singleton top layer: Mod {mod}")
```
arvidn, line 97: generating this f-string is the expensive part. this patch still allows it to be printed, but only when debug logging is enabled.

trepca (approved):

arvidn:
it turns out that the RPC API makes use of the error message from `uncurry()`. It returns it to the caller. It doesn't seem like a great loss to not print the puzzle program back in the error message though.

PR #12767, opened 2022-08-03
Fix coverage report

almogdepaz:


tests/core/full_node/full_sync/test_full_sync.py
```diff
@@ -346,5 +346,4 @@ async def test_block_ses_mismatch(self, two_nodes, default_1000_blocks, self_hos
             s.new_sub_slot_iters * 2,
         )
         await full_node_2.full_node.sync_from_fork_point(0, 500, peak1.header_hash, summaries2)
-        log.info(f"full node height {full_node_2.full_node.blockchain.get_peak().height}")
-        assert node_height_between(full_node_2, 320, 400)
+        await time_out_assert(180, node_height_exactly, True, full_node_2, 320)
```
almogdepaz, line 349: fixed some flakiness here

altendky:
Closing and reopening to trigger a new merge with `main` in hopes of some NFT test fixes helping this run to completion.

altendky (changes requested):
I think this is not processing the Windows coverage information.  For example, it seems that we are capturing coverage of importing `chia/daemon/windows_signal.py` but we lack coverage for the portion run only on Windows (lines 19-54).

I found what appears to be an import sequence that would load this file for any test run regardless of platform, but to make this a bit more explicit, it might be worth adding a canary test file for coverage where the sole purpose of the file is simply to have sections that are each only executed on one platform.  `test_coverage_canary.py` or such, perhaps.  That would provide an easy reference to remember in any coverage results.

https://github.com/Chia-Network/chia-blockchain/runs/7779353683?check_suite_focus=true#step:7:1308
```
chia/daemon/windows_signal.py                                    28     20      8      1    25.0%   19-54
```

https://github.com/Chia-Network/chia-blockchain/blob/0bc69603c073ec2037f4896a23f750bca6c2d29c/chia/daemon/windows_signal.py#L15-L54

almogdepaz:
> I think this is not processing the Windows coverage information. For example, it seems that we are capturing coverage of importing `chia/daemon/windows_signal.py` but we lack coverage for the portion run only on Windows (lines 19-54).
> 
> I found what appears to be an import sequence that would load this file for any test run regardless of platform, but to make this a bit more explicit, it might be worth adding a canary test file for coverage where the sole purpose of the file is simply to have sections that are each only executed on one platform. `test_coverage_canary.py` or such, perhaps. That would provide an easy reference to remember in any coverage results.
> 
> https://github.com/Chia-Network/chia-blockchain/runs/7779353683?check_suite_focus=true#step:7:1308
> 
> ```
> chia/daemon/windows_signal.py                                    28     20      8      1    25.0%   19-54
> ```
> 
> https://github.com/Chia-Network/chia-blockchain/blob/0bc69603c073ec2037f4896a23f750bca6c2d29c/chia/daemon/windows_signal.py#L15-L54

i dont follow, your saying the 25% is wrong ?

altendky:
Yes, I would expect the Windows-only portions of the file to be covered when imported for the Windows tests.  Though, I did not do a detailed verification that the file is really imported in the Windows tests.  Just a cursory check such that I expect the file is imported.

PR #12768, opened 2022-08-03
rpc: Improve RPC server start/stop

xdustinface:
- Make `RpcServer` properly stoppable
- Refactor `start_rpc_server` to return an `RpcServer`
- Moves the creation of the actual RPC server (`web.TCPSite`) into the method `RpcServer.start` 
- Introduces `RpcEnvironment` which gets used for the new member `RpcServer.environment`
- Adds `Service.rpc_server` and drops `Service._rpc_task`


chia/rpc/rpc_server.py
```diff
@@ -58,12 +67,41 @@ def create(
...
+        #
+        # On a dual-stack system, we want to get the (first) IPv4 port unless
+        # prefer_ipv6 is set in which case we use the IPv6 port
+        #
+        if rpc_port == 0:
+            rpc_port = select_port(root_path, runner.addresses)
+
+        self.environment = RpcEnvironment(runner, site, uint16(rpc_port))
+
+    def stop(self) -> None:
```
mariano54, line 91: Should this be called close instead of stop? Goes with `close` and `await_closed`

chia/server/start_service.py
```diff
@@ -163,19 +163,16 @@ async def start(self, **kwargs) -> None:
...
-                    self.daemon_port,
-                    uint16(rpc_port),
-                    self.stop,
-                    self.root_path,
-                    self.config,
-                    self._connect_to_daemon,
-                    max_request_body_size=self.max_request_body_size,
-                    name=self._service_name + "_rpc",
-                )
+            self.rpc_server = await start_rpc_server(
```
mariano54, line 166: This is a change in behaviour. Before we were not waiting for the rpc server to actually get created, now we are. Do you think this will cause some issues? I don't exactly remember why we did it like this, but I remember there were some issues in this area of the code

mariano54: Perhaps this might even fix the flaky issue?

mariano54:
In general I like these changes, they simplify things

chia/rpc/rpc_server.py
xdustinface: Makes sense, changed in [021f081](https://github.com/Chia-Network/chia-blockchain/pull/12768/commits/021f08174659750538675767d1b11cf991941fc9)!

chia/server/start_service.py
xdustinface: > This is a change in behaviour. Before we were not waiting for the rpc server to actually get created, now we are. Do you think this will cause some issues?

Right, i thought about it when i made the change but i didn't see a reason why it would be an issue. 

> I don't exactly remember why we did it like this, but I remember there were some issues in this area of the code

We are doing it the same way already for the other API's in `ChiaServer.start_server` so i guess its just fine? 

> Perhaps this might even fix the flaky issue?

Yeah maybe, you have your ways of reproducing it, right? Maybe give it a try with this branch :)

mariano54 (approved):
If it starts causing issues we can revert, but all the tests pass so that's a good sign

wjblanke (approved):
i am hesitant to approve this but if it helps our flakiness lets see

jack60612:
ci

PR #12779, opened 2022-08-03
Remove unused ping/pong code

emlowe:
The ClientWebSocketResponse::receive() call internally eats PING, PONG, and CLOSE, so those msgs are never seen by callers

(I checked both the aiohttp code in client_ws.py - but this is also in the documentation: "The coroutine implicitly handles PING, PONG and CLOSE without returning the message.

It process ping-pong game and performs closing handshake internally." (https://docs.aiohttp.org/en/stable/client_reference.html)

### DIFF
--- a/chia/rpc/rpc_server.py
+++ b/chia/rpc/rpc_server.py
@@ -260,6 +260,7 @@ class RpcServer:
         await ws.send_str(payload)
 
         while True:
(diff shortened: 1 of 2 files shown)


jack60612 (approved):
lgtm

mariano54 (approved):

PR #12782, opened 2022-08-04
Fix no keys loaded error by making KeychainProxy automatically reconnect when a connection is lost

### LINKED ISSUE #11175: [Bug] chia wallet service can become stuck
### What happened?

If you
• Have multiple keys
• Are in the middle of syncing a big wallet
• Execute a chia show wallet without specifying the fingerprint of the wallet that is currently syncing
it throws a `No keys loaded. Run 'chia keys generate' or import a key`
with error log:

```
ERROR    Task exception was never retrieved
future: <Task finished name='Task-4524' coro=<ClientWebSocketResponse.send_str() done, defined at /path/to/chia-blockchain/venv/lib/python3.9/site-packages/aiohttp/client_ws.py:147> exception=ConnectionResetError('Cannot write to closing transport')>
  File "/path/to/chia-blockchain/venv/lib/python3.9/site-packages/aiohttp/client_ws.py", line 150, in send_str
    await self._writer.send(data, binary=False, compress=compress)
  File "/path/to/chia-blockchain/venv/lib/python3.9/site-packages/aiohttp/http_websocket.py", line 687, in send
    await self._send_frame(message, WSMsgType.TEXT, compress)
  File "/path/to/chia-blockchain/venv/lib/python3.9/site-packages/aiohttp/http_websocket.py", line 598, in _send_frame
    raise ConnectionResetError("Cannot write to closing transport")
ConnectionResetError: Cannot write to closing transport
```

and it just stays broken
i can only chia wallet show -f to the same wallet i'm already synced to
if i pick a different fingerprint, i get a different error
`chia wallet show -f 233666666`

`Login failed: {'error': "'NoneType' object is not subscriptable", 'success': False}`

and wallet debug spits out

```
2022-04-14T01:52:21.714 wallet chia.rpc.util              : WARNING  Error while handling message: Traceback (most recent call last):
  File "/path/to/chia-blockchain/chia/rpc/util.py", line 16, in inner
    res_object = await f(request_data)
  File "/path/to/chia-blockchain/chia/rpc/wallet_rpc_api.py", line 189, in log_in
    started = await self.service._start(fingerprint)
  File "/path/to/chia-blockchain/chia/wallet/wallet_node.py", line 189, in _start
    private_key = await self.get_key_for_fingerprint(fingerprint)
  File "/path/to/chia-blockchain/chia/wallet/wallet_node.py", line 168, in get_key_for_fingerprint
    key = await keychain_proxy.get_key_for_fingerprint(fingerprint)
  File "/path/to/chia-blockchain/chia/daemon/keychain_proxy.py", line 270, in get_key_for_fingerprint
    response, success = await self.get_response_for_request(
  File "/path/to/chia-blockchain/chia/daemon/keychain_proxy.py", line 93, in get_response_for_request
    success = response["data"].get("success] False)
TypeError: 'NoneType' object is not subscriptable
```

### Version

1.3.2.dev156

Linux

CLI

### Relevant log output

_No response_

jack60612:
Also includes full type passes on `chia/daemon/client.py` & `chia/daemon/keychain_proxy.py` & unexcludes them from mypy


chia/daemon/keychain_proxy.py
```diff
@@ -89,13 +97,80 @@ def format_request(self, command: str, data: Dict[str, Any]) -> WsRpcMessage:
...
+    async def _get(self, request: WsRpcMessage) -> WsRpcMessage:
+        """
+        Overrides DaemonProxy._get() to handle the connection state
+        """
+        try:
+            if not self.disconnect:  # if we are disconnected, and we send a request we should throw original error.
+                await asyncio.wait_for(self.connection_established.wait(), timeout=10)  # is 10 seconds too long?
+            else:
+                self.log.error("Attempting to send request to a keychain-proxy that has shut down.")
+            self.log.debug(f"Sending request to keychain: {request}")
```
paninaro, line 109: There's some sensitive data that can be logged here. We can log the command name but the request's `data` should not be logged.

jack60612: ah yeah very good point.

chia/daemon/client.py
```diff
@@ -70,19 +71,17 @@ async def _get(self, request: WsRpcMessage) -> WsRpcMessage:
             raise Exception("Websocket is not connected")
         asyncio.create_task(self.websocket.send_str(string))
 
-        async def timeout():
+        async def timeout() -> None:
             await asyncio.sleep(30)
             if request_id in self._request_dict:
-                print("Error, timeout.")
-                self._request_dict[request_id].set()
+                self.response_dict.pop(request_id)
```
paninaro, line 77: Shouldn't we still call `self._request_dict[request_id].set()` to signal the caller waiting on `wait()` below?

jack60612: yeah i'm already fixing it, it's what you mentioned earlier

chia/daemon/client.py
```diff
@@ -79,10 +80,10 @@ async def timeout():
         asyncio.create_task(timeout())
         await self._request_dict[request_id].wait()
         if request_id in self.response_dict:
-            response = self.response_dict[request_id]
+            response: WsRpcMessage = self.response_dict[request_id]
             self.response_dict.pop(request_id)
         else:
-            response = None
+            response = None  # type: ignore
```
jack60612, line 86: i wanted to enable mypy for the rest of the file, however the logic already worked with this, so i left it as is.

chia/daemon/keychain_proxy.py
```diff
@@ -89,13 +97,80 @@ def format_request(self, command: str, data: Dict[str, Any]) -> WsRpcMessage:
...
 
+    async def _get(self, request: WsRpcMessage) -> WsRpcMessage:
+        """
+        Overrides DaemonProxy._get() to handle the connection state
+        """
+        try:
+            if not self.disconnect:  # if we are disconnected, and we send a request we should throw original error.
+                await asyncio.wait_for(self.connection_established.wait(), timeout=10)  # is 10 seconds too long?
+            else:
+                self.log.error("Attempting to send request to a keychain-proxy that has shut down.")
```
paninaro, line 108: After logging, should we continue trying to send the request since presumably we're trying to shut down the proxy?

jack60612: yes we should, as we check it in the super classes _get function

jack60612: I just wanted to add some additional logic for clarity

chia/daemon/keychain_proxy.py
```diff
@@ -89,13 +97,80 @@ def format_request(self, command: str, data: Dict[str, Any]) -> WsRpcMessage:
...
+            self.log.debug(f"Sending request to keychain command: {request['command']} from {request['origin']}.")
+            return await super()._get(request)
+        except asyncio.TimeoutError:
+            raise KeychainProxyConnectionFailure("Could not reconnect to Keychain!")
+
+    async def start(self) -> None:
+        self.keychain_connection_task = asyncio.create_task(self.connect_to_keychain())
+        await self.connection_established.wait()  # wait until connection is established.
+
+    async def connect_to_keychain(self) -> None:
```
paninaro, line 118: Can we add some logging in here to indicate that we're attempting to connect/reconnect?

chia/daemon/keychain_proxy.py
```diff
@@ -89,13 +97,80 @@ def format_request(self, command: str, data: Dict[str, Any]) -> WsRpcMessage:
...
+            self.connection_established.clear()
+            if self.websocket is not None:
+                await self.websocket.close()
+            if self.client_session is not None:
+                await self.client_session.close()
+            self.websocket = None
+            self.client_session = None
+            await asyncio.sleep(2)
+
+    async def connection_listener(self) -> None:
```
paninaro, line 145: The logic in here seems basically identical to DaemonProxy's listener() method, with the exception of calling `self.connection_established.set()` and the "Close signal received..." log message. Can we leverage super.listener() to avoid duplicating code?

jack60612: it calls self.close so we sadly can't

paninaro: Maybe we can move the common part into a new method? This part specifically:

```                if message.type == WSMsgType.TEXT:
                    decoded: WsRpcMessage = json.loads(message.data)
                    request_id = decoded["request_id"]

                    if request_id in self._request_dict:
                        self.response_dict[request_id] = decoded
                        self._request_dict[request_id].set()```

File: chia/daemon/keychain_proxy.py (hunk 5 above)
Comment (jack60612) [reply chia/daemon/keychain_proxy.py:117]: added

File: chia/daemon/keychain_proxy.py (hunk 6 above)
Comment (jack60612) [reply chia/daemon/keychain_proxy.py]: done

**Review by paninaro — APPROVED** [review]
lgtm. Tested a couple of artificial scenarios that would cause a timeout and the reconnect logic is working.

File: chia/daemon/keychain_proxy.py [hunk 7]
```diff
@@ -89,13 +96,70 @@ def format_request(self, command: str, data: Dict[str, Any]) -> WsRpcMessage:
...
+        """
+        try:
+            if not self.disconnect:  # if we are disconnected, and we send a request we should throw original error.
+                await asyncio.wait_for(self.connection_established.wait(), timeout=10)  # is 10 seconds too long?
+            else:
+                self.log.error("Attempting to send request to a keychain-proxy that has shut down.")
+            self.log.debug(f"Sending request to keychain command: {request['command']} from {request['origin']}.")
+            return await super()._get(request)
+        except asyncio.TimeoutError:
+            raise KeychainProxyConnectionFailure("Could not reconnect to Keychain!")
```
altendky, line 111: This is specifically a timeout which is good to express as distinct from a no response error or such.

altendky (commented):
The new reconnect code seems, at a glance, to be pretty general.  Can it not provide reconnect behavior for 'everything' instead of just the keychain proxy?  Though that question may be a good indication of my (lack of) understanding of the structure around the proxy.

chia/daemon/client.py
```diff
@@ -45,23 +45,27 @@ async def start(self):
             await self.close()
             raise
 
-        async def listener():
+        async def listener_task() -> None:
+            await self.listener()
+            await self.close()
```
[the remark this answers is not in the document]
jack60612: Yep

chia/daemon/client.py
```diff
@@ -45,23 +45,27 @@ async def start(self):
...
+
+        asyncio.create_task(listener_task())
+        await asyncio.sleep(1)
+
+    async def listener(self) -> None:
+        if self.websocket is not None:
```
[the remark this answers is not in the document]
jack60612: It would never be none, but mypy needed it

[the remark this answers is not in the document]
altendky: Then we would _really_ want to know that something impossible has happened...  Also, when you are placating mypy and you don't have a link to a mypy issue or pr that would eliminate the need to placate mypy, then it is usually an indication that there's a design issue in our code.  Albeit maybe pre-existing here.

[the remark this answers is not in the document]
altendky: Also, instead of ending up indenting everything for an impossible case, handle the impossible case on the side most likely by raising an exception.

Also, also, don't assume stuff controlled outside that method is impossible to get into a state that is in fact sometimes the state of it.  We want code that helps you when you break expectations that have been made.  Not code that silently does nothing when you modify it and end up violating an assumption.

PR #12783, opened 2022-08-04
No network type

mariano54:
Removes the concept of NetworkType, which was necessary before (when mainnet had transactions disabled). Now, there is no difference between different networks in terms of code, so this complexity can be removed permanently.

This has been tested manually with old configs, and it works fine.


chia/farmer/farmer_api.py
```diff
@@ -30,6 +29,7 @@
 from chia.ssl.create_ssl import get_mozilla_ca_crt
 from chia.types.blockchain_format.pool_target import PoolTarget
 from chia.types.blockchain_format.proof_of_space import ProofOfSpace
+from chia.types.blockchain_format.sized_bytes import bytes32
```
xdustinface, line 32:
```suggestion
```

xdustinface (dismissed):
There are still a bunch of `NETWORK_TYPE` entries in `initial_config.yaml` and you missed to drop the `bytes32` import after your changes to use `selected_network`.

jack60612:
is this ready for review?

jack60612 (dismissed):
lgtm, glad you stripped it from the config

jack60612 (approved):
LGTM, that's probably a bit clearer

PR #12642, opened 2022-07-28
Update pool config only after sync finishes

jack60612:
Now that our wallet can sync faster I think this a simple and easy change.
This has not been locally tested yet, but once I do have a chance to test it i will undraft it.
This closes #12631

PR #12796, opened 2022-08-05
add delayed pool config update feature

### LINKED ISSUE #12631: Delay updating pool config for plotnfts until the wallet is fully synced [Enhancement]
### What happened?

Currently, when a wallet resyncs from scratch, all associated pool wallets (plotnfts) "replay" their history over all plotnft changes: whenever a plotnft changes state, the pool config (in config.yaml) is updated. This results in a farmer sending partials to wrong (or even long-gone) pools until the wallet syncs past the last plotnft changes.

I suggest to defer plotnft pool config updates while the walled is still syncing and only update the plotnft pool config once, when the wallet is fully synced up.

### Version

1.5.0

Linux

CLI

### Relevant log output

_No response_

jack60612:
Now that our wallet can sync faster I think this a simple and easy change.
Tested on my personal farmer, works perfectly.
This closes https://github.com/Chia-Network/chia-blockchain/issues/12631

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -401,6 +401,7 @@ class PoolWallet:
         self.wallet_info = wallet_info
         self.target_state = None
         self.log = logging.getLogger(name if name else __name__)
+        await self.update_pool_config()
         return self
 
     @staticmethod
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -1,3 +1,4 @@
+import asyncio
 import dataclasses
 import logging
 import time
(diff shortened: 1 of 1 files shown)


jack60612:
replaces #12642

jack60612:
Tested on my personal farmer, works perfectly

mariano54 (changes requested):
Let's say you're syncing, then you sync up to the pool wallet on chain, so it get's queued up (but not written to disk) then you shut down the wallet and farmer before sync is done. 
When you start it back up, the pool config will be stuck in the old state permanently

mariano54 (approved):

PR #12798, opened 2022-08-05
optimize simple_solution_generator()

arvidn:
this is a somewhat expensive part of adding a transaction to the mempool. With the planned raise of the mempool size this is an attempt to mitigate load- and performance issues caused by higher pressure on the mempool.

The benchmark shows a small improvement on wall-clock time, but it's CPU time in the main process that's the more precious resource. We already limit validation to 2 threads.

CPU usage:

| CPU usage | before | after | after / before |
| --- | --- | --- | --- |
| mempool_manager.add_spendbundle() | 13.58% | 6.12% | 0.45 |

main:
```
add_spendbundle time: 2.1495s
2.15ms per add_spendbundle() call
```

this patch:
```
add_spendbundle time: 2.0407s
2.04ms per add_spendbundle() call
```

# profiles

main:

mempool-add-mt

this patch:

mempool-add-mt

### DIFF
--- a/chia/full_node/bundle_tools.py
+++ b/chia/full_node/bundle_tools.py
@@ -1,9 +1,6 @@
 import re
 from typing import Optional, Tuple, List, Union
 
-from clvm import SExp
-from clvm_tools import binutils
-
 from chia.full_node.generator import create_compressed_generator
 from chia.types.blockchain_format.program import SerializedProgram, Program
 from chia.types.coin_spend import CoinSpend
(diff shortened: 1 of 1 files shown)


mariano54:
Nice!

We also call `simple_solution_generator(new_spend).program` in the prevalidation, so maybe we should send that over to not have to recompute it.

arvidn:
> We also call simple_solution_generator(new_spend).program in the prevalidation, so maybe we should send that over to not have to recompute it.

Yeah, I was considering that patch also. but after making this change it disappeared from the profile, so I'm thinking it might be a wash of sending it over the pipe or regenerating it

richardkiss (dismissed):
Seems straightforward enough. Maybe keep the old code as comments to document exactly what's happening (or write in English – but keeping old code is probably easier).

arvidn:
yeah, I agree my comments are a bit sparse

richardkiss (approved):

trepca (approved):

PR #12801, opened 2022-08-05
util: Force keyring migration / Deprecate legacy keyring support

xdustinface:
After this PR the users will be forced to migrate their keyring if they didn't yet. This will only happen on CLI level if the user runs any of the following commands: `chia keys`, `chia passphrase`, `chia wallet`, `chia start`. The message they will see then is:

```
Legacy keyring support is deprecated and gets removed in version 1.5.2, you need to migrate your keyring
to continue using chia.

Your existing keys will be migrated to a new keyring that is optionally secured by a master passphrase.
Would you like to set a master passphrase now? Use 'chia passphrase set' to change the passphrase.
```

Aborting the migration will result in the CLI exiting with:

```
Migration aborted, can't continue running any chia command.
```

I tested all of this for both migration paths (full migration required and partially migration required).

### DIFF
--- a/chia/cmds/keys_funcs.py
+++ b/chia/cmds/keys_funcs.py
@@ -188,7 +188,7 @@ def migrate_keys(forced: bool = False):
     from chia.util.misc import prompt_yes_no
 
     deprecation_message = (
-        "\nLegacy keyring support is deprecated and is getting removed in version 1.5.2, "
+        "\nLegacy keyring support is deprecated and will be removed in version 1.5.2. "
         "you need to migrate your keyring to continue using chia."
     )
 
--- a/chia/cmds/keys_funcs.py
+++ b/chia/cmds/keys_funcs.py
@@ -189,7 +189,7 @@ def migrate_keys(forced: bool = False):
 
     deprecation_message = (
         "\nLegacy keyring support is deprecated and will be removed in version 1.5.2. "
-        "you need to migrate your keyring to continue using chia."
+        "You need to migrate your keyring to continue using Chia."
     )
 
     # Check if the keyring needs a full migration (i.e. if it's using the old keyring)
--- a/chia/cmds/keys_funcs.py
+++ b/chia/cmds/keys_funcs.py
@@ -183,48 +183,55 @@ def verify(message: str, public_key: str, signature: str):
     print(AugSchemeMPL.verify(public_key, messageBytes, signature))
 
 
-def migrate_keys():
+def migrate_keys(forced: bool = False):
     from chia.util.keyring_wrapper import KeyringWrapper
     from chia.util.misc import prompt_yes_no
 
+    deprecation_message = (
+        "\nLegacy keyring support is deprecated and is getting removed in version 1.5.2, "
+        "you need to migrate your keyring to continue using chia."
+    )
+
     # Check if the keyring needs a full migration (i.e. if it's using the old keyring)
     if Keychain.needs_migration():
+        print(deprecation_message)
         KeyringWrapper.get_shared_instance().migrate_legacy_keyring_interactive()
     else:
         keys_to_migrate, legacy_keyring = Keychain.get_keys_needing_migration()
         if len(keys_to_migrate) > 0 and legacy_keyring is not None:
+            print(deprecation_message)
             print(f"Found {len(keys_to_migrate)} key(s) that need migration:")
             for key, _ in keys_to_migrate:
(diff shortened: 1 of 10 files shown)


chia/util/keyring_wrapper.py
```diff
@@ -506,11 +504,9 @@ def migrate_legacy_keyring_interactive(self):
...
         from chia.cmds.passphrase_funcs import async_update_daemon_migration_completed_if_running
 
-        # Make sure the user is ready to begin migration.
-        response = self.confirm_migration()
-        if not response:
-            print("Skipping migration. Unable to proceed")
-            exit(0)
+        # Let the user know about the migration.
+        if not self.confirm_migration():
+            exit("Migration aborted, can't continue running any chia command.")
```
jack60612, line 509: maybe this can be changed to `"Migration aborted, can't run any chia commands."`

jack60612 (changes requested):
mostly looks good, what happens if we run the ci without that flag?

jack60612:
> mostly looks good, what happens if we run the ci without that flag?

i tested ci without flag and it all passes, so can we not include that change?

xdustinface:
> i tested ci without flag and it all passes, so can we not include that change?

Thanks, there were indeed some redundant flag additions left because i initially forced the migration for all `chia` commands, now it's just a subset on a subset and in this test it's now required for all `chia keys` calls.

jack60612:
ci

tests/core/cmds/test_keys.py
```diff
@@ -162,7 +178,14 @@ def test_generate_with_existing_config(self, tmp_path, empty_keyring):
         # Generate the new config
         runner = CliRunner()
         init_result: Result = runner.invoke(
-            cli, ["--root-path", os.fspath(tmp_path), "--keys-root-path", os.fspath(keys_root_path), "init"]
+            cli,
+            [
```
jack60612, line 182: you need to restore the old formatting

tests/core/cmds/test_keys.py
```diff
@@ -242,7 +283,14 @@ def test_add_interactive(self, tmp_path, empty_keyring):
 
         runner = CliRunner()
         init_result: Result = runner.invoke(
-            cli, ["--root-path", os.fspath(tmp_path), "--keys-root-path", os.fspath(keys_root_path), "init"]
+            cli,
+            [
```
jack60612, line 287: same thing

tests/core/cmds/test_keys.py
```diff
@@ -268,7 +324,14 @@ def test_add_from_mnemonic_seed(self, tmp_path, empty_keyring, mnemonic_seed_fil
...
```
jack60612, line 328: again

tests/core/cmds/test_keys.py
```diff
@@ -302,7 +366,14 @@ def test_delete(self, tmp_path, empty_keyring, mnemonic_seed_file):
...
```
jack60612, line 370: another one here. I won't mark any more because that would be kinda dumb but please fix all of them.

jack60612 (dismissed):
still just a small nitpick

wallentx:
@xdustinface if this is intended to be in the 1.5.1 release, please target the `release/1.5.1` branch

chia/cmds/keys_funcs.py
```diff
@@ -183,49 +183,56 @@ def verify(message: str, public_key: str, signature: str):
     print(AugSchemeMPL.verify(public_key, messageBytes, signature))
 
 
-def migrate_keys():
+def migrate_keys(forced: bool = False):
     from chia.util.keyring_wrapper import KeyringWrapper
     from chia.util.misc import prompt_yes_no
 
+    deprecation_message = (
+        "\nLegacy keyring support is deprecated and is getting removed in version 1.5.2, "
```
paninaro, line 191:
```suggestion
        "\nLegacy keyring support is deprecated and will be removed in version 1.5.2. "

File: chia/cmds/keys_funcs.py [hunk 7]
```diff
@@ -183,49 +183,56 @@ def verify(message: str, public_key: str, signature: str):
...
+        "you need to migrate your keyring to continue using chia."
```
paninaro, line 192:
```suggestion
        "You need to migrate your keyring to continue using Chia."
```

paninaro (dismissed):
lgtm

paninaro (approved):
lgtm

PR #12811, opened 2022-08-06
Add set wallet name API for NFT wallet

ytx1991:


chia/wallet/did_wallet/did_wallet.py
```diff
@@ -516,7 +516,6 @@ def get_my_DID(self) -> str:
 
     async def set_name(self, new_name: str):
         import dataclasses
```
trepca, line 518: why are we importing inline?

PR #12812, opened 2022-08-06
Adding Chia-Network/required-reviewers as CODEOWNERS

wallentx:
This just adds the @Chia-Network/required-reviewers user group as codeowners for the entire contents of the repo, which will allow me to set a branch protection rule to require PR approval from CODEOWNERS before merging

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*	@Chia-Network/required-reviewers


altendky:
If I understand correctly this will not allow the delayed notification to the required reviewers that we have now?  It's not necessarily an unacceptable trade-off, but might be worth considering how we could help them know when to look at a PR.

wallentx:
> If I understand correctly this will not allow the delayed notification to the required reviewers that we have now? It's not necessarily an unacceptable trade-off, but might be worth considering how we could help them know when to look at a PR.

There are some team settings that can be set to fine-tune how individuals in this group are notified (There's also a reminder setting that would be nice, but is limited to a slack integration). In the automation that I'm wanting to add, and in support of the PR triage view dashboards I want to provide, the most important thing needed is that the status be "always red" until all requirements are met, because a "green" status will be the trigger for state transition.
Our usage of `ready_for_merge` is a bit untrue at the moment, but that was a known placeholder solution until we could have a state where "green"/able-to-be-merged-without-an-override actually meant that the PR could be merged.

If we have merge requirements, I'd like them to be required at all times.

altendky:
I'll note that it appears that if other code owners are added they will _replace_ these global owners for the matched files.  As opposed to stacking.  https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

How will we address the need for different reviewers on different long lived branches?  For example, would we plan to just change the code owners file created here for atari to reference a group with myself and Earle?  I think that would work, just checking on expectations.

wallentx:
The rules would be applied via branch protection rules, with my goal being to only apply this to main, and probably release/*.
All other branches would operate as they currently do.

altendky (approved):

wjblanke (approved):
aok

PR #12815, opened 2022-08-06
Fee estimation interface

aqk:
This is the RPC API, command line interface, and the Python Protocol to hook into the Mempool.

See https://github.com/Chia-Network/chia-blockchain/pull/12827


chia/full_node/mempool.py
```diff
@@ -55,6 +62,14 @@ def remove_from_pool(self, item: MempoolItem):
         self.total_mempool_cost -= item.cost
         assert self.total_mempool_cost >= 0
 
+        mempool_info = FeeMempoolInfo(
+            uint64(self.max_size_in_cost),
+            uint64(self.minimum_fee_per_cost_to_replace),
+            uint64(self.total_mempool_cost),
+            datetime.now(),
+        )
```
paninaro, line 70: Minor suggestion -- add a `feeSample()` method to create the `FeeMempoolInfo` instance, since this logic duplicated in `add_to_pool()`.

paninaro: Also totally minor, but I was wondering if there's utility in having the caller provide the timestamp, or whether it would be better to create that in the FeeMempoolInfo constructor.

chia/policy/fee_estimation.py
```diff
@@ -0,0 +1,38 @@
...
+        current_mempool_cost (uint64):This is the current capacity of the mempool, measured in XCH per CLVM Cost
+        max_size_in_cost (uint64): This is the maximum capacity of the mempool, measured in XCH per CLVM Cost
+        minimum_fee_per_cost_to_replace (uint64): Smallest FPC that  might be accepted to replace another SpendBundle
+        time (datetime): Local time this sample was taken
+    """
+
+    max_size_in_cost: uint64  # Mempool max allowed CLVM cost total
+    minimum_fee_per_cost_to_replace: uint64
+    current_mempool_cost: uint64  # Current sum of CLVM cost of all SpendBundles in mempool (mempool "size")
+    time: datetime  # Local time this sample was taken
```
paninaro, line 22: Minor: If you wanted to make it optional to pass this in to the constructor:

`from dataclasses import dataclass, field`

`time: datetime = field(default_factory=datetime.now)`

chia/policy/fee_estimator.py
```diff
@@ -0,0 +1,33 @@
...
+    def new_block(self, block_info: FeeBlockInfo) -> None:
+        pass
+
+    def add_mempool_item(self, mempool_item_info: FeeMempoolInfo, mempool_item: MempoolItem) -> None:
+        pass
+
+    def remove_mempool_item(self, mempool_info: FeeMempoolInfo, mempool_item: MempoolItem) -> None:
+        pass
+
+    def estimate_fee(self, *, cost: int, time: int) -> uint64:
```
paninaro, line 26: Should cost be uint64?

chia/policy/fee_estimator_demo.py
```diff
@@ -0,0 +1,35 @@
...
+from chia.util.ints import uint64
+
+MIN_MOJO_PER_COST = 5
+
+
+def demo_fee_rate_function(cost: int, time_in_seconds: int) -> uint64:
+    return uint64(cost * MIN_MOJO_PER_COST * max((3600 - time_in_seconds), 1))
+
+
+class FeeEstimatorDemo:  # FeeEstimatorInterface Protocol
```
paninaro, line 13: Should this be adopting the FeeEstimatorInterface in some manner? I'm not familiar enough with python's best practices in this area.

chia/policy/fee_estimator_zero.py
```diff
@@ -0,0 +1,20 @@
+from chia.policy.fee_estimation import FeeBlockInfo, FeeMempoolInfo
+from chia.policy.fee_estimator import FeeEstimatorConfig
+from chia.util.ints import uint64
+
+
+class FeeEstimatorZero:  # FeeEstimatorInterface Protocol
```
paninaro, line 6: Missing `mempool_size()` and `mempool_max_size()` methods? Not sure it matters for this implementation.

tests/policy/test_fee_estimation_rpc.py
```diff
@@ -0,0 +1,134 @@
+from typing import Tuple
+
+import pytest
+
+from chia.rpc.full_node_rpc_api import FullNodeRpcApi
+from chia.rpc.full_node_rpc_client import FullNodeRpcClient
+from chia.types.blockchain_format.coin import Coin
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.ints import uint64
+from tests.wallet_tools import WalletTool
```
paninaro, line 10: Looks like this has moved to chia.simulator.wallet_tools

chia/policy/fee_estimator_zero.py
aqk: Yes, I'll remove this demo estimator in the final patch.

chia/policy/fee_estimator_demo.py
aqk: Ah, the way this works is:

If you declare a variable elsewhere as type FeeEstimatorInterface, python will throw if the object you pass does not conform to the protocol object. Unfortunately, that is a bit implicit.

However, I think that in addition to this python Protocol, we should also have Estimators that plug into our infrastructure inherit from an object that will enforce some of the algorithmic things we expect.

Normally I would avoid inheritance for behaviour sharing, but I think it might help to enforce some of our invariants (e.g. no negative estimates).

chia/policy/fee_estimator.py
aqk: Yes, that would be better, thanks.

PR #12817, opened 2022-08-06
Fix a migration bug of NFT table change

ytx1991:


wallentx (dismissed):
:shipit: Ty!

mariano54:
Would be better to Rebase just one commit on top of the branch, there are way too many commits in this PR

chia/wallet/wallet_nft_store.py
```diff
@@ -115,7 +115,7 @@ async def get_nft_list(
                 None if row[2] is None else LineageProof.from_json_dict(json.loads(row[2])),
                 Program.from_bytes(row[5]),
                 uint32(row[3]),
-                uint32(row[6]),
+                uint32(row[6]) if row[6] else uint32(0),
```
altendky, line 118:
```suggestion
                uint32(row[6]) if row[6] is not None else uint32(0),
```

chia/wallet/wallet_nft_store.py
```diff
@@ -139,7 +139,7 @@ async def get_nft_by_id(self, nft_id: bytes32) -> Optional[NFTCoinInfo]:
             None if row[2] is None else LineageProof.from_json_dict(json.loads(row[2])),
             Program.from_bytes(row[5]),
             uint32(row[3]),
-            uint32(row[6]),
+            uint32(row[6]) if row[6] else uint32(0),
```
altendky, line 142:
```suggestion
            uint32(row[6]) if row[6] is not None else uint32(0),
```

altendky (commented):
I would recommend being specific about the check rather than just 'any falsey anything of any type'.  In this case, yes, it's presumably either `None`, 0, or a non-zero integer and it is ok, but I think this form is better practice in general.

Anyways, thanks for the fix!

Gbolahan657:
If you truly need help contact hackerdecon936@gmail.com

wjblanke (approved):
aok

PR #12818, opened 2022-08-06
fix a few instances of coin name logging

arvidn:
that would print binary-escaped strings instead of hex.

These are regressions from when I moved `Coin` to rust. `Coin.name()` now returns `bytes` instead of `bytes32`. One difference is that `bytes32` prints as hex by default, whereas `bytes` doesn't.


mariano54 (approved):

Gbolahan657:
If you truly need help contact hackerdecon936@gmail.com,

PR #12821, opened 2022-08-06
Change client connection errors from info to error

chiameh:
Resubmitting PR #10929. Log farmer and harvester connection failures are errors instead of informational.

Connection errors from harvester and farmer should be logged as errors instead of info, so problems with farming are easier to identify as they are hidden by the default log level INFO.


mariano54:
We should check that the node type is HARVESTER or FARMER

PR #12822, opened 2022-08-07
pre-commit: Fix a venv activation issue on linux

xdustinface:
I ended up with the same issue as in https://github.com/Chia-Network/chia-blockchain/pull/12235 and using `.` instead of `source` (Which is supposed to be the same?) fixed it.

### DIFF
--- a/activated.py
+++ b/activated.py
@@ -18,7 +18,7 @@ def main(*args: str) -> int:
         command = ["powershell", os.fspath(here.joinpath(script)), *args]
     else:
         script = "activated.sh"
-        command = [os.fspath(here.joinpath(script)), *args]
+        command = ["sh", os.fspath(here.joinpath(script)), *args]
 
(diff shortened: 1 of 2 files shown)


wallentx:
> I ended up with the same issue as in #12235 and using `.` instead of `source` (Which is supposed to be the same?) fixed it.

On Ubuntu (and maybe other distros) `sh` is symlinked to `dash`, which doesn't understand `source`

altendky:
We do use `.` everywhere else.  I very much dislike the illegibility and unsearchability of it, but someone thought it was a nice shortcut way back when and so we're kinda stuck.  Anyways, I was thinking we'd switch the `#!` to `sh` as well and update the `.py` to specify `sh` explicitly similar to how it specifies `powershell`.

altendky (approved):

wjblanke (approved):
aok

PR #12827, opened 2022-08-08
Fee Estimator Wallet Protocol

aqk:
See https://github.com/Chia-Network/chia-blockchain/pull/12815


chia/full_node/mempool.py
```diff
@@ -15,6 +16,7 @@ def __init__(self, max_size_in_cost: int):
         self.removals: Dict[bytes32, MempoolItem] = {}
         self.max_size_in_cost: int = max_size_in_cost
         self.total_mempool_cost: int = 0
+        self.fee_estimator = FeeEstimatorDemo()
```
arvidn, line 19: "Demo" makes it sound like this is temporary, and not production. But, yet, it's hooked up to production code

chia/policy/fee_estimator_demo.py
```diff
@@ -0,0 +1,20 @@
+from typing import List
+
+from chia.full_node.fee_estimate import FeeEstimate
+from chia.util.ints import uint64
+
+MIN_MOJO_PER_COST = 5
+
+
+def demo_fee_rate_function(cost: int, time_in_seconds: int) -> uint64:
+    return uint64(cost * MIN_MOJO_PER_COST * max((3600 - time_in_seconds), 1))
```
arvidn, line 10: I see. This is a place holder for the actual fee-estimator logic. Right?

arvidn (approved):

chia/full_node/mempool.py
aqk: Yes, what I did was split this PR into 3 parts so it would be easier to review.

But I needed some glue code at the points where the other PRs connect.

Maybe I should also push this as one giant PR as well, so we'll have the exact code we will commit.

chia/policy/fee_estimator_demo.py
aqk: That's right.

PR #12830, opened 2022-08-08
util: Introduce `get_key`/`get_keys` in `Keychain` + `KeychainServer`

xdustinface:
This is preparation for the key name support where in a follow up the key name gets added to `KeyData`.  Adds the two new methods:
- `get_key` to fetch a `KeyData` entry based on a fingerprint
- `get_keys` to fetch the `KeyData` entries of all keys in the keychain. 

There is a parameter `include_secrets` (defaults to false) in the related methods and the server requests which allows to hide/show the secrets.

Based on: 
- #12772 
- #12752

### DIFF
--- a/chia/daemon/keychain_server.py
+++ b/chia/daemon/keychain_server.py
@@ -2,9 +2,13 @@ import logging
 
 from blspy import PrivateKey
 from chia.cmds.init_funcs import check_keys
-from chia.util.keychain import Keychain
+from chia.util.errors import KeychainException, KeychainFingerprintNotFound
+from chia.util.ints import uint32
+from chia.util.keychain import KeyData, Keychain
+from chia.util.streamable import streamable, Streamable
 from pathlib import Path
-from typing import Any, Dict, List, Optional, cast
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, cast, Type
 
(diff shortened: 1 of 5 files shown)


chia/daemon/keychain_server.py
```diff
@@ -26,6 +32,33 @@
...
+
+def get_key_command(keychain: Keychain, request: GetKeyRequest) -> Dict[str, Any]:
+    return {
+        "success": True,
+        "key": keychain.get_key(request.fingerprint, request.include_secrets).to_json_dict(),
+    }
+
+
+def get_keys_command(keychain: Keychain, request: GetKeysRequest) -> Dict[str, Any]:
+    return {
```
paninaro, line 56: I like the compactness of the new RPC implementations, though it's hard to figure out the shape of the response data beyond the top-level keys. A comment with an example output might be sufficient, or maybe we should have a GetKeysResponse(Streamable) class?

xdustinface: > or maybe we should have a GetKeysResponse(Streamable) class?

Yeah good idea, i changed it and also slightly modified the request class (it contains the `run` now which was a separate function before and passed into `run_command`.

paninaro (approved):
lgtm

PR #12843, opened 2022-08-09
util: Implement key label support for `FileKeyring` and `KeyringWrapper`

xdustinface:
This PR implements key label support for the `FileKeyring` and `KeyringWrapper` classes. The logic to restrict adding labels for key's that don't exist in the `FileKeyring` will come to `Keychain` in a follow up PR since the `FileKeyring` in its current version doesn't know about the actual data stored in it. This can be changed later after the legacy keyring support was removed. 

There will be follow up PRs for:

- `Keychain` + `KeychainServer` 
- CLI
- RPC

Based on:
- #12669


chia/util/file_keyring.py
```diff
@@ -244,6 +253,44 @@ def delete_password(self, service: str, user: str) -> None:
...
+        """
+        with self.lock_and_reload_if_required():
+            return self.cached_labels().get(fingerprint)
+
+    def set_label(self, fingerprint: int, label: str) -> None:
+        """
+        Set a label for the given fingerprint. This will force a write to keyring.yaml on success.
+        """
+        # First validate the label
+        if len(label.strip()) == 0:
```
paninaro, line 268: I would strip the label in case the user accidentally added a space or two before/after the name.

```suggestion
        clean_label = label.strip()
        if len(clean_label) == 0:
```

paninaro: The other checks (against the stripped label) look ok to me.

chia/util/file_keyring.py
```diff
@@ -19,12 +19,14 @@
...
 
 from chia.util.default_root import DEFAULT_KEYS_ROOT_PATH
+from chia.util.errors import KeychainFingerprintNotFound, KeychainLabelExists, KeychainLabelInvalid
 from chia.util.lock import Lockfile
 
 SALT_BYTES = 16  # PBKDF2 param
 NONCE_BYTES = 12  # ChaCha20Poly1305 nonce is 12-bytes
 HASH_ITERS = 100000  # PBKDF2 param
 CHECKBYTES_VALUE = b"5f365b8292ee505b"  # Randomly generated
+MAX_LABEL_LENGTH = 30
```
paninaro, line 29: I think this is unreasonably short. At a minimum I would use a length of 64 to allow users to specify a hash or XCH address as the label (as strange as that might sound).

tests/core/util/test_keyring_wrapper.py
```diff
@@ -482,3 +484,97 @@ def test_passphrase_hint_update(self):
             KeyringWrapper.get_shared_instance().get_master_passphrase_hint()
             == "Something you wouldn't expect The Shredder to say"
         )
+
```
paninaro, line 487: I would add test cases for labels containing Emoji, Japanese text, and an RTL language like Arabic/Hebrew.

paninaro (changes requested):
Looks good. Just a few things to look at.

chia/util/file_keyring.py
xdustinface: See e906fcd265c2f463daa878724b217e581d9728d9

chia/util/file_keyring.py
xdustinface: See a7fc8b9710c180146d0c9fbdc0d2cf2fb0113a65

tests/core/util/test_keyring_wrapper.py
xdustinface: See f126660655451c42c06ece4994a9cd09e2ec385a

xdustinface:
> Looks good. Just a few things to look at.

Thanks, addressed the suggestions.

paninaro (approved):
Approved!

wallentx:
This needs new CI changes for OSX in order for the build to work. Merging past this

PR #12850, opened 2022-08-09
util: Simplify `unlocks_keyring` and move it with `obtain_current_passphrase` into `cmds`

xdustinface:
- Drops the unused parameter 
- Makes it not longer be a decorator but just a normal function call. Im not sure why making it a decorator would be better here since it just makes hinting complex. Let me know if i miss the point of it. 
- Moves it alongside with `obtain_current_passphrase` into `chia.cmds.util` since this code is CLI stuff actually.

This is a first step to get rid of CLI code in keychain related parts, there are still places which need to be fixed.


paninaro (approved):
lgtm

paninaro:
> * Im not sure why making it a decorator would be better here since it just makes hinting complex

I dunno, I liked it :) :shrug:

PR #12852, opened 2022-08-09
Ms.plot sync flake

mariano54:
This fixes the flaky `plot_sync` tests which say that you cannot bind a port that is in use.
The fix is to use port 0 for the farmer instead of `find_available_port`.


mariano54:
This doesn't handle exceptions properly, harvester won't shut down.
Replaced by https://github.com/Chia-Network/chia-blockchain/pull/12862

PR #12862, opened 2022-08-09
server|tests: Delay the harvester -> farmer connection in some tests

xdustinface:
- Adds `Service.add_peer`
-  Should fix flakiness in the here adjusted tests by delayed adding of the farmer connection information to he harvester service via `Service.add_peer` and the use port `0` instead of `find_available_listen_port("farmer_server")` for the service creation.


mariano54 (approved):
Nice, I like this approach

PR #12875, opened 2022-08-10
Ms.wlt node unsynced

mariano54:
Fixes the issues with the wallet not syncing in untrusted mode, if connected to a trusted peer that is not synced.

Also some other important sycing fixes. This should be merged along with: https://github.com/Chia-Network/chia-blockchain/pull/12433

This needs to go into 1.5.1 as well


chia/wallet/util/wallet_sync_utils.py
```diff
@@ -324,27 +324,29 @@ async def request_header_blocks(
     peer: WSChiaConnection, start_height: uint32, end_height: uint32
 ) -> Optional[List[HeaderBlock]]:
     if Capability.BLOCK_HEADERS in peer.peer_capabilities:
+        log.warning(f"Peer {peer.get_peer_info()} has the capability! {peer.peer_capabilities}")
```
trepca, line 327: log.warning? is this just to test?

trepca (dismissed):
LGTM, just `log.warning` seems a bit excessive for what seems to be more debug like messages

chia/wallet/util/wallet_sync_utils.py
mariano54: Oh this was debugging, I'll revert

trepca (dismissed):

chia/wallet/util/wallet_sync_utils.py
wallentx: Resolving conversation, as this looks to be fixed.

mariano54:
Force pushed to get rid of unsigned commits

trepca (approved):

Quexington (approved):

PR #12876, opened 2022-08-10
Fix getting timestamp when local node is not synced

mariano54:


chia/wallet/wallet_node.py
```diff
@@ -929,16 +941,20 @@ async def get_timestamp_for_height(self, height: uint32) -> uint64:
...
+            self.log.debug(f"Fetching block at height: {height} from {peer.get_peer_info()}")
+            last_tx_block = await fetch_last_tx_from_peer(height, peer)
+            if last_tx_block is None:
+                continue
+
+            assert last_tx_block.foliage_transaction_block is not None
+            self.get_cache_for_peer(peer).add_to_blocks(last_tx_block)
+            return last_tx_block.foliage_transaction_block.timestamp
+
         if last_tx_block is None:
```
trepca: wouldn't `last_tx_block` always be None here?

mariano54: fixed

PR #12878, opened 2022-08-10
Xch spam

danieljperry:
Add new spam filter and tests, and make mergeable with main branch.
This is to replace PR #11894


chia/wallet/wallet_coin_store.py
```diff
@@ -139,6 +139,22 @@ async def get_unspent_coins_for_wallet(self, wallet_id: int) -> Set[WalletCoinRe
...
 
+    async def get_all_unspent_coins(self) -> Set[WalletCoinRecord]:
+        """Returns set of CoinRecords that have not been spent yet for a wallet."""
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            rows = await conn.execute_fetchall(
+                "SELECT * FROM coin_record WHERE spent_height=0"
+            )
+        return set(self.coin_record_from_row(row) for row in rows)
+
+    async def get_all_coins(self) -> Set[WalletCoinRecord]:
```
mariano54, line 150: why is this method added?

chia/wallet/wallet_state_manager.py
```diff
@@ -660,6 +660,42 @@ async def determine_coin_type(
 
         return None, None
 
+    async def filter_spam(self, new_coin_state: List[CoinState]) -> List[CoinState]:
+        all_unspent: Set[WalletCoinRecord] = await self.coin_store.get_all_unspent_coins()
```
mariano54, line 664: I think we should first get `xch_spam_amount` from config, and if it's zero, we can avoid performing the DB lookup, and just return. This means that it will be much faster for those people that disable it, and have huge non-dust wallets

chia/wallet/wallet_state_manager.py
```diff
@@ -660,6 +660,42 @@ async def determine_coin_type(
...
+            # Only apply filter to new coins being sent to our wallet, that are very small
+            if (
+                cs.created_height is not None
+                and cs.spent_height is None
+                and cs.coin.amount < xch_spam_amount
+                and (cs.coin.puzzle_hash in is_standard_wallet_phs or await self.is_standard_wallet_tx(cs))
+            ):
+                is_standard_wallet_phs.add(cs.coin.puzzle_hash)
+                if small_unspent_count < spam_filter_after_n_txs:
+                    self.log.info(f"Not filtered 1 {small_unspent_count}")
```
mariano54, line 683: remove log

chia/wallet/wallet_state_manager.py
```diff
@@ -660,6 +660,42 @@ async def determine_coin_type(
...
+                    filtered_cs.append(cs)
+                small_unspent_count += 1
+            else:
+                self.log.info(f"Not filtered 2 {small_unspent_count}")
```
mariano54, line 687: remove log

mariano54 (changes requested):
thanks for the changes!

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -230,6 +244,179 @@ async def test_num_unspent(self, bt, db_version):
...
+    # Part 1: Farm a block with the farm wallet.
+    #         Farm wallet has 2 coins, receive wallet has 0.
+    # Part 2: Send 1 mojo from farm wallet to receive wallet and farm another block.
+    #         Farm wallet has 6 coins (5 coinbase + 1 change), receive wallet has 1.
+    # Part 3: Send all money from farm wallet to receive wallet in two coins and farm another block.
+    #         Farm wallet has 2 coins, receive wallet has 3.
+    # Part 4: Send all money from receive wallet to farm wallet and farm another block.
+    #         Farm wallet has 5 coins (4 coinbase + 1 received), receive wallet has 0.
+    @pytest.mark.asyncio
+    async def test_get_all_unspent_coins(
```
mariano54, line 291: this test might be better if it didn't have wallets and nodes. Check  out the other tests in the file. We can usually just add coins manually to the coin store, and don't need all these services up and running, which takes time and decreases stability

chia/wallet/wallet_coin_store.py
danieljperry: The method used to exist. It was removed when the wallet coin store cache was removed. I'm using it in one of my tests. I could remove it, but the test wouldn't be as comprehensive.

mariano54: yeah we should remove it. If you want, you can make db queries manually from the test instead

danieljperry: This is complete in my latest commit.

chia/wallet/wallet_state_manager.py
danieljperry: Updated in the latest commit.

chia/util/initial-config.yaml
```diff
@@ -545,6 +540,11 @@ wallet:
   # Interval to resend unconfirmed transactions, even if previously accepted into Mempool
   tx_resend_timeout_secs: 1800
 
+  # After n received unspent transactions, the spam filter will be enabled, which will filter out received
+  # coins with very small value. Any standard TX under xch_spam_amount is filtered
+  spam_filter_after_n_txs: 200
```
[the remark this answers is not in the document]
danieljperry: Some have over 100k tx.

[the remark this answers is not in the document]
mariano54: Should we increase this @trepca ? I think increasing is fine since wallet perf is getting much better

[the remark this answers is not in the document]
trepca: I missed the fact it's only counting "change" coins, which makes sense. It's probably unlikely that even a power user would have that many change coins, although I guess it depends on the select coin strategy. 
What we want to avoid is making life harder for power users, that might create a lot of change coins to enable more parallel minting / operations. These might hit the limit of what we consider "spam". /cc @geoffwalmsley

mariano54:
Two more things @danieljperry.
1. Can you fix the merge conflicts?
2. How about adding a test where the user has > 200 dust coins, and then receives an NFT. I want to ensure that NFTs are never filtered (since they use 1 or few mojos).

chia/util/initial-config.yaml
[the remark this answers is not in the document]
mariano54: if they are doing something like that, they can go to the config and change it right? Or use higher value coins

mariano54 (dismissed):
LGTM pending CI and checking performance by @danieljperry

altendky (dismissed):
Needs to be reapplied to `release/1.6.0` such that it isn't bringing `main` in with it.  Happy to chat about this, but maybe not for about 9 hours or so.  Just being extra careful with requesting changes here.

mariano54 (approved):
LGTM. Dan's numbers:

1.5.1:
sync time: 10 minutes, 20 seconds
105,481 tx
1,603.37001457181 TXCH

1.5.1 + patch (default settings)
sync time: 10 seconds
4448 tx
1,603.369989050548 TXCH

emlowe (approved):

tests/core/full_node/stores/test_coin_store.py
wallentx: Added new issue https://github.com/Chia-Network/chia-blockchain/issues/13244

PR #12883, opened 2022-08-10
util: Implement key label support for `Keychain` and `KeychainServer`

xdustinface:
This PR implements key label support for the `Keychain` and `KeychainServer` classes.

Changes to the daemon websocket interface: 
- Adds the new command `set_label` to set a name for fingerprint
- Adds the new command `delete_label` to delete the name for fingerprint
- Adds the `label` parameter to the `add_private_key` method to assign a label when adding keys

Based on: 
- #12830 
- #12843 
-  #13054

### DIFF
--- a/chia/util/keychain.py
+++ b/chia/util/keychain.py
@@ -332,6 +332,7 @@ class Keychain:
         except Exception:
             if label is not None:
                 self.keyring_wrapper.delete_label(fingerprint)
+            raise
 
         return key
 
--- a/chia/util/keychain.py
+++ b/chia/util/keychain.py
@@ -192,6 +192,7 @@ class KeyDataSecrets(Streamable):
 class KeyData(Streamable):
     fingerprint: uint32
     public_key: G1Element
+    label: Optional[str]
     secrets: Optional[KeyDataSecrets]
 
     def __post_init__(self) -> None:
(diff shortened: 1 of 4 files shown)


chia/util/keychain.py
```diff
@@ -284,6 +286,7 @@ def _get_key_data(self, index: int, include_secrets: bool = True) -> KeyData:
         return KeyData(
             fingerprint=public_key.get_fingerprint(),
             public_key=public_key,
+            label=self.keyring_wrapper.get_label(public_key.get_fingerprint()),
```
paninaro, line 289: nit: assign fingerprint to local variable to avoid extra call to `get_fingerprint()`?

chia/util/keychain.py
```diff
@@ -315,13 +318,36 @@ def add_private_key(self, mnemonic: str) -> PrivateKey:
...
+        if label is not None:
+            self.keyring_wrapper.set_label(fingerprint, label)
+
+        try:
+            self.keyring_wrapper.set_passphrase(
+                self.service,
+                get_private_key_user(self.user, index),
+                bytes(key.get_g1()).hex() + entropy.hex(),
+            )
+        except Exception:
```
paninaro, line 332: We should re-raise the exception so that the user knows that their key wasn't saved properly.

chia/util/keychain.py
```diff
@@ -315,13 +318,36 @@ def add_private_key(self, mnemonic: str) -> PrivateKey:
             # Prevents duplicate add
             raise KeychainFingerprintExists(fingerprint)
 
-        self.keyring_wrapper.set_passphrase(
-            self.service,
-            get_private_key_user(self.user, index),
-            bytes(key.get_g1()).hex() + entropy.hex(),
-        )
+        # Try to set the label first, it may fail if the label is invalid or already exists.
+        # This can probably just be moved into `FileKeyring.set_passphrase` after the legacy keyring stuff was dropped.
```
paninaro, line 322: I agree that we should make this into a single operation of setting the passphrase w/ optional label. If the legacy keyring API is the limiting factor here, I'm ok waiting until we remove that legacy support. But if there's a way to combine the two operations into one now, I think that would be worth pursuing.

chia/util/keychain.py
```diff
@@ -315,13 +318,36 @@ def add_private_key(self, mnemonic: str) -> PrivateKey:
...
+        Assigns the given label to the first key with the given fingerprint.
+        """
+        self.get_key(fingerprint)  # raise if the fingerprint doesn't exist
+        self.keyring_wrapper.set_label(fingerprint, label)
+
+    def delete_label(self, fingerprint: int) -> None:
+        """
+        Removes the label assigned to the key with the given fingerprint.
+        """
+        self.keyring_wrapper.delete_label(fingerprint)
```
paninaro, line 349: It seems a little asymmetrical that setting a label has a `get_key` check, but deleting a label doesn't. I'd probably add the `get_key` check here to avoid possibly confusing the user.

paninaro (changes requested):
looks good with just a few minor suggestions

chia/util/keychain.py
xdustinface: `FileKeyring.delete_label` already raises `KeychainFingerprintNotFound` if the label doesn't exist.

chia/util/keychain.py
xdustinface: See 2fdbf835db7c6b7345de505627b28c9eaf693860

chia/util/keychain.py
xdustinface: Good point, see 77a7d898987cb905959990473360376602a7ef16

chia/util/keychain.py
xdustinface: There is a hacky way maybe like passing the fingerprint down to `FileKeyring` if it doesn't somewhere conflict with the legacy keyring APIs but i would rather not do/try this now since we anyway drop the legacy code soon and it works fine this way for now.

paninaro (approved):
lgtm

PR #12889, opened 2022-08-10
Ms.peer required + sync speedup

mariano54:
These three methods in `wallet_node`: `fetch_puzzle_solution`, `get_coin_state`, and `fetch_children`, are all called during sync, by the different wallets (and WSM). If any of them fail, the whole sync will stop. This is a problem, especially if we are connected to a peer that is not synced to the chain (for example our local node). 

This fixes it, by forcing the sync target peer to be passed in. This peer must be synced up to the chain, so they will respond with the correct data. If something goes wrong, sync stops, and it can resume again with another peer.

This seems like a big change but it's not. It's just passing in an additional peer argument everywhere.

Also changes a constant for how many concurrent requests to make, (reverts it from 20 to 10), to bring performance back up to par with 1.5.0


chia/wallet/util/wallet_sync_utils.py
```diff
@@ -68,7 +68,7 @@ async def subscribe_to_phs(
     all_coins_state: Optional[RespondToPhUpdates] = await peer.register_interest_in_puzzle_hash(msg, timeout=300)
     if all_coins_state is None:
         raise ValueError(f"None response from peer {peer.peer_host} for register_interest_in_puzzle_hash")
-    return all_coins_state.coin_states
+    return all_coins_state.coin_states[:30]
```
mariano54, line 71: this is debugging, regression

Quexington (approved):

mariano54:
This one should be merged to main after `ms.wlt_node_unsynced` @wallentx

paninaro (approved):
lgtm

PR #12890, opened 2022-08-10
Combine m1 and intel mac installers to a single workflow

cmmarslender:
Combines m1 and intel mac installers to a single matrix workflow, since we have native m1 runners now and the jobs are essentially identical. Testing on both arch mac, and the installers worked and started syncing.


build_scripts/npm_macos/package.json
```diff
@@ -10,6 +10,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "appdmg": "^0.6.4",
```
paninaro, line 13: Is this being used?

cmmarslender: Not entirely sure, but it was in one of the previous package.json files for mac, and was in both lock files.

paninaro: Ahh ok, it's a dependency of electron-installer-dmg

paninaro (approved):
lgtm

build_scripts/npm_macos/package.json
cmmarslender: Looks to be used by electron-installer-dmg, so it probably doesn't need to be specifically called out like it is, but used either way

```
├─┬ electron-installer-dmg@3.0.0
│ ├── appdmg@0.6.4 deduped
```

hoffmang9:
For old times sake - lgtm

PR #12900, opened 2022-08-11
Occasional sleep to allow wallet to respond to pings

mariano54:
I was getting a close signal from the keychain while validating weight proof:
I think this is because `Future.result()` is a blocking call.

``` 
2022-08-11T12:19:18.856 wallet chia.full_node.weight_proof: INFO     validating 6206 sub epochs
2022-08-11T12:19:18.857 wallet chia.full_node.weight_proof: INFO     validate weight proof peak height 2383374
2022-08-11T12:19:20.505 wallet wallet                     : INFO     Reconnecting to peer {'host': '127.0.0.1', 'port': 8444}
2022-08-11T12:19:20.511 wallet wallet_server              : INFO     Cannot connect to host 127.0.0.1:8444 ssl:<ssl.SSLContext object at 0x7f78f37923c0> [Connect call failed ('127.0.0.1', 8444)]
2022-08-11T12:19:52.065 wallet chia.wallet.wallet_node    : INFO     It took 33.27067971229553 time to validate the weight proof
2022-08-11T12:19:52.068 wallet wallet                     : INFO     Reconnecting to peer {'host': '127.0.0.1', 'port': 8444}
2022-08-11T12:19:52.071 wallet chia.wallet.wallet_node    : INFO     Close signal received from keychain, we probably timed out.
2022-08-11T12:19:52.072 wallet chia.wallet.wallet_node    : INFO     Reconnecting to keychain at wss://localhost:55400.
```

PR #12910, opened 2022-08-11
return to use awaitable in WP

almogdepaz:
alternative to https://github.com/Chia-Network/chia-blockchain/pull/12900


mariano54 (approved):

mariano54:
Yeah this seems better than my solution

PR #12911, opened 2022-08-11
cmds|daemon: Improve legacy keyring migration enforcement

xdustinface:
This helps to avoid the repeated unlock keyring prompts and slightly adjusts two comments. 

The fix for the unlock prompts is changing the `migrate_keys` function to use the daemon via `KeychainProxy` if it's running instead of creating a separate `Keychain` instance which needs to be unlocked every time. 

----

Note that 1f23b3d5e166a0c53bcbc4129db6992219303005 is also done in #12850 but is needed here because:
- There is imo a bug in `KeyringWrapper` constructor where we assign the legacy keyring to `KeyringWrapper.keyring` if `force_legacy=True` https://github.com/Chia-Network/chia-blockchain/blob/535d21e6796ab67895d2658ff5337781fab04048/chia/util/keyring_wrapper.py#L118
- Even if the above bug is fixed, the decorator will still try to unlock the file keyring because we use the shared instance in there https://github.com/Chia-Network/chia-blockchain/blob/535d21e6796ab67895d2658ff5337781fab04048/chia/util/keychain.py#L97

### DIFF
--- a/chia/cmds/keys.py
+++ b/chia/cmds/keys.py
@@ -1,4 +1,7 @@
+import asyncio
+
 import click
+import sys
 
 from typing import Optional, Tuple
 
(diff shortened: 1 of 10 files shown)


paninaro (approved):
lgtm. I did some testing with this PR and found that there is a noticeable delay when running `chia keys show` when the daemon is already running. I think this is reasonable in the short term as we only need to live with the migration checks for another release or two. Once we can eliminate the legacy keyring / migration code, things should be much snappier.

PR #12916, opened 2022-08-11
Wallentx/cherrypick

wallentx:
plz review carefully


xdustinface (changes requested):
[603f996](https://github.com/Chia-Network/chia-blockchain/pull/12916/commits/603f996c9bd7341d996d388ffa776558dbf3e0ce) is not correct, it just adds `wallet_weight_proof_handler.validate_weight_proof_inner` but instead should apply the changed that were applied to that code in the original PR to the related code in the release branch (there was some code moved around and merged in #12310).

However, #12910 is trying to fix an issue introduced in https://github.com/Chia-Network/chia-blockchain/pull/12310 which isn't part of the release branch so we should probably just drop 603f996.

The others look good!

xdustinface (approved):
Maybe we should still pick https://github.com/Chia-Network/chia-blockchain/pull/12310 and https://github.com/Chia-Network/chia-blockchain/pull/12910 together in another PR if we decide to pick both since #12310 has 

> avoid double validation of wp in wallet long sync

in its description . @almogdepaz has this a big impact?

almogdepaz:
> Maybe we should still pick #12310 and #12910 together in another PR if we decide to pick both since #12310 has
> 
> > avoid double validation of wp in wallet long sync
> 
> in its description . @almogdepaz has this a big impact?

not really a big impact mostly just better code and less prone to bugs

PR #12919, opened 2022-08-12
Bump chia_rs dependency

arvidn:
This bumps `chia_rs` to the most recent verision (0.1.7).

The main updates are:

* the complex logic to handle redundant leading zeroes in conditions has been removed. Now that we've soft-forked away allowing it. This also removes one of the flags to `run_as_generator()` controlling whether leading zeros are allowed or not. They are always disallowed now. https://github.com/Chia-Network/chia_rs/pull/43
* `parse_rust()` on rust types implementing the Streamable protocol now takes a buffer (rather than `bytes`). This avoids copying the whole input buffer every time `parse_rust()` is invoked. This turned out to be expensive during wallet sync. https://github.com/Chia-Network/chia_rs/pull/32
* A new function `tree_hash()` is exposed from the rust module to compute the tree-hash on a serialized program directly, avoiding parsing it into rust first. This speeds up wallet sync as we have to validate the tree hash of each incoming coin. https://github.com/Chia-Network/clvm_rs/pull/200 https://github.com/Chia-Network/chia_rs/pull/49

### DIFF
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -1,6 +1,6 @@
 import logging
 from typing import Dict, Optional
-from chia_rs import MEMPOOL_MODE, COND_CANON_INTS, NO_NEG_DIV
+from chia_rs import MEMPOOL_MODE, NO_NEG_DIV
 
 from chia.consensus.cost_calculator import NPCResult
 from chia.types.spend_bundle_conditions import SpendBundleConditions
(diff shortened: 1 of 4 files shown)


chia/full_node/mempool_check_conditions.py
```diff
@@ -42,7 +41,7 @@ def get_name_puzzle_conditions(
         # conditions must use integers in canonical encoding (i.e. no redundant
```
trepca, line 41: we should remove this comment too

arvidn: actually, this comment is still valid. The flag that was removed (`COND_CANON_INTS`) is now implied. With the soft-fork, integers are *required* to be in canoical form (i.e. no redundant leading zeroes).

Removing this flag, and making its behavior mandatory, was one of the simplifications in `chia_rs` I made recently.

trepca (approved):

PR #12923, opened 2022-08-12
rust tree_hash()

arvidn:
It's faster to serialize `Program` and run the rust `tree_hash()` function on it, than to run the python `_tree_hash()` implementation.

Also, the second parameter to `get_tree_hash()` is used in three places. Not passing in any precomputed hashes is by far the most common case, so I split the function in two. One with and one without the pre-calculated hashes.

The three calls in these functions were changed to use the new `get_tree_hash_precalc()`.
```
chia/wallet/did_wallet/did_wallet_puzzles.py
chia/wallet/puzzles/singleton_top_layer.py
chia/wallet/trading/offer.py
```

This speeds up creation of puzzle hashes in the wallet.

| CPU usage | before | after | after / before |
| --- | --- | --- | --- |
| Program.tree_hash() | 26.34% | 15.83% | 60% |

before:
chia-hotspot-0-1-tree-hash-baseline

after:

chia-hotspot-0-1-patched

### DIFF
--- a/chia/types/blockchain_format/program.py
+++ b/chia/types/blockchain_format/program.py
@@ -95,12 +95,15 @@ class Program(SExp):
         """
         return _sexp_replace(self, self.to, **kwargs)
 
-    def get_tree_hash(self, *args: bytes32) -> bytes32:
+    def get_tree_hash_precalc(self, *args: bytes32) -> bytes32:
         """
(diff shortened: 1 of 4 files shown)


trepca (approved):
LGTM 🚤

PR #12924, opened 2022-08-12
revert change to change OR REPLACE -> OR FAIL in wallet_coin_store

arvidn:
There's not enough test coverage to make this change with confidence, and it's not necessary.

I made this change as part of another PR where (as far as I can tell) we no longer rely on replacing existing entries. I made this change to ensure we aren't (in the tests).

But I'm having  second thoughts about this change. The test coverage is probably not good enough to tell us whether we still need this or not.

This was my original change, that this PR reverts: https://github.com/Chia-Network/chia-blockchain/pull/12391

### DIFF
--- a/chia/wallet/wallet_coin_store.py
+++ b/chia/wallet/wallet_coin_store.py
@@ -73,7 +73,7 @@ class WalletCoinStore:
         assert record.spent == (record.spent_block_height != 0)
         async with self.db_wrapper.writer_maybe_transaction() as conn:
             await conn.execute_insert(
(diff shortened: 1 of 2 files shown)


mariano54:
Is the FAIL also included in the 1 5 1 release? If so, we might want to revert there too

wallentx:
> Is the FAIL also included in the 1 5 1 release? If so, we might want to revert there too

https://github.com/Chia-Network/chia-blockchain/blob/release/1.5.1/chia/wallet/wallet_coin_store.py#L76 It is there

trepca (approved):

PR #12925, opened 2022-08-12
fix flakey test

almogdepaz:
better isolation for testing ses mismatch on sync, this should now run consistently


mariano54 (approved):

mariano54:
Looks like it's still broken, or maybe this is something else:
https://github.com/Chia-Network/chia-blockchain/runs/8165529186?check_suite_focus=true

mariano54:
^ @paninaro is this the same TSL cert issue as before?

mariano54:
Or maybe we need to `await async_mock` @almogdepaz ? There is an error about that in the log.f

almogdepaz:
> Or maybe we need to `await async_mock` @almogdepaz ? There is an error about that in the log.f

@mariano54 it looks like the event loop was closed mid sync and that failed the test, im not sure its related specifically to this test, looking into it

PR #12926, opened 2022-08-12
Reduce flakiness (saw 281 in a test)

mariano54:


jack60612 (dismissed):
Still within the margin of error

arvidn (dismissed):
in an ideal world, we would use deterministic test cases. with a fixed random number generator seed if we have to.

jack60612:
> in an ideal world, we would use deterministic test cases. with a fixed random number generator seed if we have to.

agreed

mariano54:
The test is fine, the issue is that the function being tested has a call to random (to try different combinations of coins). we can probably seed the rng so it becomes deterministic

mariano54:
Test is deterministic now @arvidn

jack60612:
> Test is deterministic now @arvidn

yay i guess?

jack60612 (dismissed):
lgtm

chia/wallet/coin_selection.py
```diff
@@ -135,10 +135,11 @@ def select_smallest_coin_over_target(target: uint128, sorted_coin_list: List[Coi
 # we use this to find the set of coins which have total value closest to the target, but at least the target.
 # IMPORTANT: The coins have to be sorted in descending order or else this function will not work.
 def knapsack_coin_algorithm(
-    smaller_coins: List[Coin], target: uint128, max_coin_amount: int, max_num_coins: int
+    smaller_coins: List[Coin], target: uint128, max_coin_amount: int, max_num_coins: int, seed: bytes = b"knapsack seed"
 ) -> Optional[Set[Coin]]:
     best_set_sum = max_coin_amount
     best_set_of_coins: Optional[Set[Coin]] = None
+    random.seed(seed)
```
arvidn, line 142: it's a bit of a nitpick, but I think it would be cleaner to create a new `Random` object with this seed, and then ask it for random numbers. With this change, if any other part of the program depends on pseudo random numbers, they won't be so random anymore after calling this function (because you're re-seeding the global state)

arvidn: actually, I think it's quite risky to seed the global PRNG.

jack60612: > actually, I think it's quite risky to seed the global PRNG.

agreed

mariano54: yeah I didn't realize it was global. whoops, that was a huge oversight from my part. I'll do what you suggested

arvidn (approved):

jack60612 (approved):
Nice

emlowe (approved):

wallentx (approved):

PR #12934, opened 2022-08-12
Rust coin state

arvidn:
The motivation for this patch is to speed up `subscribe_to_phs()` in the wallet. For wallet with a lot of transactions, and many puzzle hashes, parsing the response for this RPC is very slow.

| --- | --- | --- | --- |
| subscribe_to_phs() | 53.29% | 0.88% | 0.0165 |

Profile from main:
chia-hotspot-3-main

Profile with this patch:
chia-hotspot-3-patched

### DIFF
--- a/chia/protocols/wallet_protocol.py
+++ b/chia/protocols/wallet_protocol.py
@@ -1,6 +1,8 @@
 from dataclasses import dataclass
 from typing import List, Optional, Tuple
 
+from chia_rs import CoinState, RespondToPhUpdates
+
 from chia.types.blockchain_format.coin import Coin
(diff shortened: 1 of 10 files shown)


chia/protocols/wallet_protocol.py
```diff
@@ -178,29 +183,13 @@ class RespondHeaderBlocks(Streamable):
     header_blocks: List[HeaderBlock]
 
 
-@streamable
```
mariano54, line 181: I would leave this code here but commented. The protocol file acts as a sort of documentation of how wallet and node communicate.

trepca (approved):
🚀

PR #12935, opened 2022-08-12
Change DL offers to prove a branch instead of a leaf

Quexington:
Previously the DL offer chialisp assumed that you were proving a leaf in a merkle tree.  Since we instead want to prove a key/value pair where the keys and values are leaves, I've changed it to instead require a branch to prove.

The hashing algorithm in the chialisp also changed to work like sha256tree.


emlowe:
Just to comment that we matched the chialisp hashing to the DL and this was also changed for custody as well. So everything right now is all consistent with this PR

geoffwalmsley (dismissed):

emlowe (approved):

PR #12946, opened 2022-08-12
Add single_derived_address config

ytx1991:
This will enable the single derivation address feature if user set single_derived_address to true.
All coins will be managed in one address.


ytx1991:
I'm planning to test 3 cases. The baseline is the current solution, I will set the user experience score as 100.
Current experience: Will become syncing status every 3-5 mins, each time will take 1-2 mins. Making transactions will take a very long time and has a high chance of timeout (600s). Switching between tokens/NFT/Offer will also take minutes.

1. Just disabled creating new derived address after every transaction.
Score 120-150. Switching between tokens/NFT/Offer just take seconds. The other issue still exist.

2. Based on #1, move all coins in one address.
Testing

3. Based on #2, create a new wallet without legacy transactions and derivation paths.
4. Testing

mariano54:
> I'm planning to test 3 cases. The baseline is the current solution, I will set the user experience score as 100. Current experience: Will become syncing status every 3-5 mins, each time will take 1-2 mins. Making transactions will take a very long time and has a high chance of timeout (600s). Switching between tokens/NFT/Offer will also take minutes.
> 
> 1. Just disabled creating new derived address after every transaction.
>    Score 120-150. Switching between tokens/NFT/Offer just take seconds. The other issue still exist.
> 2. Based on above, move all coins in one address.
>    Testing
> 3. Based on above, create a new wallet without legacy transactions and derivation paths.
> 4. Testing

Creating one address should not take longer than 5-10 milliseconds. If it's taking longer than that, it's a bug we need to investigate.

When you get a new wallet, you do need to generate the whole list and that can take a while

wallentx:
> > I'm planning to test 3 cases. The baseline is the current solution, I will set the user experience score as 100. Current experience: Will become syncing status every 3-5 mins, each time will take 1-2 mins. Making transactions will take a very long time and has a high chance of timeout (600s). Switching between tokens/NFT/Offer will also take minutes.
> > 
> > 1. Just disabled creating new derived address after every transaction.
> >    Score 120-150. Switching between tokens/NFT/Offer just take seconds. The other issue still exist.
> > 2. Based on above, move all coins in one address.
> >    Testing
> > 3. Based on above, create a new wallet without legacy transactions and derivation paths.
> > 4. Testing
> 
> Creating one address should not take longer than 5-10 milliseconds. If it's taking longer than that, it's a bug we need to investigate.
> 
> When you get a new wallet, you do need to generate the whole list and that can take a while
> 
> Edit: what do you mean it takes minutes to switch between cat/NFT/offers. You mean clicking on the GUI? switching keys? Minting a new CAT? Either way, it should not take minutes

Please note that the timing of any `chia wallet` commands would have been inflated prior to this PR https://github.com/Chia-Network/chia-blockchain/pull/12950

The timing still impacts `chia keys` and `chia start`

mariano54:
Some things are not clear for me yet. I think the following should be done before this goes in:

1. Clarify what happens with this setting. Does the wallet still generate 100 addresses but only uses the first one? Is only one address ever generated?
2. Write tests to ensure one address is generated, and the same address is always returned.
3. Write tests to call `get_new_address` rpc and either fail, or return the same address when the new flag is passed in. (I think failure is better)
4. Identify why this setting exists and where the performance issues actually are. If we can fix the performance issues, we might not need this setting.
5. Clarify what happens if I already have 500 addresses in my wallet (where the last one used is the 400th). Will the wallet start returning the 400th address forever? Will it start returning the 1st address forever? Will new wallets use 1 address or 500? How does this affect privacy?
6. Write a test for point number 5
7. Will the main address be observer or non-observer? Will both be added to the DB or only 1 of them? A test would be good as well

trepca:
@ytx1991 this should be a draft PR, merging into `main`. The intent should be to mainly share what changes were made to gather data on how derived addresses affect performance for further discussion.

PR #12953, opened 2022-08-13
Replace `electron-packager` with `electron-builder`

ChiaMineJP:
# Prerequisite of merging this PR into the `main`
1. Review and merge [the GUI PR](https://github.com/Chia-Network/chia-blockchain-gui/pull/903)
2. Update the submodule ref to the GUI which takes in the above PR.
3. Merge this PR.

\* Until "2." is done, CI jobs for building installer in this PR would all fail

# ~Known issue~
## ~`electron-builder` does not work for `.deb` installer for arm64~
~The root cause: https://github.com/electron-userland/electron-builder/issues/3901.~
~As of 2022-08-14, it has not been resolved yet, so I decided to leave `electron-packager` for debian arm64 installer.~
This issue was resolved by locally installing `fpm`.

# What we need to do next
- Check RPM installer updates.  
  `build_linux_rpm-2-installer.sh` had code which manually update rpm's spec file for some issues.
  See https://github.com/Chia-Network/chia-blockchain/blob/main/build_scripts/build_linux_rpm-2-installer.sh#L109-L121.
  I moved the rpmbuild's `%define` macros into https://github.com/Chia-Network/chia-blockchain-gui/pull/903/files#diff-5b41a1b378a6cd7efdab84c282078232ef16a09d6400b40ea0e9ba5e1eed917fR85-R91 , and this needs to be reviewed whether this works as expected.
- Confirm whether all installer work on target OSes.
  - Case 1:  Fresh install (No chia-blockchain application is installed on target OSes.)
  - Case 2:  Update (Previous version of the app is already installed)  
\*Please note that you can download the installers from CI jobs of this PR (https://github.com/Chia-Network/chia-blockchain/pull/12954). The PR is almost the same with this PR but only differs the submodule ref to the GUI repository. (pointing to the commit which is compatible with this PR)
- Confirm whether GUI runs as expected after installing.


paninaro:
Thanks @ChiaMineJP, this is great.

Regarding the ARM64 build, I found this... maybe you could give it a try?

https://github.com/jordansissel/fpm/issues/1801#issuecomment-919877499
and:
https://www.beekeeperstudio.io/blog/electron-apps-for-arm-and-raspberry-pi

paninaro:
I've checked the macOS PoC build and it seems to be working as expected.

@wallentx or @cmmarslender do either you have the historical context for why the RPM spec file is modified by the build script?

ChiaMineJP:
> Thanks @ChiaMineJP, this is great.
> 
> Regarding the ARM64 build, I found this... maybe you could give it a try?
> 
> [jordansissel/fpm#1801 (comment)](https://github.com/jordansissel/fpm/issues/1801#issuecomment-919877499) and: https://www.beekeeperstudio.io/blog/electron-apps-for-arm-and-raspberry-pi

I'll try this solution.

paninaro (dismissed):
lgtm. Tested macOS and Windows installers.

cmmarslender (changes requested):
Working on testing the linux installers now. So far, they seem to work fine, but a few more OS to test. I think the deb package name when installed changed from `chia-blockchain` to just `chia` so I ended up with two different chia installations when testing on ubuntu from the UI package. Can we keep the package name as `chia-blockchain`?

paninaro:
> Working on testing the linux installers now. So far, they seem to work fine, but a few more OS to test. I think the deb package name when installed changed from `chia-blockchain` to just `chia` so I ended up with two different chia installations when testing on ubuntu from the UI package. Can we keep the package name as `chia-blockchain`?

Yeah, the names aren't quite right. I'll take a look today.

cmmarslender:
Tested the RPM on centos7 and fedora, and ran into no issues. The RPM spec changes were for centos7 if I remember right, so seems like those changes are retained correctly.

paninaro:
Windows and Mac installers look good with the latest product naming fixes. @cmmarslender can you take a look when you have a chance?

wallentx (changes requested):

build_scripts/assets/rpm/postinst.sh
```diff
@@ -3,5 +3,5 @@
 
 set -e
 
-ln -s /usr/lib/chia-blockchain/resources/app.asar.unpacked/daemon/chia /usr/bin/chia || true
-ln -s /usr/lib/chia-blockchain/resources/app.asar.unpacked/daemon /opt/chia || true
+ln -s /opt/chia/resources/app.asar.unpacked/daemon/chia /usr/bin/chia || true
+ln -s /opt/chia/resources/app.asar.unpacked/daemon /opt/chia || true
```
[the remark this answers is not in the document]
cmmarslender: Yeah agree with Will, since there is already a directory here now. The more important one is the link to `/usr/bin/chia` - the link for /opt/chia was to make it easier to find chia in both the UI and CLI only installers

ChiaMineJP:
OK, I checked how GUI installer of `chia-blockchain`@1.5.0 works for Ubuntu 22.04.

### The installer extract all files under `/usr/lib/chia-blockchain/`
image

### It also creates a symbolic link which makes it look like CLI binaries are saved under `/opt/chia/`  
(This is done by `build_scripts/assets/deb/postinst.sh`)
image
---
### 1.5.0 CLI only installer for .deb

image
---

On the other hand, `electron-builder` just extracts all files under `/opt/chia/` instead of `/usr/lib/chia-blockchain/`.
On and before 1.5.0, `/opt/chia/` is the folder which stores "CLI" binaries but it is going to be changed so that `/opt/chia/` stores electron GUI files.

Should we keep install location compatibility between 1.5.0 and 1.6.0?

wallentx:
> Should we keep install location compatibility between 1.5.0 and 1.6.0?

Why not put the CLI binary somewhere within PATH, such as `/use/bin/`?
See what the current up to date AUR package maintainer has done here: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=chia-bin

build_scripts/assets/rpm/postinst.sh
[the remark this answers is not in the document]
ChiaMineJP: I've removed the line.

build_scripts/assets/deb/postinst.sh
```diff
@@ -3,5 +3,5 @@
 
 set -e
 
-ln -s /usr/lib/chia-blockchain/resources/app.asar.unpacked/daemon/chia /usr/bin/chia || true
-ln -s /usr/lib/chia-blockchain/resources/app.asar.unpacked/daemon /opt/chia || true
+ln -s /opt/chia/resources/app.asar.unpacked/daemon/chia /usr/bin/chia || true
+ln -s /opt/chia/resources/app.asar.unpacked/daemon /opt/chia || true
```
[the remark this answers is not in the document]
ChiaMineJP: Thanks, I removed the line.

ChiaMineJP:
> > Windows and Mac installers look good with the latest product naming fixes. @cmmarslender can you take a look when you have a chance?
> 
> Still seeing just `chia` as the installed package name, at least for the .deb

@cmmarslender 
By adding `--config.linux.desktop.Name="Chia Blockchain"`, app name displayed in desktop menu has been changed to "Chia Blockchain".
Could you try the latest installer?

@wallentx 
I checked all GUI/CLI `.deb`, `.rpm` installers and confirmed that they created `/usr/bin/chia` which can be accessed under `PATH` env var.
Did you mention Windows? Currently Windows installer does not add the location of `chia.exe` to `PATH`

ChiaMineJP:
> Maybe this is the `packageName` option? https://www.electron.build/configuration/linux#linuxtargetspecificoptions-apk-freebsd-pacman-p5p-and-rpm-options

Thanks for investigating! I've just added `config.xxx.packageName="chia-blockchain"` to the command.
Let's see how it goes.

build_scripts/build_linux_deb-2-installer.sh
```diff
@@ -62,49 +60,64 @@ dpkg-deb --build --root-owner-group "dist/$CLI_DEB_BASE"
 cp -r dist/daemon ../chia-blockchain-gui/packages/gui
 
 # Change to the gui package
-cd ../chia-blockchain-gui/packages/gui || exit
+cd ../chia-blockchain-gui/packages/gui
```
cmmarslender, line 63: Why are you removing the `|| exit`? This ensures that if the directory doesn't exist that we expect to exist, the script exits (I think usually i see `exit 1`, but either way, if we can't change to an expected directory, we should not continue)

build_scripts/build_linux_rpm-2-installer.sh
```diff
@@ -27,11 +24,10 @@ fi
 echo "Chia Installer Version is: $CHIA_INSTALLER_VERSION"
 
 echo "Installing npm and electron packagers"
-cd npm_linux_rpm || exit
+cd npm_linux
```
cmmarslender, line 27: should retain all `|| exit` on any `cd` commands, see comment above

build_scripts/build_linux_rpm-2-installer.sh
```diff
@@ -27,11 +24,10 @@ fi
...
 npm ci
-GLOBAL_NPM_ROOT=$(pwd)/node_modules
 PATH=$(npm bin):$PATH
-cd .. || exit
+cd ..
```
cmmarslender, line 30: should retain all `|| exit` on any `cd` commands, see comment above

build_scripts/build_linux_rpm-2-installer.sh
```diff
@@ -74,64 +70,45 @@ fpm -s dir -t rpm \
 cp -r dist/daemon ../chia-blockchain-gui/packages/gui
 
 # Change to the gui package
-cd ../chia-blockchain-gui/packages/gui || exit
+cd ../chia-blockchain-gui/packages/gui
```
cmmarslender, line 73: should retain all `|| exit` on any `cd` commands, see comment above

build_scripts/build_linux_rpm-2-installer.sh
```diff
@@ -74,64 +70,45 @@ fpm -s dir -t rpm \
...
-  --config rpm-options.json
-  LAST_EXIT_CODE=$?
-  if [ "$LAST_EXIT_CODE" -ne 0 ]; then
-	  echo >&2 "electron-installer-redhat failed!"
-	  exit $LAST_EXIT_CODE
-  fi
-fi
+GUI_RPM_NAME="chia-blockchain-${CHIA_INSTALLER_VERSION}-1.${REDHAT_PLATFORM}.rpm"
+mv "dist/${PRODUCT_NAME}-${CHIA_INSTALLER_VERSION}.rpm" "../../../build_scripts/dist/${GUI_RPM_NAME}"
+cd ../../../build_scripts
```
cmmarslender, line 104: should retain all `|| exit` on any `cd` commands, see comment above

cmmarslender (changes requested):
deb arm, deb intel, rpm intel, mac intel, and mac arm all tested and look good to me now. A couple other changes requested here for the scripts. i'd suggest running shellcheck on your scripts to catch any similar issues to these.

build_scripts/build_linux_deb-2-installer.sh
ChiaMineJP: I thought `|| exit` suppress error if directory does not exist.
If directory doesn't exist, then `cd` should fail and CI action should also fails.
I'll check more.

cmmarslender: `|| true` will suppress errors. Exit ensures the script exits, though it should probably be `exit 1` if we want the actual job to fail as well, not just exit

ChiaMineJP: OK. I've updated the code using `|| exit 1`.

build_scripts/build_linux_rpm-2-installer.sh
ChiaMineJP: I picked `|| exit1`. Please check the latest code.

build_scripts/build_linux_rpm-2-installer.sh
ChiaMineJP: I picked `|| exit1`. Please check the latest code.

build_scripts/build_linux_rpm-2-installer.sh
ChiaMineJP: Updated.

build_scripts/installer-version.py
```diff
@@ -21,7 +21,7 @@ def main():
         dev_release_number = ""
     elif len(version) == 4:
         smc_patch_version = version[2]
-        dev_release_number = "." + version[3]
+        dev_release_number = "-" + version[3]
```
cmmarslender, line 24: Did the old versioning scheme not work anymore with the new builders or what was the reason for this change?

cmmarslender (approved):

paninaro (approved):
Approved!

wallentx (approved):
LGTM

PR #12966, opened 2022-08-14
default `lock` parameter of `DataStore.shift_root_generations()` to True

altendky:


altendky:
I think this is existing breakage on `long_lived/atari` so we can revisit after that is fixed.

fchirica (approved):

PR #12978, opened 2022-08-15
Integrating NFT store directly into NFT wallet

trepca:
Removes `my_nft_coins` list that was caching NFTs in NFT store. To find an NFT we iterated over `my_nft_coins` list, now using a sqlite select which lowers time complexity especially for whale wallets. 
Lot of search/replace as methods now require to be awaited.


tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -737,7 +737,18 @@ def only_ids(trades):
...
+                "7usy4u48tmafcajc4dc0dmqa4h9z5f27e3qnuzf37yr78sl6kslts9aua5zfdg3r7knncj78pzg4nvrn0a6dkjvmme7"
+                "jjzz72xmlruuhmawm0eedl7fpfjkhnf70al2tw34pdgqje0m8wt6v8uaxw8gtjlkfzlw4447fk429f42tmn9x6l4qm9u"
+                "2n404j74ls5yv2grt0tzstm2l8hukgx4v6h42908px8dh0avzhdlxw7ruj5t53etc9dt2n2wm7098ks89waeunnfexdn"
+                "dmhf5dmhyjs6wvjzvvlj0scdh3np6mgmur6m2jj2y474cwuaurph0tq28ee6y3hxahhkkfqzlc8g7hm3lllvrl2nhlm7"
+                "2hgau9lgdumy9m99hy78dv5uwdr69jfkvu6a5qc0jlzkas6cry3zh7hasdwg785nmhhsl680m4fxdseavzdk8mg93dank"
+                "88ue2hned4tarn0al7gl4wq6ct4gd3c3q5a6l2gjlvd8ftteddfxxq5v4zdu0ycv2vuwslf7rz2u56nl7guqatk0ut7cy"
+                "ga0zu096k7rhdl99kc5jmscmtdz9vme2mmg86dwq7nk088spawraxfgftl0lqkycapflf725mjht2law69wh0rq8l7ue"
+                "gztx0xnvgc8y7wvvuwv3th5pcwckkm07jacznlgeuu8kcw0yuu4utjrm2mut8ekm8rmzp6vlzcm6e4f8xzytjx3ytnye"
+                "kany0a9l4tq0zxnh3rjwhve88658nd0xwhmgectl33u3us6klkk5c7vjyuurr6yetk7ua654my4cmxmtrjazfu3ara9"
+                "yc449jqxg4mfgx0sw3p9"
```
arvidn, line 751: generally I think it's better to separate pure formatting (and minor cleanup) into separate PRs, to limit the noise on the important ones

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -874,21 +884,21 @@ async def test_nft_endpoints(wallet_rpc_environment: WalletRpcTestEnvironment):
     await time_out_assert(15, wallet_is_synced, True, wallet_1_node, full_node_api)
     nft_wallet: NFTWallet = wallet_1_node.wallet_state_manager.wallets[nft_wallet_id]
 
-    def have_nfts():
-        return len(nft_wallet.get_current_nfts()) > 0
+    async def have_nfts():
+        return len(await nft_wallet.get_current_nfts()) > 0
```
arvidn, line 888: ideally the wallet would have a function to count the number of NFTs

chia/wallet/wallet_nft_store.py
```diff
@@ -143,19 +175,66 @@ async def get_nft_by_id(self, nft_id: bytes32) -> Optional[NFTCoinInfo]:
             row[4] == IN_TRANSACTION_STATUS,
         )
 
-    async def rollback_to_block(self, height: int) -> None:
+    async def get_nft_by_coin_id(self, nft_coin_id: bytes32) -> Optional[NFTCoinInfo]:
```
arvidn, line 178: this function seems redundant. one could just call `get_nft_by_coin_ids()` instead and would only have to wrap the coin id in `[` and `]`.

If this function is used in a lot of places already, maybe it could be implemented in terms of `get_nft_by_coin_ids()`.

chia/wallet/wallet_nft_store.py
```diff
@@ -53,16 +54,49 @@ async def create(cls: Type[_T_WalletNftStore], db_wrapper: DBWrapper2) -> _T_Wal
...
+            if cursor.rowcount > 0:
+                return True
+            return False
+
+    async def update_pending_transaction(self, nft_coin_id: bytes32, pending_transaction: bool) -> bool:
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
+            c = await conn.execute(
+                """
+            UPDATE users_nfts SET status=? WHERE nft_coin_id = ?
+            """,
```
arvidn, line 90: I don't think we should use tripple-quoted strings as SQL queries

chia/wallet/wallet_nft_store.py
```diff
@@ -143,19 +175,66 @@ async def get_nft_by_id(self, nft_id: bytes32) -> Optional[NFTCoinInfo]:
...
+            sql = """SELECT nft_id, coin, lineage_proof, mint_height, status, full_puzzle, latest_height
+                from users_nfts WHERE removed_height is NULL and nft_id=?"""
+            params: List[Union[uint32, str]] = [nft_id.hex()]
+            if wallet_id:
+                sql += " and wallet_id=?"
+                params.append(wallet_id)
+            row = await execute_fetchone(
+                conn,
+                sql,
+                tuple(params),
```
arvidn, line 216: you don't have to convert this to a tuple

chia/wallet/wallet_nft_store.py
```diff
@@ -143,19 +175,66 @@ async def get_nft_by_id(self, nft_id: bytes32) -> Optional[NFTCoinInfo]:
...
+            )
+
+        if row is None:
+            return None
+        return self._to_nft_coin_info(row)
+
+    async def get_nft_by_id(self, nft_id: bytes32, wallet_id: Optional[uint32] = None) -> Optional[NFTCoinInfo]:
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            sql = """SELECT nft_id, coin, lineage_proof, mint_height, status, full_puzzle, latest_height
+                from users_nfts WHERE removed_height is NULL and nft_id=?"""
```
arvidn, line 208: it looks weird to have literal new-lines in the SQL query, in my mind

chia/wallet/wallet_nft_store.py
```diff
@@ -55,20 +59,53 @@ async def create(cls: Type[_T_WalletNftStore], db_wrapper: DBWrapper2) -> _T_Wal
...
             # Remove NFT in the users_nfts table
-            await (
-                await conn.execute("UPDATE users_nfts SET removed_height=? WHERE nft_id=?", (int(height), nft_id.hex()))
-            ).close()
+            cursor = await conn.execute(
+                "UPDATE users_nfts SET removed_height=? WHERE nft_id=?", (int(height), nft_id.hex())
+            )
+            if cursor.rowcount > 0:
+                return True
+            return False
```
arvidn, line 74: this could also be:

```
return cursor.rowcount > 0
```

chia/wallet/wallet_nft_store.py
```diff
@@ -55,20 +59,53 @@ async def create(cls: Type[_T_WalletNftStore], db_wrapper: DBWrapper2) -> _T_Wal
...
+
+    async def update_pending_transaction(self, nft_coin_id: bytes32, pending_transaction: bool) -> bool:
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
+            c = await conn.execute(
+                "UPDATE users_nfts SET status=? WHERE nft_coin_id = ?",
+                (IN_TRANSACTION_STATUS if pending_transaction else DEFAULT_STATUS, nft_coin_id.hex()),
+            )
+            if c.rowcount > 0:
+                return True
+            return False
```
arvidn, line 100: this could also be:

```
return c.rowcount > 0
```

chia/wallet/wallet_nft_store.py
```diff
@@ -87,22 +124,29 @@ async def save_nft(self, wallet_id: uint32, did_id: Optional[bytes32], nft_coin_
...
+                "DELETE FROM users_nfts WHERE removed_height is not NULL and removed_height<?",
+                (int(nft_coin_info.latest_height) - REMOVE_BUFF_BLOCKS,),
+            )
+
+    async def count(self, wallet_id: Optional[uint32] = None, did_id: Optional[bytes32] = None) -> int:
+        sql = "SELECT COUNT(nft_id) FROM users_nfts WHERE removed_height is NULL"
+        if wallet_id is not None:
+            sql += f" AND wallet_id={wallet_id}"
+        if did_id:
+            sql += f" AND did_id='{did_id}'"
```
arvidn, line 138: I really think you should not put the arguments directly into the string. make these:

`sql += " AND wallet_id=?"` and `sql += " AND did_id=?'"` instead, and keep a list of the arguments to the query separately. It doesn't have to be a tuple, a list would be simpler I think.

Not only does it protect us from sql injections, but it also plays nicely with python's sqlite query cache. We won't need to re-compile these queries every time

chia/wallet/wallet_nft_store.py
```diff
@@ -87,22 +124,29 @@ async def save_nft(self, wallet_id: uint32, did_id: Optional[bytes32], nft_coin_
...
+
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            rows = list(await conn.execute_fetchall(sql))
```
arvidn, line 141: there is free function called `execute_fetchone()` which does this in one line, and returns an `Optional[Row]`.

```
from chia.util.db_wrapper import execute_fetchone
```

chia/wallet/wallet_nft_store.py
```diff
@@ -87,22 +124,29 @@ async def save_nft(self, wallet_id: uint32, did_id: Optional[bytes32], nft_coin_
...
-                )
-            ).close()
+            await conn.execute(
+                "DELETE FROM users_nfts WHERE removed_height is not NULL and removed_height<?",
+                (int(nft_coin_info.latest_height) - REMOVE_BUFF_BLOCKS,),
+            )
+
+    async def count(self, wallet_id: Optional[uint32] = None, did_id: Optional[bytes32] = None) -> int:
+        sql = "SELECT COUNT(nft_id) FROM users_nfts WHERE removed_height is NULL"
+        if wallet_id is not None:
```
arvidn, line 135: the test of `is not None` is inconsistent here and on line 137 (which uses the shorter form). I think they should be consistent, and I'm not sure whether we have a policy, or whether pylint/flake8 has opinons

chia/wallet/wallet_nft_store.py
```diff
@@ -129,18 +173,20 @@ async def get_nft_list(
...
-                "SELECT nft_id, coin, lineage_proof, mint_height, status, full_puzzle, latest_height, minter_did"
-                " from users_nfts WHERE removed_height is NULL and nft_id=?",
-                (nft_id.hex(),),
+                "SELECT EXISTS(SELECT nft_id"
+                " from users_nfts WHERE removed_height is NULL and nft_coin_id=? LIMIT 1)",
+                (coin_id.hex(),),
             )
+            if rows and rows[0] == 1:
+                return True
+            return False
```
arvidn, line 186: this could just be:
```
return rows and rows[0] == 1
```

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -284,28 +285,20 @@ async def add_coin(
...
-            if coin_info.coin == coin:
-                my_nft_coins.remove(coin_info)
-                await self.wallet_state_manager.nft_store.delete_nft(coin_info.nft_id, height)
-                delete_off_chain_metadata(coin_info.nft_id, self.wallet_state_manager.config)
-        self.wallet_state_manager.state_changed("nft_coin_removed", self.wallet_info.id)
-        return
+        nft_coin_info = await self.nft_store.get_nft_by_coin_id(coin.name())
+        if nft_coin_info:
+            await self.nft_store.delete_nft_by_coin_id(coin.name(), height)
+            delete_off_chain_metadata(nft_coin_info.nft_id, self.wallet_state_manager.config)
```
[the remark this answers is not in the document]
trepca: it's calling `unlink`, I don't think it's simple to make unlink async.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -708,7 +678,8 @@ async def generate_unsigned_spendbundle(
             elif len(payments) > 1:
                 raise ValueError("NFTs can only be sent to one party")
 
-            nft_coin = next(c for c in self.my_nft_coins if c.coin in coins)
+            nft_coin = await self.nft_store.get_nft_by_coin_ids([x.name() for x in coins])
```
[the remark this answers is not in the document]
trepca: > It also seems a bit misleading to make coins a Set, since it's only allowed to contain a single coin at most.
we're using `coins` to stay compatible with the interface of `generated_unsigned_bundle` that all other wallets use

chia/wallet/wallet_nft_store.py
trepca: ah, yes, thanks

tests/wallet/rpc/test_wallet_rpc.py
trepca: IDEs are too smart nowadays

arvidn: hopefully it follows the same rules as our `black` configuration :)

arvidn (approved):

mariano54 (approved):

Quexington (approved):

PR #12993, opened 2022-08-15
 cleanup keychain proxy 

jack60612:
follow up to changes requested in 12782

### DIFF
--- a/chia/daemon/client.py
+++ b/chia/daemon/client.py
@@ -46,8 +46,12 @@ class DaemonProxy:
             raise
 
         async def listener_task() -> None:
-            await self.listener()
-            await self.close()
+            try:
+                await self.listener()
+                await self.close()
+            except Exception:
+                await self.close()
+                raise
 
         asyncio.create_task(listener_task())
         await asyncio.sleep(1)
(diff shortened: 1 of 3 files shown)


chia/daemon/client.py
```diff
@@ -46,25 +46,30 @@ async def start(self) -> None:
...
 
         async def listener_task() -> None:
-            await self.listener()
-            await self.close()
+            try:
+                await self.listener()
+                await self.close()
+            except Exception:
+                await self.close()
+                raise
```
altendky, line 54:
```suggestion
            try:
                await self.listener()
            finally:
                await self.close()
```

chia/daemon/client.py
```diff
@@ -73,23 +78,15 @@ async def _get(self, request: WsRpcMessage) -> WsRpcMessage:
...
             self.response_dict.pop(request_id)
-        else:
-            response = None  # type: ignore
-        self._request_dict.pop(request_id)
-
-        return response
+            return response
+        except asyncio.TimeoutError:
+            self._request_dict.pop(request_id)
+            raise Exception(f"No response from daemon for request_id: {request_id}.")
```
altendky, line 89:
```suggestion
            raise Exception(f"No response from daemon for request_id: {request_id}")
```

altendky (commented):
A couple quick comments, not a full review.

chia/daemon/client.py
jack60612: 👍

chia/daemon/client.py
jack60612: oh i forgot about that!

emlowe (approved):
Looks good

PR #12999, opened 2022-08-16
enable mypy for NFT wallet

arvidn:
This removes the nft_wallet from the mypy exemption, and fixes the type annotations.


chia/wallet/util/debug_spend_bundle.py
```diff
@@ -20,7 +20,7 @@
 # we may need also to save the `genesis_coin_mod` or its hash
 
 
-def disassemble(sexp):
+def disassemble(sexp: Program):
```
trepca, line 23: should `sexp` be renamed to `program` ?

trepca (approved):
thanks!

chia/wallet/util/debug_spend_bundle.py
arvidn: `Program` derives from `SExp`, so they are interchangeable to some extent. Perhaps the type annotation really should be `SExp`. But I don't think it really matters that much, we only use this for programs afaict.

PR #13000, opened 2022-08-16
Fix the DBWrapper test

arvidn:
The original test was incorrectly assuming co-routines would be scheduled before being waited on.


almogdepaz (dismissed):

almogdepaz (approved):
lgtm

mariano54 (approved):

PR #13018, opened 2022-08-17
Load NFT off-chain metadata on backend

ytx1991:
1. Off-chain data will be fetched asynchronously when the wallet manager discovered a NFT
2. The default cache folder is .chia/mainet/nft_cache . User can config the path by setting the "nft_metadata_cache_path"
3. It will check if there is a valid cached metadata before fetching remote metadata.
4. User can delete or modify the cached metadata. They will be fetched again when calling nft_get_nfts.
5. The cached metadata will be removed when user lost the NFT ownership.
6. Add pagination for the nft_get_nfts. Use "start_index" and "num" to control. Default value is returning all NFTs.
7. Added a new string field "off_chain_data" into the response. For saving backend resource, off-chain data will not be serialized. GUI needs to serialize it.


ytx1991:
Not sure if this should merge to 1.5.1

altendky (dismissed):
What about `nft_off_chain.py` requires that it be added to the mypy leniency list?

wallentx:
> Not sure if this should merge to 1.5.1

Since this will be for 1.6.0, I think targeting this branch is okay, then switch to the next release branch after we merge this. It might be good to convert this PR to a draft for now

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,96 @@
...
+DEFAULT_PATH.mkdir(parents=True, exist_ok=True)
+log = logging.getLogger(__name__)
+
+
+def fetch_off_chain_metadata(nft_coin_info: NFTCoinInfo) -> Optional[str]:
+    uncurried_nft: Optional[UncurriedNFT] = UncurriedNFT.uncurry(*nft_coin_info.full_puzzle.uncurry())
+    if uncurried_nft is None:
+        return None
+    for uri in uncurried_nft.meta_uris.as_python():  # pylint: disable=E1133
+        response = requests.get(uri)
```
trepca, line 26: please use `aiohttp` so it doesn't block (await)

tests/wallet/nft_wallet/test_nft_off_chain.py
```diff
@@ -0,0 +1,116 @@
...
+        SINGLETON_STRUCT,
+        innerpuz.get_tree_hash(),
+        2000,
+    )
+    curried_inner = innerpuz
+    curried_ownership_layer = construct_ownership_layer(old_did, curried_tp, curried_inner)
+    return innerpuz, curried_ownership_layer
+
+
+def test_create_full_puzzle() -> Program:
```
trepca, line 69: there are no asserts in this function, why is it prefixed with `test_`? looks like a helper function, so makes more sense to name it `create_test_full_puzzle`

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -255,10 +257,14 @@ async def add_coin(
         for coin_info in my_nft_coins:
             if coin_info.coin == coin:
                 my_nft_coins.remove(coin_info)
+                delete_off_chain_metadata(coin_info.nft_id, self.wallet_state_manager.config.get(CACHE_PATH_KEY, None))
         new_nft = NFTCoinInfo(nft_id, coin, lineage_proof, puzzle, mint_height, confirmed_height)
         my_nft_coins.append(new_nft)
         await self.wallet_state_manager.nft_store.save_nft(self.id(), self.get_did(), new_nft)
         await self.wallet_state_manager.add_interested_coin_ids([coin.name()])
+        multiprocessing.Process(
```
trepca, line 265: if user receives a lot of new coins this will fork same amount of processes and given fetching metadata is mainly I/O bound an `asyncio.create_task` should be sufficient?
another potential issue here is that if we receive 1000 NFTs, we would issue 1000 requests crushing users bandwidth/connection, so should be rate limited

altendky: I would rather not spawn lots of tasks then block them all on a semaphore or such so they sit around doing nothing.  But, that's what we do now and I haven't gotten my reference async worker pool into use so go for it for right now.  The idea being to spawn as many workers as you want working and have them just read from a queue of work to be done.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1500,9 +1507,18 @@ async def nft_get_nfts(self, request) -> EndpointResult:
         wallet_id = uint32(request["wallet_id"])
         nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
         nfts = nft_wallet.get_current_nfts()
+        start_index = request.get("start_index", 0)
+        num = request.get("num", len(nfts))
```
paninaro, line 1511: I appreciate building out support for pagination. One GUI use case that this RPC doesn't natively handle is fetching _all_ NFTs across all wallets. At the moment, the GUI is calling this RPC for each NFT wallet and aggregating the results into a single list.

I can see this pagination implementation being useful when selecting a specific wallet in the GUI/CLI, but if "All NFTs" is selected, this form of pagination is not going to work.

Would it be too much trouble to add a mode here to fetch all NFTs? Perhaps we could just make the `"wallet_id"` param optional -- if None, fetch all NFTs.

paninaro: I think it should be as simple as adding an outer loop that iterates over NFT wallets

chia/rpc/wallet_rpc_api.py
```diff
@@ -1500,9 +1507,18 @@ async def nft_get_nfts(self, request) -> EndpointResult:
...
         nft_info_list = []
+        count = 0
         for nft in nfts:
-            nft_info_list.append(nft_puzzles.get_nft_info_from_puzzle(nft))
+            if count >= start_index and count - start_index < num:
+                nft_info_list.append(
+                    nft_puzzles.get_nft_info_from_puzzle(
+                        nft, self.service.wallet_state_manager.config.get(CACHE_PATH_KEY, None)
```
paninaro, line 1518: nit: Maybe just save this value to a local variable?

paninaro: I see the same line repeated a few times throughout this file

chia/wallet/nft_wallet/nft_info.py
```diff
@@ -75,6 +75,9 @@ class NFTInfo(Streamable):
     launcher_puzhash: bytes32 = LAUNCHER_PUZZLE.get_tree_hash()
     """Puzzle hash of the singleton launcher in hex"""
 
+    off_chain_data: Optional[str] = None
```
paninaro, line 78: Any objection to naming this `off_chain_metadata`? Just in case we want to support exposing other types of off-chain data at some point.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,96 @@
+import logging
+from pathlib import Path
+from typing import Optional
+
+import requests
```
paninaro, line 5: @ytx1991 @altendky I was reading about https://www.python-httpx.org/ recently -- a drop-in replacement for `requests`, but with async support (using a different API). I don't see other uses of requests in the codebase. Is it worth considering this other package?

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,96 @@
...
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.default_root import DEFAULT_ROOT_PATH
+from chia.util.hash import std_hash
+from chia.wallet.nft_wallet.nft_info import NFTCoinInfo
+from chia.wallet.nft_wallet.uncurry_nft import UncurriedNFT
+
+PREFIX_HASH_LENGTH = 3
+CACHE_PATH_KEY = "nft_metadata_cache_path"
+DEFAULT_PATH = DEFAULT_ROOT_PATH / "nft_cache"
+DEFAULT_PATH.mkdir(parents=True, exist_ok=True)
```
paninaro, line 17: @altendky is this pattern ok? Creating a directory on module import?

chia/wallet/nft_wallet/nft_off_chain.py
paninaro, line 26: We should enforce some size limit (maybe around 1 MB) to ensure that abusive/airdropped NFTs don't fill up the cache. If the metadata is larger than that, I think it's fine for the client to support manually fetching the metadata.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,96 @@
...
+        cache = Path(cache_path)
+    file_name = cache / nft_coin_info.nft_id.hex()[:PREFIX_HASH_LENGTH] / nft_coin_info.nft_id.hex()
+    if not file_name.exists():
+        return None
+    try:
+        text = file_name.read_text()
+        if verify_metadata(text, nft_coin_info.full_puzzle):
+            return text
+        else:
+            return None
```
paninaro, line 45: Debug logging for failed metadata hash verification?

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,96 @@
...
+    except Exception:
+        return None
```
paninaro, line 47: Log the exception? If the file exists and we fail to read it, it might be useful info.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,96 @@
...
+            return response.text
+    return None
+
+
+def read_off_chain_metadata(nft_coin_info: NFTCoinInfo, cache_path: Optional[str] = None) -> Optional[str]:
+    # Read metadata from disk cache
+    cache = DEFAULT_PATH
+    if cache_path is not None:
+        cache = Path(cache_path)
+    file_name = cache / nft_coin_info.nft_id.hex()[:PREFIX_HASH_LENGTH] / nft_coin_info.nft_id.hex()
```
paninaro, line 37: Minor suggestion: create a function to get the cache filename for an NFT

chia/wallet/nft_wallet/nft_off_chain.py
paninaro: This probably isn't necessary since the directory is created in `write_off_chain_metadata()`, right?

chia/rpc/wallet_rpc_api.py
paninaro, line 1518: I think including/fetching the off-chain metadata should be optional. Can we use a request param `"include_off_chain_metadata"` to control that?

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def get_off_chain_metadata(nft_coin_info: NFTCoinInfo, cache_path: Optional[str] = None) -> Optional[str]:
+    try:
+        # Check if the metadata is in disk cache
+        metadata = read_off_chain_metadata(nft_coin_info, cache_path)
+        if metadata is not None:
+            return metadata
+        log.debug(f"{nft_coin_info.nft_id.hex()} is not in cache, downloading now ...")
+        metadata = fetch_off_chain_metadata(nft_coin_info)
```
paninaro, line 77: I'm concerned that get_off_chain_metadata is called in several locations, without the ability to disable the synchronous fetching of the data (if not already cached). For some operations I think we'll want that control (`nft_set_nft_did` for example doesn't need to fetch the data).

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,96 @@
...
+
+from chia.types.blockchain_format.program import Program
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.default_root import DEFAULT_ROOT_PATH
+from chia.util.hash import std_hash
+from chia.wallet.nft_wallet.nft_info import NFTCoinInfo
+from chia.wallet.nft_wallet.uncurry_nft import UncurriedNFT
+
+PREFIX_HASH_LENGTH = 3
+CACHE_PATH_KEY = "nft_metadata_cache_path"
```
paninaro, line 15: Shouldn't we also update initial-config.yaml to include this with a default value? Otherwise users will have a hard time figuring out how to override the default path (without looking at this code).

chia/wallet/nft_wallet/nft_wallet.py
ytx1991: Are you referring thread pooling?

altendky: No, asyncio tasks that just process work from a queue.  But sure, same idea as a pool of threads or processes.  I have an old PR I need to resurrect to give us a reusable form for this.

chia/wallet/nft_wallet/nft_off_chain.py
altendky: For now we would just use the aiohttp client wouldn't we?  But sure, my impression, without being a major http user until here, is that httpx is a nice and newer lib.  There are options other than aiohttp for serving http as well that we could consider.

chia/wallet/nft_wallet/nft_off_chain.py
altendky: I don't like it, no.  Importing should not have side effects.

Separately from what we do here probably, there are proper locations for caches vs. tmp files vs. program data vs. config...  https://pypi.org/project/appdirs/ would be relevant if at some point we decided to follow the standards for that.

chia/wallet/nft_wallet/nft_off_chain.py
altendky: What are the possible exceptions here?  We certainly don't want to consume `NameError` or such.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,96 @@
...
+def read_off_chain_metadata(nft_coin_info: NFTCoinInfo, cache_path: Optional[str] = None) -> Optional[str]:
+    # Read metadata from disk cache
+    cache = DEFAULT_PATH
+    if cache_path is not None:
+        cache = Path(cache_path)
+    file_name = cache / nft_coin_info.nft_id.hex()[:PREFIX_HASH_LENGTH] / nft_coin_info.nft_id.hex()
+    if not file_name.exists():
+        return None
+    try:
+        text = file_name.read_text()
```
altendky, line 41: Don't bother pre-checking.  You will still need to sensibly handle the exception since there's a race condition here where the file could get deleted between the existence check and the read attempt.

chia/wallet/nft_wallet/nft_off_chain.py
ytx1991: For most cases we don't need to fetch metadata on the fly for any API. The metadata should already be cached at this point of time, unless user deleted or changed them manually.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,108 @@
...
+        return True
+    else:
+        log.error(
+            f"Expect metadata hash: {uncurried_nft.meta_hash.as_python().hex()},"
+            f" actual metadata hash {std_hash(str.encode(metadata)).hex()}"
+        )
+        return False
+
+
+def get_cached_filename(nft_id: bytes32, cache_path: Optional[str] = None) -> Path:
```
altendky, line 100: I see that `cache_path` being a string is propagated from existing code so I won't request a change on it here.  But, be aware that such deserialization should be done at the boundaries rather than pushing the serialized form all throughout the code.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,108 @@
...
+    if not DEFAULT_PATH.exists():
+        DEFAULT_PATH.mkdir(parents=True, exist_ok=True)
+    cache = DEFAULT_PATH
+    if cache_path is not None and cache_path != "":
```
altendky, line 104: What's wrong with the case where `cache_path` is `""`?

altendky (commented):
It looks like at least a couple conversations, I didn't check more, were resolved without making the changes.  Is there an incoming commit with code responses to those?

chia/wallet/nft_wallet/nft_wallet.py
ytx1991: I think we don't have to create another thread for loading metadata. This thread is only for this coin right?

chia/wallet/nft_wallet/nft_off_chain.py
ytx1991: The default value is "". Can we set None in yaml?

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,108 @@
...
+from chia.types.blockchain_format.program import Program
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.default_root import DEFAULT_ROOT_PATH
+from chia.util.hash import std_hash
+from chia.wallet.nft_wallet.nft_info import NFTCoinInfo
+from chia.wallet.nft_wallet.uncurry_nft import UncurriedNFT
+
+PREFIX_HASH_LENGTH = 3
+CACHE_PATH_KEY = "nft_metadata_cache_path"
+DEFAULT_PATH = DEFAULT_ROOT_PATH / "nft_cache"
```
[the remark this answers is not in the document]
ytx1991: If there are all configs we have to pass many parameter to the methods. Can you point me the helper function you mentioned?

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,108 @@
...
+async def fetch_off_chain_metadata(nft_coin_info: NFTCoinInfo, ignore_size_limit: bool) -> Optional[str]:
+    uncurried_nft: Optional[UncurriedNFT] = UncurriedNFT.uncurry(*nft_coin_info.full_puzzle.uncurry())
+    if uncurried_nft is None:
+        log.error(f"Cannot fetch off-chain metadata, {nft_coin_info.nft_id} is not a NFT.")
+        return None
+    for uri in uncurried_nft.meta_uris.as_python():  # pylint: disable=E1133
+        async with aiohttp.ClientSession() as session:
+            async with session.get(str(uri, "utf-8")) as response:
+                if response.status == 200:
+                    text = await response.text()
```
trepca, line 31: wouldn't this load all fetched file content into memory?

trepca: let's say metadata file is 100GB in size, wouldn't this try to load the whole file content into `text` variable?

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+async def fetch_off_chain_metadata(nft_coin_info: NFTCoinInfo, ignore_size_limit: bool) -> Optional[str]:
+    uncurried_nft: Optional[UncurriedNFT] = UncurriedNFT.uncurry(*nft_coin_info.full_puzzle.uncurry())
+    if uncurried_nft is None:
+        log.error(f"Cannot fetch off-chain metadata, {nft_coin_info.nft_id} is not a NFT.")
+        return None
+    for uri in uncurried_nft.meta_uris.as_python():  # pylint: disable=E1133
+        async with aiohttp.ClientSession() as session:
+            async with session.get(str(uri, "utf-8")) as response:
```
trepca, line 31: timeout by default is 5 minutes, which has a big impact to how long it could take to load all metadata (especially over time when NFTs will have a lot of obsolete metadata URLs). 30 seconds should be enough.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,115 @@
...
+    try:
+        # Check if the metadata is in disk cache
+        metadata = read_off_chain_metadata(nft_coin_info, config)
+        if metadata is not None:
+            return metadata
+        log.debug(f"{nft_coin_info.nft_id.hex()} is not in cache, downloading now ...")
+        metadata = await fetch_off_chain_metadata(nft_coin_info, ignore_size_limit)
+        if metadata is None:
+            log.error(f"Cannot find off-chain metadata of {nft_coin_info.nft_id.hex()}.")
+            return None
```
trepca, line 79: I think if `fetch_off_chain_metadata` always returns None due to bad connection/not existing anymore, we still try to fetch it again forever whenever we call `get_nft_info_from_puzzle`, right?

trepca (changes requested):

chia/wallet/nft_wallet/nft_off_chain.py
trepca:
```
>>> yaml.safe_load("var: null")
{'var': None}
>>> yaml.safe_load("var: ")
{'var': None}
>>> yaml.safe_load("var: ''")
{'var': ''}
>>> yaml.safe_load('var: ""')
{'var': ''}
```

chia/wallet/nft_wallet/nft_off_chain.py
ytx1991: What is the solution for this in Python? I was thinking use content-length to check the file size but it is not reliable.

chia/wallet/nft_wallet/nft_wallet.py
ytx1991: I tried asyncio, but it doesn't work here because of it is in an async event loop.

chia/wallet/nft_wallet/nft_off_chain.py
ytx1991: Why we don't want to catch NameError? It will reload metadata no matter what kind of exception.

chia/wallet/nft_wallet/nft_off_chain.py
ytx1991: Yes, if there isn't a valid cached file it will keep fetching until success.

chia/wallet/nft_wallet/nft_off_chain.py
trepca: you can use the streaming API: https://docs.aiohttp.org/en/stable/streams.html

for example:
```python
async with aiohttp.ClientSession() as session:
    async with session.get(str(uri, "utf-8")) as response:
        if response.status == 200:
            async for chunk in response.content.iter_chunked(1024):
                file_obj.write(chunk)
```

paninaro:
Since the normal mode of operation for this is to fetch the metadata when an NFT coin is added, how are we planning to handle upgrades for users that already have their wallets synced?

ytx1991:
> Since the normal mode of operation for this is to fetch the metadata when an NFT coin is added, how are we planning to handle upgrades for users that already have their wallets synced?

I don't know what is the concern. For synced users, the cache will be fetched on the first time of calling nft_get_nfts.

chia/wallet/nft_wallet/nft_off_chain.py
paninaro: I'm happy with the logging addition. I'm not sure I understand what the `NameError` discussion is about @altendky

paninaro:
> > Since the normal mode of operation for this is to fetch the metadata when an NFT coin is added, how are we planning to handle upgrades for users that already have their wallets synced?
> 
> I don't know what is the concern. For synced users, the cache will be fetched on the first time of calling nft_get_nfts.

My concern is that users that currently have a large collection will have a poor experience immediately after upgrading. The GUI may appear to be unresponsive while the metadata is being fetched. I don't want to hold this PR up, but maybe there's something we can do as a follow-up to make this experience better.

paninaro (dismissed):
lgtm

chia/rpc/wallet_rpc_api.py
```diff
@@ -1497,12 +1503,29 @@ async def nft_mint_nft(self, request) -> EndpointResult:
...
+        wallet_id = request.get("wallet_id", None)
+        nfts = []
+        if wallet_id is not None:
+            nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
+            nfts.extend(nft_wallet.get_current_nfts())
+        else:
+            for wallet in self.service.wallet_state_manager.wallets.values():
+                if wallet.type() == WalletType.NFT.value:
+                    nfts.extend(wallet.get_current_nfts())
+        start_index = request.get("start_index", 0)
```
trepca, line 1515: if you're adding pagination, there should also be pagination tests

trepca (changes requested):
thanks, lets add tests for pagination and should be it

trepca (dismissed):

paninaro (dismissed):
lgtm

chia/util/initial-config.yaml
```diff
@@ -491,7 +491,10 @@ wallet:
   full_node_peer:
     host: *self_hostname
     port: 8444
-
+  # The path of NFT off-chain metadata cache
+  nft_metadata_cache_path: "./nft_cache"
```
altendky, line 495: See the other comment about `chia.util.path.path_from_root()`.

```suggestion
  nft_metadata_cache_path: "nft_cache"
```

setup.py
```diff
@@ -33,6 +33,7 @@
     "typing-extensions==4.3.0",  # typing backports like Protocol and TypedDict
     "zstd==1.5.0.4",
     "packaging==21.3",
+    "aiohttp==3.8.1",
```
altendky, line 36: We already have this above.

```suggestion
```

chia/rpc/wallet_rpc_api.py
```diff
@@ -1497,13 +1503,34 @@ async def nft_mint_nft(self, request) -> EndpointResult:
...
+                        self.service.wallet_state_manager.config,
+                        request.get("include_off_chain_metadata", False),
+                        request.get("ignore_size_limit", False),
+                    )
+                    nft_info_list.append(nft_info)
+                count += 1
+            return {"wallet_id": wallet_id, "success": True, "nft_list": nft_info_list}
+        except Exception as e:
+            log.exception("Failed to get NFTs.")
+            return {"success": False, "error": f"Failed to get NFTs: {e}"}
```
altendky, line 1533: Doesn't the RPC framework handle logging and returning errors?  If kept more as it already is, it needs to be made to not consume `asyncio.CancelledError`.

```suggestion
        wallet_id = request.get("wallet_id", None)
        nfts = []
        if wallet_id is not None:
            nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
            nfts.extend(nft_wallet.get_current_nfts())
        else:
            for wallet in self.service.wallet_state_manager.wallets.values():
                if wallet.type() == WalletType.NFT.value:
                    nfts.extend(wallet.get_current_nfts())
        start_index = request.get("start_index", 0)
        num = request.get("num", len(nfts))
        nft_info_list = []
        count = 0
        for nft in nfts:
            if count >= start_index and count - start_index < num:
                nft_info = await nft_puzzles.get_nft_info_from_puzzle(
                    nft,
                    self.service.wallet_state_manager.config,
                    request.get("include_off_chain_metadata", False),
                    request.get("ignore_size_limit", False),
                )
                nft_info_list.append(nft_info)
            count += 1
        return {"wallet_id": wallet_id, "success": True, "nft_list": nft_info_list}
```

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,121 @@
...
+
+from chia.types.blockchain_format.program import Program
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.default_root import DEFAULT_ROOT_PATH
+from chia.util.hash import std_hash
+from chia.wallet.nft_wallet.nft_info import NFTCoinInfo
+from chia.wallet.nft_wallet.uncurry_nft import UncurriedNFT
+
+DEFAULT_PREFIX_HASH_LENGTH = 3
+DEFAULT_CACHE_PATH = "./nft_cache"
```
altendky, line 15: See the other comment about `chia.util.path.path_from_root()`.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,121 @@
...
+            log.warning(f"Invalid cached metadata, NFT ID: {nft_coin_info.nft_id.hex()}")
+            return None
+    except Exception:
+        log.exception(f"Cannot load cached metadata of {nft_coin_info.nft_id.hex()}.")
+        return None
+
+
+def delete_off_chain_metadata(nft_id: bytes32, config: Optional[Dict[str, Any]] = None) -> None:
+    file_name = get_cached_filename(nft_id, config)
+    if file_name.exists():
```
altendky, line 63: Don't pre-check, just handle the exception raised in the case it isn't there.  As is, the file could still be deleted between this check and the `.unlink()` below so the exception could still happen.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,121 @@
...
+    if config is None:
+        cache_path = DEFAULT_CACHE_PATH
+        hash_length = DEFAULT_PREFIX_HASH_LENGTH
+    else:
+        cache_path = str(config.get(CACHE_PATH_KEY, DEFAULT_CACHE_PATH))
+        hash_length = int(config.get(PREFIX_HASH_LENGTH_KEY, DEFAULT_PREFIX_HASH_LENGTH))
+    if cache_path.startswith("./") or cache_path.startswith(".\\"):
+        cache = DEFAULT_ROOT_PATH / cache_path[2:]
+    else:
+        cache = Path(cache_path)
```
altendky, line 118: Take a look at uses of `chia.util.path.path_from_root()`.  Then remove the `./` from other spots.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,121 @@
...
+        return False
+
+
+def get_cached_filename(nft_id: bytes32, config: Optional[Dict[str, Any]]) -> Path:
+    if config is None:
+        cache_path = DEFAULT_CACHE_PATH
+        hash_length = DEFAULT_PREFIX_HASH_LENGTH
+    else:
+        cache_path = str(config.get(CACHE_PATH_KEY, DEFAULT_CACHE_PATH))
+        hash_length = int(config.get(PREFIX_HASH_LENGTH_KEY, DEFAULT_PREFIX_HASH_LENGTH))
```
altendky, line 114: Shouldn't these already be of these types anyways?

```suggestion
        cache_path: str = config.get(CACHE_PATH_KEY, DEFAULT_CACHE_PATH)
        hash_length: int = config.get(PREFIX_HASH_LENGTH_KEY, DEFAULT_PREFIX_HASH_LENGTH)
```

chia/wallet/nft_wallet/nft_off_chain.py
altendky: You don't want to catch `NameError` because it indicates a variable name typo or such.  That should fail out hard so we see it quickly and fix it, not hide in the logs where we forget to look while developing and only hear about it when a user sends us a log asking about it.  This is a general commentary on us catching all exceptions in lots of places without thinking about, or often even being aware of, all the relevant exceptions.  `NameError` in particular is just an example.

altendky (dismissed):

emlowe:
This is out of 1.6.0, marking as Draft to avoid any accidental merging and removed 1.6.0 label

chia/util/initial-config.yaml
ytx1991: ./ is used to distinguish the relative path and absolute path. User may not want to cache the off-chain metadata on their system disk.

chia/wallet/nft_wallet/nft_off_chain.py
paninaro: What if we just re-raise the exception here after logging it?

paninaro (approved):
lgtm

chia/wallet/nft_wallet/nft_off_chain.py
trepca: have opened a new issue re catching generic exception

chia/util/initial-config.yaml
altendky: This is no different than our other big stuff.  It need not be defaulted differently here.  Using `./` just means it will go any number of places depending what directory the user is in when they run chia.

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,125 @@
...
+def get_cached_filename(nft_id: bytes32, config: Optional[Dict[str, Any]]) -> Path:
+    cache_path: str = DEFAULT_CACHE_PATH
+    hash_length: int = DEFAULT_PREFIX_HASH_LENGTH
+    if config is not None:
+        cache_path = config.get(CACHE_PATH_KEY, DEFAULT_CACHE_PATH)
+        hash_length = config.get(PREFIX_HASH_LENGTH_KEY, DEFAULT_PREFIX_HASH_LENGTH)
+    if re.match("^(/.*)|([a-zA-z]:.*)", cache_path):
+        cache = Path(cache_path)
+    else:
+        cache = path_from_root(DEFAULT_ROOT_PATH, cache_path)
```
altendky, line 121: Why does this need to deviate from existing usage?

chia/wallet/nft_wallet/nft_off_chain.py
```diff
@@ -0,0 +1,125 @@
...
+        log.exception(f"Cannot load cached metadata of {nft_coin_info.nft_id.hex()}.")
+        return None
+
+
+def delete_off_chain_metadata(nft_id: bytes32, config: Optional[Dict[str, Any]] = None) -> None:
+    file_name = get_cached_filename(nft_id, config)
+    try:
+        file_name.unlink()
+        log.info(f"Deleted off-chain metadata of {nft_id.hex()}")
+    except Exception:
```
altendky, line 68: I think `OSError` would be a lot more apt.  That would avoid caching `AttributeError`, `NameError`, etc that we _want_ to see asap.

chia/wallet/nft_wallet/nft_off_chain.py
altendky: This in particular was mostly just a discussion.  Reraising is an thing but sometimes we end up double logging then.  Unexpected exceptions should generally bubble really far up and then be reported clearly to developers.  None of the `except Exception:` logging here provides _any_ information about what happened.

chia/wallet/nft_wallet/nft_off_chain.py
altendky: Also report some info from the exception.

PR #13021, opened 2022-08-17
implement SerializedProgram.curry() to speed up puzzle-hash generation

arvidn:
The main motivation behind this is to speed up generating new puzzle hashes. The bottleneck right now is `puzzle_for_pk()` and `get_tree_hash()`.

`puzzle_for_pk()` is slow because it has to call `curry()`, and `get_tree_hash()` is slow because it has to operate on `Program` (the python tree structure) instead of `SerializedProgram`.

This patch implements `curry()` to operate directly on `SerializedProgram`, which avoids the critical path from having to parse the program and speeds up `curry()` as well as `get_tree_hash()`.

| --- | --- | --- | --- |
| puzzle_for_pk() | 17.15% | 9.64% | 0.56 |
| get_tree_hash() | 25.40% | 6.62% | 0.25 |

The profile from current main:
chia-hotspot-0

The profile with this patch:
chia-hotspot-0-patched

### DIFF
--- a/chia/types/blockchain_format/program.py
+++ b/chia/types/blockchain_format/program.py
@@ -133,6 +133,8 @@ class Program(SExp):
     def curry(self, *args) -> "Program":
         fixed_args: Any = 1
         for arg in reversed(args):
+            if type(arg) == SerializedProgram:
+                arg = arg.to_program()
             fixed_args = [4, (1, arg), fixed_args]
         return Program.to([2, (1, self), fixed_args])
 
@@ -190,8 +192,13 @@ def _tree_hash(node: SExp, precalculated: Set[bytes32]) -> bytes32:
 
 
 def _serialize(node) -> bytes:
-    if type(node) == SerializedProgram:
+    t = type(node)
+    if t == SerializedProgram:
         return bytes(node)
+    elif t == bytes32:
+        # 0xA0 is the length prefix for a 32 byte atom
+        # 0x80 + 0x20
+        return b"\xa0" + node
     else:
         return SExp.to(node).as_bin()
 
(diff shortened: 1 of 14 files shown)


chia/types/blockchain_format/program.py
```diff
@@ -133,6 +133,8 @@ def run(self, args) -> "Program":
     def curry(self, *args) -> "Program":
         fixed_args: Any = 1
         for arg in reversed(args):
+            if type(arg) == SerializedProgram:
```
richardkiss, line 136: Is this really necessary? As long as both `Program` and `SerializedProgram` descend from `CLVMObject`, they should interoperate.

richardkiss: I see, `SerializedProgram` is a base class. What happens if you make it a subclass of `CLVMObject` (or even `Program`)? You'd have to override `.atom` and `.pair` (decorate with `@property def atom(self): ...`).

richardkiss: Check for `0xff` as the first byte to determine `atom` vs `pair`. Then call `deserialized` and populate `._atom` or `._pair` so you only have to deserialize once (this is similar to the strategy I'm trying in my rewrite).

arvidn: right, `SerializedProgram` doesn't implement `CLVMObject` or derive from `SExp`. I would expect it would be more complicated to implement `CLVMObject` than adding this special case.

richardkiss (commented):
I'm sure this is a lot faster, but I'm just too concerned about `SerializedProgram` not being a `CLVMObject` subclass. If you can do that, the changes you need to make will drop drastically, and I doubt it will be any slower.

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -375,15 +375,17 @@ async def get_new_inner_hash(self) -> bytes32:
...
 
-    async def get_new_inner_puzzle(self) -> Program:
+    async def get_new_inner_puzzle(self) -> SerializedProgram:
         return await self.standard_wallet.get_new_puzzle()
 
     async def get_new_puzzlehash(self) -> bytes32:
         return await self.standard_wallet.get_new_puzzlehash()
 
-    def puzzle_for_pk(self, pubkey: G1Element) -> Program:
+    def puzzle_for_pk(self, pubkey: G1Element) -> SerializedProgram:
```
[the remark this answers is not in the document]
arvidn: this is part of the critical path. this is the function that has to return a `SerializedProgram` in order for `get_tree_hash()` to be fast later.

chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py
```diff
@@ -61,17 +61,17 @@
...
-from chia.types.blockchain_format.program import Program
+from chia.types.blockchain_format.program import Program, SerializedProgram
 from chia.types.blockchain_format.sized_bytes import bytes32
 
-from .load_clvm import load_clvm
+from .load_clvm import load_serialized_clvm
 from .p2_conditions import puzzle_for_conditions
 
-DEFAULT_HIDDEN_PUZZLE = Program.from_bytes(bytes.fromhex("ff0980"))
+DEFAULT_HIDDEN_PUZZLE = SerializedProgram.from_bytes(b"\xff\x09\x80")
```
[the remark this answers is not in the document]
arvidn: I suppose this only contributes to start-up cost

richardkiss:
@arvidn  I believe this can be closed as we solved this in a different way.

PR #13029, opened 2022-08-17
skip validating GTElement in mempool

arvidn:
the mempool has a `ProcessExecutor` with two subprocesses. These processes are use to run validation of incoming transactions, possibly to be added to the mempool. Validation includes executing clvm and checking BLS signatures.

In order to accelerate new blocks that are likely to include transactions in our mempool, we have a BLS cache that allows us to not do all the work again to validate these signatures. The validation processes pass back the `GTElement` objects to the main thread, which then stashes them in the BLS cache.

Since the subprocesses performing validation is part of our program, and can be trusted not to send back invalid `GTElement` objects, we don't need to perform the validation check on them before adding them to the BLS cache.

This patch saves a fair amount of CPU in the main thread when adding transactions to the mempool.

| --- | --- | --- | --- |
| add_spend_bundles() | 18.08% | 10.56% | 0.58 |

Profile of main:
mempool-add-mt-main

Profile of this patch:
mempool-add-mt-patched

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -276,7 +276,7 @@ class MempoolManager:
         if err is not None:
             raise ValidationError(err)
         for cache_entry_key, cached_entry_value in new_cache_entries.items():
-            LOCAL_CACHE.put(cache_entry_key, GTElement.from_bytes(cached_entry_value))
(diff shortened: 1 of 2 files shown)


mariano54 (approved):
very nice change! I had never measured this before. I'm glad the measurement stuff is paying off

PR #13034, opened 2022-08-17
Prep for `puzzle_hash_for_pk` optimization.

richardkiss:


arvidn:
this is a good idea and seems to have a significant speed-up. There should be tests for the new functions though.

chia/wallet/util/generate_curry_and_treehash_f.py
```diff
@@ -0,0 +1,72 @@
...
+            shatree_pair(curried_values_tree_hash(arguments[1:]), NULL_TREEHASH),
+        ),
+    )
+
+
+# The curry pattern is `(a . ((q . F)  . (E . 0)))` == `(a (q . F) E)
+# where `F` is the `mod` and `E` is the curried environment
+
+
+def generate_curry_and_treehash_f(mod_hash: bytes32) -> CurryHashFunction:
```
arvidn, line 61: the `_f` suffix is meant to indicate it returns a function?

chia/wallet/util/generate_curry_and_treehash_f.py
```diff
@@ -0,0 +1,72 @@
+from hashlib import sha256
+from typing import Callable, List
+
+from chia.types.blockchain_format.sized_bytes import bytes32
+
+CurryHashFunction = Callable[..., bytes32]
```
arvidn, line 6: There's no way to type the parameters to the `Callable` to be `bytes32`? Something like: `Callable[[bytes32...], bytes32]`

chia/wallet/wallet_state_manager.py
```diff
@@ -83,6 +83,19 @@
 from chia.wallet.wallet_user_store import WalletUserStore
 
 
+# this is a placeholder function to bridge wallets that don't yet have
+# the `puzzle_hash_for_pk` optimization in them
```
arvidn, line 87: If we were to define a Protocol for wallets, and require wallets to implement it. Would mypy be able to tell us about wallets not implementing new functions like this at build-time?

tests/clvm/test_generate_curry_and_treehash.py
```diff
@@ -0,0 +1,35 @@
...
+        args = [v, v * v, v * v * v]
+        # we don't really care about the arguments either
+        puzzle = arbitrary_mod.curry(*args)
+        puzzle_hash_via_curry = puzzle.get_tree_hash()
+        hashed_args = [Program.to(_).get_tree_hash() for _ in args]
+        puzzle_hash_via_f = f(*hashed_args)
+        assert puzzle_hash_via_curry == puzzle_hash_via_f
+
+
+def mod_api_check(module: Any) -> None:
```
arvidn, line 26: would it work to make this `object` instead of `Any`?
It looks like a WalletProtocol` would come in handy here

arvidn (dismissed):

chia/wallet/util/generate_curry_and_treehash_f.py
richardkiss: The problem is I'm using `(*arguments)` to allow a variable number of `bytes32` arguments. There are mypy extensions for this (`Vararg`), it involves importing from mypy which I don't want to do as it seems like a bad idea to require mypy at runtime. I couldn't figure out a better way, and I think the varargs form is a better api, at the cost of a little type-checking.

chia/wallet/util/generate_curry_and_treehash_f.py
richardkiss: Yes, exactly.

richardkiss: It might actually be overkill... we could probably just export the inner function and have `mod_hash` be a parameter, and assume the caller to be smart enough to pre-cache it where appropriate. It would probably make it easier to understand for the casual observer too. Do you have an opinion on this?

chia/wallet/wallet_state_manager.py
richardkiss: Yes, for sure. I think it's a good idea, but it's probably out of scope of this patch (and I wouldn't call myself an expert on this wallet code).

tests/clvm/test_generate_curry_and_treehash.py
richardkiss: At first I thought this function would be property of a module, but it turns out to be property of `Wallet` subclasses (not real subclasses, just duck-typing subclasses), and they are pretty hard to create in a vacuum, so this test either has to use a `mock` or import a lot of junk just to set up.

chia/wallet/util/generate_curry_and_treehash_f.py
arvidn: I have a weak preference for exporting the inner function directly. I think it's simpler and more straight-forward. I'm pretty sure all wallets do cache the tree-hash of their puzzle already as well.

chia/wallet/util/generate_curry_and_treehash_f.py
arvidn: I see. makes sense. However, this problem goes away if you would export the curry_treehash function directly as well (iiuc)

chia/wallet/util/generate_curry_and_treehash_f.py
arvidn: I see, it's not the treehash of the puzzle directly, it's the hash of the quoted puzzle

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -59,6 +60,10 @@
 
 # This should probably not live in this file but it's for experimental right now
 
+CAT_MOD_HASH = CAT_MOD.get_tree_hash()
+CAT_MOD_HASH_HASH = Program.to(CAT_MOD_HASH).get_tree_hash()
```
arvidn, line 64: I don't understand this. `Program.to()` will create a single atom containing the hash. the tree hash of that atom will be different from `(q . <CAT_MOD>)`, right?

richardkiss: Yes, the `q` addition is done in `generate_curry_and_treehash_f`.

chia/wallet/util/generate_curry_and_treehash_f.py
richardkiss: Right, so that's a detail that the caller would have to deal with. It would have to be clearly documented.

arvidn (dismissed):

arvidn:
we'll need this optimization for the DID wallet too

richardkiss:
> we'll need this optimization for the DID wallet too

This one is too involved; I'll have to delegate to a Matt.

chia/wallet/did_wallet/did_wallet_puzzles.py
```diff
@@ -75,6 +85,21 @@ def create_fullpuz(innerpuz: Program, launcher_id: bytes32) -> Program:
...
+    """
+    Create a full puzzle of DID
+    :param innerpuz_hash: DID inner puzzle tree hash
+    :param launcher_id: launcher coin name
+    :return: DID full puzzle hash
+    """
+    mod_hash = SINGLETON_TOP_LAYER_MOD.get_tree_hash()
+    quoted_mod_hash = calculate_hash_of_quoted_mod_hash(SINGLETON_TOP_LAYER_MOD.get_tree_hash())
+    # singleton_struct = (MOD_HASH . (LAUNCHER_ID . LAUNCHER_PUZZLE_HASH))
+    singleton_struct = Program.to((mod_hash, (launcher_id, LAUNCHER_PUZZLE.get_tree_hash())))
```
arvidn, line 98: it's a bit of a nit-pick, but it seems unnecessary to compute tree-hashes of constants every time we want a new puzzle hash. We do that 3 times in here.

arvidn (approved):
maybe it would make sense to have @Quexington or @matt-o-how review some of this too. looks good to me

chia/wallet/rl_wallet/rl_wallet.py
```diff
@@ -418,6 +418,13 @@ def puzzle_for_pk(self, pk) -> Optional[Program]:
             clawback_pk=self.rl_info.admin_pubkey,
         )
 
+    def puzzle_hash_for_pk(self, pubkey: G1Element) -> Optional[bytes32]:
+        # TODO: rewrite using curry_and_treehash
+        puzzle = self.puzzle_for_pk(pubkey)
+        if puzzle is None:
+            return puzzle
+        return puzzle.get_tree_hash()
+
```
Quexington, line 427: How is this wallet still around 🤦

Quexington (approved):

trepca (approved):

chia/wallet/did_wallet/did_wallet_puzzles.py
trepca: lets fix this in a separate PR

tests/clvm/test_curry_and_treehash.py
```diff
@@ -0,0 +1,21 @@
...
+
+def test_curry_and_treehash() -> None:
+    arbitrary_mod = p2_delegated_puzzle_or_hidden_puzzle.MOD
+    arbitrary_mod_hash = arbitrary_mod.get_tree_hash()
+
+    # we don't really care what `arbitrary_mod` is. We just need some code
+
+    quoted_mod_hash = calculate_hash_of_quoted_mod_hash(arbitrary_mod_hash)
+
+    for v in range(500):
```
matt-o-how, line 14: Shouldn't we test edgecases like large strings and negatives rather than just positive integers 500 times?

PR #13037, opened 2022-08-18
cmds: Don't `exit` in `migrate_legacy_keyring_interactive`

xdustinface:
Exiting leads to the actual start command not running after the mirgation was done successfully. Also we have a daemon connection open here if we trigger the migration via `chia start` commands and exiting in there leaves the connection opened which leads to exception traceback outputs.


paninaro (approved):
lgtm

PR #13044, opened 2022-08-18
coin interest optimization for trade records

trepca:
this one makes a big difference for users with a lot of offers (tested on bulk minter)

we currently load all trade records, extract their coins of interest and then filter through all. This patch creates a lookup table  so we can handle many more trade records as we're not loading everything in memory and reduces time complexity.

### DIFF
--- a/chia/wallet/trade_manager.py
+++ b/chia/wallet/trade_manager.py
@@ -96,7 +96,7 @@ class TradeManager:
 
     async def get_coins_of_interest(
         self,
-    ) -> List[bytes32]:
+    ) -> Set[bytes32]:
         """
         Returns list of coins we want to check if they are included in filter,
         These will include coins that belong to us and coins that that on other side of treade
--- a/chia/wallet/trade_manager.py
+++ b/chia/wallet/trade_manager.py
@@ -96,18 +96,15 @@ class TradeManager:
 
     async def get_coins_of_interest(
         self,
-    ) -> Dict[bytes32, Coin]:
+    ) -> List[bytes32]:
         """
         Returns list of coins we want to check if they are included in filter,
         These will include coins that belong to us and coins that that on other side of treade
         """
-        coins = await self.trade_store.get_coins_of_interest_with_trade_statuses(
+        coin_ids = await self.trade_store.get_coin_ids_of_interest_with_trade_statuses(
             trade_statuses=[TradeStatus.PENDING_ACCEPT, TradeStatus.PENDING_CONFIRM, TradeStatus.PENDING_CANCEL]
         )
-        interested_dict = {}
-        for coin in coins:
-            interested_dict[coin.name()] = coin
-        return interested_dict
+        return coin_ids
 
     async def get_trade_by_coin(self, coin: Coin) -> Optional[TradeRecord]:
         all_trades = await self.get_all_trades()
--- a/chia/wallet/trade_manager.py
+++ b/chia/wallet/trade_manager.py
@@ -19,6 +19,7 @@ from chia.wallet.nft_wallet.nft_wallet import NFTWallet
 from chia.wallet.outer_puzzles import AssetType
 from chia.wallet.payment import Payment
 from chia.wallet.puzzle_drivers import PuzzleInfo
+from chia.wallet.puzzles.load_clvm import load_clvm
 from chia.wallet.trade_record import TradeRecord
 from chia.wallet.trading.offer import NotarizedPayment, Offer
 from chia.wallet.trading.trade_status import TradeStatus
@@ -28,7 +29,6 @@ from chia.wallet.util.transaction_type import TransactionType
 from chia.wallet.util.wallet_types import WalletType
 from chia.wallet.wallet import Wallet
 from chia.wallet.wallet_coin_record import WalletCoinRecord
-from chia.wallet.puzzles.load_clvm import load_clvm
 
 OFFER_MOD = load_clvm("settlement_payments.clvm")
 
@@ -101,19 +101,12 @@ class TradeManager:
         Returns list of coins we want to check if they are included in filter,
         These will include coins that belong to us and coins that that on other side of treade
         """
-        all_pending = []
-        pending_accept = await self.get_offers_with_status(TradeStatus.PENDING_ACCEPT)
-        pending_confirm = await self.get_offers_with_status(TradeStatus.PENDING_CONFIRM)
-        pending_cancel = await self.get_offers_with_status(TradeStatus.PENDING_CANCEL)
-        all_pending.extend(pending_accept)
-        all_pending.extend(pending_confirm)
-        all_pending.extend(pending_cancel)
+        coins = await self.trade_store.get_coins_of_interest_with_trade_statuses(
+            trade_statuses=[TradeStatus.PENDING_ACCEPT, TradeStatus.PENDING_CONFIRM, TradeStatus.PENDING_CANCEL]
+        )
         interested_dict = {}
-
-        for trade in all_pending:
-            for coin in trade.coins_of_interest:
-                interested_dict[coin.name()] = coin
-
+        for coin in coins:
+            interested_dict[coin.name()] = coin
         return interested_dict
 
     async def get_trade_by_coin(self, coin: Coin) -> Optional[TradeRecord]:
--- a/chia/wallet/trading/trade_store.py
+++ b/chia/wallet/trading/trade_store.py
@@ -9,7 +9,7 @@ from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.mempool_inclusion_status import MempoolInclusionStatus
 from chia.util.db_wrapper import DBWrapper2
 from chia.util.errors import Err
-from chia.util.ints import uint8, uint32
+from chia.util.ints import uint8, uint32, uint64
 from chia.wallet.trade_record import TradeRecord
 from chia.wallet.trading.trade_status import TradeStatus
 
@@ -18,9 +18,7 @@ async def migrate_coin_of_interest(log: logging.Logger, db: aiosqlite.Connection
     log.info("Beginning migration of coin_of_interest_to_trade_record lookup table")
 
     start_time = perf_counter()
-    cursor = await db.execute("SELECT trade_record, trade_id from trade_records")
-    rows = await cursor.fetchall()
-    await cursor.close()
+    rows = await db.execute_fetchall("SELECT trade_record, trade_id from trade_records")
 
     inserts: List[Tuple[bytes32, bytes, str]] = []
     for row in rows:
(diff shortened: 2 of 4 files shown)


chia/wallet/trading/trade_store.py
```diff
@@ -236,6 +301,25 @@ async def get_trade_record_with_status(self, status: TradeStatus) -> List[TradeR
...
 
+    async def get_coins_of_interest_with_trade_statuses(self, trade_statuses: List[TradeStatus]) -> List[Coin]:
+        """
+        Checks DB for TradeRecord with id: id and returns it.
+        """
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            cursor = await conn.execute(
+                "SELECT distinct c.coin from coin_of_interest_to_trade_record c, trade_records t "
+                "WHERE t.status in (%s) AND c.trade_id = t.trade_id" % ",".join("?" * len(trade_statuses)),
+                ([x.value for x in trade_statuses]),
```
arvidn, line 312: you don't actually have to convert the list to a tuple. the sqlite API accepts anything `Iterable`

chia/wallet/trading/trade_store.py
```diff
@@ -236,6 +301,25 @@ async def get_trade_record_with_status(self, status: TradeStatus) -> List[TradeR
...
         return records
 
+    async def get_coins_of_interest_with_trade_statuses(self, trade_statuses: List[TradeStatus]) -> List[Coin]:
+        """
+        Checks DB for TradeRecord with id: id and returns it.
+        """
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            cursor = await conn.execute(
+                "SELECT distinct c.coin from coin_of_interest_to_trade_record c, trade_records t "
+                "WHERE t.status in (%s) AND c.trade_id = t.trade_id" % ",".join("?" * len(trade_statuses)),
```
arvidn, line 311: generally I think we need to be moving away from building SQL query strings in code (because it's risky and may defeat the compiler query cache in python's sqlite binding). However, the pattern of using `in (?,?,...)` with a variable number of arguments is hard to get away from (afaik). However, we seem to have a convention to build the list of question marks with an f-string in other places. It might make sense to conform to that just for consistency.

i.e.

```
f"WHERE t.status in ({','.join('?' * len(trade_statuses))}) AND c.trade_id = t.trade_id"
```

Come to think of it, we should probably factor out this pattern into a utility function, since we use it in quite a few places. (out of scope for you patch clearly)

chia/wallet/trading/trade_store.py
```diff
@@ -236,6 +301,25 @@ async def get_trade_record_with_status(self, status: TradeStatus) -> List[TradeR
 
         return records
 
+    async def get_coins_of_interest_with_trade_statuses(self, trade_statuses: List[TradeStatus]) -> List[Coin]:
+        """
+        Checks DB for TradeRecord with id: id and returns it.
+        """
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            cursor = await conn.execute(
```
arvidn, line 309: if you use `execute_fetchall()` here instead, you can avoid the extra two round-trips to the sqlite thread in the `fetchall()` and `close()` calls

chia/wallet/trading/trade_store.py
```diff
@@ -78,6 +108,28 @@ async def create(
                 )
             )
 
+            await conn.execute(
+                (
+                    "CREATE TABLE IF NOT EXISTS coin_of_interest_to_trade_record("
+                    " trade_id text,"
```
arvidn, line 114: I don't think we should create new tables where we store binary data as hex strings

chia/wallet/trading/trade_store.py
```diff
@@ -78,6 +108,28 @@ async def create(
                 )
             )
 
+            await conn.execute(
+                (
+                    "CREATE TABLE IF NOT EXISTS coin_of_interest_to_trade_record("
```
arvidn, line 113: is it no possible to add an index to the database to fill the function of this new table?
It might be more space efficient to let sqlite take care of the indexing

chia/wallet/trading/trade_store.py
```diff
@@ -13,6 +14,35 @@
 from chia.wallet.trading.trade_status import TradeStatus
 
 
+async def migrate_coin_of_interest(log: logging.Logger, db: aiosqlite.Connection) -> None:
```
arvidn, line 17: it seems it would be important for the creation of the table and the migration all be part of the same transaction. Is that the case?
If migration fails half-way through, will the table still have been created?

arvidn (changes requested):
This looks like a manual index of an existing table (but I'm not sure). I would like to be convinced that an index (which I think would be a simpler solution) is insufficient.

Does the trade manager have any unit test coverage? If not, or if it's sparse, it should probably be beefed up. At least the new functions should have unit tests.

chia/wallet/trading/trade_store.py
trepca: given one trade record can point to many coins it would require an `array` like data type for a proper index to be applied. Jsonb data type is an option, but was introduced this year which might affect our backwards compatibility.

chia/wallet/trading/trade_store.py
```diff
@@ -1,18 +1,49 @@
...
+    log.info("Beginning migration of coin_of_interest_to_trade_record lookup table")
+
+    start_time = perf_counter()
+    rows = await db.execute_fetchall("SELECT trade_record, trade_id from trade_records")
+
+    inserts: List[Tuple[bytes32, bytes, str]] = []
+    for row in rows:
+        record: TradeRecord = TradeRecord.from_bytes(row[0])
+        for coin in record.coins_of_interest:
+            inserts.append((coin.name(), coin.to_bytes(), record.trade_id.hex()))
```
arvidn, line 27: you still store `trade_id` as hex, instead of binary.
I think you meant to remove the `coin.to_bytes()` here too, right?

chia/wallet/trading/trade_store.py
```diff
@@ -1,18 +1,49 @@
...
+
+    if not inserts:
+        # no trades to migrate
+        return
+    try:
+        await db.executemany(
+            "INSERT INTO coin_of_interest_to_trade_record " "(coin_id, trade_id) " "VALUES(?, ?, ?)", inserts
```
arvidn, line 34: these strings should probably be joined
the table only has two columns now, but you insert 3 values

chia/wallet/trading/trade_store.py
```diff
@@ -1,18 +1,49 @@
...
 from chia.wallet.trading.trade_status import TradeStatus
 
 
+async def migrate_coin_of_interest(log: logging.Logger, db: aiosqlite.Connection) -> None:
+    log.info("Beginning migration of coin_of_interest_to_trade_record lookup table")
+
+    start_time = perf_counter()
+    rows = await db.execute_fetchall("SELECT trade_record, trade_id from trade_records")
+
+    inserts: List[Tuple[bytes32, bytes, str]] = []
```
arvidn, line 23: should be: `List[Tuple[bytes32, bytes32]]`

chia/wallet/trading/trade_store.py
```diff
@@ -115,6 +165,15 @@ async def add_trade_record(self, record: TradeRecord) -> None:
                 ),
             )
             await cursor.close()
+            # remove all current coin ids
+            await conn.execute("DELETE FROM coin_of_interest_to_trade_record WHERE trade_id=?", (record.trade_id,))
```
arvidn, line 169: here, you don't turn `trade_id` into hex (which is correct). But I would expect there to be a test catching this error, since you added the entry as hex in the migration function. I think you should add one that ensures entries are correctly removed.

Perhaps it's the migration function that doesn't have test coverage.

chia/wallet/trading/trade_store.py
```diff
@@ -236,6 +295,27 @@ async def get_trade_record_with_status(self, status: TradeStatus) -> List[TradeR
...
+        Checks DB for TradeRecord with id: id and returns it.
+        """
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            rows = await conn.execute_fetchall(
+                "SELECT cr.coin_parent, cr.puzzle_hash, cr.amount "
+                "from coin_record cr, coin_of_interest_to_trade_record cl, trade_records t "
+                "WHERE "
+                "t.status in (%s) "
+                "AND LOWER(hex(cl.trade_id)) = t.trade_id "
+                "AND cr.coin_name = LOWER(hex(cl.coin_id))" % (",".join("?" * len(trade_statuses)),),
```
arvidn, line 309: nice with the hex conversion in sqlite

arvidn (changes requested):
the migration function doesn't look right, and doesn't seem to have a test

chia/wallet/trading/trade_store.py
trepca: true, migration function was off. I added migration and trade record replacement tests.

chia/wallet/trading/trade_store.py
trepca: fixed

chia/wallet/trading/trade_store.py
trepca: looks like my IDE has a reformatting bug, fixed

chia/wallet/trading/trade_store.py
trepca: yep, fixed

chia/wallet/trading/trade_store.py
trepca: fixed

chia/wallet/trading/trade_store.py
```diff
@@ -236,6 +295,27 @@ async def get_trade_record_with_status(self, status: TradeStatus) -> List[TradeR
 
         return records
 
+    async def get_coins_of_interest_with_trade_statuses(self, trade_statuses: List[TradeStatus]) -> List[Coin]:
```
arvidn, line 298: I think this function should return a list of coin_ids instead. It would make the sql query simpler and it would save the caller from computing the name

chia/wallet/trade_manager.py
```diff
@@ -101,19 +101,12 @@ async def get_coins_of_interest(
...
+            trade_statuses=[TradeStatus.PENDING_ACCEPT, TradeStatus.PENDING_CONFIRM, TradeStatus.PENDING_CANCEL]
+        )
         interested_dict = {}
-
-        for trade in all_pending:
-            for coin in trade.coins_of_interest:
-                interested_dict[coin.name()] = coin
-
+        for coin in coins:
+            interested_dict[coin.name()] = coin
```
arvidn, line 109: the only thing we do with the coins is to call `name()`. we could avoid a DB join and this call

trepca: but `coin` is still needed as value to the dict? or are you saying doing another select here to fetch coins from coin_ids?

arvidn: oops. I missed that :)

arvidn: actually.. as far as I can tell, this function could return a set of coin_ids. there are only two callers of it, and both of them only care about the coin_id (afaict)

arvidn (dismissed):

arvidn (approved):

mariano54 (approved):

PR #13046, opened 2022-08-18
instrumentation of asyncio tasks

arvidn:
This adds a developer tool that measures where tasks spend their wall-clock time. This can help pin-point delays that are not caused by needing to use a lot of CPU.

Here's an example output from syncing the wallet:
Task-33 dot

### DIFF
--- a/chia/server/start_wallet.py
+++ b/chia/server/start_wallet.py
@@ -1,4 +1,5 @@
 import pathlib
+import os
 from multiprocessing import freeze_support
 import sys
 from typing import Dict, Optional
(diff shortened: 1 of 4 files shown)


chia/server/start_wallet.py
```diff
@@ -112,6 +114,8 @@ async def async_main() -> int:
 
 def main() -> int:
     freeze_support()
+    start_task_instrumentation()
+    atexit.register(stop_task_instrumentation)
```
arvidn, line 118: this needs to be done conditionally, like with a command like flag. However, the wallet is started by the daemon, and it's not trivial to pass on that here. Any help would be appreciated

richardkiss: A couple options I can think of:

- environment variables. I believe the daemon doesn't sanitize this, so stop the daemon, then do `CHIA_WALLET_INSTRUMENT=1 chia start wallet`
- magic file. Look for presence of `chia-wallet-instrument-magic` in the current directory. But I guess this presupposes that the `PWD` environment variable is passed through, so if this works, the first way would work

richardkiss: Stop the daemon with `chia stop -d all`

mariano54: Config option?

arvidn: It has to happen before we start the asyncio run loop, and we load the config after that. I liked the idea of using an environment variable

mariano54 (approved):

PR #7249, opened 2021-07-08
Keyring passphrase protection

paninaro:
This PR includes changes that define a new encrypted keyring format to be used to store keys, and which is optionally encrypted to a user-supplied passphrase. In addition, when multiple Chia services are running, keyring operations are now delegated to the daemon so that only a single entity needs to be in possession of the user's passphrase.

NOTE: Usage of the new keyring (and passphrase support) is disabled until GUI work is finalized.

Background:
Linux key storage has been using the cryptfile keyring backend with a hard-coded key, without the option for the user to change the passphrase. In discussions with other Chia devs, it was suggested that file synchronization issues with the Linux keyring have led to data corruption, and that a new keyring implementation should be pursued. Additionally, we should centralize keyring access in the daemon so that only a single passphrase entry is required and eliminate the need for each Chia service to directly access the keyring.

New keyring format:
The new keyring file is stored in ~/.chia_keys/keyring.yaml. The keyring is stored outside of ~/.chia to avoid accidental deletion when testing and removing the ~/.chia directory. The contents include an unencrypted outer-payload, as well as an encrypted inner-payload in which the keys are secured. The contents of the inner-payload are encrypted and authenticated using ChaCha20Poly1305, with a PBKDF2 key derived from the user's passphrase. When a user-provided passphrase is not set, the keys are still encrypted to a default passphrase, thereby avoiding the need for the user to interactively unlock the keyring (unattended unlocking is also available via the --passphrase-file CLI option)

CLI additions:
The 'chia' executable has two new global options:

    $ chia -keys-root-path PATH # Overrides the default keyring directory ~/.chia_keys

    $ chia --passphrase-file FILENAME # Reads the passphrase from a file or file descriptor

    Example: $ chia --passphrase-file <(echo 'my super secure passphrase') init --set-passphrase

The 'chia init' command has a new --set-passphrase option:

    $ chia init --set-passphrase # As part of bootstrapping, a keyring is created and immediately secured

A new top-level 'passphrase' command has been added that supports setting/updating and removing a passphrase from the keyring:

    $ chia passphrase remove # Removes the user-supplied passphrase, the default passphrase will be used instead (no interaction required)

    $ chia passphrase set # Sets or updates the keyring passphrase

If a passphrase has been set, operations that require access to the keyring will interactively prompt for passphrase entry (unless using an option that reads the passphrase from a file, like --passphrase-file). Once the passphrase has been entered, it will be cached in memory to avoid further prompts.

Migration:
[...]

PR #13052, opened 2022-08-18
daemon: Recover `sys.stdout.close()` call

xdustinface:
Fixing the following output on macOS when starting/stopping the daemon on CLI:

```
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
BrokenPipeError: [Errno 32] Broken pipe
```

This was commented out here https://github.com/Chia-Network/chia-blockchain/pull/7249/files#diff-b8e70126694c0407702f1d792f6b3845e9cb1e6b8a78f9387a368b3d9e8c9900R1164 (db87ff31372c6f13228175ea39e932ad4970f676 this commit), let me know if this was for some reason @paninaro.

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -1363,7 +1363,7 @@ async def async_run_daemon(root_path: Path, wait_for_unlock: bool = False) -> in
             await ws_server.start()
             await shutdown_event.wait()
             log.info("Daemon WebSocketServer closed")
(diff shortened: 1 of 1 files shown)


paninaro (approved):
I vaguely remember trying to suppress that output. Let's give it a try... lgtm

PR #13079, opened 2022-08-19
atari - cancel offer rpc

altendky:
Draft for:
- [x] Offer cancellation fails to consistently block taking the offer in the tests.
- [x] Handle removal of pending root in case of off-chain/not-secure cancellation


fchirica (dismissed):

chia/data_layer/data_layer_util.py
```diff
@@ -574,3 +574,42 @@ def marshal(self) -> Dict[str, Any]:
             "error": self.error,
             "fee": None if self.fee is None else int(self.fee),
         }
+
+
+@dataclasses.dataclass(frozen=True)
+class CancelOfferRequest:
+    trade_id: bytes32
+    # cancel on chain (secure) vs. just locally
+    secure: bool
```
fchirica, line 583: if `secure=False`, we'll need to delete our pending root locally.

altendky: agreed, thanks, will address.

chia/rpc/data_layer_rpc_util.py
```diff
@@ -45,11 +45,16 @@ def decorator(route: UnboundMarshalledRoute) -> UnboundRoute:
         request_class: Type[MarshallableProtocol] = hints["request"]
 
         async def wrapper(self: object, request: Dict[str, object]) -> Dict[str, object]:
+            # import json
```
fchirica, line 48: do we want to keep these comments? (and `print(f"\n ==== {name} response.json\n{json.dumps(marshalled_response, indent=4)}"`) below)

altendky: It's naughty, you can make me change this.  I used them for dumping the reference request and response values.  https://gist.github.com/altendky/4b57b25d711f4b33035857ede6388a5b  I think it should become a formal feature in general so we can have tests that validate the request and response and generate easily usable reference.

But...  commented out code is...  dirty.  If you want it removed, I will.

fchirica (approved):

PR #13081, opened 2022-08-20
atari - transactions for offers

altendky:
Draft for:
- [x] The new `DataLayer._update_confirmation_status()` call needs consideration and maybe another PR to add that to any other missing places.  It's not clear why adding the locking exposed this.


chia/data_layer/data_layer.py
```diff
@@ -236,6 +241,8 @@ async def get_root(self, store_id: bytes32) -> Optional[SingletonRecord]:
         return latest
 
     async def get_local_root(self, store_id: bytes32) -> Optional[bytes32]:
+        async with self.lock:
+            await self._update_confirmation_status(tree_id=store_id)
```
altendky, line 245: @fchirica, for some reason, adding the locking made a test fail until I added this.  I'm not clear why, but should we review all the functions to see which should have this?  Or do you think there's some other issue.  The test was `tests/core/data_layer/test_data_rpc.py::test_create_insert_get`.

altendky: Addressed in https://github.com/Chia-Network/chia-blockchain/pull/13083.

fchirica (approved):

PR #13094, opened 2022-08-22
enable wallet to read memos from incoming txs

trepca:


mariano54 (commented):
This might slow down sync significantly, since request_puzzle_solution is a very slow method in the full node. We might want to run a toilet wallet sync before and after and compare (trusted sync).

trepca:
> This might slow down sync significantly, since request_puzzle_solution is a very slow method in the full node. We might want to run a toilet wallet sync before and after and compare (trusted sync).

good point, will test. I also added caching for `request_puzzle_solution` on wallet_node as noticed a few duplicated calls for it.

PR #13098, opened 2022-08-22
Only try and install aws cli if its not already present, since this i…

cmmarslender:
…s really slow on 10.14


wallentx (approved):
:shipit:

PR #13115, opened 2022-08-23
Add `farmer-no-node` and `farmer-no-node-no-wallet` start modes

xearl4:
This add two new start modes, which are handy for chia-docker use:

- `farmer-no-node`: starts farmer + harvester + wallet
- `farmer-no-node-no-wallet`: starts farmer + harvester

This supplements `farmer` (farmer + harvester + node + wallet) and `farmer-only` (farmer) modes.


xdustinface:
Hm probably ok to add this but is it really required to add more combinations here since you can also just call it with combinations like `chia start farmer-only harvester` instead of `chia start farmer-no-node-no-wallet`?

xchdata1:
Good point, forgot about that possibility. I'll check if that passes through chia-docker correctly (and fix that there, if it does not).

PR #13116, opened 2022-08-23
remove DBWrapper

arvidn:
now that all uses of DBWrapper have been updated to DBWrapper2


trepca (approved):

arvidn:
I'm reverting this since it's still used by the atari branch

PR #13119, opened 2022-08-23
Fix pending root when changelist is empty.

fchirica:
This change intends pending root checking to work even if we don't explicitly check it in python code.

For the case when there's at least one operation and one pending root, the operation will fail as `(generation, tree_id)` root primary key will be duplicate.

However, when there is a pending root, but the changelist is empty, we'll get to call `rollback_to_generation`, which will delete our pending root. Then, the 2nd pending root will override the first one. This change throws before calling `rollback_to_generation`.

Even if checking for pending root is possible after this change without the python code, I left it in the code as it is more explicit than throwing `sqlite3.IntegrityError: UNIQUE constraint failed: root.tree_id, root.generation`


altendky:
Had a direct discussion about this and decided that 1) this is a good catch and we should respond, 2) i think that a single exception would be better than two, and 3) this risk is presently caught by an existing check so this can wait until after release.

emlowe:
Should this be retargeted to main?

chia/data_layer/data_store.py
```diff
@@ -981,6 +981,8 @@ async def insert_batch(
         async with self.db_wrapper.locked_transaction(lock=lock):
             hint_keys_values = await self.get_keys_values_dict(tree_id, lock=False)
             old_root = await self.get_tree_root(tree_id, lock=False)
+            if len(changelist) == 0:
```
almogdepaz, line 984: i would move this check and error raise to before we acquire the lock

altendky (approved):

emlowe (approved):

PR #13120, opened 2022-08-23
atari - include wallet maker offer rpc call in transaction rollback region

altendky:


chia/data_layer/data_layer.py
```diff
@@ -663,34 +663,34 @@ async def make_offer(
...
-        if wallet_offer is None:
-            raise Exception("offer is None despite validate_only=False")
-
-        offer = Offer(
-            trade_id=trade_record.trade_id,
-            offer=bytes(wallet_offer),
-            taker=taker,
-            maker=tuple(our_store_proofs.values()),
-        )
+            # Excluding wallet from transaction since failures in the wallet may occur
```
emlowe, line 666: I assume comment doesn't apply anything

altendky: Quite so.  Thanks for catching the details...

fchirica (approved):

PR #13131, opened 2022-08-23
Adding release/pre-release detection

wallentx:
Swap the tag creation trigger for `release.published` trigger, so that we can leverage conditionals based on `release` vs `pre-release`


.github/workflows/build-linux-arm64-installer.yml
```diff
@@ -101,12 +102,22 @@ jobs:
...
     - name: Get latest bladebit plotter
+      if: '!github.event.release.prerelease'
       run: |
         mkdir "$GITHUB_WORKSPACE/bladebit"
         wget -O /tmp/bladebit.tar.gz https://github.com/Chia-Network/bladebit/releases/download/${{ steps.latest-bladebit.outputs.result }}/bladebit-${{ steps.latest-bladebit.outputs.result }}-ubuntu-arm64.tar.gz
         tar -xvzf /tmp/bladebit.tar.gz -C $GITHUB_WORKSPACE/bladebit
         chmod +x "$GITHUB_WORKSPACE/bladebit/bladebit"
 
+    - name: Get latest prerelease bladebit plotter
+      if: github.event.release.published && '!github.event.release.prerelease'
```
cmmarslender, line 113:
```suggestion
      if: github.event.release.published && 'github.event.release.prerelease'
```

I think for prereleases we'd want the prerelease bladebit (I think this logic might be backwards?)

.github/workflows/upload-pypi-source.yml
```diff
@@ -6,8 +6,8 @@ on:
       - 'long_lived/**'
       - main
       - 'release/**'
-    tags:
-        - '**'
+  release:
```
cmmarslender, line 9: For the last step in here (when we publish to pypi) I'm wondering if theres a way we can mark these as prerelease there when its a pre-release

wallentx: https://github.com/pypa/gh-action-pypi-publish
Just looking at this, it doesn't say anything specific, but if it handles tags, we could just maybe enforce a tag standard syntax for pre-release?

wallentx: Actually, this may answer it:
https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#pre-release-versioning
What I gather from that is that it is considered a pre-release based on the contents of the tag.
https://pypi.org/project/chia-blockchain/#history
It looks like it even already has some pre-releases flagged. I think this should be good to go, so long as our tagging format for pre-releases follows one of the conventions that pypi sets.

PR #13133, opened 2022-08-24
Fix several bugs with untrusted sync, and correct sync status

mariano54:
This should fix missing wallets and incorrect balance when local node is not synced, and wallet is connected to untrusted nodes.

## Fix 1
Scenario:
Global Blockchain is at 2.4 million height.
Local node is partially synced to 1.9 million. The wallet connects to the local node, and a few other untrusted nodes. The wallet subscribes to standard wallet puzzle hashes. The local node responds with some of the coins (only standard XCH at this point) up to it's synced height. The wallet DOES receive these states, even though local node isn't synced. Note that we are not checking for created CATs, NFT wallets, etc, since we aren't fully syncing, just responding to some PH subscriptions.
The wallet mistakenly sets `finished_sync_up_to`. (see the diff of this PR).
The wallet doesn't sync to the local node since it's not synced. But it starts syncing to another node.
Since the height is at 1.9 million, it will only process things that happened after that height. Since the CATs and NFT wallets were created at an earlier height, these are completely skipped. 
The height will be set to 2.4 million now.
secondary sync might find the missing items though.

## Fix 2
Even if the secondary sync finds the new stuff, there is another issue. The syncing full node will be sending state updates to the wallet whenever it sees a new TX. These state updates can trigger a rollback of DB to the fork point (which is the outdated local node's height) meaning all new stuff is reverted. 

## Fix 3
After rolling back, why didn't the new peaks from the untrusted peers trigger another long sync? The reason is that there are two peak heights in the wallet: `peak_height` and `finished_sync_up_to`. When we rollback, we only rollback the latter. however when we receive a new peak, we use the `peak_height`, which is not correct... I changed that in this PR. However this needs to be tested, as I'm not 100% sure all edge cases are covered.

## Fix 4
Don't show wallet synced until the wallet is actually synced. This means processing all pending subscriptions. 

## Fix 5
Change some constants to reduce issues. Only do a long sync if we are 300 blocks behind. Consider wallet "not synced" unless the latest peak is 5 old at most.

Going forward we need to change wallet sync to be reliable and do things in an established order, with correct behavior when things go wrong. Sync is way too complex, as can be seen by weird bugs like this.

TODO: add tests

### DIFF
--- a/chia/wallet/util/new_peak_queue.py
+++ b/chia/wallet/util/new_peak_queue.py
@@ -53,18 +53,28 @@ class NewPeakItem:
 class NewPeakQueue:
     def __init__(self, inner_queue: asyncio.PriorityQueue):
         self._inner_queue: asyncio.PriorityQueue = inner_queue
+        self._pending_data_process_items: int = 0
 
     async def subscribe_to_coin_ids(self, coin_ids: List[bytes32]):
+        self._pending_data_process_items += 1
         await self._inner_queue.put(NewPeakItem(NewPeakQueueTypes.COIN_ID_SUBSCRIPTION, coin_ids))
 
     async def subscribe_to_puzzle_hashes(self, puzzle_hashes: List[bytes32]):
+        self._pending_data_process_items += 1
         await self._inner_queue.put(NewPeakItem(NewPeakQueueTypes.PUZZLE_HASH_SUBSCRIPTION, puzzle_hashes))
 
     async def full_node_state_updated(self, coin_state_update: CoinStateUpdate, peer: WSChiaConnection):
+        self._pending_data_process_items += 1
         await self._inner_queue.put(NewPeakItem(NewPeakQueueTypes.FULL_NODE_STATE_UPDATED, (coin_state_update, peer)))
 
     async def new_peak_wallet(self, new_peak: NewPeakWallet, peer: WSChiaConnection):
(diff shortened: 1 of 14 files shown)


chia/wallet/util/new_peak_queue.py
```diff
@@ -53,18 +53,28 @@ def __ge__(self, other):
...
         await self._inner_queue.put(NewPeakItem(NewPeakQueueTypes.NEW_PEAK_WALLET, (new_peak, peer)))
 
     async def get(self) -> NewPeakItem:
-        return await self._inner_queue.get()
+        item: NewPeakItem = await self._inner_queue.get()
+        if item.item_type != NewPeakQueueTypes.NEW_PEAK_WALLET:
+            self._pending_data_process_items -= 1
+        return item
+
+    def has_pending_data_process_items(self) -> bool:
```
trepca, line 79: is it possible to never receive an update to one of those subscriptions? like maybe a peer is overwhelmed etc? how would the wallet recover from something like that since counter would never get to zero?

mariano54: this is a local counter for local queue items. We pull an item from the queue (decreasing the counter), and then we ask the full node. If the full node does not respond, that's it, we forget about it. The item was still pulled from the queue. Does that make sense?

trepca: ah, yes 🤦‍♂️

trepca (approved):
💯

PR #13142, opened 2022-08-24
 lookup the coin records in bulk

arvidn:
When adding coins to the wallet. This shaves off about 9% from sync-time the toilet wallet:

| --- | --- | --- | --- |
| wallet sync | 89.77s | 81.85s | 91.2% |

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -876,12 +876,16 @@ class WalletStateManager:
         used_up_to = -1
         ph_to_index_cache: LRUCache = LRUCache(100)
 
-        for coin_state in coin_states:
+        local_records: List[Optional[WalletCoinRecord]] = await self.coin_store.get_coin_records(
+            [st.coin.name() for st in coin_states]
+        )
+
+        assert len(local_records) == len(coin_states)
+        for coin_state, local_record in zip(coin_states, local_records):
             coin_name: bytes32 = coin_state.coin.name()
             wallet_info: Optional[Tuple[uint32, WalletType]] = await self.get_wallet_id_for_puzzle_hash(
                 coin_state.coin.puzzle_hash
             )
-            local_record: Optional[WalletCoinRecord] = await self.coin_store.get_coin_record(coin_name)
             self.log.debug("%s: %s", coin_name, coin_state)
 
             # If we already have this coin, and it was spent and confirmed at the same heights, then we return (done)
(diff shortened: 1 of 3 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -876,12 +876,16 @@ async def new_coin_state(
         used_up_to = -1
         ph_to_index_cache: LRUCache = LRUCache(100)
 
-        for coin_state in coin_states:
+        local_records: List[Optional[WalletCoinRecord]] = await self.coin_store.get_coin_records(
+            [st.coin.name() for st in coin_states]
```
arvidn, line 880: this makes us compute the coin name twice. It's still a net improvement though, and there's no simple way of avoiding it. The `WalletCoinRecord` does not contain a name field.

chia/wallet/wallet_state_manager.py
```diff
@@ -876,12 +876,16 @@ async def new_coin_state(
...
+        )
+
+        assert len(local_records) == len(coin_states)
+        for coin_state, local_record in zip(coin_states, local_records):
             coin_name: bytes32 = coin_state.coin.name()
             wallet_info: Optional[Tuple[uint32, WalletType]] = await self.get_wallet_id_for_puzzle_hash(
                 coin_state.coin.puzzle_hash
             )
-            local_record: Optional[WalletCoinRecord] = await self.coin_store.get_coin_record(coin_name)
```
arvidn, line 884: this is the lookup we're saving

chia/wallet/wallet_state_manager.py
```diff
@@ -967,10 +971,10 @@ async def new_coin_state(
                     )
                     await self.coin_store.add_coin_record(record)
                     # Coin first received
-                    coin_record: Optional[WalletCoinRecord] = await self.coin_store.get_coin_record(
+                    parent_coin_record: Optional[WalletCoinRecord] = await self.coin_store.get_coin_record(
                         coin_state.coin.parent_coin_info
                     )
-                    if coin_record is not None and wallet_type.value == coin_record.wallet_type:
+                    if parent_coin_record is not None and wallet_type.value == parent_coin_record.wallet_type:
```
arvidn, line 977: this is an unrelated renaming. I think this is a better na,e

trepca (approved):

chia/wallet/wallet_state_manager.py
trepca: you could build a dict lookup for names, but probably marginal

PR #13146, opened 2022-08-24
Fix comments in initial config where puzzle hash should be receive address

hugepants:
These instructions should be suggesting an address rather than a puzzle hash.


paninaro (approved):
lgtm

PR #13161, opened 2022-08-25
Add atari CLI help.

fchirica:


chia/cmds/data.py
```diff
@@ -263,10 +272,20 @@ def get_root_history(
...
+@data_cmd.command("add_missing_files", short_help="Manually reconstruct server files from the data layer database")
+@click.option(
+    "-i",
+    "--ids",
+    help="List of stores to reconstruct. If not specified, all stores will be reconstructed",
+    type=str,
+    required=False,
+)
+@click.option(
+    "-o/-n", "--override/--no-override", help="Specify if already existing files need to be overridden by this command"
```
altendky, line 284: 'replace' might be a generally better word to consider here in the future?

```suggestion
    "-o/-n", "--override/--no-override", help="Specify if already existing files need to be overwritten by this command"
```

chia/cmds/data.py
```diff
@@ -263,10 +272,20 @@ def get_root_history(
...
 
-@data_cmd.command("add_missing_files", short_help="")
-@click.option("-i", "--ids", help="", type=str, required=False)
-@click.option("-o/-n", "--override/--no-override")
-@click.option("-f", "--foldername", type=str, required=False)
+@data_cmd.command("add_missing_files", short_help="Manually reconstruct server files from the data layer database")
+@click.option(
+    "-i",
+    "--ids",
+    help="List of stores to reconstruct. If not specified, all stores will be reconstructed",
```
altendky, line 279: A point for the future to make this not be JSON.

chia/cmds/data.py
```diff
@@ -281,10 +300,10 @@ def add_missing_files(ids: Optional[str], override: bool, foldername: Optional[s
...
 
 
-@data_cmd.command("add_mirror", short_help="")
-@click.option("-i", "--id", help="", type=str, required=True)
-@click.option("-a", "--amount", help="", type=int, required=True)
-@click.option("-u", "--url", "urls", help="", type=str, multiple=True)
+@data_cmd.command("add_mirror", short_help="Publish mirror urls on chain")
+@click.option("-i", "--id", help="Store id", type=str, required=True)
+@click.option("-a", "--amount", help="Amount for this mirror", type=int, required=True)
+@click.option("-u", "--url", "urls", help="List of urls published for the coin", type=str, multiple=True)
```
altendky, line 306: Especially given that we have cases where you do actually provide a JSON array (list), I wouldn't use list here.  Maybe this...?

```suggestion
@click.option("-u", "--url", "urls", help="URL to publish on the new coin, multiple accepted and will be published to a single coin.", type=str, multiple=True)
```

chia/cmds/data.py
```diff
@@ -301,8 +320,8 @@ def add_mirror(id: str, amount: int, urls: List[str], fee: Optional[str], data_r
     )
 
 
-@data_cmd.command("delete_mirror", short_help="")
-@click.option("-i", "--id", help="", type=str, required=True)
+@data_cmd.command("delete_mirror", short_help="Delete an owned mirror by its coin id")
+@click.option("-i", "--id", help="Store id", type=str, required=True)
```
altendky, line 324: Isn't this the mirror coin id?

```suggestion
@click.option("-i", "--id", help="Coin id", type=str, required=True)
```

altendky (changes requested):
Thanks for filling these in.  It's definitely nice to have entries like this in `--help`.

chia/cmds/data.py
fchirica: agreed, using `multiple=True` as everything else would be nice

altendky (approved):

PR #13163, opened 2022-08-25
fix test flakes

almogdepaz:
assert spend is in mempool before farming blocks

reduce farmed blocks in tests

reduce timeouts

make tests consistent

### DIFF
--- a/tests/wallet/rpc/test_wallet_rpc.py
+++ b/tests/wallet/rpc/test_wallet_rpc.py
@@ -1,3 +1,4 @@
+import asyncio
 import dataclasses
 import json
 import logging
@@ -71,6 +72,9 @@ async def farm_transaction_block(full_node_api: FullNodeSimulator, wallet_node: 
     await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(bytes32(b"\00" * 32)))
     await time_out_assert(20, wallet_is_synced, True, wallet_node, full_node_api)
 
+async def check_spends_in_mempool(full_node_api: FullNodeSimulator, num_of_spends = 1):
+    await asyncio.sleep(0.1)
+    return len(full_node_api.full_node.mempool_manager.mempool.sorted_spends) == num_of_spends
 
 async def farm_transaction(full_node_api: FullNodeSimulator, wallet_node: WalletNode, spend_bundle: SpendBundle):
     await time_out_assert(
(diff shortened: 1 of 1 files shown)


altendky (commented):
The reduced timeouts concern me in terms of CI runners liking to just randomly stop running our stuff for awhile, at least on macOS in my previous experiences, but otherwise I'm good with this.  I can give it a big rerun blast in another PR to see if my concerns are well founded or not.  If not I'm happy to merge this into the `coverage` branch.

altendky:
Also...  thanks!

almogdepaz:
> The reduced timeouts concern me in terms of CI runners liking to just randomly stop running our stuff for awhile, at least on macOS in my previous experiences, but otherwise I'm good with this. I can give it a big rerun blast in another PR to see if my concerns are well founded or not. If not I'm happy to merge this into the `coverage` branch.

i dont mind changing back but i feel like this is just being to defensive i think the timeouts here are just a result of hiding the real issue/copy paste from other tests

there is no reason for 20 sec, can we run this like 100 times and see? i hate giving long timeouts it just hides stuff

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -571,12 +575,17 @@ async def test_cat_endpoints(wallet_rpc_environment: WalletRpcTestEnvironment):
     assert should_be_none is None
     assert name == next(iter(DEFAULT_CATS.items()))[1]["name"]
 
-    # TODO: Investigate why farming only one block here makes it flaky
-    for _ in range(5):
+    # make sure spend is in mempool before farming tx block
+    await time_out_assert(5, check_mempool_spend_count, True, full_node_api, 1)
+    for i in range(5):
+        if check_mempool_spend_count(full_node_api, 0):
```
almogdepaz, line 581: this is still a hack, im not sure yet why we sometimes farm empty blocks and not include the spend even though it is in the mempool

but this is a better hack then what we had

altendky:
Testing robustness in https://github.com/Chia-Network/chia-blockchain/pull/13217.  That is running only the `wallet.rpc` job and creating 60x duplicates.  That hits 240 jobs in the linux and windows workflows where 255 (or 256) is the limit per matrix.

If that works out, ok.  I both agree with disliking long timeouts and also have spent too much time dealing with macOS runners out of the blue doing nothing at all for 30 seconds at a time.  Anyways, we'll see what the run above says.

tests/wallet/rpc/test_wallet_rpc.py
altendky: Thanks for digging into all of this.

PR #13171, opened 2022-08-25
Add some anti spam measures for DL mirrors

Quexington:
Open to other measures but these seem like some common sense options.


almogdepaz:
some mechanism to blacklist unresponsive mirrors ?

PR #13181, opened 2022-08-26
Cherry-pick sync fixes from main

wallentx:
This adds:
https://github.com/Chia-Network/chia-blockchain/commit/99be9be871f149728a964aafe0c55231001dbd31
https://github.com/Chia-Network/chia-blockchain/commit/dc5c44accd2d5f36b942791d48e376641ee406eb
https://github.com/Chia-Network/chia-blockchain/commit/805e18fe0eb25b4fd503b7918a0ceda63bde95bb
https://github.com/Chia-Network/chia-blockchain/commit/2d551dd21b6710b9677aee764716bb6899ec3d82

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -283,8 +283,6 @@ class Blockchain(BlockchainInterface):
                         fetched_block_record.header_hash,
                         fetched_block_record.sub_epoch_summary_included,
                     )
-                if state_change_summary is not None:
-                    self._peak_height = block_record.height
             except BaseException as e:
                 self.block_store.rollback_cache_block(header_hash)
                 log.error(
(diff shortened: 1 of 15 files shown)


wallentx:
I need to grab an additional cherrypick from somewhere, because tests are failing due to undefined asyncio

altendky (changes requested):
I would like to see cherry picks done individually unless maybe if they are tightly intertwined.  And even then I'm not entirely sure.  Like if you need commit X but it had a bug that was fixed in commit Y then _maybe_ they go in one PR.  Or they just go in two sequenced one after the other...

wallentx:
> I would like to see cherry picks done individually unless maybe if they are tightly intertwined. And even then I'm not entirely sure. Like if you need commit X but it had a bug that was fixed in commit Y then _maybe_ they go in one PR. Or they just go in two sequenced one after the other...

For this, these are actually closely intertwined. I found the other missing commit, which should resolve the asyncio undefined issue. This is also related to https://github.com/Chia-Network/chia-blockchain/pull/13200 (should be targeting release/1.6.0 soon), which is the last(?) missing part that all these commits have attempted to solve.

PR #13190, opened 2022-08-26
Puzzle driver type annotations

arvidn:
for bite sized reviews, each commit can be reviewed independently

### DIFF
--- a/chia/wallet/cat_wallet/cat_outer_puzzle.py
+++ b/chia/wallet/cat_wallet/cat_outer_puzzle.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Any, Callable, List, Optional
+from typing import Callable, List, Optional
 
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.program import Program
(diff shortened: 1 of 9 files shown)


chia/wallet/cat_wallet/cat_outer_puzzle.py
```diff
@@ -45,16 +44,18 @@ def get_inner_puzzle(self, constructor: PuzzleInfo, puzzle_reveal: Program) -> O
         if args is None:
             raise ValueError("This driver is not for the specified puzzle reveal")
         _, _, inner_puzzle = args
-        if constructor.also() is not None:
-            deep_inner_puzzle: Optional[Program] = self._get_inner_puzzle(constructor.also(), inner_puzzle)
+        also = constructor.also()
```
arvidn, line 47: the reason to name a variable for `also` is that mypy needs to know that we checked it against `None` in order to consider it a `PuzzleInfo` object. the method call `also()` always returns an `Optional[PuzzleInfo]` which isn't compatible with `_get_inner_puzzle()` (for example). It doesn't take an `Optional`

Quexington (approved):
The last commit will have to be undone if a future outer puzzle comes along that needs to make use of calculating other asset IDs but this will likely be refactored anyway before that happens, if ever.

trepca (approved):

jack60612 (commented):
sorry for the delayed review, looks good

PR #13192, opened 2022-08-26
catchup: into release/1.6.0 from long_lived/atari  @ 72491c5cdce1a15c3dbd1fc21cfd52e38f8d1894

altendky:
Source hash: 72491c5cdce1a15c3dbd1fc21cfd52e38f8d1894

Draft for:
- [x] GUI pinned to merge of https://github.com/Chia-Network/chia-blockchain-gui/pull/967
- [x] May want a different GUI pin.  If not we'll need to cherry pick other things into release first to deal with the installer.


emlowe (dismissed):
Approved

emlowe (approved):

chia/cmds/configure.py
```diff
@@ -92,6 +93,13 @@ def configure(
             config["full_node"]["target_peer_count"] = int(set_peer_count)
             print("Target peer count updated")
             change_made = True
+        if enable_data_server:
+            config["data_layer"]["run_server"] = str2bool(enable_data_server)
```
aqk, line 97: Are you guaranteed to have the `data_layer` yaml node at this point?

chia/cmds/configure.py
```diff
@@ -260,6 +270,12 @@ def configure(
     help="configures the seeder nameserver setting. Ex: `example.com.`",
     type=str,
 )
+@click.option(
+    "--enable-data-server",
+    "--data-server",
```
aqk, line 275: I see that `click.Choice` accepts any argument in the array.
`--enable-data-server false` is more clear to me than `--data-server false`
I would only keep one of them.

If you wanted to keep both, it might be good to have `--enable-data-server` and `--disable-data-server`

chia/cmds/data_funcs.py
```diff
@@ -0,0 +1,309 @@
...
+        exc: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> None:
+        if self._client is None:
+            return
+        self._client.close()
+        await self._client.await_closed()
+
+
+async def create_data_store_cmd(rpc_port: Optional[int], fee: Optional[str]) -> None:
```
aqk, line 47: It's probably okay to force the caller to give you `rpc_port` - you could remove `Optional` here in that case.

chia/cmds/data_funcs.py
```diff
@@ -0,0 +1,309 @@
...
+
+async def get_owned_stores_cmd(rpc_port: Optional[int]) -> None:
+    try:
+        async with get_client(rpc_port) as (client, rpc_port):
+            res = await client.get_owned_stores()
+            print(res)
+    except aiohttp.ClientConnectorError:
+        print(f"Connection error. Check if data is running at {rpc_port}")
+    except Exception as e:
+        print(f"Exception from 'data': {e}")
```
aqk, line 309: I really like how clear the functions in `data_funcs.py` are.

chia/cmds/data_funcs.py
```diff
@@ -0,0 +1,309 @@
...
+
+async def create_data_store_cmd(rpc_port: Optional[int], fee: Optional[str]) -> None:
+    final_fee = None
+    if fee is not None:
+        final_fee = uint64(int(Decimal(fee) * units["chia"]))
+    try:
+        async with get_client(rpc_port) as (client, rpc_port):
+            res = await client.create_data_store(fee=final_fee)
+            print(res)
+    except aiohttp.ClientConnectorError:
```
aqk, line 55: Error handling code would benefit from  de-duping.

chia/cmds/wallet_funcs.py
```diff
@@ -58,12 +58,12 @@ def print_transaction(tx: TransactionRecord, verbose: bool, name, address_prefix
...
 def get_mojo_per_unit(wallet_type: WalletType) -> int:
     mojo_per_unit: int
-    if wallet_type == WalletType.STANDARD_WALLET or wallet_type == WalletType.POOLING_WALLET:
+    if wallet_type in {WalletType.STANDARD_WALLET, WalletType.POOLING_WALLET, WalletType.DATA_LAYER}:
         mojo_per_unit = units["chia"]
     elif wallet_type == WalletType.CAT:
         mojo_per_unit = units["cat"]
     else:
-        raise LookupError("Only standard wallet, CAT wallets, and Plot NFTs are supported")
+        raise LookupError(f"Operation is not supported for Wallet type {wallet_type.name}")
```
aqk, line 66: Nice and generic.

chia/data_layer/data_layer_errors.py
```diff
@@ -0,0 +1,40 @@
+from typing import Iterable, List
+
+from chia.types.blockchain_format.sized_bytes import bytes32
+
+
+class IntegrityError(Exception):
```
aqk, line 6: Better log this. Not connected yet?

chia/data_layer/data_layer_server.py
```diff
@@ -0,0 +1,60 @@
...
+    root_path: Path
+    config: Dict[str, Any]
+    log: logging.Logger
+
+    async def start(self) -> None:
+        self.log.info("Starting Data Layer Server.")
+        self.port = self.config["host_port"]
+
+        # Setup UPnP for the data_layer_service port
+        self.upnp: UPnP = UPnP()
```
aqk, line 24: Does the UPnP library we are using allow separate instances of the object without causing conflicts?

chia/data_layer/data_layer_server.py
```diff
@@ -0,0 +1,60 @@
...
+        app.add_routes([web.get("/{filename}", self.file_handler)])
+        self.runner = web.AppRunner(app)
+        await self.runner.setup()
+        self.site = web.TCPSite(self.runner, self.config["host_ip"], port=self.port)
+        await self.site.start()
+        self.log.info("Started Data Layer Server.")
+
+    async def stop(self) -> None:
+        self.upnp.release(self.port)
+        # this is a blocking call, waiting for the UPnP thread to exit
```
aqk, line 42:
```suggestion
        # upnp.release() is a blocking call, waiting for the UPnP thread to exit
```

chia/data_layer/dl_wallet_store.py
```diff
@@ -0,0 +1,329 @@
...
+        uint32(row[7]),
+        uint64(row[8]),
+    )
+
+
+def _row_to_mirror(row: Row) -> Mirror:
+    urls: List[bytes] = []
+    byte_list: bytes = row[3]
+    while byte_list != b"":
+        length = uint16.from_bytes(byte_list[0:2])
```
aqk, line 34: Could use `struct.unpack` here

chia/data_layer/dl_wallet_store.py
```diff
@@ -0,0 +1,329 @@
...
+    @classmethod
+    async def create(cls: Type[_T_DataLayerStore], db_wrapper: DBWrapper2) -> _T_DataLayerStore:
+        self = cls()
+
+        self.db_wrapper = db_wrapper
+
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
+            await conn.execute(
+                (
+                    "CREATE TABLE IF NOT EXISTS singleton_records("
```
aqk, line 57: Using a version field in this schema might help later.

chia/data_layer/dl_wallet_store.py
```diff
@@ -0,0 +1,329 @@
...
+        self,
+        launcher_id: bytes32,
+        min_generation: Optional[uint32] = None,
+        max_generation: Optional[uint32] = None,
+        num_results: Optional[uint32] = None,
+    ) -> List[SingletonRecord]:
+        """
+        Returns stored singletons with a specific launcher ID.
+        """
+        query_params: List[Union[bytes32, uint32]] = [launcher_id]
```
aqk, line 129: Nice that all objects are the same size here. Much easier to debug.

chia/data_layer/download_data.py
```diff
@@ -0,0 +1,184 @@
...
+        node_hash = bytes32(bytes.fromhex(raw_node_hash))
+        generation = int(raw_generation)
+    except ValueError:
+        return False
+
+    if len(rest) > 0:
+        return False
+
+    # TODO: versions should probably be centrally defined
+    if raw_version != "v1.0.dat":
```
aqk, line 39: nice

chia/data_layer/download_data.py
```diff
@@ -0,0 +1,184 @@
...
+async def insert_into_data_store_from_file(
+    data_store: DataStore,
+    tree_id: bytes32,
+    root_hash: Optional[bytes32],
+    filename: Path,
+) -> None:
+    with open(filename, "rb") as reader:
+        while True:
+            chunk = b""
+            while len(chunk) < 4:
```
aqk, line 60: If we switch to `struct.unpack`, we could version the format strings to match our data store version.

chia/data_layer/download_data.py
```diff
@@ -0,0 +1,184 @@
...
+    data_store: DataStore,
+    tree_id: bytes32,
+    root: Root,
+    foldername: Path,
+    override: bool = False,
+) -> bool:
+    if root.node_hash is not None:
+        node_hash = root.node_hash
+    else:
+        node_hash = bytes32([0] * 32)  # todo change
```
aqk, line 97: Is there one global root, or is it per-singleton?

chia/data_layer/download_data.py
```diff
@@ -0,0 +1,184 @@
...
+
+    await data_store.insert_root_with_ancestor_table(tree_id=tree_id, node_hash=root_hash, status=Status.COMMITTED)
+
+
+async def write_files_for_root(
+    data_store: DataStore,
+    tree_id: bytes32,
+    root: Root,
+    foldername: Path,
+    override: bool = False,
```
aqk, line 92:
```suggestion
    overwrite: bool = False,
```

`override = True` is a lot bigger danger to their data; better to make this flag stand out.

chia/data_layer/download_data.py
```diff
@@ -0,0 +1,184 @@
...
+        except Exception:
+            await data_store.server_misses_file(tree_id, server_info, timestamp)
+            raise
+
+        log.info(f"Successfully downloaded delta file {filename}.")
+        try:
+            await insert_into_data_store_from_file(
+                data_store,
+                tree_id,
+                None if root_hash == bytes32([0] * 32) else root_hash,
```
aqk, line 159:
```suggestion
                None if root_hash == DEFAULT_ROOT_HASH else root_hash,
```

Should be a constant so you can be sure you catch them all when you replace.

chia/cmds/configure.py
altendky: Yeah, there's that hazard.  I'm on the fence about fixing it ad-hoc here.  It's both more of a hazard with data layer being new and also an existing hazard for everything here I think.  Anyways, https://github.com/Chia-Network/chia-blockchain/pull/13199.

chia/wallet/wallet_puzzle_store.py
```diff
@@ -39,11 +39,12 @@ async def create(cls, db_wrapper: DBWrapper2, cache_size: uint32 = uint32(600000
                     "CREATE TABLE IF NOT EXISTS derivation_paths("
                     "derivation_index int,"
                     " pubkey text,"
-                    " puzzle_hash text PRIMARY KEY,"
```
aqk, line 42: Did we ever release the table version with `puzzle_hash text PRIMARY KEY`?

If so, we'll have to do a DB migration for clients that already have the `derivation_paths` table.

aqk (approved):
I understand this is mainly a review to see if merging into main / release won't affect other code. ie. No merge errors. So, I'm not requesting changes for this merge.

Since this is one of the last merges before release, I wanted to take a closer look. Everything looks great, except for a change to the database schema in the `derivation_paths` table. Please take a look at my comment in chia/wallet/wallet_puzzle_store.py

We might need to tackle that in another PR if it is an issue.

chia/wallet/wallet_puzzle_store.py
altendky: That's what's on `main` for quite awhile and multiple releases, yes.  It was only discovered as a bug because data layer had another bug where it reused other wallet's puzzle hashes.  This has been discussed and deemed ok as is.  Nobody _needs_ the change.  When they rebuild their db they will get the change.  If we actually need this in the future we can do an official migration.

chia/cmds/data_funcs.py
altendky: The entire CLI system needs a big overhaul in this regard, absolutely.  Also, replace the callback-based helpers with context managers so they don't need to be aware of the code 'inside' them.

chia/cmds/data_funcs.py
altendky: I'm a bit tempted to agree that moving the config handling out of here and into the CLI might be overall sensible.  There's definitely a relationship between config, CLI options, and environment variables where they all mix and override each other.  A quick search seems to show many `config` references in both the CLI and 'func' files so I'm not immediately sure which is more normal presently.

Overall, I think it would be ok to leave this for an overhaul of the entire CLI.

chia/data_layer/data_layer_errors.py
altendky: I'm not quite sure what you mean.  This is a base class to group some of the other exceptions below in the exception hierarchy.

chia/data_layer/data_layer_server.py
altendky: I don't know.  While the data layer file propagation service 'structure' is in flux, it does not and will not exist in the full node process.  Is that what you were concerned about?  Or concerned about two instances even in separate processes on the same computer?

chia/data_layer/data_layer_server.py
altendky: Pretty sure this does refer to the shutdown call below it.  Clarified in https://github.com/Chia-Network/chia-blockchain/pull/13216.

chia/data_layer/download_data.py
altendky: More like per-singleton.  Though a singleton has history and there's a root for each generation in that history.

chia/cmds/configure.py
altendky: Or perhaps https://github.com/Chia-Network/chia-blockchain/pull/13220 instead.

chia/cmds/configure.py
fchirica: https://github.com/Chia-Network/chia-blockchain/pull/13222

chia/data_layer/download_data.py
fchirica: Made a PR to change `override` to `overwrite`: https://github.com/Chia-Network/chia-blockchain/pull/13224

chia/data_layer/data_layer_errors.py
aqk: Ah, I see

chia/data_layer/data_layer_server.py
aqk: I was concerned even in the separate process case, in the unlikely case that they both choose the same port. But otherwise, I think separate process handles all the other potential issues.

chia/data_layer/download_data.py
fchirica: use `struct.unpack` only for this line:`size = int.from_bytes(chunk, byteorder="big")`?

chia/data_layer/data_layer_server.py
altendky: I may not know enough about how UPnP works.  The processes pick their own ports as normal, which will be different for different services by design.  Separately UPnP connects to the router to try to request the port forwarding from service's port on the router to the service's port on the local system.  Is one of those ports what you were worried about colliding?  Or is there another port involved in UPnP.

chia/data_layer/dl_wallet_store.py
altendky: @Quexington, ping on a specific comment from Adam.

PR #13193, opened 2022-08-26
correct port reference in data layer test

altendky:
Draft for:
- [x] based on https://github.com/Chia-Network/chia-blockchain/pull/13192, merge after it is


altendky:
We don't have the data file propagation server setup so the data won't get across the wallets, but the root hash updates can still be monitored.

AmineKhaldi (approved):

emlowe (approved):

PR #13203, opened 2022-08-27
use rust parser for Program.from_bytes()

arvidn:
This significantly speeds up parsing of puzzles in the wallet. Especially with generalized NFTs with nested puzzles.

I consider this a stop-gap until we get @richardkiss new clvm parser that's compatible with the existing python API.

To quantify this gain:

| --- | --- | --- | --- |
| Offer.from_bytes() | 8.05s | 6.55s | 81.4% |

note how the `to_program()` frame disappears from the profile:

before:
offer-parsing-main

after:
offer-parsing-patched

### DIFF
--- a/chia/types/blockchain_format/program.py
+++ b/chia/types/blockchain_format/program.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import io
 from typing import Callable, Dict, List, Set, Tuple, Optional, Any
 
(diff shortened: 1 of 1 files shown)


chia/types/blockchain_format/program.py
```diff
@@ -32,11 +33,18 @@ def stream(self, f):
         sexp_to_stream(self, f)
 
     @classmethod
-    def from_bytes(cls, blob: bytes) -> "Program":
-        f = io.BytesIO(blob)
-        result = cls.parse(f)  # noqa
-        assert f.read() == b""
-        return result
+    def from_bytes(cls, blob: bytes) -> Program:
```
richardkiss, line 36:
```suggestion
    def from_bytes(cls, blob: bytes) -> "Program":
```

I thought we had to quote `Program` here because it's not in scope until the entire `Program` class is parsed.

richardkiss (approved):
What a glorious hack! I love it.

chia/types/blockchain_format/program.py
arvidn: that's true, unless you `from __future__ import annotations` at the top

trepca (approved):
💯

PR #13204, opened 2022-08-27
implement a more efficient Program.uncurry()

arvidn:
This significantly speeds up parsing generalized offers, especially with nested puzzles. We do a lot of uncurrying when matching puzzles to types of coins.

This also drops the dependency of clvm_tools.uncurry.

To quantify the speed-up:

| --- | --- | --- | --- |
| Offer.from_bytes() | 7.8s | 3.73s | 47.8% |
| uncurry CPU | 53.93% | 5.24% | 9.72% |

before:
offer-parsing-main

after:

offer-parsing-patched

### DIFF
--- a/chia/types/blockchain_format/program.py
+++ b/chia/types/blockchain_format/program.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import io
 from typing import Callable, Dict, List, Set, Tuple, Optional, Any
 
@@ -6,7 +7,6 @@ from clvm.casts import int_from_bytes
 from clvm.EvalError import EvalError
 from clvm.serialize import sexp_from_stream, sexp_to_stream
 from chia_rs import MEMPOOL_MODE, run_chia_program, serialized_length, run_generator, tree_hash
-from clvm_tools.curry import uncurry
 
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.hash import std_hash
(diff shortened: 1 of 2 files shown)


chia/types/blockchain_format/program.py
```diff
@@ -136,11 +136,34 @@ def curry(self, *args) -> "Program":
...
+            ev = next(i)
+            # the first instruction is eval
+            if ev.atom != b"\x02":
+                return self, self.to(0)
+            quoted_inner = next(i)
+            # then the quoted inner mod
+            if quoted_inner.pair[0].atom != b"\x01":
+                return self, self.to(0)
+            mod = quoted_inner.pair[1]
+            uncurried_args = next(i)
```
arvidn, line 151: notably, I'm not checking that `i` is now at the end. I get the impression the only way to do that is to call `next()` again and expect a `StopIteration` exception. Is it important to ensure there's no garbage at the end of this list?

The same thing goes for the `ai` iterator below.

chia/types/blockchain_format/program.py
```diff
@@ -136,11 +136,37 @@ def curry(self, *args) -> "Program":
             fixed_args = [4, (1, arg), fixed_args]
         return Program.to([2, (1, self), fixed_args])
 
-    def uncurry(self) -> Tuple["Program", "Program"]:
-        r = uncurry(self)
-        if r is None:
+    def uncurry(self) -> Tuple[Program, Program]:
+        def match(o: SExp, expected: bytes) -> None:
```
trepca, line 140: `assert_match` maybe?

prozacchiwawa (approved):

chia/types/blockchain_format/program.py
arvidn: I was considering using actual `assert` (and catch it), but it would be a bit of an abuse. Perhaps it's me by "assert" implies that failure is a bug (like in systems languages).

trepca (approved):

PR #13205, opened 2022-08-27
don't run the get_offered_coins post-init check

arvidn:
It's very expensive.

@Quexington Is this check a security feature? I get the impression it's mostly a debug assertion

To quantify the cost:

| --- | --- | --- | --- |
| Offer.from_bytes() | 7.73s | 3.57s | 46.18% |

before:
offer-parsing-main

after:
offer-parsing-patched

### DIFF
--- a/chia/wallet/trading/offer.py
+++ b/chia/wallet/trading/offer.py
@@ -97,13 +97,6 @@ class Offer:
         return announcements
 
     def __post_init__(self) -> None:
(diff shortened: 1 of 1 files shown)


arvidn:
I updated this patch to make the same change as what's in `release/1.6.0`. So they should merge well.

Quexington (approved):

trepca (approved):

PR #13211, opened 2022-08-28
introduce an UncurriedPuzzle class

arvidn:
containing the mod and args from a puzzle that has already been uncurried. Pass this around to functions that want to access the uncurried puzzle, to avoid uncurrying the same puzzle multiple times.

This speeds up the wallet, especially for generalized NFTs where puzzles are matched against multiple kinds.

| --- | --- | --- | --- |
| Offer.from_bytes() |  7.82s |  4.99s | 63.81% |
| uncurry() CPU | 54% | 26.36% | 48.81% |
| uncurry() calls | 4600 | 1300 | 28.26% |

Since these measurements were made, other optimizations to `Offer.from_bytes()` as well as `uncurry()` have landed in `main`. The number of calls benchmark should still accurately represent the effect of this patch.

before:
offer-parsing-main

after:

offer-parsing-patched


trepca (approved):
lgtm

PR #13213, opened 2022-08-28
fix performance issue in wallet with offers

arvidn:
Skip parsing (and computing) offers just to get to the involved coins, instead use the DB field for coins_of_interest. This fixes a serious performance issue in wallets containing offers.

@Quexington please confirm that this field is in fact the same as what you get from `get_involved_coins()` on an offer.

| -- | -- | -- | -- |
| chia wallet show | 376s | 0.75s | 0.2% |

Before this change, the wallet profile for `chia wallet show` in the test wallet looked like this:
chia-hotspot-main

### DIFF
--- a/chia/wallet/trade_manager.py
+++ b/chia/wallet/trade_manager.py
@@ -199,7 +199,7 @@ class TradeManager:
 
         coins_of_interest = []
         for trade_offer in all_pending:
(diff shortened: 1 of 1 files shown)


geoffwalmsley (approved):

Quexington (approved):
Yeah this is my bad, nice catch

trepca (approved):

PR #13217, opened 2022-08-28
--only wallet/rpc --duplicates 60

altendky:
Testing out the reliability of https://github.com/Chia-Network/chia-blockchain/pull/13163.


tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -906,7 +916,8 @@ def have_nfts():
     await farm_transaction_block(full_node_api, wallet_1_node)
     await time_out_assert(5, check_mempool_spend_count, True, full_node_api, 0)
     await time_out_assert(5, wallet_is_synced, True, wallet_1_node, full_node_api)
-
+    assert wallet_height_exactly(wallet_1_node, 8)
+    await time_out_assert(5, check_wallet_entries, True, wallet_1_node, WalletType.NFT)
     nft_wallet_id_1 = (
```
almogdepaz, line 921: i feel like theres a bug here, im checking that the nft wallet entry exists but still in some tests we fail at line 921 on a index out of range

so either im missing something or we clear the wallet record (that can happen in a reorg is seems)

PR #13218, opened 2022-08-29
Debug get tree hash

arvidn:
this patch factors out calls to `get_tree_hash()` to avoid redundant computations of tree hashes. It also avoids computing tree hashes intended to be debug-logged, when not in debug-logging mode.


trepca (approved):
lgtm

PR #13219, opened 2022-08-29
Minimized a chance where `sudo` prompts users for password

ChiaMineJP:
If `python`, `sqlite`, `openssl` are already available, we don't need to run `sudo apt install xxx`.
This PR tries to minimize the chance to prompt users for their password.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -129,15 +129,12 @@ check_python_version() {
   PYTHON_MINOR_VER=$(echo "$PYTHON_VERSION" | cut -d'.' -f2)
 
   if [ "$PYTHON_MAJOR_VER" -ne "3" ] || [ "$PYTHON_MINOR_VER" -lt "7" ] || [ "$PYTHON_MINOR_VER" -ge "11" ]; then
-    echo "Chia requires Python version >= 3.7 and  < 3.11.0" >&2
-    echo "Current Python version = $PYTHON_VERSION" >&2
-    # If Arch, direct to Arch Wiki
-    if type pacman >/dev/null 2>&1 && [ -f "/etc/arch-release" ]; then
-      echo "Please see https://wiki.archlinux.org/title/python#Old_versions for support." >&2
-    fi
-    exit 1
+    set -e
+    return 1
   fi
+
   set -e
+  return 0
 }
 
 find_python() {
@@ -147,10 +144,6 @@ find_python() {
     if command -v python$V >/dev/null; then
       if [ "$BEST_VERSION" = "" ]; then
         BEST_VERSION=$V
-        if [ "$BEST_VERSION" = "3" ]; then
-          PY3_VERSION=$(python$BEST_VERSION --version | cut -d ' ' -f2)
-          check_python_version "$PY3_VERSION"
-        fi
       fi
     fi
   done
@@ -173,6 +166,11 @@ if [ -n "$INSTALL_PYTHON_VERSION" ]; then
   INSTALL_PYTHON_PATH=python${INSTALL_PYTHON_VERSION}
   PY3_VERSION=$($INSTALL_PYTHON_PATH --version | cut -d ' ' -f2)
   check_python_version "$PY3_VERSION"
+  CHECK_PYTHON_VER_RESULT=$?
+  if [ "$CHECK_PYTHON_VER_RESULT" != "0" ]; then
+    INSTALL_PYTHON_VERSION=
+    INSTALL_PYTHON_PATH=
+  fi
 fi
 
 find_sqlite() {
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -122,30 +122,103 @@ install_python3_and_sqlite3_from_source_with_yum() {
...
+SQLITE_MINOR_VER=
+OPENSSL_VERSION_STRING=
+OPENSSL_VERSION_INT=
+
+if [ -z "$INSTALL_PYTHON_VERSION" ]; then
+  INSTALL_PYTHON_VERSION=$(find_python)
+fi
+
+if [ -n "$INSTALL_PYTHON_VERSION" ]; then
+  INSTALL_PYTHON_PATH=python${INSTALL_PYTHON_VERSION}
```
altendky, line 163: Should this have the 3.7 default from below?  I'll note that `Install.ps1` actually tries each supported version from newest to oldest.  But that should probably be done separately if of interest.

altendky (dismissed):
The openssl bit was specifically intended to trigger an update, not just an install when missing.  It was added when there was a critical ssl update a few months ago.  I'm not sure if that is still considered important.

Thanks for the work on this!

install.sh
```diff
@@ -122,30 +122,103 @@ install_python3_and_sqlite3_from_source_with_yum() {
...
+fi
+
+find_sqlite() {
```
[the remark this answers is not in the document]
ChiaMineJP: Thanks for the comment. I updated my code.

install.sh
```diff
@@ -122,30 +122,103 @@ install_python3_and_sqlite3_from_source_with_yum() {
...
+  if [ -n "$INSTALL_PYTHON_PATH" ]; then
+    # Check sqlite3 version bound to python
+    SQLITE_VERSION=$($INSTALL_PYTHON_PATH -c 'import sqlite3; print(sqlite3.sqlite_version)')
+    SQLITE_MAJOR_VER=$(echo "$SQLITE_VERSION" | cut -d'.' -f1)
+    SQLITE_MINOR_VER=$(echo "$SQLITE_VERSION" | cut -d'.' -f2)
+  fi
+}
+
+find_openssl() {
```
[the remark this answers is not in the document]
ChiaMineJP: See the latest code. I've taken your suggestion.

install.sh
```diff
@@ -122,30 +122,103 @@ install_python3_and_sqlite3_from_source_with_yum() {
...
+  if [ "$PYTHON_VENV_INSTALLED" -lt 1 ]; then
+    PACKAGE_INSTALL_REQUIRED=1
+  fi
+fi
+
 # Manage npm and other install requirements on an OS specific basis
 if [ "$(uname)" = "Linux" ]; then
   #LINUX=1
-  if [ "$UBUNTU_PRE_20" = "1" ]; then
+  if [ "$UBUNTU_PRE_20" = "1" ] && [ "$PACKAGE_INSTALL_REQUIRED" = "1" ]; then
```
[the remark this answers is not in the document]
ChiaMineJP: I updated the code in response to your suggestion.

install.sh
ChiaMineJP: Code updated.

hoffmang9:
As python just uses the system installed OpenSSL we want to be forcing an upgrade to the most current OpenSSL on an ongoing basis. This will certainly not be the last time OpenSSL has a vulnerability...

altendky:
I had initially proposed a CLI flag to skip the system stuff.  I don't want to be entering my password into scripts all the time.  I also don't want to wait around for apt all the time.  A CLI flag would leave the option for default users to still get the baby sitting we have here.  Someone recently got banned from apt connections or something and so they couldn't use the installer for a few hours.  That's kind of the top level perspective that brought up this topic.  I wonder if the full suite of mature installers at this time has any bearing on the trade offs here.

ChiaMineJP:
@hoffmang9 @altendky 
Thank you for the comment.
I'll change the PR to:
- Add a cli flag to set `PACKAGE_INSTALL_REQUIRED` forcibly to `0`
- Set `PACKAGE_INSTALL_REQUIRED=1` when OpenSSL version is old if the flag above is not specified.

altendky:
I think to check if openssl is old I would basically expect to do an `apt update` and then ask it.  At that point we maybe may as well just do the install instead?  Unless you can ask apt if there are new versions without doing the global/`sudo`-requiring `apt update`?

ChiaMineJP:
OpenSSL will be updated only when version < 3.0.2 or 1.1.1n or 1.0.2zd is installed.
We won't update openssl just because there are new versions.

altendky:
I thought Gene wanted us to do that.  Otherwise we have to push an updated `install.sh` to everyone to get the benefit of this when there's another important update to openssl.  I'm on the fence about it being our place to update the users system, but the upside is certainly pretty straightforward.

ChiaMineJP:
Overhauled entire construct of `install.sh`.
Added `-s` option. (Skip python install)
If you don't specify `-s`, then `install.sh` works just the same as usual.
If you specify `-s`, Python installation (as well as openssl/sqlite) will be skipped.

emlowe (approved):
I like this idea of a skip option

install.sh
```diff
@@ -3,10 +3,11 @@
 set -o errexit
 
 USAGE_TEXT="\
-Usage: $0 [-adh]
+Usage: $0 [-adsh]
 
   -a                          automated install, no questions
   -d                          install development dependencies
+  -s                          skip python package installation and just do pip install
```
altendky, line 10: I'm wondering if it would be good to add this to `Install.ps1` as well.  I'm not entirely sure.  On the one hand, it is pointless right now afaik, on the other, maintaining parity in the accepted options is useful.

ChiaMineJP: As to `Install.ps1`, it does not try to install `python.exe`.
So by default `-s` is `true`.
Windows user should install Python by themselves, so it doesn't require Administrative previledge.

PR #13228, opened 2022-08-29
Add a new protocol message for faster puzzle/solution fetching

Quexington:
Basically we make the current puzzle/solution request with a coin name instead of the full coin info.  The full node finds the block generator which has all of the coin spends in it, but importantly the coin spends have only the following information [parent_id, puzzle_reveal, amount].  This means in order to identify if a particular spend is the one the wallet is requesting it has to tree hash the puzzle_reveal then calculate the coin name using that info.  Since tree hashing was recently found to be an expensive operation and since that call basically forces a full node to do probably 20-30 treehashes on average before it finds the correct spend, I would imagine that's one of the things that makes the call pretty slow.

### DIFF
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -3,7 +3,7 @@ import dataclasses
 import time
 import traceback
 from secrets import token_bytes
-from typing import Dict, List, Optional, Tuple, Set
+from typing import Dict, List, Optional, Tuple, Set, Union
 
 from blspy import AugSchemeMPL, G2Element
 from chiabip158 import PyBIP158
@@ -14,7 +14,10 @@ from chia.consensus.block_record import BlockRecord
 from chia.consensus.pot_iterations import calculate_ip_iters, calculate_iterations_quality, calculate_sp_iters
 from chia.full_node.bundle_tools import best_solution_generator_from_template, simple_solution_generator
 from chia.full_node.full_node import FullNode
-from chia.full_node.mempool_check_conditions import get_puzzle_and_solution_for_coin
+from chia.full_node.mempool_check_conditions import (
+    get_puzzle_and_solution_for_coin,
+    get_puzzle_and_solution_for_coin_with_full_info,
+)
 from chia.full_node.signage_point import SignagePoint
 from chia.protocols import farmer_protocol, full_node_protocol, introducer_protocol, timelord_protocol, wallet_protocol
 from chia.protocols.full_node_protocol import RejectBlock, RejectBlocks
@@ -1267,7 +1270,23 @@ class FullNodeAPI:
 
     @api_request
     async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSolution) -> Optional[Message]:
-        coin_name = request.coin_name
+        return await self._request_puzzle_solution(request)
+
+    @api_request
+    async def request_puzzle_solution_with_coin_info(
+        self, request: wallet_protocol.RequestPuzzleSolutionWithCoinInfo
+    ) -> Optional[Message]:
+        return await self._request_puzzle_solution(request)
+
+    async def _request_puzzle_solution(
+        self,
+        request: Union[wallet_protocol.RequestPuzzleSolution, wallet_protocol.RequestPuzzleSolutionWithCoinInfo],
+    ) -> Optional[Message]:
+        if isinstance(request, wallet_protocol.RequestPuzzleSolutionWithCoinInfo):
+            requested_coin = Coin(request.parent_id, request.puzzle_hash, request.amount)
+            coin_name: bytes32 = requested_coin.name()
+        else:
+            coin_name = request.coin_name
         height = request.height
         coin_record = await self.full_node.coin_store.get_coin_record(coin_name)
         reject = wallet_protocol.RejectPuzzleSolution(coin_name, height)
@@ -1286,9 +1305,14 @@ class FullNodeAPI:
 
         block_generator: Optional[BlockGenerator] = await self.full_node.blockchain.get_block_generator(block)
         assert block_generator is not None
-        error, puzzle, solution = get_puzzle_and_solution_for_coin(
-            block_generator, coin_name, self.full_node.constants.MAX_BLOCK_COST_CLVM
-        )
+        if isinstance(request, wallet_protocol.RequestPuzzleSolutionWithCoinInfo):
+            error, puzzle, solution = get_puzzle_and_solution_for_coin_with_full_info(
+                block_generator, requested_coin, self.full_node.constants.MAX_BLOCK_COST_CLVM
+            )
+        else:
+            error, puzzle, solution = get_puzzle_and_solution_for_coin(
+                block_generator, coin_name, self.full_node.constants.MAX_BLOCK_COST_CLVM
+            )
 
         if error is not None:
             return reject_msg
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -1283,8 +1283,7 @@ class FullNodeAPI:
         request: Union[wallet_protocol.RequestPuzzleSolution, wallet_protocol.RequestPuzzleSolutionWithCoinInfo],
     ) -> Optional[Message]:
         if isinstance(request, wallet_protocol.RequestPuzzleSolutionWithCoinInfo):
-            requested_coin = Coin(request.parent_id, request.puzzle_hash, request.amount)
-            coin_name: bytes32 = requested_coin.name()
+            coin_name: bytes32 = request.coin.name()
         else:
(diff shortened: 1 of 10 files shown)


chia/full_node/mempool_check_conditions.py
```diff
@@ -70,6 +74,36 @@ def get_puzzle_and_solution_for_coin(generator: BlockGenerator, coin_name: bytes
...
+                raise ValueError(f"coin spend could not be found for coin {bytes32(coin.name())} in specified block")
+            coin_spend = coin_spend_list.first()
+            if (
+                coin_spend.at("f") == requested_parent
+                and coin_spend.at("rrf") == requested_amount
+                and coin_spend.at("rf").get_tree_hash() == requested_puzzle_hash  # tree hash last for performance
+            ):
+                return None, coin_spend.at("rf"), coin_spend.at("rrrf")
+            else:
+                coin_spend_list = coin_spend_list.rest()
```
arvidn, line 101: this block can be simplified I think. Consider:
```
for spend in coin_spend_list.as_iter():
    parent, puzzle, amount, solution = spend.as_iter()
    if parent == requested_parent
        and amount == requested_amount
        and Program.to(puzzle).get_tree_hash() == requested_puzzle_hash:
        return None, Program.to(puzzle), Program.to(solution)
    return ValueError("coin not found"), None, None
```

chia/wallet/util/wallet_sync_utils.py
```diff
@@ -389,3 +390,29 @@ async def fetch_header_blocks_in_range(
         assert res_h_blocks is not None
         blocks.extend([bl for bl in res_h_blocks.header_blocks if bl.height >= start])
     return blocks
+
+
+async def fetch_puzzle_solution_with_coin_info(height: uint32, coin: Coin, peer: WSChiaConnection) -> CoinSpend:
```
arvidn, line 395: I think while we're adding a new message, we should make it have a list of coins, so we can perform bulk lookups

chia/wallet/util/wallet_sync_utils.py
```diff
@@ -389,3 +390,29 @@ async def fetch_header_blocks_in_range(
...
+            )
+        )
+    else:
+        solution_response = await peer.request_puzzle_solution(
+            wallet_protocol.RequestPuzzleSolution(coin.name(), height)
+        )
+    if solution_response is None or not isinstance(solution_response, wallet_protocol.RespondPuzzleSolution):
+        raise ValueError(f"Was not able to obtain solution {solution_response}")
+    assert solution_response.response.puzzle.get_tree_hash() == coin.puzzle_hash
+    assert solution_response.response.coin_name == coin.name()
```
arvidn, line 412: I would prefer we move away from using `assert` as a data sanitization mechanism. Normally `assert` describes a condition that's always true, short of bugs.

chia/full_node/mempool_check_conditions.py
```diff
@@ -70,6 +74,36 @@ def get_puzzle_and_solution_for_coin(generator: BlockGenerator, coin_name: bytes
         return e, None, None
 
 
+def get_puzzle_and_solution_for_coin_with_full_info(generator: BlockGenerator, coin: Coin, max_cost: int):
```
arvidn, line 77: we should have tests for this function too

chia/wallet/util/wallet_sync_utils.py
Quexington: This might complicate the rate limiting, are you sure it's a good idea?  You can essentially make a bunch of these requests as one request and it only counts as a single request according to the rate limit.

arvidn: I'm not sure it's a good idea. but we also currently suffer from the issue of only ever having a single request outstanding (afaict) in the wallet. We can solve that by pipelining these requests, but since we would need to spawn a bunch of tasks to do that, it won't be very elegant.

I suppose making the rate limit count requests internal to messages wouldn't be very elegant either.

However, consider the case where the wallet has 100 coins it wants to look up in the same block. It would be a net win for the node to look them all up at the same time.

Maybe the trade-off is that all coins requested to be looked up has to be in the same block, i.e. a single `height` field in the message.

chia/full_node/mempool_check_conditions.py
```diff
@@ -5,15 +5,19 @@
...
 from chia.types.generator_types import BlockGenerator
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.errors import Err
 from chia.util.ints import uint32, uint64, uint16
 from chia.wallet.puzzles.generator_loader import GENERATOR_FOR_SINGLE_COIN_MOD
+from chia.wallet.puzzles.load_clvm import load_clvm
 from chia.wallet.puzzles.rom_bootstrap_generator import get_generator
 
 GENERATOR_MOD = get_generator()
+DESERIALIZE_MOD = load_clvm("chialisp_deserialisation.clvm", package_or_requirement="chia.wallet.puzzles")
```
arvidn, line 20: this should use `load_serialized_clvm()` to avoid having to serialize it every time we use it

chia/wallet/util/wallet_sync_utils.py
Quexington: > However, consider the case where the wallet has 100 coins it wants to look up in the same block. It would be a net win for the node to look them all up at the same time.

I think this is probably not very likely almost ever.  I _could_ see it happening in the case of like a bulk mint but that being a one time event means it's probably not the end of the world if the wallet slows down while processing that one block.  And I'm not sure the wallet would have the coordination to batch the request in that scenario either, it would probably end up making a bunch of requests anyways.

Quexington: Also I think we used to have a "pipeline" concept with the WalletActionStore so maybe that's something we can revisit in a more elegant way

arvidn: yeah, I'm a bit ignorant of what situations are likely to happen in real life. I don't believe any of these lookups in the toilet wallet are in the same block for instance.

chia/server/rate_limit_numbers.py
```diff
@@ -111,6 +111,7 @@ def compose_rate_limits(old_rate_limits: Dict[str, Any], new_rate_limits: Dict[s
             ProtocolMessageTypes.request_peers: RLSettings(10, 100),
             ProtocolMessageTypes.respond_peers: RLSettings(10, 1 * 1024 * 1024),
             ProtocolMessageTypes.request_puzzle_solution: RLSettings(1000, 100),
+            ProtocolMessageTypes.request_puzzle_solution_with_coin_info: RLSettings(1000, 100),
```
arvidn, line 114: I think we should allow more than 1000 per minute here. I would think closer to 3000 (that's 50 per second).

Quexington: @mariano54 thoughts?

mariano54: Agreed,  this might become a bottleneck in syncing cats,  if you can only do 50/second.
Its a complicated issue because we have to balance DOS attacks as well as wallet performance. I'm not sure

arvidn (approved):

chia/full_node/full_node_api.py
```diff
@@ -1267,7 +1270,22 @@ async def send_transaction(self, request: wallet_protocol.SendTransaction, *, te
...
+    async def _request_puzzle_solution(
+        self,
+        request: Union[wallet_protocol.RequestPuzzleSolution, wallet_protocol.RequestPuzzleSolutionWithCoinInfo],
+    ) -> Optional[Message]:
+        if isinstance(request, wallet_protocol.RequestPuzzleSolutionWithCoinInfo):
+            coin_name: bytes32 = request.coin.name()
+        else:
+            coin_name = request.coin_name
         height = request.height
         coin_record = await self.full_node.coin_store.get_coin_record(coin_name)
```
trepca, line 1290: given `coin_record` is always fetched, doesn't that provide coin info in any case? (if provided through protocol request or not)

arvidn: that is a good observation. We can probably get away without introducing a new message type

arvidn (changes requested):
it seems we don't need the new message type after all

chia/full_node/full_node_api.py
Quexington: :) good catch

arvidn:
the rate limit of the message is still an issue though, most likely

Quexington:
As @trepca pointed out, this is actually uneccessary because the node already has the full info, it doesn't need to come across in a message.

Quexington:
@arvidn is going to make a PR that encompasses the underlying optimization here though.

arvidn:
https://github.com/Chia-Network/chia-blockchain/pull/13237

PR #13229, opened 2022-08-29
minor fixes for Serialized program

arvidn:
minor fixes related to running serialized programs


chia/full_node/mempool_check_conditions.py
```diff
@@ -49,6 +49,7 @@ def get_name_puzzle_conditions(
         if err is not None:
             return NPCResult(uint16(err), None, uint64(0))
         else:
+            assert result is not None
```
arvidn, line 52: I'm not sure why, but something made mypy realize that `result` is an `Optional` here. There's nothing in the type system describing that *either* `err` or `result` is set.

mypy.ini
```diff
@@ -17,7 +17,7 @@ no_implicit_reexport = True
...
...ue,tests.util.test_misc,tests.util.test_network,tests.util.test_network_protocol_files,tests.wallet.cat_wallet.test_cat_lifecycle,tests.wallet.cat_wallet.test_cat_wallet,tests.wallet.cat_wallet.test_offer_lifecycle,tests.wallet.cat_wallet.test_trades,tests.wallet.did_wallet.test_did,tests.wallet.did_wallet.test_did_rpc,tests.wallet.did_wallet.test_nft_rpc,tests.wallet.did_wallet.test_nft_wallet,tests.wallet.rl_wallet.test_rl_rpc,tests.wallet.rl_wallet.test_rl_wallet,tests.wallet.rpc.test_wallet_rpc,tests.wallet.simple_sync.test_simple_sync_protocol,tests.wallet.sync.test_wallet_sync,tests.wallet.test_bech32m,tests.wallet.test_chialisp,tests.wallet.test_puzzle_store,tests.wallet.test_singleton,tests.wallet.test_singleton_lifecycle,tests.wallet.test_singleton_lifecycle_fast,tests.wallet.test_taproot,tests.wallet.test_wallet_blockchain,tests.wallet.test_wallet_interested_store,tests.wallet.test_wallet_key_val_store,tests.wallet.test_wallet_user_store,chia.simulator.wallet_tools,tests.weight_proof.test_weight_proof,tools.analyze-chain,tools.run_block,tools.test_full_sync,tests.wallet.nft_wallet.test_nft_wallet,chia.wallet.nft_wallet.nft_puzzles,tests.wallet.nft_wallet.test_nft_puzzles]
```
arvidn, line 20: this is removing the `rom_bootstrap_generator` from the exemption list

richardkiss:
Seems reasonable.

richardkiss (approved):

chia/types/blockchain_format/program.py
```diff
@@ -267,7 +267,7 @@ def run_as_generator(
         self, max_cost: int, flags: int, *args
     ) -> Tuple[Optional[int], Optional[SpendBundleConditions]]:
 
-        serialized_args = b""
+        serialized_args = bytearray()
```
richardkiss, line 270: I imagine operating on `bytearray` is faster than on `bytes`, so this seems good.

chia/types/blockchain_format/program.py
```diff
@@ -279,7 +279,7 @@ def run_as_generator(
 
         err, ret = run_generator(
             self._buf,
-            serialized_args,
+            bytes(serialized_args),
```
richardkiss, line 282: Is it strictly necessary to do this cast for `run_generator`? Maybe we can add `Union[bytes, bytearray]` to this argument and everything still works without the cast.

richardkiss: That said, even without that optimization, this is probably still faster.

chia/types/blockchain_format/program.py
arvidn: that's my understanding. in this case it's probably quite marginal

chia/types/blockchain_format/program.py
arvidn: yes, this is strictly necessary. PyO3 `PyBytes` does not support the buffer protocol. It would need a similar treatment as `parse_rust()` in order to accept a `memoryview` or `bytearrat`.

`PyBytes` fail without this conversion

trepca (approved):

PR #13232, opened 2022-08-30
🍒2 - Fix several bugs with untrusted sync, and correct sync status (#13133)

wallentx:
* Hopefully fix bug with untrusted sync

* Try removing the cache

* Add cache back, and don't rollback

* Compare to finished_sync_up_to

* Fix plotnfts, and tests

* Async function

* fix wallet blockchain tests

* Fix test, and improve detection of synced

* Correct sync status

* Fix more tests

* Fix more tests

* Fix NFT tests

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -728,7 +728,7 @@ class PoolWallet:
             history: List[Tuple[uint32, CoinSpend]] = await self.get_spend_history()
             last_height: uint32 = history[-1][0]
             if (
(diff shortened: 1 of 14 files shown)


altendky (commented):
I'm going to defer to @mariano54 to inspect this for any concerning differences between `main` and `release/1.6.0`.  Probably not a big deal, but just in case something catches their eye.

mariano54 (approved):

emlowe (approved):

PR #13233, opened 2022-08-30
🍒3 - update _peak_height after committing the block to the DB (#13157)

wallentx:
Cherry pick 2d551dd


altendky (approved):
Seems to be doing the same thing as the original.

emlowe (approved):

PR #13237, opened 2022-08-30
Implement generator_for_single_coin() in python instead of clvm

arvidn:
and apply quex' optimization to only computing the puzzle hash if the parent and amount matches.


tests/core/test_cost_calculation.py
```diff
@@ -155,10 +158,12 @@ async def test_mempool_mode(self, bt):
         )
         assert npc_result.error is None
 
-        coin_name = npc_result.conds.spends[0].coin_id
-        error, puzzle, solution = get_puzzle_and_solution_for_coin(
-            generator, coin_name, test_constants.MAX_BLOCK_COST_CLVM
+        coin = Coin(
+            bytes32.fromhex("3d2331635a58c0d49912bc1427d7db51afe3f20a7b4bcaffa17ee250dcbcbfaa"),
+            bytes32.fromhex("14947eb0e69ee8fc8279190fc2d38cb4bbb61ba28f1a270cfd643a0e8d759576"),
+            300,
```
arvidn, line 164: we kind of construct the generator program by hand above, so it seems fine to construct the coin by hand too

chia/full_node/mempool_check_conditions.py
```diff
@@ -55,17 +60,26 @@ def get_name_puzzle_conditions(
...
+
+        coin_spends = result.first()
+        for spend in coin_spends.as_iter():
+            parent, puzzle, amount, solution = spend.as_iter()
+            if (
+                parent == requested_parent
+                and amount == requested_amount
+                and Program.to(puzzle).get_tree_hash() == coin.puzzle_hash
+            ):
+                return None, Program.to(puzzle), Program.to(solution)
```
Quexington, line 81: I don't think these `Program.to()`'s are necessary since `as_iter()` returns an iterator of programs right?

chia/full_node/mempool_check_conditions.py
```diff
@@ -55,17 +60,26 @@ def get_name_puzzle_conditions(
...
-        return None, puzzle, solution
+        args = [bytes(a) for a in generator.generator_refs]
+        cost, result = generator.program.run_with_cost(DEFAULT_CONSTANTS.MAX_BLOCK_COST_CLVM, DESERIALIZE_MOD, args)
+
+        requested_parent = Program.to(coin.parent_coin_info)
+        requested_amount = Program.to(coin.amount)
+
+        coin_spends = result.first()
+        for spend in coin_spends.as_iter():
+            parent, puzzle, amount, solution = spend.as_iter()
```
Quexington, line 75: This may be over the top but is there a chance that `as_iter()` does some extra work that we wouldn't need if the spend we're looking for is one of the first ones?  I can't imagine `first()` and `rest()` are that expensive but we _could_ technically do fewer of them if we only do `rest()` after we have decided `first()` is not the spend we're looking for.  Anyways, probably not worth it from a code complexity standpoint.

Quexington (approved):
A couple of maybe changes but otherwise this is fine as is

chia/full_node/mempool_check_conditions.py
arvidn: actually, it returns `SExp` objects. iirc, mypy complained without this

chia/full_node/mempool_check_conditions.py
arvidn: I'm hoping to replace this with a rust function where all of this state and searching it will stay on the rust side

arvidn: hm. removing it now does *not* make mypy complain. But I don't fully trust that mypy always find things like this

trepca (approved):

PR #13238, opened 2022-08-30
fix run_block

arvidn:
to actually pass in the chialisp_deserialization program, as required by consensus.

This is my understanding of the consensus rules of what a generator program expects

### DIFF
--- a/tools/run_block.py
+++ b/tools/run_block.py
@@ -45,7 +45,6 @@ from clvm.casts import int_from_bytes
 
 from chia.consensus.constants import ConsensusConstants
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
-from chia.full_node.generator import create_generator_args
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.program import SerializedProgram
 from chia.types.blockchain_format.sized_bytes import bytes32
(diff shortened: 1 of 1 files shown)


tools/run_block.py
```diff
@@ -98,8 +100,8 @@ def npc_to_dict(npc: NPC):
 
 def run_generator(block_generator: BlockGenerator, constants: ConsensusConstants, max_cost: int) -> List[CAT]:
 
-    args = create_generator_args(block_generator.generator_refs).first()
-    _, block_result = block_generator.program.run_with_cost(max_cost, 0, args)
```
arvidn, line 102: as far as I can tell, passing `0` as the first argument here means that the block generator does not have access to the clvm parser. if it tries to parse one of the block-ref arguments, presumably this would fail.

richardkiss: Yes, I believe that's true – the standard says a (buggy and therefore rather useless) piece of code that deserializes clvm is passed to the generator. We can't change it without a hard fork.

So I will write a simple drop-in block generator that fails if this argument is wrong. We can use this for validation here and possibly other places... essentially "copy-protection" to ensure that the ROM is compatible.

arvidn: I see. that explains why nothing seems to be failing, despite passing 0 here

emlowe (approved):

tools/run_block.py
richardkiss: https://gist.github.com/richardkiss/e4a0e6ceb97489f7a9f1d3938b1a0d43

arvidn: there should be a test against that block, yes. More importantly, the main blockchain should be tested against it.

PR #13261, opened 2022-08-31
On chain wallet notifications

Quexington:
The idea here is to have a way in which someone can pay someone else on chain in such a way that it sends the recipient wallet a notification.


chia/wallet/notification_manager.py
```diff
@@ -0,0 +1,97 @@
...
+                await self.notification_store.add_notification(
+                    Notification(
+                        coin_state.coin.name(),
+                        coin_memos[1],
+                        uint64(coin_state.coin.amount),
+                    )
+                )
+            return True
+
+    async def send_new_notification(
```
trepca, line 72: how would sender know what's the minimal amount `target` is requiring?

Quexington: In my mind, they don't.  Sending a notification doesn't guarantee acknowledgement anyways, so you pay as much as you are willing to in order to get the message across.

chia/wallet/notification_store.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        coin_ids_str_list = "("
+        for _ in coin_ids:
+            coin_ids_str_list += "?"
+            coin_ids_str_list += ","
+        coin_ids_str_list = coin_ids_str_list[:-1] if len(coin_ids_str_list) > 1 else "("
+        coin_ids_str_list += ")"
+
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            cursor = await conn.execute(f"SELECT * from notifications WHERE coin_id IN {coin_ids_str_list}", coin_ids)
```
trepca, line 76: use `conn.execute_fetchall`, saves you next 2 lines

chia/wallet/notification_store.py
```diff
@@ -0,0 +1,125 @@
...
+from typing import List, Optional
+
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.db_wrapper import DBWrapper2
+from chia.util.ints import uint32, uint64
+
+
+@dataclasses.dataclass(frozen=True)
+class Notification:
+    coin_id: bytes32
```
trepca, line 12: should we also include coin's puzzle hash so we can easily filter by type of notifications?

Quexington: The puzzle hash is parameterized by 4 things:
* The notification mod
* The wallet address it was sent to
* The message
* The amount

I don't see how any of those things would be helpful except the message and amount which are already on this class

Quexington: I'm not making any biases in the wallet for what kind of notifications we sync.  I see this as a UI task: delete the notficiations you don't want.  It doesn't prevent any spam because you can always make a notification of a certain format that has a dead URL for example.

chia/wallet/puzzles/notification.clvm
```diff
@@ -0,0 +1,8 @@
+(mod (
+    TARGET
+    MSG_HASH
```
trepca, line 3: why do we need `MSG_HASH` here?

Quexington: I think when I wrote it, I just wanted this to be distinct from a regular ACS puzzle, make something about it "notification" specific but I think you're right, that's probably silly

mariano54:
should we add reorg support to remove the notifications if they get reorged?

Quexington:
> should we add reorg support to remove the notifications if they get reorged?

Seems like more trouble than it's worth, and the idea here is actually just to get your message across, you don't really care if it persists on chain.  This does mean that potentially a reorg could make it so the notification didn't cost you anything, but the main point of the cost is to prevent spam and it's hard to take advantage of reorgs to spam someone.

trepca:
A user would be under impression that notification was sent when it wasn't due to reorg no?

I feel like long term we need to solve the reorg issue on a level above. Wallet should just implement an interface for new and reverted coins or something like that.

Quexington:
> A user would be under impression that notification was sent when it wasn't due to reorg no?
> 
> I feel like long term we need to solve the reorg issue on a level above. Wallet should just implement an interface for new and reverted coins or something like that.

Mariano was referring to the receiving side which is primarily what I responded to.  On the sender side, there is a wallet transaction for the notification so if it gets reorged, that transaction will get deleted and of course no funds will be lost.

chia/util/initial-config.yaml
```diff
@@ -545,6 +545,11 @@ wallet:
   # Interval to resend unconfirmed transactions, even if previously accepted into Mempool
   tx_resend_timeout_secs: 1800
 
+  # Accept notifications from parties on chain
+  accept_notifications: False
+  # The amount someone has to pay you in mojos for you to see their notification
+  required_notification_amount: 1000000000
```
trepca, line 551: @esaung ran a poll and people felt most comfortable with 100M mojos

trepca:
backend lgtm, but we need a CLI support too

trepca (approved):

PR #13263, opened 2022-08-31
optimize get_block_store

arvidn:
for `get_puzzle_and_solution_for_coin()` to not parse the full block. We just need the block generator, pre_hash and generator refs.

Once we switch over to the rust implementation to pull out the puzzle and solution for a coin, the `get_full_block()` call is a significant cost. This patch reduces it.

### DIFF
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -9,11 +9,12 @@ from chia.types.blockchain_format.program import SerializedProgram
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.full_block import FullBlock
 from chia.types.weight_proof import SubEpochChallengeSegment, SubEpochSegments
-from chia.util.db_wrapper import DBWrapper2
+from chia.util.db_wrapper import DBWrapper2, execute_fetchone
 from chia.util.errors import Err
-from chia.util.full_block_utils import generator_from_block
+from chia.util.full_block_utils import block_info_from_block, generator_from_block
 from chia.util.ints import uint32
 from chia.util.lru_cache import LRUCache
+from chia.util.full_block_utils import GeneratorBlockInfo
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


chia/full_node/block_store.py
```diff
@@ -329,24 +359,23 @@ async def get_generator(self, header_hash: bytes32) -> Optional[SerializedProgra
...
-                try:
-                    return generator_from_block(block_bytes)
-                except Exception as e:
-                    log.error(f"cheap parser failed for block at height {row[1]}: {e}")
-                    # this is defensive, on the off-chance that
-                    # generator_from_block() fails, fall back to the reliable
-                    # definition of parsing a block
-                    b = FullBlock.from_bytes(block_bytes)
-                    return b.transactions_generator
+            row = await execute_fetchone(conn, formatted_str, (self.maybe_to_hex(header_hash),))
```
arvidn, line 362: this is a somewhat unrelated change to make `get_generator()` just have a single round-trip to the DB thread (by using `execute_fetchone()`)

chia/full_node/full_node_api.py
```diff
@@ -1279,7 +1279,7 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
         if header_hash is None:
             return reject_msg
 
-        block: Optional[FullBlock] = await self.full_node.block_store.get_full_block(header_hash)
+        block = await self.full_node.block_store.get_block_info(header_hash)
```
arvidn, line 1282: the object returned here is one that implements the `BlockInfo` protocol. essentially a class that has the 3 members of `FullBlock` required by `get_block_generator()`. We don't want an actual full block here, because that's expensive to construct.

tests/util/test_full_block_utils.py
```diff
@@ -234,8 +234,8 @@ def get_full_blocks() -> Iterator[FullBlock]:
 
 
 @pytest.mark.asyncio
-@pytest.mark.skip("This test is expensive and has already convinced us the parser works")
-async def test_parser(self):
+#@pytest.mark.skip("This test is expensive and has already convinced us the parser works")
```
arvidn, line 237: I enabled this test again, since I'm adding a test

chia/full_node/block_store.py
```diff
@@ -320,6 +321,37 @@ async def get_full_blocks_at(self, heights: List[uint32]) -> List[FullBlock]:
...
+                return None
+            if self.db_wrapper.db_version == 2:
+                block_bytes = zstd.decompress(row[0])
+            else:
+                block_bytes = row[0]
+
+            try:
+                return block_info_from_block(block_bytes)
+            except Exception as e:
+                log.error(f"cheap parser failed for block at height {row[1]}: {e}")
```
trepca, line 346: `log.exception` might work better there as also includes the traceback

chia/full_node/full_node_api.py
mariano54: Should we use the type annotation here `GeneratorBlockInfo`?

tests/util/test_full_block_utils.py
```diff
@@ -217,25 +224,26 @@ def get_full_blocks() -> Iterator[FullBlock]:
...
+                                            transactions_info,
+                                            generator,  # transactions_generator
+                                            refs_list,  # transactions_generator_ref_list
+                                        )
 
 
 @pytest.mark.asyncio
-@pytest.mark.skip("This test is expensive and has already convinced us the parser works")
-async def test_parser(self):
+# @pytest.mark.skip("This test is expensive and has already convinced us the parser works")
```
mariano54, line 245: Is this intentional? Is this test even slower now?

chia/full_node/full_node_api.py
arvidn: maybe. I would probably be more tempted to annotate it with the `BlockInfo` protocol though. the concrete type is really an implementation detail

tests/util/test_full_block_utils.py
arvidn: yes. in my first incarnation of this patch I had a bug that wasn't caught by this test (because it didn't cover the ref-list). So, regrettably, this test now takes longer. But we can probably disable it again after this lands. It's here now to demonstrate that the parser works correctly

mariano54 (approved):

PR #13266, opened 2022-09-01
Catch exception while tracking singleton.

fchirica:


chia/data_layer/data_layer.py
```diff
@@ -480,6 +480,8 @@ async def periodically_manage_data(self) -> None:
                     pass
                 except asyncio.CancelledError:
                     raise
+                except Exception as e:
+                    self.log.error(f"Exception while wallet tracking subscription's singleton: {type(e)} {e}")
```
altendky, line 484:
```suggestion
                    self.log.error(f"Exception while requesting wallet track subscription: {type(e)} {e}")
```

altendky (approved):

emlowe (approved):

PR #13268, opened 2022-09-01
Add data_layer_http to pyinstaller

emlowe:
Made sure to add DL to pyinstaller


altendky (approved):
reads ok here, happy if you've tested the builds and they work

PR #13274, opened 2022-09-01
Remove boolean from atari DB.

fchirica:
The Linux Debian installer build now uses an older version of SQLite that does not support `BOOLEAN`, `TRUE`, and `FALSE`.

```
Exception from 'data': {'error': "{'error': 'no such column: TRUE', 'success': False}", 'success': False}
```
```
2022-09-01T15:21:20.324 data_layer chia.rpc.util          : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia/rpc/util.py", line 16, in inner
  File "chia/rpc/data_layer_rpc_api.py", line 143, in get_keys_values
  File "chia/data_layer/data_layer.py", line 212, in get_keys_values
  File "chia/data_layer/data_layer.py", line 269, in _update_confirmation_status
  File "chia/rpc/wallet_rpc_client.py", line 739, in dl_latest_singleton
  File "chia/rpc/rpc_client.py", line 49, in fetch
ValueError: {'error': 'no such column: TRUE', 'success': False}
```

### DIFF
--- a/chia/data_layer/data_store.py
+++ b/chia/data_layer/data_store.py
@@ -133,7 +133,7 @@ class DataStore:
                     url TEXT,
                     ignore_till INTEGER,
                     num_consecutive_failures INTEGER,
-                    from_wallet BOOLEAN,
+                    from_wallet tinyint CHECK(from_wallet == 0 OR from_wallet == 1),
                     PRIMARY KEY(tree_id, url)
                 )
                 """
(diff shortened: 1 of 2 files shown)


altendky (approved):
fyi @Quexington

Quexington:
> fyi @Quexington

noted :+1:

emlowe:
Does this mean I need to delete my DB with this build?

emlowe (approved):

altendky:
Pretty sure no.  `TRUE` is just `1`, etc.  Now the column might be slightly different, `tinyint` vs. `BOOLEAN`, but that doesn't cause any trouble.  You won't have the check until you rebuild, but that's also not critical, just an extra guardrail.

PR #13298, opened 2022-09-02
Add `running_services` command to list all running services

### LINKED ISSUE #13313: Optimize daemon `is_running` command to return all services
Currently the `is_running` command takes a single service name and returns a response indicating whether that service is running or not. To reduce traffic/logging, the GUI should be able to request status for all registered services in a single request.

paninaro:
Issue #13313 

Added a `running_services` command to return the list of running services registered with the daemon. This complements the existing `is_running` command which only returns a response for a single requested service name.

The request doesn't take any params.

Request:
```{"ack": false, "command": "running_services", "data": {}, "destination": "daemon", "origin": "client", "request_id": "43dc226cef76963ddd56c7068972947f373918c24b0fdf5bfa767a1340271da6"}```

Response:
```{"ack": true, "command": "running_services", "data": {"running_services": ["chia_wallet", "chia_farmer", "chia_harvester", "chia_full_node"], "success": true}, "destination": "client", "origin": "daemon", "request_id": "43dc226cef76963ddd56c7068972947f373918c24b0fdf5bfa767a1340271da6"}```

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -1104,7 +1104,9 @@ class WebSocketServer:
     async def is_running(self, request: Dict[str, Any]) -> Dict[str, Any]:
         service = request.get("service", "*")
         services_to_check = (
-            service.split(",") if service != "*" else list(set([*self.services.keys(), *self.connections.keys()]))
+            list(map(str.strip, service.split(",")))
+            if service != "*"
+            else list(set([*self.services.keys(), *self.connections.keys()]))
         )
         responses = []
         for service_name in services_to_check:
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -326,6 +326,8 @@ class WebSocketServer:
             response = await self.stop_plotting(cast(Dict[str, Any], data))
         elif command == "stop_service":
             response = await self.stop_service(cast(Dict[str, Any], data))
+        elif command == "running_services":
+            response = await self.running_services(cast(Dict[str, Any], data))
         elif command == "is_running":
             response = await self.is_running(cast(Dict[str, Any], data))
         elif command == "is_keyring_locked":
@@ -1101,43 +1103,39 @@ class WebSocketServer:
         response = {"success": result, "service_name": service_name}
         return response
 
-    async def is_running(self, request: Dict[str, Any]) -> Dict[str, Any]:
-        service = request.get("service", "*")
-        services_to_check = (
-            list(map(str.strip, service.split(",")))
-            if service != "*"
+    def is_service_running(self, service_name: str) -> bool:
+        if service_name == service_plotter:
+            processes = self.services.get(service_name)
+            is_running = processes is not None and len(processes) > 0
+        else:
+            process = self.services.get(service_name)
+            is_running = process is not None and process.poll() is None
+            if not is_running:
+                # Check if we have a connection to the requested service. This might be the
+                # case if the service was started manually (i.e. not started by the daemon).
+                service_connections = self.connections.get(service_name)
+                if service_connections is not None:
+                    is_running = len(service_connections) > 0
+        return is_running
+
+    async def running_services(self, request: Dict[str, Any]) -> Dict[str, Any]:
+        requested_services = request.get("services", [])
+        services = (
+            requested_services
+            if len(requested_services) > 0
             else list(set([*self.services.keys(), *self.connections.keys()]))
         )
-        responses = []
-        for service_name in services_to_check:
-            if service_name == service_plotter:
-                processes = self.services.get(service_name)
-                is_running = processes is not None and len(processes) > 0
-                response = {
-                    "service_name": service_name,
-                    "is_running": is_running,
-                }
-            else:
-                process = self.services.get(service_name)
-                is_running = process is not None and process.poll() is None
-                if not is_running:
-                    # Check if we have a connection to the requested service. This might be the
-                    # case if the service was started manually (i.e. not started by the daemon).
-                    service_connections = self.connections.get(service_name)
-                    if service_connections is not None:
-                        is_running = len(service_connections) > 0
-                response = {
-                    "service_name": service_name,
-                    "is_running": is_running,
-                }
-            responses.append(response)
+        running_services = []
+        for service_name in services:
+            if self.is_service_running(service_name):
+                running_services.append(service_name)
 
-        if len(responses) == 1:
-            response = {"success": True, **responses[0]}
-        else:
-            response = {"success": True, "services": responses}
+        return {"success": True, "running_services": running_services}
 
-        return response
+    async def is_running(self, request: Dict[str, Any]) -> Dict[str, Any]:
+        service_name = request["service"]
+        is_running = self.is_service_running(service_name)
+        return {"success": True, "service_name": service_name, "is_running": is_running}
 
     async def exit(self) -> None:
         if self.websocket_runner is not None:
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -1123,7 +1123,7 @@ class WebSocketServer:
         services = (
             requested_services
             if len(requested_services) > 0
-            else list(set([*self.services.keys(), *self.connections.keys()]))
+            else list({*self.services.keys(), *self.connections.keys()})
         )
         running_services = [service_name for service_name in services if self.is_service_running(service_name)]
 
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -1102,30 +1102,38 @@ class WebSocketServer:
         return response
 
     async def is_running(self, request: Dict[str, Any]) -> Dict[str, Any]:
-        service_name = request["service"]
+        service = request.get("service", "*")
+        services_to_check = (
+            service.split(",") if service != "*" else list(set([*self.services.keys(), *self.connections.keys()]))
+        )
+        responses = []
+        for service_name in services_to_check:
+            if service_name == service_plotter:
+                processes = self.services.get(service_name)
+                is_running = processes is not None and len(processes) > 0
+                response = {
+                    "service_name": service_name,
+                    "is_running": is_running,
+                }
+            else:
+                process = self.services.get(service_name)
+                is_running = process is not None and process.poll() is None
+                if not is_running:
+                    # Check if we have a connection to the requested service. This might be the
+                    # case if the service was started manually (i.e. not started by the daemon).
+                    service_connections = self.connections.get(service_name)
+                    if service_connections is not None:
+                        is_running = len(service_connections) > 0
+                response = {
+                    "service_name": service_name,
+                    "is_running": is_running,
+                }
+            responses.append(response)
 
-        if service_name == service_plotter:
-            processes = self.services.get(service_name)
-            is_running = processes is not None and len(processes) > 0
-            response = {
-                "success": True,
-                "service_name": service_name,
-                "is_running": is_running,
-            }
+        if len(responses) == 1:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by paninaro** [comment]
The intent here is to allow the GUI to make a single call to query all registered services, rather than requesting the status of each service (whose name needs to be known beforehand) one-by-one. This should help cut down some of the daemon websocket traffic and logging.

As for using the value "*" as a wildcat to indicate all services -- I'm open to alternatives.

I tried omitting the entire `"service"` param, but that causes the server code to barf and I didn't feel like making more extensive changes to support that:

`> {"ack": false, "command": "is_running", "data": {}, "destination": "daemon", "origin": "client", "request_id": "43dc226cef76963ddd56c7068972947f373918c24b0fdf5bfa767a1340271da6"}`

`< {"ack": true, "command": "is_running", "data": {"error": "is_running requires \"data\"", "success": false}, "destination": "client", "origin": "daemon", "request_id": "43dc226cef76963ddd56c7068972947f373918c24b0fdf5bfa767a1340271da6"}`

**Review by xdustinface — CHANGES_REQUESTED** [review]
I think it would be good to add some sanity checks for the split result / service names and then also tests for them (seems like there is no test yet for `is_running` though)? :) Maybe even strip the split entries so that `service_a, service_b` works the same like `service_a,service_b`, not necessarily requesting this though, just had the thought.

**Comment by paninaro** [comment]
> I think it would be good to add some sanity checks for the split result / service names and then also tests for them (seems like there is no test yet for `is_running` though)? :) Maybe even strip the split entries so that `service_a, service_b` works the same like `service_a,service_b`, not necessarily requesting this though, just had the thought.

See d71a9168d692e54709dfefb0aa1944c8c7578ac6

File: chia/daemon/server.py [hunk 1]
```diff
@@ -1102,30 +1102,40 @@ async def stop_service(self, request: Dict[str, Any]) -> Dict[str, Any]:
...
-
-        if service_name == service_plotter:
-            processes = self.services.get(service_name)
-            is_running = processes is not None and len(processes) > 0
-            response = {
-                "success": True,
-                "service_name": service_name,
-                "is_running": is_running,
-            }
+        service = request.get("service", "*")
```
altendky, line 1105: What if this were a list of strings instead and we didn't have to apply nested parsing to the existing JSON parsed result?  Or perhaps `service` remains entirely as is, there's a new key named `services` that's a list, we just prepend `service` to `services` and move forward with only looking at `services`.  I don't think I've quite thought through all the little corners.

chia/daemon/server.py
```diff
@@ -1102,30 +1102,40 @@ async def stop_service(self, request: Dict[str, Any]) -> Dict[str, Any]:
...
-                # case if the service was started manually (i.e. not started by the daemon).
-                service_connections = self.connections.get(service_name)
-                if service_connections is not None:
-                    is_running = len(service_connections) > 0
-            response = {
-                "success": True,
-                "service_name": service_name,
-                "is_running": is_running,
-            }
+            response = {"success": True, "services": responses}
```
altendky, line 1138: I guess there's some argument for this in terms of backwards compatibility.  Maybe though we could define a consistent form?  Put the 'first' result into the root object as before but also always provide the `services` attribute?  Then callers don't need to special case one result.  They can just always use the new form from now on.

tests/core/daemon/test_daemon.py
```diff
@@ -169,6 +170,100 @@ async def reader(ws, queue):
...
+
+    # Mock daemon server that forwards to WebSocketServer
+    @dataclass
+    class Daemon:
+        # Instance variables used by WebSocketServer.is_running()
+        services: Dict[str, Union[List[Service], Service]]
+        connections: Dict[str, Optional[List[Any]]]
+
+        async def is_running(self, request: Dict[str, Any]) -> Dict[str, Any]:
+            return await WebSocketServer.is_running(self, request)  # type: ignore
```
altendky, line 191: Please provide the full ignore form such as `# type: ignore[no-any-return]` or whatever it is as well as a comment on the previously line either with a link to a bug or PR that would address the need for the ignore or a long-term justification for it.

tests/core/daemon/test_daemon.py
```diff
@@ -169,6 +170,100 @@ async def reader(ws, queue):
...
+
+    # Mock daemon server without any registered services/connections
+    lonelyDaemon = Daemon(services={}, connections={})
```
altendky, line 194: Looks like we need to let pylint complain about camel case.  Anyways, `lonely_daemon`.

tests/core/daemon/test_daemon.py
```diff
@@ -169,6 +170,100 @@ async def reader(ws, queue):
     assert blockchain_state_found
 
 
+@pytest.mark.asyncio
+async def test_is_running_wildcard():
```
altendky, line 174: I think this could be several separate tests without inducing much overhead.  Move the classes to the global scope, maybe make a fixture for each daemon form used here.  This allows granular results and also complete results even when the first assertion fails.

altendky (changes requested):

tests/core/daemon/test_daemon.py
paninaro: ack, been spending too much time in JavaScript lately.

chia/daemon/server.py
paninaro: @xdustinface and I have been discussing the merits of just creating a new `running_services` command rather than trying to shoehorn this new behavior into `is_running`. I was initially going for something small and quick, but I'm onboard with leaving `is_running` as-is and moving this new functionality to a new command.

tests/core/daemon/test_daemon.py
paninaro: Added separate tests.

tests/core/daemon/test_daemon.py
paninaro: Updated

chia/daemon/server.py
paninaro: No longer applicable. `is_running` behaves as it did before.

chia/daemon/server.py
paninaro: Added `running_services` instead of modifying `is_running`

tests/core/daemon/test_daemon.py
```diff
@@ -102,6 +139,57 @@ def assert_response(response: aiohttp.http_websocket.WSMessage, expected_respons
...
+            assert response_dict[k] == v
+
+
+def assert_is_running_response(response_dict: Dict[str, Any], expected_response_dict: Dict[str, Any]) -> None:
+    for k, v in expected_response_dict.items():
+        assert response_dict[k] == v
+
+
+@pytest.fixture(scope="session")
+def mock_lonely_daemon():
```
altendky, line 158: I personally wouldn't call this a mock since we aren't messing with any existing code, just passing a different thing to it.  But, I haven't reviewed the terminology around this recently.

altendky (changes requested):
The way that `/running_services` ends up working here I'm wondering if there's any use in ever passing it a list of services anyways.  If it responded with a failure when you requested a service it is unaware of that would be one thing but, as is, even if you ask about `blueberry` you still have to check the result to see if it contains `blueberry`.  The upside seems to be that you can avoid the cost of checking each service you don't care about.  But, is it expensive to check which services are running?

The other aspect I wonder about at an API design level, rather than just the intended scope of this PR, is whether there is other information we might like about the services in the future.  For example, rather than just returning a list of names we could return `{"full_node": {"running": True, "ping_response_time": 0.0027, "uptime": 4923.18}}`.  Some statuses we add in the future might be costly enough that we would want to retain the 'request specific services' ability I questioned above.  This would be more of a `/service_status` or such route I guess.  Costly checks could also be enabled via an optional selection in the request as to what statuses to actually collect.  This may be drifting into the exporter behavior though.

paninaro:
> The way that `/running_services` ends up working here I'm wondering if there's any use in ever passing it a list of services anyways. If it responded with a failure when you requested a service it is unaware of that would be one thing but, as is, even if you ask about `blueberry` you still have to check the result to see if it contains `blueberry`. The upside seems to be that you can avoid the cost of checking each service you don't care about. But, is it expensive to check which services are running?
> 
> The other aspect I wonder about at an API design level, rather than just the intended scope of this PR, is whether there is other information we might like about the services in the future. For example, rather than just returning a list of names we could return `{"full_node": {"running": True, "ping_response_time": 0.0027, "uptime": 4923.18}}`. Some statuses we add in the future might be costly enough that we would want to retain the 'request specific services' ability I questioned above. This would be more of a `/service_status` or such route I guess. Costly checks could also be enabled via an optional selection in the request as to what statuses to actually collect. This may be drifting into the exporter behavior though.

I don't foresee any cases where it would be especially useful to limit the list of services to query, so I'm onboard with removing that option.

I like the idea of returning stats etc. That does seem like it could be expensive to query, and not really in the scope of this endpoint, so I agree that it's probably better served by a `service_status` (or similar) command.

chia/daemon/server.py
```diff
@@ -1121,13 +1116,26 @@ async def is_running(self, request: Dict[str, Any]) -> Dict[str, Any]:
...
-                "service_name": service_name,
-                "is_running": is_running,
-            }
+        return is_running
+
+    async def running_services(self, request: Dict[str, Any]) -> Dict[str, Any]:
+        requested_services = request.get("services", [])
+        services = (
+            requested_services
+            if len(requested_services) > 0
```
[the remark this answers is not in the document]
paninaro: Removed the option to specify the list of "services" to query.

tests/core/daemon/test_daemon.py
```diff
@@ -18,6 +19,42 @@
...
+@dataclass
+class Daemon:
+    # Instance variables used by WebSocketServer.is_running()
+    services: Dict[str, Union[List[Service], Service]]
+    connections: Dict[str, Optional[List[Any]]]
+
+    def is_service_running(self, service_name: str) -> bool:
+        # Squelch mypy error:
+        # Argument 1 to "is_service_running" of "WebSocketServer" has incompatible type "Daemon";
+        #   expected "WebSocketServer"
```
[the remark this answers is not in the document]
paninaro: Went with the cast approach

tests/core/daemon/test_daemon.py
```diff
@@ -169,6 +257,264 @@ async def reader(ws, queue):
...
+    ],
+)
+async def test_is_running_no_services(mock_lonely_daemon, service_request, expected_result, expected_exception):
+    daemon = mock_lonely_daemon
+    if expected_exception is not None:
+        with pytest.raises(expected_exception):
+            await daemon.is_running(service_request)
+    else:
+        response = await daemon.is_running(service_request)
+        assert_is_running_response(response, expected_result)
```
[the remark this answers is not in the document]
paninaro: Yeah this works. Updated.

chia/daemon/server.py
```diff
@@ -326,6 +326,8 @@ async def handle_message(
             response = await self.stop_plotting(cast(Dict[str, Any], data))
         elif command == "stop_service":
             response = await self.stop_service(cast(Dict[str, Any], data))
+        elif command == "running_services":
+            response = await self.running_services(cast(Dict[str, Any], data))
```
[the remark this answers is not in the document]
paninaro: I removed the cast here.

altendky (approved):
thanks!

xdustinface (approved):

tests/core/daemon/test_daemon.py
paninaro: Punting per an offline discussion.

emlowe (approved):

PR #13301, opened 2022-09-02
Puzzle solution benchmark

arvidn:
This adds a benchmark measuring the cost of calling `get_puzzle_and_solution_for_coin()`. This test is implemented by having picked a large block from testnet, list all the spends in it, then measure the time it takes to query the puzzle and solution for them all (3 times).

This is a step to quantify an optimization to this function that I'm preparing.


trepca (approved):
nice!

PR #13307, opened 2022-09-02
60x core.data_layer for flake collection

altendky:
Draft for:
- [ ] just testing

### DIFF
--- a/tests/core/data_layer/test_data_rpc.py
+++ b/tests/core/data_layer/test_data_rpc.py
@@ -843,6 +843,15 @@ async def offer_setup_fixture(
 
         await full_node_api.farm_blocks(count=1, wallet=wallet)
 
+    message = "\n".join(
+        [
+            "We have farmed more blocks than expected which will likely result in the",
+            "store IDs being inconsistent with the hard coded test data.  This",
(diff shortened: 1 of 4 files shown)


tests/core/data_layer/test_data_rpc.py
```diff
@@ -843,6 +843,15 @@ async def offer_setup_fixture(
...
         await full_node_api.farm_blocks(count=1, wallet=wallet)
 
+    message = "\n".join(
+        [
+            "We have farmed more blocks than expected which will likely result in the",
+            "store IDs being inconsistent with the hard coded test data.  This",
+            "flakiness should get fixed but for now here is a direct explanation.",
+        ]
+    )
+    assert full_node_api.full_node.blockchain.get_peak_height() == 4, message
```
almogdepaz, line 853: @altendky did you see this fail ?
it looks like this should be constant unless we have a wrong count in two_wallet_node_and_rpc

altendky: I had a theory that it was not being consistent and was triggering other symptoms like the singleton IDs being 'shifted'.  I don't think I've seen this directly fail yet.

PR #13314, opened 2022-09-02
Ms.mempool simplify

mariano54:
Improves some code in the mempool to prepare for more extensive changes.

- Additions was not used anywhere, so it's removed
- We were running the CLVM twice for no reason, this is now removed
- Pending cache usage changed
- Validation is now separated from addition to pool, to improve code modularity
- A few files are now being checked by mypy
- change `remove_from_pool` to take a list of ids instead of 1 full item (for future changes)
- rename `add_spendbundle` to `add_spend_bundle`


chia/full_node/mempool.py
```diff
@@ -37,37 +37,37 @@ def get_min_fee_rate(self, cost: int) -> float:
...
-            del self.sorted_spends[item.fee_per_cost]
-        self.total_mempool_cost -= item.cost
-        assert self.total_mempool_cost >= 0
-
-    def add_to_pool(
-        self,
-        item: MempoolItem,
-    ):
+        for spend_bundle_id in items:
+            item = self.spends[spend_bundle_id]
```
arvidn, line 45: I'm not sure if this matters, but if `spend_bundle_id` cannot be found in `spends`, this will throw and terminate the loop. would it make sense to at least try to remove the remaining items too?

chia/full_node/mempool.py
```diff
@@ -37,37 +37,37 @@ def get_min_fee_rate(self, cost: int) -> float:
...
         Adds an item to the mempool by kicking out transactions (if it doesn't fit), in order of increasing fee per cost
         """
 
         while self.at_full_capacity(item.cost):
             # Val is Dict[hash, MempoolItem]
             fee_per_cost, val = self.sorted_spends.peekitem(index=0)
-            to_remove = list(val.values())[0]
-            self.remove_from_pool(to_remove)
+            to_remove: MempoolItem = list(val.values())[0]
+            self.remove_from_pool([to_remove.name])
```
arvidn, line 70: so we kick out transactions without checking that the new `MempoolItem` has a higher fee_per_cost?

chia/full_node/mempool_manager.py
```diff
@@ -286,42 +285,94 @@ async def pre_validate_spendbundle(
...
+            Optional[Err]: None iff status is FAILED
+        """
+
+        # Skip if already added
+        if spend_name in self.mempool.spends:
+            cost: Optional[uint64] = self.mempool.spends[spend_name].cost
+            assert cost is not None
+            return uint64(cost), MempoolInclusionStatus.SUCCESS, None
+
+        cost, status, err, item, remove_items = await self.validate_spend_bundle(new_spend, npc_result, spend_name)
```
arvidn, line 317: I have a feeling the return values can be simplified here. specifically, we just pass through `err` here. Do we really want to do the work below if the validation returns an error?
I suspect that perhaps there's some implied invariant between whether we get an error and the `status` this returns.

chia/full_node/mempool_manager.py
```diff
@@ -286,42 +285,94 @@ async def pre_validate_spendbundle(
...
+
+        Args:
+            new_spend: spend bundle to validate
+            npc_result: result of running the clvm transaction in a fake block
+            spend_name: hash of the spend bundle data, passed in as an optimization
+
+        Returns:
+            Optional[uint64]: cost of the entire transaction, None iff status is FAILED
+            MempoolInclusionStatus:  SUCCESS (should add to pool), FAILED (cannot add), and PENDING (can add later)
+            Optional[Err]: None iff status is FAILED
```
arvidn, line 349: I think you mean `Err` is *set* iff status is FAILED

chia/full_node/mempool_manager.py
```diff
@@ -286,42 +285,94 @@ async def pre_validate_spendbundle(
...
+            Optional[MempoolItem]: the item to add (None iff status is FAILED)
+            List[bytes32]: conflicting mempool items to remove empty unless status is SUCCESS
```
arvidn, line 351: I think you meant to add a "." between "remove" and "empty". It took a minute to parse :)

chia/full_node/mempool_manager.py
```diff
@@ -286,42 +285,94 @@ async def pre_validate_spendbundle(
...
+
+    async def validate_spend_bundle(
+        self,
+        new_spend: SpendBundle,
+        npc_result: NPCResult,
+        spend_name: bytes32,
+    ) -> Tuple[Optional[uint64], MempoolInclusionStatus, Optional[Err], Optional[MempoolItem], List[bytes32]]:
+        """
+        Validates new_spend with the given NPCResult, and spend_name, and the current mempool. The mempool should
+        be locked during this call (blockchain lock).
```
arvidn, line 339: If I understand correctly, the purpose of separating out `validate_spend_bundle()` from `add_spend_bundle()` is to have the validation not mutate the mempool object itself, right?

chia/full_node/mempool_manager.py
```diff
@@ -389,50 +429,32 @@ async def add_spendbundle(
...
-                return (
-                    uint64(cost),
-                    MempoolInclusionStatus.PENDING,
-                    Err.MEMPOOL_CONFLICT,
-                )
 
-        elif fail_reason:
-            return None, MempoolInclusionStatus.FAILED, fail_reason
+        if fail_reason is not None and fail_reason is not Err.MEMPOOL_CONFLICT:
+            return None, MempoolInclusionStatus.FAILED, fail_reason, None, []
```
arvidn, line 457: so, I take it you deliberately continue here in the case of `MEMPOOL_CONFLICT`. Can you comment on that?
The code below used to not be run under that condition, but now is.

chia/full_node/mempool.py
mariano54: That was checked in validation

chia/full_node/mempool.py
mariano54: good point, will do that.

chia/full_node/mempool_manager.py
mariano54: OK i'll try that

chia/full_node/mempool_manager.py
mariano54: Yes exactly. This will become more useful in the future as we have more complex logic, mutating spend bundles.

chia/full_node/mempool_manager.py
mariano54: OK, added comment

chia/full_node/mempool_manager.py
arvidn: I can't think of anything obvious. I think because of the lack of (or awkward use of) variant types in python, we just put all possible return types in a tuple, even if some are mutually exclusive.

chia/full_node/mempool_manager.py
```diff
@@ -314,26 +314,27 @@ async def add_spend_bundle(
...
-            assert cost is not None and item is not None
-            self.mempool.remove_from_pool(remove_items)
+        err, item, remove_items = await self.validate_spend_bundle(new_spend, npc_result, spend_name)
+        if err is None:
+            # No error, immediately add to mempool, after removing conflicting TXs.
+            assert item is not None
             self.mempool.add_to_pool(item)
-        elif status == MempoolInclusionStatus.PENDING:
-            assert cost is not None and item is not None
+            self.mempool.remove_from_pool(remove_items)
```
arvidn, line 322: the add and remove changed order here. I would think it's relevant, right?

arvidn: or maybe it doesn't matter anymore because of the separation of "add" and "validate"

chia/full_node/mempool_manager.py
```diff
@@ -344,19 +345,17 @@ async def validate_spend_bundle(
...
 
         Returns:
-            Optional[uint64]: cost of the entire transaction, None iff status is FAILED
-            MempoolInclusionStatus:  SUCCESS (should add to pool), FAILED (cannot add), and PENDING (can add later)
-            Optional[Err]: None iff status is FAILED
-            Optional[MempoolItem]: the item to add (None iff status is FAILED)
-            List[bytes32]: conflicting mempool items to remove empty unless status is SUCCESS
+            Optional[Err]: Err is set if we cannot add to the mempool, None if we will immediately add to mempool
+            Optional[MempoolItem]: the item to add (to mempool or pending pool)
+            List[bytes32]: conflicting mempool items to remove, if no Err
```
arvidn, line 350: nice, good idea

arvidn (approved):

trepca (approved):

chia/full_node/mempool_manager.py
```diff
@@ -540,27 +562,27 @@ async def new_peak(
...
-                    item.spend_bundle, item.npc_result, item.spend_bundle_name, item.program
-                )
+                _, result, _ = await self.add_spend_bundle(item.spend_bundle, item.npc_result, item.spend_bundle_name)
                 # If the spend bundle was confirmed or conflicting (can no longer be in mempool), it won't be
                 # successfully added to the new mempool. In this case, remove it from seen, so in the case of a reorg,
                 # it can be resubmitted
-                if result != MempoolInclusionStatus.SUCCESS:
-                    self.remove_seen(item.spend_bundle_name)
+                if result == MempoolInclusionStatus.SUCCESS:
+                    self.seen(item.spend_bundle_name)
```
trepca, line 579: why do we call this method here? it only checks if spend bundle has been seen recently so kind of no-op

chia/full_node/mempool_manager.py
```diff
@@ -540,27 +562,27 @@ async def new_peak(
...
         else:
             old_pool = self.mempool
             self.mempool = Mempool(self.mempool_max_total_cost)
+            self.seen_bundle_hashes = {}
             for item in old_pool.spends.values():
-                _, result, _ = await self.add_spendbundle(
-                    item.spend_bundle, item.npc_result, item.spend_bundle_name, item.program
-                )
+                _, result, _ = await self.add_spend_bundle(item.spend_bundle, item.npc_result, item.spend_bundle_name)
                 # If the spend bundle was confirmed or conflicting (can no longer be in mempool), it won't be
```
trepca, line 575: we probably should update this comment

chia/full_node/mempool_manager.py
mariano54: It avoids re-processing transactions that we have already processed.

trepca: but `self.seen()` doesn't change any state? there is a `add_and_maybe_pop_seen` but it's never used I think

PR #13315, opened 2022-09-02
use get_puzzle_and_solution_for_coin() from chia_rs

arvidn:
This bumps the `chia_rs` dependency to 0.1.9, which has an implementation of `get_puzzle_and_solution_for_coin()`. It also switches over `chia-blockchain` to use that function.

One noteworthy difference is that the rust implementation returns `SerializedProgram` for both the puzzle and the solution. This ends up being more efficient in most cases since we mostly just turn around and serializes these over the wallet protocol.

To quantify the speedup, the `tests/core/test_cost_calculation.py::test_get_puzzle_and_solution_for_coin_performance` benchmark has the improvement.

| benchmark | before | after | after / before |
| --- | --- | --- | --- |
| get_puzzle_and_solution_for_coin() | 8.67s | 2.86s | 32.99% |

The output from the benchmark can be found here: https://github.com/Chia-Network/chia-blockchain/runs/8163694887?check_suite_focus=true#step:10:35

The rust implementation can be found here: https://github.com/Chia-Network/chia_rs/blob/main/src/gen/get_puzzle_and_solution.rs#L48

### DIFF
--- a/chia/clvm/spend_sim.py
+++ b/chia/clvm/spend_sim.py
@@ -7,7 +7,6 @@ from typing import Optional, List, Dict, Tuple, Any, Type, TypeVar
 
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.blockchain_format.coin import Coin
-from chia.types.blockchain_format.program import Program, SerializedProgram
 from chia.types.mempool_item import MempoolItem
 from chia.util.ints import uint64, uint32
 from chia.util.hash import std_hash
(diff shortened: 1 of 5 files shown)


chia/full_node/full_node_api.py
```diff
@@ -1291,10 +1290,10 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
         if error is not None:
             return reject_msg
 
-        pz = Program.to(puzzle)
-        sol = Program.to(solution)
```
arvidn, line 1295: these are the correct type already, no need to convert

chia/clvm/spend_sim.py
```diff
@@ -386,9 +385,9 @@ async def get_puzzle_and_solution(self, coin_id: bytes32, height: uint32) -> Opt
         if error:
             return None
         else:
-            puzzle_ser: SerializedProgram = SerializedProgram.from_program(Program.to(puzzle))
-            solution_ser: SerializedProgram = SerializedProgram.from_program(Program.to(solution))
```
arvidn, line 390: these are the correct type now, no need to convert

tests/core/test_cost_calculation.py
```diff
@@ -296,7 +296,7 @@ async def test_get_puzzle_and_solution_for_coin_performance():
     # benchmark the function to pick out the puzzle and solution for a specific
     # coin
     generator = BlockGenerator(LARGE_BLOCK.transactions_generator, [], [])
-    with assert_runtime(seconds=20, label="get_puzzle_and_solution_for_coin"):
+    with assert_runtime(seconds=7, label="get_puzzle_and_solution_for_coin"):
```
arvidn, line 299: the benchmark can have a lower limit now

trepca (approved):
lgtm

Issue #13304, opened 2022-09-02
Identify table and indexiness on all SQL indexes


altendky:
`{table_name}_{column_name}_index`?  Or maybe extra `_` since table and column names can have those which leaves some ambiguity.

Thanks @neurosis69

emlowe:
Moved to jira backlog - closing GH issue

PR #13316, opened 2022-09-02
Dl wallet store index fix

neurosis69:
- renamed the indexes according to #13304
- removed index `coin_id` as there is already an internal index for the PK: sqlite_autoindex_singleton_records_1
- removed index `inner_puzzle_hash` as I found no SQL query using it
- fixed wrong indexed column for index `confirmed_at_height`, indexed column was `root`
- removed index `id` as there is already an internal index for the PK: sqlite_autoindex_launchers_1
- added index `mirrors_launcher_id_index` as there is a query [here](https://github.com/Chia-Network/chia-blockchain/blob/98c44fdd15ea0f06948a393d237142c71bd9d532/chia/data_layer/dl_wallet_store.py#L303) which would benefit from it

### DIFF
--- a/chia/data_layer/dl_wallet_store.py
+++ b/chia/data_layer/dl_wallet_store.py
@@ -83,9 +83,9 @@ class DataLayerStore:
             await conn.execute("CREATE INDEX IF NOT EXISTS singleton_records_confirmed_at_height_index on singleton_records(confirmed_at_height)")
             await conn.execute("CREATE INDEX IF NOT EXISTS singleton_records_generation_index on singl)eton_records(generation)")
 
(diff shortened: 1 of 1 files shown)


emlowe:
Close and reopen for new CI run

altendky (dismissed):
Tracking here that I looked over this, but this PR isn't going in what with the unsigned commits.  We can create new signed commits in our own PR.

neurosis69:
> Tracking here that I looked over this, but this PR isn't going in what with the unsigned commits. We can create new signed commits in our own PR.

signed my commits according to a guide from @wallentx

wjblanke:
This will go in after branch

wjblanke (approved):
aok but for main after release branch

altendky:
I'm going to draft this for now for clarity that it should not be merged yet.  I'll also kick CI for a fresh run.

PR #13324, opened 2022-09-04
Replace localhost with 127.0.0.1 for tests

altendky:
This is an effort to mitigate the flaky 404 errors (and others) in our tests.  This is intended to avoid exercising the problem in our tests and another real fix to handle it should be implemented.

From what I can tell the issue is that when we start a server using port 0 in an effort to get any unused port, enabling concurrent test execution, we then only track the port but use `localhost` as the host when connecting with the client.  I am unsure of the detailed mechanics of resolving `localhost`, but I did observe requests from a client in one test process being handled by a server in another test in another process.  When this happened I was able to confirm that the incorrectly reached server _did_ have that port open but on IPv6 `::1` while the intended server had that port for IPv4 `127.0.0.1`.  A real solution would be to track not just the port but the host and port or full address tuples (2-tuple for IPv4 and 5-tuple for IPv6, I believe).  That is...  not trivial and not quick, though I did start for a bit in https://github.com/Chia-Network/chia-blockchain/pull/13328.  This also reaches into the old `self_hostname` topic that we all were displeased with the state of.  But anyways, this hard coding of `127.0.0.1` seems to have eliminated the 404s entirely, perhaps along with other flakes where the crossing of servers still yielded a server of the expected type, hence no 404, but also could trigger failures of any number of other sorts.

I believe this is not really an issue in at least most production setups since there we will specify port numbers which will require that whatever socket is opened, IPv4 or IPv6 or both, would have that port number.

I think this is a good fix for the 404 and related issues.  Since it is both ready and minimally intrusive, I think we can consider it for merging to `main` and then cherry picking to `release/1.6.0`.  From there we can continue, with hopefully a bit less noise, addressing other flaky tests.  The test results below suggest the DataLayer and NFT tests are good candidates to start with.  Though note that DataLayer tests should be checked and worked on on `release/1.6.0`.

PR #13328, opened 2022-09-04
a whole mess of keeping hosts and ports together

altendky:
Relates to https://github.com/Chia-Network/chia-blockchain/pull/13324.

Draft for:
- [ ] This is a ginormous mess that shouldn't be done mid-release, not even on the `main` branch what with all the cross merging going on.
- [ ] This is still cheating by not keeping the full IPv6 5(?)-tuple.
- [ ] This mixes with eliminating the `self_hostname` passing all over in the tests since that was our 'shortcut' for the host.

### DIFF
--- a/chia/rpc/rpc_server.py
+++ b/chia/rpc/rpc_server.py
@@ -19,7 +19,7 @@ from chia.types.peer_info import PeerInfo
 from chia.util.byte_types import hexstr_to_bytes
 from chia.util.ints import uint16
 from chia.util.json_util import dict_to_json_str
-from chia.util.network import select_port
+from chia.util.network import Address, select_address
 from chia.util.ws_message import WsRpcMessage, create_payload, create_payload_dict, format_response, pong
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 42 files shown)


altendky:
blasting `core` in https://github.com/Chia-Network/chia-blockchain/pull/13329

arvidn:
a nit-pick; typically "address" just refers to the IP-address. address + port is referred to as "endpoint"

altendky:
As I looked it up, 'socket' was the actual answer.  I expected that to be too much for people to handle what with the name collision with objects that reference actual open things you can send bytes through.  Also, we already have `PeerInfo` which _might_ be appropriate to just use instead of a new class.  Also also, it was pointed out that the IPv6 side gives you maybe a 5-tuple though I'm not sure how you stick that all into a URL.

I expect this PR to die and maybe get replaced by something else.

https://en.m.wikipedia.org/wiki/Network_socket suggests 'socket address'.  Also referenced at https://datatracker.ietf.org/doc/html/rfc793.

PR #13329, opened 2022-09-04
Replace localhost with 127.0.0.1 in block tools config  - just `core`

altendky:
Draft for:
- [ ] just exploration


altendky:
Kicking for another run after https://github.com/Chia-Network/chia-blockchain/actions/runs/2988790233.

altendky:
Kicking for another run after https://github.com/Chia-Network/chia-blockchain/actions/runs/2988980429.

PR #13350, opened 2022-09-06
Simple script for generating mnemonics with known words

paninaro:
Since we share some test keys that are referred-to by the first word in their mnemonic, why not make it easy to generate wallets with known words. Now anyone can easily create their own "toilet" wallet.

### DIFF
--- a/tools/mnemonic_generator.py
+++ b/tools/mnemonic_generator.py
@@ -5,30 +5,59 @@
 #  $ python tools/mnemonic_generator.py exotic toilet dance
 #  exotic toilet dance <rest of mnemonic>
 
+import argparse
 import sys
 
 from bitstring import BitArray
 from chia.util.keychain import bip39_word_list, bytes_to_mnemonic
 from secrets import token_bytes
 
-if len(sys.argv) < 2:
-    print("Usage: mnemonic_generator.py <word1> <word2> ... <word23>")
-    sys.exit(1)
 
-word_list = bip39_word_list().splitlines()
-bitarray = BitArray(token_bytes(32))
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-l", "--list", action="store_true")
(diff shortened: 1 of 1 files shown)


tools/mnemonic_generator.py
```diff
@@ -0,0 +1,63 @@
...
+import sys
+
+from bitstring import BitArray
+from chia.util.keychain import bip39_word_list, bytes_to_mnemonic
+from secrets import token_bytes
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-l", "--list", action="store_true")
```
altendky, line 18: This is more of a subcommand situation since it just does something entirely different.

tools/mnemonic_generator.py
```diff
@@ -0,0 +1,63 @@
...
+
+    for i, word in enumerate(words):
+        try:
+            word_index = word_list.index(word)
+        except ValueError:
+            print(f"Word not found in BIP39 word list: {word}")
+            sys.exit(1)
+        start = i * 11
+        end = start + 11
+        bitarray[start:end] = word_index
```
altendky, line 54: I don't recall the details of the bitstring library, but in general slice assignment can extend the size of the container.

```python-console
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
>>> x = []
>>> x[3:5] = [1, 2]
>>> x
[1, 2]
```
This could happen here if we specified a full 24 words on the command line.

altendky (changes requested):
The slice assignment extension risk is the only comment I think really needs to be addressed (or my errors pointed out).  Given the 'unofficial' and 'non-critical' nature of this I think the others can just be FYIs.

I do not know the details of mnemonics generation so that portion is educational for me and I think someone else should review it for correctness.

emlowe:
Closing PR - this doesn't seem to benefit anyone in general other than "fun" - and I am not a fun person

PR #13353, opened 2022-09-07
fix datalayer upsert test flakes

altendky:
Just the upsert was having issues because we only waited for the key to be present, but it's an upsert so the only thing changing is the value...

### DIFF
--- a/tests/core/data_layer/test_data_rpc.py
+++ b/tests/core/data_layer/test_data_rpc.py
@@ -3,7 +3,7 @@ import contextlib
 import copy
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, AsyncIterator, Dict, List, Tuple
+from typing import Any, AsyncIterator, Dict, List, Optional, Tuple
 
 import pytest
 import pytest_asyncio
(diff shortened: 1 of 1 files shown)


tests/core/data_layer/test_data_rpc.py
```diff
@@ -966,16 +966,18 @@ async def process_for_data_layer_keys(
...
         try:
-            await data_layer.get_key_value_hash(store_id=store_id, key=expected_key)
+            value = await data_layer.get_value(store_id=store_id, key=expected_key)
         except Exception as e:
             # TODO: more specific exceptions...
             if "Key not found" not in str(e):
                 raise
         else:
-            break
+            if expected_value is None or value == expected_value:
```
almogdepaz, line 979: what's the expected behavior of expected_value = None but value != None ?

altendky: If the `expected_value` is not specified (is `None`), then we don't care what the `value` is.  We only want to make sure the key is present.  If `expected_value` is specified (is `bytes`), then we want to verify that the key is present and also that the value matches `expected_value`.

fchirica (approved):
LGTM

tests/core/data_layer/test_data_rpc.py
almogdepaz: ok, on first look that seems like it will hide issues, why not just specify what is expected ?
dose this case have any use ?

altendky: It is intentional.  It may not be the right choice and isn't a big cost to require the extra specificity at the only other use site.  There it is used when building up the store which uses only inserts so just checking key presence is sufficient.

mariano54 (approved):

emlowe (approved):

PR #13361, opened 2022-09-07
ask the DB to count unspent small coins

arvidn:
instead of fetching every unspent coin from the DB, just to count in python. It's a lot faster.

main:

chia-hotspot-98-107-main-spam-filter

patched:

chia-hotspot-78-81-patched

### DIFF
--- a/chia/wallet/wallet_coin_store.py
+++ b/chia/wallet/wallet_coin_store.py
@@ -4,7 +4,7 @@ import sqlite3
 
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.sized_bytes import bytes32
-from chia.util.db_wrapper import DBWrapper2
+from chia.util.db_wrapper import DBWrapper2, execute_fetchone
 from chia.util.ints import uint32, uint64
 from chia.wallet.util.wallet_types import WalletType
 from chia.wallet.wallet_coin_record import WalletCoinRecord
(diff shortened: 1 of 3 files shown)


chia/wallet/wallet_coin_store.py
```diff
@@ -51,8 +51,18 @@ async def create(cls, wrapper: DBWrapper2):
 
             await conn.execute("CREATE INDEX IF NOT EXISTS wallet_id on coin_record(wallet_id)")
 
+            await conn.execute("CREATE INDEX IF NOT EXISTS coin_amount on coin_record(amount)")
```
mariano54, line 54: could this index slow down the DB? or is it just more space on disk? is it a large amount of space for mainnet?

arvidn: it's more space and it slows down inserts

arvidn: this only scales by the size of the wallet though, not the size of the full blockchain. But it's not ideal, a better solution would be to slightly alter the definition of the tx limit to something that's easier to query or maintain in memory.

mariano54: oh right I forgot this was wallet. not a big deal then

mariano54 (dismissed):

mariano54 (approved):

PR #13369, opened 2022-09-07
Ignore FileNotFoundError when checking SSL file permissions

altendky:


hoffmang9 (approved):
lgtm and solves the issue for me

emlowe (approved):
Approved - not required to print a message about permissions on non-existent files

PR #13371, opened 2022-09-07
Add a try/except for individual coin states

### LINKED ISSUE #13321: Wallet syncing in an infinity loop [[Bug]
### What happened?

Hi,

My wallet has over 6k NFTs, and suddenly it's not able to get synced. I'm getting always: "ERROR    Error adding states" and my it tries to "perform_atomic_rollback" without success in an infinity loop...

I tried deleting my db and differents PCs with fresh installation, but the issue remain the same on every system.

Now, I'm not able to mint new NFTs or at least transfer my did because never get synced.

### Version

1.5.1

Linux

CLI

### Relevant log output

```shell
2022-08-31T18:18:39.971 wallet chia.wallet.wallet_state_manager: INFO     Setting tx_id: 02a9c0cf3d2d97532f8f48029d73a819c0dcee8b085fc1904f21cbbb95b0a7ba to confirmed
2022-08-31T18:18:39.988 wallet DID did:chia:13q0lkt30jysru9078xplncw2x8xm8m3kjsxgk2fdt3w7fa05aymqzqn86a: INFO     DID wallet has been notified that coin was added: b'\x0e-5_\x15f\x9aq\xb47\xdb\x8dpoA\xa2\x9f\x15\xfc\xc5Nu/\xb8\xb5\x85\xccL"&o\xdd':Coin { parent_coin_info: 4ffd9924bec8d58b2236e48482253bc2ac35952cfb364883a27e33bc13e3d024, puzzle_hash: 913064fff26074f8400b70c1fb8a22bdb99c20709e8874b64679d55734b76e74, amount: 1000000 }
2022-08-31T18:18:40.493 wallet DID did:chia:13q0lkt30jysru9078xplncw2x8xm8m3kjsxgk2fdt3w7fa05aymqzqn86a: INFO     Adding parent b'\x0e-5_\x15f\x9aq\xb47\xdb\x8dpoA\xa2\x9f\x15\xfc\xc5Nu/\xb8\xb5\x85\xccL"&o\xdd': {'amount': 1000000,
 'inner_puzzle_hash': '0x5b6d7e41c46099b3bff4d0bfa40f385ff46653f6f7692be7890baf6ca7128db2',
 'parent_name': '0x4ffd9924bec8d58b2236e48482253bc2ac35952cfb364883a27e33bc13e3d024'}
2022-08-31T18:18:41.002 wallet chia.wallet.wallet_node    : ERROR    Error adding states..  Traceback (most recent call last):
  File "chia/wallet/wallet_node.py", line 791, in receive_state_from_peer
  File "chia/wallet/wallet_state_manager.py", line 934, in new_coin_state
  File "chia/wallet/wallet_state_manager.py", line 1330, in coin_added
  File "chia/wallet/did_wallet/did_wallet.py", line 386, in coin_added
AssertionError

2022-08-31T18:18:42.223 wallet chia.wallet.wallet_blockchain: INFO     Peak set to: 2479065 timestamp: 1661983771
2022-08-31T18:18:42.225 wallet chia.wallet.wallet_node    : INFO     Starting sync trusted: True to peer 192.168.21.123
2022-08-31T18:18:42.226 wallet chia.wallet.wallet_node    : INFO     perform_atomic_rollback to 2467068
```

Quexington:
This looks like a big change but really it's just a big tab to put a large chunk of code inside a try/except.  Currently, if a single state fails during new_coin_state, an exception will be raised to the wallet node which will log an error and then return that it failed to receive the state at all (even though my understanding is that DB changes may have already been made).  This potentially results in not being able to progress past the block with the error causing state in it.  Instead, we should catch exceptions for individual states and just move past them gracefully while logging that we failed to add them.


trepca (dismissed):

mariano54 (changes requested):
Discussed with Quex. This change breaks trusted sync invariants. Trusted sync should always show the right balance if the sync finished, and it never retries anything.

chia/wallet/wallet_state_manager.py
```diff
@@ -932,331 +937,355 @@ async def new_coin_state(
...
-                            coin_name,
-                        )
-                    await self.add_interested_coin_ids([coin_name])
+                            await self.add_interested_coin_ids([coin_name])
 
+                    else:
+                        raise RuntimeError("All cases already handled")  # Logic error, all cases handled
+            except Exception as e:
+                tb = traceback.format_exc()
+                self.log.error(f"Error adding state.. {e} {tb}")
```
trepca, line 1283: lets change this to `self.log.exception`

chia/wallet/wallet_node.py
```diff
@@ -377,6 +382,42 @@ async def _messages_to_resend(self) -> List[Tuple[Message, Set[bytes32]]]:
...
+                            peer = matching_peer[0]
+                        async with self.wallet_state_manager.db_wrapper.writer():
+                            self.log.info(f"retrying coin_state: {state}")
+                            try:
+                                await self.wallet_state_manager.new_coin_state(
+                                    [state], peer, None if fork_height == 0 else fork_height, False
+                                )
+                            except Exception as e:
+                                tb = traceback.format_exc()
+                                self.log.error(f"Exception while adding states.. : {e} {tb}")
```
trepca, line 414: `self.log.exception`

Quexington:
The behavior of this PR now is to try every coin state in a DB transaction and try/except block.  If it fails for whatever reason, the DB is rolled back for that state.  If the exception is of a specfic type (network or DB error) then it will add the coin state into a store to retry periodically.  If the exception is any other type, the coin state is skipped.  I've tested this on a wallet with many thousands of coin states and I've gotten the same balance every time whether trusted or untrusted.  I've also tried it in a wallet with what used to be a fatal error, and the sync moves past it after this patch just fine.

chia/wallet/wallet_node.py
```diff
@@ -381,6 +387,41 @@ async def _messages_to_resend(self) -> List[Tuple[Message, Set[bytes32]]]:
...
+                    self.last_state_retry_time = current_time
+                    if self.wallet_state_manager is None:
+                        continue
+                    states_to_retry = await self.wallet_state_manager.retry_store.get_all_states_to_retry()
+                    for state, peer_id, fork_height in states_to_retry:
+                        matching_peer = tuple(
+                            p for p in self.server.get_full_node_connections() if p.peer_node_id == peer_id
+                        )
+                        if len(matching_peer) == 0:
+                            peer = self.get_full_node_peer()
```
mariano54, line 405: what if there are no peers connected?

chia/wallet/wallet_retry_store.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        Adds object to key val store. Obj MUST support __bytes__ and bytes() methods.
+        """
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
+            cursor = await conn.execute(
+                "INSERT OR IGNORE INTO retry_store VALUES(?, ?, ?)",
+                (bytes(state), peer_id, 0 if fork_height is None else fork_height),
+            )
+            await cursor.close()
+
```
mariano54, line 48: Shouldn't we also have a rollback? What if we end up retrying infinitely, something that is already on chain, because of a reorg

mariano54 (commented):
Should we also raise PeerRequestException on the other types of failures?
For example, request_header_blocks, request_additions, etc etc

chia/wallet/wallet_retry_store.py
Quexington: If it retries and it fails because it's already on chain (_if_ it fails, I think it might just make redundant updates) it will not be one of the two blessed exceptions and will therefore be removed.

Quexington:
> Should we also raise PeerRequestException on the other types of failures?
> For example, request_header_blocks, request_additions, etc etc
This PR is focused on improving the flakiness of coin states, we don't make these requests from wallet state manager or any of the wallets at all.  We certainly could but it doesn't help or hurt us for the problem we're trying to fix here.

chia/wallet/wallet_state_manager.py
```diff
@@ -934,335 +937,363 @@ async def new_coin_state(
...
-            else:
-                raise RuntimeError("All cases already handled")  # Logic error, all cases handled
+                    else:
+                        raise RuntimeError("All cases already handled")  # Logic error, all cases handled
+            except Exception as e:
+                self.log.exception(f"Error adding state... {e}")
+                if isinstance(e, PeerRequestException) or isinstance(e, aiosqlite.Error):
+                    await self.retry_store.add_state(coin_state, peer.peer_node_id, fork_height)
+                else:
+                    await self.retry_store.remove_state(coin_state)
```
trepca, line 1295: it's removing the retry state when it starts the `new_coin_state` update why remove it again?

Quexington: The first removal is in the transaction that gets rolled back if there's a raise.  We need to add another command to remove it here in case an exception is thrown but it's not one of the blessed ones.

trepca (approved):

mariano54 (approved):

PR #13374, opened 2022-09-07
DL: stop using fee config setting and remove from initial config

### LINKED ISSUE #13365: Remove or Set to 0 DataLayer default fee

emlowe:
Stop using DataLayer fee setting in config and remove the setting from the initial config. Note any previous installs will still have a fee setting, but it is ignored
Tests are updated to specify the old default fee in order to keep reference offer data the same

Fixes #13365 

Draft for:
- [x] Discussion with Michael.


altendky (changes requested):
Need to discuss with Michael.

altendky (approved):

TheLastCicada (approved):
I think this is the right way to go.  No known datalayer apps rely on the default fee.

wjblanke (approved):
aok

PR #13376, opened 2022-09-08
Recompress generators

richardkiss:


tools/analyze-chain.py
```diff
@@ -49,59 +49,74 @@ def run_gen(env_data: bytes, block_program_args: bytes, flags: int):
...
+
+
 @click.command()
 @click.argument("file", type=click.Path(), required=True)
 @click.option(
     "--mempool-mode", default=False, is_flag=True, help="execute all block generators in the strict mempool mode"
 )
-def main(file: Path, mempool_mode: bool):
+@click.option("--precache", default=False, is_flag=True, help="pre-cache block heights")
+@click.option("--start", default=225000, help="first block to examine")
```
richardkiss, line 73: Blocks before this point have no generators on mainnet.

tools/analyze-chain.py
```diff
@@ -49,59 +49,74 @@ def run_gen(env_data: bytes, block_program_args: bytes, flags: int):
         return 117, None, 0
 
 
+def generator_references_from_indices(c, height_to_hash_cache, indices: List[int]) -> List[bytes]:
+    ref_block_blobs = []
+    for h in indices:
+        if h not in height_to_hash_cache:
+            ref = c.execute("SELECT header_hash FROM full_blocks WHERE height=?", (h,))
+            height_to_hash_cache[h] = ref.fetchone()[0]
```
arvidn, line 57: the first block may not be the one we want. the first might be an orphan

tools/analyze-chain.py
```diff
@@ -49,59 +49,74 @@ def run_gen(env_data: bytes, block_program_args: bytes, flags: int):
...
+            ref.close()
+
+        ref = c.execute("SELECT block FROM full_blocks WHERE header_hash=?", (height_to_hash_cache[h],))
```
arvidn, line 60: I believe you can just change the condition here to be `height=? AND in_main_chain=1` (and pass in `h`). Then you won't need the height-to-hash cache

tools/analyze-chain.py
```diff
@@ -49,59 +49,74 @@ def run_gen(env_data: bytes, block_program_args: bytes, flags: int):
...
+    rows = c.execute("SELECT max(height) FROM full_blocks")
+    (max_height,) = next(rows)
+    end_height = max_height if end is None else min(max_height, end)
 
-    height_to_hash: List[bytes] = []
+    if precache:
+        rows = c.execute(f"SELECT header_hash, height FROM full_blocks WHERE height <= {end_height}")
+        for hh, height in rows:
+            height_to_hash_cache[height] = hh
+            print(f"precaching {height}         \r", end="")
```
arvidn, line 88: I think we should remove the `height_to_hash_cache` entirely. I don't think we need it anymore, if we can just require a v2 blockchain database instead

tools/analyze-chain.py
```diff
@@ -49,59 +49,74 @@ def run_gen(env_data: bytes, block_program_args: bytes, flags: int):
...
+
+    rows = c.execute(
+        f"SELECT header_hash, height, block FROM full_blocks "
+        f"WHERE height >= {start} and height <= {end_height} ORDER BY height"
```
arvidn, line 92: this would have to also have the `AND in_main_chain=1` condition added to it

tools/analyze-chain.py
richardkiss: Seems to work, seems to be fast still, so this is a good change. Greatly simplifies the code.

arvidn:
looks good

tools/compression_v2_stats.py
```diff
@@ -0,0 +1,77 @@
...
+while the current directory is the root `chia-blockchain` folder.
+
+It will iterate over transaction blocks, recompress them and print the new cost ("agony rating")
+along with a relative percentage, where 100% is par and smaller is better.
+"""
+
+import time
+from typing import List
+
+from chia_rs import compression  # type: ignore[attr-defined]
```
arvidn, line 20: is this warning about `chia_rs` being imported as well, somewhere?

arvidn (commented):
It would be great to run this in CI, since the only (reliable) way to know whether python code is broken is to run it. There is a test database you could run it against at `tests/tools/test-blockchain-db.sqlite`.

This PR really *should* be red, since it uses functionality that isn't exposed by `chia_rs` yet.

Also, what's the results when you run this with the compression?

arvidn (dismissed):
this code will likely bitrot pretty quickly without being exercised by CI. maybe that's fine and we can fix it up if we need it again in the future

trepca (dismissed):

trepca (dismissed):

tools/compression_v2_stats.py
richardkiss: The wheel I installed didn't have updated mypy headers. I'll have to figure out how to do that.

altendky:
```console
$ venv/bin/python -c 'from chia_rs import compression'
  File "<string>", line 1, in <module>
ImportError: cannot import name 'compression' from 'chia_rs' (/home/altendky/repos/chia-blockchain/venv/lib/python3.10/site-packages/chia_rs/__init__.py)
```

altendky (dismissed):
I think Richard said they were going to fix this, but to avoid premature merging, here's a change request.

richardkiss:
> I think Richard said they were going to fix this, but to avoid premature merging, here's a change request.

Right. At the moment, `chia_rs` doesn't have the `compression` submodule. There is a separate PR for that over there.

tools/compression_v2_stats.py
richardkiss: Correct, this is a feature preview. The compression module only exists in a branch of `chia_rs` at the moment.

altendky:
This is broken.  It's not clear to me why it should get merged that way but I've gotten multiple DMs about this so apparently it is important and urgent?  It seems like at the very least there should be direct acknowledgement and explanation here and in the code.  Is there a reason we can't wait until this is not broken?  If we want certain people testing this then we could update the chia_rs dependency to refer to the needed branch of chia_rs until it is merged and released.

richardkiss:
@altendky I've removed the file with the problem. Should be uncontroversial to merge now.

trepca (approved):

arvidn (approved):

PR #13378, opened 2022-09-08
Ms.no hang

mariano54:
Based on the experiments from:
https://github.com/Chia-Network/chia-blockchain/pull/13087

It seems to fix the two issues:
- (Maybe) Hanging on ubuntu (and maybe other platforms)
- The windows Worker crash (by not running that test on windows)

### DIFF
--- a/tests/core/test_farmer_harvester_rpc.py
+++ b/tests/core/test_farmer_harvester_rpc.py
@@ -12,6 +12,7 @@ from typing import Any, Awaitable, Callable, Dict, List, Union, cast
 
 import pytest
 import pytest_asyncio
+from _pytest.outcomes import xfail
 
 from chia.consensus.coinbase import create_puzzlehash_for_pk
 from chia.farmer.farmer import Farmer
@@ -520,6 +521,7 @@ async def test_farmer_get_harvester_plots_endpoints(
 
 
 @pytest.mark.asyncio
+@pytest.mark.skip("This test causes hangs occasionally. TODO: fix this.")
 async def test_harvester_add_plot_directory(harvester_farmer_environment) -> None:
     (
         farmer_service,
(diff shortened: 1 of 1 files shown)


tests/core/test_farmer_harvester_rpc.py
```diff
@@ -414,6 +416,9 @@ async def test_farmer_get_harvester_plots_endpoints(
     reverse: bool,
     expected_plot_count: int,
 ) -> None:
+    if platform.system() == "Windows":
+        # Windows has an issue on Github CI with crashing workers
+        return
```
altendky, line 421: Decorate with `@pytest.mark.skipif(sys.platform == "win32", reason="avoiding crashes on windows until we fix this").  Or whatever message.

tests/core/test_farmer_harvester_rpc.py
```diff
@@ -534,7 +540,8 @@ async def assert_added(path: Path) -> None:
         await harvester_rpc_client.fetch("add_plot_directory", {})
 
     root_path = harvester_service._node.root_path
-    test_path = Path(root_path / "test_path").resolve()
+    random_name = token_bytes(32).hex()
+    test_path = Path(root_path / f"test_path_{random_name}").resolve()
```
altendky, line 544: `Path()`ing a `/` joined `Path` object is redundant.  Could just drop the `Path` though I would write it as:

```suggestion
    test_path = root_path.joinpath(f"test_path_{random_name}").resolve()
```

Though really if we want a temporary directory we shouldn't shove it into the root path at all.  `tmp_path` is a fixture for getting a temp dir for the test.  Then don't clean them up.  pytest manages that for you, albeit that it keeps a few runs around before deleting the old temporary directories.

tests/core/test_farmer_harvester_rpc.py
```diff
@@ -1,10 +1,12 @@
 import dataclasses
 import logging
 import operator
+import sys
 import time
 from math import ceil
 from os import mkdir
 from pathlib import Path
+from secrets import token_bytes
```
altendky, line 9:
```suggestion
```

altendky (approved):

mariano54:
Please approve @emlowe @wjblanke

emlowe (approved):

PR #13383, opened 2022-09-08
Fix wrapper test, release

almogdepaz:
#13000 into 1.6.0

since it flakes https://github.com/Chia-Network/chia-blockchain/runs/8250240811?check_suite_focus=true

### DIFF
--- a/tests/core/util/test_db_wrapper.py
+++ b/tests/core/util/test_db_wrapper.py
@@ -23,6 +23,20 @@ async def increment_counter(db_wrapper: DBWrapper2) -> None:
         await connection.execute("UPDATE counter SET value = :value", {"value": new_value})
 
 
+async def decrement_counter(db_wrapper: DBWrapper2) -> None:
+    async with db_wrapper.writer_maybe_transaction() as connection:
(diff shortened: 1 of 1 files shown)


tests/core/util/test_db_wrapper.py
```diff
@@ -209,17 +223,25 @@ async def test_mixed_readers_writers(acquire_outside: bool) -> None:
...
+            async with connection.execute("SELECT value FROM counter") as cursor:
+                row = await cursor.fetchone()
+                assert row is not None
+                assert row[0] == 1
+
+    # it's possible all increments or all decrements are run first
     assert len(values) == concurrent_task_count
+    for v in values:
+        assert v > -99
+        assert v <= 100
```
arvidn, line 247: these constants seem to assume we loop 100 times above, but we loop 200 times. So, there's a small risk these may fail

arvidn (approved):

altendky:
I wouldn't bother with bringing this over to release.  I don't believe this is flaking enough to cause significant hassle.  I just want DataLayer test fixes here on release and other test fixes if the failures are frequent enough to keep us from being able to process important fixes if needed.

PR #13388, opened 2022-09-08
fill_missing_services=True for data cli

altendky:


altendky:
Untested, but my guess for that CLI error.

```
$ ./chia data create_data_store -m 1000
Exception from 'data': 'data_layer'
```

emlowe (approved):
Tested and this works for me

wjblanke (approved):
aok

PR #13390, opened 2022-09-08
Update test_nft_wallet.py

altendky:
- `__tracebackhide__` provides more clear tracebacks on failure when writing assert-like helpers
- `coroutine` is the name of the type of object you get back after calling an async def
- no need to repeat the timing condition after it has already failed


trepca (approved):
thanks

PR #13394, opened 2022-09-08
[Draft] Retry network requests infinitely if trusted sync

Quexington:
This is an alternative to https://github.com/Chia-Network/chia-blockchain/pull/13371 where instead of retrying every bad state infinitely, we just retry the network request that the wallet is making infinitely until we get it.  Any bad state that gets added otherwise is skipped.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -821,7 +821,7 @@ class WalletNode:
         self, peer: WSChiaConnection, header_block: HeaderBlock, request_time: uint64
     ) -> Optional[uint64]:
         # Get last timestamp
-        last_tx: Optional[HeaderBlock] = await fetch_last_tx_from_peer(header_block.height, peer)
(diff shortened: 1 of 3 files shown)


chia/wallet/wallet_node.py
```diff
@@ -1539,11 +1539,19 @@ async def validate_block_inclusion(
...
-        )
-        if solution_response is None or not isinstance(solution_response, wallet_protocol.RespondPuzzleSolution):
-            raise ValueError(f"Was not able to obtain solution {solution_response}")
+        while True:
+            solution_response = await peer.request_puzzle_solution(
+                wallet_protocol.RequestPuzzleSolution(coin.name(), height)
+            )
+            if solution_response is None or not isinstance(solution_response, wallet_protocol.RespondPuzzleSolution):
+                if self.is_trusted(peer):
+                    await asyncio.sleep(0.5)
```
trepca, line 1548: shouldn't we stop at some point and tell the user they should fix their full node?

PR #13399, opened 2022-09-09
Heap profiler

arvidn:
The profiler, like the CPU profiler, is enabled for either the wallet or full node in `config.yaml`.

example output:

chia-heap-41

### DIFF
--- a/chia/util/memory_profiler.py
+++ b/chia/util/memory_profiler.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 import asyncio
 import logging
 import pathlib
-import time
 import tracemalloc
+from datetime import datetime
 from typing import Dict, List, Optional, Set
 
 from chia.util.path import path_from_root
--- a/chia/util/memory_profiler.py
+++ b/chia/util/memory_profiler.py
@@ -12,7 +12,7 @@ from chia.util.path import path_from_root
 
 async def mem_profile_task(root_path: pathlib.Path, service: str, log: logging.Logger) -> None:
 
-    profile_dir = path_from_root(root_path, f"memory-profile-{service}-{int(time.time())}")
+    profile_dir = path_from_root(root_path, f"memory-profile-{service}") / datetime.now().strftime("%Y-%m-%d_%H:%M:%S")
     log.info("Starting memory profiler. saving to %s" % profile_dir)
     profile_dir.mkdir(parents=True, exist_ok=True)
 
--- a/chia/util/memory_profiler.py
+++ b/chia/util/memory_profiler.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import pathlib
+import time
 import tracemalloc
 from typing import Dict, List, Optional, Set
 
@@ -11,7 +12,7 @@ from chia.util.path import path_from_root
 
 async def mem_profile_task(root_path: pathlib.Path, service: str, log: logging.Logger) -> None:
 
-    profile_dir = path_from_root(root_path, f"memory-profile-{service}")
+    profile_dir = path_from_root(root_path, f"memory-profile-{service}-{int(time.time())}")
     log.info("Starting memory profiler. saving to %s" % profile_dir)
     profile_dir.mkdir(parents=True, exist_ok=True)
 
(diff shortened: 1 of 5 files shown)


xdustinface (changes requested):
Nice! Few comments and also see this branch https://github.com/xdustinface/chia-blockchain/tree/suggestions-13399 (if you want to pick those commits, i can PR them separately later if not). It makes use of `click` and moves the analyze parts into `tools` and the actual profiling task into `chia.util.profile` (there is still the analyze stuff for CPU in there but that should be moved into `tools` in a separate PR).

chia/util/memory_profiler.py
```diff
@@ -0,0 +1,244 @@
...
+import pathlib
+import tracemalloc
+from typing import Dict, List, Optional, Set
+
+from chia.util.path import path_from_root
+
+
+async def mem_profile_task(root_path: pathlib.Path, service: str, log: logging.Logger) -> None:
+
+    profile_dir = path_from_root(root_path, f"memory-profile-{service}")
```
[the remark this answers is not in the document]
arvidn: good idea

chia/util/memory_profiler.py
```diff
@@ -0,0 +1,245 @@
...
+                current = all_lines[line]
+                if " def " in current or current.startswith("def "):
+                    return current.split("def ")[1].split("(")[0]
+            return file.rsplit("/", 1)[1]
+        except Exception:
+            return "<unknown>"
+
+    def analyze_slot(slot: int) -> None:
+        file = str(profile_dir / f"heap-{slot:05d}.profile")
+        output_file = f"chia-heap-{slot}"
```
xdustinface, line 105:
```suggestion
        output_file = str(profile_dir / f"heap-{slot:05d}")
```

This lets the resulting files end up next to their input files in the profile dir.

chia/util/memory_profiler.py
```diff
@@ -0,0 +1,245 @@
...
+import time
+import tracemalloc
+from typing import Dict, List, Optional, Set
+
+from chia.util.path import path_from_root
+
+
+async def mem_profile_task(root_path: pathlib.Path, service: str, log: logging.Logger) -> None:
+
+    profile_dir = path_from_root(root_path, f"memory-profile-{service}-{int(time.time())}")
```
xdustinface, line 15:
```suggestion
    profile_dir = path_from_root(root_path, f"memory-profile-{service}") / datetime.now().strftime("%Y-%m-%d_%H:%M:%S")
```

Maybe ^ ? This way its easier to understand when the profiling happened only by looking at the folder name and also it collects the results inside `memory-profile-<service>` and doesn't clutter the root dir.

chia/util/memory_profiler.py
```diff
@@ -0,0 +1,245 @@
+from __future__ import annotations
+
+import asyncio
+import logging
+import pathlib
+import time
+import tracemalloc
+from typing import Dict, List, Optional, Set
```
xdustinface, line 8:
```suggestion
import tracemalloc
from datetime import datetime
from typing import Dict, List, Optional, Set
```

xdustinface (changes requested):

chia/util/memory_profiler.py
arvidn: I don't have a strong opinion about this. my workflow of using these is that the resulting png file is temporary anyway and it's easy to open from the terminal's cwd. It's possible this workflow could be improved, and certainly when running A/B tests it's good to have labels (not just timestamps) of which branch was tested.

It's not obvious that the `~/.chia/mainnet/` folder is the best place to put the output of the analysis though. I would think it might be more appropriate in a directory under cwd.

arvidn: I think it would be reasonable to improve this workflow as part of, or after, the tools have been moved out of the production code

xdustinface: > It's not obvious that the `~/.chia/mainnet/` folder is the best place to put the output of the analysis though.

Well, we also store the profiling data there currently so imo we should either move the profiling data to where the results are currently or the results to where the profiling data is currently?

chia/util/memory_profiler.py
arvidn: I'm actually increasingly preferring overwriting the previous profiles. it makes it simpler to reuse command lines in your shell history, and avoid entering new timestamps for every run.

xdustinface: Hm, i think both ways have its pros, right. However, if we overwrite we should at least clear the old results on start. Im not sure whats the best but i feel like keeping old results might be a good thing.

xdustinface (approved):

emlowe (approved):

PR #13400, opened 2022-09-09
separate process global setup for daemon

altendky:
This separates the process global setup, just signals for now, so that it can be not called when running tests.  This form is modeled after the same behavior on the the `Service` class.

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -169,15 +169,6 @@ class WebSocketServer:
                 ssl.OPENSSL_VERSION,
             )
 
-        def master_close_cb():
-            asyncio.create_task(self.stop())
-
-        try:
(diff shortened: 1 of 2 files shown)


arvidn:
does this make it cleaner to shut down tests?

arvidn (approved):

altendky:
I didn't explore what observable effects it had.  Just made it more consistent with the other code, more correct I believe, and observed that it did _not_ fix the hangs.  `:[`

altendky:
Well, I guess there was the inspiration for this.  Running locally I ran into an exception about setting signals outside the primary thread so it should fix that, but I've only seen that once.  Running outside the main thread is a thing xdist does to us sometimes and has caused other issues here and elsewhere and everyone wants xdist to not do that but I guess it's down in their execnet dependency and...  so on and so on.

emlowe (approved):

PR #13405, opened 2022-09-09
run expensive API requests in separate thread

arvidn:
This moves two expensive API request calls into a 1-thread `ThreadPoolExecutor` in `full_node_api`.

1. request puzzle and solution from coin
2. get block header

Both of these are expensive because they have to run the block program (and pull referenced blocks from DB, but those are async). Running the block generator program is currently done in the main thread, blocking the main loop.

### DIFF
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -3,6 +3,7 @@ import dataclasses
 import logging
 import time
 import traceback
+import functools
 from secrets import token_bytes
 from typing import Dict, List, Optional, Tuple, Set
 
@@ -47,10 +48,11 @@ from chia.types.transaction_queue_entry import TransactionQueueEntry
 from chia.types.unfinished_block import UnfinishedBlock
 from chia.util.api_decorators import api_request, peer_required, bytes_required, execute_task, reply_type
 from chia.util.full_block_utils import header_block_from_block
-from chia.util.generator_tools import get_block_header
+from chia.util.generator_tools import get_block_header, tx_removals_and_additions
 from chia.util.hash import std_hash
 from chia.util.ints import uint8, uint32, uint64, uint128
 from chia.util.merkle_set import MerkleSet
+from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions
 
 
 class FullNodeAPI:
@@ -1106,15 +1108,39 @@ class FullNodeAPI:
             msg = make_msg(ProtocolMessageTypes.reject_header_request, RejectHeaderRequest(request.height))
             return msg
         block: Optional[FullBlock] = await self.full_node.block_store.get_full_block(header_hash)
-        if block is not None:
-            tx_removals, tx_additions, _ = await self.full_node.blockchain.get_tx_removals_and_additions(block)
-            header_block = get_block_header(block, tx_additions, tx_removals)
-            msg = make_msg(
-                ProtocolMessageTypes.respond_block_header,
-                wallet_protocol.RespondBlockHeader(header_block),
+        if block is None:
+            return None
+
+        tx_removals: List[bytes32] = []
+        tx_additions: List[Coin] = []
+
+        if block.transactions_generator is not None:
+
+            block_generator: Optional[BlockGenerator] = await self.full_node.blockchain.get_block_generator(block)
+            # get_block_generator() returns None in case the block we specify
+            # does not have a generator (i.e. is not a transaction block).
+            # in this case we've already made sure `block` does have a
+            # transactions_generator, so the block_generator should always be set
+            assert block_generator is not None, "failed to get block_generator for tx-block"
+
+            npc_result = await asyncio.get_running_loop().run_in_executor(
+                self.executor,
+                functools.partial(
+                    get_name_puzzle_conditions,
+                    block_generator,
+                    self.full_node.constants.MAX_BLOCK_COST_CLVM,
+                    cost_per_byte=self.full_node.constants.COST_PER_BYTE,
+                    mempool_mode=False,
+                ),
             )
-            return msg
-        return None
+
+            tx_removals, tx_additions = tx_removals_and_additions(npc_result.conds)
+        header_block = get_block_header(block, tx_additions, tx_removals)
+        msg = make_msg(
+            ProtocolMessageTypes.respond_block_header,
+            wallet_protocol.RespondBlockHeader(header_block),
+        )
+        return msg
 
     @api_request
     async def request_additions(self, request: wallet_protocol.RequestAdditions) -> Optional[Message]:
@@ -1298,10 +1324,9 @@ class FullNodeAPI:
 
         block_generator: Optional[BlockGenerator] = await self.full_node.blockchain.get_block_generator(block)
         assert block_generator is not None
-        error, puzzle, solution = await asyncio.get_running_loop().run_in_executor(self.executor,
-            get_puzzle_and_solution_for_coin,
-            block_generator,
-            coin_record.coin)
+        error, puzzle, solution = await asyncio.get_running_loop().run_in_executor(
+            self.executor, get_puzzle_and_solution_for_coin, block_generator, coin_record.coin
+        )
 
         if error is not None:
             return reject_msg
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -29,6 +29,7 @@ from chia.protocols.wallet_protocol import (
     RespondSESInfo,
 )
 from chia.server.server import ChiaServer
+from concurrent.futures import ThreadPoolExecutor
 from chia.types.block_protocol import BlockInfo
 from chia.server.outbound_message import Message, make_msg
 from chia.types.blockchain_format.coin import Coin, hash_coin_ids
@@ -54,9 +55,11 @@ from chia.util.merkle_set import MerkleSet
 
 class FullNodeAPI:
     full_node: FullNode
+    executor: ThreadPoolExecutor
 
     def __init__(self, full_node: FullNode) -> None:
         self.full_node = full_node
+        self.executor = ThreadPoolExecutor(max_workers=1)
 
     @property
     def server(self) -> ChiaServer:
@@ -1295,7 +1298,10 @@ class FullNodeAPI:
 
         block_generator: Optional[BlockGenerator] = await self.full_node.blockchain.get_block_generator(block)
         assert block_generator is not None
-        error, puzzle, solution = get_puzzle_and_solution_for_coin(block_generator, coin_record.coin)
+        error, puzzle, solution = await asyncio.get_running_loop().run_in_executor(self.executor,
+            get_puzzle_and_solution_for_coin,
+            block_generator,
+            coin_record.coin)
 
         if error is not None:
             return reject_msg
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ dependencies = [
     "chiapos==1.0.10",  # proof of space
     "clvm==0.9.7",
     "clvm_tools==0.4.5",  # Currying, Program.to, other conveniences
-    "chia_rs==0.1.9",
+    "chia_rs==0.1.10",
     "clvm-tools-rs==0.1.19",  # Rust implementation of clvm_tools' compiler
     "aiohttp==3.8.1",  # HTTP server for full node rpc
     "aiosqlite==0.17.0",  # asyncio wrapper for sqlite, to store blocks


chia/full_node/full_node_api.py
```diff
@@ -1103,15 +1107,37 @@ async def request_block_header(self, request: wallet_protocol.RequestBlockHeader
...
-            msg = make_msg(
-                ProtocolMessageTypes.respond_block_header,
-                wallet_protocol.RespondBlockHeader(header_block),
+        if block is None:
+            return None
+
+        tx_removals: List[bytes32] = []
+        tx_additions: List[Coin] = []
+
+        if block.transactions_generator:
```
altendky, line 1116: I think this is what is meant here?  Just reading as is I wasn't sure if it was `None` exclusion or skipping an empty collection or...

```suggestion
        if block.transactions_generator is not None:
```

chia/full_node/full_node_api.py
```diff
@@ -1103,15 +1107,37 @@ async def request_block_header(self, request: wallet_protocol.RequestBlockHeader
...
+
+            block_generator: Optional[BlockGenerator] = await self.full_node.blockchain.get_block_generator(block)
+            assert block_generator is not None
```
altendky, line 1119: While we're touching this, perhaps you would be willing to add an explanation of why this would be a bug?

```suggestion
            assert block_generator is not None, "explanation here"
```

chia/full_node/full_node_api.py
```diff
@@ -1103,15 +1107,37 @@ async def request_block_header(self, request: wallet_protocol.RequestBlockHeader
...
+                    self.full_node.constants.MAX_BLOCK_COST_CLVM,
+                    cost_per_byte=self.full_node.constants.COST_PER_BYTE,
+                    mempool_mode=False,
+                ),
+                block_generator,
             )
-            return msg
-        return None
+
+            assert npc_result is not None
```
altendky, line 1132: When would this return `None`?

https://github.com/Chia-Network/chia-blockchain/blob/ae99b987cc93b2a59b33cfe686a9d0bddf194c92/chia/full_node/mempool_check_conditions.py#L33-L35

chia/full_node/full_node_api.py
```diff
@@ -1103,15 +1107,37 @@ async def request_block_header(self, request: wallet_protocol.RequestBlockHeader
...
+
+            npc_result = await asyncio.get_running_loop().run_in_executor(
+                self.executor,
+                lambda x: get_name_puzzle_conditions(
+                    x,
+                    self.full_node.constants.MAX_BLOCK_COST_CLVM,
+                    cost_per_byte=self.full_node.constants.COST_PER_BYTE,
+                    mempool_mode=False,
+                ),
+                block_generator,
```
altendky, line 1129: I prefer to avoid closures.  Regardless, the lambda was an interesting mix of passing one positional arg through the executor run call which can take several and others through closures from the lambda.  Could also pass all parameters by name with `functools.partial()`.

```suggestion
                functools.partial(
                    get_name_puzzle_conditions,
                    block_generator,
                    self.full_node.constants.MAX_BLOCK_COST_CLVM,
                    cost_per_byte=self.full_node.constants.COST_PER_BYTE,
                    mempool_mode=False,
                ),
```

setup.py
```diff
@@ -8,7 +8,7 @@
     "chiapos==1.0.10",  # proof of space
     "clvm==0.9.7",
     "clvm_tools==0.4.5",  # Currying, Program.to, other conveniences
-    "chia_rs==0.1.9",
+    "chia_rs==0.1.10",
```
altendky, line 11: This seems unrelated?

chia/full_node/full_node_api.py
```diff
@@ -46,17 +47,20 @@
...
+from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions
 
 
 class FullNodeAPI:
     full_node: FullNode
+    executor: ThreadPoolExecutor
 
     def __init__(self, full_node: FullNode) -> None:
         self.full_node = full_node
+        self.executor = ThreadPoolExecutor(max_workers=1)
```
altendky, line 63: This could only leverage a single extra thread at this point?  I do think if we do a lot of these things we might prefer a single overall thread pool for the process, but I'm fine with thinking that through later.

altendky (changes requested):
Do you know what percent of time is used in Python code (just gaining pre-emption) vs. in GIL-releasing extensions (getting more CPU time)?

altendky:
Oh yeah, my other general vague...  not quite a concern but...  is breaking this function apart and doing so only in one place.  I wonder about the applicability in other places and if the original function itself should be broken into multiple pieces to avoid the repetition in reuse.  I don't raise this as a thing that I think _needs_ to change but I did want to say it to make sure it was thought about in case there are easy opportunities.

setup.py
arvidn: I meant to comment on this in the PR. the `Coin` class wasn't declared "sendable" in rust, which means it wasn't allowed to be sent to a different thread. I had to fix that in order to pass in a coin to the `get_puzzle_and_solution_for_coin()` function, with it running in a thread pool

chia/full_node/full_node_api.py
arvidn: using a single thread here was a deliberate conservative choice. We perform work on behalf of random wallets on the internet, we want to limit the resources we spend on that.

chia/full_node/full_node_api.py
arvidn: are you sure this works? if I'm reading the documentation correctly, you bind arguments from the beginning of the parameter list, not the end.

chia/full_node/full_node_api.py
arvidn: if this is a policy, we should really get flake8 to flag it

arvidn:
> Do you know what percent of time is used in Python code (just gaining pre-emption) vs. in GIL-releasing extensions (getting more CPU time)?

in the measurements I made, all the CPU usage related to these calls disappeared. I'm not sure why but I'm guessing cProfile only picks up samples from the thread it was started in.

altendky:
cProfile is indeed single threaded.  Awhile ago I used yappi for multithreaded profiling.  py-spy looks to do threads, and has some related options, but is a sampling profiler so needs more runs potentially.

chia/full_node/full_node_api.py
```diff
@@ -46,17 +47,20 @@
...
 from chia.util.merkle_set import MerkleSet
+from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions
 
 
 class FullNodeAPI:
     full_node: FullNode
+    executor: ThreadPoolExecutor
 
     def __init__(self, full_node: FullNode) -> None:
         self.full_node = full_node
```
mariano54, line 62: It might be better to put this in Full Node. Full Node Api is just a wrapper class and shouldn't have any state

chia/full_node/full_node_api.py
mariano54: Yeah agreed, it should be a policy

mariano54 (commented):
I like the idea

chia/full_node/full_node_api.py
altendky: I've got a small entry at https://gist.github.com/altendky/7c423a63d5091b2ccf2c418624317f93#idioms about this so we will end up discussing it and can explore enforcement options.

chia/full_node/full_node_api.py
altendky: I don't think I understand the concern.

https://github.com/Chia-Network/chia-blockchain/blob/d66abf04436db9129413c5ab2f5f88a9ce52418c/chia/full_node/mempool_check_conditions.py#L33-L35

So a 'regular' call would be like this, right?

```python
get_name_puzzle_conditions(
    block_generator,
    self.full_node.constants.MAX_BLOCK_COST_CLVM,
    cost_per_byte=self.full_node.constants.COST_PER_BYTE,
    mempool_mode=False,
)
```

So the same order would apply with `functools.partial()`.  Though just using names could alleviate order concerns everywhere.

```python-console
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
>>> import functools
>>> def f(a, b):
...     print(a, b)
... 
>>> p = functools.partial(f, 1, 2)
>>> p()
1 2
```

chia/full_node/full_node_api.py
arvidn: yeah, I was considering that. It's not clear cut in my mind. The API boundary of the full node isn't high level enough (today) to encompass this, I would say. There's a fair amount of work, and assembling, happening in the full_node_api.

Taken all the way, the `FullNodeAPI` perhaps should just be a protocol that `FullNode` implements (i.e. there's literally nothing left in `FullNodeAPI`). I get the impression the main purpose of the `API` classes are to provide limited views onto their underlying objects.

chia/full_node/full_node_api.py
arvidn: the concern is demonstrated by this example:
```
Python 3.9.14 (main, Sep  6 2022, 23:16:16) 
[Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
>>> import functools
>>> def f(a, b):
...     print(a, b)
... 
>>> p = functools.partial(f, 1)
>>> p(2)
1 2
```

In the code above, we need to bind the *last* 3 parameters of `get_name_puzzle_conditions()`, not the *first* 3.

arvidn: I suppose we could bind all parameters. The risk then is to make sure `block_generator` is transferred efficiently into the thread. But since this is just a thread, and not a process, presumably it's just transferred by reference, so it wouldn't be an issue.

arvidn: I'm curious; what's wrong with closures?

chia/full_node/full_node_api.py
arvidn: mypy complained in an earlier incarnation of my patch. A more interesting question though; why isn't mypy reporting this assert as redundant?

chia/full_node/full_node_api.py
altendky: Closures make reference tracking and teardown more complicated and error prone I believe and also are more likely to confuse developers when used in loops (not here, but in general) due to their late binding behavior.  If you are interested in more details about either case maybe a chat would be good.

Yes, `functools.partial()` does bind as you expected and as you see.  For that exact topic alone, fix it by passing by name even for the parameters that don't require it.  The thing is that there was no need to be passing some positional items via the executor and some by closure and the call in the lambda.  Since some are keyword only, you can't pass them via the executor since it only handles positional arguments.  So they ought have just gone all into the lambda.  But then there are even more closures, hence the partial suggestion.

While not quite the core of the topic here, the 'here is a callable and arguments will be provided later' topic is generally kind of hard.  It makes for a complicated API to figure out how to have a callable like `.run_in_executor()` take it's own positional and keyword args while also taking the positional and keyword args for the callable it is passed.  We could just handle the callable's parameters via lambdas or partials and take _none_ via the executor call.  Could have the executor take it's own and return a callable that takes the other params like `.run_in_executor(the_function, an_executor_argument)(the_function_argument_1, the_function_argument_2)`.  Could do as it does and just discourage people from using the clarity of keyword arguments thus maximizing bug risk.  For a particularly extended discussion you could check out https://github.com/python-trio/trio/issues/470.

chia/full_node/full_node_api.py
altendky: So there is a strict equality check option that will complain if you try to compare equality of unlike types.  I am not familiar offhand with such an option for `is not None` against a non-optional type.  You could dig into the docs and issues to see if it has been discussed though.

chia/full_node/full_node_api.py
altendky: Your concern about order is a bit more critical when it comes to a map-like function where the callable may be called multiple times with different parameters and you can't include _all_ parameters in the partial.  In that case, yes, you have to deal with the order when the map-like can't handle keyword arguments.  I would default to using keyword arguments to create the partial.

chia/full_node/full_node_api.py
```diff
@@ -1295,7 +1324,9 @@ async def request_puzzle_solution(self, request: wallet_protocol.RequestPuzzleSo
 
         block_generator: Optional[BlockGenerator] = await self.full_node.blockchain.get_block_generator(block)
         assert block_generator is not None
-        error, puzzle, solution = get_puzzle_and_solution_for_coin(block_generator, coin_record.coin)
+        error, puzzle, solution = await asyncio.get_running_loop().run_in_executor(
+            self.executor, get_puzzle_and_solution_for_coin, block_generator, coin_record.coin
```
altendky, line 1328: Consider consistency with a partial here?  Even kwargs for all params in each case?  `:]`

altendky (approved):
All remaining points are either general discussion or non-mandatory change suggestions.  Feel free to close my conversations when they are complete.

chia/full_node/full_node_api.py
mariano54: They used to be both together, but I split them up because it was getting confusing to see where the API methods were. It's true that the boundaries are not clear cut though..

mariano54 (approved):

PR #13419, opened 2022-09-10
set segment download timeout for cache action steps

altendky:
Even the blocks and plots cache appears to finish in its entirety in 15 seconds, so the 1 minute 'segment' (not sure exactly what that means) timeout seems reasonable.

https://github.com/actions/cache/issues/810


wallentx (approved):
Nice

cmmarslender (approved):

PR #13430, opened 2022-09-12
Lots of blocks are being synced, extend the timeouts

mariano54:
400 blocks are being synced to the wallet so it can take a bit of time


aqk (approved):

almogdepaz:
was this degradation expected due to some change ?
it doesn't look like the test changed recently so maybe we are burring some performance issue ?

mariano54:
Nothing changed here recently. Our wallet sync has actually improved significantly. It just seems like CI sometimes stops or gets extremely slow for like 10 or 15 seconds, and causes flakiness. On average these take 1 second on my PC.

emlowe (approved):

PR #13433, opened 2022-09-12
Correct /get_routes and /healthz to return a boolean for success

altendky:
Draft for:
- [x] back-compat considerations
  - reviewers can chatter about this

Note: we should note in the release notes this is an API change to the response for these. All the other RPCs return success as a boolean, so this change makes these work consistently with the other (which is a good thing)

### DIFF
--- a/chia/rpc/rpc_server.py
+++ b/chia/rpc/rpc_server.py
@@ -159,7 +159,7 @@ class RpcServer:
 
     async def _get_routes(self, request: Dict[str, Any]) -> EndpointResult:
         return {
-            "success": "true",
+            "success": True,
             "routes": list(self.get_routes().keys()),
         }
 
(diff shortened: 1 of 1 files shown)


arvidn (approved):
I would expect that if we have many users of these calls, someone would have told us about this mistake. If nobody did, it's probably safe to fix it without much consideration to backwards compatibility

xdustinface (approved):

emlowe:
This does concern me marginally in that anyone using these endpoints will need to change. I imagine rather than report any issue, they just assumed we returned what we wanted to and worked with that.
I did check our docker repo and although it uses /healthz it doesn't parse the response, it's happy as long as a HTTP success response code (200) is returned (uses `curl --fail`) - so it would be ok

emlowe (approved):

PR #13436, opened 2022-09-13
cmds: Allow `--label/-l` to be `""` to skip the labeling

xdustinface:


cmmarslender (approved):
tested and was able to skip the label without interaction with `-l ""`

paninaro (approved):

PR #13437, opened 2022-09-13
make a WalletProtocol, defining the shared API of all wallets

arvidn:
and fix type checking

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -152,8 +152,8 @@ class DataLayerWallet:
         self.log = logging.getLogger(name if name else __name__)
         self.standard_wallet = wallet
 
-        for _, wallet in self.wallet_state_manager.wallets.items():
-            if wallet.type() == uint8(WalletType.DATA_LAYER):
+        for _, w in self.wallet_state_manager.wallets.items():
+            if w.type() == uint8(WalletType.DATA_LAYER):
                 raise ValueError("DataLayer Wallet already exists for this key")
 
         assert name is not None
@@ -1039,10 +1039,11 @@ class DataLayerWallet:
     # WALLET #
     ##########
 
-    def puzzle_hash_for_pk(self, pubkey: G1Element) -> Optional[bytes32]:
+    def require_derivation_paths(self) -> bool:
+        return True
+
+    def puzzle_hash_for_pk(self, pubkey: G1Element) -> bytes32:
         puzzle: Program = self.puzzle_for_pk(pubkey)
-        if puzzle is None:
-            return puzzle
         return puzzle.get_tree_hash()
 
     def puzzle_for_pk(self, pubkey: G1Element) -> Program:
@@ -1059,8 +1060,8 @@ class DataLayerWallet:
     async def new_peak(self, peak: BlockRecord) -> None:
         pass
 
-    async def get_confirmed_balance(self, record_list: Optional[Set[WalletCoinRecord]] = None) -> uint64:
-        return uint64(0)
+    async def get_confirmed_balance(self, record_list: Optional[Set[WalletCoinRecord]] = None) -> uint128:
+        return uint128(0)
 
     async def get_unconfirmed_balance(self, record_list: Optional[Set[WalletCoinRecord]] = None) -> uint128:
         return uint128(0)
@@ -1288,6 +1289,15 @@ class DataLayerWallet:
                         )
                     summary["offered"].append(singleton_summary)
         return summary
+
+    async def select_coins(
+        self,
+        amount: uint64,
+        exclude: Optional[List[Coin]] = None,
+        min_coin_amount: Optional[uint64] = None,
+        max_coin_amount: Optional[uint64] = None,
+    ) -> Set[Coin]:
+        raise RuntimeError("DataLayerWallet does not support select_coins()")
 
 
 def verify_offer(
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -1374,3 +1374,9 @@ def verify_offer(
 
     if taker_from_offer != taker_from_reference:
         raise OfferIntegrityError("taker: reference and offer inclusions do not match")
+
+
+if TYPE_CHECKING:
+    from chia.wallet.wallet_protocol import WalletProtocol
+
+    _dummy: WalletProtocol = DataLayerWallet()
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -56,6 +56,8 @@ from chia.wallet.util.transaction_type import TransactionType
 from chia.wallet.util.wallet_types import AmountWithPuzzlehash, WalletType
 from chia.wallet.wallet_info import WalletInfo
 from chia.wallet.wallet_node import WalletNode
+from chia.wallet.wallet import Wallet
+from chia.wallet.wallet_protocol import WalletProtocol
 
 # Timeout for response from wallet/full node for sending a transaction
 TIMEOUT = 30
@@ -585,13 +587,15 @@ class WalletRpcApi:
                 did_id: Optional[bytes32] = None
                 if "did_id" in request and request["did_id"] is not None:
                     did_id = decode_puzzle_hash(request["did_id"])
-                if wallet.type() == WalletType.NFT and wallet.get_did() == did_id:
-                    log.info("NFT wallet already existed, skipping.")
-                    return {
-                        "success": True,
-                        "type": wallet.type(),
-                        "wallet_id": wallet.id(),
-                    }
+                if wallet.type() == WalletType.NFT:
+                    assert isinstance(wallet, NFTWallet)
+                    if wallet.get_did() == did_id:
+                        log.info("NFT wallet already existed, skipping.")
+                        return {
+                            "success": True,
+                            "type": wallet.type(),
+                            "wallet_id": wallet.id(),
+                        }
 
             async with self.service.wallet_state_manager.lock:
                 nft_wallet: NFTWallet = await NFTWallet.create_new_nft_wallet(
@@ -620,6 +624,7 @@ class WalletRpcApi:
                     max_pwi = 1
                     for _, wallet in self.service.wallet_state_manager.wallets.items():
                         if wallet.type() == WalletType.POOLING_WALLET:
+                            assert isinstance(wallet, PoolWallet)
                             pool_wallet_index = await wallet.get_pool_wallet_index()
                             if pool_wallet_index > max_pwi:
                                 max_pwi = pool_wallet_index
@@ -694,6 +699,7 @@ class WalletRpcApi:
                 if self.service.logged_in_fingerprint is not None:
                     wallet_balance["fingerprint"] = self.service.logged_in_fingerprint
                 if wallet.type() == WalletType.CAT:
+                    assert isinstance(wallet, CATWallet)
                     wallet_balance["asset_id"] = wallet.get_asset_id()
         else:
             async with self.service.wallet_state_manager.lock:
@@ -723,6 +729,7 @@ class WalletRpcApi:
                 if self.service.logged_in_fingerprint is not None:
                     wallet_balance["fingerprint"] = self.service.logged_in_fingerprint
                 if wallet.type() == WalletType.CAT:
+                    assert isinstance(wallet, CATWallet)
                     wallet_balance["asset_id"] = wallet.get_asset_id()
                 self.balance_cache[wallet_id] = wallet_balance
 
@@ -790,9 +797,11 @@ class WalletRpcApi:
         selected = self.service.config["selected_network"]
         prefix = self.service.config["network_overrides"]["config"][selected]["address_prefix"]
         if wallet.type() == WalletType.STANDARD_WALLET:
+            assert isinstance(wallet, Wallet)
             raw_puzzle_hash = await wallet.get_puzzle_hash(create_new)
             address = encode_puzzle_hash(raw_puzzle_hash, prefix)
         elif wallet.type() == WalletType.CAT:
+            assert isinstance(wallet, CATWallet)
             raw_puzzle_hash = await wallet.standard_wallet.get_puzzle_hash(create_new)
             address = encode_puzzle_hash(raw_puzzle_hash, prefix)
         else:
@@ -810,8 +819,10 @@ class WalletRpcApi:
         wallet_id = uint32(request["wallet_id"])
         wallet = self.service.wallet_state_manager.wallets[wallet_id]
 
-        if wallet.type() == WalletType.CAT:
-            raise ValueError("send_transaction does not work for CAT wallets")
+        if wallet.type() != WalletType.STANDARD_WALLET:
+            raise ValueError("send_transaction only works for standard wallets")
+
+        assert isinstance(wallet, Wallet)
 
         if not isinstance(request["amount"], int) or not isinstance(request["fee"], int):
             raise ValueError("An integer amount or fee is required (too many decimals)")
@@ -847,6 +858,7 @@ class WalletRpcApi:
 
         wallet_id = uint32(request["wallet_id"])
         wallet = self.service.wallet_state_manager.wallets[wallet_id]
+        assert isinstance(wallet, Wallet)
 
         async with self.service.wallet_state_manager.lock:
             transaction: Dict = (await self.create_signed_transaction(request, hold_lock=False))["signed_tx"]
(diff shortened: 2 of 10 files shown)


Quexington (dismissed):
It'd be nice if there was a way to make `wallet.type()` contain the type information so we can get rid of all the `assert isinstance(wallet, BlahWallet)` but this looks good regardless.

arvidn:
> It'd be nice if there was a way to make `wallet.type()` contain the type information so we can get rid of all the `assert isinstance(wallet, BlahWallet)` but this looks good regardless.

yeah, I imagine that we would want to replace some of our checks with `isinstance()`, to make the checks more clear, for humans and mypy.

chia/data_layer/data_layer_wallet.py
```diff
@@ -140,7 +140,7 @@ def id(self) -> uint32:
     async def create_new_dl_wallet(
         cls: Type[_T_DataLayerWallet],
         wallet_state_manager: WalletStateManager,
-        wallet: Wallet,
+        wall: Wallet,
```
altendky, line 143: I would make the interface read well and the local variables do any needed gymnastics.

chia/data_layer/data_layer_wallet.py
```diff
@@ -1289,6 +1287,15 @@ async def get_offer_summary(offer: Offer) -> Dict[str, Any]:
...
         return summary
 
+    async def select_coins(
+        self,
+        amount: uint64,
+        exclude: Optional[List[Coin]] = None,
+        min_coin_amount: Optional[uint64] = None,
+        max_coin_amount: Optional[uint64] = None,
+    ) -> Set[Coin]:
+        raise RuntimeError("DataLayerWallet does not support select_coins()")
```
altendky, line 1297: I reserve the right to explore alternatives in the future.  `:]`

chia/rpc/wallet_rpc_api.py
```diff
@@ -624,13 +625,15 @@ async def create_new_wallet(self, request: Dict) -> EndpointResult:
...
                     did_id = decode_puzzle_hash(request["did_id"])
-                if wallet.type() == WalletType.NFT and wallet.get_did() == did_id:
-                    log.info("NFT wallet already existed, skipping.")
-                    return {
-                        "success": True,
-                        "type": wallet.type(),
-                        "wallet_id": wallet.id(),
-                    }
+                if wallet.type() == WalletType.NFT:
+                    assert isinstance(wallet, NFTWallet)
```
altendky, line 629: Why not just do the type check as the if?

Also, I don't mean to request this change here, but maybe this behavior can be shifted onto the wallets?

chia/rpc/wallet_rpc_api.py
```diff
@@ -984,13 +997,15 @@ async def get_cat_list(self, request) -> EndpointResult:
 
     async def cat_set_name(self, request) -> EndpointResult:
         wallet_id = uint32(request["wallet_id"])
-        wallet: CATWallet = self.service.wallet_state_manager.wallets[wallet_id]
+        wallet = self.service.wallet_state_manager.wallets[wallet_id]
+        assert isinstance(wallet, CATWallet)
```
altendky, line 1001: Maybe we want helper functions for this where you pass in the type you want and it returns an instance of that type or raises?  Then we can centralize this check.

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -176,20 +176,21 @@ async def create_new_cat_wallet(
     @staticmethod
     async def create_wallet_for_cat(
         wallet_state_manager: WalletStateManager,
-        wallet: Wallet,
+        wall: Wallet,
```
altendky, line 179: As above, clean interface naming.

chia/wallet/wallet_protocol.py
```diff
@@ -0,0 +1,63 @@
...
+from chia.util.ints import uint8, uint32, uint64, uint128
+from chia.wallet.wallet_coin_record import WalletCoinRecord
+
+if TYPE_CHECKING:
+    from chia.wallet.wallet_state_manager import WalletStateManager
+
+
+class WalletProtocol(Protocol):
+
+    # TODO: it seems like this should return WalletType instead
```
altendky, line 20: Just verifying this is meant to be left in when merged.

chia/wallet/wallet_protocol.py
```diff
@@ -0,0 +1,63 @@
...
+from chia.types.blockchain_format.coin import Coin
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.ints import uint8, uint32, uint64, uint128
+from chia.wallet.wallet_coin_record import WalletCoinRecord
+
+if TYPE_CHECKING:
+    from chia.wallet.wallet_state_manager import WalletStateManager
+
+
+class WalletProtocol(Protocol):
```
altendky, line 18: Did you verify that all the wallet classes actually get complained about if you tweak the protocol in each of a few incompatible ways?

chia/wallet/wallet_protocol.py
```diff
@@ -0,0 +1,63 @@
...
+        ...
+
+    async def get_max_send_amount(self, unspent_records: Optional[Set[WalletCoinRecord]] = None) -> uint128:
+        ...
+
+    def puzzle_hash_for_pk(self, pubkey: G1Element) -> bytes32:
+        ...
+
+    # pylint says:
+    # E0601: Using variable 'WalletStateManager' before assignment (used-before-assignment)
```
altendky, line 62: I think explaining why this is ok might be more useful?

```suggestion
    # WalletStateManager is only imported for type hinting thus leaving pylint unable to process this
```

chia/wallet/wallet_state_manager.py
```diff
@@ -210,7 +211,10 @@ def sql_trace_callback(req: str) -> None:
         self.private_key = private_key
         self.main_wallet = await Wallet.create(self, main_wallet_info)
 
-        self.wallets = {main_wallet_info.id: self.main_wallet}
+        # mypy is saying:
+        # Dict entry 0 has incompatible type "uint32": "Wallet"; expected "uint32": "WalletProtocol"
+        # but Wallet implements WalletProtocol
+        self.wallets = {main_wallet_info.id: self.main_wallet}  # type: ignore[dict-item]
```
altendky, line 217: That sounds like a complaint that `Wallet` does not satisfy `WalletProtocol`.  How was it verified that it does?  I'm trying to think through what else would trigger this.

altendky (changes requested):
I'm not certain that any of this needs changed in this PR, but I would like to at least talk it through before merging.

chia/wallet/wallet_state_manager.py
altendky: I added:

```python
        x: Wallet
        y: WalletProtocol = x
```

And uncommented the ignore here and got:

```
chia/wallet/wallet_state_manager.py:215: error: Incompatible types in assignment (expression has type "Wallet", variable has type "WalletProtocol")  [assignment]
chia/wallet/wallet_state_manager.py:215: note: "Wallet" is missing following "WalletProtocol" protocol member:
chia/wallet/wallet_state_manager.py:215: note:     coin_added
chia/wallet/wallet_state_manager.py:215: note: Following member(s) of "Wallet" have conflicts:
chia/wallet/wallet_state_manager.py:215: note:     Expected:
chia/wallet/wallet_state_manager.py:215: note:         def get_confirmed_balance(self, record_list: Optional[Set[WalletCoinRecord]] = ...) -> Coroutine[Any, Any, uint64]
chia/wallet/wallet_state_manager.py:215: note:     Got:
chia/wallet/wallet_state_manager.py:215: note:         def get_confirmed_balance(self, unspent_records: Any = ...) -> Coroutine[Any, Any, uint128]
chia/wallet/wallet_state_manager.py:215: note:     Expected:
chia/wallet/wallet_state_manager.py:215: note:         def get_max_send_amount(self, unspent_records: Optional[Set[WalletCoinRecord]] = ...) -> Coroutine[Any, Any, uint128]
chia/wallet/wallet_state_manager.py:215: note:     Got:
chia/wallet/wallet_state_manager.py:215: note:         def get_max_send_amount(self, records: Optional[Set[WalletCoinRecord]] = ...) -> Coroutine[Any, Any, int]
chia/wallet/wallet_state_manager.py:220: error: Dict entry 0 has incompatible type "uint32": "Wallet"; expected "uint32": "WalletProtocol"  [dict-item]
```

altendky: (and yes, you shouldn't have to trick the protocol into telling you what is wrong...  there are options we can talk about.)

chia/rpc/wallet_rpc_api.py
altendky: Not fully thought through, but something of this sort.

```python
T_Wallet = TypeVar("T_Wallet")

def get_wallet_of_type(self, id: int, cls: Type[T_Wallet]) -> T_Wallet:
    wallet = self[id]
    if not isinstance(wallet, cls):
        raise Something()
    return wallet
```

chia/rpc/wallet_rpc_api.py
arvidn: I agree with both. We should use the language mechanism to check types, to let mypy (and other generic python programmers) keep up.

I consider this patch a first step to highlight this need.

chia/data_layer/data_layer_wallet.py
arvidn: right, this name is part of the function signature..

chia/rpc/wallet_rpc_api.py
arvidn: that sounds reasonable, at least after exhausted the possibility to move logic into the wallets themselves. (it seems like a code smell to constantly down cast the wallet type)

chia/wallet/wallet_protocol.py
arvidn: I kind of meant to leave it in. I would like to explore making that change, but I would need mypy coverage first.

chia/wallet/wallet_protocol.py
arvidn: I did not, and I'm not sure it matters. No wallet declares that it implements the protocol, this is just a hint to mypy. mypy did complain in a bunch of places. but mypy is still partially disabled in a lot of the wallet code, so I wouldn't expect it to catch everything yet either.

chia/wallet/wallet_protocol.py
arvidn: that's closer to an explanation. But the error message from pylint is still nonsensical. There is no "use" nor "assignment".

arvidn: or is pylint referring to the *type* `WalletStateManager` as a "variable"? and the definition of that type as "assignment"?

chia/wallet/wallet_state_manager.py
arvidn: oh, I was certain mypy would have told me if `Wallet` did not implement `WalletProtocol`..

chia/wallet/wallet_protocol.py
altendky: Yep, you're using `WalletStateManager` before pylint knows what it is since it is only conditionally imported due to `if typing.TYPE_CHECKING:`.  In some sense it is useful to think of Python as not having variables.  You have names that reference objects and that's about it.  As distinct from, say, C variables where an int variable is a block of memory holding the int.  Functions are objects.  As are classes, modules, methods, stack frames, etc.

chia/wallet/wallet_state_manager.py
altendky: Hence my other question about having made sure all the proper complaints happened.  `:]`  I think it took me like three refactor PRs for the RPC hinting stuff before I got to the point that any of the RPC classes were actually checked.  Speaking of which, I ought to look back and see how far I got with that.

chia/wallet/wallet_protocol.py
arvidn: I added checks

chia/wallet/wallet_protocol.py
```diff
@@ -0,0 +1,65 @@
...
+
+    async def get_max_send_amount(self, records: Optional[Set[WalletCoinRecord]] = None) -> uint128:
+        ...
+
+    # not all wallet supports this. To signal support, make
+    # require_derivation_paths() return true
+    def puzzle_hash_for_pk(self, pubkey: G1Element) -> bytes32:
+        ...
+
+    def require_derivation_paths(self) -> bool:
```
arvidn, line 60: I added this as it seemed less hacky and more type-safe than using `hasattr(wallet, "puzzle_hash_for_pk")` as we did previously (or actually, we checked for the wrong name, "puzzle_for_pk").

chia/wallet/wallet_protocol.py
```diff
@@ -0,0 +1,65 @@
...
+    async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection) -> None:
+        ...
+
+    async def select_coins(
+        self,
+        amount: uint64,
+        exclude: Optional[List[Coin]] = None,
+        min_coin_amount: Optional[uint64] = None,
+        max_coin_amount: Optional[uint64] = None,
+    ) -> Set[Coin]:
```
arvidn, line 37: This function had a discrepancy with the `DIDWallet`, which used to return an `Optional` instead of throwing on errors. Most wallets used this signature, so I harmonised them to this one.

see: https://github.com/Chia-Network/chia-blockchain/pull/13437/files#diff-ff2dff17865341c5c8414c162e4d4031faf867bece63a892edf43ff4629d2631L309

chia/rpc/wallet_rpc_api.py
```diff
@@ -1300,19 +1322,21 @@ async def did_update_metadata(self, request) -> EndpointResult:
...
+        assert isinstance(wallet, DIDWallet)
         my_did: str = encode_puzzle_hash(bytes32.fromhex(wallet.get_my_DID()), AddressType.DID.hrp(self.service.config))
         async with self.service.wallet_state_manager.lock:
-            coins = await wallet.select_coins(uint64(1))
-        if coins is None or coins == set():
-            return {"success": True, "wallet_id": wallet_id, "my_did": my_did}
-        else:
-            coin = coins.pop()
-            return {"success": True, "wallet_id": wallet_id, "my_did": my_did, "coin_id": coin.name()}
+            try:
```
arvidn, line 1329: this is the new uniform interface for `select_coins()` for all wallets. It doesn't return an `Optional`, it throws on errors.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -306,7 +306,7 @@ async def select_coins(
         exclude: Optional[List[Coin]] = None,
         min_coin_amount: Optional[uint64] = None,
         max_coin_amount: Optional[uint64] = None,
-    ) -> Optional[Set[Coin]]:
+    ) -> Set[Coin]:
```
Quexington, line 309: I think DID should be one of the wallets that raises on this method, it looks like right here the guaranteed behavior is just to return your DID every time.  And amount passed to this > 1 would fail I think.  Maybe out of scope of this PR though.

Quexington (dismissed):

altendky (commented):
Should we have a protocol check for the DataLayer wallet?

chia/wallet/did_wallet/did_wallet.py
arvidn: I don't think I follow. In this patch I make this function throw instead of returning `None`. The only condition where it used to return `None` was if the `amount` > `spendable_amount`. It falls back on the generic `select_coins()` function though, but is it the case that there will always only be a single coin in a DID wallet?

arvidn:
> Should we have a protocol check for the DataLayer wallet?

yes

arvidn:
I made the last change as a new commit to make it easier to review.

altendky (approved):

mariano54 (approved):

chia/rpc/wallet_rpc_api.py
```diff
@@ -1587,7 +1624,8 @@ async def nft_get_by_did(self, request) -> EndpointResult:
 
     async def nft_get_wallet_did(self, request) -> EndpointResult:
         wallet_id = uint32(request["wallet_id"])
-        nft_wallet: NFTWallet = self.service.wallet_state_manager.wallets[wallet_id]
+        nft_wallet = self.service.wallet_state_manager.wallets[wallet_id]
+        assert isinstance(nft_wallet, NFTWallet)
```
altendky, line 1628: I do here-by amend my review.  We can't be asserting on external input.  This should be an error like `requested wallet id <ID> is a <wallet type> not the needed nft wallet type`.  Or...  something.

Looks like there may be a few of these to address.

This came up in support based on a response of `Failed to mint NFT: {'error': '', 'success': False}` and log traceback.

```python-traceback
wallet chia.rpc.util : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia/rpc/util.py", line 18, in inner
  File "chia/rpc/wallet_rpc_api.py", line 1743, in nft_get_wallet_did
AssertionError
```

altendky: Exploring in https://github.com/Chia-Network/chia-blockchain/pull/14030.

PR #13438, opened 2022-09-13
fix wallet fork test

almogdepaz:
changed the cases where the case can change to be dynamic
added assertions 
added a close case with 10 blocks apart to make sure we still test that case

### DIFF
--- a/tests/wallet/sync/test_wallet_sync.py
+++ b/tests/wallet/sync/test_wallet_sync.py
@@ -576,6 +576,7 @@ class TestWalletSync:
 
     @pytest.mark.asyncio
     async def test_get_wp_fork_point(self, default_10000_blocks):
+
         blocks = default_10000_blocks
         header_cache, height_to_hash, sub_blocks, summaries = await load_blocks_dont_validate(blocks)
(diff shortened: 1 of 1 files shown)


tests/wallet/sync/test_wallet_sync.py
```diff
@@ -584,20 +585,46 @@ async def test_get_wp_fork_point(self, default_10000_blocks):
...
         fork23 = get_wp_fork_point(test_constants, wp3, wp2)
         fork24 = get_wp_fork_point(test_constants, wp4, wp2)
         fork34 = get_wp_fork_point(test_constants, wp3, wp4)
         fork45 = get_wp_fork_point(test_constants, wp4, wp5)
-        assert fork14 == 8700
-        assert fork24 == 8700
-        assert fork12 == 9000
+        fork16 = get_wp_fork_point(test_constants, wp1, wp6)
+
+        # overlap between recent chain in wps, fork point should is the tip of the shorter wp
```
altendky, line 598:
```suggestion
        # overlap between recent chain in wps, fork point should be the tip of the shorter wp
```

mariano54 (approved):
nice.

PR #13439, opened 2022-09-13
remove uses of cast() in chia.daemon.server

altendky:
Draft for:
- [x] let https://github.com/Chia-Network/chia-blockchain/pull/13298 go in first


arvidn (approved):
is `data` always a string? Did something change recently to make this unnecessary, or was this a mistake from the beginning?

altendky:
You mean a `Dict[str, Any]`?  I didn't dig into the history.  Just noticed a new entry being added, I don't like casting, so I questioned it and noticed the items already had the type they were being cast too.  Might have changed when we added the `WsMessage` `TypedDict` type?

emlowe (approved):

PR #13443, opened 2022-09-13
Disconnect untrusted peers when trusted connects

Quexington:


mariano54 (changes requested):
We should only close connection if the trusted peer is synced.

PR #13458, opened 2022-09-14
Better proof to singleton mapping

Quexington:


trepca:
@Quexington what makes it a better proof? (it'd be great if you can add more guidance in PR descriptions in general)

Quexington:
This solves an issue where trying to prove the same k,v pair in multiple singletons in a DL offer results in a ASSERT_ANNOUNCE_CONSUMED error on chain.  The check was heuristic in that if it saw a valid proof for a value it was looking for a proof for, it would just use it.  However, you can have two proper proofs for a value that result in a different root.  So I added a mapping of singletons to their roots and then added an extra check that not only does the proof check out, but it comes out to the root inside the singleton that the graftroot puzzle is inside.

emlowe (approved):

PR #13461, opened 2022-09-15
util: Remove the assigned label if a key gets removed

xdustinface:
The labels are currently left dangling in the keyring after key deletion.

### DIFF
--- a/chia/util/keychain.py
+++ b/chia/util/keychain.py
@@ -6,6 +6,7 @@ import unicodedata
 from bitstring import BitArray  # pyright: reportMissingImports=false
 from blspy import AugSchemeMPL, G1Element, PrivateKey  # pyright: reportMissingImports=false
 from chia.util.errors import (
+    KeychainException,
     KeychainNotSet,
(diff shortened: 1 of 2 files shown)


chia/util/keychain.py
```diff
@@ -437,6 +437,11 @@ def delete_key_by_fingerprint(self, fingerprint: int) -> int:
             try:
                 key_data = self._get_key_data(index, include_secrets=False)
                 if key_data.fingerprint == fingerprint:
+                    try:
+                        self.keyring_wrapper.delete_label(key_data.fingerprint)
+                    except Exception:
```
paninaro, line 442: I'm channeling my inner-Kyle here, but should we be catching a specific exception, like `KeychainUserNotFound`?

xdustinface: If so it would rather be `KeychainError` but i decided to not because we two lines below also catch `Exception` when deleting the key. Anyway, you got me :)

paninaro (dismissed):
lgtm

paninaro (approved):

PR #13462, opened 2022-09-16
Add resync API

ytx1991:
https://github.com/Chia-Network/issue-tracker/issues/36


chia/rpc/wallet_rpc_api.py
```diff
@@ -895,6 +896,23 @@ async def get_current_derivation_index(self, request) -> Dict[str, Any]:
 
         return {"success": True, "index": index}
 
+    async def resync_wallets(self, request) -> Dict[str, Any]:
```
mariano54, line 899: Can you write a comment describing what the function does?

chia/rpc/wallet_rpc_api.py
```diff
@@ -895,6 +896,23 @@ async def get_current_derivation_index(self, request) -> Dict[str, Any]:
...
+        assert self.service.wallet_state_manager is not None
+        fingerprint = self.service.logged_in_fingerprint
+        self.service._close()
+        log.info("Stopping wallet ...")
+        # Clean tables
+        await self.service.wallet_state_manager.clean_resync_tables()
+        log.info("Cleaned tables ...")
+        peers_close_task: Optional[asyncio.Task] = await self.service._await_closed(shutting_down=False)
```
mariano54, line 907: Why delete the tables before calling await close?

chia/rpc/wallet_rpc_api.py
```diff
@@ -109,6 +109,7 @@ def get_routes(self) -> Dict[str, Endpoint]:
             "/select_coins": self.select_coins,
             "/get_current_derivation_index": self.get_current_derivation_index,
             "/extend_derivation_index": self.extend_derivation_index,
+            "/resync_wallets": self.resync_wallets,
```
mariano54, line 112: I would call it 'reset DB' or 'reset wallets'

chia/wallet/wallet_state_manager.py
```diff
@@ -1478,6 +1478,14 @@ async def get_wallet_for_coin(self, coin_id: bytes32) -> Any:
         wallet = self.wallets[wallet_id]
         return wallet
 
+    async def clean_resync_tables(self) -> None:
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
+            await conn.execute("DELETE FROM coin_record")
```
mariano54, line 1483: Why not delete all tables? For example transaction records

mariano54: If we add new tables we will forget to delete them here in the future, guaranteed. I recommend using code that wipes everything all at once instead.

mariano54 (dismissed):

mariano54:
If you resync without wiping transactions, I'm sure a lot of things can get messed up.
We might want to make it an optional flag (or always delete everything)

chia/rpc/wallet_rpc_api.py
ytx1991: Since the connection will be closed.

chia/wallet/wallet_state_manager.py
ytx1991: That is the requirement, we want to save the offer and transaction records. Because we are doing this during the wallet running, drop table will cause "table not exist" errors.

ytx1991:
> If you resync without wiping transactions, I'm sure a lot of things can get messed up. For example we will have confirmed TX but we will be synced up to height 0,which is not consistent. We might want to make it an optional flag (or always delete everything)

I have same concerns and discussed with @trepca . So far based on the testnet experiments, the transaction looks good. We are doing a "Insert or Replace" for the transaction table. But I agree partially resync will cause a lot of edge cases.

trepca (changes requested):
please remove rl_wallet files and add a unit test for resync

chia/rpc/wallet_rpc_api.py
arvidn: it seems safer to delete the tables *after* the wallet has shut down. what if it's trying to access them while shutting down?

arvidn: I don't understand how you can assign the return value from `_await_closed()` to a variable of type `Optional[asyncio.Task]`. `_await_closed()` is an `async` function that returns `None`. On this line, you `await` the call to it, so presumably that would return `None`.

chia/rpc/wallet_rpc_api.py
```diff
@@ -909,6 +910,28 @@ async def get_current_derivation_index(self, request) -> Dict[str, Any]:
 
         return {"success": True, "index": index}
 
+    async def reset_wallets_db(self, request) -> Dict[str, Any]:
```
arvidn, line 913: this name is a little bit misleading. it sounds like it's clearing the whole database.
maybe something like "reset wallet sync" would make more sense.

chia/wallet/wallet_state_manager.py
arvidn: if we add a new table, that's also derived from the chain, we would need to delete it here too, as mariano points out. Can you think of a way to avoid it being forgotten? ideally some check that fails CI. Maybe an assert at the end of this function that only the tables we expect are left. That way the test for this function would fail if a new table is added.

Also, regarding `drop table` vs `delete from`. It really seems risky to do this while the wallet is running. Is there a good reason to not shut down the wallet node first, then drop the tables? There are so many scenarios and interleavings of logic that would have to be tested if we allow a running wallet to suddenly lose all entries in these tables.

chia/rpc/wallet_rpc_api.py
ytx1991: It's copy from _stop_wallet()

ytx1991: The requirement is to reset the database without shutdown the wallet.

chia/wallet/wallet_state_manager.py
ytx1991: I tried drop table, it doesn't work while the connection is still open. The requirement is not shutdown the node. If we shutdown the wallet node, we need a daemon or some external thing to recreate a node.

arvidn: when you say "connection", do you mean the database connection?
How did it fail? 
It seem like that failure could be an indication of some additional synchronization that needs to happen, and your solution of just removing all the rows is just masking an underlying problem.

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -1125,6 +1125,15 @@ async def test_select_coins_rpc(wallet_rpc_environment: WalletRpcTestEnvironment
         assert coin != coin_300[0]
 
 
+@pytest.mark.asyncio
+async def test_reset_wallet_sync(wallet_rpc_environment: WalletRpcTestEnvironment):
+    env: WalletRpcTestEnvironment = wallet_rpc_environment
+    full_node_api: FullNodeSimulator = env.full_node.api
+    client: WalletRpcClient = env.wallet_1.rpc_client
+    await generate_funds(full_node_api, env.wallet_1)
+    await client.reset_wallet_sync()
```
arvidn, line 1134: there's no test here. If you make `reset_wallet_sync()` an empty function, this test will still pass.

It needs to ensure both that the intended state was achieved and also that it's safe to do while performing any other wallet operation. This test isn't covering any interesting edge cases that (in my mind) are quite likely to cause problems.

arvidn (changes requested):
It might be a good idea to (in a separate PR, to land before this one) remove `wallet_rpc_api.py` from the mypy exemption list, in order to catch trivial mistakes.

tests/wallet/rpc/test_wallet_rpc.py
ytx1991: TBH, I don't know how to simulate such complicated cases in unit tests. If you know how to do it let me know.

chia/wallet/wallet_state_manager.py
ytx1991: Here are the reasons I don't drop tables:
1. It's hard to reuse the existing code to recreate tables. Fully restarting the wallet can solve the problem but it is against the requirement.
2. It will throw exceptions while the wallet is still running. The insert will fail if it happens after we delete the table and before we create the new table.

tests/wallet/rpc/test_wallet_rpc.py
arvidn: this functionality is likely to break in the future unless we have an automated test for it.
maybe you could have the test check the wallet log to ensure it resynced

PR #13463, opened 2022-09-16
plot_sync: Change log level to `INFO` in `Receiver.reset`

xdustinface:
It's not necessarily an error, could be just a harvester restart.


altendky (approved):
In so much as a disconnect should be reported at a higher level that should probably be the job of the disconnect itself, not this response to it.

emlowe (approved):

PR #13465, opened 2022-09-16
less `cast()` in keychain server

altendky:


xdustinface (approved):
Looks good. Note that the `command` hint is unrelated and the `Any` could be an `object` it seems :)

emlowe (approved):

PR #13466, opened 2022-09-16
feat: enable `create_signed_transaction` and `push_transaction`

stmharry:
This PR does two things:
- CAT requires a `create_signed_transaction` equivalent to be spent with flexibility
- The wallet requires and endpoint that not only forwards `SpendBundle`'s, but also record and track their status.


chia/rpc/wallet_rpc_api.py
```diff
@@ -1905,21 +1938,21 @@ async def create_signed_transaction(self, request, hold_lock=True) -> EndpointRe
...
-                exclude_coins=exclude_coins,
-                ignore_max_send_amount=True,
-                primaries=additional_outputs,
-                memos=memos_0,
-                coin_announcements_to_consume=coin_announcements,
-                puzzle_announcements_to_consume=puzzle_announcements,
-                min_coin_amount=min_coin_amount,
-            )
-        return {"signed_tx": signed_tx.to_json_dict_convenience(self.service.config)}
+                return {"signed_tx": tx.to_json_dict_convenience(self.service.config)}
```
Quexington, line 1941: We should return `{"signed_tx": blah, "signed_txs": [blah]}` here for backwards compatibility but also reducing the complexity on the client moving forward as most wallets will return multiple transactions.

Quexington (changes requested):
Can you also add a test for `push_transactions` and `create_signed_transaction` with a CAT into `tests/wallet/rpc/test_wallet_rpc.py`?

chia/rpc/wallet_rpc_api.py
stmharry: Added backwards compatibility

stmharry:
Test cases added for both `push_transactions` and `create_signed_transaction` RPC endpoints.

chia/rpc/wallet_rpc_api.py
Quexington: I meant the other way around.  The CAT side can only return signed_txs, we should just make the XCH case return signed_txs as well.  That was you don't need the Union and if/else in the RPC client.

stmharry: Oh I see. That certainly makes sense.

stmharry: Marked as resolved.

Quexington (dismissed):

mariano54 (dismissed):

wallentx:
@stmharry there are a few things that are failing the pre-commit checks and our linting checks. Are you able to correct for those?

chia/rpc/wallet_rpc_api.py
```diff
@@ -414,6 +416,23 @@ async def push_tx(self, request: Dict) -> EndpointResult:
         await self.service.push_tx(SpendBundle.from_bytes(hexstr_to_bytes(request["spend_bundle"])))
         return {}
 
+    async def push_transactions(self, request: Dict) -> EndpointResult:
```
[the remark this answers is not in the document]
stmharry: Some wallets like CAT naturally generated multiple txs, and I think other potential plugin wallets would also love to do that. Having all txs pushed in one call seems like a good solution.
Compared to `/push_tx`, typically one can just aggregate all spend bundles so the endpoint is in singular form.

stmharry:
> @stmharry there are a few things that are failing the pre-commit checks and our linting checks. Are you able to correct for those?

Just ran the linter. The history looks a bit dirty since there's a lingering "Merging branch main" event that doesn't seem to go away. That said the only difference between this commit and the last one is the linter run.

stmharry:
`pre-commit` items run without complaints. 
Added a `create_signed_transactions` wallet RPC client function to facilitate the new "multi-tx" return value by CAT wallet. Calls to `create_signed_transaction` will remain singular for backwards compatibility.

mariano54 (dismissed):

paninaro (approved):

PR #13468, opened 2022-09-16
add a transactional reader as `DBWrapper2.reader()`

altendky:
Draft for:
- [x] fixing it
- [x] fix the WAL situation in https://github.com/Chia-Network/chia-blockchain/pull/13510 first, then adjust here
- [x] self review


chia/util/db_wrapper.py
```diff
@@ -259,6 +259,20 @@ async def writer_maybe_transaction(self) -> AsyncIterator[aiosqlite.Connection]:
                 finally:
                     self._current_writer = None
 
+    @contextlib.asynccontextmanager
+    async def reader(self) -> AsyncIterator[aiosqlite.Connection]:
+        async with self.reader_no_transaction() as connection:
+            if connection.in_transaction:
+                yield connection
+            else:
+                await connection.execute("BEGIN DEFERRED;")
```
arvidn, line 268: Does `SAVEPOINT` cause `in_transaction` to be true? If not, that would be problematic since I don't believe `BEGIN` is allowed to nest `SAVEPOINT`. (my reading of [this](https://sqlite.org/lang_savepoint.html))

chia/util/db_wrapper.py
```diff
@@ -259,6 +259,20 @@ async def writer_maybe_transaction(self) -> AsyncIterator[aiosqlite.Connection]:
...
+                try:
+                    yield connection
+                finally:
+                    # close the transaction with a rollback instead of commit just in
+                    # case any modifications were submitted through this reader
+                    await connection.rollback()
```
arvidn, line 274: similar to my comment above, I believe this will roll back *all* transactions, not just this one.
consider the case where we have one task that has a write-connection (i.e. `SAVEPOINT) open, and writing to it. Within this transaction it opens a `reader` transaction which throws. Just leaving *this* context shouldn't roll back all previous changes within the outer transaction.

I think this only works in the scenario where `reader_no_transaction()` returns one of the reader-connections. But remember that if the calling task currently also has the writer, you get the writer connection back.

I suspect to get this right it might be simpler to not rely on `reader_no_transaction()`

arvidn (changes requested):
I believe there is one problematic case that isn't covered by a test:

a single task, something like this I think:

```
async with db_wrapper.writer() as writer:
   await writer.execute("UPDATE counter SET value = 1337")
   try:
      async with db_wrapper.reader() as reader:
         raise RuntimeError("")
   except Exception:
      pass
   assert await query_value(connection=writer) == 1337
```

arvidn:
at a second look, I realize you only `rollback` in the case where you `BEGIN`. So, assuming `in_transaction` is true even for a `SAVEPOINT` (which does sound quite likely), I don't think there are any problems.

arvidn:
digging a bit in python's `in_transaction`, it looks like it's relying on [sqlite3_get_autocommit()](https://www.sqlite.org/c3ref/get_autocommit.html) (but they also have confusing wording about there being uncommitted changes or not, which the sqlite documentation says nothing about).

However, the sqlite documentation does *not* in fact enumerate `SAVEPOINT` as a function that disables autocommit (but I believe it does). The documentation doesn't explicitly call it out, but says it behaves like `BEGIN` ([here](https://sqlite.org/lang_savepoint.html))

arvidn (approved):

emlowe (approved):

PR #13473, opened 2022-09-17
switch to context manager for task timing instrumentation

altendky:


chia/util/task_timing.py
```diff
@@ -335,6 +336,24 @@ def stop_task_instrumentation() -> None:
             f.write("}\n")
 
 
+@contextlib.contextmanager
+def manage_task_instrumentation() -> Iterator[None]:
```
arvidn, line 340: I don't see this one used anywhere

arvidn (dismissed):

chia/util/task_timing.py
altendky: Yes, it's a non-optional usage for places where you just want to enable instrumentation such as when testing code locally or if we end up with mandatory test cases to collect data in CI or such.  I could see collapsing down to the single interface with `enable` defaulted to `True`.  Does that sound better?  Or perhaps swap to `disable: bool = False`.

arvidn: I don't think it really matters, I would just expect to not have this function until we need it somewhere

arvidn (approved):

Issue #13209, opened 2022-08-27
New lockfile `/tmp/clvm_compile/p2_conditions.clvm.lock` prevents second harvester being started


neurosis69:
Would it be possible to add the node_id to the path to get uniqueness?
https://github.com/Chia-Network/chia-blockchain/blob/093115311176e56388fb40334003a20e440da721/chia/wallet/puzzles/load_clvm.py#L72

I run two harvester instances on the same machine, both are connected to different farmers to simulate kind of H/A.
With versions < 1.5.1, this was possible by just changing the port of the secondary harvester and farmers to let them connect.

Since 1.5.1 I cannot start a second harvester instance on the same machine and the reason is that both want to acquire the same lock file.

```
Traceback (most recent call last):
  File "/home/chia/chia-blockchain/venv/bin/chia", line 5, in <module>
    from chia.cmds.chia import main
  File "/home/chia/chia-blockchain/chia/cmds/chia.py", line 18, in <module>
    from chia.cmds.plotters import plotters_cmd
  File "/home/chia/chia-blockchain/chia/cmds/plotters.py", line 2, in <module>
    from chia.plotters.plotters import call_plotters
  File "/home/chia/chia-blockchain/chia/plotters/plotters.py", line 5, in <module>
    from chia.plotters.bladebit import get_bladebit_install_info, plot_bladebit
  File "/home/chia/chia-blockchain/chia/plotters/bladebit.py", line 10, in <module>
    from chia.plotting.create_plots import resolve_plot_keys
  File "/home/chia/chia-blockchain/chia/plotting/create_plots.py", line 10, in <module>
    from chia.daemon.keychain_proxy import KeychainProxy, connect_to_keychain_and_validate, wrap_local_keychain
  File "/home/chia/chia-blockchain/chia/daemon/keychain_proxy.py", line 8, in <module>
    from chia.cmds.init_funcs import check_keys
  File "/home/chia/chia-blockchain/chia/cmds/init_funcs.py", line 10, in <module>
    from chia.consensus.coinbase import create_puzzlehash_for_pk
  File "/home/chia/chia-blockchain/chia/consensus/coinbase.py", line 6, in <module>
    from chia.wallet.puzzles.p2_delegated_puzzle_or_hidden_puzzle import puzzle_for_pk
  File "/home/chia/chia-blockchain/chia/wallet/puzzles/p2_delegated_puzzle_or_hidden_puzzle.py", line 68, in <module>
[...]

PR #13478, opened 2022-09-19
only compile clvm for developers

altendky:
Set the `CHIA_DEV_COMPILE_CLVM_ON_IMPORT` environment variable to anything except `""` to trigger automatic recompilation of the Chialisp on load.  This will also be triggered when running tests based on the pytest module having been loaded.

Should help with https://github.com/Chia-Network/chia-blockchain/issues/13209.

Draft for:
- [ ] General Chialisp developer input


chia/wallet/puzzles/load_clvm.py
```diff
@@ -86,36 +89,59 @@ def load_serialized_clvm(
...
     clvm_blob = bytes.fromhex(clvm_hex)
     return SerializedProgram.from_bytes(clvm_blob)
 
 
-def load_clvm(clvm_filename, package_or_requirement=__name__, include_standard_libraries: bool = False) -> Program:
+def load_clvm(
+    clvm_filename,
+    package_or_requirement=__name__,
+    include_standard_libraries: bool = False,
+    recompile: bool = True,
```
arvidn, line 122: is there a good reason to not just default this to?
```
recompile: bool = recompile_requested,
```

And not bother to introduce a new name `load_clvm_maybe_recompile()`. Are there still use cases for the function where we try to recompile unconditionally?

If so, I would still think that would be the uncommon case that should have a longer name.

tools/manage_clvm.py
```diff
@@ -164,4 +181,52 @@ def main() -> int:
     return 1 if overall_fail else 0
 
 
+@main.command()
+def build() -> int:
```
arvidn, line 185: is this function triggered by CI? i.e. is it demonstrated to work and produce identical output?

arvidn (commented):
looks reasonable. It seems reasonable to me to leave the `load_clvm()` name as-is, and change its behavior to not, unconditionally, attempt a recompilation. It would make the patch smaller and lower risk or conflicts with other PRs.

At a higher level, it seems odd to me to have this logic in `load_clvm()` to begin with. Building and hashing of the clvm programs would make a lot more sense to do in pre-commit, never at import-time. Is that general consensus and this is a step in that direction, or is there a good reason to do it at import-time?

tools/manage_clvm.py
altendky: Nope.  This was a side step that wasn't removed.

chia/wallet/puzzles/load_clvm.py
altendky: The concern was that other projects might be using it.  The fact it is buried down in our wallet puzzles directory seems both bad and an indication against that.  @trepca brought up the concern.

altendky:
I agree that this shouldn't be an import-time activity at all.  I have been peeling off pieces around this topic for awhile to clean it up and yes, this is just the next piece.  The goal here is to at least relegate the compilation to "developers".  Though a compromise was made to trigger it when running tests as well so that chialisp devs don't have to set an environment variable nor otherwise trigger compilation in their workflows.

chia/wallet/puzzles/load_clvm.py
arvidn: although, come to think of it, the NFT wallet does de-compile and re-compile clvm code. I'm not sure why, but there is a dependency there too.

arvidn (approved):

trepca (approved):

chia/wallet/puzzles/load_clvm.py
trepca: It'd be great to separate into dev libraries, but currently even `chia-dev-tools` depends on `chia-blockchain`. For any serious coin development, developers need access to simulator which is in `chia-blockchain` too and all the standard puzzles.

PR #13460, opened 2022-09-15
Wallentx/prerelease

wallentx:
The working state of code in this PR:
A `PR check` from `<dev branch>` to `main` - https://github.com/wallentx/billy-blockchain/pull/2/checks
https://github.com/wallentx/billy-blockchain/actions
https://github.com/wallentx/billy-blockchain/releases

PR #13489, opened 2022-09-21
remove noise from lead in to changelog

altendky:


hoffmang9:
That intro is there to describe why the log is the format it is and to largely comply with the standard. What's the reason to change that?

altendky:
I don't think any of this is useful to those reading the changelog.  If we want to document the format then maybe markdown has a comment feature?

We start with four things that I think are odd distractions.

1) The definition of a changelog
2) The format which is only particularly useful to those editing the changelog
3) An always empty 'unreleased' section that is generally supposed to be deleted when you make a release anyways.  Since we don't edit it during development (conflict central, why changelog tools exist) it never has anything in it at all.
4) The "what's changed" section that is not a useful description since these are all changes, we don't use in any other release sections, and also is a sub-heading inside the always empty unreleased section.

altendky:
I forgot the comment about us not following semantic versioning.  How is that PEP440s fault and in what particular way do we not follow it?

hoffmang9:
I would suggest you read the linked standard which is what we are following. This PR makes the decision to abandon the standard. https://keepachangelog.com/en/1.0.0/

altendky:
I think it does not and our present usage does not understand the intent, for at least some counts.  The point isn't to have this noise.  Also, how do we not follow semantic versioning (I'm not claiming we do) and how is it PEP440s fault?

hoffmang9:
The PEP-440 issue is primarily around what behavior python forces in pre release versions. Since the launch, we've been able to adhere to the standard and SemVer so that extra line can mostly go as long as we don't formally start doing named pre releases. The standard includes declaring the standard however.

altendky:
Prereleases are being worked on in https://github.com/Chia-Network/chia-blockchain/pull/13460.  I'm not sure what exact bit of prereleases is incompatible between semver and PEP440.  In some sense I think it could be argued this is not a relevant detail, but it is relevant enough to place blame so it seems important enough to understand.  I looked over https://github.com/semver/semver/issues/483 and they point out some differences.  Is it the `-`?  Or...  It sounds like the intent is to pretty much follow semver in terms of the semantics of the version but we have a corner or two here the generated version is not exactly in spec?  Though I would wonder what would qualify for a Chia v2 not backwards compatible change.  Without knowing that it's hard to say we follow semver I think (we don't say that now but I'm trying to figure out what we really ought to be expressing).  We certainly have made various small CLI and RPC changes that are not backwards compatible, probably even in micro/patch versions.

On the unreleased section topic, the point of that section is to build the changelog along the way.  To add entries into the unreleased section as the changes are made.  We do not do this.  We should build our changelog as we go, but we should not do it into this single file since that will result in piles of trivial merge conflicts that must be resolved.  That is why tools like towncrier and reno exist.  Editing the changelog directly is really only plausible with few developers and a small number of concurrent changes.  Since we do not, and should not, use the unreleased section, it seems a bad thing to have.

hoffmang9:
I can spend a bunch of time re researching why pre release PEP440 isn’t SemVer compliant but that seems like an incredible waste of time. Dropping unreleased, sure. However the rest of this change needs to go as the only thing left is to ignore the issue and state we fully follow Semver to remain in compliance with the standard

altendky:
This is exactly why these things need to be clearly documented instead of just vaguely passing blame.  It is unclear that there even is an actual problem beyond our presumption that we correctly remember a problem that we don't remember.  In lots of cases like this there has just been misunderstanding or the issue has been fixed in the mean time.  Clearly documenting exact issues would save a lot of time while also allowing us to move forward when the issue has in fact been resolved.

I would argue that we do not particularly try to follow semver just in terms of the actual semantics of it.

The unreleased section removal has been submitted at https://github.com/Chia-Network/chia-blockchain/pull/13495.

PR #13491, opened 2022-09-21
Add Signing message APIs

ytx1991:
https://github.com/Chia-Network/issue-tracker/issues/17
Add three new APIs for signing message.


Quexington (dismissed):

chia/rpc/wallet_rpc_api.py
```diff
@@ -999,6 +1001,15 @@ async def send_notification(self, request) -> EndpointResult:
         await self.service.wallet_state_manager.add_pending_transaction(tx)
         return {"tx": tx.to_json_dict_convenience(self.service.config)}
 
+    async def sign_message(self, request) -> EndpointResult:
+        puzzle_hash: bytes32 = decode_puzzle_hash(request["address"])
```
mariano54, line 1005: why address and not `puzzle_hash`?

chia/rpc/wallet_rpc_api.py
```diff
@@ -999,6 +1001,15 @@ async def send_notification(self, request) -> EndpointResult:
...
+        hex_message = request["message"]
+        if hex_message.startswith("0x") or hex_message.startswith("0X"):
```
mariano54, line 1007: I think we have a bytes32.fromhex method

chia/rpc/wallet_rpc_api.py
```diff
@@ -1902,6 +1925,35 @@ async def nft_calculate_royalties(self, request) -> EndpointResult:
...
+                        nft = nft
+                        break
+                if nft_wallet is not None:
+                    break
+        if nft_wallet is None or nft is None:
+            return {"success": False, "error": f"NFT wallet {nft_id.hex()} doesn't exist."}
+
+        assert isinstance(nft_wallet, NFTWallet)
+        hex_message = request["message"]
+        if hex_message.startswith("0x") or hex_message.startswith("0X"):
```
mariano54, line 1951: we should not be repeating this hex code everywhere

chia/rpc/wallet_rpc_api.py
```diff
@@ -999,6 +1001,15 @@ async def send_notification(self, request) -> EndpointResult:
...
         return {"tx": tx.to_json_dict_convenience(self.service.config)}
 
+    async def sign_message(self, request) -> EndpointResult:
+        puzzle_hash: bytes32 = decode_puzzle_hash(request["address"])
+        hex_message = request["message"]
+        if hex_message.startswith("0x") or hex_message.startswith("0X"):
+            hex_message = hex_message[2:]
+        message = bytes.fromhex(hex_message)
+        pubkey, signature = await self.service.wallet_state_manager.main_wallet.sign_message(message, puzzle_hash)
+        return {"success": True, "pubkey": str(pubkey), "signature": str(signature)}
```
mariano54, line 1011: I think `bytes(pubkey).hex()` is better. It's not clear what `str` does.

chia/rpc/wallet_rpc_api.py
```diff
@@ -999,6 +1001,15 @@ async def send_notification(self, request) -> EndpointResult:
         await self.service.wallet_state_manager.add_pending_transaction(tx)
         return {"tx": tx.to_json_dict_convenience(self.service.config)}
 
+    async def sign_message(self, request) -> EndpointResult:
```
mariano54, line 1004: Some comments would be nice as well

chia/rpc/wallet_rpc_api.py
ytx1991: Address is more popular among the users.

chia/rpc/wallet_rpc_api.py
ytx1991: hex_message is not a bytes32

chia/rpc/wallet_rpc_api.py
ytx1991: They are same

chia/rpc/wallet_rpc_api.py
Quexington: We also have chia.util.byte_types.hexstr_from_bytes

chia/rpc/wallet_rpc_api.py
mariano54: I know, I wrote that code. However, one is more clear than the other.

chia/rpc/wallet_rpc_api.py
mariano54: Makes sense. I just want to ensure that we have consistency (always accepting address or always accepting puzzle hash), in all the apis. It looks like some are using one and some are using the other, which is not great. I suppose its fine to use address then.

chia/rpc/wallet_rpc_api.py
```diff
@@ -115,6 +115,7 @@ def get_routes(self) -> Dict[str, Endpoint]:
             "/get_notifications": self.get_notifications,
             "/delete_notifications": self.delete_notifications,
             "/send_notification": self.send_notification,
+            "/sign_message": self.sign_message,
```
paninaro, line 118: Is there a need to have 3 separate RPC variants for signing messages? If we want to support data layer singleton signing, will that require a 4th variant?

If we can, I would recommend that we consolidate `nft_sign_message` and `did_sign_message` into this one `sign_message` RPC.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1539,13 +1552,27 @@ async def did_transfer_did(self, request) -> EndpointResult:
...
 
     async def did_sign_message(self, request) -> EndpointResult:
-        wallet_id = uint32(request["wallet_id"])
-        did_wallet: WalletProtocol = self.service.wallet_state_manager.wallets[wallet_id]
+        """
+        Given a DID ID, sign the message by the DID's private key.
+        :param request:
+        :return:
+        """
+        if request["did_id"].startswith("did"):
```
paninaro, line 1560: It would be nice to use the `AddressType` utils to avoid hard-coding bech32m prefixes.

```suggestion
        if is_valid_address(request["did_id"], {AddressType.DID), self.service.config):
```

chia/rpc/wallet_rpc_api.py
```diff
@@ -1902,6 +1929,37 @@ async def nft_calculate_royalties(self, request) -> EndpointResult:
             {asset["asset"]: uint64(asset["amount"]) for asset in request.get("fungible_assets", [])},
         )
 
+    async def nft_sign_message(self, request) -> EndpointResult:
+        """
+        Given a NFT ID, sign the message by the NFT's private key.
+        :param request:
+        :return:
+        """
+        if request["nft_id"].startswith("nft"):
```
paninaro, line 1938: It would be nice to use the `AddressType` utils to avoid hard-coding bech32m prefixes.

```suggestion
        if is_valid_address(request["nft_id"], {AddressType.NFT), self.service.config):
```

chia/rpc/wallet_rpc_api.py
```diff
@@ -999,6 +1001,17 @@ async def send_notification(self, request) -> EndpointResult:
...
         return {"tx": tx.to_json_dict_convenience(self.service.config)}
 
+    async def sign_message(self, request) -> EndpointResult:
+        """
+        Given a derived P2 address, sign the message by its private key.
+        :param request:
+        :return:
+        """
+        puzzle_hash: bytes32 = decode_puzzle_hash(request["address"])
+        message = hexstr_to_bytes(request["message"])
```
paninaro, line 1011: If I were calling this RPC and didn't read the docs, I would assume that `"message"` would be a regular string, and not a hex string.

I'd recommend renaming `"message"` to `"message_data"` to at least give a hint that a normal string isn't expected here. Or choose a name that self-documents its type like `"message_hex_string"` or something.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1902,6 +1929,37 @@ async def nft_calculate_royalties(self, request) -> EndpointResult:
...
+        nft: Optional[NFTCoinInfo] = None
+        for wallet in self.service.wallet_state_manager.wallets.values():
+            if wallet.type() == WalletType.NFT.value:
+                assert isinstance(wallet, NFTWallet)
+                nfts: List[NFTCoinInfo] = await wallet.get_current_nfts()
+                for nft in nfts:
+                    if nft.nft_id == nft_id:
+                        nft_wallet = wallet
+                        nft = nft
+                        break
```
paninaro, line 1952: `wallet.get_current_nfts()` could be really expensive, right?

Couldn't we use `wallet.get_nft(nft_id)` instead?

chia/rpc/wallet_rpc_api.py
```diff
@@ -1902,6 +1929,37 @@ async def nft_calculate_royalties(self, request) -> EndpointResult:
...
+                if nft_wallet is not None:
+                    break
+        if nft_wallet is None or nft is None:
+            return {"success": False, "error": f"NFT wallet {nft_id.hex()} doesn't exist."}
```
paninaro, line 1956:
```suggestion
            return {"success": False, "error": f"NFT wallet for {nft_id.hex()} doesn't exist."}
```

chia/rpc/wallet_rpc_api.py
```diff
@@ -1539,13 +1552,27 @@ async def did_transfer_did(self, request) -> EndpointResult:
...
+            did_id = bytes32.from_hexstr(request["did_id"])
+        did_wallet: Optional[WalletProtocol] = None
+        for wallet in self.service.wallet_state_manager.wallets.values():
+            if wallet.type() == WalletType.DECENTRALIZED_ID.value:
+                assert isinstance(wallet, DIDWallet)
+                assert wallet.did_info.origin_coin is not None
+                if wallet.did_info.origin_coin.name() == did_id:
+                    did_wallet = wallet
+        if did_wallet is None:
+            return {"success": False, "error": f"DID wallet {did_id.hex()} doesn't exist."}
```
paninaro, line 1572:
```suggestion
            return {"success": False, "error": f"DID wallet for {did_id.hex()} doesn't exist."}
```

chia/rpc/wallet_rpc_api.py
```diff
@@ -1539,13 +1552,27 @@ async def did_transfer_did(self, request) -> EndpointResult:
...
+            did_id: bytes32 = decode_puzzle_hash(request["did_id"])
+        else:
+            did_id = bytes32.from_hexstr(request["did_id"])
+        did_wallet: Optional[WalletProtocol] = None
+        for wallet in self.service.wallet_state_manager.wallets.values():
+            if wallet.type() == WalletType.DECENTRALIZED_ID.value:
+                assert isinstance(wallet, DIDWallet)
+                assert wallet.did_info.origin_coin is not None
+                if wallet.did_info.origin_coin.name() == did_id:
+                    did_wallet = wallet
```
paninaro, line 1570: break after finding the wallet?

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -531,6 +532,19 @@ async def get_puzzle_info(self, nft_id: bytes32) -> PuzzleInfo:
...
+    async def sign_message(self, message: bytes, nft: NFTCoinInfo) -> Tuple[G1Element, G2Element]:
+        uncurried_nft = UncurriedNFT.uncurry(*nft.full_puzzle.uncurry())
+        if uncurried_nft is not None:
+            p2_puzzle = uncurried_nft.p2_puzzle
+            puzzle_hash = p2_puzzle.get_tree_hash()
+            pubkey, private = await self.wallet_state_manager.get_keys(puzzle_hash)
+            synthetic_secret_key = calculate_synthetic_secret_key(private, DEFAULT_HIDDEN_PUZZLE_HASH)
+            synthetic_pk = synthetic_secret_key.get_g1()
+            prefix = f"\x18Chia Signed Message:\n{len(message)}"
+            return synthetic_pk, AugSchemeMPL.sign(synthetic_secret_key, std_hash(prefix.encode("utf-8") + message))
```
paninaro, line 544: Is there a standard that we're basing this off of? Because it seems odd to me that the length of the message is being encoded in the prefix string as a sequence of characters. That is, if my message is "hello world!" is the expectation that the length (12) is represented as `'1'` and `'2'`?

I would normally expect a construction like the following, but I could be missing something...

`std_hash(len("Chia Signed Message:\n") + "Chia Signed Message:\n".encode("utf-8") + len(message))`

And for length values that don't fit in a single byte, I would assume (again, maybe incorrectly) network byte ordering.

chia/rpc/wallet_rpc_api.py
ytx1991: I discussed this with @trepca. We decided to build multiple APIs because of the following reasons.
1. We want to have different CLI tools for different types of wallets.
2. Different types of wallets have different ways to get private keys.
The redundant code is just a small part, it's more clear to separate them into different functions.

chia/wallet/nft_wallet/nft_wallet.py
ytx1991: Pending on CHIP0002

chia/rpc/wallet_rpc_api.py
trepca: we could break it into two to make managing signing a bit easier on GUI, `sign_message` and `coin_sign_message(type, id, message)`  ?

chia/rpc/wallet_rpc_api.py
```diff
@@ -1002,6 +1003,59 @@ async def send_notification(self, request) -> EndpointResult:
...
+        elif is_valid_address(request["id"], {AddressType.NFT}, self.service.config):
+            target_nft: Optional[NFTCoinInfo] = None
+            for wallet in self.service.wallet_state_manager.wallets.values():
+                if wallet.type() == WalletType.NFT.value:
+                    assert isinstance(wallet, NFTWallet)
+                    nft: Optional[NFTCoinInfo] = await wallet.get_nft(entity_id)
+                    if nft is not None:
+                        selected_wallet = wallet
+                        target_nft = nft
+                    break
```
paninaro, line 1048: Shouldn't the `break` be indented once more?  Otherwise this will stop looping on the first NFT wallet, which may not hold the specified NFT.

paninaro (approved):
lgtm

PR #13509, opened 2022-09-22
parse api method hints earlier, and less, using get_type_hints()

altendky:


chia/server/ws_connection.py
```diff
@@ -53,6 +54,15 @@ def __init__(
         # Local properties
         self.ws: Any = ws
         self.local_type = local_type
+        self.annotations: Dict[str, Type[Streamable]] = {}
+        for name, method in vars(class_for_type(self.local_type)).items():
+            is_api_function = getattr(method, "api_function", False)
+            if not is_api_function:
+                continue
+
+            hint = next(hint for name, hint in get_type_hints(method).items() if name not in {"return", "peer"})
```
altendky, line 63: Shouldn't this be the last one, not the first one?  In terms of matching existing behavior that is.

chia/server/ws_connection.py
```diff
@@ -53,6 +54,19 @@ def __init__(
         # Local properties
         self.ws: Any = ws
         self.local_type = local_type
+        self.annotations: Dict[str, Type[Streamable]] = {}
```
xdustinface, line 57:
```suggestion
        self.request_types: Dict[str, Type[Streamable]] = {}
```

Maybe? Or something else but `annotations` isn't really meaningful here imo.

xdustinface (changes requested):

chia/server/ws_connection.py
altendky: Agreed and fixed per your suggestion.

xdustinface (approved):

emlowe (approved):

PR #13510, opened 2022-09-23
make DBWrapper2 handle more connection setup

altendky:
Optional follow ups:
- only open and close the sql trace log file once
- read the db version in the wrapper, though defaults must be provided still

Draft for:
- [x] self review

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -12,7 +12,6 @@ import traceback
 from pathlib import Path
 from typing import Any, Awaitable, Callable, Dict, List, Optional, Set, Tuple, Union
 
-import aiosqlite
 import sqlite3
 from blspy import AugSchemeMPL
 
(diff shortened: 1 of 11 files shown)


chia/full_node/full_node.py
```diff
@@ -221,37 +219,28 @@ async def _start(self) -> None:
         # These many respond_transaction tasks can be active at any point in time
         self.respond_transaction_semaphore = asyncio.Semaphore(200)
         # create the store (db) and full node instance
-        db_connection = await aiosqlite.connect(self.db_path)
-        db_version: int = await lookup_db_version(db_connection)
+        # TODO: is this standardized and thus able to be handled by DBWrapper2?
```
arvidn, line 222: to answer this question: the DB lookup could be handled by `DBWrapper2`, but the version number that's read from the DB is required to construct the DBWrapper object. So, we could create a temporary DBWrapper with version 1, read the version and the create the final DBWrapper with the correct version.

This is because storing the DB version in the DBWrapper object is quite convenient

arvidn (dismissed):

chia/full_node/full_node.py
altendky: But if the wrapper handled this, maybe only in the `.create()` method, then it could fill it itself 'during creation'.

altendky: I guess there may be a need for 'default' in case of initial creation.

arvidn: > But if the wrapper handled this, maybe only in the `.create()` method, then it could fill it itself 'during creation'.

only for the full-node's blockchain database. We use the `DBWrapper2` for other databases also, where we don't use the version number (yet)

altendky: Could be optional.

emlowe (dismissed):

arvidn (approved):

emlowe (approved):

PR #13512, opened 2022-09-23
server: Create ssl contexts earlier + store them instead of all paths

xdustinface:


chia/server/server.py
```diff
@@ -158,14 +169,43 @@ def __init__(
...
+            _private_cert_path, _private_key_path = None, None
+
         if self._local_type is not NodeType.HARVESTER:
-            self.p2p_crt_path, self.p2p_key_path = public_ssl_paths(root_path, config)
+            p2p_crt_path, p2p_key_path = public_ssl_paths(root_path, config)
+        else:
+            p2p_crt_path, p2p_key_path = None, None
+
+        # If node has public cert use that one for id, if not use private.
+        self.node_id = calculate_node_id(p2p_crt_path or _private_cert_path)
```
altendky, line 182: I feel like this could use some reorganization.  Make the conditions check what is rather than what isn't (also use `==` not `is`), use `if`/`elif`, and then accordingly avoid the `or` object selection at the end.  I personally dislike that form.

xdustinface: Not really what you requested and im not sure where you would put an `elif` there, but i refactored all a bit more with the last force push, let me know!

chia/server/server.py
```diff
@@ -157,15 +164,47 @@ def __init__(
...
-        self.chia_ca_crt_path, self.chia_ca_key_path = chia_ca_crt_key
-        self.node_id = self.my_id()
+            # Public servers
+            public_cert_path, public_key_path = public_ssl_paths(root_path, config)
+            self.ssl_context = ssl_context_for_server(
+                chia_ca_crt_path, chia_ca_key_path, public_cert_path, public_key_path, log=self.log
+            )
+
+        # If node has public cert use that one for id, if not use private.
+        self.node_id = calculate_node_id(private_cert_path if public_cert_path is None else public_cert_path)
```
altendky, line 207: Should probably apply https://gist.github.com/altendky/2d316b1e93f99352868f73623bb85ce3 (or I can PR separately).  It points out that `private_cert_path` is `Optional` at this point when expected to be a `Path`.

chia/server/server.py
```diff
@@ -157,15 +164,47 @@ def __init__(
...
 
-        if self._local_type is not NodeType.INTRODUCER:
-            self._private_cert_path, self._private_key_path = private_ssl_paths(root_path, config)
-        if self._local_type is not NodeType.HARVESTER:
-            self.p2p_crt_path, self.p2p_key_path = public_ssl_paths(root_path, config)
+        ca_private_crt_path, ca_private_key_path = private_ca_crt_key
+        chia_ca_crt_path, chia_ca_key_path = chia_ca_crt_key
+
+        private_cert_path, private_key_path = None, None
+        public_cert_path, public_key_path = None, None
```
altendky, line 171: I personally would rather see a line per for this.  I don't think the build-a-tuple-then-unpack-it does much good here.

```suggestion
        private_cert_path = None
        private_key_path = None
        public_cert_path = None
        public_key_path = None
```

Or even add the hints here.

chia/server/server.py
```diff
@@ -99,6 +99,13 @@ def ssl_context_for_client(
     return ssl_context
 
 
+def calculate_node_id(cert_path: Path) -> bytes32:
+    pem_cert = x509.load_pem_x509_certificate(cert_path.read_bytes(), default_backend())
```
altendky, line 103: I think this loads `cert_path` a second time, though I also think that the other load is buried down in Python stdlib code so we may not be able to save the double load.  Unless maybe we already have the `pem_cert` object from the previous load and could pass it in?  I'm not sure what all is in the `ssl_context`.

altendky (changes requested):
The hinting concern from the previous review is the only one that needs addressed.  The others are comments that may or may not inspire your interest in any change.

chia/server/server.py
altendky: The situation here is non-trivial but there doesn't appear to be an actual bug.  Hinting can be dealt with later.  In short, `self._local_type` is always in either `authenticated_client_types` or `authenticated_server_types` (`private_cert_path` is not `None`), or it is in neither (`public_cert_path` is not `None`).

altendky (approved):

emlowe (approved):

PR #13513, opened 2022-09-23
server: Drop some `ChiaServer` attributes

xdustinface:


altendky (commented):
The removal of the unused task attributes seems good.  The shift towards direct use of global defaults doesn't feel like an improvement to me.  It seems to mostly reduce control and visibility.  But, fine for someone else to approve if they like it.  I don't think it causes a problem.

altendky (approved):

emlowe (approved):
nice to remove unused variables

PR #13514, opened 2022-09-23
util: Refactor `select_port` to not load the config

xdustinface:


chia/rpc/rpc_server.py
```diff
@@ -429,7 +429,8 @@ async def start_rpc_server(
 
         rpc_server = RpcServer.create(rpc_api, rpc_api.service_name, stop_cb, root_path, net_config)
         rpc_server.rpc_api.service._set_state_changed_callback(rpc_server.state_changed)
-        await rpc_server.start(root_path, self_hostname, rpc_port, max_request_body_size)
+        prefer_ipv6 = bool(net_config.get("prefer_ipv6", False))
```
altendky, line 432: `bool()` isn't a great converter.  For example, `bool("false")` is `True` since `"false"` is a non-empty string.  But hey, maybe we look at this as just be explicit about it earlier and someday we'll have a schema that rejects the string `"false"` entirely.

altendky (dismissed):

chia/server/server.py
```diff
@@ -279,7 +279,7 @@ async def start_server(self, on_connect: Callable = None):
         # prefer_ipv6 is set in which case we use the IPv6 port
         #
         if self._port == 0:
-            self._port = select_port(self.root_path, self.runner.addresses)
+            self._port = select_port(self.config.get("prefer_ipv6", False), self.runner.addresses)
```
emlowe, line 282: I don't believe this works since prefer_ipv6 is a global config item at the top level of the config, and the config entry passed into the ChiaServer is the specific entry for that service, so this will always be false and never read the actual prefer_ipv6 setting.

emlowe (changes requested):

chia/server/server.py
altendky: Sounds right.  `:[`  Thanks for catching this.

xdustinface: Thanks! Fixed in 23992f7c3f93f732213a28c44383ec21fe50433e.

chia/rpc/rpc_server.py
xdustinface: Good point, see 5caa80becffb3a53b6d9ecd662c8547944ea925f

emlowe:
I'm not entirely sure this is much of an improvement, but I guess it ends up in one less read of the config file.

xdustinface:
> it ends up in one less read of the config file.

Thats the point, no reason to reload it there if we can just pass the stuff in and its 2 times per service start which makes it a bunch in tests.

emlowe (approved):

PR #13515, opened 2022-09-23
util|rpc: Refactor `RpcEnvironment` -> `WebServer`

xdustinface:
I want to use it in `ChiaServer`, `DataLayerServer` and `WebSocketServer` also.. this PR changes:
- Move `rpc.rpc_server.RpcEnviroment` -> `util.network.WebServer`
- Move the creation into `WebServer`
- Make sure the shutdown does what it's supposed to do
- Drop not needed `site` attribute

Based on: 
- #13514 
- #13518

### DIFF
--- a/chia/rpc/rpc_server.py
+++ b/chia/rpc/rpc_server.py
@@ -188,7 +188,6 @@ class RpcServer:
             await self.client_session.close()
         if self.webserver is not None:
             await self.webserver.await_closed()
-            self.webserver = None
         if self.daemon_connection_task is not None:
             await self.daemon_connection_task
             self.daemon_connection_task = None
--- a/chia/rpc/rpc_server.py
+++ b/chia/rpc/rpc_server.py
@@ -21,7 +21,7 @@ from chia.util.byte_types import hexstr_to_bytes
 from chia.util.config import str2bool
 from chia.util.ints import uint16
 from chia.util.json_util import dict_to_json_str
-from chia.util.network import select_port
+from chia.util.network import WebServer, select_port
 from chia.util.ws_message import WsRpcMessage, create_payload, create_payload_dict, format_response, pong
 
 log = logging.getLogger(__name__)
@@ -30,13 +30,6 @@ max_message_size = 50 * 1024 * 1024  # 50MB
 
 EndpointResult = Dict[str, Any]
 Endpoint = Callable[[Dict[str, object]], Awaitable[EndpointResult]]
-
-
-@dataclass(frozen=True)
-class RpcEnvironment:
-    runner: web.AppRunner
-    site: web.TCPSite
-    listen_port: uint16
 
 
 class StateChangedProtocol(Protocol):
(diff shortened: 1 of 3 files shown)


chia/util/network.py
```diff
@@ -1,12 +1,73 @@
...
+        await self.runner.cleanup()
+
+    def close(self) -> None:
+        self._close_task = asyncio.create_task(self._close())
+
+    async def await_closed(self) -> None:
+        if self._close_task is None:
+            raise RuntimeError("WebServer stop not triggered")
+        await self._close_task
+        self._close_task = None
```
altendky, line 68: As above with clearing the attribute.  Not sure it is important or valuable or leading to a good place.

chia/util/network.py
```diff
@@ -1,12 +1,73 @@
...
+        shutdown_timeout: int = 60,  # Default `shutdown_timeout` from web.TCPSite
+        prefer_ipv6: bool = False,
+        logger: logging.Logger = web_logger,
+    ) -> WebServer:
+        app = web.Application(client_max_size=max_request_body_size, logger=logger)
+        runner = web.AppRunner(app, access_log=None, keepalive_timeout=keepalive_timeout)
+
+        runner.app.add_routes(routes)
+        await runner.setup()
+        site = web.TCPSite(runner, hostname, int(port), ssl_context=ssl_context, shutdown_timeout=shutdown_timeout)
```
altendky, line 45: It feels awkward to drop the reference to this.  I _think_ I dug through this and `runner` is passed in so that the site can register itself onto the runner or something, so there is a thing holding a reference to the object.  But.  meh.  Seems like it wouldn't be that bad to just hold references to each thing created?  Application, runner, and site.  Feels both kinda pointless and kinda cleaner.

chia/rpc/rpc_server.py
```diff
@@ -171,38 +164,31 @@ def create(
...
         if self.websocket is not None:
             await self.websocket.close()
         if self.client_session is not None:
             await self.client_session.close()
-        if self.environment is not None:
-            await self.environment.runner.cleanup()
-            self.environment = None
+        if self.webserver is not None:
+            await self.webserver.await_closed()
+            self.webserver = None
```
altendky, line 191: Is this useful?  Feels like drifting towards object reuse which is hard to get right and I don't think we need it.  If kept, should it be cleared in a finally block to be sure it happens regardless of shutdown failure?  Or maybe it explicitly shouldn't, but what then?

altendky (commented):
Not sure if I'm asking for changes here or not.  Let me know what you think.

chia/util/network.py
xdustinface: Hm i don't see why we should add them if we anyway don't need to access them again + yeah, the `AppRunner` holds them in 
https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.AppRunner.app
https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRunner.sites

altendky (dismissed):

altendky (dismissed):

altendky (approved):

emlowe (approved):
I like this

PR #13519, opened 2022-09-23
Fix a breaking change in get_puzzle_and_solution RPC

Quexington:


chia/rpc/full_node_rpc_api.py
```diff
@@ -702,9 +701,10 @@ async def get_puzzle_and_solution(self, request: Dict) -> EndpointResult:
         if error is not None:
             raise ValueError(f"Error: {error}")
 
-        puzzle_ser: SerializedProgram = SerializedProgram.from_program(Program.to(puzzle))
-        solution_ser: SerializedProgram = SerializedProgram.from_program(Program.to(solution))
```
arvidn, line 706: avoiding these conversions look like an optimization, does it affect correctness?

arvidn:
@Quexington can you explain what the problem was?

chia/rpc/full_node_rpc_api.py
arvidn: `Program.to()` does not support converting from `SerializedProgram`

arvidn (approved):

Quexington:
Explaining for github purposes: the RPC wasn't tested at all which is how this slipped through, but `get_puzzle_and_solution_for_coin` switched its API to return `SerializedProgram` which was causing an error when passed to `Program.to()`

trepca (approved):

PR #13533, opened 2022-09-25
drop reply_type api decorator

altendky:
Draft for:
- [ ] Discuss the historical or future need for `@reply_type`


altendky:
So it appears that the decorator was unused even when introduced in https://github.com/Chia-Network/chia-blockchain/commit/e9bf0ec12b6ae727e7103a1556358d6623549786.  But it seems a decent alternative to the mappings that were added there as well.  I think I'll plan on exploring that route rather than removal of the decorator here.

PR #13542, opened 2022-09-26
more explicit and complete handling of api decorator data

altendky:


xdustinface (commented):
Thoughts about adding 960ea95d3dbcc366f00f3247a18fa45fc202a146 on top here which requires some reordering though? Like forcing the decorator order, dropping the `api_function` any only apply the metadata when applying the `api_request` decorator. Also getting rid of this "try to set the metadata always and set it even again in `api_request` on the final result" which confused me while reviewing :)

xdustinface (dismissed):
I still think the commit above would be helpful but let's go without assumed this will soon become a single `api_request` decorator with the other decorators merged in as parameter.

xdustinface (approved):

emlowe (approved):

PR #13557, opened 2022-09-27
Use codespell to catch some typos

hoffmang9:
I can also add or create a subsequent PR to add codespell to the pre-commit but it will require a few manual exceptions.


altendky (approved):
They all read well, at least for a quick read.

paninaro (approved):
loooks good to me

PR #13559, opened 2022-09-28
embrace click 8

altendky:
likely replacing https://github.com/Chia-Network/chia-blockchain/pull/11361

### DIFF
--- a/chia/cmds/chia.py
+++ b/chia/cmds/chia.py
@@ -32,22 +32,6 @@ from typing import Optional
 CONTEXT_SETTINGS = dict(help_option_names=["-h", "--help"])
 
 
-def monkey_patch_click() -> None:
-    # this hacks around what seems to be an incompatibility between the python from `pyinstaller`
-    # and `click`
-    #
-    # Not 100% sure on the details, but it seems that `click` performs a check on start-up
-    # that `codecs.lookup(locale.getpreferredencoding()).name != 'ascii'`, and refuses to start
-    # if it's not. The python that comes with `pyinstaller` fails this check.
-    #
-    # This will probably cause problems with the command-line tools that use parameters that
-    # are not strict ascii. The real fix is likely with the `pyinstaller` python.
-
-    import click.core
-
-    click.core._verify_python3_env = lambda *args, **kwargs: 0  # type: ignore[attr-defined]
-
-
 @click.group(
     help=f"\n  Manage chia blockchain infrastructure ({__version__})\n",
     epilog="Try 'chia start node', 'chia netspace -d 192', or 'chia show -s'",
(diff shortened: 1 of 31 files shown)


chia/cmds/chia.py
```diff
@@ -149,7 +133,6 @@ def run_daemon_cmd(ctx: click.Context, wait_for_unlock: bool) -> None:
 
 
 def main() -> None:
-    monkey_patch_click()
```
hoffmang9, line 152: This monkeypatch was added as having a non standard locale meant that a large minority of macos users could not plot using the GUI as the command would fail when passed from the GUI. Are we comfortable that that is no longer an issue?

altendky: To shortcut a few links, https://github.com/pallets/click/commit/9072f1809e559898814d58da77b9b9d5a73233fb is the commit that entirely deletes the function we monkey patch.

hoffmang9: Them dropping locale certainly seems like it would help. On the testing side @richardkiss would have been the test writer on this if one was written.

altendky: I was more curious what group of users would be affected by this in case we have one in house that could recreate the issues with click 7 without monkey patching and then see if this pr works in that same situation.

emlowe: I manually tested this build on Intel with macOS 12.5.1 (Monterey), macOS 11.6.4 (Big Sur), macOS 10.15.7 (Catalina) and macOS 10.14.6 (Mojave) - all worked fine running and making a plot from the GUI. I'll note a caveat that I only testing the English/North America locale/region setting.

altendky: From source?  Or via the dmgs.

emlowe: DMGs from this PR

emlowe (approved):

chia/cmds/chia.py
richardkiss: @emlowe Were you able to reproduce the problem with Click 7?

emlowe: I didn't try that but I suppose I could - but that would require a build

emlowe: I could not reproduce any problems with a build that just removes the monkey path and uses click 7 (tested on Monterey and Mojave).

emlowe: I'm not sure what combination of python/pyinstaller/click were involved in the original need to monkey-patch click. It seems that possibly hasn't been needed due to other changes elsewhere (pyinstaller or the build systems?). In any event, I think we should move forward with this PR. To that end, I am resolving this and undrafting.

PR #13565, opened 2022-09-28
refactor `@api_request` decorator and family to use a single decorator

altendky:


chia/util/api_decorators.py
```diff
@@ -1,29 +1,21 @@
...
-if TYPE_CHECKING:
-    from chia.server.ws_connection import WSChiaConnection
-
-    converted_api_f_type = Union[
-        Callable[[Union[bytes, _T_Streamable]], Coroutine[Any, Any, Optional[Message]]],
-        Callable[[Union[bytes, _T_Streamable], WSChiaConnection], Coroutine[Any, Any, Optional[Message]]],
-    ]
+P = ParamSpec("P")
+R = TypeVar("R")
+S = TypeVar("S", bound=Streamable)
```
xdustinface, line 16:
```suggestion
S = TypeVar("S", bound=Streamable)
```

Why are you dropping the one from the streamable module? We could just make it public there maybe?

chia/util/api_decorators.py
```diff
@@ -34,104 +26,66 @@ class ApiMetadata:
...
     message_class: Optional[Any] = None
 
 
-def get_metadata(function: Callable[..., Any]) -> ApiMetadata:
+def get_metadata(function: Callable[..., object]) -> ApiMetadata:
     maybe_metadata: Optional[ApiMetadata] = getattr(function, metadata_attribute_name, None)
     if maybe_metadata is None:
         return ApiMetadata()
 
     return maybe_metadata
```
xdustinface, line 38:
```suggestion
def get_metadata(function: Callable[..., object]) -> Optional[ApiMetadata]:
    return getattr(function, metadata_attribute_name, None)
```

I think ^ to return an `Optional[ApiMetadata]` and then drop `ApiMetadata.api_function` member is cleaner so that we can just check for methods which have an actual metadata assigned instead of returning this default object. Thoughts?

xdustinface (approved):
Looks good, two maybe changes though :)

chia/util/api_decorators.py
altendky: There is simplicity in both options.  I haven't managed to come up with a clear winner.  I'll switch it.

chia/util/api_decorators.py
altendky: I don't have a fully formed plan or policy around handling these `TypeVar`s, but I don't think that at least trivial ones like this benefit much from being shared.  Happy to chat about it towards the end of coming up with preferred practices and their reasons.

emlowe (approved):

PR #13568, opened 2022-09-29
commonize signal setup

altendky:
the minimal fix https://github.com/Chia-Network/chia-blockchain/pull/13570

This provides a single set of helpers to configure either synchronous or asynchronous signal handlers for shutdown and replaces all existing signal handling configuration with this one consistent and cross platform solution.

I did run this on an ephemeral testnet and got Chris' stamp of 'pass on basic functionality' there.

Draft for:
- [x] self consideration
- [x] ack!
- [x] tests like https://github.com/Chia-Network/chia-blockchain/pull/13576 (and more)
- [x] seeder too https://github.com/Chia-Network/chia-blockchain/pull/16079


chia/data_layer/data_layer_server.py
```diff
@@ -110,7 +96,12 @@ async def file_handler(self, request: web.Request) -> web.Response:
...
         return response
 
-    def _accept_signal(self, signal_number: int, stack_frame: Any = None) -> None:
+    def _accept_signal(
+        self,
+        signal_number: int,
+        stack_frame: Optional[FrameType],
+        loop: asyncio.AbstractEventLoop,
+    ) -> None:
         self.log.info("Got SIGINT or SIGTERM signal - stopping")
```
arvidn, line 105: it might even be nice to do something like: `f"Got signal {signal_number} - stopping"`

chia/data_layer/data_layer_server.py
```diff
@@ -110,7 +96,12 @@ async def file_handler(self, request: web.Request) -> web.Response:
...
 
         self.close()
```
arvidn, line 107: I take it we're fine with calling this multiple times, yeah?

chia/server/start_service.py
```diff
@@ -238,23 +236,14 @@ async def setup_process_global_state(self) -> None:
...
+        setup_sync_signal_handler(handler=self._accept_signal)
 
-    def _accept_signal(self, signal_number: int, stack_frame: Optional[FrameType] = None) -> None:
+    def _accept_signal(
+        self,
+        signal_number: int,
+        stack_frame: Optional[FrameType],
+        loop: asyncio.AbstractEventLoop,
+    ) -> None:
         self._log.info(f"got signal {signal_number}")
```
arvidn, line 247: I think this signal handler is problematic. It's kind of unrelated to your patch though. I think this might be the reason I sometimes have to `kill -9` left-over processes. if we won't tell the main process that the user wants to shut down, we should at least shut down *this* process, not just drop it.

chia/util/misc.py
```diff
@@ -165,3 +170,75 @@ def to_batches(to_split: Collection[T], batch_size: int) -> Iterator[Batch[T]]:
...
+    else:
+        for signal_ in [signal.SIGINT, signal.SIGTERM]:
+            loop.add_signal_handler(
+                signal_,
+                functools.partial(handler, signal_number=signal_, stack_frame=None, loop=loop),
+            )
+
+
+# signals are a global thing so...  global task reference maintenance is "ok" i guess
+signal_handler_tasks: List[asyncio.Task[None]] = []
```
arvidn, line 202: I would have thought the point of having this list would be to join them when shutting down, but I don't see that anywhere

arvidn (commented):
looks good in general. I just had some minor questions

altendky (commented):
Thanks for the ideas, I think some points are worth not addressing here and others are.

chia/util/misc.py
altendky: https://github.com/Chia-Network/chia-blockchain/commit/e020ff2e191e2d3b37587c948214b6d31cfc73e6 I think there's more refactor that could be worthwhile around this, but maybe this is an ok layer to stop at for this PR.

chia/server/start_service.py
arvidn: > Are you referring to having to kill workers? Or main processes.

I believe worker processes

>  Why would the subprocesses get signals to begin with? Does the signal to the main process get propagated to all? 

I think there are certain failure modes where we don't properly shut down the workers. Those worker processes can then not be killed with `SIGINT`, but only `SIGKILL`

arvidn (approved):

PR #13569, opened 2022-09-29
util: Set log levels per handler / Fix the log level in beta mode

xdustinface:
Prior to this PR the log level from the beta config (`DEBUG`) wasn't used. So the beta mode still was depending on the config level of the service being `DEBUG`.

This PR refactors the logging setup a bit to make it even possible to set a specific log level for each handler which is what we need to make the `beta.log` level independent from the normal `debug.log` level.

### DIFF
--- a/chia/util/chia_logging.py
+++ b/chia/util/chia_logging.py
@@ -92,7 +92,7 @@ def initialize_logging(service_name: str, logging_config: Dict, root_path: Path,
     # the potentially smaller log levels of specific handlers.
     root_logger.setLevel(min(handler.level for handler in handlers))
 
-    if root_logger.level < logging.INFO:
+    if root_logger.level <= logging.DEBUG:
         logging.getLogger("aiosqlite").setLevel(logging.INFO)  # Too much logging on debug level
 
 
--- a/chia/util/chia_logging.py
+++ b/chia/util/chia_logging.py
@@ -73,18 +73,19 @@ def initialize_logging(service_name: str, logging_config: Dict, root_path: Path,
         handlers.append(get_file_log_handler(file_log_formatter, beta_root_path, get_beta_logging_config()))
 
     root_logger = logging.getLogger()
-    log_level_invalid = False
+    log_level_exceptions = set()
     for handler in handlers:
         try:
             handler.setLevel(log_level)
-        except Exception:
+        except Exception as e:
             handler.setLevel(default_log_level)
-            log_level_invalid = True
+            log_level_exceptions.add(str(e))
         root_logger.addHandler(handler)
 
-    if log_level_invalid:
+    if len(log_level_exceptions) > 0:
         root_logger.error(
-            f"Invalid log level found in {service_name} config: {log_level}. Defaulting to: {default_log_level}"
+            f"Invalid log level found in {service_name} config: {log_level}. Defaulting to: {default_log_level}. "
+            f"Errors: {', '.join(log_level_exceptions)}"
         )
 
     # Adjust the root logger to the smallest used log level since its default level is WARNING which would overwrite
--- a/chia/util/chia_logging.py
+++ b/chia/util/chia_logging.py
@@ -73,19 +73,19 @@ def initialize_logging(service_name: str, logging_config: Dict, root_path: Path,
         handlers.append(get_file_log_handler(file_log_formatter, beta_root_path, get_beta_logging_config()))
 
     root_logger = logging.getLogger()
-    log_level_exceptions = set()
+    log_level_exceptions = {}
     for handler in handlers:
         try:
             handler.setLevel(log_level)
         except Exception as e:
             handler.setLevel(default_log_level)
-            log_level_exceptions.add(str(e))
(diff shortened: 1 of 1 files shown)


chia/util/chia_logging.py
```diff
@@ -38,56 +40,59 @@ def get_file_log_handler(
...
     if beta_root_path is not None:
-        logger.addHandler(get_file_log_handler(file_log_formatter, beta_root_path, get_beta_logging_config()))
+        handlers.append(get_file_log_handler(file_log_formatter, beta_root_path, get_beta_logging_config()))
+
+    root_logger = logging.getLogger()
+    log_level_invalid = False
+    for handler in handlers:
+        try:
+            handler.setLevel(log_level)
+        except Exception:
```
altendky, line 80:
```
>>> handler.setLevel("lkajdsflk")
  File "/usr/lib/python3.8/logging/__init__.py", line 916, in setLevel
    self.level = _checkLevel(level)
  File "/usr/lib/python3.8/logging/__init__.py", line 198, in _checkLevel
    raise ValueError("Unknown level: %r" % level)
ValueError: Unknown level: 'lkajdsflk'
```

Let's at least restrict this to handling `ValueError`.  Maybe the message check too.

```suggestion
        except ValueError as e:
            if "unknown level" not in str(e).lower():
                raise
```

chia/util/chia_logging.py
```diff
@@ -38,56 +40,59 @@ def get_file_log_handler(
...
+    if log_level_invalid:
+        root_logger.error(
+            f"Invalid log level found in {service_name} config: {log_level}. Defaulting to: {default_log_level}"
+        )
+
+    # Adjust the root logger to the smallest used log level since its default level is WARNING which would overwrite
+    # the potentially smaller log levels of specific handlers.
+    root_logger.setLevel(min(handler.level for handler in handlers))
+
+    if root_logger.level < logging.INFO:
```
altendky, line 94: I think we want this.  The log level could be in between info and debug without triggering the aiosqlite debug so the boundary for this should be where debug actually triggers.

```suggestion
    if root_logger.level <= logging.DEBUG:
```

altendky (changes requested):

chia/util/chia_logging.py
xdustinface: Hm we then depend on the error message and there is also a `TypeError` btw https://github.com/python/cpython/blob/4652093e1b816b78e9a585d671a807ce66427417/Lib/logging/__init__.py#L208-L218. Initially i had `(ValueError, TypeError)` but went back to `Exception` to not depend on the errors they raise there. Isn't logging an error about the conversion failure enough here if i maybe just add the exception to the log?

altendky: I suppose this is a corner where not raising, enabling as many loggers as we can, and then logging the error to all of them may be better.  I guess instead of `log_level_invalid` you could just collect a list of exceptions that need reported.

xdustinface: > enabling as many loggers as we can, and then logging the error to all of them may be better

Not sure if i follow this, thats what the code currently does, no? (without the exception in the log yet though, which is supposed to be the same for all loggers anyway, no?)

altendky: Consuming an exception and reporting what you intended it to be is importantly different than reporting what the error actually was.  But sure, not all that different.

xdustinface: > Isn't logging an error about the conversion failure enough here if i maybe just add the exception to the log?

I wonder if you read this question ^ in my reply above? :) If so, i don't get the response, but anyway, im adding the exception to the log.

altendky: The direct answer is yes.  The more confusing reply that includes my various considerations...  is above.

xdustinface: Haha, i see. Okay here we go https://github.com/Chia-Network/chia-blockchain/pull/13569/commits/9035286326dde3a5a6ca3c7ba1a40e5a9bb4d7fd. There is no connection between the error and the handler now but.. its anyway always the same log level and with that the same error unless we have some custom handler with special validation in `setLevel` but... 

Example: 

```
2022-09-29T20:24:16.474 harvester root                    : ERROR    Invalid log level found in harvester config: Isdjfo. Defaulting to: WARNING. Errors: Unknown level: 'Isdjfo'
```

chia/util/chia_logging.py
xdustinface: Ops, actually could have just accepted here but see https://github.com/Chia-Network/chia-blockchain/pull/13569/commits/2e54d82d2a0e077ce6633b444fb50ef91aa1bada

chia/util/chia_logging.py
xdustinface: Okay well, i just added the handler also, now it prints multiple errors but at least there is a connection between the handler and the error 3712a2bebde930cc5d7770c3e0eea6ec5625e994

altendky (approved):

emlowe (approved):

PR #13570, opened 2022-09-29
minimal fixup for daemon signal handling regression

altendky:
the break point https://github.com/Chia-Network/chia-blockchain/pull/13400

the...  not-minimal "fix" https://github.com/Chia-Network/chia-blockchain/pull/13568

`killall chia_daemon` on main doesn't seem to do much.  With this the daemon process terminated promptly.  (in linux)

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -1,4 +1,5 @@
 import asyncio
+import functools
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


xdustinface (approved):

emlowe (commented):
I'll note that `chia/server/start_server.py:setup_process_global_state` has some specific code for windows that is not in this code.

altendky:
Yes, this is a focused fix for the regression introduced in https://github.com/Chia-Network/chia-blockchain/pull/13400.  An overhaul is being explored in https://github.com/Chia-Network/chia-blockchain/pull/13568 that introduces parity across operating systems and services.

emlowe (approved):

PR #13578, opened 2022-09-30
server: Use `get_extra_info` to get the `ssl_object`

xdustinface:


chia/server/server.py
```diff
@@ -310,7 +318,12 @@ async def incoming_connection(self, request):
         ws = web.WebSocketResponse(max_msg_size=max_message_size)
         await ws.prepare(request)
         close_event = asyncio.Event()
-        cert_bytes = request.transport._ssl_protocol._extra["ssl_object"].getpeercert(True)
+        ssl_object = request.get_extra_info("ssl_object")
```
emlowe, line 321: I think you can directly get the peercert here with `request.get_extra_info("peercert") - https://docs.python.org/3/library/asyncio-protocol.html?highlight=peercert#asyncio.BaseTransport.get_extra_info

xdustinface: Im on mobile so i cant verify but looks like then its called with `binary_form=False` and we would need to convert it?

xdustinface: https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.getpeercert

emlowe: Ah I see, we want `binary_form=True`, but `request.get_extra_info("peercert")` will call it as the default which is `False`. Ok resolving comment

emlowe (approved):

PR #13582, opened 2022-10-01
convert datalayer to DBWrapper2

altendky:
Draft for:
- [x] Use `.reader()` when appropriate

### DIFF
--- a/chia/util/db_wrapper.py
+++ b/chia/util/db_wrapper.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
 from types import TracebackType
-from typing import Any, AsyncIterator, Dict, Generator, Iterable, Optional, Type, Union
+from typing import Any, AsyncIterator, Dict, Generator, Iterable, List, Optional, Type, Union
 
 import aiosqlite
 from typing_extensions import final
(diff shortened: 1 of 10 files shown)


chia/util/db_wrapper.py
```diff
@@ -163,12 +117,18 @@ async def create(
         log_path: Optional[Path] = None,
         journal_mode: str = "WAL",
         synchronous: Optional[str] = None,
+        foreign_keys: bool = False,
+        row_factory: Optional[Type[aiosqlite.Row]] = None,
```
arvidn, line 121: would you mind putting a `*,` somewhere high up here too? I think we have too many parameters now to allow non-named ones

tests/core/data_layer/test_data_store.py
```diff
@@ -195,15 +203,14 @@ async def test_insert_internal_node_does_nothing_if_matching(data_store: DataSto
     ancestors = await data_store.get_ancestors(node_hash=kv_node.hash, tree_id=tree_id)
     parent = ancestors[0]
 
-    async with data_store.db_wrapper.locked_transaction():
-        cursor = await data_store.db.execute("SELECT * FROM node")
+    async with data_store.db_wrapper.reader() as reader:
+        cursor = await reader.execute("SELECT * FROM node")
```
arvidn, line 207: probably not for this PR, but there is an `execute_fetchall()` function which would save you one round-trip to the sqlite thread

arvidn (dismissed):
very nice!

arvidn (approved):

emlowe (approved):

PR #13587, opened 2022-10-03
include windows coverage

altendky:
https://github.com/nedbat/coveragepy/issues/991

Draft for:
- [x] See if it works
- [x] If it does, wait for it to be released


emlowe (approved):
nice

PR #13603, opened 2022-10-05
rework around tx records and offer takes

altendky:
Some adjustments to reduce flakes by waiting for transactions to be fully processed instead of just farming blocks and hoping.

Draft for:
- Break outs
  - [x] https://github.com/Chia-Network/chia-blockchain/pull/13629


altendky:
Kick CI

altendky:
I did chat with quex to get an approval for the conceptual change here returning the transaction records along with the offer record.  They did suggest exposing that through the RPCs as well, though I would rather save that for another PR given the size of this already.

A 'real' PR review is still required for the details.

chia/wallet/trade_manager.py
```diff
@@ -788,7 +790,7 @@ async def respond_to_offer(
         for tx in tx_records:
             await self.wallet_state_manager.add_transaction(tx)
 
-        return True, trade_record, None
+        return True, trade_record, [push_tx, *tx_records], None
```
altendky, line 793: @Quexington, Almog reviewed this PR but called me out on including the `push_tx` here.  Neither of us are really sure what it is and if it should be included.  I see the comment above `Dummy transaction for the sake of the wallet push` but i'm not quite clear on the details from that.  Could you help us understand this?  Thanks.

Quexington: It's been a while so I'm not 100% but I think there was an issue with calling add_pending_transaction for all of the generated TX records causing unexpected statuses of the transaction.  I'm not sure if the wallet still works the way it did when I wrote it to make this necessary, but the dummy TX was supposed to contain the spend bundle that went out to the network while the rest of the actually visible transactions just got added to the database.  Looking at it now, it _seems_ like you should just be able to stick the spend bundle onto the first generated tx record and then add all of them with `add_pending_transaction`.  Regardless, for what the rest of this PR is trying to do, you need to return `push_tx` here or stuff will not work.

altendky: @Quexington, should we only be returning the `push_tx`?  It sounds like it is the only real transaction here?  I feel like there should be some comment around this but I'm not sure what should really be said.

Quexington: It depends what you're looking for.  The UI for example only would only care about tx_records because that's the actual summary of the balances changed in the wallets.  But the tests care about the actual on chain objects which are contained within the dummy tx.

altendky: I guess it sounds like these are two separate things that don't belong in a list together.  `push_tx` is sent to the network and is what the tests should wait on.  `tx_records` is irrelevant to this use in the tests.  A local run with just `[push_tx]` seems to be passing.  I think.  Does that sound sensible instead?

```suggestion
        # push_tx is for the network, tx_records is for local wallet bookkeeping
        return True, trade_record, [push_tx], None
```

Quexington: Is this PR just for making the tests good or is the plan to return these from the RPC?  If so we need to return the actual tx records too.  If not then yeah we can only return push_tx

altendky: This PR was intended to be for the tests but you suggested/requested that the records should be returned through the RPC as well.  I was game for that but wanted it as a separate PR.  So, this prepping for that at this layer is fine by me.  So, they should all be returned.  But, they should be separate things, right?  So that the caller can pick which group of things they want without indexing the first item in the list or slicing the rest of the list.

almogdepaz: that sounds good to me

altendky: This feels like further mixing the separate data.  Am I misunderstanding something?

altendky: So for the next layer, exposing this out the RPCs, we'll need separate returning of spend bundles and transaction records anyways, it seems.  All the more reason for them to be separate here?  Return a class with a `.spend_bundles` attribute and another for `.transaction_records`?

Quexington: When shipped over the RPCs, the spendbundles will get dropped.  Anybody looking at tx records on the RPC will have no need for the actual spend bundle.

almogdepaz: i like the idea of a class much better then whenever we use RPC's we must call to_json_dict_convenience or otherwise might get some dummy tx retuned with our spendbundle

theres lots of hidden information here not sure if its described anywhere

Quexington: I'm suggesting we remove the dummy TX so the only consequence of not calling `to_json_dict_convenience` would be that you get the spend bundle that you don't need, but you could still just ignore that.  There's other reasons for using that method that would likely make any user of the RPC request that you start using it (consistency between the other RPCs for example)

altendky: I think there's room for more cleanup around this but I'm not sure I see it all clearly enough to do it right now.  I'd suggest we go ahead with reviews on this knowing that there's likely some second pass to take on these return values at some point.  Also maybe some adjustments to the spend-bundle/transaction-record relationships, but that's where I don't have enough insight presently.

altendky: Since we have multiple people in this conversation maybe we can get a thumbs up from each here to confirm if closure of the conversation is ok or not.  Otherwise I'll leave it open and it blocks the PR.  Which is the point if you don't feel the conversation should be over here.

Quexington (approved):
We have one item that we need to revisit, but otherwise this seems like a good, carefully done refactor

emlowe (approved):

PR #13606, opened 2022-10-05
simulator: Bump plot refresh duration check to 10s

xdustinface:
Try to avoid flakiness like in 
https://github.com/Chia-Network/chia-blockchain/actions/runs/3185807311/jobs/5195866349#step:16:487


xdustinface:
There is already #13602

PR #13612, opened 2022-10-05
Minor Simulation CI test adjustments

emlowe:
Rearranged some code. Check return codes in order to fail faster on errors. Adjusted peak height checks such that the actual height will be reported in a failure (as opposed to just true/false).
Ignore an RPC timeout in one specific RPC call. Possibly this is incorrect, however, if the subsequent tests pass following this timeout, it would seem to be ok. 
Adjust the start simulator tests to use a temporary keyring and to properly cleanup any created temporary directories

### DIFF
--- a/chia/simulator/simulator_test_tools.py
+++ b/chia/simulator/simulator_test_tools.py
@@ -28,20 +28,21 @@ from chia.util.ints import uint32
 from chia.util.keychain import Keychain
 from chia.util.lock import Lockfile
 from chia.wallet.derive_keys import master_sk_to_wallet_sk
+from tests.util.keyring import TempKeyring
 
 """
 These functions are used to test the simulator.
 """
 
 
-def mnemonic_fingerprint() -> Tuple[str, int]:
+def mnemonic_fingerprint(keychain: Keychain) -> Tuple[str, int]:
     mnemonic = (
         "today grape album ticket joy idle supreme sausage "
         "oppose voice angle roast you oven betray exact "
         "memory riot escape high dragon knock food blade"
     )
     # add key to keychain
-    sk = Keychain().add_private_key(mnemonic)
+    sk = keychain.add_private_key(mnemonic)
     fingerprint = sk.get_g1().get_fingerprint()
     return mnemonic, fingerprint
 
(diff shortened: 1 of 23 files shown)


chia/simulator/simulator_test_tools.py
```diff
@@ -28,20 +28,21 @@
 from chia.util.keychain import Keychain
 from chia.util.lock import Lockfile
 from chia.wallet.derive_keys import master_sk_to_wallet_sk
+from tests.util.keyring import TempKeyring
```
jack60612, line 31: We can't import from tests in this file

jack60612: The simulator tests are designed to run standalone, not with the not exposed tests library

tests/simulation/test_start_simulator.py
```diff
@@ -37,9 +37,9 @@ async def test_start_simulator(
         self, get_chia_simulator: Tuple[FullNodeSimulator, Path, Dict[str, Any], str, int]
```
jack60612, line 37: All these changes look good

jack60612 (changes requested):
all files in the simulator folder & test_start_simulator.py can not rely on packages other then the ones that are "exported" in setup.py

chia/simulator/simulator_test_tools.py
```diff
@@ -154,3 +165,5 @@ async def get_full_chia_simulator(
 
         await ws_server.stop()
         await shutdown_event.wait()  # wait till shutdown is complete
+        if temp_dir is not None:
+            temp_dir.cleanup()
```
altendky, line 169: Context manager please.  It's a touch more complicated since you only conditionally need this, but you have two basic choices.  Either conditionally choose the manager from either https://docs.python.org/3.10/library/contextlib.html#contextlib.nullcontext or the temporary directory if you need it and then always enter the chosen manager, or you can always enter https://docs.python.org/3.10/library/contextlib.html#contextlib.ExitStack and then only conditionally have it enter the temporary directory manager.

Or...  you skip this complexity and just make the caller give you a directory to work in.  It's not that hard for them to make a temp dir if that's where they want this function to operate.  It looks like we only call this one place and we pass a root there so the optionality is entirely, well, optional.

chia/simulator/simulator_test_tools.py
emlowe: Not sure I understand this - this code is located in `chia/simulator/simulator_test_tools.py` which you would get by cloning chia-blockchain, in which case you would have `tests.util.keyring`. Under what scenario would it not run?

chia/simulator/simulator_test_tools.py
emlowe: It seems to me that two lines of manual cleanup is preferable to more complicated code simply to use a context manager. but I'll consider further changes....

chia/simulator/simulator_test_tools.py
altendky: Any packaged result for example.  My personal projects include the tests in the main package (`chia._tests` for example) but we do not (and many Python projects don't).  Anyways, the last time this was an issue with the simulator the relevant code was moved out of the tests and into the main application.

chia/simulator/simulator_test_tools.py
altendky: There's `try:`/`finally:` as well, but the ugliness suggests to me that making the caller do it if needed is the clean solution.

chia/simulator/simulator_test_tools.py
emlowe: I modified the code to avoid this problem, I think the changes now are an improvement. - resolving the conversation

chia/simulator/simulator_test_tools.py
emlowe: Rearranged the code to make the caller specify it, which in our test code was simplified by using tmp_path fixture

tests/simulation/test_start_simulator.py
```diff
@@ -27,19 +28,19 @@ class TestStartSimulator:
...
-        self, automated_testing: bool = False, chia_root: Optional[Path] = None, config: Optional[Dict[str, Any]] = None
-    ) -> AsyncGenerator[Tuple[FullNodeSimulator, Path, Dict[str, Any], str, int], None]:
-        async for simulator_args in get_full_chia_simulator(automated_testing, chia_root, config):
+        self, tmp_path: Path, empty_keyring: Keychain
+    ) -> AsyncGenerator[Tuple[FullNodeSimulator, Path, Dict[str, Any], str, int, Keychain], None]:
+        async for simulator_args in get_full_chia_simulator(tmp_path, empty_keyring):
             yield simulator_args
 
     @pytest.mark.asyncio
     async def test_start_simulator(
```
jack60612, line 37: WE need to pass args to get_chia_simulator using pytest.

jack60612 (changes requested):

tests/simulation/test_start_simulator.py
emlowe: pytest wasn't passing any arguments at all previously.

altendky: I'm not sure what you are referring to, or suggesting, here Jack.  Could you say a bit more about it?

tests/simulation/test_start_simulator.py
```diff
@@ -27,19 +28,19 @@ class TestStartSimulator:
...
-        self, get_chia_simulator: Tuple[FullNodeSimulator, Path, Dict[str, Any], str, int]
+        self, get_chia_simulator: Tuple[FullNodeSimulator, Path, Dict[str, Any], str, int, Keychain]
     ) -> None:
-        simulator, root_path, config, mnemonic, fingerprint = get_chia_simulator
-        ph_1 = get_puzzle_hash_from_key(fingerprint, key_id=1)
-        ph_2 = get_puzzle_hash_from_key(fingerprint, key_id=2)
-        dummy_hash = std_hash(b"test")
+        simulator, root_path, config, mnemonic, fingerprint, keychain = get_chia_simulator
+        ph_1: bytes32 = get_puzzle_hash_from_key(keychain, fingerprint, key_id=1)
+        ph_2: bytes32 = get_puzzle_hash_from_key(keychain, fingerprint, key_id=2)
```
altendky, line 42:
```suggestion
        ph_1: bytes32 = get_puzzle_hash_from_key(keychain=keychain, fingerprint=fingerprint, key_id=1)
        ph_2: bytes32 = get_puzzle_hash_from_key(keychain=keychain, fingerprint=fingerprint, key_id=2)
```

chia/simulator/simulator_test_tools.py
```diff
@@ -34,20 +34,23 @@
...
         "today grape album ticket joy idle supreme sausage "
         "oppose voice angle roast you oven betray exact "
         "memory riot escape high dragon knock food blade"
     )
     # add key to keychain
-    sk = Keychain().add_private_key(mnemonic)
+    try:
+        sk = keychain.add_private_key(mnemonic)
+    except KeychainFingerprintExists:
+        pass
```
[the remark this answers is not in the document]
emlowe: This code tries to add a fixed key - if you say reuse a keychain (like the default one .chia_keys) this code will only work one time, the second time you try to use it you will see this error.  This wasn't intended by this code and when this code was written, you could add a key again without it throwing any errors, so this was introduced recently. The chia test code avoids this by using a temp keyring for the test.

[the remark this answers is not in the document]
altendky: Mm, I was thinking in terms of tests, but this is used otherwise now so sure, makes sense.  Thanks.

altendky (approved):
I'm good with this, but I do think we should chat with @jack60612 to make sure we aren't overlooking something.

jack60612 (approved):
lgtm, still a bit concerned about empty_keyring, but i'm happy for now.

paninaro (approved):

PR #13615, opened 2022-10-06
full_node: Introduce `sync_store.Peak`

xdustinface:


chia/full_node/sync_store.py
```diff
@@ -66,7 +61,7 @@ def peer_has_block(
         Adds a record that a certain peer has a block.
         """
 
-        if header_hash == self.sync_target_header_hash:
+        if self.target_peak is not None and header_hash == self.target_peak.header_hash:
```
aqk, line 64: What happens in the `self.target_peak == None` case?
Is that always the start of chain, or can that happen when we are starting up?

chia/full_node/sync_store.py
```diff
@@ -75,11 +70,11 @@ def peer_has_block(
             if len(self.peak_to_peer) > 256:  # nice power of two
                 item = self.peak_to_peer.popitem(last=False)  # Remove the oldest entry
                 # sync target hash is used throughout the sync process and should not be deleted.
-                if item[0] == self.sync_target_header_hash:
+                if self.target_peak is not None and item[0] == self.target_peak.header_hash:
                     self.peak_to_peer[item[0]] = item[1]  # Put it back in if it was the sync target
                     self.peak_to_peer.popitem(last=False)  # Remove the oldest entry again
         if new_peak:
-            self.peer_to_peak[peer_id] = (header_hash, height, weight)
+            self.peer_to_peak[peer_id] = Peak(header_hash, height, weight)
```
aqk, line 77: We could add an invariant here and throw if we expect the components of `Peak` (header_hash, height, weight) to be non-null. Or perhaps in the Peak constructor?

chia/full_node/full_node.py
```diff
@@ -716,21 +719,20 @@ async def new_peak(self, request: full_node_protocol.NewPeak, peer: ws.WSChiaCon
 
         if self.sync_store.get_sync_mode():
             # If peer connects while we are syncing, check if they have the block we are syncing towards
-            peak_sync_hash = self.sync_store.get_sync_target_hash()
-            peak_sync_height = self.sync_store.get_sync_target_height()
-            if peak_sync_hash is not None and request.header_hash != peak_sync_hash and peak_sync_height is not None:
-                peak_peers: Set[bytes32] = self.sync_store.get_peers_that_have_peak([peak_sync_hash])
+            target_peak = self.sync_store.target_peak
+            if target_peak is not None and request.header_hash != target_peak.header_hash:
```
aqk, line 723: It would be good to have a test or invariant assert that ensures that if either `height` or `header_hash` was None in the old code, that `target_peak` is always None is the new code.

chia/full_node/sync_store.py
xdustinface: Unless i miss something, it just does the same like before, it skips. Previously, in the case where now `target_peak` is `None` we checked `header_hash == None` because we had it defaulting to `None` too, see 
https://github.com/Chia-Network/chia-blockchain/blob/d18f205bd033b3997e19393e90ab4b0cf0c63802/chia/full_node/sync_store.py#L27-L28

which fails like `self.target_peak is not None` because `header_hash` is  `bytes32`?

chia/full_node/sync_store.py
xdustinface: Why do you think we need this here since they are supposed to be non-optional by their definition `header_hash: bytes32, peer_id: bytes32, weight: uint128, height: uint32`?

chia/full_node/full_node.py
xdustinface: There were either both `None` or none of them was `None` since they were initially `None` and only set via https://github.com/Chia-Network/chia-blockchain/blob/bf4759be503bf46c997b41e39a3c22a53316e404/chia/full_node/sync_store.py#L41-L43 which sets both. Now they are just packed into `Peak` so the behaviour is the same, isn't it? Either `target_peak` is non, which would be the same as both of them being `None` before or `target_peak` is not non and then both of them are also not `None` because they are supposed to be `bytes32` and `uint32` by their definition.

almogdepaz: having hash None but hight not None or the other way around doesn't make sense 
if i had to guess the checks in the old code are only for lint purposes.

i think the new code is better because it enforces that fact that if we know the peak than we must know its hash and its hight and the wight.

chia/full_node/full_node.py
```diff
@@ -291,7 +291,10 @@ def get_connections(self, request_node_type: Optional[NodeType]) -> List[Dict[st
             peak_store = None
         for con in connections:
             if peak_store is not None and con.peer_node_id in peak_store:
-                peak_hash, peak_height, peak_weight = peak_store[con.peer_node_id]
+                peak = peak_store[con.peer_node_id]
+                peak_height = peak.height
+                peak_hash = peak.header_hash
```
almogdepaz, line 296: im not sure where and how get_connections is used
but i think it would be nicer to just use Peak here and not separate to each field

xdustinface: It's done like this to retain the current behaviour because it's used by an RPC endpoint.

chia/full_node/sync_store.py
```diff
@@ -93,38 +88,34 @@ def get_peers_that_have_peak(self, header_hashes: List[bytes32]) -> Set[bytes32]
...
+    def get_peak_of_each_peer(self) -> Dict[bytes32, Peak]:
         """
         Returns: dictionary of peer id to peak information.
         """
 
         ret = {}
-        for peer_id, v in self.peer_to_peak.items():
-            if v[0] not in self.peak_to_peer:
+        for peer_id, peak in self.peer_to_peak.items():
+            if peak.header_hash not in self.peak_to_peer:
```
almogdepaz, line 98: i know this is not related to the pr but this discrepancy between the two Dictionaries is wrong
i guess we ignore ones not in peak_to_peer because there is no way of finding them in peer_to_peak
but maybe we should add a warning ?
i dont think this should happen unless something went wrong

chia/full_node/sync_store.py
```diff
@@ -93,38 +88,34 @@ def get_peers_that_have_peak(self, header_hashes: List[bytes32]) -> Set[bytes32]
...
         if len(self.peer_to_peak) == 0:
             return None
-        heaviest_peak_hash: Optional[bytes32] = None
-        heaviest_peak_weight: uint128 = uint128(0)
-        heaviest_peak_height: Optional[uint32] = None
-        for peer_id, (peak_hash, height, weight) in self.peer_to_peak.items():
-            if peak_hash not in self.peak_to_peer:
+        heaviest_peak: Optional[Peak] = None
+        for peak in self.peer_to_peak.values():
+            if peak.header_hash not in self.peak_to_peer:
```
almogdepaz, line 113: same as above, i feel we are ignoring something here

chia/full_node/full_node.py
almogdepaz: yes i know just saying would be nicer to use your new class, not insisting on this as im not sure how much code depends on this

xdustinface: Agree and i thought about it too but i think we shouldn't just break compatibility. At the end it could have been done with a `peak` object before this PR too without the `Peak` class.

chia/full_node/sync_store.py
xdustinface: Yeah thats unrelated to here, i will put it on my list, might be worth investigating later.

almogdepaz (dismissed):

chia/full_node/full_node.py
aqk: Yes, agreed. I noticed the same thing. This way covers more cases of possible future misuse. :)

chia/full_node/sync_store.py
aqk: We don't need it - it was just a suggestion. Since we enforce running mypy in the calling code, this is good enough.

aqk (approved):
A nice improvement.

emlowe (approved):

PR #13619, opened 2022-10-06
full_node: Drop unused ultra priority lock

xdustinface:


mariano54 (approved):
There is no reason to have it if we're not using it.

PR #13629, opened 2022-10-11
process_transaction_records() -> process_transactions()

altendky:
Adding flexibility to wait for spend bundles in addition to transaction records.

Break out from https://github.com/Chia-Network/chia-blockchain/pull/13603.

### DIFF
--- a/chia/simulator/full_node_simulator.py
+++ b/chia/simulator/full_node_simulator.py
@@ -409,10 +409,7 @@ class FullNodeSimulator(FullNodeAPI):
 
             await asyncio.sleep(0.050)
 
-    async def process_transaction_records(
-        self,
-        records: Collection[TransactionRecord] = (),
-    ) -> None:
+    async def process_transaction_records(self, records: Collection[TransactionRecord] = ()) -> None:
         """Process the specified transaction records and wait until they have been
         included in a block.
 
@@ -431,10 +428,7 @@ class FullNodeSimulator(FullNodeAPI):
 
         return await self.process_coin_spends(coins=coins_to_wait_for)
 
-    async def process_spend_bundles(
-        self,
-        bundles: Collection[SpendBundle] = (),
-    ) -> None:
+    async def process_spend_bundles(self, bundles: Collection[SpendBundle] = ()) -> None:
         """Process the specified spend bundles and wait until they have been included
         in a block.
 
@@ -445,10 +439,7 @@ class FullNodeSimulator(FullNodeAPI):
         coins_to_wait_for: Set[Coin] = {addition for bundle in bundles for addition in bundle.additions()}
         return await self.process_coin_spends(coins=coins_to_wait_for)
 
-    async def process_coin_spends(
-        self,
-        coins: Collection[Coin] = (),
-    ) -> None:
+    async def process_coin_spends(self, coins: Collection[Coin] = ()) -> None:
         """Process the specified coin names and wait until they have been created in a
         block.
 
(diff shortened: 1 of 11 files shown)


chia/simulator/full_node_simulator.py
```diff
@@ -408,14 +409,19 @@ async def wait_transaction_records_entered_mempool(self, records: Collection[Tra
...
+    ) -> None:
         """Process the specified transaction records and wait until they have been
         included in a block.
 
         Arguments:
             records: The transaction records to process.
+            bundles: The spend bundles to process.
         """
-        coins_to_wait_for: Set[Coin] = set()
+        coins_to_wait_for: Set[Coin] = {addition for bundle in bundles for addition in bundle.additions()}
```
almogdepaz, line 424: can you explain what this does ?
this syntax is confusing i would avoid it

tests/simulation/test_simulation.py
```diff
@@ -353,9 +355,15 @@ async def test_process_transaction_records(
...
                 coins={coin},
             )
+            assert tx.spend_bundle is not None, "the above created transaction is missing the expected spend bundle"
             await wallet.push_transaction(tx)
 
-            await full_node_api.process_transaction_records(records=[tx])
+            if record_or_bundle == "record":
+                await full_node_api.process_transactions(records=[tx])
+            elif record_or_bundle == "bundle":
+                await full_node_api.process_transactions(bundles=[tx.spend_bundle])
```
almogdepaz, line 364: we should test with multiple spend bundles that have multiple additions to exercise the code in process_transactions

almogdepaz:
im not entirely sure why this is needed can you explain ?
i think separating handling of coins_to_wait_for into an "inner function" and calling it from 2 separate cases 
one for records and one for spend bundles would be easier to read

chia/simulator/full_node_simulator.py
altendky: It reads better on multiple lines but yeah, there's a bit to be aware of.

```
x = {
    addition
    for bundle in bundles
    for addition in bundle.additions()
}
```
Is equivalent, at least in result, to this.
```
x = set()
for bundle in bundles:
    for addition in bundle.additions():
        x.add(addition)
```

Having the value (`addition`) first often trips people up in terms of ordering.  The order, or nesting, of the for loops is the same in both cases.  In this case this is being done to flatten all the 'additions' from the various spend bundles out into a single flat set.

altendky:
This is a breakout from https://github.com/Chia-Network/chia-blockchain/pull/13603 where I noticed that some places had transaction records to wait for and others had spend bundles to wait for.

I had considered two separate functions but didn't come up with a strong preference either way.  The one thing that a single function does is to allow both types of things to be waited as a single group.  That's not likely to be a critical thing, at least not now.

I'll implement this as two separate functions and see how it looks.

chia/simulator/full_node_simulator.py
altendky: Or, Python's description.  https://docs.python.org/3.10/tutorial/datastructures.html#list-comprehensions

almogdepaz: ok yes the multiline is much clearer,  the first value was the thing that confused me.
personally i would avoid or maybe force the multiline for readability but it might just be me.

almogdepaz (approved):

emlowe (approved):

chia/simulator/full_node_simulator.py
altendky: I agree.  I used to say a lot that all my one liner's were four or five lines.  I don't think we can make black do that though.

PR #13638, opened 2022-10-15
treat Semaphore._waiters as length zero when it is None

altendky:
We access the `._waiters` private attribute of the Python asyncio `Semaphore` class.  This was changed in Python 3.10.8 (and other versions) to be initialized to `None` instead of an empty deque.  Our existing unconditional length checks failed on the new `None` default.  This seems to block syncing.

https://github.com/python/cpython/pull/97020

https://github.com/python/cpython/compare/v3.10.7..v3.10.8#diff-0fee1befb15023abc0dad2623effa93a304946796929f6cb445d11a57821e737

Reported traceback:
```python-traceback
2022-10-12T20:03:59.367 full_node full_node_server        : INFO     Connected with full_node {'host': '65.34.144.6', 'port': 8444}
2022-10-12T20:03:59.370 full_node full_node_server        : ERROR    Exception: object of type 'NoneType' has no len(), {'host': '65.34.144.6', 'port': 8444}. Traceback (most recent call last):
  File "/home/summa/chia-blockchain/chia/server/server.py", line 598, in wrapped_coroutine
    result = await coroutine
  File "/home/summa/chia-blockchain/chia/full_node/full_node_api.py", line 114, in new_peak
    waiter_count = len(self.full_node.new_peak_sem._waiters)
TypeError: object of type 'NoneType' has no len()

2022-10-12T20:03:59.371 full_node full_node_server        : ERROR    Exception: object of type 'NoneType' has no len() <class 'TypeError'>, closing connection {'host': '65.34.144.6', 'port': 8444}. Traceback (most recent call last):
  File "/home/summa/chia-blockchain/chia/server/server.py", line 608, in api_call
    response: Optional[Message] = await asyncio.wait_for(wrapped_coroutine(), timeout=timeout)
  File "/usr/lib/python3.10/asyncio/tasks.py", line 408, in wait_for
    return await fut
  File "/home/summa/chia-blockchain/chia/server/server.py", line 605, in wrapped_coroutine
    raise e
  File "/home/summa/chia-blockchain/chia/server/server.py", line 598, in wrapped_coroutine
    result = await coroutine
  File "/home/summa/chia-blockchain/chia/full_node/full_node_api.py", line 114, in new_peak
    waiter_count = len(self.full_node.new_peak_sem._waiters)
TypeError: object of type 'NoneType' has no len()

[...]

PR #13639, opened 2022-10-16
use a semaphore to limit waiters for NewPeak and NewCompactVDF

altendky:
Solves the same issue as https://github.com/Chia-Network/chia-blockchain/pull/13638 but in a more intrusive and better way.  Well, better assuming I got this right at least.  It does lose the count of the waiters for debug logging.  Perhaps a more custom solution to wrap up the counting or such would be better even though this is only used in two places.

Draft for:
- [ ] probably replaced by https://github.com/Chia-Network/chia-blockchain/pull/13642

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -125,7 +125,9 @@ class FullNode:
     _sync_task: Optional[asyncio.Task[None]]
     _transaction_queue: Optional[asyncio.PriorityQueue[Tuple[int, TransactionQueueEntry]]]
     _compact_vdf_sem: Optional[asyncio.Semaphore]
+    _compact_vdf_waiting_sem: Optional[asyncio.Semaphore]
     _new_peak_sem: Optional[asyncio.Semaphore]
+    _new_peak_waiting_sem: Optional[asyncio.Semaphore]
     _respond_transaction_semaphore: Optional[asyncio.Semaphore]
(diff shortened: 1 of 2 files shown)


chia/full_node/full_node.py
```diff
@@ -125,7 +125,9 @@ class FullNode:
     _sync_task: Optional[asyncio.Task[None]]
     _transaction_queue: Optional[asyncio.PriorityQueue[Tuple[int, TransactionQueueEntry]]]
     _compact_vdf_sem: Optional[asyncio.Semaphore]
+    _compact_vdf_waiting_sem: Optional[asyncio.Semaphore]
     _new_peak_sem: Optional[asyncio.Semaphore]
+    _new_peak_waiting_sem: Optional[asyncio.Semaphore]
```
arvidn, line 130: it seems to me that you could just use plain `int` here, and call them `_new_peak_waiting_counter` and `_compact_vdf_waiting_counter`. I suppose with a `Semaphore` you get the context manager to increment and decrement the counter. But you don't need any of the heavy machinery to manage a queue of waiting tasks and to resume them.

i.e. you don't need the mutex-feature, just the counter-feature.

altendky: https://github.com/Chia-Network/chia-blockchain/pull/13642

PR #13642, opened 2022-10-17
Add LimitedSemaphore

altendky:


almogdepaz:
this looks good, why still a draft ?

chia/util/limited_semaphore.py
```diff
@@ -0,0 +1,39 @@
...
+            _available_count=active_limit + waiting_limit,
+        )
+
+    @contextlib.asynccontextmanager
+    async def acquire(self) -> AsyncIterator[int]:
+        if self._available_count < 1:
+            raise LimitedSemaphoreFullError()
+
+        self._available_count -= 1
+        try:
```
almogdepaz, line 35: can this throw anything ?

altendky: Being a context manager, the body (which runs at the `yield`) can raise anything.  Using `finally` will let any exceptions propagate further but makes sure we _always_ decrement.  I'm trusting that the semaphore itself will never acquire but not release.  Seems that would be a bug in Python that we don't need to account for.

almogdepaz (approved):

arvidn (approved):

PR #13650, opened 2022-10-18
debugging extra tx block needed

altendky:


chia/simulator/full_node_simulator.py
```diff
@@ -373,7 +373,7 @@ async def farm_blocks_to_wallet(
             expected_peak_height = 0 if original_peak_height is None else original_peak_height
 
             # TODO: why two final transaction blocks and not just one?
-            for to_wallet in [*([True] * count), False, False]:
+            for to_wallet in [*([True] * count), False]:
```
altendky, line 376: Previously commented on at https://github.com/Chia-Network/chia-blockchain/pull/11535#discussion_r896092474.  Even with the extra checks I'm unclear why the reward transactions are missing.

PR #13658, opened 2022-10-19
Fixing s3 upload conditional for windows builds

wallentx:
The s3 upload step had an extra conditional that was preventing it from uploading to dev builds folder.


cmmarslender (approved):

wjblanke (approved):
aok

Issue #13282, opened 2022-09-01
Optional[Credential] = keyring.get_credential(keychain_service, current_user) -> AttributeError: 'NoneType' object has no attribute 'get_credential'


seanvetter:
OS: Freebsd 13.1(TrueNAS), using a jail

I currently have 1.5.0 working in this jail and wanted to update to the latest, so I created a new directory and cloned the latest release and did the following build steps. When I try to start the services and enter my passphrase I get this traceback. In the previous 1.5.0 install the passphrase seems to work correctly and I do not get this traceback. I'm not sure if I'm missing a step because it's freebsd or if this is a bug, kinda seems like a bug to me?

Build steps:
1. git clone https://github.com/Chia-Network/chia-blockchain.git -b latest
2. cd chia-blockchain
3. python3 -m venv venv
4. source venv/bin/activate
5. pip install --upgrade pip
6. pip install clvm-tools-rs maturin (also tried clvm-tools-rs==0.1.19)
7. bash ./install.sh
8. chia init

**CMD: chia start farmer**
```
Traceback (most recent call last):
  File "/root/chia-1.5.1v2/chia-blockchain/venv/bin/chia", line 8, in <module>
    sys.exit(main())
  File "/root/chia-1.5.1v2/chia-blockchain/chia/cmds/chia.py", line 147, in main
    cli()  # pylint: disable=no-value-for-parameter
  File "/root/chia-1.5.1v2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/root/chia-1.5.1v2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/root/chia-1.5.1v2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/root/chia-1.5.1v2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/root/chia-1.5.1v2/chia-blockchain/venv/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
[...]

PR #13398, opened 2022-09-09
util: Remove legacy keyring support

xdustinface:

PR #13666, opened 2022-10-19
Freebsd improvements when checking for legacy key

seanvetter:
Issue: #13282 Optional[Credential] = keyring.get_credential(keychain_service, current_user) -> AttributeError: 'NoneType' object has no attribute 'get_credential'

Freebsd platform fails to migrate in previous releases or in 1.5.1+ versions it fails to start farmer because there is a NoneType object from keyring.get_credential. This just checks for that when trying to set the legacy_keyring.

### DIFF
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -112,7 +112,7 @@ class KeyringWrapper:
         self.keys_root_path = keys_root_path
         if force_legacy:
             legacy_keyring = get_legacy_keyring_instance()
-            if check_legacy_keyring_keys_present(legacy_keyring):
+            if legacy_keyring is not None and check_legacy_keyring_keys_present(legacy_keyring):
                 self.legacy_keyring = legacy_keyring
         else:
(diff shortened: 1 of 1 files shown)


emlowe (approved):

xdustinface (changes requested):
So this just silently ignores the legacy keyring on systems which aren't supported it seems (hm, actually that shouldn't be an issue since they are not supported so they won't have a legacy keyring around anyway i guess), lets maybe just merge #13398 instead now that we have the 1.6.1 release branch? @paninaro thoughts?

xdustinface (commented):
Actually, let's maybe retarget this to the 1.6.1 release branch?

emlowe:
1.6.1 is pretty set now. I see, if we merge in 13398 to main, this small fix becomes moot. I might be inclined to take it anyway in main.

xdustinface (approved):

xdustinface:
I mean yeah, the related code will be gone then, but sure, we can merge this if you wish.

PR #13668, opened 2022-10-19
stop using deprecated set-output

altendky:

1. `"::set-output name=(.*)::(.*)"` -> `"$1=$2" >> \$GITHUB_OUTPUT`
2. `::set-output name=(.*)::(.*)` -> `"$1=$2" >> \$GITHUB_OUTPUT`


altendky:
fyi @wallentx and @cmmarslender

altendky:
Apparently I totally didn't understand something...

zcomputerwiz (commented):
I had opened a pull request ( #13730 ) to update this as well - and I encountered the same problems with the Windows build and test workflows at first but did get them all working.

One solution is to use the Windows PowerShell syntax instead, the other would be to specify shell: Bash in the get pip and npm cache directory steps. I have not tried the second since the first seems to work just fine.

altendky:
Yeah, your PR was pointed out to me this morning.  Thanks for the contribution.  I have run projects with bash on all systems for the workflows but I think we're not doing that here, and I indeed fail to handle the Windows case properly.  We'll look through these PRs when we get a chance and get something in.

PR #13688, opened 2022-10-21
extend wallet.nft_wallet timeout to 45 minutes

altendky:


altendky:
Welp, I like the 45 minutes given the 30 minute Windows runtime...  and it's already merged in https://github.com/Chia-Network/chia-blockchain/pull/13675 so done and done.

PR #13694, opened 2022-10-24
Don't hold removals twice in mempool items as they're already part of the underlying spend bundle

AmineKhaldi:


arvidn (approved):
good catch!

PR #13704, opened 2022-10-25
Made error message of `chia plotters version` more traceable

ChiaMineJP:


wallentx (commented):
@ChiaMineJP Should this PR also handle passthrough of error messages when the user is missing a system library?
For example:
```
root18:07bladebit$ ./bladebit
./bladebit:` error while loading shared libraries: libnuma.so.1: cannot open shared object file: No such file or directory
root18:07bladebit$ echo $?
127
```

wallentx (dismissed):

paninaro (dismissed):
lgtm

paninaro (approved):
lgtm

PR #13716, opened 2022-10-26
add target_coin amount to coin selection cli (chia wallet coins) 

jack60612:


jack60612:
tested on testnet10

chia/cmds/coin_funcs.py
```diff
@@ -127,27 +128,37 @@ async def async_combine(args: Dict[str, Any], wallet_client: WalletRpcClient, fi
...
-    if largest:
-        conf_coins.sort(key=lambda r: r.coin.amount, reverse=True)
+    final_target_coin_amount = uint64(int(target_coin_amount * mojo_per_unit))
+    if final_target_coin_amount != 0:  # if we have a set target, just use standard coin selection.
+        removals: List[Coin] = await wallet_client.select_coins(
+            amount=final_target_coin_amount + final_fee,
+            wallet_id=wallet_id,
+            max_coin_amount=final_max_dust_amount,
+            min_coin_amount=final_min_coin_amount,
+            excluded_amounts=final_excluded_amounts + [final_target_coin_amount],  # dont reuse coins of same amount.
```
arvidn, line 138: this comment is cryptic. I don't understand what it's trying to say

chia/cmds/coin_funcs.py
```diff
@@ -109,6 +109,7 @@ async def async_combine(args: Dict[str, Any], wallet_client: WalletRpcClient, fi
     excluded_amounts = args["excluded_amounts"]
     number_of_coins = args["number_of_coins"]
     max_dust_amount = Decimal(args["max_dust_amount"])
+    target_coin_amount = Decimal(args["target_coin_amount"])
```
arvidn, line 112: is `target_coin_amount` not optional? if it's `None`, passing it to `Decimal()` would fail, right?

chia/cmds/coin_funcs.py
```diff
@@ -127,27 +128,37 @@ async def async_combine(args: Dict[str, Any], wallet_client: WalletRpcClient, fi
...
-    )
-    if len(target_coin_ids) > 0:
-        conf_coins = [cr for cr in conf_coins if cr.name in target_coin_ids]
-    if len(conf_coins) <= 1:
-        print("No coins to combine.")
-        return
-    if largest:
-        conf_coins.sort(key=lambda r: r.coin.amount, reverse=True)
+    final_target_coin_amount = uint64(int(target_coin_amount * mojo_per_unit))
+    if final_target_coin_amount != 0:  # if we have a set target, just use standard coin selection.
```
arvidn, line 132: I would have expected we use `None` here instead of 0, as in the value not being present. Or is 0 significant?

arvidn (commented):
it makes me nervous that we don't have any tests for this logic. it *is* important that it works, right?

chia/cmds/coin_funcs.py
jack60612: the default value is zero, that's what we normally use for cli.

chia/cmds/coin_funcs.py
jack60612: 0 is the default value

chia/cmds/coin_funcs.py
jack60612: if we have a coin of amount a but we want to make another coin of amount a, the default coin selection algorithm will just use that existing coin of amount a, instead of choosing a set of new coins to create amount a. So we tell the algorithm to filter out all coins of amount a so that the coins outputted from the algorithm can result in amount a but don't only contain a coin of amount a.

jack60612:
> it makes me nervous that we don't have any tests for this logic. it _is_ important that it works, right?
I don't think we have cli tests & the logic is almost all from the existing coin selection algo.

chia/cmds/coins.py
```diff
@@ -97,6 +97,13 @@ def list_cmd(
 )
 @click.option("-f", "--fingerprint", help="Set the fingerprint to specify which wallet to use", type=int)
 @click.option("-i", "--id", help="Id of the wallet to use", type=int, default=1, show_default=True, required=True)
+@click.option(
+    "-a",
+    "--target-coin-amount",
+    help="An optional target coin amount, if you want to create a coin with a specific amount.",
+    type=str,
+    default="0",
```
jack60612, line 105: default set here

arvidn (approved):

PR #13720, opened 2022-10-27
Add bulk set NFT DID API

ytx1991:


chia/rpc/wallet_rpc_api.py
```diff
@@ -1727,6 +1729,30 @@ async def nft_set_nft_did(self, request):
         spend_bundle = await nft_wallet.set_nft_did(nft_coin_info, did_id, fee=fee)
         return {"wallet_id": wallet_id, "success": True, "spend_bundle": spend_bundle}
 
+    async def nft_set_bulk_nft_did(self, request):
+        if len(request["nft_coin_list"]) > MAX_NFT_CHUNK_SIZE:
+            return {"success": False, "error": f"You can only set {MAX_NFT_CHUNK_SIZE} NFTs at once"}
+        wallet_id = uint32(request["wallet_id"])
+        nft_wallet = self.service.wallet_state_manager.wallets[wallet_id]
+        assert isinstance(nft_wallet, NFTWallet)
+        did_id = request.get("did_id", "")
```
trepca, line 1738: you can set the default value to `b""` here so you don't have to set it in the if

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -981,6 +981,62 @@ async def make_nft1_offer(
...
+
+        if len(spend_bundles) > 0:
+            spend_bundle = SpendBundle.aggregate(spend_bundles)
+            # Add all spend bundles to the first tx
+            refined_tx_list[0] = dataclasses.replace(refined_tx_list[0], spend_bundle=spend_bundle)
+            for tx in refined_tx_list:
+                await self.wallet_state_manager.add_pending_transaction(tx)
+            for coin in coin_ids:
+                await self.update_coin_status(coin, True)
+            self.wallet_state_manager.state_changed("nft_coin_did_set", self.wallet_info.id)
```
trepca, line 1035: should we wait for spend bundle to be processed before setting the new wallet id?

chia/wallet/wallet_state_manager.py
```diff
@@ -921,7 +921,7 @@ async def handle_nft(
                         self.log.info(f"No NFT, deleting wallet {nft_wallet.did_id.hex()} ...")
                         await self.user_store.delete_wallet(nft_wallet.wallet_info.id)
                         self.wallets.pop(nft_wallet.wallet_info.id)
-            if nft_wallet_info.did_id == new_did_id:
+            if nft_wallet_info.did_id == new_did_id and new_derivation_record is not None:
```
trepca, line 924: what's the reasoning behind all these derivation record checks?

tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -1169,6 +1169,126 @@ async def test_update_metadata_for_nft_did(two_wallet_nodes: Any, trusted: Any)
...
+    coins_response = await wait_rpc_state_condition(
+        30, api_0.nft_get_nfts, [{"wallet_id": nft_wallet_0_id}], lambda x: len(x["nft_list"]) > 1
+    )
+    coins = coins_response["nft_list"]
+    assert len(coins) == 2
+    assert coins[0].owner_did is None
+    nft_coin_list = [coins[0].nft_coin_id.hex(), coins[1].nft_coin_id.hex()]
+    resp = await api_0.nft_set_bulk_nft_did(
+        dict(wallet_id=nft_wallet_0_id, did_id=hmr_did_id, nft_coin_list=nft_coin_list)
+    )
```
trepca, line 1271: can you check that nfts in `nft_coin_list` have a pending state set. Once they're in pending state, should be block any further attempts to bulk set them?

chia/wallet/nft_wallet/nft_wallet.py
ytx1991: new wallet id? It's for notifying the GUI there are some NFTs of this wallet are in transaction.

chia/wallet/wallet_state_manager.py
ytx1991: This is for handling the unexpected wallet creation/deletion issue.

tests/wallet/nft_wallet/test_nft_wallet.py
ytx1991: It doesn't have lockers on the backend side. The GUI will prohibit further operations.

trepca: but if you call `get_nft_info` it should return back if nft is pending or not right?

chia/rpc/wallet_rpc_api.py
```diff
@@ -1727,6 +1727,28 @@ async def nft_set_nft_did(self, request):
         spend_bundle = await nft_wallet.set_nft_did(nft_coin_info, did_id, fee=fee)
         return {"wallet_id": wallet_id, "success": True, "spend_bundle": spend_bundle}
 
+    async def nft_set_bulk_nft_did(self, request):
+        if len(request["nft_coin_list"]) > MAX_NFT_CHUNK_SIZE:
+            return {"success": False, "error": f"You can only set {MAX_NFT_CHUNK_SIZE} NFTs at once"}
+        wallet_id = uint32(request["wallet_id"])
```
paninaro, line 1733: Would it be possible to assign NFTs from different wallets to a DID? The UI has an aggregated view that shows all NFTs, so I can imagine that users might try to select NFTs across multiple NFT wallets to perform a bulk DID assignment.

tests/wallet/nft_wallet/test_nft_wallet.py
ytx1991: It is checked.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1727,6 +1727,66 @@ async def nft_set_nft_did(self, request):
         spend_bundle = await nft_wallet.set_nft_did(nft_coin_info, did_id, fee=fee)
         return {"wallet_id": wallet_id, "success": True, "spend_bundle": spend_bundle}
 
+    async def nft_set_bulk_nft_did(self, request):
```
paninaro, line 1730: I won't hold up the review, but my only comments are related to documenting the params and validating those params earlier with a useful error message to the caller. If, for example, the DID isn't set, or `nft_coin_list` is missing, or some nested param is missing, the caller isn't likely to get a very useful error message.

trepca:
@ytx1991 thanks, looks good, but agree with @paninaro, lets add docstring to RPC method to document params required and check for missing required params and return a sensible error message to the user

trepca (approved):

PR #13727, opened 2022-10-27
Call close callback for already closed connections

emlowe:
Explorations into calling the close callback in all cases, even if the connection has already been closed.
This may help with some issues around the timing of closed connections resulting in "zombie" connections in the list which can't be removed.
This also makes sure we run the peer through the banning logic.


bramv-chia (dismissed):

emlowe:
This code seems to cause substantial flakes in the wallet sync tests - needs some investigation, so moving back to draft

chia/server/server.py
```diff
@@ -258,7 +260,10 @@ async def garbage_collect_connections_task(self) -> None:
                             to_remove.append(connection)
             for connection in to_remove:
                 self.log.debug(f"Garbage collecting connection {connection.peer_host} due to inactivity")
-                await connection.close()
+                if connection.closed:
+                    self.all_connections.pop(connection.peer_node_id)
+                else:
+                    await connection.close()
```
altendky, line 266: I'm guessing there's a reason we shouldn't pop it when we close it?

```suggestion
                if not connection.closed:
                    await connection.close()
                self.all_connections.pop(connection.peer_node_id)
```

chia/server/server.py
```diff
@@ -377,6 +382,10 @@ async def incoming_connection(self, request):
     async def connection_added(self, connection: WSChiaConnection, on_connect=None):
         # If we already had a connection to this peer_id, close the old one. This is secure because peer_ids are based
         # on TLS public keys
+        if connection.closed:
+            self.log.debug(f"Curious: ignoring request to add closed connection {connection.peer_host} ")
```
altendky, line 386: Maybe?  Or am I being too business serious here...

```suggestion
            self.log.debug(f"ignoring unexpected request to add closed connection {connection.peer_host} ")
```

chia/server/ws_connection.py
```diff
@@ -202,13 +202,22 @@ async def perform_handshake(
         self.outbound_task = asyncio.create_task(self.outbound_handler())
         self.inbound_task = asyncio.create_task(self.inbound_handler())
 
+    def do_close_callback(self, ban_time: int, closed_connection: bool = False) -> None:
+        try:
+            self.close_callback(self, ban_time, closed_connection)
+        except Exception:
+            error_stack = traceback.format_exc()
+            self.log.error(f"Error calling close callback: {error_stack}")
+
```
altendky, line 211: I think this layer may be able to be eliminated by using the `chia.util.log_exceptions.log_exceptions()` context manager instead.  Almost nobody knows about this and I keep forgetting about it too.  I should go on a quest and apply it as much as possible.  Maybe.

https://github.com/Chia-Network/chia-blockchain/blob/0befdec071f49708e26c7638656874492c52600a/chia/util/log_exceptions.py

Anyways, with that out of the way I think we could drop the `closed_connection` parameter and just have a ban function and a close function.

tests/connection_utils.py
```diff
@@ -29,9 +29,12 @@ async def disconnect_all(server: ChiaServer) -> None:
     for con in cons:
         await con.close()
 
+    await asyncio.sleep(5)  # 5 seconds to allow connections and tasks to all drain
+
```
altendky, line 33: Can we monitor for the desired state instead of just sleeping and hoping?

tests/connection_utils.py
```diff
@@ -29,9 +29,12 @@ async def disconnect_all(server: ChiaServer) -> None:
...
 
 async def disconnect_all_and_reconnect(server: ChiaServer, reconnect_to: ChiaServer, self_hostname: str) -> bool:
     await disconnect_all(server)
+
```
altendky, line 37: Diff noise be-gone!

```suggestion
```

altendky (changes requested):

tests/connection_utils.py
emlowe: I'd like to but struggled to find a way.

chia/server/server.py
emlowe: They get popped when the close() call comes back around and calls the close_callback - so I didn't change how this part of the code worked.

chia/server/ws_connection.py
emlowe: Can you pseudo code this out how it might look? Not sure I follow

emlowe: I think my problem is the "close_callback" is implemented in server.py, which manages the connection list and the banned list - so this would require 2 callbacks into server.py?

emlowe: I removed this `do_close_callback` helper function and use log_exceptions context manager

tests/connection_utils.py
altendky: It seems that `WSChiaConnection.close()` is not great.  It cancels tasks without actually waiting for them to finish.  Add another PR to the list...  if you want to maybe save some time you could gather all the `.close()` calls concurrently.  Also not sure why we're slicing the fresh list either.

```python
    async def disconnect_all(server: ChiaServer) -> None:
        connections = list(server.all_connections.values())
        await asyncio.gather(*(connection.close() for connection in connections))
        await asyncio.sleep(5)  # 5 seconds to allow connections and tasks to all drain
```

I also made a note over in https://github.com/Chia-Network/chia-blockchain/pull/13778 to make this timeout adjusted.

altendky (changes requested):
Since we're changing the callback, let's at least hint it.  here's a patch to consider.

```patch
diff --git a/chia/server/ws_connection.py b/chia/server/ws_connection.py
index 41224cc62..38fae2769 100644
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -8,6 +8,7 @@ import traceback
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
 from aiohttp import WSCloseCode, WSMessage, WSMsgType
+from typing_extensions import Protocol
 
 from chia.cmds.init_funcs import chia_full_version_str
 from chia.protocols.protocol_message_types import ProtocolMessageTypes
@@ -29,6 +30,16 @@ from chia.util.network import class_for_type, is_localhost
 LENGTH_BYTES: int = 4
 
 
+class ConnectionClosedCallbackProtocol(Protocol):
+    def __call__(
+        self,
+        connection: WSChiaConnection,
+        ban_time: int,
+        closed_connection: bool = ...,
+    ) -> None:
+        pass
+
+
 class WSChiaConnection:
     """
     Represents a connection to another node. Local host and port are ours, while peer host and
@@ -46,7 +57,7 @@ class WSChiaConnection:
         is_feeler: bool,  # Special type of connection, that disconnects after the handshake.
         peer_host,
         incoming_queue,
-        close_callback: Callable,
+        close_callback: ConnectionClosedCallbackProtocol,
         peer_id,
         inbound_rate_limit_percent: int,
         outbound_rate_limit_percent: int,
```

It turns up at least a couple failures.

```
tests/connection_utils.py:68: error: Argument 9 to "WSChiaConnection" has incompatible type "Callable[[Any, Any], Any]"; expected "ConnectionClosedCallbackProtocol"  [arg-type]
tests/core/ssl/test_ssl.py:32: error: Argument 9 to "WSChiaConnection" has incompatible type "Callable[[Any, Any], Any]"; expected "ConnectionClosedCallbackProtocol"  [arg-type]
```

I'll note that those are already being touched up in https://github.com/Chia-Network/chia-blockchain/pull/13878 so expect some conflicts if you don't wait.  The hint for the callback parameter is as well.  Anyways, I just approved that so you can check it out and help get it merged

tests/connection_utils.py
emlowe: I think it's a little more complicated since WSChiaConnection doesn't know about all the tasks in use - I don't think we should be calling the close method directly on the connections, this should be done at the service/client level since there we can wait on all the tasks.

emlowe:
Implemented code suggestions

tests/connection_utils.py
altendky: Yeah, there probably is more.  That's what happens with _un_structured concurrency.  You can't really tell if something is done or not.

altendky (approved):
thanks

paninaro (approved):

PR #13730, opened 2022-10-27
Update workflows to use GITHUB_OUTPUT

zcomputerwiz:
Use GITHUB_OUTPUT in workflows instead of set-output.

GitHub is deprecating set-output as per their blog:

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -41,7 +41,7 @@ jobs:
           fetch-depth: 0
 
       - name: Setup Python environment
-        uses: actions/setup-python@v4
+        uses: Chia-Network/actions/setup-python@main
         with:
           python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -47,7 +47,7 @@ jobs:
 
       - name: Get pip cache dir (Windows)
         if: matrix.os.matrix == 'Windows'
-        id: pip-cache-win
+        id: pip-cache
         run: |
           echo "dir=$(pip cache dir)" | Out-File -FilePath $Env:GITHUB_OUTPUT -Encoding utf8 -Append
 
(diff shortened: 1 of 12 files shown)


zcomputerwiz:
I added checks for OS in the Tests and Benchmarks workflows to use PowerShell syntax for Windows or Bash for others.
An alternative could be to specify shell: Bash instead though I have not tested that.

I also added workflow_dispatch to the MacOS and Linux ARM64 build workflows for the convenience of on-demand runs and consistency with the other build workflows.

altendky (changes requested):
1) I created conflicts for you (sorry...) by changing the setup Python actions to all use our wrapper where you had updated at least some of them.
2) The benchmarks workflow is Linux only so no need for compatibility complexity there.  If we need that I would prefer the change to be done in a separate PR.
3) I think I would rather see the `shell: bash` option so we can reduce the duplication around the caching.  I may at some point push for using bash everywhere since the point of our CI setup isn't to test if we can write PowerShell.  `:]`

Thanks again for the work here and for identifying the failures in my similar PR.  I'm very much looking forward to eliminating all those annotations on the build.

altendky:
I'm also happy to just make the changes if you prefer that, I don't need to dump tedious work on you.  I just try not to take over other people's contributions without they're approval.

zcomputerwiz:
No worries, I'll make the changes!

altendky (approved):
Thanks for helping with the cleanup.  With this I think we're getting close to dealing with all the annotations.  Well, other than the closed event loop stuff...

altendky:
Sorry for the wait here, I'm going to try to get this the rest of the way through the process.

emlowe (approved):

cmmarslender (approved):

wallentx:
@zcomputerwiz it looks like your commits in this PR are unsigned, which is blocking us from merging due to branch protection rules that we have in place. To save you the trouble of attempting to clean all these up for resubmission, I've created a separate PR, where you will see that you are fully accredited for authorship https://github.com/Chia-Network/chia-blockchain/pull/13911
Please feel free to review it for accuracy.

I'll work on getting that one merged in, and will close this PR for now.

PR #13732, opened 2022-10-28
Add verify signature API

ytx1991:


trepca:
the challenging part here is to figure out the public key especially for coins. It might make sense that this API does the work for them and only coin is provided. @paninaro what do you think?

ytx1991:
> the challenging part here is to figure out the public key especially for coins. It might make sense that this API does the work for them and only coin is provided. @paninaro what do you think?

I remember the requirement is adding an API to verify general signatures. The pubkey can be owned by anyone. It's just for answering "If this pubkey signed this message?"

paninaro:
> the challenging part here is to figure out the public key especially for coins. It might make sense that this API does the work for them and only coin is provided. @paninaro what do you think?

Yes, that was my feeling as well. This RPC effectively determines if a given signature is valid given the signed message and pubkey, but equally important is determining whether that pubkey is the pubkey that I expect as part of proving ownership.

The scenario I keep coming back to is this:

Prove to me that you control the XCH address `xch1xgpfy7uahfua5ndndm8pl0a0p8xks3ptcfkpgn73ukg78h3ldqyqlz870h`. 
I can provide a proof of this by using the `sign_message_by_address` RPC and returning the pubkey and signature to the challenger.

The challenger can use this RPC to verify that the signature is valid given the message and pubkey, but the challenger *also* needs to calculate the XCH address from the pubkey. If the calculated XCH address doesn't match the original XCH that the challenger was asking for proof of ownership, then it doesn't matter that the signature is valid -- it was created by a totally unrelated key.

I'm not sure if the right answer is to support passing in a coin or an XCH address, but verification is currently a two step process, and I think it would be nice to have a single RPC that can check both the signature and the pubkey's validity.

paninaro (dismissed):
looks good, thanks for add this!

trepca (dismissed):

trepca (approved):

paninaro (approved):

PR #13739, opened 2022-10-28
I can set a Chia Server to download a Data Store from last full data file,

Chida82:
Let's consider DataStores that are changed often and the keys overwritten often. An application that connects via RPC with Chia (a project of mine) must download these DataStores, but starting from item 1 the time taken to download all versions is very long. Furthermore, in "ecological" terms, unloading more than necessary has an impact.
This development allows you to download the latest version at the first request. This way you can build a dedicated server for this. I developed in server terms with a backward compatible explicit key so that it is explicit that the files located "\data_layer\db\server_files_location_mainnet" are not useful for mirroring. Its best use and with a docker to have a fast DataStore reader.
This is a first step sufficient to solve the problem, if accepted in the future it is also possible to reduce the files created in '\data_layer\db\server_files_location_mainnet'.

### DIFF
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -31,7 +31,7 @@ from chia.data_layer.data_layer_util import (
 )
 from chia.data_layer.data_layer_wallet import DataLayerWallet, Mirror, SingletonRecord, verify_offer
 from chia.data_layer.data_store import DataStore
-from chia.data_layer.download_data import insert_from_delta_file, write_files_for_root
+from chia.data_layer.download_data import insert_from_delta_file, insert_from_full_file, write_files_for_root
 from chia.rpc.rpc_server import default_get_connections
 from chia.rpc.wallet_rpc_client import WalletRpcClient
 from chia.server.outbound_message import NodeType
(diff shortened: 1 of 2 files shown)


chia/data_layer/data_layer.py
```diff
@@ -375,16 +375,28 @@ async def fetch_and_validate(self, tree_id: bytes32) -> None:
...
-                    self.data_store,
-                    tree_id,
-                    root.generation,
-                    [record.root for record in reversed(to_download)],
-                    server_info,
-                    self.server_files_location,
-                    timeout,
-                    self.log,
-                )
+                if self.config.get("use_last_full_tree_file", False) and root.generation == 0:
```
fchirica, line 378: I think this will not work. The way our download works is as following: let `root.generation` the local generation and `singleton_record.generation` the target generation. We download all trees by deltas corresponding to `root.generation + 1`, `root.generation + 2` ... all the way up to `singleton_record.generation`.

What this code will do is, at the first run, download the latest tree and insert it with `generation=1`. Then, when the next loop iteration will be made, we'll need to download all generations `2`, `3`, ... all the way up to the new `singleton_record.generation`. Basically, I _think_ it'll still download everything, but this time having `generation=1`
being calculated wrongly.

As a proposed fix, I suggest moving this line at the very beginning of the loop (right after retrieving `url` and `root`), so we can skip the `generation` checks. The new checks should be having no tree downloaded (when `root.generation = 0`) or having the root hash changed from the previously downloaded tree. Here's a code sketch (note that I didn't test it, but I feel it's easier to explain this by writing some code):

```
for server_info in servers_info:
    url = server_info.url
    root = await self.data_store.get_tree_root(tree_id=tree_id)
    if self.config.get("use_last_full_tree_file", False):
        root_hash = self.none_bytes if root.node_hash is None else root.node_hash
        # A new download is needed, since either we have no data for this store
        # or our previously stored data has changed.
        if root.generation == 0 or root_hash != singleton_record.root:
            success = await insert_from_full_file(
                self.data_store,
                tree_id,
                singleton_record.generation,
                to_download[0].root,
                server_info,
                self.server_files_location,
                timeout,
                self.log,
             )
             if success:
                 self.log.info(f"Successfully downloaded the latest tree state for {tree_id}: {singleton_record.root}")
                 break
             else:
                 continue
    else:
        self.log.info(f"Fetch data: no download needed for {tree_id}."
        break
```

fchirica (changes requested):
Thanks for the contribution! This feature is really nice. I wrote below my feedback.

chia/data_layer/data_layer.py
Chida82: thanks for the tip.  I create a more suitable datasource for testing.

Chida82: I'm deepening and I'm at a good point. Basically it's like developing the snapshot concept. I was able to debug and analyze better. I have to pass "Generation" in some functions and maybe modify "await self.upload_files (subscription.tree_id)".

Chida82:
@emlowe i create a new PR 13855 for verify commit.
https://github.com/Chia-Network/chia-blockchain/pull/13739

PR #13754, opened 2022-10-29
only open SQLite log file once per db wrapper

altendky:
Opening and closing files is extremely slow compared to just writing a line.  Now that we have log setup etc centralized inside the wrapper, we can better manage this and make sure that only a single file is opened and then used for each call.

Draft for:
- [x] Is there any easy benchmarking?
- [x] Should a periodic flush be added, or buffering otherwise considered and configured?
- [x] ~Since we only use a single file for all sql logging, this PR is probably doomed~

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -69,7 +69,7 @@ from chia.util.bech32m import encode_puzzle_hash
 from chia.util.check_fork_next_block import check_fork_next_block
 from chia.util.condition_tools import pkm_pairs
 from chia.util.config import PEER_DB_PATH_KEY_DEPRECATED, process_config_start_method
-from chia.util.db_wrapper import DBWrapper2, create_connection
+from chia.util.db_wrapper import DBWrapper2, manage_connection
 from chia.util.errors import ConsensusError, Err, ValidationError
 from chia.util.ints import uint8, uint32, uint64, uint128
 from chia.util.path import path_from_root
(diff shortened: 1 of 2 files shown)


chia/full_node/full_node.py
```diff
@@ -323,7 +323,7 @@ async def _start(self) -> None:
         self._respond_transaction_semaphore = asyncio.Semaphore(200)
         # create the store (db) and full node instance
         # TODO: is this standardized and thus able to be handled by DBWrapper2?
-        async with create_connection(self.db_path) as db_connection:
+        async with manage_connection(self.db_path) as db_connection:
```
altendky, line 326: Also shifting this so it ends up on the SQLite log as well.  https://github.com/Chia-Network/chia-blockchain/pull/13755

arvidn: is this the only call to `manage_connection`? there's no `log_path` or `name` ever passed in here.

chia/util/db_wrapper.py
```diff
@@ -130,13 +130,13 @@ async def create(
 
         write_connection.row_factory = row_factory
 
-        self = cls(connection=write_connection, db_version=db_version)
+        self = cls(connection=write_connection, db_version=db_version, log_file=log_file)
 
         for index in range(reader_count):
-            read_connection = await create_connection(
+            read_connection = await _create_connection(
```
arvidn, line 136: I see, this one passes on log_file and name

arvidn (approved):
looks good. We don't have any automated tests for the sql logging, so I take it you checked it still works manually, yeah?

chia/full_node/full_node.py
altendky: From my searching, yes, this is the only use outside of the wrapper module itself.  After #13755 it should end up specifying both `log_path` and `name`.

emlowe (approved):

Issue #13132, opened 2022-08-24
[Bug] Tons of "WARNING  Rate limiting ourselves. message type: respond_peers" messages in log 


jayhohoho2019:
### What happened?

After upgrade to 1.5.1 (from 1.5.0), full node shows a large number of "WARNING  Rate limiting ourselves. message type: respond_peers" messages in debug.log.  Full node is in sync.  Most connected peers are in sync, have even farmed a block since the upgrade, but these warning messages are filling up the log.  

### Version

1.5.1

### What platform are you using?

Linux

### What ui mode are you using?

CLI

### Relevant log output

```shell
2022-08-23T22:56:57.376 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_proof_of_weight, peer: 31.19.126.71
2022-08-23T22:56:58.384 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_proof_of_weight, peer: 31.19.126.71
2022-08-23T22:56:59.391 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_proof_of_weight, peer: 31.19.126.71
2022-08-23T22:57:01.625 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_peers, peer: 190.16.85.253
2022-08-23T22:57:01.626 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_peers, peer: 109.111.131.6

Last 5 lines of log

Since upgrade, 1851 lines of this message type has shown up in log, with a total of 14 unique IPs.
```

jayhohoho2019:
2022-08-23T16:59:22.007 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_peers, peer: 88.26.224.173
2022-08-23T16:59:22.263 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_peers, peer: 88.26.224.173
2022-08-23T17:04:02.144 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_peers, peer: 24.96.61.222

First 3 lines in log, to provide a timeframe for the total of 1851 lines of messages.

jayhohoho2019:
As of now Wed 24 Aug 2022 12:51:47 PM EDT, 4293 lines of this “Rate limiting" warning messages in log.  So that's during a 20 hour time span.

alnuweiri:
The same problem here.

[...]

PR #13758, opened 2022-10-30
Revert rate limiting messages from `warning` back to `debug`

hugepants:
In 1.5.1 this was changed from `debug` to `warning` causing a lot of excess noise in default logging as reported in https://github.com/Chia-Network/chia-blockchain/issues/13132

If there is an underlying issue that is causing this noise, it can still be identified with debug logging, but for the majority of users configured with `log_level: WARNING` or `INFO` this change suppresses the noise which is inflating log files a lot.


emlowe (approved):
Yea, I think this the best plan for now

PR #13759, opened 2022-10-30
strip input before bech32 decoding

altendky:
this avoids leading and trailing whitespace causing offer processing to fail


altendky:
Separately, this really should get descriptive exceptions instead of completely opaque nothing returns.

paninaro (approved):
lgtm

PR #13760, opened 2022-10-30
include memos in get_puzzle_and_solution return value

dkackman:
I am working on something similar in concept to [marmot minter](https://github.com/Yostra/marmot_minter). Part of he workflow entails passing information in the memo of a transaction from the sender to the receiver. In order to enable and simplify that via RPC, this PR patches that ability into `get_puzzle_and_solution`.

1) This code was added to the `CoinSpend`:
https://github.com/Yostra/marmot_minter/blob/40ff182ee21e3566d3d071aa5f9f3592c4b950eb/marmot_server.py#L56

2) `get_puzzle_and_solution` returns an additional field along with the serialized `CoinSpend`:
`return {"coin_solution": coin_solution, "memos": memos}`

This is nonbreaking both in inputs and outputs. It could be made opt-in with an `include_memos` if that's better.

### DIFF
--- a/chia/rpc/full_node_rpc_api.py
+++ b/chia/rpc/full_node_rpc_api.py
@@ -712,7 +712,10 @@ class FullNodeRpcApi:
         assert puzzle is not None
         assert solution is not None
 
-        return {"coin_solution": CoinSpend(coin_record.coin, puzzle, solution)}
+        coin_solution = CoinSpend(coin_record.coin, puzzle, solution)
+        memos = coin_solution.get_memos()
+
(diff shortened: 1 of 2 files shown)


trepca:
Hi Don, thanks for this, but can you also add tests for the new method.

dkackman:
> Hi Don, thanks for this, but can you also add tests for the new method.

you bet.

dkackman:
I'm gonna close this. The use case I'm working on is better supported by notifications. Thanks for taking a look at it @trepca

PR #13774, opened 2022-10-31
Bump for new clvm_tools and clvm_tools_rs versions

prozacchiwawa:


chia/wallet/trading/offer.py
```diff
@@ -417,7 +417,7 @@ def to_valid_spend(self, arbitrage_ph: Optional[bytes32] = None) -> SpendBundle:
                     sibling_spends: str = "("
                     sibling_puzzles: str = "("
                     sibling_solutions: str = "("
-                    disassembled_offer_mod: str = disassemble(OFFER_MOD)  # type: ignore
+                    disassembled_offer_mod: str = disassemble(OFFER_MOD)
```
arvidn, line 420: do you understand why we do this? Why do we have a string with the source code representation at all?
(if we absolutely need this, this particular case appears to be a constant, and should probably be pre-computed)

arvidn (approved):
looks reasonable. Do you know why we may have needed those mypy suppressions?

chia/wallet/trading/offer.py
prozacchiwawa: Kyle added the suppressions because types were added to clvm_tools and the changes haven't been reflected in chia-blockchain yet.  These cause failures in run-py-tests and the CI due to extra types now being available that weren't before.
I will peek at why offer does disassembly, not code i wrote.

prozacchiwawa: It definitely looks like sibling puzzles contains a list of copies of the disassembly of just OFFER_MOD.  I'll have to get in touch (probably with quexington) to know more about how this functions, but it's seeming from my read that these lists, sibling_spends, sibling_puzzles and sibling_solutions are used concurrently.  As for factoring out the constant, i'd defer to the writers of this code rather than fixing it myself.

prozacchiwawa: My read of it is that the Solver is multipurpose and takes data in a certain textural format with sexpressions.  I am not in the offer stack, so my knowledge is limited, but it appears that this supports discovery of the form a solution will take given the data named in the offer.  One imagines that design considerations led to this textural format and reuse led to disassembly here.

PR #13775, opened 2022-10-31
dont validate wp on log_in

almogdepaz:
remove redundant wp validation on log_in

### DIFF
--- a/chia/wallet/wallet_blockchain.py
+++ b/chia/wallet/wallet_blockchain.py
@@ -46,6 +46,8 @@ class WalletBlockchain(BlockchainInterface):
         self.constants = constants
         self.CACHE_SIZE = constants.SUB_EPOCH_BLOCKS * 3
         self.synced_weight_proof = await self._basic_store.get_object("SYNCED_WEIGHT_PROOF", WeightProof)
+        self._sub_slot_iters = await self._basic_store.get_object("SUB_SLOT_ITERS", uint64)
+        self._difficulty = await self._basic_store.get_object("DIFFICULTY", uint64)
         self._finished_sync_up_to = await self._basic_store.get_object("FINISHED_SYNC_UP_TO", uint32)
         if self._finished_sync_up_to is None:
             self._finished_sync_up_to = uint32(0)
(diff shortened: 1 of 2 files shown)


emlowe:
Almog, when this code was added (PR #12310) you said : 

> most of the validation is skipped using the flag, we do this to get the block records from the proof after restart since we don't cache them.
> 
> this happens only on a restart and replaces the initialization we used to do in wallet_blockchain.py"


https://github.com/Chia-Network/chia-blockchain/pull/12310#discussion_r940069089

almogdepaz:
> Almog, when this code was added (PR #12310) you said :
> 
> > most of the validation is skipped using the flag, we do this to get the block records from the proof after restart since we don't cache them.
> > this happens only on a restart and replaces the initialization we used to do in wallet_blockchain.py"
> 
> [#12310 (comment)](https://github.com/Chia-Network/chia-blockchain/pull/12310#discussion_r940069089)

that comment is correct, that validation used to happen when we called create on wallet_blockchain.py 
that was called by create in wallet_state_managed.py which was called by the start method in wallet_node.py
 
the logic stayed the same but i removed the need to propagate the weight_proof_handler to wallet_state_manager and wallet_blockchain, making the code simpler and having only one place where WP validation is done (this is similar to the full node)
 
 in this pr im adding caching to the values i talked about in the comment, looking into whats going on inside i realized we only need the last sub slot iterations and difficulty for initialization (the latest peak was already being cached)

chia/wallet/wallet_blockchain.py
```diff
@@ -78,6 +80,8 @@ async def new_valid_weight_proof(self, weight_proof: WeightProof, records: List[
 
         self._sub_slot_iters = records[-1].sub_slot_iters
         self._difficulty = uint64(records[-1].weight - records[-2].weight)
+        await self._basic_store.set_object("SUB_SLOT_ITERS", self._sub_slot_iters)
```
almogdepaz, line 83: @altendky 
@altendky 
i want to make sure the 3 different writes to the basic_store are under the same transaction here do i just call them under 
        async with self.block_store.db_wrapper.writer():
?

altendky: Yep, that's it.  It tracks connections and nesting etc internally.  The only caveat is that maybe someday we want to make the wrapper object private and so on, but for now, yup.

emlowe (approved):

Issue #13723, opened 2022-10-27
[Bug] PermissionError: [WinError 21]


risksoft:
### What happened?

After error PermissionError: [WinError 21] farmer does not see any plots. Updating the plots does not help, only restarting the program. 

### Version

1.6

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output

```shell
2022-10-27T14:39:57.283 full_node chia.full_node.full_node: INFO      Updated peak to height 2740037, weight 5700708424, hh 5e38809896ef3a037e236b3b69028f8516c3d6997d195aab02200cc30cf868a9, forked at 2740036, rh: 2bb1ced85d05799daf0e3bd45b5ebc4688f55e3b5b16348bedb43e0750ba16af, total iters: 11107282522647, overflow: False, deficit: 0, difficulty: 1776, sub slot iters: 137363456, Generator size: No tx, Generator ref list size: No tx
2022-10-27T14:39:57.337 wallet chia.wallet.wallet_blockchain: INFO     Peak set to: 2740037 timestamp: 1666870752
2022-10-27T14:39:57.410 farmer chia.farmer.farmer         : INFO     GET /farmer response: {'authentication_public_key': '0x8ca07f4d625fa8fc11bcf8abda1d08463469329e4e997235316366900192e728db5da8f52810c91f5468ac7eb0601897', 'payout_instructions': '39493dc9d87535b7309f3f7b6ded5b4643fa9f602cffc4920c12f320dfb9399c', 'current_difficulty': 134, 'current_points': 20116}
2022-10-27T14:39:59.359 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_peers, peer: 59.16.41.55
2022-10-27T14:39:59.359 full_node full_node_server        : WARNING  Rate limiting ourselves. message type: respond_peers, peer: 84.112.15.54
2022-10-27T14:40:03.950 full_node chia.full_node.full_node: INFO       Finished signage point 55/64: CC: f33ec5dee434fd3bef84e08c0303a7a77af207c6034d6e00e382a6a941239490 RC: d12a9c154816aa95b4e7a15f928fdc0f06e762dad73eaf361d79d4ecacf1bfd4 
2022-10-27T14:40:05.380 harvester chia.harvester.harvester: INFO     16 plots were eligible for farming 84112f64c7... Found 0 proofs. Time: 1.39869 s. Total 6472 plots
[...]

PR #13786, opened 2022-11-01
plotting: Catch exceptions raised when resolving plot directories

xdustinface:
Fixes issues like https://github.com/Chia-Network/chia-blockchain/issues/13723

### DIFF
--- a/chia/plotting/util.py
+++ b/chia/plotting/util.py
@@ -78,8 +78,12 @@ def get_plot_filenames(root_path: Path) -> Dict[Path, List[Path]]:
     config = load_config(root_path, "config.yaml")
     recursive_scan: bool = config["harvester"].get("recursive_plot_scan", False)
     for directory_name in get_plot_directories(root_path, config):
-        directory = Path(directory_name).resolve()
-        all_files[directory] = get_filenames(directory, recursive_scan)
+        try:
+            directory = Path(directory_name).resolve()
+        except (OSError, RuntimeError):
+            log.exception(f"Failed to resolve {directory_name}")
+        else:
+            all_files[directory] = get_filenames(directory, recursive_scan)
     return all_files
 
 
--- a/chia/plotting/util.py
+++ b/chia/plotting/util.py
@@ -82,8 +82,8 @@ def get_plot_filenames(root_path: Path) -> Dict[Path, List[Path]]:
             directory = Path(directory_name).resolve()
         except (OSError, RuntimeError):
             log.exception(f"Failed to resolve {directory_name}")
-        else:
-            all_files[directory] = get_filenames(directory, recursive_scan)
+            continue
+        all_files[directory] = get_filenames(directory, recursive_scan)
     return all_files
 
 


chia/plotting/util.py
```diff
@@ -78,8 +78,12 @@ def get_plot_filenames(root_path: Path) -> Dict[Path, List[Path]]:
...
     recursive_scan: bool = config["harvester"].get("recursive_plot_scan", False)
     for directory_name in get_plot_directories(root_path, config):
-        directory = Path(directory_name).resolve()
-        all_files[directory] = get_filenames(directory, recursive_scan)
+        try:
+            directory = Path(directory_name).resolve()
+        except (OSError, RuntimeError):
+            log.exception(f"Failed to resolve {directory_name}")
+        else:
+            all_files[directory] = get_filenames(directory, recursive_scan)
```
arvidn, line 86: it seems quite odd to me to use the `else` clause here. partly because it's an odd feature in python, but also because it defeats one of the main purposes of using exceptions. You can just move this line immediately following line 82, and you'll have (almost) the same behavior.

The main difference is that if `get_filenames()` throws, you still continue looping. I think that's better behavior. With this patch, `get_filenames()` throwing will terminate the loop, which I don't think is deliberate.

altendky (changes requested):
Requesting changes since there's at least discussion to be had.

chia/plotting/util.py
altendky: What's the `RuntimeError` that can happen there?  Indicating the error would be great.  Not found, permissions, etc.  Also the directory that's having trouble.  Otherwise I have no information to take action on.

Without saying it shouldn't be done here, I think that handling exceptions from `get_filenames()` is a separate enhancement.  What's here addresses the directly reported failure.  But yeah, handling some more exceptions would probably be nice while we're here.

Putting them in the same `try:` block means you can't differentiate the source of the error and have to provide a generic message that covers both.  I might do separate `try:`/`except:` blocks with a continue in the first `except:` after logging?  Or both for consistency.

altendky (commented):
and...  here's my comment i forgot to submit...

chia/plotting/util.py
xdustinface: > What's the `RuntimeError` that can happen there? Indicating the error would be great. Not found, permissions, etc. Also the directory that's having trouble. Otherwise I have no information to take action on.

https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve

See the docs for the `RuntimeError` and since in the linked issue we get the `PermissionError` which is a `OSError` i just picked `OSError` in case similar related errors coming up in `Path.resolve`.


@arvidn @altendky About your other concerns: I moved the `get_filenames` out there, yeah, to just address the actual issue. At least from my understanding and in its current implementation `get_filenames` doesn't raise really. The alternative then to have a `else` is to have a `continue` after the logging to have the correct behaviour. At the end i think the related code here can improved in various places and could be rewritten to not catch some exceptions in the one method and few in the other because it's all kind of related anyway but yeah, just trying to address the linked issue here and not open any rabbit holes :) Do we really need a change here? Let me know.

altendky: I now see the `except Exception as e:` and...  `*runs and hides under a rock*`  so yeah.  I think reporting the directory that had an issue is important and reporting the error is important.  I think a `continue` fits better here since it avoids the pattern of nesting code in a post-`try:` block.

Maybe wait for Arvid to weigh in so we don't make you flip flop a bunch.

arvidn: logging the error makes sense, and I think the current logging does that well. I still would like some justification for using the `else` block on a `try` block. it adds an extra line of code and I think it also adds confusion.

xdustinface: @arvidn I changed it in a9d1621ef1d63286deb51e06cf9382a8866b617b so we can just merge it but it doesn't really add a line of code if we want to distinguish the failure vs having a general error message, i.e. we only catch errors in `resolve` here. Its:

```python
try:
    directory = Path(directory_name).resolve()
except (OSError, RuntimeError):
    log.exception(f"Failed to resolve {directory_name}")
else:
    all_files[directory] = get_filenames(directory, recursive_scan)
```

vs 

```python
try:
    directory = Path(directory_name).resolve()
except (OSError, RuntimeError):
    log.exception(f"Failed to resolve {directory_name}")
    continue
all_files[directory] = get_filenames(directory, recursive_scan)
```

and the `else` is not more confusing than having a `continue` imo, actually i think the first block looks cleaner. I mean.. like other language specific things, if you got it, you got it and "if it doesn't raise, do this instead" doesn't sound that confusing, does it?

arvidn: in my mind, the natural way of writing this would be:
```
try:
    directory = Path(directory_name).resolve()
    all_files[directory] = get_filenames(directory, recursive_scan)
except (OSError, RuntimeError):
    log.exception(f"Failed to resolve {directory_name}")
```

altendky: The exception handling is intended only for the `.resolve()` call.  We expect `get_filenames()` to handle it's own errors presently (the code seems to).  If we put both things in the `try:` then we have to have a more generic error.  Also, as is, we still don't know if the path doesn't exist, or it's a permission error, or...  something more arcane.

arvidn: I see

arvidn (approved):

chia/plotting/util.py
altendky: I was overlooking the use of `log.exception()` that will include the information I'm interested in.  Sorry for that.

altendky (approved):

PR #13799, opened 2022-11-02
Identical spend aggregation

AmineKhaldi:


chia/full_node/mempool_manager.py
```diff
@@ -491,13 +519,25 @@ async def validate_spend_bundle(
...
-                    sb: MempoolItem = self.mempool.spends[c_sb_id]
-                    conflicting_pool_items[sb.name] = sb
-            log.warning(f"Conflicting pool items: {len(conflicting_pool_items)}")
-            if not self.can_replace(conflicting_pool_items, removal_record_dict, fees, fees_per_cost):
-                return Err.MEMPOOL_CONFLICT, potential, []
+            conflicts = []
+            for coin in potential_conflicts:
+                # See if this coin spend is eligible for deduplication
+                coin_id = coin.name()
+                eligible_coin_spends_in_item = [
```
arvidn, line 526: I'm not sure exactly what you're doing here. you need to know whether a coin spend is eligible or not. You can just build a set of all eligible coins up-front, before looping over `potential_conflicts`.

Right now, the mempool has an invariant that no coin is spent multiple times, right? This is a big change, and should have some comments added in a few places to make this clear.

chia/full_node/mempool_manager.py
```diff
@@ -142,34 +143,61 @@ async def create_bundle_from_mempool(
...
+                        # Process this item with this reduced cost in mind
+                        duplicate_cost = dedup_coin_spends[coin_spend_solution_hash]
+                        item_cost -= duplicate_cost
+                        dedup_spends = True
+                    else:
+                        # Process this item normally but store the isolated spend cost future reference
+                        # TODO: Ask Arvid whether INFINITE_COST is adequate here (and whether this is how to get the cost in the first place)
+                        coin_spend_cost = uint64(
+                            coin_spend_in_bundle.puzzle_reveal.run_with_cost(
+                                INFINITE_COST, coin_spend_in_bundle.solution
```
arvidn, line 178: the cost limit here should be the cost of the whole spend-bundle. Just running one solution should never exceed that. Technically, a solution shouldn't make it in here if it's too costly either, but using `INFINITE` makes me nervous. We should set it to the true upper limit of this solution.

arvidn (commented):
I think it would be good to split some of this up into separate functions that can be tested independently

chia/full_node/mempool_manager.py
```diff
@@ -142,35 +143,82 @@ async def create_bundle_from_mempool(
...
-                    removals.extend(item.removals)
-                    additions.extend(item.additions)
-                else:
-                    broke_from_inner_loop = True
-                    break
-        if len(spend_bundles) > 0:
+                consider_this_item = True
+                cost_saving = 0
+                spends_to_dedup: List[CoinSpend] = []
+                spends_to_run_for_cost: List[Tuple[bytes32, CoinSpend]] = []
```
arvidn, line 162: I really don't think we should compute the isolated cost for a puzzle+solution until we need to. so, the first time we find a duplicate, that's when we should compute this cost. I imagine removing this list and making the cost field optional in `dedup_coin_spends`. The first time you need the cost, compute it then.

Also, computing the cost should be its own function that you can unit test independently.

chia/full_node/mempool_manager.py
```diff
@@ -142,35 +143,82 @@ async def create_bundle_from_mempool(
...
+                        else:
+                            # It wasn't, so let's skip this whole transaction
+                            consider_this_item = False
+                            cost_saving = 0
+                            break
+                    else:
+                        # We didn't process an item with this coin before
+                        # If we end up including this transaction, add this item to
+                        # the ones we need to run in isolation and store the cost of
+                        spends_to_run_for_cost.append((coin_id, coin_spend_in_bundle))
```
arvidn, line 185: consider the case where you have a duplicate spend in the same bundle. Would this work then? (that should be a test case)

chia/full_node/mempool_manager.py
```diff
@@ -142,35 +143,82 @@ async def create_bundle_from_mempool(
...
+                if consider_this_item:
+                    # Run the spends that we're processing for the first time, for cost
+                    new_dedup_coin_spends: Dict[bytes32, Tuple[bytes32, uint64]] = {}
+                    for coin_id, coin_spend in spends_to_run_for_cost:
+                        spend_cost = uint64(coin_spend.puzzle_reveal.run_with_cost(item.cost, coin_spend.solution)[0])
```
arvidn, line 190: this should not fail, but if it does, you probably want to ignore the whole spend bundle and move on. Right now it looks like you don't consider errors returned from this function

chia/full_node/mempool_manager.py
```diff
@@ -142,35 +143,82 @@ async def create_bundle_from_mempool(
...
+                        new_dedup_coin_spends[coin_id] = (coin_spend.solution.get_tree_hash(), spend_cost)
+                        cost_saving += spend_cost
```
arvidn, line 192: this is the first time we include this spend, isn't it? I don't see how this could give us any cost savings. This should be a unit test as well; having only one spend that is eligible for dedup (but doesn't have any duplicates). It should not affect the cost of the resulting bundle.

chia/full_node/mempool_manager.py
```diff
@@ -142,35 +143,82 @@ async def create_bundle_from_mempool(
...
+                        and item.fee + fee_sum <= self.constants.MAX_COIN_AMOUNT
+                    ):
+                        # Update the deduplicated coin spends map
+                        dedup_coin_spends.update(new_dedup_coin_spends)
+                        # Add all the spends in the transaction except the duplicated ones
+                        for cs in item.spend_bundle.coin_spends:
+                            if cs not in spends_to_dedup:
+                                coin_spends.append(cs)
+                                removals.append(cs.coin)
+                        additions.update(item.additions)
```
arvidn, line 206: do you mutate `item.additions`? it seems unintuitive to have removed duplicates from this object.
This looks wrong. It looks like you include the duplicate spends here that have been removed.

chia/full_node/mempool_manager.py
```diff
@@ -142,35 +143,82 @@ async def create_bundle_from_mempool(
...
+        # This is a map of coin ID to a pair of the coin spend solution hash and its isolated cost
+        dedup_coin_spends: Dict[bytes32, Tuple[bytes32, uint64]] = {}
+        coin_spends: List[CoinSpend] = []
+        sigs: List[G2Element] = []
         broke_from_inner_loop = False
         log.info(f"Starting to make block, max cost: {self.constants.MAX_BLOCK_COST_CLVM}")
         for dic in reversed(self.mempool.sorted_spends.values()):
             if broke_from_inner_loop:
                 break
             for item in dic.values():
```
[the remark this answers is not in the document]
AmineKhaldi: I attempted to do that via `check_item_for_dedup` and co.

chia/full_node/mempool_manager.py
```diff
@@ -82,6 +84,69 @@ def validate_clvm_and_signature(
     return None, bytes(result), new_cache_entries
 
 
+def compute_spend_cost(
+    puzzle_reveal: SerializedProgram, solution: SerializedProgram, maximum_cost: uint64
+) -> Optional[uint64]:
+    try:
+        cost, _ = puzzle_reveal.run_with_cost(maximum_cost, solution)
```
arvidn, line 91: you actually get the `CREATE_COIN` conditions out of running this here, so you could subtract these from the additions when you remove the duplicate

chia/full_node/mempool_manager.py
```diff
@@ -82,6 +84,69 @@ def validate_clvm_and_signature(
...
+        return uint64(cost)
+    except Exception:
+        return None
+
+
+def check_item_for_dedup(
+    item: MempoolItem, dedup_coin_spends: Dict[bytes32, Tuple[bytes32, Optional[uint64]]]
```
arvidn, line 98: the `Tuple[]` at the end here, the solution hash and cost, I think it might be appropriate to actually make a small dataclass for this, to get named fields.
Also, storing a `SerializedProgram` here is cheaper than the hash, since it's just a pointer to the buffer we already have stored in a mempool item

chia/full_node/mempool_manager.py
```diff
@@ -82,6 +84,69 @@ def validate_clvm_and_signature(
...
+    new_dedups: Dict[bytes32, Tuple[bytes32, Optional[uint64]]] = {}
+    # See if this item has coin spends that are eligible for deduplication
+    assert item.npc_result.conds is not None
+    eligible_coin_ids = [bytes32(s.coin_id) for s in item.npc_result.conds.spends if s.eligible_for_dedup]
+    for coin_id in eligible_coin_ids:
+        # Pick the first coin spend with coin_id
+        # npc_result.conds.spends should be consistent with
+        # spend_bundle.coin_spends at this point, guaranteeing the presence of
+        # this coin spend in this bundle as a result
+        coin_spend_in_bundle = next(cs for cs in item.spend_bundle.coin_spends if cs.coin.name() == coin_id)
```
arvidn, line 112: presumably there's exactly one coin whose coin ID is `coin_id`. Instead of performing a linear search here, how about `eligible_coin_ids` just contain a pointer to the `CoinSpend` object instead?

chia/full_node/mempool_manager.py
```diff
@@ -82,6 +84,69 @@ def validate_clvm_and_signature(
...
+        return None
+
+
+def check_item_for_dedup(
+    item: MempoolItem, dedup_coin_spends: Dict[bytes32, Tuple[bytes32, Optional[uint64]]]
+) -> Optional[Tuple[List[CoinSpend], uint64, Dict[bytes32, Tuple[bytes32, Optional[uint64]]]]]:
+    consider_this_item = True
+    cost_saving = uint64(0)
+    spends_to_dedup: List[CoinSpend] = []
+    new_dedups: Dict[bytes32, Tuple[bytes32, Optional[uint64]]] = {}
```
arvidn, line 103: instead of creating a new dictionary here, can't we just update `dedup_coin_spends` in place? that seems simpler in a lot of ways. But it should be made clear at the call-site

chia/full_node/mempool_manager.py
```diff
@@ -157,61 +222,36 @@ async def create_bundle_from_mempool(
...
+                    # the same solution), so those additions get deduplicated
+                    # naturally on set update.
+                    # Ideally we would have more explicit checks
+                    additions.update(item.additions)
+                    sigs.append(item.spend_bundle.aggregated_signature)
+                    cost_sum += item_cost
+                    fee_sum += item.fee
+                    processed_spend_bundles += 1
+                else:
+                    broke_from_inner_loop = True
```
arvidn, line 253: when I see this patter, of recording that we need to break another level of loop, I think that the code would probably benefit from being pulled out into a function that can just `return` out of the two loops instead,

chia/full_node/mempool_manager.py
```diff
@@ -80,6 +93,84 @@ def validate_clvm_and_signature(
...
+                    # If we never ran this before, additions should be empty
+                    assert len(duplicate_additions) == 0
+                    result = run_for_cost_and_additions(
+                        coin_id, coin_spend_in_bundle.puzzle_reveal, coin_spend_in_bundle.solution, item.cost
+                    )
+                    if result is None:
+                        # We failed to run this puzzle, skip this whole item
+                        consider_this_item = False
+                        cost_saving = uint64(0)
+                        break
```
arvidn, line 149: it would be easier to read (and fewer lines of code) to just say `return None` here. There's no need to assign values to those variables and break. The effect is that we'll return `None`

chia/full_node/mempool_manager.py
```diff
@@ -80,6 +93,84 @@ def validate_clvm_and_signature(
...
+                    else:
```
arvidn, line 150: this `else` block is redundant. You can remove `else` and reduce the indentation level by one of the code below

chia/full_node/mempool_manager.py
```diff
@@ -80,6 +93,84 @@ def validate_clvm_and_signature(
...
+                        # If we end up including this item, update this entry's cost and additions
+                        new_dedups[coin_id] = DedupCoinSpend(processed_solution, spend_cost, duplicate_additions)
+                cost_saving = uint64(cost_saving + duplicate_cost)
+                dedup_additions.update(duplicate_additions)
+                spends_to_dedup.append(coin_spend_in_bundle)
+            else:
+                # It wasn't, so let's skip this whole transaction
+                consider_this_item = False
+                cost_saving = uint64(0)
+                break
```
arvidn, line 163: same thing here; you can just return `None`, no need to assign to the variables. I would also recommend inverting the condition in the if-statement to make this an early exit. Then you can reduce the level of indentation by one of the remaining code.

chia/full_node/mempool_manager.py
```diff
@@ -80,6 +93,84 @@ def validate_clvm_and_signature(
...
+        else:
+            # We didn't process an item with this coin before. If we end up including
+            # this item, add this pair to dedup_coin_spends
+            new_dedups[coin_id] = DedupCoinSpend(solution_in_bundle, None, set())
+    if consider_this_item:
```
arvidn, line 168: If you have early exits, you don't need this variable anymore, and you can just unconditionally return the 4 return values here

chia/full_node/mempool_manager.py
```diff
@@ -80,6 +93,84 @@ def validate_clvm_and_signature(
     return None, bytes(result), new_cache_entries
 
 
+def run_for_cost_and_additions(
```
arvidn, line 96: this function should have a unit test

chia/full_node/mempool_manager.py
```diff
@@ -80,6 +93,84 @@ def validate_clvm_and_signature(
...
+    coin_id: bytes32, puzzle_reveal: SerializedProgram, solution: SerializedProgram, maximum_cost: uint64
+) -> Optional[Tuple[uint64, List[Coin]]]:
+    try:
```
arvidn, line 99: if you make `check_item_for_dedup()` signal errors via an exception instead of returning `None`, you can make this function `run_for_cost_and_additions()` also signal errors via an exception instead of returning `None`. i.e. you could remove this `try`-`catch` block

chia/full_node/mempool_manager.py
```diff
@@ -80,6 +93,84 @@ def validate_clvm_and_signature(
...
+                spends_to_dedup.append(coin_spend_in_bundle)
+            else:
+                # It wasn't, so let's skip this whole transaction
+                consider_this_item = False
+                cost_saving = uint64(0)
+                break
+        else:
+            # We didn't process an item with this coin before. If we end up including
+            # this item, add this pair to dedup_coin_spends
+            new_dedups[coin_id] = DedupCoinSpend(solution_in_bundle, None, set())
```
arvidn, line 167: since this block is short, if you invert the condition in the if-statement and add a `continue` here, you can remove the (new) `else`-block and reduce its indentation by one level.

chia/full_node/mempool_manager.py
```diff
@@ -80,6 +93,84 @@ def validate_clvm_and_signature(
...
+            duplicate_additions = dedup_coin_spend.additions
+            if processed_solution == solution_in_bundle:
+                # Let's calculate the saved cost if we never did that before
+                if duplicate_cost is None:
+                    # If we never ran this before, additions should be empty
+                    assert len(duplicate_additions) == 0
+                    result = run_for_cost_and_additions(
+                        coin_id, coin_spend_in_bundle.puzzle_reveal, coin_spend_in_bundle.solution, item.cost
+                    )
+                    if result is None:
```
arvidn, line 145: if `run_for_cost_and_additions()` throw to indicate an error, you can just focus on the happy-path here, and not even bother to turn `result is None` into `return None`. The error is propagated for you.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -136,7 +178,92 @@ async def test_block_cost_exceeds_max() -> None:
...
+    with pytest.raises(ValueError, match="('cost exceeded', '01')"):
+        run_for_cost_and_additions(TEST_COIN_ID, IDENTITY_PUZZLE, solution, uint64(1))
+    created_coin = Coin(TEST_COIN_ID, IDENTITY_PUZZLE_HASH, 1)
+    expected_cost, expected_coins = run_for_cost_and_additions(TEST_COIN_ID, IDENTITY_PUZZLE, solution, uint64(50))
+    assert expected_cost == 44
+    assert expected_coins == [created_coin]
+
+
+@pytest.mark.asyncio
+async def test_check_item_for_dedup() -> None:
```
arvidn, line 199: I think you should split this function up into a separate function for each case that you test

tests/core/mempool/test_mempool_manager.py
```diff
@@ -136,7 +178,92 @@ async def test_block_cost_exceeds_max() -> None:
...
+    mempool_manager = await instantiate_mempool_manager(get_coin_record)
+
+    # No eligible coins, nothing to deduplicate, item gets considered normally
+    conditions = [[ConditionOpcode.AGG_SIG_UNSAFE, G1Element(), IDENTITY_PUZZLE_HASH]]
+    _, sb_name, res = await generate_and_add_spendbundle(mempool_manager, TEST_COIN, conditions)
+    assert res == (2873056, MempoolInclusionStatus.SUCCESS, None)
+    mempool_item = get_mempool_item(mempool_manager, sb_name)
+    dedup_coin_spends: Dict[bytes32, DedupCoinSpend] = {}
+    result = check_item_for_dedup(mempool_item, dedup_coin_spends)
+    assert result == ([], 0, {}, set())
```
arvidn, line 217: it really seems unnecessarily complicated to create a mempool and add an item to it just to test `check_item_for_dedup()`. I think you should just create the `MempoolItem` independently. It might sound like a nit, but we want to make each test case as simple and small as possible do lower the friction to add more cases.

chia/full_node/mempool_manager.py
```diff
@@ -81,6 +94,67 @@ def validate_clvm_and_signature(
...
+        if condition[0] == ConditionOpcode.CREATE_COIN:
+            puzzle_hash = bytes32(condition[1])
+            amount = uint64(int.from_bytes(condition[2], "big"))
+            created_coins.append(Coin(coin_id, puzzle_hash, amount))
+    return (uint64(cost), created_coins)
+
+
+def check_item_for_dedup(
+    item: MempoolItem, dedup_coin_spends: Dict[bytes32, DedupCoinSpend]
+) -> Tuple[List[CoinSpend], uint64, Dict[bytes32, DedupCoinSpend], Set[Coin]]:
```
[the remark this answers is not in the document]
AmineKhaldi: Done.

chia/full_node/mempool_manager.py
```diff
@@ -81,6 +94,67 @@ def validate_clvm_and_signature(
...
+        if coin_id not in dedup_coin_spends:
+            # We didn't process an item with this coin before. If we end up including
+            # this item, add this pair to dedup_coin_spends
+            new_dedups[coin_id] = DedupCoinSpend(solution_in_bundle, None, set())
+            continue
+        # See if the solution was identical
+        dedup_coin_spend = dedup_coin_spends[coin_id]
+        processed_solution = dedup_coin_spend.solution
+        duplicate_cost = dedup_coin_spend.cost
+        duplicate_additions = dedup_coin_spend.additions
```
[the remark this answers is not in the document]
AmineKhaldi: Done.

tests/core/mempool/test_mempool_manager.py

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Done.

chia/full_node/mempool_manager.py
```diff
@@ -103,6 +116,81 @@ def compute_assert_height(
     return height
 
 
+def run_for_cost_and_additions(
+    coin_id: bytes32, puzzle_reveal: SerializedProgram, solution: SerializedProgram, maximum_cost: uint64
+) -> Tuple[uint64, List[Coin]]:
+    cost, conditions = puzzle_reveal.run_with_cost(maximum_cost, solution)
+    created_coins = []
+    for condition in conditions.as_python():
```
arvidn, line 124: it's not safe to use `as_python()`. It will (recursively) convert the *entire* lisp tree into python objects. If the puzzle is malicious, that's more than you have RAM.

Instead, you need to iterate the list of `SExp` objects, pairs and atoms. You could try to call `as_iter()` on the `SExp` object. It will give you an iterator over `SExp` items in the list.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -213,7 +247,121 @@ async def get_coin_record(coin_id: bytes32) -> Optional[CoinRecord]:
...
+    expected_spends_to_dedup = []
+    expected_spends_to_dedup.extend(sb1.coin_spends)
+    expected_spends_to_dedup.extend(sb2.coin_spends)
+    assert spends_to_dedup == expected_spends_to_dedup
+    saved_cost2 = uint64(44)
+    created_coins2 = {Coin(TEST_COIN_ID2, IDENTITY_PUZZLE_HASH, 3)}
+    expected_new_dedups = {TEST_COIN_ID2: DedupCoinSpend(second_solution, saved_cost2, created_coins2)}
+    assert cost_saving == saved_cost + saved_cost2
+    assert new_dedups == expected_new_dedups
+    assert dedup_additions == created_coins.union(created_coins2)
```
arvidn, line 367: I think it's clearer if you specify exactly what you expect the result to be, rather than computing it with `union`. There are only 3 items involved, right?

tests/core/mempool/test_mempool_manager.py
```diff
@@ -213,7 +247,121 @@ async def get_coin_record(coin_id: bytes32) -> Optional[CoinRecord]:
...
+    sb2 = spend_bundle_from_conditions(TEST_COIN2, second_conditions)
+    sb3_conditions = [[ConditionOpcode.AGG_SIG_UNSAFE, G1Element(), IDENTITY_PUZZLE_HASH]]
+    sb3 = spend_bundle_from_conditions(TEST_COIN3, sb3_conditions)
+    sb = SpendBundle.aggregate([sb1, sb2, sb3])
+    mempool_item = mempool_item_from_spendbundle(sb)
+    spends_to_dedup, cost_saving, new_dedups, dedup_additions = find_duplicate_spends(mempool_item, dedup_coin_spends)
+    expected_spends_to_dedup = []
+    expected_spends_to_dedup.extend(sb1.coin_spends)
+    expected_spends_to_dedup.extend(sb2.coin_spends)
+    assert spends_to_dedup == expected_spends_to_dedup
```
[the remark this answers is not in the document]

tests/core/mempool/test_mempool_manager.py
```diff
@@ -213,7 +247,121 @@ async def get_coin_record(coin_id: bytes32) -> Optional[CoinRecord]:
...
     assert error == expected_error
+
+
+def test_run_for_cost_and_additions() -> None:
+    conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, 1]]
+    solution = Program.to(conditions)
+    with pytest.raises(ValueError, match="('cost exceeded', '01')"):
+        run_for_cost_and_additions(TEST_COIN_ID, IDENTITY_PUZZLE, solution, uint64(1))
+    created_coin = Coin(TEST_COIN_ID, IDENTITY_PUZZLE_HASH, 1)
+    expected_cost, expected_coins = run_for_cost_and_additions(TEST_COIN_ID, IDENTITY_PUZZLE, solution, uint64(50))
```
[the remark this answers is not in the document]

tests/core/mempool/test_mempool_manager.py

tests/core/mempool/test_mempool_manager.py
```diff
@@ -213,7 +247,121 @@ async def get_coin_record(coin_id: bytes32) -> Optional[CoinRecord]:
...
     _, status, error = result[2]
     assert status == expected_status
     assert error == expected_error
+
+
+def test_run_for_cost_and_additions() -> None:
+    conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, 1]]
+    solution = Program.to(conditions)
+    with pytest.raises(ValueError, match="('cost exceeded', '01')"):
+        run_for_cost_and_additions(TEST_COIN_ID, IDENTITY_PUZZLE, solution, uint64(1))
```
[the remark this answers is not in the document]
AmineKhaldi: I added a couple more.

chia/full_node/mempool_manager.py
```diff
@@ -103,6 +117,84 @@ def compute_assert_height(
...
+        if duplicate_cost is None:
+            # If we never ran this before, additions should be empty
+            assert len(duplicate_additions) == 0
+            spend_cost, created_coins = run_for_cost_and_additions(
+                coin_id, coin_spend_in_bundle.puzzle_reveal, coin_spend_in_bundle.solution, item.cost
+            )
+            duplicate_cost = spend_cost
+            duplicate_additions.update(created_coins)
+            # If we end up including this item, update this entry's cost and additions
+            new_dedups[coin_id] = DedupCoinSpend(current_solution, spend_cost, duplicate_additions)
```
arvidn, line 191: it might make sense to keep this around in the `MempoolItem` as well. in case this won't make it into the block, we wouldn't want to have to run the puzzle again next time.

chia/full_node/mempool_manager.py
```diff
@@ -103,6 +117,84 @@ def compute_assert_height(
     return height
 
 
+def run_for_cost_and_additions(
+    coin_id: bytes32, puzzle_reveal: SerializedProgram, solution: SerializedProgram, max_cost: uint64
+) -> Tuple[uint64, List[Coin]]:
+    cost, conditions_program = puzzle_reveal.run_mempool_with_cost(max_cost, solution)
+    created_coins = []
+    for s in conditions_program.as_iter():
```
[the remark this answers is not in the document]
arvidn: as you pointed out on keybase; this function should also account for the cost of `CREATE_COIN`, but we should probably not account for the cost-per-byte of the puzzle. We'll soon have compression that would deduplicate any identical puzzle reveals in the block anyway, so that's probably not a significant enough saving to count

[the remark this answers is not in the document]
AmineKhaldi: Done.

chia/full_node/mempool_manager.py
```diff
@@ -118,19 +119,25 @@ def compute_assert_height(
 
 
 def run_for_cost_and_additions(
-    coin_id: bytes32, puzzle_reveal: SerializedProgram, solution: SerializedProgram, max_cost: uint64
+    coin_id: bytes32,
+    puzzle_reveal: SerializedProgram,
+    solution: SerializedProgram,
+    max_cost: uint64,
```
arvidn, line 123: instead of passing in `max_cost`, I think you could use `npc_result.cost`. Would that work?

chia/full_node/mempool_manager.py
```diff
@@ -118,19 +119,25 @@ def compute_assert_height(
...
-        atoms = s.as_atom_list()
-        if len(atoms) == 0:
-            raise ValueError("Invalid condition while running for cost and additions")
-        if atoms[0] == ConditionOpcode.CREATE_COIN:
-            puzzle_hash = bytes32(atoms[1])
-            amount = uint64(int_from_bytes(atoms[2]))
-            created_coins.append(Coin(coin_id, puzzle_hash, amount))
-    return (uint64(cost), created_coins)
+    for spend in npc_result.conds.spends:
+        if spend.coin_id == coin_id:
```
arvidn, line 130: you'll only ever find a single spend matching the coin ID, you don't need to keep searching once you've found it.
I think a more clear way of saying this is to first find that spend, then do the loop you have below, to copy the created coins.

iirc, there's an `index()` function in python that returns the index to a matching element in a list. I'm not sure that would work here. But something like that I think would make sense.

chia/full_node/mempool_manager.py
AmineKhaldi: I checked how mempool items' cost is defined and indeed npc_result's cost can be used here. Done.

chia/full_node/mempool_manager.py
AmineKhaldi: Done.

chia/full_node/mempool_manager.py
```diff
@@ -254,15 +383,18 @@ def always(bundle_name: bytes32) -> bool:
             item_inclusion_filter = always
 
         log.info(f"Starting to make block, max cost: {self.max_block_clvm_cost}")
-        spend_bundles, cost_sum, additions, removals = self.process_mempool_items(item_inclusion_filter)
-        if len(spend_bundles) == 0:
+        processed_spend_bundles, sigs, coin_spends, cost_sum, additions, removals = self.process_mempool_items(
```
arvidn, line 386: I think returning this many values from a function is a code smell. Can you think of ways to simplify this? Maybe the interface to this function can shift a little to require fewer return values.

For instance, it could return an `Optional[SpendBundle]` instead of `coin_spends`, `sigs` and `processed_spend_bundles`.
It could print the info log message instead of returning `cost_sum`.

chia/full_node/mempool_manager.py
```diff
@@ -722,8 +860,17 @@ def can_replace(
         # fee than AB therefore kicking out A altogether. The better way to solve this would be to keep a cache
         # of booted transactions like A, and retry them after they get removed from mempool due to a conflict.
         for coin in item.removals:
-            if coin.name() not in removal_names:
-                log.debug(f"Rejecting conflicting tx as it does not spend conflicting coin {coin.name()}")
+            coin_id = coin.name()
+            if coin_id not in removal_names:
+                # NOTE: Coins that are eligible for deduplication are excluded from the superset rule. This
+                # prevents an attacker from replacing a transaction with one that spends an eligible coin
+                # additionally, rendering it impossible to replace the original transaction by fee later on.
```
arvidn, line 867: I don't think this comment is correct. It is possible to replace such transaction, it's just that our wallet doesn't support it. But it doesn't support replace-by-fee anyway.
I think it would be a mistake to exclude coins that are eligible for de-duplication from the superset rule.
unless I'm missing something (in which case this comment should probably describe the attack in greater details)

chia/full_node/mempool_manager.py
```diff
@@ -143,6 +155,99 @@ def compute_assert_height(
...
+            # spending this coin with a different solution and that would
+            # conflict with the coin spends that we're deduplicating already
+            raise ValueError("Solution is different from what we're deduplicating on")
+        # Let's calculate the saved cost if we never did that before
+        if duplicate_cost is None:
+            # If we never ran this before, additions should be empty
+            assert len(duplicate_additions) == 0
+            # See first if this mempool item had this cost computed before
+            # This can happen if this item didn't get included in the previous block
+            if coin_id in item.dedup_spend_info:
```
arvidn, line 229: I think it might make sense to avoid the double-lookup of `coin_id` in here, by using `item.dedup_spend_info.get(coin_id)`

chia/full_node/mempool.py
```diff
@@ -101,7 +105,13 @@ def _row_to_item(self, row: sqlite3.Row) -> MempoolItem:
...
 
         return MempoolItem(
-            item.spend_bundle, uint64(fee), item.npc_result, name, uint32(item.height_added_to_mempool), assert_height
+            item.spend_bundle,
+            uint64(fee),
+            item.npc_result,
+            name,
+            uint32(item.height_added_to_mempool),
+            assert_height,
+            dedup_spend_info=item.dedup_spend_info,
```
arvidn, line 114: it's a bug not to set the `assert_before_*` fields. I'll file a separate PR to fix that

arvidn (commented):
I think it would be good to run this patch through the mempool benchmark, comparing it to main. Maybe even amend the benchmark to cover some more interesting edge cases, like spend bundles with a lot of coin spends.

chia/full_node/mempool.py
```diff
@@ -312,3 +314,61 @@ def at_full_capacity(self, cost: int) -> bool:
         """
 
         return self.total_mempool_cost() + cost > self.mempool_info.max_size_in_cost
+
+    def get_items_spending_coin_ids(self, spent_coin_ids: List[bytes32]) -> List[MempoolItem]:
```
arvidn, line 318: I think this function needs a unit test

chia/full_node/mempool_manager.py
```diff
@@ -425,18 +388,35 @@ async def validate_spend_bundle(
...
+        removal_names_from_coin_spends: Set[bytes32] = set()
+        bundle_coin_spends: Dict[bytes32, BundleCoinSpend] = {}
+        for coin_spend in new_spend.coin_spends:
+            coin_id = coin_spend.coin.name()
+            removal_names_from_coin_spends.add(coin_id)
+            eligible_for_dedup, spend_additions = eligibility_and_additions.get(coin_id, (False, []))
+            bundle_coin_spends[coin_id] = BundleCoinSpend(coin_spend, eligible_for_dedup, spend_additions)
+
+        if removal_names != removal_names_from_coin_spends:
+            # If you reach here it's probably because your program reveal doesn't match the coin's puzzle hash
```
arvidn, line 418: do we have a test covering the case described in this comment?
do we have a test exercising this error in some other way?

chia/full_node/mempool_manager.py
```diff
@@ -425,18 +388,35 @@ async def validate_spend_bundle(
...
+            if not is_eligible:
+                non_eligible_coin_ids.append(coin_id)
+            eligibility_and_additions[coin_id] = (is_eligible, spend_additions)
+        removal_names_from_coin_spends: Set[bytes32] = set()
+        bundle_coin_spends: Dict[bytes32, BundleCoinSpend] = {}
+        for coin_spend in new_spend.coin_spends:
+            coin_id = coin_spend.coin.name()
+            removal_names_from_coin_spends.add(coin_id)
+            eligible_for_dedup, spend_additions = eligibility_and_additions.get(coin_id, (False, []))
+            bundle_coin_spends[coin_id] = BundleCoinSpend(coin_spend, eligible_for_dedup, spend_additions)
```
arvidn, line 415: it looks like this new logic, added to `validate_spend_bundle()` does not have a test. unless I'm missing it.

arvidn (commented):
it's really starting to take shape

chia/full_node/mempool_manager.py
AmineKhaldi: This comment existed before and when I peeked at the time I noticed that if we succeed pre-validation, this inconsistency between npc_result spends and spend bundle coin spends wouldn't be possible, so I left the comment and kept this logic.

chia/full_node/mempool.py
AmineKhaldi: Done.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -1113,3 +1153,221 @@ async def test_sufficient_total_fpc_increase() -> None:
...
+        )
+        assert result[1] == MempoolInclusionStatus.SUCCESS
+    assert len(mempool_manager.mempool.get_spends_by_coin_id(coin_id)) == 3
+    assert mempool_manager.mempool.size() == 3
+    assert len(list(mempool_manager.mempool.spends_by_feerate())) == 3
+    # Setup a new peak where the incoming block has spent the coin
+    # Mark this coin as spent
+    test_coin_records = {coin_id: CoinRecord(coin, uint32(0), TEST_HEIGHT, False, uint64(0))}
+    block_record = create_test_block_record(height=new_height)
+    npc_result = NPCResult(
```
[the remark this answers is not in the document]
AmineKhaldi: I don't see a `make_item()` that creates `NPCResult`, but I modified the recently added `make_test_conds()` to use it here and elsewhere where spends need to be honored.

chia/types/eligible_coin_spends.py
```diff
@@ -0,0 +1,102 @@
...
+                        max_cost=item.npc_result.cost,
+                    )
+                    # Update this mempool item's coin spends map
+                    item.bundle_coin_spends[coin_id] = BundleCoinSpend(
+                        spend_data.coin_spend, spend_data.eligible_for_dedup, spend_data.additions, spend_cost
+                    )
+                duplicate_cost = spend_cost
+                # If we end up including this item, update this entry's cost
+                new_eligible_spends[coin_id] = DedupCoinSpend(current_solution, duplicate_cost)
+            cost_saving = uint64(cost_saving + duplicate_cost)
```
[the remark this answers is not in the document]
AmineKhaldi: Done.

chia/full_node/mempool.py
```diff
@@ -312,3 +314,61 @@ def at_full_capacity(self, cost: int) -> bool:
...
+        cost_sum = 0  # Checks that total cost does not exceed block maximum
+        fee_sum = 0  # Checks that total fees don't exceed 64 bits
+        processed_spend_bundles = 0
+        additions: List[Coin] = []
+        # This contains a map of coin ID to a coin spend solution and its isolated cost
+        eligible_coin_spends = EligibleCoinSpends()
+        coin_spends: List[CoinSpend] = []
+        sigs: List[G2Element] = []
+        with self._db_conn:
+            cursor = self._db_conn.execute("SELECT name, fee FROM tx ORDER BY fee_per_cost DESC")
```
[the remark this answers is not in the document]
AmineKhaldi: Done.

chia/full_node/mempool.py
```diff
@@ -328,38 +324,60 @@ def at_full_capacity(self, cost: int) -> bool:
 
         return self.total_mempool_cost() + cost > self.mempool_info.max_size_in_cost
 
+    def get_items_spending_coin_ids(self, spent_coin_ids: List[bytes32]) -> List[MempoolItem]:
+        items: List[MempoolItem] = []
+        for coin_ids in chunks(spent_coin_ids, SQLITE_MAX_VARIABLE_NUMBER):
+            args = ",".join(["?"] * len(coin_ids))
+            with self._db_conn:
```
arvidn, line 331: I was going to comment on how this context manager seems like it belong outside of the loop. but then I started thinking that the only thing (afaik) the context manager does is to commit when exiting. Since all we do here is reading, maybe we don't even need it. Do you know?

chia/full_node/mempool.py
```diff
@@ -328,38 +324,60 @@ def at_full_capacity(self, cost: int) -> bool:
...
-        log.info(f"Starting to make block, max cost: {self.mempool_info.max_block_clvm_cost}")
-        for item in self.spends_by_feerate():
-            if not item_inclusion_filter(item.name):
+        # This contains a map of coin ID to a coin spend solution and its isolated cost
+        eligible_coin_spends = EligibleCoinSpends()
+        coin_spends: List[CoinSpend] = []
+        sigs: List[G2Element] = []
+        with self._db_conn:
+            cursor = self._db_conn.execute("SELECT name, fee FROM tx ORDER BY fee_per_cost DESC, seq ASC")
+        for row in cursor:
```
arvidn, line 351: it looks a bit funky to have this loop outside of the context manager. I'm sure it's correct, but I'm also thinking that when we just read the DB, maybe we don't even need the context manager

chia/full_node/mempool.py
arvidn: I've used the context manager in a bunch of places in here too, where I don't think we need it. Let's punt on changing this until after our mempool PRs land

arvidn (dismissed):

arvidn (dismissed):

chia/full_node/mempool.py
```diff
@@ -390,31 +386,47 @@ def create_bundle_from_mempool_items(
     ) -> Optional[Tuple[SpendBundle, List[Coin]]]:
         cost_sum = 0  # Checks that total cost does not exceed block maximum
         fee_sum = 0  # Checks that total fees don't exceed 64 bits
-        spend_bundles: List[SpendBundle] = []
+        processed_spend_bundles = 0
         additions: List[Coin] = []
+        # This contains a map of coin ID to a coin spend solution and its isolated cost
+        eligible_coin_spends = EligibleCoinSpends()
```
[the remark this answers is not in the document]
AmineKhaldi: I do, I sent a comment documenting the reason.

chia/types/eligible_coin_spends.py
```diff
@@ -0,0 +1,108 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class DedupCoinSpend:
+    solution: SerializedProgram
+    cost: Optional[uint64]
+
+
+@dataclasses.dataclass(frozen=True)
+class EligibleCoinSpends:
+    eligible_spends: Dict[bytes32, DedupCoinSpend] = dataclasses.field(default_factory=dict)
```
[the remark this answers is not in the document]
AmineKhaldi: Indeed.

arvidn (approved):

PR #13802, opened 2022-11-02
Adding new templates

wallentx:
This is to aid in release note generation.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,17 @@
+* **What is the purpose of the changes in this PR?**
+
+
+
+* **What is the current behavior?** (You can also link to an open issue here)
+
+
+
+* **What is the new behavior (if this is a feature change)?**
+
+
+
+* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
+
+
+
+* **Other information**:
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,4 +14,4 @@
 
 
 
-* **Other information**:
+* **Testing notes (is this code covered by tests, or equivalent manual testing?)**```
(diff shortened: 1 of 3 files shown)


.github/workflows/require-labels.yml
```diff
@@ -0,0 +1,25 @@
...
+    outputs:
+      status: ${{ steps.check-labels.outputs.status }}
+    steps:
+      - id: check-labels
+        uses: mheap/github-action-required-labels@v2
+        with:
+          mode: exactly
+          count: 1
+          labels: "Added, Changed, Fixed"
+          exit_type: success
```
cmmarslender, line 21: We'd need to allow the other non-changelog labels too right? (CI, etc)

cmmarslender (changes requested):

.github/workflows/require-labels.yml
wallentx: My thinking was that even release-note-excluded changes can fit into one of these 3 categories.
My 🧠
- Maintain baseline requirements: always
- Add exceptions as an additional layer: can be tweaked later on without disrupting our baseline truth

wallentx: Que 👍 or 👎

cmmarslender: Yeah I guess that makes sense. We can always change it later if we need.

cmmarslender (dismissed):

wallentx:
Some other templates for Pull Requests if anyone has feelings on fine-tuning `.github/PULL_REQUEST_TEMPLATE.md`

https://github.com/axolo-co/pull_request_template
https://github.com/devspace/awesome-github-templates

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+
+* **What is the new behavior (if this is a feature change)?**
+
+
+
+* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
+
+
+
+* **Other information**:
```
paninaro, line 17: Personally I would like to see this swapped with
```suggestion
* **Testing notes (is this code covered by tests, or equivalent manual testing?)**```

paninaro (dismissed):

paninaro (dismissed):

paninaro (approved):

PR #13809, opened 2022-11-03
util: Some cleanups for the memory profiler 

xdustinface:
See commit messages.


chia/util/memory_profiler.py
```diff
@@ -6,8 +6,14 @@
 import tracemalloc
 from dataclasses import dataclass, field
 from datetime import datetime
+from functools import lru_cache
+from subprocess import check_call
+from sys import stdout
 from typing import Dict, List, Optional, Set
 
+import click
+from colorama import Back, Fore, Style, init
```
arvidn: this will add a dependency on `colorama` to the full node. I was trying to avoid that. This would require mentioning `colorama` in `setup.py`, right?

xdustinface: Like here? :) https://github.com/Chia-Network/chia-blockchain/blob/5c38a39077c47f6d5527cacb1a60b68a9931a07d/setup.py#L20?

arvidn: I see

arvidn (approved):
I take it you've manually tested this, right? since we don't have any automated testing

PR #13812, opened 2022-11-03
moved rate limiting log to info from warning

wjblanke:


hugepants:
Setting it to `info` is not helpful for farmers because most will be using monitoring tools like Chiadog which requires `info` logging. So they will still encounter the very noisy rate limit entries. This is why I proposed for it to be set to debug in my PR https://github.com/Chia-Network/chia-blockchain/pull/13758

emlowe:
Yea, I think moving back to debug is the simplest. We might need to level set on the rates though

wjblanke:
ok closing this in favor of https://github.com/Chia-Network/chia-blockchain/pull/13758

PR #13825, opened 2022-11-04
Fix trades between fungible assets <-> fungible assets + nfts

Quexington:
Some bad code means that royalties attempt to get split between 0 NFTs too which result is a division by zero error.


geoffwalmsley (approved):

arvidn (approved):
I take it the test actually fails without the fix in `nft_wallet.py`

PR #13826, opened 2022-11-04
Ensure `chia-blockchain` still opens UI on Linux

### LINKED ISSUE #13847: [Bug]  Absence of "chia-blockchain" binary in  64-bit Linux deb
### What happened?

dpkg -i chia-blockchain_1.6.1_amd64.deb

Gui client is now "/opt/chia/@chiagui" rather than  "chia-blockchain".

Unsure if this is intentional since it's not documented in the  release notes (at least as far as I can see!)

### Version

1.6.1

Linux

GUI

### Relevant log output

_No response_

cmmarslender:
Resolves https://github.com/Chia-Network/chia-blockchain/issues/13847

Tested on amd64/arm64 ubuntu and amd64 fedora (for coverage on all the debs and the rpm) and the symlink for chia-blockchain opens the UI as expected.

### DIFF
--- a/build_scripts/assets/deb/postinst.sh
+++ b/build_scripts/assets/deb/postinst.sh
@@ -4,3 +4,4 @@
 set -e
 
 ln -s /opt/chia/resources/app.asar.unpacked/daemon/chia /usr/bin/chia || true
+ln -s /opt/chia/chia-blockchain /usr/bin/chia-blockchain || true
--- a/build_scripts/assets/deb/prerm.sh
+++ b/build_scripts/assets/deb/prerm.sh
@@ -4,3 +4,4 @@
 set -e
 
 unlink /usr/bin/chia || true
+unlink /usr/bin/chia-blockchain || true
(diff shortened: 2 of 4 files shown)


wallentx (dismissed):

emlowe:
I think we need to look into why the output binary name is `/opt/bin/@chiagui` because this seems like a bug to me.

and generally isn't linking things into `/usr/bin` considered somewhat bad form now?

cmmarslender:
I believe it changed because of the new installer builder.

If we can have it end up named properly by default, I'm also fine with that. /usr/bin vs /usr/local/bin are probably the two options. /usr/bin/ was in PATH already, so I believe thats why we put the `chia` link there initially, so thats why chia-blockchain also got added there. Could do either though

cmmarslender:
1.6.0 (and older, I assume) also had `chia-blockchain` at `/usr/bin/chia-blockchain` so putting it back there makes this more or less consistent with where people have been expecting to find this all along.

emlowe:
Consistency is good - I suspect there is something wrong somewhere with the scripts that results in a name `@chiagui`

wallentx (approved):

emlowe (approved):

PR #13833, opened 2022-11-05
Made it able to RAM plot with bladebit2

ChiaMineJP:
# Changelog

## Removed
- Dropped support for bladebit < 2
- `chia plotters bladebit2` command has been removed and integrated to `chia plotters bladebit`
## Changed
- `chia plotters bladebit` command now requires sub command `ramplot` / `diskplot`.  

```
$ chia plotters bladebit diskplot -h
usage: chia plotters bladebit diskplot [-h] [-r THREADS] [-n COUNT] [-f FARMERKEY] [-p POOL_KEY] [-c CONTRACT] [-i ID] [-w] [--nonuma] [-v] -d FINALDIR
                                       [--no-cpu-affinity] [--cache CACHE] [--f1-threads F1_THREADS] [--fp-threads FP_THREADS] [--c-threads C_THREADS]        
                                       [--p2-threads P2_THREADS] [--p3-threads P3_THREADS] [--alternate] -t TMPDIR [-2 TMPDIR2] [-u BUCKETS] [-m MEMO]        
                                       [--no-t1-direct] [--no-t2-direct]

optional arguments:
  -h, --help            show this help message and exit
  -r THREADS, --threads THREADS
                        Num threads.
  -n COUNT, --count COUNT
                        Number of plots to create (default = 1)
  -f FARMERKEY, --farmerkey FARMERKEY
                        Farmer Public Key (48 bytes)
  -p POOL_KEY, --pool-key POOL_KEY
                        Pool Public Key (48 bytes)
  -c CONTRACT, --contract CONTRACT
                        Pool Contract Address (64 chars)
  -i ID, --id ID        Plot id
  -w, --warmstart       Warm start
  --nonuma              Disable numa
  -v, --verbose         Set verbose
  -d FINALDIR, --final_dir FINALDIR
                        Final directory.
  --no-cpu-affinity     Disable assigning automatic thread affinity
  --cache CACHE         Size of cache to reserve for I/O
  --f1-threads F1_THREADS
                        Override the thread count for F1 generation
  --fp-threads FP_THREADS
                        Override the thread count for forward propagation
  --c-threads C_THREADS
                        Override the thread count for C table processing
  --p2-threads P2_THREADS
                        Override the thread count for Phase 2
  --p3-threads P3_THREADS
                        Override the thread count for Phase 3
  --alternate           Halves the temp2 cache size requirements by alternating bucket writing methods between tables
  -t TMPDIR, --tmp_dir TMPDIR
                        Temporary directory 1.
  -2 TMPDIR2, --tmp_dir2 TMPDIR2
                        Temporary directory 2.
  -u BUCKETS, --buckets BUCKETS
                        Number of buckets.
  -m MEMO, --memo MEMO  Memo variable.
  --no-t1-direct        Disable direct I/O on the temp 1 directory
  --no-t2-direct        Disable direct I/O on the temp 2 directory
```

```
$ chia plotters bladebit ramplot -h 
usage: chia plotters bladebit ramplot [-h] [-r THREADS] [-n COUNT] [-f FARMERKEY] [-p POOL_KEY] [-c CONTRACT] [-i ID] [-w] [--nonuma] [--no-cpu-affinity]
                                      [-v] -d FINALDIR

optional arguments:
  -h, --help            show this help message and exit
  -r THREADS, --threads THREADS
                        Num threads.
  -n COUNT, --count COUNT
                        Number of plots to create (default = 1)
  -f FARMERKEY, --farmerkey FARMERKEY
                        Farmer Public Key (48 bytes)
  -p POOL_KEY, --pool-key POOL_KEY
                        Pool Public Key (48 bytes)
  -c CONTRACT, --contract CONTRACT
                        Pool Contract Address (64 chars)
  -i ID, --id ID        Plot id
  -w, --warmstart       Warm start
  --nonuma              Disable numa
  --no-cpu-affinity     Disable assigning automatic thread affinity
  -v, --verbose         Set verbose
  -d FINALDIR, --final_dir FINALDIR
                        Final directory.
```

### DIFF
--- a/chia/plotters/bladebit.py
+++ b/chia/plotters/bladebit.py
@@ -153,7 +153,7 @@ def get_bladebit_install_info(plotters_root_path: Path) -> Optional[Dict[str, An
     return info
 
 
-progress_bladebit1 = {
+progress_bladebit_ram = {
     "Finished F1 sort": 0.01,
     "Finished forward propagating table 2": 0.06,
     "Finished forward propagating table 3": 0.12,
@@ -175,7 +175,7 @@ progress_bladebit1 = {
 }
 
 
-progress_bladebit2 = {
+progress_bladebit_disk = {
     # "Running Phase 1": 0.01,
     "Finished f1 generation in ": 0.01,
     "Completed table 2 in ": 0.06,
@@ -210,23 +210,11 @@ def plot_bladebit(args, chia_root_path, root_path):
         print(f"Error: {version_or_exception}")
         return
 
-    version = None
-    actual_version = None
-    if args.plotter == "bladebit":
-        version = actual_version = 1
-        if found and version_or_exception[0] != "1":
-            actual_version = 2
-    elif args.plotter == "bladebit2":
-        version = actual_version = 2
-        if found and version_or_exception[0] != "2":
-            print(
-                f"You're trying to run bladebit version 2"
-                f" but currently version {'.'.join(version_or_exception)} is installed"
-            )
-            return
-
-    if version is None:
-        print(f"Unknown version of bladebit: {args.plotter}")
+    if found and int(version_or_exception[0]) < 2:
+        print(
+            f"Version {'.'.join(version_or_exception)} is detected."
+            f"bladebit < 2 is not supported any more."
+        )
         return
 
     bladebit_executable_path = get_bladebit_executable_path(root_path)
@@ -248,6 +236,7 @@ def plot_bladebit(args, chia_root_path, root_path):
             args.connect_to_daemon,
         )
     )
+    plot_type = "ramplot" if args.plot_type == "ramplot" else "diskplot"
     call_args = [
         os.fspath(bladebit_executable_path),
         "-t",
@@ -275,10 +264,9 @@ def plot_bladebit(args, chia_root_path, root_path):
     if "memo" in args and args.memo is not None and args.memo != b"":
         call_args.append("--memo")
         call_args.append(args.memo)
-    if version > 1:
-        call_args.append("diskplot")
-    elif version == 1 and actual_version > 1:
-        call_args.append("ramplot")
+
+    call_args.append(plot_type)
+
     if "buckets" in args and args.buckets:
         call_args.append("-b")
         call_args.append(str(args.buckets))
@@ -318,7 +306,7 @@ def plot_bladebit(args, chia_root_path, root_path):
     call_args.append(args.finaldir)
 
     try:
-        progress = progress_bladebit1 if version == 1 else progress_bladebit2
+        progress = progress_bladebit_ram if plot_type == "ramplot" else progress_bladebit_disk
         asyncio.run(run_plotter(chia_root_path, args.plotter, call_args, progress))
     except Exception as e:
         print(f"Exception while plotting: {e} {type(e)}")
--- a/chia/plotters/bladebit.py
+++ b/chia/plotters/bladebit.py
@@ -211,10 +211,8 @@ def plot_bladebit(args, chia_root_path, root_path):
         return
 
     if found and int(version_or_exception[0]) < 2:
-        print(
-            f"Version {'.'.join(version_or_exception)} is detected."
-            f"bladebit < 2 is not supported any more."
-        )
+        print(f"Version {'.'.join(version_or_exception)} is detected.")
+        print(f"bladebit < 2 is not supported any more.")
         return
 
     bladebit_executable_path = get_bladebit_executable_path(root_path)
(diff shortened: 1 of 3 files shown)


ChiaMineJP:
Relevant PR: https://github.com/Chia-Network/chia-blockchain-gui/pull/1160

chia/plotters/bladebit.py
```diff
@@ -211,16 +211,13 @@ def plot_bladebit(args, chia_root_path, root_path):
         return
 
     version = None
+    actual_version = None
```
paninaro, line 214: While I can follow the logic here to see how disk or RAM plotting is determined, it does feel a bit odd.

Just to explore an alternative, if we had an enum that offered `DISK` or `RAM` plotting options:

```
class BBPlottingType(Enum):
  RAM = 'ramplot' # Supported by BB1 and BB2
  DISK = 'diskplot' # BB2 only
```

Then I think you could do something like:

```
plotting_type = BBPlottingType.RAM # Assume RAM plotting by default

if args.plotter == "bladebit2":
  plotting_type = BBPlottingType.DISK

...

call_args.append(plotting_type.value)
```

Would that work?

ChiaMineJP: Actually, command syntax of bladebit1 and bladebit2 with ramplot option are a bit different.

### Bladebit 1 (Always RAM plot)
```
bladebit <general options>
```

### Bladebit 2 with RAM plot option
```
bladebit <general options> ramplot [additional options]
```

If you want to do ramplot with bladebit2, you need to specify `ramplot` flag while bladebit1 doesn't recognize it.

And to make the code simple, when you want to do ramplot with bladebit2, GUI posts request to chia daemon as if there is bladebit1 installed, because whole option syntax is the same with bladebit1 except for `ramplot` flag.

Maybe it makes easy to understand how GUI recognize ramplot option when bladebit2 is installed.
image

chia/plotters/bladebit.py
```diff
@@ -275,45 +272,47 @@ def plot_bladebit(args, chia_root_path, root_path):
         call_args.append("-v")
     if args.nonuma:
         call_args.append("-m")
-    if args.memo is not None and args.memo != b"":
+    if "memo" in args and args.memo is not None and args.memo != b"":
```
altendky, line 275: What is making the parameters maybe not present now?

ChiaMineJP: `args` here is `arpparse`'s `Namespace` object and it raises error if specified property does not exist.
`memo` is an option of bladebit2 and bladebit1 does not expect the option.

https://github.com/Chia-Network/chia-blockchain/blob/cmj.bb2-ramplot/chia/plotters/plotters.py#L91-L131
Check these lines for your convinience.
`--memo` is an option for bladebit2 but not for bladebit1.

altendky: I would expect we won't be developing bladebit1 anymore so having the code support it just stay and remain unchanged seems ok.  New bladebit2 code can handle just bladebit2 without having to worry about backwards compatibility with 1.  Also, checking this way cascades from version selection to arg definition to arg usage rather than just using the single layer of version selection as the condition everywhere.  But, I admittedly have my whole own other way that I've handled plotters in plotman so... `:]`

I won't block merging on this topic, feel free to close this conversation if you would rather keep this all as is.

ChiaMineJP:
As we discussed, I'll drop support for old bladebit1. I'll add commits for it.

chia/plotters/bladebit.py
ChiaMineJP: Thanks for your comment Kyle.
I was thinking whether I should create another PR to drop bladebit1 support but I decide to manage that in this PR.

emlowe:
Just a note that typically "Deprecated" means that something still works, but will stop working in a future release. However, in this case, this not working at all - so it's just outright removed so it's technically a breaking change
I'm ok with this breaking change without some period of deprecation mostly because BB1 was a pretty small set of users and I'm not bothered too much by it.

paninaro (dismissed):
lgtm

ChiaMineJP:
> Just a note that typically "Deprecated" means that something still works, but will stop working in a future release. However, in this case, this not working at all - so it's just outright removed so it's technically a breaking change I'm ok with this breaking change without some period of deprecation mostly because BB1 was a pretty small set of users and I'm not bothered too much by it.

Thanks Earle! I fixed the words.

ChiaMineJP:
@paninaro I rebased to `main` for some failed CI jobs. Please re-approve this PR.

paninaro (approved):

PR #13855, opened 2022-11-07
Start subscription with a snapshot

Chida82:
Let's consider DataStores that are changed often and the keys overwritten often. An application that connects via RPC with Chia (a project of mine) must download these DataStores, but starting from item 1 the time taken to download all versions is very long. Furthermore, in "ecological" terms, unloading more than necessary has an impact.
This development allows you to download the latest version at the first request. This way you can build a dedicated server for this. I developed in server terms with a backward compatible explicit key so that it is explicit that the files located "\data_layer\db\server_files_location_mainnet" are not useful for mirroring. Its best use and with a docker to have a fast DataStore reader.
in other words, "WE TREAT THE LAST FULL FILE AS A SNAPSHOT"

### DIFF
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -31,7 +31,7 @@ from chia.data_layer.data_layer_util import (
 )
 from chia.data_layer.data_layer_wallet import DataLayerWallet, Mirror, SingletonRecord, verify_offer
 from chia.data_layer.data_store import DataStore
-from chia.data_layer.download_data import insert_from_delta_file, write_files_for_root
+from chia.data_layer.download_data import insert_from_delta_file, insert_from_full_file, write_files_for_root
 from chia.rpc.rpc_server import default_get_connections
 from chia.rpc.wallet_rpc_client import WalletRpcClient
 from chia.server.outbound_message import NodeType
@@ -375,16 +375,28 @@ class DataLayer:
 
             try:
                 timeout = self.config.get("client_timeout", 15)
-                success = await insert_from_delta_file(
-                    self.data_store,
-                    tree_id,
-                    root.generation,
-                    [record.root for record in reversed(to_download)],
-                    server_info,
-                    self.server_files_location,
-                    timeout,
-                    self.log,
-                )
+                if self.config.get("use_last_full_tree_file", False) and root.generation == 0:
+                    success = await insert_from_full_file(
+                        self.data_store,
+                        tree_id,
+                        singleton_record.generation,
+                        to_download[0].root,
+                        server_info,
+                        self.server_files_location,
+                        timeout,
+                        self.log,
+                    )
+                else:
+                    success = await insert_from_delta_file(
+                        self.data_store,
+                        tree_id,
+                        root.generation,
+                        [record.root for record in reversed(to_download)],
+                        server_info,
+                        self.server_files_location,
+                        timeout,
+                        self.log,
+                    )
                 if success:
                     self.log.info(
                         f"Finished downloading and validating {tree_id}. "
@@ -533,7 +545,7 @@ class DataLayer:
                     try:
                         await self.update_subscriptions_from_wallet(subscription.tree_id)
                         await self.fetch_and_validate(subscription.tree_id)
-                        await self.upload_files(subscription.tree_id)
+# await self.upload_files(subscription.tree_id)
                     except asyncio.CancelledError:
                         raise
                     except Exception as e:
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -375,7 +375,7 @@ class DataLayer:
 
             try:
                 timeout = self.config.get("client_timeout", 15)
-                if self.config.get("use_last_full_tree_file", False) and root.generation == 0:
+                if self.config.get("use_snapshot", False) and root.generation == 0:
                     success = await insert_from_full_file(
                         self.data_store,
                         tree_id,
--- a/chia/data_layer/download_data.py
+++ b/chia/data_layer/download_data.py
@@ -55,7 +55,7 @@ async def insert_into_data_store_from_file(
     tree_id: bytes32,
     root_hash: Optional[bytes32],
     filename: Path,
-    generation: Optional[int] = None,
+    snapshot_generation: Optional[int] = None,
 ) -> None:
     with open(filename, "rb") as reader:
         while True:
@@ -82,9 +82,13 @@ async def insert_into_data_store_from_file(
             serialized_node = SerializedNode.from_bytes(serialize_nodes_bytes)
 
             node_type = NodeType.TERMINAL if serialized_node.is_terminal else NodeType.INTERNAL
-            await data_store.insert_node(node_type, serialized_node.value1, serialized_node.value2)
 
-    await data_store.insert_root_with_ancestor_table(tree_id, root_hash, Status.COMMITTED, generation)
+            print(f"insert node. Serialized:{serialized_node}")
+            await data_store.insert_node(node_type, serialized_node.value1, serialized_node.value2)
+            print("post insert node.")
+    print("Pre Ancestor")
+    await data_store.insert_root_with_ancestor_table(tree_id, root_hash, Status.COMMITTED, snapshot_generation)
+    print("Post Ancestor")
 
 
 async def write_files_for_root(
--- a/chia/data_layer/download_data.py
+++ b/chia/data_layer/download_data.py
@@ -55,6 +55,7 @@ async def insert_into_data_store_from_file(
     tree_id: bytes32,
     root_hash: Optional[bytes32],
     filename: Path,
+    generation: Optional[int] = None,
 ) -> None:
     with open(filename, "rb") as reader:
         while True:
@@ -83,7 +84,7 @@ async def insert_into_data_store_from_file(
             node_type = NodeType.TERMINAL if serialized_node.is_terminal else NodeType.INTERNAL
             await data_store.insert_node(node_type, serialized_node.value1, serialized_node.value2)
 
-    await data_store.insert_root_with_ancestor_table(tree_id=tree_id, node_hash=root_hash, status=Status.COMMITTED)
+    await data_store.insert_root_with_ancestor_table(tree_id, root_hash, Status.COMMITTED, generation)
 
 
 async def write_files_for_root(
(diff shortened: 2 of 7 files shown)


chia/data_layer/data_layer.py
```diff
@@ -533,7 +545,7 @@ async def periodically_manage_data(self) -> None:
                     try:
                         await self.update_subscriptions_from_wallet(subscription.tree_id)
                         await self.fetch_and_validate(subscription.tree_id)
-                        await self.upload_files(subscription.tree_id)
+# await self.upload_files(subscription.tree_id)
```
Chida82, line 548: This now is for test. If works, is necessary to change this to not create files prior to the snapshot

chia/data_layer/download_data.py
```diff
@@ -81,9 +82,13 @@ async def insert_into_data_store_from_file(
             serialized_node = SerializedNode.from_bytes(serialize_nodes_bytes)
 
             node_type = NodeType.TERMINAL if serialized_node.is_terminal else NodeType.INTERNAL
-            await data_store.insert_node(node_type, serialized_node.value1, serialized_node.value2)
 
-    await data_store.insert_root_with_ancestor_table(tree_id=tree_id, node_hash=root_hash, status=Status.COMMITTED)
+            print(f"insert node. Serialized:{serialized_node}")
```
Chida82, line 86: printf is only for resolve the issue of test. when resolve is possible to remove

tests/core/data_layer/test_data_store.py
```diff
@@ -1226,6 +1226,87 @@ async def test_data_server_files(data_store: DataStore, tree_id: bytes32, test_d
...
+    data_store: DataStore, tree_id: bytes32, test_delta: bool, tmp_path: Path
+) -> None:
+    roots: List[Root] = []
+    num_batches = 10
+    num_ops_per_batch = 100
+    generation_of_snapshot = 5
+    key_check_all_generation = hexstr_to_bytes("746573745F636865636B")
+    # key_check_after_snapshot = hexstr_to_bytes("746573745F6265666F7265")
+    db_path = tmp_path.joinpath("dl_server_util.sqlite")
+    print('Test start')
```
Chida82, line 1244: only for investigation, when resolve remove printf

tests/core/data_layer/test_data_store.py
```diff
@@ -1226,6 +1226,87 @@ async def test_data_server_files(data_store: DataStore, tree_id: bytes32, test_d
...
+                    changelist.append({"action": "insert", "key": key, "value": value})
+                else:
+                    key = random.choice(keys)
+                    keys.remove(key)
+                    changelist.append({"action": "delete", "key": key})
+                counter += 1
+            value = countergeneration.to_bytes(4, byteorder="big")
+            changelist.append({"action": "delete", "key": key_check_all_generation})
+            changelist.append({"action": "insert", "key": key_check_all_generation, "value": value})
+            # if countergeneration == 3:
```
Chida82, line 1271: for try to reproduce problem of delta

Chida82 (commented):
at the moment there is an error in test if after snapshot it use delta file.

Chida82:
@emlowe  i need help for these. the test works if test_delta is False, if use only full file after snapshot.
if we use delta_file after snapshot, after a full file, there is "FOREIGN KEY constraint failed". 
It seems that a full_file in some cases does not have all the "INTERNAL" records and when use a delta_file it try to add a "INTERNAL" node but without having the appropriate hashed record. 
Can someone help me understand how the "internal" works and how I can solve this problem

Chida82:
TEST Error Output
```
================================== FAILURES ===================================
_________________ test_data_server_files_from_snapshot[True] __________________
[gw0] win32 -- Python 3.8.10 c:\github\Chida82\chia-blockchain\venv\Scripts\python.exe
tests\core\data_layer\test_data_store.py:1299: in test_data_server_files_from_snapshot
    await insert_into_data_store_from_file(data_store, tree_id, root.node_hash, tmp_path.joinpath(filename))
chia\data_layer\download_data.py:87: in insert_into_data_store_from_file
    await data_store.insert_node(node_type, serialized_node.value1, serialized_node.value2)
chia\data_layer\data_store.py:254: in insert_node
    await self._insert_node(node_hash, node_type, bytes32(value1), bytes32(value2), None, None)
chia\data_layer\data_store.py:235: in _insert_node
    await writer.execute(
venv\lib\site-packages\aiosqlite\core.py:184: in execute
    cursor = await self._execute(self._conn.execute, sql, parameters)
venv\lib\site-packages\aiosqlite\core.py:129: in _execute
    return await future
venv\lib\site-packages\aiosqlite\core.py:102: in run
    result = function()
E   sqlite3.IntegrityError: FOREIGN KEY constraint failed
---------------------------- Captured stdout call -----------------------------
Test start
insert node. Serialized:{'is_terminal': True, 'value1': '0x00000107', 'value2': '0x0000020e'}
post insert node.
insert node. Serialized:{'is_terminal': True, 'value1': '0x000001ca', 'value2': '0x00000394'}
post insert node.
insert node. Serialized:{'is_terminal': False,
 'value1': '0x7a68293d08b6d92d677a7d2108de095ad14b8f0d53faf4ce81ba805dec555a00',
 'value2': '0x9300f083e11c9dbb6582719d479571c17b41c13ddcd21986765c564689aad140'}
post insert node.
insert node. Serialized:{'is_terminal': True, 'value1': '0x00000155', 'value2': '0x000002aa'}
post insert node.
insert node. Serialized:{'is_terminal': True, 'value1': '0x000001d4', 'value2': '0x000003a8'}


.......................

insert node. Serialized:{'is_terminal': False,
 'value1': '0x7c456afd63691f224c1542206a1d978f06ccefebf7cc37fc1a2543e8e8d5f5e9',
 'value2': '0xd94494c180f65feecbf5edb9398931d7c59805a8234edc589334ec30bdc226d8'}
post insert node.
insert node. Serialized:{'is_terminal': False,
 'value1': '0x383052304608a9cdf48cdb21c687f15c27063310db4150fbda9eae5181c0553e',
 'value2': '0x46102347dcb0ea6b63bdd01884d3f0ca3f43b7a13d9878869422c883f3444287'}
- generated xml file: C:\Users\dchini\AppData\Local\Temp\tmp-10288j3cnIzc30B0A.xml -
======================== 1 failed in 281.25s (0:04:41) ========================
```

the last serialize node generates the error.
[https://github.com/Chida82/chia-blockchain/tree/dl_fetch_from_last_full_file_filetest/testDelta](https://github.com/Chida82/chia-blockchain/tree/dl_fetch_from_last_full_file_filetest/testDelta)
here are all the log data, the starting data and the result before the error

fchirica:
Delta files skip already seen `(key, value)` pairs for terminal nodes, or `(left, right)` pairs for internal nodes, anywhere up to that point. So if any of those pairs appeared before your snapshot, but are not present in the snapshot, they will not be present in any of the next delta files - and anything referencing them will throw. Since full files contain everything, the same pairs will appear there (and be referenced correctly). Are there any pairs appearing before the snapshot, not being present in the snapshot, then being referenced in generations after the snapshot?

fchirica:
Another possibility is to have a generation mismatch: i.e. if `snapshot_generation=5`, all delta files used must have generation > 5 from that point on (given the condition I've written above) - we must insert first generation 6, generation 7, ... all the way up to singleton's generation, without any gaps. We can't use any generation lower than 5 after the snapshot moved us to generation=5 - those record will miss. Also not inserting contiguous generations after the snapshot will produce missing records.

Chida82:
@fchirica As you say, the "full" file should have it all. Confirm?
The test shows that if you create a full database with generation 5 and then apply a delta 6 to it, this doesn't work, it lacks references.
The only thing that comes to mind is that there is a bug in the creation of the full and that has not emerged because there is no method that loads starting from the full.
Here are the evidences, that is the database resulting from the full v5 + delta v6 up to the error
https://github.com/Chida82/chia-blockchain/tree/dl_fetch_from_last_full_file_filetest/testDelta
Do you need additional logs? can I facilitate the investigation?

fchirica:
Alright, so we're not in the case #2, that's good. Please refer to the first comment. Are all keys and values unique? Say `key=1 value=1` appeared in `generation=1`, but it's not present in `generation=5`. Then, somewhere say in `generation=11` the same pair `key=1 value=1` appears again. This will be skipped in delta file corresponding to `generation=11` and we'll not know it since it wasn't present in `generation=5` snapshot. Is it possible we run into a case like this?

Chida82:
@fchirica  I have also implemented the case you suggested and we are not in that.
see latest commit.
We are also not even in the case where you generate the key and at each Generation, you delete and recreate it.
the tests pass if you set "num_ops_per_batch" = 0.
With the random entropy of 100, the problem arises.
the original test "test_data_server_files" does not highlight the problem because it does not start from a full and then applies a delta as in my case.

Please @fchirica  open this Sqlite which is random generation 10 , as it comes out of the "test_data_server_files" test. 
https://github.com/Chida82/chia-blockchain/blob/dl_fetch_from_last_full_file_filetest/testDelta/source/dl_server_util.sqlite
and search the internal node gneration 6
insert node. Serialized:{'is_terminal': False,
 'value1': '0x383052304608a9cdf48cdb21c687f15c27063310db4150fbda9eae5181c0553e',
 'value2': '0x46102347dcb0ea6b63bdd01884d3f0ca3f43b7a13d9878869422c883f3444287'}
 this node i htink is related to something in 3 and 4 generation.
 after that, open this:
https://github.com/Chida82/chia-blockchain/blob/dl_fetch_from_last_full_file_filetest/testDelta/result.sqlite
 It's the sql of a full 5 generation ( same file of other db ) and find somethink correlated, please. There isn't the 3 and 4 internal generation, notthing correlate to this node. 
 WHY?

fchirica:
I ran your test, it seems `383052304608a9cdf48cdb21c687f15c27063310db4150fbda9eae5181c0553e` appears in generation 3, in generation 4, misses in generation 5 and appears again in generation 6. the only delta it'll be written to is the one for generation 3 - so there's no way to get this node, since it will miss from generation 6 delta and from generation 5 full tree.

Chida82:
@fchirica My conclusion is this too, but I don't understand why it's how to reproduce it.  because if I apply in delta 6 and not present in full 5 it is right that it is a mistake.  If it is not a mistake, how to reproduce it on time?  do I generate a key to generation 3, then delete to generation 4 and then insert to generation 6?  because I've tried and can't play it.  it emerges only in the chaos of random.

fchirica:
Imagine having the tree a -> b and a -> c in generation 4 (hash a). now in generation 5 we get a tree a' -> b', a' -> c, b' -> b, b' -> d (basically inserting leaf d, changing the hash to a'). in generation 6 we delete key d. the tree gets back to a -> b, a -> c (and hash reverses to a from a'). note how hash a never appeared in generation 5 (since the hash was a' for the subtree), but it won't appear in generation 6 delta file either (because we've seen it in generation 4). this is a possible example how a hash can miss a generation, but reappear later and get skipped in the delta file.

Chida82:
@fchirica 
1) i can do a test specific, but can gli give me the order of bulk ( insert and delete) for generation?
for example just do something and testing:
```
            #     changelist.append({"action": "insert", "key": key_check_after_snapshot, "value": value})
            # if countergeneration == 4:
            #     changelist.append({"action": "delete", "key": key_check_after_snapshot})
            #     changelist.append({"action": "insert", "key": key_check_after_snapshot, "value": value})
            # if countergeneration == 6:
            #     changelist.append({"action": "delete", "key": key_check_after_snapshot})
            #     changelist.append({"action": "insert", "key": key_check_after_snapshot, "value": value})
```
2) Why does the "full" not have the previous information? the "full" should have everything and all references, or I am wrong assumptions.

Chida82:
@fchirica  I apologize in advance for my English and if I seem impolite.
I understand that a "Full" file may not have all the information and subsequent deltas are not usable. Now I wonder, what's the point of having a "Full" file if we then have to use the sequence of "Deltas" from the beginning?

fchirica:
the full file contains everything - but only related to the tree. If say pair (1, 2) was at some point in the tree, but it's currently deleted, we won't have any reference to it in the full tree - unless it reappears in a future generation, then the full tree file will contain it again, but not the delta file.

Chida82:
@fchirica if we use only full tree files, it works.
But it is not the final goal, the "full" are more weights and we would consume more resources using only the "full".
Using only "delta" is also not as green as downloading the latest "snapshot" and then using the deltas.  Moreover you have the benefit of immediately having consistent and updated data.

emlowe:
Closing old unfinished PR

Issue #13741, opened 2022-10-28
[Bug] Cannot open at least XXX files, please raise maximum open file limit in OS.


yllierop:
### What happened?

When trying to use: `./chia plotters madmax -c....` I'm seeing the following error message: ` Cannot open at least XXX files, please raise maximum open file limit in OS.`

I've adjusted the `ulimit` from the defaults. This issue doesn't occur on the directly compiled version of the `Mad Max` plotter.

### Version

Chia 1.6.0

### What platform are you using?

macOS

### What ui mode are you using?

CLI

### Relevant log output

```shell
Cannot open at least XXX files, please raise maximum open file limit in OS.
```

madMAx43v3r:
How many buckets are you using? Try a lower number.

yllierop:
`-u 512` and `-v 256` the issue is it works fine in the directly downloaded version from the @madMAx43v3r repo just not in the Chia app wrapped version.

madMAx43v3r:
Try 128 buckets for each

yllierop:
`./chia plotters madmax -c` works fine with `128` buckets each. But, I guess what I'm wondering is... why does it behave differently when invoked from a build of: `https://github.com/madMAx43v3r/chia-plotter`

yllierop:
I had originally thought the issue was the same as what is described in:  https://github.com/madMAx43v3r/chia-plotter/issues/385

yllierop:
I think that I might have found the issue with `setrlimit` being hard-coded versus, using the systems `ulimit` settings on the following line of code available at: https://github.com/Chia-Network/chia-blockchain/commit/f550555c7d24202d7a98b435fe59560c5b61104f#diff-8acf0d600784f10ea705fc00078cdd0bc807d5805348e03e6f5cd303a7159898R172

```
  # madMAx has a ulimit -n requirement > 296:
        # "Cannot open at least 296 files, please raise maximum open file limit in OS."
        resource.setrlimit(resource.RLIMIT_NOFILE, (512, 512))
```

yllierop:
@MumfMeisterT — I don't think that this is a `MacOS` only issue and it also effects the `GUI` as well.

wallentx:
@yllierop Did you modify that line on your local system, and did it resolve your situation?

yllierop:
@wallentx — yes, seems to do the trick. But, I think a better solution might be to check the current `ulimit` setting before blindly overwriting it.

[...]

PR #13856, opened 2022-11-07
upped the number of file handles for madmax

wjblanke:
this won't affect the hard limit but only soft limits set previously

### DIFF
--- a/chia/plotters/madmax.py
+++ b/chia/plotters/madmax.py
@@ -160,7 +160,9 @@ def plot_madmax(args, chia_root_path: Path, plotters_root_path: Path):
 
         # madMAx has a ulimit -n requirement > 296:
         # "Cannot open at least 296 files, please raise maximum open file limit in OS."
-        resource.setrlimit(resource.RLIMIT_NOFILE, (1024, 1024))
+        soft_limit, hard_limit = resource.getrlimit(resource.RLIMIT_NOFILE)
+        # Set soft limit to max (hard limit)
+        resource.setrlimit(resource.RLIMIT_NOFILE, (hard_limit, hard_limit))
     else:
         reset_loop_policy_for_windows()
 
--- a/chia/plotters/madmax.py
+++ b/chia/plotters/madmax.py
@@ -160,7 +160,7 @@ def plot_madmax(args, chia_root_path: Path, plotters_root_path: Path):
 
         # madMAx has a ulimit -n requirement > 296:
         # "Cannot open at least 296 files, please raise maximum open file limit in OS."
-        resource.setrlimit(resource.RLIMIT_NOFILE, (512, 512))
+        resource.setrlimit(resource.RLIMIT_NOFILE, (1024, 1024))
     else:
         reset_loop_policy_for_windows()
 


paninaro (dismissed):

wallentx:
There appears to be some legit CI failures

arvidn:
doesn't `setrlimit()` be protected with a `try`-`catch` block in case 1024 exceeds the hard limit for the process?

wallentx:
> doesn't `setrlimit()` be protected with a `try`-`catch` block in case 1024 exceeds the hard limit for the process?

@wjblanke consideration before I merge?

wjblanke:
Let me make this better

wallentx (dismissed):
marking for changes so I don't accidentally merge

wjblanke:
I changed the code so that it only sets the soft limit up to the hard limit amount. I am still confused why this line is needed at all, unless python or the OS soft limits us earlier on. Jeff said it did avoid an issue when this was added, but I am definitely not raising the hard limit now.

paninaro:
> I changed the code so that it only sets the soft limit up to the hard limit amount. I am still confused why this line is needed at all, unless python or the OS soft limits us earlier on. Jeff said it did avoid an issue when this was added, but I am definitely not raising the hard limit now.

I've tested the change, and it does work... But it also works for me without calling setrlimit at all. If I run madmax directly from the CLI, I get the following:

```
./venv/bin/chia_plot -k 25 -n 1 -t ~/tmp/ -2 ~/tmp2/ -d ~/final/ -f 936bb15faeb9f7d30fdd3598eacd90977050b5bb2e3cacb75250e17ad76c5ae4062e486576de68781ef4a457f10f8a52 -p 8b1a2f76aefa27921bd9574d75f26fe3a3bef8f0f897930ef872820a8ecd356e0608714aeb0830da826ecaa3b661735d
Cannot open at least 296 files, please raise maximum open file limit in OS.
```

paninaro (approved):
Approving as there doesn't appear to be a downside, but as I mentioned, I'm able to run madmax without the call to setrlimit at all. Perhaps the behavior was different with an older version of python or in Catalina.

wallentx:
https://stackoverflow.com/a/16529317

arvidn (approved):

arvidn:
the default soft limit in MacOS is 256, so I would expect us to have to raise this limit

wallentx:
I wrote a python script to do some testing to help figure out what's happening here. I named it `chia_plot` and put it in my `venv/bin/` so I could call it in place of madmax.
Here is the output when running it just on its own:
```
:~/chia-blockchain/../bin > ./chia_plot                                                 ## This prints the inherited resource limits,                                           ## then sets some specific values,                                                      ## then prints the results after.                                                                                                                                               getrlimit before: (1024, 524288)                                                                                                                                                ## Running 'whoami; ulimit -n' in a subprocess with 'shell=True'                        subprocess: william                                                                     1024                                                                                                                                                                            ## The command 'ulimit' is a shell builtin,                                             ## and can't be called as a command in a subprocess with 'shell=False',                 ## but a script that executes it can.                                                   subprocess 'whoami' command:                                                             
 william                                                                                                                                                                        subprocess bash script:                                                                  
 ulimit -n: 1024                                                                        
 ulimit -Sn: 1024                                                                        
 ulimit -Hn: 524288                                                                                                                                                              subprocess python script:                                                                
 (1024, 524288)                                                                                                                                                                 ## Back to checking within this script.
getrlimit: (1024, 524288)                                                                                                                                                       ## Now I'm going to execute:
## resource.setrlimit(resource.RLIMIT_NOFILE, (524288, 524288))
## and I'll re-run all of the things I just ran.                                        
getrlimit after: (524288, 524288)
                                                                                        ## Running 'whoami; ulimit -n' in a subprocess with 'shell=True'                        
 subprocess: william
 524288
                                                                                                                                                                                ## Running scripts again
 subprocess 'whoami' command:                                                             
 william
                                                                                        subprocess bash script:                                                                  
 ulimit -n: 524288
 ulimit -Sn: 524288                                                                      
 ulimit -Hn: 524288                                                                                                                                                              subprocess python script:                                                                
 (524288, 524288)                                                                       
                                                                                        ## Back to checking within this script.                                                 getrlimit: (524288, 524288)
## setting: resource.setrlimit(resource.RLIMIT_NOFILE, (131072, 131072))
getrlimit after: (131072, 131072)
```

I then called it from the chia cli while on `main`:
```
:~/chia-blockchain/../bin > chia plotters madmax -t /mnt/ssdarray1/ -d /mnt/ssdarray1/  ## This prints the inherited resource limits,                                           ## then sets some specific values,
## then prints the results after.
                                                                                        getrlimit before: (512, 512)                                                                                                                                                    ## Running 'whoami; ulimit -n' in a subprocess with 'shell=True'                        subprocess: william                                                                     512                                                                                                                                                                             ## The command 'ulimit' is a shell builtin,                                             ## and can't be called as a command in a subprocess with 'shell=False',                 ## but a script that executes it can.                                                   subprocess 'whoami' command:
 william                                                                                                                                                                        subprocess bash script:
 ulimit -n: 512
 ulimit -Sn: 512                                                                         
 ulimit -Hn: 512

subprocess python script:                                                                (512, 512)                                                                             
## Back to checking within this script.
getrlimit: (512, 512)                                                                                                                                                           ## Now I'm going to execute:                                                            ## resource.setrlimit(resource.RLIMIT_NOFILE, (524288, 524288))                         ## and I'll re-run all of the things I just ran.                                        STDERR: Traceback (most recent call last):                                                                                                                                      STDERR:   File "/home/william/chia-blockchain/venv/bin/chia_plot", line 53, in <module>                                                                                         STDERR:     resource.setrlimit(resource.RLIMIT_NOFILE, (524288, 524288))                                                                                                        STDERR: ValueError: not allowed to raise maximum limit
```

Finally, I checked out this PR and ran it once more:
```
:~/chia-blockchain/../bin > gh pr checkout 13856                                        M       chia-blockchain-gui                                                             branch 'ulimitwjb' set up to track 'origin/ulimitwjb'.
Switched to a new branch 'ulimitwjb'.                                                                :~/chia-blockchain/../bin > chia plotters madmax -t /mnt/ssdarray1/ -d /mnt/ssdarray1/
## This prints the inherited resource limits,                                           ## then sets some specific values,
## then prints the results after.                                                                                                                                               getrlimit before: (524288, 524288)
                                                                                        ## Running 'whoami; ulimit -n' in a subprocess with 'shell=True'                        subprocess: william                                                                     524288                                                                                                                                                                          ## The command 'ulimit' is a shell builtin,
## and can't be called as a command in a subprocess with 'shell=False',
## but a script that executes it can.                                                   subprocess 'whoami' command:
 william
                                                                                        subprocess bash script:
 ulimit -n: 524288
 ulimit -Sn: 524288                                                                      
 ulimit -Hn: 524288                                                                      
subprocess python script:
 (524288, 524288)                                                                       
## Back to checking within this script.
getrlimit: (524288, 524288)                                                             
## Now I'm going to execute:                                                            ## resource.setrlimit(resource.RLIMIT_NOFILE, (524288, 524288))                         ## and I'll re-run all of the things I just ran.
                                                                                        getrlimit after: (524288, 524288)                                                                                                                                               ## Running 'whoami; ulimit -n' in a subprocess with 'shell=True'                        subprocess: william                                                                     524288
                                                                                                                                                                                ## Running scripts again
subprocess 'whoami' command:
 william                                                                                
subprocess bash script:
 ulimit -n: 524288                                                                      
 ulimit -Sn: 524288                                                                      
 ulimit -Hn: 524288

subprocess python script:                                                                
 (524288, 524288)

## Back to checking within this script.                                                 getrlimit: (524288, 524288)
## setting: resource.setrlimit(resource.RLIMIT_NOFILE, (131072, 131072))
getrlimit after: (131072, 131072)
```

So the current fixes in this PR shows that it does set the soft limit to the same value as the hard limit... But I would maybe suggest that we not attempt to set anything on behalf of the user, and allow their system settings to pass through untouched?

wallentx:
I did test excluding those lines entirely just to verify Madmax doesn't attempt to bump nofiles, and it indeed leaves it up to the user. I can see how this would be pretty annoying/confusing to someone not familiar with how to do this. I think this is the way to go

chia/plotters/madmax.py
```diff
@@ -160,7 +160,9 @@ def plot_madmax(args, chia_root_path: Path, plotters_root_path: Path):
 
         # madMAx has a ulimit -n requirement > 296:
         # "Cannot open at least 296 files, please raise maximum open file limit in OS."
```
altendky, line 162: These comments are clearly out of date.

chia/plotters/madmax.py
```diff
@@ -160,7 +160,9 @@ def plot_madmax(args, chia_root_path: Path, plotters_root_path: Path):
...
-        resource.setrlimit(resource.RLIMIT_NOFILE, (512, 512))
+        soft_limit, hard_limit = resource.getrlimit(resource.RLIMIT_NOFILE)
+        # Set soft limit to max (hard limit)
+        resource.setrlimit(resource.RLIMIT_NOFILE, (hard_limit, hard_limit))
```
altendky, line 165: I'm not really sure that jumping to a limit of 1M (maybe) on my system is really an appropriate response?

wjblanke: i'm not sure what madmax needs wrt/file handle, especially at different bucket sizes so hard to say what number to use aside from the hard limit. i'm not worried about it going off the rails wrt/file handles. i think that is all the soft limit is trying to ensure.

altendky: Seems like something to fix upstream?

wallentx:
This fixes https://github.com/Chia-Network/chia-blockchain/issues/13741

chia/plotters/madmax.py
madMAx43v3r: Should be fixed now.

PR #13857, opened 2022-11-07
Remove incorrect log warning message.

aqk:
The Bitcoin fee estimator is designed to ignore calling remove_tx multiple times for the same tx.


arvidn:
why do we end up calling `remove_tx()` on the same transaction multiple times?

arvidn:
It makes me nervous that we have such bare-bones tests of this feature. the tests we have are not particularly convincing that the estimator works well

emlowe (approved):
Thanks

emlowe:
Arvid, we just want in this PR to eliminate this spamming and bad log message. Adam has some more PRs in-flight to add better testing and fix up remaining issues.

Issue #13839, opened 2022-11-05
Multi NFT Offer Failure


L4bb3rs:
### What happened?

Offers created for more then 1 NFT in exchange for XCH fail. Resulting in error can't create 2 identical coins.

### Version

1.6.1

### What platform are you using?

Windows

### What ui mode are you using?

GUI

kjlynch1974:
Here is a screenshot of the error. ![Chia 1 6 1 error](https://user-images.githubusercontent.com/117544790/200141950-2ac90394-e3a4-44bb-b28e-ac44783396de.png)

Lowestofttim:
I can confirm this is the case.

9x9NFT:
Debug log from Light Wallet 1.6.1 on attempt to accept the offer: (getting same error in gui as mentioned above) 

https://dexie.space/offers/A4mgwi5w6BNQTjGY3ydtgdJY4qMTky9LZFnKAkYJgagY

2022-11-05T23:11:46.194 wallet chia.wallet.wallet         : DEBUG    About to select coins for amount 214999999997
2022-11-05T23:11:46.195 wallet chia.wallet.wallet         : DEBUG    Selected closest greater coin: b'\xf2W/k\xf4\x8fv\x98\x9d\x0cf\x96\xa2\x82\xf3\x17|}&N\x03YN\xb8)\xd7\x1f\xa2g\xc4\xa8\xa6'
2022-11-05T23:11:46.242 wallet chia.wallet.wallet         : DEBUG    Generating transaction for: bae24162efbd568f89bc7a340798a6118df0189eb9e3f8697bcea27af99f8f79 200000000000 {Coin { parent_coin_info: 37ac90fea1742249dcb3bca0430c533ec0916218879b3713a40e2520628981d6, puzzle_hash: 13103db141c654b2129d907a1ecf19fad30da75c2fceaedd8fa46eeac8df8976, amount: 899000000000 }}
2022-11-05T23:11:46.252 wallet chia.wallet.wallet         : DEBUG    Got back max send amount: 960220719623
2022-11-05T23:11:46.253 wallet chia.wallet.wallet         : INFO     coins is not None {Coin { parent_coin_info: 37ac90fea1742249dcb3bca0430c533ec0916218879b3713a40e2520628981d6, puzzle_hash: 13103db141c654b2129d907a1ecf19fad30da75c2fceaedd8fa46eeac8df8976, amount: 899000000000 }}
[...]

PR #13858, opened 2022-11-07
Create royalty payments in fewer coins

Quexington:
https://github.com/Chia-Network/chia-blockchain/issues/13839

This patch fixes a problem in some complex NFT offers where royalties being split between NFTs that take the same percentage result in the creation of a duplicate coin.  I realized it was possible to do the royalty payments all with one coin except in the case where the address is also the same in which case instead of creating all of the offer coins at once, we have the offer coins create further offer coins that are subsequently used to pay out the duplicate royalty payments.

As a side effect, this will usually result in cheaper NFT offers except in the case where all the royalty payments that need to be made are all the same.

### DIFF
--- a/chia/wallet/nft_wallet/nft_wallet.py
+++ b/chia/wallet/nft_wallet/nft_wallet.py
@@ -899,7 +899,11 @@ class NFTWallet:
                         Offer.ph(),
                         primaries=[
                             AmountWithPuzzlehash(
-                                {"amount": sum(p.amount for _, p in payments), "puzzlehash": Offer.ph(), "memos": []}
+                                {
+                                    "amount": uint64(sum(p.amount for _, p in payments)),
+                                    "puzzlehash": Offer.ph(),
+                                    "memos": [],
+                                }
                             )
                         ],
                         fee=fee,
@@ -959,7 +963,7 @@ class NFTWallet:
                                     None,
                                     [
                                         Payment(
-                                            Offer.ph(), sum(p.amount for _, p in duplicate_payments), []
+                                            Offer.ph(), uint64(sum(p.amount for _, p in duplicate_payments)), []
                                         ).as_condition_args()
                                     ],
                                 )
@@ -987,6 +991,8 @@ class NFTWallet:
                                     else:
                                         continue
                                     break
+                        assert royalty_coin is not None
+                        assert parent_spend is not None
                         if asset is None:  # If XCH
                             royalty_sol = inner_royalty_sol
                         else:
--- a/chia/wallet/nft_wallet/nft_wallet.py
+++ b/chia/wallet/nft_wallet/nft_wallet.py
@@ -898,8 +898,9 @@ class NFTWallet:
                         abs(amount),
                         Offer.ph(),
                         primaries=[
-                            AmountWithPuzzlehash({"amount": p.amount, "puzzlehash": Offer.ph(), "memos": []})
-                            for _, p in payments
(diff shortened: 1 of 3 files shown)


chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -898,8 +898,13 @@ async def make_nft1_offer(
...
                         primaries=[
-                            AmountWithPuzzlehash({"amount": p.amount, "puzzlehash": Offer.ph(), "memos": []})
-                            for _, p in payments
+                            AmountWithPuzzlehash(
+                                {
+                                    "amount": uint64(sum(p.amount for _, p in payments)),
+                                    "puzzlehash": Offer.ph(),
+                                    "memos": [],
+                                }
+                            )
```
Quexington, line 907: Instead of a bunch of payments, one payment that is the sum of all their amounts

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -929,29 +934,65 @@ async def make_nft1_offer(
...
-                        # ((nft_launcher_id . ((ROYALTY_ADDRESS, royalty_amount, memos))))
-                        inner_royalty_sol = Program.to([(launcher_id, [payment.as_condition_args()])])
+                    # Create a coin_spend for the royalty payout from OFFER MOD
+
+                    # We cannot create coins with the same puzzle hash and amount
+                    # So if there's multiple NFTs with the same royalty puzhash/percentage, we must create multiple
+                    # generations of offer coins
+                    royalty_coin: Optional[Coin] = None
+                    parent_spend: Optional[CoinSpend] = None
+                    while True:
```
Quexington, line 944: This repeats until there are no more duplicate payments

chia/wallet/trading/offer.py
```diff
@@ -156,25 +156,35 @@ def get_offered_coins(self) -> Dict[Optional[bytes32], List[Coin]]:
 
             parent_puzzle: UncurriedPuzzle = uncurry_puzzle(parent_spend.puzzle_reveal.to_program())
             parent_solution: Program = parent_spend.solution.to_program()
-            additions: List[Coin] = [a for a in parent_spend.additions() if a not in self.bundle.removals()]
+            additions: List[Coin] = parent_spend.additions()
```
Quexington, line 159: I had to change this method because it was a little too heuristic and was picking up the wrong coin as "offered".  We now filter the additions to just unspent coins at the end of the function.

chia/wallet/trading/offer.py
```diff
@@ -156,25 +156,35 @@ def get_offered_coins(self) -> Dict[Optional[bytes32], List[Coin]]:
...
             if puzzle_driver is not None:
                 asset_id = create_asset_id(puzzle_driver)
                 inner_puzzle: Optional[Program] = get_inner_puzzle(puzzle_driver, parent_puzzle)
                 inner_solution: Optional[Program] = get_inner_solution(puzzle_driver, parent_solution)
                 assert inner_puzzle is not None and inner_solution is not None
+
+                # We're going to look at the conditions created by the inner puzzle
                 conditions: Program = inner_puzzle.run(inner_solution)
-                matching_spend_additions: List[Coin] = []  # coins that match offered amount and are sent to offer ph.
+                expected_num_matches: int = 0
```
Quexington, line 170: It also could only match one coin being offered

geoffwalmsley (approved):

trepca (approved):

PR #13859, opened 2022-11-07
adjustments for new mypy

altendky:


chia/consensus/blockchain_interface.py
```diff
@@ -19,22 +19,27 @@ def get_peak_height(self) -> Optional[uint32]:
         pass
 
     def block_record(self, header_hash: bytes32) -> BlockRecord:
-        pass
+        # ignoring hinting error until we handle our interfaces more formally
```
arvidn, line 22: does "more formally" mean making this a protocol, and these function bodies: `...`?

arvidn (approved):

chia/consensus/blockchain_interface.py
altendky: That's a relevant direction, yeah.  Though as I recall there may still be some inconsistency around these and we at least have to stop inheriting from them and maybe more.

emlowe: @arvidn Can you resolve this conversation if Kyles's comment is ok?

PR #13863, opened 2022-11-08
full_node: Manage subscriptions via `ph_subscriptions`/`coin_subscriptions` only

xdustinface:
Avoid housekeeping of the other members, should not be relevant for the 
perfomance?

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -115,12 +115,6 @@ class FullNode:
     coin_subscriptions: Dict[bytes32, Set[bytes32]]
     # Puzzle Hash : Set[Peer ID]
     ph_subscriptions: Dict[bytes32, Set[bytes32]]
-    # Peer ID: Set[Coin ids]
-    peer_coin_ids: Dict[bytes32, Set[bytes32]]
-    # Peer ID: Set[puzzle_hash]
-    peer_puzzle_hash: Dict[bytes32, Set[bytes32]]
-    # Peer ID: subscription count
-    peer_sub_counter: Dict[bytes32, int]
     _transaction_queue_task: Optional[asyncio.Task[None]]
     simulator_transaction_callback: Optional[Callable[[bytes32], Awaitable[None]]]
     _sync_task: Optional[asyncio.Task[None]]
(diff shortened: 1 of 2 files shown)


chia/full_node/full_node_api.py
```diff
@@ -1463,13 +1457,8 @@ async def register_interest_in_puzzle_hash(
             hint_coin_ids.extend(ph_hint_coins)
             if puzzle_hash not in self.full_node.ph_subscriptions:
                 self.full_node.ph_subscriptions[puzzle_hash] = set()
-            if (
-                peer.peer_node_id not in self.full_node.ph_subscriptions[puzzle_hash]
-                and self.full_node.peer_sub_counter[peer.peer_node_id] < max_items
-            ):
+            if self.full_node.peer_subscriptions_count(peer) < max_items:
```
arvidn, line 1460: `peer_subscription_count()` is O(n) in your patch, and it's called in a loop computing almost the same value over and over again. you could hoist this out of the loop, but the wallet may still be making these requests in a loop, and you're basically back to what you have now.

it would seem reasonable to keep the counter in the peer object itself. We don't preserve the subscriptions across disconnects, right?

chia/full_node/full_node_api.py
```diff
@@ -1490,22 +1479,12 @@ async def register_interest_in_puzzle_hash(
...
-            self.full_node.peer_sub_counter[peer.peer_node_id] = 0
         max_items = self.full_node.config.get("max_subscribe_items", 200000)
         for coin_id in request.coin_ids:
             if coin_id not in self.full_node.coin_subscriptions:
                 self.full_node.coin_subscriptions[coin_id] = set()
-            if (
-                peer.peer_node_id not in self.full_node.coin_subscriptions[coin_id]
-                and self.full_node.peer_sub_counter[peer.peer_node_id] < max_items
-            ):
+            if self.full_node.peer_subscriptions_count(peer) < max_items:
```
arvidn, line 1486: same comment here as above

chia/full_node/full_node.py
```diff
@@ -886,25 +877,35 @@ def on_disconnect(self, connection: ws.WSChiaConnection) -> None:
...
+
+        # Remove all coin id subscriptions and drop the coin_id key if the node_ids are empty.
+        remove_coin_ids = set()
+        for coin_id, node_ids in self.coin_subscriptions.items():
+            if node_id in node_ids:
+                node_ids.remove(node_id)
+            if len(node_ids) == 0:
+                remove_coin_ids.add(coin_id)
+        for coin_id in remove_coin_ids:
+            self.coin_subscriptions.pop(coin_id)
```
arvidn, line 908: I'm hesitant to make every peer connect cost be proportional to the total number of subscriptions. I would expect it to be especially noticeable on the RPi, where memory access is slow.

arvidn (commented):
I would prefer to maintain (or improve) the current Big-Oh complexity, but in a more space efficient and encapsulated way. i.e. perhaps moving the counters into the `WSChiaConnection` objects, and maybe using weak references to the `WSChiaConnection` objects instead of (bulky) peer IDs

PR #13867, opened 2022-11-08
server: Drop some unused code

xdustinface:


arvidn (approved):
looks good to me. I think it would be good to hear from the person who added this code too

hoffmang9:
There is a request for me to review this but this is not something I originally wrote or edited. I may be in blame due to some earlier co-merges.

wallentx:
> There is a request for me to review this but this is not something I originally wrote or edited. I may be in blame due to some earlier co-merges.

Ah, your name did appear to pop up when I checked. Looking at the history it seems like it may have been just as you described

xdustinface:
`Delivery` was added by @mariano54 and the rest by @Yostra but all this was added 2-3 years back and i assume this is either leftover from development / not longer used because of some earlier refactoring. In any case, it's not used anymore.

PR #13869, opened 2022-11-08
server: Use `get_extra_info` directly

xdustinface:


xdustinface:
Only works for the `ClientWebSocketResponse`, not for `WebSocketResponse`.

PR #13870, opened 2022-11-08
util: Hint `ApiMetadata.message_class` with `Optional[Type[Streamable]]`

xdustinface:
Im working towards more strict mypy in `chia.server.ws_connection` and  this change helps there while (i think) it still retains the same  behaviour.

### DIFF
--- a/chia/util/api_decorators.py
+++ b/chia/util/api_decorators.py
@@ -27,7 +27,7 @@ class ApiMetadata:
     bytes_required: bool = False
     execute_task: bool = False
     reply_types: List[ProtocolMessageTypes] = field(default_factory=list)
-    message_class: Optional[Any] = None
+    message_class: Any = None
 
 
 def get_metadata(function: Callable[..., object]) -> ApiMetadata:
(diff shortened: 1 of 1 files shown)


chia/util/api_decorators.py
```diff
@@ -27,7 +27,7 @@ class ApiMetadata:
     bytes_required: bool = False
     execute_task: bool = False
     reply_types: List[ProtocolMessageTypes] = field(default_factory=list)
-    message_class: Optional[Any] = None
+    message_class: Any = None
```
arvidn, line 30: I think `Optional[]` adds value here; it's a more specific type than just `Any`.
A better type would be:
```
message_class: Optional[Type[Streamable]] = None
```

arvidn (changes requested):
I would like to hear from @altendky too

chia/util/api_decorators.py
xdustinface: Right, `Type[Streamable]` makes sense.

altendky:
> I would like to hear from @altendky too

I don't remember anything offhand. More specific sounds good if mypy is happy.  If I come up with a problem and this has been merged... Well, we aren't exactly running out of PRs.

arvidn (approved):

PR #13873, opened 2022-11-08
Fix mypy typing issues in clvm_tools

aqk:


altendky (commented):
Looks good, happy to approve once `clvm_tools` is released.  Thanks for the cleanup!

PR #13874, opened 2022-11-08
Connection compression

jiellse:
This adds connection compression between full nodes in order to save bandwidth.
During the handshake they announce their capability to decompress messages if received.
The sending side checks if the receiver is able to decompress and the size of the message is large enough for it to be beneficial, and if so compresses the message using zstandard.
Both compression and decompression is settable in the config.


chia/server/ws_connection.py
```diff
@@ -391,34 +400,135 @@ async def _wait_and_retry(self, msg: Message, queue: asyncio.Queue):
             self.log.debug(f"Exception {e} while waiting to retry sending rate limited message")
             return None
 
+    async def _potentially_compress(self, message: Message) -> Message:
```
arvidn, line 403: this function should probably not be `async`

chia/server/ws_connection.py
```diff
@@ -391,34 +400,135 @@ async def _wait_and_retry(self, msg: Message, queue: asyncio.Queue):
...
+
+        message_to_send = await self._potentially_compress(message)
+
+        if message_to_send != message:
+            encoded = bytes(message_to_send)
+            size = len(encoded)
+
         if not self.outbound_rate_limiter.process_msg_and_check(
-            message, self.local_capabilities, self.peer_capabilities
+            message_to_send, self.local_capabilities, self.peer_capabilities
```
arvidn, line 463: it would seem simpler to just pass `message` here, just like in the current logic. it would create a cleaner layering of where messages are compressed. I don't think you would need to consider the case where an already compressed message is passed to `_potentially_compress()` either

chia/server/ws_connection.py
```diff
@@ -391,34 +400,135 @@ async def _wait_and_retry(self, msg: Message, queue: asyncio.Queue):
...
+        # The 'test_mode' is only to make testing easier
+        try:
+            if Capability.CAN_DECOMPRESS_MESSAGES in self.local_capabilities and len(full_message_loaded.data) > 0:
+                # Check the uncompressed size before doing the decompression (must be less than 4 GiB)
+                decompressed_size = get_decompressed_size(full_message_loaded.data[1:])
+                self.log.debug(
+                    f"Received compressed message {ProtocolMessageTypes(full_message_loaded.data[0]).name}: "
+                    f"compressed {len(full_message_loaded.data)}, decompressed {decompressed_size}"
+                )
+                if decompressed_size > 0 and decompressed_size < (2 ** (LENGTH_BYTES * 8)):
```
arvidn, line 501: we already have a message size limit (that I believe is a lot smaller than 4 GiB). you should check against that

chia/server/ws_connection.py
```diff
@@ -391,34 +400,135 @@ async def _wait_and_retry(self, msg: Message, queue: asyncio.Queue):
...
+                    # Replace the message so it appears as if it was sent uncompressed
+                    # The rate limiter has already checked the message (using the size of the compressed message)
+                    full_message_loaded = Message(
+                        uint8(full_message_loaded.data[0]),
+                        full_message_loaded.id,
+                        zstd.decompress(full_message_loaded.data[1:]),
+                    )
+                    return full_message_loaded
+                else:
+                    self.log.debug(" -> It is either not using Zstandard or it is too big")
```
arvidn, line 511: I believe this should be a disconnect and probably a ban

chia/server/ws_connection.py
```diff
@@ -391,34 +400,135 @@ async def _wait_and_retry(self, msg: Message, queue: asyncio.Queue):
...
+            else:
+                # We received a compressed message but we are not configured to do decompression
+                # Discard this message, don't let it slip through (unless we are doing pytest)
```
arvidn, line 514: this should also be a disconnect and ban, right? what legitimate case would the peer have to send us a message it should know we don't support?

chia/server/ws_connection.py
```diff
@@ -391,34 +400,135 @@ async def _wait_and_retry(self, msg: Message, queue: asyncio.Queue):
...
+                if len(full_message_loaded.data) > 0:
+                    self.log.debug(
+                        "Received compressed message "
+                        f"{ProtocolMessageTypes(full_message_loaded.data[0]).name}, "
+                        "but we are not configured to decompress"
+                    )
+                else:
+                    self.log.debug("Received compressed message, but it is totally empty")
+                if test_mode:
+                    return full_message_loaded
```
arvidn, line 524: this makes me nervous, why do we need a special case for tests?

chia/server/ws_connection.py
```diff
@@ -461,7 +571,13 @@ async def _read_one_message(self) -> Optional[Message]:
             self.bytes_read += len(data)
             self.last_message_time = time.time()
             try:
-                message_type = ProtocolMessageTypes(full_message_loaded.type).name
+                if (
+                    full_message_loaded.type == ProtocolMessageTypes.wrapped_compressed.value
+                    and Capability.CAN_DECOMPRESS_MESSAGES in self.local_capabilities
+                ):
+                    message_type = ProtocolMessageTypes(full_message_loaded.data[0]).name
```
arvidn, line 578: you don't seem to be using the protocol message type here either.

chia/protocols/full_node_protocol.py
```diff
@@ -203,3 +203,12 @@ class RequestPeers(Streamable):
 @dataclass(frozen=True)
 class RespondPeers(Streamable):
     peer_list: List[TimestampedPeerInfo]
+
+
+@streamable
+@dataclass(frozen=True)
+class WrappedCompressed(Streamable):
+    # The inner (wrapped) type's ProtocolMessageTypes enum
+    inner_type: uint8
```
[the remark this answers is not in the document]
arvidn: this type is not used anywhere, it makes the protocol less robust as changing this type won't highlight any of the places that need to agree on this protocol. nor will it actually change the protocol.

arvidn (changes requested):
I didn't see a test exercising the attack of sending a small compressed message that balloons once decompressed. (what currently is limited to 4 GiB, but should be set much lower).

Also, I would like to see some justification for compression. What are the savings on average? are there specific use cases where it's especially valuable?

chia/protocols/full_node_protocol.py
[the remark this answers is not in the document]
jiellse: uint8 is used because the field in Message has that type.

chia/server/ws_connection.py
```diff
@@ -391,34 +400,135 @@ async def _wait_and_retry(self, msg: Message, queue: asyncio.Queue):
...
+        # then compress it by transforming the original message into a new compressed message
+        if (
+            self.sending_compressed_enabled
+            and Capability.CAN_DECOMPRESS_MESSAGES in self.peer_capabilities
+            and len(message.data) >= self.compress_if_at_least_size
+        ):
+            compressed = Message(
+                uint8(ProtocolMessageTypes.wrapped_compressed.value),
+                message.id,
+                bytes([message.type]) + bytes(zstd.compress(message.data))
```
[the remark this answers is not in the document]
jiellse: Not a good reason, no. While coding this up I didn't have a WrappedCompressed, it was only when preparing for the PR when I got an error that a class matching the ProtocolMessageTypes was missing that I added it. Will fix this.

chia/server/ws_connection.py
jiellse: I agree the async can go.

chia/server/ws_connection.py
jiellse: The message_to_send is needed here to use the updated message size when determining if it should be rate limited or not
Regarding point two: If a compressed message was rate-limited it is put back on the outgoing_queue, it shouldn't be compressed once more.

chia/server/ws_connection.py
jiellse: I used the assert that was already at the beginning of _send_message as a guide for this. Will use the value max_message_size (50MiB) from server.py instead

chia/server/ws_connection.py
jiellse: Yes, disconnect and ban is a valid outcome.

chia/server/ws_connection.py
jiellse: Agree here also. Should I use Err.INVALID_PROTOCOL_MESSAGE or add a new value?

chia/server/ws_connection.py
jiellse: I wanted a way to read the specific bytes on the wire to be sure the message was sent compressed during testing. I may find another solution for this however, so the test_mode can disappear.

chia/server/ws_connection.py
jiellse: Same reason as above, will fix.

jiellse:
Compression is helpful in different ways:
Some ISP's have a different upload speed compared to download.
Some ISP's have limits on how much can be downloaded and/or uploaded per month (might lower the connection speed
if above the limit), or can charge per GB/month (might be more common with mobile plans).

The messages sent can be quite large, for example anything that contains a FullBlock can be over 400 KB. When long-syncing 32 FullBlocks are sent at a time, so the size can add up quickly.

I ran a check through the db, and doing a compression on the blocks that are over 8KB (the value I have suggested in this PR in the config) yields about 65% savings overall, less for small blocks.
Less bytes to send means it reaches the destination faster and can also help with users' data plans.

chia/server/ws_connection.py
arvidn: the cost of re-compressing might be low compared to the complexity of not having a clean boundary separating compressed and uncompressed messages. but it's a judgment call.

chia/server/ws_connection.py
arvidn: INVALID_PROTOCOL_MESSAGE seems appropriate

chia/protocols/full_node_protocol.py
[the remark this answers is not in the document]
altendky: Arvid, I believe this would come down to `Streamable` not supporting enumerations.  I had suggested previously that we make it so it does since this would avoid hazards around hand-coded type selection when processing the messages, but there were other concerns.

emlowe:
aiohttp websockets has built in compression support following RFC 7692 - both ws_connect has a compress parameter as does WebSocketReponse. The RFC only specified deflate support, so this is a minor downside, however, the ability to turn on an RFC-compliant already implemented feature in aiohttp likely outweighs a roll-your-own version.

I recommend trying the aiohttp built-in support - this also makes it easier to make configurable.

chia/server/ws_connection.py
```diff
@@ -391,34 +400,135 @@ async def _wait_and_retry(self, msg: Message, queue: asyncio.Queue):
...
+                # the immediate line above is in practice a serialized WrappedCompressed
+            )
+            # Validating and safe-guarding
+            if get_decompressed_size(compressed.data[1:]) == len(message.data):
```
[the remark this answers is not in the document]
jiellse: I check the decompressed size before that happens, and depend on the zstd module to honor the size set in the frame. But I haven't validated the code in the Python zstd module, or facebook's code.

emlowe:
I used your script to run some timing tests on a couple of systems

TL;DR: while zstd is considerably faster, I don't think introducing a 50ms delay is likely to be a problem for using the standard configuration deflate. Even zstd longest was in the 16-20ms range.

Any such setting does need to be configura

### Pi4:

### deflate/zlib, level==1

```
                                  Sum (bytes)                       Total time (s)          Average (ms)
  Range (bytes)    Count   uncompressed   compressed   Saving %   compress  decompress   compress  decompress
 --------------  -------   ------------  -----------   --------   --------  ----------   --------  ----------
400000 -      0    11309     4843898164   1904804877       60.7      244.9       43.48     21.655       3.845

Longest to compress: 50.195ms
Longest to decompress: 6.698ms
```

### deflate/zlib, level==-1
```
                                   Sum (bytes)                       Total time (s)          Average (ms)
  Range (bytes)    Count   uncompressed   compressed   Saving %   compress  decompress   compress  decompress
 --------------  -------   ------------  -----------   --------   --------  ----------   --------  ----------
400000 -      0    11309     4843898164   1825988496       62.3      331.3       38.07     29.296       3.366

Longest to compress: 56.800ms
Longest to decompress: 6.032ms
```

### zstd level==1
```
                                   Sum (bytes)                       Total time (s)          Average (ms)
  Range (bytes)    Count   uncompressed   compressed   Saving %   compress  decompress   compress  decompress
 --------------  -------   ------------  -----------   --------   --------  ----------   --------  ----------
400000 -      0    11329     4852196332   1304767134       73.1       10.9        8.42      0.958       0.743

Longest to compress: 16.693ms
Longest to decompress: 7.218ms
```

### zstd level==3
```
                                   Sum (bytes)                       Total time (s)          Average (ms)
  Range (bytes)    Count   uncompressed   compressed   Saving %   compress  decompress   compress  decompress
 --------------  -------   ------------  -----------   --------   --------  ----------   --------  ----------
400000 -      0    11309     4843898164   1293737577       73.3       38.8        9.92      3.433       0.877

Longest to compress: 10.113ms
Longest to decompress: 3.561ms
```

### Core2Duo P8600 (2008 cpu):

### deflate/zlib, level==1

```
                                  Sum (bytes)                       Total time (s)          Average (ms)
  Range (bytes)    Count   uncompressed   compressed   Saving %   compress  decompress   compress  decompress
 --------------  -------   ------------  -----------   --------   --------  ----------   --------  ----------
400000 -      0    11329     4852196332   1907718960       60.7      144.7       27.29     12.774       2.409

Longest to compress: 31.357ms
Longest to decompress: 16.522ms
```

### deflate/zlib, level==-1
```
                                   Sum (bytes)                       Total time (s)          Average (ms)
  Range (bytes)    Count   uncompressed   compressed   Saving %   compress  decompress   compress  decompress
 --------------  -------   ------------  -----------   --------   --------  ----------   --------  ----------
400000 -      0    11329     4852196332   1828818326       62.3      195.7       23.64     17.274       2.087

Longest to compress: 38.092ms
Longest to decompress: 16.113ms
```

### zstd level==1
```
                                   Sum (bytes)                       Total time (s)          Average (ms)
  Range (bytes)    Count   uncompressed   compressed   Saving %   compress  decompress   compress  decompress
 --------------  -------   ------------  -----------   --------   --------  ----------   --------  ----------
400000 -      0    11329     4852196332   1304767134       73.1       10.9        8.42      0.958       0.743

Longest to compress: 16.693ms
Longest to decompress: 7.218ms
```

### zstd level==3
```
                                   Sum (bytes)                       Total time (s)          Average (ms)
  Range (bytes)    Count   uncompressed   compressed   Saving %   compress  decompress   compress  decompress
 --------------  -------   ------------  -----------   --------   --------  ----------   --------  ----------
400000 -      0    11329     4852196332   1295995407       73.3       16.7        6.62      1.474       0.584

Longest to compress: 20.567ms
Longest to decompress: 8.236ms
```

PR #13875, opened 2022-11-09
server: Return earlier in `WSChiaConnection.__getattr__`

xdustinface:


chia/server/ws_connection.py
```diff
@@ -319,24 +319,24 @@ async def invoke(*args, **kwargs):
 
             msg: Message = Message(uint8(getattr(ProtocolMessageTypes, attr_name).value), None, args[0])
             request_start_t = time.time()
-            result = await self.send_request(msg, timeout)
+            message = await self.send_request(msg, timeout)
```
arvidn, line 322: a nitpick: this `message` is the response from the peer. given that the request message is called `msg` it might be a bit confusing to call the response the same thing. I would think `request` and `response` would be more descriptive names

arvidn (dismissed):

chia/server/ws_connection.py
xdustinface: Fair nitpick :) didn't realise the `msg` there. See 6c5ef48c68b6da51b41dbe4d51b1816b8fcd7516.

arvidn (approved):

PR #13876, opened 2022-11-09
server: Introduce `WSChiaConnection._get_extra_info`

xdustinface:
Makes it only one place where `mypy` complains if enabled to be more strict.

### DIFF
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -65,7 +65,7 @@ class WSChiaConnection:
         # Remote properties
         self.peer_host = peer_host
 
-        peername = self.ws._writer.transport.get_extra_info("peername")
+        peername = self._get_extra_info("peername")
 
         if peername is None:
             raise ValueError(f"Was not able to get peername from {self.peer_host}")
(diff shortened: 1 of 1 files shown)


chia/server/ws_connection.py
```diff
@@ -117,6 +117,9 @@ def __init__(
         self.version = ""
         self.protocol_version = ""
 
+    def _get_extra_info(self, name: str) -> Optional[Any]:
+        return self.ws._writer.transport.get_extra_info(name)
```
arvidn, line 121: I don't understand how this can help mypy with anything. `get_extra_info()` is part of the standard library, right? I would expect that to be typed already.

arvidn (approved):

chia/server/ws_connection.py
xdustinface: It's just wrapping so that this assert https://github.com/Chia-Network/chia-blockchain/pull/13878/commits/30d0e1dee21d1154f68e09b27547d54040783747#diff-076ad87bb2203f6728105a8ac4aab4e126ba0113a38e4efff07884ed4934cabfR127 is required once, not thrice. `_writer` is optional and while `ClientWebSocketResponse` contains a `get_extra_info` wrapper, `WebSocketResponse` doesn't but in our case `ws` can be either of both.

PR #13877, opened 2022-11-09
server: Don't pass the queue into `WSChiaConnection._wait_and_retry`

xdustinface:


arvidn:
this looks like an obvious and straight-forward improvement. does mypy cover this file? would it tell us if there are any other calls to `_wait_and_retry()` than the one you updated the call site for?

arvidn (approved):

xdustinface:
>  does mypy cover this file? 

Not in here but this PR is basically a breakout of https://github.com/Chia-Network/chia-blockchain/pull/13878 where `mypy` gets enabled for this file.

PR #13879, opened 2022-11-09
server: Add optional `outbound` parameter to `get_connections`

xdustinface:


emlowe:
So: `outbound==None`, return all connections. `outbound==False` return inbound connections, `outbound==True` return outbound connections

I might think about a "Direction" enum with in, out, both - but maybe only slightly more readable

xdustinface:
@emlowe Hm, yeah it might be better. I had a different approach before i ended up with this PR which also contained a `Direction` enum but only with `Inbound` and `Outbound`, not the in/out you mentioned. See the two most recent commits of this branch https://github.com/xdustinface/chia-blockchain/tree/direction. The first commit though also changes the `is_outbound: bool` parameter of `WSChiaConnection` to `direction: Direction`. I wasn't soo convinced at the end so i came up with this PR, but let me know your thoughts.

emlowe (approved):
Approved as is -

PR #13880, opened 2022-11-09
refactor test utils for import availability outside of repo

almogdepaz:
moved tests/util/keyring.py, test/util/setup_nodes.py, test.utils.setup_services.py under chia/simulator 

that way they can be imported and used for testing from other repos like the minting tool or clawback


xdustinface (changes requested):
Hm, might be only me but it feels wrong to mix all those test helpers like `tests/setup_nodes.py` and `tests/util/keyring.py` just into `chia.util` where we from my understanding only collect stuff which is used by production code. From the branch name i assume you want to import this stuff in the simulator so maybe moving them to `chia.simulator`  would be better?

almogdepaz:
> Hm, might be only me but it feels wrong to mix all those test helpers like `tests/setup_nodes.py` and `tests/util/keyring.py` just into `chia.util` where we from my understanding only collect stuff which is used by production code. From the branch name i assume you want to import this stuff in the simulator so maybe moving them to `chia.simulator` would be better?

i haven't decided yet if thats the appropriate place, this is just the first attempt the goal is to be able to use this things outside of the chia-blockchain repo, i don't think simulator is appropriate for stuff like setup_nodes though maybe this needs a new folder ?

xdustinface (changes requested):
Some more thoughts:
- Two of the moved files are also in `isort.cfg`, they should be dropped there now.
- Seems like there are a bunch of unrelated changes, see the comments. And since you also renamed the files in `mypy.ini` the partial hint fixes in the moved files are probably not required? I assume this was an attempt to fix all hints before you renamed them in `mypy.ini`?
- Seems like the LGTM alert above it legit and can be fixed?

It's not that all the extra hint changes are bad at the end, just a bit confusing in the context of the PR imo. I think it would at least makes sense to drop the unrelated ones i commented on and also fix the isort config and LGTM, for the hint changes in the moved files im not absolutely requesting a change.

almogdepaz:
> Some more thoughts:
> 
> * Two of the moved files are also in `isort.cfg`, they should be dropped there now.
> * Seems like there are a bunch of unrelated changes, see the comments. And since you also renamed the files in `mypy.ini` the partial hint fixes in the moved files are probably not required? I assume this was an attempt to fix all hints before you renamed them in `mypy.ini`?
> * Seems like the LGTM alert above it legit and can be fixed?
> 
> It's not that all the extra hint changes are bad at the end, just a bit confusing in the context of the PR imo. I think it would at least makes sense to drop the unrelated ones i commented on and also fix the isort config and LGTM, for the hint changes in the moved files im not absolutely requesting a change.

@xdustinface i dont see any harm in fixing some of the easy mypy issues, i can revert if you insist on this.
i tried fixing the whole thing but the tech debt here is too big.

xdustinface (approved):
> i dont see any harm in fixing some of the easy mypy issues, i can revert if you insist on this.

Approved, but yeah, i still think those changes shouldn't be there. Sure there is no harm but all the ones i commented on are still in files completely unrelated to this PR and i think we should have as clear scopes as possible for each PR and not just mix in changes in random files.

almogdepaz:
> > i dont see any harm in fixing some of the easy mypy issues, i can revert if you insist on this.
> 
> Approved, but yeah, i still think those changes shouldn't be there. Sure there is no harm but all the ones i commented on are still in files completely unrelated to this PR and i think we should have as clear scopes as possible for each PR and not just mix in changes in random files.

generally i agree but his is test infra and typing so no real changes,  im trying to get these files out of the mypy ignore list, i am currently handling the service creation issue in another pr, hope this will allow to clean everything here

emlowe (approved):

PR #13887, opened 2022-11-09
add new, optimized tx queue

jack60612:
- [x] Working new queue
- [x] Priority for local tx's 
- [x] Unit Tests


chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+
+    async def put(self, tx: TransactionQueueEntry, high_priority: bool = False) -> None:
+        if tx.peer is None or high_priority:  # when it's local there is no peer.
+            await self._high_priority_queue.put(tx)
+        else:
+            if tx.peer not in self._queue_dict:
+                self._queue_dict[tx.peer] = asyncio.Queue(500)
+                self._index_to_peer_map.append(tx.peer)
+            try:
+                self._queue_dict[tx.peer].put_nowait(tx)
```
arvidn, line 48: did you consider making `put()` sleep here if the queue is full? That might simplify the API, you may no longer need `empty()`, `full()` or `qsize()`

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+            except asyncio.QueueFull:
+                self.log.warning(f"Transaction queue full for peer {tx.peer.peer_host}")
+                return
+        self._queue_length.release()  # increment semaphore to indicate that we have a new item in the queue
+
+    async def get(self) -> TransactionQueueEntry:
```
arvidn, line 54: I think the item is removed here too, right? I think `pop()` would be a better name, but maybe if the python convention is `get()` we should go with that.

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+                return self._queue_dict[peer].get_nowait()
+            self._list_iterator += 1
+            peer = self._index_to_peer_map[self._list_iterator]
+            if self._list_iterator > list_length:
+                # reset iterator
+                self._list_iterator = 0
+
+    async def clean_up_queue(self) -> None:
+        while True:
+            await asyncio.sleep(600)
```
arvidn, line 71: this doesn't look right. is this really necessary?

arvidn (changes requested):
You need unit tests too

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+from chia.server.ws_connection import WSChiaConnection
+from chia.types.transaction_queue_entry import TransactionQueueEntry
+
+
+@dataclass
+class TransactionQueue:
+    _list_iterator: int
+    _queue_length: asyncio.Semaphore
+    _index_to_peer_map: List[WSChiaConnection]
+    _queue_dict: Dict[WSChiaConnection, asyncio.Queue[TransactionQueueEntry]]
```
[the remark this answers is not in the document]
jack60612: both async and normal queue have built in syncing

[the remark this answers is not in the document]
jack60612: good point about the peer id

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+    _high_priority_queue: asyncio.Queue[TransactionQueueEntry]
+    max_size: int
```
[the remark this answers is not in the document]
jack60612: i had a limit on both but ill get rid of the main limit

chia/full_node/tx_processing_queue.py
jack60612: yeah i just changed it up and simplified the api

chia/full_node/tx_processing_queue.py
jack60612: ill change it to pop

chia/full_node/tx_processing_queue.py
jack60612: don't we need to delete old queues?

arvidn: I thought your idea of removing empty queues once per loop was good. It strikes a balance between capping resource use and still give peers some time to reuse the previous one if there's another incoming transaction. I also think it is a simpler design to not have yet another task running

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+                # reset iterator
+                self._list_iterator = 0
+
+    async def clean_up_queue(self) -> None:
+        while True:
+            await asyncio.sleep(600)
+            for peer_id in self._queue_dict:
+                if self._queue_dict[peer_id].empty():
+                    self._queue_dict.pop(peer_id)
+                    self._index_to_peer_map.remove(peer_id)
```
arvidn, line 73: this ends up being a linear search still, which isn't great. On the other hand, the size of this list will probably rarely be very long anyway, so if you can't think of a simple fix, it's probably fine to leave it like this

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+            if self._list_iterator > len(self._index_to_peer_map):
+                self._list_iterator = 0
```
arvidn, line 75: this will bias the queues we pick from to favor lower indices. I don't think this is OK. Rather than doing this, make sure to update the `_list_iterator` when removing an item with a lower or equal index.

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+@dataclass
+class TransactionQueue:
+    """
+    This class replaces one queue by using a high priority queue for local transactions and separate queues for peers.
+    Local transactions are processed first.
+    Then the next transaction is taken from the next non-empty queue after the last processed queue. (round-robin)
+    This decreases the effects of one peer spamming your node with transactions.
+    """
+
+    _list_iterator: int
```
arvidn, line 21: I think the term "iterator" refers to specific kinds of objects in python (and many other languages too). I think "cursor" would be a better name. maybe even have a comment to describe what it's used for.

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+    _queue_length: asyncio.Semaphore
+    _index_to_peer_map: List[bytes32]
+    _queue_dict: Dict[bytes32, asyncio.Queue[TransactionQueueEntry]]
+    _high_priority_queue: asyncio.Queue[TransactionQueueEntry]
```
arvidn, line 25: I think this should be a `queue.SimpleQueue`. We don't use any of the synchronization that comes with `asyncio.Queue`

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+        await self._queue_length.acquire()
+        list_length = len(self._index_to_peer_map)
+        if not self._high_priority_queue.empty():
+            return self._high_priority_queue.get_nowait()
+        peer_id = self._index_to_peer_map[self._list_iterator]
+        while True:
+            if not self._queue_dict[peer_id].empty():
+                return self._queue_dict[peer_id].get_nowait()
+            self._list_iterator += 1
+            peer_id = self._index_to_peer_map[self._list_iterator]
```
arvidn, line 62: I think this would throw an `IndexError` exception. but you'll discover that when you add the tests.

chia/full_node/tx_processing_queue.py
jack60612: ok ill revert back to that.

chia/full_node/tx_processing_queue.py
```diff
@@ -0,0 +1,75 @@
...
+
+    async def put(self, tx: TransactionQueueEntry, high_priority: bool = False) -> None:
+        if tx.peer is None or high_priority:  # when it's local there is no peer.
+            await self._high_priority_queue.put(tx)
+        else:
+            if tx.peer.peer_node_id not in self._queue_dict:
+                self._queue_dict[tx.peer.peer_node_id] = asyncio.Queue(self.peer_size_limit)
+                self._index_to_peer_map.append(tx.peer.peer_node_id)
+            try:
+                self._queue_dict[tx.peer.peer_node_id].put_nowait(tx)
```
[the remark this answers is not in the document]
jack60612: before these changes, if the queue was full we didnt add it, i thought this was a good idea, unless you think we dont want that.

chia/full_node/tx_processing_queue.py
jack60612: yeah there should always be under 100 peers so it should be fine.

chia/full_node/tx_processing_queue.py
jack60612: SimpleQueue Doesn't have a peer transaction limit

chia/full_node/tx_processing_queue.py
jack60612: it will, oversight on my part.

jack60612: > I think this would throw an `IndexError` exception. but you'll discover that when you add the tests.

i don't think it will, however it is broken, so im fixing it.

chia/full_node/tx_processing_queue.py
jack60612: yeah your right.

chia/full_node/tx_processing_queue.py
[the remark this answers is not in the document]
arvidn: actually, in an ideal world, we would apply back-pressure to the peer and reject transactions once the queue is full. Maybe that's better. In that case, none of the queues should be `asyncio.Queue` but just `queue.SimpleQueue`

chia/full_node/tx_processing_queue.py
arvidn: you increment `_list_iterator` without checking if you just past the last item of the list, then you access the list at that index.

chia/full_node/tx_processing_queue.py
arvidn: that's fine. it has a length

chia/full_node/tx_processing_queue.py
arvidn: I would assume closer to 1000 for a big node. Anyway, one simple change to turn this from O(n^2) to O(n) would be to not update `_index_to_peer_map` as you go, but rebuild it from scratch when you're done.

jack60612: > I would assume closer to 1000 for a big node. Anyway, one simple change to turn this from O(n^2) to O(n) would be to not update `_index_to_peer_map` as you go, but rebuild it from scratch when you're done.

i did realize that i was doing remove, so i changed the loop a bit to loop off of the list instead of the dict.

jack60612: @arvidn after looking here: [https://wiki.python.org/moin/TimeComplexity](https://wiki.python.org/moin/TimeComplexity) I don't think I have found a better way to do it.

jack60612: @arvidn after looking here: [https://wiki.python.org/moin/TimeComplexity](https://wiki.python.org/moin/TimeComplexity) I don't think I have found a better way to do it.

jack60612:
im going to finish this and add unit tests later, ill let you know when i want another review @arvidn

chia/full_node/tx_processing_queue.py
arvidn: iterating over the list would have better complexity. this is untested, but something like this:

```
new_peer_map = []
for idx, peer_id in enumerate(self._index_to_peer_map):
    if self._queue_dict[peer_id].empty():
        self._queue_dict.pop(peer_id)
        if idx < self._list_iterator:
            self._list_iterator -= 1
    else:
        new_peer_map.append(peer_id)
self._index_to_peer_map = new_peer_map
```

jack60612: > iterating over the list would have better complexity. this is untested, but something like this:
> 
> ```
> new_peer_map = []
> for idx, peer_id in enumerate(self._index_to_peer_map):
>     if self._queue_dict[peer_id].empty():
>         self._queue_dict.pop(peer_id)
>         if idx < self._list_iterator:
>             self._list_iterator -= 1
>     else:
>         new_peer_map.append(peer_id)
> self._index_to_peer_map = new_peer_map
> ```

i kept the list building part, but dropped the enumerate part, as we are resetting to zero anyway.

arvidn: right

jack60612:
> im going to finish this and add unit tests later, ill let you know when i want another review @arvidn

rdy for review

arvidn (commented):
I think the tests could be simpler (and a bit easier to follow). I believe there are a few cases missing:

* test when a peer id's queue is full. expect some kind of feedback (probably an exception)
* test that one peer posting 2 items and a different peer posts 1 item, still are serviced fairly
* it might make sense to ensure the `pop()` function does in fact sleep when the queue is empty, and wakes up when an item is pushed.

tests/core/full_node/test_tx_processing_queue.py
```diff
@@ -0,0 +1,240 @@
...
+    return TransactionQueueEntry(
+        sb,
+        None,
+        sb.name(),
+        peer,
+        False,
+    )
+
+
+class TestTransactionQueue:
```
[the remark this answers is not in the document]
jack60612: > I think you should remove this class and make all the functions free functions. that way you'll save one level of indentation too

we always use classes when using tests, so i think we should keep it.

chia/full_node/tx_processing_queue.py
```diff
@@ -18,58 +19,56 @@ class TransactionQueue:
...
+                self._queue_dict[tx.peer.peer_node_id] = SimpleQueue()
                 self._index_to_peer_map.append(tx.peer.peer_node_id)
-            try:
-                self._queue_dict[tx.peer.peer_node_id].put_nowait(tx)
-            except asyncio.QueueFull:
+            if self._queue_dict[tx.peer.peer_node_id].qsize() <= self.peer_size_limit:
+                self._queue_dict[tx.peer.peer_node_id].put(tx)
+            else:
                 self.log.warning(f"Transaction queue full for peer {tx.peer.peer_node_id}")
                 return
```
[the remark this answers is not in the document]
jack60612: I'm going to add this and then add an error to the message queue for the local wallet, but the current blockchain protocol doesn't have responses for these types of messages and our node simply dropped the tx's if they were full.

jack60612:
> I think the tests could be simpler (and a bit easier to follow). I believe there are a few cases missing:
> 
> * test when a peer id's queue is full. expect some kind of feedback (probably an exception)
> * test that one peer posting 2 items and a different peer posts 1 item, still are serviced fairly
> * it might make sense to ensure the `pop()` function does in fact sleep when the queue is empty, and wakes up when an item is pushed.

test 1 and 3 added, and existing were cleaned up.

chia/full_node/full_node_api.py
```diff
@@ -247,9 +248,12 @@ async def respond_transaction(
...
         # TODO: Use fee in priority calculation, to prioritize high fee TXs
-        await self.full_node.transaction_queue.put(
-            TransactionQueueEntry(tx.transaction, tx_bytes, spend_name, peer, test)
-        )
+        try:
+            await self.full_node.transaction_queue.put(
+                TransactionQueueEntry(tx.transaction, tx_bytes, spend_name, peer, test), peer.peer_node_id
+            )
+        except TransactionQueueFull:
+            pass  # we can't do anything here, the tx will be dropped. We might do something in the future.
```
arvidn, line 256: right, I see. peers aren't sending these TXs to us, we're requesting them. In the future we might want to stop requesting more TXs once we get close to the limit

tests/core/full_node/test_tx_processing_queue.py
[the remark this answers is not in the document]
arvidn: it's used in many places, most likely because of some original misunderstandings followed by copy+paste. We shouldn't keep bad habits just because they're habits.

tests/core/full_node/test_tx_processing_queue.py
```diff
@@ -0,0 +1,212 @@
...
+
+def get_transaction_queue_entry(peer_id: bytes32) -> TransactionQueueEntry:
+    sb: SpendBundle = standard_spend_bundle
+    return TransactionQueueEntry(
+        sb,
+        peer_id,  # we cheat a bit here by reusing transaction_bytes to store our peer_id
+        sb.name(),
+        None,
+        False,
+    )
```
arvidn, line 58: I think all of the code above is a distraction, diluting these tests and making them harder to follow. the `TransactionQueueEntry` does not need to contain a peer_id or a `SpendBundle`. I think you should create a dummy class that just has some identifier in it (to tell them apart), then cast that to a `TransactionQueueEntry`. It should work at run-time because the queue shouldn't be looking at any of its fields.

tests/core/full_node/test_tx_processing_queue.py
```diff
@@ -0,0 +1,212 @@
...
+
+        resulting_txs = []
+        for _ in range(num_txs):
+            resulting_txs.append(await transaction_queue.pop())
+
+        assert transaction_queue._queue_length._value == 0  # check that all are removed
+        for i in range(num_txs):
+            assert list_txs[i] == resulting_txs[i]
+
+        with pytest.raises(asyncio.TimeoutError):
```
arvidn, line 117: is this actually waiting in wall-clock time? If so, I think we should test this without sleeping. e.g. we could create a task and ensure it's blocked in the `pop()`

tests/core/full_node/test_tx_processing_queue.py
```diff
@@ -0,0 +1,212 @@
...
+        list_txs = [get_transaction_queue_entry(peer_id) for _ in range(num_txs)]
+        for tx in list_txs:
+            await transaction_queue.put(tx, peer_id)
+
+        assert transaction_queue._queue_length._value == num_txs  # check that all are included
+        assert transaction_queue._index_to_peer_map == [peer_id]  # sanity checking
+
+        resulting_txs = []
+        for _ in range(num_txs):
+            resulting_txs.append(await transaction_queue.pop())
```
arvidn, line 111: it seems to me that the first part of this test, pushing 100 items and then popping them is redundant. You already have a test that does that

tests/core/full_node/test_tx_processing_queue.py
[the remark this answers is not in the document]
jack60612: > > I think you should remove this class and make all the functions free functions. that way you'll save one level of indentation too
> 
> we always use classes when using tests, so i think we should keep it.

unless there is a specific reason not to?

tests/core/full_node/test_tx_processing_queue.py
```diff
@@ -0,0 +1,212 @@
...
+    )
+
+
+class TestTransactionQueue:
+    @pytest.fixture(scope="function")
+    def transaction_queue(self) -> TransactionQueue:
+        return TransactionQueue(1000, log)
+
+    @pytest.mark.asyncio
+    async def test_local_txs(self, transaction_queue: TransactionQueue) -> None:
```
arvidn, line 67: I don't see a test that ensures local TXs take precedence over TXs we receive from peers

arvidn (commented):
generally it's not great practice to have tests verify internal state (and implementation details). Normally you want to just verify the observable behavior of the class

tests/core/full_node/test_tx_processing_queue.py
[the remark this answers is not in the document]
arvidn: in my mind, the main reasons not to are:
1. save one level of indentation
2. save lines of code
3. make it clear to the reader that the tests do not in fact share state (which could be a legitimate reason to use a class)

tests/core/full_node/test_tx_processing_queue.py
jack60612: i was about to add to add that lol, good catch.

tests/core/full_node/test_tx_processing_queue.py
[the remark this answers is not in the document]
jack60612: ok that makes sense, will do!

chia/full_node/full_node_api.py
jack60612: i could add that

arvidn: that should be a separate PR. we don't want to tie too many independent changes together

tests/core/full_node/test_tx_processing_queue.py
jack60612: we are testing the non high priority queue, all local / high priority txs bypass the round robin logic.

tests/core/full_node/test_tx_processing_queue.py
jack60612: changed that, thanks!

tests/core/full_node/test_tx_processing_queue.py
```diff
@@ -0,0 +1,184 @@
...
+
+    resulting_txs = []
+    for _ in range(total_txs):
+        resulting_txs.append(await transaction_queue.pop())
+
+    assert transaction_queue._index_to_peer_map == []  # we should have removed all the peer ids
+
+
+@pytest.mark.asyncio
+async def test_queue_cleanup_and_fairness() -> None:
```
arvidn, line 150: this test could be as simple as:

* pushing 2 entries for peer A, one entry for B and 2 entries for C
* pop 3 items (ensuring it's from A, B and C)
* pop 2 items, ensuring it's from A and C (in that order)

It's problematic to have tests assert state of implementation details. One of the purposes of tests is to ensure the class behaves correctly after refactoring it, or changing the implementation strategy. With tests like this, one would have to re-write the tests as well, which defeats an important point of having them.

arvidn:
This looks a lot better. I can't find anything obvious that's not coverered by a test. However, I think it might make sense to put a bit more work into making the tests simple (after all, we don't test the tests, we rely on them being simple enough to ensure their correctness by manual inspection).

Also, tests that access private members and test implementation details are fragile and not as useful as tests that ensure the observed behavior only.

arvidn (approved):

PR #13911, opened 2022-11-15
Update workflows to use GITHUB_OUTPUT (signed commits)

wallentx:
This PR is a copy of https://github.com/Chia-Network/chia-blockchain/pull/13730 that includes signed commits.

Use GITHUB_OUTPUT in workflows instead of set-output.

GitHub is deprecating set-output as per their blog:

There is no change in behavior with these changes from the prior.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -41,7 +41,7 @@ jobs:
           fetch-depth: 0
 
       - name: Setup Python environment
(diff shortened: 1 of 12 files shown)


cmmarslender (dismissed):

emlowe (dismissed):

emlowe:
There seems to be something wrong with building the Windows installer.

paninaro (approved):

PR #13914, opened 2022-11-15
server: Drop broken message types logging

xdustinface:
Seems like the idea was to log an overview of message types for the  currently active api calls whenever we have multiples of 100 running.  This currently isn't working at all because `len(message_types) % 100 ==  0` will only trigger if there are 100 different message types in the  counter (we don't even have that much different messages). A fix would  be to check `message_types.total() % 100 == 0`. But then, if we hover  around 100 active messages this will repeatedly trigger logs since we  decrement for each processed message, im not sure this all makes much  sense and we anyway log all received message types with `DEBUG` logs so  i decided to just drop this. If someone thinks we should keep it with  the fix or has ideas to make this a more reasonable log instead of  dropping it, let me know :)

### DIFF
--- a/chia/server/server.py
+++ b/chia/server/server.py
@@ -5,14 +5,11 @@ import logging
 import ssl
 import time
 import traceback
-from collections import Counter
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


arvidn (approved):
seems reasonable to me

PR #13916, opened 2022-11-15
Simplify `ChiaServer.close_all_connections`

xdustinface:

### DIFF
--- a/chia/server/server.py
+++ b/chia/server/server.py
@@ -714,12 +714,9 @@ class ChiaServer:
         return result
 
     async def close_all_connections(self) -> None:
-        keys = [a for a, b in self.all_connections.items()]
-        for node_id in keys:
+        for connection in self.all_connections.copy().values():
             try:
-                if node_id in self.all_connections:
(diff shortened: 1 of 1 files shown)


chia/server/server.py
```diff
@@ -714,14 +714,11 @@ def get_connections(self, node_type: Optional[NodeType] = None) -> List[WSChiaCo
...
-                if node_id in self.all_connections:
-                    connection = self.all_connections[node_id]
-                    await connection.close()
-            except Exception as e:
-                self.log.error(f"Exception while closing connection {e}")
+        try:
+            for connection in self.all_connections.copy().values():
+                await connection.close()
+        except Exception as e:
+            self.log.error(f"Exception while closing connection {e}")
```
arvidn, line 721: one change you introduce here is that if a call to `close()` throws, you abort, previously we would keep trying to close the remaining connections

xdustinface: Right, thanks, changed!

emlowe (approved):

Issue #13765, opened 2022-10-31
[Bug] Failed to set DID on NFT


Jsewill:
### What happened?

When attempting to move an NFT to a DID to which I've moved other NFTs in the past, I received the error, `The NFT move failed: Error: Failed to set DID on NFT:`. This has been a problem since at least 1.5.0.

### Version

1.6.0

### What platform are you using?

Linux

### What ui mode are you using?

GUI

### Relevant log output

```shell
wallet chia.rpc.wallet_rpc_api    : ERROR    Failed to set DID on NFT:~
Traceback (most recent call last):
  File "/redacted/path/chia-blockchain/chia/rpc/wallet_rpc_api.py", line 1581, in nft_set_nft_did
    spend_bundle = await nft_wallet.set_nft_did(nft_coin_info, did_id, fee=fee)
  File "/redacted/path/chia-blockchain/chia/wallet/nft_wallet/nft_wallet.py", line 912, in set_nft_did
    did_inner_hash, did_bundle = await self.get_did_approval_info(nft_id, bytes32(did_id))
  File "/redacted/path/chia-blockchain/chia/wallet/nft_wallet/nft_wallet.py", line 292, in get_did_approval_info
    did_bundle = await wallet.create_message_spend(puzzle_announcements=[nft_id])
  File "/redacted/path/chia-blockchain/chia/wallet/did_wallet/did_wallet.py", line 703, in create_message_spend
    assert coins is not None
AssertionError
```

ytx1991:
Did you move any NFT before this? You need to want the movement confirmed on the chain before moving the next one.

Jsewill:
@ytx1991 Yes, just as I said in the issue description. I've moved many NFTs to the DID in question, all confirmed on chain. I had to resync, and now the wallet won't let me move to or from said DID.

Others had said that a full resync would help, so I tried removing my chia directory and reinitializing it a couple times to no avail.

Jsewill:
No, I will give that a try and report back.

ytx1991:
So you are able to see the DID wallet after the resync?

Jsewill:
I have been able to see the DID wallet the whole time, both in settings->profiles, and viewing the NFTs within them. I simply can't move NFTs to or from it.

ytx1991:
[...]

Issue #13691, opened 2022-10-23
DID wallet disappeared / stopped working


codephunk:
### What happened?

I created a DID for my NFT collection on my farmer. I assigned that DID to a new NFT wallet I wanted to use for minting.
Minting started just fine, but after a while my minting script stopped.
Since I wasn't able to access my farmer for a while I wanted to move my minting operations to a new node.
But on that new node my DID wallet never showed up, even though I used the same key as on my farmer. The NFT wallet did show up and had the correct DID assigned to it, but if I try to continue minting with it, it says the assigned DID's wallet could not be found.

I did some research and found a thread in the Chia forum, where someone had a similar problem and solved it by copying the original wallet db that has been used to create the DID to the new machine.
I tried that as well and that actually made the wallet show up in my wallet list.

But since then the DID wallet has a balance of zero while my other DID wallets all have a balance of 1.0. I guess that means the DID smart coin isn't even really there, right?
Minting also doesn't work anymore. Trying to continue minting my collection gives me the attached warning and error.
So, it seems like the DID wallet only exists in that original wallet db which was in use when I created the DID. Even if I delete the wallet db on my farmer and sync it again without accepting unknown peers, the DID wallet doesn't show up.

The DID is did:chia:12e7kac8e2grp82t2p765y2wuqlzq505jtcafmddj6qx3sp2n4p0qx449ze if that helps.

I already tried everything I found on the web several times. Waiting, raising the derivation index, restarting the client, deleting and rebuilding the databases, creating a new DID, creating a POOL NFT and deleting unconfirmed transactions.

### Version

1.6.0

### What platform are you using?

Linux

### What ui mode are you using?

CLI

### Relevant log output

```shell
2022-10-22T21:23:33.600 wallet chia.wallet.did_wallet.did_wallet: WARNING  Can't select 1, from spendable 0 for wallet id 38
2022-10-22T21:23:33.612 wallet chia.rpc.util              : WARNING  Error while handling message: Traceback (most recent call last):
  File "/home/pi/chia-blockchain/chia/rpc/util.py", line 16, in inner
    res_object = await f(request_data)
  File "/home/pi/chia-blockchain/chia/rpc/wallet_rpc_api.py", line 1549, in nft_mint_nft
    spend_bundle = await nft_wallet.generate_new_nft(
  File "/home/pi/chia-blockchain/chia/wallet/nft_wallet/nft_wallet.py", line 384, in generate_new_nft
[...]

PR #13925, opened 2022-11-16
Fix DID unnecessary wallet deletion

ytx1991:
Fix a bug causing unnecessary DID wallet deletion

The DID wallet will be deleted and then recreated if users transfer their DID to a new address.

DID wallet will not be deleted unless there is an additional spend to prove the new address is not belong to the wallet.

no

Covered by Unit tests


chia/wallet/wallet_state_manager.py
```diff
@@ -787,7 +787,7 @@ async def handle_did(
 
         launch_id: bytes32 = bytes32(bytes(singleton_struct.rest().first())[1:])
         if derivation_record is None:
-            if new_p2_puzhash == old_inner_puzhash or p2_puzzle == new_p2_puzhash:
+            if new_p2_puzhash != old_inner_puzhash and p2_puzzle != new_p2_puzhash:
```
trepca, line 790: can you explain this change?

ytx1991: When the wallet cannot find a key for the hint (new_p2_puzhash), it will check if it's because of the hint bug.
 In this case, if the hint doesn't equal the previous didDIDinner puzhash or p2_puzzle hash, we know the user may update the p2_address. We have to wait for an additional spend to confirm the p2_puzzle is changed then delete the wallet.

arvidn:
I really think you should add unit tests covering the fixes. It seems unlikely the change to the test covers both the true and false conditions of the new production logic

chia/wallet/wallet_state_manager.py
trepca: thanks, lets add a comment here to explain the logic

chia/rpc/wallet_rpc_api.py
```diff
@@ -1764,6 +1773,153 @@ async def did_get_info(self, request) -> EndpointResult:
             "hints": hints,
         }
 
+    async def did_find_lost_did(self, request) -> EndpointResult:
+        """
+        Recover a bugged DID wallet by a coin id of the DID
```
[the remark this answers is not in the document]
ytx1991: Yes, it can be any coin id of the DID

ytx1991:
Related issues
#13691 #13765

ytx1991:
To make the fix testable and concise, I removed the modification of the syncing logic and try to keep it clean. Instead, a new API for recovering lost DID is added. User should be able to find their lost DID by calling this API.

trepca (approved):

chia/rpc/wallet_rpc_api.py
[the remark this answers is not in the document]
wallentx: Verified with @trepca that this convo is resolved

PR #13947, opened 2022-11-18
tools: Implement and test `tools/legacy_keyring.py`

xdustinface:
Adds a tool which allows to access to the legacy keyring after #13398 has been merged which drops the legacy keyring support from all production code.

To use the tool (or run the tests) on linux it's additionally required to run `install.sh` with the `-l` flag to install the legacy keyring dependency.  

### Help message
```
Usage: legacy_keyring.py [OPTIONS] COMMAND [ARGS]...

  Manage the keys in the legacy keyring.

Options:
  --help  Show this message and exit.

Commands:
  clear  Remove all keys
  show   Show all available keys
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,7 +7,7 @@ Usage: $0 [-adlsph]
 
   -a                          automated install, no questions
   -d                          install development dependencies
-  -l                          install legacy keyring dependencies (required for linux only)
+  -l                          install legacy keyring dependencies (linux only)
   -s                          skip python package installation and just do pip install
   -p                          additional plotters installation
   -h                          display this help and exit
--- a/install.sh
+++ b/install.sh
@@ -3,10 +3,11 @@
 set -o errexit
 
 USAGE_TEXT="\
-Usage: $0 [-adsph]
+Usage: $0 [-adlsph]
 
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -70,6 +70,10 @@
     "types-setuptools",
 ]
 
+legacy_keyring_dependencies = [
+    "keyrings.cryptfile @ git+https://github.com/frispete/keyrings.cryptfile@master",
```
arvidn, line 74: it makes me nervous to depend on `master` rather than a specific commit

xdustinface: This is still a draft for that reason :)

install.sh
```diff
@@ -3,10 +3,11 @@
 set -o errexit
 
 USAGE_TEXT="\
-Usage: $0 [-adsph]
+Usage: $0 [-adlsph]
 
   -a                          automated install, no questions
   -d                          install development dependencies
+  -l                          install legacy keyring dependencies (required for linux only)
```
paninaro, line 10:
```suggestion
  -l                          install legacy keyring dependencies (linux only)
```

paninaro (approved):
lgtm

PR #13948, opened 2022-11-18
Revert "Return minted NFT ID & Allow transfer NFT with NFT ID"

altendky:
Reverts Chia-Network/chia-blockchain#13928


altendky:
fix was merged

PR #13951, opened 2022-11-18
Preserve correct `MempoolItem` block height when rebuilding mempool

aqk:
Fix the following: First seen block_height was not being preserved in MempoolItems during mempool rebuild

No, it's a fixing change.

Yes.

No, no UI behaviour

### DIFF
--- a/benchmarks/mempool.py
+++ b/benchmarks/mempool.py
@@ -148,7 +148,7 @@ async def run_mempool_benchmark(single_threaded: bool) -> None:
             for tx in spend_bundles:
                 npc = await mempool.pre_validate_spendbundle(tx, None, tx.name())
                 assert npc is not None
-                _, status, error = await mempool.add_spend_bundle(tx, npc, tx.name())
+                _, status, error = await mempool.add_spend_bundle(tx, npc, tx.name(), rec.height)
                 assert status == MempoolInclusionStatus.SUCCESS
                 assert error is None
 
--- a/benchmarks/mempool.py
+++ b/benchmarks/mempool.py
@@ -148,7 +148,7 @@ async def run_mempool_benchmark(single_threaded: bool) -> None:
             for tx in spend_bundles:
                 npc = await mempool.pre_validate_spendbundle(tx, None, tx.name())
                 assert npc is not None
-                _, status, error = await mempool.add_spend_bundle(tx, npc, tx.name(), rec.height)
+                _, status, error = await mempool.add_spend_bundle(tx, npc, tx.name(), height)
                 assert status == MempoolInclusionStatus.SUCCESS
                 assert error is None
 
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -361,7 +361,9 @@ class MempoolManager:
             new_spend: spend bundle to validate
             npc_result: result of running the clvm transaction in a fake block
             spend_name: hash of the spend bundle data, passed in as an optimization
-            first_added_height: The block height when this SpendBundle was first added
+            first_added_height: The block height that `new_spend`  first entered this node's mempool.
+                Used to estimate how long a spend has taken to be included on the chain.
+                This value could differ node to node. Not preserved across full_node restarts.
 
         Returns:
             Optional[Err]: Err is set if we cannot add to the mempool, None if we will immediately add to mempool
@@ -599,8 +601,6 @@ class MempoolManager:
                 )
                 # Only add to `seen` if inclusion worked, so it can be resubmitted in case of a reorg
                 if result == MempoolInclusionStatus.SUCCESS:
-                    new_item = self.get_mempool_item(item.spend_bundle_name)
-                    assert item == new_item
                     self.add_and_maybe_pop_seen(item.spend_bundle_name)
                 # If the spend bundle was confirmed or conflicting (can no longer be in mempool), it won't be
                 # successfully added to the new mempool.
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -7,7 +7,7 @@ import time
 from concurrent.futures import Executor
 from concurrent.futures.process import ProcessPoolExecutor
 from multiprocessing.context import BaseContext
-from typing import Dict, List, Optional, Set, Tuple
+from typing import Callable, Dict, List, Optional, Set, Tuple
 
 from blspy import GTElement
 from chiabip158 import PyBIP158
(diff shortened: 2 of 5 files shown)


chia/full_node/mempool_manager.py
```diff
@@ -350,6 +362,7 @@ async def validate_spend_bundle(
             new_spend: spend bundle to validate
             npc_result: result of running the clvm transaction in a fake block
             spend_name: hash of the spend bundle data, passed in as an optimization
+            first_added_height: The block height when this SpendBundle was first added
```
arvidn, line 365: I don't think this description is super helpful. The variable name says the same thing. My understanding is that this is the full node's peak height at the time this transaction was received from the network. Different nodes may have different heights for the same transactions (right?). It's not obvious what it's useful for, that would be nice to have an explanation for also, especially given that it's not a network-wide constant.

chia/full_node/mempool_manager.py
```diff
@@ -582,9 +595,13 @@ async def new_peak(
...
             self.seen_bundle_hashes = {}
             for item in old_pool.spends.values():
-                _, result, err = await self.add_spend_bundle(item.spend_bundle, item.npc_result, item.spend_bundle_name)
+                _, result, err = await self.add_spend_bundle(
+                    item.spend_bundle, item.npc_result, item.spend_bundle_name, item.height_added_to_mempool
+                )
                 # Only add to `seen` if inclusion worked, so it can be resubmitted in case of a reorg
                 if result == MempoolInclusionStatus.SUCCESS:
+                    new_item = self.get_mempool_item(item.spend_bundle_name)
+                    assert item == new_item
```
arvidn, line 604: this looks like debugging left-overs. did you mean to leave it in?

chia/full_node/mempool_manager.py
```diff
@@ -131,7 +133,9 @@ def shut_down(self) -> None:
         self.pool.shutdown(wait=True)
 
     async def create_bundle_from_mempool(
-        self, last_tb_header_hash: bytes32
+        self,
+        last_tb_header_hash: bytes32,
+        item_inclusion_filter: Optional[Callable[[MempoolManager, MempoolItem], bool]] = None,
```
arvidn, line 138: this looks unrelated to preserving the block height when rebuilding mempool. in an ideal world, it would be a separate PR or at least commit. This is necessary because the coin-selection logic has a filter, right? Or is this a debug-only feature?

arvidn (commented):
I don't see any non-test uses of the inclusion filter. I can imagine that we need it, but it's a bit suspicious.
It would also be nice to have unit tests that *just* instantiate the class we test (i.e. `MempoolManager`), with so many other moving parts it's hard to be certain that the tests are correct.

chia/full_node/mempool_manager.py
aqk: Yes, I'll put some thought into a better explanation.

chia/full_node/mempool_manager.py
aqk: This is essentially a debug-only feature. I was hoping to leave this out of this particular PR, because the testing support code is much larger than the actual fix. But I do next it for 2 other PRs, and this test would also be very difficult without it.

Here's why we need the filter: the MempoolManager algorithm to select spends for a block is not modular.
Also, the code to create new MempoolItem instances is deep inside `validate_spend_bundle`, which is several calls down from the main action.

What needs to happen eventually, is we need to refactor some of the giant methods in MempoolManager, but definitely not now - this is as small as I can make it and still test the production code.

At first, I was reluctant to include this is main - I didn't want to have people messing around with the Mempool selection algorithm. But on further reflection, that only matters for consensus.

aqk:
> I don't see any non-test uses of the inclusion filter. I can imagine that we need it, but it's a bit suspicious. 

Yes, good eye. It is not needed for normal operation, because our very simple transaction selection algorithm is baked into   `create_bundle_from_mempool`. However, we **need** it for testing.

I could have fully modularized the inclusion algorithm in this patch by moving the following checks into a default implementation of `item_inclusion_filter`, but I chose not to do that, partly to make this change as small as possible, and partly as a safeguard that cannot be overridden.

```
                    item.cost + cost_sum <= self.limit_factor * self.constants.MAX_BLOCK_COST_CLVM
                    and item.fee + fee_sum <= self.constants.MAX_COIN_AMOUNT
```

> It would also be nice to have unit tests that _just_ instantiate the class we test (i.e. `MempoolManager`)

Hahahaha, yes, I agree! However, if you attempt to write a test that only instantiates MempoolManager, you will soon find yourself re-implementing half of `Server` and `FullNode`. `Mempool` is very easy to test on its own, but `MempoolManager` is not. I suspect that's why this is the first test of `MempoolManager`.

chia/full_node/mempool_manager.py
aqk: You know, it is debugging leftover, but I did intend to leave it in, as an invariant.

This invariant expresses the exact problem that this PR fixes - that the `MempoolItem` included into the "new" Mempool by `add_spendbundle` is different than the MempoolItem in the "old" Mempool when rebuilding the mempool during `new_peak`.

benchmarks/mempool.py
```diff
@@ -148,7 +148,7 @@ async def add_spend_bundles(spend_bundles: List[SpendBundle]) -> None:
             for tx in spend_bundles:
                 npc = await mempool.pre_validate_spendbundle(tx, None, tx.name())
                 assert npc is not None
-                _, status, error = await mempool.add_spend_bundle(tx, npc, tx.name())
+                _, status, error = await mempool.add_spend_bundle(tx, npc, tx.name(), rec.height)
```
almogdepaz, line 151: i would use hight instead of rec.height.

unrelated to the pr this whole fake_block_record thing smells, dont think we should pass a block record if it is not needed, the correct thing would probably be pass only the vars you need or a new class that contains only those

chia/full_node/mempool_manager.py
almogdepaz: are we calling get_mempool_item on the item we just added to see if they are the same ?
if thats the case i think i would do some kind of unittest instead of leaving it here

aqk: Almog, you read my mind. I was just coming back to ask if you two thought I should move this out into a test.

aqk: This is checked in `test_mempoolitem_height_added`

benchmarks/mempool.py
aqk: Thanks - changed

almogdepaz (approved):
lgtm

arvidn (approved):

PR #13967, opened 2022-11-21
dont wait for timeout on None response

almogdepaz:
Allow None responses from failed calls to other nodes to not wait-out the timeout

currently if a request fails the serving node doesn't respond with a message which makes the  receiving node wait the full request timeout before we return 

this changes that so on the serving node we catch failures and return an empty message of a new type,
and on the calling node catch this empty message and return  None to the calling method (similar behavior to timeout)

no

this is covered by existing tests, a unittest was added to make sure this actually works as intended


arvidn (commented):
this is just implementing support for the requestor to consider this new message a failure. As far as I can tell, there's no request that actually responds with this new message when it fails. I was under the impression that most requests already have a failure message type. Is that not the case?

chia/server/server.py
```diff
@@ -639,6 +639,13 @@ async def wrapped_coroutine() -> Optional[Message]:
                     if response is not None:
                         response_message = Message(response.type, full_message.id, response.data)
                         await connection.send_message(response_message)
+                    elif metadata.reply_types:
```
almogdepaz, line 642: @arvidn this is where we catch the failed request and return a failure message instead of just not responding

almogdepaz:
> this is just implementing support for the requestor to consider this new message a failure. As far as I can tell, there's no request that actually responds with this new message when it fails. I was under the impression that most requests already have a failure message type. Is that not the case?

this is incorrect, currently inside server py we only catch methods that resulted in something other then None and we send a message with the result, i added a section where if the result was None we now make a new "none response" message
so the calling node is notified and doesn't wait the full timeout, we dont have this machnisem for any of our methods as far as i know except for request_block_headers that has a reject response

chia/server/server.py
```diff
@@ -639,6 +639,14 @@ async def wrapped_coroutine() -> Optional[Message]:
                     if response is not None:
                         response_message = Message(response.type, full_message.id, response.data)
                         await connection.send_message(response_message)
+                    # check that this call needs a reply
+                    elif metadata.reply_types is not None and len(metadata.reply_types) > 0:
```
almogdepaz, line 643: @arvidn this is where we catch the failed request and return a failure message instead of just not responding

altendky: So the `.reply_types` aren't actually used anywhere else.  In https://github.com/Chia-Network/chia-blockchain/pull/13565 I refactored from multiple api decorators, one for each piece of information, to a single decorator.  While doing so I noticed that the information from the reply type decoration wasn't used.  It seemed like it should be used instead of the separately defined mapping.

https://github.com/Chia-Network/chia-blockchain/blob/f9f74481713b0edce85fcc6dbb000fe840b2a7db/chia/protocols/protocol_state_machine.py#L14-L45

But...  I hadn't gotten back to refactoring that duplication/dead code.  I hesitate to use `.reply_types` in the mean time since it may not be in sync with the other definition of the data.

altendky (dismissed):

chia/server/server.py
almogdepaz: asserting the correct message type is done here

https://github.com/Chia-Network/chia-blockchain/blob/4db7fa956c72d9295a2d58b0ddf7651b33ed93af/chia/server/ws_connection.py#L366

arvidn (approved):

arvidn (dismissed):
part of me feels like this is an awkward protocol. Generally, every request ought to have well defined success and failure messages. It's fine to have generic failure messages shared by different requests. But this looks like covering the case "oops, we have a bug; we better respond with *something*". But maybe I'm misunderstanding.

almogdepaz:
> part of me feels like this is an awkward protocol. Generally, every request ought to have well defined success and failure messages. It's fine to have generic failure messages shared by different requests. But this looks like covering the case "oops, we have a bug; we better respond with _something_". But maybe I'm misunderstanding.

agree on the awkward part, i feel this whole thing needs a refactor mostly to get all the working parts in one coherent place so its easy to follow .

this covers all the cases of us failing to come out with a response to a call that expects something back (that can happen for various reasons although rare )

im not sure requiring a dedicated reject response for each call (as done in request_block_headers) is better since that leaves it to the devs to remember to add it and its needed for every call that expects something back (or at least i cant think of where we dont want it)

chia/server/ws_connection.py
```diff
@@ -558,3 +558,6 @@ def get_peer_logging(self) -> PeerInfo:
             return PeerInfo(self.peer_host, port)
         else:
             return info
+
+    def has_capability(self, Capability) -> bool:
+        return Capability in self.peer_capabilities
```
altendky, line 563: https://github.com/Chia-Network/chia-blockchain/actions/runs/3592717699/jobs/6048798710#step:9:100
> `chia/server/ws_connection.py:562: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]`


```suggestion
    def has_capability(self, capability: Capability) -> bool:
        return capability in self.peer_capabilities
```

almogdepaz: damn, thanks

arvidn (approved):

PR #9443, opened 2021-12-02
Add support for pending transactions to get_mempool_item_by_tx_id().

rwarren:
The original implementation of `get_mempool_item_by_tx_id()` only gets
transactions that have actually made it into the mempool. Transactions that
have been queued up for the future with `MempoolInclusionStatus.PENDING`
(e.g. those with `ASSERT_HEIGHT_*` conditions) are completely missed by the
function.

Adding the optional `include_pending` argument allows for the retrieval of
these PENDING transactions as well. The default value of `False` ensures that
no old code will have problems.

This change resolves some confusion that can result after pushing a tx with
`push_tx`. It is typical to monitor progress of the transaction by checking
the mempool status with `get_mempool_item_by_tx_id()`... except that PENDING
transactions (that are relegated to `MempoolManager.potential_cache`) were
always reporting back as "not in mempool", even though they are queued up to
potentially enter the mempool in the next round. Code that interpreted "not
in mempool" as a trigger to check `Coin.spent` for included coins to see if
the tx was successful or not would fail in this case. Also worth noting is
that, although `push_tx` returns the `MempoolInclusionStatus` to let you know
that the tx is `PENDING`, there are still timing considerations when polling
to see if the tx is still in the mempool. Detection of mempool presence could
be missed if polling timing is off.  The `include_pending` flag solves all of
these issues since the result can now be interpreted as "waiting for the
mempool to process the tx", whether or not the tx is in the mempool, or
PENDING.

PR #13976, opened 2022-11-22
Add support for pending transactions to `get_mempool_item_by_tx_id()` #9443

wallentx:
Co-Authored-By: @rwarren
Co-Authored-By: @emlowe 

Purpose: This PR is a clone of #9443 that had an unsigned commit

Current Behavior: https://github.com/Chia-Network/chia-blockchain/pull/9443 cannot be merged due to the first commit being unsigned

New Behavior: I took https://patch-diff.githubusercontent.com/raw/Chia-Network/chia-blockchain/pull/9443.diff and applied it on top of `main` in my branch, while also attributing credit to all who contributed in the original PR. There should be no drift in the intended effect of the original PR, and this one.

Breaking changes: None

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -553,11 +553,17 @@ class MempoolManager:
             return self.mempool.spends[bundle_hash].spend_bundle
         return None
 
-    def get_mempool_item(self, bundle_hash: bytes32) -> Optional[MempoolItem]:
-        """Returns a MempoolItem if it's inside one the mempools"""
-        if bundle_hash in self.mempool.spends:
-            return self.mempool.spends[bundle_hash]
-        return None
+    def get_mempool_item(self, bundle_hash: bytes32, include_pending: bool = False) -> Optional[MempoolItem]:
+        """
+        Returns a MempoolItem if it's inside one the mempools.
+
+        If include_pending is specified, also check the PENDING cache.
+        """
+        item = self.mempool.spends.get(bundle_hash, None)
+        if not item and include_pending:
+            # no async lock needed since we're not mutating the potential_cache
+            item = self.potential_cache._txs.get(bundle_hash, None)
+        return item
 
     async def new_peak(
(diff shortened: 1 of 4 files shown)


chia/full_node/mempool_manager.py
```diff
@@ -553,11 +553,17 @@ def get_spendbundle(self, bundle_hash: bytes32) -> Optional[SpendBundle]:
...
         return None
 
-    def get_mempool_item(self, bundle_hash: bytes32) -> Optional[MempoolItem]:
-        """Returns a MempoolItem if it's inside one the mempools"""
-        if bundle_hash in self.mempool.spends:
-            return self.mempool.spends[bundle_hash]
-        return None
+    def get_mempool_item(self, bundle_hash: bytes32, include_pending: bool = False) -> Optional[MempoolItem]:
+        """
+        Returns a MempoolItem if it's inside one the mempools.
```
arvidn, line 558: I can't parse this sentence. is it supposed to say "inside one *of* the mempools"?
But we only have one mempool, so I still wouldn't know what that means really. presumably "it" refers to the `bundle_hash` we pass in here, right?

arvidn (commented):
looks correct to me (apart from the docstring). Do we have any documentation for our RPCs? If so, I would have expected to see an update to those in this PR as well.

rwarren:
@arvidn  - That bit of the docstring is 100% unchanged from before my original [PR](https://github.com/Chia-Network/chia-blockchain/pull/9443/files) and, as you say, does make no sense.

When I authored the original PR here for the `include_pending` logic, I had explicitly chosen to not change the original docstring (aside from the bit about the new arg) since I didn't want to mix the new feature with a documentation fix, even if required.  Just stuck with what the PR was about.

arvidn (approved):

emlowe (approved):

chia/full_node/mempool_manager.py
wallentx: I've spawned https://github.com/Chia-Network/chia-blockchain/issues/13984 for future cleanup to resolve this convo

PR #13981, opened 2022-11-22
Opt for a function instead of the broke from inner loop pattern

AmineKhaldi:


chia/full_node/mempool_manager.py
```diff
@@ -150,40 +175,16 @@ def always(mm: MempoolManager, mi: MempoolItem) -> bool:
...
         log.info(f"Starting to make block, max cost: {self.constants.MAX_BLOCK_COST_CLVM}")
-        for dic in reversed(self.mempool.sorted_spends.values()):
-            if broke_from_inner_loop:
-                break
-            for item in dic.values():
-                log.info(f"Cumulative cost: {cost_sum}, fee per cost: {item.fee / item.cost}")
-                if (
-                    item.cost + cost_sum <= self.limit_factor * self.constants.MAX_BLOCK_COST_CLVM
-                    and item.fee + fee_sum <= self.constants.MAX_COIN_AMOUNT
-                    and item_inclusion_filter(self, item)
```
arvidn, line 168: I'm pretty sure the inclusion filter is meant to filter out single items, but the way it's used here (and still in the function you factored out) it will abort the whole loop. I believe this is wrong. We need a test for this also.

arvidn: @aqk could you take a look? ^^

AmineKhaldi: @aqk I sent a [fix suggestion](https://github.com/Chia-Network/chia-blockchain/pull/13981/commits/0b64ed4286a5bf2ad7530fba3b1ed7a8aec75475) if you could kindly check. Thanks.

aqk: Thanks - see you in a few mins

aqk: Do you want to merge https://github.com/Chia-Network/chia-blockchain/pull/13981/commits/0b64ed4286a5bf2ad7530fba3b1ed7a8aec75475 first?

aqk: Okay, we decided to put in the small fix first, and do more work on tests on this PR.

arvidn (dismissed):

chia/full_node/mempool_manager.py
AmineKhaldi: PR #13992 ^

chia/full_node/mempool_manager.py
```diff
@@ -173,18 +153,39 @@ def always(mm: MempoolManager, mi: MempoolItem) -> bool:
...
-                    break
-        if len(spend_bundles) > 0:
-            log.info(
-                f"Cumulative cost of block (real cost should be less) {cost_sum}. Proportion "
-                f"full: {cost_sum / self.constants.MAX_BLOCK_COST_CLVM}"
-            )
-            agg = SpendBundle.aggregate(spend_bundles)
-            return agg, additions, removals
-        else:
+                    return (spend_bundles, uint64(cost_sum), additions, removals)
```
arvidn, line 156: this is a nit-pick: if you invert the if-statement and return, you don't need an else-block, and the code on line 150-154 no longer need to be indented. I think code generally reads easier with early-exits and fewer levels of indentation.

chia/full_node/mempool_manager.py
```diff
@@ -173,18 +153,39 @@ def always(mm: MempoolManager, mi: MempoolItem) -> bool:
...
+        return (spend_bundles, uint64(cost_sum), additions, removals)
+
+    async def create_bundle_from_mempool(
```
arvidn, line 159: this function does not need to be `async`. That's an old issue though, so it's probably best to address it separately.

arvidn (dismissed):

arvidn (approved):

PR #13992, opened 2022-11-23
Skip only the item that doesn't satisfy the inclusion filter

AmineKhaldi:


arvidn:
this is a testing feature, and I don't believe this testing feature was mentioned in @aqk's original patch (in the changelog) so I don't think this PR should be part of the changelog either

aqk (approved):
I Agree with Arvid - this codepath is only used in testing, and is not visible to users, so it's not appropriate for the changelog.

arvidn (approved):

PR #13993, opened 2022-11-23
Avoid recomputing additions names

AmineKhaldi:


arvidn (commented):
I don't think the `Dict` property is really necessary. You could just use a `List[Tuple[Bytes32, Coin]]` instead. And I don't see why we would need both `additions_dict` and `additions` either.

But the most important thing; is there a test where we try to add a TX creating a duplicate output? There really needs to be one, to ensure these checks either still work or are redundant.

emlowe:
closing old PR - code has largely moved to chia_rs

PR #14008, opened 2022-11-27
server: Move api call processing into `WSChiaConnection`

xdustinface:
We currently have one queue `ChiaServer.incoming_queue` which gets passed to each new connection and all connections just push their messages into this queue as they come in. Whenever messages are put into the queue they are read by `ChiaServer.imcoming_task` and transformed into async tasks to complete the related api method with now limit like how many tasks will be created at the time. From my understanding this may lead to unfair prioritising for connections sending more/faster messages because its not really possible to limit concurrent tasks per connection in a reasonable way?

With this PR the api call processing gets moved into the `WSChiaConnection` class where each connection gets its own queue so that we can better coordinate tasks from all connections and get some limiting of active task per connection but this can/should be done in a separate PR.

### DIFF
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -6,7 +6,8 @@ import logging
 import time
 import traceback
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Optional, Tuple, Union
+from secrets import token_bytes
+from typing import Any, Awaitable, Callable, Dict, List, Optional, Set, Tuple, Union
 
 from aiohttp import ClientSession, WSCloseCode, WSMessage, WSMsgType
 from aiohttp.client import ClientWebSocketResponse
@@ -15,8 +16,8 @@ from typing_extensions import Protocol, final
 
 from chia.cmds.init_funcs import chia_full_version_str
 from chia.protocols.protocol_message_types import ProtocolMessageTypes
-from chia.protocols.protocol_state_machine import message_response_ok
-from chia.protocols.protocol_timing import INTERNAL_PROTOCOL_ERROR_BAN_SECONDS
+from chia.protocols.protocol_state_machine import message_requires_reply, message_response_ok
+from chia.protocols.protocol_timing import API_EXCEPTION_BAN_SECONDS, INTERNAL_PROTOCOL_ERROR_BAN_SECONDS
 from chia.protocols.shared_protocol import Capability, Handshake
 from chia.server.outbound_message import Message, NodeType, make_msg
 from chia.server.rate_limits import RateLimiter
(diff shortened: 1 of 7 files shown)


chia/server/ws_connection.py
```diff
@@ -74,8 +77,11 @@ class WSChiaConnection:
     is_outbound: bool
 
     # Messaging
-    incoming_queue: asyncio.Queue[Tuple[Message, WSChiaConnection]]
+    received_message_callback: Optional[ConnectionCallback]
+    incoming_queue: asyncio.Queue[Message] = field(default_factory=asyncio.Queue)
     outgoing_queue: asyncio.Queue[Message] = field(default_factory=asyncio.Queue)
+    api_tasks: Dict[bytes32, asyncio.Task[None]] = field(default_factory=dict)
```
arvidn, line 83: I think some comments here may be warranted. what's the distinction between an "api_task" and an "execute_task"?

chia/server/ws_connection.py
```diff
@@ -310,6 +329,89 @@ async def outbound_handler(self) -> None:
                 self.log.error(f"Exception: {e} with {self.peer_host}")
                 self.log.error(f"Exception Stack: {error_stack}")
 
+    async def _api_call(self, full_message: Message, task_id: bytes32) -> None:
```
arvidn, line 332: was this code moved verbatim?

chia/server/ws_connection.py
```diff
@@ -310,6 +329,89 @@ async def outbound_handler(self) -> None:
...
+            # TODO: actually throw one of the errors from errors.py and pass this to close
+            await self.close(API_EXCEPTION_BAN_SECONDS, WSCloseCode.PROTOCOL_ERROR, Err.UNKNOWN)
+        finally:
+            if task_id in self.api_tasks:
+                self.api_tasks.pop(task_id)
+            if task_id in self.execute_tasks:
+                self.execute_tasks.remove(task_id)
+
+    async def incoming_message_handler(self) -> None:
+        while True:
```
arvidn, line 409: so, with this patch there is no limit to the number of tasks we create. Is that the case currently also, prior to this task? (if not, this would be a regression)

arvidn (commented):
I agree with the general plan to move the queues into the peer connections, and also to make these changes incrementally to make it easier to review.

chia/server/ws_connection.py
xdustinface: There was no limit before. Tasks were just created out of the queue like they appeared.

arvidn (approved):

chia/server/ws_connection.py
wallentx: Resolving, as this appears to have been since approved.

chia/server/ws_connection.py
wallentx: Resolving, as this PR has since been approved.

chia/server/ws_connection.py
wallentx: Resolving, as this PR has since been approved, and comments have been added.

PR #14011, opened 2022-11-28
server: Drop unused `WSChiaConnection.is_feeler`

xdustinface:


arvidn:
I'm pretty sure this used to be used. at least in the crawler. Do you know of a recent patch making this unused?

xdustinface:
> I'm pretty sure this used to be used. at least in the crawler. Do you know of a recent patch making this unused?

Looking at 1.1.3 and 1.3.4 there was no usage of this either.

chia/server/server.py
```diff
@@ -480,7 +479,6 @@ async def start_client(
                 self._port,
                 self.log,
                 True,
-                False,
```
arvidn, line 483: I imagine the original intention was to pass `is_feeler` here

arvidn (approved):
I see. The parameter to `start_client()` is primarily used to initiate a `close()` task immediately after the connection is established. I imagine the original idea was for the connection to be treated differently in some cases if it was about to be disconnected, but it doesn't appear to be

PR #14012, opened 2022-11-28
Adding shell completion files

wallentx:
This PR adds the generated shell completion files which can be generated by the following commands:
`_CHIA_COMPLETE=bash_source chia`
`_CHIA_COMPLETE=zsh_source chia`
`_CHIA_COMPLETE=fish_source chia`

Currently, there is no documented mention of this being possible, as it may have been unknown (I stumbled upon this https://click.palletsprojects.com/en/8.1.x/shell-completion/#enabling-completion).

The user may source these files once their venv is activated.
For possible future tasks, we may want to do one (or many) of the following:
- For users who install from src, as a final step of `install.sh`, it may be desirable to inject a `source chia-complete.bash` into the relevant `venv/bin/activate` script, so that the completion is activated when the venv is activated
- When packaging the installers, we may want to unpack these to a standardized location (for example: `/usr/share/chia/chia-complete.bash` for bash)
- We may want to add a CLI command to generate this file, i.e.: `chia completion --bash`
- It may be best to simply document the availability of this feature, and perhaps recommend how this can get added to their dotfiles, or what not to do (going the `eval` route would lead to a sluggish experience)

https://user-images.githubusercontent.com/8990544/204255728-69109215-a314-47cb-8200-e697871bc1c6.mp4

### DIFF
--- a/chia-complete.bash
+++ b/chia-complete.bash
@@ -1,0 +1,29 @@
+_chia_completion() {
+    local IFS=$'\n'
+    local response
+
+    response=$(env COMP_WORDS="${COMP_WORDS[*]}" COMP_CWORD=$COMP_CWORD _CHIA_COMPLETE=bash_complete $1)
+
+    for completion in $response; do
+        IFS=',' read type value <<< "$completion"
+
+        if [[ $type == 'dir' ]]; then
+            COMPREPLY=()
+            compopt -o dirnames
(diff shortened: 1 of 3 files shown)


wallentx:
I've worked this into a CLI command in https://github.com/Chia-Network/chia-blockchain/pull/14782

PR #14015, opened 2022-11-28
Simplify the mempool manager instantiation by not making it depend on a coin store

AmineKhaldi:


chia/clvm/spend_sim.py
```diff
@@ -157,7 +158,7 @@ def new_coin_record(self, coin: Coin, coinbase: bool = False) -> CoinRecord:
 
     async def all_non_reward_coins(self) -> List[Coin]:
         coins = set()
-        async with self.mempool_manager.coin_store.db_wrapper.reader_no_transaction() as conn:
+        async with self.coin_store.db_wrapper.reader_no_transaction() as conn:
```
arvidn, line 161: you might as well just use `self.db_wrapper` directly here too, right?

chia/clvm/spend_sim.py
```diff
@@ -369,20 +366,16 @@ async def get_additions_and_removals(self, header_hash: bytes32) -> Tuple[List[C
             filter(lambda br: br.header_hash == header_hash, self.service.block_records)
         )[0]
         block_height: uint32 = selected_block.height
-        additions: List[CoinRecord] = await self.service.mempool_manager.coin_store.get_coins_added_at_height(
-            block_height
-        )  # noqa
-        removals: List[CoinRecord] = await self.service.mempool_manager.coin_store.get_coins_removed_at_height(
-            block_height
-        )  # noqa
+        additions: List[CoinRecord] = await self.service.coin_store.get_coins_added_at_height(block_height)  # noqa
```
arvidn, line 369: you could just use `self.coin_store` here too, right?
and below. Maybe that would let you remove the `# noqa`

chia/clvm/spend_sim.py
```diff
@@ -369,20 +366,16 @@ async def get_additions_and_removals(self, header_hash: bytes32) -> Tuple[List[C
...
+        removals: List[CoinRecord] = await self.service.coin_store.get_coins_removed_at_height(block_height)  # noqa
         return additions, removals
 
     async def get_puzzle_and_solution(self, coin_id: bytes32, height: uint32) -> Optional[CoinSpend]:
         filtered_generators = list(filter(lambda block: block.height == height, self.service.blocks))
         # real consideration should be made for the None cases instead of just hint ignoring
         generator: BlockGenerator = filtered_generators[0].transactions_generator  # type: ignore[assignment]
         coin_record: CoinRecord
-        coin_record = await self.service.mempool_manager.coin_store.get_coin_record(  # type: ignore[assignment]
+        coin_record = await self.service.coin_store.get_coin_record(  # type: ignore[assignment]
```
arvidn, line 378: it looks like you could just use `self.coin_store` here as well. Also, do you know why the variable type definition is a separate statement from the declaration?
Also, do you know why we suppress type checking on this line, it looks suspicious

arvidn (commented):
looks good. just some nit-picks

chia/clvm/spend_sim.py
AmineKhaldi: Yeap.

chia/clvm/spend_sim.py
AmineKhaldi: Same about `SimClient` and I replaced this suppression and type definition with just an assertion as the type is easily deducted (I can make it explicit if needed).

arvidn (approved):

chia/full_node/full_node.py
```diff
@@ -380,7 +380,7 @@ async def _start(self) -> None:
         )
 
         self._mempool_manager = MempoolManager(
-            coin_store=self.coin_store,
+            get_coin_record=self.coin_store.get_coin_record,
```
aqk, line 383: Fantastic. Another chunk of complexity removed from MempoolManager.

aqk (approved):
I like this change very much.

trepca (approved):

PR #14016, opened 2022-11-28
Add Wallet Action Manager (only to support payment <-> dl_inclusion offers currently)

Quexington:
This PR introduces the WalletActionManager which is a utility similar to the TradeManager but designed to be more generally applicable for any action a wallet can perform.  The general flow is as such:
1) User constructs a summary (see summary format below for what that entails)
2) User submits summary to wallet action manager which turns it into an unsigned & incomplete spend
3) User submits unsigned & incomplete spend to wallet for signing (currently this is still integrated into the wallet action manager, but this could be a separate airgapped machine for example)
4) User is returned a signed & incomplete spend
5) User submits signed & incomplete spend and an environment of variables (similar to a clvm solution) to wallet action manager which returns a signed and complete spend ready to be pushed to the network
6) User pushes spend to the network

The separation of transaction creation, signing, and completion allows for clients to construct more powerful applications without needing explicit support on the wallet backend.

### Backwards Compatibility
This PR is only designed to supply the underlying infrastructure, it does not yet expose the full functionality at the RPC level for users to use.  Instead, the trade manager calls methods from the wallet action manager in response to its normal calls like `/create_offer_for_ids`.  It leverages a library of functions to convert legacy requests/responses <-> action manager requests/responses.  Furthermore, this PR does not even use the underlying infrastructure for any supported use case, it only calls action manager methods in exactly the case where there is a DL singleton in the offer and also some other type of asset.  This means the new infrastructure only gets called to create/respond to a payment <-> dl_inclusion offer.  All other offers still use the same code paths in the trade manager.  The new summary format does not need to be used for pay-to-include offers, there is back compat code for converting requests and responses for those as well.

The idea is that in the future, we can do a slow cut-over to the new infrastructure and slowly (emphasis on slow) deprecate the trade manager and the offers it creates being careful to add unit tests for the old use cases in the new action manager.  While spends made in the new infrastructure may not be byte for byte similar to the same spend made on the old infrastructure, the intention is that new infrastructure spends will be able to be interpreted by old infrastructure clients.

Despite the following functionality not yet being available, here's a description of the intended interaction with the action manager when we do start to uncover it.

### Solvers
Many variables in this PR are typed with the class `Solver`.  This is a pre-existing type already in use in offer `driver_dict`s but it's poorly documented.  The intention of the class is to give a medium for users to specify exactly what they mean when they are referring to things handled in CLVM.  Rather than forcing literal CLVM for every call of an RPC (for example), `Solver` is kind of a middle ground in that it is a dictionary of strings mapped to one of three things:

1) Another dictionary
2) A list
3) A CLVM string

The python class operates similar to a dictionary where you can get an index of the object (i.e. `solver["puzzle_hash"]`) and behind the scenes, the class will interpret the CLVM and return to you its best guess of the python object that represents it.  For example `{"puzzle_hash": "0xcafef00d"}` will return you the equivalent of `bytes.fromhex("cafef00d")` whereas `{"puzzle_hash": "cafef00d"}` will return `"cafef00d"`.  All CLVM is represented as a string so things representing an integer would work like this: `{"amount": "1000"}`.

### Summary Format
```
{
  "actions" : [
    {
      "with": {*coin specification*},
      "do": [
        {*wallet action*},
        ...
      ],
    },
    ...
  ],
  "bundle_actions": [
    {*wallet action*}
  ],
  *additional keys*
}
```

The summary to build a spend is passed in as a CLVM dict similar to the `driver_dict` argument in `/create_offer_for_ids`.  Keys are strings, values are strings interpreted as CLVM.

The `"actions"` key is the main part of the summary.  Each action specifies a `"with"` key that contains a specification for the types and quantity of assets that should be selected in order to perform the spend.  For example `{"amount": "10"}` will select a plain XCH coin or coin(s) combining to at least 10 mojos.  `{"asset_id": "0xlauncher_id"}` will select the current generation of a singleton with the specified asset ID (no amount required).  Previous arguments to select_coin work as well like "min_coin_amount".  Each action also specifies a `"do"` key that is a list of arbitrary dictionaries each with a `"type"` key that specifies the action name so the action manager knows what parser to use.  An example action is `{"type": "offered_amount", "amount": "100"}` which will send 100 mojos to the offer puzzle hash allowing someone to claim it.

`"bundle_actions"` is simply a list of actions that have no `"with"` spec.  The idea is that the user does not care which coins complete the action, just that it get completed.  A `"request_payment"` action is currently the most common use case of this, or `"require_dl_inclusion"` for DL offers.

Additional keys can be specified that are additional instructions for how to construct the spend.  For example, the action manager currently will automatically add in a direct payment of the leftover amount to the local wallet as change.  However, this can be disabled with `"change": "()"`.

### Wallet Integration

This PR also introduces a few protocols that a wallet can implement in order to automatically integrate with the action manager:
* `OuterDriver` is a class whose instance can be used to spend coins with a specific outer puzzle or series of outer puzzles
* `InnerDriver` is a class whose instance can be used to spend coins with a specific inner puzzle
* `WalletAction` is a class that can parse an action from the request summary and that a `-Driver` class knows how to handle (right now there are only two of these for our standard inner puzzle: `Condition` and `Graftroot`).
* `ActionAlias` is a class that can parse an action from the request summary and then can digest that into a `WalletAction`.

### Securing Bundles with Annoucements

As good practice, spend bundles are generally made to be cohesive with announcements rather than relying on signatures alone to be inseparable.  In our standard wallet transactions, we secure with announcements by choosing an "origin coin" that creates all of the outputs and have it create an announcement that any coins supplying the amount can assert.  This assures that an attacker can never push the supply spends without the origin coin, however, it does not prevent them from pushing the origin coin without the supply spends.  This is fine for simple transactions, but from a general perspective, it does not satisfy the requirement that the bundle be cohesive where no spend can be pushed without the others.

The action manager makes the decision to be more general by having the coins form a ring of announcements (like how CATs do it).  This ensures that if one spend is omitted, the whole ring collapses and no spends are valid anymore.  This comes at the additional cost of ~32 bytes per coin spend, but also makes cancellation of multi asset spends cheaper since fewer coins have to be spent.  It also gives more flexibility to which coin can be spent to cancel the whole batch.

### New Chialisp

Reviewers may notice a few new chialisp files in the diff.  The `curry_*` files can be mostly ignored, they are utility functions and are not designed to hit the blockchain.  This leaves `add_wrapped_announcement.clsp` which is a new graftroot puzzle that does the functionality of requesting payments.  Using asset type specifications (covered in the next section) it can be solved with missing information in order to request payment of an asset type where some information must be concrete and some information is flexible.  For example, currently requesting an NFT requires that the NFT metadata be exactly the same as the time you requested it in order to fill the offer for that NFT.  It would be nicer if you could request an NFT with a specific launcher ID, but leave it to the taker to fill in what the metadata is at the time of taking.

The `add_wrapped_announcement.clsp` makes this possible by allowing some information to be curried in, but other information to be solved, and then it uses all of the information to build the puzzle hash of the asset it's expecting to hear from. Then, it adds a dynamic announcement assertion of a specific message from that puzzle hash.  This puzzle is not used in any current use cases as they all require complete knowledge of the puzzle hash at the time of making.  When a `RequestPayment` is parsed, it can see whether it has all of the information necessary to complete the puzzle hash and instead of using this graftroot puzzle with a pretty large overhead, it simplifies the puzzle to one that simply prepends the calculated announcement to the output of the inner delegated puzzle.

### Asset Type Specifications

One problem that arose while implementing this and thinking ahead to future use cases was a general way to describe the type of an asset.  Currently, we rely on asset ids and internal mappings of them to understand what asset the user intends to spend/request.  However, as described in the section above, what if we don't have all of the information?

The construct that was designed to solve for this is an "asset type spec" which consists of three pieces of information:
1) A "mod" or uncurried base puzzle with no environment
2) A "solution_template" which is a template of the environment, indicating which elements are concrete and which are flexible.  An example template for most puzzles would look like this: `(1 -1 0 . $)` which corresponds to the following chialisp `(mod (COMMITTED_ARG SOLVABLE_ARG INNER_PUZZLE . rest_of_the_solution) ...)`
3) "committed_args" which is a mirror of the solution template but with committed arguments where they should be (everything else is ignored so might as well use `()`).  Here's an example for the template above `(0xlauncher_id_or_something () () . ())`.

When it is time to turn an asset type into a puzzle hash, the above information will be used as well as another mirror of the solution template called "solution_args" which looks like a "committed_args" template but filling in all of `-1` spots instead.  Continuing with the example: `(() ("metadata" "and" "stuff") () . ())`

### This PR

- [x] Create WalletActionManager, move all relevant trade manager functions to it
- [x] Merge main in and resolve conflicts
- [x] Divide offer_request.py into action_manager_offer_backcompat.py and the rest -> wallet_action_manager.py
- [x] Get all new protocols into their own file and make them up to date with the wallets
- [x] Delete add_conditions.clsp
- [x] Bring CATWallet fully into conformity with `ActionProtocol`
- [x] Mypy
- [x] Add more comments

### Follow-ups
- [ ] Create `OuterDriver`s for remaining wallets (currently the standard wallet is the only inner wallet)
- [ ] Implement `TradeRecord` style tracking of spends
- [ ] Add RPCs for action manager
- [ ] Implement Minting for CATs

### Migration
- [ ] Bring other offer types over to new infrastructure (backwards compatibility unit tests!)
- [ ] Move other action-like RPCs to call action manager

### Lofty Long-Term Plans
- [ ] Allow > 1 outer wallet per coin (Recursive matching)
- [ ] Bring Singletons and NFT metadata layers out of NFT, DL, and DID wallets, make them their own wallet
- [ ] Refactor wallet state management and sync to work in similarly general ways with input from wallets
- [ ] Define a wallet protocol that once implemented automatically integrates fully with the whole infrastructure
- [ ] Create a plugin protocol


wallentx:
Snyk is broken at the moment. We should consider this as passing CI.

xdustinface (changes requested):
- I think you should start factor out changes into smaller PRs to get this properly reviewed
- This adds almost 4600 LOC and i only see a few things changed in tests. I didn't really look at it much so im not sure if that somehow magically tests all the new code :) If not i think there should be way more test coverage for a PR like this

PR #14023, opened 2022-11-29
Improve performance of wallets with many NFTs 

trepca:
**What is the purpose of the changes in this PR?**
It improves performance of wallets with many NFTs.

**What is the current behavior?**
Large wallets currently use excessive amounts of memory.

**What is the new behavior (if this is a feature change)?**
It decreases memory and CPU usage. 

**Does this PR introduce a breaking change?**
No.

### DIFF
--- a/chia/wallet/wallet_nft_store.py
+++ b/chia/wallet/wallet_nft_store.py
@@ -157,6 +157,19 @@ class WalletNftStore:
                 return int(count_row[0])
         return -1
 
+    async def is_empty(self, wallet_id: Optional[uint32] = None) -> bool:
+        sql = "SELECT 1 FROM users_nfts WHERE removed_height is NULL"
+        params: List[Union[uint32, bytes32]] = []
+        if wallet_id is not None:
+            sql += " AND wallet_id=?"
+            params.append(wallet_id)
+        sql += " LIMIT 1"
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            count_row = await execute_fetchone(conn, sql, params)
+            if count_row:
+                return False
+        return True
+
     async def get_nft_list(
         self, wallet_id: Optional[uint32] = None, did_id: Optional[bytes32] = None
     ) -> List[NFTCoinInfo]:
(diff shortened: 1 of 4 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -954,8 +954,8 @@ async def handle_nft(
                 assert isinstance(nft_wallet, NFTWallet)
                 if parent_coin_state.spent_height is not None:
                     await nft_wallet.remove_coin(coin_spend.coin, uint32(parent_coin_state.spent_height))
-                    num = await nft_wallet.get_current_nfts()
```
arvidn, line 957: wow... :/

chia/wallet/wallet_nft_store.py
```diff
@@ -157,6 +157,19 @@ async def count(self, wallet_id: Optional[uint32] = None, did_id: Optional[bytes
                 return int(count_row[0])
         return -1
 
+    async def is_empty(self, wallet_id: Optional[uint32] = None) -> bool:
+        sql = "SELECT 1 FROM users_nfts WHERE removed_height is NULL"
```
arvidn, line 161: does `1` refer to column 1?
would it make sense to query `COUNT(*)` instead? (still with `LIMIT 1`). It might be optimized and would require less data to be passed back from sqlite and the aiosqlite thread

chia/wallet/wallet_nft_store.py
```diff
@@ -157,6 +157,19 @@ async def count(self, wallet_id: Optional[uint32] = None, did_id: Optional[bytes
                 return int(count_row[0])
         return -1
 
+    async def is_empty(self, wallet_id: Optional[uint32] = None) -> bool:
```
arvidn, line 160: if we have a unit test for `wallet_nft_store`, we should probably have a few more cases to cover this new function

chia/wallet/wallet_nft_store.py
trepca: it returns `1` as a constant for each row that is returned, so it's more performant than `COUNT`

chia/wallet/wallet_nft_store.py
trepca: ah, yes

chia/wallet/wallet_nft_store.py
emlowe, line 161: `select exists(SELECT 1 FROM users_nfts WHERE removed_height is NULL)` returns either 0 or 1, unclear if it's any better but it is more typical

trepca: Thanks, I was trying to make it as simple as possible and to avoid nesting.

emlowe (approved):

PR #14025, opened 2022-11-29
enable memos in transactions

trepca:
**What is the purpose of the changes in this PR?**
To populate transactions with memos that were sent with coins. 

**What is the current behavior?**
The memos are not read from the blockchain so no memos are present in `Transaction` object.

**What is the new behavior (if this is a feature change)?**
Memos will appear when printing transactions in CLI and `Transaction` object will have memos filled. 

**Testing notes**
We should test if this has significant impact on performance and if there are cases where we could miss filling memos.

_I tested toilet wallet (on testnet) and memos make syncing about **3X SLOWER**_


Quexington (commented):
This looks like it is semantically fine, but the real question here is whether or not this slows down sync (Getting up to 2x the coin states & running CLVM in sync process) / drastically increases full node load (Lots more requests to fetch_puzzle_and_solution).  Have you done any benchmarking with results you can share?

trepca:
closing as impact to sync performance too big (we also found better ways to do this)

PR #14030, opened 2022-11-29
Add WalletStateManager.get_wallet() to centralize wallet type checking

altendky:
Just cleaning up the assertions that are strewn about.  They are inappropriate since they provide users with no messages and are often triggerable by user input via the RPCs.  This should also make the code a bit simpler and more consistent.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1677,11 +1677,7 @@ class WalletRpcApi:
 
     async def did_set_wallet_name(self, request) -> EndpointResult:
         wallet_id = uint32(request["wallet_id"])
-        wallet = self.service.wallet_state_manager.get_wallet(
-            id=wallet_id,
-            required_type=DIDWallet,
-            message_format="Wallet id {id} is not a DID wallet",
-        )
+        wallet = self.service.wallet_state_manager.get_wallet(id=wallet_id, required_type=DIDWallet)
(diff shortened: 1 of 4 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1428,8 +1420,7 @@ async def cat_spend(self, request, hold_lock=True) -> EndpointResult:
 
     async def cat_get_asset_id(self, request) -> EndpointResult:
         wallet_id = uint32(request["wallet_id"])
-        wallet = self.service.wallet_state_manager.wallets[wallet_id]
-        assert isinstance(wallet, CATWallet)
+        wallet = self.service.wallet_state_manager.get_wallet(id=wallet_id, required_type=CATWallet)
```
arvidn, line 1423: does mypy understand that `wallet` is of type `CATWallet` here?

chia/wallet/wallet_state_manager.py
```diff
@@ -269,6 +271,15 @@ async def get_keys(self, puzzle_hash: bytes32) -> Optional[Tuple[G1Element, Priv
         pubkey = private.get_g1()
         return pubkey, private
 
+    def get_wallet(self, id: uint32, required_type: Type[TWalletType]) -> TWalletType:
```
arvidn, line 274: I see

arvidn (approved):

PR #14035, opened 2022-11-30
Remove redundant checks in validate_spend_bundle()

AmineKhaldi:
This simplifies the mempool by removing `COIN_AMOUNT_NEGATIVE` and `COIN_AMOUNT_EXCEEDS_MAXIMUM` checks that are already performed on the Rust side.
The Rust side currently emits `Err.INVALID_CONDITION` for these checks. An exploration to make it return more appropriate values is in https://github.com/Chia-Network/chia_rs/pull/99
The added tests pass before and after introducing the code changes.

### DIFF
--- a/tests/core/mempool/test_mempool_manager.py
+++ b/tests/core/mempool/test_mempool_manager.py
@@ -6,7 +6,10 @@ import pytest
 from blspy import G2Element
 
 from chia.consensus.block_record import BlockRecord
+from chia.consensus.cost_calculator import NPCResult
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
+from chia.full_node.bundle_tools import simple_solution_generator
+from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions
 from chia.full_node.mempool_manager import MempoolManager
 from chia.types.blockchain_format.classgroup import ClassgroupElement
 from chia.types.blockchain_format.coin import Coin
@@ -16,7 +19,7 @@ from chia.types.coin_record import CoinRecord
 from chia.types.coin_spend import CoinSpend
 from chia.types.condition_opcodes import ConditionOpcode
 from chia.types.spend_bundle import SpendBundle
-from chia.util.errors import ValidationError
+from chia.util.errors import Err, ValidationError
 from chia.util.ints import uint8, uint32, uint64, uint128
 
 IDENTITY_PUZZLE = Program.to(1)
@@ -72,6 +75,13 @@ def spend_bundle_from_conditions(conditions: List[List[Any]]) -> SpendBundle:
     return SpendBundle([coin_spend], G2Element())
 
 
+def get_name_puzzle_conditions_on_spendbundle(mempool_manager: MempoolManager, sb: SpendBundle) -> NPCResult:
+    program = simple_solution_generator(sb)
+    max_cost = int(mempool_manager.limit_factor * mempool_manager.constants.MAX_BLOCK_COST_CLVM)
+    cost_per_byte = mempool_manager.constants.COST_PER_BYTE
+    return get_name_puzzle_conditions(program, max_cost=max_cost, cost_per_byte=cost_per_byte, mempool_mode=True)
+
+
 @pytest.mark.asyncio
 async def test_negative_addition_amount() -> None:
     mempool_manager = await instantiate_mempool_manager(zero_calls_get_coin_record)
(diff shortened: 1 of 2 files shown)


tests/core/mempool/test_mempool_manager.py
```diff
@@ -0,0 +1,100 @@
...
+    return SpendBundle([coin_spend], G2Element())
+
+
+@pytest.mark.asyncio
+async def test_addition_amount(mempool_manager: MempoolManager) -> None:
+    # Negative amount
+    conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, -1]]
+    sb = spend_bundle_from_conditions(conditions)
+    sb_name = sb.name()
+    with pytest.raises(ValidationError, match="Err.INVALID_CONDITION"):
```
arvidn, line 88: it's not obvious why this would trigger an `INVALID_CONDITION`. Shouldn't it be `COIN_AMOUNT_NEGATIVE`?

tests/core/mempool/test_mempool_manager.py
```diff
@@ -0,0 +1,100 @@
...
+    uint64(0),
+    None,
+    None,
+    None,
+    None,
+    None,
+)
+
+
+async def get_coin_record(_: bytes32) -> Optional[CoinRecord]:
```
arvidn, line 59: I think it would be great if each test ensured that this callback was called with the expected coin ID as well, and not just ignored

tests/core/mempool/test_mempool_manager.py
```diff
@@ -0,0 +1,100 @@
...
+    # Big but valid amount
+    max_amount = mempool_manager.constants.MAX_COIN_AMOUNT
+    conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, max_amount]]
+    sb = spend_bundle_from_conditions(conditions)
+    npc_result = await mempool_manager.pre_validate_spendbundle(sb, None, sb.name())
+    assert npc_result.error is None
+    # Too big amount
+    conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, max_amount + 1]]
+    sb = spend_bundle_from_conditions(conditions)
+    with pytest.raises(ValidationError, match="Err.INVALID_CONDITION"):
```
arvidn, line 99: likewise, how come this error isn't `COIN_AMOUNT_EXCEEDS_MAXIMUM`?

arvidn (commented):
would you mind splitting up the test into multiple ones, so each test only tests one thing?
i.e. one test for negative coin amount, one positive test, one for coin-amount-exceeds-maximum.

For these specific cases you might be able to parameterize the test too, by the input condition and the expected error

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: I'm exploring that through https://github.com/Chia-Network/chia_rs/pull/97

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: I created a record for this test coin and made this callback return it from a dictionary of records.

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Exploring this one also in https://github.com/Chia-Network/chia_rs/pull/97

arvidn:
you should add comments next to the unexpected error codes and explain why they're wrong. Then we can fix them once we've cut a new `chia_rs` release

tests/core/mempool/test_mempool_manager.py
```diff
@@ -0,0 +1,112 @@
...
+IDENTITY_PUZZLE = Program.to(1)
+IDENTITY_PUZZLE_HASH = IDENTITY_PUZZLE.get_tree_hash()
+
+TEST_TIMESTAMP = uint64(1616108400)
+TEST_COIN = Coin(IDENTITY_PUZZLE_HASH, IDENTITY_PUZZLE_HASH, uint64(1000000000))
+TEST_COIN_ID = TEST_COIN.name()
+TEST_COIN_RECORD = CoinRecord(TEST_COIN, uint32(0), uint32(0), False, TEST_TIMESTAMP)
+TEST_COIN_RECORDS = {TEST_COIN_ID: TEST_COIN_RECORD}
+TEST_HEIGHT = uint32(1)
+TEST_BLOCK_RECORD = BlockRecord(
```
arvidn, line 33: do you anticipate that future tests will need access to this block record? If not, I think it should be moved into the fixture function

arvidn (commented):
I think you should also test that the `get_coin_record` callback is called exactly the number of times you'd expect and with the coin IDs you would expect

tests/core/mempool/test_mempool_manager.py

arvidn (dismissed):

arvidn (approved):

arvidn (approved):

arvidn:
@AmineKhaldi I don't think this PR should be mentioned in the changelog. you need another label for that

PR #14042, opened 2022-12-01
Stop creating unecessary index in notification store

### LINKED ISSUE #13955: Wallet - Index `coin_id_index` for column `coin_id` on `notifications` table is redundant
The columns used in a Primary Key are automatically indexed, in this case for column `coin_id`.

So this create index statement can be removed:
https://github.com/Chia-Network/chia-blockchain/blob/22a1d1b1c20c8360d24ec1c87d926e2c0d2f3459/chia/wallet/notification_store.py#L47

To remove the redundant index add:
```sql
DROP INDEX IF EXISTS coin_id_index;
```

To verify that `sqlite_autoindex_notifications_1` is the automatically created index for coin_id:
(my wallets notifications table is empty, so I used a random number for the selection criteria)
```sql
sqlite> select * from sqlite_master where tbl_name = 'notifications';
type   name                              tbl_name       rootpage  sql                                                        
-----  --------------------------------  -------------  --------  -------------------------------------------------------------------------
table  notifications                     notifications  834       CREATE TABLE notifications(coin_id blob PRIMARY KEY,msg blob,amount blob)
index  sqlite_autoindex_notifications_1  notifications  835                                                                  
index  coin_id_index                     notifications  836       CREATE INDEX coin_id_index on notifications(coin_id)     
  
sqlite> drop index coin_id_index;

sqlite> .eqp on

sqlite> select * from notifications where coin_id=2;
QUERY PLAN
`--SEARCH TABLE notifications USING INDEX sqlite_autoindex_notifications_1 (coin_id=?)

```

Quexington:


chia/wallet/notification_store.py
```diff
@@ -44,7 +44,7 @@ async def create(
                 "CREATE TABLE IF NOT EXISTS notifications(" "coin_id blob PRIMARY KEY," "msg blob," "amount blob" ")"
             )
 
-            await conn.execute("CREATE INDEX IF NOT EXISTS coin_id_index on notifications(coin_id)")
+            await conn.execute("DROP INDEX IF EXISTS coin_id_index")
```
arvidn, line 47: this is redundant because `coin_id` is already the `PRIMARY KEY` of the table, and has an implicit index.
I think it would make sense to leave a comment here explaining the history a bit, and maybe in which version we switched from creating this index to dropping it. because at some point in the future we may want to remove this line entirely

arvidn (dismissed):

arvidn (approved):

chia/wallet/notification_store.py
wallentx: Comment added

PR #14049, opened 2022-12-03
also upload the `-cli` `.deb` and `.rpm` packages to GitHub releases

altendky:
Totally untested.  Just providing a concrete thing to move on.


cmmarslender (dismissed):

cmmarslender (approved):

emlowe (approved):
Seems ok to add in an upload step for these, since we already have the artifact.

PR #14052, opened 2022-12-05
server: Replace `WSChiaConnection.__getattr__` with `call_api`

xdustinface:
Make API calls obvious and let `mypy` detect access to non-existing attributes in `WSChiaConnection`.

Based on:

- #14051

### DIFF
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -9,7 +9,7 @@ import traceback
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime, timezone
 from secrets import token_bytes
-from typing import Dict, List, Optional, Set, Tuple
+from typing import TYPE_CHECKING, Dict, List, Optional, Set, Tuple
 
 from blspy import AugSchemeMPL, G1Element, G2Element
 from chiabip158 import PyBIP158
@@ -20,7 +20,6 @@ from chia.consensus.pot_iterations import calculate_ip_iters, calculate_iteratio
 from chia.full_node.bundle_tools import best_solution_generator_from_template, simple_solution_generator
 from chia.full_node.fee_estimate import FeeEstimate, FeeEstimateGroup
 from chia.full_node.fee_estimator_interface import FeeEstimatorInterface
-from chia.full_node.full_node import FullNode
 from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions, get_puzzle_and_solution_for_coin
 from chia.full_node.signage_point import SignagePoint
 from chia.full_node.tx_processing_queue import TransactionQueueFull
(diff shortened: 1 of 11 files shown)


chia/util/api_decorators.py
```diff
@@ -71,6 +72,7 @@ def wrapper(self: Self, original: Union[bytes, S], *args: P.args, **kwargs: P.kw
         message_name_bytes = f"{message_name}_bytes"
 
         metadata = ApiMetadata(
+            endpoint=f.__name__,
```
xdustinface, line 75: I tried to somehow get the name of the decorated methods passed into `call_api`'s `request_method` parameter but i couldn't find a reasonable way so i ended up with `ApiMetadata.endpoint`. We get the wrapped method there so a direct `__name__` access would only result in `wrapper` as name since we get the below:

```
<function chia.util.api_decorators.api_request.<locals>.inner.<locals>.wrapper(self: 'Self', original: 'Union[bytes, S]', *args: 'P.args', **kwargs: 'P.kwargs') -> 'R'>
```

chia/full_node/full_node_api.py
```diff
@@ -61,6 +60,11 @@
 from chia.util.limited_semaphore import LimitedSemaphoreFullError
 from chia.util.merkle_set import MerkleSet
 
+if TYPE_CHECKING:
+    from chia.full_node.full_node import FullNode
```
arvidn, line 64: would you mind leaving a comment explaining what this is for?
Is it to avoid cyclical dependencies, or just to keep unnecessary imports out of production?

tests/plot_sync/test_receiver.py
```diff
@@ -78,7 +78,7 @@ def create_payload(payload_type: Any, start: bool, *args: Any) -> Any:
 def assert_error_response(plot_sync: Receiver, error_code: ErrorCodes) -> None:
     connection = plot_sync.connection()
     assert connection is not None
-    message = connection.last_sent_message
+    message = connection.last_sent_message  # type: ignore[attr-defined]
```
arvidn, line 81: I don't understand why you need the mypy suppression here. If there's a good reason, and it needs to stay there, it should probably be a comment in the code

arvidn (commented):
this is a great improvement!

chia/full_node/full_node_api.py
altendky: I would suggest that this should be a matter of general practice and not a thing that is commented each place.  I generally only do this for known cyclic imports.  Could make a pr to the pretty good practices doc if we want to 'formalize' this.

arvidn: mypy doesn't mind circular dependencies then I take it?

altendky: Nope.

tests/plot_sync/test_receiver.py
xdustinface: Added!

arvidn (approved):

PR #14057, opened 2022-12-05
Fix did test

ytx1991:
Fix test_get_info


arvidn:
do you have a theory of why this is taking longer than it used to? or has this test never worked reliably?

PR #14058, opened 2022-12-06
add typehints to MempoolManager members

arvidn:


AmineKhaldi (dismissed):

chia/full_node/mempool_manager.py
```diff
@@ -96,7 +106,7 @@ def __init__(
         # Keep track of seen spend_bundles
         self.seen_bundle_hashes: Dict[bytes32, bytes32] = {}
 
-        self.get_coin_record = get_coin_record
+        setattr(self, "get_coin_record", get_coin_record)
```
xdustinface, line 109: Whats the reason for the `setattr` and `getattr` below?

arvidn: mypy doesn't like to assign function objects. it complaints that:
```
self.get_coin_record = get_coin_record
```

is a malformed attempt to *call* `self.get_coin_record`.

Similarly, trying to call this function with this syntax:
```
self.get_coin_record(coin_id)
```

fails (at least according to mypy) because it looks like a member function and it passes in `self` as the implied first argument, which is an argument mismatch. I tried like this:
```
(self.get_coin_record)(coin_id)
```
But that also did not work.

xdustinface: Hm, this doesn't look different to any other callback we assign in other places. I guess what you mean happens if you `def` it? It passes here with

```
diff --git a/chia-blockchain-gui b/chia-blockchain-gui
index 83bbca6ff..a6724910a 160000
--- a/chia-blockchain-gui
+++ b/chia-blockchain-gui
@@ -1 +1 @@
-Subproject commit 83bbca6ff2c425792eb738279ab3876a00322474
+Subproject commit a6724910a82e1e719784a7a0885da13b30533c66-dirty
diff --git a/chia/full_node/mempool_manager.py b/chia/full_node/mempool_manager.py
index 4c597c0e2..aed0fb5dc 100644
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -106,7 +106,7 @@ class MempoolManager:
         # Keep track of seen spend_bundles
         self.seen_bundle_hashes: Dict[bytes32, bytes32] = {}
 
-        setattr(self, "get_coin_record", get_coin_record)
+        self.get_coin_record = get_coin_record
 
         # The fee per cost must be above this amount to consider the fee "nonzero", and thus able to kick out other
         # transactions. This prevents spam. This is equivalent to 0.055 XCH per block, or about 0.00005 XCH for two
@@ -432,7 +432,7 @@ class MempoolManager:
         removal_record_dict: Dict[bytes32, CoinRecord] = {}
         removal_amount: int = 0
         for name in removal_names:
-            removal_record = await getattr(self, "get_coin_record")(name)
+            removal_record = await self.get_coin_record(name)
             if removal_record is None and name not in additions_dict:
                 return Err.UNKNOWN_UNSPENT, None, []
             elif name in additions_dict:
```

arvidn: it failed with the message I copy-pasted on CI. But now it seems the previous commits may have been garbage collected by github, so I can't find it

arvidn: Locally, for me, I also get these errors, that are similar:
```
chia/plotting/manager.py:67: error: Cannot assign to a method  [assignment]
chia/plotting/manager.py:86: error: Cannot assign to a method  [assignment]
```

arvidn: but that file is exempt in `mypy.ini`. mempool_manager.py is not

xdustinface: So, this here https://mypy-play.net/?mypy=0.971&python=3.11&gist=39dddce541e89f7fbb827bf5ee2375d8 fails for `mypy` versions <= 0.971 but passes for all above.

arvidn:
```
$ mypy --version
mypy 0.971 (compiled: yes)
```

arvidn: what matters, though, is what version we have in CI

xdustinface: Thats always the latest afaik?

arvidn: I don't know why it was failing earlier. It just passed on CI in a test. I hope the version of mypy is not flaky.

xdustinface (approved):

paninaro (approved):

PR #14063, opened 2022-12-07
Add type hinting to Mempool

aqk:
No user-visible changes. Code maintenance.


chia/types/spend_bundle.py
```diff
@@ -59,6 +60,10 @@ def fees(self) -> int:
 
         return amount_in - amount_out
 
+    @property
+    def id(self) -> SpendBundleID:
```
aqk, line 64: Take a look at this line that adds a new `SpendBundleID` property to `SpendBundle`

chia/types/borderlands.py
```diff
@@ -0,0 +1,41 @@
...
+PrivateKeyBytes = NewType("PrivateKeyBytes", bytes32)
+PublicKeyBytes = NewType("PublicKeyBytes", bytes48)
+SignatureBytes = NewType("SignatureBytes", bytes96)
+PuzzleHash = NewType("PuzzleHash", bytes32)
+CoinID = NewType("CoinID", bytes32)
+SpendBundleID = NewType("SpendBundleID", bytes32)  # Also used as MempoolItemID
+BlockRecordHeaderHash = NewType("BlockRecordHeaderHash", bytes32)
+
+
+def bytes_to_PrivateKeyBytes(input_bytes: bytes) -> PrivateKeyBytes:
```
aqk, line 16: I had a big comment in here earlier - I should probably add that back.

chia/full_node/mempool_manager.py
```diff
@@ -22,6 +22,14 @@
 from chia.full_node.pending_tx_cache import PendingTxCache
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.sized_bytes import bytes32, bytes48
+from chia.types.borderlands import (
+    BlockRecordHeaderHash,
```
aqk, line 26: mempool_manager.py was the main motivation to add these types differentiating various `byte32` types.

chia/types/borderlands.py
```diff
@@ -0,0 +1,41 @@
...
+    return PrivateKeyBytes(bytes32(input_bytes))
+
+
+def bytes_to_PublicKeyBytes(input_bytes: bytes) -> PublicKeyBytes:
```
aqk, line 20: `bytes_to_PrivateKeyBytes` etc are needed to avoid importing `bytes32` and then writing `PrivateKeyBytes(bytes32(foo))`.

chia/types/spend_bundle.py
aqk: `SpendBundle.name()` has many many uses in many many files.

chia/types/spend_bundle.py
```diff
@@ -59,6 +60,10 @@ def fees(self) -> int:
...
+        return bytes_to_SpendBundleID(self.get_hash())
+
     def name(self) -> bytes32:
```
aqk, line 67: `SpendBundle.name()` has many many uses in many many files.

chia/types/borderlands.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+from typing import NewType
+
+from chia.types.blockchain_format.sized_bytes import bytes32, bytes48, bytes96
+
+PrivateKeyBytes = NewType("PrivateKeyBytes", bytes32)
```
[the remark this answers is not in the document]
arvidn: I don't see this type used anywhere (yet). I would have expected it to be called `SecretKeyBytes`. I think it's a bit risky to introduce the type separate from transitioning to using it.

tests/clvm/coin_store.py
```diff
@@ -69,12 +70,15 @@ def validate_spend_bundle(
         if result.error is not None:
             raise BadSpendBundleError(f"condition validation failure {Err(result.error)}")
 
-        ephemeral_db = dict(self._db)
+        ephemeral_db = {}
+        for k, v in self._db.items():
+            ephemeral_db[bytes_to_CoinID(k)] = v
```
arvidn, line 75:
```suggestion
        ephemeral_db = {bytes_to_CoinID(k): v for k, v in self._db.items()}
```

arvidn: would that work?

chia/types/borderlands.py
arvidn: I would hope there would be a way to "convert" types here without actually cloning/re-allocating the bytes in memory. Do you know whether this has a run-time cost or if it's just a mypy check conversion?

chia/types/spend_bundle.py
arvidn: generally I think it's a mistake to add properties that hare costly to compute. We (used to) compute header hashes and coin ids unnecessarily, without the code making it look like it, because of this.
I think it's reasonable to make it visible at the call-site that computing the spend bundle ID is not free, but is in fact a function call.

tests/clvm/coin_store.py
arvidn: did you explore changing the type of `self._db` instead?

chia/types/spend_bundle.py
arvidn: In the long run, I imagine we'd want to make all those places aware of any type casting they do too

chia/full_node/full_node.py
```diff
@@ -2216,45 +2221,48 @@ async def respond_transaction(
         if not test and not (await self.synced()):
             return MempoolInclusionStatus.FAILED, Err.NO_TRANSACTIONS_WHILE_SYNCING
 
+        spend_bundle_id = bytes_to_SpendBundleID(spend_name)
```
arvidn, line 2224: why not change the type of `spend_name` instead?

chia/full_node/full_node.py
```diff
@@ -2216,45 +2221,48 @@ async def respond_transaction(
...
             except ValidationError as e:
-                self.mempool_manager.remove_seen(spend_name)
+                self.mempool_manager.remove_seen(spend_bundle_id)
                 return MempoolInclusionStatus.FAILED, e.code
             except Exception:
-                self.mempool_manager.remove_seen(spend_name)
+                self.mempool_manager.remove_seen(spend_bundle_id)
                 raise
             async with self._blockchain_lock_low_priority:
                 if self.mempool_manager.get_spendbundle(spend_name) is not None:
```
arvidn, line 2249: it would seem reasonable for `get_spendbundle()` to take a `SpendBundleID` too

arvidn (commented):
I really appreciate this direction of more type checks. However, renaming of variables, and copying them to have them represented with different types makes me nervous. It's not obvious semantics aren't changing. Perhaps it would make sense to split this up into smaller pieces, where single types are introduced (but then also propagated throughout the code base), minimizing the number of places where you have to convert back and forth with the existing `Bytes32`

chia/full_node/full_node.py
aqk: > why not change the type of spend_name instead?

Because `respond_transaction` is near the interface to the one-the-wire protocol. More thought has to be put in before changing types up that call chain, so I figured this is a good place to cut the type propagation for now.

chia/full_node/full_node.py
aqk: I did start doing that, but `get_spendbundle()` is called about 236 times, so that one has to wait.

tests/clvm/coin_store.py
aqk: >    ephemeral_db = {bytes_to_CoinID(k): v for k, v in self._db.items()}

Hey nice, that works just like a list comprehension! Sweet.

aqk: > did you explore changing the type of `self._db` instead?

Since this is `tests.clvm.coin_store` and not `chia.full_node.coin_store`, I will definitely take this suggestion, thanks.

aqk:
> I really appreciate this direction of more type checks. However, renaming of variables, and copying them to have them represented with different types makes me nervous. It's not obvious semantics aren't changing. Perhaps it would make sense to split this up into smaller pieces, where single types are introduced (but then also propagated throughout the code base), minimizing the number of places where you have to convert back and forth with the existing `Bytes32`

I could split this up if it makes it easier to review. I don't think there's "back-and-forth" conversion, though. Well, except in `get_spendbundle` - that change was getting way too big. If there is, I need to change this patch - once we are converted into these types, we should stay in there. Of course, that's not possible unless we change all the code at the same time.

The main reason I'm not as worried about this change is my current understanding of NewType:

```
    # Replacing int with Y will have no effect at runtime (except a small performance penalty)
    Y = NewType("Y", int)

    n : int = 5
    m: Y = 5
    # m & n should act the same at runtime.
```

aqk:
By the way, guys - I don't expect all of you to review it, but I wanted to tag anyone who might have a need to know. @xdustinface

chia/types/borderlands.py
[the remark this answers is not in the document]
aqk: Ya, this is leftover. Removed.

chia/types/borderlands.py
aqk: I think there is a small runtime cost, but I don't know if the inner object is wrapped at runtime.
I don't think it is, because the checks are definitely not performed at runtime. (The typecheck-time and runtime "instantiations" are totally different implementations)

chia/types/spend_bundle.py
aqk: Ya, I can reverse this property, and make it a method call.

aqk:
Okay, I'll split this into 3 PRs at Arvid's request.

aqk:
Moved to the following PRs:

https://github.com/Chia-Network/chia-blockchain/pull/14074
https://github.com/Chia-Network/chia-blockchain/pull/14076
https://github.com/Chia-Network/chia-blockchain/pull/14077
https://github.com/Chia-Network/chia-blockchain/pull/14078
https://github.com/Chia-Network/chia-blockchain/pull/14079
https://github.com/Chia-Network/chia-blockchain/pull/14080
https://github.com/Chia-Network/chia-blockchain/pull/14084

PR #14066, opened 2022-12-07
Deprecate `curry-and-treehash` and replace it with `curry`.

richardkiss:
See title.

### DIFF
--- a/chia/wallet/puzzles/curry-and-treehash.clinc
+++ b/chia/wallet/puzzles/curry-and-treehash.clinc
@@ -1,3 +1,8 @@
+;;
+;; THIS FILE IS DEPRECATED and is only for legacy contracts.
+;; Please use `curry.clinc` going forward. It has a less stupid API.
+;;
+
 (
   ;; The code below is used to calculate of the tree hash of a curried function
   ;; without actually doing the curry, and using other optimization tricks
@@ -49,6 +54,10 @@
                              (sha256 TWO environment-hash (sha256 ONE 0))))
   )
 
+  ;; DO NOT USE THIS FUNCTION GOING FORWARD. Having to pass the arguments in reverse order is stupid.
+  ;; When I (RK) wrote this, I thought it was necessary, but of course, I was just being dumb
+  ;; and it is not necessary. It's very confusing.
+  ;;
   ;; function-hash:
   ;;   the hash of a puzzle function, ie. a `mod`
   ;;
@@ -65,6 +74,7 @@
   ;; the hashes of the desired parameters, but their treehash representation since
   ;; that's the form we're assuming they take in the actual curried program.
 
+  ;; DO NOT USE
   (defun puzzle-hash-of-curried-function (function-hash . reversed-curry-parameter-hashes)
      (tree-hash-of-apply function-hash
                          (build-curry-list reversed-curry-parameter-hashes (sha256 ONE ONE)))
--- a/chia/wallet/puzzles/curry-and-treehash.clinc
+++ b/chia/wallet/puzzles/curry-and-treehash.clinc
@@ -1,6 +1,6 @@
 ;;
 ;; THIS FILE IS DEPRECATED and is only for legacy contracts.
-;; Please use `curry.clinc` going forward. It has a less stupid API.
+;; Please use `curry.clib` going forward. It has a less stupid API.
 ;;
 
 (
--- a/chia/wallet/puzzles/curry.clib
+++ b/chia/wallet/puzzles/curry.clib
@@ -4,7 +4,7 @@
   ;; like unrolling `sha256tree`.
 
   (defconstant TWO 2)
-  (defconstant constant-tree (
+  (defconstant constant_tree (
      (0x4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a .   ;  = `(sha256 1)`
       0x9dcf97a184f32623d11a73124ceb99a5709b083721e878a16d78f596718ba7b2) .  ;  = `(sha256 1 1)` = `(sha256 1 #q)`
      (0x02a12871fee210fb8619291eaea194581cbd2531e4b23759d225f6806923f63222 . ;  = `(concat 2 (sha256 1 #a))`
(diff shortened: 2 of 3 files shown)


chia/wallet/puzzles/curry-and-treehash.clinc
```diff
@@ -66,6 +75,7 @@
   ;; that's the form we're assuming they take in the actual curried program.
 
   (defun puzzle-hash-of-curried-function (function-hash . reversed-curry-parameter-hashes)
+     ;; DO NOT USE THIS FUNCTION. Use
```
arvidn, line 78: this comment seems truncated

Quexington (changes requested):
Can we also call it `.clib` instead of `.clinc`, I think that's what we're standardizing on.

chia/wallet/puzzles/curry.clinc
```diff
@@ -0,0 +1,68 @@
...
+  ;;   a list of pre-hashed trees representing parameters to be curried into the puzzle.
+	;;
+  ;; we return the hash of the curried expression
+  ;;   (a (q . mod-hash) (c (cp1 (c cp2 (c ... 1)...))))
+  ;;
+  ;; Note that from a user's perspective the hashes passed in here aren't simply
+  ;; the hashes of the desired parameters, but their treehash representation since
+  ;; that's the form we're assuming they take in the acutal curried program.
+
+  (defun curry (mod-hash . curry-parameter-hashes)
```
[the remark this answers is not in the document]
richardkiss: Done

chia/wallet/puzzles/curry-and-treehash.clinc
richardkiss: Fixed

chia/wallet/puzzles/curry.clinc
```diff
@@ -0,0 +1,68 @@
...
+  ;; The code below is used to calculate of the tree hash of a curried function
+  ;; without actually doing the curry, and using other optimization tricks
+  ;; like unrolling `sha256tree`.
+
+  (defconstant ONE 1)
+  (defconstant TWO 2)
+  (defconstant A_KW #a)
+  (defconstant Q_KW #q)
+  (defconstant C_KW #c)
+	(defconstant sha256_one 0x4bf5122f344554c53bde2ebb8cd2b7e3d1600ad631c385a5d7cce23c7785459a)
```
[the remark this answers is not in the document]
richardkiss: So writing `(sha 1)` in multiple places would expand out to the constant. I figured since I can't keep the long constant out of the code, I may as well just include it once, as a `defconstant`.

chia/wallet/puzzles/curry.clinc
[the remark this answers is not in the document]
richardkiss: I undid this because it doesn't seem to work (probably due to a shortcoming in the compiler, although I'm using the old python compiler).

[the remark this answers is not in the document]
richardkiss: I'll add a `curry2` function which takes exactly two parameters (instead of "varargs" style), and then it should be inline-able.

chia/wallet/puzzles/curry-and-treehash.clinc
```diff
@@ -1,3 +1,8 @@
+;;
+;; THIS FILE IS DEPRECATED and is only for legacy contracts.
+;; Please use `curry.clinc` going forward. It has a less stupid API.
```
Quexington, line 3: ~clinc~ clib

chia/wallet/puzzles/curry.clib
```diff
@@ -0,0 +1,95 @@
...
+  ;;   a list of pre-hashed trees representing parameters to be curried into the puzzle.
+  ;;
+  ;; we return the hash of the curried expression
+  ;;   (a (q . mod-hash) (c (cp1 (c cp2 (c ... 1)...))))
+  ;;
+  ;; Note that from a user's perspective the hashes passed in here aren't simply
+  ;; the hashes of the desired parameters, but their treehash representation since
+  ;; that's the form we're assuming they take in the acutal curried program.
+
+  ;; inline functions that take varargs don't seem to work, so we can't inline `curry`
```
Quexington, line 82: You could inline it if you went `(defun curry (mod_hash_and_param_hashes)` then just use `(f` and `(r` to access the two args that are supposed to be there.  I'm pretty sure inlining will end up making a significant enough decrease in reveal size that it's worth it.

chia/wallet/puzzles/curry.clib
```diff
@@ -0,0 +1,95 @@
...
+
+  (defun curry (mod-hash . curry-parameter-hashes)
```
Quexington, line 84: should we call it `curry_treehash` or something since it specifically only handles the already hashed branches?

chia/wallet/puzzles/curry.clib
richardkiss: I tried

```
  (defun-inline curry mod-hash-and-curry-parameter-hashes
     (tree-hash-of-apply (f mod-hash-and-curry-parameter-hashes)
                         (calculate-hash-of-curried-parameters (r mod-hash-and-curry-parameter-hashes)))
  )
```

and I can't get this to work either, likely because of compiler shortcomings.

Quexington: Have you tried in a file that `(include *standard-cl-21*)` is in?  It might be fine if it only works there since we should be using that going forward anyhow.  But otherwise yeah I don't see a way out.

chia/wallet/puzzles/curry.clib
richardkiss: How about `curry_hashes` (non-inlined) and `curry_hashes_inline` (inlined)?

Quexington (dismissed):

arvidn:
do we not have tests for these functions?

Quexington (approved):

richardkiss:
> do we not have tests for these functions?

There are no tests for any of this. But I've tried it and it seems to give the same results as the thing it replaces (except using more sane parameter-passing standards).

trepca (approved):

PR #14067, opened 2022-12-07
tests: Use the `self_hostname` fixture in more places

xdustinface:


arvidn (approved):
I'm a bit skeptical of the value of making "localhost" configurable. Virtually all of these tests want to just connect to some service the test itself has stood up, they're not supposed to *also* test that we can replace "localhost" with "127.0.0.1" or "::1". If we want to test that, we should have dedicated tests for that (in my opinion).

PR #14068, opened 2022-12-07
util: Rework `get_host_addr`

xdustinface:
Start of a series of PR's working towards using

```
IPAddress = Union[IPv4Address, IPv6Address]
```

wherever we use string hosts except in some places where they are needed because it can be a domain also. One example would be in `PeerInfo` to use `address: IPAddress` instead of `host: str` which is a big example since it will require a bunch of other changes and will also be done in few steps but i think at the end if this is all done things will be much cleaner.

### DIFF
--- a/chia/timelord/timelord_launcher.py
+++ b/chia/timelord/timelord_launcher.py
@@ -4,7 +4,7 @@ import pathlib
 import signal
 import time
 import os
-from typing import Dict, List, Optional
+from typing import Dict, List
 
 import pkg_resources
 
@@ -39,7 +39,7 @@ def find_vdf_client() -> pathlib.Path:
     raise FileNotFoundError("can't find vdf_client binary")
 
 
-async def spawn_process(host: str, port: int, counter: int, lock: asyncio.Lock, prefer_ipv6: Optional[bool]):
+async def spawn_process(host: str, port: int, counter: int, lock: asyncio.Lock, prefer_ipv6: bool):
     global stopped
     global active_processes
     path_to_vdf_client = find_vdf_client()
@@ -85,7 +85,9 @@ async def spawn_all_processes(config: Dict, net_config: Dict, lock: asyncio.Lock
     if process_count == 0:
         log.info("Process_count set to 0, stopping TLauncher.")
         return
-    awaitables = [spawn_process(hostname, port, i, lock, net_config.get("prefer_ipv6")) for i in range(process_count)]
+    awaitables = [
+        spawn_process(hostname, port, i, lock, net_config.get("prefer_ipv6", False)) for i in range(process_count)
+    ]
     await asyncio.gather(*awaitables)
 
 
--- a/chia/timelord/timelord_launcher.py
+++ b/chia/timelord/timelord_launcher.py
@@ -39,7 +39,7 @@ def find_vdf_client() -> pathlib.Path:
     raise FileNotFoundError("can't find vdf_client binary")
 
 
-async def spawn_process(host: str, port: int, counter: int, lock: asyncio.Lock, prefer_ipv6: bool):
+async def spawn_process(host: str, port: int, counter: int, lock: asyncio.Lock, *, prefer_ipv6: bool):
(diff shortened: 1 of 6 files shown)


chia/util/network.py
```diff
@@ -108,32 +109,25 @@ def class_for_type(type: NodeType) -> Any:
     raise ValueError("No class for type")
 
 
-def get_host_addr(host: Union[PeerInfo, str], prefer_ipv6: Optional[bool]) -> str:
-    # If there was no preference passed in (from config), set the system-wise
+def get_host_addr(host: str, prefer_ipv6: bool = False) -> IPAddress:
```
arvidn, line 112: would it be really cumbersome to require this bool parameter to be named?

arvidn, line 112:
```suggestion
def get_host_addr(host: str, *, prefer_ipv6: bool = False) -> IPAddress:
```

chia/util/network.py
```diff
@@ -108,32 +109,25 @@ def class_for_type(type: NodeType) -> Any:
...
     for t in addrset:
         if prefer_ipv6 and t[0] == socket.AF_INET6:
-            return t[4][0]
+            return ip_address(t[4][0])
         if not prefer_ipv6 and t[0] == socket.AF_INET:
-            return t[4][0]
+            return ip_address(t[4][0])
     # If neither matched preference, just return the first available
-    return addrset[0][4][0]
+    return ip_address(addrset[0][4][0])
```
arvidn, line 130: I realize you didn't really change this code, but it doesn't look correct. The list of IP addresses may contain multiple IPv4, multiple IPv6 and multiple of other address families. Just picking the first one here doesn't seem right. We should pick the first one that is either `AF_INET6` or `AF_INET`.

Maybe the loop should remember the first address of each type it finds, and then pick one of them that isn't `None` depending on the preference

chia/timelord/timelord_launcher.py
```diff
@@ -39,7 +39,7 @@ def find_vdf_client() -> pathlib.Path:
     raise FileNotFoundError("can't find vdf_client binary")
 
 
-async def spawn_process(host: str, port: int, counter: int, lock: asyncio.Lock, prefer_ipv6: Optional[bool]):
+async def spawn_process(host: str, port: int, counter: int, lock: asyncio.Lock, prefer_ipv6: bool):
```
arvidn, line 42:
```suggestion
async def spawn_process(host: str, port: int, counter: int, lock: asyncio.Lock, *, prefer_ipv6: bool):
```

arvidn, line 42: would this break a lot of call sites?

arvidn (commented):
looks generally good

chia/util/network.py
xdustinface: Funny that right now, all our call sites already refer to "prefer_ipv6" by either variable names or config fetches anyway but sure we can make it named, makes sense, see [d581b92](https://github.com/Chia-Network/chia-blockchain/pull/14068/commits/d581b9277fba2dbf7bed1ec56fb8d2e06024994d).

chia/timelord/timelord_launcher.py
xdustinface: Not more than i anyway already touched with this PR, so i also changed it in [d581b92](https://github.com/Chia-Network/chia-blockchain/pull/14068/commits/d581b9277fba2dbf7bed1ec56fb8d2e06024994d).

chia/util/network.py
xdustinface: Agree and yeah it makes sense to adjust this while reworking this method, see [bf72dbd](https://github.com/Chia-Network/chia-blockchain/pull/14068/commits/bf72dbd25a500cc7cbbeb5742b21328193c759af).

arvidn (approved):

arvidn:
I think it might make sense to add the `cleanup` label to this PR. But if you don't, make sure the PR title makes sense in the changelog

PR #14069, opened 2022-12-07
server: Resolve config peers earlier

xdustinface:
Don't pass around `PeerInfo` objects with a non-IP host.

Based on:
- #14068

### DIFF
--- a/chia/server/start_farmer.py
+++ b/chia/server/start_farmer.py
@@ -16,6 +16,7 @@ from chia.util.chia_logging import initialize_service_logging
 from chia.util.config import load_config, load_config_cli
 from chia.util.default_root import DEFAULT_ROOT_PATH
 from chia.util.keychain import Keychain
+from chia.util.network import get_host_addr
 
 # See: https://bugs.python.org/issue29288
 "".encode("idna")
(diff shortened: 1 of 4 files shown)


chia/server/start_farmer.py
```diff
@@ -36,7 +37,9 @@ def create_farmer_service(
     connect_peers = []
     fnp = service_config.get("full_node_peer")
     if fnp is not None:
-        connect_peers.append(PeerInfo(fnp["host"], fnp["port"]))
+        connect_peers.append(
+            PeerInfo(str(get_host_addr(fnp["host"], prefer_ipv6=config.get("prefer_ipv6", False))), fnp["port"])
```
arvidn, line 41: is `get_host_addr()` not `async`? this looks suspicious. If it's a blocking call, we'll block the whole mainloop

chia/server/start_wallet.py
```diff
@@ -56,8 +57,10 @@ def create_wallet_service(
     fnp = service_config.get("full_node_peer")
 
     if fnp:
-        connect_peers = [PeerInfo(fnp["host"], fnp["port"])]
-        node.full_node_peer = PeerInfo(fnp["host"], fnp["port"])
+        node.full_node_peer = PeerInfo(
+            str(get_host_addr(fnp["host"], prefer_ipv6=config.get("prefer_ipv6", False))), fnp["port"]
```
arvidn, line 61: same comment here. it looks suspicious to do `get_host_addr()` witoout `await`

chia/server/start_farmer.py
xdustinface: Its not `async`. https://github.com/Chia-Network/chia-blockchain/blob/cd4874c584046daa9bdf77ed05c37f3d42de0a8a/chia/util/network.py#L112
so yeah blocking but does it matter where exactly we block on startup? I mean, right now it happens in `Service.run` and after this PR its just moved few layers up in the startup process into `create_<whatever>_service`

chia/server/start_wallet.py
xdustinface: https://github.com/Chia-Network/chia-blockchain/pull/14069#discussion_r1047931543

arvidn (approved):

PR #14071, opened 2022-12-07
server: Integrate reconnect tasks into `ChiaServer`

xdustinface:
Just feels like it belongs in there rather than in its own file because we pass in an `ChiaServer` anyway and access its connections and starting the client.

Based on:
- #14069 
- #14165


arvidn:
this is not an obvious improvement to me. for example, if we ever were to write unit tests for this function, having it be a free function would make that a lot simpler (we would just need to come up with a protocol to mock the `ChiaServer` we pass in, or pass in the two members we access separately).

There are a few other changes in here that seems like obvious improvements though (and also somewhat unrelated to moving the free function into the class). for example, moving the exception type into the error file

xdustinface:
> this is not an obvious improvement to me. for example, if we ever were to write unit tests for this function, having it be a free function would make that a lot simpler (we would just need to come up with a protocol to mock the `ChiaServer` we pass in, or pass in the two members we access separately).

Actually it seemed clear to me that it's part of `ChiaServer`. Since it's not obvious to you, can you explain where you draw the line of making a function accessing internals of a class a method vs. free function? 

Not sure why a maybe coming unit test would justify leaving this method in its own file with a public interface vs integrating it as private method into the related class? Also i don't really get how the "pass in the two members separately" would look like with this being a task which access them at runtime (not only on task creation) other than holding a reference to the passed in members in the task which seems risky/unintuitive?

> There are a few other changes in here that seems like obvious improvements though (and also somewhat unrelated to moving the free function into the class). for example, moving the exception type into the error file

There are few somewhat unrelated simplifications, right, like getting rid of the inner wrapping method and instead create the task in place or simplify the "are we still connected check". But moving the exception isn't unrelated imo. It's just, there is a `ServiceException` before and now there is a `ServerException` because the PR integrated the reconnecting from the service into the server moving the exception is just to not have an exception in the wrong place.

tests/farmer_harvester/test_farmer_harvester.py
```diff
@@ -60,7 +60,7 @@ async def handshake_done() -> bool:
     # Start both services and wait a bit
     await farmer_service.start()
     await harvester_service.start()
-    harvester_service.add_peer(PeerInfo(str(farmer_service.self_hostname), farmer_service._server.get_port()))
+    harvester_service._server.add_peer(PeerInfo(str(farmer_service.self_hostname), farmer_service._server.get_port()))
```
arvidn, line 63: accessing the internal member her will make it harder to change implementation details of `harvester_service`, this increases technical debt. I would prefer tests use the public interface.

chia/server/server.py
```diff
@@ -265,12 +281,20 @@ async def garbage_collect_connections_task(self) -> None:
             for peer_ip in to_remove_ban:
                 del self.banned_peers[peer_ip]
 
+    def add_peer(self, peer: PeerInfo) -> None:
+        if self._reconnect_tasks.get(peer) is not None:
+            raise ServerError(f"Peer {peer} already added")
+        self._reconnect_tasks[peer] = asyncio.create_task(self._reconnect_task_handler(peer))
```
arvidn, line 287: do you know why we have a reconnect task for each peer? It seems really wasteful for every peer to iterate over all peers just to find itself. It would seem simpler an more efficient to just have one task that iterates over all peers and retries the ones that should be retried

chia/server/reconnect_task.py
```diff
@@ -1,29 +0,0 @@
-from __future__ import annotations
-
-import asyncio
-from logging import Logger
-
-from chia.server.server import ChiaServer
-from chia.types.peer_info import PeerInfo
-
-
-def start_reconnect_task(server: ChiaServer, peer_info: PeerInfo, log: Logger) -> asyncio.Task[None]:
```
arvidn, line 10: nothing this function accesses from `server` starts with an underscore. I would think `all_connections` and `start_client()` are meant to be public

arvidn (commented):
is `start_service.py` the only user of `ChiaServer`?
By moving this logic from start_service into ChiaServer you seem to suggest that is the case. If not, this is a behavior change

arvidn:
> Actually it seemed clear to me that it's part of ChiaServer. Since it's not obvious to you, can you explain where you draw the line of making a function accessing internals of a class a method vs. free function?

I think the short answer is: where it makes sense to chop up logic to be unit tested.

> Not sure why a maybe coming unit test would justify leaving this method in its own file with a public interface vs integrating it as private method into the related class?

I would say it's increasing the coupling between those two parts. Generally I think we should be moving in the direction of splitting classes into smaller pieces. That makes them much easier to unit test.

> Also i don't really get how the "pass in the two members separately" would look like with this being a task which access them at runtime (not only on task creation) other than holding a reference to the passed in members in the task which seems risky/unintuitive?

For example, the peer list could be passed in separately from the function to reconnect a peer. This way it would make a test much simpler because you wouldn't have to create something that looks like a `ChiaServer` sharing references to mutable data isn't very nice, but it's how it works now. But perhaps more importantly, it doesn't look like a very good idea to have a task per peer for this to begin with.

xdustinface:
Closing in favour of #14997.

PR #14072, opened 2022-12-07
Make `--number-of-coins` required for the `chia coins split` CLI

altendky:
```python-traceback
chia wallet coins split -f 353135768 -a 1 -t 0xd34ab3bec670fc473acda616bcab1f05665ced504c7c058d18ddb836ff7e6b12
Exception from 'wallet' '>' not supported between instances of 'NoneType' and 'int':
  File "/home/william/chia-blockchain/chia/cmds/cmds_util.py", line 100, in get_any_service_client
  File "/home/william/chia-blockchain/chia/cmds/cmds_util.py", line 211, in execute_with_wallet
    await function(extra_params, wallet_client, new_fp)
  File "/home/william/chia-blockchain/chia/cmds/coin_funcs.py", line 184, in async_split
    if number_of_coins > 500:
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

I think that this command doesn't make sense without a number of coins to create.  There is probably no sensible default so the option should just be required.

### DIFF
--- a/chia/cmds/coins.py
+++ b/chia/cmds/coins.py
@@ -198,6 +198,7 @@ def combine_cmd(
     "--number-of-coins",
     type=int,
     help="The number of coins we are creating.",
+    required=True,
 )
 @click.option(
     "-m",


jack60612 (approved):
oops, lgtm

emlowe (approved):

PR #14074, opened 2022-12-07
Add PublicKeyBytes type to represent public key bytes

aqk:
No user-visible changes. Code maintenance.


aqk:
Smaller PR at Arvid's request.

chia/types/borderlands.py
```diff
@@ -0,0 +1,28 @@
...
+```
+
+Native modules are only able to return native Python types by default, thus we get `bytes`, and not `bytes32`
+from those libraries.
+
+"""
+
+
+def bytes_to_PublicKeyBytes(input_bytes: bytes) -> PublicKeyBytes:
+    return PublicKeyBytes(bytes48(input_bytes))
```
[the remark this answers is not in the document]
aqk: True. To change this, we can add a runtime check like this

```
def bytes_to_PublicKeyBytes(input_bytes: bytes) -> PublicKeyBytes:
    return PublicKeyBytes(bytes48(input_bytes))
```

to get the minor benefit of encouraging conversions closer to the source, or we can keep a more primitive type conversion, keeping the TypeName constructors, and tossing the bytes_to_TypeName conversions.

@altendky @arvidn Thoughts?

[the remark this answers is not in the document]
altendky: Also, if we want the types converted closer then maybe that should be done in the library?  I've suggested a few times that it would be nice for all our projects to have the outer layer be a Python package.  In this case that would let us add wrapper functions that could do this hinting/runtime conversion and checks for us.  For things that don't need wrapping, they can just be re-exported to avoid any runtime overhead or the need to write wrapper methods.

Another direction, make our fixed width bytes and integer types just `NewType`s as well.  Then, at the boundaries of serialization (network, db, etc) we can deal with the serialization.  This has some complexity around checks and is not particularly tied to this change here.  Just commenting in case discussion would be of interest.

chia/rpc/full_node_rpc_api.py
```diff
@@ -749,8 +749,8 @@ async def get_mempool_item_by_tx_id(self, request: Dict) -> EndpointResult:
             raise ValueError("No tx_id in request")
         include_pending: bool = request.get("include_pending", False)
         tx_id: bytes32 = bytes32.from_hexstr(request["tx_id"])
-
-        item = self.service.mempool_manager.get_mempool_item(tx_id, include_pending)
+        spend_bundle_id = tx_id
```
almogdepaz, line 752: rename tx_id to spend_bundle_id instead ?

tests/clvm/coin_store.py
```diff
@@ -70,7 +70,8 @@ def validate_spend_bundle(
         if result.error is not None:
             raise BadSpendBundleError(f"condition validation failure {Err(result.error)}")
 
-        ephemeral_db = dict(self._db)
+        ephemeral_db = dict(self._db.items())
```
almogdepaz, line 73: whats going on here?, was this wrong ?
i dont see any changes to what we take out of ephemeral_db and i assume the tests passed before

altendky: Creating a dict from a dict like it was before does what you might expect.  Creating a dict from an iterable of two-tuples, such as from `.items()`, uses the first element of each tuple as the key and the second as the value.  So the code does the same thing.  Still leaves open the question of what this has to do with the rest of the PR.

chia/full_node/mempool.py
```diff
@@ -95,7 +95,7 @@ def add_to_pool(self, item: MempoolItem) -> None:
         self.sorted_spends[item.fee_per_cost][item.name] = item
 
         for coin in item.removals:
-            coin_id = coin.name()
+            coin_id: bytes32 = coin.name()
```
arvidn, line 98: This patch is already very large. I think it would benefit from not also including unrelated fixups

chia/clvm/spend_sim.py
```diff
@@ -270,15 +269,16 @@ def __init__(self, service: SpendSim) -> None:
...
             cost_result: NPCResult = await self.service.mempool_manager.pre_validate_spendbundle(
-                spend_bundle, None, spend_bundle.name()
+                spend_bundle, None, spend_bundle_id
             )
         except ValidationError as e:
             return MempoolInclusionStatus.FAILED, e.code
         assert self.service.mempool_manager.peak
         cost, status, error = await self.service.mempool_manager.add_spend_bundle(
-            spend_bundle, cost_result, spend_bundle.name(), self.service.mempool_manager.peak.height
+            spend_bundle, cost_result, spend_bundle_id, self.service.mempool_manager.peak.height
```
arvidn, line 281: this file is also an unrelated fixup

chia/simulator/full_node_simulator.py
```diff
@@ -444,7 +445,7 @@ async def process_coin_spends(self, coins: Collection[Coin] = ()) -> None:
         block.
 
         Arguments:
-            coin_names: The coin names to process.
+            coins: The coins to process.
```
arvidn, line 448: this is unrelated, and not obvious that it's an improvement

arvidn (changes requested):
there's a lot of changes in here that are not "Add PublicKeyBytes type". I think you should distill this patch down to introduce one type (and its use)

altendky (changes requested):
I guess I should have "requested changes" before.  Doing it here.

Primarily https://github.com/Chia-Network/chia-blockchain/pull/14074#pullrequestreview-1209262214.

tests/clvm/coin_store.py
almogdepaz: so if it does the same thing why the change ?
also yes not related to this pr

chia/clvm/spend_sim.py
aqk: Okay - I'll split that out.

chia/simulator/full_node_simulator.py
aqk: Will revert.

chia/types/borderlands.py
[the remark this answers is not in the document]
aqk: > Also, if we want the types converted closer then maybe that should be done in the library? I've suggested a few times that it would be nice for all our projects to have the outer layer be a Python package.

I think everyone would prefer that, but it takes work, and it would have to be higher priority than other things.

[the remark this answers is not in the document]
aqk: > It's still not clear to me when bytes_to_PublicKeyBytes() will do something useful. Both for hinting and at runtime it will accept bytes, bytes48, and PublicKeyBytes as I understand it. At runtime, PublicKeyBytes will do what bytes48 does so bytes_to_PublicKeyBytes(x) will be the same as bytes48(bytes48(x)) I think.

No, I don't think you are correct.

```
$ cat x.py
from typing import NewType


b = b"b"
b32 = bytes32(b"b" * 32)

T = NewType("T", bytes32)

a = T(b)
b = T(b32)

$ mypy x.py
x.py:10: error: Argument 1 to "T" has incompatible type "bytes"; expected "bytes32"  [arg-type]
Found 1 error in 1 file (checked 1 source file)

```

tests/clvm/coin_store.py
aqk: You are both correct - this crept in when I was slimming this PR down. I'll revert this change, of course.

chia/simulator/full_node_simulator.py
aqk: This is unrelated, but notice that this is changing the documentation to match the actual parameter name and type. I'll break this out into a separate PR.

aqk: Done.

chia/full_node/mempool.py
aqk: Thanks, done.

chia/full_node/mempool_manager.py
```diff
@@ -60,10 +61,11 @@ def validate_clvm_and_signature(
         if result.error is not None:
             return Err(result.error), b"", {}
 
-        pks: List[bytes48] = []
+        pks: List[PublicKeyBytes] = []
         msgs: List[bytes] = []
         assert result.conds is not None
         pks, msgs = pkm_pairs(result.conds, additional_data)
+        pks = [bytes_to_PublicKeyBytes(pk) for pk in pks]
```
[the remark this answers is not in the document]
aqk: You are correct - this line is no longer needed.

chia/full_node/mempool_manager.py
```diff
@@ -608,19 +610,19 @@ async def new_peak(
...
             if status == MempoolInclusionStatus.SUCCESS:
-                txs_added.append((item.spend_bundle, item.npc_result, item.spend_bundle_name))
+                txs_added_to_mempool_from_pending_pool.append((item.spend_bundle, item.npc_result, item.name))
         log.info(
             f"Size of mempool: {len(self.mempool.spends)} spends, cost: {self.mempool.total_mempool_cost} "
             f"minimum fee rate (in FPC) to get in for 5M cost tx: {self.mempool.get_min_fee_rate(5000000)}"
         )
         self.mempool.fee_estimator.new_block(FeeBlockInfo(new_peak.height, included_items))
-        return txs_added
+        return txs_added_to_mempool_from_pending_pool
```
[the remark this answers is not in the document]
aqk: Thanks.

chia/full_node/mempool_manager.py
```diff
@@ -539,22 +541,22 @@ async def check_removals(self, removals: Dict[bytes32, CoinRecord]) -> Tuple[Opt
...
         Returns a MempoolItem if it's inside one the mempools.
 
         If include_pending is specified, also check the PENDING cache.
         """
-        item = self.mempool.spends.get(bundle_hash, None)
+        item = self.mempool.spends.get(spend_bundle_id, None)
         if not item and include_pending:
             # no async lock needed since we're not mutating the potential_cache
-            item = self.potential_cache._txs.get(bundle_hash, None)
+            item = self.potential_cache._txs.get(spend_bundle_id, None)
```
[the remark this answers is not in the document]
aqk: reverted

chia/full_node/mempool_manager.py
```diff
@@ -580,10 +582,10 @@ async def new_peak(
...
                 for spend in last_npc_result.conds.spends:
                     if spend.coin_id in self.mempool.removals:
-                        c_ids: List[bytes32] = self.mempool.removals[bytes32(spend.coin_id)]
-                        self.mempool.remove_from_pool(c_ids)
-                        for c_id in c_ids:
-                            self.remove_seen(c_id)
+                        spendbundle_ids: List[bytes32] = self.mempool.removals[bytes32(spend.coin_id)]
+                        self.mempool.remove_from_pool(spendbundle_ids)
+                        for spendbundle_id in spendbundle_ids:
+                            self.remove_seen(spendbundle_id)
```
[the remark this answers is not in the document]
aqk: reverted, thanks

altendky:
> @altendky, could you mention the changes you are requesting?

Sorry, certainly.  Mostly https://github.com/Chia-Network/chia-blockchain/pull/14074#pullrequestreview-1209262214.  I added that link to the change request comment at https://github.com/Chia-Network/chia-blockchain/pull/14074#pullrequestreview-1210788217 as well for easy reference from there.

I see you've provided a response that I need to look at.  Thanks.

chia/types/borderlands.py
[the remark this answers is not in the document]
altendky: I think you are right and have now gotten me caught up.  I think I was off on my thinking about inheritance and also the direction I was looking at this from.  Thanks.

Now I'm wondering if it would be worth avoiding the function call overheads for the hint-only benefit.  It both seems like over-optimization and also appropriate given that we are optimizing this to the point of writing the implementations in rust.  `:]`


```python
from typing import TypeVar
from typing_extensions import Protocol

_T = TypeVar("_T", covariant=True)

class BytesToProtocol(Protocol[_T]):
    def __call__(self, __positional_only: bytes) -> _T:
        ...

bytes_to_PublicKeyBytes: BytesToProtocol[PublicKeyBytes] = bytes48
```

Please do question the appropriateness of covariance there.  mypy wanted it and I have to refresh myself every time I deal with this.

These examples show what works and what complains and needs an ignore (for the example, in reality this is what you intend to not allow, you want the mypy complaint)

```python
x: PublicKeyBytes

x = bytes_to_PublicKeyBytes(b"")
x = bytes_to_PublicKeyBytes(bytes48(b""))
x = bytes_to_PublicKeyBytes(PublicKeyBytes(bytes48(b"")))

x = PublicKeyBytes(b"")  # type: ignore[arg-type]
x = PublicKeyBytes(bytes48(b""))
x = PublicKeyBytes(PublicKeyBytes(bytes48(b"")))
```

Perhaps we don't even care about the actual object type and would be happy to not even convert it and trust that the library gave us the right thing.  That's what we have now in at least some cases I think where rust claims `bytes32` but really has only created `bytes`.  Am I remembering correctly @arvidn?  If we're ok with that, then instead of `= bytes48` it could be `= lambda __x: __x`.  (I don't know of an optimized no-op that can be written as a function call)

I have no evidence that these are faster.  Just "reasoning", which isn't always right.  And even if it is, the amount may not be of interest.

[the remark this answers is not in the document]
altendky: I'm totally on board with the core concept of these PRs.  That's why I care to talk through the implementation.  `:]`

I don't think rust, as we handle it now, can literally give us `bytes32` etc.  If nothing else, that'd make a circular type dependency.  Though I think Arvid has commented more about it being an aspect of maturin.  But...  I know I was displeased about the hints lying about the type somewhere.  Looking elsewhere in the same `.pyi` you find things like `def tree_hash(program: ReadableBuffer) -> bytes32: ...`.  To be clear, displeased means I hope we can find something better, not that I know this was the wrong choice.

So, in those cases we ought to be able to just update the existing hint generation to reference these new types instead?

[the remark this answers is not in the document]
arvidn: There are two main challenges in making rust code return the "rich" python types like `bytes32` or `PublicKeyBytes`:

1. Since those types are defined in `chia-blockchain`, it would create a circular dependency
2. It's not trivial to construct native python types and return on the rust side

Those challenges aren't impossible, and one way to solve them (although, perhaps, heavy handed) would be to move `Bytes32` et.al. into `chia_rs` and make them rust native.

If it would help, we could define these python types in `chia_rs.pyi`. It might still be problematic constructing actual plain `bytes` objects and hint them to be `bytes32` or `PublicKeyBytes`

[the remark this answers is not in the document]
arvidn: I think it's reasonable to have to cast at the python/rust boundary for now, in the interest of making progress. I'm not so sure about casting in other places though.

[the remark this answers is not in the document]
altendky: By cast, do you mean a hint-only thing?  Or actually creating the `bytes32` etc objects.  In at least some cases now we are avoiding the object creation overhead and just doing a hint-only 'cheat' in the `.pyi`.  If we leave that and then _add_ any call in chia-blockchain we are adding new overhead.  Even if just `lambda _x: _x`.  This may not be critical.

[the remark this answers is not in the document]
aqk: I think the best next-step following the theme of these patches is to move bytes32, etc. over to `clvm_tools_rs`

It will be a big pain initially, but once they are over there, it will be easy to export types like these that build on them.

As for these patches, though, if we keep the changes to the Python side only, I think the right thing to do is do any checks (both static and dynamic) once at the boundary, and just change the type annotations everywhere else.

 In its current state, this patch only does static checks at the interface, there are no dynamic checks, and there are no conversions or object creations afterward. You can check that for yourself by searching for `PublicKeyBytes(`

chia/simulator/full_node_simulator.py
aqk: @arvidn This was moved here: https://github.com/Chia-Network/chia-blockchain/pull/14112/files

aqk:
@almogdepaz @arvidn @altendky Thanks for your reviews. I believe all issues are addressed. Anyone see anything else?

chia/full_node/mempool_manager.py
```diff
@@ -549,10 +550,10 @@ async def check_removals(self, removals: Dict[bytes32, CoinRecord]) -> Tuple[Opt
         # 5. If coins can be spent return list of unspents as we see them in local storage
         return None, []
 
-    def get_spendbundle(self, bundle_hash: bytes32) -> Optional[SpendBundle]:
+    def get_spendbundle(self, spend_bundle_id: bytes32) -> Optional[SpendBundle]:
```
arvidn, line 553: this changes the function signature. any call site naming this parameter will need to be changed as well. It's unrelated to the PR and non-trivial. Did you inspect all call sites of `get_spendbundle()`?

chia/full_node/mempool_manager.py
```diff
@@ -605,11 +606,11 @@ async def new_peak(
             self.seen_bundle_hashes = {}
             for item in old_pool.spends.values():
                 _, result, err = await self.add_spend_bundle(
-                    item.spend_bundle, item.npc_result, item.spend_bundle_name, item.height_added_to_mempool
```
arvidn, line 608: I don't see this patch renaming `spend_bundle_name` in `MempoolItem`. Is this a synonym?
It seems unrelated to this PR

chia/rpc/full_node_rpc_api.py
```diff
@@ -748,11 +748,10 @@ async def get_mempool_item_by_tx_id(self, request: Dict) -> EndpointResult:
...
             raise ValueError("No tx_id in request")
         include_pending: bool = request.get("include_pending", False)
-        tx_id: bytes32 = bytes32.from_hexstr(request["tx_id"])
-
-        item = self.service.mempool_manager.get_mempool_item(tx_id, include_pending)
+        spend_bundle_id: bytes32 = bytes32.from_hexstr(request["tx_id"])
+        item = self.service.mempool_manager.get_mempool_item(spend_bundle_id, include_pending)
         if item is None:
-            raise ValueError(f"Tx id 0x{tx_id.hex()} not in the mempool")
+            raise ValueError(f"Tx id 0x{spend_bundle_id.hex()} not in the mempool")
```
arvidn, line 754: renaming of this also seems unrelated

chia/rpc/full_node_rpc_api.py
```diff
@@ -768,10 +767,9 @@ async def get_fee_estimate(self, request: Dict) -> Dict[str, Any]:
 
         cost = 0
         if "spend_bundle" in request:
-            spend_bundle = SpendBundle.from_json_dict(request["spend_bundle"])
-            spend_name = spend_bundle.name()
+            spend_bundle: SpendBundle = SpendBundle.from_json_dict(request["spend_bundle"])
             npc_result: NPCResult = await self.service.mempool_manager.pre_validate_spendbundle(
-                spend_bundle, None, spend_name
+                spend_bundle, None, spend_bundle.name()
```
arvidn, line 772: this looks like it's undoing another PR you already landed

chia/types/border_types.py
```diff
@@ -0,0 +1,37 @@
...
+```
+    def bytes_to_PublicKeyBytes(input_bytes: bytes) -> PublicKeyBytes:
+```
+
+Native modules are only able to return native Python types by default, thus we get `bytes`, and not `bytes32`
+from those libraries.
+
+"""
+
+bytes_to_PublicKeyBytes: BytesToProtocol[PublicKeyBytes] = bytes48
```
altendky, line 37: I don't know if this makes it better or not.  There will be a batch of these after the other PRs and maybe just repeating the name, without keeping it properly associated with it's supertype, might be easier?  Or maybe not.

```suggestion
bytes_to_PublicKeyBytes: BytesToProtocol[PublicKeyBytes] = PublicKeyBytes.__supertype__
```

chia/types/borderlands.py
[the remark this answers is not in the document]
altendky: Well, I apparently entirely misunderstood the runtime behavior of `PublicKeyBytes()`.  `:[`  I thought it acted like the constructor of the type it was "created from" as opposed to being a runtime no-op.  I'm sorry that I missed that.  It's right there...

https://docs.python.org/3.11/library/typing.html#newtype
> At runtime, the statement `Derived = NewType('Derived', Base)` will make `Derived` a callable that immediately returns whatever parameter you pass it.

I'm good with resolving this discussion.  I'll add a thumbs up here and once you to do as well then whoever can mark it resolves.

I think I would need to hear more about `bytes32` etc going into `clvm_tools_rs`, but that's another discussion we can take up elsewhere.  Thanks for talking through this all with me.

altendky (changes requested):
In addition to Arvid's comments about unrelated changes, there are a few `__init__.py` files being added that don't seem relevant.

PR #14076, opened 2022-12-08
Add BlockRecordHeaderHash type to reduce use of bytes32 type

aqk:
Code maintenance. No user visible changes.

### DIFF
--- a/chia/consensus/block_record.py
+++ b/chia/consensus/block_record.py
@@ -24,7 +24,7 @@ class BlockRecord(Streamable):
     difficulty adjustments, etc, without saving the whole header block in memory.
     """
 
-    header_hash: bytes32
+    header_hash: BlockRecordHeaderHash
     prev_hash: bytes32  # Header hash of the previous block
     height: uint32
     weight: uint128  # Total cumulative difficulty of all ancestor blocks since genesis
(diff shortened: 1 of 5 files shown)


chia/consensus/block_record.py
```diff
@@ -12,6 +12,8 @@
 from chia.util.ints import uint8, uint32, uint64, uint128
 from chia.util.streamable import Streamable, streamable
 
+BlockRecordHeaderHash = NewType("BlockRecordHeaderHash", bytes32)
+
 
 @streamable
 @dataclass(frozen=True)
```
aqk, line 19: Note that `BlockRecord` is not the serialized block datastructure.

arvidn:
I think you should also use this new type in this PR.

chia/consensus/block_record.py
```diff
@@ -22,7 +24,7 @@ class BlockRecord(Streamable):
     difficulty adjustments, etc, without saving the whole header block in memory.
     """
 
-    header_hash: bytes32
+    header_hash: BlockRecordHeaderHash
     prev_hash: bytes32  # Header hash of the previous block
```
arvidn, line 28: `prev_hash` is also a `BlockRecordHeaderHash`, right?

arvidn (commented):
is there a difference between block-*record*-header-hash and just block-header-hash?
I believe there's only one kind of header hashes for blocks. It would be a shorter name to remove  `Record`

PR #14077, opened 2022-12-08
Replace uses of bytes32 with CoinID

aqk:


arvidn (changes requested):
I think you should make this patch only do what the title says: "Replace uses of bytes32 with CoinID"

It looks like there's a fair number of changes in here that you've also posted in other PRs now

PR #14079, opened 2022-12-08
Introduce TransactionRecordID

aqk:
Code Maintenance. No user visible changes.


benchmarks/coin_store.py
```diff
@@ -26,13 +26,13 @@ def make_coin() -> Coin:
...
-def make_coins(num: int) -> Tuple[List[Coin], List[bytes32]]:
+def make_coins(num: int) -> Tuple[List[Coin], List[CoinID]]:
     additions: List[Coin] = []
-    hashes: List[bytes32] = []
+    hashes: List[CoinID] = []
     for i in range(num):
         c = make_coin()
         additions.append(c)
-        hashes.append(c.name())
+        hashes.append(CoinID(c.name()))
```
arvidn, line 35: The title says `TransactionRecordID`, so I assume all these `CoinID` changes are mistakes, is that right?

PR #14083, opened 2022-12-08
pools: Fix delayed pool config updates

### LINKED ISSUE #13872: [Bug] Switching Pool Does not Work Until Restarting
### What happened?

I click to Join or leave a pool. Press submit.
The status changes to "Pending" and then to "Pooling" for joining.
However, nothing updates after that, points stay at zero, and pool does not receive partials. The view pool URL is empty. 
This remains until the client is restarted, I only waited about 10 minutes each time,.. It used to work right away once Pooling is displayed on previous clients.
When going to self-pooling from Pooling. The "Pooling" status remains until restarted, even after transaction is confirmed and the pool shows many invalid partials.
I tried 1.6.0 and 1.6.1

### Version

1.6.1

Windows

GUI

### Relevant log output

_No response_

xdustinface:
We are in a DB write transaction when we are inside `apply_state_transition` where we then create a new task which fetches the plotnft state from the database. Now, if the wallet is already synced / if we call the update https://github.com/Chia-Network/chia-blockchain/blob/2dea800491302b06f555cbc545607d184f5ea1e6/chia/pools/pool_wallet.py#L321 before the DB transaction was closed we don't get the latest plotnft state since its not yet committed to the DB.  9771183265df2957cac0baa3dce5cfa9bb29851b fixes this by creating the config dict within the same task which has the write transaction open so that we get the latest data still and then it puts it into the task.

2aced3ca39a9b66c62ed8b5c747e762dd7ce073b adds a test and the other commits are some cleanups while at it. 

Fixes issues like https://github.com/Chia-Network/chia-blockchain/issues/13872

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -314,10 +314,8 @@ class PoolWallet:
         if self._update_pool_config_after_sync_task is None or self._update_pool_config_after_sync_task.done():
 
             async def update_pool_config_after_sync_task():
-                synced = await self.wallet_state_manager.synced()
-                while not synced:
-                    await asyncio.sleep(5)  # we sync pretty quickly, so I think this is ok.
-                    synced = await self.wallet_state_manager.synced()
+                while not await self.wallet_state_manager.synced():
+                    await asyncio.sleep(5)
                 await self.update_pool_config()
                 self.log.info("Updated pool config after syncing finished.")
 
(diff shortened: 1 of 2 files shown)


jack60612:
this fix looks good, i'm going to test it shortly.

xdustinface:
I think there is a issue/race now with multiple wallets because of the way the delayed update fetches the config upfront now but im on mobile, cant check for the next few hours.

chia/pools/pool_wallet.py
```diff
@@ -302,26 +306,30 @@ async def apply_state_transition(self, new_state: CoinSpend, block_height: uint3
                     self.next_transaction_fee = uint64(0)
                 break
 
-        await self.update_pool_config_after_sync()  # Update pool config after we finish syncing.
+        if self.wallet_state_manager.synced():
```
emlowe, line 309: This needs an await
```suggestion
        if await self.wallet_state_manager.synced():
```

emlowe (changes requested):

jack60612:
Closed in favor of removing offending code #14090

PR #12795, opened 2022-08-04
checkpoint: into long_lived/atari from main  @ 8ec56d1901315a5efb13202023e15930349ae504

AmineKhaldi:
Source hash: 8ec56d1901315a5efb13202023e15930349ae504
Remaining commits: 75

PR #14090, opened 2022-12-08
revert delayed pool config handling (#12795, #13154)

altendky:
- Revert "We did not await the async function (#13154)"
- Revert "add delayed pool config update feature (#12796)"

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -314,10 +314,10 @@ class PoolWallet:
         if self._update_pool_config_after_sync_task is None or self._update_pool_config_after_sync_task.done():
 
             async def update_pool_config_after_sync_task():
-                synced = await self.wallet_state_manager.synced()
+                synced = self.wallet_state_manager.synced()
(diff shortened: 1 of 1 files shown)


emlowe:
This works in my testing pretty well, and seems a reasonable revert of the code.

I think this is a good plan for 1.7.0 - and then we can work on fixing the fix for another day with some better code, especially if we feel there might be some more dragons in there somewhere.

emlowe:
I tested this and it seems solid. I create two plotnfts and set one to pool and one to self - I then at one time reversed that (well two commands with &&) - set the one to self and the other to pool.

All went through fine and everything in config is ok.
I'm testing a wallet resync to see what happen while we walk the chain, but I do expect the old behaviour that we tried to fix.

I'm pretty good with this

emlowe:
Testing complete, syncing from 0 as expected in the wallet wrt config changes

emlowe (approved):

jack60612 (approved):
Looks good!

PR #14093, opened 2022-12-08
handle incoming unknown capabilities

altendky:


emlowe (approved):

arvidn (approved):
this is still an issue when a 1.7.0 peer connects to an older peer though, right? because the older peer will still have the issue of unknown capabilities

emlowe:
Not 1.7.0 - because we didn't introduce a new capability in 1.7.0.
There is a new capability in `main` that would cause problems with older versions - we will likely revert that though

arvidn:
I see. I thought it had made it into 1.7.

PR #14096, opened 2022-12-09
remove (redundant) OR FAIL from UPDATE sql statements

### LINKED ISSUE #10008: [Bug] Errant SQL directive does not work
### What happened?

The SQL directive `OR FAIL` does not work as anticipated. These should be removed. The following code demonstrates that no exception or failure occurs when errant behavior is encountered.

```
import asyncio
import aiosqlite

async def example():
    async with aiosqlite.connect("test_db.db") as conn:
        try:
            await conn.execute("CREATE TABLE example(id INT, name text)")
            await conn.execute("INSERT INTO example VALUES(0, 'satoshi')")
            await conn.execute(
                "UPDATE OR FAIL example SET name='bram' WHERE id=1"
            )
            await conn.commit()
            print('no error') # 'no error' printed
        except Exception as e:
            print("error: " + str(e))

asyncio.run(example())
```

### Version

protocols_and_cats_rebased

macOS

CLI

### Relevant log output

_No response_

arvidn:
remove (redundant) OR FAIL from UPDATE sql statements. When relevant, assert the number of affected rows (which was the original intent).

~Also, enable SQLITE strict mode, to catch more typos in the future.~

### DIFF
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -155,15 +155,17 @@ class BlockStore:
         if self.db_wrapper.db_version == 2:
             async with self.db_wrapper.writer_maybe_transaction() as conn:
                 await conn.execute(
-                    "UPDATE OR FAIL full_blocks SET in_main_chain=0 WHERE height>? AND in_main_chain=1", (height,)
+                    "UPDATE full_blocks SET in_main_chain=0 WHERE height>? AND in_main_chain=1", (height,)
                 )
 
     async def set_in_chain(self, header_hashes: List[Tuple[bytes32]]) -> None:
         if self.db_wrapper.db_version == 2:
             async with self.db_wrapper.writer_maybe_transaction() as conn:
-                await conn.executemany(
-                    "UPDATE OR FAIL full_blocks SET in_main_chain=1 WHERE header_hash=?", header_hashes
-                )
+                async with await conn.executemany(
+                    "UPDATE full_blocks SET in_main_chain=1 WHERE header_hash=?", header_hashes
+                ) as cursor:
+                    if cursor.rowcount != len(header_hashes):
+                        raise RuntimeError(f"The blockchain database is corrupt. All of {header_hashes} should exist")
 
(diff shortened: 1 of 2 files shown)


AmineKhaldi (dismissed):

chia/full_node/block_store.py
```diff
@@ -155,15 +155,16 @@ async def rollback(self, height: int) -> None:
...
     async def set_in_chain(self, header_hashes: List[Tuple[bytes32]]) -> None:
         if self.db_wrapper.db_version == 2:
             async with self.db_wrapper.writer_maybe_transaction() as conn:
-                await conn.executemany(
-                    "UPDATE OR FAIL full_blocks SET in_main_chain=1 WHERE header_hash=?", header_hashes
-                )
+                async with await conn.executemany(
+                    "UPDATE full_blocks SET in_main_chain=1 WHERE header_hash=?", header_hashes
+                ) as cursor:
+                    assert cursor.rowcount > 0
```
altendky, line 167: Asserts should have a comment expressing what must have happened and why this shouldn't be possible and what sort of bug might trigger this.  Or some part of that. 
 Remember that the CLI result for relevant asserts is `{"success": False, "error": ""}` when there's no message.  Tracebacks from pyinstallered code don't even contain the code.  So you end up having to verify the version that was running and pull up that version in code to find what was even asserted.

Side note, I should find a linter setting to require this...  Albeit as a stop-gap until I get my way and get to ban `assert` entirely in runtime code.  `:]`

altendky (dismissed):

chia/full_node/block_store.py
arvidn: I changed this to a `RuntimeError` with an informative error message

arvidn:
I dropped the `strict=on` commit

altendky (approved):

jack60612 (approved):

AmineKhaldi (approved):

trepca (approved):

PR #14099, opened 2022-12-09
Cleanup options for chia wallet coins commands

emlowe:
Attempt to cleanup CLI commands primarily for `chia wallet coins`

Removed the majority of short-codes.
Removed the word "coin" from most commands as these are already within `chia wallet coins` so it would seem clear that everything refers to coins
Options that are `multiple=True` take a single item and therefore the option should not be plural.

Needs testing


chia/cmds/coins.py
```diff
@@ -25,30 +25,26 @@ def coins_cmd(ctx: click.Context) -> None:
...
-    "--max-coin-amount",
+    "--max-amount",
     help="Ignore coins worth more then this much XCH or CAT units",
     type=str,
     default="0",
 )
 @click.option(
-    "-e",
-    "--excluded-coin-ids",
+    "--exclude-id",
```
altendky, line 42: Make the parameter/variable name also sensible
```suggestion
    "--exclude-id",
    "excluded_ids",
```

chia/cmds/coins.py
```diff
@@ -64,20 +60,20 @@ def list_cmd(
...
     id: int,
     show_unconfirmed: bool,
-    min_coin_amount: str,
-    max_coin_amount: str,
-    excluded_coin_ids: Tuple[str],
-    excluded_coin_amounts: Tuple[int],
+    min_amount: str,
+    max_amount: str,
+    exclude_id: Tuple[str],
+    exclude_amount: Tuple[int],
```
altendky, line 66: Not an error in this PR, but I think these would be lists not tuples.

```suggestion
    excluded_ids: Tuple[str],
    excluded_amounts: Tuple[int],
```

chia/cmds/coins.py
```diff
@@ -25,30 +25,26 @@ def coins_cmd(ctx: click.Context) -> None:
...
-    "-e",
-    "--excluded-coin-ids",
+    "--exclude-id",
     multiple=True,
     help="prevent this coin from being included.",
 )
 @click.option(
-    "-x",
-    "--excluded-coin-amounts",
+    "--exclude-amount",
```
altendky, line 47:
```suggestion
    "--exclude-amount",
    "excluded_amounts",
```

chia/cmds/coins.py
```diff
@@ -64,20 +60,20 @@ def list_cmd(
...
     extra_params = {
         "id": id,
-        "max_coin_amount": max_coin_amount,
-        "min_coin_amount": min_coin_amount,
-        "excluded_amounts": excluded_coin_amounts,
-        "excluded_coin_ids": list(excluded_coin_ids),
+        "max_amount": max_amount,
+        "min_amount": min_amount,
+        "excluded_amounts": exclude_amount,
+        "excluded_coin_ids": list(exclude_id),
```
altendky, line 76:
```suggestion
        "excluded_amounts": excluded_amounts,
        "excluded_coin_ids": list(excluded_ids),
```

chia/cmds/coins.py
```diff
@@ -99,21 +95,19 @@ def list_cmd(
...
-    "--min-coin-amount",
+    "--min-amount",
     help="Ignore coins worth less then this much XCH or CAT units",
     type=str,
     default="0",
 )
 @click.option(
-    "-e",
-    "--excluded-coin-amounts",
+    "--exclude-amount",
```
altendky, line 110:
```suggestion
    "--exclude-amount",
    "excluded_amounts",
```

chia/cmds/coins.py
```diff
@@ -216,22 +207,22 @@ def combine_cmd(
     type=str,
     required=True,
 )
-@click.option("-t", "--target-coin-id", type=str, required=True, help="The coin id of the coin we are splitting.")
+@click.option("-t", "--target-id", type=str, required=True, help="The coin id of the coin we are splitting.")
```
altendky, line 210: Does `-id` distinguish anything here?  Almost makes it sound like we're saying it's not a coin but an id instead.

altendky (changes requested):
A spattering, take a look at them and see where else the ideas apply.  Decided not to create suggestions for all of them.

emlowe:
Replaced by #14102

Issue #14002, opened 2022-11-24
[Bug] Wallet 1.7.0-b2 is missing many transactions


atomsymbol:
### What happened?

After updating from 1.6.2 to 1.7.0-b2, if the Chia 1.7.0-b2 node is started with the wallet DB from 1.6.2:

The wallet fails to properly record new transactions. The wallet balance becomes invalid after just a few new transactions.

### Version

1.7.0-b2

### What platform are you using?

Linux

### What ui mode are you using?

GUI

nqzhang:
Also confirmed in 1.7.0-rc2
@wallentx

wallentx:
Are you able to provide any logs that might help point us in the right direction?

nqzhang:
> Are you able to provide any logs that might help point us in the right direction?

This is the transaction record for wallet  1.6.1

![1 6 1_algold](https://user-images.githubusercontent.com/12559853/206616244-1338a795-7c90-4f6d-87b0-8e1f5df8d8c0.png)

Then upgrade chia to 1.7.0-rc2
delete wallet db
delete log in .chia/mainnet/log/debug.log
resync wallet db
Here are all logs during wallet sync
[debug.log](https://github.com/Chia-Network/chia-blockchain/files/10191578/debug.log)

paninaro:
Can you try to adjust the following setting in your config.yaml?
`  max_subscribe_items: 200000`

Try bumping the value to something like 400000

Because you appear to have an especially large number of CAT wallets, it may be that your full node is hitting this `max_subscribe_items` limit.

Also set `connect_to_unknown_peers` to false to avoid hitting untrusted full node peers that may have a lower limit set.
`  connect_to_unknown_peers: false`

paninaro:
@Quexington the debug log includes a bunch of exceptions like the following:

```2022-12-09T11:20:28.490 wallet chia.wallet.cat_wallet.cat_wallet: WARNING  Not creating wallet for already existing CAT wallet
2022-12-09T11:20:28.491 wallet chia.wallet.wallet_state_manager: ERROR    Error adding state... Wallet already exists
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/chia/wallet/wallet_state_manager.py", line 1034, in new_coin_state
    wallet_id, wallet_type = await self.determine_coin_type(peer, coin_state, fork_height)
[...]

PR #14101, opened 2022-12-09
Return existing CAT wallet instead of raising

Quexington:
There's no reason to raise here, all the caller is looking for is an instance of the CATWallet that they're looking for and to create it if it doesn't exist.

https://github.com/Chia-Network/chia-blockchain/issues/14002

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -586,7 +586,7 @@ class WalletRpcApi:
 
             elif request["mode"] == "existing":
                 async with self.service.wallet_state_manager.lock:
-                    cat_wallet = await CATWallet.create_wallet_for_cat(
+                    cat_wallet = await CATWallet.get_or_create_wallet_for_cat(
                         wallet_state_manager, main_wallet, request["asset_id"], name
                     )
                 self.service.wallet_state_manager.state_changed("wallet_created")
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -589,7 +589,6 @@ class WalletRpcApi:
                     cat_wallet = await CATWallet.get_or_create_wallet_for_cat(
                         wallet_state_manager, main_wallet, request["asset_id"], name
                     )
-                self.service.wallet_state_manager.state_changed("wallet_created")
                 return {"type": cat_wallet.type(), "asset_id": request["asset_id"], "wallet_id": cat_wallet.id()}
 
             else:  # undefined mode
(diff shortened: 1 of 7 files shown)


arvidn:
I think it would be good to rename the function to indicate that it has this new semantic. Perhaps `get_or_create_cat_wallet()` or something like that.
I think that would also requiring updating all call sites, which helps a reviewer to also ensure these new semantics make sense in all cases where it's currently called.

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -192,7 +192,7 @@ async def create_wallet_for_cat(
                 assert isinstance(w, CATWallet)
                 if w.get_asset_id() == limitations_program_hash_hex:
                     self.log.warning("Not creating wallet for already existing CAT wallet")
-                    raise ValueError("Wallet already exists")
+                    return w
```
Quexington, line 195: Here's the main change

chia/rpc/wallet_rpc_api.py
```diff
@@ -586,7 +586,7 @@ async def create_new_wallet(self, request: Dict) -> EndpointResult:
 
             elif request["mode"] == "existing":
                 async with self.service.wallet_state_manager.lock:
-                    cat_wallet = await CATWallet.create_wallet_for_cat(
+                    cat_wallet = await CATWallet.get_or_create_wallet_for_cat(
                         wallet_state_manager, main_wallet, request["asset_id"], name
                     )
                 self.service.wallet_state_manager.state_changed("wallet_created")
```
emlowe, line 592: this state changed message will fire now in cases where the wallet wasn't actually created, but perhaps that is harmless

Quexington: Moved the notification inside of the function just in case

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -208,6 +208,7 @@ async def create_wallet_for_cat(
 
         self.lineage_store = await CATLineageStore.create(self.wallet_state_manager.db_wrapper, self.get_asset_id())
         await self.wallet_state_manager.add_new_wallet(self, self.id())
+        self.wallet_state_manager.state_changed("wallet_created")
```
emlowe, line 211: Interestingly, `add_new_wallet` already calls `state_changed` - not sure if all the previous calls were also redundant. https://github.com/Chia-Network/chia-blockchain/blob/c209f886b1e8c30902c0b6e05884d55ad1c9591f/chia/wallet/wallet_state_manager.py#L1687

emlowe (approved):
Looks good to me

chia/wallet/cat_wallet/cat_wallet.py
wallentx: Resolving, as this has been reviewed, and approved

PR #14102, opened 2022-12-09
Cleanup cli options for chia wallet coins {list,split,combine}

emlowe:
Attempt to cleanup CLI commands for chia wallet coins {list,split,combine}

Removed the majority of short-codes.
Simplified most long options
Options that are multiple=True take a single item and therefore the option should not be plural.

Needs testing

### DIFF
--- a/chia/cmds/coin_funcs.py
+++ b/chia/cmds/coin_funcs.py
@@ -36,7 +36,7 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
         return
     final_min_coin_amount: uint64 = uint64(int(min_coin_amount * mojo_per_unit))
     final_max_coin_amount: uint64 = uint64(int(max_coin_amount * mojo_per_unit))
-    final_excluded_amounts: List[uint64] = [uint64(int(amount * mojo_per_unit)) for amount in excluded_amounts]
+    final_excluded_amounts: List[uint64] = [uint64(int(Decimal(amount) * mojo_per_unit)) for amount in excluded_amounts]
     conf_coins, unconfirmed_removals, unconfirmed_additions = await wallet_client.get_spendable_coins(
         wallet_id=wallet_id,
         max_coin_amount=final_max_coin_amount,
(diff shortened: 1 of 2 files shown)


chia/cmds/coins.py
```diff
@@ -64,20 +62,20 @@ def list_cmd(
...
     id: int,
     show_unconfirmed: bool,
-    min_coin_amount: str,
-    max_coin_amount: str,
-    excluded_coin_ids: Tuple[str],
-    excluded_coin_amounts: Tuple[int],
+    min_amount: str,
+    max_amount: str,
+    coins_to_exclude: Tuple[str],
+    amounts_to_exclude: Tuple[int],
```
altendky, line 68: `str`?  In XCH?  And is it really a `Tuple`?  Or a `List`.

chia/cmds/coin_funcs.py
```diff
@@ -36,7 +36,7 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
         return
     final_min_coin_amount: uint64 = uint64(int(min_coin_amount * mojo_per_unit))
     final_max_coin_amount: uint64 = uint64(int(max_coin_amount * mojo_per_unit))
-    final_excluded_amounts: List[uint64] = [uint64(int(amount * mojo_per_unit)) for amount in excluded_amounts]
+    final_excluded_amounts: List[uint64] = [uint64(int(Decimal(amount) * mojo_per_unit)) for amount in excluded_amounts]
```
altendky, line 39: I don't know what we usually do, but in general I would rather this happen right at the CLI.  But, why only this one and not the other amounts?

chia/cmds/coins.py
```diff
@@ -64,20 +62,20 @@ def list_cmd(
...
     address_prefix = selected_network_address_prefix(config)
     extra_params = {
         "id": id,
-        "max_coin_amount": max_coin_amount,
-        "min_coin_amount": min_coin_amount,
-        "excluded_amounts": excluded_coin_amounts,
-        "excluded_coin_ids": list(excluded_coin_ids),
+        "max_coin_amount": max_amount,
+        "min_coin_amount": min_amount,
+        "excluded_amounts": amounts_to_exclude,
```
altendky, line 77: This isn't actually going to the RPC, so could change the keys as well if you wanted.  (someday this indirection just goes away)

altendky (changes requested):

chia/cmds/coin_funcs.py
emlowe: The others happen a few lines up at the top of the function:
```
    min_coin_amount = Decimal(args["min_coin_amount"])
    max_coin_amount = Decimal(args["max_coin_amount"])
```

chia/cmds/coins.py
emlowe: I believe it should be a List.

chia/cmds/coin_funcs.py
altendky: I see, they are handled further up.  Probably good to do them all in one place.  `Decimal` conversion above and mojo conversion here.

chia/cmds/coins.py
emlowe: Changes made Tuple->List. I believe they should all be Lists as changed

altendky (approved):

jack60612 (approved):
lgtm

paninaro (approved):

PR #14105, opened 2022-12-11
patch capabilities for old clients

almogdepaz:
make sure we dont send new capabilities for clients under 1.6.2
no breaking changes, fixes the issue where old clients would disconnect new one but only for the older clients outbound connections 

example:
a 1.5 peer initiates a connection to a 1.6.2 peer, the 1.6.2 knows what 1.5 supports and only sends those to avoid the disconnect

this cant be done the other way because of who sends the first handshake message with the version and supported capabilties

### DIFF
--- a/chia/protocols/shared_protocol.py
+++ b/chia/protocols/shared_protocol.py
@@ -49,3 +49,10 @@ capabilities = [
     (uint16(Capability.RATE_LIMITS_V2.value), "1"),
     (uint16(Capability.NONE_RESPONSE.value), "1"),
 ]
+
+# capabilities to send to nodes prior to 1.7
+limitedcapabilties = [
+    (uint16(Capability.BASE.value), "1"),
+    (uint16(Capability.BLOCK_HEADERS.value), "1"),
+    (uint16(Capability.RATE_LIMITS_V2.value), "1"),
+]
--- a/chia/protocols/shared_protocol.py
+++ b/chia/protocols/shared_protocol.py
@@ -50,7 +50,7 @@ capabilities = [
     (uint16(Capability.NONE_RESPONSE.value), "1"),
 ]
 
-# capabilities to send to nodes prior to 1.7
+# capabilities to send to nodes prior to 1.6.2
 limitedcapabilties = [
     (uint16(Capability.BASE.value), "1"),
     (uint16(Capability.BLOCK_HEADERS.value), "1"),
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -693,6 +693,9 @@ class WSChiaConnection:
         version = software_version.split(".")
         major = int(version[0])
         minor = int(version[1])
-        if major == 1 and minor < 7:
+        patch_version = 0
+        if len(version) > 2:
+            patch_version = int(version[2])
+        if major == 1 and minor < 6 and patch_version < 2:
             return limitedcapabilties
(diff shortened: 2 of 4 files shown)


altendky (changes requested):
Maybe parametrize the tests both against the edges right below and above the boundary as well as further away including even future versions.

chia/server/ws_connection.py
```diff
@@ -672,3 +687,12 @@ def get_peer_logging(self) -> PeerInfo:
...
     def has_capability(self, capability: Capability) -> bool:
         return capability in self.peer_capabilities
+
+    # only send limitedcapabilties to peers before 1.7
+    def get_capabilties_for_version(self, software_version: str) -> list[tuple[uint16, str]]:
+        version = software_version.split(".")
+        major = int(version[0])
+        minor = int(version[1])
+        if major == 1 and minor < 7:
+            return limitedcapabilties
```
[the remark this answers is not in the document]
almogdepaz: i like what i have better where its clear we only check the major and minor version numbers.

as for the malformed version, i think we should just ignore, other implementations are allowed so we dont want to disconnect them or make them report some fake version number just for handshake sake

almogdepaz:
> Maybe parametrize the tests both against the edges right below and above the boundary as well as further away including even future versions.

i changed the tests to be more close to the check case, i dont think there's use in adding more cases

emlowe:
Almog, clarification.
This works if the old peer initiates the connection, in which case this code makes sure to not send a new capability

This does not work if the new peer initiates the connection, as it will send the new capability that will be rejected by the old node.

Is my understanding correct here?

chia/protocols/shared_protocol.py
```diff
@@ -49,3 +49,10 @@ class Handshake(Streamable):
     (uint16(Capability.RATE_LIMITS_V2.value), "1"),
     (uint16(Capability.NONE_RESPONSE.value), "1"),
 ]
+
+# capabilities to send to nodes prior to 1.7
```
arvidn, line 53: I believe this is now version 1.6.2

chia/server/ws_connection.py
```diff
@@ -675,3 +691,14 @@ def get_peer_logging(self) -> PeerInfo:
...
         return capability in self.peer_capabilities
+
+    # only send limitedcapabilties to peers before 1.6.2
+    def get_capabilties_for_version(self, software_version: str) -> list[tuple[uint16, str]]:
+        major, minor, patch, _ = get_version_numbers(software_version)
+        patch_number = 0
+        path_split = re.findall("[0-9]+", patch)  # extract number from patch string
+        if len(path_split) > 1:
+            patch_number = path_split[0]
+        if int(major) == 1 and int(minor) <= 6 and int(patch_number) < 2:
```
arvidn, line 702: could you write this as: `(major, minor, patch_number) < (1, 6, 2)`?

chia/protocols/shared_protocol.py
almogdepaz: thanks, leftover comment from before the change

almogdepaz:
@altendky suggested using https://packaging.pypa.io/en/stable/version.html for parsing im up for that if we are ok with moving towards using that everywhere for versions, otherwise i feel this will only cause more confusion
@emlowe any thoughts on this?

emlowe:
closing old PR

PR #14106, opened 2022-12-12
Capabilities fix

aqk:


arvidn (approved):
I think this test would be a lot more useful if it would make a plain WebSocket connection and start speaking just enough of the handshake of being able to send a custom list of capabilities. I imagine that we'll want more tests where we send unexpected values

almogdepaz:
this does the same thing as https://github.com/Chia-Network/chia-blockchain/pull/14093 which was already merged to 1.7 no ?

aqk:
Yes, this is almost the same as https://github.com/Chia-Network/chia-blockchain/pull/14093

PR #14107, opened 2022-12-12
Remove duplicate '-e' command line flag in 'chia wallet send'

aqk:


arvidn (commented):
my main comment is: is this the right `-e` to remove, or should the one associated with `--exclude-coin` be removed instead? The latter seems newer, but I could be wrong. Maybe the final state is better to allow excluding with a short option.

arvidn (approved):
you should give this PR a more specific title, since it will be part of the changelog

aqk:
I think we should remove many more of the single-letter flags.

aqk:
> my main comment is: is this the right `-e` to remove, or should the one associated with `--exclude-coin` be removed instead? The latter seems newer, but I could be wrong. Maybe the final state is better to allow excluding with a short option.

I think '-e' fits much better for exclude. I think we should remove all the single-letter options that are not clear.

aqk:
> you should give this PR a more specific title, since it will be part of the changelog

Done, thanks!

emlowe:
This is fixed by https://github.com/Chia-Network/chia-blockchain/pull/14100 - which is already merged, so closing this PR

PR #14108, opened 2022-12-12
don't throw unnecessary exception on peer connect

### LINKED ISSUE #14094: [Bug] Bad error message when manually adding a peer that fails to connect
### What happened?

when added manually a peer , if we can't connect, the backend code throws a very generic and unhelpful exception:
```
2022-12-08T15:18:08.115 full_node full_node_server        : INFO     Cannot connect to host 10.1.1.23:3435 ssl:<ssl.SSLContext object at 0x0000018289A90F40> [The semaphore timeout period has expired]
2022-12-08T15:18:08.116 full_node chia.rpc.util           : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia\rpc\util.py", line 18, in inner
  File "chia\rpc\rpc_server.py", line 273, in open_connection
ValueError: Start client failed, or server is not set
```
The CLI handles this exception though, and just returns "Failed to connect":
```
→ C:\Users\elowe\AppData\Local\Programs\Chia\resources\app.asar.unpacked\daemon› chia peer -a 10.1.1.23:3435 full_node
Connecting to 10.1.1.23, 3435
Failed to connect to 10.1.1.23:3435
```
The GUI however, takes the ValueError text of `Start client filed, or server is not set` and displays it to the end-user. Which is of course, entirely unhelpful. 

I think probably several fixes are needed here. Both in the back end and in the GUI

### Version

1.7.0

Windows

GUI

### Relevant log output

```shell
2022-12-08T15:18:08.115 full_node full_node_server        : INFO     Cannot connect to host 10.1.1.23:3435 ssl:<ssl.SSLContext object at 0x0000018289A90F40> [The semaphore timeout period has expired]
2022-12-08T15:18:08.116 full_node chia.rpc.util           : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia\rpc\util.py", line 18, in inner
  File "chia\rpc\rpc_server.py", line 273, in open_connection
ValueError: Start client failed, or server is not set
```

almogdepaz:
dont throw a redundant exception when using the cli to manually connect to a peer
added logs to start_client return paths that had none 
this doesn't change behavior just removes the redundant throw 

fixes https://github.com/Chia-Network/chia-blockchain/issues/14094#event-8001123775

### DIFF
--- a/chia/cmds/peer_funcs.py
+++ b/chia/cmds/peer_funcs.py
@@ -17,7 +17,10 @@ async def add_node_connection(rpc_client: RpcClient, add_connection: str) -> Non
         )
         print(f"Connecting to {ip}, {port}")
         try:
-            await rpc_client.open_connection(ip, int(port))
+            result = await rpc_client.open_connection(ip, int(port))
+            err = result.get("error")
+            if result["success"] is False or err is not None:
+                print(err)
         except Exception:
             print(f"Failed to connect to {ip}:{port}")
 
--- a/chia/rpc/rpc_server.py
+++ b/chia/rpc/rpc_server.py
@@ -270,8 +270,8 @@ class RpcServer:
         if hasattr(self.rpc_api.service, "on_connect"):
             on_connect = self.rpc_api.service.on_connect
         if not await self.rpc_api.service.server.start_client(target_node, on_connect):
-            raise ValueError("Start client failed, or server is not set")
-        return {}
+            return {"success": False, "error": f"could not connect to {target_node}"}
+        return {"success": True}
 
     async def close_connection(self, request: Dict[str, Any]) -> EndpointResult:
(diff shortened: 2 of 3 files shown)


emlowe:
I do like a nice simple change. I do wonder if start_client should return (or raise) different errors (instead of boolean) so we can differentiate why something failed.

almogdepaz:
> I do like a nice simple change. I do wonder if start_client should return (or raise) different errors (instead of boolean) so we can differentiate why something failed.

i first went down that path and its not a nice one, we have some exceptions we raise there, but we suppress lots of others in favor of logging, also theres some nested try/catch there and in the places its being called.

so for the bug this is the correct fix i think, but yes theres lots of tech debt to handle in start_client if we want to do that better.

emlowe (approved):

PR #14111, opened 2022-12-12
Rename confusing variables. 'c_id' suggests coin IDs, but these values are actually SpendBundle IDs.

aqk:
Code cleanup.

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -590,10 +590,10 @@ class MempoolManager:
             if last_npc_result.conds is not None:
                 for spend in last_npc_result.conds.spends:
                     if spend.coin_id in self.mempool.removals:
-                        c_ids: List[bytes32] = self.mempool.removals[bytes32(spend.coin_id)]
-                        self.mempool.remove_from_pool(c_ids)
-                        for c_id in c_ids:
-                            self.remove_seen(c_id)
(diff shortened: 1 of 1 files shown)


almogdepaz (approved):

chia/full_node/mempool_manager.py
```diff
@@ -590,10 +590,10 @@ async def new_peak(
             if last_npc_result.conds is not None:
                 for spend in last_npc_result.conds.spends:
                     if spend.coin_id in self.mempool.removals:
-                        c_ids: List[bytes32] = self.mempool.removals[bytes32(spend.coin_id)]
-                        self.mempool.remove_from_pool(c_ids)
-                        for c_id in c_ids:
-                            self.remove_seen(c_id)
+                        spendbundle_ids: List[bytes32] = self.mempool.removals[bytes32(spend.coin_id)]
```
arvidn, line 593: the name `removals` is also confusing then. That name usually refers to a list of coins that were removed. I guess in this case it's a `removed_coin` -> `SpendBundle` mapping

arvidn (approved):

chia/full_node/mempool_manager.py
aqk: Yep, it sure is. I'll put that in another PR!

PR #14112, opened 2022-12-12
Make docstring match process_coin_spends parameter

aqk:
Code cleanup.


arvidn:
@aqk I just noticed this. I think the docstring right now is fairly meaningless. the function is called `process_coins`, the question isn't whether this function processes coins, it's "what does it do to the coins when it processes them?"

PR #14117, opened 2022-12-13
tools: Add `-v/--verbose` to `manage_clvm.py check`

xdustinface:
Let it only print errors by default so that the `pre-commit` output  doesn't get spammed on success.

### DIFF
--- a/tools/manage_clvm.py
+++ b/tools/manage_clvm.py
@@ -103,7 +103,12 @@ def main() -> None:
 
 
 @main.command()
-def check() -> int:
+@click.option("-v", "--verbose", is_flag=True, help="Increase verbosity")
+def check(verbose: bool) -> int:
+    def print_maybe(message: str = "") -> None:
+        if verbose:
+            print(message)
+
     used_excludes = set()
     overall_fail = False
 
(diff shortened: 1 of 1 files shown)


altendky (changes requested):
How about at least have a successful run provide a summary like mypy, black, and isort?  Something more than the present just-pre-commit output.

https://github.com/Chia-Network/chia-blockchain/actions/runs/3681554402/jobs/6228363263#step:9:40
```
.clvm.hex files..........................................................Passed
- hook id: clvm_hex
- duration: 9.45s
```

For the record, I chatted on keybase with Kevin about using logging and a counting verbosity and such.  They didn't feel it was warranted here.  I would like to see consistent usage of complete solutions, but I also foresee some complexities in initially figuring out using logging, or whatever more formal system, for output and verbosity control.  As such I'm forgoing that effort now in favor of focusing on other opportunities to improve our code.

xdustinface:
As for the verbosity levels, yeah, i don't see why we would would need more than stdout/stderr here. It would probably make more sense to use `click.echo` for normal logs and then set `err=True` for error logs in here but im currently not sure if we even need to change something since it seems stdout is already suppressed by pre-commit if run without `--verbose`. Though, im pretty sure i sometimes still see the output.. and im also pretty sure i could reproduce it before i opened this PR but right now i can't anymore :/ I guess now i first need to figure out why and when it happens :)

PR #14122, opened 2022-12-13
tests: Drop unused `PeerInfo`s

xdustinface:


arvidn:
@fchirica do you know if these are supposed to be used in the test (and not doing so is the problem) or whether they are left-overs that shouldn't be there?

fchirica (approved):
they look like leftovers.

wjblanke (approved):
aok

PR #14123, opened 2022-12-13
server: Drop redundant `VettedPeer.__init__`

xdustinface:


arvidn (approved):
this is unnecessary because the class is already a `dataclass`, right?

xdustinface:
> this is unnecessary because the class is already a `dataclass`, right?

Yeah correct.

PR #14125, opened 2022-12-13
Correct grammatical error

aqk:


chia/full_node/mempool_manager.py
```diff
@@ -557,7 +557,7 @@ def get_spendbundle(self, bundle_hash: bytes32) -> Optional[SpendBundle]:
 
     def get_mempool_item(self, bundle_hash: bytes32, include_pending: bool = False) -> Optional[MempoolItem]:
         """
-        Returns a MempoolItem if it's inside one the mempools.
```
arvidn, line 560: we only have a single mempool, right? (the only exception is while updating the peak, where we temporarily creates a new one to replace the one mempool we have).

but the `bundle_hash` is used to look up a mempool item from `self.mempool.spends`, isn't that a dictionary indexing mempool items by individual spends? From just looking at this code, it looks like `bundle_hash` really is a coin ID that's being spent.

aqk: > we only have a single mempool, right? (the only exception is while updating the peak, where we temporarily creates a new one to replace the one mempool we have).

We have one conceptual mempool. You could also say we only have one actual mempool, if you distinguish it from `MempoolManager.potential_cache`. Whoever wrote this comment was probably including the pending cache in the idea of "mempool".


> but the `bundle_hash` is used to look up a mempool item from `self.mempool.spends`, isn't that a dictionary indexing mempool items by individual spends? From just looking at this code, it looks like `bundle_hash` really is a coin ID that's being spent.

Well, I'm not telling until we get to my PRs that add the `SpendBundleID` and `CoinID` types.

emlowe:
Closing PR as this text was updated in #14300

PR #13325, opened 2022-09-04
Coin Selection CLI / RPC Expansion

jack60612:
Add New Cli Commands...

- [x] `chia wallet coins list`: List all coins
- [x] `chia wallet coins combine`: allow users to manage dust and aggregate smaller coins.
- [x]  `chia wallet coins split`: Allow users to split a larger coin into a large amount of smaller coins.

Also:

- [x] expand existing RPC's with max coin amount
- [x] add ability to exclude specific coin when using `chia wallet send`
- [x] Add RPC that outputs all coins that pass a set of filters.
- [x] Add wallet coin record rpc & various coin record functions that interact between datastores

Please test the commands if you can!

PR #14127, opened 2022-12-13
Fees should always be in XCH mojos units

### LINKED ISSUE #14118: Sending a CAT with a fee in the Windows CLI is not producing the same effect as sending a CAT with a fee in the Windows GUI (this might be true on other systems)
CAT sent with a fee from the GUI:
Image

CAT sent with a fee from the CLI:
Image

emlowe:
Fixes bug introduced in #13325 where the CLI would use CAT units for fee calculations instead of XCH mojos.
Fees should always be in XCH mojos

Currently, when sending CATs using the CLI, the fee will be calculated incorrectly and is multiplied by the incorrect multiplier

This fix puts back the proper behaviour.

This resolves #14118 

This is a fix to a bug introduced in not-yet-released code.

### DIFF
--- a/chia/cmds/wallet_funcs.py
+++ b/chia/cmds/wallet_funcs.py
@@ -219,7 +219,7 @@ async def send(args: dict, wallet_client: WalletRpcClient, fingerprint: int) -> 
         print(f"Wallet id: {wallet_id} not found.")
         return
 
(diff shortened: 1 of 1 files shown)


wallentx (approved):

altendky (approved):

wjblanke (approved):
aok

PR #14128, opened 2022-12-13
Fix display issues when listing coins from a CAT wallet

emlowe:
When using `chia wallet coins list` on a CAT wallet, the display would indicate the amount of the coins in XCH/TXCH and then apply the CAT conversion to the mojos, so you would get strange output like

```
Coin ID: 0xa7a087b1d3f49937dcada65f343a0797ebc75c39176c58629689e426355a1baa
	Address: txch1fckd972smrl3pxn5z5ekeqp2nzlfnumjk8l7uulen8rc0kaur9gscy9fq8 Amount: 9.000000000000 txch (9000 mojo), Confirmed in block: 1930591
```

Where clearly 9 TXCH does not equal 9000 mojos.

This PR fixes the display to use the word CAT instead of XCH/TXCH

```
Coin ID: 0xa7a087b1d3f49937dcada65f343a0797ebc75c39176c58629689e426355a1baa
	Address: txch1fckd972smrl3pxn5z5ekeqp2nzlfnumjk8l7uulen8rc0kaur9gscy9fq8 Amount: 9.000000000000 CAT (9000 mojo), Confirmed in block: 1930591
```

This is a fix for a new feature not yet released

### DIFF
--- a/chia/cmds/coin_funcs.py
+++ b/chia/cmds/coin_funcs.py
@@ -12,6 +12,7 @@ from chia.types.coin_record import CoinRecord
 from chia.util.bech32m import decode_puzzle_hash, encode_puzzle_hash
 from chia.util.ints import uint64, uint128
 from chia.wallet.transaction_record import TransactionRecord
+from chia.wallet.util.wallet_types import WalletType
 
 
 async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, fingerprint: int) -> None:
@@ -26,7 +27,7 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
     if paginate is None:
         paginate = sys.stdout.isatty()
     try:
-        wallet_type = await get_wallet_type(wallet_id=wallet_id, wallet_client=wallet_client)
+        wallet_type: WalletType = await get_wallet_type(wallet_id=wallet_id, wallet_client=wallet_client)
         mojo_per_unit = get_mojo_per_unit(wallet_type)
     except LookupError:
         print(f"Wallet id: {wallet_id} not found.")
@@ -44,6 +45,10 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
         excluded_amounts=final_excluded_amounts,
         excluded_coin_ids=excluded_coin_ids,
     )
+    coin_type: str = "XCH" if addr_prefix == "xch" else "TXCH"
+    if wallet_type == WalletType.CAT:
+        coin_type = "CAT"  # override XCH/TXCH prefix for CAT wallet
+
     print(f"There are a total of {len(conf_coins) + len(unconfirmed_additions)} coins in wallet {wallet_id}.")
     print(f"{len(conf_coins)} confirmed coins.")
     print(f"{len(unconfirmed_additions)} unconfirmed additions.")
@@ -54,6 +59,7 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
         [(cr.coin, str(cr.confirmed_block_index)) for cr in conf_coins],
         mojo_per_unit,
         addr_prefix,
+        coin_type,
         paginate,
     )
     if show_unconfirmed:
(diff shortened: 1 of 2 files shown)


wallentx (dismissed):

altendky (changes requested):
I thought we had code to output the proper CAT ticker/name in the CLI.  Also, a CAT amount shouldn't be shown with 12 decimal places, but instead three.  Feel free to handle that separately though.

chia/cmds/coin_funcs.py
```diff
@@ -71,12 +78,18 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
...
         )
 
 
 def print_coins(
-    target_string: str, coins: List[Tuple[Coin, str]], mojo_per_unit: int, addr_prefix: str, paginate: bool
+    target_string: str,
+    coins: List[Tuple[Coin, str]],
+    mojo_per_unit: int,
+    addr_prefix: str,
+    coin_type: str,
```
altendky, line 91: This seems more like the ticker or name than the 'type'?

chia/cmds/coin_funcs.py
```diff
@@ -44,6 +45,10 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
         excluded_amounts=final_excluded_amounts,
         excluded_coin_ids=excluded_coin_ids,
     )
+    coin_type: str = "XCH" if addr_prefix == "xch" else "TXCH"
+    if wallet_type == WalletType.CAT:
+        coin_type = "CAT"  # override XCH/TXCH prefix for CAT wallet
```
altendky, line 50: https://github.com/Chia-Network/chia-blockchain/blob/d13f8b837fd585c122b4c7366bf723aea1d4affd/chia/cmds/show_funcs.py#L131

I'd probably prefer an `if`/`elif`/`else` over picking the distinctly wrong thing first.

```suggestion
    if wallet_type == WalletType.XCH:
        coin_type: str = "XCH" if addr_prefix == "xch" else "TXCH"
    elif wallet_type == WalletType.CAT:
        coin_type = "CAT"  # override XCH/TXCH prefix for CAT wallet
    else:
        <something, i dunno what we want to do>
```

This is also relevant I think.

https://github.com/Chia-Network/chia-blockchain/blob/d13f8b837fd585c122b4c7366bf723aea1d4affd/chia/cmds/wallet_funcs.py#L87-L100

chia/cmds/coin_funcs.py
```diff
@@ -26,7 +27,7 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
     if paginate is None:
         paginate = sys.stdout.isatty()
     try:
-        wallet_type = await get_wallet_type(wallet_id=wallet_id, wallet_client=wallet_client)
+        wallet_type: WalletType = await get_wallet_type(wallet_id=wallet_id, wallet_client=wallet_client)
```
altendky, line 30:
```suggestion
        wallet_type = await get_wallet_type(wallet_id=wallet_id, wallet_client=wallet_client)
```

chia/cmds/coin_funcs.py
emlowe: yea, I worried about getting the entire CAT name to display here. since it could be quite long to display in each coin

emlowe: This is also why I went with the simpler form of naming it first - it should be XCH/TXCH in most cases, except only for CAT wallets so an alternative would be 
```
    if wallet_type == WalletType.CAT:
        coin_type = "CAT"  # override XCH/TXCH prefix for CAT wallet
    else 
        coin_type: str = "XCH" if addr_prefix == "xch" else "TXCH"
```

chia/cmds/coin_funcs.py
emlowe: changed variable name

chia/cmds/coin_funcs.py
emlowe: changed

chia/cmds/coin_funcs.py
emlowe: changed

emlowe:
Closing due to removable of feature

PR #14140, opened 2022-12-14
Add diff-cover report

altendky:
One measure of code coverage when making a change is whether the lines that were modified are covered by tests.  This PR uses the [diff-cover](https://pypi.org/project/diff-cover/) package to provide this data.  Aside from running this, it would be helpful to surface it better to the developers and reviewers.  At some point we can also consider making a CI check that requires coverage of all lines.

https://github.com/Chia-Network/chia-blockchain/pull/14141 adds some not covered code and branching as a test for this PR.  See https://github.com/Chia-Network/chia-blockchain/actions/runs/4000252355 for a sample output.

Draft for:
- [x] https://github.com/Chia-Network/chia-blockchain/pull/14164


altendky:
fyi @arvidn @wallentx

arvidn:
is this to be expected?
```
Run diff-cover --compare-branch=origin/main --html-report coverage-reports/diff-cover.html --markdown-report coverage-reports/diff-cover.md coverage-reports/coverage.xml | tee coverage-reports/diff-cover-stdout
-------------
Diff Coverage
-------------
No lines with coverage information in this diff.
-------------
```

arvidn (approved):

altendky:
This changes no Python code so yes, there is no relevant diff against `main` for diff-cover to check.  That's why I provided the links to the other test PR in the OP so an actual report could be seen including not-covered changes.

PR #14144, opened 2022-12-14
Improve mempool conflict message when replacement spendbundle is detected

aqk:


chia/full_node/mempool_manager.py
```diff
@@ -512,7 +512,10 @@ async def validate_spend_bundle(
                     sb: MempoolItem = self.mempool.spends[c_sb_id]
                     conflicting_pool_items[sb.name] = sb
             log.debug(f"Replace attempted. number of MempoolItems: {len(conflicting_pool_items)}")
-            if not self.can_replace(conflicting_pool_items, removal_record_dict, fees, fees_per_cost):
+            can_replace = self.can_replace(conflicting_pool_items, removal_record_dict, fees, fees_per_cost)
+            log.warning(f"Conflicting pool items: {len(conflicting_pool_items)} can_replace={can_replace}")
+            if not can_replace:
+                log.warning(f"Cannot replace SpendBundles: {[c for c in conflicting_pool_items]}")
```
arvidn, line 518: I don't think this should be a warning. The person running this node most likely couldn't care less about someone attempting a replace and failing.

Also, the `can_replace()` function already logs the reason why it couldn't replace, but as a debug message (iirc)

arvidn (changes requested):

chia/full_node/mempool_manager.py
aqk: Yes, definitely should not be a warning.

aqk: Okay, since `can_replace()` log it, I'm closing this PR.

PR #14145, opened 2022-12-14
Supply a reason when removing an item from the mempool

aqk:
No user visible changes. This change will let the mempool know why a caller is removing a SpendBundle.

This is a required part of the Fee Estimator backend

### DIFF
--- a/chia/full_node/mempool.py
+++ b/chia/full_node/mempool.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 from datetime import datetime
+from enum import IntEnum
 from typing import Dict, List, Optional
 
 from sortedcontainers import SortedDict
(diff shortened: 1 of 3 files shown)


chia/full_node/mempool.py
```diff
@@ -17,6 +18,18 @@
 from chia.util.ints import uint64
 
 
+class MempoolRemoveReason(IntEnum):
```
arvidn, line 21: this looks just like an enum. does this have any benefits over an enum?

chia/full_node/mempool.py
```diff
@@ -73,7 +86,8 @@ def remove_from_pool(self, items: List[bytes32]) -> None:
             self.total_mempool_cost -= item.cost
             assert self.total_mempool_cost >= 0
             mempool_info = self.get_mempool_info()
-            self.fee_estimator.remove_mempool_item(mempool_info, item)
+            if reason != MempoolRemoveReason.BLOCK_INCLUSION:
```
arvidn, line 89: this looks like a change in behavior. would you mind explaining it?

chia/full_node/mempool.py
aqk: Nope! I just got in the habit of using IntEnum from us using Streamable to manage on-the-wire values. I'll change that to Enum.

chia/full_node/mempool.py
aqk: Yes. This is the first in a series of small fixes that fix broken behaviour in the fee estimator, specifically how and where the code is called from.

This particular line fixes the issue that MempoolItems were being marked as removed in the fee estimator even though the reason they were being removed was that they had successfully gotten into a block.

aqk: I could take that line out and put it in the PR that does have a user-visible fix.

Let me know what you think.

aqk: Yeah, I think I'll do that.

arvidn (dismissed):

almogdepaz (approved):

PR #14148, opened 2022-12-15
Document reasoning for type of MempoolItem.spend_bundle_name

aqk:
No user visible changes. No code changes. No effect. Comment.


almogdepaz (approved):

arvidn:
The `SpendBundleID` *is* a `bytes32`, right? it has the `parse()`, `stream()`, `__bytes__()`, `from_bytes()` and `get_hash()` methods on it, I would expect. What fails if you use this type instead of `bytes32`?

PR #14151, opened 2022-12-15
server:  `peer_host` + `peer_port` -> `peer_info` in `WSChiaConnection`

xdustinface:


almogdepaz:
add a  host and port @property to ws_connection instead would be nicer transition and shorter calls all over

xdustinface:
> add a host and port @Property to ws_connection instead, would be nicer transition and shorter calls all over.

Thought about it but i don't feel like it's really required here, sure it would be a smaller diff here but its not a lot places and this way we get away without the two extra properties. At the end there are anyway a bunch of places where `host` gets logged where i think we could also in another PR just print the whole `PeerInfo` instead.

almogdepaz:
> > add a host and port @Property to ws_connection instead, would be nicer transition and shorter calls all over.
> 
> Thought about it but i don't feel like it's really required here, sure it would be a smaller diff here but its not a **lot** places and this way we get away without the two extra properties. At the end there are anyway a bunch of places where `host` gets logged where i think we could also in another PR just print the whole `PeerInfo` instead (although this might not be very useful for the most part because ports are kind of fix anyway).

i think its nicer to have access to the host/port from connection without needing to go through the extra parameter in the way, not sure why the objection or what the cost here ?

almogdepaz:
> @almogdepaz I think i have no rational objections at this point other than my personal preference of not having the names `peer_host` and `peer_port` separately spread around like this, actually i even liked getting rid of those with this PR :) But thats maybe only me and the reason why i came up with it like this rather might be buried somewhere down the road in the other branches i work on in this area to refactor away from using `host: str` all over the codebase to use `ip: IPAddress` instead and another thing which is to split up `WSChiaConnection` into two classes, `Peer` and `Connection`. So yeah, i want to get away from the `host` thing and i don't like the `peer_` prefixes much for them this way. I guess it will be some back and forth with renaming and whatnot either way. I get your point and if you want to see the change still let me know, i can add it in few hours when im back, if not i can come up with other properties later down the road.

you are still using the names peer_info.host and  peer_info.port everywhere so i dont really get your point, for the other reason you mentioned not sure i understand why that conflicts with using property on ws_connection if you say it will only make sense with the other prs merged than i think they should all be merged together

xdustinface:
> you are still using the names peer_info.host and peer_info.port everywhere so i dont really understand the difference. 

Right but when reading `connection.peer_info.host` its obvious that we access `peer_info` (which contains two bundled attributes) while it isn't if you read `connection.peer_host` which is maybe not groundbreaking (but at least i like it) and at the end the difference is just that explicit case has 5 characters more.. so i don't really understand the concerns.

Its not that it is a huge or complex PR without those properties.. but yeah, its not a big thing to add them either so if you still can't follow my thoughts and want to see this change feel free to leave a change request :)

> but if you say it will make sense together with the other change's merged than i dont think they should be merged separately

Not sure i get why you have concerns about this. I mean, in this case now its just our personal preference is different and there is no "makes sense" in either of the both ways imo.. so, i don't see how this makes it bad to split up changes.

emlowe (approved):
looks fine

PR #14163, opened 2022-12-15
Add a height timestamp for wallet notifications

Quexington:
Probably good to test a wallet that already has notifications before this patch starts up correctly with the patch.  I would write a unit test for it, but I'm not sure how to restart the wallet in the middle of a test (suggestions welcome)

### DIFF
--- a/chia/wallet/notification_store.py
+++ b/chia/wallet/notification_store.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
+import sqlite3
 import logging
 from typing import List, Optional, Tuple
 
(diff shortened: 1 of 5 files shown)


chia/wallet/notification_store.py
```diff
@@ -44,6 +46,11 @@ async def create(
                 "CREATE TABLE IF NOT EXISTS notifications(" "coin_id blob PRIMARY KEY," "msg blob," "amount blob" ")"
             )
 
+            try:
+                await conn.execute("ALTER TABLE notifications ADD COLUMN height bigint DEFAULT 0")
+            except sqlite3.OperationalError:
+                pass  # ignore what is likely Duplicate column error
```
arvidn, line 52: I would feel better if this was checked specifically for the error we want to ignore, and any other errors were re-thrown.

I suspect we also don't have any automates tests exercising this case, which makes it extra risky to ignore errors

Quexington: I checked for a string in the error message does that work?

trepca (approved):

PR #14165, opened 2022-12-16
simulator: Replace `localhost` -> `127.0.0.1` in farmer/harvester config

xdustinface:
We create new configs in `setup_farmer` and `setup_harvester` and they  contain `localhost` defaults. Replace it here also like we do it in  `BlockTools` to consistently use `127.0.0.1` instead of `localhost` in  tests.

This fixes test failures in #14152 and #14071.

### DIFF
--- a/chia/simulator/setup_services.py
+++ b/chia/simulator/setup_services.py
@@ -4,9 +4,10 @@ import asyncio
 import logging
 import signal
 import sqlite3
+from contextlib import contextmanager
 from pathlib import Path
 from secrets import token_bytes
-from typing import Any, AsyncGenerator, List, Optional, Tuple
+from typing import Any, AsyncGenerator, Dict, Iterator, List, Optional, Tuple
 
 from chia.cmds.init_funcs import init
 from chia.consensus.constants import ConsensusConstants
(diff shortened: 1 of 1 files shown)


chia/simulator/setup_services.py
```diff
@@ -39,6 +40,16 @@
 log = logging.getLogger(__name__)
 
 
+@contextmanager
+def create_lock_and_load_config(certs_path: Path, root_path: Path) -> Iterator[Dict[str, Any]]:
+    init(None, root_path)
+    init(certs_path, root_path)
+    path = config_path_for_filename(root_path=root_path, filename="config.yaml")
+    path.write_text(path.read_text().replace("localhost", "127.0.0.1"))
```
arvidn, line 48: it doesn't look like this only affects tests. does this make sense to do in production too?

xdustinface: Actually this is meant to be for tests only, there was some flakiness issue a while ago with using `localhost` and then we switched over to using the IP instead, see https://github.com/chia-network/chia-blockchain/pull/13324.

In this case here it's not about flakiness but rather about the recent PR to resolve the peer outside the reconnect task which we don't do in tests hence the two PRs linked in here fail in CI without this PR.

I guess in production we don't want to get rid of the `localhost` because we also want to cover ipv6?

emlowe:
Using `localhost` is important for IPv6 hosts - I think even in the tests we should maybe consider using `localhost` so we can get some IPv6 testing in CI. We don't do that now but we probably should.

arvidn:
> Using `localhost` is important for IPv6 hosts - I think even in the tests we should maybe consider using `localhost` so we can get some IPv6 testing in CI. We don't do that now but we probably should.

If we want to test IPv6, I think we should have specific tests that do that deliberately. We can't run every test both over IPv4 and IPv6, and let some arbitrary CI configuration decide which we test over might make it quite hard to know what the true cause of a failure is. It's the classic problem of having a single test test too many things, so when it fails all you know is that something isn't working.

arvidn:
I think it's reasonable to use 127.0.0.1 in tests, but this patch affects the production code, not just tests

xdustinface:
>  but this patch affects the production code, not just tests

Where/how?

chia/simulator/setup_services.py
arvidn: ok, I didn't realize this was in the simulator at first. So, this primarily affects tests. However, I would expect it to also affect users of the "stand-alone" simulator. It probably doesn't matter much though

chia/simulator/setup_services.py
```diff
@@ -39,6 +40,17 @@
 log = logging.getLogger(__name__)
 
 
+@contextmanager
+def create_lock_and_load_config(certs_path: Path, root_path: Path) -> Iterator[Dict[str, Any]]:
+    init(None, root_path)
+    init(certs_path, root_path)
+    path = config_path_for_filename(root_path=root_path, filename="config.yaml")
+    # Using localhost leads to flakiness on CI
```
arvidn, line 48: that's a pretty sparse comment. I assume because some machines resolve it to IPv6, while still not supporting IPv6 or not binding our listen socket to both IPv4 and IPv6.

arvidn (approved):

PR #14166, opened 2022-12-16
types: `PeerInfo` doesn't need to be streamable

xdustinface:


arvidn:
This looks right, but would our tests really fail if `PeerInfo` is used as a member of a streamable type?
We do some set-up work with streamable types, does that ensure that all members are themselves streamable, transitively?

arvidn (approved):

xdustinface:
>  but would our tests really fail if `PeerInfo` is used as a member of a streamable type?

Yes this would lead to a failure on startup in the streamable decorator when the class which contains the `PeerInfo` would be loaded.

PR #14167, opened 2022-12-16
types: Use `IPAddress` instead of `str` in `PeerInfo`

xdustinface:
Working towards using `IPAddress` instead of `str` all over / where it makes sense.

Based on:

- #14166

### DIFF
--- a/chia/server/address_manager.py
+++ b/chia/server/address_manager.py
@@ -300,7 +300,7 @@ class AddressManager:
     def mark_good_(self, addr: PeerInfo, test_before_evict: bool, timestamp: int) -> None:
         self.last_good = timestamp
         (info, node_id) = self.find_(addr)
-        if not addr.is_valid(self.allow_private_subnets):
+        if addr.ip.is_private and not self.allow_private_subnets:
             return None
         if info is None:
             return None
@@ -365,7 +365,7 @@ class AddressManager:
             addr.host,
             addr.port,
         )
-        if not peer_info.is_valid(self.allow_private_subnets):
+        if peer_info.ip.is_private and not self.allow_private_subnets:
             return False
         (info, node_id) = self.find_(peer_info)
         if info is not None and info.peer_info == peer_info:
@@ -554,7 +554,7 @@ class AddressManager:
             rand_pos = randrange(len(self.random_pos) - n) + n
             self.swap_random_(n, rand_pos)
             info = self.map_info[self.random_pos[n]]
-            if not info.peer_info.is_valid(self.allow_private_subnets):
+            if info.peer_info.ip.is_private and not self.allow_private_subnets:
                 continue
             if not info.is_terrible():
                 cur_peer_info = TimestampedPeerInfo(
--- a/chia/server/node_discovery.py
+++ b/chia/server/node_discovery.py
@@ -25,6 +25,7 @@ from chia.server.ws_connection import WSChiaConnection
 from chia.types.peer_info import PeerInfo, TimestampedPeerInfo
 from chia.util.hash import std_hash
 from chia.util.ints import uint16, uint64
+from chia.util.network import IPAddress, get_host_addr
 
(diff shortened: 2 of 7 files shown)


chia/server/address_manager.py
```diff
@@ -300,7 +300,7 @@ def clear_new_(self, bucket: int, pos: int) -> None:
     def mark_good_(self, addr: PeerInfo, test_before_evict: bool, timestamp: int) -> None:
         self.last_good = timestamp
         (info, node_id) = self.find_(addr)
-        if not addr.is_valid(self.allow_private_subnets):
+        if addr.ip.is_private and not self.allow_private_subnets:
```
arvidn, line 303: it's not obvious that the new condition is the same as the old. Is being in a private subnet while private subnets are disallowed the only way an IP can be invalid in?
The left side suggests the IP address can be invalid in more ways

chia/server/node_discovery.py
```diff
@@ -62,7 +63,7 @@ def __init__(
         random.shuffle(dns_servers)  # Don't always start with the same DNS server
         if introducer_info is not None:
             self.introducer_info: Optional[PeerInfo] = PeerInfo(
-                introducer_info["host"],
+                str(get_host_addr(introducer_info["host"], prefer_ipv6=False)),
```
arvidn, line 66: is it OK to make a blocking call here? I assume `get_host_addr()` is blocking in case it needs to perform a DNS lookup. If it's OK, I think it warrants a comment in the code to make it clear to readers

chia/server/server.py
```diff
@@ -419,14 +419,8 @@ async def start_client(
...
-
-            try:
-                if type(ip_address(target_node.host)) is IPv6Address:
-                    target_node = PeerInfo(f"[{target_node.host}]", target_node.port)
-            except ValueError:
-                pass
-
-            url = f"wss://{target_node.host}:{target_node.port}/ws"
+            ip = f"[{target_node.ip}]" if target_node.ip.is_v6 else f"{target_node.ip}"
+            url = f"wss://{ip}:{target_node.port}/ws"
```
arvidn, line 423: maybe in the future it would make sense to make `PeerInfo` have a `__str__()` that contains this to-string logic

tests/simulation/test_simulator.py
```diff
@@ -251,7 +251,7 @@ async def test_create_coins_with_invalid_amounts_raises(
 ) -> None:
     [[full_node_api], [[wallet_node, wallet_server]], _] = simulator_and_wallet
 
-    await wallet_server.start_client(PeerInfo("localhost", uint16(full_node_api.server._port)), None)
+    await wallet_server.start_client(PeerInfo("127.0.0.1", uint16(full_node_api.server._port)), None)
```
arvidn, line 254: is this change necessary because we no longer support hostnames or is it a matter of reducing test flakiness?

arvidn (commented):
looks good generally. Just a few questions

tests/simulation/test_simulator.py
xdustinface: The former, yes.

chia/server/address_manager.py
xdustinface: It's just, before we validated the host in [PeerInfo.is_valid](https://github.com/Chia-Network/chia-blockchain/blob/99abc2d66f0136bf6d0420a9c3163664987a04ee/chia/types/peer_info.py#L16), see below. But with this PR we store valid IP addresses in `PeerInfo` so `is_valid` is not longer needed and gets dropped. The condition `addr.ip.is_private and not self.allow_private_subnets` is just the private subnet check which was nested into `PeerInfo.is_valid` for some reason.

chia/server/node_discovery.py
xdustinface: See 5009987591390d12aee466339e5d30e02709bcae.

arvidn (approved):

PR #14168, opened 2022-12-16
Rename Mempool.removals, which is actually a map from coin_id to a list of SpendBundleIDs

aqk:
Code maintenance.


arvidn (approved):
do we allow the same coin to be spent in more than one SpendBundle?
We're working on identical spend deduplication (which would allow this).

PR #14172, opened 2022-12-16
Fix fee and unit amounts when coin splitting or viewing coins

### LINKED ISSUE #14120: "Coinsplitting: fees showing up in transaction list as a different fee than specified."

### LINKED ISSUE #14131: Various CLI issues with chia wallet coins {list,split.combine}
Fixed by
https://github.com/Chia-Network/chia-blockchain/pull/14102
https://github.com/Chia-Network/chia-blockchain/pull/14100

jack60612:
This PR fixes the fee being improperly converted to cat's instead of chia.
Additionally, this pr removes the unit type in `chia coins list` that was causing confusion 
This resolves issue #14120.


jack60612:
I think cli tests are a bigger issue and should not be included in this pr.

chia/cmds/coin_funcs.py
```diff
@@ -176,7 +176,7 @@ async def async_combine(args: Dict[str, Any], wallet_client: WalletRpcClient, fi
 async def async_split(args: Dict[str, Any], wallet_client: WalletRpcClient, fingerprint: int) -> None:
     wallet_id: int = args["id"]
     number_of_coins = args["number_of_coins"]
-    fee = Decimal(args["fee"])
+    final_fee = uint64(int(Decimal(args["fee"] * units["chia"])))
```
arvidn, line 179: what's the point of creating a `Decimal` object and an `int` object here? `args` is presumably some `int`-like object since you can multiply it. So, after the multiplication, I would expect we would already have `int` and could just pass it directly to `uint64()`. What am I missing?

chia/cmds/coin_funcs.py
```diff
@@ -88,7 +89,7 @@ def print_coins(
                 break
             coin, conf_height = coins[i + j]
             address = encode_puzzle_hash(coin.puzzle_hash, addr_prefix)
-            amount_str = print_balance(coin.amount, mojo_per_unit, addr_prefix, decimal_only=True)
+            amount_str = print_balance(coin.amount, mojo_per_unit, "", decimal_only=True)
```
arvidn, line 92: right. the address prefix doesn't tell us what kind of asset this is. We *could* have a lookup table to recognize known assets though. But it makes sense to just not say anything for now

arvidn (commented):
looks right. I just had one question

chia/cmds/coin_funcs.py
jack60612: args is a float / decimal because we express it in xch, once we convert it to chia thats not the case and its always an int.

arvidn: is it not the same to say this:
```
 final_fee = uint64(args["fee"] * units["chia"])
```
?

jack60612: mypy complains if you do that, we use decimal because args["fee"] is a str

arvidn: in that case, I'm pretty sure that multiplication is a bug

jack60612: > in that case, I'm pretty sure that multiplication is a bug

?

arvidn: you want to convert the string to a number and then multiply it, right?
What this line does currently is to "multiply" the string, which will duplicate it. You may end up with a string containing multiple decimal points. I would hope `Decimal()` throws if it's passed such string though

arvidn (dismissed):

chia/cmds/wallet.py
```diff
@@ -925,8 +924,7 @@ def nft_get_info_cmd(
 
 
 # Keep at bottom.
-# Remove coins subcommand from wallet for now pending further testing
-# wallet_cmd.add_command(coins_cmd)
+wallet_cmd.add_command(coins_cmd)
```
arvidn, line 927: This adds back a lot more commands than the ones you just fixed. Are you confident that the other ones are OK?

jack60612: they should be, and work when i checked but ill have someone double check right now.

chia/cmds/coin_funcs.py
```diff
@@ -176,7 +176,7 @@ async def async_combine(args: Dict[str, Any], wallet_client: WalletRpcClient, fi
 async def async_split(args: Dict[str, Any], wallet_client: WalletRpcClient, fingerprint: int) -> None:
```
arvidn, line 176: if `args` had been typed as `Dict[str, str]` it would have been easier to spot this bug, Are there examples of args that are not strings?

jack60612: yep, wallet id.

jack60612: > yep, wallet id.

only floats are strings, normal ints are ints.

arvidn (dismissed):

emlowe (dismissed):
I think the other commands were already fixed up. This might be a good first candidate for click CliRunner tests

jack60612:
> I think the other commands were already fixed up. This might be a good first candidate for click CliRunner tests

me and @danieljperry are still working on it, not sure how to re mark it as draft.

emlowe (approved):
looks ok

PR #14174, opened 2022-12-17
Don't rely on noticing self.limit_factor, instead do the block limit calculation once.

aqk:
Correct some non-essential block size calculations by computing the value once.

This is a required part of the Fee Estimator backend.

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -520,7 +520,7 @@ class MempoolManager:
         log.log(
             logging.DEBUG if duration < 2 else logging.WARNING,
             f"add_spendbundle {spend_name} took {duration:0.2f} seconds. "
-            f"Cost: {cost} ({round(100.0 * cost/self.max_block_clvm_cost, 3)}% of max block cost)",
+            f"Cost: {cost} ({round(100.0 * cost/self.constants.MAX_BLOCK_COST_CLVM, 3)}% of max block cost)",
         )
 
         return None, potential, list(conflicting_pool_items.keys())
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -86,7 +86,6 @@ class MempoolManager:
     seen_bundle_hashes: Dict[bytes32, bytes32]
     get_coin_record: Callable[[bytes32], Awaitable[Optional[CoinRecord]]]
     nonzero_fee_minimum_fpc: int
-    limit_factor: float
     mempool_max_total_cost: int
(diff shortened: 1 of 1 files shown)


chia/full_node/mempool_manager.py
```diff
@@ -134,7 +134,7 @@ def __init__(
         self.mempool: Mempool = Mempool(
             self.mempool_max_total_cost,
             uint64(self.nonzero_fee_minimum_fpc),
-            uint64(self.constants.MAX_BLOCK_COST_CLVM),
+            uint64(self.max_block_clvm_cost),
```
arvidn, line 137: this is a bit gnarly. by making this depend on the fill rate, you also make the fee estimator depend on the fill rate. i.e. a node with a fill rate of 1 will give much lower estimates than a node with fill rate 0.5.
In a way it makes it more accurate now, assuming no farmers change this constant

chia/full_node/mempool_manager.py
```diff
@@ -520,7 +520,7 @@ async def validate_spend_bundle(
         log.log(
             logging.DEBUG if duration < 2 else logging.WARNING,
             f"add_spendbundle {spend_name} took {duration:0.2f} seconds. "
-            f"Cost: {cost} ({round(100.0 * cost/self.constants.MAX_BLOCK_COST_CLVM, 3)}% of max block cost)",
+            f"Cost: {cost} ({round(100.0 * cost/self.max_block_clvm_cost, 3)}% of max block cost)",
```
arvidn, line 523: I'm not sold on this change. I think it still makes sense to log the fill rate relative to the consensus rules

chia/full_node/mempool_manager.py
aqk: Well, if the estimator doesn't have the true size for max mempool size and max block size, that can make a lot of trouble (Hmm, even more trouble if the mempool doesn't have the correct size).

chia/full_node/mempool_manager.py
aqk: Yes, I could go either way here. Or even tell them both (X% of consensus, Y% of YOUR configuration).

chia/full_node/mempool_manager.py
arvidn: yeah, this rate limit is probably close enough to a constant though. I think the point I was trying to make is that what the fee estimator *really* is interested in is the next winning farmer's block cost limit. And this is probably a good estimate for that (until we bump the rate limit to 1)

arvidn (approved):

PR #14175, opened 2022-12-17
Fix fee estimator being re-created every block

aqk:
Hold a ref to the fee estimator when the Mempool is destroyed and re-created. Unify the sharing of information between Mempool and MempoolManager.

This is a required part of the Fee Estimator backend.


chia/full_node/mempool.py
```diff
@@ -70,10 +64,10 @@ def remove_from_pool(self, items: List[bytes32]) -> None:
             dic = self.sorted_spends[item.fee_per_cost]
             if len(dic.values()) == 0:
                 del self.sorted_spends[item.fee_per_cost]
-            self.total_mempool_cost -= item.cost
-            assert self.total_mempool_cost >= 0
-            mempool_info = self.get_mempool_info()
-            self.fee_estimator.remove_mempool_item(mempool_info, item)
+            current_mempool_cost = self.mempool_info.current_mempool_cost - item.cost
+            self.mempool_info = dataclasses.replace(self.mempool_info, current_mempool_cost=current_mempool_cost)
```
arvidn, line 68: this is a bit awkward. it looks like this should be a member function on `mempool_info`

chia/full_node/mempool.py
```diff
@@ -99,23 +93,14 @@ def add_to_pool(self, item: MempoolItem) -> None:
             if coin_id not in self.removals:
                 self.removals[coin_id] = []
             self.removals[coin_id].append(item.name)
-        self.total_mempool_cost += item.cost
 
-        mempool_info = self.get_mempool_info()
-        self.fee_estimator.add_mempool_item(mempool_info, item)
+        current_mempool_cost = self.mempool_info.current_mempool_cost + item.cost
+        self.mempool_info = dataclasses.replace(self.mempool_info, current_mempool_cost=current_mempool_cost)
```
arvidn, line 98: here too. should `mempool_info` just be part of the fee estimator?

arvidn (commented):
it does look like a simplification to some extent, but it also complicates a few places. it looks like there is a better data structure or abstraction surface

chia/full_node/mempool.py
aqk: Yes, this part could definitely be better. I might be able to split this struct up.

I'll think about this part and update this patch.

chia/full_node/mempool.py
aqk: Not as-is, no. There are parts that update on every call from the mempool into the estimator. (current mempool size, for example).

aqk: So far, I'm thinking that all I need to do is pull current_mempool_cost out - the other members are set at MempoolManager construction time. I'll try that first.

chia/full_node/mempool.py
aqk: I moved `current_mempool_cost` / `total_mempool_cost` back to the way it was, and capture it only when we need to send the value down into the Fee Estimator. This way should work much better.

chia/full_node/mempool.py
```diff
@@ -1,32 +1,30 @@
...
 from chia.types.blockchain_format.sized_bytes import bytes32
-from chia.types.clvm_cost import CLVMCost
-from chia.types.fee_rate import FeeRate
 from chia.types.mempool_item import MempoolItem
-from chia.util.ints import uint64
 
 
 class Mempool:
-    def __init__(self, max_size_in_cost: int, minimum_fee_per_cost_to_replace: uint64, max_block_cost_clvm: uint64):
+    def __init__(self, mempool_info: FeeMempoolInfo, fee_estimator: FeeEstimatorInterface):
```
aqk, line 17: By the way, I'm not too jazzed about passing more info into the Mempool constructor than it needs, but as the Americans say "it is what it is".

Right now, we re-create the Mempool every block. The correct place for the calls into the fee estimator is behind `add_to_pool` and `remove_from_pool`, but the MempoolManager has to hold the fee_estimator ref to ensure it lives for the lifetime of the FullNode.

If we want to decouple more, we could pass some lambdas at Mempool construction time, or something else. I'm open to ideas!

chia/full_node/mempool.py
aqk: I like your idea of letting FeeMempoolInfo live in the fee estimator - I think that would be better overall.
As you can see, these are mostly constant throughout the life of the FullNode.

A couple downsides are:
* Data is different places, possibly duplicate
* No ability to dynamically set some of these values (like max mempool size) without restart

I'm not tied to point number 2 - the scenario I thought it would be useful in is to have a full-system test on a testnet where all nodes set very low blocksize and mempool sizes.

My experience with this project is that most of the bugs have been from integrating the fee estimator with the code that calls it - it is much harder to test multiple connected full nodes, or even a single full node (although that is becoming easier with `FullNodeSimulator`).

chia/full_node/mempool.py
arvidn: presumably the majority of blocks hit the "optimized" path though, where we just incrementally remove mempool items that were confirmed in the most recent block.

My understanding is that we only recreate the mempool when we experience a reorg

chia/full_node/bitcoin_fee_estimator.py
```diff
@@ -71,6 +62,9 @@ def get_tracker(self) -> FeeTracker:
         """
         return self.tracker
 
+    def get_mempool_info(self) -> FeeMempoolInfo:
+        return self.last_mempool_info if self.last_mempool_info else EmptyFeeMempoolInfo
```
arvidn, line 66: I don't think `self.last_mempool_info` is ever "falsely", is it?
It would be good to have mypy type annotations for the members, but as far as I can tell it's always initialized.

(I would expect mypy to alert on this, if it had a non-`Optional` type)

chia/full_node/mempool.py
```diff
@@ -6,27 +6,24 @@
...
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.clvm_cost import CLVMCost
-from chia.types.fee_rate import FeeRate
 from chia.types.mempool_item import MempoolItem
 from chia.util.ints import uint64
 
 
 class Mempool:
-    def __init__(self, max_size_in_cost: int, minimum_fee_per_cost_to_replace: uint64, max_block_cost_clvm: uint64):
+    def __init__(self, mempool_info: MempoolInfo, fee_estimator: FeeEstimatorInterface):
```
arvidn, line 19: I like using a `FeeEstimatorInterface` instead of the mempool creating the fee estimator

chia/full_node/mempool_manager.py
```diff
@@ -184,13 +189,13 @@ def always(mm: MempoolManager, mi: MempoolItem) -> bool:
...
 
-        log.info(f"Starting to make block, max cost: {self.constants.MAX_BLOCK_COST_CLVM}")
+        log.info(f"Starting to make block, max cost: {self.max_block_clvm_cost}")
         spend_bundles, cost_sum, additions, removals = self.process_mempool_items(item_inclusion_filter)
         if len(spend_bundles) == 0:
             return None
         log.info(
             f"Cumulative cost of block (real cost should be less) {cost_sum}. Proportion "
-            f"full: {cost_sum / self.constants.MAX_BLOCK_COST_CLVM}"
+            f"full: {cost_sum / self.max_block_clvm_cost}"
```
arvidn, line 198: I think it's important to restore this. People reading the log will assume that this metric is the same as it has always been

arvidn (changes requested):
The check if `self.last_mempool_info` is `None` seems like a mistake. Also, I really think the logging of %-full needs to be restored. I think that's something a lot of people are looking at.

I also think it would be really helpful to add type annotations to the members of both `BitcoinFeeEstimator` as well as `Mempool`, but maybe that's too much to be folded into this change.

However, now that you have an interface for the fee estimator and since the mempool can be created without a full chain, it should be straight-forward to add actual (small) unit tests to ensure the interaction with the fee estimator are correct

chia/full_node/mempool.py
aqk: Yes, that's right - we only run the other path during reorgs or at startup.

chia/full_node/mempool_manager.py
aqk: Done.

chia/full_node/mempool.py
aqk: Thanks!

chia/full_node/bitcoin_fee_estimator.py
aqk: You know, it used to be, when we relied on it being initialized with the first real mempoolInfo being passed in.

But now, since we init it in the constructor, you are correct. Done.

aqk:
> The check if `self.last_mempool_info` is `None` seems like a mistake. Also, I really think the logging of %-full needs to be restored. I think that's something a lot of people are looking at.

Done.

> I also think it would be really helpful to add type annotations to the members of both `BitcoinFeeEstimator` as well as `Mempool`, but maybe that's too much to be folded into this change.

No problem at all - see `__init__` for `BitcoinFeeEstimator` and `Mempool`

arvidn:
> > I also think it would be really helpful to add type annotations to the members of both `BitcoinFeeEstimator` as well as `Mempool`, but maybe that's too much to be folded into this change.
> 
> No problem at all - see `__init__` for `BitcoinFeeEstimator` and `Mempool`

I meant declaring the members with types directly under `class`. I might be mistaken, but I believe that increases mypy's ability to enforce types of members, since `__init__()` is dynamic and may assign to any members under any conditions.

But again, not necessarily for this PR. I point it out because one of the mistakes (likely) would have been caught by mypy.

arvidn (dismissed):
I think it would be good to have a few unit tests ensuring that the mempool notifies the fee estimator correctly though

aqk:
> I think it would be good to have a few unit tests ensuring that the mempool notifies the fee estimator correctly though

Yeah, no problem. I can only do the checks for `add` without pulling in fixes from later PRs. `remove` requires the change that adds `MempoolRemoveReason`, and `new_block` requires a full MempoolManager, not just the Mempool.

Let me know your preference.

aqk:
> I meant declaring the members with types directly under `class`. I might be mistaken, but I believe that increases mypy's ability to enforce types of members, since `__init__()` is dynamic and may assign to any members under any conditions.

Done,  thanks.

chia/full_node/mempool_manager.py
```diff
@@ -131,11 +133,13 @@ def __init__(
...
-            self.mempool_max_total_cost,
-            uint64(self.nonzero_fee_minimum_fpc),
-            uint64(self.max_block_clvm_cost),
+        self.fee_estimator: FeeEstimatorInterface = create_bitcoin_fee_estimator(self.max_block_clvm_cost)
+        mempool_info = MempoolInfo(
+            CLVMCost(uint64(self.mempool_max_total_cost)),
+            FeeRate(uint64(self.nonzero_fee_minimum_fpc)),
+            CLVMCost(uint64(self.max_block_clvm_cost)),
         )
+        self.mempool: Mempool = Mempool(mempool_info, self.fee_estimator)
```
aqk, line 142: `self.mempool: Mempool = Mempool(mempool_info, self.fee_estimator)`

This line is the fix.

arvidn (approved):

PR #14177, opened 2022-12-17
add cache to speed up manage_clvm/clvm_hex pre-commit check

altendky:
Runs were taking +/- 10 seconds.  When the cache is fully hit runs are less than one second.

Draft for:
- [x] cache version

### DIFF
--- a/tools/manage_clvm.py
+++ b/tools/manage_clvm.py
@@ -185,24 +185,28 @@ def main() -> None:
 
 
 @main.command()
-def check() -> int:
+@click.option("--use-cache/--no-cache", default=True, show_default=True, envvar="USE_CACHE")
+def check(use_cache: bool) -> int:
     used_excludes = set()
     overall_fail = False
 
     cache: Cache
-    try:
-        print(f"Attempting to load cache from: {cache_path}")
-        with cache_path.open(mode="r") as file:
-            cache = load_cache(file=file)
-    except FileNotFoundError:
-        print("Cache not found, starting fresh")
+    if not use_cache:
         cache = create_empty_cache()
-    except NoCacheVersionError:
-        print("Ignoring cache due to lack of version")
-        cache = create_empty_cache()
-    except WrongCacheVersionError as e:
-        print(f"Ignoring cache due to incorrect version, expected {e.expected_version!r} got: {e.found_version!r}")
-        cache = create_empty_cache()
+    else:
+        try:
+            print(f"Attempting to load cache from: {cache_path}")
+            with cache_path.open(mode="r") as file:
+                cache = load_cache(file=file)
+        except FileNotFoundError:
+            print("Cache not found, starting fresh")
+            cache = create_empty_cache()
+        except NoCacheVersionError:
+            print("Ignoring cache due to lack of version")
+            cache = create_empty_cache()
+        except WrongCacheVersionError as e:
+            print(f"Ignoring cache due to incorrect version, expected {e.expected_version!r} got: {e.found_version!r}")
+            cache = create_empty_cache()
 
     cache_entries = cache["entries"]
     cache_modified = False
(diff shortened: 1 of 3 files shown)


tools/manage_clvm.py
```diff
@@ -34,6 +37,29 @@
...
 
 
+class CacheEntry(typing.TypedDict):
+    clvm: str
+    hex: str
+    hash: str
+
+
+# PathString = typing.NewType(str)
+# HashString = typing.NewType(str)
```
altendky, line 47:
```suggestion
```

tools/manage_clvm.py
```diff
@@ -37,21 +37,60 @@
 top_levels = {"chia"}
 
 
+class ManageClvmError(Exception):
```
xdustinface, line 40: I wonder why you didn't move this down to the others or the others up to here :)

tools/manage_clvm.py
```diff
@@ -37,21 +37,60 @@
...
 CacheEntries = typing.Dict[str, CacheEntry]
+CacheVersion = typing.List[int]
+current_cache_version: CacheVersion = [1]
+
+
+class CacheVersionError(ManageClvmError):
+    pass
+
+
+class NoCacheVersionError(CacheVersionError):
```
xdustinface, line 59: We could get rid of this and it's usage by using `WrongCacheVersionError(None, current_cache_version)` if there is no version since `found_version` is an `object`, maybe.

xdustinface (dismissed):
Thanks! Nice to see this step finally just pass by 🎉

arvidn:
I don't see where we invoke the compiler. I'm under the impression that this test is primarily here to ensure the compiler keeps generating the same output. Is the version (or commit hash) of the compiler used in the key in the cache?

There is no way to change which version of the compiler is used and still get a cache hit, right?

prozacchiwawa:
The code is compiled at line 252 in the no cache hit case.

I don't think this is for tests specifically, as you'd want the cache disabled to do the recompile check (i do it in clvm_tools_rs' tests, but adam added a similar check in chia-blockchain's tests.  I'll have to see how that particular test works with this.  It removes (iirc) the hex files and causes them to be regenerated by using load_clvm on the clvm files.

You should always get a cache hit with a new version of the compiler (as compile output is meant not to change).  Not to do so means that it's not generating the same output.

That having been said, I wonder if clvm_tools_rs.get_version() could be used as part of the cache version (perhaps only in the test scenario) so that the cache is invalidated when it changes as a hedge against caching compiled clvm when it'd be better to expose an error early.

altendky:
The purpose of this check is to make sure that the files in the repo are clean.  We had lots of cases where people were committing with trailing newlines where they shouldn't be there or missing trailing newlines where there should or whatever else and this resulted in dirty working copies when other people not working on clvm files were doing stuff.

But...  I feel like someone should have requested changes here until we get this discussed?

altendky:
Or, well, that is a key requirement from me.  But yes, there is more going on here as you say.  Basically, instead of having a repository managing these critical things and doing releases of them as standard artifacts and then future compiler checks compare to those well controlled artifacts...  we instead have pushed artifact management on the blockchain git repo and compiler testing on the blockchain repo ci.

CI already doesn't use a cache, so long as we don't commit one.  Is adding a `--no-cache` option that we use in CI sufficient?  Or do you want local pre-commit also compiling?

altendky:
I went ahead and added the cache skip.  I didn't find quite yet a way to pass options through pre-commit but only in CI so I went with an environment variable via the Click mechanism.  I don't like it since any mismatch would result in silent lack of disabling the cache whereas a passed command line option would result in a failure if it were wrong.  But...  it's something and I think it's good to have the control even if we decide something else is also needed to address the concerns.

prozacchiwawa (approved):

arvidn (approved):

PR #14184, opened 2022-12-18
add rate limits numbers to none response type

### LINKED ISSUE #14162: [Bug] Message type ProtocolMessageTypes.none_response not found in rate limits
### What happened?

There are a lot of similar entries in the log.

### Version

1.7.0rc3-dev63

Linux

CLI

### Relevant log output

```shell
2022-12-15T20:19:39.297 full_node chia.server.rate_limits : WARNING  Message type ProtocolMessageTypes.none_response not found in rate limits
2022-12-15T20:19:40.793 full_node chia.server.rate_limits : WARNING  Message type ProtocolMessageTypes.none_response not found in rate limits
2022-12-15T20:19:41.373 full_node chia.server.rate_limits : WARNING  Message type ProtocolMessageTypes.none_response not found in rate limits
```

almogdepaz:
added rate limit numbers to the new none_response message type, doesn't change behavior 
https://github.com/Chia-Network/chia-blockchain/issues/14162

### DIFF
--- a/chia/server/rate_limit_numbers.py
+++ b/chia/server/rate_limit_numbers.py
@@ -178,6 +178,7 @@ rate_limits = {
             ProtocolMessageTypes.request_puzzle_solution: RLSettings(5000, 100),
             ProtocolMessageTypes.respond_puzzle_solution: RLSettings(5000, 1024 * 1024),
             ProtocolMessageTypes.reject_puzzle_solution: RLSettings(5000, 100),
+            ProtocolMessageTypes.none_response: RLSettings(500, 100),
         },
         "rate_limits_other": {  # These will have a lower cap since they don't scale with high TPS (NON_TX_FREQ)
             ProtocolMessageTypes.request_header_blocks: RLSettings(5000, 100),


arvidn:
we have a fair amount of test infrastructure around protocol messages to ensure we don't miss things like this.
Could you add a test in CI to make sure all message types have rate limits? (possibly with a hard coded whitelist of exceptions).

almogdepaz:
> we have a fair amount of test infrastructure around protocol messages to ensure we don't miss things like this. Could you add a test in CI to make sure all message types have rate limits? (possibly with a hard coded whitelist of exceptions).

not in this pr since we have lots of other messages that are missing rate limits

altendky:
Those tests need a lot of rework that needs to be based on a lot of API tracking rework before they'll be robust.

altendky:
I guess the numbers for this would be dependent on which messages respond with this and how frequent they are?  Or maybe that's too detailed.

almogdepaz:
> I guess the numbers for this would be dependent on which messages respond with this and how frequent they are? Or maybe that's too detailed.

not sure i understand what you are saying here

altendky:
I was guessing at how to pick the proper numbers since that seemed like the most important part of this review.  How did you select the frequency of 500 and max size of 100?

almogdepaz:
> I was guessing at how to pick the proper numbers since that seemed like the most important part of this review. How did you select the frequency of 500 and max size of 100?

i used other message types as reference, this should be a rare occasion so i dont think we will hit

altendky (approved):

emlowe (approved):

PR #14186, opened 2022-12-18
wallet: Avoid duplicates in `get_puzzle_hashes_to_subscribe`

xdustinface:
We currently store the derived puzzle hashes in both, puzzle store and  interested store and here we combine them as list. There is some code to  mitigate duplicate subscriptions but its based on chunks here  https://github.com/Chia-Network/chia-blockchain/blob/39d733599d170b340b0ae4c13ada8896a78a0d29/chia/wallet/wallet_node.py#L662  so if there are less than 1000 puzzle hashes it obviously still can lead to duplicated entries in the subscription calls?

Based on:
- https://github.com/Chia-Network/chia-blockchain/pull/14187

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -1261,13 +1261,13 @@ class WalletNode:
         return weight_proof, summaries, block_records
 
     async def get_puzzle_hashes_to_subscribe(self) -> List[bytes32]:
-        all_puzzle_hashes = list(await self.wallet_state_manager.puzzle_store.get_all_puzzle_hashes())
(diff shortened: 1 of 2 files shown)


xdustinface:
@arvidn I made this PR based on #14187 now and added https://github.com/Chia-Network/chia-blockchain/pull/14186/commits/da5fe47b5f48ed436a9aa4180892180d8578d5a1 instead of modifying an existing test. How do you feel about just leaving the length asserts in #14187 in combination with the test in here?

arvidn (approved):

PR #14188, opened 2022-12-18
add local max subscription config

### LINKED ISSUE #14142: Should max_subscribe_items check be bypassed if is_in_network?
I spent the last few days troubleshooting a broken wallet. The derivation index is over 20,000 and it has 7 CAT wallets, so we are well over the default `max_subscribe_items` limit according to this issue:

https://github.com/Chia-Network/chia-blockchain/issues/14002#issuecomment-1344654073

I noticed all kinds of weird behavior, especially offer files with multiple CATs. Some times some of the offer transactions would show up, sometimes not. Sometimes they would confirm, sometimes they'd be "stuck" forever. Sometimes restarting the wallet would work, sometimes I'd have to delete transactions and try again. There was nothing in the logs about going past this limit, but I methodically tried different settings in the config.yaml file (I thought it might be timeouts at first) until I figured out that this is the one that did it. By changing it from the default of `200000` to `2000000`, my wallet can sync from empty and new offers/transactions all show up correctly as expected.

I wonder if these issues are related:

#13963
#14002 

After digging into the code a bit more, I have some suggestions. First, I think this error should be logged. It doesn't seem to get logged or throw an exception, which means the client has no idea what is happening and we get the weird behavior as described above.

https://github.com/Chia-Network/chia-blockchain/blob/dfd888dc58e86f2a9b482912b2795ba813235e0f/chia/full_node/full_node_api.py#L1501

Secondly: I don't necessarily think that the default should be increased, as we don't want our poor full-nodes getting DOS'd with external wallet subscriptions BUT I think it is fair to assume that a local wallet running against the local trusted full node should not be subject to this max, similar to the `is_in_network` & `is_localhost` check done for max connections here:

https://github.com/Chia-Network/chia-blockchain/blob/dfd888dc58e86f2a9b482912b2795ba813235e0f/chia/server/server.py#L331

This would mean that the rare large wallets wouldn't be limited here by default when talking to their own full node, but untrusted wallets that connect from outside would still be unable to DOS us.

I am not confident enough yet in my python fu to know if this is the correct approach, or if I'm missing something else entirely. Thanks!

almogdepaz:
add a different config for local max subscriptions  so we have a higher limit for local wallets, the default is 2000000 instead of 200000 we use for remote wallets

add trusted wallet config so we can use the new max subscription config with wallets defined as trusted

in response to https://github.com/Chia-Network/chia-blockchain/issues/14142

### DIFF
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -1466,13 +1466,12 @@ class FullNodeAPI:
             max_items = self.full_node.config.get("max_subscribe_items", 200000)
         for puzzle_hash in request.puzzle_hashes:
 
+            hint_coin_ids.extend(await self.full_node.hint_store.get_coin_ids(puzzle_hash))
+
             # check subscription limit
             if self.full_node.peer_sub_counter[peer.peer_node_id] >= max_items:
                 self.log.warning(f"wallet {peer.peer_node_id} exceeded max puzzle_hash subscription limit")
-                break
-
-            ph_hint_coins = await self.full_node.hint_store.get_coin_ids(puzzle_hash)
-            hint_coin_ids.extend(ph_hint_coins)
+                continue
 
             if puzzle_hash not in self.full_node.ph_subscriptions:
                 self.full_node.ph_subscriptions[puzzle_hash] = set()
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -1461,8 +1461,8 @@ class FullNodeAPI:
 
         hint_coin_ids = []
         # Add peer to the "Subscribed" dictionary
-        if is_in_network(peer_host=peer.peer_host, networks=self.server.exempt_peer_networks):
-            max_items = self.full_node.config.get("local_max_subscribe_items", 2000000)
+        if self.is_trusted(peer):
+            max_items = self.full_node.config.get("trusted_max_subscribe_items", 2000000)
         else:
             max_items = self.full_node.config.get("max_subscribe_items", 200000)
         for puzzle_hash in request.puzzle_hashes:
@@ -1470,16 +1470,13 @@ class FullNodeAPI:
             hint_coin_ids.extend(ph_hint_coins)
             if puzzle_hash not in self.full_node.ph_subscriptions:
                 self.full_node.ph_subscriptions[puzzle_hash] = set()
-            if (
-                peer.peer_node_id not in self.full_node.ph_subscriptions[puzzle_hash]
-                and self.full_node.peer_sub_counter[peer.peer_node_id] >= max_items
-            ):
+            if self.full_node.peer_sub_counter[peer.peer_node_id] >= max_items:
                 self.log.warning(f"wallet {peer.peer_node_id} reached max puzzle_hash subscription limit")
                 break
-
-            self.full_node.ph_subscriptions[puzzle_hash].add(peer.peer_node_id)
-            self.full_node.peer_puzzle_hash[peer.peer_node_id].add(puzzle_hash)
-            self.full_node.peer_sub_counter[peer.peer_node_id] += 1
+            if peer.peer_node_id not in self.full_node.ph_subscriptions[puzzle_hash]:
+                self.full_node.ph_subscriptions[puzzle_hash].add(peer.peer_node_id)
+                self.full_node.peer_puzzle_hash[peer.peer_node_id].add(puzzle_hash)
+                self.full_node.peer_sub_counter[peer.peer_node_id] += 1
 
         # Send all coins with requested puzzle hash that have been created after the specified height
         states: List[CoinState] = await self.full_node.coin_store.get_coin_states_by_puzzle_hashes(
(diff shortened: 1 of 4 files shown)


chia/full_node/full_node_api.py
```diff
@@ -1498,14 +1504,20 @@ async def register_interest_in_coin(
...
-        max_items = self.full_node.config.get("max_subscribe_items", 200000)
+        if self.is_trusted(peer):
+            max_items = self.full_node.config.get("trusted_max_subscribe_items", 2000000)
+        else:
+            max_items = self.full_node.config.get("max_subscribe_items", 200000)
         for coin_id in request.coin_ids:
+
+            # check subscription limit
+            if self.full_node.peer_sub_counter[peer.peer_node_id] >= max_items:
+                self.log.warning(f"wallet {peer.peer_node_id} exceeded max coin subscription limit")
```
arvidn, line 1515: I think this should be an `INFO`. The person running the full node probably doesn't care *that* much about one of its wallet peers

chia/full_node/full_node_api.py
```diff
@@ -1460,16 +1460,22 @@ async def register_interest_in_puzzle_hash(
...
+            max_items = self.full_node.config.get("max_subscribe_items", 200000)
         for puzzle_hash in request.puzzle_hashes:
-            ph_hint_coins = await self.full_node.hint_store.get_coin_ids(puzzle_hash)
-            hint_coin_ids.extend(ph_hint_coins)
+
+            hint_coin_ids.extend(await self.full_node.hint_store.get_coin_ids(puzzle_hash))
+
+            # check subscription limit
+            if self.full_node.peer_sub_counter[peer.peer_node_id] >= max_items:
+                self.log.warning(f"wallet {peer.peer_node_id} exceeded max puzzle_hash subscription limit")
```
arvidn, line 1473: I think this should be `INFO`. It's not a problem for the full node

chia/simulator/setup_services.py
```diff
@@ -105,6 +105,7 @@ async def setup_full_node(
     config = local_bt.config
     service_config = config["full_node"]
     service_config["database_path"] = db_name
+    service_config["testing"] = True
```
arvidn, line 108: these changes look unrelated. Are they deliberate? if so, can you explain why you need to make this change?

chia/util/initial-config.yaml
```diff
@@ -423,6 +426,9 @@ full_node:
   network_overrides: *network_overrides
   selected_network: *selected_network
 
+  trusted_peers:
+    0ThisisanexampleNodeID7ff9d60f1c3fa270c213c0ad0cb89c01274634a7c3cb7: Does_not_matter
```
arvidn, line 430: I think this could be explained in a more user-friendly way. Instead of having a dummy entry, maybe you could add a comment describing what one can put in this list.

I take it we're not adding our own local wallet's peer ID in here automatically then

chia/simulator/setup_services.py
almogdepaz: deliberate otherwise we throw an error here https://github.com/Chia-Network/chia-blockchain/blob/a9b3e25d2ae037e29a0acd7b0fae7450e139e087/chia/server/server.py#L688

this is whole thing is done in a bad way but im not going to fix that 
here, i can set the config inside the test itself if you think its better but i figured id save other people the trouble.

chia/full_node/full_node_api.py
almogdepaz: i dont think so, this could be your own wallet, and it really shouldn't happen this means the receiving wallet will miss coins

chia/full_node/full_node_api.py
almogdepaz: same as above

chia/full_node/full_node_api.py
joshpainter: Agree - a user with a large wallet/lots of CATs will most likely be running their own full node. When they hit the limit and things start acting weird, they'll check their logs or support will ask them for the logs.

joshpainter:
This is great, thank you! It still feels a bit like "kicking the can down the road" though by just raising the max 10x. See here for a possible optimization: #14192

emlowe:
I also have a small issue with the logging. A remote wallet might cause this to happen to some random node - the person running that node will get a warning in the logs they can't do anything about (well they are unlikely to increase their limit anyway) - it's not like that person can somehow contact the wallet owner to tell them they might be missing coins. And the wallet owner will never see the log entry. So in this case, this log entry doesn't seem very helpful - I guess it might be useful from a Chia network diagnostic perspective to understand if this happens but that also would require people to report it.

Perhaps this is rare enough that it doesn't matter either way, but I think the logging should likely only be done for trusted peers.

altendky:
What about the logging just being info?  Or info for non-trusted peers?  Though it does seem like the important thing is for the wallet to be aware and to notify.  I don't know if this is already the case or not (and probably wouldn't belong in this PR).

almogdepaz:
> nd the wallet owner will never see the log entry. So in this case, this log entry doesn't seem very helpful - I guess it might be useful from a Chia network diagnostic perspective to understand if this happens but that also would require people to report it.
> 
> Perhaps this is rare enough that it doesn't matter either way, but I think the logging should likely only be done for tr

@emlowe i can change to info if you think thats better but wouldn't you want to know that some node subscribed to the max amount ? 
that feels like and edge case to me and can have effects on your node performance especially if there are more than one

nqzhang:
mark

emlowe:
I guess if your node was having performance issues, such a log entry might be useful, although still not sure what the node operator would do about it. I'm not really a fan of warnings that are not actionable - perhaps it's a pet peeve.

I'll approve as I don't want to hold this up over a quibble - when this change could materially help those with large wallets while running local noded

emlowe (approved):

PR #14193, opened 2022-12-19
Fixes the case when a fee estimation request is made before the FullNode has initialized a peak

aqk:
Fixes the case when a fee estimation request is made before the FullNode has initialized a peak

### DIFF
--- a/chia/rpc/full_node_rpc_api.py
+++ b/chia/rpc/full_node_rpc_api.py
@@ -789,13 +789,15 @@ class FullNodeRpcApi:
         mempool_max_size = estimator.mempool_max_size()
         blockchain_state = await self.get_blockchain_state({})
         synced = blockchain_state["blockchain_state"]["sync"]["synced"]
-        peak_height = blockchain_state["blockchain_state"]["peak"].height
-        last_peak_timestamp = blockchain_state["blockchain_state"]["peak"].timestamp
+        peak = blockchain_state["blockchain_state"]["peak"]
+        peak_height = peak.height if peak else uint32(0)
+        last_peak_timestamp = peak.timestamp if peak else uint64(0)
         peak_with_timestamp = peak_height
-        while last_peak_timestamp is None:
-            peak_with_timestamp -= 1
-            block_record = self.service.blockchain.height_to_block_record(peak_with_timestamp)
-            last_peak_timestamp = block_record.timestamp
+        if peak:
+            while last_peak_timestamp is None:
+                peak_with_timestamp -= 1
+                block_record = self.service.blockchain.height_to_block_record(peak_with_timestamp)
+                last_peak_timestamp = block_record.timestamp
 
         dt = datetime.now(timezone.utc)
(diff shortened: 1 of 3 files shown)


tests/fee_estimation/test_fee_estimation_rpc.py
```diff
@@ -44,6 +44,26 @@ async def setup_node_and_rpc(
     return client, full_node_rpc_api
 
 
+@pytest_asyncio.fixture(scope="function")
+async def setup_node_and_rpc_no_blocks_farmed(
+    two_wallet_nodes_services: Tuple[List[Service[FullNode]], List[Service[WalletNode]], BlockTools],
```
altendky, line 49: We could avoid setting up two wallets here I think?

chia/rpc/full_node_rpc_api.py
```diff
@@ -789,13 +789,15 @@ async def get_fee_estimate(self, request: Dict) -> Dict[str, Any]:
...
         peak_with_timestamp = peak_height
-        while last_peak_timestamp is None:
-            peak_with_timestamp -= 1
-            block_record = self.service.blockchain.height_to_block_record(peak_with_timestamp)
-            last_peak_timestamp = block_record.timestamp
+        if peak:
+            while last_peak_timestamp is None:
+                peak_with_timestamp -= 1
+                block_record = self.service.blockchain.height_to_block_record(peak_with_timestamp)
+                last_peak_timestamp = block_record.timestamp
```
altendky, line 800: How would you feel about this?  Seems a bit more clear as to the handling of the two cases.

```suggestion
        if peak is None:
            peak_height = uint32(0)
            last_peak_timestamp = uint64(0)
        else:
            peak_height = peak.height
            last_peak_timestamp = peak.timestamp
            peak_with_timestamp = peak_height
            while last_peak_timestamp is None:
                peak_with_timestamp -= 1
                block_record = self.service.blockchain.height_to_block_record(peak_with_timestamp)
                last_peak_timestamp = block_record.timestamp
```

aqk: Yes, looks clearer to me.

tests/fee_estimation/test_fee_estimation_rpc.py
aqk: Ya, agreed

chia/rpc/full_node_rpc_api.py
aqk: @altendky Something is preventing me from using the github `Commit suggestion` button - I just pushed these changes up from local.

arvidn (approved):
I take it those tests failed before the fix, right?

chia/rpc/full_node_rpc_api.py
altendky: Yeah, I wasn't sure if it would work, sorry.  I cheated since it wouldn't let me create it over a deleted line.

altendky (approved):

PR #14206, opened 2022-12-20
Handle None from `get_keys()`

emlowe:
Part of the series of PRs to eventually type hint `WalletStateManager`

`get_keys()` returns an Optional, so make sure to handle possible None return

### DIFF
--- a/chia/wallet/did_wallet/did_wallet.py
+++ b/chia/wallet/did_wallet/did_wallet.py
@@ -1179,7 +1179,14 @@ class DIDWallet:
 
             p2_puzzle, _, _, _, _ = puzzle_args
             puzzle_hash = p2_puzzle.get_tree_hash()
-            pubkey, private = await self.wallet_state_manager.get_keys(puzzle_hash)
+            maybe = await self.wallet_state_manager.get_keys(puzzle_hash)
+            if maybe is None:
+                error_msg = f"Wallet couldn't find keys for puzzle_hash {puzzle_hash}"
+                self.log.error(error_msg)
+                raise ValueError(error_msg)
+
+            pubkey, private = maybe
+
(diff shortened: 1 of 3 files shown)


arvidn (commented):
it really looks like this code is asking for being turned into a function :)

arvidn:
I take it the reason mypy isn't complaining about this is because these files are still exempt from mypy checks, is that right?

emlowe:
Yes, mypy doesn't complain.... yet. 

Yes, this is really calling out for some de-duplication - not sure what the best approach is though

altendky (changes requested):
Raise and log inside `.get_keys()` itself?  Just consider the exceptions that were explicitly or implicitly raised before and might be handled up the chain...  or...  is it just me or is the hint on `.get_keys()` wrong and it already won't ever return `None` but instead raises an exception...

https://github.com/Chia-Network/chia-blockchain/blob/50af5dd0ee7e726c3c470d621e9a6058c4f2fd2c/chia/wallet/wallet_state_manager.py#L262-L272

If I'm reading this right then we just delete the `Optional` bit of the hint and then check the call sites to make sure they aren't checking for `None` pointlessly.

altendky:
> I take it the reason mypy isn't complaining about this is because these files are still exempt from mypy checks, is that right?

At least the majority of wallet state managers are hinted as `Any` since there wasn't awareness around how to handle circular dependencies for hinting.  So, basically all uses of wallet state manager or anything anyone gets from it (we often access multiple layers of attributes from the manager) are entirely unchecked even in the otherwise 'more strict' files.  Once we get all files 'more strict', we get to go on an `Any` elimination binge.

emlowe:
You are right, `get_keys` doesn't actually return Optional - so I'm closing this PR and will make a different one

PR #14207, opened 2022-12-21
Add get_new_puzzlehash to WalletProtocol

emlowe:
Part of a series of PRs to add type checking to WalletStateManager

Add `get_new_puzzlehash()` to WalletProtocol - this is used primarily by trade_manager code.

The specific implementation for Pool, DID, and NFT wallets was taken from the existing CAT wallet implementation

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -978,6 +978,9 @@ class PoolWallet:
     def get_name(self) -> str:
         return self.wallet_info.name
 
+    async def get_new_puzzlehash(self) -> bytes32:
+        return await self.standard_wallet.get_new_puzzlehash()
+
 
 if TYPE_CHECKING:
     from chia.wallet.wallet_protocol import WalletProtocol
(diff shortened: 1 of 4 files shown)


arvidn:
it's a bit strange to add this function without adding any users of it. If it's an attempt to make the PR smaller, another approach might be to include more commits in the PR where each commit is bite size, but the reviewer can still see the big picture

arvidn:
I'm not sure it's relevant here; but one source of performance problems in the wallet used to be creating new puzzle hashes. We used to (and still do in some wallet types) build the full puzzle and then run tree-hash on it. The optimized version hashes the puzzle as it's built (and all the parts are discarded immediately, so it isn't actually built. It really only need to hash the curried-in arguments, and include the mod-hash.

I think we should move towards an interface where all wallets do that.

emlowe:
Currently, trade_manager.py uses this, but because it does `wallet_state_manager: Any` - I can add the Types to trade_manager and make a bigger PR if that would make more sense

altendky:
With all the other wallets just passing through to the standard wallet, I'm a bit on the fence about this being particular useful.  But, I'm willing to fall on the optimistic side of the fence here.  But, I don't want to accidentally bypass Arvid's discussion (that they started without requesting changes... `:]`) so I'll hold off on actually dropping an approval here.

emlowe:
Closing in preference of alternative implementation

PR #14208, opened 2022-12-21
Add get_asset_id to WalletProtocol

emlowe:
Another PR in the type hint WalletStateManager series.

Add `get_asset_id` to WalletProtocol. The only implementation is the CAT wallet
Other wallets will raise an exception
Used by the TradeManager

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -1105,6 +1105,9 @@ class DataLayerWallet:
     def get_name(self) -> str:
         return self.wallet_info.name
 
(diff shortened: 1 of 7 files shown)


altendky (changes requested):
It seems like the asset ID is actually not a general wallet property but rather a CAT wallet specific property.  Maybe it should be handled accordingly and only accessed when we have confirmed we have a CAT wallet instead of adding it to the general wallet interface?

I don't know if there are other expected cases for asset IDs.

emlowe:
Closing in preference of an alternative implementation

PR #14213, opened 2022-12-21
Add SigningMode and update verify_signature RPC to work with sign_message_by_* RPCs

paninaro:
The `verify_signature` RPC currently doesn't verify signatures made by the `sign_message_by_address`/`sign_message_by_id` RPCs, which follow the CHIP-0002 signing method. The signatures made by those RPCs calculate a tree hash of ("Chia Signed Message", message), and then sign that hash. This change adds support for verifying those signatures by computing the same tree hash and verifying the signature over that hash.

The `verify_signature` RPC was originally written to complement the `chia keys sign` CLI, which directly signs the message content (as a hex string) using the BLS message augmentation mode.

A new RPC request param, `signing_mode`, can be set to `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_AUG:CHIP-0002_` to indicate that the signature should be validated as a CHIP-0002 signature. In this signing mode, the tree hash of ("Chia Signed Message", message) should be computed the signature should be verified against that hash.

The `sign_message_by_*` RPCs now return the `signing_mode` that was used (CHIP-0002 in both cases) to assist with verification.

A new SigningMode enum has been added to `chia/types` to indicate how a particular message was signed, and similarly how it should be verified. Enum values are provided for signing in the CHIP-0002 method, as well as values for direct signing of the content using BLS message augmentation mode, where the message is either a hex string or UTF-8 string.

The`chia keys sign` and `chia keys verify` CLIs has been updated to output and accept JSON data respectively. The signing CLI output now also includes the original message and the signing mode that was used. The verification CLI now accepts the `-b as_bytes` option to match the same option on the signing side.

Testing notes:
`verify_signature` wasn't previously covered by any tests, nor is it exposed to the CLI. This change includes tests to cover both behaviors (`signing_mode = chip_0002|None`) as well as some negative tests.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1238,11 +1238,14 @@ class WalletRpcApi:
         signing_mode_str: Optional[str] = request.get("signing_mode")
         # Default to BLS_MESSAGE_AUGMENTATION_HEX_INPUT as this RPC was originally designed to verify
         # signatures made by `chia keys sign`, which uses BLS_MESSAGE_AUGMENTATION_HEX_INPUT
-        signing_mode = (
-            SigningMode(signing_mode_str)
-            if signing_mode_str is not None
-            else SigningMode.BLS_MESSAGE_AUGMENTATION_HEX_INPUT
-        )
+        if signing_mode_str is None:
+            signing_mode = SigningMode.BLS_MESSAGE_AUGMENTATION_HEX_INPUT
+        else:
+            try:
+                signing_mode = SigningMode(signing_mode_str)
+            except ValueError:
+                raise ValueError(f"Invalid signing mode: {signing_mode_str}")
+
         if signing_mode == SigningMode.CHIP_0002:
             # CHIP-0002 message signatures are made over the tree hash of:
             #   ("Chia Signed Message", message)
@@ -1253,6 +1256,8 @@ class WalletRpcApi:
         elif signing_mode == SigningMode.BLS_MESSAGE_AUGMENTATION_UTF8_INPUT:
             # Message is expected to be a UTF-8 string
             message_to_verify = bytes(input_message, "utf-8")
+        else:
+            raise ValueError(f"Unsupported signing mode: {signing_mode_str}")
 
         # Verify using the BLS message augmentation scheme
         is_valid = AugSchemeMPL.verify(
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1233,9 +1233,19 @@ class WalletRpcApi:
         :param request:
         :return:
         """
+        input_message: str = request["message"]
+        if request.get("is_chia_signed_message", False):
+            # When signing a message with the sign_message_by_address/sign_message_by_id RPCs,
+            # the actual content being signed is the tree hash of:
+            #   ("Chia Signed Message", message)
+            message_to_verify: bytes32 = Program.to(("Chia Signed Message", input_message)).get_tree_hash()
+        else:
+            # Signed message is expected to be a hex string
+            message_to_verify = bytes.fromhex(input_message)
+
         is_valid = AugSchemeMPL.verify(
             G1Element.from_bytes(bytes.fromhex(request["pubkey"])),
-            bytes.fromhex(request["message"]),
+            message_to_verify,
             G2Element.from_bytes(bytes.fromhex(request["signature"])),
         )
         if "address" in request:
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1257,7 +1257,7 @@ class WalletRpcApi:
             # Message is expected to be a UTF-8 string
             message_to_verify = bytes(input_message, "utf-8")
         else:
-            raise ValueError(f"Unsupported signing mode: {signing_mode_str}")
+            raise ValueError(f"Unsupported signing mode: {signing_mode_str!r}")
 
         # Verify using the BLS message augmentation scheme
         is_valid = AugSchemeMPL.verify(
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1244,7 +1244,7 @@ class WalletRpcApi:
             try:
                 signing_mode = SigningMode(signing_mode_str)
             except ValueError:
-                raise ValueError(f"Invalid signing mode: {signing_mode_str}")
+                raise ValueError(f"Invalid signing mode: {signing_mode_str!r}")
 
         if signing_mode == SigningMode.CHIP_0002:
             # CHIP-0002 message signatures are made over the tree hash of:
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -25,6 +25,7 @@ from chia.types.blockchain_format.program import Program
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.coin_record import CoinRecord
 from chia.types.coin_spend import CoinSpend
+from chia.types.signing_mode import SigningMode
 from chia.types.spend_bundle import SpendBundle
 from chia.util.bech32m import decode_puzzle_hash, encode_puzzle_hash
 from chia.util.byte_types import hexstr_to_bytes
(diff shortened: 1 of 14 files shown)


tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -1294,3 +1305,113 @@ async def test_notification_rpcs(wallet_rpc_environment: WalletRpcTestEnvironmen
     notification = (await client_2.get_notifications())[0]
     assert await client_2.delete_notifications([notification.coin_id])
     assert [] == (await client_2.get_notifications([notification.coin_id]))
+
+
+@pytest.mark.parametrize(
+    "rpc_request, rpc_response",
```
altendky, line 1311: I would prefer the form that leverages Python syntax more.

```suggestion
    ["rpc_request", "rpc_response"],
```

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -1294,3 +1305,113 @@ async def test_notification_rpcs(wallet_rpc_environment: WalletRpcTestEnvironmen
...
+    [
+        # Valid signatures
```
altendky, line 1313: Recreating these would require access to a particular set of mnemonics I would think?  I'm not sure if it is better to have a public set for stuff like this or if we just accept that any modification will often be a complete change including the address and mnemonics used.

altendky (changes requested):

chia/rpc/wallet_rpc_api.py
```diff
@@ -1228,15 +1228,28 @@ async def verify_signature(self, request) -> EndpointResult:
...
         """
+        input_message: str = request["message"]
+        signing_mode: Optional[str] = request.get("signing_mode", None)
+        if signing_mode == "chip_0002":
+            # CHIP-0002 message signatures are made over the tree hash of:
+            #   ("Chia Signed Message", message)
+            message_to_verify: bytes = Program.to((CHIP_0002_SIGN_MESSAGE_PREFIX, input_message)).get_tree_hash()
+        else:
+            # Signed message is expected to be a hex string
+            message_to_verify = hexstr_to_bytes(input_message)
```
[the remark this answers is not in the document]
paninaro: Added a SigningMode enum to handle the CHIP-0002 case as well as the default case (now denoted as `SigningMode.BLS_MESSAGE_AUGMENTATION_HEX_INPUT`). A third case is also handled when dealing with a UTF-8 message `SigningMode.BLS_MESSAGE_AUGMENTATION_UTF8_INPUT`.

I've left out the `else:` branch because an instance of SigningMode is constructed from a string, raising an exception if the string is not a valid signing mode value. Including the `else:` branch would lead to unreachable code in this case.

chia/wallet/wallet.py
```diff
@@ -445,10 +449,12 @@ async def sign_transaction(self, coin_spends: List[CoinSpend]) -> SpendBundle:
         )
 
     async def sign_message(self, message: str, puzzle_hash: bytes32) -> Tuple[G1Element, G2Element]:
+        # CHIP-0002 message signing as documented at:
+        # https://github.com/Chia-Network/chips/blob/main/CHIPs/chip-0002.md#signmessage
```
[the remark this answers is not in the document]
paninaro: Updated

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -1294,3 +1305,113 @@ async def test_notification_rpcs(wallet_rpc_environment: WalletRpcTestEnvironmen
...
+                    "91246eb1c7f1b66a6a"
+                ),
+                "signing_mode": "chip_0002",
+                "address": "xch1d0rekc2javy5gpruzmcnk4e4qq834jzlvxt5tcgl2ylt49t26gdsjen7t0",
+            },
+            {"isValid": False, "error": "Public key doesn't match the address"},
+        ),
+    ],
+)
+@pytest.mark.parametrize("prefix_hex_strings", [True, False])
```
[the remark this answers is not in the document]
paninaro: Updated

tests/wallet/rpc/test_wallet_rpc.py
paninaro: Updated

tests/wallet/rpc/test_wallet_rpc.py
paninaro: I'm mostly in the camp of tests should run against static data, so that was my reasoning for baking in pre-computed signature data here.

As you mention, recreating this exact set of test data would require using the same mnemonic, but my expectation is any updates to this would simply create new test data using a new throw-away mnemonic. I could add a comment explaining that in more detail if you think that would be helpful. I included the CLI for generating the test data, but maybe that's not clear on its own.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1233,15 +1234,39 @@ async def verify_signature(self, request) -> EndpointResult:
...
+        if signing_mode == SigningMode.CHIP_0002:
+            # CHIP-0002 message signatures are made over the tree hash of:
+            #   ("Chia Signed Message", message)
+            message_to_verify: bytes = Program.to((CHIP_0002_SIGN_MESSAGE_PREFIX, input_message)).get_tree_hash()
+        elif signing_mode == SigningMode.BLS_MESSAGE_AUGMENTATION_HEX_INPUT:
+            # Message is expected to be a hex string
+            message_to_verify = hexstr_to_bytes(input_message)
+        elif signing_mode == SigningMode.BLS_MESSAGE_AUGMENTATION_UTF8_INPUT:
+            # Message is expected to be a UTF-8 string
+            message_to_verify = bytes(input_message, "utf-8")
```
altendky, line 1255: I would have expected something to complain about the maybe not defined `message_to_verify` variable.  I think an else here with an exception would be good to indicate that we are not handling a signing mode that we have defined.

I played around a bit and while I was able to trigger both mypy and flake8 to complain when I removed all references to creating the variable, they seem to be doing a really bad job of detecting this...  `:[`

tests/wallet/rpc/test_wallet_rpc.py
altendky: Not a huge deal, but a comment might save someone some time looking around to see if there's an existing key to use.

altendky (dismissed):

chia/rpc/wallet_rpc_api.py
```diff
@@ -1233,15 +1234,39 @@ async def verify_signature(self, request) -> EndpointResult:
         :param request:
         :return:
         """
+        input_message: str = request["message"]
+        signing_mode_str: Optional[str] = request.get("signing_mode")
+        # Default to BLS_MESSAGE_AUGMENTATION_HEX_INPUT as this RPC was originally designed to verify
+        # signatures made by `chia keys sign`, which uses BLS_MESSAGE_AUGMENTATION_HEX_INPUT
+        signing_mode = (
+            SigningMode(signing_mode_str)
```
emlowe, line 1242: I believe this will throw/raise some enum exception here if `signing_mode_str` is not one of the enums values - is that what you want?

emlowe (changes requested):

chia/rpc/wallet_rpc_api.py
altendky: It does provide at least a legible error message.

```python-console
Python 3.10.8 (main, Oct 15 2022, 17:48:51) [GCC 9.4.0] on linux
>>> import enum
>>> class E(enum.Enum):
...     ab = "ab"
...     cd = "cd"
... 
>>> E("ab")
<E.ab: 'ab'>
>>> E("cd")
<E.cd: 'cd'>
>>> E("ef")
  File "/home/altendky/.pyenv/versions/3.10.8/lib/python3.10/enum.py", line 385, in __call__
    return cls.__new__(cls, value)
  File "/home/altendky/.pyenv/versions/3.10.8/lib/python3.10/enum.py", line 710, in __new__
    raise ve_exc
ValueError: 'ef' is not a valid E
```

paninaro: Yes, this was deliberate. I tried the same experiment as @altendky and felt it was a reasonable error message. But perhaps it's asking a bit much to rely on this behavior. I can wrap the call in a try/except and provide our own error message.

chia/rpc/wallet_rpc_api.py
paninaro: Added an else case

tests/wallet/rpc/test_wallet_rpc.py
paninaro: Added

chia/rpc/wallet_rpc_api.py
```diff
@@ -1233,15 +1234,44 @@ async def verify_signature(self, request) -> EndpointResult:
...
+        signing_mode_str: Optional[str] = request.get("signing_mode")
+        # Default to BLS_MESSAGE_AUGMENTATION_HEX_INPUT as this RPC was originally designed to verify
+        # signatures made by `chia keys sign`, which uses BLS_MESSAGE_AUGMENTATION_HEX_INPUT
+        if signing_mode_str is None:
+            signing_mode = SigningMode.BLS_MESSAGE_AUGMENTATION_HEX_INPUT
+        else:
+            try:
+                signing_mode = SigningMode(signing_mode_str)
+            except ValueError:
+                raise ValueError(f"Invalid signing mode: {signing_mode_str}")
```
altendky, line 1247: This modification will print the `repr()` of the string making it easier to see special cases like trailing spaces or newlines etc.  I guess it could be argued that converting it to a JSON encoded string might be more language agnostic.

```python-console
>>> s = "abc "
>>> print(f"{s}")
abc 
>>> print(f"{s!r}")
'abc '
```

```suggestion
                raise ValueError(f"Invalid signing mode: {signing_mode_str!r}")
```

emlowe (approved):

PR #11407, opened 2022-05-03
Add support for Python 3.11

altendky:
Draft for:
- [x] Update other packages
  - https://github.com/Chia-Network/chiavdf/pull/119
  - https://github.com/Chia-Network/chiabip158/pull/16
  - https://github.com/Chia-Network/chiapos/pull/330
  - https://github.com/Chia-Network/chia_rs/pull/16
  - https://github.com/Chia-Network/clvm/pull/117
  - https://github.com/Chia-Network/clvm_tools/pull/87
  - https://github.com/Chia-Network/clvm_tools_rs/pull/24
  - https://github.com/Chia-Network/bls-signatures/pull/332
- [x] Release other projects (check mark indicates we are referencing a new enough version)
  - [x] https://github.com/Chia-Network/chiavdf/
  - [x] https://github.com/Chia-Network/chiabip158/
  - [x] https://github.com/Chia-Network/chiapos/commits/1.0.11
  - [x] https://github.com/Chia-Network/chia_rs/
  - [x] https://github.com/Chia-Network/clvm/
  - [x] https://github.com/Chia-Network/clvm_tools/
  - [x] https://github.com/Chia-Network/clvm_tools_rs/
  - [x] https://github.com/Chia-Network/bls-signatures/commits/1.0.16
- [ ] Upstream wheel releases (these are "optional" and we have filled the gaps on pypi.chia.net)
  - [x] frozenlist https://github.com/aio-libs/frozenlist/issues/342
  - [x] memory_profiler https://pypi.org/project/memory-profiler/0.61.0/
  - [x] multidict https://github.com/aio-libs/multidict/issues/778
  - [ ] psutil https://github.com/giampaolo/psutil/issues/2089 https://github.com/giampaolo/psutil/pull/2102
  - [ ] setproctitle https://github.com/dvarrazzo/py-setproctitle/pull/120
  - [x] yarl https://github.com/aio-libs/yarl/issues/781
  - [ ] zstd https://github.com/sergey-dryabzhinsky/python-zstd/issues/83 https://github.com/Chia-Network/build-wheels/pull/9
- [x] If still at a Python 3.11 pre-release, how to handle distributing pre-release wheels.
- [ ] However many fixes
  - [x] https://github.com/Chia-Network/chia-blockchain/pull/13848
  - [x] black 3.11 target in `pyproject.toml` (after https://github.com/Chia-Network/chia-blockchain/pull/14159 is merged)
[...]

PR #14214, opened 2022-12-22
use default_factory= for SpendableCAT mutable defaults

altendky:
extracted from https://github.com/Chia-Network/chia-blockchain/pull/11407

Purpose:

Current Behavior:
New Behavior:

Testing Notes:


arvidn (approved):
ouch, good catch.

altendky:
For some reason 3.11 triggered catching it.  I hadn't gotten into the details, but interesting behavior.  I would note that we have various frozen classes that have mutable attributes and as such aren't all that immutable so be wary how safe you feel around that.  This also makes them unhashable such that they can't be dict keys nor used in a set etc.

PR #14222, opened 2022-12-22
factor out coin- and PH subscription handling in order to add unit tests

arvidn:
The data structure we use to track subscriptions to coins and puzzle hashes is complex and deserves to be isolated in its own class and unit tested.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -35,6 +35,7 @@ from chia.full_node.hint_store import HintStore
 from chia.full_node.lock_queue import LockClient, LockQueue
 from chia.full_node.mempool_manager import MempoolManager
 from chia.full_node.signage_point import SignagePoint
+from chia.full_node.subscriptions import PeerSubscriptions
 from chia.full_node.sync_store import SyncStore
 from chia.full_node.tx_processing_queue import TransactionQueue
 from chia.full_node.weight_proof import WeightProofHandler
@@ -106,17 +107,7 @@ class FullNode:
     multiprocessing_context: Optional[BaseContext]
     _ui_tasks: Set[asyncio.Task[None]]
     db_path: Path
-    # TODO: use NewType all over to describe these various uses of the same types
-    # Puzzle Hash : Set[Peer ID]
-    coin_subscriptions: Dict[bytes32, Set[bytes32]]
-    # Puzzle Hash : Set[Peer ID]
-    ph_subscriptions: Dict[bytes32, Set[bytes32]]
-    # Peer ID: Set[Coin ids]
-    peer_coin_ids: Dict[bytes32, Set[bytes32]]
-    # Peer ID: Set[puzzle_hash]
-    peer_puzzle_hash: Dict[bytes32, Set[bytes32]]
-    # Peer ID: subscription count
-    peer_sub_counter: Dict[bytes32, int]
+    subscriptions: PeerSubscriptions
     _transaction_queue_task: Optional[asyncio.Task[None]]
     simulator_transaction_callback: Optional[Callable[[bytes32], Awaitable[None]]]
     _sync_task: Optional[asyncio.Task[None]]
@@ -179,11 +170,8 @@ class FullNode:
 
         db_path_replaced: str = config["database_path"].replace("CHALLENGE", config["selected_network"])
         self.db_path = path_from_root(root_path, db_path_replaced)
-        self.coin_subscriptions = {}
-        self.ph_subscriptions = {}
-        self.peer_coin_ids = {}
-        self.peer_puzzle_hash = {}
-        self.peer_sub_counter = {}
+
+        self.subscriptions = PeerSubscriptions()
         self.db_path.parent.mkdir(parents=True, exist_ok=True)
         self._transaction_queue_task = None
         self.simulator_transaction_callback = None
@@ -887,27 +875,8 @@ class FullNode:
         self._state_changed("sync_mode")
         if self.sync_store is not None:
             self.sync_store.peer_disconnected(connection.peer_node_id)
-        self.remove_subscriptions(connection)
-
-    def remove_subscriptions(self, peer: WSChiaConnection) -> None:
         # Remove all ph | coin id subscription for this peer
-        node_id = peer.peer_node_id
-        if node_id in self.peer_puzzle_hash:
-            puzzle_hashes = self.peer_puzzle_hash[node_id]
-            for ph in puzzle_hashes:
-                if ph in self.ph_subscriptions:
-                    if node_id in self.ph_subscriptions[ph]:
-                        self.ph_subscriptions[ph].remove(node_id)
-
-        if node_id in self.peer_coin_ids:
-            coin_ids = self.peer_coin_ids[node_id]
-            for coin_id in coin_ids:
-                if coin_id in self.coin_subscriptions:
-                    if node_id in self.coin_subscriptions[coin_id]:
-                        self.coin_subscriptions[coin_id].remove(node_id)
-
-        if peer.peer_node_id in self.peer_sub_counter:
-            self.peer_sub_counter.pop(peer.peer_node_id)
+        self.subscriptions.remove_peer(connection.peer_node_id)
 
     def _num_needed_peers(self) -> int:
         assert self.server.all_connections is not None
@@ -1164,7 +1133,9 @@ class FullNode:
                     assert peak is not None
                     # Hints must be added to the DB. The other post-processing tasks are not required when syncing
                     hints_to_add, lookup_coin_ids = get_hints_and_subscription_coin_ids(
-                        state_change_summary, self.coin_subscriptions, self.ph_subscriptions
+                        state_change_summary,
+                        self.subscriptions.has_coin_sub,
+                        self.subscriptions.has_ph_sub,
                     )
                     await self.hint_store.add_hints(hints_to_add)
                     await self.update_wallets(state_change_summary, hints_to_add, lookup_coin_ids)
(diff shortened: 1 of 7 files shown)


chia/full_node/subscriptions.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def has_coin_sub(self, coin_id: bytes32) -> bool:
+        return coin_id in self._coin_subscriptions
+
+    def __init__(self) -> None:
+        self._coin_subscriptions = {}
+        self._ph_subscriptions = {}
+        self._peer_coin_ids = {}
+        self._peer_puzzle_hash = {}
+        self._peer_sub_counter = {}
```
altendky, line 44: Should be a dataclass.

altendky (dismissed):

chia/full_node/subscriptions.py
arvidn: reading up on dataclasses a bit, it doesn't look like they're meant for this. It looks like `@dataclass` is meant for value types.

arvidn: It's not clear to me what the benefit is to add:
```
@dataclass(init=False, repr=False, eq=False, order=False, unsafe_hash=True, match_args=False)
```

to the class.

altendky: I'm not sure why there is a tendency to think that a clear and consistently structured class is only useful for 'small' or 'trivial' or 'attribute only' classes.  Not that the doc is definitively correct or anything, but since I wrote up my reasons there, here's a link https://github.com/Chia-Network/chia-blockchain/blob/b8068a856d8e2ccd6660f5f6488182e3c1589be8/PRETTY_GOOD_PRACTICES.md#dataclasses.

Why would you write that big huge dataclass decorator that defeats the basic benefit of a dataclass by still using your hand coded init that duplicates the list of hinted attributes?  I'll add a separate comment with a suggestion to show the nominal starting point I would use.

altendky, line 44:
```suggestion
@dataclass
class PeerSubscriptions:
    # TODO: use NewType all over to describe these various uses of the same types
    # Puzzle Hash : Set[Peer ID]
    _coin_subscriptions: Dict[bytes32, Set[bytes32]] = field(default_factory=dict)
    # Puzzle Hash : Set[Peer ID]
    _ph_subscriptions: Dict[bytes32, Set[bytes32]] = field(default_factory=dict)
    # Peer ID: Set[Coin ids]
    _peer_coin_ids: Dict[bytes32, Set[bytes32]] = field(default_factory=dict)
    # Peer ID: Set[puzzle_hash]
    _peer_puzzle_hash: Dict[bytes32, Set[bytes32]] = field(default_factory=dict)
    # Peer ID: subscription count
    _peer_sub_counter: Dict[bytes32, int] = field(default_factory=dict)

    def has_ph_sub(self, ph: bytes32) -> bool:
        return ph in self._ph_subscriptions

    def has_coin_sub(self, coin_id: bytes32) -> bool:
        return coin_id in self._coin_subscriptions
```

chia/full_node/subscriptions.py
```diff
@@ -0,0 +1,137 @@
...
+        return ret
+
+    ret2: Set[bytes32] = set()
+    c[key] = ret2
+    return ret2
+
+
+# The PeerSubscriptions class is essentially a multi-index container. It can be
+# indexed by peer_id, coin_id and puzzle_hash.
+@dataclass
```
xdustinface, line 21:
```suggestion
@dataclass(frozen=True)
```

chia/full_node/full_node.py
```diff
@@ -179,11 +170,8 @@ def __init__(
 
         db_path_replaced: str = config["database_path"].replace("CHALLENGE", config["selected_network"])
         self.db_path = path_from_root(root_path, db_path_replaced)
-        self.coin_subscriptions = {}
-        self.ph_subscriptions = {}
-        self.peer_coin_ids = {}
-        self.peer_puzzle_hash = {}
-        self.peer_sub_counter = {}
+
```
xdustinface, line 173:
```suggestion
```
nit nit..

xdustinface (changes requested):
Looks good but few comments below. Also, why not just make `max_items` a member of `PeerSubscriptions`?

chia/full_node/subscriptions.py
altendky: But this isn't frozen, it's entirely about mutation.  I guess it still blocks assignment of new dicts to the attributes...  I dunno.

tests/core/full_node/test_hint_management.py
```diff
@@ -97,33 +101,38 @@ async def test_lookup_coin_ids(bt: BlockTools, empty_blockchain: Blockchain) ->
     scs = StateChangeSummary(br, uint32(0), [], npc_res, rewards)
 
     # Removal ID and addition PH
-    coin_subscriptions = {coin_ids[1]: {bytes32(b"2" * 32)}}
-    ph_subscriptions = {phs[4]: {bytes32(b"3" * 32)}}
+    has_coin_sub = lambda c: c == coin_ids[1]  # noqa
```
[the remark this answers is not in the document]
altendky: I suggested that maybe `get_hints_subscription_coin_ids()` should be a method.  Not sure if Arvid got to checking out that option quite yet.

chia/full_node/subscriptions.py
```diff
@@ -0,0 +1,137 @@
...
+    _peer_coin_ids: Dict[bytes32, Set[bytes32]] = field(default_factory=dict, init=False)
+    # Peer ID: Set[puzzle_hash]
+    _peer_puzzle_hash: Dict[bytes32, Set[bytes32]] = field(default_factory=dict, init=False)
+    # Peer ID: subscription count
+    _peer_sub_counter: Dict[bytes32, int] = field(default_factory=dict, init=False)
+
+    def has_ph_sub(self, ph: bytes32) -> bool:
+        return ph in self._ph_subscriptions
+
+    def has_coin_sub(self, coin_id: bytes32) -> bool:
```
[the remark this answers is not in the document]
altendky: I do prefer not having to guess what abbreviation to use or what an abbreviation means.

chia/full_node/subscriptions.py
xdustinface: Its only about mutation of the members though, what is still possible when its frozen?

altendky: Frozen only blocks reassignment of attributes, not access to them or modification of them.  When I think about frozen, I think of classes that have immutable attributes all the way down the tree.  A whole immutable tree of data once created.  Maybe that's too closed minded.

I guess frozen also blocks addition of new attributes, but slots can trigger that as well and mypy should protect against it generally.  Frozen also adjusts expectations around hashing where you'll get an exception about the dicts being unhashable instead of the instance of this class being unhashable.

chia/full_node/subscriptions.py
```diff
@@ -0,0 +1,137 @@
...
+        # decrement this counter as we go, to know if we've hit the limit of
+        # number of subscriptions
+        subscriptions_left = max_items - existing_sub_count
+
+        for ph in phs:
+            ph_sub: Optional[Set[bytes32]] = self._ph_subscriptions.get(ph)
+            if ph_sub is None:
+                ph_sub = set()
+                self._ph_subscriptions[ph] = ph_sub
+            elif peer_id in ph_sub:
```
almogdepaz, line 62: we should add some logging for the return paths, will help people with lots of coins/ph to understand why/if they don't get all their coins when syncing from a local node

chia/full_node/subscriptions.py
```diff
@@ -0,0 +1,137 @@
...
+    return ret2
+
+
+# The PeerSubscriptions class is essentially a multi-index container. It can be
+# indexed by peer_id, coin_id and puzzle_hash.
+@dataclass
+class PeerSubscriptions:
+    # TODO: use NewType all over to describe these various uses of the same types
+    # Puzzle Hash : Set[Peer ID]
+    _coin_subscriptions: Dict[bytes32, Set[bytes32]] = field(default_factory=dict, init=False)
```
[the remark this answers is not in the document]
arvidn: `init=False` on a member means it won't be a parameter to the generated `__init__()` function. In this class, these fields are private and the class invariant rely on them not being initialized from the outside.

tests/core/full_node/test_hint_management.py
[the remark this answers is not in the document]
arvidn: yeah, it sounds like it could be a good idea, however I'm hesitant to explore that in this PR because:
1. I'd like to stay focused factoring out the subscription data structure, and pulling this function in as a member would significantly increase the size of the patch (the test would have to be updated)
2. It's not obvious that it's better to make it a member. As a free function it's much simpler to unit test.

chia/full_node/subscriptions.py
arvidn: good point.

chia/full_node/subscriptions.py
arvidn: I must have misunderstood what frozen meant. I thought it meant that the object's value was immutable, but it doesn't seem to mean that. I'll add `frozen=True`

xdustinface (changes requested):
There is a GUI submodule update which you should drop. Also seems like you missed it, so, do you have any thought about the max item question here https://github.com/Chia-Network/chia-blockchain/pull/14222#pullrequestreview-1228469161?

arvidn:
> Looks good but few comments below. Also, why not just make `max_items` a member of `PeerSubscriptions`?

I had reviewed almog's PR before starting on this, so I knew we were going to have different peers have different limits. His patch has now landed and I've rebased this on top of it. You can see now that `max_items` depends on whether the peer is trusted or not.

xdustinface (approved):

chia/full_node/subscriptions.py
```diff
@@ -0,0 +1,133 @@
...
+        existing_sub_count = self._peer_sub_counter.setdefault(peer_id, 0)
+
+        # if we've reached the limit on number of subscriptions, just bail
+        if existing_sub_count >= max_items:
+            log.info(
+                "peer_id: %s reached max number of puzzle-hash subscriptions. "
+                "Not all its coin states will be reported",
+                peer_id,
+            )
+            return
```
trepca, line 45: maybe return `False` or `True` so function caller could potentially report to the peer that not all subscriptions were processed.
Not sure how else will wallet/peer know this.

arvidn: right, it would be nice to have some feedback apart from logging. The wallet protocol doesn't really support that though, so that's a bigger change

trepca (approved):

PR #14227, opened 2022-12-24
cleanup some wallet polymorphism

altendky:
some issues exposed in https://github.com/Chia-Network/chia-blockchain/pull/10893

Purpose:

Current Behavior:
New Behavior:

Testing Notes:


altendky:
As far as coverage of this change, the error cases are almost all missed as well as the handling of a DataLayer wallet in `NFTWallet.make_nft1_offer()`.  Seems off that the offer creation involves the NFT wallet having to handle the other wallets.  Also, line coverage is certainly not the whole picture.

PR #14229, opened 2022-12-24
only send_to_all(), drop send_to_all_except()

altendky:
Purpose:
Remove duplicate code by just providing a default for the optional parameter.

Current Behavior:
New Behavior:

Testing Notes:

### DIFF
--- a/chia/server/server.py
+++ b/chia/server/server.py
@@ -573,14 +573,12 @@ class ChiaServer:
                         )
                 raise ProtocolError(Err.INTERNAL_PROTOCOL_ERROR, [message.type])
 
-    async def send_to_all(self, messages: List[Message], node_type: NodeType) -> None:
-        await self.validate_broadcast_message_type(messages, node_type)
-        for _, connection in self.all_connections.items():
-            if connection.connection_type is node_type:
-                for message in messages:
-                    await connection.send_message(message)
-
-    async def send_to_all_except(self, messages: List[Message], node_type: NodeType, exclude: bytes32) -> None:
+    async def send_to_all(
+        self,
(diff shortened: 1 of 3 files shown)


chia/server/server.py
```diff
@@ -573,14 +573,12 @@ async def validate_broadcast_message_type(self, messages: List[Message], node_ty
...
-            if connection.connection_type is node_type:
-                for message in messages:
-                    await connection.send_message(message)
-
-    async def send_to_all_except(self, messages: List[Message], node_type: NodeType, exclude: bytes32) -> None:
+    async def send_to_all(
+        self,
+        messages: List[Message],
+        node_type: NodeType,
+        exclude: Optional[bytes32] = None,
```
arvidn, line 580: I think it might look good at call sites to call this parameter `except` and make it mandatory to name it

arvidn (approved):

chia/server/server.py
altendky: I'm mostly hoping that https://github.com/PyCQA/pylint/pull/7852 evolves to something usable for enforcing named argument passing in nearly all cases.  I don't like having to try to decide on a case by case basis which are ok as positional and which deserve to be passed by name.  Do you have some rules in mind for when we should make a function definition itself enforce the used of named args?

Since this is green and approved and I am hoping to deal with pass by name in a different way, I'll let this get merged as is.  But, if you want to raise a concern, I'll happily make the modification here.

arvidn: I don't have a very solid rule, but I think an obvious example is where you have multiple (but maybe even one) `bool` parameter to a function

PR #14251, opened 2023-01-02
update GUI when wallets are removed

trepca:
Purpose:
update GUI when wallets are removed so it updates it's internal cache 

Current Behavior: 
if wallets are removed, GUI is unaware and still shows them 
New Behavior:
GUI will be able to update and remove a wallet 

Testing Notes:
best way to test is to remove DID so that wallet gets removed automatically


paninaro:
@trepca is there a simple way to repro the wallet deletion behavior? I'd like to be able to repro the behavior before/after the fix to verify that the GUI is behaving properly.

emlowe:
Close and reopen for new CI run

emlowe (approved):

PR #14253, opened 2023-01-02
Try each Chia DNS Server in list before trying introducers

### LINKED ISSUE #14252: [Enhancement] Support multiple DNS seeders and do not connect unless necessary for node and wallet
### What happened?

We have dns seeders working for node and wallet but they currently depend on addresses in the *.chia.net domain. Documentation for running a [chia seeder is here](https://github.com/Chia-Network/chia-blockchain/wiki/Chia-Seeder-User-Guide). [initial-config.yaml](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/util/initial-config.yaml#L408) has a list data structure but node and wallet only try the first one on the list and if it times out in two seconds, move on to classic/dumb introducer which is also hardcoded to an address at *.chia.net.

We need to enhance both wallet and node to first attempt to connect to peers from their peer dbs and then and only then should they go out to the dns seeders. This is both a privacy enhancement so that nodes aren't leaking IP addresses to these semi-centralized hosts but is also a more efficient use of the network and these services.

Why? This is the last point of dependence on CNI - though of course this could be fixed by the community in a fork later. However it is best to go ahead and get this working correctly and then we'll create a short list of community sourced dns seeders into `initial-config.yaml` to be used at random to obtain additional node peers as needed.

Some Bitcoin PRs and documentation as a reference jumping off point: https://bitcoincore.reviews/16939

### Version

8ea3f933395a462440d02f10fa44665c0ac675a0

macOS

CLI

### Relevant log output

```shell
# Log from node startup where community dns seeder is first and intentionally disabled with dns-introducer.chia.net listed second in config.yaml
2022-12-31T16:51:30.482 full_node chia.server.address_manager_store: INFO     Creating new address_manager
2022-12-31T16:51:30.491 full_node full_node_server        : INFO     Started listening on port: 8444
2022-12-31T16:51:30.491 full_node full_node               : INFO     Started full_node service on network_id: mainnet
2022-12-31T16:51:30.502 daemon chia.daemon.server         : INFO     Register service {'service': 'chia_full_node'}
2022-12-31T16:51:30.503 daemon chia.daemon.server         : INFO     registered for service chia_full_node
2022-12-31T16:51:30.503 daemon chia.daemon.server         : INFO     {'success': True}
2022-12-31T16:51:34.127 full_node chia.full_node.full_node: WARNING  querying DNS introducer failed: All nameservers failed to answer the query chia.hoffmang.com. IN A: Server 100.100.100.100 UDP port 53 answered The DNS operation timed out after 2.000 seconds; Server 100.100.100.100 UDP port 53 answered SERVFAIL
2022-12-31T16:51:36.351 full_node full_node_server        : INFO     Connected with introducer PeerInfo(host='introducer.chia.net', port=8444)
2022-12-31T16:51:36.353 full_node full_node_server        : INFO     Connection closed: introducer.chia.net, node id: d8fbd3bf3819fd9586a9f68e5869baf762f59fe1f239fad7c9b65ab30bc99bd9
2022-12-31T16:51:36.354 full_node chia.full_node.full_node: INFO     peer disconnected PeerInfo(host='100.21.23.17', port=8444)
2022-12-31T16:51:45.536 full_node full_node_server        : INFO     Connected with introducer PeerInfo(host='introducer.chia.net', port=8444)
2022-12-31T16:51:45.538 full_node full_node_server        : INFO     Connection closed: introducer.chia.net, node id: d8fbd3bf3819fd9586a9f68e5869baf762f59fe1f239fad7c9b65ab30bc99bd9
2022-12-31T16:51:45.539 full_node chia.full_node.full_node: INFO     peer disconnected PeerInfo(host='44.227.187.109', port=8444)
```

aqk:
This changes the behaviour of the full node: 

1. Try each dns server in the list
2. Try the introducer
3. Repeat until success, obeying the exponential backoff with each attempt of either type.

Further, do not always start with the head of the dns server list. This makes sense for the common user, to reduce dns server load, and prevent DNS server capture by an opponent. Note that this removes the ability to use the dns server list as an ad-hoc priority list.

### DIFF
--- a/chia/server/node_discovery.py
+++ b/chia/server/node_discovery.py
@@ -272,10 +272,18 @@ class FullNodeDiscovery:
         next_feeler = self._poisson_next_send(time.time() * 1000 * 1000, 240, random)
         retry_introducers = False
         introducer_attempts: int = 0
+        dns_start_index: int = 0
         dns_server_index: int = 0
+        tried_all_dns_servers: bool = False
         local_peerinfo: Optional[PeerInfo] = await self.server.get_peer_info()
         last_timestamp_local_info: uint64 = uint64(int(time.time()))
         last_collision_timestamp = 0
+
+        if len(self.dns_servers) > 0:
+            # Don't always start with the same DNS server
+            dns_start_index = random.randint(0, len(self.dns_servers) - 1)
+            dns_server_index = dns_start_index
+
         if self.initial_wait > 0:
             await asyncio.sleep(self.initial_wait)
 
@@ -291,13 +299,16 @@ class FullNodeDiscovery:
                         await asyncio.sleep(introducer_backoff)
                     except asyncio.CancelledError:
                         return None
-                    # Run dual between DNS servers and introducers. One time query DNS server,
-                    # next two times query the introducer.
-                    if introducer_attempts % 3 == 0 and len(self.dns_servers) > 0:
+                    # Alternate between DNS servers and introducers.
+                    # First try all the DNS servers in the list once. Then try the introducers once.
+                    if len(self.dns_servers) > 0 and not tried_all_dns_servers:
                         dns_address = self.dns_servers[dns_server_index]
-                        dns_server_index = (dns_server_index + 1) % len(self.dns_servers)
+                        next_index = (dns_server_index + 1) % len(self.dns_servers)
+                        tried_all_dns_servers = next_index == dns_start_index
+                        dns_server_index = next_index
                         await self._query_dns(dns_address)
                     else:
+                        tried_all_dns_servers = False
                         await self._introducer_client()
                         # there's some delay between receiving the peers from the
                         # introducer until they get incorporated to prevent this
--- a/chia/server/node_discovery.py
+++ b/chia/server/node_discovery.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import math
+import random
 import time
 import traceback
 from logging import Logger
@@ -58,6 +59,7 @@ class FullNodeDiscovery:
         self.legacy_peer_db_migrated = False
         self.peers_file_path = peer_store_resolver.peers_file_path
         self.dns_servers = dns_servers
+        random.shuffle(dns_servers)  # Don't always start with the same DNS server
         if introducer_info is not None:
             self.introducer_info: Optional[PeerInfo] = PeerInfo(
                 introducer_info["host"],
@@ -271,18 +273,11 @@ class FullNodeDiscovery:
     async def _connect_to_peers(self, random: Random) -> None:
         next_feeler = self._poisson_next_send(time.time() * 1000 * 1000, 240, random)
         retry_introducers = False
-        introducer_attempts: int = 0
-        dns_start_index: int = 0
         dns_server_index: int = 0
         tried_all_dns_servers: bool = False
         local_peerinfo: Optional[PeerInfo] = await self.server.get_peer_info()
         last_timestamp_local_info: uint64 = uint64(int(time.time()))
         last_collision_timestamp = 0
-
-        if len(self.dns_servers) > 0:
-            # Don't always start with the same DNS server
-            dns_start_index = random.randint(0, len(self.dns_servers) - 1)
-            dns_server_index = dns_start_index
 
         if self.initial_wait > 0:
             await asyncio.sleep(self.initial_wait)
@@ -303,9 +298,8 @@ class FullNodeDiscovery:
                     # First try all the DNS servers in the list once. Then try the introducers once.
                     if len(self.dns_servers) > 0 and not tried_all_dns_servers:
                         dns_address = self.dns_servers[dns_server_index]
-                        next_index = (dns_server_index + 1) % len(self.dns_servers)
-                        tried_all_dns_servers = next_index == dns_start_index
-                        dns_server_index = next_index
+                        dns_server_index = (dns_server_index + 1) % len(self.dns_servers)
+                        tried_all_dns_servers = dns_server_index == 0
                         await self._query_dns(dns_address)
                     else:
                         tried_all_dns_servers = False
@@ -320,7 +314,6 @@ class FullNodeDiscovery:
                             return None
 
                     retry_introducers = False
-                    introducer_attempts += 1
                     # keep doubling the introducer delay until we reach 5
                     # minutes
                     if introducer_backoff < 300:


aqk:
See https://github.com/Chia-Network/chia-blockchain/issues/14252

hoffmang9:
I don't think this is getting at the functionality change we need. In my tests this appears to just make us fall back and use the dumb old introducer which, frankly, we should rip out (I suspect the only reason we don't is ease of testnet but I digress.)

With chia.hoffmang.com and dns-introducer.chia.net as my two dns introducers in config.yaml, my node only seems to connect to the old and busted introducer. Optimally we would want to start from the peer db (I have nuked it in my test configuration) and then if we can't get connections go to the list of dns introducers and if we hit ultimate futility - try the old and busted style.

One of the drivers of this is that the dns introducers do a bit of peer quality checking so the ones they're handing out are known to be pretty good and to have ports available for new incoming connections. Old and busted introducer just dumbly hands out any IP it could connect back to that connected to it at some point with a time decay.

chia/server/node_discovery.py
```diff
@@ -291,9 +292,13 @@ async def _connect_to_peers(self, random: Random) -> None:
...
                         return None
-                    # Run dual between DNS servers and introducers. One time query DNS server,
-                    # next two times query the introducer.
-                    if introducer_attempts % 3 == 0 and len(self.dns_servers) > 0:
+                    # Alternate between DNS servers and introducers.
+                    # First (self.dns_query_ratio - 1) times query the introducer.
+                    # Query the next Chia DNS server in the list once every self.dns_query_ratio attempts.
+                    if (
+                        len(self.dns_servers) > 0
+                        and (introducer_attempts % self.dns_query_ratio) == self.dns_query_ratio - 1
```
arvidn, line 300: is the movement of `len(self.dns_servers) > 0` to the front of the expression relevant? is it to save a modulus when we don't have any dns servers?

Other than trying the DNS last, you also change the ratio from 3 to 10. Do these changes all have the objective to reduce load on the introducers?

arvidn (approved):
looks reasonable, assuming the main objective is to reduce load on our introducers.

Unless I'm misunderstanding, I think you saying "local" introducer is misleading. We have a client object that talks to an introducer, right?

arvidn (changes requested):
reading the ticket, this PR does not seem to implement the desired behavior.

chia/server/node_discovery.py
```diff
@@ -38,6 +38,7 @@
 
 class FullNodeDiscovery:
     resolver: Optional[dns.asyncresolver.Resolver]
+    dns_query_ratio = 10  # Contact introducer ("chia dns server") once per dns_query_ratio attempts to get a peer
```
arvidn, line 41: this comment is a bit misleading. the "introducer" and "DNS seed" are distinct services. My impression from reading the ticket is that we should:
1. remove the affinity towards finding peers from the DNS seeds (and only fall back onto DNS if we fail to find a peer after a sufficient number of attempts from our peer DB)
2. shuffle the `dns_servers` list in order to pick one at random
3. probably never contact the introducer service (but maybe a safer first step would be to log a warning if we do)

chia/server/node_discovery.py
aqk: > I don't think this is getting at the functionality change we need. In my tests this appears to just make us fall back and use the dumb old introducer which, frankly, we should rip out (I suspect the only reason we don't is ease of testnet but I digress.)
> 
> With chia.hoffmang.com and dns-introducer.chia.net as my two dns introducers in config.yaml, my node only seems to connect to the old and busted introducer. Optimally we would want to start from the peer db (I have nuked it in my test configuration) and then if we can't get connections go to the list of dns introducers and if we hit ultimate futility - try the old and busted style.

> One of the drivers of this is that the dns introducers do a bit of peer quality checking so the ones they're handing out are known to be pretty good and to have ports available for new incoming connections. Old and busted introducer just dumbly hands out any IP it could connect back to that connected to it at some point with a time decay.

Thanks. I think the issue here is that I'm not understanding the full issue (eg. There is an old, broken introducer, and newer, better introducers -- what is the name of the old introducer?), and also misunderstanding what we are calling "introducers" and "DNS Servers" in the code.

I'll look up the docs on what the intent / desired situation is.

With the incomplete knowledge I have, I'm all for removing the "bad" introducer client side - we can wait to remove the server itself until usage dwindles to nothing.

aqk: > is the movement of `len(self.dns_servers) > 0` to the front of the expression relevant? is it to save a modulus when we don't have any dns servers?

No, the change has to do with automatic code formatting. I believe this code has the same meaning, even in the face of short-circuiting `and`.

> Other than trying the DNS last, you also change the ratio from 3 to 10. Do these changes all have the objective to reduce load on the introducers?

Yes, that is what I understood from the ticket, but see my above comment to Gene. I was not aware of one of the issues when I put this together.

chia/server/node_discovery.py
aqk: Thanks - this exactly what I was misunderstanding. I was aware that we had introducers that handed out peers, but I did not know we had a "DNS seed" service - can you give a brief synopsis of the "DNS Seed" idea?

arvidn: it's the same mechanism bitcoin uses, where the network of caching DNS servers is leveraged to hand out peer IPs.

our root server probably responds with a ranom subset of nodes we know of to each query, and the cache in the network ensures our load stays reasonable. Since we want to encourage community-run DNS seeds, we want to pick which one to use at random.

hoffmang9: The code seems to call the dns introducer the dns server and the old and busted introducer - just introducer

aqk:
I misunderstood what was wanted in the ticket. After talking with Arvid & Earle, I believe what we want short-term is:

1. Try each dns server in the list
2. Try the introducer
3. Repeat until success, obeying the exponential backoff with each attempt of either type.

Further, do not always start with the head of the dns server list. This makes sense for the common user, to reduce dns server load, and prevent DNS server capture by an opponent. Note that this removes the ability to use the dns server list as an ad-hoc priority list.

hoffmang9:
> I misunderstood what was wanted in the ticket. After talking with Arvid & Earle, I believe what we want short-term is:
> 
> 1. Try each dns server in the list
> 2. Try the introducer
> 3. Repeat until success, obeying the exponential backoff with each attempt of either type.
> 
> Further, do not always start with the head of the dns server list. This makes sense for the common user, to reduce dns server load, and prevent DNS server capture by an opponent. Note that this removes the ability to use the dns server list as an ad-hoc priority list.

Exactly this.

chia/server/node_discovery.py
```diff
@@ -272,10 +272,18 @@ async def _connect_to_peers(self, random: Random) -> None:
         next_feeler = self._poisson_next_send(time.time() * 1000 * 1000, 240, random)
         retry_introducers = False
         introducer_attempts: int = 0
```
arvidn, line 274: it doesn't look like we need this counter anymore

chia/server/node_discovery.py
```diff
@@ -291,13 +299,16 @@ async def _connect_to_peers(self, random: Random) -> None:
...
                         return None
-                    # Run dual between DNS servers and introducers. One time query DNS server,
-                    # next two times query the introducer.
-                    if introducer_attempts % 3 == 0 and len(self.dns_servers) > 0:
+                    # Alternate between DNS servers and introducers.
+                    # First try all the DNS servers in the list once. Then try the introducers once.
+                    if len(self.dns_servers) > 0 and not tried_all_dns_servers:
                         dns_address = self.dns_servers[dns_server_index]
-                        dns_server_index = (dns_server_index + 1) % len(self.dns_servers)
+                        next_index = (dns_server_index + 1) % len(self.dns_servers)
```
arvidn, line 306: `next_index` seems redundant. it looks like you could just increment `dns_server_index` directly

arvidn (commented):
I think you could simplify this patch a bit by removing a few variables and lines of code.

In the bigger picture, this patch is also supposed to ensure that the peer database is always loaded and considered *before* getting here. Whether we try a DNS seed or the introducer. I'm not very familiar with the details, but from gene's log it looked like we were very eager to contact the introducer.

Another minor observation is that if we only have one DNS seed in our list, we will still contact the introducer every other time we get here. That's still a lot better than our current 2 out of 3 times, but perhaps we would like to have a stronger affinity towards DNS.

chia/server/node_discovery.py
```diff
@@ -272,10 +272,18 @@ async def _connect_to_peers(self, random: Random) -> None:
...
         dns_server_index: int = 0
+        tried_all_dns_servers: bool = False
         local_peerinfo: Optional[PeerInfo] = await self.server.get_peer_info()
         last_timestamp_local_info: uint64 = uint64(int(time.time()))
         last_collision_timestamp = 0
+
+        if len(self.dns_servers) > 0:
+            # Don't always start with the same DNS server
+            dns_start_index = random.randint(0, len(self.dns_servers) - 1)
+            dns_server_index = dns_start_index
```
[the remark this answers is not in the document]
emlowe: oh TIL about `random.shuffle(list)`

chia/server/node_discovery.py
aqk: Thanks

chia/server/node_discovery.py
aqk: True. I just thought it was more clear naming it, to clarify that we intended to use the old value, not the new value.

aqk: (That is, use the old value to index `dns_address`)

aqk:
@arvidn Take a look when you get a chance - addressed all your above comments.

arvidn (approved):

PR #14254, opened 2023-01-02
Use better check that we are on mainnet when deciding to use default Chia DNS server

aqk:
We previously checked if the full node was using port 8444 to decide this. Now we use our network name.


emlowe (approved):
Seems reasonable

PR #14263, opened 2023-01-03
Add MempoolRemoveReason, which represents why we are removing a MempoolItem

aqk:
The Mempool interface did not have the concept of why an item was being removed from the Mempool (e.g. Mempool Full, Replace by Fee, etc). This patch adds that. We can now use this information in lower layers. This patch also declines to call `FeeEstimatorInterface.remove_mempool_item` when the reason is `MempoolRemoveReason.BLOCK_INCLUSION`. The semantics of the FeeEstimatorInterface dictate that that case be handled by `FeeEstimatorInterface.new_block`.

### DIFF
--- a/chia/full_node/mempool.py
+++ b/chia/full_node/mempool.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 from datetime import datetime
+from enum import Enum
 from typing import Dict, List, Optional
 
 from sortedcontainers import SortedDict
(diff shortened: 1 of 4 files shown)


arvidn (approved):

arvidn:
I envision a future incarnation of the fee estimator interface being a more generic *observer* (yes, from the Gang of four Design Patterns :) ).

Where the knowledge to disregard the block-inclusion case is in the fee estimator rather than the mempool.

PR #14266, opened 2023-01-03
Add current block height member to FeeMempoolInfo

aqk:
No user visible changes. We add the concept of current block height to `FeeMempoolInfo`, a struct which holds information about the current state of the Mempool. This can be used by the Fee Estimator to gauge the passing "blockchain time" as the methods in its interface are called.

### DIFF
--- a/chia/full_node/fee_estimation.py
+++ b/chia/full_node/fee_estimation.py
@@ -44,6 +44,7 @@ class FeeMempoolInfo:
     mempool_info: MempoolInfo
     current_mempool_cost: CLVMCost  # Current sum of CLVM cost of all SpendBundles in mempool (mempool "size")
     time: datetime  # Local time this sample was taken
+    height: uint32  # Current blockchain height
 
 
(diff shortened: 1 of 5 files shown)


arvidn (commented):
it seems a bit awkward to pass in the height into these functions just for them to pass it on to the `FeeEstimator` observer. It seems it might make more sense to have a separate callback on the FeeEstimator interface to let in know we have changed peak (both a new block or a reorg)

aqk:
> it seems a bit awkward to pass in the height into these functions just for them to pass it on to the FeeEstimator observer. 

Yeah, I didn't like that either.

> It seems it might make more sense to have a separate callback on the FeeEstimator interface to let in know we have changed peak (both a new block or a reorg)

Thanks, that sounds better to me too. I'll definitely take that suggestion.

aqk:
Implemented suggestions and Moved to new PR:
https://github.com/Chia-Network/chia-blockchain/pull/14277

PR #14270, opened 2023-01-04
Refactor get_bucket_index

aqk:
Code maintenance.

### DIFF
--- a/chia/full_node/fee_tracker.py
+++ b/chia/full_node/fee_tracker.py
@@ -61,6 +61,16 @@ def get_estimate_time_intervals() -> List[uint64]:
     return [uint64(blocks * SECONDS_PER_BLOCK) for blocks in get_estimate_block_intervals()]
 
 
+def get_bucket_index(sorted_buckets: SortedDict, fee_rate: float) -> int:
+    if fee_rate in sorted_buckets:
+        bucket_index = sorted_buckets[fee_rate]
+    else:
+        # Choose the bucket to the left if we do not have exactly this fee rate
+        bucket_index = sorted_buckets.bisect_left(fee_rate) - 1
+
+    return int(bucket_index)
+
+
 # Implementation of bitcoin core fee estimation algorithm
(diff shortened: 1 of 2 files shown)


chia/full_node/fee_tracker.py
```diff
@@ -61,6 +61,16 @@ def get_estimate_time_intervals() -> List[uint64]:
     return [uint64(blocks * SECONDS_PER_BLOCK) for blocks in get_estimate_block_intervals()]
 
 
+def get_bucket_index(sorted_buckets: SortedDict, fee_rate: float) -> int:
+    if fee_rate in sorted_buckets:
+        bucket_index = sorted_buckets[fee_rate]
+    else:
+        # Choose the bucket to the left if we do not have exactly this fee rate
+        bucket_index = sorted_buckets.bisect_left(fee_rate) - 1
```
arvidn, line 69: it seems like it would be simpler and faster to *just* use `bisect_left()`. your keys are floats after all, an exact hit is presumably unlikely, no?

arvidn: why do you need to do `- 1` btw? presumably `bisect_left()` will give you the first element.
It seems there would be a risk you end up with index `-1` here

arvidn (changes requested):
now that you have factored this our to be a free function is a perfect opportunity to add unit tests to it. Specifically, I would like to see tests where you exercise the `bisect_left()` path and try to get to the first and last bucket (so prove that the -1 is correct)

chia/full_node/fee_tracker.py
aqk: It's funny you mention it. The NEXT PR after this one fixes that and adds tests.

I just want to remove this refactor from the actual bugfix commit.

I would prefer to put this refactor in first, and then the bugfix - let me know which way you want me to go.

aqk: Yes. This patch is a prelude to a fix for that. Take a look at https://github.com/Chia-Network/chia-blockchain/pull/14270#discussion_r1061940380

arvidn (approved):

PR #14274, opened 2023-01-04
return failure if mempool manager has no peak

altendky:
Purpose:

This avoids an exception that seems to block ... well, general functioning anyways.  I made a trivial "fix" and Adam thought it was ok.  Anyone else have opinions?

Current Behavior:

```python-traceback
Jan 04 15:18:54 fullnode chia_full_node[3315992]: 2023-01-04T15:18:54.301 full_node chia.full_node.full_node: ERROR    Error in _handle_one_transaction, closing: Traceback (most recent call last):
Jan 04 15:18:54 fullnode chia_full_node[3315992]:   File "/farm/chia-blockchain/chia/full_node/full_node.py", line 458, in _handle_one_transaction
Jan 04 15:18:54 fullnode chia_full_node[3315992]:     inc_status, err = await self.respond_transaction(entry.transaction, entry.spend_name, peer, entry.test)
Jan 04 15:18:54 fullnode chia_full_node[3315992]:   File "/farm/chia-blockchain/chia/full_node/full_node.py", line 2262, in respond_transaction
Jan 04 15:18:54 fullnode chia_full_node[3315992]:     assert self.mempool_manager.peak
Jan 04 15:18:54 fullnode chia_full_node[3315992]: AssertionError
```

New Behavior:

Testing Notes:

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -2259,10 +2259,11 @@ class FullNode:
                 if self.mempool_manager.get_spendbundle(spend_name) is not None:
                     self.mempool_manager.remove_seen(spend_name)
                     return MempoolInclusionStatus.SUCCESS, None
-                if self.mempool_manager.peak is not None:
-                    cost, status, error = await self.mempool_manager.add_spend_bundle(
-                        transaction, cost_result, spend_name, self.mempool_manager.peak.height
-                    )
+                if self.mempool_manager.peak is None:
+                    return MempoolInclusionStatus.FAILED, Err.MEMPOOL_NOT_INITIALIZED
+                cost, status, error = await self.mempool_manager.add_spend_bundle(
+                    transaction, cost_result, spend_name, self.mempool_manager.peak.height
+                )
             if status == MempoolInclusionStatus.SUCCESS:
                 self.log.debug(
                     f"Added transaction to mempool: {spend_name} mempool size: "
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -2259,10 +2259,10 @@ class FullNode:
                 if self.mempool_manager.get_spendbundle(spend_name) is not None:
                     self.mempool_manager.remove_seen(spend_name)
                     return MempoolInclusionStatus.SUCCESS, None
-                assert self.mempool_manager.peak
-                cost, status, error = await self.mempool_manager.add_spend_bundle(
-                    transaction, cost_result, spend_name, self.mempool_manager.peak.height
-                )
+                if self.mempool_manager.peak is not None:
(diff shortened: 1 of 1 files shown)


aqk (dismissed):

chia/full_node/full_node.py
```diff
@@ -2259,10 +2259,10 @@ async def respond_transaction(
...
                     return MempoolInclusionStatus.SUCCESS, None
-                assert self.mempool_manager.peak
-                cost, status, error = await self.mempool_manager.add_spend_bundle(
-                    transaction, cost_result, spend_name, self.mempool_manager.peak.height
-                )
+                if self.mempool_manager.peak is not None:
+                    cost, status, error = await self.mempool_manager.add_spend_bundle(
+                        transaction, cost_result, spend_name, self.mempool_manager.peak.height
+                    )
             if status == MempoolInclusionStatus.SUCCESS:
```
arvidn, line 2266: it looks like, with this change, the next error will be that `status` is not defined.

arvidn (commented):
I'm not super familiar, but I would expect that whoever put that `assert` there considered this a bug.
Are we confident that we're not just sweeping a bug under the carpet by doing this?

AmineKhaldi:
`self.mempool_manager.peak` being `None` seems to indicate mempool not initialized, so I would replace the assert with `return MempoolInclusionStatus.FAILED, Err.MEMPOOL_NOT_INITIALIZED` when that's the case.

AmineKhaldi (commented):
`self.mempool_manager.peak` being `None` seems to indicate mempool not initialized, so I would replace the assert with `return MempoolInclusionStatus.FAILED, Err.MEMPOOL_NOT_INITIALIZED` when that's the case.

altendky:
Yeah, I'm confused as to why this made my node ok...  thanks.

altendky:
> I'm not super familiar, but I would expect that whoever put that `assert` there considered this a bug. Are we confident that we're not just sweeping a bug under the carpet by doing this?

That was @aqk in https://github.com/Chia-Network/chia-blockchain/commit/5cbc415589e13c7e2be4769ca1cd44c96076c161.

I just pushed Amine's suggestion.  Is that returning 'mempool not initialized' to a peer?  Seems like we shouldn't be accepting peer connections to begin with...  But I think that is getting into the discussion I had with Adam and we can take a bigger pass at that another time.

arvidn:
while syncing we need peers but not a mempool. In that case, ignoring transactions seems like the right thing to do. But transactions aren't pushed, they're pulled, so we should probably not be requesting transactions from peers while syncing either

arvidn (approved):

emlowe:
@aqk Adam, any thoughts on whether returning an error here might be covering up some edge case we should try to otherwise handle?

PR #14277, opened 2023-01-04
Add new_block_height to FeeEstimatorInterface

aqk:
No user visible changes. We add the ability to update the current block height via `FeeEstimatorInterface.new_block_height`. This can be used by the Fee Estimator to gauge the passing "blockchain time" at a finer level of detail, rather than only updating it in `FeeEstimatorInterface.new_block`, as was done before.


arvidn:
you need to make `mempool_manager.py` call this new callback too, right?
The change to `test_mempool_manager.py` seems incomplete.

arvidn (approved):

aqk:
> you need to make `mempool_manager.py` call this new callback too, right? The change to `test_mempool_manager.py` seems incomplete.

Ya, thanks. I didn't have the test for it yet.

PR #14278, opened 2023-01-04
Limit rate of spammy log messages when farmer is disconnected from pool

aqk:
Limit rate of spammy log messages when farmer is disconnected from pool

### DIFF
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -5,8 +5,9 @@ import json
 import logging
 import time
 import traceback
+from collections import defaultdict
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Set, Tuple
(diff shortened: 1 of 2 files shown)


arvidn (commented):
this looks like a reasonable feature. The duplication of code is a bit troubling. Is there no python logging adaptor that does something like this off-the-shelf?

aqk:
> this looks like a reasonable feature. The duplication of code is a bit troubling. Is there no python logging adaptor that does something like this off-the-shelf?

You know, last time I looked, I couldn't find one. I'll take a look again. Might not be worth another dependency if it's not already in `logging.Logger`

I also debated putting something a bit more reusable in a utility function.

aqk:
New PR here: https://github.com/Chia-Network/chia-blockchain/pull/14344

PR #14280, opened 2023-01-05
Configure fullnode peer

ardikars:
Purpose:
Configure the wallet so that it can connect to a certain full node that has a good network connection.

Current Behavior:
Currently the wallet is connected to the local node by default, this can be changed by modifying the config.yml manually.

New Behavior:
Changing full node peer can be done via CLI (`${fullnode_address}:${fullnode_port}`).

Testing Notes:
This PR is related to [#179](https://github.com/Chia-Network/chia-docker/pull/179)


Starttoaster (dismissed):
My mistake on not seeing this sooner. I don't have a good system set up to track notifications on here, usually I'm pinged in keybase. I played around with this locally. The code changes make sense and I can confirm it works as designed.

PR #14281, opened 2023-01-05
Add proxy support to Data Layer download client

Chida82:
Purpose: Add a possibility to use a proxy when download Data Layer files. 
It is useful for future development and analysis of protocols, such as gzip 

No breaking change. If you don't configure the parameter, everything works as before.

Testing Notes: If you want to test, you can use a proxy like Fiddler and configure the "proxy_url" parameter under "data_layer" in the config file. Then you need to subscribe to a Data layer ( eg: chia data subscribe --id b143010e278e55389f426e3f81ac8ba1dfcbd0354bb5d59d49fabf83115986a6 ) and from that moment you should see the calls in the proxy. Obviously first you have to delete the folder and the db, to force you to download everything.

This is an example from proxy:
image


Chida82:
> were

Solved

emlowe (approved):
looks good to me - I wonder if similar support for the other services might also be worthwhile (although they use websockets)

Chida82:
> looks good to me - I wonder if similar support for the other services might also be worthwhile (although they use websockets)

Great suggestion. I think it's useful for outbound calls. I'll think about it and I think it's better to continue with incremental releases. At the moment I need to merge this in order to develop a performance improvement with a subsequent PR

PR #14285, opened 2023-01-05
It is okay to return a result in some cases when median == -1

aqk:
Fixes an issue where we were too conservative in returning fee estimator results.


emlowe (dismissed):

wallentx:
@aqk All of the Fee Estimation tests seem to be failing

wallentx (changes requested):
@aqk All of the Fee Estimation tests seem to be failing

wallentx (dismissed):

emlowe (approved):

PR #14287, opened 2023-01-05
wallet: Drop duplicate puzzle hashes assert

xdustinface:
While #14187 fixes duplicated entries in new DBs, existing DBs still  contain them and will trigger this assert so lets remove it again.


wjblanke (approved):
aok

PR #14288, opened 2023-01-05
Fix computation of moving averages

aqk:
Fix computation of moving averages in Fee Estimator


emlowe (approved):

arvidn:
@aqk I think you should add a test that would have caught this bug too

aqk:
> @aqk I think you should add a test that would have caught this bug too

Yes, will do.

PR #14294, opened 2023-01-06
Test get_bucket_index

aqk:
Test boundary conditions on fee bucket calculation.

### DIFF
--- a/chia/full_node/fee_estimator_constants.py
+++ b/chia/full_node/fee_estimator_constants.py
@@ -2,9 +2,9 @@
 from __future__ import annotations
 
 MIN_FEE_RATE = 0  # Value of first bucket
-INITIAL_STEP = 100  # First bucket after zero value
-MAX_FEE_RATE = 40000000  # Mojo per 1000 cost unit
-INFINITE_FEE_RATE = 1000000000
+INITIAL_STEP = 5.0  # First bucket after zero value
+MAX_FEE_RATE = 40000000.0  # Mojo per 1000 cost unit
+INFINITE_FEE_RATE = 1000000000.0
 
 STEP_SIZE = 1.05  # bucket increase by 1.05
 
--- a/chia/full_node/fee_tracker.py
+++ b/chia/full_node/fee_tracker.py
@@ -1,10 +1,9 @@
 from __future__ import annotations
 
 import logging
+from bisect import bisect_left
 from dataclasses import dataclass
 from typing import List, Optional, Tuple
-
-from sortedcontainers import SortedDict
 
 from chia.full_node.fee_estimate_store import FeeStore
 from chia.full_node.fee_estimator_constants import (
@@ -61,32 +60,10 @@ def get_estimate_time_intervals() -> List[uint64]:
     return [uint64(blocks * SECONDS_PER_BLOCK) for blocks in get_estimate_block_intervals()]
 
 
-def get_bucket_index(sorted_buckets: SortedDict, fee_rate: float) -> int:
-    if len(sorted_buckets) < 1:
-        raise RuntimeError("get_bucket_index: sorted_buckets is invalid")
-
-    if fee_rate in sorted_buckets:
-        bucket_index = sorted_buckets[fee_rate]
-    else:
-        # Choose the bucket to the left if we do not have exactly this fee rate
-        # Python's list.bisect_left returns the index to insert a new element into a sorted list
-        bucket_index = sorted_buckets.bisect_left(fee_rate) - 1
-
-    if bucket_index < 0:
-        return 0
-
-    if bucket_index >= len(sorted_buckets):
-        # The fee rate is greater than our max fee rate - clamp to the highest rate
-        return len(sorted_buckets) - 1
-
-    return int(bucket_index)
-
-
 # Implementation of bitcoin core fee estimation algorithm
 # https://gist.github.com/morcos/d3637f015bc4e607e1fd10d8351e9f41
 class FeeStat:  # TxConfirmStats
-    buckets: List[float]
-    sorted_buckets: SortedDict  # key is upper bound of bucket, val is index in buckets
+    buckets: List[float]  # These elements represent the upper-bound of the range for the bucket
 
     # For each bucket xL
     # Count the total number of txs in each bucket
@@ -122,7 +99,6 @@ class FeeStat:  # TxConfirmStats
     def __init__(
         self,
         buckets: List[float],
-        sorted_buckets: SortedDict,
         max_periods: int,
         decay: float,
         scale: int,
@@ -130,7 +106,6 @@ class FeeStat:  # TxConfirmStats
         my_type: str,
     ):
         self.buckets = buckets
-        self.sorted_buckets = sorted_buckets
         self.confirmed_average = [[] for _ in range(0, max_periods)]
         self.failed_average = [[] for _ in range(0, max_periods)]
         self.decay = decay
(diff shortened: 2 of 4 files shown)


arvidn (commented):
thinking some more about this; it seems questionable to use a float as a key in a dictionary.
you call the dictionary "buckets", but you don't really bucketize the fee rates.

At first I was thinking that (probably) the correct solution here us to run both `bisect_left()` and `bisect_right()` and then see which key is closest to the one passed in. And maybe that's preferred, but it also seems like you could quantize the fee rates and actually use these as buckets.

The logic now looks more of a patchwork of checks and fix-ups just to get it to work, rather than having a clear idea of what the semantics aught to be

aqk:
> thinking some more about this; it seems questionable to use a float as a key in a dictionary.

I 100% agree. If I was writing this from scratch, I wouldn't make that choice.

When porting code, I like to keep it close to the original, and do fixes and code cleanups separately, so it's easier to find bugs or behaviour differences.

> you call the dictionary "buckets", but you don't really bucketize the fee rates.

Hmm, I don't agree. If you take a peek at https://github.com/bitcoin/bitcoin/blob/master/src/policy/fees.cpp#L76
The scheme they are using is to use `buckets[n]` to represent the upper bound value of that bucket, and `bucketMap` as an extra indirection to help them bucketize.

> At first I was thinking that (probably) the correct solution here us to run both `bisect_left()` and `bisect_right()` and then see which key is closest to the one passed in. And maybe that's preferred, but it also seems like you could quantize the fee rates and actually use these as buckets.

Yeah, any stable grouping would work. Using their scheme has one nice advantage: you can make the buckets vary in size.
So you could have e.g. logarithmic buckets.

> The logic now looks more of a patchwork of checks and fix-ups just to get it to work, rather than having a clear idea of what the semantics aught to be

Well, yes. Right now in `get_bucket_index`, I'm just trying to match their semantics exactly, specifically `map::lower_bound`
https://cplusplus.com/reference/map/map/lower_bound/

After it's working, we can go in and improve the code. Also, since I put FeeEstimatorInterface in before porting their estimator, we can replace it with a better or easier to understand one if we want.

chia/full_node/fee_tracker.py
```diff
@@ -400,6 +411,25 @@ def estimate_median_val(
...
+        sorted_buckets[fee_rate] = index
+        if fee_rate == 0:
+            fee_rate = INITIAL_STEP
+        else:
+            fee_rate = fee_rate * STEP_SIZE
+        index += 1
+    buckets.append(INFINITE_FEE_RATE)
+    sorted_buckets[INFINITE_FEE_RATE] = index
+    assert len(sorted_buckets.keys()) == len(buckets)
+    return buckets, sorted_buckets
```
arvidn, line 430: I see how this is done now. This is where we define the bucket sizes. you don't create entries at arbitrary fee rates. This makes more sense to me now.

However, since you use a `SortedDict` (which is essentially a `std::map` or AVL-tree), it doesn't look like you need the separate `List[float]`. i.e. `buckets` is always the same as `sorted_buckets.keys()`, isn't it?

chia/full_node/fee_tracker.py
```diff
@@ -62,12 +62,23 @@ def get_estimate_time_intervals() -> List[uint64]:
 
 
 def get_bucket_index(sorted_buckets: SortedDict, fee_rate: float) -> int:
+    if len(sorted_buckets) < 1:
+        raise RuntimeError("get_bucket_index: sorted_buckets is invalid")
+
     if fee_rate in sorted_buckets:
         bucket_index = sorted_buckets[fee_rate]
```
arvidn, line 69: I still think you should remove this case, and always exercise the `bisec_left()` case. Is there a good reason to keep this? I fear it may obscure issues.

chia/full_node/fee_tracker.py
```diff
@@ -62,12 +62,23 @@ def get_estimate_time_intervals() -> List[uint64]:
...
     else:
         # Choose the bucket to the left if we do not have exactly this fee rate
+        # Python's list.bisect_left returns the index to insert a new element into a sorted list
         bucket_index = sorted_buckets.bisect_left(fee_rate) - 1
```
arvidn, line 73: let's assume that the fee rate computed by a fee divided by cost is never exactly on the boundary of buckets (that seems like a fairly safe assumption, is it?). Any fee rate lower than `INITIAL_STEP` will return 0 here. i.e. the first bucket. Otherwise you get the bucket whose range you fall into. e.g. if the fee is between `INITIAL_STEP`and `INITIAL_STEP` + `STEP_SIZE`, you get bucket 1.

The edge case is if the fee rate is greater than the largest fee you can imagine (i.e. `INFINITE_FEE_RATE`), then you get back an index that doesn't exist, one past the end. (I believe that's what `bisect_left()` returns).

It seem to me the `-1` is suspect, and the check to see if `bucket_index` < 0 is suspect. The check to see if you're one past the end is unavoidable I think. This all assumes fee rates can't be negative of course, so maybe there should be an assert for that.

My assumption is that the fee "key" defines the upper bound of fees that belong in that bucket. But perhaps the intention is that the keys defined the lower bound, and that's why you subtract 1 from the bucket index. I would think a more intuitive way to do this would be to actually create the bottom bucket using the `MIN_FEE_RATE` constant (which is 0). I notice that this constant isn't used anywhere, suggesting that perhaps that was the intention.

tests/fee_estimation/test_fee_estimation_unit_tests.py
```diff
@@ -145,3 +148,36 @@ def test_fee_estimation_inception() -> None:
...
+        result_index = get_bucket_index(sorted_buckets, rate)
+        print(rate, expected_index, result_index)
+        assert expected_index == result_index
+
+
+def test_get_bucket_index() -> None:
+    sorted_buckets = SortedDict({1.0: 0, 2.0: 1})
+    print()
+    print(sorted_buckets)
+    for rate, expected_index in ((0.5, 0), (1.0, 0), (1.5, 0), (2.0, 1), (2.1, 1)):
```
arvidn, line 180: I think it's risky to have the tests use fee rates that fall exactly on the bucket boundaries. It may make them flaky since floating point comparisons don't (typically) have completely portable behavior.

arvidn (commented):
I understand this code a lot better now, and I feel better about it. It looks like you could simplify it by removing the list of sorted bucket fee rate boundaries though, since the sorted dict already includes that.

arvidn (dismissed):

chia/full_node/fee_estimator_constants.py
```diff
@@ -2,9 +2,9 @@
 from __future__ import annotations
 
 MIN_FEE_RATE = 0  # Value of first bucket
-INITIAL_STEP = 100  # First bucket after zero value
-MAX_FEE_RATE = 40000000  # Mojo per 1000 cost unit
-INFINITE_FEE_RATE = 1000000000
+INITIAL_STEP = 5.0  # First bucket after zero value
```
aqk, line 5: These are just int float type changes.
`INITIAL_STEP = 5` is in main

chia/full_node/fee_tracker.py
aqk: That's correct! They don't need both. I went ahead and simplified their code - no reason to have a duplicate array.

If you look in `init_buckets`, you'll see that we do have a non-linear series making up the bucket high-water marks.

chia/full_node/fee_tracker.py
aqk: I took your suggestion here. I believe the code is okay in both cases. All the `equals` case gives us is the property that the bucket upper bound is inclusive. One float epsilon won't make a difference in such large buckets, for sure!

As you point out, it is risky in general to use `float equals`, but this is not one of those cases - failure of an expected `float equals here` will always run the bisect case.

arvidn: I agree with that. I don't think the risk is that the case where the float equals the boundary is a bug, I think having two separate ways of doing the same thing adds extra test burden with no upside (afaict).

chia/full_node/fee_tracker.py
aqk: True, the `bucket_index < 0` check is only there because of the `-1`

arvidn (approved):

PR #14297, opened 2023-01-06
optimize mempool's potential cache

arvidn:
split potential cache into conflict- and pending cache.
The pending cache can be ordered by height so we only have to consider transactions whose assert-height would make them valid with the new peak. This improves big-O complexity and hence allows the pending cache to have a greater capacity.

That's the rationale to separate the conflict case from the assert-height case.

The existing `PendingTxCache` is preserved but renamed `ConflictTxCache`. The new `PendingTxCache` orders the `MempoolItem`s by assert height, and its `drain()` function only trains up to the peak.

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -20,7 +20,7 @@ from chia.full_node.fee_estimation import FeeBlockInfo, MempoolInfo
 from chia.full_node.fee_estimator_interface import FeeEstimatorInterface
 from chia.full_node.mempool import Mempool, MempoolRemoveReason
 from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions, mempool_check_time_locks
-from chia.full_node.pending_tx_cache import PendingTxCache
+from chia.full_node.pending_tx_cache import ConflictTxCache, PendingTxCache
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.sized_bytes import bytes32, bytes48
 from chia.types.clvm_cost import CLVMCost
@@ -29,6 +29,7 @@ from chia.types.fee_rate import FeeRate
 from chia.types.mempool_inclusion_status import MempoolInclusionStatus
 from chia.types.mempool_item import MempoolItem
 from chia.types.spend_bundle import SpendBundle
+from chia.types.spend_bundle_conditions import SpendBundleConditions
 from chia.util import cached_bls
 from chia.util.cached_bls import LOCAL_CACHE
 from chia.util.condition_tools import pkm_pairs
@@ -81,6 +82,27 @@ def validate_clvm_and_signature(
     return None, bytes(result), new_cache_entries
 
 
+def compute_assert_height(
+    removal_coin_records: Dict[bytes32, CoinRecord],
+    conds: SpendBundleConditions,
+) -> uint32:
+    """
+    Computes the most restrictive height assertion in the spend bundle. Relative
+    height assertions are resolved using the confirmed heights from the coin
+    records.
+    """
+
+    height: uint32 = uint32(conds.height_absolute)
+
+    for spend in conds.spends:
+        if spend.height_relative is None:
+            continue
+        h = uint32(removal_coin_records[bytes32(spend.coin_id)].confirmed_block_index + spend.height_relative)
+        height = max(height, h)
+
+    return height
+
+
 class MempoolManager:
     pool: Executor
     constants: ConsensusConstants
@@ -88,7 +110,10 @@ class MempoolManager:
     get_coin_record: Callable[[bytes32], Awaitable[Optional[CoinRecord]]]
     nonzero_fee_minimum_fpc: int
     mempool_max_total_cost: int
-    potential_cache: PendingTxCache
+    # a cache of MempoolItems that conflict with existing items in the pool
+    _conflict_cache: ConflictTxCache
+    # cache of MempoolItems with height conditions making them not valid yet
+    _pending_cache: PendingTxCache
     seen_cache_size: int
     peak: Optional[BlockRecord]
     mempool: Mempool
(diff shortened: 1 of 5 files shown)


chia/full_node/mempool_manager.py
```diff
@@ -81,14 +82,38 @@ def validate_clvm_and_signature(
...
+    height assertions are resolved using the confirmed heights from the coin
+    records.
+    """
+
+    height: uint32 = uint32(conds.height_absolute)
+
+    for spend in conds.spends:
+        if spend.height_relative is None:
+            continue
+        h = uint32(removal_coin_records[bytes32(spend.coin_id)].confirmed_block_index + spend.height_relative)
```
Quexington, line 100: I'm not positive but should this have a special case for 0?  I seem to recall that `(ASSERT_HEIGHT_RELATIVE 0)` would result in being unable to spend the coin ephemerally basically acting as a `(ASSERT_HEIGHT_RELATIVE 1)`.  Am I mistaken in recalling this?

Quexington: Or is it actually +1 because the timestamp is based off of the previous block maybe?

chia/full_node/pending_tx_cache.py
```diff
@@ -40,3 +47,68 @@ def drain(self) -> Dict[bytes32, MempoolItem]:
...
+        if name in self._txs:
+            return None
+
+        self._txs[name] = item
+        self._cache_cost += item.cost
+        self._by_height.setdefault(item.assert_height, {})[name] = item
+
+        while self._cache_cost > self._cache_max_total_cost or len(self._txs) > self._cache_max_size:
+
+            # we start removing items with the highest assert_height first
```
Quexington, line 81: It may not be this simple, but is this what a perfectly greedy farmer would do?  Wouldn't they keep the ones with the higher fee around?  I don't think that the probability of them winning a block now versus later is any different, right?

chia/full_node/mempool_manager.py
arvidn: your recollection is correct. the special case is the check on line 98. a relative height of 0 means the coin has to be spent in at least the block after it was confirmed. the `assert_height` being computed here has the same semantics as the `ASSERT_HEIGHT_ABSOLUTE`, meaning the block height has to be *greater than* the specified one for the spend to be valid. i.e. the same block is still too soon.

chia/full_node/pending_tx_cache.py
arvidn: you're right. This is likely not the optimal eviction policy. I don't think its worse than what we have now though (which is FIFO). I imagine we'll revisit this once we make the mempool more sophisticated. The balance is that there's a cost to keeping these transactions in memory, and transactions that won't be valid for a very long time are more expensive to keep in RAM, because they will use RAM for longer.

But the fee-per-cost probably should be taken into account for a better strategy.

Quexington (dismissed):
One item I noted is that the eviction process for conflicting / expiring spends is likely not what a perfectly greedy farmer would do (although it is optimal in terms of hardware usage so it's complicated).  It's better than what we have now though and everything else looks good to me.

chia/full_node/mempool_manager.py
```diff
@@ -550,8 +588,11 @@ def get_mempool_item(self, bundle_hash: bytes32, include_pending: bool = False)
         """
         item = self.mempool.spends.get(bundle_hash, None)
         if not item and include_pending:
-            # no async lock needed since we're not mutating the potential_cache
-            item = self.potential_cache._txs.get(bundle_hash, None)
+            # no async lock needed since we're not mutating the pending_cache
+            item = self._pending_cache.get(bundle_hash)
+        if not item and include_pending:
```
aqk, line 593: May as well merge this condition with the one above:

`if not item and include_pending:`

aqk: Oh, are these caches meant to be overlapping, or not? Better comment on that somewhere.

chia/full_node/mempool_manager.py
```diff
@@ -351,9 +377,15 @@ async def add_spend_bundle(
...
+            # The transaction has a conflict with another item in the
+            # mempool, put it aside and re-try it later
+            self._conflict_cache.add(item)
+            return item.cost, MempoolInclusionStatus.PENDING, err
         elif item is not None:
-            # There is an error,  but we still returned a mempool item, this means we should add to the pending pool.
-            self.potential_cache.add(item)
+            # The transasction has a height assertion and is not yet valid.
+            # remember it to try it again later
+            self._pending_cache.add(item)
```
aqk, line 388: Can an item enter the `_pending_cache` for a reason other than a height assertion?

chia/full_node/mempool_manager.py
aqk: Oh, I see. we first try pending, then conflicting.

arvidn: both `_conflict_cache` and `_pending_cache` are considered "pending" for the purposes of this API call. an item should only ever be in one or the other (or neither)

chia/full_node/mempool_manager.py
arvidn: it's not supposed to be possible, and I don't see how it would be possible. we only add it to one or the other. and the only way an item is coming back here is if it was first drained from one of the caches.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -70,6 +71,32 @@ async def instantiate_mempool_manager(
...
+    # 42 is the absolute height condition
+    conds = SpendBundleConditions([Spend(coin_id, bytes32(b"c" * 32), None, 0, [], [], 0)], 0, 42, 0, [], 0)
+    assert compute_assert_height(coin_records, conds) == 42
+
+    # 1 is a relative height, but that only amounts to 13, so the absolute
+    # height is more restrictive
+    conds = SpendBundleConditions([Spend(coin_id, bytes32(b"c" * 32), 1, 0, [], [], 0)], 0, 42, 0, [], 0)
+    assert compute_assert_height(coin_records, conds) == 42
+
+    # 100 is a relative height, and sinec the coin was confirmed at height 12,
```
aqk, line 90:
```suggestion
    # 100 is a relative height, and since the coin was confirmed at height 12,
```

aqk (dismissed):

arvidn:
I had to rebase because of a conflict in the test

Quexington (approved):
Re-stamping my approval

aqk (approved):

trepca (approved):

PR #14301, opened 2023-01-06
add dust warning message to chia coins commands & cleanup code

### LINKED ISSUE #14330: Coin combine command issue when used with CATs and a fee[Bug]
### What happened?

When using the `chia wallet coins combine` command to combine CATs, if I include a fee, two new coins get created (the command should only output one coin). When using the same command on CATs without a fee, or with XCH (regardless of fee), the command works correctly.

To reproduce:

1. List coins in a CAT wallet (in this case, it's a Spacebucks wallet and there is one coin)
```
chia wallet coins list -f 3792481086 -i 3

There are a total of 1 coins in wallet 3.
1 confirmed coins.
0 unconfirmed additions.
0 unconfirmed removals.
Confirmed coins:
Coin ID: 0x1f110e4a843922163dc02d200cf0d89600d2f1f7405d21f38017b640710f3cb2
        Address: xch12tevpt8em3acyydu5k662mms7sk446857mmjpsuzzlw27phlns9ql9jsvk Amount: 9446.259000000000  (9446259 mojo), Confirmed in block: 3088835
```

2. Split the CAT into multiple coins (I create 100 here, but I don't think the number matters), optionally include a fee (the `split` command works regardless of whether there's a fee)
```
chia wallet coins split -f 3792481086 -i 3 --number-of-coins 100 --amount-per-coin 10 --target-coin-id 

0x1f110e4a843922163dc02d200cf0d89600d2f1f7405d21f38017b640710f3cb2 --fee 0.000000001
Transaction sent: d469fc86612fc234a581eab004d165433c0df914f84a539843ab46012b1dfc27
To get status, use command: chia wallet get_transaction -f 3792481086 -tx 0xd469fc86612fc234a581eab004d165433c0df914f84a539843ab46012b1dfc27
```

3. After the transaction has completed, list the coins again. (In this case, there are 101 coins, as expected)
```
chia wallet coins list -f 3792481086 -i 3

There are a total of 101 coins in wallet 3.
101 confirmed coins.
0 unconfirmed additions.
0 unconfirmed removals.
Confirmed coins:
Coin ID: 0x5fcd6a69f11cd9bdd111f1a54336adc8fd1278144c3b455f05253b2c74df4e6f
        Address: xch1rvz0k6g0tayz0kzcyuevacp3gq0eqcmh7pd00ds25jhrzvqlz76qp7cdu9 Amount: 10.000000000000  (10000 mojo), Confirmed in block: 3088839
...
...
```

4. Call the `combine` command **and include a fee**. (In this case, all 101 coins should be combined.)
```
chia wallet coins combine -f 3792481086 -i 3 --fee 0.000000001

Combining 101 coins.
Would you like to Continue? (y/n): y
Transaction sent: 55d6916f96d2348ef5766953e38ca7bb8cea5c2c3823e32966663ac4604028bd
To get status, use command: chia wallet get_transaction -f 3792481086 -tx 0x55d6916f96d2348ef5766953e38ca7bb8cea5c2c3823e32966663ac4604028bd 
```

5. List the coins again. There should be only one coin, but **there are two**. Note that neither coin existed before.
```
chia wallet coins list -f 3792481086 -i 3

There are a total of 2 coins in wallet 3.
2 confirmed coins.
0 unconfirmed additions.
0 unconfirmed removals.
Confirmed coins:
Coin ID: 0xcb47a1a9f30aa0707a359a01444563d81b52dacdb01ace76b4d27842852ed0db
        Address: xch1hej4aewgmjph4nq034ykl950re3dxqh6xwjj40jvsdsve0xqc60slcm9xu Amount: 1.000000000000  (1000 mojo), Confirmed in block: 3088847

Coin ID: 0x6743596e1e1332d2a237e1eff3abce1200c0f8f44554a82bc5228b3f786c07d5
        Address: xch1e3gyc7xky86lthe7z9f8zrsm3jng3t9m898vsqmahrq8t0u3hnus3ey9jn Amount: 9445.259000000000  (9445259 mojo), Confirmed in block: 3088854
```

6. Run the same `combine` command again, still without a fee. The same extra coin still exists, and it's still unspent.
```
chia wallet coins combine -f 3792481086 -i 3 --fee 0.000000001

Combining 2 coins.
Would you like to Continue? (y/n): y
Transaction sent: 225e21f80cf29d375990f5b726ad727a7091c9640ce49a2f6df4d72ae0fb0482
To get status, use command: chia wallet get_transaction -f 3792481086 -tx 0x225e21f80cf29d375990f5b726ad727a7091c9640ce49a2f6df4d72ae0fb0482
```
After the transaction has completed...
```
chia wallet coins list -f 3792481086 -i 3

There are a total of 2 coins in wallet 3.
2 confirmed coins.
0 unconfirmed additions.
0 unconfirmed removals.
Confirmed coins:
Coin ID: 0xcb47a1a9f30aa0707a359a01444563d81b52dacdb01ace76b4d27842852ed0db
        Address: xch1hej4aewgmjph4nq034ykl950re3dxqh6xwjj40jvsdsve0xqc60slcm9xu Amount: 1.000000000000  (1000 mojo), Confirmed in block: 3088847

Coin ID: 0x6743596e1e1332d2a237e1eff3abce1200c0f8f44554a82bc5228b3f786c07d5
        Address: xch1e3gyc7xky86lthe7z9f8zrsm3jng3t9m898vsqmahrq8t0u3hnus3ey9jn Amount: 9445.259000000000  (9445259 mojo), Confirmed in block: 3088854
```

7. Attempt to combine the coins again, but this time don't include a fee.
```
chia wallet coins combine -f 3792481086 -i 3

Combining 2 coins.
Would you like to Continue? (y/n): y
Transaction sent: 67be9a7f1f028cc03b8b9207f3c2dfbfeb3fa8aaa6c707f9c7c7d2734d941bb1
To get status, use command: chia wallet get_transaction -f 3792481086 -tx 0x67be9a7f1f028cc03b8b9207f3c2dfbfeb3fa8aaa6c707f9c7c7d2734d941bb1
```

8. List the coins again. This time there is only one coin, which is correct.
```
chia wallet coins list -f 3792481086 -i 3

There are a total of 1 coins in wallet 3.
1 confirmed coins.
0 unconfirmed additions.
0 unconfirmed removals.
Confirmed coins:
Coin ID: 0x2a8ce0dfe502ee143691f0ab5208c38d5c34f2100a9d75337c71ffbd1b2bcef5
        Address: xch18rpn6v0xg9avkqz8nxtwufrnxn3pw9zy0q77k0afg9q8ex8cuh8qj7tcpg Amount: 9446.259000000000  (9446259 mojo), Confirmed in block: 3088886
```

A few notes:
- The same problem doesn't exist when combining XCH and including a fee. In this case, all coins are correctly combined.
- When combining CATs, it doesn't seem to matter how many coins I start with or what their value is. I always end up with two coins when I specify a fee, and one coin when I don't.
- I have verified this with Spacebucks, but not with other CATs.

### Version

jn.add-dust-warning branch (1.7.0b5.dev51)

Windows

CLI

### Relevant log output

_No response_

jack60612:
Purpose:
This adds a warning to the user letting them know that the dust filter might prevent them from seeing their coins.

Current Behavior:
No warning when making coins below the dust limit, user might be confused as wallet ignores coins.
New Behavior:
A warning was added to stop possible user confusion.

Testing Notes:
CLI Testing will be added in the coming weeks.

### DIFF
--- a/chia/cmds/coin_funcs.py
+++ b/chia/cmds/coin_funcs.py
@@ -252,3 +252,12 @@ async def async_split(
         tx_id = transaction.name.hex()
         print(f"Transaction sent: {tx_id}")
         print(f"To get status, use command: chia wallet get_transaction -f {fingerprint} -tx 0x{tx_id}")
+        dust_threshold = config.get("xch_spam_amount", 1000000)  # min amount per coin in mojo
+        spam_filter_after_n_txs = config.get("spam_filter_after_n_txs", 200)  # how many txs to wait before filtering
+        if final_amount_per_coin < dust_threshold and wallet_type == WalletType.STANDARD_WALLET:
+            print(
+                f"WARNING: The amount per coin: {amount_per_coin} is less than the dust threshold: "
+                f"{dust_threshold / mojo_per_unit}. Some or all of the Coins "
+                f"{'will' if number_of_coins > spam_filter_after_n_txs else 'may'} not show up in your wallet unless "
+                f"you decrease the dust limit to below {final_amount_per_coin} mojos or disable it by setting it to 0."
+            )


chia/cmds/coin_funcs.py
```diff
@@ -217,3 +220,9 @@ async def async_split(args: Dict[str, Any], wallet_client: WalletRpcClient, fing
     tx_id = transaction.name.hex()
     print(f"Transaction sent: {tx_id}")
     print(f"To get status, use command: chia wallet get_transaction -f {fingerprint} -tx 0x{tx_id}")
+    if final_amount_per_coin < dust_threshold and wallet_type == WalletType.STANDARD_WALLET:
+        print(
+            f"WARNING: The amount per coin: {amount_per_coin} is less than the dust threshold: {dust_threshold / mojo_per_unit}."
+            f"Some or all of the Coins {'will' if number_of_coins > spam_filter_after_n_txs else 'may'} not show up in your wallet unless you "
+            f"decrease the dust limit to below {final_amount_per_coin} mojos or disable it by setting it to 0."
```
arvidn: It would probably be more helpful to actually name the configuration option here. `xch_spam_amount`

chia/cmds/coin_funcs.py
```diff
@@ -36,6 +27,15 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
...
         return
+    min_coin_amount = Decimal(args["min_coin_amount"])
+    max_coin_amount = Decimal(args["max_coin_amount"])
+    excluded_coin_ids = args["excluded_coin_ids"]
+    excluded_amounts = args["excluded_amounts"]
+    addr_prefix = args["addr_prefix"]
+    show_unconfirmed = args["show_unconfirmed"]
+    paginate = args["paginate"]
+    if paginate is None:
+        paginate = sys.stdout.isatty()
```
arvidn, line 40: what's the reason to move this until after checking the wallet type?
Is it to fail earlier if we have the wrong wallet ID?

chia/cmds/coin_funcs.py
```diff
@@ -27,23 +27,22 @@ async def async_list(args: Dict[str, Any], wallet_client: WalletRpcClient, finge
...
-    final_max_coin_amount: uint64 = uint64(int(max_coin_amount * mojo_per_unit))
-    final_excluded_amounts: List[uint64] = [uint64(int(Decimal(amount) * mojo_per_unit)) for amount in excluded_amounts]
     conf_coins, unconfirmed_removals, unconfirmed_additions = await wallet_client.get_spendable_coins(
         wallet_id=wallet_id,
-        max_coin_amount=final_max_coin_amount,
-        min_coin_amount=final_min_coin_amount,
-        excluded_amounts=final_excluded_amounts,
+        max_coin_amount=max_coin_amount,
+        min_coin_amount=min_coin_amount,
+        excluded_amounts=excluded_amounts,
```
arvidn, line 45: all of these are now specified in mojos, whereas they used to be specified in whatever CAT the wallet manages.

This is changing the interface, and behavior, of a public class. You can't update all 3rd party uses of this. I don't think this is safe. Am I missing something?

arvidn (dismissed):
I don't think the 3rd commit is safe

chia/cmds/coin_funcs.py
jack60612: yep, also looks cleaner

chia/cmds/coin_funcs.py
jack60612: > all of these are now specified in mojos, whereas they used to be specified in whatever CAT the wallet manages.
> 
> This is changing the interface, and behavior, of a public class. You can't update all 3rd party uses of this. I don't think this is safe. Am I missing something?

I'm a bit confused, the internals of a cli command should never have any 3rd party uses & all i did was simplify the variables

arvidn: perhaps I'm misreading this, but `wallet_client` really sounds like our client library that speaks the RPC protocol. The library that we would expect 3rd party developers to use to communicate with a wallet. Is it not?

jack60612: > perhaps I'm misreading this, but `wallet_client` really sounds like our client library that speaks the RPC protocol. The library that we would expect 3rd party developers to use to communicate with a wallet. Is it not?

yeah it is, however I changed the variables above, so final_excluded_amounts on main, is the same as excluded_amounts on this pr.

jack60612: all i did was change the internals, but what its giving to the rpc is the same thing.

chia/cmds/coin_funcs.py
```diff
@@ -131,7 +132,7 @@ async def async_combine(args: Dict[str, Any], wallet_client: WalletRpcClient, fi
     largest = bool(args["largest"])
     if target_coin_amount != 0:  # if we have a set target, just use standard coin selection.
         removals: List[Coin] = await wallet_client.select_coins(
-            amount=target_coin_amount + fee,
+            amount=target_coin_amount + fee if is_xch else target_coin_amount,
```
arvidn, line 135: this doesn't look right. you add `target_coin_amount` twice for non-standard wallet coins?

chia/cmds/coins.py
```diff
@@ -169,7 +169,7 @@ def combine_cmd(
         "min_coin_amount": min_amount,
         "excluded_amounts": amounts_to_exclude,
         "number_of_coins": number_of_coins,
-        "max_amount": max_amount,
+        "max_coin_amount": max_amount,
```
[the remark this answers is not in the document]
jack60612: I have had Dan test every variation and it does work

chia/cmds/coin_funcs.py
```diff
@@ -163,11 +164,11 @@ async def async_combine(args: Dict[str, Any], wallet_client: WalletRpcClient, fi
     if input("Would you like to Continue? (y/n): ") != "y":
         return
     total_amount: uint128 = uint128(sum(coin.amount for coin in removals))
-    if total_amount - fee <= 0:
+    if is_xch and total_amount - fee <= 0:
```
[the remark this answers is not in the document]
jack60612: If we are using cats then the cats obviously couldn't be used as a tx fee so we shouldn't worry about it

chia/cmds/coin_funcs.py
```diff
@@ -163,11 +164,11 @@ async def async_combine(args: Dict[str, Any], wallet_client: WalletRpcClient, fi
...
         print("Total amount is less than 0 after fee, exiting.")
         return
     target_ph: bytes32 = decode_puzzle_hash(await wallet_client.get_next_address(wallet_id, False))
-    additions = [{"amount": total_amount - fee, "puzzle_hash": target_ph}]
+    additions = [{"amount": total_amount - fee if is_xch else total_amount, "puzzle_hash": target_ph}]
```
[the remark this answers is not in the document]
jack60612: Meant to use parentheses

chia/cmds/coin_funcs.py
jack60612: > given that we don't have any test coverage for this, it seems risky to move things around like this. It's not trivial to manually inspect to ensure it's correct. (e.g. moved verbatim and no values mutate in the collapsed section of the patch that changes meanings of variables from one place to the other)

if you view commit by commit its really easy to see this change.

chia/cmds/coin_funcs.py
```diff
@@ -195,7 +197,9 @@ async def async_split(args: Dict[str, Any], wallet_client: WalletRpcClient, fing
     target_coin_id: bytes32 = bytes32.from_hexstr(args["target_coin_id"])
     dust_threshold = int(args["dust_threshold"])  # min amount per coin in mojo
     spam_filter_after_n_txs = int(args["spam_filter_after_n_txs"])
-    total_amount = (amount_per_coin * number_of_coins) + fee
+    total_amount = amount_per_coin * number_of_coins
```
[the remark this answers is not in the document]
jack60612: look at line 202

chia/cmds/coin_funcs.py
arvidn: what happened here? it still doesn't look right? did you change this in another commit?

chia/cmds/coin_funcs.py
arvidn: it's hard to review commit-by-commit because later amendments weren't squashed into the original commits

chia/rpc/wallet_rpc_api.py
```diff
@@ -1380,21 +1383,22 @@ async def cat_spend(self, request, hold_lock=True) -> EndpointResult:
             exclude_coin_amounts = [uint64(a) for a in exclude_coin_amounts]
         exclude_coin_ids: Optional[List] = request.get("exclude_coin_ids")
         if exclude_coin_ids is not None:
-            exclude_coins: Set[Coin] = {
+            exclude_coins: Optional[Set[Coin]] = {
```
arvidn, line 1386: you change this type, but I don't see you changing the signature for `generate_signed_transaction()` (which this variable is passed into). Do you understand why mypy doesn't issue a warning?

arvidn: I see, it took an optional all along

chia/rpc/wallet_rpc_api.py
```diff
@@ -1370,6 +1370,9 @@ async def cat_spend(self, request, hold_lock=True) -> EndpointResult:
             puzzle_hashes.append(decode_puzzle_hash(request["inner_address"]))
             if "memos" in request:
                 memos.append([mem.encode("utf-8") for mem in request["memos"]])
+        coins = None
+        if "coins" in request and len(request["coins"]) > 0:
+            coins = set([Coin.from_json_dict(coin_json) for coin_json in request["coins"]])
```
arvidn, line 1375: I take it this is one of the CLI bugs this PR is supposed to fix. is that right?
is `cat_spend()` an internal function of the `wallet_rpc_api.py`? If so, it makes me uneasy to pass the arguments as a `dict`. If not (and it's public), the types should be validated. e.g. `request["coins"]` could be a string, but you'd end up iterating over characters instead of json entries.

The fact that `request` is allowed to be untyped suggests that this whole file doesn't have mypy turned up to max yet either, which is another cause for concern (genrally)

arvidn (commented):
A lot of this code is not controversial, and I would be happy to approve it. But since this PR does so many things (dust filter, cleanup, fixing several bugs), it makes it harder to land the non-controversial parts.

Specifically, the smallest possible changes to fix the existing bugs, I believe would be simple enough to approve and land before we gave tests.

chia/rpc/wallet_rpc_api.py
jack60612: cat_spend is not an internal function, it is a usable rpc.

chia/cmds/coin_funcs.py
```diff
@@ -252,3 +252,12 @@ async def async_split(
...
         print(f"To get status, use command: chia wallet get_transaction -f {fingerprint} -tx 0x{tx_id}")
+        dust_threshold = config.get("xch_spam_amount", 1000000)  # min amount per coin in mojo
+        spam_filter_after_n_txs = config.get("spam_filter_after_n_txs", 200)  # how many txs to wait before filtering
+        if final_amount_per_coin < dust_threshold and wallet_type == WalletType.STANDARD_WALLET:
+            print(
+                f"WARNING: The amount per coin: {amount_per_coin} is less than the dust threshold: "
+                f"{dust_threshold / mojo_per_unit}. Some or all of the Coins "
+                f"{'will' if number_of_coins > spam_filter_after_n_txs else 'may'} not show up in your wallet unless "
+                f"you decrease the dust limit to below {final_amount_per_coin} mojos or disable it by setting it to 0."
+            )
```
arvidn, line 263: there's no test coverage for this block

jack60612: all the cli tests are going to follow my proof of concept pr.

arvidn (approved):

PR #14302, opened 2023-01-06
infrastructure for upcoming soft-fork

arvidn:
This patch restores and fixes up the soft-fork infrastructure from our previous soft-fork.

It primarily entails passing the block height to various relevant places, but mostly adding a block height test fixture to run some tests twice, once before and once after the soft-fork.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -132,6 +132,11 @@ def db_version(request):
     return request.param
 
 
+@pytest.fixture(scope="function", params=[1000000, 3630000])
+def softfork_height(request):
+    return request.param
+
+
 saved_blocks_version = "rc5"
 
 
--- a/tests/core/mempool/test_mempool.py
+++ b/tests/core/mempool/test_mempool.py
@@ -1859,6 +1859,7 @@ def generator_condition_tester(
     mempool_mode: bool = False,
     quote: bool = True,
     max_cost: int = MAX_BLOCK_COST_CLVM,
+    height: uint32,
 ) -> NPCResult:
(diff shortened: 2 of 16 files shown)


tests/core/mempool/test_mempool.py
```diff
@@ -2041,49 +2053,60 @@ def test_create_coin_different_parent(self):
...
+            (True, None),
+            (False, 2300000),
+            (False, 3630000),
+        ],
+    )
+    def test_unknown_condition(self, mempool: bool, height: uint32):
         for c in ['(2 100 "foo" "bar")', "(100)", "(4 1) (2 2) (3 3)", '("foobar")']:
-            npc_result = generator_condition_tester(c, mempool_mode=mempool)
+            npc_result = generator_condition_tester(c, mempool_mode=mempool, height=height)
             print(npc_result)
```
trepca, line 2110: this should be removed

tests/conftest.py
```diff
@@ -132,6 +132,11 @@ def db_version(request):
     return request.param
 
 
+@pytest.fixture(scope="function", params=[1000000, 3630000])
```
trepca, line 135: shouldn't second param be a constant from `ConsensusConstants`?

arvidn: it could be, but that would add a dependency on the default_constants module here, which I don't think is worth it. my understanding is that all tests would pay for that

tests/core/mempool/test_mempool.py
arvidn: this can be pretty handy if the test fails. pytest doesn't show stdout by default

richardkiss (approved):
With the understanding that this is a temporary hack that will be removed once it takes effect, approved.

One idea would be to add some comments with this information.

emlowe (approved):

PR #14306, opened 2023-01-07
Fixed condition to check for wallet_id

steppsr:
The second condition for the OR statement was the same as the first. It should be verifying that a wallet_id was passed in.

### Purpose:
The nft bulk set did is looping through each coin_id/nft_id. The first action is to check if both the coin/nft id AND a wallet id were passed in. The check was looking for coin/nft id twice and not check for a wallet id. This will update the second condition to ensure a wallet id is given.

### Current Behavior:
wallet_id **_is not_** checked to ensure it is passed into the RPC.

### New Behavior:
wallet_id **_is_** checked to ensure it is passed into the RPC.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -2215,7 +2215,7 @@ class WalletRpcApi:
         nft_ids = []
         fee = uint64(request.get("fee", 0))
         for nft_coin in request["nft_coin_list"]:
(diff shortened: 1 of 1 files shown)


trepca:
looks good, thanks. We should also update the unit test to cover for this case.

ytx1991 (approved):
lgtm

trepca (approved):

PR #14310, opened 2023-01-08
Enable setting time between blocks in full node sim

geoffwalmsley:
### Purpose:
Fix a small bug in the full node sim that was reseting the time_per_block parameter to 1. 

### Current Behavior:
When you set the time_per_block, it resets back to 1 so can't be used for testing cases where we want a certain time to pass between blocks

### New Behavior:
To set a custom number of seconds between sim blocks,  set:
`full_node_api.time_per_block = #seconds`

### Testing Notes:
Added a test in: tests/simulation/test_simulation.py::TestSimulation::test_simulation_farm_blocks


wjblanke (approved):
aok

PR #14327, opened 2023-01-10
Update version of clvm_tools_rs to 0.1.30

prozacchiwawa:
This bumps chia-blockchain to use clvm_tools_rs 0.1.30, which contains bug fixes from bugs detected by me and others.

### Purpose:
Memory leak fix
Fix for modern chialisp when parsing improper lists
Avoid a panic when compiling code that doesn't give enough arguments to an inline function.
Purpose:
Pulls in a version of the chialisp compiler with some bug fixes, which will lead to more freedom for chialisp devs.

Not a breaking change.

### Testing Notes:
See tests:
https://github.com/Chia-Network/clvm_tools_rs/blob/283f38af052fab4373920a29634eaef32186b791/src/tests/compiler/compiler.rs#L1138
https://github.com/Chia-Network/clvm_tools_rs/blob/283f38af052fab4373920a29634eaef32186b791/src/tests/compiler/compiler.rs#L1151

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ dependencies = [
     "clvm==0.9.7",
     "clvm_tools==0.4.6",  # Currying, Program.to, other conveniences
     "chia_rs==0.2.0",
-    "clvm-tools-rs==0.1.25",  # Rust implementation of clvm_tools' compiler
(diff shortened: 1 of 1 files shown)


trepca (approved):
Lgtm

PR #14339, opened 2023-01-13
Reorganize reusable singleton wallet code

aqk:
No change, no user visible change.

### DIFF
--- a/chia/wallet/did_wallet/did_wallet.py
+++ b/chia/wallet/did_wallet/did_wallet.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple
 
 from blspy import AugSchemeMPL, G1Element, G2Element
 
+import chia.wallet.singleton
 from chia.full_node.full_node_api import FullNodeAPI
 from chia.protocols import wallet_protocol
 from chia.protocols.wallet_protocol import CoinState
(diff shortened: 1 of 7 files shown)


aqk:
This reorg is to support sharing of code between Singleton wallets, in preparation for the DAO Wallet.

chia/wallet/did_wallet/did_wallet_puzzles.py
```diff
@@ -114,15 +90,6 @@ def is_did_innerpuz(inner_f: Program) -> bool:
     return inner_f == DID_INNERPUZ_MOD
 
 
-def is_did_core(inner_f: Program) -> bool:
```
matt-o-how, line 117: Glad to see the back of this function.
Once upon a time the singleton layer was called the did core, which is why this function is called that. 
Hopefully that's the last of references to did_core

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -10,6 +10,7 @@
 
 from blspy import AugSchemeMPL, G1Element, G2Element
 
+import chia.wallet.singleton
```
Quexington, line 13: I'm not a huge fan of this, can we either throw the other singleton functions that we need into the import below or import this `as singleton`?

Quexington (approved):
Stamping this with approval because it's an improvement and fine to merge, but I will note that it's not entirely complete as we also have chia.wallet.puzzles.singleton_top_layer_v1_1.py which is the same as this file but with more and different functions.  Moving those functions in here might be a good idea as well.

matt-o-how (approved):
Good changes.
Looks good to me.

chia/wallet/singleton.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import annotations
+
+from typing import Optional
+
+from chia.types.blockchain_format.program import Program
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.wallet.puzzles.load_clvm import load_clvm_maybe_recompile
+from chia.wallet.util.curry_and_treehash import calculate_hash_of_quoted_mod_hash, curry_and_treehash
+
+SINGLETON_TOP_LAYER_MOD = load_clvm_maybe_recompile("singleton_top_layer_v1_1.clvm")
```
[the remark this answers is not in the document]
aqk: Yes, I can move that here, too

chia/wallet/did_wallet/did_wallet.py
aqk: Oh, sure.

aqk:
> Stamping this with approval because it's an improvement and fine to merge, but I will note that it's not entirely complete as we also have chia.wallet.puzzles.singleton_top_layer_v1_1.py which is the same as this file but with more and different functions. Moving those functions in here might be a good idea as well.

Thanks, I'll open a new PR for that

emlowe (approved):

PR #14344, opened 2023-01-13
Limit rate of log messages when farmer is disconnected from pool

aqk:
Limit rate of log messages when farmer is disconnected from pool

### DIFF
--- a/chia/util/logging.py
+++ b/chia/util/logging.py
@@ -13,7 +13,7 @@ class TimedDuplicateFilter(logging.Filter):
 
     def __init__(self, regex_str: str, min_time_wait_secs: int, name: str = ""):
         super(TimedDuplicateFilter, self).__init__(name)
-        self.last_log_time = time.time()
+        self.last_log_time = time.monotonic()
         self.regex = re.compile(regex_str)
         self.min_time_wait_secs = min_time_wait_secs
 
@@ -25,7 +25,7 @@ class TimedDuplicateFilter(logging.Filter):
         msg = record.getMessage()
 
         if self.regex.match(msg):
-            now = time.time()
+            now = time.monotonic()
             if now - self.last_log_time > self.min_time_wait_secs:
                 self.last_log_time = now
                 return True
--- a/chia/util/logging.py
+++ b/chia/util/logging.py
@@ -1,0 +1,35 @@
+from __future__ import annotations
+
+import logging
+import re
+import time
+from typing import Pattern
+
+
+class TimedDuplicateFilter(logging.Filter):
+    last_log_time: float
+    regex: Pattern[str]
+    min_time_wait_secs: int
+
+    def __init__(self, regex_str: str, min_time_wait_secs: int, name: str = ""):
+        super(TimedDuplicateFilter, self).__init__(name)
+        self.last_log_time = time.time()
+        self.regex = re.compile(regex_str)
+        self.min_time_wait_secs = min_time_wait_secs
+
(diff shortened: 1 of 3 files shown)


chia/util/logging.py
```diff
@@ -0,0 +1,35 @@
...
+import time
+from typing import Pattern
+
+
+class TimedDuplicateFilter(logging.Filter):
+    last_log_time: float
+    regex: Pattern[str]
+    min_time_wait_secs: int
+
+    def __init__(self, regex_str: str, min_time_wait_secs: int, name: str = ""):
```
aqk, line 14: This forces a wait time of `min_time_wait_secs` between similar messages. Would this be more user-friendly as a rate (e.g. X msgs per minute) ?

aqk:
@arvidn This is a more general solution in response to your review on https://github.com/Chia-Network/chia-blockchain/pull/14278

chia/util/logging.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        _ = super(TimedDuplicateFilter, self).filter(record)
+        if not _:
+            return False
+
+        msg = record.getMessage()
+
+        if self.regex.match(msg):
+            now = time.time()
```
arvidn, line 28: it's minor, but if you'd use `time.monotonic()` instead, this would work as expected regardless of what the user (or ntpd) does with its system clock (the other places where you use `time.time()` would also need to be changed in that case)

tests/util/test_logging_filter.py
```diff
@@ -0,0 +1,40 @@
...
+
+import pytest
+
+from chia.util.logging import TimedDuplicateFilter
+
+
+def test_logging_filter(caplog: pytest.LogCaptureFixture) -> None:
+
+    log_interval_secs = 10
+    num_logs = 11
```
arvidn, line 15: I think it would make sense to use as small numbers as possible here, to make the test run as quickly as possible. e.g. set the limit to 1 second and sleep for 95ms.

tests/util/test_logging_filter.py
```diff
@@ -0,0 +1,40 @@
...
+
+    log = logging.getLogger()
+    log.addFilter(TimedDuplicateFilter("Filter this log message.*", log_interval_secs))
+
+    for n in range(num_logs):
+        with caplog.at_level(logging.WARNING):
+            log.warning(f"Filter this log message {n}")
+            sleep(sleep_secs)
+
+    assert len(split("\n", caplog.text)) <= ((num_logs * sleep_secs) / log_interval_secs) + 1
```
arvidn, line 26: I think this test is flaky. each iteration of the loop will add rounding error to the total time. you may want to check how much time has passed each loop iteration as well, and just sleep long enough to get to the next second. It would still be flaky because there are no guarantees, but probably less so

arvidn (dismissed):

chia/util/logging.py
aqk: Thanks - changed!

tests/util/test_logging_filter.py
aqk: Thanks, done.

tests/util/test_logging_filter.py
aqk: Good idea.

arvidn (approved):
it's a bit strange that the function called `filter()`, returning a `bool` filters the message when it returns `False`. as if the function really wanted to be called `allow()`.

PR #14345, opened 2023-01-13
Only the standard wallet generates derivation paths

Quexington:
### Purpose:
Currently, we will generate full puzzle hashes for each type of asset in the wallet.  I believe this logic is leftover from when we did not have hints.  With hints, every coin will hint to a standard XCH address and will be picked up that way.  Once it's been picked up, it gets sorted to the appropriate wallet.  This means we only need to derive puzzle hashes for innermost puzzles of which we only have one type right now: p2_delegated_or_hidden_puzzle.

### Testing Notes:
Make sure you still receive and sync all CATs, DIDs, DL singletons etc.  This was made in response to an error subscribing to too many puzzle hashes so maybe make a bunch of standard puzzle hashes and then add a bunch of wallets and see if you get that error again.

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -1059,7 +1059,7 @@ class DataLayerWallet:
     ##########
 
     def require_derivation_paths(self) -> bool:
-        return True
+        return False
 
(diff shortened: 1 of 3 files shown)


wjblanke (dismissed):
aok but really curious to see CI :-)

wjblanke (approved):
aok

PR #14356, opened 2023-01-17
Only subscribe to inner wallet puzzle hashes

Quexington:
(recreation of https://github.com/Chia-Network/chia-blockchain/pull/14345 for simplicity in fixing the immediate issue)

Dependent on:
- #15176
- #15179

### Purpose:

Currently, we will subscribe to full puzzle hashes for each type of asset in the wallet. I believe this logic is leftover from when we did not have hints. With hints, every coin will hint to a standard XCH address and will be picked up that way. Once it's been picked up, it gets sorted to the appropriate wallet. This means we only need to subscribe to puzzle hashes for innermost puzzles of which we only have one type right now: p2_delegated_or_hidden_puzzle.

### Testing Notes:

Make sure you still receive and sync all CATs, DIDs, DL singletons etc. This was made in response to an error subscribing to too many puzzle hashes so maybe make a bunch of standard puzzle hashes and then add a bunch of wallets and see if you get that error again.


chia/wallet/wallet_state_manager.py
```diff
@@ -384,9 +384,10 @@ async def create_more_puzzle_hashes(
                 self.log.info(f"Done: {creating_msg} Time: {time.time() - start_t} seconds")
             await self.puzzle_store.add_derivation_paths(derivation_paths)
             if len(derivation_paths) > 0:
-                await self.wallet_node.new_peak_queue.subscribe_to_puzzle_hashes(
-                    [record.puzzle_hash for record in derivation_paths]
-                )
+                if wallet_id == self.main_wallet.id():
+                    await self.wallet_node.new_peak_queue.subscribe_to_puzzle_hashes(
+                        [record.puzzle_hash for record in derivation_paths]
+                    )
```
Quexington, line 390: This is the main change, only subscribing to newly derived puzzle hashes for wallet ID 1.  Everything else is fixing the wallet to be reliant on the correct way to pick up subscriptions of non XCH coins rather than relying on the secondary derived puzzle hash.

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -208,6 +208,12 @@ async def get_or_create_wallet_for_cat(
 
         self.lineage_store = await CATLineageStore.create(self.wallet_state_manager.db_wrapper, self.get_asset_id())
         await self.wallet_state_manager.add_new_wallet(self, self.id())
+
+        for state in await self.wallet_state_manager.interested_store.get_unacknowledged_states_for_asset_id(
+            limitations_program_hash, delete_after=True
+        ):
+            await self.wallet_state_manager.retry_store.add_state(*state)
```
trepca, line 215: so this means we store coin states for CATs that we didn't accept, and once accepted we replay them to get them added to the wallet?

chia/wallet/wallet_interested_store.py
```diff
@@ -30,6 +31,10 @@ async def create(cls, wrapper: DBWrapper2):
             fields = "asset_id text PRIMARY KEY, name text, first_seen_height integer, sender_puzzle_hash text"
             await conn.execute(f"CREATE TABLE IF NOT EXISTS unacknowledged_asset_tokens({fields})")
 
+            # Table for coin states of unknown CATs
+            fields = "coin_state blob PRIMARY KEY, asset_id text, peer blob, fork_height int"
```
trepca, line 35: I'd add an index for `asset_id` since that's primary filter when querying 
and you can save ~50% of space if you make `asset_id` a blob

chia/wallet/wallet_state_manager.py
trepca: given this is global, I'm surprised it's not affecting DID wallet, maybe it's not using derived puzzles anymore

chia/wallet/cat_wallet/cat_wallet.py
Quexington: That's right

chia/wallet/wallet_state_manager.py
Quexington: The DID uses derived puzzles for its own internal spend logic, but picks everything up via hint like it should. Most of the problems I fixed in this PR were related to change, and since singletons don't make change, they were unaffected.

aqk: Does the DID wallet need to have correct state for its previously used derived puzzles in the wallet db to work correctly (after a fresh sync)?

chia/wallet/wallet_state_manager.py
```diff
@@ -707,6 +715,8 @@ async def handle_cat(
         parent_coin_state: CoinState,
         coin_state: CoinState,
         coin_spend: CoinSpend,
+        peer: WSChiaConnection,
```
xdustinface, line 718: Besides the fact that i think we probably shouldn't store the peer id, maybe just pass in `peer_id: bytes32` here.

xdustinface (changes requested):
This currently only does what you want to during puzzle hash creation i think? We still subscribe to all puzzle hashes stored in the puzzle store (which include all the ones you want to exclude with this PR) during normal sync operations: https://github.com/Chia-Network/chia-blockchain/blob/4e34ff79822350efb52b1050857d83edbbe71463/chia/wallet/wallet_node.py#L749-L759

Also left some more comments.

chia/wallet/wallet_state_manager.py
trepca: @aqk no, it should be able to resync itself, in the past versions we forgot to add a hint in a few cases and had some issues with it

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -368,7 +368,7 @@ async def test_create_signed_transaction(
     # Farm the transaction and make sure the wallet balance reflects it correct
     spend_bundle = tx.spend_bundle
     assert spend_bundle is not None
-    push_res = await full_node_rpc.push_tx(spend_bundle)
+    push_res = await wallet_1_rpc.push_transactions([tx])
```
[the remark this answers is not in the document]
Quexington: Yeah the wallet has this concept of subscribing to the additions and removals of transactions that it creates.  Previously, this test was pushing the bundle directly to the full node, denying the wallet the opportunity to do that.

chia/wallet/wallet_state_manager.py
Quexington: I like this because of how it matches the other similar functions like `handle_did`.  Passing in peer_id to one and a full peer into another seems more confusing than them all having the same API.

Quexington: This is now actually required to be a peer so this point is moot

xdustinface (changes requested):
So this PR seems to try to solve four independent things from my understanding, right? Like:

1. Only subscribe to inner wallet puzzle hashes which is what the PR should be about only from the title. 
2. Add some new table to make adding unacknowledged CATs more reliable.
3. Fix some place where we still forgot to hint some CAT output when a new CAT gets minted. Which i btw, also figured to be an issue over in #15274.
4. In offer code you added some extra coin_id subscriptions for non offer related coins and you added additions/removals to some `TransactionRecord`. 

I think it would be easier to understand and discuss whats happening and what belongs to what if you would just create separate PRs for 2-4. Still, some thoughts about them for now:

2. Wouldn't it be safer to just subscribe to the puzzle hashes again when adding a unacknowledged CAT since with this table its possible to attack (fill up the disk) for free (same goes for the retry store probably), i think? While its possible to dust a wallet too we at least have a dust filter now.. Maybe its not so critical, didn't do any math or think a lot about it yet but still i want to raise the concern :)
3. Seems reasonable but for me it feels like this is so far still a blocker for this PR and we need to handle this backwards compatible somehow like the to offer a way to retrospectively hint unhinted coins because now, while it's fixed by the changes in here for newly minted CATs there are still unhinted ones out in the wild which can not be synced/processed during initial sync because `determine_coin_type` will fail?
4. Im not so sure whats going on here and why it is required here or in general.

chia/wallet/trade_manager.py
```diff
@@ -391,6 +391,16 @@ async def cancel_pending_offers(
 
     async def save_trade(self, trade: TradeRecord, offer_name: bytes32) -> None:
         await self.trade_store.add_trade_record(trade, offer_name)
+        offer: Offer = Offer.from_bytes(trade.offer)
```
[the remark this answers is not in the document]
Quexington: This is kind of a bit of wallet magic that I'm not able to fully explain.  You can see here: https://github.com/Chia-Network/chia-blockchain/blob/4603b9d7688030f2cf300140c9592bd003647978/chia/wallet/wallet_state_manager.py#L1545-L1547 that we do this for regular transactions.  If you don't subscribe to the additions and removals, you start getting balance failures in the tests.  It's possible this is a hack, but it predates my involvement in this code so it's a hack that's been around a long time.  Regardless, not something I intend to diagnose and fix in this PR.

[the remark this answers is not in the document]
Quexington: I did make it so that the offer gets passed through to the save_trade function though

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -213,6 +213,22 @@ async def get_or_create_wallet_for_cat(
...
+
+        delete: bool = False
+        for state in await self.wallet_state_manager.interested_store.get_unacknowledged_states_for_asset_id(
+            limitations_program_hash
+        ):
+            new_peer = self.wallet_state_manager.wallet_node.get_full_node_peer()
+            if new_peer is not None:
+                delete = True
+                peer_id: bytes32 = new_peer.peer_node_id
+                await self.wallet_state_manager.retry_store.add_state(state[0], peer_id, state[1])
```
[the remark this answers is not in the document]
Quexington: 1) You might create the wallet directly instead of specifically adding it as an unacknowledged asset.  In this case, we still want to pick up the CATs that we've already synced but that we are just not showing to the user.  The core problem here is that CATs are special in that we sync them fully, we just at the end decide whether or not we want to keep them.  Previously, we would just forget them entirely, and when the wallet was added, we would resubscribe to a bunch of stuff and sync them again.  Now, we need some way to go back in time and resync the states that we skipped the first time.

2) I'm wary of directly calling the `add_coin_states` method outside of the normal process that calls it.  The wallet node has all of this careful task batching/cancelling and transactionality and whatnot that I don't fully understand, but basically the retry store is the easiest way to insert the coins to be processed in the normal order of things.  If someone makes changes to how coin states are processed in the future, I want to make sure it applies to these too and not create yet another entry point for coin states.

Quexington:
> So this PR seems to try to solve four independent things from my understanding, right?

Not really, we're changing one thing and then fixing all of the pieces of code that were relying on the old way of doing things.  This PR makes sense as a single PR because it's "one core change, and here's everything it affects".

> Wouldn't it be safer to just subscribe to the puzzle hashes again when adding a unacknowledged CAT since with this table its possible to attack (fill up the disk) for free (same goes for the retry store probably), i think? While its possible to dust a wallet too we at least have a dust filter now.. Maybe its not so critical, didn't do any math or think a lot about it yet but still i want to raise the concern :)

Perhaps, but it's a bit less of an issue than dust used to be.  We can't subscribe because we'll just end up with the same problem this PR is attempting to fix.  The dust filter doesn't stop us from syncing anything, it just stops us from putting info in the DB about it and leaving it available for coin selection and display to the user.  Obviously there's no coin selection or user display for these unacknowledged CATs so it shouldn't affect normal wallet operations, just a little bit of DB bloat.  The "attack" isn't really free either with blockchain fees taken into account.  The retry store is designed to only hold flaky states so it's not susceptible to this.

> Seems reasonable but for me it feels like this is so far still a blocker for this PR and we need to handle this backwards compatible somehow like the to offer a way to retrospectively hint unhinted coins because now, while it's fixed by the changes in here for newly minted CATs there are still unhinted ones out in the wild which can not be synced/processed during initial sync because determine_coin_type will fail

The function that's fixed here is a test function and is not a supported use case in the wild.  Unhinted CATs that are minted would already have failed on resync without the changes in this PR, the reason our tests weren't susceptible to that is that we don't ever resync in the tests.

chia/wallet/trade_manager.py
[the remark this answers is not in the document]
xdustinface: > that we do this for regular transactions. If you don't subscribe to the additions and removals, you start getting balance failures in the tests. It's possible this is a hack, but it predates my involvement in this code so it's a hack that's been around a long time. Regardless, not something I intend to diagnose and fix in this PR.

I think it's always important to understand why you add a change to the code, especially if maybe adds more subscriptions for no reason, and not just do something without understanding why you do it only to make tests happy, even if this requires diagnosing an issue or hack which predates your time. 

> I did make it so that the offer gets passed through to the save_trade function though

Thanks!

xdustinface:
> Not really, we're changing one thing and then fixing all of the pieces of code that were relying on the old way of doing things. This PR makes sense as a single PR because it's "one core change, and here's everything it affects".

Well thats the whole point of splitting up changes. To merge "everything it affects" independently because its not only an issue within this PR but also on current main even though this PR kind of uncovered some of the issues. Which then makes finding issues in the commit history easier if not a bunch of changes are squashed into one thing, gets certain stuff merged faster than others which might unblock other PRs, avoids discussions about different topics in PRs and what else i don't recall right now. Its always "one thing" you want to implement but still there are always other things which are affected and its good practice (not only in my mind) to get this other affected or refactored things merged in independently and upfront to avoid what we have here now.. mixed changes and discussions about other things in a PR which is supposed to be about "inner puzzle hash subscriptions". I tried many times already to communicate this to you and while this is not a big PR and its not super important its also not a big deal to just do it and start to follow good practices to get used to it. But i will resign here for now..
 
> We can't subscribe because we'll just end up with the same problem this PR is attempting to fix. 

Why? We anyway need to subscribe, even after this PR or does this PR magically let us keep our wallet up to date without any subscription? In fact we anyway currently downloading all coin states from height 0 and drop all below our synced height every time we go into a long sync which basically happens for each new untrusted connection and also for each trusted connection so i think we could do better than just adding yet another table which doesn't even take reorgs into account and might provide us reorged coins later which then at the end get pushed into another table from where it gets "retried" and after it gets pushed into the final table if we succeed to process the coins after all, if its still valid.

> The dust filter doesn't stop us from syncing anything, it just stops us from putting info in the DB

Yes but the unacknowledged CATs will still be put into the DB then with this change no matter what. 

> The function that's fixed here is a test function and is not a supported use case in the wild.

Isn't it used inside the `create_new_wallet` RPC command and with that exposed to whoever wants to use it?

Quexington:
> Why? We anyway need to subscribe, even after this PR or does this PR magically let us keep our wallet up to date without any subscription?

We're subscribing only to standard puzzle hashes, if we subscribe to all the cat puzzle hashes when a new unacknowledged CAT wallet is added, we end up with the multiplication factor (# of puzzles hashes x # of CAT wallets) that we're trying to get rid of.

Quexington:
> Isn't it used inside the create_new_wallet RPC command and with that exposed to whoever wants to use it?

We also have `/farm_block` which clearly isn't meant for general use.  Removing that function from the RPC and codebase in general is a piece of tech debt that we have yet to pay off.  The supported ways to mint CATs are well documented and do not refer people to this RPC.

xdustinface:
> We're subscribing only to standard puzzle hashes, if we subscribe to all the cat puzzle hashes when a new unacknowledged CAT wallet is added, we end up with the multiplication factor (# of puzzles hashes x # of CAT wallets) that we're trying to get rid of.

Why would we need to subscribe to all CAT puzzle hashes? Isn't that what hints are there for and what this PR is going to leverage for the CAT sync anyway or how is syncing "unackknowleged CATs" different?

Also, no thought about the reorg issue i mentioned above? And in general about this new "store" and the "retry store".. In my mind, this is rather something we should get rid of instead of giving it more attention because it certainly leads to coin states being processed out of order only if there was some "flake" somewhere in between and we depend on processing them in order so its basically something which inevitable leads to issues in certain cases.

Quexington:
> Why would we need to subscribe to all CAT puzzle hashes? Isn't that what hints are there for and what this PR is going to leverage for the CAT sync anyway or how is syncing "unackknowleged CATs" different?

I explained how they are different above.  We've already "synced" unacknowledged CATs.  The full node already gave us everything we subscribed to.  Adding a new CAT wallet doesn't trigger us to reprocess every state we've ever seen.  So if we don't store the ones we _might_ want to process, we'll never add the coin states at all.

> Also, no thought about the reorg issue i mentioned above? 

So many things to respond to I guess it got lost :)  It's a good catch.  I added a rollback to the interested store and made a separate PR for doing the same to the retry store: https://github.com/Chia-Network/chia-blockchain/pull/15303.

>  it certainly leads to coin states being processed out of order only if there was some "flake" somewhere in between and we depend on processing them in order so its basically something which inevitable leads to issues in certain cases

I've explained this on a separate PR, but the current wallet paradigm is supposed to assume that the order we get the coins in doesn't matter.  Any wallet that does depend on the order is bugged, and should be fixed so that is not the case.  We shouldn't compromise general flexibility for a few wallets that can't handle it, we should bring them in compliance with the standard.

xdustinface:
> I explained how they are different above. We've already "synced" unacknowledged CATs. The full node already gave us everything we subscribed to. Adding a new CAT wallet doesn't trigger us to reprocess every state we've ever seen. So if we don't store the ones we _might_ want to process, we'll never add the coin states at all.

Of course right now adding a new CAT wallet doesn't trigger us reprocessing every state thats why i brought up just subscribing again to all puzzle hashes (what we anyway already do permanently) and filter/process the ones we need once a unknown CAT gets acknowledged which would avoid us storing things we don't care about and not opening up the ability for someone to blow up the DB with unknown CATs which sounds risky to me. But seems like you don't share the same concerns or just don't want to change things again..

So, if we really go with this table you could maybe at least add unit tests for the new table's methods and ideally also a tests where you retrospectively acknowledge a CAT to at least make sure the things work as expected. 

>  I added a rollback to the interested store and made a separate PR for doing the same to the retry store: #15303.

Thanks!

> Any wallet that does depend on the order is bugged

Why is wallet is bugged if it expects to receive coins in order? Sounds actually reasonable to me. In general it's the reality a coin creation exists before a spend of the same coin so shouldn't we also expect and process it this way?

chia/wallet/trade_manager.py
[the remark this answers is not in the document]
Quexington: > I think it's always important to understand why you add a change to the code, especially if maybe adds more subscriptions for no reason, and not just do something without understanding why you do it only to make tests happy, even if this requires diagnosing an issue or hack which predates your time

Okay I dove into this, and the reason we do this is because basically _removals_ don't have hints.  The flow is as follows:
1) We receive an unspent coin because we have subscribed to its hint
2) We spend that coin and subscribe to its ID as we push the transaction (the behavior that was in question here)
3) We pick up the coin state for the coin in its unspent state via the ID that we just subscribed to

If we didn't subscribe to all of the removals, we would never pick up the coin state with a `spent_height is not None` and handle it accordingly.

Quexington:
> Why is wallet is bugged if it expects to receive coins in order? Sounds actually reasonable to me. In general it's the reality a coin creation exists before a spend of the same coin so shouldn't we also expect and process it this way?

It's a dependency that we don't need, it is entirely possible to sync every wallet with out of order coins, and since this is better and puts less pressure on the wallet protocol/node to get us everything in order, I think we should preserve that.

xdustinface (approved):
Its obviously covered somehow, but i think there is no tests which specifically tests this new "retrospectively add CAT states" feature, would have been nice to see this too :)

Quexington:
> Its obviously covered somehow, but i think there is no tests which specifically tests this new "retrospectively add CAT states" feature, would have been nice to see this too :)

https://github.com/Chia-Network/chia-blockchain/blob/9e8d6ebd71ac690a35b959d8f2e51bffb08b57d3/tests/wallet/cat_wallet/test_cat_wallet.py#L737 <- tests autodiscovery including after you have already received a CAT. This test failing is what prompted the need for the unacknowledged states table.

trepca (approved):

PR #14357, opened 2023-01-17
Add a wallet RPC for converting heights to timestamps

Quexington:
Title is pretty self explanatory

### DIFF
--- a/chia/rpc/wallet_rpc_client.py
+++ b/chia/rpc/wallet_rpc_client.py
@@ -90,6 +90,9 @@ class WalletRpcClient(RpcClient):
     async def farm_block(self, address: str) -> Dict[str, Any]:
         return await self.fetch("farm_block", {"address": address})
 
+    async def get_timestamp_for_height(self, height: uint32) -> uint64:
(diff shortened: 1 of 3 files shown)


chia/rpc/wallet_rpc_client.py
```diff
@@ -90,6 +90,9 @@ async def push_transactions(self, txs: List[TransactionRecord]):
     async def farm_block(self, address: str) -> Dict[str, Any]:
         return await self.fetch("farm_block", {"address": address})
 
+    async def get_timestamp_for_height(self, height: uint32) -> uint64:
+        return uint64((await self.fetch("get_timestamp_for_height", {"height": height}))["timestamp"])
```
trepca, line 94: it'd be good to handle errors for invalid blocks and return user friendly error

Quexington: From this call to the wallet node I'm pretty sure it has no way f knowing if it failed because that height doesn't exist or because the peers were just being flaky.  They will receive `{"error": "Error fetching timestamp from all peers"}` which I think is the best that we can do.

trepca (approved):

PR #14362, opened 2023-01-19
wallet: Drop `WalletNode._primary_peer_sync_task`, just await the sync

xdustinface:
### Purpose:

Simplify away the task member which isn't required in this case.

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -127,7 +127,6 @@ class WalletNode:
     _shut_down: bool = False
     _process_new_subscriptions_task: Optional[asyncio.Task] = None
     _retry_failed_states_task: Optional[asyncio.Task] = None
-    _primary_peer_sync_task: Optional[asyncio.Task] = None
     _secondary_peer_sync_task: Optional[asyncio.Task] = None
 
     @property
@@ -303,8 +302,6 @@ class WalletNode:
             self._process_new_subscriptions_task.cancel()
         if self._retry_failed_states_task is not None:
             self._retry_failed_states_task.cancel()
-        if self._primary_peer_sync_task is not None:
-            self._primary_peer_sync_task.cancel()
         if self._secondary_peer_sync_task is not None:
             self._secondary_peer_sync_task.cancel()
 
@@ -1041,11 +1038,7 @@ class WalletNode:
             if peer.peer_node_id not in self.synced_peers:
                 if new_peak_hb.height - current_height > self.LONG_SYNC_THRESHOLD:
                     self.wallet_state_manager.set_sync_mode(True)
-                self._primary_peer_sync_task = asyncio.create_task(
-                    self.long_sync(new_peak_hb.height, peer, uint32(max(0, current_height - 256)), rollback=True)
-                )
-                await self._primary_peer_sync_task
-                self._primary_peer_sync_task = None
+                await self.long_sync(new_peak_hb.height, peer, uint32(max(0, current_height - 256)), rollback=True)
                 self.wallet_state_manager.set_sync_mode(False)
 
     async def new_peak_from_untrusted(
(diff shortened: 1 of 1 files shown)


chia/wallet/wallet_node.py
```diff
@@ -303,8 +302,6 @@ def _close(self):
             self._process_new_subscriptions_task.cancel()
         if self._retry_failed_states_task is not None:
             self._retry_failed_states_task.cancel()
-        if self._primary_peer_sync_task is not None:
```
almogdepaz, line 306: wouldn't we now wait for long sync before we shutdown ?

xdustinface: From my understanding the underlying API call task will be anyway canceled on shutdown/disconnect via https://github.com/Chia-Network/chia-blockchain/blob/9f0220771f12549a9b566bc6e59fa10c527e8041/chia/server/ws_connection.py#L306  which should lead to cancelation of the long sync unless we suppress the cancelation somewhere what i don't think.

arvidn (commented):
as far as I can tell, this is a much more straight-forward way of saying what the existing code does. Although, I do find it unexpected that calling `new_peak_from_trusted()` would hi-jack the co-routine for the duration of the long-sync. But as far as I can tell, that's what we do.

I would approve this, but I'd like someone else to confirm our understanding is correct first.

chia/wallet/wallet_node.py
almogdepaz: maybe im missing something, in the current code we explicitly cancel _primary_peer_sync_task what would cancel the long sync now that we dont do that explicitly ?

xdustinface: Its just canceled like before, but not via the API call like i suspected above because the API calls are queued in the `new_peak_queue`, so instead the long sync gets triggered in here https://github.com/Chia-Network/chia-blockchain/blob/1d4700dceb3524e7d9be2c71c986be9a03165754/chia/wallet/wallet_node.py#L977 which gets called here https://github.com/Chia-Network/chia-blockchain/blob/1d4700dceb3524e7d9be2c71c986be9a03165754/chia/wallet/wallet_node.py#L467 after a new peak message was put into the queue via an API call.

The new peak queue processing gets canceled here still https://github.com/Chia-Network/chia-blockchain/blob/1d4700dceb3524e7d9be2c71c986be9a03165754/chia/wallet/wallet_node.py#L303

chia/wallet/wallet_node.py
```diff
@@ -1041,11 +1038,7 @@ async def new_peak_from_trusted(self, new_peak_hb: HeaderBlock, latest_timestamp
             if peer.peer_node_id not in self.synced_peers:
                 if new_peak_hb.height - current_height > self.LONG_SYNC_THRESHOLD:
                     self.wallet_state_manager.set_sync_mode(True)
-                self._primary_peer_sync_task = asyncio.create_task(
-                    self.long_sync(new_peak_hb.height, peer, uint32(max(0, current_height - 256)), rollback=True)
-                )
-                await self._primary_peer_sync_task
-                self._primary_peer_sync_task = None
+                await self.long_sync(new_peak_hb.height, peer, uint32(max(0, current_height - 256)), rollback=True)
```
almogdepaz, line 1041: @xdustinface i mean now if you shut down when we are waiting for long sync i dont' think there is any place where we break, it looks like we would wait for the sync to complete

chia/wallet/wallet_node.py
almogdepaz: im not talking about t _process_new_subscriptions_task im talking about the call to long_sync which used to be a task but now we just do await self_long_sync() and im not sure that we break anywhere in the middle of long_sync in case of a shutdown

here https://github.com/Chia-Network/chia-blockchain/pull/14362#discussion_r1090534336

chia/wallet/wallet_node.py
xdustinface: We still cancel the task which is awaiting the `long_sync`, the `_process_new_subscriptions_task` and once this is canceled there will be a canceled exception raised just the same like if we canceled the `_primary_peer_sync_task` before, right?

chia/wallet/wallet_node.py
xdustinface: Lets continue the conversation here, but see https://github.com/Chia-Network/chia-blockchain/pull/14362#discussion_r1090552770 and i think we might need [9cea0de](https://github.com/Chia-Network/chia-blockchain/pull/14362/commits/9cea0deca7a212999ea5cb481177166a7b36adde) to make it work unless it also raises on any other await inside this task.. maybe @altendky has some more thoughts about this but from my understanding the long sync should be canceled like before now.

chia/wallet/wallet_node.py
almogdepaz: if thats the case then thats fine, it looks like someone went into some effort to make ut possible we can explicitly cancel that part i dident debug if you say this shuts down ok during long sync that thats good with me

arvidn (dismissed):

arvidn (approved):

PR #14365, opened 2023-01-19
RPC call to reset wallet sync data

trepca:
### Purpose:
Make it easy to reset wallet sync db to restart syncing in case  something goes wrong. 

### Current Behavior:
User needs to delete the sqlite file to reset sync, which means they lose user generate content with that too.

### New Behavior:
They can use the new RPC call `set_wallet_resync` to only delete sync data. It sets a flag in config that toggles reset when wallet is restarted. 

### Testing Notes:
You'll need to restart the wallet for reset to happen.


chia/wallet/wallet_node.py
```diff
@@ -225,6 +227,60 @@ async def get_private_key(self, fingerprint: Optional[int]) -> Optional[PrivateK
 
         return key
 
+    def set_resync_flag(self, fingerprint, flag: bool = True):
+        config = load_config(self.root_path, "config.yaml")
```
paninaro, line 231: Should probably use the context manager version:
`with lock_and_load_config(self.root_path, "config.yaml") as config:`

chia/wallet/wallet_node.py
```diff
@@ -225,6 +227,60 @@ async def get_private_key(self, fingerprint: Optional[int]) -> Optional[PrivateK
...
+        conn: aiosqlite.Connection
+        async with manage_connection(db_path) as conn:
+            self.log.info("Resetting wallet sync data...")
+            rows = list(await conn.execute_fetchall("SELECT name FROM sqlite_master WHERE type='table'"))
+            try:
+                # if less than 20 it means we're running on old database and that's
+                # handled, but not if there's more
+                assert len(rows) < 20, (
+                    f"Expected 20 tables, found {len(rows)} tables."
+                    " Please check if the new table needs to be added in clean_resync_tables."
```
paninaro, line 254:
```suggestion
                    " Please check if the new table needs to be added in reset_sync_db."
```

chia/wallet/wallet_node.py
```diff
@@ -225,6 +227,60 @@ async def get_private_key(self, fingerprint: Optional[int]) -> Optional[PrivateK
...
+                    await conn.execute("DELETE FROM interested_coins")
+                if "interested_puzzle_hashes" in tables:
+                    await conn.execute("DELETE FROM interested_puzzle_hashes")
+                if "key_val_store" in tables:
+                    await conn.execute("DELETE FROM key_val_store")
+                if "users_nfts" in tables:
+                    await conn.execute("DELETE FROM users_nfts")
+            except aiosqlite.Error:
+                self.log.exception("Error resetting sync tables")
+                await conn.execute("ROLLBACK")
```
paninaro, line 275: nit: I would probably move the rollback into the finally block in an else statement if commit is false.

chia/util/initial-config.yaml
```diff
@@ -188,12 +188,12 @@ harvester:
   selected_network: *selected_network
 
   # Plots are searched for in the following directories
-  plot_directories: []
+  plot_directories: [ ]
```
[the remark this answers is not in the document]
trepca: no, my formatter has strange ideas it seems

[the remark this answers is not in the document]
paninaro: Does this file need to be reverted/removed from the PR? It still shows up as modified for me.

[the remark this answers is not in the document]
trepca: strange, it's changed for me, check files changed tab directly

chia/wallet/wallet_node.py
```diff
@@ -225,6 +232,65 @@ async def get_private_key(self, fingerprint: Optional[int]) -> Optional[PrivateK
...
+            finally:
+                try:
+                    if commit:
+                        self.log.info("Reset wallet sync data completed.")
+                        await conn.execute("COMMIT")
+                    else:
+                        self.log.info("Reverting reset resync changes")
+                        await conn.execute("ROLLBACK")
+                except aiosqlite.Error:
+                    self.log.exception("Error finishing reset resync db")
```
arvidn, line 290: I think printing the error code/exception would be very helpful here

chia/wallet/wallet_node.py
```diff
@@ -225,6 +232,65 @@ async def get_private_key(self, fingerprint: Optional[int]) -> Optional[PrivateK
...
+                # handled, but not if there's more
+                assert len(rows) < 20, (
+                    f"Expected 20 tables, found {len(rows)} tables."
+                    " Please check if the new table needs to be added in reset_sync_db."
+                )
+            except AssertionError:
+                self.log.exception("Incompatible database to reset")
+                return
+            await conn.execute("BEGIN")
+            commit = True
```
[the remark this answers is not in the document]
arvidn: I can see how you ended up here though. The DBWrapper is created in `WalletStateManager` and then it's immediately used to populate our store classes. If you would move the db wrapper out into `WalletNode` and pass it into the `WalletStateManager` create-function instead, you would be able to manipulate the database with it before anyone is reading from it

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -1294,3 +1294,83 @@ async def test_notification_rpcs(wallet_rpc_environment: WalletRpcTestEnvironmen
...
+    # use second node to start the same wallet, reusing config and db
+    await wallet_node_2._start_with_fingerprint(fingerprint)
+    assert wallet_node_2._wallet_state_manager
+    after_txs = await wallet_node_2.wallet_state_manager.tx_store.get_all_transactions()
+    # transactions should be the same
+    assert after_txs == before_txs
+    # only coin_store was populated in this case, but now should be empty
+    assert len(await wallet_node_2._wallet_state_manager.coin_store.get_all_unspent_coins()) == 2
+    wallet_node_2._close()
+    await wallet_node_2._await_closed()
```
arvidn, line 1376: both these tests don't seem to push the limits very much. They are both testing the "happy path".

In addition to failure cases, I think it would make sense to actually verify that the database contains the tables you expect it to before, and then that they were emptied afterwards. Also, given that you use the production code to set-up the database means that if someone happens to have a previous version of the DB, that case is not exercised. Consider the case where we add a new table in the future, these tests will only cover the case where the new table exists.

chia/wallet/wallet_node.py
```diff
@@ -225,6 +232,65 @@ async def get_private_key(self, fingerprint: Optional[int]) -> Optional[PrivateK
...
+
+    async def reset_sync_db(self, db_path, fingerprint):
+        conn: aiosqlite.Connection
+        async with manage_connection(db_path) as conn:
+            self.log.info("Resetting wallet sync data...")
+            rows = list(await conn.execute_fetchall("SELECT name FROM sqlite_master WHERE type='table'"))
+            try:
+                # if less than 20 it means we're running on old database and that's
+                # handled, but not if there's more
+                assert len(rows) < 20, (
```
arvidn, line 257: I don't think this is an appropriate use of `assert`. At least in my book, `assert` is meant to check invariants that are supposed to be maintained by the program, and them being violated is a bug. This is not that.

It would seem more appropriate to just *check* the number of tables and if they're fewer than 20, print the error log below and return.

I also think it would make sense to add some details to this check, it doesn't seem super reliable. What if we optimize away a few tables in the future? Will we remember to change this number?

arvidn (dismissed):
I think at the very least the assertion for 20 tables or more should not use an `assert`, the tests also seem pretty meager.

chia/wallet/wallet_node.py
trepca: `log.exception` will print the whole stack trace

chia/rpc/wallet_rpc_api.py
```diff
@@ -492,6 +493,23 @@ async def delete_all_keys(self, request: Dict) -> EndpointResult:
...
     ##########################################################################################
+    async def set_wallet_resync(self, request) -> Dict[str, Any]:
+        """
+        Resync the current logged in wallet. The transaction and offer records will be kept.
+        :param request: optionally pass in `enable` as bool to enable/disable resync
+        :return:
+        """
+        assert self.service.wallet_state_manager is not None
+        try:
+            enable = bool(request.get("enable", True))
```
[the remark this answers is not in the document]
paninaro: I had wondered about this as well, but I think it makes sense as an "undo" feature. If a CLI user enabled resync, and then decided they didn't mean to do that, they could disable resync. As long as the user didn't restart the wallet between those operations, they could effectively undo the request to resync the wallet.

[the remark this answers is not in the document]
arvidn: the fact that the resync only happens after restart is not (or at least poorly) documented. I assumed it happened immediately when I started my review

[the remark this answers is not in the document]
arvidn: and I agree with you btw. but It could be more clearly explained in the CLI help for instance

chia/wallet/wallet_node.py
arvidn: ah, right, it has access to the exception without it being captured in the `except` clause.

chia/wallet/wallet_node.py
```diff
@@ -225,6 +232,94 @@ async def get_private_key(self, fingerprint: Optional[int]) -> Optional[PrivateK
...
+                names = names - set(required_tables)
+                for name in names:
+                    for ignore_name in ignore_tables:
+                        if name.startswith(ignore_name):
+                            break
+                    else:
+                        raise AssertionError(
+                            f"Mismatch in expected schema to reset, found unexpected table: {name}. "
+                            "Please check if you've run all migration scripts."
+                        )
```
arvidn, line 288: it would seem simpler and more straight-forward to just print this message to the log here and return `False`, rather than throwing this exception and catching it below in order to return `False` from there. There doesn't seem to be really any opportunity for other asserts to fail here, to be caught below.

You could even remove the whole `try`/`except` block and save a level of indentation

paninaro (dismissed):

chia/wallet/wallet_node.py
```diff
@@ -225,6 +232,92 @@ async def get_private_key(self, fingerprint: Optional[int]) -> Optional[PrivateK
...
+            else:
+                self.log.debug(
+                    "Trying to disable resync: %s [%s]", fingerprint, config["wallet"].get("reset_sync_for_fingerprint")
+                )
+                if config["wallet"].get("reset_sync_for_fingerprint") == fingerprint:
+                    del config["wallet"]["reset_sync_for_fingerprint"]
+                    self.log.info("Disabled resync for wallet fingerprint: %s", fingerprint)
+            save_config(self.root_path, "config.yaml", config)
+
+    async def reset_sync_db(self, db_path, fingerprint):
```
arvidn, line 249: you don't have any type annotations in this signature (both parameters and return value). I suppose mypy isn't fully enabled on this file. I think it would be good to add it anyway though. The sane goes for `set_resync_on_startup()`. `fingerprint` and return value lack types.

arvidn (commented):
I think this looks good. but I think you should add type annotations too, to make sure mypy agrees

arvidn (approved):

PR #14366, opened 2023-01-19
Add the expected offer ID to some RPCs that take offer blobs

Quexington:
Requested for front end purposes so there is a hook to check if the offer is already in the wallet and what it's state may be.


trepca:
lets also update the tests to check for ID

trepca (approved):

PR #14368, opened 2023-01-19
Ensure fee estimates monotonically decrease with time

aqk:
Ensure that fee estimates farther in the future are equal or lower than estimates near in time.

### DIFF
--- a/chia/rpc/full_node_rpc_api.py
+++ b/chia/rpc/full_node_rpc_api.py
@@ -26,6 +26,7 @@ from chia.types.unfinished_header_block import UnfinishedHeaderBlock
 from chia.util.byte_types import hexstr_to_bytes
 from chia.util.ints import uint32, uint64, uint128
 from chia.util.log_exceptions import log_exceptions
+from chia.util.math import monotonically_decrease
 from chia.util.ws_message import WsRpcMessage, create_payload_dict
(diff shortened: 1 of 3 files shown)


prozacchiwawa (dismissed):
I think monotonically_decreasing could be named better if a better name is at hand, but otherwise LGTM.

chia/rpc/full_node_rpc_api.py
```diff
@@ -782,11 +783,13 @@ async def get_fee_estimate(self, request: Dict) -> Dict[str, Any]:
...
             cost = uint64(request["cost"])
 
-        target_times = request["target_times"]
+        target_times: List[int] = request["target_times"]
         estimator: FeeEstimatorInterface = self.service.mempool_manager.mempool.fee_estimator
+        target_times.sort()
         estimates = [
             estimator.estimate_fee_rate(time_offset_seconds=time).mojos_per_clvm_cost * cost for time in target_times
         ]
+        estimates = monotonically_decrease(estimates)
```
arvidn, line 792: If you understand how the fee estimator can end up with estimates that are not monotonically decreasing, this might be a good place to explain that in a comment, to future readers

emlowe:
close and reopen to get full CI with target change

emlowe (approved):

PR #14377, opened 2023-01-21
Get memo api

ytx1991:
### Purpose:
Add a new API to get transaction memo

### Current Behavior:
none

### New Behavior:
Return memos based on additional coins

### Testing Notes:
Unit test


paninaro:
Thanks for adding this. The RPC looks good to me.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -1092,7 +1092,7 @@ async def set_bulk_nft_did(
         for tx in nft_tx_record:
             if tx.spend_bundle is not None:
                 spend_bundles.append(tx.spend_bundle)
-                refined_tx_list.append(dataclasses.replace(tx, spend_bundle=None))
+            refined_tx_list.append(dataclasses.replace(tx, spend_bundle=None))
```
trepca, line 1095: why is this part of this PR?

trepca (changes requested):
we also need a test for XCH transaction. API should have support now to add memo to an XCH transaction.

chia/rpc/wallet_rpc_api.py
```diff
@@ -839,6 +840,15 @@ async def get_transaction(self, request: Dict) -> EndpointResult:
...
         }
 
+    async def get_transaction_memo(self, request: Dict) -> EndpointResult:
+        transaction_id: bytes32 = bytes32(hexstr_to_bytes(request["transaction_id"]))
+        tr: Optional[TransactionRecord] = await self.service.wallet_state_manager.get_transaction(transaction_id)
+        if tr is None:
+            raise ValueError(f"Transaction 0x{transaction_id.hex()} not found")
+        if tr.spend_bundle is None or len(tr.spend_bundle.coin_spends) == 0:
+            raise ValueError(f"Transaction 0x{transaction_id.hex()} doesn't have any coin spend.")
+        return {transaction_id.hex(): compute_memos(tr.spend_bundle)}
```
arvidn, line 850: the memo is the *list* of additional parameters, right? only the first item in that list is the *hint*. We have a database where the hint can be looked up (which probably would be more efficient than executing the CLVM code of the spend bundle). But in this case you really want the whole memo I take it.

chia/rpc/wallet_rpc_api.py
```diff
@@ -2261,7 +2271,7 @@ async def nft_set_did_bulk(self, request):
         for tx in tx_list:
             if tx.spend_bundle is not None:
                 spend_bundles.append(tx.spend_bundle)
-                refined_tx_list.append(dataclasses.replace(tx, spend_bundle=None))
+            refined_tx_list.append(dataclasses.replace(tx, spend_bundle=None))
```
arvidn, line 2274: is this deliberate? If so, what does it fix?
(same question applies to the other similar changes)

chia/wallet/nft_wallet/nft_wallet.py
ytx1991: It disappeared after merged the main branch.

chia/rpc/wallet_rpc_api.py
ytx1991: It's gone now.

chia/rpc/wallet_rpc_api.py
trepca: @arvidn  good point, thanks, @paninaro since memos are [hint, other memos ... ] you might want to think about how to display this in UX (if it matters)

trepca: @ytx1991 afaik `get_transaction` only fetches transactions that user sent/created, but not for transactions received from other wallets?

tests/wallet/test_wallet.py
```diff
@@ -409,6 +409,72 @@ async def test_wallet_make_transaction_with_fee(
...
+        tx = await wallet.generate_signed_transaction(uint64(tx_amount), ph_2, uint64(tx_fee), memos=[ph_2])
+        tx_id = tx.name.hex()
+        assert tx.spend_bundle is not None
+
+        fees = tx.spend_bundle.fees()
+        assert fees == tx_fee
+
+        await wallet.push_transaction(tx)
+        await full_node_1.wait_transaction_records_entered_mempool(records=[tx])
+        memos = await api_0.get_transaction_memo(dict(transaction_id=tx_id))
```
trepca, line 463: this should also test that running this on `wallet2` will also return memos

chia/rpc/wallet_rpc_api.py
```diff
@@ -1799,9 +1824,6 @@ async def did_find_lost_did(self, request) -> EndpointResult:
         p2_puzzle, recovery_list_hash, num_verification, singleton_struct, metadata = curried_args
 
         hint_list = compute_coin_hints(coin_spend)
-        old_inner_puzhash = DID_INNERPUZ_MOD.curry(
```
trepca, line 1802: why is removing this code relevant for memos?

ytx1991: removed

trepca (approved):

PR #14384, opened 2023-01-23
Fixed wallet state if db is rolled back

wjblanke:
### Purpose:

Problem is sqlite rollbacks because of exceptions thrown during writer() in new_coin_state. that would explain the aliasing of wallet IDs as well as the missing derivation_paths. if you look through logs there are lots of "Error adding state..." exceptions thrown in new_coin_state. that may roll back the db but i don't think it handles wallets that may have been created by new_coin_state in the interim with erroneous ids

### Current Behavior:

2023-01-19T10:26:38.295 wallet chia.wallet.wallet_state_manager: ERROR    Error adding state... 
  File "chia/wallet/wallet_state_manager.py", line 1065, in new_coin_state
  File "chia/wallet/wallet_state_manager.py", line 1519, in coin_added
  File "chia/wallet/did_wallet/did_wallet.py", line 389, in coin_added
AssertionError

this assertion happens after the did wallet is created so easy to see why wallet disappears from the db due to rollback but is probably still hanging around in memory causing wallet aliasing as long as the running instance is around. 

### New Behavior:

if we nuke these wallets during the exception, i think this will be fixed. but there are still probably issues with why the exceptions happen in the first place

### Testing Notes:

Try to force exceptions in new_coin_state

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1011,6 +1011,7 @@ class WalletStateManager:
 
         assert len(local_records) == len(coin_states)
         for coin_state, local_record in zip(coin_states, local_records):
+            rollback_wallets=self.wallets.copy()  # Shallow copy of wallets in case writer rolls back the db
             try:
                 async with self.db_wrapper.writer():
                     # This only succeeds if we don't raise out of the transaction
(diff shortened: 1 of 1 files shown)


arvidn (approved):
I hope that `assert` that's failing isn't meant to validate input, but to actually indicate a violation of some invariant, i.e. a bug. This looks good, it would also be good to understand why we have an assertion failure

PR #14386, opened 2023-01-23
dont announce none response capability to peers

almogdepaz:
### Purpose:

remove none response capability 

### Current Behavior:
 node announces none response capability

### New Behavior:
 dont announce

### DIFF
--- a/tests/core/full_node/full_sync/test_full_sync.py
+++ b/tests/core/full_node/full_sync/test_full_sync.py
@@ -372,46 +372,46 @@ class TestFullSync:
         # assert we failed somewhere between sub epoch 0 to sub epoch 1
         assert node_height_between(full_node_2, summary_heights[0], summary_heights[1])
 
-    # todo uncomment when none response capability is enabled
-    # @pytest.mark.asyncio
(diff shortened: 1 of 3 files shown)


tests/core/full_node/full_sync/test_full_sync.py
```diff
@@ -372,45 +372,46 @@ async def async_mock():
...
-        peers: List = [c for c in full_node_3.full_node.server.all_connections.values()]
-        request = full_node_protocol.RequestProofOfWeight(
-            blocks_950[-1].height + 1, default_1000_blocks[-1].header_hash
-        )
-        start = time.time()
-        res = await peers[0].call_api(FullNodeAPI.request_proof_of_weight, request, timeout=6)
-        assert res is None
-        duration = time.time() - start
-        assert duration > 5
+    # todo uncomment when none response capability is enabled
```
altendky, line 375: Could we put this into the skip list instead of commenting it out?  This code is just off the top of my head, it might not be quite exactly right.

```suggestion
    @pytest.mark.skip(message="skipping until we re-enable the capability in chia.protocols.shared_protocol")
```

emlowe (approved):

PR #14388, opened 2023-01-23
Fix chia wallet coins cli bugs

jack60612:
Purpose:
This adds a warning to the user letting them know that the dust filter might prevent them from seeing their coins.
This PR also makes the most sense when viewed by commit.
This PR also fixes issues where the fee was not properly dealt with & specific coin filtering rules were not respected for CATs.

Current Behavior:
No warning when making coins below the dust limit, user might be confused as wallet ignores coins.
New Behavior:
A warning was added to stop possible user confusion.

Testing Notes:
CLI Testing will be added in the coming weeks & tests were expanded for the RPC's modifed

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1370,6 +1370,9 @@ class WalletRpcApi:
             puzzle_hashes.append(decode_puzzle_hash(request["inner_address"]))
             if "memos" in request:
                 memos.append([mem.encode("utf-8") for mem in request["memos"]])
+        coins = None
+        if "coins" in request and len(request["coins"]) > 0:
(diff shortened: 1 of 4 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1370,6 +1370,9 @@ async def cat_spend(self, request, hold_lock=True) -> EndpointResult:
             puzzle_hashes.append(decode_puzzle_hash(request["inner_address"]))
             if "memos" in request:
                 memos.append([mem.encode("utf-8") for mem in request["memos"]])
+        coins = None
```
arvidn, line 1373: Would you mind adding type annotation?

arvidn (commented):
this looks good! Just two small comments.

arvidn (approved):

PR #14390, opened 2023-01-23
Annotate wallet_node

aqk:
No user visible change
No code behaviour change


chia/wallet/wallet_node.py
```diff
@@ -81,27 +82,27 @@ def get_wallet_db_path(root_path: Path, config: Dict[str, Any], key_fingerprint:
...
     root_path: Path
     constants: ConsensusConstants
     local_keychain: Optional[Keychain] = None
 
     log: logging.Logger = logging.getLogger(__name__)
 
     # Normal operation data
-    cached_blocks: Dict = dataclasses.field(default_factory=dict)
-    future_block_hashes: Dict = dataclasses.field(default_factory=dict)
+    # cached_blocks: Dict = dataclasses.field(default_factory=dict)
```
aqk, line 93: @almogdepaz Do we intend to use these commented out variables?
```
cached_blocks
future_block_hashes
proof_hashes
```

chia/wallet/wallet_node.py
```diff
@@ -444,10 +445,10 @@ async def _process_new_subscriptions(self):
                     puzzle_hashes: List[bytes32] = item.data
                     for peer in self.server.get_connections(NodeType.FULL_NODE):
                         # Puzzle hash subscription
-                        coin_states: List[CoinState] = await subscribe_to_phs(puzzle_hashes, peer, uint32(0))
```
aqk, line 447: `coin_states` was redefined here.

chia/wallet/wallet_node.py
```diff
@@ -461,10 +462,10 @@ async def _process_new_subscriptions(self):
                     # This can take a VERY long time, because it might trigger a long sync. It is OK if we miss some
                     # subscriptions or state updates, since all subscriptions and state updates will be handled by
                     # long_sync (up to the target height).
-                    request: wallet_protocol.NewPeakWallet = item.data[0]
```
aqk, line 464: mypy: `request` was "redefined" here.

chia/wallet/wallet_node.py
```diff
@@ -379,11 +380,11 @@ async def _messages_to_resend(self) -> List[Tuple[Message, Set[bytes32]]]:
 
         return messages
 
-    async def _retry_failed_states(self):
+    async def _retry_failed_states(self) -> None:
         while not self._shut_down:
             try:
                 await asyncio.sleep(5)
-                current_time = time.time()
+                current_time = int(time.time())
```
aqk, line 387: Note that we assign `self.last_state_retry_time = int(time.time())` at line
https://github.com/Chia-Network/chia-blockchain/pull/14390/files#diff-8387032f76cb8397134e314bda2f2ebbe6dd1c4aa473feac200bdbd4f27051b4L278

chia/wallet/wallet_node.py
```diff
@@ -81,27 +82,27 @@ def get_wallet_db_path(root_path: Path, config: Dict[str, Any], key_fingerprint:
...
-    cached_blocks: Dict = dataclasses.field(default_factory=dict)
-    future_block_hashes: Dict = dataclasses.field(default_factory=dict)
+    # cached_blocks: Dict = dataclasses.field(default_factory=dict)
+    # future_block_hashes: Dict = dataclasses.field(default_factory=dict)
 
     # Sync data
-    proof_hashes: List = dataclasses.field(default_factory=list)
-    state_changed_callback: Optional[Callable] = None
+    # proof_hashes: List = dataclasses.field(default_factory=list)
+    state_changed_callback: Optional[Callable[..., Any]] = None
```
altendky, line 98: Don't we know a bit better than this what is going to be passed?  And on the return, would we really be able to handle an `Any` back or could this be `object` or `None` instead?

Maybe this deserves a centralized type definition since I think we'll use it many places.

chia/wallet/wallet_node.py
```diff
@@ -324,19 +325,19 @@ async def _await_closed(self, shutting_down: bool = True):
             await asyncio.sleep(0.5)  # https://docs.aiohttp.org/en/stable/client_advanced.html#graceful-shutdown
         self.wallet_peers = None
 
-    def _set_state_changed_callback(self, callback: Callable):
+    def _set_state_changed_callback(self, callback: Callable[..., Any]) -> None:
```
altendky, line 328: As above.

chia/wallet/wallet_node.py
altendky: I saw some inconsistency like this somewhere recently.  Happy to have it cleaned up.  But...  (of course `:]`) Do we actually want to drop the float resolution down to just integer seconds?

chia/wallet/wallet_node.py
arvidn, line 328: do we really not have typing for this callback?

chia/wallet/wallet_node.py
```diff
@@ -379,11 +380,11 @@ async def _messages_to_resend(self) -> List[Tuple[Message, Set[bytes32]]]:
...
 
-    async def _retry_failed_states(self):
+    async def _retry_failed_states(self) -> None:
         while not self._shut_down:
             try:
                 await asyncio.sleep(5)
-                current_time = time.time()
+                current_time = int(time.time())
                 if self.last_state_retry_time < current_time - 10:
                     self.last_state_retry_time = current_time
```
arvidn, line 389: I don't see type annotation for `self.last_state_retry_time`. I assumed it was `int` since you changed `current_time` from `float` to `int`. This is a semantic change though, not just annotation. Are you confident this is right?

In other places where we explicitly treat time as integral, we use `uint64`.

chia/wallet/wallet_node.py
altendky: Maybe we should rename both for consistency?

arvidn (commented):
looks good generally. I would really prefer complete type annotation for the callback, and maybe some justification for the cast-to-int of the timestamp.

It also looks like the timestamp member variable lacks type annoation

altendky (changes requested):
Just a few small things to discuss and maybe change.

chia/wallet/wallet_node.py
altendky: Oof, arbitrary attribute creation that isn't being caught by the tools.  Maybe using `slots=True` should be a norm for us even on non-frozen classes.  Perhaps we do need our own dataclass decorator that defaults some of this stuff how we want it.

chia/wallet/wallet_node.py
almogdepaz: good question but i have no idea what they are for, github says @altendky was the one that added these

altendky: I was just refactoring to a dataclass in https://github.com/Chia-Network/chia-blockchain/commit/75115f2475e1e5f2421bccb8fbb80f7bafe93740.  In `.__init__()` there was `self.cached_blocks: Dict = {}` so i added `cached_blocks: Dict = dataclasses.field(default_factory=dict)`.  Other uses for `cached_blocks` at least were removed prior to mainnet https://github.com/Chia-Network/chia-blockchain/blame/88cf51ad82b6cddf4e47b26f50de54729db09be5/src/wallet/wallet_node.py.  Likewise for `future_block_hashes` and `proof_hashes`.  Well, at least within `src/wallet/wallet_node.py`.

aqk: Okay, sounds like it is safe for me to remove these. Thanks!

chia/wallet/wallet_node.py
aqk: I don't think we need to. But the reason I went in this direction was to not change the behaviour of a part of the code I'm not working on right now.

aqk:
> looks good generally. I would really prefer complete type annotation for the callback, 

Me too. Don't want this change to spread to other files, though. Have to keep these annotation PRs focused for these large messy files.

> and maybe some justification for the cast-to-int of the timestamp.

Heh. I actually think we should leave it as float. I only added that cast to get it to type-agree (see already-existing cast https://github.com/Chia-Network/chia-blockchain/pull/14390/files#diff-8387032f76cb8397134e314bda2f2ebbe6dd1c4aa473feac200bdbd4f27051b4L278).

I also wanted to preserve the current behavior. But now that I think about what Python is doing at runtime, removing the initializing cast is more behavior preserving. 


> It also looks like the timestamp member variable lacks type annoation

Yes, a lot of members introduced in `_start_with_fingerprint` still do.

chia/wallet/wallet_node.py
aqk: This is used in multiple places with multiple arguments, unfortunately. I'd rather not do this in this PR.

chia/wallet/wallet_node.py
aqk: Thanks!

aqk:
@arvidn @altendky Take a peek when you get a chance.

Pay special attention to `last_state_retry_time`, `last_wallet_tx_resend_time` and the specific reason I didn't annotate `_set_state_changed_callback`.

chia/wallet/wallet_node.py
aqk: It is a semantic change either way. Yes, these three variables are used elsewhere, but always for timeout checks.
I reversed this to remove the initialization cast to int.

chia/wallet/wallet_node.py
aqk: No, we don't. This is used in multiple places with function signatures that do not agree.

Seems out of scope for this PR.

chia/wallet/wallet_node.py
aqk: See https://github.com/Chia-Network/chia-blockchain/pull/14464

chia/wallet/wallet_node.py
aqk: See https://github.com/Chia-Network/chia-blockchain/pull/14464

arvidn:
looks good to me, but it's probably a good idea to land https://github.com/Chia-Network/chia-blockchain/pull/14464 first and rebase this

chia/wallet/wallet_node.py
altendky: Yep, didn't mean that should be considered to change here.  I'm just not sure why we dislike sub-second time resolution.

chia/wallet/wallet_node.py
altendky: Adding `slots=True` to the dataclasses decorator does indeed point out this, and another.  I would think they should get annotated and included in the dataclass definition.

```
chia/wallet/wallet_node.py:277: error: Trying to assign name "last_state_retry_time" that is not in "__slots__" of type "chia.wallet.wallet_node.WalletNode"  [misc]
chia/wallet/wallet_node.py:286: error: Trying to assign name "sync_event" that is not in "__slots__" of type "chia.wallet.wallet_node.WalletNode"  [misc]
chia/wallet/wallet_node.py:389: error: Trying to assign name "last_state_retry_time" that is not in "__slots__" of type "chia.wallet.wallet_node.WalletNode"  [misc]
```

aqk: Hey, that's nice. I'll deal with those.

arvidn (approved):

PR #14391, opened 2023-01-24
move SerializedProgram to its own file

arvidn:
The new file is no longer exempt from full mypy checks, so those are fixed.

This patch clarifies the dependency relationship between `Program` and `SerializedProgram`. the latter depends on the former, and not the other way around. This is required to avoid cyclical imports. This also has the consequence that `Program` cannot have the `to_serialized_program()` member.

This patch is split into 3 commits:

1. move `SerializedProgram` into its own file
2. fix mypy annotations in the new file
3. update all module imports to refer to the new file

It's advised to review each commit independently.

My long-term vision is to transition `SerializedProgram` into being the `chia_rs` rust implementation (currently called `chia_rs.Program`). That, in turn, would enable transitioning `CoinSpend` into using the rust implementation, which in-turn enabled transitioning `SpendBundle` into using the rust implementation.

`SpendBundle` contains a `G2Element` however, which is not solved how to transition yet.

### DIFF
--- a/benchmarks/block_ref.py
+++ b/benchmarks/block_ref.py
@@ -15,7 +15,7 @@ from chia.consensus.blockchain import Blockchain
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
 from chia.full_node.block_store import BlockStore
 from chia.full_node.coin_store import CoinStore
-from chia.types.blockchain_format.program import SerializedProgram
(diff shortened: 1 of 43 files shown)


aqk (approved):
Thanks

emlowe (approved):

PR #14392, opened 2023-01-24
Allow sending of zero coins

Quexington:
### Purpose:
0 coins currently cannot be sent from the standard wallet. There isn't really a reason for this that I can see, so this removes that restriction.

### New Behavior:
Calls to the `/send_transaction` RPC will now create a 0 coin if `"amount": 0` is specified.
`chia wallet send` now guards zero-amount spends behind the `--override` flag


Quexington:
This is going to require a more sophisticated fix than I assumed, closing for a more sophisticated approach at a later date.

PR #14397, opened 2023-01-25
Rename notification config

ytx1991:
### Purpose:
Rename notification config
https://github.com/Chia-Network/issue-tracker/issues/225

### Current Behavior:
accept_notifications = False

### New Behavior:
enable_notifications = True

### Testing Notes:
Uint Test

### DIFF
--- a/chia/wallet/notification_manager.py
+++ b/chia/wallet/notification_manager.py
@@ -46,7 +46,7 @@ class NotificationManager:
         if (
             coin_state.spent_height is None
             or not self.wallet_state_manager.wallet_node.config.get("enable_notifications", True)
-            or self.wallet_state_manager.wallet_node.config.get("required_notification_amount", 0)
+            or self.wallet_state_manager.wallet_node.config.get("required_notification_amount", 100000000)
             > coin_state.coin.amount
         ):
             return False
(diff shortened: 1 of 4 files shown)


chia/wallet/notification_manager.py
```diff
@@ -45,7 +45,7 @@ async def create(
     async def potentially_add_new_notification(self, coin_state: CoinState, parent_spend: CoinSpend) -> bool:
         if (
             coin_state.spent_height is None
-            or not self.wallet_state_manager.wallet_node.config.get("accept_notifications", False)
+            or not self.wallet_state_manager.wallet_node.config.get("enable_notifications", True)
             or self.wallet_state_manager.wallet_node.config.get("required_notification_amount", 0)
```
Quexington, line 49: If we're going to have this on by default, we also need a reasonable filter number so people can't spam us for free.

Quexington (changes requested):

chia/wallet/notification_manager.py
trepca: filter number?

Quexington: `"required_notification_amount"`

trepca (dismissed):

Quexington (dismissed):

trepca (approved):

PR #14398, opened 2023-01-25
Add some extra checks for size and existence to notification backend

Quexington:
### Purpose
Some extra careful steps that should probably be taken when receiving a notification.


wjblanke (approved):
ok but sebastjan should take a look too

trepca (approved):

PR #14403, opened 2023-01-26
move wallet_is_synced/wallets_are_synced to FullNodeSimulator methods

altendky:
### Purpose:

I was going to use the wait for synced feature in another simulator method so it needed to move out of the tests anyways.  Figured I'd add them as methods as well as direct `.wait_for_*()` methods to avoid having to nest them within `time_out_assert()` calls.

Notes:
- [ ] Could revisit in another PR all of the uses to make a few cleanups.
  - [ ] Wait for synced then assert balances instead of waiting for balances then waiting for synced.
  - [ ] Some places are waiting for the wallet to reach a height, then nodes to sync, then the wallet to sync.  Could probably drop at least the wallet at height check?  Maybe also the node height checks?  Or maybe just wait for the node that wasn't farming?
  - [ ] Timeouts could probably be reduced due to the automatic adjustment for hanging runners that is now made in CI.  Maybe don't specify them at all and use the default 5 second timeout in most places.
  - [ ] Some spots use `.get_finished_sync_up_to()` before waiting for the wallet to be synced.
  - [ ] Some places use a block farming helper beforehand that will, in another pr, be made to make sure the target wallet is synced.

Draft for:
- [x] Let https://github.com/Chia-Network/chia-blockchain/pull/14196 go in first and catch up with its changes.


aqk (approved):
This is great. Much easier to see what is intended.

emlowe (approved):

PR #14404, opened 2023-01-26
Simplify request_mempool_transactions() and get_items_not_in_filter()

AmineKhaldi:
* Only deal with the relevant subset of the `MempoolItem` (which is `SpendBundle` here).
* `get_items_not_in_filter()` doesn't need to be async.
* Use len(items) instead of counter for the limit check.
* Update tests to reflect the improvements.

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -669,17 +669,15 @@ class MempoolManager:
 
     def get_items_not_in_filter(self, mempool_filter: PyBIP158, limit: int = 100) -> List[SpendBundle]:
         items: List[SpendBundle] = []
-        counter = 0
 
         assert limit > 0
 
         # Send 100 with the highest fee per cost
         for dic in reversed(self.mempool.sorted_spends.values()):
             for item in dic.values():
-                if counter == limit:
+                if len(items) == limit:
                     return items
                 if mempool_filter.Match(bytearray(item.spend_bundle_name)):
                     continue
                 items.append(item.spend_bundle)
-                counter += 1
         return items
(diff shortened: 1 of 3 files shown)


chia/full_node/mempool_manager.py
```diff
@@ -667,22 +667,16 @@ async def new_peak(
...
             for item in dic.values():
                 if counter == limit:
-                    broke_from_inner_loop = True
-                    break
+                    return items
                 if mempool_filter.Match(bytearray(item.spend_bundle_name)):
                     continue
-                items.append(item)
+                items.append(item.spend_bundle)
                 counter += 1
```
arvidn, line 681: you could even remove `counter` and just use `len(items)` instead, right?

chia/full_node/mempool_manager.py
```diff
@@ -667,22 +667,16 @@ async def new_peak(
         self.mempool.fee_estimator.new_block(FeeBlockInfo(new_peak.height, included_items))
         return txs_added
 
-    async def get_items_not_in_filter(self, mempool_filter: PyBIP158, limit: int = 100) -> List[MempoolItem]:
-        items: List[MempoolItem] = []
+    def get_items_not_in_filter(self, mempool_filter: PyBIP158, limit: int = 100) -> List[SpendBundle]:
```
arvidn, line 670: given that you could make this change without updating a test suggest we don't have any tests for this function. Do you know if any test exercises this functionality?

arvidn (commented):
looks good

chia/full_node/mempool_manager.py
AmineKhaldi: Yeap, done.

chia/full_node/mempool_manager.py
AmineKhaldi: I found none so I sent https://github.com/Chia-Network/chia-blockchain/pull/14416

arvidn (approved):

PR #14410, opened 2023-01-27
fix sending of the `new_on_chain_notification` state change message

paninaro:
### Purpose:

The `new_on_chain_notification` state change message is intended to notify the UI that new notifications are available for processing. Wallet state change messages need to be handled by wallet_rpc_api.py's `_state_changed` method to ensure that the payload is properly constructed to ensure delivery to the UI.

### Current Behavior:

Without this fix, the state change message is never sent to the UI.

### New Behavior:

With this fix, the state change message is sent and seen by the UI.

### Testing Notes:

Manually tested by watching RPC messages in the UI.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -234,6 +234,7 @@ class WalletRpcApi:
             "added_stray_cat",
             "pending_transaction",
             "tx_update",
+            "new_on_chain_notification",
         }:
             payloads.append(create_payload_dict("state_changed", change_data, self.service_name, "wallet_ui"))
 


arvidn (approved):
we should think of ways to make these kinds of mistakes impossible or caught by CI. for instance, could the "change" be an enum instead of a free-form string? Maybe we could have an assert that `change` matches at least one of these cases. e.g. `assert len(payloads) > 0`

Quexington (approved):

paninaro:
> we should think of ways to make these kinds of mistakes impossible or caught by CI. for instance, could the "change" be an enum instead of a free-form string? Maybe we could have an assert that `change` matches at least one of these cases. e.g. `assert len(payloads) > 0`

Agreed. I had thought of adding a log message to catch unhandled state change messages, but I was concerned that it might end up being a source of annoyance in the logs without some additional processing.

Issue #14389, opened 2023-01-23
Pressing ctrl+c during CLI commands spits out traceback


wallentx:
![Image](https://user-images.githubusercontent.com/8990544/214121319-5ebef3b2-7846-4046-b7f5-9a26ff7c20e0.png)

![Image](https://user-images.githubusercontent.com/8990544/214121318-a0b2e519-8d46-483e-b8ec-b058c7835a02.png)

altendky:
I think there are likely two separate cases to be considered here.  The general import-time cancellation which would be handled generically and the specific case of the client session in the first shown traceback.

altendky:
I think there are two cases to be handled here separately.  The first traceback shows an unclosed client session that should be dealt with specifically.  The second traceback is a general cancellation while importing and would involve a fairly general keyboard interrupt consumption mechanism.  Maybe something with `sys.excepthook` when run from an entry points script or such.

wallentx:
```
Traceback (most recent call last):
  File "chia\cmds\chia.py", line 9, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 499, in exec_module
  File "chia\cmds\beta.py", line 10, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 499, in exec_module
  File "chia\cmds\beta_funcs.py", line 9, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "PyInstaller\loader\pyimod02_importers.py", line 499, in exec_module
  File "chia\util\chia_logging.py", line 11, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
[...]

PR #14415, opened 2023-01-27
close the keychain proxy in more failure and cancellation cases

altendky:
### Purpose:

One of the two cases of loud tracebacks during shutdown noted in https://github.com/Chia-Network/chia-blockchain/issues/14389 is caused by the keychain proxy client session not being closed.  This PR makes it more likely that the keychain proxy is closed properly in more situations.  For one of the cases anyways, this was introduced in 1.6.2 and seemingly associated with https://github.com/Chia-Network/chia-blockchain/commit/ca536d7002f9fd87f612ec7ee622e032582b63fa.

### Current Behavior:

Only happy path termination results in closure of the keychain proxy.

### New Behavior:

The keychain proxy is also closed in more cases for sad path termination.

### Draft for:

- [x] Maybe still make a context manager for the cases that can use it.
  - not right now


paninaro (approved):
lgtm

PR #14416, opened 2023-01-27
Add unit tests for get_items_not_in_filter

AmineKhaldi:


tests/core/mempool/test_mempool_manager.py
```diff
@@ -329,3 +330,51 @@ async def test_sb_twice_with_eligible_coin_and_different_spends_order() -> None:
...
+    assert result == [mempool_item3, mempool_item2, mempool_item1]
+
+    # Filter everything
+    full_filter = PyBIP158([bytearray(sb1_name), bytearray(sb2_name), bytearray(sb3_name)])
+    result = await mempool_manager.get_items_not_in_filter(full_filter)
+    assert result == []
+
+    # Negative limit
+    result = await mempool_manager.get_items_not_in_filter(empty_filter, limit=-1)
+    assert result == [mempool_item3, mempool_item2, mempool_item1]
```
arvidn, line 360: this return value seems a bit odd. is specifying a negative limit sensible? It would seem to make more sense to throw an exception. Maybe `assert` that `limit >= 0`

arvidn (dismissed):

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: That's exactly why I left that path covered here. I didn't know whether this behavior was intended or not so I thought to expose it.

arvidn: you might be able to find out whether it's deliberate or not by reviewing git history and call sites. I suspect it's not, and that we should turn it into an error

arvidn (approved):

Issue #14411, opened 2023-01-27
Notifications are regenerated after deletion when switching between wallet keys


paninaro:

paninaro:
```
$ chia wallet notifications get -f 3339549250

ID: 7cf74176b5d35de4d8c45b9cb21e8596364047768af5f26f65cd2036573cf072
message: {"v":1,"t":1,"d":{"u":"https://raw-testnet.dexie.space/HiwZR93RrXzyAawBeKd6JM34gufD74XgyijfDFBcywnX","ph":"d78d740ecd42aaab981034a893213d75d39cb4e298c032213e8bbc7a0aa79d27"}}
amount: 10000000

ID: 8d6969ced07767367a0c34f6f27ecb8ee33f2c1b5194f2fb1272a9e028283a8e
message: {"v":1,"t":1,"d":{"u":"https://testnet.dexie.space/offers/GRmaTAcZxNncFHKqoyyTC6bapKxBWFx9RsWYuq9pu4iN","ph":"ffe88bf6298ff1701e7a56402bb6e9b27e40732c899db0ce2cb73914d88abcf0"}}
amount: 10000000

ID: 420a1311a2b90ec83997d70c76fca979db08abf6d694211afaa24165348cde29
message: {"v":1,"t":1,"d":{"u":"https://raw-testnet.dexie.space/4GWM9kNw3HXNqUesZPWfNiH29zyW2qKQztB6TxBi2tPU","ph":"4a9a2dedab47322f83b0363ef473ea23143051bbdbaa08d6561a3768f57d15db"}}
amount: 10000000

ID: 5fc70c154be7255c8078aa456cff4a93621562bc7040dbcc5eeb91578614adce
message: {"v":1,"t":1,"d":{"u":"https://raw-testnet.dexie.space/69k5LSedoXWxMpozztznwZ3iTYNCPVMKXMh54DcMF2Md","ph":"d6b28c798bb38b8cfc4902362421e23ef3495a1154edaf0e8a3cfc04d5e81592"}}
amount: 10000000
```

```
$ chia wallet notifications delete --all -f 3339549250
Success: True
```

```
chia wallet notifications get -f 3339549250
<no output>
```

_Switch to another wallet key and wait for sync..._

_Switch back to original wallet key and wait for sync..._

```
chia wallet notifications get -f 3339549250

ID: 7cf74176b5d35de4d8c45b9cb21e8596364047768af5f26f65cd2036573cf072
[...]

PR #14418, opened 2023-01-27
Add a persitent id table to notification store

Quexington:
### Purpose:
Add a list of IDs for notification we've synced separate from the main notifications table.  This means that even if a notification is deleted, we don't forget that we've already synced it previously.

### Testing Notes:
Reproduce this issue and see if it exists: https://github.com/Chia-Network/chia-blockchain/issues/14411

### DIFF
--- a/chia/wallet/notification_store.py
+++ b/chia/wallet/notification_store.py
@@ -46,6 +46,8 @@ class NotificationStore:
                 "CREATE TABLE IF NOT EXISTS notifications(" "coin_id blob PRIMARY KEY," "msg blob," "amount blob" ")"
             )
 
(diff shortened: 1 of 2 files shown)


chia/wallet/notification_store.py
```diff
@@ -46,6 +46,8 @@ async def create(
                 "CREATE TABLE IF NOT EXISTS notifications(" "coin_id blob PRIMARY KEY," "msg blob," "amount blob" ")"
             )
 
+            await conn.execute("CREATE TABLE IF NOT EXISTS all_ids(coin_id blob PRIMARY KEY)")
```
trepca, line 49: table name is a bit generic, can we use `notifications_all_ids` or something similar?

trepca (changes requested):
needs tests

trepca (approved):

PR #14419, opened 2023-01-27
Add additional state_changed events to metrics destination

cmmarslender:
### Purpose:
Adds additional metrics events so chia-exporter can reliably report additional data related to connections for each service as well as know when to update harvester info for remote harvesters (from the farmer)

### Current Behavior:
Some events not passed to metrics

### New Behavior:
* Additional events passed to metrics destination. 
* `add_connection` event added to harvester on_connect method
* `close_connection` event added to wallet_node

### DIFF
--- a/chia/harvester/harvester.py
+++ b/chia/harvester/harvester.py
@@ -98,7 +98,7 @@ class Harvester:
         return default_get_connections(server=self.server, request_node_type=request_node_type)
 
     async def on_connect(self, connection: WSChiaConnection):
-        pass
+        self.state_changed("add_connection")
 
     def _set_state_changed_callback(self, callback: Callable):
         self.state_changed_callback = callback
--- a/chia/rpc/farmer_rpc_api.py
+++ b/chia/rpc/farmer_rpc_api.py
@@ -118,6 +118,14 @@ class FarmerRpcApi:
                     "wallet_ui",
                 )
             )
+            payloads.append(
+                create_payload_dict(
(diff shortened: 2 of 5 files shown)


aqk:
@cmmarslender Just curious - why did you name it `metrics` instead of `service_name` ?

aqk:
Do you want to add an `add_connection` event here, too?

https://github.com/Chia-Network/chia-blockchain/pull/14419/files#diff-8387032f76cb8397134e314bda2f2ebbe6dd1c4aa473feac200bdbd4f27051b4R560

aqk (approved):

cmmarslender:
> @cmmarslender Just curious - why did you name it `metrics` instead of `service_name` ?

For this PR, its to be consistent with the other metrics that were added previously. But in general, its because all the data that goes to the "metrics" group is for metrics. Could have called it chia-exporter, but there isn't a reason that someone else couldn't also connect and consume metrics events, so the more general name seemed appropriate

cmmarslender:
> Do you want to add an `add_connection` event here, too?
>
> https://github.com/Chia-Network/chia-blockchain/pull/14419/files#diff-8387032f76cb8397134e314bda2f2ebbe6dd1c4aa473feac200bdbd4f27051b4R560


add_connection for wallet already existed here https://github.com/Chia-Network/chia-blockchain/blob/main/chia/wallet/wallet_node.py#L575

arvidn (approved):

PR #14421, opened 2023-01-27
fix reporting of deprecation of plotter install and bladebit2 commands

altendky:
### Purpose:

Make sure that the ~deprecation~ removal messages for `chia plotters install` and `chia plotters bladebit2` are presented even when arguments such as `-h` are passed.

### Current Behavior:

Presently you just get an empty help output for `-h`.

### New Behavior:

You get the message regardless of anything following `chia plotters bladebit2`

### Testing Notes:

Make sure to test out other commands as well since this fix injects a pre-parsing of the arguments.

### DIFF
--- a/chia/plotters/plotters.py
+++ b/chia/plotters/plotters.py
@@ -452,12 +452,19 @@ def call_plotters(root_path: Path, args):
     deprecation_warning = (
         "[DEPRECATED] 'chia plotters install' is no longer available. Use install-plotter.sh/ps1 instead."
     )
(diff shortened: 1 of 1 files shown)


ChiaMineJP:
LGTM

At least the following command ran successfully so it seems it doesn't change original plotting behavior.
```
chia plotters bladebit diskplot -d 'X:\chia_plot\dst' -t 'X:\chia_plot\tmp1'
```

emlowe (approved):

PR #14422, opened 2023-01-27
Stop sending change to a new address

ytx1991:
### Purpose:
Stop sending change to a new address to improve performance and achieve staking and other functions.

### Current Behavior:
The change of XCH and CAT will be saved to a new address

### New Behavior:
The change of XCH and CAT will not be saved to a new address.

### Testing Notes:
unit test


chia/wallet/wallet.py
```diff
@@ -401,7 +401,14 @@ async def _generate_unsigned_transaction(
                 else:
                     primaries.append({"puzzlehash": newpuzzlehash, "amount": uint64(amount), "memos": memos})
                 if change > 0:
-                    change_puzzle_hash: bytes32 = await self.get_new_puzzlehash()
+                    if (
+                        self.wallet_state_manager.config.get("reuse_public_key_for_change", False)
+                        and newpuzzlehash != coin.puzzle_hash
+                    ):
+                        print(f"{newpuzzlehash.hex()} {coin.puzzle_hash.hex()}")
```
trepca, line 408: let's remove the print

chia/wallet/wallet.py
```diff
@@ -401,7 +401,14 @@ async def _generate_unsigned_transaction(
                 else:
                     primaries.append({"puzzlehash": newpuzzlehash, "amount": uint64(amount), "memos": memos})
                 if change > 0:
-                    change_puzzle_hash: bytes32 = await self.get_new_puzzlehash()
+                    if (
+                        self.wallet_state_manager.config.get("reuse_public_key_for_change", False)
```
trepca, line 405: can we also override this from `generate_signed_transaction` parameter?

tests/wallet/nft_wallet/test_nft_offers.py
```diff
@@ -149,12 +159,28 @@ async def test_nft_offer_with_fee(
...
+        old_maker_offer if forwards_compat else Offer.from_bytes(trade_make.offer),
+        peer,
+        fee=taker_fee,
+        reuse_puzhash=reuse_puzhash and not forwards_compat,
     )
 
     assert trade_take is not None
     assert tx_records is not None
-
+    if reuse_puzhash and not forwards_compat:
```
trepca, line 170: you should also add an `else` condition here that checks derivation records have changed in case `reuse_puzhash=False`

trepca (approved):

PR #14430, opened 2023-01-30
break untrusted long sync if connected to synced trusted node

almogdepaz:
### Purpose:

dont wait for long sync from untrusted if we are connected to a trusted synced node

### New Behavior:

break mid sync and sync from trusted node

### Testing Notes:

currently theres a test making sure we actually break mid sync, the breaks are consistent with how we return in case of an error so there should not be any added db inconsistency risk


xdustinface (commented):
While reviewing this PR i realised there is an imo better approach of doing this, see #14640. But if we for some reason still go with this PR here note that the included test also passes on current main so it doesn't really test what its supposed to test?

almogdepaz:
> While reviewing this PR i realised there is an imo better approach of doing this, see #14640. But if we for some reason still go with this PR here note that the included test also passes on current main so it doesn't really test what its supposed to test?

this functionality is supposed to already happen in main, this is an attempt to make it better so the test should work (its not currently tested on main)

almogdepaz:
> > this functionality is supposed to already happen in main, this is an attempt to make it better so the test should work (its not currently tested on main)
> 
> It is supposed to happen but the logic there isn't working, one reason is because we process messages off the new peak queue, i explained this in the PR description of #14640.
> 
> Your test passes on `main` but the line you linked above is not the reason why it passes. What happens there is that we just never start the untrusted long sync and bail out early at the line you linked and then just do the trusted sync.
> 
> Even if you remove this whole block here on `main`
> 
> https://github.com/Chia-Network/chia-blockchain/blob/41ca9622a3721356e324daba711d767b2f817536/chia/wallet/wallet_node.py#L1158-L1160
> 
> and run your test, it still passes (you need to fix the issue that coroutines can only be awaited once which exists there because of the `MagicMock` returns the same coroutine multiple times) because it just gracefully ends the untrusted sync and then does the trusted sync, during which it disconnects the untrusted which is why it's not in `synced_peers` at the end.. but this doesn't imply the sync didn't happen/finish. There is not really anything which tests that the untrusted sync was started first and then aborted.

thanks i was looking at the logs and made sure that we get "Cancelling untrusted sync, we are connected to a trusted peer"

should add an assertion that the sync to untrusted started

emlowe:
closing old PR

PR #14431, opened 2023-01-30
rpc: Dont't expect `WalletNode._await_closed` to maybe return a task

xdustinface:
### Purpose:

Simplify some code because `WalletNode._await_closed` doesn't potentially return a task unless there is some hacky thing around somewhere which leads to it which i didn't see?

### New Behavior:

No change of behaviour.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import asyncio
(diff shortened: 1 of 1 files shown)


altendky (approved):
Looks fine to me.  Might be worth explicitly hinting all the `._await_closed()` definitions to return `None`?  Most do, a few are unhinted.

emlowe (approved):

PR #14432, opened 2023-01-30
full_node: Hint the sync store and create it earlier

xdustinface:
### Purpose:

Simplify.

### New Behavior:

No change in behaviour.


altendky (approved):
Only question is whether the `SyncStore.peers_changed` `asyncio.Event` ([docs](https://docs.python.org/3.11/library/asyncio-sync.html#asyncio.Event), [source](https://github.com/python/cpython/blob/v3.11.1/Lib/asyncio/locks.py#L158)) needs to be created while there is an event loop active.  At least at first glance it doesn't appear that anything loop related happens at the time of construction.  Just a `collections.deque()` gets created which is a non-async data structure.

emlowe (approved):

PR #14433, opened 2023-01-30
Add unit tests for ephemeral time locks

AmineKhaldi:


tests/core/mempool/test_mempool_manager.py
```diff
@@ -329,3 +329,54 @@ async def test_sb_twice_with_eligible_coin_and_different_spends_order() -> None:
...
+        ),
+    ],
+)
+async def test_ephemeral_timelock(
+    opcode: ConditionOpcode,
+    lock_value: int,
+    expected_status: MempoolInclusionStatus,
+    expected_error: Optional[Err],
+) -> None:
+    mempool_manager = await instantiate_mempool_manager(get_coin_record_for_test_coins)
```
arvidn, line 372: Absolute height 4 is OK, but height 3 not. As a reviewer of this test I can't tell whether that's right now not, There's nothing telling me what block height this mempool is at. Ideally, that height would be visible in this test and not require digging into abstractions to find it.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -329,3 +329,54 @@ async def test_sb_twice_with_eligible_coin_and_different_spends_order() -> None:
...
     assert mempool_manager.get_spendbundle(reordered_sb_name) is None
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "opcode,lock_value,expected_status,expected_error",
+    [
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, -2, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, -1, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, 0, MempoolInclusionStatus.SUCCESS, None),
```
arvidn, line 340: I think the discrepancy between `SECONDS_RELATIVE` and `HEIGHT_RELATIVE` warrants a comment, to remind a reviewer of this test what the rules are

tests/core/mempool/test_mempool_manager.py
```diff
@@ -37,7 +37,7 @@
 TEST_COIN3 = Coin(IDENTITY_PUZZLE_HASH, IDENTITY_PUZZLE_HASH, TEST_COIN_AMOUNT3)
 TEST_COIN_ID3 = TEST_COIN3.name()
 TEST_COIN_RECORD3 = CoinRecord(TEST_COIN3, uint32(0), uint32(0), False, TEST_TIMESTAMP)
-TEST_HEIGHT = uint32(1)
+TEST_HEIGHT = uint32(5)
```
arvidn, line 40: I think this constant is a mistake. apparently it wasn't used for anything until now

tests/core/mempool/test_mempool_manager.py
```diff
@@ -329,3 +329,54 @@ async def test_sb_twice_with_eligible_coin_and_different_spends_order() -> None:
...
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, 1, MempoolInclusionStatus.FAILED, Err.ASSERT_SECONDS_RELATIVE_FAILED),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, -2, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, -1, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, 0, MempoolInclusionStatus.PENDING, Err.ASSERT_HEIGHT_RELATIVE_FAILED),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, 1, MempoolInclusionStatus.PENDING, Err.ASSERT_HEIGHT_RELATIVE_FAILED),
+        (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 4, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 5, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 6, MempoolInclusionStatus.PENDING, Err.ASSERT_HEIGHT_ABSOLUTE_FAILED),
+        (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 7, MempoolInclusionStatus.PENDING, Err.ASSERT_HEIGHT_ABSOLUTE_FAILED),
+        (ConditionOpcode.ASSERT_SECONDS_ABSOLUTE, TEST_TIMESTAMP - 1, MempoolInclusionStatus.SUCCESS, None),
```
arvidn, line 350: it's not clear what `TEST_TIMESTAMP` is or why this is correct.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -53,7 +53,7 @@ async def get_coin_record_for_test_coins(coin_id: bytes32) -> Optional[CoinRecor
     return test_coin_records.get(coin_id)
 
 
-def create_test_block_record(*, height: uint32 = TEST_HEIGHT) -> BlockRecord:
+def create_test_block_record(*, height: uint32 = TEST_HEIGHT, timestamp: uint64 = TEST_TIMESTAMP) -> BlockRecord:
```
arvidn, line 56: it doesn't look like you're using this variable anywhere.
if you add `, *` before the `timestamp` parameter, you'll prevent it being specified as a positional argument, but has to be named if specified

tests/core/mempool/test_mempool_manager.py
```diff
@@ -84,10 +84,12 @@ def create_test_block_record(*, height: uint32 = TEST_HEIGHT) -> BlockRecord:
 
 
 async def instantiate_mempool_manager(
-    get_coin_record: Callable[[bytes32], Awaitable[Optional[CoinRecord]]]
+    get_coin_record: Callable[[bytes32], Awaitable[Optional[CoinRecord]]],
+    current_block_height: uint32 = TEST_HEIGHT,
+    current_block_timestamp: uint64 = TEST_TIMESTAMP,
```
arvidn, line 89: since both of these are integers, I think it would make sense to force them to be named as well

tests/core/mempool/test_mempool_manager.py
```diff
@@ -329,3 +331,61 @@ async def test_sb_twice_with_eligible_coin_and_different_spends_order() -> None:
...
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, 0, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, 1, MempoolInclusionStatus.FAILED, Err.ASSERT_SECONDS_RELATIVE_FAILED),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, -2, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, -1, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, 0, MempoolInclusionStatus.PENDING, Err.ASSERT_HEIGHT_RELATIVE_FAILED),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, 1, MempoolInclusionStatus.PENDING, Err.ASSERT_HEIGHT_RELATIVE_FAILED),
+        (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 4, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 5, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 6, MempoolInclusionStatus.PENDING, Err.ASSERT_HEIGHT_ABSOLUTE_FAILED),
+        (ConditionOpcode.ASSERT_HEIGHT_ABSOLUTE, 7, MempoolInclusionStatus.PENDING, Err.ASSERT_HEIGHT_ABSOLUTE_FAILED),
```
arvidn, line 351: these `ASSERT_HEIGHT_ABSOLUTE` tests use concrete numbers like 4, 5, 6 and 7. When you create the mempool you create it at the height `TEST_HEIGHT`. it seems unnecessary to force the reader to then look up what `TEST_HEIGHT` is just to ensure that these tests make sense. using `TEST_HEIGHT` obfuscate the true height. It would be simpler and easier if you would just create the mempool with a concrete height.

Alternatively, you could use the same approach as you do with the timestamps, where both the tests and the creation of the mempool refers to `TEST_TIMESTAMP`, to kind of make the concrete value of `TEST_TIMESTAMP` unimportant. But I don't think the name `TEST_TIMESTAMP` or `TEST_HEIGHT` adds anything here (but possibly dilutes the test).

tests/core/mempool/test_mempool_manager.py
```diff
@@ -329,3 +331,61 @@ async def test_sb_twice_with_eligible_coin_and_different_spends_order() -> None:
...
+    sb1 = spend_bundle_from_conditions(conditions)
+    sb2 = spend_bundle_from_conditions(conditions, created_coin)
+    # sb spends TEST_COIN and creates created_coin which gets spent too
+    sb = SpendBundle.aggregate([sb1, sb2])
+    # We shouldn't have a record of this ephemeral coin
+    assert await get_coin_record_for_test_coins(created_coin.name()) is None
+    # This ephemeral coin has entered the mempool at the current block height
+    assert mempool_manager.peak is not None
+    assert mempool_manager.peak.height == TEST_HEIGHT
+    _, status, error = await add_spendbundle(mempool_manager, sb, sb.name(), first_added_height=TEST_HEIGHT)
```
arvidn, line 390: I'm pretty sure `first_added_height` is irrelevant here. The fact that you specify it suggests that it matters. Does it? If so, I think it warrants a comment

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Block height and timestamp are more explicit now.

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Done.

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Block height and timestamp are more explicit now.

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Done.

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Done.

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Done.

tests/core/mempool/test_mempool_manager.py
arvidn: the `ASSERT_SECONDS_RELATIVE` condition requires the block timestamp to be *greater than* the coin timestamp + this argument. meaning that 0 still won't accept being spent in the same block.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -329,3 +332,61 @@ async def test_sb_twice_with_eligible_coin_and_different_spends_order() -> None:
...
+    "opcode,lock_value,expected_status,expected_error",
+    [
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, -2, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, -1, MempoolInclusionStatus.SUCCESS, None),
+        # Relative 0 seconds is okay
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, 0, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_SECONDS_RELATIVE, 1, MempoolInclusionStatus.FAILED, Err.ASSERT_SECONDS_RELATIVE_FAILED),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, -2, MempoolInclusionStatus.SUCCESS, None),
+        (ConditionOpcode.ASSERT_HEIGHT_RELATIVE, -1, MempoolInclusionStatus.SUCCESS, None),
+        # Relative 0 height is not okay (unlike relative 0 seconds)
```
arvidn, line 348: for `ASSERT_HEIGHT_RELATIVE`, the block height must be *greater than* the coin creation height + the argument, which means the coin cannot be spent in the same block (where the height would be the same)

arvidn (approved):

PR #14440, opened 2023-01-31
full_node: Reduce blockchain lock time

xdustinface:
### Purpose:

Reduce time we lock the blockchain lock during peak processing.

### Current Behavior:

- The signage point post processing (which updates the nodes/farmers, the signage point times and the UI) currently always runs under the blockchain lock.

- `FullNode.peak_post_processing_2` is handling the data propagation to peers after processing peaks and is not supposed to run under the blockchain lock. Right now there are two places where this happens though:
https://github.com/Chia-Network/chia-blockchain/blob/c26902a6ece013cf0a22fc5e5ff8131b9a1ef28a/chia/full_node/full_node.py#L593 
https://github.com/Chia-Network/chia-blockchain/blob/c26902a6ece013cf0a22fc5e5ff8131b9a1ef28a/chia/full_node/full_node.py#L1324

### New Behavior:
 
- The signage point post processing is part of `FullNode.peak_post_processing_2`
- `FullNode.peak_post_processing_2` is not longer called under the blockchain lock.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -578,6 +578,7 @@ class FullNode:
                 if not response:
                     raise ValueError(f"Error short batch syncing, invalid/no response for {height}-{end_height}")
                 async with self._blockchain_lock_high_priority:
+                    ppp_result: Optional[PeakPostProcessingResult] = None
                     state_change_summary: Optional[StateChangeSummary]
                     success, state_change_summary = await self.receive_block_batch(response.blocks, peer, None)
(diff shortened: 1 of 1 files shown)


arvidn:
I don't think we have very thorough tests for the interleaving that may happen if we don't lock this mutex correctly. Making a change like this needs to be made very carefully and with tests to ensure it doesn't introduce any regressions.

Also, can the upside of making this change be quantified somehow? is the risk worth the benefit?

PR #14447, opened 2023-01-31
Always Log service startup and shutdown.

aqk:
### Purpose:

Always Log service startup and shutdown. This is important information when trying to understand log files.

### Current Behavior:

Does not log startup and shutdown


hoffmang9 (dismissed):
lgtm

emlowe (approved):

PR #14448, opened 2023-01-31
Extend sign API to support hex string

ytx1991:
### Purpose:
To simplify the Chialisp code using sign API, we need to support signing hex message.

### Current Behavior:
The message will be regarded as a normal string.

### New Behavior:
If the is_hex = true then the message will be convert to bytes

### Testing Notes:
unit test

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1368,9 +1368,7 @@ class WalletRpcApi:
                 return {"success": False, "error": f"NFT for {entity_id.hex()} doesn't exist."}
 
             assert isinstance(selected_wallet, NFTWallet)
-            pubkey, signature = await selected_wallet.sign_message(
-                request["message"], target_nft, is_hex
-            )
+            pubkey, signature = await selected_wallet.sign_message(request["message"], target_nft, is_hex)
             latest_coin_id = target_nft.coin.name()
         else:
             return {"success": False, "error": f'Unknown ID type, {request["id"]}'}
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1317,7 +1317,7 @@ class WalletRpcApi:
         if isinstance(is_hex, str):
             is_hex = bool(is_hex)
         pubkey, signature = await self.service.wallet_state_manager.main_wallet.sign_message(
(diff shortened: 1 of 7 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1313,8 +1313,11 @@ async def sign_message_by_address(self, request) -> EndpointResult:
         :return:
         """
         puzzle_hash: bytes32 = decode_puzzle_hash(request["address"])
+        is_hex = request.get("is_hex", False)
+        if isinstance(is_hex, str):
+            is_hex = bool(is_hex)
         pubkey, signature = await self.service.wallet_state_manager.main_wallet.sign_message(
-            request["message"], puzzle_hash
+            request["message"], puzzle_hash, request.get("is_hex", is_hex)
```
trepca, line 1320: shouldn't `request.get("is_hex", is_hex)` be just `is_hex` ?

trepca (changes requested):

chia/rpc/wallet_rpc_api.py
```diff
@@ -1362,7 +1368,9 @@ async def sign_message_by_id(self, request) -> EndpointResult:
                 return {"success": False, "error": f"NFT for {entity_id.hex()} doesn't exist."}
 
             assert isinstance(selected_wallet, NFTWallet)
-            pubkey, signature = await selected_wallet.sign_message(request["message"], target_nft)
+            pubkey, signature = await selected_wallet.sign_message(
+                request["message"], target_nft, request.get("is_hex", False)
```
trepca, line 1372: this needs to be sanitized too

tests/wallet/test_wallet.py
```diff
@@ -847,6 +848,18 @@ async def test_sign_message(
             puzzle.get_tree_hash(),
             G2Element.from_bytes(bytes.fromhex(response["signature"])),
         )
+        # Test hex string
+        message = "0123456789ABCDEF"
+        response = await api_0.sign_message_by_address(
+            {"address": encode_puzzle_hash(ph, "xch"), "message": message, "is_hex": True}
```
trepca, line 854: please add tests with badly formed `is_hex` parameter too

trepca (approved):

altendky:
Note that discussions need to be resolved.

PR #14450, opened 2023-01-31
Check incoming notification hints

Quexington:
### Purpose:

We weren't checking whether an incoming notification has a hint that matches us, so we were accidentally sending every notification to ourselves.


Quexington:
https://github.com/Chia-Network/chia-blockchain-gui/issues/1393

trepca (approved):

Issue #14453, opened 2023-01-31
[Bug] wallet falls out of sync


esaung:
### What happened?

Seeing these errors in the log files around the same time the GUI is going from synced (green) to syncing (orange).

### Version

1.7.0rc5

### What platform are you using?

macOS

### What ui mode are you using?

GUI

### Relevant log output

```shell
2023-01-31T17:01:25.672 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia/rpc/rpc_server.py", line 331, in safe_handle
  File "chia/rpc/rpc_server.py", line 322, in ws_api
  File "chia/rpc/wallet_rpc_api.py", line 2512, in nft_get_info
  File "chia/util/byte_types.py", line 53, in from_hexstr
  File "chia/util/byte_types.py", line 35, in __init__
ValueError: bad bytes32 initializer b''

2023-01-31T17:01:26.734 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia/rpc/rpc_server.py", line 331, in safe_handle
  File "chia/rpc/rpc_server.py", line 322, in ws_api
  File "chia/rpc/wallet_rpc_api.py", line 2512, in nft_get_info
  File "chia/util/byte_types.py", line 53, in from_hexstr
ValueError: non-hexadecimal number found in fromhex() arg at position 0

2023-01-31T17:01:27.752 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia/rpc/rpc_server.py", line 331, in safe_handle
  File "chia/rpc/rpc_server.py", line 322, in ws_api
  File "chia/rpc/wallet_rpc_api.py", line 2512, in nft_get_info
  File "chia/util/byte_types.py", line 53, in from_hexstr
  File "chia/util/byte_types.py", line 35, in __init__
ValueError: bad bytes32 initializer b''
```

PR #14454, opened 2023-01-31
Return a better error message when the 'coin_id' argument for nft_get_info cannot be decoded into a 32 byte hex string

aqk:
Return a better error message when the 'coin_id' argument for nft_get_info cannot be decoded into a 32 byte hex string.

### Purpose:

Fix error reporting.

### Current Behavior:

Leave an exception in the log when interpreting 'coin_id' as a hex encoded bytes32

### New Behavior:

Return error message `Invalid Coin ID format for 'coin_id': <<contents>>`

### Testing Notes:

See https://github.com/Chia-Network/chia-blockchain/pull/new/ak.fix-14453

Reporter: @esaung

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -2530,7 +2530,10 @@ class WalletRpcApi:
         if coin_id.startswith(AddressType.NFT.hrp(self.service.config)):
             coin_id = decode_puzzle_hash(coin_id)
         else:
-            coin_id = bytes32.from_hexstr(coin_id)
+            try:
+                coin_id = bytes32.from_hexstr(coin_id)
+            except ValueError:
+                return {"success": False, "error": f"Invalid Coin ID format for 'coin_id' {request['coin_id']}"}
         # Get coin state
         peer: Optional[WSChiaConnection] = self.service.get_full_node_peer()
         assert peer is not None
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -2533,7 +2533,7 @@ class WalletRpcApi:
             try:
                 coin_id = bytes32.from_hexstr(coin_id)
             except ValueError:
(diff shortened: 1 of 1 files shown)


paninaro (dismissed):

chia/rpc/wallet_rpc_api.py
```diff
@@ -2530,7 +2530,10 @@ async def nft_get_info(self, request: Dict) -> EndpointResult:
         if coin_id.startswith(AddressType.NFT.hrp(self.service.config)):
             coin_id = decode_puzzle_hash(coin_id)
         else:
-            coin_id = bytes32.from_hexstr(coin_id)
+            try:
+                coin_id = bytes32.from_hexstr(coin_id)
+            except ValueError:
+                return {"success": False, "error": f"Invalid Coin ID format for 'coin_id' {request['coin_id']}"}
```
altendky, line 2536: I have two basic thoughts around this.  I only expect action here on the first as I think the second could easily turn into a rabbit hole and don't feel the need to force it's exploration.

1. Let's make the information presented a touch more 'formally' to make odd cases of leading spaces etc more apparent.  I'll admit that using `!r` results in a Python representation and that something like `json.dumps()` would be a bit more language agnostic.  Maybe something to consider applying across the whole code base.

```suggestion
                return {"success": False, "error": f"Invalid Coin ID format for 'coin_id': {request['coin_id']!r}"}
```

2. There were two cases listed in https://github.com/Chia-Network/chia-blockchain/issues/14453.  `bad bytes32 initializer b''` and `non-hexadecimal number found in fromhex() arg at position 0`.  This change improves things by always presenting the problematic input, but it does lose the explanation of what the complaint was.

aqk: Good idea for when the values can contain whitespace, yeah.

Quexington:
I think this needs to be rebased, it looks like it brings a lot of main commits with it that are not necessary for the PR.

Quexington (approved):

emlowe (approved):

emlowe:
close and reopen for CI

PR #14458, opened 2023-02-01
full_node: Rename some `receive`/`respond` prefixes to `add`

xdustinface:
### Purpose:

- The prefixes `respond` and `receive` are confusing here imo, we have full node API methods prefixed `respond` but the methods changed here in the `FullNode` class are methods add the received data to the internal data structures so i think `add` prefix is the better choice? Maybe it's only me but when i read `receive` or `respond` it more reads like something where communication with a peer is involved.
- Also drop some parameter wrappings which were there to follow the API Call response naming

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -2438,7 +2438,7 @@ class FullNode:
             )
             response = await peer.call_api(FullNodeAPI.request_compact_vdf, peer_request, timeout=10)
             if response is not None and isinstance(response, full_node_protocol.RespondCompactVDF):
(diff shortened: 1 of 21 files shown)


arvidn:
I think this renaming makes sense. I don't think we should land this until 1.7.0 is released and we no longer have an active release branch. Fixes from the release branch should have the right of way to land with minimal conflicts.

arvidn (approved):
I think this is an improvement. I seem to recall some people not being convinced. can we make sure there's support for this?

PR #14464, opened 2023-02-01
Typecheck _set_state_changed_callback

aqk:
### Purpose:

Code cleanup. No behaviour change.

### DIFF
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -6,7 +6,7 @@ import random
 import time
 import traceback
 from pathlib import Path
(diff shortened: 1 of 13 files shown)


aqk:
Note the widening of the types at 
https://github.com/Chia-Network/chia-blockchain/blob/671255996e40062fdf110aeb6997a798e26e4c98/chia/rpc/rpc_server.py#L36 and at the declaration sites. The decl site `Optional` /  `None` is used to compare against in some places.

The callsite Optional could be done away with more easily by requiring the Dict, with a little loss of information to the downstream code.

arvidn (approved):

PR #14474, opened 2023-02-02
Correct fee estimate calculation to be in mojos, not millimojos

### LINKED ISSUE #14457: [Bug] Fee estimates are very high
### What happened?

Fee estimates are very high - hundreds of mojos per cost when the blocks are fairly empty.

```
'current_fee_rate': 454, 'estimates': [454, 454, 0], 'fee_rate_last_block': 7.621014857016043e-06, 'fees_last_block': 1000, 'full_node_synced': True, 'last_block_cost': 131216120, 'last_peak_timestamp': 1675212533, 'last_tx_block_height': 3186157, 'mempool_fees': 0, 'mempool_max_size': 550000000000, 'mempool_size': 0, 'node_time_utc': 1675212567, 'num_spends': 0, 'peak_height': 3186157, 'success': True, 'target_times': [60, 120, 300]}


  Mempool max cost: 550000000000 CLVM cost
      Mempool cost:            0 CLVM cost
     Mempool count:            0 spends
   Fees in Mempool:            0 mojos

Stats for last transaction block:
      Block height:      3186157
        Block fees:         1000 mojos
        Block cost:    131216120 CLVM cost
          Fee rate:    7.621e-06 mojos per CLVM cost

Fee Rate Estimates:
    1  minute: 454 mojo per CLVM cost
    2 minutes: 454 mojo per CLVM cost
    5 minutes: 0 mojo per CLVM cost
```

### Version

1.7.0 rc4

Windows

GUI

### Relevant log output

```shell
N/A
```

aqk:

### DIFF
--- a/chia/cmds/show_funcs.py
+++ b/chia/cmds/show_funcs.py
@@ -181,7 +181,7 @@ async def print_fee_info(node_client: FullNodeRpcClient) -> None:
     print("\nFee Rate Estimates:")
     max_name_len = max(len(name) for name in target_times_names)
     for (n, e) in zip(target_times_names, res["estimates"]):
-        print(f"    {n:>{max_name_len}}: {e} mojo per CLVM cost")
+        print(f"    {n:>{max_name_len}}: {e:.3f} mojo per CLVM cost")
     print("")
 
 
--- a/chia/full_node/bitcoin_fee_estimator.py
+++ b/chia/full_node/bitcoin_fee_estimator.py
@@ -6,7 +6,7 @@ from chia.full_node.fee_estimator import SmartFeeEstimator
 from chia.full_node.fee_estimator_interface import FeeEstimatorInterface
 from chia.full_node.fee_tracker import FeeTracker
 from chia.types.clvm_cost import CLVMCost
-from chia.types.fee_rate import FeeRate
+from chia.types.fee_rate import FeeRateV2
 from chia.types.mempool_item import MempoolItem
 from chia.util.ints import uint32, uint64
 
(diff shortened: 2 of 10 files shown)


arvidn:
I get the impression the reason to introduce a v2 type is to record that you have corrected the unit. If you didn't, you may want to consider running a one-time pass over the database to perform the conversion, and then create a table to indicate that you've performed the conversion (or some other marker/version number). It could simplify the remaining code and just let you correct the bug.

aqk:
> I get the impression the reason to introduce a v2 type is to record that you have corrected the unit. If you didn't, you may want to consider running a one-time pass over the database to perform the conversion, and then create a table to indicate that you've performed the conversion (or some other marker/version number). It could simplify the remaining code and just let you correct the bug.

Ah, unfortunately, no.

The reason I did it this way is that I used `FeeRate` and `FeeEstimate` both in the node <-> wallet protocol, AND as the return types for the internal call that feeds the RPC. (Look at where `fee_rate_v2_to_v1` is used).

To properly upgrade the `FeeRate` and `FeeEstimate` data structures for the wire protocol, I need a much bigger PR.

This is the smallest I can see how make a change that carries through the float from the underlying estimator machinery, but I'm definitely open to a different way, and avoids pulling in other issues.

Accepting this patch leaves us open to changing the internals of `FeeRateV2` and `FeeEstimateV2` like you suggest, because I'm not changing what goes over the wire in this patch.

I'm also working on the tests we discussed, but I wanted to get this smaller PR here as a stopgap.

emlowe:
Right now in 1.7.0 the estimates are considerably off by a factor of 1000 - I think we should merge this PR in (fix up the pre-commit) and get it into 1.7.0 as it makes the estimates somewhat saner - there are still plenty of bugs, but this is much closer to sane fees than what we have now. I've been running it, and although far from perfect, it is much better

arvidn:
it seems like "fixed" would be a better label for this PR

emlowe (approved):

PR #14476, opened 2023-02-03
Separate mempool item's JSON response value from the internal structure

AmineKhaldi:


arvidn (dismissed):

arvidn:
you don't provide an explanation of why you want to make `MempoolItem` not streamable. I consider it a pretty good feature as I might want to store `MempoolItems` in sqlite blobs.

Separating the publicly visible structure from the internal one makes sense. It's not obvious to me that this is the simplest way. We only use the public type for conversion to- and from- JSON, right? for instance, it might be simpler to do the json dict conversion and just remove the non-public fields.

arvidn (commented):
I think this looks reasonable

arvidn:
@altendky what do you think?

arvidn (changes requested):

altendky:
> @altendky what do you think?

I've chatted with both Arvid and Amine about this topic in general.  I don't feel the need to block this PR but I have expressed that while it is moving in the direction of one option around being in control of the interface, it is just one piece of many on one possible path.  Can I be a bit more vague?  Probably not.

aqk (approved):

arvidn (approved):

PR #14478, opened 2023-02-03
Quex.offer mod tests

Quexington:
### Purpose:
Adds a bunch of forwards compatibility tests for offers


chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -37,7 +37,19 @@
     calculate_synthetic_secret_key,
     puzzle_for_pk,
 )
-from chia.wallet.trading.offer import OFFER_MOD, OFFER_MOD_HASH, NotarizedPayment, Offer
+from chia.wallet.trading.offer import (
+    OFFER_MOD,
+<<<<<<< HEAD
```
emlowe, line 42: This seems to have merge conflict items left in the file?

emlowe (changes requested):

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -782,7 +794,10 @@ async def make_nft1_offer(
         fee: uint64,
         min_coin_amount: Optional[uint64] = None,
         max_coin_amount: Optional[uint64] = None,
+        old: bool = False,
     ) -> Offer:
+        DESIRED_OFFER_MOD = OFFER_MOD_OLD if old else OFFER_MOD
```
emlowe, line 799: any thought about wrapping MOD and MOD_HASH into a Class or other structure in order to keep them together?

chia/wallet/trading/offer.py
```diff
@@ -539,7 +547,11 @@ def from_spend_bundle(cls, bundle: SpendBundle) -> Offer:
         requested_payments: Dict[Optional[bytes32], List[NotarizedPayment]] = {}
         driver_dict: Dict[bytes32, PuzzleInfo] = {}
         leftover_coin_spends: List[CoinSpend] = []
+        old: bool = False
         for coin_spend in bundle.coin_spends:
+            if not old and OFFER_MOD_OLD_BYTES in bytes(coin_spend):
```
emlowe, line 552: This seems the only place this is used - can you just do bytes() call here

emlowe (changes requested):

chia/wallet/nft_wallet/nft_wallet.py
Quexington: Can you elaborate on the benefit of this? Or is this just a general thought?

emlowe: Mostly a general thought - you can't mix up the two and you don't need to set them individually. I was mostly thinking it avoids a class of possible typos where you use the wrong hash

emlowe: a NamedTuple maybe

altendky: `namedtuple` is, at least in my mind, primarily a migration tool for moving from regular `tuple` to a full class, presumably a dataclass.  It provides both the old indexing interface that you would be deprecating and the new attribute access you want to move towards.  But I don't think that applies here.

altendky: This sounds maybe like the `NewType` stuff that Adam is playing around with.

emlowe (dismissed):

chia/wallet/nft_wallet/nft_wallet.py
emlowe: this isn't that critical so resolving conversation

arvidn (approved):

emlowe (approved):

PR #14480, opened 2023-02-03
fix crossed parameters for setup_simulators_and_wallets_inner() calling setup_full_node()

altendky:
### Purpose:

Correct the test code.

### Current Behavior:

The self hostname is passed as the db name while the db name is used as the self hostname.

### New Behavior:

Parameters go to the proper places.

### DIFF
--- a/chia/simulator/setup_nodes.py
+++ b/chia/simulator/setup_nodes.py
@@ -283,10 +283,10 @@ async def setup_simulators_and_wallets_inner(
             )
         )  # block tools modifies constants
         sim = setup_full_node(
-            bt_tools[index].constants,
-            bt_tools[index].config["self_hostname"],
-            db_name,
-            bt_tools[index],
+            consensus_constants=bt_tools[index].constants,
+            db_name=db_name,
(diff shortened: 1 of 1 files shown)


chia/simulator/setup_nodes.py
```diff
@@ -283,10 +283,10 @@ async def setup_simulators_and_wallets_inner(
             )
         )  # block tools modifies constants
         sim = setup_full_node(
-            bt_tools[index].constants,
-            bt_tools[index].config["self_hostname"],
-            db_name,
```
altendky, line 288: https://github.com/Chia-Network/chia-blockchain/blob/d1a4c579f4dd2b058fcd46f2249a876a60b1dec5/chia/simulator/setup_services.py#L80-L83

emlowe (dismissed):
This looks correct to me - we should retarget to main and undraft it?

emlowe (approved):

chia/simulator/setup_nodes.py
wallentx: assuming this can be resolved.

PR #14487, opened 2023-02-04
feat: make recurse_jsonify support for float

birdca:
### Purpose:
To extend the type of number support in recurse_jsonify.

### Current Behavior:
for numbers, currently recurse_jsonify only supports for int

### New Behavior:
make recurse_jsonify also supports for float numbers

### DIFF
--- a/chia/util/streamable.py
+++ b/chia/util/streamable.py
@@ -328,6 +328,8 @@ def recurse_jsonify(d: Any) -> Any:
         return d
     elif isinstance(d, int):
         return int(d)
+    elif isinstance(d, float):
+        return float(d)
     elif d is None or type(d) == str:
         return d
     elif hasattr(d, "to_json_dict"):


arvidn:
@birdca can you expand a bit on the justification to do this?
our binary serialization format ("Streamable") does not support floating point numbers, for pretty good reasons I think. If I'm not mistaken, we only use JSON serialization for types that are also Streamable.
Floating point numbers don't really round-trip well in general.

birdca:
@arvidn I saw this function from data_layer and think it's really useful and use it elsewhere. I thought this function simply check the data type and output corresponding data type, did not do casting or type conversion, so not facing the floating point issues.
But yeah, you are right, Streamable did not support floating point numbers.

emlowe:
Closing PR - we don't believe the proposed code is useful to chia-blockchain in general.

PR #14506, opened 2023-02-06
Revert "pin mypy to < 1 (#14504)"

altendky:
### Purpose:

Fixup to account for new complaints from mypy v1.

```
chia/util/files.py:74: error: Incompatible types in assignment (expression has type "Union[str, bytes, PathLike[str], PathLike[bytes], int]", variable has type "PathLike[Any]")  [assignment]
chia/util/files.py:75: error: Argument 1 to "write" of "AsyncTextIOWrapper" has incompatible type "Union[str, bytes]"; expected "str"  [arg-type]
chia/util/files.py:75: error: Argument 1 to "write" of "_UnknownAsyncBinaryIO" has incompatible type "Union[str, bytes]"; expected "Union[bytes, Union[bytearray, memoryview, array[Any], mmap, _CData, PickleBuffer]]"  [arg-type]
chia/util/files.py:80: error: Argument 1 to "move_file_async" has incompatible type "PathLike[Any]"; expected "Path"  [arg-type]
```

### Current Behavior:

Pinned to mypy <1.

### New Behavior:

Working with mypy 1.

### Draft For:
- [ ] Actually dealing with the complaints.

### DIFF
--- a/chia/util/files.py
+++ b/chia/util/files.py
@@ -71,8 +71,8 @@ async def write_file_async(file_path: Path, data: Union[str, bytes], *, file_mod
     mode: Literal["w+", "w+b"] = "w+" if type(data) == str else "w+b"
     temp_file_path: Path
     async with tempfile.NamedTemporaryFile(dir=file_path.parent, mode=mode, delete=False) as f:
-        temp_file_path = f.name
-        await f.write(data)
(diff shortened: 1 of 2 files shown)


emlowe (approved):
A reasonable approach

PR #14515, opened 2023-02-07
just tuple() instead of tuple([])

altendky:


aqk (approved):

arvidn (approved):
does it currently produce a 1-tuple of a list?

altendky:
No, no bug here, just a minor waste of time.  `tuple()` takes an iterable.  Existing code created a tuple from a list from a generator, of sorts at least.  Modified, it creates a tuple from the generator directly.  Just saving an intermediate list.

PR #14522, opened 2023-02-08
Fix removing spend bundles from the mempool when a new peak is available

AmineKhaldi:
* In the optimized case, it's not safe to remove from `removal_coin_id_to_spendbundle_ids` while looping over it. We'd skip items as a result.
* Optimize this optimized case a little bit further by collecting all the spend bundle IDs to be removed and removing them in one call instead of many.

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -621,17 +621,19 @@ class MempoolManager:
         if use_optimization and last_npc_result is not None:
             # We don't reinitialize a mempool, just kick removed items
             if last_npc_result.conds is not None:
+                spendbundle_ids_to_remove = []
                 for spend in last_npc_result.conds.spends:
                     if spend.coin_id in self.mempool.removal_coin_id_to_spendbundle_ids:
                         spendbundle_ids: List[bytes32] = self.mempool.removal_coin_id_to_spendbundle_ids[
                             bytes32(spend.coin_id)
-                        ].copy()
+                        ]
+                        spendbundle_ids_to_remove.extend(spendbundle_ids)
                         for spendbundle_id in spendbundle_ids:
                             item = self.mempool.spends.get(spendbundle_id)
                             if item:
                                 included_items.append(item)
                             self.remove_seen(spendbundle_id)
-                        self.mempool.remove_from_pool(spendbundle_ids, MempoolRemoveReason.BLOCK_INCLUSION)
+                self.mempool.remove_from_pool(spendbundle_ids_to_remove, MempoolRemoveReason.BLOCK_INCLUSION)
         else:
             old_pool = self.mempool
             self.mempool = Mempool(old_pool.mempool_info, old_pool.fee_estimator)
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -625,7 +625,7 @@ class MempoolManager:
                     if spend.coin_id in self.mempool.removal_coin_id_to_spendbundle_ids:
                         spendbundle_ids: List[bytes32] = self.mempool.removal_coin_id_to_spendbundle_ids[
                             bytes32(spend.coin_id)
-                        ]
+                        ].copy()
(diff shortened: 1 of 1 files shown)


tests/core/mempool/test_mempool_manager.py
```diff
@@ -257,6 +263,22 @@ async def generate_and_add_spendbundle(
     return (sb, sb_name, result)
 
 
+def mempool_item_from_spendbundle(spend_bundle: SpendBundle) -> MempoolItem:
+    generator = simple_solution_generator(spend_bundle)
+    npc_result = get_name_puzzle_conditions(
+        generator=generator, max_cost=INFINITE_COST, cost_per_byte=COST_PER_BYTE, mempool_mode=True
```
arvidn, line 269: is `INFINITE_COST` important here? I would prefer some reasonable limit

tests/core/mempool/test_mempool_manager.py
```diff
@@ -552,3 +574,45 @@ async def test_get_items_not_in_filter() -> None:
...
+@pytest.mark.asyncio
+async def test_new_peak_remove_from_pool() -> None:
+    mempool_manager = await instantiate_mempool_manager(get_coin_record_for_test_coins)
+    # Create a bunch of mempool items that spend TEST_COIN in order to setup
+    # a mapping of that coin's ID to several spend bundles
+    sb1_conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, 1]]
+    sb1 = spend_bundle_from_conditions(sb1_conditions)
+    sb2_conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, 2]]
+    sb2 = spend_bundle_from_conditions(sb2_conditions, TEST_COIN2)
+    sb3_conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, 3]]
```
arvidn, line 588: it seems like a lot of this could be simplified by storing them in lists and create them with list-comprehension

arvidn (commented):
good catch! what do you think of making the test a bit shorter? I think the repetitions are a bit distracting from the core idea of the test

tests/core/mempool/test_mempool_manager.py

tests/core/mempool/test_mempool_manager.py

tests/core/mempool/test_mempool_manager.py
```diff
@@ -552,3 +575,43 @@ async def test_get_items_not_in_filter() -> None:
...
+    sb12 = SpendBundle.aggregate([sb1, sb2])
+    sb12_name = sb12.name()
+    mi12 = mempool_item_from_spendbundle(sb12)
+    sb13 = SpendBundle.aggregate([sb1, sb3])
+    sb13_name = sb13.name()
+    mi13 = mempool_item_from_spendbundle(sb13)
+    sb123 = SpendBundle.aggregate([sb1, sb2, sb3])
+    sb123_name = sb123.name()
+    mi123 = mempool_item_from_spendbundle(sb123)
+    mempool_manager.mempool.removal_coin_id_to_spendbundle_ids = {
```
arvidn, line 598: this doesn't look OK. tests really should be using the public interface for the classes they test

AmineKhaldi: We'll add tests for this when we activate the 1 to n mapping of coin IDs to spend bundles.

arvidn (approved):
looks good. ~~would you mind squashing these commits to make the change smaller~~ (never mind)

arvidn:
the title of this PR is misleading now, and will cause a bad changelog entry. you should probably add the "cleanup" tag

AmineKhaldi:
> the title of this PR is misleading now, and will cause a bad changelog entry. you should probably add the "cleanup" tag

Added the label.

aqk (approved):
This is great, thank you.

Also, `spendbundle_ids_to_remove` is much more clear than doing two things with `spendbundle_ids`

PR #14524, opened 2023-02-08
wallet: Fix sync mode indication

xdustinface:
### Purpose:

Make sure to always have the sync mode set whenever we are long syncing and also make sure to reset the sync flag in any case.

### Current Behavior:

There are currently a few scenarios where we can end up not properly resetting the flag when we for example bail out from the sync after setting the flag or after the untrusted long sync succeeded. 

The GUI is blocked during trusted catch-up sync.
- The balance view in the GUI shows loading spinners all over the place during the initial sync
- The NFT page only shows a big loading spinner until the initial sync is done
- The connections view just shows a big loading spinner until the initial sync is done

### New Behavior:

The sync flag will always be reset whenever the sync context ends.

The GUI is released during trusted catch-up sync. 
- The balance view shows 0 during the initial sync
- The NFT page shows the existing NFTs during the initial sync with a "loading" indication in the corner
- The connections screen shows the connections properly during the initial sync

### Testing notes:

Just sync your wallet to a trusted node, restart the wallet and let it sync again. Do this on `main` and do the same with this PR to see the differences i described above.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -1111,8 +1111,7 @@ class WalletNode:
             await self.wallet_state_manager.new_peak(new_peak)
 
     async def new_peak_from_trusted(self, new_peak_hb: HeaderBlock, latest_timestamp: uint64, peer: WSChiaConnection):
-        current_height: uint32 = await self.wallet_state_manager.blockchain.get_finished_sync_up_to()
-        async with self.wallet_state_manager.set_sync_mode(new_peak_hb.height):
+        async with self.wallet_state_manager.set_sync_mode(new_peak_hb.height) as current_height:
             await self.wallet_state_manager.blockchain.set_peak_block(new_peak_hb, latest_timestamp)
             # Disconnect from all untrusted peers if our local node is trusted and synced
             await self.disconnect_and_stop_wpeers()
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -382,7 +382,7 @@ class WalletNode:
 
         self.sync_event = asyncio.Event()
         self.log_in(private_key)
-        self.wallet_state_manager.set_sync_mode(False)
+        self.wallet_state_manager.state_changed("sync_changed")
 
         async with self.wallet_state_manager.puzzle_store.lock:
             index = await self.wallet_state_manager.puzzle_store.get_last_derivation_path()
@@ -1112,7 +1112,7 @@ class WalletNode:
 
     async def new_peak_from_trusted(self, new_peak_hb: HeaderBlock, latest_timestamp: uint64, peer: WSChiaConnection):
         current_height: uint32 = await self.wallet_state_manager.blockchain.get_finished_sync_up_to()
-        async with self.wallet_state_manager.lock:
+        async with self.wallet_state_manager.set_sync_mode(new_peak_hb.height):
             await self.wallet_state_manager.blockchain.set_peak_block(new_peak_hb, latest_timestamp)
             # Disconnect from all untrusted peers if our local node is trusted and synced
             await self.disconnect_and_stop_wpeers()
@@ -1120,10 +1120,7 @@ class WalletNode:
             # disconnected), we assume that the full node will continue to give us state updates, so we do
             # not need to resync.
             if peer.peer_node_id not in self.synced_peers:
-                if new_peak_hb.height - current_height > self.LONG_SYNC_THRESHOLD:
-                    self.wallet_state_manager.set_sync_mode(True)
                 await self.long_sync(new_peak_hb.height, peer, uint32(max(0, current_height - 256)), rollback=True)
-                self.wallet_state_manager.set_sync_mode(False)
 
     async def new_peak_from_untrusted(
         self, new_peak_hb: HeaderBlock, peer: WSChiaConnection, request_time: uint64
@@ -1146,7 +1143,6 @@ class WalletNode:
         syncing = False
         if far_behind or len(self.synced_peers) == 0:
             syncing = True
-            self.wallet_state_manager.set_sync_mode(True)
 
         secondary_sync_running = (
             self._secondary_peer_sync_task is not None and self._secondary_peer_sync_task.done() is False
@@ -1164,8 +1160,6 @@ class WalletNode:
         except Exception:
             self.log.exception(f"Error syncing to {peer.get_peer_info()}")
             await peer.close()
-            if syncing:
-                self.wallet_state_manager.set_sync_mode(False)
             return False
         return True
 
@@ -1188,10 +1182,8 @@ class WalletNode:
         await self.wallet_state_manager.blockchain.new_valid_weight_proof(weight_proof, block_records)
 
         if syncing:
-            async with self.wallet_state_manager.lock:
-                self.log.info("Primary peer syncing")
+            async with self.wallet_state_manager.set_sync_mode(new_peak_hb.height):
                 await self.long_sync(new_peak_hb.height, peer, fork_point, rollback=True)
-            self.log.info(f"New peak wallet.. {new_peak_hb.height} {peer.get_peer_info()} 12")
             return
 
         # we exit earlier in the case where syncing is False and a Secondary sync is running
@@ -1239,7 +1231,6 @@ class WalletNode:
                     await self.receive_state_from_peer(list(self.race_cache[header_hash]), peer)
 
             self.wallet_state_manager.state_changed("new_block")
-            self.wallet_state_manager.set_sync_mode(False)
             self.log.info(f"Finished processing new peak of {new_peak_hb.height}")
             return True
 
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -538,7 +538,8 @@ class WalletStateManager:
 
     @asynccontextmanager
     async def set_sync_mode(self, target_height: uint32) -> AsyncIterator[None]:
-        self.log.debug(f"set_sync_mode enter {await self.blockchain.get_finished_sync_up_to()}-{target_height}")
+        if self.log.level == logging.DEBUG:
+            self.log.debug(f"set_sync_mode enter {await self.blockchain.get_finished_sync_up_to()}-{target_height}")
         async with self.lock:
             start_time = time.time()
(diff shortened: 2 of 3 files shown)


chia/wallet/wallet_node.py
```diff
@@ -1129,18 +1130,15 @@ async def new_peak_wallet(self, new_peak: wallet_protocol.NewPeakWallet, peer: W
...
-        async with self.wallet_state_manager.lock:
+        async with self.wallet_state_manager.set_sync_mode(new_peak_hb.height):
             await self.wallet_state_manager.blockchain.set_peak_block(new_peak_hb, latest_timestamp)
             # Disconnect from all untrusted peers if our local node is trusted and synced
             await self.disconnect_and_stop_wpeers()
             # Sync to trusted node if we haven't done so yet. As long as we have synced once (and not
             # disconnected), we assume that the full node will continue to give us state updates, so we do
             # not need to resync.
             if peer.peer_node_id not in self.synced_peers:
-                if new_peak_hb.height - current_height > self.LONG_SYNC_THRESHOLD:
```
arvidn, line 1140: you ended up removing this logic. It doesn't look right. I would have expected the context manager to just wrap the call to `long_sync()`, just like we do today.

It looks like an important distinction to allow holding the wallet state manager mutex while also not setting `sync_mode`.

chia/wallet/wallet_state_manager.py
```diff
@@ -529,13 +528,37 @@ async def synced(self):
...
+    def sync_target(self) -> Optional[uint32]:
+        return self._sync_target
+
+    @asynccontextmanager
+    async def set_sync_mode(self, target_height: uint32) -> AsyncIterator[None]:
+        self.log.debug(f"set_sync_mode enter {await self.blockchain.get_finished_sync_up_to()}-{target_height}")
+        async with self.lock:
+            start_time = time.time()
+            start_height = await self.blockchain.get_finished_sync_up_to()
+            self._sync_target = target_height
```
arvidn, line 545: previously we would not set `sync_mode = True` if `start_height - current_height <= self.LONG_SYNC_THRESHOLD`
. It seems this logic is gone now, and it's not obvious that's for the better, Is it deliberate?

chia/wallet/wallet_node.py
```diff
@@ -1129,18 +1130,15 @@ async def new_peak_wallet(self, new_peak: wallet_protocol.NewPeakWallet, peer: W
 
     async def new_peak_from_trusted(self, new_peak_hb: HeaderBlock, latest_timestamp: uint64, peer: WSChiaConnection):
         current_height: uint32 = await self.wallet_state_manager.blockchain.get_finished_sync_up_to()
```
arvidn, line 1132: you end up calling this `async` function twice now. Once here and once inside `set_sync_mode()`. Since it's `async`, it might be a good idea to just call it once.

You could yield this value in the context manager for instance, to make it available to the caller

chia/wallet/wallet_state_manager.py
```diff
@@ -529,13 +528,37 @@ async def synced(self):
...
+    def sync_mode(self) -> bool:
+        return self._sync_target is not None
+
+    @property
+    def sync_target(self) -> Optional[uint32]:
+        return self._sync_target
+
+    @asynccontextmanager
+    async def set_sync_mode(self, target_height: uint32) -> AsyncIterator[None]:
+        self.log.debug(f"set_sync_mode enter {await self.blockchain.get_finished_sync_up_to()}-{target_height}")
```
arvidn, line 541: you end up calling this `async` function here unconditionally (for the 3rd time), regardless of whether we have debug logging enabled or not. I think you should try to avoid these redundant calls

chia/wallet/wallet_state_manager.py
```diff
@@ -529,13 +528,37 @@ async def synced(self):
...
-        self.sync_mode = mode
-        self.sync_target = sync_height
-        self.state_changed("sync_changed")
+    @property
+    def sync_mode(self) -> bool:
+        return self._sync_target is not None
+
+    @property
+    def sync_target(self) -> Optional[uint32]:
+        return self._sync_target
```
arvidn, line 537: the previous member `sync_target` is not an `Optional`, there's no risk for issues changing it into an optional?

chia/wallet/wallet_node.py
xdustinface: Yeah i removed that intentionally because `LONG_SYNC_THRESHOLD` is 300 and we roll back 256 blocks anyway in this case so why not just set the sync flag.. because, thats what we do?

chia/wallet/wallet_state_manager.py
xdustinface: It wasn't used in any place so i don't think there is a risk.

chia/wallet/wallet_node.py
arvidn: I'm not following this logic. Under what circumstance do we roll back 256 blocks?
That still means we can be closer than 300 blocks to the peak, in which case this flag did not used to be set. My impression is that this is distinguishing between long and short sync. is that not the case?

chia/wallet/wallet_state_manager.py
arvidn: oh, so we could just use a `bool` then?

chia/wallet/wallet_node.py
xdustinface: I wanted to keep it simple and extend the sync context in a later PR because i think there is a bunch of stuff we could move there. But i added the height yielding here fbbd632dec5d148c2717c18869033ca632abbb24 :)

chia/wallet/wallet_state_manager.py
xdustinface: But only in trusted case.. in untrusted case we still set it even if we are not far behind. In case `synced_peers` is empty here https://github.com/Chia-Network/chia-blockchain/blob/bdbfc810a4ebedee506605dfa313043c5b11690b/chia/wallet/wallet_node.py#L1147

chia/wallet/wallet_state_manager.py
xdustinface: We could, the reason i picked the height still was because im thinking about adding it in some follow up PR to the sync status in the RPC endpoint and since and since i anyway wanted to pass the height into the context to be able to log it i just did it like this now. Want to change it rather? I don't mind.

chia/wallet/wallet_state_manager.py
xdustinface: Its not that this gets called frequently but i wrapped them with log level checks in [edcb194](https://github.com/Chia-Network/chia-blockchain/pull/14524/commits/edcb1940dd8c41b5bedccaf9523b4f3fe3d23336). I have a branch where i work on some improvements to this status, like keeping track of how many times we synced up to the latest height, to allow dropping the secondary sync task which also includes making this cached values, then we can drop the log level checks again.

chia/wallet/wallet_node.py
arvidn: I think we should get a better understanding of what the intention was, and make less assumptions by just reading the code. Testing this patch and explore the different scenarios would help build confidence in the patch being correct. Someone clearly meant to not set this flag sometimes, for some reason. It seems odd, but it's risky to assume that it was the original author who was confused, and not us.

chia/wallet/wallet_state_manager.py
```diff
@@ -529,13 +528,39 @@ async def synced(self):
...
+
+    @asynccontextmanager
+    async def set_sync_mode(self, target_height: uint32) -> AsyncIterator[uint32]:
+        if self.log.level == logging.DEBUG:
+            self.log.debug(f"set_sync_mode enter {await self.blockchain.get_finished_sync_up_to()}-{target_height}")
```
arvidn, line 542: given that you log this before we acquire the lock, the actual start height may vary, right?
It seems like it would make sense to log this after the acquire the lock, and also only call `get_finished_sync_up_to()` just once, instead of twice.

arvidn (commented):
looks good. just one last question

chia/wallet/wallet_state_manager.py
xdustinface: The idea was to log the height before we try to acquire the lock and then again once we got it to see in the debug logs if there was something happening in between.

chia/wallet/wallet_node.py
xdustinface: It's not that i didn't try to understand the intention for this line. If you search for the usage of `sync_mode` you can see it's so far only used in wallet RPC in two places, one to show the sync status in `get_sync_status` and the other to show the cached/empty balances during sync in `get_wallet_balance`.

So basically when we are doing the long sync with the wallet state manager lock held but the sync mode not set we are blocking `get_wallet_balance` calls and with this the GUI until the sync is done.

You can just just restart your trusted wallet on main and with this PR to see the behaviour difference during the initial sync. The main points i figured are:

### On `main`

- The balance view in the GUI shows loading spinners all over the place during the initial sync
- The NFT page only shows a big loading spinner until the initial sync is done
- The connections view just shows a big loading spinner until the initial sync is done

### With this PR

- The balance view shows 0 during the initial sync
- The NFT page shows the existing NFTs during the initial sync with a "loading" indication in the corner
- The connections screen shows the connections properly during the initial sync

(updated the initial description with the above)

So basically the UI is blocked and shows loading spinners all over the place during syncing if we don't set the sync mode (like currently on `main`). In my case here this means a blocked UI on startup for at least 15s with my current wallet and with this PR it shows at least some of the available information right after start.

The 0 balances are probably improvable by another PR to maybe generate the cached balanced before the sync on startup but for now its at least the same behaviour like if we are >300 blocks behind on `main` so it should not lead to confusion or questions? I asked mariano about his intention and i report back if there was something important to it but i feel like it just makes sense to merge this either way?

xdustinface: > The 0 balances are probably improvable by another PR to maybe generate the cached balanced before the sync on startup 

Like here: #14631. If you apply this patch on top of this PR the wallet shows the last balance on start during the catchup sync also instead of showing just the loading spinners (`main`) or 0s (this PR only).

arvidn (dismissed):

xdustinface:
Rebased + tweaked the included tests a bit to simplify and use a different fixture to avoid adding another new one.

arvidn (approved):

PR #14525, opened 2023-02-08
[DO NOT MERGE] Review DAO chialisp

Quexington:
A few high level thoughts as well:
* **DAO treasury**: I'm fairly confident this can be implemented as the standard puzzle, except instead of a signature, it takes an announcement from specifically a proposal coin telling it what delegated puzzle it should run.  It seems like this is sort of the behavior it is exhibiting now, except there's a lot of different spend paths for everything you might possibly want the treasury to do.  This could be simplified/generalized by telling the treasury just a puzzle to run.

* **DAO proposal**: I don't think this has to be a singleton.  The singleton layer's job is to make sure that its unknown inner puzzle isn't trying to duplicate the singleton, but the proposal does not have an inner puzzle.  It just takes CAT votes and recreates itself.  There's nothing malicious that could be attempting to duplicate or melt the proposal.

* **DAO Lockup**: Not being able to trade these CATs while they're voting is pretty unfortunate.  I think it may be possible to have this lockup layer travel with the CAT without letting it double vote.  Assuming the stack (CAT -> Lockup -> innerpuz), then Lockup _should_ only have to do three things:
  * enforce that all child coins also get wrapped in Lockup
  * enforce that the sum of all child coins >= the current amount of the coin
  * allow adding/removal from a list of proposal IDs, announcing the ones that get added as they get added
this negates you from being able to melt this CAT for amount as part of a larger spend, but does still at least allow you to trade/rekey it while it has already voted on a proposal.  The lockup could keep track of multiple proposals that it has voted on too just like it does now.  And a proposal would only let CATs with Lockup wrappers vote on them (that way they could check whether or not they had their ID in the Lockup's list).

### DIFF
--- a/chia/wallet/puzzles/dao_lockup.clvm
+++ b/chia/wallet/puzzles/dao_lockup.clvm
@@ -41,9 +41,11 @@
     (if conditions
       (if (= (f (f conditions)) CREATE_COIN)  ; this guarantees that the new coin is obeying the rules - other coins are banned to avoid re-voting
         (if (= (f (r (f conditions))) vote_added_puzhash)
+          ; COMMENT: do we care? We're enforcing the amount is correct so what do we care if they make 0 CAT or use some value from a non voted CAT to make this?
           (if seen_vote  ; assert we haven't already made a coin with the new vote included
             (x)
             (if (= (f (r (r (f conditions)))) my_amount)  ; we vote with all our value
+              ; COMMENT: If I'm not mistaken this should be called "seen_any_vote", yeah?
               (if seen_no_vote  ; assert that we haven't already recreated ourself in some fashion
                 (x)
                 (c (f conditions) (check_conditions (r conditions) vote_added_puzhash my_amount message vote_amount my_puzhash 1 1))
@@ -59,6 +61,7 @@
           (x)
          )
         )
+        ; COMMENT: Should we do this within a namespace like CATs so that other announcements not in that namespace are allowed through?
         (if (= (f (f conditions)) CREATE_PUZZLE_ANNOUNCEMENT)  ; this secures the values used to generate message - other messages are banned in case of LIES
           (if (= (f (r (f conditions))) message)
             (c (f conditions) (check_conditions (r conditions) vote_added_puzhash my_amount message vote_amount my_puzhash seen_vote seen_no_vote))
--- a/chia/wallet/puzzles/dao_proposal.clvm
+++ b/chia/wallet/puzzles/dao_proposal.clvm
@@ -14,6 +14,7 @@
 
 
 (mod (
+  ; COMMENT: When using the modern compiler, you can destructure this like so (@ SINGLETON_STRUCT (SINGLETON_MOD_HASH SINGLETON_ID . LAUNCHER_PUZZLE_HASH))
   SINGLETON_STRUCT  ; (SINGLETON_MOD_HASH, (SINGLETON_ID, LAUNCHER_PUZZLE_HASH))
   PROPOSAL_MOD_HASH
   PROPOSAL_TIMER_MOD_HASH ; proposal timer needs to know which proposal created it, AND
@@ -45,6 +46,7 @@
   (defconstant dao_finished_state 0xfb015415f2e6a09c1141f880fc2135beec6adf2a19e4d02a191432846db16559)
 
   (defun-inline calculate_win_percentage (CURRENT_CAT_ISSUANCE PROPOSAL_PASS_PERCENTAGE)
+    ; COMMENT: this code would be clearer if the divmod was performed on the proposal_pass_percente/ten_thousand first, then multiplied to current_cat_issuance (maybe more optimal too not sure)
     (f (divmod (* CURRENT_CAT_ISSUANCE PROPOSAL_PASS_PERCENTAGE) TEN_THOUSAND))
   )
 
@@ -262,6 +264,7 @@
     (c
       (list CREATE_COIN dao_finished_state ONE (c (f (r SINGLETON_STRUCT)) 0))  ; move to announce finished mode. See previous comment about proposal amounts
       (c
+        ; COMMENT: should this number 5 be configurable?
         (list ASSERT_HEIGHT_RELATIVE 5)  ; soft close so all votes can get in
         (c
           (list ASSERT_PUZZLE_ANNOUNCEMENT  ; make sure that we actually have a matching treasury spend
@@ -307,10 +310,12 @@
               )
               (if
                 (all
+                  ; COMMENT: If this is used for both things, the function could use a different name probably
                   (> TOTAL_VOTES (calculate_win_percentage vote_coin_id_or_current_cat_issuance lockup_innerpuzhash_or_attendance_required))
                   (> YES_VOTES (calculate_win_percentage TOTAL_VOTES previous_votes_or_pass_margin))
                 )
                 (a INNERPUZ vote_amount_or_solution)  ; run_chialisp: run proposed chialisp ; XXX Matt, Geoff: should we also re-create ourself with previous_votes=0 in this case?
+                ; COMMENT: Is there even a reason to ping the treasury in this case? Nothing happens on it, it can just stay the same right?
                 (list (list CREATE_PUZZLE_ANNOUNCEMENT (sha256tree (list "u" 0))))  ; dont_run_chialisp: we dont run the chialisp, but we still move to the finished state
                                                                                     ;    and let the treasury know to remain the same
               )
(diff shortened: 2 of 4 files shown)


chia/wallet/puzzles/dao_proposal.clvm
```diff
@@ -14,6 +14,7 @@
```
Quexington, line 14: I'm actually nor sure this puzzle is doing a lineage proof to some original state which is bad news right because someone an just create an already completed proposal out of thin air?

matt-o-how:
The thing that I think you've missed is the aspect of the proposal timers.
The proposal timers are uniquely tied to the proposal (via singleton ID) and are only created when the proposal is spent with 0 votes in it. This prevents the attack where you create an already full proposal.

As for the treasury, I think we want it to be locked down. Providing certain information and certain restrictions. 
I agree in theory it *could* be a standard puzzle, but with DAOs we're going to want the treasury to be restricted and only usable in a very few specific ways.

Quexington:
> The thing that I think you've missed is the aspect of the proposal timers. The proposal timers are uniquely tied to the proposal (via singleton ID) and are only created when the proposal is spent with 0 votes in it. This prevents the attack where you create an already full proposal.
> 
> As for the treasury, I think we want it to be locked down. Providing certain information and certain restrictions. I agree in theory it _could_ be a standard puzzle, but with DAOs we're going to want the treasury to be restricted and only usable in a very few specific ways.

I think I did misunderstand that the timer was providing the proof that the proposal once upon a time had zero votes.  However, this can still be done without a singleton layer.  You can borrow the whole launcher/unique ID concept from singletons without paying all of their lineage proof & output policing overhead.  The main point is that you _know_ that a proposal coin is only going to create one output so why would you reveal and run chialisp that ensures that?

matt-o-how:
I get what you're saying, and have argued similar things to other people elsewhere, but I think the advantages of code re-use here combined with the large enough overlap of functionality make it worthwhile.

Quexington:
> As for the treasury, I think we want it to be locked down. Providing certain information and certain restrictions. I agree in theory it _could_ be a standard puzzle, but with DAOs we're going to want the treasury to be restricted and only usable in a very few specific ways.

The problem with this is that you're paying _massive_ overhead to have a puzzle that is far less featured and much harder to upgrade.  A proposal must dictate exactly what it wants the DAO to do regardless and requires a majority vote, I don't see why the DAO should have extra built in restrictions on what it can do.  It heavily limits the applications of a DAO without some changes to the treasury which is maybe riskier because now you have to re-write the sensitive puzzle that policies access to the DAO.

matt-o-how:
I think there is a valuable security in having the proposal run the proposed code and having the interface to the whole DAO be in limited ways.
The Treasury can still be changed, but the heavy limits are a feature not a bug. We want to be very protective of the money pot in my opinion.

trepca:
Let's take this offline and discuss further.

chia/wallet/puzzles/dao_lockup.clvm
```diff
@@ -41,9 +41,11 @@
     (if conditions
       (if (= (f (f conditions)) CREATE_COIN)  ; this guarantees that the new coin is obeying the rules - other coins are banned to avoid re-voting
         (if (= (f (r (f conditions))) vote_added_puzhash)
+          ; COMMENT: do we care? We're enforcing the amount is correct so what do we care if they make 0 CAT or use some value from a non voted CAT to make this?
```
matt-o-how, line 44: The spend cases here are:
- You are voting with all your value
- You are voting with part of your value

In both cases we want to make sure that you are only creating locked up cats, and that all of your value is accounted for.

chia/wallet/puzzles/dao_lockup.clvm
```diff
@@ -59,6 +61,7 @@
           (x)
          )
         )
+        ; COMMENT: Should we do this within a namespace like CATs so that other announcements not in that namespace are allowed through?
```
matt-o-how, line 64: Yes actually. I'll do that.

matt-o-how: That will help with trading locked DAO CATs too

chia/wallet/puzzles/dao_proposal.clvm
```diff
@@ -45,6 +46,7 @@
   (defconstant dao_finished_state 0xfb015415f2e6a09c1141f880fc2135beec6adf2a19e4d02a191432846db16559)
 
   (defun-inline calculate_win_percentage (CURRENT_CAT_ISSUANCE PROPOSAL_PASS_PERCENTAGE)
+    ; COMMENT: this code would be clearer if the divmod was performed on the proposal_pass_percente/ten_thousand first, then multiplied to current_cat_issuance (maybe more optimal too not sure)
```
matt-o-how, line 49: with our division flooring, I think it's more accurate and safer to calculate the big number first

chia/wallet/puzzles/dao_proposal.clvm
```diff
@@ -262,6 +264,7 @@
     (c
       (list CREATE_COIN dao_finished_state ONE (c (f (r SINGLETON_STRUCT)) 0))  ; move to announce finished mode. See previous comment about proposal amounts
       (c
+        ; COMMENT: should this number 5 be configurable?
```
matt-o-how, line 267: Yeah, this could be another value in the treasury actually

chia/wallet/puzzles/dao_proposal.clvm
```diff
@@ -307,10 +310,12 @@
               )
               (if
                 (all
+                  ; COMMENT: If this is used for both things, the function could use a different name probably
                   (> TOTAL_VOTES (calculate_win_percentage vote_coin_id_or_current_cat_issuance lockup_innerpuzhash_or_attendance_required))
                   (> YES_VOTES (calculate_win_percentage TOTAL_VOTES previous_votes_or_pass_margin))
                 )
                 (a INNERPUZ vote_amount_or_solution)  ; run_chialisp: run proposed chialisp ; XXX Matt, Geoff: should we also re-create ourself with previous_votes=0 in this case?
+                ; COMMENT: Is there even a reason to ping the treasury in this case? Nothing happens on it, it can just stay the same right?
```
matt-o-how, line 318: we still need the values that the treasury stores about things like "proposal length" and "pass requirements"

chia/wallet/puzzles/dao_treasury.clvm
```diff
@@ -134,9 +137,11 @@
     my_amount
     my_singleton_id
   )
+    ; COMMENT: this whole if tree can be on just the CREATE_COIN's puzhash since the amount and hint are the same in every case
     (if (= type 'u')  ; unchanged
       (list
         (list CREATE_COIN new_puzhash (+ my_amount new_amount_change) (c my_singleton_id 0))
+        ; COMMENT: this ends up double wrapping the singleton right? You need to recreate with the current inner puzzle not full puzzle.
```
matt-o-how, line 144: oh good catch

chia/wallet/puzzles/dao_treasury.clvm
```diff
@@ -208,14 +213,19 @@
...
       ; this is the add money spend case
+      ; COMMENT: This should probably be 0 or something configurable to prevent spaming the add money case a DOS
       (if (> new_amount_change -1)
         (merge_list
+          ; COMMENT: maybe this should be a loop so you can do more than one p2_singleton at once
           (if proposal_innerpuz_or_optional_p2_singleton_coin_id
+            ; COMMENT: An attacker here could give you a p2_singleton to this DAO, then specify an amount increase of 0 or 1 and melt the p2_singleton's value without actually giving it to the DAO
             (list
               (list CREATE_PUZZLE_ANNOUNCEMENT proposal_innerpuz_or_optional_p2_singleton_coin_id)
+              ; COMMENT: Using the existing p2_singleton here pretty much eliminates support for CATs or NFTs being sent to the DAO
```
matt-o-how, line 224: This isn't using a rigid p2_singleton mod, it's taking a coin ID in from the solution and just asserts that it is making the `$` announcement

chia/wallet/puzzles/dao_treasury.clvm
```diff
@@ -208,14 +213,19 @@
...
               (list ASSERT_COIN_ANNOUNCEMENT (sha256 proposal_innerpuz_or_optional_p2_singleton_coin_id '$'))
             )
             (list
+              ; COMMENT: should this be a coin announcement? Is it possible that you might not want to do this spend if someone else increases the amount before you?
```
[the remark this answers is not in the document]
Quexington: Lets say Alice sees the DAO is running a little low on funds and wants to contribute $1,000 to it.  But before she gets her spend through, Elon comes along and adds $1,000,000,000 to the treasury.  Does Alice still want her $1000 to be able to go through?

PR #14529, opened 2023-02-09
Offer CLI command help now shows --fee units as XCH

### LINKED ISSUE #14528: CLI: Update help for offer commands to show fee units
The CLI for offers doesn't indicate what units a `--fee` option should be specified in (mojos or XCH)

paninaro:
### Purpose:
Clarify in the CLI help how to specify a fee (as XCH or mojos) when using the offers CLI commands: make_offer, take_offer, cancel_offer.

chia wallet -h will now also indicate that NFTs can be part of offers in the 'offers' subcommand description.

Additionally, including a maker fee when using `chia wallet make_offer` will now include a line item in the offer summary. Previously this line was omitted.

### Current Behavior:
Currently the CLI help just states that --fee is used to attach a fee to the operation, without mentioning whether the fee should be specified in XCH or mojos.

### New Behavior:
The CLI help will now indicate "in XCH" for the fee option.

### Testing Notes:
Manually tested make_offer, take_offer, and cancel_offer with and without a fee specified.

### DIFF
--- a/chia/cmds/wallet.py
+++ b/chia/cmds/wallet.py
@@ -382,7 +382,7 @@ def add_token_cmd(wallet_rpc_port: Optional[int], asset_id: str, token_name: str
     asyncio.run(execute_with_wallet(wallet_rpc_port, fingerprint, extra_params, add_token))
 
 
-@wallet_cmd.command("make_offer", short_help="Create an offer of XCH/CATs for XCH/CATs")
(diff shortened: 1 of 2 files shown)


Quexington (approved):

arvidn (approved):

arvidn:
this patch demonstrates how easy we've made it to run the puzzle program many times. In this patch each call to `fees()` will run the puzzle to compute the output coins and subtract the spent coins.

I'm working on a separate patch to cache the results so we don't have to run it over and over again. So there's no need to make a change here.

PR #14533, opened 2023-02-09
bump chia_rs to 0.2.2

arvidn:
the new `chia_rs` version introduces a number of changes:

1. In order to support the new `ASSERT_BEFORE_*` conditions it adds two new fields to `SpendBundleConditions` and `Spend`, as the output from parsing conditions.
2. It adds validation of `MINTING_COIN` and `RESERVE_FEE` to the conditions parser. i.e. it is now ensured that the sum of spent coin amounts are greater than or equal to created coins (additions). Additionally, it's asserted that the `reserve_fee` conditions fit within difference between amount spent and amount added.
3. The `NO_NEG_DIV` flag that was used in the soft-fork transition to disallow division with negative numbers has been removed. This flag is now implied and cannot be specified.

The new conditions also come with new error codes and condition codes to support them.

This patch is best reviewed one commit at a time. Each of the points above are addressed individually. Specifically, the new validation for `MINTING_COIN` and `RESERVE_FEE` caused some tests to fail, that ended up running a spend bundle or block without ensuring it's spending a sufficiently large coin.


tests/pools/test_pool_puzzles_lifecycle.py
```diff
@@ -242,7 +242,7 @@ def test_pool_lifecycle(self):
         # filter for only the singleton solution
         singleton_coinsol: CoinSpend = list(
             filter(
-                lambda e: e.coin.amount == START_AMOUNT,
+                lambda e: e.coin.amount != START_AMOUNT,
```
arvidn, line 245: this is perhaps a bit unintuitive. the idea here is to pick the larger coin, to avoid `Err.MINTING_COIN` further down.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -346,10 +357,14 @@ async def get_coin_record(coin_id: bytes32) -> Optional[CoinRecord]:
 
     mempool_manager = await instantiate_mempool_manager(get_coin_record)
     conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, amount]]
-    result = await generate_and_add_spendbundle(mempool_manager, conditions)
-    _, status, error = result[2]
-    assert status == expected_status
-    assert error == expected_error
+    try:
+        result = await generate_and_add_spendbundle(mempool_manager, conditions)
```
arvidn, line 361: the reason this can throw now is because some failures (specifically `MINTING_COIN` and `RESERVE_FEE`) happen earlier, in prevalidate mempool item. Failures in that funcion throw, whereas failures in the main validation function return an error code.

AmineKhaldi (dismissed):
LGTM aside from the opportunity to simplify a couple tests (sent in another PR) and that unintuitive coin selection to avoid mint coin validation error that you pointed out.

arvidn:
> aside from the [...] that unintuitive coin selection to avoid mint coin validation error that you pointed out.

Do you think it's incorrect?

AmineKhaldi:
> > aside from [...] that unintuitive coin selection to avoid mint coin validation error that you pointed out.
> 
> Do you think it's incorrect?

No, I was referring to https://github.com/Chia-Network/chia-blockchain/pull/14533#discussion_r1101569795 which (as you pointed out) made me pause and peek. I found there is a second coin with amount `1750000000000` that we're picking now.

tests/pools/test_pool_puzzles_lifecycle.py
aqk: This would make a great comment

richardkiss: I'm not sure I understand this. What's failure case causes `Err.MINTING_COIN`?

richardkiss: I see, it refers to an imbalance of outgoing vs incoming that causes unapproved coins to be minted.

richardkiss (approved):
Seems reasonable.

AmineKhaldi (approved):

tests/pools/test_pool_puzzles_lifecycle.py
arvidn: exactly. the `START_AMOUNT` is 1023, but the dummy coin we try to create further down has amount 1337, which isn't allowed (and it used to not be checked this early)

trepca (approved):

PR #14535, opened 2023-02-09
Purposefully break offer compression for older versions

Quexington:
This results in more desirable behavior when old versions try to take new offers.  Immediate error rather than on chain failure.

### DIFF
--- a/chia/wallet/util/puzzle_compression.py
+++ b/chia/wallet/util/puzzle_compression.py
@@ -35,7 +35,7 @@ ZDICT = [
     + bytes(NFT_TRANSFER_PROGRAM_DEFAULT),
     bytes(CAT_MOD),
     bytes(OFFER_MOD),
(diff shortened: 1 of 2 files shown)


emlowe (dismissed):

chia/wallet/util/puzzle_compression.py
```diff
@@ -35,6 +35,7 @@
     + bytes(NFT_TRANSFER_PROGRAM_DEFAULT),
     bytes(CAT_MOD),
     bytes(OFFER_MOD),
+    b"break",  # purposefully break compatibility with older versions
```
arvidn, line 38: a minor suggestion I had was to make this an empty string. That way we won't be feeding any more (arbitrary) bytes into the zdict (potentially reducing compression ratio). It would also let use make this quirk disappear in the future, when we no longer support old versions

arvidn (approved):

PR #14536, opened 2023-02-09
No need to test install scripts on Mint

emlowe:
Testing the install scripts on Linux Mint has very little value and simply causes longer CI times along with more chances of flaky failure due to network issues during OS setup.

We are transitioning away from users running the install manually to more and more encouraging users to use the binary packages and Mint users can use the Ubuntu DEB package.

This distro is only minimally popular and there is no rationale for specific testing on Mint as opposed to other distros (Gentoo, etc)

### DIFF
--- a/.github/workflows/test-install-scripts.yml
+++ b/.github/workflows/test-install-scripts.yml
@@ -112,41 +112,6 @@ jobs:
           type: ubuntu
           # https://packages.ubuntu.com/jammy/python3 (22.04, 3.10)
           url: "docker://ubuntu:jammy"
-        - name: linuxmintd/mint19.1-amd64 (Tessa)
-          type: mint
-          # 3.6 default with an option for 3.7
-          url: "docker://linuxmintd/mint19.1-amd64"
-        - name: linuxmintd/mint19.2-amd64 (Tina)
-          type: mint
-          # 3.6 default with an option for 3.7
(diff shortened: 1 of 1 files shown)


altendky (changes requested):
As long as we are agreeing to not fix issues with source installs on Mint, alrighty.  Though supporting a platform but not supporting it for source leaves you in a position where you just sorta make guesses, wait for a build, and check the output since you can't source install.  So maybe we should just make Mint entirely unsupported?  Also, is there any code left in `install.sh` that is meant for Mint?  It should probably get removed.

emlowe:
There are any number of ubuntu-based distros that likely work with the packaged installer. We don't have to call out that we don't support them and we don't have to call out we do support them. We don't list Mint specifically anywhere as working in our wiki or docs.

There is only one minor reference to Mint in the install.sh script which explains why we use "20" instead of "20.04" in a version check - this is hardly worth changing.

altendky (approved):
I don't know about the others, but Mint now deviates even more than before with Mint 21 coming from Ubuntu 22 or something.  If we're dropping testing here for this awkward flavor of Ubuntu then I expect us to not maintain logic for the source install to work there.

wallentx (approved):

trepca (approved):

PR #14537, opened 2023-02-09
update black to 23.3.0

altendky:
### Draft For:
- [ ] coordination around merging to avoid as many crossed PRs as we can
- [x] separating leading blank line removal to https://github.com/Chia-Network/chia-blockchain/pull/14653 for clarity here
- [x] separating unneeded parenthesis removal to https://github.com/Chia-Network/chia-blockchain/pull/14654 for clarity here
- [x] separating removal of some commas to avoid new behavior in https://github.com/Chia-Network/chia-blockchain/pull/14655


arvidn:
It would seem intuitive to me to pin it to the most recent version rather than removing the pin altogether. Without pinning the version, CI may turn red when there's a new release of black, right?

altendky:
Yes, that is how we run test dependencies (and all indirect dependencies as well).  I don't want to maintain pins without help and 1) we don't pin properly now anyways as well as 2) it's not clear that we actually are really using automated update PR tooling reliably.  Any interest in learning poetry so we can reduce the bus factor around switching to that?  The PR is broken at the moment since we need pypi.chia.net for more things so I need to improve handling around that.

altendky:
Hmm, maybe the pinning was because of the use in pre-commit as well where to use the pre-commit preset you have to specify a version.  But, we have mypy at least in pre-commit just skipping the whole pre-commit thing and running from our env so maybe black should switch over as well.

arvidn:
I don't understand the argument about the cost of maintaining a pin. As I understand it the cost of not pinning is greater.
Pinning black means that we can put exactly as little effort into maintaining it as we'd like. We can upgrade and fixup in the same PR at our leisure.
Unpinning it means we will have broken CI sometimes and be forced to drop everything to adapt to a new version of black.

This is of course assuming that the marginal benefit we derive from running a newer version is less than the cost of breaking CI sometimes. I'm also assuming that the whole point of updating black is that it has improved some formatting, that may now disagree with what it did before, hence requiring fix-up.

altendky:
Had an separate discussion about the general topic.  tl;dr on the cost of not pinning, the default path is to not update and you end up stuck in dependency hell because when you update one thing you find issues brought in with newer versions of other things.

Anyways, to retain the present approach including pre-commit black, I'll go ahead and just pin to latest in both places.

altendky:
This now just pins the update in both the dependencies and pre-commit.  Now we'll see if the formatting changes are considered controversial.  I really didn't realize we made so many blank lines to start code blocks.

chia/full_node/weight_proof.py
```diff
@@ -527,7 +525,14 @@ def handle_block_vdfs(self, curr: HeaderBlock, blocks: Dict[bytes32, BlockRecord
...
             cc_sp_vdf_info = curr.reward_chain_block.challenge_chain_sp_vdf
             if not curr.challenge_chain_sp_proof.normalized_to_identity:
-                (_, _, _, _, cc_vdf_iters, _,) = get_signage_point_vdf_info(
+                (
+                    _,
+                    _,
+                    _,
+                    _,
+                    cc_vdf_iters,
+                    _,
```
arvidn, line 534: I'm not sure this is an improvement :/
maybe we need to remove the parentheses

arvidn (dismissed):

arvidn:
I suspect `main` will go red once this lands, and will need another follow up to fix PRs that landed in the meantime

chia/full_node/weight_proof.py
altendky: I would guess it's the trailing `,`.  Or maybe now regardless if the _whole_ thing doesn't fit on one line it does this.  I would also blame the ickiness here entirely on returning a tuple.

aqk:
I don't mind changing whitespace everywhere.

I don't like running different versions of tools on CI and dev machines.

We need to keep these tools pinned.

You might want to land the whitespace fixes first, then the black config changes - it's hard to separate them in the diff.

altendky:
> I don't like running different versions of tools on CI and dev machines.

Regardless of pinning or not, the way you do that at this point is to delete your venv and run the install script again.

> We need to keep these tools pinned.

The more interested you are in pinning the more you should make sure people take an interest in https://github.com/Chia-Network/chia-blockchain/pull/11057 and become familiar with poetry.

> You might want to land the whitespace fixes first, then the black config changes - it's hard to separate them in the diff.

There was only a version bump, no config change, but yes...  there was a _lot_ of changes here of a few particular types.  Most have been separated into:

- https://github.com/Chia-Network/chia-blockchain/pull/14653
- https://github.com/Chia-Network/chia-blockchain/pull/14654
- https://github.com/Chia-Network/chia-blockchain/pull/14655

arvidn (approved):

PR #14540, opened 2023-02-10
Remove ability for client code to specify wallet_id during wallet creation.

aqk:
Remove ability for client code to specify wallet_id during wallet creation. wallet_id is used as an external long-lived identifier in the wallet RPCs, and they must be unique.

### Purpose:

This reduces code complexity, and removes a dangerous internal interface. No externally visible changes.

### Current Behavior:

Allows specifying wallet_id during wallet creation.

### New Behavior:

Disallows specifying wallet_id during wallet creation.

### DIFF
--- a/chia/wallet/wallet_user_store.py
+++ b/chia/wallet/wallet_user_store.py
@@ -51,13 +51,12 @@ class WalletUserStore:
         name: str,
         wallet_type: int,
         data: str,
-        id: Optional[int] = None,
     ) -> WalletInfo:
 
         async with self.db_wrapper.writer_maybe_transaction() as conn:
             cursor = await conn.execute(
-                "INSERT INTO users_wallets VALUES(?, ?, ?, ?)",
-                (id, name, wallet_type, data),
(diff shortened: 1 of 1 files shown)


trepca (commented):
Thanks, can you also add a test for this?

emlowe:
We rely on sqlite and the column `id INTEGER PRIMARY KEY AUTOINCREMENT` - this instructs sqlite to make an alias between `id` and `rowid` 

> With AUTOINCREMENT, rows with automatically selected ROWIDs are guaranteed to have ROWIDs that have never been used before by the same table in the same database. And the automatically generated ROWIDs are guaranteed to be monotonically increasing

> If the AUTOINCREMENT keyword appears after INTEGER PRIMARY KEY, that changes the automatic ROWID assignment algorithm to prevent the reuse of ROWIDs over the lifetime of the database. In other words, the purpose of AUTOINCREMENT is to prevent the reuse of ROWIDs from previously deleted rows

There is, however, one important exception that I think we do hit: `ROWID values that were rolled back are ignored and can be reused.` - This can cause some conflicts between our in-memory cache and the DB

PR #14541, opened 2023-02-10
mypy chia.util.files

aqk:
### Purpose:

Code cleanup


arvidn (dismissed):

chia/util/files.py
```diff
@@ -71,8 +73,8 @@ async def write_file_async(file_path: Path, data: Union[str, bytes], *, file_mod
     mode: Literal["w+", "w+b"] = "w+" if type(data) == str else "w+b"
     temp_file_path: Path
     async with tempfile.NamedTemporaryFile(dir=file_path.parent, mode=mode, delete=False) as f:
-        temp_file_path = f.name  # type: ignore[assignment]
-        await f.write(data)  # type: ignore[arg-type]
+        temp_file_path = f.name
+        await f.write(data)
```
altendky, line 77: I twiddled with these just the other day and decided not to bother with them.  It's not obvious how to tie the mode chosen by the type of the incoming data to the type of file object we'll get out.  Plus other bits.  If you want to chat about it we can.  Otherwise feel free to just put the ignores back in.

https://github.com/Chia-Network/chia-blockchain/pull/14506/files

```suggestion
        temp_file_path = f.name  # type: ignore[assignment]
        await f.write(data)  # type: ignore[arg-type]
```

aqk: Thanks - I tried exactly this locally, but there seems to be a bug in `mypy`. When I replace these ignores, but remove the file from the mypy.ini ignores list, it gives "unused ignore statement" for those two lines.

I think I'll report this to mypy, and close this PR for now.

altendky: Are you sure you were running mypy v1.0 locally?  That's what introduced the complaints here.

altendky (dismissed):

chia/util/files.py
```diff
@@ -71,6 +73,8 @@ async def write_file_async(file_path: Path, data: Union[str, bytes], *, file_mod
     mode: Literal["w+", "w+b"] = "w+" if type(data) == str else "w+b"
     temp_file_path: Path
     async with tempfile.NamedTemporaryFile(dir=file_path.parent, mode=mode, delete=False) as f:
+        # Ignoring type error since it is not obvious how to tie the type of the data
+        # being passed in to the type of the file object, etc.
         temp_file_path = f.name  # type: ignore[assignment]
         await f.write(data)  # type: ignore[arg-type]
```
arvidn, line 79: don't you get a warning about "unused ignore" still?

arvidn: I assumed you wanted to make this change because of these:
```
chia/util/files.py:74: error: Unused "type: ignore" comment
chia/util/files.py:75: error: Unused "type: ignore" comment
Found 2 errors in 1 file (checked 748 source files)
```

altendky: With mypy < 1 you will get such a message.  With mypy 1 you will not.  As far as I know.  One of those transitional things where you need to make sure your env is updated.

aqk: @arvidn Unfortunately, we don't have the mypy version pinned, because they are still making fast progress, so you need the latest mypy in you venv.

altendky (approved):

emlowe (approved):

PR #14559, opened 2023-02-13
Simplify MINTING_COIN and RESERVE_FEE_CONDITION_FAILED unit tests

AmineKhaldi:


tests/core/mempool/test_mempool_manager.py
```diff
@@ -339,61 +339,21 @@ async def test_double_spend_prevalidation() -> None:
...
-        assert error == expected_error
-    except ValidationError as e:
-        assert expected_status == MempoolInclusionStatus.FAILED
-        assert e.code == expected_error
+async def test_minting_coin() -> None:
+    mempool_manager = await instantiate_mempool_manager(zero_calls_get_coin_record)
+    conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, TEST_COIN_AMOUNT + 1]]
+    sb = spend_bundle_from_conditions(conditions)
+    with pytest.raises(ValidationError, match="MINTING_COIN"):
+        await mempool_manager.pre_validate_spendbundle(sb, None, sb.name())
```
arvidn, line 347: as far as I can tell, the positive test was removed, your version only has a negative test. I think we should have positive tests too

tests/core/mempool/test_mempool_manager.py
```diff
@@ -339,61 +339,21 @@ async def test_double_spend_prevalidation() -> None:
...
-        assert error == expected_error
-    except ValidationError as e:
-        assert expected_status == MempoolInclusionStatus.FAILED
-        assert e.code == expected_error
+async def test_reserve_fee_condition() -> None:
+    mempool_manager = await instantiate_mempool_manager(zero_calls_get_coin_record)
+    conditions = [[ConditionOpcode.RESERVE_FEE, TEST_COIN_AMOUNT + 1]]
+    sb = spend_bundle_from_conditions(conditions)
+    with pytest.raises(ValidationError, match="RESERVE_FEE_CONDITION_FAILED"):
+        await mempool_manager.pre_validate_spendbundle(sb, None, sb.name())
```
arvidn, line 356: same here, I think we should have a positive test

arvidn (changes requested):

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Done.

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: Done.

arvidn (approved):

trepca:
This is more of a FYI, but as a chialisp developer, I always get confused by`MINTIN_COIN` error and need to dig through the code to understand what is the problem. It'd be great to rename it or provide additional error message like "Tried to create coins with more value than was spent" or something similar.

arvidn:
I'm in favor of renaming the error code. However, I can't think of a succinct way of saying the same thing. maybe `COIN_OUT_OF_THIN_AIR` or `ADDITIONS_EXCEED_REMOVALS`. I suppose one issue the current `MINTING_COIN` might suggest that there's a new coin being created, but it's really just the *value* that matters

arvidn (approved):

PR #14561, opened 2023-02-13
nft paging at sqlite level

trepca:
### Purpose:
`nft_get_nfts` is very slow on large NFT wallets as pagination is applied only after we load all NFTs from the database. This patch moves pagination to sqlite level. 

### Current Behavior:
Load all NFTs from db, then paginate.

### New Behavior:
Load only one page from sqlite.

### DIFF
--- a/chia/wallet/wallet_nft_store.py
+++ b/chia/wallet/wallet_nft_store.py
@@ -196,9 +196,9 @@ class WalletNftStore:
         if wallet_id is not None or did_id is not None:
             sql += " and"
         sql += " removed_height is NULL"
-        sql += " LIMIT %s OFFSET %s " % (count, start_index)
+        sql += " LIMIT ? OFFSET ?"
         async with self.db_wrapper.reader_no_transaction() as conn:
-            rows = await conn.execute_fetchall(sql)
+            rows = await conn.execute_fetchall(sql, (count, start_index))
 
         return [
             NFTCoinInfo(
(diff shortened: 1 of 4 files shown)


Quexington (dismissed):

paninaro (dismissed):
lgtm

chia/wallet/wallet_nft_store.py
```diff
@@ -183,6 +196,7 @@ async def get_nft_list(
         if wallet_id is not None or did_id is not None:
             sql += " and"
         sql += " removed_height is NULL"
+        sql += " LIMIT %s OFFSET %s " % (count, start_index)
```
arvidn, line 199: it makes me nervous when we use string formatting and concatenation to generate SQL queries. Over time I think we should phase that out in favor of using parameterized sql statements.

Also, you use `%s` here, but `count` and `start_index` are integers, aren't they?
Given that python can't guarantee they be integers, I think this is actually quite risky, inviting sql injection attacks

arvidn (changes requested):

chia/wallet/wallet_nft_store.py
trepca: right, I do convert user input from RPC call into int so injection attack from that is not possible, but can convert into params in any case.

arvidn (approved):

steppsr:
Since the CLI is using the RPC command, setting the num to 50 in the except now causes the CLI to only return a max of 50 NFTs and no way to know which 50 from your wallet. This isn't in the help for the CLI command 'chia wallet nft list'.

PR #14565, opened 2023-02-14
start fixing up windows hints

altendky:

### DIFF
--- a/chia/server/chia_policy.py
+++ b/chia/server/chia_policy.py
@@ -161,7 +161,6 @@ class PausableServer(BaseEventsServer):
         """Pause future calls to accept()."""
         self._paused = True
         if sys.platform == "win32" and sys.version_info >= (3, 8):
-            assert isinstance(self._loop, ChiaProactorEventLoop)
             # proactor
             self._loop.disable_connections()
         else:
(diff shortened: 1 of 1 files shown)


chia/server/chia_policy.py
```diff
@@ -275,7 +275,7 @@ async def _chia_accept_loop(self, listener: socket.socket) -> Tuple[socket.socke
 
         def _chia_accept(self, listener: socket.socket) -> asyncio.Future[Tuple[socket.socket, Tuple[object, ...]]]:
             self._register_with_iocp(listener)
-            conn = self._get_accept_socket(listener.family)
+            conn = self._get_accept_socket(listener.family)  # pylint: disable=assignment-from-no-return
```
arvidn, line 277: is mypy mistaken in thinking that `_get_accept_socket()` doesn't return anything?

arvidn:
looks reasonable so far. there seem to be some legitimate failures in CI: https://github.com/Chia-Network/chia-blockchain/actions/runs/4169394633/jobs/7218892064#step:16:239

chia/server/chia_policy.py
altendky: pylint is the confused one.  It doesn't special case the `typing.TYPE_CHECKING` afaik so would treat it as the `False` it is coded to for runtime.  This means it misses all of our hinting shims above.  There may be more to look at here.

altendky:
otherwise made their way into https://github.com/Chia-Network/chia-blockchain/pull/14575

PR #14567, opened 2023-02-14
parameterize TestConditions

arvidn:
This is cleaning up the tests in preparation for extending them with `ASSERT_BEFORE_*` tests.

### DIFF
--- a/tests/core/full_node/test_conditions.py
+++ b/tests/core/full_node/test_conditions.py
@@ -6,7 +6,6 @@ or that they're failing for the right reason when they're invalid.
 from __future__ import annotations
 
 import logging
-import time
 from typing import List, Optional, Tuple
 
 import pytest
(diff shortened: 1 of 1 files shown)


tests/core/full_node/test_conditions.py
```diff
@@ -148,20 +157,6 @@ async def test_valid_my_id(self, bt):
         conditions = Program.to(assemble(f"(({ConditionOpcode.ASSERT_MY_COIN_ID[0]} 0x{coin.name().hex()}))"))
         await check_conditions(bt, conditions)
 
-    @pytest.mark.asyncio
-    async def test_invalid_seconds_absolute(self, bt):
-        # TODO: make the test suite not use `time.time` so we can more accurately
-        # set `time_now` to make it minimal while still failing
```
arvidn, line 154: I fixed this TODO by forcing the time stamps on the generated blocks (see `genesis_timestamp` in the call to `get_consecutive_blocks()` above). Using the wall clock time won't work in the parameters because the name of the test would then depend on the wall clock time, which doesn't work when running tests in parallel

AmineKhaldi (approved):
LGTM thanks.

trepca (approved):

PR #14570, opened 2023-02-14
rpc endpoint to count nfts per wallet or all wallets

trepca:
### Purpose:
Add RPC endpoint to count how many NFTs a wallet has so frontend can prepare when loading 

### Current Behavior:
No RPC to count.


paninaro (approved):
lgtm

arvidn (approved):

PR #14572, opened 2023-02-14
wallet: Some rework/cleanup of weight proof validation

xdustinface:
### Purpose:

See the individual commits.

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/wallet/wallet_weight_proof_handler.py
+++ b/chia/wallet/wallet_weight_proof_handler.py
@@ -6,7 +6,7 @@ import tempfile
 import time
 from concurrent.futures.process import ProcessPoolExecutor
 from multiprocessing.context import BaseContext
-from typing import IO, List, Optional, Tuple
+from typing import IO, List, Optional
 
 from chia.consensus.block_record import BlockRecord
 from chia.consensus.constants import ConsensusConstants
@@ -37,13 +37,8 @@ class WalletWeightProofHandler:
             initializer=setproctitle,
             initargs=(f"{getproctitle()}_worker",),
         )
-        self._weight_proof_tasks: List[asyncio.Task[Tuple[bool, List[BlockRecord]]]] = []
 
     def cancel_weight_proof_tasks(self) -> None:
-        for task in self._weight_proof_tasks:
-            if not task.done():
-                task.cancel()
-        self._weight_proof_tasks = []
         self._executor_shutdown_tempfile.close()
         self._executor.shutdown(wait=True)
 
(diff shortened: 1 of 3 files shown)


arvidn (approved):
looks good!
I seem to recall @almog had reviewed this patch too. Did you end up moving it to a new PR?

xdustinface:
> I seem to recall @almog had reviewed this patch too. Did you end up moving it to a new PR?

Nope nothing was moved here. This PR was just draft for a while because i waited for #14574 to get merged first and i don't think @almogdepaz saw this patch here yet.

chia/wallet/wallet_weight_proof_handler.py
```diff
@@ -56,22 +51,17 @@ async def validate_weight_proof(
...
+            self._num_processes,
+            weight_proof,
+            summaries,
+            sub_epoch_weight_list,
+            skip_segment_validation,
+            validate_from,
         )
-        self._weight_proof_tasks.append(task)
-        valid, block_records = await task
-        self._weight_proof_tasks.remove(task)
```
arvidn, line 74: this transformation is only valid if there's only ever a single task that gets here. Is that a safe assumption?

chia/wallet/wallet_weight_proof_handler.py
```diff
@@ -37,13 +37,8 @@ def __init__(
             initializer=setproctitle,
             initargs=(f"{getproctitle()}_worker",),
         )
-        self._weight_proof_tasks: List[asyncio.Task[Tuple[bool, List[BlockRecord]]]] = []
 
     def cancel_weight_proof_tasks(self) -> None:
-        for task in self._weight_proof_tasks:
-            if not task.done():
-                task.cancel()
-        self._weight_proof_tasks = []
```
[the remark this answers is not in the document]
xdustinface: Yes the weight proof validation runs as part of the new peak processing / long sync which gets canceled here https://github.com/Chia-Network/chia-blockchain/blob/5b3343b90e38c02c07d9a6387899c6d5e97f370e/chia/wallet/wallet_node.py#L407 so there is no need to pass down the cancelation to some additionally created tasks since the asyncio cancelation will take care of it yes.

chia/wallet/wallet_weight_proof_handler.py
xdustinface: There will always only be one task in there yes because we only long sync to one peer at the time and the validation only happens as part of the long sync. But i think even if there could potentially be more tasks added i think it would be safe to drop this because we add/await/remove still or why do you think it would not be safe?

PR #14573, opened 2023-02-14
Add unit tests for process_mempool_items

AmineKhaldi:


tests/core/mempool/test_mempool_manager.py
```diff
@@ -539,3 +541,131 @@ async def test_get_items_not_in_filter() -> None:
...
+    mempool_manager = await instantiate_mempool_manager(get_coin_record_for_test_coins)
+    conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, 1]]
+    _, _, result = await generate_and_add_spendbundle(mempool_manager, conditions)
+    expected_cost = uint64(2897056)
+    assert result == (expected_cost, MempoolInclusionStatus.SUCCESS, None)
+    spend_bundles, cost_sum, additions, removals = mempool_manager.process_mempool_items(item_inclusion_filter=never)
+    assert spend_bundles == []
+    assert cost_sum == 0
+    assert additions == []
+    assert removals == []
```
arvidn, line 560: it's hard to tell, but it looks like there's a fair amount of duplication between these tests. did you consider making it fewer tests that are parameterized?
It's also not clear what each of these are testing. I think comment describing the test case (unless it's obvious) would be helpful.
For example, the expected cost are large numbers, how can a reader of this code be confident these costs are correct?
if they are made up of multiple different terms, it would make sense to split them apart.

AmineKhaldi: Each of these tests cover a scenario that we go through, so there is no duplication between the tests. I added more comments to further describe each scenario.

arvidn: I was not suggesting that the they tested the same case, but that if the differences between the cases are small enough, those could be made parameters to the test, to make it easier to see which cases are tested, and highlight the important parts.

AmineKhaldi: I made the max cost/fee tests more focused and used custom coins, to highlight the covered code paths.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -534,3 +535,89 @@ async def test_get_items_not_in_filter() -> None:
...
+            assert removals == [coin1, coin2]
+        else:
+            # The second item has a higher FPC so it should get picked first
+            assert spend_bundles == [sb2]
+            # The first item hits the maximum fee/block clvm cost and gets skipped
+            assert cost_sum == expected_cost2
+            if test_for == "max_cost":
+                assert additions == [Coin(coin2.name(), IDENTITY_PUZZLE_HASH, coin2_amount - 2)]
+            elif test_for == "max_fee":
+                assert additions == [Coin(coin2.name(), IDENTITY_PUZZLE_HASH, 2)]
```
aqk, line 622:
```suggestion
                assert additions == [Coin(coin2.name(), IDENTITY_PUZZLE_HASH, 2)]
            else:
                assert False, "not all cases covered"
```

aqk: In case someone adds more later.

AmineKhaldi: Ah thanks, that was why I was explicit with the elif but this is better.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -534,3 +535,89 @@ async def test_get_items_not_in_filter() -> None:
...
+def always(mm: MempoolManager, mi: MempoolItem) -> bool:
+    return True
+
+
+def never(mm: MempoolManager, mi: MempoolItem) -> bool:
+    return False
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("test_for", ["max_cost", "max_fee", "normal_path"])
```
aqk, line 549: Not needed, but you could make these an enum.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -534,3 +535,89 @@ async def test_get_items_not_in_filter() -> None:
...
+@pytest.mark.parametrize("inclusion_filter", [always, never])
+async def test_process_mempool_items(
+    test_for: str, inclusion_filter: Callable[[MempoolManager, MempoolItem], bool]
+) -> None:
+    coin1_amount = uint64(10000000)
```
aqk, line 554: Since you are testing multiple cases in this test, it would be a little easier to read if you defined the expected outputs up here. eg.

```
coin1_child = Coin(coin1.name(), IDENTITY_PUZZLE_HASH, coin1_amount - 42)
```

tests/core/mempool/test_mempool_manager.py

tests/core/mempool/test_mempool_manager.py
```diff
@@ -534,3 +535,93 @@ async def test_get_items_not_in_filter() -> None:
...
+        else:
+            conditions = [[ConditionOpcode.CREATE_COIN, IDENTITY_PUZZLE_HASH, coin2_amount - 2]]
+            coin2_child = Coin(coin2.name(), IDENTITY_PUZZLE_HASH, coin2_amount - 2)
+            expected_cost2 = uint64(3065056)
+        sb2, _, result = await generate_and_add_spendbundle(mempool_manager, conditions, coin2)
+        assert result == (expected_cost2, MempoolInclusionStatus.SUCCESS, None)
+        spend_bundles, cost_sum, additions, removals = mempool_manager.process_mempool_items(
+            item_inclusion_filter=inclusion_filter
+        )
+        if test_for == TestFor.NORMAL_PATH:
```
arvidn, line 613: it not obvious what benefits you have from the parameterization here. it seems to obfuscate the fact that inclution_filter=never and test_for=NORMAL_PATH is redundant. In other words, these test cases aren't orthogonal.

typically, if you have different cases with different expected results, the result would also be a parameter.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -534,3 +535,93 @@ async def test_get_items_not_in_filter() -> None:
...
+
+class TestFor(Enum):
+    MAX_COST = 0
+    MAX_FEE = 1
+    NORMAL_PATH = 2
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("test_for", [t for t in TestFor])
+@pytest.mark.parametrize("inclusion_filter", [always, never])
```
arvidn, line 556: I don't think testing the inclusion filter is important. We only use it in one test, never in production

tests/core/mempool/test_mempool_manager.py
```diff
@@ -534,3 +535,93 @@ async def test_get_items_not_in_filter() -> None:
...
+        assert len(additions) == 0
+        assert len(removals) == 0
+    elif inclusion_filter == always:
+        # Make sure this mempool item was included by item_inclusion_filter
+        assert spend_bundles == [sb1]
+        assert cost_sum == expected_cost1
+        assert additions == [coin1_child]
+        assert removals == [coin1]
+        # Create a second item that when combined with the first one we exceed the maximum fee/block clvm cost
+        if test_for == TestFor.MAX_FEE:
```
arvidn, line 600: it looks like inclusion_filter=never and test_for=MAX_FEE also is redundant

tests/core/mempool/test_mempool_manager.py
```diff
@@ -534,3 +535,93 @@ async def test_get_items_not_in_filter() -> None:
...
+
+
+class TestFor(Enum):
+    MAX_COST = 0
+    MAX_FEE = 1
+    NORMAL_PATH = 2
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("test_for", [t for t in TestFor])
```
arvidn, line 555: these appear to be 3 independent tests. not really ideal to parameterize like this. It would be different if they would run through the same code and test some expected outcome (that can be specified as a parameter, and vary across the test cases).

arvidn (commented):
I think it would be better to test `create_bundle_from_mempool()` as that's the public facing function. `process_mempool_items()` looks like an internal function (which also seems a bit unnecessary).

I would think the most interesting cases to test would be:

* do we actually pick the transactions with the highest fee per cost?
* do we cap the number of items we pick to stay within our cost limit?

there are other interesting test cases that you cover here, but they are tests for `add_spend_bundle()`, such as:
* do we reject spend bundles with too high a cost?
* do we reject spend bundles with too high a fee?

There are a number of other interesting cases such as does the filter we generate accurately represent the items in our mempool?

AmineKhaldi:
> 

Created in https://github.com/Chia-Network/chia-blockchain/pull/14700

PR #14574, opened 2023-02-14
wallet: Drop leaking weight proof validation cache

xdustinface:
### Purpose:

Fix memory leak by dropping the leaking weight proof validation cache `WalletNode.valid_wp_cache`. 

### Current Behavior:

We currently add 2 new cache entries (~2MB per entry) every time we long sync to untrusted peers (primary + secondary) which might happen whenever depending on the setup/network because we always long sync from new connections if we lost connection to synced untrusted peers. Currently we don't remove those entries anywhere so the weight proof validation cache may just keep growing without limit.

### New Behavior:

Initially i thought about making it an LRU with a limit of 2 entries, see 0784dc8e5c35208523e20e987aa674beabdfb229. But then after thinking more about it.. it seems to me that we never even would get a cache hit here because we never end up requesting the weight proof for the same peak twice and each peak leads to a different weight proof because of the `recent_chain_data`? Let me know if i miss something, i can drop 93df1f81b099fabe1dc23dc1d1a6dd7e8a1e83c5 and maybe even adjust the size if we still need the cache for some reason.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -48,7 +48,6 @@ from chia.util.db_wrapper import manage_connection
 from chia.util.errors import KeychainIsEmpty, KeychainIsLocked, KeychainKeyNotFound, KeychainProxyConnectionFailure
 from chia.util.ints import uint32, uint64
 from chia.util.keychain import Keychain
-from chia.util.lru_cache import LRUCache
 from chia.util.path import path_from_root
 from chia.util.profiler import mem_profile_task, profile_task
 from chia.wallet.transaction_record import TransactionRecord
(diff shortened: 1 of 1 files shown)


almogdepaz:
thats true this cache has no meaning maybe if we concurrently get the same peak or something

almogdepaz (dismissed):

almogdepaz:
we can cache the latest chain blocks and validation and avoid that so we only validate the latest added block.
so we still validate the sub_epoch segments because i think we cant really write a good cache strategy for that but lets say we want to do block inclusion for a block that is hight + 10 from our latest WP we would only need to validate 10 block out of the local chain

almogdepaz (approved):

almogdepaz:
> > we can cache the latest chain blocks and validation and avoid that so we only validate the latest added block.
> > so we still validate the sub_epoch segments because i think we cant really write a good cache strategy for that but lets say we want to do block inclusion for a block that is hight + 10 from our latest WP we would only need to validate 10 block out of the local chain
> 
> @almogdepaz Sounds like this involves more changes. Can you come up with a PR for this in the near future? If not i think it still makes sense to merge this PR here because it fixes a memory leak.

i approved since my suggested changes are just for validation optimization and this cache is meaningless as far as i can see

arvidn:
@xdustinface did you smoke-test your theory that there's never a cache hit? you could set a breakpoint or assert if there's a cache hit and test to sync

xdustinface:
> @xdustinface did you smoke-test your theory that there's never a cache hit? you could set a breakpoint or assert if there's a cache hit and test to sync

I tested yes and you can just follow to logic to see it also, we don't call this for the same peak twice.

arvidn (approved):

PR #14581, opened 2023-02-14
fix typo introduced in bumping of chia-rs dependency

arvidn:
These error codes are supposed to match the corresponding error codes in `chia_rs`, that were introduced recently.
See: https://github.com/Chia-Network/chia_rs/blob/main/src/gen/validation_error.rs#L98


richardkiss (approved):
Oops

trepca (approved):

PR #14584, opened 2023-02-15
Ignore unused config value in the DataLayer CLI commands

AmineKhaldi:


altendky (approved):
Looks fine.  For the record, I would like to see things return class instances instead of tuples.

emlowe (approved):

PR #14586, opened 2023-02-15
Move imports to the top of the file

AmineKhaldi:


arvidn:
is there a good reason to make this change?
Given that this is a 3rd party file we've included, I think we should try to preserve it in its original form

AmineKhaldi:
> is there a good reason to make this change? Given that this is a 3rd party file we've included, I think we should try to preserve it in its original form

Ruff flagged it.

arvidn (approved):

PR #14588, opened 2023-02-15
add nft id to rpcs

altendky:
- full node optimizations (#14562)
- fixup event loop method testing (#14566)
- add the nft id to the rpc returns

### Draft For:
- [x] not 'including' those other commits, i'm not clear why they still show
- [x] consideration of implementation, like does the id really need recalculated or is it already available somewhere at that point?


wallentx:
I want

paninaro (approved):
lgtm. Thanks for adding this. The GUI has been computing this property, so this will allow us to remove that code.

PR #14591, opened 2023-02-16
server: Introduce `ChiaServer.trusted_peers`

xdustinface:
### Purpose:

- Improve trusted peers handling by storing the parsed trusted peers in the server vs always pass in the config entry to `ChiaServer.is_trusted_peer`
- Add some tests for trusted peers config handling
- Cleanup a bunch of not required assignments in tests - all the `config["trusted_peers"] = {}`

### New Behavior:

No change in behaviour.

### DIFF
--- a/tests/core/util/test_file_keyring_synchronization.py
+++ b/tests/core/util/test_file_keyring_synchronization.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import os
+import time
 from multiprocessing import Pool
 from pathlib import Path
 from sys import platform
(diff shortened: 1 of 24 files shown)


tests/core/util/test_file_keyring_synchronization.py
```diff
@@ -62,13 +61,14 @@ def dummy_set_passphrase(service, user, passphrase, keyring_path, index, num_wor
 
 class TestFileKeyringSynchronization:
     # When: using a new empty keyring
+    @pytest.mark.benchmark
```
altendky, line 64: Let's make sure we talk through this before tossing stuff on the benchmark runner for reasons other than them being tests verifying performance.

xdustinface: Im not really intending to merge it like this, its still a draft :)

PR #14592, opened 2023-02-16
server: Drop outdated config update

xdustinface:
### Purpose:

Drop some unused code, it just seems to me that we don't need this anymore, its coming from here #1598.

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/server/start_wallet.py
+++ b/chia/server/start_wallet.py
@@ -39,12 +39,6 @@ def create_wallet_service(
 
     overrides = service_config["network_overrides"]["constants"][service_config["selected_network"]]
     updated_constants = consensus_constants.replace_str_to_bytes(**overrides)
-    # add local node to trusted peers if old config
-    if "trusted_peers" not in service_config:
-        full_node_config = config["full_node"]
-        trusted_peer = full_node_config["ssl"]["public_crt"]
-        service_config["trusted_peers"] = {}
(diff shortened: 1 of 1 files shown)


almogdepaz:
how do we know if there are still users without this config entry ?

or so we just not care at this point and they should upgrade ?

xdustinface:
> how do we know if there are still users without this config entry ?

Thanks, i checked the accesses to `trusted_peers` and just opened #14643 but other than that i don't see why adding this `local_node` entry still would help someone? I might be confused.. if there is still something let me know what scenario you are thinking about!

almogdepaz:
> > how do we know if there are still users without this config entry ?
> 
> Thanks, i checked the accesses to `trusted_peers` and just opened #14643 but other than that i don't see why adding this `local_node` entry still would help someone? I might be confused.. if there is still something let me know what scenario you are thinking about!

they still would not have the default trusted node after #14643 no ?

xdustinface:
> they still would not have the default trusted node after #14643 no ?

Whats the point of having it if it's not even used? I guess it would make sense if you tell the exact scenario which could lead to issues here.

emlowe:
This code seems all wrong anyway, as it's using the old "point to certificate file" way to identify trusted peers which we changed to use the node-id, so even if you had this setting, nothing checks it correctly.

So I think this is safe to remove. We use the loopback address to trust the local node

emlowe (approved):

PR #14593, opened 2023-02-16
Mempool private implementation

arvidn:
This patch prepares the `Mempool` class for a new implementation.
This step is to make the current implementation details private and provide a public interface. The key changes are in the `mempool.py` file, and all other changes are made to use the new public interface rather than accessing members directly.

One exception is the tests where we used to compare spend bundles using the `is` operator rather than the `==` operator. This is technically unrelated to the changes made to the `Mempool` class, except in that the `Mempool` interface should not guarantee that objects are the same, just that values are the same.

My long-term ambition is:
1. replace the `Mempool` implementation with an in-memory sqlite database
2. have `Mempool` absorb the existing conflict- and pending caches (their data structures will be simpler to implement on top of a DB)
3. implement the new `ASSERT_BEFORE_*` conditions in the new `Mempool`, which depends on being able to add an index for expiration heights and timestamps.

Strictly speaking, (2) does not necessarily need to happen before (3).

### DIFF
--- a/chia/clvm/spend_sim.py
+++ b/chia/clvm/spend_sim.py
@@ -14,6 +14,7 @@ from chia.consensus.cost_calculator import NPCResult
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
 from chia.full_node.bundle_tools import simple_solution_generator
 from chia.full_node.coin_store import CoinStore
+from chia.full_node.mempool import Mempool
 from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions, get_puzzle_and_solution_for_coin
 from chia.full_node.mempool_manager import MempoolManager
 from chia.types.blockchain_format.coin import Coin
@@ -225,9 +226,8 @@ class SpendSim:
     ) -> Tuple[List[Coin], List[Coin]]:
         # Fees get calculated
         fees = uint64(0)
-        if self.mempool_manager.mempool.spends:
-            for _, item in self.mempool_manager.mempool.spends.items():
-                fees = uint64(fees + item.spend_bundle.fees())
+        for item in self.mempool_manager.mempool.all_spends():
+            fees = uint64(fees + item.fee)
 
         # Rewards get created
         next_block_height: uint32 = uint32(self.block_height + 1) if len(self.block_records) > 0 else self.block_height
@@ -251,7 +251,7 @@ class SpendSim:
         generator_bundle: Optional[SpendBundle] = None
         return_additions: List[Coin] = []
         return_removals: List[Coin] = []
-        if (len(self.block_records) > 0) and (self.mempool_manager.mempool.spends):
+        if (len(self.block_records) > 0) and (self.mempool_manager.mempool.size() > 0):
             peak = self.mempool_manager.peak
             if peak is not None:
                 result = self.mempool_manager.create_bundle_from_mempool(peak.header_hash, item_inclusion_filter)
@@ -300,7 +300,8 @@ class SpendSim:
         self.block_records = new_br_list
         self.blocks = new_block_list
         await self.coin_store.rollback_to_block(block_height)
-        self.mempool_manager.mempool.spends = {}
+        old_pool = self.mempool_manager.mempool
+        self.mempool_manager.mempool = Mempool(old_pool.mempool_info, old_pool.fee_estimator)
         self.block_height = block_height
         if new_br_list:
             self.timestamp = new_br_list[-1].timestamp
@@ -429,12 +430,12 @@ class SimClient:
             return CoinSpend(coin_record.coin, puzzle, solution)
 
     async def get_all_mempool_tx_ids(self) -> List[bytes32]:
-        return list(self.service.mempool_manager.mempool.spends.keys())
+        return self.service.mempool_manager.mempool.all_spend_ids()
 
     async def get_all_mempool_items(self) -> Dict[bytes32, MempoolItem]:
         spends = {}
-        for tx_id, item in self.service.mempool_manager.mempool.spends.items():
-            spends[tx_id] = item
+        for item in self.service.mempool_manager.mempool.all_spends():
+            spends[item.name] = item
         return spends
 
     async def get_mempool_item_by_tx_id(self, tx_id: bytes32) -> Optional[Dict[str, Any]]:
--- a/chia/full_node/mempool.py
+++ b/chia/full_node/mempool.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import logging
 from datetime import datetime
 from enum import Enum
 from typing import Dict, List, Optional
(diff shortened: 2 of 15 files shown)


chia/clvm/spend_sim.py
```diff
@@ -225,9 +226,8 @@ async def farm_block(
     ) -> Tuple[List[Coin], List[Coin]]:
         # Fees get calculated
         fees = uint64(0)
-        if self.mempool_manager.mempool.spends:
-            for _, item in self.mempool_manager.mempool.spends.items():
-                fees = uint64(fees + item.spend_bundle.fees())
+        for item in self.mempool_manager.mempool.all_spends():
+            fees = uint64(fees + item.spend_bundle.fees())
```
AmineKhaldi, line 230: Should be just `fees = uint64(fees + item.fee)`

chia/clvm/spend_sim.py
```diff
@@ -429,12 +430,12 @@ async def get_puzzle_and_solution(self, coin_id: bytes32, height: uint32) -> Opt
             return CoinSpend(coin_record.coin, puzzle, solution)
 
     async def get_all_mempool_tx_ids(self) -> List[bytes32]:
-        return list(self.service.mempool_manager.mempool.spends.keys())
+        return list(self.service.mempool_manager.mempool.all_spend_ids())
```
AmineKhaldi, line 433: Should be just `return self.service.mempool_manager.mempool.all_spend_ids()`

chia/full_node/mempool_manager.py
```diff
@@ -366,8 +362,9 @@ async def add_spend_bundle(
         """
 
         # Skip if already added
-        if spend_name in self.mempool.spends:
-            cost: Optional[uint64] = self.mempool.spends[spend_name].cost
+        existing_item = self.mempool.get_spend_by_id(spend_name)
+        if existing_item is not None:
+            cost: Optional[uint64] = existing_item.cost
             assert cost is not None
             return uint64(cost), MempoolInclusionStatus.SUCCESS, None
```
AmineKhaldi, line 369: Mempool items don't have an optional cost so this should be simplified by using `existing_item.cost` directly.

chia/full_node/mempool.py
```diff
@@ -24,26 +23,58 @@ class MempoolRemoveReason(Enum):
...
+        return list(self._spends.values())
+
+    def all_spend_ids(self) -> List[bytes32]:
+        return list(self._spends.keys())
+
+    def spends_by_feerate(self) -> List[MempoolItem]:
+        ret: List[MempoolItem] = []
+        for spends_with_fpc in reversed(self._sorted_spends.values()):
+            ret.extend(spends_with_fpc.values())
+        return ret
```
AmineKhaldi, line 50: For these and the ones below, I'd vote for taking this opportunity to properly name these with "mempool items" instead of "spends".

chia/full_node/mempool.py
```diff
@@ -24,26 +23,58 @@ class MempoolRemoveReason(Enum):
...
+
+    def size(self) -> int:
+        return len(self._spends)
+
+    def get_spend_by_id(self, spend_bundle_id: bytes32) -> Optional[MempoolItem]:
+        return self._spends.get(spend_bundle_id, None)
+
+    def get_spends_by_coin_id(self, spent_coin_id: bytes32) -> List[MempoolItem]:
```
AmineKhaldi, line 58: Mempool item oriented naming opportunity as mentioned above.

chia/full_node/mempool_manager.py
```diff
@@ -614,24 +612,19 @@ async def new_peak(
...
-                            if item:
-                                included_items.append(item)
-                            self.remove_seen(spendbundle_id)
-                self.mempool.remove_from_pool(spendbundle_ids_to_remove, MempoolRemoveReason.BLOCK_INCLUSION)
+                    items: List[MempoolItem] = self.mempool.get_spends_by_coin_id(bytes32(spend.coin_id))
+                    for item in items:
+                        included_items.append(item)
+                        self.remove_seen(item.name)
+                        spendbundle_ids_to_remove.append(item.name)
+                    self.mempool.remove_from_pool(spendbundle_ids_to_remove, MempoolRemoveReason.BLOCK_INCLUSION)
```
AmineKhaldi, line 622: This one got altered from being called once in the end, to being called several times. I doubt it's intentional?

tests/wallet/test_wallet_retry.py
```diff
@@ -36,7 +36,8 @@ def assert_sb_not_in_pool(node: FullNodeAPI, sb: SpendBundle) -> None:
 
 
 def evict_from_pool(node: FullNodeAPI, sb: SpendBundle) -> None:
-    mempool_item = node.full_node.mempool_manager.mempool.spends[sb.name()]
+    mempool_item = node.full_node.mempool_manager.mempool.get_spend_by_id(sb.name())
+    assert mempool_item
```
AmineKhaldi, line 40: `assert mempool_item is not None` ?

AmineKhaldi (commented):
Looks good. I left some comments.

chia/full_node/mempool.py
arvidn: I'm not so invested in these variable names as I already have a patch that replaces it all (locally)

arvidn: oh, you mean the member function should be `mempool_items_by_feerate()`?

chia/full_node/mempool.py
arvidn: so, `mempool_items_by_coin_id()`?

chia/full_node/mempool_manager.py
arvidn: last minute merge conflict..

chia/full_node/mempool.py
AmineKhaldi: Yep, interface level proper naming.

chia/full_node/mempool.py
AmineKhaldi: Indeed.

chia/full_node/mempool.py
arvidn: `mempool_item` is a bit long. is `tx` too short?

chia/full_node/mempool_manager.py
```diff
@@ -366,10 +362,9 @@ async def add_spend_bundle(
         """
 
         # Skip if already added
-        if spend_name in self.mempool.spends:
-            cost: Optional[uint64] = self.mempool.spends[spend_name].cost
-            assert cost is not None
-            return uint64(cost), MempoolInclusionStatus.SUCCESS, None
+        existing_item = self.mempool.get_spend_by_id(spend_name)
+        if existing_item is not None:
+            return uint64(existing_item.cost), MempoolInclusionStatus.SUCCESS, None
```
AmineKhaldi, line 367: Cost is already a `uint64` so I'd just `return existing_item.cost, MempoolInclusionStatus.SUCCESS, None`

chia/full_node/mempool.py
AmineKhaldi: Sure, sounds like a plan.

AmineKhaldi (dismissed):

tests/core/data_layer/test_data_rpc.py
```diff
@@ -116,7 +116,7 @@ async def farm_block_with_spend(
 
 
 def check_mempool_spend_count(full_node_api: FullNodeSimulator, num_of_spends: int) -> bool:
-    return len(full_node_api.full_node.mempool_manager.mempool.sorted_spends) == num_of_spends
```
arvidn, line 119: note how the current version counts the number of *sorted_spends*, which means counting the number of fee rate *levels*, not the number of items in the mempool. The patched version does away with easy access to the fee rate levels and instead count the actual number of mempool items. This also requires a few tests to change their numbers

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -84,7 +84,7 @@ async def farm_transaction_block(full_node_api: FullNodeSimulator, wallet_node:
 
 
 def check_mempool_spend_count(full_node_api: FullNodeSimulator, num_of_spends):
-    return len(full_node_api.full_node.mempool_manager.mempool.sorted_spends) == num_of_spends
```
arvidn, line 87: this is the other example of changing from counting *levels* to number of items

tests/core/data_layer/test_data_rpc.py
aqk: Yes, that old variable name was misleading.

tests/core/data_layer/test_data_rpc.py
```diff
@@ -116,7 +116,7 @@ async def farm_block_with_spend(
 
 
 def check_mempool_spend_count(full_node_api: FullNodeSimulator, num_of_spends: int) -> bool:
```
aqk, line 118: Hey, while you're in here, could you change `num_of_spends` to something like `num_of_fee_rate_levels` ?

tests/wallet/rpc/test_wallet_rpc.py
aqk: Nice. `num_of_spends` and `sorted_spends` are both misleading here as well.

aqk (approved):
Look great!

By the way, I'm sure you already know, but for future reference: we have many places in the code that use uint64 to hold mojos, but the max theoretical number of mojos won't fit in 64 bits.

tests/core/data_layer/test_data_rpc.py
arvidn: but with my change this value now refers to the number of spends in the mempool. the public interface to the mempool class no longer gives you access to the `sorted_spends` list

AmineKhaldi (approved):

trepca (approved):

PR #14606, opened 2023-02-17
run the connection DoS tests on the benchmark CI infrastructure

arvidn:
This is an attempt to improve the flakiness of the tests by isolating the ones requiring a lot of file descriptors to the benchmark runner, where tests are run one at a time.

### DIFF
--- a/tests/core/server/test_event_loop.py
+++ b/tests/core/server/test_event_loop.py
@@ -11,6 +11,7 @@ from chia.server.chia_policy import ChiaSelectorEventLoop, PausableServer, _chia
 
 
 @pytest.mark.asyncio
+@pytest.mark.benchmark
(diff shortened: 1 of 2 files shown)


xdustinface:
I also have a problem right now, obviously with too many processes/files over in https://github.com/Chia-Network/chia-blockchain/pull/14591 where i simply add some new tests to `core.util` and because of this `test_multiple_writer` is getting super flaky and stops passing at all for linux.. I tried moving `test_multiple_writers` to benchmark which fixed it.. but i guess we should have some other way to isolate such tests instead of pushing it to the benchmark HW because that one only runs python 3.9 and only linux?

arvidn:
I no longer think these tests affect CI flakiness

PR #14612, opened 2023-02-17
make SpendBundle.additions() count complete cost

arvidn:
apply the cost limit across all `CoinSpend`s whose additions are computed.

### DIFF
--- a/chia/types/spend_bundle.py
+++ b/chia/types/spend_bundle.py
@@ -9,10 +9,11 @@ from blspy import AugSchemeMPL, G2Element
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.sized_bytes import bytes32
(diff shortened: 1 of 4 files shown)


richardkiss (approved):
Looks good.

chia/types/spend_bundle.py
```diff
@@ -43,10 +44,14 @@ def aggregate(cls, spend_bundles) -> "SpendBundle":
         return cls(coin_spends, aggregated_signature)
 
     # TODO: this should be removed
```
emlowe, line 46: Mostly curious why this has a `this should be removed` todo?

emlowe (approved):

chia/types/spend_bundle.py
arvidn: the wallet calls this function as if it's cheap. It's frequently called unnecessarily as well. I envision this being a free function whose name makes it clear that it's not cheap.

PR #14613, opened 2023-02-17
exploring loop test adjustments

altendky:

### DIFF
--- a/tests/core/server/test_loop.py
+++ b/tests/core/server/test_loop.py
@@ -23,7 +23,6 @@ here = pathlib.Path(__file__).parent
 IP = "127.0.0.1"
 PORT = 8444
 NUM_CLIENTS = 500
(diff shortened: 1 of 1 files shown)


arvidn:
what's the objective with this patch?

arvidn:
this test consistently fail on my computer, because I run a chia full node.

```
OSError: [Errno 98] error while attempting to bind on address ('127.0.0.1', 8444): address already in use
```

altendky:
> what's the objective with this patch?

I hope the result is to make these tests more deterministic, less timing sensitive.  In so doing i felt comfortable reducing the repetition and connection counts so the tests could complete in less time.  We may well also want a test more like what this is changing away from but it might belong on the benchmark runner instead.

I'll take a look at the failure you also pointed out.

arvidn (approved):

altendky:
The address in use issue is pre-existing.  I'll follow up on that.

PR #14615, opened 2023-02-18
make conditions_for_solution() support untrusted input

arvidn:
This patch fixes `parse_sexp_to_condition()` to convert to python as it parses, rather than converting the entire tree to python types up-front. This means we can fail parsing and cut argument lists short early.

Without this patch, it's not safe to use these functions on untrusted input, we don't use these in the full node for instance. However, hardening this function against untrusted input protects against future mistakes where it may be assumed to handle it.

### DIFF
--- a/chia/util/condition_tools.py
+++ b/chia/util/condition_tools.py
@@ -25,12 +25,28 @@ def parse_sexp_to_condition(
     Takes a ChiaLisp sexp and returns a ConditionWithArgs.
     If it fails, returns an Error
     """
-    as_atoms = sexp.as_atom_list()
-    if len(as_atoms) < 1:
+    first = sexp.pair
+    if first is None:
         return Err.INVALID_CONDITION, None
-    opcode = as_atoms[0]
-    opcode = ConditionOpcode(opcode)
-    return None, ConditionWithArgs(opcode, as_atoms[1:])
+    op = first[0].atom
+    if op is None or len(op) != 1:
+        return Err.INVALID_CONDITION, None
+
+    # since the ConditionWithArgs only has atoms as the args, we can't parse
+    # hints and memos with this function. We just exit the loop if we encounter
+    # a pair instead of an atom
+    vars: List[bytes] = []
+    for arg in Program(first[1]).as_iter():
+        a = arg.atom
+        if a is None:
+            break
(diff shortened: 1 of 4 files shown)


cmmarslender (dismissed):
Makes sense to me, and looks good in general. I don't have a super in depth knowledge of all the opcodes to say for certain that "oh yeah, 4 arg limit is correct" - but otherwise, the rest looks good to me

chia/util/condition_tools.py
```diff
@@ -25,12 +25,28 @@ def parse_sexp_to_condition(
...
+    # hints and memos with this function
+    vars: List[bytes] = []
+    for arg in i:
+        a = arg.atom
+        if a is None:
+            break
+        vars.append(a)
+        # no condition (currently) has more than 3 arguments. Additional
+        # arguments are allowed but ignored
+        if len(vars) > 3:
```
cmmarslender, line 46: Should this actually be `>= 3` if we only support 3 args? Currently, this ends up with a max of 4 I believe

richardkiss (dismissed):

chia/util/condition_tools.py
```diff
@@ -25,12 +25,28 @@ def parse_sexp_to_condition(
     Takes a ChiaLisp sexp and returns a ConditionWithArgs.
     If it fails, returns an Error
     """
-    as_atoms = sexp.as_atom_list()
-    if len(as_atoms) < 1:
+    i = sexp.as_iter()
```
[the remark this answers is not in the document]
arvidn: I made this change

chia/util/condition_tools.py
```diff
@@ -25,12 +25,28 @@ def parse_sexp_to_condition(
     Takes a ChiaLisp sexp and returns a ConditionWithArgs.
     If it fails, returns an Error
     """
-    as_atoms = sexp.as_atom_list()
-    if len(as_atoms) < 1:
+    first = sexp.pair
```
richardkiss, line 28: The name `first` is a bit misleading, since it's actually a pair. Really, `first` is `sexp.pair[0]`.

richardkiss (approved):
I approved it anyway.

emlowe (approved):

PR #14624, opened 2023-02-20
require npc_result to be set in mempool items added to the mempool

arvidn:
This has always technically been a requirement. The assert makes sure all tests honor this requirement.

In a future patch I will use the `npc_results.conds.spends` for the removal coin IDs rather than `spend_bundle.removals()`. There are various reasons, one is that the coin IDs are already computed.


AmineKhaldi (approved):
Looks good.

aqk (approved):

trepca (approved):

PR #14625, opened 2023-02-20
Softfork2 infrastructure

arvidn:
introduce the new constant for the new upcoming soft-fork (for 1.8.0). Also enable parsing the `ASSERT_BEFORE_*` conditions after that height.

This flag also enables the other new conditions like `ASSERT_CONCURRENT_SPEND` and `ASSERT_CONCURRENT_PUZZLE`.

Note that these are already enabled in `MEMPOOL_MODE` unconditionally.

### DIFF
--- a/tests/core/test_cost_calculation.py
+++ b/tests/core/test_cost_calculation.py
@@ -141,7 +141,7 @@ class TestCostCalculation:
         program = SerializedProgram.from_bytes(
             binutils.assemble(
                 f"(q ((0x3d2331635a58c0d49912bc1427d7db51afe3f20a7b4bcaffa17ee250dcbcbfaa {disassembly} 300"
-                f"  (() (q . ((65 '00000000000000000000000000000000' 0x0cbba106e000))) ()))))"
+                f"  (() (q . ((48 '00000000000000000000000000000000' 0x0cbba106e000))) ()))))"
             ).as_bin()
         )
         generator = BlockGenerator(program, [], [])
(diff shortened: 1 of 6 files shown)


tests/core/test_cost_calculation.py
```diff
@@ -141,7 +141,7 @@ async def test_mempool_mode(self, softfork_height, bt):
         program = SerializedProgram.from_bytes(
             binutils.assemble(
                 f"(q ((0x3d2331635a58c0d49912bc1427d7db51afe3f20a7b4bcaffa17ee250dcbcbfaa {disassembly} 300"
-                f"  (() (q . ((65 '00000000000000000000000000000000' 0x0cbba106e000))) ()))))"
+                f"  (() (q . ((48 '00000000000000000000000000000000' 0x0cbba106e000))) ()))))"
```
aqk, line 144: What does this line do? ('A' -> '0')

richardkiss: I also cannot figure this out.

richardkiss:
This seems simple enough.

tests/core/test_cost_calculation.py
arvidn: I should have highlighted this change. the hint is in the commit message for this one change. This test is checking that unknown condition codes are ignored in normal consensus mode but rejected in mempool mode. The issue is that with this softfork, 65 is now a *known* condition code. This is just changing it to still be something unknown

richardkiss (approved):

trepca (approved):

PR #14629, opened 2023-02-20
tests: Merge some fixtures and drop other redundant ones

xdustinface:
### Purpose:

- Drop some redundant fixtures in `test_dl_wallet.py`
- Drop `two_wallet_nodes_five_freeze` and use `two_wallet_nodes` instead
- Replace all plain redundant "1 node, 1 wallet" fixtures with `simulator_and_wallet`.
- Drop `one_wallet_node` fixture in `test_simulation.py` and use `simulator_and_wallet`
- Replace `wallet_node` (which is the last usage of `setup_node_and_wallet`, hence it gets dropped here) with `simulator_and_wallet`

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/simulator/setup_nodes.py
+++ b/chia/simulator/setup_nodes.py
@@ -134,45 +134,6 @@ async def setup_n_nodes(
         keyring.cleanup()
 
 
-async def setup_node_and_wallet(
-    consensus_constants: ConsensusConstants,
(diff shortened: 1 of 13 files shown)


arvidn:
looks reasonable. perhaps @altendky should take a look too

arvidn (approved):

PR #14631, opened 2023-02-21
wallet: Improve balance caching

xdustinface:
### Purpose:

- Introduce the class `Balance`, use it for the balance cache and move the cache into `WalletNode` where it gets used/update via the here added `WalletNode.get_balance` method.
- Populate balance caches on wallet startup/login to avoid showing zero balances during catch-up sync

### Current Behavior:

The balance during long sync after startup is zero even though there is already a balance in the DB if the sync gets started before the first `get_wallet_balance` RPC call. 

### New Behavior:

The balance caches will be populated on startup so we always return the last available balance. 

### Testing Notes:

If this patch is applied on top of #14524 you can see how the wallet shows the last cached balances after startup vs just 0s without this patch. This PR also adds a new test `test_get_balance` for `WalletNode.get_balance`.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -46,10 +46,11 @@ from chia.util.config import (
 )
 from chia.util.db_wrapper import manage_connection
 from chia.util.errors import KeychainIsEmpty, KeychainIsLocked, KeychainKeyNotFound, KeychainProxyConnectionFailure
-from chia.util.ints import uint32, uint64
+from chia.util.ints import uint32, uint64, uint128
 from chia.util.keychain import Keychain
 from chia.util.path import path_from_root
 from chia.util.profiler import mem_profile_task, profile_task
+from chia.util.streamable import Streamable, streamable
 from chia.wallet.transaction_record import TransactionRecord
 from chia.wallet.util.new_peak_queue import NewPeakItem, NewPeakQueue, NewPeakQueueTypes
 from chia.wallet.util.peer_request_cache import PeerRequestCache, can_use_peer_request_cache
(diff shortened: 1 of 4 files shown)


chia/wallet/wallet_node.py
```diff
@@ -86,6 +87,18 @@ def get_wallet_db_path(root_path: Path, config: Dict[str, Any], key_fingerprint:
...
+@streamable
+@dataclasses.dataclass(frozen=True)
+class Balance(Streamable):
+    confirmed_wallet_balance: uint128 = uint128(0)
+    unconfirmed_wallet_balance: uint128 = uint128(0)
+    spendable_balance: uint128 = uint128(0)
+    pending_change: uint64 = uint64(0)
+    max_send_amount: uint128 = uint128(0)
+    unspent_coin_count: uint32 = uint32(0)
+    pending_coin_removal_count: uint32 = uint32(0)
```
arvidn, line 99: previously we would also store `wallet_type` and `wallet_id` here. I suppose the `wallet_id` is now used as the dictionary index to get to this object, but what about wallet type, did we not use that for anything previously?

chia/wallet/wallet_node.py
```diff
@@ -1657,3 +1676,30 @@ async def push_tx(self, spend_bundle):
...
         for peer in full_nodes:
             await peer.send_message(msg)
+
+    async def get_balance(self, wallet_id: uint32) -> Balance:
+        self.log.debug(f"get_balance - wallet_id: {wallet_id}")
+        if not self.wallet_state_manager.sync_mode:
+            self.log.debug(f"get_balance - Updating cache for {wallet_id}")
+            async with self.wallet_state_manager.lock:
+                wallet = self.wallet_state_manager.wallets[wallet_id]
+                unspent_records = await self.wallet_state_manager.coin_store.get_unspent_coins_for_wallet(wallet_id)
```
arvidn, line 1686: this could be millions of coins. This is not a new issue though

arvidn (dismissed):

chia/wallet/wallet_node.py
xdustinface: Yeah this was just moved.

chia/wallet/wallet_node.py
xdustinface: Good catch, thanks! See [1cd08da](https://github.com/Chia-Network/chia-blockchain/pull/14631/commits/1cd08dab746621c6f6766fe7a6c97fae10cb9db3) i missed to add those two to the RPC call. So i fixed it and also added a test for the RPC call!

arvidn (approved):

PR #14638, opened 2023-02-22
avoid excessive logging filter registration

altendky:
### Purpose:

Avoid gradually increasing CPU usage in the full node process that eventually causes loss of sync.

### Current Behavior:

For every message that was maybe to be logged we added a new filter object.  They presumably piled up and resulted in every debug message to that log being regexed against each filter that had ever been registered.

https://github.com/Chia-Network/chia-blockchain/pull/14435

### New Behavior:

Directly track the last logged time for that message being sent to over the connection.  If less than 60 seconds have elapsed, don't log it.

### Testing Notes:

1. graphana cpu usage data progressively growing over 12+ hours
2. profiling data pointing to debug logging and specifically the filter function
3. running with this and observing no issues with increasing cpu usage (in progess)

### DIFF
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import logging
+import math
 import time
 import traceback
 from dataclasses import dataclass, field
@@ -28,7 +29,6 @@ from chia.util.api_decorators import get_metadata
 from chia.util.errors import Err, ProtocolError
 from chia.util.ints import uint8, uint16
 from chia.util.log_exceptions import log_exceptions
-from chia.util.logging import TimedDuplicateFilter
 
 # Each message is prepended with LENGTH_BYTES bytes specifying the length
 from chia.util.network import class_for_type, is_localhost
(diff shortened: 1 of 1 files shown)


chia/server/ws_connection.py
```diff
@@ -560,10 +569,12 @@ async def _send_message(self, message: Message) -> None:
             message, self.local_capabilities, self.peer_capabilities
         ):
             if not is_localhost(self.peer_host):
-                msg = f"Rate limiting ourselves. message type: {ProtocolMessageTypes(message.type).name}, "
-                f"peer: {self.peer_host}"
-                self.log.debug(msg)
-                self.log.addFilter(TimedDuplicateFilter(msg, 60))
```
arvidn, line 566: is the `TimedDuplicateFilter` used anywhere else, or can that be removed now?

aqk (approved):
@arvidn: This looks quite similar to the first PR I proposed for this issue!

chia/server/ws_connection.py
altendky: It is still used in the farmer.  The uses there are less concerning though I still hesitate with 1) the cleanup that 'should' be done and 2) the regexing of every debug message.  But those aren't 'real' issues over there like the use here with repeated registration was.

arvidn (approved):

PR #14640, opened 2023-02-22
wallet: Disconnect untrusted peers if we find a trusted synced one

xdustinface:
### Purpose:

Disconnect untrusted peers as soon as we find a trusted synced peer, to continue with a trusted long sync instead of sticking with the untrusted sync until it's done.

We currently just continue syncing to untrusted peers even if we are already connected to a trusted peer. There is some effort in current `main` to try disconnecting untrusted peers but the logic isn't really working because we process messages from `WalletNode.new_peak_queue` which gets blocked whenever we start a sync, so the trusted peaks are not processed until the sync is done which is why we don't see the wallet switching from untrusted to trusted sync right now.

This PR adds some checking for trusted/synced peers before putting the `new_peak_wallet` messages into the queue and just disconnects the untrusted peers right away if there is a trusted synced one. The actual fix happens in 6048b76fb8f9a4438ba9ba06b9f0f9f678fbb4a5, the others are preparation/tests.

This PR is an alternative to #14430 where the idea is to check in few places during long sync if we are connected to a trusted peer and then bail out early form the sync if there is a trusted peer connected but i think it makes more sense to do it like here, to run this checks before we even put the message to the queue.

### Current Behavior:

The wallet just continues to do an untrusted sync even if its connected to a synced trusted peer.

### New Behavior:

The wallet will switch from the untrusted sync to the trusted sync whenever possible.

### Testing Notes:

There is a test added in this PR which makes sure the untrusted disconnect works as expected but you can test it manually also by starting a untrusted sync and then connecting to a trusted peer.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -1004,29 +1004,6 @@ class WalletNode:
                 neither.append(node)
         return synced_and_trusted + synced + trusted + neither
 
-    async def disconnect_and_stop_wpeers(self) -> None:
-        if self._server is None:
-            return
-
-        # Close connection of non-trusted peers
-        full_node_connections = self.server.get_connections(NodeType.FULL_NODE)
-        if len(full_node_connections) > 1:
-            for peer in full_node_connections:
-                if not self.is_trusted(peer):
-                    await peer.close()
-
-        if self.wallet_peers is not None:
-            await self.wallet_peers.ensure_is_closed()
-            self.wallet_peers = None
-
-    async def check_for_synced_trusted_peer(self, header_block: HeaderBlock) -> bool:
-        if self._server is None:
-            return False
-        for peer in self.server.get_connections(NodeType.FULL_NODE):
-            if self.is_trusted(peer) and await self.is_peer_synced(peer, header_block.height):
-                return True
-        return False
-
     async def get_timestamp_for_height(self, height: uint32) -> uint64:
         """
         Returns the timestamp for transaction block at h=height, if not transaction block, backtracks until it finds
@@ -1111,8 +1088,6 @@ class WalletNode:
         current_height: uint32 = await self.wallet_state_manager.blockchain.get_finished_sync_up_to()
         async with self.wallet_state_manager.lock:
             await self.wallet_state_manager.blockchain.set_peak_block(new_peak_hb, latest_timestamp)
-            # Disconnect from all untrusted peers if our local node is trusted and synced
-            await self.disconnect_and_stop_wpeers()
             # Sync to trusted node if we haven't done so yet. As long as we have synced once (and not
             # disconnected), we assume that the full node will continue to give us state updates, so we do
             # not need to resync.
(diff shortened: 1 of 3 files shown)


almogdepaz:
i am not sure we want to disconnect untrusted peers as opposed to just not syncing from them.
there is overhead in connecting to new peers and doing peer discovery  i think you want to be able to still get peaks if your trusted node has issues without doing peer discovery from the start.

this comment is based on the explanation i didn't go through the code yet

xdustinface:
> i am not sure we want to disconnect untrusted peers as opposed to just not syncing from them. 

Hm, thats what we already do here 

https://github.com/Chia-Network/chia-blockchain/blob/41ca9622a3721356e324daba711d767b2f817536/chia/wallet/wallet_node.py#L1117-L1118

and if we are synced to a trusted peer we immediately disconnect from untrusted ones here 

https://github.com/Chia-Network/chia-blockchain/blob/41ca9622a3721356e324daba711d767b2f817536/chia/wallet/wallet_node.py#L664-L666

This PR here basically just makes sure the first above linked disconnecting case works as expected. One thing which is maybe not ideal is that we with this just cancel the sync in which ever state it is right now but since we now only set the synced height after the sync since #14520 it should be fine i think?

xdustinface:
> support yes, optimal not sure, you saying we disconnect all untrusted if we have trusted peers so why is there even a secondary sync ?

Not sure how the secondary sync is related here now but we have it to not only rely on one untrusted node, so to validate the first untrusted node we synced to isn't lying. 

> i want to add the the trusted mode is not that important im my opinion, the goal of the light wallet was to allow people to not run a full node and still not having to trust the peer like in metamask.

Right, as soon as we have something better in place its not longer important. But so far there is a big difference especially if you have many coins to validate and it's probably important for many people.

emlowe:
Right now, the typical trusted node is going to be running on the same system as the wallet (`localhost`) - in this case (fast, local, trusted), there does seem to be a fairly big difference in trusted and untrusted syncing. So I think some stopgaps to prefer trusted are good. I would expect that a "slow, remote, and trusted node" is no different than a "slow, remote, and untrusted node" and hopefully we can eventually get to where all such nodes work the same

xdustinface:
> Right now, the typical trusted node is going to be running on the same system as the wallet (`localhost`) 

Im trusting my farmer node wherever i am, so im using a remote trusted peer :) 

> I would expect that a "slow, remote, and trusted node" is no different than a "slow, remote, and untrusted node" and hopefully we can eventually get to where all such nodes work the same

Why do you think so? There is always this big difference. I mean, in both cases the untrusted sync needs to validate the weight proof and validate all coin states where especially the coin state and block inclusion validation will always take its time. For wallets with a handful of coins it doesn't make a big difference, but for active users it will.

emlowe:
I'm probably wrong because I don't understand the paths correctly. 

They should end up at the same place though - but often don't. Which is the question really

xdustinface:
> They should end up at the same place though - but often don't. Which is the question really

They should and if they don't we should fix it but lets maybe get the focus back to the content of this PR.

emlowe (approved):

PR #14642, opened 2023-02-22
wallet: Handle some peer closed cases better

xdustinface:
### Purpose:

If a peer connection was closed during sync by either the client or the server we end up with a bunch of scary error logs in the debug log. This PR changes some log levels to `DEBUG` (should it maybe be `WARNING` instead?) based on the connection state and returns early in few places for disconnected peers to avoid running into code which will anyway fail.  

### Current Behavior:

`ERROR` logs for disconnected peer failures.

### New Behavior:

`DEBUG` logs for disconnected peer failures.

### Testing Notes:

 Just disconnect a peer which is in a long sync and watch the debug logs.

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -5,6 +5,7 @@ import json
 import logging
 import multiprocessing.context
 import time
+import traceback
 from contextlib import asynccontextmanager
 from pathlib import Path
 from secrets import token_bytes
@@ -988,7 +989,7 @@ class WalletStateManager:
             wallet_type = WalletType.NFT
         return wallet_id, wallet_type
 
-    async def add_coin_states(
+    async def _add_coin_states(
         self,
         coin_states: List[CoinState],
         peer: WSChiaConnection,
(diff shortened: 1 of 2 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -1014,6 +1014,8 @@ async def new_coin_state(
 
         assert len(local_records) == len(coin_states)
         for coin_state, local_record in zip(coin_states, local_records):
+            if peer.closed:
```
almogdepaz, line 1017: looking at where coin_state is called it looks like it is expected to throw if it fails

we should add a test to see that the state recovers correctly when this happens

xdustinface: Thanks, good point. I changed it to raise instead and i also reworked the PR a bit overall.

almogdepaz (dismissed):

chia/wallet/wallet_state_manager.py
```diff
@@ -1398,6 +1401,23 @@ async def new_coin_state(
                     await self.retry_store.remove_state(coin_state)
                 continue
 
+    async def new_coin_state(
+        self,
+        coin_states: List[CoinState],
+        peer: WSChiaConnection,
+        fork_height: Optional[uint32],
+    ) -> bool:
+        try:
```
aqk, line 1410: Maybe later we can move the rollback recovery out here, as well.

xdustinface: This would require splitting it up into also a function which handles single coin states because the rollback recovers is done per loop cycle. But i will come up with some other cleanups for this method anyway and there are already open PRs for it!

arvidn (dismissed):

arvidn (approved):

PR #14651, opened 2023-02-22
introduce run_chia_program()

arvidn:
and remove the member functions to run a program from SerialiedProgram.
This is preparing to make `SerializedProgram` a rust type as well as cutting down on "middle layer" functions between the underlying rust call and calling it from python.

The main change is in `chia/types/blockchain_format/serialized_program.py`. The other changes is accommodations for the new interface.

The new function takes `flags` and `max_cost` as optional arguments, which covers both the "mempool" and "consensus" cases in the same function.

Making `SerializedProgram` a rust type is a step towards making `CoinSpend` a rust type, which would improve serialization and deserialization performance of a lot of types

### DIFF
--- a/chia/full_node/generator.py
+++ b/chia/full_node/generator.py
@@ -69,14 +69,3 @@ def create_compressed_generator(
 def setup_generator_args(self: BlockGenerator) -> Tuple[SerializedProgram, Program]:
     args = create_generator_args(self.generator_refs)
     return self.program, args
-
-
-def run_generator_mempool(self: BlockGenerator, max_cost: int) -> Tuple[int, SerializedProgram]:
-    program, args = setup_generator_args(self)
-    return GENERATOR_MOD.run_mempool_with_cost(max_cost, program, args)
-
-
-def run_generator_unsafe(self: BlockGenerator, max_cost: int) -> Tuple[int, SerializedProgram]:
-    """This mode is meant for accepting possibly soft-forked transactions into the mempool"""
-    program, args = setup_generator_args(self)
-    return GENERATOR_MOD.run_with_cost(max_cost, program, args)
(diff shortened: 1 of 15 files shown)


chia/full_node/generator.py
```diff
@@ -69,14 +69,3 @@ def create_compressed_generator(
...
-
-def run_generator_mempool(self: BlockGenerator, max_cost: int) -> Tuple[int, SerializedProgram]:
-    program, args = setup_generator_args(self)
-    return GENERATOR_MOD.run_mempool_with_cost(max_cost, program, args)
-
-
-def run_generator_unsafe(self: BlockGenerator, max_cost: int) -> Tuple[int, SerializedProgram]:
-    """This mode is meant for accepting possibly soft-forked transactions into the mempool"""
-    program, args = setup_generator_args(self)
-    return GENERATOR_MOD.run_with_cost(max_cost, program, args)
```
arvidn, line 82: this function is only used by (two) tests, so it was moved there.

chia/full_node/generator.py
```diff
@@ -69,14 +69,3 @@ def create_compressed_generator(
 def setup_generator_args(self: BlockGenerator) -> Tuple[SerializedProgram, Program]:
     args = create_generator_args(self.generator_refs)
     return self.program, args
-
-
-def run_generator_mempool(self: BlockGenerator, max_cost: int) -> Tuple[int, SerializedProgram]:
-    program, args = setup_generator_args(self)
-    return GENERATOR_MOD.run_mempool_with_cost(max_cost, program, args)
```
arvidn, line 76: this function isn't used anywhere anymore

emlowe:
closing old PR

PR #14652, opened 2023-02-22
remove some redundant fields from MempoolItem

arvidn:
Apart from lowering memory footprint of the mempool, the main objective of this change is to (slightly) improve the performance of serializing and deserialializing `MempoolItem`s. A forthcoming PR storing these in a sqlite database rely in serializing and deserializing these.

This change shaves off a small amount of time deserializing:

| before | after | after / before |
| --- | --- | --- |
| 131.7 | 120.68 | 91.6 % |

before:
```
Profiling create_bundle_from_mempool()
  output written to: mempool-create-mt.png
  time: 65.8534s
  per call: 131.71ms
```

after:
```
Profiling create_bundle_from_mempool()
  output written to: mempool-create-mt.png
  time: 60.3389s
  per call: 120.68ms
```

### DIFF
--- a/chia/types/mempool_item.py
+++ b/chia/types/mempool_item.py
@@ -7,6 +7,7 @@ from chia.consensus.cost_calculator import NPCResult
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.spend_bundle import SpendBundle
+from chia.util.generator_tools import additions_for_npc
 from chia.util.ints import uint32, uint64
 from chia.util.streamable import Streamable, streamable
 
@@ -17,9 +18,7 @@ class MempoolItem(Streamable):
     spend_bundle: SpendBundle
     fee: uint64
     npc_result: NPCResult
-    cost: uint64
     spend_bundle_name: bytes32
-    additions: List[Coin]
     height_added_to_mempool: uint32
 
     # If present, this SpendBundle is not valid at or before this height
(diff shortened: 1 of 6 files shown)


chia/types/mempool_item.py
```diff
@@ -36,6 +35,15 @@ def fee_per_cost(self) -> float:
     def name(self) -> bytes32:
         return self.spend_bundle_name
 
+    @property
+    def cost(self) -> uint64:
+        assert self.npc_result.conds is not None
+        return uint64(self.npc_result.conds.cost)
+
+    @property
+    def additions(self) -> List[Coin]:
```
AmineKhaldi, line 44: This alters some RPCs as they won't include these in their responses anymore.

chia/types/mempool_item.py
```diff
@@ -36,6 +35,15 @@ def fee_per_cost(self) -> float:
...
+        return additions_for_npc(self.npc_result)
```
AmineKhaldi, line 45: I wonder if we can include additions (and removals) in `NPCResult` instead of computing on the python side.

arvidn: removals are already included in both `SpendBundle` and `SpendBundleConditions`.
additions are already included in `SpendBundleConditions`, available by iterating each `Spend`

chia/types/mempool_item.py
arvidn: you have a PR to address that, do you think we could get that landed?

AmineKhaldi: I think so: https://github.com/Chia-Network/chia-blockchain/pull/14476

chia/types/mempool_item.py
AmineKhaldi: Right, but I meant the need for `additions_for_npc` and any similar computation functions on the python side, on `NPCResult`.

aqk (approved):
This is good - I've certainly wanted the conditions for a MempoolItem before.

chia/types/mempool_item.py
arvidn: `additions_for_npc` is cheap, it just loops over the `spends` and appends the `create_coin` lists together.

AmineKhaldi (approved):

trepca (approved):

PR #14657, opened 2023-02-23
sqlite-backed mempool

arvidn:
# overview

The objective of this change is to allow to *reliably* add more sophisticated rules to the mempool.

This patch adds an sqlite database as the *index* for mempool items. In order to not repeat values from the index in the items themselves, it uses an `InternalMempoolItem` with fewer fields. the `_items` dictionary (which stores the mempool items) is kept in sync with the sqlite database, referencing items in the dict, by *name* (SpendBundle hash).

# rationale

The immediate features to add are the `assert_before_height` and `assert_before_seconds` fields on mempool items, ensuring they expire once those heights or timestamps have been reached.

I'm also envisioning a future where the "pending_cache" and "conflict_cache" are also folded into the database. That would also simplify the mempool data structure (or really, offload the complex parts to sqlite).

With that, I also imagine it would be easy to add support for `ASSERT_SECONDS`, which currently is not supported by the mempool (presumably to keep complexity in check).

# benefits

Fundamentally, the mempool is a multi-index container. A collection of items that we want efficient access to via different fields and ordering. e.g. we want to be able to iterate by fee rate, look up by spend bundle hash, look up by spent coin ID.

There is no native support for multi index cointainers in python, and implementing one is complex and error prone. Using a sqlite database greatly simplifies the data structure we need to support, by offloading all indices. This is especially significant as we add more indices (as mentioned under rationale). The complexity of the mempool will scale better with the number of indices and logic we add to the mempool.

# drawbacks

The main drawback of using a sqlite database is a slight performance degradation. The least common operation (farming a block) is slowed down and takes about 30% more time. See benchmarks below.

Another noteworthy consequence of this patch is that to total fee (i.e. the sum of all fees of all spend bundles) is not allowed to exceed `INT64_MAX`. This is because sqlite uses signed int64 as its internal representation for integers, so in order for the `fee_per_cost` field to be computed (and implied), sqlite need to be able to interpret the integers. This also allows the total fee and total cost to simply be sqlite queries for `SUM(<field>)`.

This patch, additionally, adds a fee limit of 2^50 mojos for an individual `MempoolItem`.

I don't think this restriction is problematic, since you would need astronomical fees in order to hit this ceiling, and it would only push out other transactions for one block (and then require more astronomical fees to keep it up).

# benchmarks

## Ubuntu AMD64 (CPU)

| function | before | after | after / before
| --- | --- | --- | --- |
| add_spend_bundle() | 6.3161s | 6.4140s | 101.55% |
| add_spend_bundle() replace | 6.2743s | 6.3782s | 101.66 % |
| create_bundle_from_mempool() | 4.5330s | 5.9528s | 131.32 % |
| new_peak() optimized | 16.6200s | 17.2157s | 103.58 % |
| new_peak() reorg | 16.2587s | 16.9127s | 104 % |

## Ubuntu AMD64 (memory)

The RAM pressure is pretty crude, just measuring the heap pressure of the mempool benchmark.
The benchmark logic itself probably contributes half of the memory pressure.

| metric | before | after | after / before |
| --- | --- | --- |--- |
| Max RSS | 78864 | 79908 | 101.32% |

## Raspberry PI (CPU)

| function | before | after | after / before
| --- | --- | --- | --- |
| add_spend_bundle() | 18.99s | 19.09s | 100.5 % |
| add_spend_bundle() replace | 18.99s | 19.09s | 100.54 % |
| create_bundle_from_mempool() | 16.98s | 21.53s | 126.78 % |
| new_peak() optimized | 47.12 s | 46.48s | 98.63 % |
| new_peak() reorg | 45.69 s | 46.42s | 101.6 % |

## Ubuntu amd64 (CPU)

current `main`:
```
Profiling add_spend_bundle()
  output written to: mempool-add-mt.png
  time: 6.3161s
  per call: 3.16ms

Profiling add_spend_bundle() with replace-by-fee
  output written to: mempool-replace-mt.png
  time: 6.2743s
  per call: 3.14ms

Profiling create_bundle_from_mempool()
  output written to: mempool-create-mt.png
  time: 4.5330s
  per call: 9.07ms

Profiling new_peak() (optimized)
  output written to: mempool-new-peak-mt.png
  time: 16.6200s
  per call: 8.31ms

Profiling new_peak() (reorg)
  output written to: mempool-new-peak-reorg-mt.png
  time: 16.2587s
  per call: 8.13ms
```

with sqlite:
```
Profiling add_spend_bundle()
  output written to: mempool-add-mt.png
  time: 6.4140s
  per call: 3.21ms

Profiling add_spend_bundle() with replace-by-fee
  output written to: mempool-replace-mt.png
  time: 6.3782s
  per call: 3.19ms

Profiling create_bundle_from_mempool()
  output written to: mempool-create-mt.png
  time: 5.9528s
  per call: 11.91ms

Profiling new_peak() (optimized)
  output written to: mempool-new-peak-mt.png
  time: 17.2157s
  per call: 8.61ms

Profiling new_peak() (reorg)
  output written to: mempool-new-peak-reorg-mt.png
  time: 16.9127s
  per call: 8.46ms
```

the profile of the most-impacted operation with sqlite backing:
mempool-create-mt

## Ubuntu AMD64 (memory)

current `main`:
```
        Command being timed: "python mempool.py"
        User time (seconds): 198.14
        System time (seconds): 1.24
        Percent of CPU this job got: 109%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 3:02.68
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 78864
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 92787
        Voluntary context switches: 34294
        Involuntary context switches: 861
        Swaps: 0
        File system inputs: 0
        File system outputs: 3232
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

with sqlite:
```
        Command being timed: "python mempool.py"
        User time (seconds): 209.45
        System time (seconds): 0.87
        Percent of CPU this job got: 111%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 3:09.48
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 79908
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 120195
        Voluntary context switches: 37254
        Involuntary context switches: 918
        Swaps: 0
        File system inputs: 0
        File system outputs: 4152
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

## Raspberry PI (CPU)

current `main`:
```
Profiling add_spend_bundle()
  output written to: mempool-add-mt.png
  time: 18.9916s
  per call: 9.50ms

Profiling add_spend_bundle() with replace-by-fee
  output written to: mempool-replace-mt.png
  time: 18.9871s
  per call: 9.49ms

Profiling create_bundle_from_mempool()
  output written to: mempool-create-mt.png
  time: 16.9782s
  per call: 33.96ms

Profiling new_peak() (optimized)
  output written to: mempool-new-peak-mt.png
  time: 47.1240s
  per call: 23.56ms

Profiling new_peak() (reorg)
  output written to: mempool-new-peak-reorg-mt.png
  time: 45.6839s
  per call: 22.84ms
```

with sqlite:
```
Profiling add_spend_bundle()
  output written to: mempool-add-mt.png
  time: 19.0876s
  per call: 9.54ms

Profiling add_spend_bundle() with replace-by-fee
  output written to: mempool-replace-mt.png
  time: 19.0887s
  per call: 9.54ms

Profiling create_bundle_from_mempool()
  output written to: mempool-create-mt.png
  time: 21.5251s
  per call: 43.05ms

Profiling new_peak() (optimized)
  output written to: mempool-new-peak-mt.png
  time: 46.4771s
  per call: 23.24ms

Profiling new_peak() (reorg)
  output written to: mempool-new-peak-reorg-mt.png
  time: 46.4199s
  per call: 23.21ms
```

### DIFF
--- a/chia/full_node/mempool.py
+++ b/chia/full_node/mempool.py
@@ -1,18 +1,26 @@
 from __future__ import annotations
 
+import sqlite3
+from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
-from typing import Dict, List, Optional
+from typing import Dict, Iterator, List, Optional
 
-from sortedcontainers import SortedDict
-
+from chia.consensus.cost_calculator import NPCResult
 from chia.full_node.fee_estimation import FeeMempoolInfo, MempoolInfo, MempoolItemInfo
 from chia.full_node.fee_estimator_interface import FeeEstimatorInterface
-from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.clvm_cost import CLVMCost
 from chia.types.mempool_item import MempoolItem
-from chia.util.ints import uint64
+from chia.types.spend_bundle import SpendBundle
+from chia.util.chunks import chunks
+from chia.util.db_wrapper import SQLITE_MAX_VARIABLE_NUMBER
+from chia.util.ints import uint32, uint64
+
+# We impose a limit on the fee a single transaction can pay in order to have the
+# sum of all fees in the mempool be less than 2^63. That's the limit of sqlite's
+# integers, which we rely on for computing fee per cost as well as the fee sum
+MEMPOOL_ITEM_FEE_LIMIT = 2**50
 
 
 class MempoolRemoveReason(Enum):
@@ -21,48 +29,117 @@ class MempoolRemoveReason(Enum):
     POOL_FULL = 3
 
 
+@dataclass(frozen=True)
+class InternalMempoolItem:
+    spend_bundle: SpendBundle
+    npc_result: NPCResult
+    height_added_to_mempool: uint32
+
+
 class Mempool:
+    _db_conn: sqlite3.Connection
+    # it's expensive to serialize and deserialize G2Element, so we keep those in
+    # this separate dictionary
+    _items: Dict[bytes32, InternalMempoolItem]
+
     def __init__(self, mempool_info: MempoolInfo, fee_estimator: FeeEstimatorInterface):
-        self._spends: Dict[bytes32, MempoolItem] = {}
-        self._sorted_spends: SortedDict = SortedDict()
+        self._db_conn = sqlite3.connect(":memory:")
+        self._items = {}
+
+        with self._db_conn:
+            # name means SpendBundle hash
+            # assert_height may be NIL
+            self._db_conn.execute(
+                """CREATE TABLE tx(
+                name BLOB PRIMARY KEY,
+                cost INT NOT NULL,
+                fee INT NOT NULL,
+                assert_height INT,
+                fee_per_cost REAL GENERATED ALWAYS AS (fee / cost) VIRTUAL)
+                """
+            )
+            self._db_conn.execute("CREATE INDEX fee_sum ON tx(fee)")
+            self._db_conn.execute("CREATE INDEX cost_sum ON tx(cost)")
+            self._db_conn.execute("CREATE INDEX feerate ON tx(fee_per_cost)")
+
+            # This table maps coin IDs to spend bundles hashes
+            self._db_conn.execute(
+                """CREATE TABLE spends(
+                coin_id BLOB NOT NULL,
+                tx BLOB NOT NULL,
+                UNIQUE(coin_id, tx))
+                """
+            )
+            self._db_conn.execute("CREATE INDEX spend_by_coin ON spends(coin_id)")
+            self._db_conn.execute("CREATE INDEX spend_by_bundle ON spends(tx)")
+
         self.mempool_info: MempoolInfo = mempool_info
         self.fee_estimator: FeeEstimatorInterface = fee_estimator
-        self._removal_coin_id_to_spendbundle_ids: Dict[bytes32, List[bytes32]] = {}
-        self._total_mempool_cost: CLVMCost = CLVMCost(uint64(0))
-        self._total_mempool_fees: int = 0
+
+    def __del__(self) -> None:
+        self._db_conn.close()
+
+    def _row_to_item(self, row: sqlite3.Row) -> MempoolItem:
+        name = bytes32(row[0])
+        fee = int(row[1])
+        assert_height = row[2]
+        item = self._items[name]
+
+        return MempoolItem(
+            item.spend_bundle, uint64(fee), item.npc_result, name, uint32(item.height_added_to_mempool), assert_height
+        )
 
     def total_mempool_fees(self) -> int:
-        return self._total_mempool_fees
+        with self._db_conn:
+            cursor = self._db_conn.execute("SELECT SUM(fee) FROM tx")
+            val = cursor.fetchone()[0]
+            return uint64(0) if val is None else uint64(val)
 
     def total_mempool_cost(self) -> CLVMCost:
-        return self._total_mempool_cost
+        with self._db_conn:
+            cursor = self._db_conn.execute("SELECT SUM(cost) FROM tx")
+            val = cursor.fetchone()[0]
+            return CLVMCost(uint64(0) if val is None else uint64(val))
 
-    def all_spends(self) -> List[MempoolItem]:
-        return list(self._spends.values())
+    def all_spends(self) -> Iterator[MempoolItem]:
+        with self._db_conn:
+            cursor = self._db_conn.execute("SELECT name, fee, assert_height FROM tx")
+            for row in cursor:
+                yield self._row_to_item(row)
 
     def all_spend_ids(self) -> List[bytes32]:
-        return list(self._spends.keys())
+        with self._db_conn:
+            cursor = self._db_conn.execute("SELECT name FROM tx")
+            return [bytes32(row[0]) for row in cursor]
 
-    def spends_by_feerate(self) -> List[MempoolItem]:
-        ret: List[MempoolItem] = []
-        for spends_with_fpc in reversed(self._sorted_spends.values()):
-            ret.extend(spends_with_fpc.values())
-        return ret
+    # TODO: move "process_mempool_items()" into this class in order to do this a
+    # bit more efficiently
+    def spends_by_feerate(self) -> Iterator[MempoolItem]:
+        with self._db_conn:
+            cursor = self._db_conn.execute("SELECT name, fee, assert_height FROM tx ORDER BY fee_per_cost DESC")
+            for row in cursor:
+                yield self._row_to_item(row)
 
     def size(self) -> int:
-        return len(self._spends)
+        with self._db_conn:
+            cursor = self._db_conn.execute("SELECT Count(name) FROM tx")
+            val = cursor.fetchone()
+            return 0 if val is None else int(val[0])
 
     def get_spend_by_id(self, spend_bundle_id: bytes32) -> Optional[MempoolItem]:
-        return self._spends.get(spend_bundle_id, None)
+        with self._db_conn:
+            cursor = self._db_conn.execute("SELECT name, fee, assert_height FROM tx WHERE name=?", (spend_bundle_id,))
+            row = cursor.fetchone()
+            return None if row is None else self._row_to_item(row)
 
+    # TODO: we need a bulk lookup function like this too
     def get_spends_by_coin_id(self, spent_coin_id: bytes32) -> List[MempoolItem]:
-        spend_bundle_ids = self._removal_coin_id_to_spendbundle_ids.get(spent_coin_id)
-        if spend_bundle_ids is None:
-            return []
-        ret: List[MempoolItem] = []
-        for spend_bundle_id in spend_bundle_ids:
-            ret.append(self._spends[spend_bundle_id])
-        return ret
+        with self._db_conn:
+            cursor = self._db_conn.execute(
+                "SELECT name, fee, assert_height FROM tx WHERE name in (SELECT tx FROM spends WHERE coin_id=?)",
+                (spent_coin_id,),
+            )
+            return [self._row_to_item(row) for row in cursor]
 
     def get_min_fee_rate(self, cost: int) -> float:
         """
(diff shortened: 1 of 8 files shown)


chia/full_node/mempool.py
```diff
@@ -1,18 +1,26 @@
...
-from chia.util.ints import uint64
+from chia.types.spend_bundle import SpendBundle
+from chia.util.chunks import chunks
+from chia.util.db_wrapper import SQLITE_MAX_VARIABLE_NUMBER
+from chia.util.ints import uint32, uint64
+
+# We impose a limit on the fee a single transaction can pay in order to have the
+# sum of all fees in the mempool be less than 2^63. That's the limit of sqlite's
+# integers, which we rely on for computing fee per cost as well as the fee sum
+MEMPOOL_ITEM_FEE_LIMIT = 2**50
```
aqk, line 23: Would be nice to compute this from `(2^63)` and `MAX_BLOCK_COST_CLVM * BLOCK_SIZE_LIMIT_FACTOR`.

But this comment is more than sufficient.

chia/full_node/mempool.py
```diff
@@ -21,65 +29,139 @@ class MempoolRemoveReason(Enum):
...
-        self._sorted_spends: SortedDict = SortedDict()
+        self._db_conn = sqlite3.connect(":memory:")
+        self._items = {}
+
+        with self._db_conn:
+            # name means SpendBundle hash
+            # assert_height may be NIL
+            self._db_conn.execute(
+                """CREATE TABLE tx(
+                name BLOB PRIMARY KEY,
```
aqk, line 54: Thank you for formatting this SQL statement.

aqk:
Great comment.
https://github.com/Chia-Network/chia-blockchain/pull/14657/files#diff-cad97bc75b80762f85e79ff71c5640bdeb58cfbb97992c82b08b64df958d262aR550

aqk:
It would be nice to compute this from (2^63) and MAX_BLOCK_COST_CLVM * BLOCK_SIZE_LIMIT_FACTOR.

But this comment is more than sufficient.

https://github.com/Chia-Network/chia-blockchain/pull/14657/files#diff-1cd73731eeaeb06087e341a00a41d5a6e819772688cb56434f6687895da82dffR23

chia/full_node/mempool.py
```diff
@@ -90,66 +172,76 @@ def remove_from_pool(self, items: List[bytes32], reason: MempoolRemoveReason) ->
...
+            return
+
+        removed_items: List[MempoolItemInfo] = []
+        if reason != MempoolRemoveReason.BLOCK_INCLUSION:
+
+            for spend_bundle_ids in chunks(items, SQLITE_MAX_VARIABLE_NUMBER):
+                args = ",".join(["?"] * len(spend_bundle_ids))
+                with self._db_conn:
+                    cursor = self._db_conn.execute(
+                        f"SELECT name, cost, fee FROM tx WHERE name in ({args})", spend_bundle_ids
```
aqk, line 185: Arvid points out that `remove_from_pool` can now throw `KeyError`.
The important call sites are:
mempool_manager.py:new_peak
mempool_manager.py:add_spend_bundle
mempool.py:add_to_pool

Let's go over those in the meeting.

chia/full_node/mempool.py
```diff
@@ -90,66 +172,76 @@ def remove_from_pool(self, items: List[bytes32], reason: MempoolRemoveReason) ->
...
+                        item = MempoolItemInfo(int(row[1]), int(row[2]), internal_item.height_added_to_mempool)
+                        removed_items.append(item)
+
+        for name in items:
+            self._items.pop(name)
+
+        for spend_bundle_ids in chunks(items, SQLITE_MAX_VARIABLE_NUMBER):
+            args = ",".join(["?"] * len(spend_bundle_ids))
+            with self._db_conn:
+                self._db_conn.execute(f"DELETE FROM tx WHERE name in ({args})", spend_bundle_ids)
```
aqk, line 199: What happens if we try to delete a nonexistent tx?

It's probably worth checking in tests with a malformed DB in each of these new SQL query cases, so we can test that the code calling these methods do what we expect.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -534,3 +534,36 @@ async def test_get_items_not_in_filter() -> None:
     sb2_and_3_filter = PyBIP158([bytearray(sb2_name), bytearray(sb3_name)])
     result = mempool_manager.get_items_not_in_filter(sb2_and_3_filter)
     assert result == [sb1]
+
```
aqk, line 537: I think we need a new test (maybe in benchmarks) that simulates the mempool running for a long time (several weeks)

chia/full_node/mempool.py
```diff
@@ -90,66 +172,76 @@ def remove_from_pool(self, items: List[bytes32], reason: MempoolRemoveReason) ->
...
+                self.remove_from_pool([name], MempoolRemoveReason.POOL_FULL)
+
+            self._db_conn.execute(
+                "INSERT INTO tx VALUES(?, ?, ?, ?)", (item.name, item.cost, item.fee, item.assert_height)
+            )
 
-        self._sorted_spends[item.fee_per_cost][item.name] = item
+            all_coin_spends = [(s.coin_id, item.name) for s in item.npc_result.conds.spends]
+            for spends in chunks(all_coin_spends, SQLITE_MAX_VARIABLE_NUMBER):
+                self._db_conn.executemany("INSERT INTO spends VALUES(?, ?)", all_coin_spends)
```
aqk, line 233: typo

chia/full_node/mempool_manager.py
```diff
@@ -470,6 +474,13 @@ async def validate_spend_bundle(
         if cost == 0:
             return Err.UNKNOWN, None, []
 
+        # this is not very likely to happen, but it's here to ensure SQLite
+        # never runs out of precision in its computation of fees.
+        # sqlite's integers are signed int64, so the max value they can
+        # represent is 2^63-1
+        if fees > MEMPOOL_ITEM_FEE_LIMIT or 2**63 - self.mempool.total_mempool_fees() <= fees:
```
arvidn, line 481: give int64 max a name, like: SQLITE_INT_MAX

chia/full_node/mempool_manager.py
```diff
@@ -180,14 +180,18 @@ def process_mempool_items(
         for item in self.mempool.spends_by_feerate():
             if not item_inclusion_filter(item.name):
                 continue
-            log.info(f"Cumulative cost: {cost_sum}, fee per cost: {item.fee / item.cost}")
+            log.info("Cumulative cost: %d, fee per cost: %0.4f", cost_sum, item.fee / item.cost)
```
arvidn, line 183: use `item.fee_per_cost`

chia/full_node/mempool_manager.py
```diff
@@ -180,14 +180,18 @@ def process_mempool_items(
...
             if item.cost + cost_sum > self.max_block_clvm_cost or item.fee + fee_sum > self.constants.MAX_COIN_AMOUNT:
                 return (spend_bundles, uint64(cost_sum), additions, removals)
             spend_bundles.append(item.spend_bundle)
             cost_sum += item.cost
             fee_sum += item.fee
             removals.extend(item.removals)
-            additions.extend(item.additions)
+            if item.npc_result.conds is not None:
```
arvidn, line 190: this should be `assert item.npc_result.conds`

chia/full_node/mempool.py
arvidn: yes, but I will remove the loop, so this typo will become correct

chia/full_node/mempool_manager.py
arvidn: done

chia/full_node/mempool.py
```diff
@@ -90,66 +172,76 @@ def remove_from_pool(self, items: List[bytes32], reason: MempoolRemoveReason) ->
...
-
-        # _sorted_spends is Dict[float, Dict[bytes32, MempoolItem]]
-        if item.fee_per_cost not in self._sorted_spends:
-            self._sorted_spends[item.fee_per_cost] = {}
+        # TODO: this block could be simplified by removing all items in a single
+        # SQL query. Or at least figure out which items to remove and then
+        # remove them all in a single call to remove_from_pool()
+        with self._db_conn:
+            while self.at_full_capacity(item.cost):
+                # pick the item with the lowest fee per cost
```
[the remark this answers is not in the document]
arvidn: done

chia/full_node/mempool.py
arvidn: it's not an error to include names in the `in (...)` statements that don't match any rows in the table

arvidn: but we delete out of the python dict as well, which *does* fail if we try to delete an item that doesn't exist

chia/full_node/mempool.py
aqk: `add_to_pool`: this call site seems safe because if we throw from here, we will not add a tx we should not have.
`add_spend_bundle`: seems safe, because we will throw before `add_to_pool`

`new_peak`: This one may not be OK. We can throw in the middle of a loop that is removing spends from the mempool. It seems that removing fewer spends than intended in this case is safe, but may prevent new spends being added that otherwise would have been if all the other spends had been removed.

arvidn: you mean the `new_peak()` function in `mempool_manager.py`, right? I'm confident that's a place where we won't throw, since we only collect transactions that already exist in the mempool, and then remove them.

AmineKhaldi (approved):

tests/core/mempool/test_mempool_manager.py
aqk: What do you think?

aqk (approved):

tests/core/mempool/test_mempool_manager.py
arvidn: it's a bit nebulous. We could have a test similar to the benchmark that causes churn through the mempool. but the test would still (presumably) be human inspection of how the process is doing.

arvidn: I made a program that churns transactions through the mempool for about a week worth of blocks ([here](https://github.com/Chia-Network/chia-blockchain/pull/14728)) and running that on this patch seems to reach the stable state (when the mempool is full). without leaking memory.

image

wjblanke:
We should add some exception handling in the off case that sqlite borks, but this seems ok for now

wjblanke (approved):
aok excited to see this in action

chia/full_node/mempool.py
```diff
@@ -21,65 +29,139 @@ class MempoolRemoveReason(Enum):
...
+                name BLOB PRIMARY KEY,
+                cost INT NOT NULL,
+                fee INT NOT NULL,
+                assert_height INT,
+                fee_per_cost REAL GENERATED ALWAYS AS (fee / cost) VIRTUAL)
+                """
+            )
+            self._db_conn.execute("CREATE INDEX fee_sum ON tx(fee)")
+            self._db_conn.execute("CREATE INDEX cost_sum ON tx(cost)")
+            self._db_conn.execute("CREATE INDEX feerate ON tx(fee_per_cost)")
```
neurosis69, line 63: you could consider adding this index to support `spends_by_feerate`:
```
CREATE INDEX feerate_desc ON tx(fee_per_cost desc);
```

My benchmarks didn't bring a big improvement, but still the time for spends_by_feerate was reduced from ~24.5% to 24% on my testing machine.

arvidn: interesting. I assumed indices are ordered, without the traversal order mattering.

PR #14661, opened 2023-02-23
anyio is needed as a standard dependency for the simulator

altendky:


arvidn:
`git grep anyio` suggests that it's only used by testing code (and possibly simulator code).

```
chia/simulator/full_node_simulator.py
tests/core/server/test_loop.py
```

that seems like a dev dependency to me, but maybe I'm missing some subtle detail

arvidn:
ok, the simulator is not a developer feature I take it

altendky:
It's in the runtime codebase, yeah.

arvidn (approved):

PR #14662, opened 2023-02-23
simplify and improve daemon connection handling

emlowe:
Simplified daemon connection handling to use one single data structure and removed `remote_address_map` as this wasn't used for anything consequential and introduced bugs. Also, change a List into a Set to avoid managing duplicate connections

Previously the daemon did not handle well the following cases:
- using the same websocket for two different services (calling `register_service` twice with different service names). In this case, error handling was incorrect and would result in an old websockets hanging around and being used to send traffic which would result in exceptions. Note that `chia-exporter` uses this pattern
- registering the same service multiple times over the same websocket. this would result in duplicate messages being sent to the service for each message, as the websocket would be appended to a List for each `register_service` call

The PR code is simpler and handles the above cases. Added some additional Logging and some sanity checks on sending connections

Tests added to cover the mentioned areas


chia/daemon/server.py
```diff
@@ -135,8 +135,7 @@ def __init__(
         self.log = log
         self.services: Dict = dict()
         self.plots_queue: List[Dict] = []
-        self.connections: Dict[str, List[WebSocketResponse]] = dict()  # service_name : [WebSocket]
-        self.remote_address_map: Dict[WebSocketResponse, str] = dict()  # socket: service_name
+        self.connections: Dict[str, Set[WebSocketResponse]] = dict()  # service_name : [WebSocket]
```
altendky, line 138: The comment needs updated I guess.

chia/daemon/server.py
```diff
@@ -231,74 +230,84 @@ async def incoming_connection(self, request):
...
+                    service_name = "Unknown"
+                    connections = {ws}  # send error back to the sender
+
+                for connection in connections:
+                    if connection.closed:
+                        # strange, connection is closed, but we still have it in the list
+                        self.remove_connection(connection)
+                        continue
+                    try:
+                        await connection.send_str(response)
```
altendky, line 249: There might still be a race where the connection has closed by now.  Maybe there's a distinct exception for that that we could catch here to respond to specifically instead of pre-checking?

chia/daemon/server.py
```diff
@@ -231,74 +230,84 @@ async def incoming_connection(self, request):
...
+                        self.log.info(f"Closing websocket with {service_name}")
+                        self.remove_connection(connection)
+                        await connection.close()
             else:
-                service_name = "Unknown"
-                if ws in self.remote_address_map:
-                    service_name = self.remote_address_map[ws]
+                service_names = self.find_service_names(ws)
+                if not service_names:
+                    service_names = "Unknown"
```
altendky, line 259: `Union[str, List[str]]` isn't a beautiful hint.  Maybe if we're more explicit?

```suggestion
                service_names = self.find_service_names(ws)
                if len(service_names) > 0:
                    formatted_service_names = ", ".join(name for name in service_names)
                else:
                    formatted_service_names = "unknown service"
```

tests/core/daemon/test_daemon_register.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import Any
+
+import aiohttp
+import pytest
+
+from chia.util.ws_message import create_payload
+
+
+@pytest.mark.asyncio
+async def test_multiple_register_same(get_daemon: Any, bt: Any) -> None:
```
altendky, line 12: I think those `Any` are easily replaced.

tests/core/daemon/test_daemon_register.py
```diff
@@ -0,0 +1,96 @@
...
+
+    connections = ws_server.connections.get(service_name, {})
+    assert len(connections) == 1
+
+    await client.close()
+    await ws_server.stop()
+
+
+@pytest.mark.asyncio
+async def test_multiple_register_different(get_daemon: Any, bt: Any) -> None:
```
altendky, line 51: I think those `Any` are easily replaced.

altendky (changes requested):

chia/daemon/server.py
emlowe: It probably throws ConnectionReset - so I was just hoping to avoid the traceback in the log for this case. Although I could write better logging in the ConnectionReset case, since that is a rather typical networking error

chia/daemon/server.py
emlowe: I changed the code a bit - how about just using a list with ["Unknown"] - simple short and only for logging.

chia/daemon/server.py
emlowe: I removed it

cmmarslender (dismissed):
You have my upvote functionally speaking. I tested this and all the random daemon/websocket bugs I was aware of before are no longer happening 🥳

tests/core/daemon/test_daemon.py
```diff
@@ -777,3 +777,49 @@ async def test_key_label_methods(
...
+    message = json.loads(response.data.strip())
+    assert message["data"]["success"] is False
+    assert message["data"]["error"].startswith("Expecting property name")
+
+    # properly register a service
+    service_name = "test_service"
+    data = {"service": service_name}
+    payload = create_payload("register_service", data, service_name, "daemon")
+    await ws.send_str(payload)
+    await ws.receive()
```
altendky, line 800: Worth making sure the response indicates success so as to catch any issues here early and avoid confusion by an error further down?

tests/core/daemon/test_daemon.py
```diff
@@ -777,3 +777,49 @@ async def test_key_label_methods(
     keychain.add_private_key(test_key_data.mnemonic_str(), "key_0")
     await ws.send_str(create_payload(method, parameter, "test", "daemon"))
     assert_response(await ws.receive(), response_data_dict)
+
+
+@pytest.mark.asyncio
+async def test_bad_json(daemon_connection_and_temp_keychain: Tuple[aiohttp.ClientWebSocketResponse, Keychain]) -> None:
```
altendky, line 783: Do we need the temp keychain for these tests?

tests/core/daemon/test_daemon_register.py
```diff
@@ -0,0 +1,113 @@
...
+            ssl_context=bt.get_daemon_ssl_context(),
+            max_msg_size=100 * 1024 * 1024,
+        )
+
+        service_name = "test_service"
+        data = {"service": service_name}
+        payload = create_payload("register_service", data, service_name, "daemon")
+        for _ in range(4):
+            await ws.send_str(payload)
+            await ws.receive()
```
altendky, line 33: Verify indication of success?

tests/core/daemon/test_daemon_register.py
```diff
@@ -0,0 +1,113 @@
...
+
+    connections = ws_server.connections.get(service_name, set())
+    assert len(connections) == 1
```
altendky, line 36: How about verifying 0 prior to registering, and then 1 after each registration in the loop?  Or collect the lengths in a list and `assert that_list = [0, 1, 1, 1, 1]` to get them all asserted at once at the end.

tests/core/daemon/test_daemon_register.py
```diff
@@ -0,0 +1,113 @@
...
+
+        test_service_names = ["service1", "service2", "service3"]
+
+        for service_name in test_service_names:
+            data = {"service": service_name}
+            payload = create_payload("register_service", data, service_name, "daemon")
+            await ws.send_str(payload)
+            await ws.receive()
+
+        assert len(ws_server.connections.keys()) == len(test_service_names)
```
altendky, line 64:
```suggestion
        assert sorted(ws_server.connections.keys()) == test_service_names
```

tests/core/daemon/test_daemon_register.py
```diff
@@ -0,0 +1,113 @@
...
+
+        for service_name in test_service_names:
+            connections = ws_server.connections.get(service_name, set())
+            assert len(connections) == 1
+
+        await ws.close()
```
altendky, line 70: No context manager for a client?  If not I guess we need to `try`/`finally` or make a helper context manager.  There's https://docs.python.org/3.10/library/contextlib.html#contextlib.aclosing but it's 3.10+.

I think this might apply other places.

tests/core/daemon/test_daemon_register.py
```diff
@@ -0,0 +1,113 @@
...
+
+        for service_name in test_service_names:
+            connections = ws_server.connections.get(service_name, set())
+            assert len(connections) == 0
```
altendky, line 74: Could apply this other places as well, unless maybe you are concerned about it switching from a set to something else?

```suggestion
            assert connections == set()
```

tests/core/daemon/test_daemon_register.py
```diff
@@ -0,0 +1,113 @@
...
+            max_msg_size=100 * 1024 * 1024,
+        )
+
+        test_service_names = ["service1", "service2", "service3", "service4", "service5"]
+
+        for service_name in test_service_names:
+            data = {"service": service_name}
+            payload = create_payload("register_service", data, service_name, "daemon")
+            await ws.send_str(payload)
+            await ws.receive()
```
altendky, line 100: Verify indication of success?

tests/core/daemon/test_daemon_register.py
```diff
@@ -0,0 +1,113 @@
...
+    connections = ws_server.connections.get(test_service_names[0], set())
+    assert len(connections) == 1
+    ws_to_remove = next(iter(connections))
+
+    removed_names = ws_server.remove_connection(ws_to_remove)
+    assert removed_names == test_service_names
+
+    # remove again, should return empty set and not raise any errors
+    removed_names = ws_server.remove_connection(ws_to_remove)
+    assert len(removed_names) == 0
```
altendky, line 113: Same sort of stuff as above could be considered here.

altendky (commented):
Do we have a usable 'self hostname' thing for tests at this point instead of 127.0.0.1?

tests/core/daemon/test_daemon_register.py
emlowe: I don't think it's needed, the test will fail if this didn't work because the connections won't match in the next line or so and it would only be slightly more helpful to fail earlier.

tests/core/daemon/test_daemon_register.py
emlowe: good idea, implemented

tests/core/daemon/test_daemon_register.py
emlowe: added

tests/core/daemon/test_daemon_register.py
emlowe: changed

chia/daemon/server.py
```diff
@@ -233,79 +232,106 @@ async def incoming_connection(self, request):
...
+            except Exception as e:
+                service_names = self.remove_connection(connection)
+                if len(service_names) == 0:
+                    service_names = ["Unknown"]
+
+                tb = ""
+                if not isinstance(e, ConnectionResetError):
+                    tb = traceback.format_exc()
+
+                self.log.error(f"Unexpected exception trying to send to {service_names} (websocket: {e} {tb})")
```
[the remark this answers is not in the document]
emlowe: I was being a little lazy here. Yes, ConnectionReset is afaik a pretty typical error message and is fairly expected and ordinary.

tests/core/daemon/test_daemon.py
emlowe: IMO no

tests/core/daemon/test_daemon_register.py
emlowe: I don't think it's needed

tests/core/daemon/test_daemon_register.py
altendky: Yes, the idea is clarity of the failure.  That it isn't that the server didn't track the connection but that you didn't even talk to the server successfully.

tests/core/daemon/test_daemon.py
altendky: Mm, I was thinking this triggered extra setup work but it seems maybe it doesn't.

https://github.com/Chia-Network/chia-blockchain/blob/69a91d24ef1ef60e1e88de353857c8f0d641b02c/tests/conftest.py#L599-L602

altendky (approved):
Woohoo!

paninaro (approved):
lgtm

PR #14663, opened 2023-02-23
Make sure to await properly during Daemon shutdown

emlowe:
Running a simple test like `test_daemon_rpc` with `pytest -n 0 tests/core/test_daemon_rpc.py` results in various not awaited errors and task destroyed messages like :
```
sys:1: RuntimeWarning: coroutine 'WebServer._close' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
Task was destroyed but it is pending!
task: <Task pending name='Task-16' coro=<WebServer._close() running at /Users/elowe/source/chia-blockchain/chia/util/network.py:91> cb=[Task.task_wakeup()]>
```

This PR attempts to make all the ways the daemon can be stopped work cleanly. The following have been tested to work without errors in logs or console. There are still tasks created that are not awaited, so there may still be edges cases

Testing scenarios - testing limited to macOS
- running pytest. daemon service stop method properly stops and awaits and tests exit clean
- using `chia stop -d`. Daemon stops all started services and cleanly exits with no errors in the log
- HUP'ing the service (`kill -HUP`). Daemon stops all started services and cleanly exits with no errors in the log
- running daemon in foreground `chia run_daemon` and using `ctrl-c`. Daemon stops all started services and cleanly exits with no errors in the log.

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -213,7 +213,7 @@ class WebSocketServer:
         if stop_service_jobs:
             await asyncio.wait(stop_service_jobs)
         self.services.clear()
-        asyncio.create_task(self.exit())
+        await self.exit()
         log.info(f"Daemon Server stopping, Services stopped: {service_names}")
         return {"success": True, "services_stopped": service_names}
 
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -204,7 +204,12 @@ class WebSocketServer:
             except Exception as e:
                 self.log.error(f"Error while canceling task.{e} {task}")
 
-    async def stop(self) -> Dict[str, Any]:
+    def daemon_exit_message(self, message: WsRpcMessage) -> bool:
+        command = message["command"]
+        destination = message["destination"]
+        return command == "exit" and destination == "daemon"
+
+    async def stop_services(self) -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


xdustinface (changes requested):
This is a problem because you are with this waiting for the server to shutdown during processing the `exit` message which runs inside `incoming_connection` which runs as part of the server context. So this `await` you added leads to some weird "the task waits inside itself for itself to be done" situation i think?

You can see if you try to stop the daemon via `chia stop all -d`, it will fail with a timeout or open/close the GUI, it will stuck because there will no more response from the daemon.

emlowe:
Yea, daemon shutdown is messy and agreed this isn't the correct way to solve it. Although it does make the tests cleanup correct. But the current code also isn't great since it just creates the task and never awaits it. I'll move to draft to explore other options

emlowe:
Latest commit I think works in a variety of tests I have done.

xdustinface (changes requested):
Seems to work with this now but this still leads to a dangling, not awaited task and while reviewing i realised we can actually just do https://github.com/Chia-Network/chia-blockchain/pull/14674.

emlowe:
Closing PR as was replaced by https://github.com/Chia-Network/chia-blockchain/pull/14674.

PR #14668, opened 2023-02-24
Fix a DID Wallet issue where an old DID coin causes us to assert during wallet creation

aqk:
### Purpose:

Fix an issue where an old DID coin causes us to assert during wallet creation.

Short story: During `handle_did`, `create_new_did_wallet_from_coin_spend` is called. Soon after, `DIDWallet.coin_added` is called with the same coin.

We need to talk about what the right way to fix this is. Should we move more of the checks from coin_added up into `create_new_did_wallet_from_coin_spend`, and reject this coin earlier, not creating the wallet at all?

Or should we create this wallet, and refine `coin_added` so that if reads the spend solution, and distingushes recovery spends from the other type of spend?

Note that the calculated inner puzzle does not match the on-chain inner puzzle.

### Current Behavior:

```
01:09:55 DID did:chia:1jz3gkneglep0whvan3pl6e2hg2u4gw7s325m4j5k26840qtp7y6sc232jc: WARNING DID 90a28b4f28fe42f75d9d9c43fd655742b9543bd08aa9baca96568f578161f135 has a recovery list hash but missing a reveal, you may need to reset the recovery info.
01:09:56 chia.wallet.wallet_state_manager: ERROR Error adding state...
  File "/Users/aqk/chia-blockchain/chia/wallet/wallet_state_manager.py", line 1081, in new_coin_state
    await self.coin_added(
  File "/Users/aqk/chia-blockchain/chia/wallet/wallet_state_manager.py", line 1534, in coin_added
    await self.wallets[wallet_id].coin_added(coin, height, peer)
  File "/Users/aqk/chia-blockchain/chia/wallet/did_wallet/did_wallet.py", line 422, in coin_added
```

### New Behavior:

We don't throw in this case. Note that this patch does not fix all issues in DID wallet creation and coin processing.


ytx1991:
I think DID sync part need a deeper cleanup. It has two big design changes which caused many edge cases. The goal of the cleanup is to simplify the logic. I will make a draft.

aqk:
> I think DID sync part need a deeper cleanup. It has two big design changes which caused many edge cases. The goal of the cleanup is to simplify the logic. I will make a draft.

Okay, super. I think we can commit this to main in the meantime.

wjblanke (approved):
aok

wallentx (approved):

PR #14669, opened 2023-02-24
simplify create_bundle_from_mempool() interface

arvidn:
Currently `create_bundle_from_mempool()` returns 3 values, the spend bundle, additions and removals. removals are available in the SpendBundle and so do not provide great benefits of returning.

This removes the `removals` return value to just return the SpendBundle and additions. Note that additions would still be expensive to re-compute from the SpendBundle, so they make sense to return


emlowe:
closing old PR - create_bundle_from_mempool has been changed already to not return removals

PR #14674, opened 2023-02-25
daemon: Fix server shutdown / wait for it

xdustinface:
### Purpose:

Make sure the daemon server gets always closed properly. 

This is an alternative to #14663.

### Current Behavior:

The daemon server shutdown isn't awaited which leads to spammy logs in some cases.

### New Behavior:

The daemon server shutdown gets awaited properly. 

### Testing Notes:

I tried all ways to start/stop a daemon (CLI/GUI) and all seemed to work fine.

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -128,7 +128,6 @@ class WebSocketServer:
         ca_key_path: Path,
         crt_path: Path,
         key_path: Path,
-        shutdown_event: asyncio.Event,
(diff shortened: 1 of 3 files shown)


xdustinface:
> See `setup_daemon` in `setup_services.py` - I think that needs to call both the stop and exit functions.

Good point yeah! See the two commits, i reworked the PR to make it contextmanager. I think now all cases should be covered.

emlowe (approved):
looks good to me

PR #14689, opened 2023-02-27
Bump chia-blockchain-gui to 71995c6eb722e8d92008e2d3df38fd71f1231171

paninaro:
Includes:
Tooltips added to buttons in the app header
spendCAT WalletConnect command now shows the CAT name/amount instead of XCH
Wallet transaction list column updates
Removed drop shadow from NFT "Update Pending" text in dark mode
NFT gallery: limit NFT selection to 10 when creating an offer
Fix EPERMS issue on Windows when loading NFTs
Absorb non-critical errors from getRemoteFileSize
support restoreMessages npm script on Windows
Updated localizations
Support editing wallet name from layout header
Expand offered fee section by default for imported offers
Expose suppressShareOnCreate as user visible pref in Notifications
Update window title and show an app status header when using testnet


wallentx (approved):
doot

wjblanke (approved):
aok

PR #14711, opened 2023-02-28
Fix Pool Wallet check for invalid state

aqk:
### Purpose:
Fix bug in Pool Wallet

### Current Behavior:
Pool wallet will reject a state change directly from one pool to another (when not passing through self-farming)

### New Behavior:
Fixed.

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -892,9 +892,9 @@ class PoolWallet:
 
         if self.target_state is None:
             return
-        if self.target_state == pool_wallet_info.current.state:
+        if self.target_state == pool_wallet_info.current:
             self.target_state = None
-            raise ValueError("Internal error")
+            raise ValueError(f"Internal error. Pool wallet {self.wallet_id} state: {pool_wallet_info.current}")
 
         if (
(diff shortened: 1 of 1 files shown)


altendky (approved):

chia/pools/pool_wallet.py
```diff
@@ -892,9 +892,9 @@ async def new_peak(self, peak_height: uint64) -> None:
 
         if self.target_state is None:
             return
-        if self.target_state == pool_wallet_info.current.state:
```
arvidn, line 895: was this just a typo? is there no field `state` in `pool_wallet_info.current`?

aqk: It was a typo. There is a field state in both, but what was intended is to compare the whole `PoolState`, NOT `PoolState.state`, which is a  `PoolSingletonState`. Better naming would help here.

The state machine for `PoolSingletonState` is very simple, but two `PoolState`s with different members are not equivalent.

emlowe (approved):

arvidn (approved):

wallentx (approved):

PR #14712, opened 2023-03-01
Soft fork2 tests

arvidn:
This patch makes two categories of changes:
1. it introduces aliases for `MempoolInclusionStatus`, `ConditionOpcode` and `ReceivedBlockResult` in tests. This reduces noise in the test parameters. In the upcoming assert-before patch, these test will see a doubling in the number of test cases, so this makes that a bit more palatable.
2. It makes some tests cover both with and without soft-fork 2 activated. Since we can't reasonably make a test chain reach all the way up to the soft-fork 2 cut-off, this means changing the constants instead, which means passing them down to `get_name_puzzle_condition()`. Since there aren't any test cases that depend on soft-fork2 (yet), all the existing tests still pass without change. This is preparing for additional test cases for assert-before conditions.

These commits are best reviewed one at a time. The change to `test_blockchain.py` that fall into category (2) introduce an extra level of indentation, and are hence best reviewed ignoring white space.

### DIFF
--- a/chia/consensus/block_body_validation.py
+++ b/chia/consensus/block_body_validation.py
@@ -319,6 +319,7 @@ async def validate_block_body(
                     cost_per_byte=constants.COST_PER_BYTE,
                     mempool_mode=False,
                     height=curr.height,
+                    constants=constants,
                 )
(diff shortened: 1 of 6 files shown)


AmineKhaldi (approved):
Looks good.

emlowe (approved):

PR #14717, opened 2023-03-02
work around issue in Python 3.9 import

aqk:


arvidn (approved):

altendky (changes requested):
Can we chat about this?

emlowe:
As an FYI, setproctitle was updated recently to 1.3.2 in https://github.com/Chia-Network/chia-blockchain/pull/13920

aqk:
Hey, this PR was started because I didn't understand how module in our code was being implicitly included in our include namespace. Kyle explained what is happening here.

This is an issue where Python auto-adds files in your local directory to your import path, possibly conflicting with the toplevel imports in Python's stdlib.

Happens if you are in the directory with a file that uses the same name of a module in the stdlib.

We'll try to find another way to work around this confusing behaviour.

PR #14719, opened 2023-03-02
fixup ephemeral coin test in test_mempool

arvidn:
This patch does 2 things:
1. uses shorter names for the constants in the test parameters for `test_ephemeral_timelock` (to make it easier to read)
2. fixes an old typo where the condition to test was applied both to the ephemeral coin *and* the coin creating it. It's meant to just test the ephemeral coin


AmineKhaldi (approved):
Looks good.

trepca (approved):

PR #14720, opened 2023-03-02
Bump chia_rs to 0.2.3. and ASSERT_MY_BIRTH_*

arvidn:
this patch bumps the chia_rs dependency to 0.2.3, which includes a number fixes and new features. Most importantly, it fixes a bug in the edge case parsing of `ASSERT_BEFORE_* conditions.

(The full changelog can be found [here](https://github.com/Chia-Network/chia_rs/releases/tag/0.2.3))

* The `Spend` type (as the result of parsing conditions) now include two optional integers, `birth_height` and `birth_seconds`, in case those conditions were specified. This affects tests that create objects of this type.
* The `SpendBundleConditions` now include two new fields, the sum of all addition amounts and removal amounts for the spend bundle. This can be used to easily compute the fee. This affects tests that create objects of this type.
* There are two new condition codes, `ASSERT_MY_BIRTH_HEIGHT` and `ASSERT_MY_BIRTH_SECONDS`
* There are two new error codes, `ASSERT_MY_BIRTH_SECONDS_FAILED` and `ASSERT_MY_BIRTH_HEIGHT_FAILED`
* There is new logic that implements the behavior of these new conditions. (it's fairly straight-forward, it just amends `mempool_check_time_lock()`

### DIFF
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -135,6 +135,12 @@ def mempool_check_time_locks(
 
     for spend in bundle_conds.spends:
         unspent = removal_coin_records[bytes32(spend.coin_id)]
+        if spend.birth_height is not None:
+            if spend.birth_height != unspent.confirmed_block_index:
+                return Err.ASSERT_MY_BIRTH_HEIGHT_FAILED
+        if spend.birth_seconds is not None:
+            if spend.birth_seconds != unspent.timestamp:
+                return Err.ASSERT_MY_BIRTH_SECONDS_FAILED
         if spend.height_relative is not None:
             if prev_transaction_block_height < unspent.confirmed_block_index + spend.height_relative:
                 return Err.ASSERT_HEIGHT_RELATIVE_FAILED
--- a/chia/types/condition_opcodes.py
+++ b/chia/types/condition_opcodes.py
@@ -33,6 +33,8 @@ class ConditionOpcode(bytes, enum.Enum):
     ASSERT_MY_PARENT_ID = bytes([71])
     ASSERT_MY_PUZZLEHASH = bytes([72])
     ASSERT_MY_AMOUNT = bytes([73])
+    ASSERT_MY_BIRTH_SECONDS = bytes([74])
+    ASSERT_MY_BIRTH_HEIGHT = bytes([75])
 
     # the conditions below ensure that we're "far enough" in the future
 
--- a/tests/blockchain/test_blockchain.py
+++ b/tests/blockchain/test_blockchain.py
@@ -1892,6 +1892,15 @@ class TestBodyValidation:
     @pytest.mark.parametrize(
         "opcode,lock_value,expected,with_garbage",
         [
+            (co.ASSERT_MY_BIRTH_HEIGHT, -1, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_HEIGHT, 0x100000000, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_HEIGHT, 2, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_HEIGHT, 3, rbr.NEW_PEAK, False),
+            (co.ASSERT_MY_BIRTH_SECONDS, -1, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_SECONDS, 0x10000000000000000, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_SECONDS, 10029, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_SECONDS, 10030, rbr.NEW_PEAK, False),
+            (co.ASSERT_MY_BIRTH_SECONDS, 10031, rbr.INVALID_BLOCK, False),
             (co.ASSERT_SECONDS_RELATIVE, -2, rbr.NEW_PEAK, False),
             (co.ASSERT_SECONDS_RELATIVE, -1, rbr.NEW_PEAK, False),
             (co.ASSERT_SECONDS_RELATIVE, 0, rbr.NEW_PEAK, False),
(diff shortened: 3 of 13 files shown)


chia/types/condition_opcodes.py
```diff
@@ -33,6 +33,8 @@ class ConditionOpcode(bytes, enum.Enum):
     ASSERT_MY_PARENT_ID = bytes([71])
     ASSERT_MY_PUZZLEHASH = bytes([72])
     ASSERT_MY_AMOUNT = bytes([73])
+    ASSERT_MY_BIRTH_SECONDS = bytes([74])
+    ASSERT_MY_BIRTH_HEIGHT = bytes([75])
```
arvidn, line 37: matching these: https://github.com/Chia-Network/chia_rs/blob/main/src/gen/opcodes.rs#L29-L30

AmineKhaldi (approved):
Looks good.

tests/core/mempool/test_mempool_manager.py
```diff
@@ -207,6 +225,46 @@ def test_seconds_absolute(
...
+        ],
+    )
+    def test_assert_my_birth_seconds(
+        self,
+        value: uint64,
+        expected_error: Optional[Err],
+    ) -> None:
+        conds = make_test_conds(birth_seconds=value)
+        assert (
+            mempool_check_time_locks(self.REMOVALS, conds, self.PREV_BLOCK_HEIGHT, self.PREV_BLOCK_TIMESTAMP)
```
trepca, line 264: is it possible to provide actual coin birth height when returning the error to simplify debugging?

arvidn: you mean it could return something like a `ValidationErr()` with the error code and include additional fields for the value? the function checks the `ASSERT_HEIGHT_*` and `ASSERT_SECONDS_*` conditions as well (and, as you can imagine, will check `ASSERT_BEFORE_*` conditions in a future patch).

This function is used in two places: the mempool, to gate transactions that aren't valid and in the block body validation for consensus, validating blocks. Any value returned from this function would need to be returned out of those contexts too, before reaching the user that's debugging. It's not obvious how to do that. logging might work, but on the other hand, when running a node you don't necessarily want to log just because one of your peers is submitting invalid transactions.

tests/blockchain/test_blockchain.py
```diff
@@ -1892,6 +1892,15 @@ async def test_aggsig_garbage(self, empty_blockchain, opcode, with_garbage, expe
...
         "opcode,lock_value,expected,with_garbage",
         [
+            (co.ASSERT_MY_BIRTH_HEIGHT, -1, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_HEIGHT, 0x100000000, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_HEIGHT, 2, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_HEIGHT, 3, rbr.NEW_PEAK, False),
+            (co.ASSERT_MY_BIRTH_SECONDS, -1, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_SECONDS, 0x10000000000000000, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_SECONDS, 10029, rbr.INVALID_BLOCK, False),
+            (co.ASSERT_MY_BIRTH_SECONDS, 10030, rbr.NEW_PEAK, False),
```
aqk, line 1902: How do we know `10030` ?

chia/full_node/mempool_check_conditions.py
```diff
@@ -135,6 +135,12 @@ def mempool_check_time_locks(
 
     for spend in bundle_conds.spends:
         unspent = removal_coin_records[bytes32(spend.coin_id)]
+        if spend.birth_height is not None:
+            if spend.birth_height != unspent.confirmed_block_index:
+                return Err.ASSERT_MY_BIRTH_HEIGHT_FAILED
+        if spend.birth_seconds is not None:
+            if spend.birth_seconds != unspent.timestamp:
+                return Err.ASSERT_MY_BIRTH_SECONDS_FAILED
```
aqk, line 143: I'm assuming `ASSERT_MY_BIRTH_SECONDS_FAILED` is from the timestamp in the block, ya?

tests/core/full_node/test_conditions.py
```diff
@@ -131,9 +131,19 @@ class TestConditions:
...
             # which end of the range they exceed
+            (co.ASSERT_MY_BIRTH_HEIGHT, -1, Err.ASSERT_MY_BIRTH_HEIGHT_FAILED),
+            (co.ASSERT_MY_BIRTH_HEIGHT, 0x100000000, Err.ASSERT_MY_BIRTH_HEIGHT_FAILED),
+            (co.ASSERT_MY_BIRTH_HEIGHT, 3, Err.ASSERT_MY_BIRTH_HEIGHT_FAILED),
+            (co.ASSERT_MY_BIRTH_HEIGHT, 2, None),
+            # genesis timestamp is 10000 and each block is 10 seconds
+            (co.ASSERT_MY_BIRTH_SECONDS, -1, Err.ASSERT_MY_BIRTH_SECONDS_FAILED),
+            (co.ASSERT_MY_BIRTH_SECONDS, 0x10000000000000000, Err.ASSERT_MY_BIRTH_SECONDS_FAILED),
+            (co.ASSERT_MY_BIRTH_SECONDS, 10019, Err.ASSERT_MY_BIRTH_SECONDS_FAILED),
+            (co.ASSERT_MY_BIRTH_SECONDS, 10020, None),
```
aqk, line 145: I see. We must be adding 10 seconds per virtual block somewhere above.

chia/full_node/mempool_check_conditions.py
arvidn: I'm not sure I know what you mean. the `spend` are the conditions imposed on the spend being valid. the `unspent` is the `CoinRecord` of the coin we're spending. The timestamp in the `CoinRecord` is supposed to match the timestamp in the block that confirmed that coin. (in fact, we store this timestamp in the coin_store for every record, which is quite wasteful)

tests/blockchain/test_blockchain.py
arvidn: yeah, this isn't super visible in this test unfortunately. there are some comments further down:

> genesis timestamp is 10000 and each block is 10 seconds

Also, on line 1851, we specify 10 seconds per block, and we generate 3 blocks. So this is the 4th block (height 3)

chia/full_node/mempool_check_conditions.py
aqk: Thanks, I meant to say ASSERT_MY_BIRTH_SECONDS. I couldn't find a good place to put this question when I asked it.

arvidn: both `ASSERT_MY_BIRTH_*` conditions ensure that the coin being spent was confirmed on chain at a specific height (or the timestamp of the block at that height)

aqk (approved):

emlowe (approved):

PR #14730, opened 2023-03-03
Add support to cat_spend RPC for running TAIL

Quexington:
### Purpose:
This adds support for running the TAIL (for minting/melting) to the standard RPC for spending CATs.  A tail reveal and solution are required for this call.  Potentially more ergonomic actions could be added that make running of specific TAILs easier.


chia/rpc/wallet_rpc_api.py
```diff
@@ -1487,12 +1487,31 @@ async def cat_spend(self, request, hold_lock=True) -> EndpointResult:
             }
         else:
             exclude_coins = None
+        cat_discrepancy_params: Tuple[Optional[int], Optional[str], Optional[str]] = (
```
trepca, line 1490: why "discrepancy"?

Quexington: "discrepancy in accounting" is how we describe extra delta and the related calculations that need to be adjusted while that CATs are making their accounting announcements

trepca (approved):

Issue #14557, opened 2023-02-13
[Bug] Seeder is not listing on IPv6


hoffmang9:
### What happened?

DNS Seeder is only binding to port 53 on the IPv4 stack and thus if it is queried on it's IPv6 interface, the connection times out.

https://github.com/Chia-Network/chia-blockchain/blob/main/chia/seeder/dns_server.py#L96

It does fall back to answering on v4 to give out v6 addresses.

Compare `dig -t AAAA chia.hoffmang.com` and `dig -4 chia.hoffmang.com @chia-dns.hoffmang.com` to `dig -6 chia.hoffmang.com @chia-dns.hoffmang.com`

HT to @denisu for catching this.

### Version

1.6.2 and main

### What platform are you using?

Linux

### What ui mode are you using?

CLI

emlowe:
Fixed in 1.7.1 by #14731

PR #14731, opened 2023-03-03
Listen on ::0 so we can respond to both ipv4 and ipv6

cmmarslender:
### Purpose:
Have the DNS server listen on both ipv4 and ipv6. Also moves the 60 second sleep between refreshing peer IPs to AFTER the first attempt to load them, so that if we already have a crawler DB present, we don't have to wait 60 seconds to start responding to DNS requests.

### Current Behavior:
Currently, it only listens on ipv4

### New Behavior:
Listens on both ipv4 and ipv6

### Testing Notes:
~I'm not entirely sure what happens on an ipv4 only host with this change~. Tested on a host with no public ipv6 address, and it all still worked fine on ipv4.

I tested empty string, `None`, and a few other variants, and this was the only variant that ended up supporting both ipv4 and ipv6.

### DIFF
--- a/chia/seeder/dns_server.py
+++ b/chia/seeder/dns_server.py
@@ -100,8 +100,6 @@ class DNSServer:
     async def periodically_get_reliable_peers(self):
         sleep_interval = 0
         while True:
-            sleep_interval = min(15, sleep_interval + 1)
-            await asyncio.sleep(sleep_interval * 60)
             try:
                 # TODO: double check this. It shouldn't take this long to connect.
                 crawl_db = await aiosqlite.connect(self.db_path, timeout=600)
@@ -142,6 +140,9 @@ class DNSServer:
                 )
             except Exception as e:
                 log.error(f"Exception: {e}. Traceback: {traceback.format_exc()}.")
+
+            sleep_interval = min(15, sleep_interval + 1)
+            await asyncio.sleep(sleep_interval * 60)
 
     async def get_peers_to_respond(self, ipv4_count, ipv6_count):
         peers = []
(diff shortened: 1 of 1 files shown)


hoffmang9:
Running now in "production" on chia.hoffmang.com and everything is working great. `dig -6 chia.hoffmang.com @chia-dns.hoffmang.com` now works just fine. Fixes [14557](https://github.com/Chia-Network/chia-blockchain/issues/14557)

hoffmang9 (approved):
works good for me pending CI

chia/seeder/dns_server.py
```diff
@@ -143,6 +141,9 @@ async def periodically_get_reliable_peers(self):
             except Exception as e:
                 log.error(f"Exception: {e}. Traceback: {traceback.format_exc()}.")
 
+            sleep_interval = min(15, sleep_interval + 1)
```
wjblanke, line 144: is there a reason this is moved to the end?

cmmarslender: Yeah, mentioned above, but it didn't seem like we should be waiting 60 seconds every time we start the seeder to even try to look up reliable peers, so I moved the first sleep to the _end_ of the first loop. This way, when we're restarting the seeder or starting from a DB backup of existing peers, we can start serving DNS replies sooner. I can split this out to a separate PR if you want to keep them isolated, it was just something that had been bugging me for a while that I fixed while I was in there.

wjblanke: the DB backup of existing peers is supposed to thwart eclipse attack. just not sure if this change unwinds that

cmmarslender: All the moving of the sleep does is change startup from:

1. Sleep 60 seconds
2. Check DB for reliable peers
3. Start serving DNS responses, if there were peers in the DB

To:

1. Check DB for reliable peers
2. Start serving DNS responses, if there were peers in the DB
3. Sleep for 60 seconds

wjblanke (approved):
ok just making sure

PR #14740, opened 2023-03-04
correct Announcement.__str__() to present hex

altendky:
### Purpose:

A hash is quite unlikely to be valid UTF-8 data and even if it does decode it seems unlikely to be a useful representation.

### Current Behavior:

Converting an `Announcement` to a `str` is likely to result in an exception.

### New Behavior:

Converting an `Announcement` to a `str` results in the hexadecimal representation of the bytes.


arvidn (approved):
wow. I imagine this would just throw an error in most cases

PR #14767, opened 2023-03-07
catch up with mypy 1.1.1

altendky:

### DIFF
--- a/chia/full_node/weight_proof.py
+++ b/chia/full_node/weight_proof.py
@@ -1221,7 +1221,7 @@ def validate_recent_blocks(
     ses_blocks, sub_slots, transaction_blocks = 0, 0, 0
     challenge, prev_challenge = recent_chain.recent_chain_data[0].reward_chain_block.pos_ss_cc_challenge_hash, None
     tip_height = recent_chain.recent_chain_data[-1].height
-    prev_block_record = None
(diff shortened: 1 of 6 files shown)


chia/full_node/weight_proof.py
```diff
@@ -1245,10 +1245,10 @@ def validate_recent_blocks(
         if (challenge is not None) and (prev_challenge is not None):
             overflow = is_overflow_block(constants, block.reward_chain_block.signage_point_index)
             if not adjusted:
+                assert prev_block_record is not None
```
emlowe, line 1248: At this line, isn't prev_block_record guaranteed to be None, since it was set to None on line 1224?

emlowe (changes requested):

chia/full_node/weight_proof.py
altendky: Looks like a couple paths anyways hit an `except Exception:` and others don't, but I didn't find a case where it would catch `TypeError` but not `AssertionError` so I think this is a no-op change give or take.

emlowe: The dataclasses.replace line would raise on None in any event, so this is ok

emlowe (approved):

Issue #14744, opened 2023-03-04
[Bug] Token-vs-NFT offer accept fails when token is just 1 mojo


andyharless:
### What happened?

I created an offer file ([posted on Dexie](https://dexie.space/offers/FUw88cavoMZtd1RSVCdjzBz755adwnkEEMxhxwH6iSbp)) to sell an NFT for 0.001 of a token (Asset ID 0fe711bf57d0222603a8d3e13df54b29382373ab2b05261c69667360a5738d64). I tried to accept the offer form another account. After clicking the "accept offer" button, I got "Error ('clvm raise', '80')". (That's from the Chia GUI. I got a different error trying to accept it from a different account on Pawket.) The offer is apparently still active, and the GUI still lets me click the button and gives the same error. Factors that may be relevant:

- There was a 50000 mojo fee included in the offer—thus many more mojos than the offered transaction itself.
- There was a 4% royalty on the NFT. It appeared that this was rounded down to zero, as it should have been, but perhaps the inability to calculate a positive royalty created a problem.

### Version

1.7.0

### What platform are you using?

Linux

### What ui mode are you using?

GUI

### Relevant log output

```shell
2023-03-04T15:57:52.701 wallet chia.wallet.wallet_state_manager: INFO     Start: Creating puzzle hashes from 730 to 730 for wallet_id: 11
2023-03-04T15:57:52.707 wallet chia.wallet.wallet_state_manager: INFO     Done: Creating puzzle hashes from 730 to 730 for wallet_id: 11 Time: 0.07550764083862305 seconds
2023-03-04T15:57:52.710 wallet chia.wallet.wallet_state_manager: INFO     Updating last used derivation index: 305
2023-03-04T15:57:52.717 wallet chia.wallet.cat_wallet.cat_wallet: INFO     Confirmed balance for cat wallet 11 is 11
2023-03-04T15:57:52.767 wallet chia.wallet.trade_manager  : ERROR    Error with creating trade offer: <class 'ValueError'>Traceback (most recent call last):
  File "chia/wallet/trade_manager.py", line 519, in _create_offer_for_ids
  File "chia/wallet/trade_manager.py", line 812, in check_for_special_offer_making
  File "chia/wallet/nft_wallet/nft_wallet.py", line 1048, in make_nft1_offer
  File "chia/wallet/outer_puzzles.py", line 58, in solve_puzzle
  File "chia/wallet/cat_wallet/cat_outer_puzzle.py", line 119, in solve
[...]

PR #14769, opened 2023-03-07
Don't create a coin for 0 amount royalty payments

Quexington:
### Purpose:

An inadvertent side effect of upgrading the OFFER_MOD was that 0 coins are no longer allowed as outputs.  This breaks our current system of paying royalties that are so small that they get rounded down to 0.  We used to create a coin for this anyway, so this PR just removes that behavior and simply does not recognize the zero payment as a payment that needs to be made.  This makes offers with royalties involved a little bit cheaper in some instances.

https://github.com/Chia-Network/chia-blockchain/issues/14744

### DIFF
--- a/tests/wallet/nft_wallet/test_nft_1_offers.py
+++ b/tests/wallet/nft_wallet/test_nft_1_offers.py
@@ -4,7 +4,7 @@ import asyncio
 import logging
 import sys
 from secrets import token_bytes
-from typing import Any, Callable, Coroutine, Tuple, Optional
+from typing import Any, Callable, Coroutine, Optional, Tuple
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


tests/wallet/nft_wallet/test_nft_1_offers.py
```diff
@@ -55,14 +55,16 @@ async def get_nft_count(wallet: NFTWallet) -> int:
     [True, False],
 )
 @pytest.mark.parametrize(
-    "forwards_compat",
-    [True, False],
+    "forwards_compat_and_zero",
+    [(True, False), (False, False), (False, True)],
```
arvidn, line 59: you can make these separate parameters to the test function directly, by:
```
    "forwards_compat,zero",
    [(True, False), (False, False), (False, True)],
```

and:

```
async def test_nft_offer_sell_nft(
    self_hostname: str, two_wallet_nodes: Any, trusted: Any, forwards_compat: bool, zero: bool
) -> None:
```

tests/wallet/nft_wallet/test_nft_1_offers.py
```diff
@@ -1390,9 +1400,9 @@ async def test_complex_nft_offer(
     target_puzhash_taker = ph_taker
     royalty_puzhash_maker = ph_maker
     royalty_puzhash_taker = ph_taker
-    royalty_basis_pts_maker = uint16(200)
-    royalty_basis_pts_taker_1 = uint16(500)
-    royalty_basis_pts_taker_2 = uint16(500)
+    royalty_basis_pts_maker = uint16(0 if zero_royalties else 200)
+    royalty_basis_pts_taker_1 = uint16(0 if zero_royalties else 500)
+    royalty_basis_pts_taker_2 = uint16(0 if zero_royalties else 500)
```
arvidn, line 1405: it might make sense to make these numbers separate test parameters. that way you can easily add more combinations to test

tests/wallet/nft_wallet/test_nft_1_offers.py
```diff
@@ -4,7 +4,7 @@
 import logging
 import sys
 from secrets import token_bytes
-from typing import Any, Callable, Coroutine, Optional
+from typing import Any, Callable, Coroutine, Tuple, Optional
```
altendky, line 7: For isort.

```suggestion
from typing import Any, Callable, Coroutine, Optional, Tuple
```

trepca (approved):

PR #14776, opened 2023-03-08
Cleanup tests

arvidn:
This PR offers some simplifications, and one bugfix, to tests.

1. the `test_ephemeral_timelock()` ensures unknown additional "garbage" parameters are ignored on known conditions, in consensus mode. This `with_garbage` test parameter was split out to be independent of the other test parameters, to no steal attention from the core of the test cases.
2. The mempool manager's `test_ephemeral_timelock()` had a bug where it would apply the conditions under test to both the coin creating the ephemeral coin, as well as the ephemeral coin itself. It's only supposed to apply the conditions to the ephemeral coin, since that's what's being tested
3. a test that deliberately uses an unknown condition opcode was made clearer what the intention is (and to change the condition opcode away from a value that's about to become a known opcode)


AmineKhaldi (approved):
Looks good.

tests/core/test_cost_calculation.py
```diff
@@ -138,10 +138,11 @@ async def test_mempool_mode(self, softfork_height, bt):
         )
         puzzle = p2_delegated_puzzle_or_hidden_puzzle.puzzle_for_pk(G1Element.from_bytes(pk))
         disassembly = binutils.disassemble(puzzle)
+        unknown_opcode = "15"
```
trepca, line 141: why is it better to have a variable for this vs hardcoding?

trepca (approved):

PR #14777, opened 2023-03-08
full_node: Drop cache logs in `BlockStore`

xdustinface:
### Purpose:

Reduce some `DEBUG` log spam from `BlockStore`. 

### Current Behavior:

At least in tests i found this cache hit messages pretty spammy and i don't think it brings much value to know we got a cache hit in this cases? 

### New Behavior:

Instead this PR adds two more cache miss logs and adds the name of the function to the other cache miss logs.

### DIFF
--- a/chia/full_node/block_store.py
+++ b/chia/full_node/block_store.py
@@ -279,9 +279,7 @@ class BlockStore:
     async def get_full_block(self, header_hash: bytes32) -> Optional[FullBlock]:
         cached: Optional[FullBlock] = self.block_cache.get(header_hash)
         if cached is not None:
-            log.debug(f"cache hit for block {header_hash.hex()}")
             return cached
-        log.debug(f"cache miss for block {header_hash.hex()}")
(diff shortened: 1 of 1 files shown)


arvidn:
I'm not opposed to reducing the log spam. However, the rationale for these was to be able to get an idea of the cache hit ratio. I still think that would be useful (and just logging cache miss might not be so useful without it).

Unless there is a good reason to log cache misses only, I would be in favor of removing the logging altogether.

If we need the logs, I would also suggest you move the formatting to use a format string, to avoid that cost when debug logging is disabled.

xdustinface:
> Unless there is a good reason to log cache misses only, I would be in favor of removing the logging altogether.

Not so sure if its a _good_ reason but the thought i had was just, if there are lot cache misses popping up it might be good to look into it no matter how many cache hits we got. But sure, im fine dropping it all together also, let me know what you think, i can change it.

arvidn:
I'm in favor of just removing it

arvidn (approved):

PR #14782, opened 2023-03-08
Added `chia completion` command

wallentx:
### Purpose:
Add CLI command to generate shell completion for Chia

### Current Behavior:
No guidance on how to do this

### New Behavior:
Now provides a way

### DIFF
--- a/chia/cmds/completion.py
+++ b/chia/cmds/completion.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import os
 import subprocess
+import sys
+from pathlib import Path
 
 import click
 
@@ -45,4 +47,4 @@ def generate(shell: str) -> None:
     """
     # Could consider calling directly in the future.
     # https://github.com/pallets/click/blob/ef11be6e49e19a055fe7e5a89f0f1f4062c68dba/src/click/shell_completion.py#L17
-    subprocess.run([sys.executable, "-m", "chia"], check=True, env={**os.environ, "_CHIA_COMPLETE": f"{shell}_source")
+    subprocess.run([sys.executable, "-m", "chia"], check=True, env={**os.environ, "_CHIA_COMPLETE": f"{shell}_source"})
--- a/chia/cmds/completion.py
+++ b/chia/cmds/completion.py
@@ -5,9 +5,10 @@ import subprocess
 
 import click
 
-SHELLS = ['bash', 'zsh', 'fish']
-shell = os.environ.get('SHELL', '').split('/')[-1]
-cmd = f'_CHIA_COMPLETE={shell}_source chia'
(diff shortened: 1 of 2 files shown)


chia/cmds/completion.py
```diff
@@ -0,0 +1,35 @@
...
+    help="Shell type to generate completion for",
+)
+def generate(shell: str) -> None:
+    """
+    Generate shell completion script
+    """
+    if shell is None:
+        shell = os.environ.get('SHELL', '').split('/')[-1]
+    if shell not in SHELLS:
+        raise click.UsageError(f'Invalid shell type: {shell}')
```
altendky, line 29: I think adjusting above eliminates both of these checks.

chia/cmds/completion.py
```diff
@@ -0,0 +1,35 @@
...
+    cmd = f'_CHIA_COMPLETE={shell}_source chia'
+    print("Save the following output as '~/.chia-completion."+shell+"' and source it according to your shell:\n")
```
altendky, line 31: We should be able to just get the output for use in automation like `chia completion generate >> place` or to just call on the fly like `eval $(chia completion generate)` or whatever it would be.

chia/cmds/completion.py
```diff
@@ -0,0 +1,35 @@
...
+    subprocess.run(cmd, shell=True, check=True)
+
+if __name__ == '__main__':
+    completion()
```
altendky, line 35: Shouldn't be run directly, in my opinion.

```suggestion
```

altendky (changes requested):

xdustinface (commented):
I didn't try it or look close at it so maybe i miss some detail but after a quick scan i wonder if it wouldn't make more sense to automatically set this up as part of the install process?

wallentx:
> I didn't try it or look close at it so maybe i miss some detail but after a quick scan i wonder if it wouldn't make more sense to automatically set this up as part of the install process?

@xdustinface I tried this in several different ways, but there's a problem of accurate shell detection when executing the install script. It would be ideal if we could write this to the `activate` file, but that doesn't solve for people who are installing from our installers.

I think we might be able to figure out something that is an automatic solution that this PR will enable us to explore, but perhaps in a different PR.

chia/cmds/completion.py
```diff
@@ -0,0 +1,42 @@
...
+    pass
+
+
+@completion.command(short_help="Generate shell completion code")
+@click.option(
+    "-s",
+    "--shell",
+    type=click.Choice(SHELLS),
+    default=shell,
+    show_default=False,
```
altendky, line 26:
```suggestion
    show_default=True,
```

altendky (approved):

paninaro (approved):

PR #14786, opened 2023-03-09
disable soft-fork 2 (for now)

arvidn:
A few questions have come up about how to handle ephemeral coin spends in combination with time-lock conditions that we need some more time to resolve. This disables the soft-fork 2, for now. We will enable it again once we know how we want to move forward.

### DIFF
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 from typing import Dict, List, Optional, Tuple
 
-from chia_rs import ENABLE_ASSERT_BEFORE, LIMIT_STACK, MEMPOOL_MODE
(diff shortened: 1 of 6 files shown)


chia/full_node/mempool_check_conditions.py
```diff
@@ -53,8 +53,9 @@ def get_name_puzzle_conditions(
     else:
         flags = 0
 
-    if height is not None and height >= constants.SOFT_FORK2_HEIGHT:
-        flags = flags | ENABLE_ASSERT_BEFORE
+    # soft-fork2 is disabled (for now)
+    # if height is not None and height >= constants.SOFT_FORK2_HEIGHT:
+    #    flags = flags | ENABLE_ASSERT_BEFORE
```
arvidn, line 58: this is the material change

AmineKhaldi (approved):
Looks reasonable.

emlowe (approved):

PR #14792, opened 2023-03-09
Wallet class interface discussion

aqk:
This PR is being used to discuss an improvement to the current `Wallet` class interface.

### DIFF
--- a/chia/wallet/singleton_wallet.py
+++ b/chia/wallet/singleton_wallet.py
@@ -1,0 +1,92 @@
+from __future__ import annotations
+
+from typing import List, Tuple
+
+from chia_rs import CoinSpend
+
+from chia.types.blockchain_format.coin import Coin
+from chia.types.blockchain_format.program import Program
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.ints import uint32
+from chia.wallet.transaction_record import TransactionRecord
+from chia.wallet.wallet import Wallet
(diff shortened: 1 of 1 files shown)


chia/wallet/singleton_wallet.py
```diff
@@ -0,0 +1,92 @@
...
+    - create new wallet for singleton that already exists (we have not yet received the coin)
+    - spending when you don't own the singleton, but you can spend
+      - Examples:
+    """
+
+    ###################### Class methods ######################################
+
+    # Name: detect_singleton_coin_type ?
+    @classmethod
+    def coin_is_my_singleton_wallet_type(cls, coin: Coin, inner_puzzle: Program) -> bool:
```
aqk, line 52: Matt mentioned that we can use this approach to deal with all outer / inner style coin wallets, not just Singleton Wallets, and make the code even more uniform.

chia/wallet/singleton_wallet.py
```diff
@@ -0,0 +1,92 @@
...
+        """Match the singleton inner puzzle for this type of wallet"""
+        ...
+
+    # def register_singleton_inner_puzzle(self) -> Program:
+    #    ...
+
+    @classmethod
+    def create_wallet_from_coin(
```
aqk, line 60: Do we want the in-memory wallet : in-db state table relationship to be 1:1 ?

Dow we want to allow one `Wallet` object to handle multiple layered coins in the simple interface?

PR #14798, opened 2023-03-10
Add check_ids API

ytx1991:
### Purpose:
https://github.com/Chia-Network/issue-tracker/issues/332
Add an new API to check asset ownership.

### Current Behavior:
None

### New Behavior:
Call check_ids with a list of IDs and asset type

### Testing Notes:
Unit test

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -141,6 +141,7 @@ class WalletRpcApi:
             "/sign_message_by_address": self.sign_message_by_address,
             "/sign_message_by_id": self.sign_message_by_id,
             "/verify_signature": self.verify_signature,
+            "/check_ids": self.check_ids,
             "/get_transaction_memo": self.get_transaction_memo,
             # CATs and trading
             "/cat_set_name": self.cat_set_name,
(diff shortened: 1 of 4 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1334,6 +1335,46 @@ async def verify_signature(self, request) -> EndpointResult:
         else:
             return {"isValid": False, "error": "Signature is invalid."}
 
+    async def check_ids(self, request) -> EndpointResult:
+        """
+        Check if a list of IDs is owned by the current wallet.
+        :param request: {"ids":["ID1", "ID2", ...], "type":"COIN/NFT/DID"}
```
trepca, line 1341: wouldn't it be easier to provide a list of `(coin_id, coin_type)` vs doing multiple requests depending on the coin type?

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -896,7 +896,63 @@ def only_ids(trades):
...
+    nft_id = (await nft_wallet.get_current_nfts())[0].nft_id.hex()
+    res = await wallet_1_rpc.check_ids([nft_id], "NFT")
+    assert len(res) == 1
+    assert res[0] == nft_id
+
+    # COIN
+    coin_id = (await nft_wallet.get_current_nfts())[0].coin.name().hex()
+    res = await wallet_1_rpc.check_ids([coin_id], "COIN")
+    assert len(res) == 1
+    assert res[0] == coin_id
```
trepca, line 955: add a test to check if it doesn't return coins that it doesn't own but they should exist

chia/rpc/wallet_rpc_api.py
```diff
@@ -1334,6 +1335,46 @@ async def verify_signature(self, request) -> EndpointResult:
...
+        if "type" not in request:
+            raise ValueError("Type is required by check_ids.")
+        if request["type"].upper() == "COIN":
+            ids: List[bytes32] = [bytes32.fromhex(coin_id) for coin_id in request["ids"]]
+            coin_records: List[
+                WalletCoinRecord
+            ] = await self.service.wallet_state_manager.coin_store.get_multiple_coin_records(ids)
+            valid_ids = [r.coin.name().hex() for r in coin_records]
+        elif request["type"].upper() == "NFT":
+            ids = [
```
trepca, line 1355: @paninaro how many ids do you plan to check at a time?

paninaro: The primary use case I have in mind would be for checking if NFTs in an offer are owned by the user. We have a cap of ~10 NFTs in an offer when created from the GUI. So I would say ~20 would probably be the max.

chia/rpc/wallet_rpc_api.py
paninaro: I agree that being able to query a mix of coin types in a single request would be a nice.

Do we even need to specify the type if we can infer the coin type by the bech32m prefix? `nft1...abcxyz` for NFT, `did:chia:1...def123` for DID, and anything else could be looked up as a coin. Or is there a reason to have the type specified by the caller?

ytx1991: The question is if we want to support hex string for NFT and DID. Also, if the type is a known thing, providing the type will speed up the query.

tests/wallet/rpc/test_wallet_rpc.py
ytx1991: I thought only the coins we owned will be saved in the table. What is the reason for saving the coin we don't own?

PR #13320, opened 2022-09-03
Ms.mypy enable

mariano54:
Full node directory and a few utilities.
I have manually performed each change carefully, so this change is low risk.

PR #14803, opened 2023-03-11
avoid ValueErrors when FullNode.fetch_block_batches() when removing already removed peer

altendky:
### Purpose:

Please weigh in on whether this should be applied to the release or retargeted to `main`.

Avoid both the exception message below and also unexpected termination of the function.

```
2023-03-11T04:54:23.167 full_node chia.full_node.full_node: ERROR    Exception fetching 452768 to 452800 from peer list.remove(x): x not in list
```

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -1075,12 +1075,18 @@ class FullNode:
                     fetched = False
                     for peer in random.sample(new_peers_with_peak, len(new_peers_with_peak)):
                         if peer.closed:
-                            peers_with_peak.remove(peer)
+                            try:
+                                peers_with_peak.remove(peer)
+                            except ValueError:
+                                pass
                             continue
                         response = await peer.call_api(FullNodeAPI.request_blocks, request, timeout=30)
                         if response is None:
                             await peer.close()
-                            peers_with_peak.remove(peer)
+                            try:
+                                peers_with_peak.remove(peer)
+                            except ValueError:
+                                pass
                         elif isinstance(response, RespondBlocks):
                             await batch_queue.put((peer, response.blocks))
                             fetched = True


xdustinface (changes requested):
Hm, looking at the surrounding code this removal and the usage of `peers_with_peak ` seems strange to begin with. We remove from `peers_with_peak` which is created outside the task and not longer used anywhere else because we copy it into `new_peers_with_peak` in the task https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1070
Yes its "used" in the validation task to remove peers but this doesn't really make sense imo. I think we can also just drop those two lines?
https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1117-L1118
It seems its enough to just close the connection there, what we already do https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1119 which then triggers the `SyncStore.peer_changed` event which then leads to `new_peers_with_peak` being  overwritten here https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1093

almogdepaz:
> Hm, looking at the surrounding code this removal and the usage of `peers_with_peak ` seems strange to begin with. We remove from `peers_with_peak` which is created outside the task and not longer used anywhere else because we copy it into `new_peers_with_peak` in the task
> 
> https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1070
> 
> 
> Yes its "used" in the validation task to remove peers but this doesn't really make sense imo. I think we can also just drop those two lines?
> https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1117-L1118
> 
> 
> It seems its enough to just close the connection there, what we already do
> https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1119
> 
> which then triggers the `SyncStore.peer_changed` event which then leads to `new_peers_with_peak` being overwritten here
> https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1093

i think i agree with you, when i wrote it originally we dident use this new_peers_with_peak it looks like Mariano added at at some point but im not sure i understand why

altendky:
My goal here was to quiet down the logs so when monitoring for errors we can see errors.  This is certainly just one step, but it's a step.  Does this make things significantly worse?  Would you like to provide another refactor that avoids the irrelevant `ERROR` log entry?

almogdepaz:
> My goal here was to quiet down the logs so when monitoring for errors we can see errors. This is certainly just one step, but it's a step. Does this make things significantly worse? Would you like to provide another refactor that avoids the irrelevant `ERROR` log entry?

actually  we use the copy of the list so we dont remove from the same list while we are iterating on it, i would just merge as is

almogdepaz:
> > Hm, looking at the surrounding code this removal and the usage of `peers_with_peak ` seems strange to begin with. We remove from `peers_with_peak` which is created outside the task and not longer used anywhere else because we copy it into `new_peers_with_peak` in the task
> > https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1070
> > 
> > Yes its "used" in the validation task to remove peers but this doesn't really make sense imo. I think we can also just drop those two lines?
> > https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1117-L1118
> > 
> > It seems its enough to just close the connection there, what we already do
> > https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1119
> > 
> > which then triggers the `SyncStore.peer_changed` event which then leads to `new_peers_with_peak` being overwritten here
> > https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1093
> 
> i think i agree with you, when i wrote it originally we dident use this new_peers_with_peak it looks like Mariano added at at some point but im not sure i understand why, was changed here #13320

@xdustinface actually we use the copy of the list so we dont remove from the same list while we are iterating on it, i would just merge as is

almogdepaz (approved):

xdustinface:
> @xdustinface actually we use the copy of the list so we dont remove from the same list while we are iterating on it, in python that would shift items to the left while we are still working on the list, i would just merge as is

I don't see where we would remove while iterating. We iterate over `random.sample(new_peers_with_peak)` which returns a new list. Also, like i pointed out initially, `new_peers_with_peak` gets anyway overwritten every time `SyncStore.peers_changed` gets triggered so removal from `peers_with_peak` doesn't make sense at all. See https://github.com/Chia-Network/chia-blockchain/blob/aafe234cd29c0c4aeef7b53a8a0875477cc918a6/chia/full_node/full_node.py#L1093

I just opened #15165 as alternative to this PR.

PR #14804, opened 2023-03-11
avoid FullNode.fetch_block_batches() from putting two `None`s into the batch queue

altendky:
### Purpose:

Please weigh in on whether this should be applied to the release or retargeted to main.

The `finally:` clause will already result in `None` being put into the queue.  I expect that the other `None` pushed before the `return` is unnecessarily doubling this up.  I don't know if the double `None` has any negative effect.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -1087,7 +1087,6 @@ class FullNode:
                             break
                     if fetched is False:
                         self.log.error(f"failed fetching {start_height} to {end_height} from peers")
-                        await batch_queue.put(None)
                         return
                     if self.sync_store.peers_changed.is_set():
                         new_peers_with_peak = self.get_peers_with_peak(peak_hash)


xdustinface (dismissed):
Make sense to me but i don't feel like it's important for release since from my understanding nothing changes, the validation task will return if a `None` gets pulled from the queue. If there is a duplicated `None` in there it will just stay in there but since the queue is temporarily created and not reused it doesn't matter?

almogdepaz (approved):

almogdepaz:
> Make sense to me but i don't feel like it's important for release since from my understanding nothing changes, the validation task will return if a `None` gets pulled from the queue. If there is a duplicated `None` in there it will just stay in there but since the queue is temporarily created and not reused it doesn't matter?

doesn't do anything but its a bug

emlowe (approved):

PR #14812, opened 2023-03-13
better handling of offer status and failed txs

### LINKED ISSUE #14632: [Bug] cancel_offer in 1.7 of an offer in pending cancel state reports error
### What happened?

I am unable to cancel an offer that is stuck in a Pending Cancel state with Chia 1.7.0 wallet.

```
chia wallet cancel_offer -id 55a6c69a9b335a30933abf2b2d04e08f3d94a1a37f82bb5f4b180dbd8459cc3d -m 0.001 -f 1096298047
```

I get the following error within the terminal:
```
Exception from 'wallet' 6:
  File "/opt/homebrew/lib/python3.10/site-packages/chia/cmds/cmds_util.py", line 100, in get_any_service_client
  File "/opt/homebrew/lib/python3.10/site-packages/chia/cmds/cmds_util.py", line 211, in execute_with_wallet
    await function(extra_params, wallet_client, new_fp)
  File "/opt/homebrew/lib/python3.10/site-packages/chia/cmds/wallet_funcs.py", line 691, in cancel_offer
    trade_record = await wallet_client.get_offer(id, file_contents=True)
  File "/opt/homebrew/lib/python3.10/site-packages/chia/rpc/wallet_rpc_client.py", line 706, in get_offer
    offer_str = bytes(Offer.from_bech32(res["offer"])).hex() if file_contents else ""
  File "/opt/homebrew/lib/python3.10/site-packages/chia/wallet/trading/offer.py", line 599, in from_bech32
    return cls.try_offer_decompression(decoded_bytes)
  File "/opt/homebrew/lib/python3.10/site-packages/chia/wallet/trading/offer.py", line 582, in try_offer_decompression
    return cls.from_compressed(offer_bytes)
  File "/opt/homebrew/lib/python3.10/site-packages/chia/wallet/trading/offer.py", line 577, in from_compressed
    return Offer.from_bytes(decompress_object_with_puzzles(compressed_bytes))
  File "/opt/homebrew/lib/python3.10/site-packages/chia/wallet/util/puzzle_compression.py", line 72, in decompress_object_with_puzzles
    raise CompressionVersionError(version)
chia.wallet.util.puzzle_compression.CompressionVersionError: 6
```
 This needs better error handling to tell me what's going on.

1.6.2 wallet cannot see the offer file correctly. Does not mention a CAT2.
```
chia wallet get_offers -f 1096298047 -id 55a6c69a9b335a30933abf2b2d04e08f3d94a1a37f82bb5f4b180dbd8459cc3d -s

Record with id: 55a6c69a9b335a30933abf2b2d04e08f3d94a1a37f82bb5f4b180dbd8459cc3d
---------------
Created at: 2023-02-20 14:53:05
Confirmed at: Not confirmed
Accepted at: N/A
Status: PENDING_CANCEL
Summary:
  OFFERED:
  REQUESTED:
    - XCH (Wallet ID: 1): 5.455785405774 (5455785405774 mojos)
Pending Outbound Balances:
    - Unknown: 240000 mojos
Included Fees: 0
---------------
```

1.7.0 see's the offer file in a pending state within the Offers table and in the CLI.

```
chia wallet get_offers -f 1096298047 -id 55a6c69a9b335a30933abf2b2d04e08f3d94a1a37f82bb5f4b180dbd8459cc3d -s
Exception from 'wallet' 6:
  File "/opt/homebrew/lib/python3.10/site-packages/chia/cmds/cmds_util.py", line 100, in get_any_service_client
  File "/opt/homebrew/lib/python3.10/site-packages/chia/cmds/cmds_util.py", line 211, in execute_with_wallet
    await function(extra_params, wallet_client, new_fp)
  File "/opt/homebrew/lib/python3.10/site-packages/chia/cmds/wallet_funcs.py", line 564, in get_offers
    records = [await wallet_client.get_offer(bytes32.from_hexstr(id), file_contents)]
  File "/opt/homebrew/lib/python3.10/site-packages/chia/rpc/wallet_rpc_client.py", line 706, in get_offer
    offer_str = bytes(Offer.from_bech32(res["offer"])).hex() if file_contents else ""
  File "/opt/homebrew/lib/python3.10/site-packages/chia/wallet/trading/offer.py", line 599, in from_bech32
    return cls.try_offer_decompression(decoded_bytes)
  File "/opt/homebrew/lib/python3.10/site-packages/chia/wallet/trading/offer.py", line 582, in try_offer_decompression
    return cls.from_compressed(offer_bytes)
  File "/opt/homebrew/lib/python3.10/site-packages/chia/wallet/trading/offer.py", line 577, in from_compressed
    return Offer.from_bytes(decompress_object_with_puzzles(compressed_bytes))
  File "/opt/homebrew/lib/python3.10/site-packages/chia/wallet/util/puzzle_compression.py", line 72, in decompress_object_with_puzzles
    raise CompressionVersionError(version)
chia.wallet.util.puzzle_compression.CompressionVersionError: 6
```

Log states that it's double spent. 

### Version

1.7.0

macOS

CLI

### Relevant log output

```shell
2023-02-21T09:12:33.735 wallet chia.wallet.wallet_node    : WARNING  SpendBundle has been rejected by the FullNode. {'error': 'DOUBLE_SPEND', 'status': 3, 'txid': '0xfb83a6a363a2f9229cb7383b4a6ade8c154dbefb4f7d5fea5435c27592e76413'}
2023-02-21T09:12:33.778 wallet chia.wallet.wallet_node    : WARNING  SpendBundle has been rejected by the FullNode. {'error': 'DOUBLE_SPEND', 'status': 3, 'txid': '0x9d15fa3a2e097f9e5774c609888a8f3350aa4b1cba91596c82f661ee010a9c04'}
```

### LINKED ISSUE #14714: Dummy txns created for Offers get stuck, causes slowness & DOUBLE_SPEND errors forever
### **Summary**

I've been hunting this one for months and I think I've finally cracked it - this will be long-winded but hopefully it will help others searching for similar issues.

### **Wut?**

I've noticed for several months now that wallets inexplicably seem to get "slower" over time. The wallets that show this behavior are very active and they may create and/or accept several offers within the same block. They may be contending with other users trying to accept the same offer at the same time, as in the case with mint events. Deviation indexes of 20,000 and above are common for these larger, active wallets.

At first I thought it might be related to the deviation index. It seemed to get slower at accepting offers as the deviation index climbed. But the strange thing is that I could delete the wallet sqlite database file and let it resync, and then the wallet would be nice and fast again, even with a very high deviation index. So this meant that the derivation index itself was not the root cause!

Something must be accruing, or "leaking," over time. Let's take a look through the wallet database! I ran across some odd looking records in the transaction_record table:

image
What is this?!? Several transaction records with weird blank values, ```wallet_id``` of 0 (doesn't exist), etc. After lots of spelunking, I found this bit of code:

https://github.com/Chia-Network/chia-blockchain/blob/f29eb44ffc14b79b9ad8ee28fb8d9de0c87514ba/chia/wallet/trade_manager.py#L769-L787

When an offer is accepted, this "dummy" txn is created. Once the offer is confirmed, it appears that these dummy txns get cleaned up. Here's the problems:

- These dummies are not cleaned up if the offer is canceled.
- They are also not cleaned up if the "Delete Unconfirmed Transactions" is used.
- Failed offers (another user snipes your accept) mean more orphaned dummy txns in this table.
- This means they build up over time

### **How does this affect performance?** 

**It appears that these dummy txns are submitted to the full node over and over, resulting in ```DOUBLE_SPEND``` errors and ```pre_validate_spendbundle``` warnings in the log.** 

image

As these orphan dummies build up over time, the wallet gets slower and slower at accepting offers and sending txns because everytime it does, it goes through that huge list of stuck orphan dummies and tries to submit them to full node again. In fact, **full node will eventually try to ban the local wallet** for this nonsense!

### **How to see if you are affected by this issue?** 

Run this query in your wallet db:

```SELECT * FROM transaction_record WHERE wallet_id = 0```

If you have no pending offers or transactions, but you still get results from this query, it means you have orphaned dummies. 

### **How to temporarily fix this issue?**

Use this statement to clean them up manually (log out of wallet in UI/CLI first):

```DELETE FROM transaction_record WHERE wallet_id = 0```

The more dummy records you cleaned up, the faster your wallet should feel after you next start it up!

I'm not sure about a longer term fix or I would have submitted a PR, but I think I'd start by just including the ```wallet_id``` when making these dummy txns instead of setting it to zero:

https://github.com/Chia-Network/chia-blockchain/blob/f29eb44ffc14b79b9ad8ee28fb8d9de0c87514ba/chia/wallet/trade_manager.py#L781

This would mean that they would naturally get cleaned up by ```delete_unconfirmed_transactions```:

https://github.com/Chia-Network/chia-blockchain/blob/f29eb44ffc14b79b9ad8ee28fb8d9de0c87514ba/chia/wallet/wallet_transaction_store.py#L356

Longer term, maybe adding some logic to ```trade_store.set_status``` so that when offers are canceled or failed, these dummy txns get cleaned up would be nice: https://github.com/Chia-Network/chia-blockchain/blob/f29eb44ffc14b79b9ad8ee28fb8d9de0c87514ba/chia/wallet/trading/trade_store.py#L177

### **Fin**
Anyway, hopefully this detail helps this bug get fixed extra-quick, cause I have a sneaky suspicion that it has been one of the main causes of wallet slowness and ```DOUBLE_SPEND``` errors in logs!

Thank you for attending my New Issue Talk.

### PR DESCRIPTION
**Pull request by trepca**
### Purpose:
to better handle offer status in case offers fails due to bad spend, conflict or failures to be added to mempool

### Current Behavior:
offer can be retried forever, we also have a bug where offers don't go into accepted state, 

### New Behavior:
offer state should work as expected
transaction that fail for more than 3 times become confirmed as failed and will not be retried anymore

### Testing Notes:
try accepting offers that were already accepted

### DIFF
--- a/chia/wallet/transaction_record.py
+++ b/chia/wallet/transaction_record.py
@@ -111,8 +111,7 @@ class TransactionRecord(Streamable):
         }
         return formatted
 
-    @property
-    def is_spendable(self) -> bool:
+    def is_valid(self) -> bool:
         past_receipts = self.sent_to
         if len(past_receipts) < 6:
             # we haven't tried enough peers yet
--- a/chia/wallet/transaction_record.py
+++ b/chia/wallet/transaction_record.py
@@ -114,7 +114,7 @@ class TransactionRecord(Streamable):
     @property
     def is_spendable(self) -> bool:
         past_receipts = self.sent_to
-        if len(past_receipts) < 3:
+        if len(past_receipts) < 6:
             # we haven't tried enough peers yet
             return True
         if any([x[0] for x in past_receipts if x[0] == MempoolInclusionStatus.SUCCESS.value]):
--- a/chia/wallet/transaction_record.py
+++ b/chia/wallet/transaction_record.py
@@ -9,6 +9,7 @@ from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.mempool_inclusion_status import MempoolInclusionStatus
 from chia.types.spend_bundle import SpendBundle
 from chia.util.bech32m import decode_puzzle_hash, encode_puzzle_hash
+from chia.util.errors import Err
 from chia.util.ints import uint8, uint32, uint64
 from chia.util.streamable import Streamable, streamable
 from chia.wallet.util.transaction_type import TransactionType
@@ -109,3 +110,17 @@ class TransactionRecord(Streamable):
             if memo is not None
         }
         return formatted
+
+    @property
+    def is_spendable(self) -> bool:
+        past_receipts = self.sent_to
+        if len(past_receipts) < 3:
+            # we haven't tried enough peers yet
+            return True
+        if any([x[0] for x in past_receipts if x[0] == MempoolInclusionStatus.SUCCESS.value]):
+            # we managed to push it to mempool at least once
+            return True
+        if any([x[1] for x in past_receipts if x[1] in (Err.INVALID_FEE_LOW_FEE, Err.INVALID_FEE_TOO_CLOSE_TO_ZERO)]):
+            # we tried to push it to mempool and got a fee error so it's a temporary error
+            return True
+        return False
--- a/chia/wallet/wallet.py
+++ b/chia/wallet/wallet.py
@@ -372,7 +372,8 @@ class Wallet:
             change = max(0, change)
         if change < 0 < fee:
             fee_coins = await self.select_coins(
-                uint64(fee),
+                # change already includes fee amount
+                uint64(abs(change)),
                 excluded_coin_amounts=exclude_coin_amounts,
                 exclude=([] if exclude_coins is None else list(exclude_coins)) + list(coins or []),
             )
@@ -380,8 +381,7 @@ class Wallet:
             spend_value = sum([coin.amount for coin in coins])
             self.log.info(f"Updated spend_value is {spend_value} and total_amount is {total_amount}")
             change = spend_value - total_amount
-
-        assert change >= 0
+        assert change >= 0, f"change is negative: {change}"
 
         if coin_announcements_to_consume is not None:
             coin_announcements_bytes: Optional[Set[bytes32]] = {a.name() for a in coin_announcements_to_consume}
--- a/chia/wallet/wallet.py
+++ b/chia/wallet/wallet.py
@@ -370,7 +370,8 @@ class Wallet:
         change = spend_value - total_amount
         if negative_change_allowed:
             change = max(0, change)
-        if change < 0 < fee:
+        # only kicks in if fee is missing
+        if change < 0 and fee + amount == total_amount:
             fee_coins = await self.select_coins(
                 # change already includes fee amount
                 uint64(abs(change)),
--- a/chia/wallet/wallet.py
+++ b/chia/wallet/wallet.py
@@ -50,7 +50,6 @@ from chia.wallet.wallet_info import WalletInfo
 if TYPE_CHECKING:
     from chia.server.ws_connection import WSChiaConnection
 
-
 # https://github.com/Chia-Network/chips/blob/80e4611fe52b174bf1a0382b9dff73805b18b8c6/CHIPs/chip-0002.md#signmessage
 CHIP_0002_SIGN_MESSAGE_PREFIX = "Chia Signed Message"
 
@@ -344,7 +343,7 @@ class Wallet:
         if not ignore_max_send_amount:
             max_send = await self.get_max_send_amount()
             if total_amount > max_send:
-                raise ValueError(f"Can't send more than {max_send} mojos in a single transaction")
+                raise ValueError(f"Can't send more than {max_send} mojos in a single transaction, got {total_amount}")
             self.log.debug("Got back max send amount: %s", max_send)
         if coins is None:
             if total_amount > total_balance:
@@ -363,13 +362,24 @@ class Wallet:
             )
         elif exclude_coins is not None:
             raise ValueError("Can't exclude coins when also specifically including coins")
+
         assert len(coins) > 0
         self.log.info(f"coins is not None {coins}")
         spend_value = sum([coin.amount for coin in coins])
-
+        self.log.info(f"spend_value is {spend_value} and total_amount is {total_amount}")
         change = spend_value - total_amount
         if negative_change_allowed:
             change = max(0, change)
+        if change < 0 < fee:
+            fee_coins = await self.select_coins(
+                uint64(fee),
+                excluded_coin_amounts=exclude_coin_amounts,
+                exclude=([] if exclude_coins is None else list(exclude_coins)) + list(coins or []),
+            )
+            coins = coins.union(fee_coins)
+            spend_value = sum([coin.amount for coin in coins])
+            self.log.info(f"Updated spend_value is {spend_value} and total_amount is {total_amount}")
+            change = spend_value - total_amount
 
         assert change >= 0
 
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1623,10 +1623,10 @@ class WalletStateManager:
                             # offer to failed
                             trade_coins_removed.add(removed_coin.name())
                     if trade and trade_coins_removed:
-                        if not tx.is_spendable:
-                            # we've already sent this transaction to full node and it was accepted
-                            # or we've sent it with low fee and it was rejected
-                            # we don't need to subscribe to these coins anymore
+                        if not tx.is_valid():
+                            # we've tried to send this transaction to a full node multiple times
+                            # but failed, it's safe to assume that it's not going to be accepted
+                            # we can mark this offer as failed
                             self.log.info("This offer can't be posted, removing it from pending offers")
                             assert trade is not None
                             await self.trade_manager.fail_pending_offer(trade.trade_id)
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1030,7 +1030,7 @@ class WalletStateManager:
 
         trade_removals = await self.trade_manager.get_coins_of_interest()
         all_unconfirmed: List[TransactionRecord] = await self.tx_store.get_all_unconfirmed()
-        trade_coin_removed: List[CoinState] = []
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: chia/wallet/wallet.py [hunk 1]
```diff
@@ -363,13 +362,24 @@ async def _generate_unsigned_transaction(
...
+        if change < 0 < fee:
+            fee_coins = await self.select_coins(
+                uint64(fee),
+                excluded_coin_amounts=exclude_coin_amounts,
+                exclude=([] if exclude_coins is None else list(exclude_coins)) + list(coins or []),
+            )
+            coins = coins.union(fee_coins)
+            spend_value = sum([coin.amount for coin in coins])
+            self.log.info(f"Updated spend_value is {spend_value} and total_amount is {total_amount}")
+            change = spend_value - total_amount
```
Quexington, line 382: Can you comment what is going on here, I'm not sure why we would do this (I don't think we have many code paths that hit negative change)

chia/wallet/wallet_transaction_store.py
```diff
@@ -152,6 +155,11 @@ async def increment_sent(
         sent_to.append(append_data)
 
         tx: TransactionRecord = dataclasses.replace(current, sent=sent_count, sent_to=sent_to)
+        if not tx.is_spendable:
+            # if the tx is not spendable, it is confirmed that we can't spend it, but
+            # we need to adjust the GUI to show that it is failed
+            log.info(f"Marking tx={tx.name} as confirmed but failed, since it is not spendable due to errors")
```
Quexington, line 161: Does TransactionRecord have a "failed" state? I thought we only had confirmed or not.

chia/wallet/wallet.py
trepca: right, wasn't sure where to put this, but when we generate offer coins, we don't include `fee` into it, then we call generated_signed tx and we pass `coins` to it so additional coins are not selected if we don't have enough coin. When fee is low, this is not a problem because we have enough change to use it for fee, but when fee is larger or user has a lot of small amount coins this errors out.

chia/wallet/wallet_transaction_store.py
trepca: it doesn't but I defined a failed state as having `confirmed=1 and confirmed_at_height=0` which we can use to display as failed in UX/GUI

chia/wallet/wallet.py
```diff
@@ -363,13 +362,24 @@ async def _generate_unsigned_transaction(
...
         self.log.info(f"coins is not None {coins}")
         spend_value = sum([coin.amount for coin in coins])
-
+        self.log.info(f"spend_value is {spend_value} and total_amount is {total_amount}")
         change = spend_value - total_amount
         if negative_change_allowed:
             change = max(0, change)
+        if change < 0 < fee:
+            fee_coins = await self.select_coins(
+                uint64(fee),
```
Quexington, line 375: Should this be fee + change (which is sematically the same as fee - abs(change)).  You might be able to partially complete the fee with leftover change, where this requires you to have amount + change + fee in your wallet rather than amount + fee.

Quexington (dismissed):

Quexington (dismissed):

chia/wallet/transaction_record.py
```diff
@@ -109,3 +110,17 @@ def to_json_dict_convenience(self, config: Dict) -> Dict:
             if memo is not None
         }
         return formatted
+
+    @property
+    def is_spendable(self) -> bool:
```
arvidn, line 115: it makes me nervous to have properties like this. The risk is that it looks cheap (just querying a state) whereas in fact it's not. I suppose the only way for this function to be expensive is if `self.sent_to` is a very long list.

I'm not familiar with this code, but this name seems odd to me. it looks like it's answering the question "do we still think there is a chance for this transaction to land in a block?". the name `is_spendable()` implies a few other things in my mind. e.g. (1) is it valid, (2) is it unspent.

chia/wallet/wallet.py
```diff
@@ -363,15 +362,26 @@ async def _generate_unsigned_transaction(
...
         self.log.info(f"coins is not None {coins}")
         spend_value = sum([coin.amount for coin in coins])
-
+        self.log.info(f"spend_value is {spend_value} and total_amount is {total_amount}")
         change = spend_value - total_amount
         if negative_change_allowed:
             change = max(0, change)
-
-        assert change >= 0
+        if change < 0 < fee:
```
arvidn, line 373: I think I'm missing something. perhaps I don't understand this syntax. But `change` has to be < 0 for this condition to be true, right? Is that used as a place-holder to ask the wallet to pick coins to spend? And why do we not allow this path for zero-fee transactions?

Did we have a problem here before, where we would sometimes fail to spend because of `MINTING_COIN`?

chia/wallet/wallet_state_manager.py
```diff
@@ -1597,28 +1598,43 @@ async def remove_from_queue(
...
+                        ):
+                            # offer was tied to these coins, lets subscribe to them to get a confirmation to
+                            # cancel it if it's confirmed
+                            # we send transactions to multiple peers, and in cases when mempool gets
+                            # fragmented, it's safest to wait for confirmation from blockchain before setting
+                            # offer to failed
+                            trade_coins_removed.add(removed_coin.name())
+                    if trade and trade_coins_removed:
+                        if not tx.is_spendable:
+                            # we've already sent this transaction to full node and it was accepted
```
arvidn, line 1627: this comment makes me a bit confused. I think you mean to say "and it failed"

chia/wallet/wallet_transaction_store.py
arvidn: this feels a bit scary. it should be clearly documented, because it's not obvious from the code that we're treating this special case as such

chia/wallet/wallet.py
trepca: this should only kick in if we're missing the fee and coins were not selected in this method but passed through. This happens when we generate offers as trade_manager only selects coins that are needed to complete the offer, not fee. But you have a point that this condition could be clearer.

arvidn (approved):

aaronscientiae:
this is still happening for me in 1.8.0rc1. I have 5 'pending accept' nfts that are already in a different wallet, but not cancellable even with a fee.

PR #14813, opened 2023-03-14
wallet: Don't just subscribe to all coin ids from the DB

xdustinface:
### Purpose:

Reduce number of subscriptions used from wallets on the full nodes. 

### Current Behavior:

Besides subscribing to the coin ids from `WalletInterestedStore` and `TradeManager` the wallet also subscribes to all coin ids included in the `WalletCoinStore` which from my understanding is not required because they are all already covered by the wallet's puzzle hash subscriptions. 

### New Behavior:

The wallet only subscribes to coin ids from `WalletInterestedStore` or `TradeManager`.

### Testing Notes:

Should be covered by all tests which test wallet sync/generate/send coins because the coins or state changes would not even make it into the DB if the puzzle hash subscription would not cover them. Also, this subscribing only happens during long sync at which point the coins are already in the DB because they were covered by the puzzle hash subscriptions.

### DIFF
--- a/chia/wallet/wallet_coin_store.py
+++ b/chia/wallet/wallet_coin_store.py
@@ -181,19 +181,6 @@ class WalletCoinStore:
             rows = await conn.execute_fetchall("SELECT * FROM coin_record WHERE spent_height=0")
         return set(self.coin_record_from_row(row) for row in rows)
 
-    async def get_coin_names_to_check(self, check_height) -> Set[bytes32]:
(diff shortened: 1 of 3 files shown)


Quexington (approved):
I think this is fine, sometimes we subscribe to coin IDs but all of that should happen during sync anyways and be added to the interested store

arvidn (approved):

PR #14816, opened 2023-03-14
fix: set default value with optional argument

birdca:
### Purpose: 
make upgrading to >=1.6.1 more friendly

### Current Behavior: 
it shows `missing 1 required positional argument: 'root_hash'`

### New Behavior: 
It won't cause errors in DataLayer when upgrading to version >= 1.6.1.

### DIFF
--- a/chia/rpc/data_layer_rpc_client.py
+++ b/chia/rpc/data_layer_rpc_client.py
@@ -26,7 +26,7 @@ class DataLayerRpcClient(RpcClient):
         response = await self.fetch("batch_update", {"id": store_id.hex(), "changelist": changelist, "fee": fee})
         return response
 
-    async def get_keys_values(self, store_id: bytes32, root_hash: Optional[bytes32]) -> Dict[str, Any]:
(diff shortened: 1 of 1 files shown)


emlowe:
I'm not clear on what the problem is here. The RPC endpoint takes care of this, so this could only be a problem if you are using the code directly and calling this within your own code.

birdca:
@emlowe Sorry for being unclear. Yes, we're using this within code. As I look in the code, if `root_hash` isn't being set, it'll set it to None. Does this RPC design for assigning a particular `root_hash`?

emlowe:
I'm not sure I understand the question. But you can call the RPC as follows
`chia rpc data_layer get_keys_values '{"id":"xxx"}'` and root_hash will be set to None internally when this is called
`chia rpc data_layer get_keys_values '{"id":"xxx", "root_hash":"yyy"}'` and it will use the provided root_hash

You may refer to the code in `data_layer_rpc_api.py`

PR #14828, opened 2023-03-15
wallet: Return a dict in `WalletCoinStore.get_coin_records`

xdustinface:
### Purpose:

Just to simplify/cleanup. I think its more intuitive to just return the dict instead of returning a list where we expect a `None` entry at the same index where we provided a coin id which was not found in the DB?

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -985,13 +985,10 @@ class WalletRpcApi:
             exclude_coin_amounts = [uint64(a) for a in exclude_coin_amounts]
         exclude_coin_ids: Optional[List] = request.get("exclude_coin_ids")
         if exclude_coin_ids is not None:
-            exclude_coins: Set[Coin] = {
-                wr.coin
-                for wr in await self.service.wallet_state_manager.coin_store.get_coin_records(
-                    [bytes32.from_hexstr(hex_id) for hex_id in exclude_coin_ids]
-                )
-                if wr is not None
-            }
+            coin_records = await self.service.wallet_state_manager.coin_store.get_coin_records(
+                [bytes32.from_hexstr(hex_id) for hex_id in exclude_coin_ids]
+            )
+            exclude_coins = {wr.coin for wr in coin_records.values()}
         else:
             exclude_coins = set()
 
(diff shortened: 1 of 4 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -985,13 +985,10 @@ async def send_transaction(self, request) -> EndpointResult:
...
-                wr.coin
-                for wr in await self.service.wallet_state_manager.coin_store.get_coin_records(
-                    [bytes32.from_hexstr(hex_id) for hex_id in exclude_coin_ids]
-                )
-                if wr is not None
-            }
+            coin_records = await self.service.wallet_state_manager.coin_store.get_coin_records(
+                [bytes32.from_hexstr(hex_id) for hex_id in exclude_coin_ids]
+            )
+            exclude_coins = {wr.coin for wr in coin_records.values()}
```
arvidn, line 991: it looks very inefficient to create a bunch of `CoinRecord`s, just to pick out the `Coin`s later. We should probably just get coins from the DB

chia/wallet/wallet_state_manager.py
```diff
@@ -1034,12 +1034,10 @@ async def new_coin_state(
         used_up_to = -1
         ph_to_index_cache: LRUCache = LRUCache(100)
 
-        local_records: List[Optional[WalletCoinRecord]] = await self.coin_store.get_coin_records(
-            [st.coin.name() for st in coin_states]
-        )
+        local_records = await self.coin_store.get_coin_records([st.coin.name() for st in coin_states])
 
-        assert len(local_records) == len(coin_states)
```
arvidn, line 1041: this assert seems relevant still, no?

chia/wallet/wallet_state_manager.py
```diff
@@ -1034,12 +1034,10 @@ async def new_coin_state(
...
-        for coin_state, local_record in zip(coin_states, local_records):
+        for coin_state in coin_states:
+            local_record = local_records.get(coin_state.coin.name())
```
arvidn, line 1040: I notice we call `Coin.name()` 4 times per coin state now. we used to call it 3 times. it's not trivial to compute.

arvidn (approved):

chia/wallet/wallet_state_manager.py
xdustinface: No because we not longer include the `None` entries. If a local record isn't found the assertion isn't true anymore.

chia/wallet/wallet_coin_store.py
```diff
@@ -137,7 +137,7 @@ async def get_coin_records(
         include_spent_coins: bool = True,
         start_height: uint32 = uint32(0),
         end_height: uint32 = uint32((2**32) - 1),
-    ) -> List[Optional[WalletCoinRecord]]:
+    ) -> Dict[bytes32, WalletCoinRecord]:
         """Returns CoinRecord with specified coin id."""
         async with self.db_wrapper.reader_no_transaction() as conn:
             rows = list(
```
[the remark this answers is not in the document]
xdustinface: Looks like yes. I will PR it separately.

PR #14833, opened 2023-03-16
wallet: Handle trade coins in the `try` block of `new_coin_state`

xdustinface:
### Purpose:

Maybe i miss the point why this was put at the outside the `try` block but the idea here is to avoid aborting the whole sync if the processing of a trade removal raises which might happen because it for example involves peer communication and DB access. Also this lets them be added to the retry store if the relevant exceptions are raised.

One reason i can think of why it was put at the end (doesn't explain why its outside the `try` block though) would be that it depends on the state being processed already but from looking inside `coins_of_interest_farmed` it doesn't seem like? 

### Current Behavior:

All trade related coins are processed together at the end of `new_coin_state` outside the `try`/`except` block.

### New Behavior:

All trade related coins are processed separately in the `try`/`except` block.

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1030,7 +1030,6 @@ class WalletStateManager:
 
         trade_removals = await self.trade_manager.get_coins_of_interest()
         all_unconfirmed: List[TransactionRecord] = await self.tx_store.get_all_unconfirmed()
-        trade_coin_removed: List[CoinState] = []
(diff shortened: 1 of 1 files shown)


Quexington (approved):
Yeah I can't think of the reason it would have been outside either.  Perhaps the wallet used to have some behavior that resulted in this decision but I can't think of any _current_ logic that would necessitate this.

arvidn (approved):

PR #14839, opened 2023-03-16
remove pytest as a CLVM compilation trigger

altendky:
### Purpose:

Presently, it takes many seconds (20-30 per a couple devs) to do the CLVM recompilation for every single pytest run.  There may be a bug somewhere, there are certainly other alternative solutions to this.  I would like it to be improved.  As is, there is still the `CHIA_DEV_COMPILE_CLVM_ON_IMPORT` environment variable that can be set (presumably once per user per computer) to still make it always happen as desired.

### Current Behavior:

Take forever to start processes for testing.

### New Behavior:

Doesn't take forever on CLVM stuff.

### DIFF
--- a/chia/wallet/puzzles/load_clvm.py
+++ b/chia/wallet/puzzles/load_clvm.py
@@ -17,7 +17,7 @@ from chia.util.lock import Lockfile
 
 compile_clvm_py = None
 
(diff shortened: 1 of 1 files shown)


Quexington:
I would like to suggest https://github.com/Chia-Network/chia-blockchain/pull/14840 as an alternative to this

PR #14840, opened 2023-03-16
Only compile CLVM if source newer than hex

Quexington:
Running tests is really slow right now because _every_ clvm file is recompiled on _every_ test run.  This reverts to the old behavior of only recompiling if the source is newer than the hex.  The behavior remains guarded behind an environment variable or running during pytest.

### DIFF
--- a/chia/wallet/puzzles/load_clvm.py
+++ b/chia/wallet/puzzles/load_clvm.py
@@ -103,7 +103,9 @@ def load_serialized_clvm(
                     search_paths = [full_path.parent]
                     if include_standard_libraries:
                         # we can't get the dir, but we can get a file then get its parent.
-                        chia_puzzles_path = pathlib.Path(pkg_resources.resource_filename(__name__, "__init__.py")).parent
+                        chia_puzzles_path = pathlib.Path(
+                            pkg_resources.resource_filename(__name__, "__init__.py")
+                        ).parent
                         search_paths.append(chia_puzzles_path)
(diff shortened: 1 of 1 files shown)


emlowe (approved):
Looks good to me

arvidn:
we check in the compiled code too, right? so this test (as far as I understand it) is mostly to ensure the checked-in compiled programs match the corresponding source code.

Is looking at timestamps a good proxy for these to possibly be out of sync?
I suppose it would detect a commit *failing* to compile, but not to detect whether it's *correctly* compiled.

arvidn:
I think @AmineKhaldi and @prozacchiwawa should see this before we land

prozacchiwawa:
i don't think this breaks anything.  will approve if wanted.

altendky:
It doesn't particularly matter, but I'm curious when this behavior got lost.  I don't think I removed it in https://github.com/Chia-Network/chia-blockchain/commit/a20ee08fce6d608bf0f886b8edba35031dabb1c1.  I looked a bit farther, just real quickly, and didn't see anything else.

Quexington:
> It doesn't particularly matter, but I'm curious when this behavior got lost. I don't think I removed it in [a20ee08](https://github.com/Chia-Network/chia-blockchain/commit/a20ee08fce6d608bf0f886b8edba35031dabb1c1). I looked a bit farther, just real quickly, and didn't see anything else.

The behavior used to be part of the python clvm_tools so maybe switching to clvm_tools_rs at some point changed this behavior.

PR #14842, opened 2023-03-17
wallet: Raise in `WalletNode.get_full_node_peer` and move some of its calls

xdustinface:
### Purpose:

Code cleanup which also moves peer fetching closer to the actual call/usage in some cases.

### Current Behavior:

The method returns `Optional[WSChiaConnection]`.

### New Behavior:

The method returns `WSChiaConnection` and in some cases the call of the method was moved closer to the actual call/usage to skip some async calls inbetween.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -266,10 +266,8 @@ class WalletRpcApi:
         )
 
     async def get_latest_singleton_coin_spend(
-        self, peer: Optional[WSChiaConnection], coin_id: bytes32, latest: bool = True
+        self, peer: WSChiaConnection, coin_id: bytes32, latest: bool = True
     ) -> Tuple[CoinSpend, CoinState]:
-        if peer is None:
-            raise ValueError("No peers to get info from")
         coin_state_list: List[CoinState] = await self.service.wallet_state_manager.wallet_node.get_coin_state(
             [coin_id], peer=peer
         )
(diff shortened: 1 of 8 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -3140,11 +3131,6 @@ async def dl_track_new(self, request) -> Dict:
         """Initialize the DataLayer Wallet (only one can exist)"""
         if self.service.wallet_state_manager is None:
             raise ValueError("The wallet service is not currently initialized")
-
-        peer: Optional[WSChiaConnection] = self.service.get_full_node_peer()
-        if peer is None:
-            raise ValueError("No peer connected")
```
arvidn, line 3146: this code snippet appear to have changed behavior. It used to end here if we weren't connected, but in your patch we just keep going. are you sure that's OK?

With your change we'll create the wallet before throwing, whereas before we would skip creating the wallet in the first place when we're not connected

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -470,12 +470,9 @@ async def load_parent(self, did_info: DIDInfo):
             did_wallet_puzzles.metadata_to_program(json.loads(self.did_info.metadata)),
         )
         wallet_node = self.wallet_state_manager.wallet_node
-        peer: WSChiaConnection = wallet_node.get_full_node_peer()
-        if peer is None:
-            raise ValueError("Could not find any peers to request puzzle and solution from")
-
         parent_coin: Coin = did_info.origin_coin
         while True:
+            peer = wallet_node.get_full_node_peer()
```
arvidn, line 475: is there a good reason to move this into the loop? it means we'll drop our reference to the peer, and then re-acquire it the next iteration. And we'll have more opportunities for this to throw

arvidn (commented):
I'm in favor of this general idea, but I think there are two places where you change the behavior without a clear reason

chia/wallet/did_wallet/did_wallet.py
xdustinface: The idea was to make sure we have a connected peer before each all since there are async calls in the loop.

chia/rpc/wallet_rpc_api.py
xdustinface: Hm true that it wouldn't create the wallet in case there is no peer but since there anyway can only be one wallet and a user/script would re-try this anyway if it fails it probably doesn't matter at what point the wallet gets created? Also its only really relevant for the first usage. I can change it if you think its bad but i feel like its ok.

arvidn (approved):

PR #14847, opened 2023-03-18
slightly faster nft complex offer tests

trepca:
nft_wallet tests are timing out after adding more offer tests


arvidn (dismissed):

chia/wallet/nft_wallet/metadata_outer_puzzle.py
```diff
@@ -45,8 +43,8 @@ def match(self, puzzle: UncurriedPuzzle) -> Optional[PuzzleInfo]:
             _, metadata, updater_hash, inner_puzzle = curried_args
             constructor_dict = {
                 "type": "metadata",
-                "metadata": disassemble(metadata),
-                "updater_hash": "0x" + updater_hash.as_python().hex(),
+                "metadata": metadata,
```
arvidn, line 46: It's great to not have to disassemble. But I was under the impression this field *had* to be disassembled.
tracing this, it looks like we end up passing this metadata into `puzzle_for_metadata_layer()` which has taken a `Program` all along. I don't understand why it was correct to `disassemble()` this to begin with. I would expect there to be a corresponding `assemble()` call removed somewhere.

arvidn (commented):
looks good. are you sure it's ok to not disassemble the program in the metadata?
and is there no test coverage of "forwards_compat" now?

chia/wallet/trading/offer.py
```diff
@@ -78,6 +78,8 @@ class Offer:
     # this is a cache of the coin additions made by the SpendBundle (_bundle)
     # ordered by the coin being spent
     _additions: Dict[Coin, List[Coin]] = field(init=False)
+    _offered_coins: Dict[Optional[bytes32], List[Coin]] = field(init=False)
```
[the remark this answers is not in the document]
trepca: doesn't post_init populate it?

tests/wallet/nft_wallet/test_nft_offers.py
```diff
@@ -488,7 +472,7 @@ async def test_nft_offer_with_metadata_update(
     updated_nft = coins_maker[0]
     updated_nft_info = match_puzzle(uncurry_puzzle(updated_nft.full_puzzle))
 
-    assert url_to_add in updated_nft_info.also().info["metadata"]  # type: ignore
+    assert url_to_add in disassemble(updated_nft_info.also().info["metadata"])  # type: ignore
```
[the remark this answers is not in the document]
trepca: PuzzleDriver actually decodes everything back into Program so I just had to add this: https://github.com/Chia-Network/chia-blockchain/pull/14847/files#diff-c1840ab535355caa37a34dfbb07017e20f08b713378102b8fd4c1a96c574466dR102 to support metadata as Program. Before it was disassembled and assembled back. I assume because of how we use it in other cases.

tests/wallet/nft_wallet/test_nft_1_offers.py
```diff
@@ -1276,7 +1267,6 @@ async def get_trade_and_status(trade_manager: Any, trade: Any) -> TradeStatus:
         (True, (200, 500, 500)),
         (False, (200, 500, 500)),
         (False, (0, 0, 0)),  # test that we can have 0 royalty
-        (False, (65000, 65534, 65535)),  # test that we can reach max royalty
```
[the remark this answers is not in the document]
trepca: this case is very slow yes, about 50 seconds due to amount of coins it needs to fetch/mint to pay for royalty and in general tests a fairly complex case of nft trade

chia/wallet/trading/offer.py
```diff
@@ -152,6 +154,8 @@ def __post_init__(self) -> None:
             if max_cost < 0:
                 raise ValidationError(Err.BLOCK_COST_EXCEEDS_MAX, "compute_additions for CoinSpend")
         object.__setattr__(self, "_additions", adds)
+        object.__setattr__(self, "final_spend_bundle", self.to_spend_bundle())
+        object.__setattr__(self, "_offered_coins", self._get_offered_coins())
```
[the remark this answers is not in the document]
trepca: I was having some problems with dataclass field overriding, but I can't remember anymore. I can try doing equals.

chia/wallet/nft_wallet/metadata_outer_puzzle.py
Quexington: I agree that this _should_ be disassembled because other calls generally to this API will not get the right output.  PuzzleInfo is just bad at type checking itself and if it was better, it would likely throw an error due to this change.  However, in context, I would sort of like to deprecate this code (or make `PuzzleInfo` better so that this would actually not be a problem) so if all of the calls to this method still work, I'm actually fine with changing this.

Quexington: I see you changed `decode_info_value` to be able to handle this actually so that's perfect :+1:

arvidn: We need to move away from using `disassemble()` and `assemble()` in general

Quexington: I agree in the context of such a haphazard manner, but some RPCs need to accept/return CLVM them and there's really no way around those methods in that case.

Quexington (dismissed):

arvidn (dismissed):

wallentx (changes requested):
@trepca this is consistently failing benchmark tests. This should be good to go once those tests are green

wallentx (approved):

arvidn (approved):

PR #14849, opened 2023-03-19
exclude CompressorArg.generator from repr() due to size

altendky:
### Purpose:

Make the log entries smaller.

### Current Behavior:

103KB log entries

### New Behavior:

Sane log entries.

### DIFF
--- a/chia/types/generator_types.py
+++ b/chia/types/generator_types.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


arvidn (approved):
looks reasonable. But it raises the question; where do we log `CompressorArg` objects, and why?

altendky:
https://github.com/Chia-Network/chia-blockchain/blob/ea2da6ce6e7c692ca22e9a01504ec6c2da81d9a8/chia/full_node/full_node_api.py#L762-L769

PR #14852, opened 2023-03-20
wallet: Simplify `WalletStateManager.coin_added`

xdustinface:
### Purpose:

Just some cleanup.

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1472,26 +1472,23 @@ class WalletStateManager:
             wallet_id,
             wallet_type,
         )
-        farmer_reward = False
-        pool_reward = False
-        if self.is_farmer_reward(height, coin):
-            farmer_reward = True
-        elif self.is_pool_reward(height, coin):
-            pool_reward = True
 
-        farm_reward = False
+        if self.is_pool_reward(height, coin):
+            tx_type = TransactionType.COINBASE_REWARD
+        elif self.is_farmer_reward(height, coin):
+            tx_type = TransactionType.FEE_REWARD
+        else:
+            tx_type = TransactionType.INCOMING_TX
+
+        coinbase = tx_type != TransactionType.INCOMING_TX
+
         parent_coin_record: Optional[WalletCoinRecord] = await self.coin_store.get_coin_record(coin.parent_coin_info)
         if parent_coin_record is not None and wallet_type.value == parent_coin_record.wallet_type:
             change = True
         else:
             change = False
 
-        if farmer_reward or pool_reward:
-            farm_reward = True
-            if pool_reward:
-                tx_type: int = TransactionType.COINBASE_REWARD.value
-            else:
-                tx_type = TransactionType.FEE_REWARD.value
+        if coinbase:
             timestamp = await self.wallet_node.get_timestamp_for_height(height)
 
             tx_record = TransactionRecord(
@@ -1540,7 +1537,7 @@ class WalletStateManager:
                     wallet_id=wallet_id,
                     sent_to=[],
                     trade_id=None,
-                    type=uint32(TransactionType.INCOMING_TX.value),
+                    type=uint32(tx_type),
                     name=coin_name,
                     memos=[],
                 )
@@ -1548,7 +1545,7 @@ class WalletStateManager:
                     await self.tx_store.add_transaction_record(tx_record)
 
         coin_record_1: WalletCoinRecord = WalletCoinRecord(
-            coin, height, uint32(0), False, farm_reward, wallet_type, wallet_id
+            coin, height, uint32(0), False, coinbase, wallet_type, wallet_id
         )
         await self.coin_store.add_coin_record(coin_record_1, coin_name)
 
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1511,17 +1511,13 @@ class WalletStateManager:
             )
             await self.tx_store.add_transaction_record(tx_record)
         else:
-            records: List[TransactionRecord] = []
+            coin_confirmed_transaction = False
             for record in all_unconfirmed_transaction_records:
-                for add_coin in record.additions:
-                    if add_coin == coin:
-                        records.append(record)
+                if coin in record.additions and not record.confirmed:
+                    await self.tx_store.set_confirmed(record.name, height)
+                    coin_confirmed_transaction = True
 
-            if len(records) > 0:
-                for record in records:
-                    if record.confirmed is False:
-                        await self.tx_store.set_confirmed(record.name, height)
-            elif not change:
+            if not coin_confirmed_transaction and not change:
                 timestamp = await self.wallet_node.get_timestamp_for_height(height)
                 tx_record = TransactionRecord(
                     confirmed_at_height=uint32(height),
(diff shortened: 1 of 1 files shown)


xdustinface:
> @AmineKhaldi we need to carefully compare the TX Record at https://github.com/Chia-Network/chia-blockchain/pull/14852/files#diff-a367f650aed3b60eba2c253ce15852630c1baf5e9891223d1557ac9cb3333d20R1220
> 
> with the one that is removed.

@aqk Not sure if its relevant for this PR? What you linked to there is in `new_coin_state` - this PR is about `coin_added`.

chia/wallet/wallet_state_manager.py
```diff
@@ -1474,32 +1474,30 @@ async def coin_added(
...
-            else:
-                tx_type = TransactionType.FEE_REWARD.value
-            timestamp = await self.wallet_node.get_timestamp_for_height(height)
+        coinbase = tx_type != TransactionType.INCOMING_TX
+        coin_confirmed_transaction = False
+        if not coinbase:
+            for record in all_unconfirmed_transaction_records:
+                if coin in record.additions and not record.confirmed:
+                    await self.tx_store.set_confirmed(record.name, height)
+                    coin_confirmed_transaction = True
```
arvidn, line 1491: at this point, we won't ever find any other record with this same coin, will we?
it seems to me that we could `break` the for loop here

chia/wallet/wallet_state_manager.py
```diff
@@ -1474,32 +1474,30 @@ async def coin_added(
...
-            pool_reward = True
-
-        farm_reward = False
-        parent_coin_record: Optional[WalletCoinRecord] = await self.coin_store.get_coin_record(coin.parent_coin_info)
-        if parent_coin_record is not None and wallet_type.value == parent_coin_record.wallet_type:
-            change = True
+
+        if self.is_pool_reward(height, coin):
+            tx_type = TransactionType.COINBASE_REWARD
+        elif self.is_farmer_reward(height, coin):
```
arvidn, line 1480: I suppose a single coin can't be both `pool_reward` and `farmer_reward`, right? if so, the order of checking here would matter

chia/wallet/wallet_state_manager.py
```diff
@@ -1474,32 +1474,30 @@ async def coin_added(
...
-        elif self.is_pool_reward(height, coin):
-            pool_reward = True
-
-        farm_reward = False
-        parent_coin_record: Optional[WalletCoinRecord] = await self.coin_store.get_coin_record(coin.parent_coin_info)
-        if parent_coin_record is not None and wallet_type.value == parent_coin_record.wallet_type:
-            change = True
+
+        if self.is_pool_reward(height, coin):
+            tx_type = TransactionType.COINBASE_REWARD
```
arvidn, line 1479: is `tx_type` an `int`? it was prior to this patch. I think keeping the type annotation would be a good idea.
Also, previously we ask the enum constants for their `.value`, you dropped that, which may have different behavior.
I imagine restoring the type annotation might highlight this.

xdustinface: Its an `IntEnum` yes.

chia/wallet/wallet_state_manager.py
xdustinface: Yeah right we probably won't unless we do something strange with the unconfirmed transactions i don't know about. Added https://github.com/Chia-Network/chia-blockchain/pull/14852/commits/8f1d4f066c33ec7231117adbb79382bf489a724a.

chia/wallet/wallet_state_manager.py
arvidn: prior to this patch, if `is_farmer_reward()` is true we don't ask about `is_pool_reward()`, so `tx_type` becomes `TransactionType.FEE_REWARD.value`.

In your patch, if `is_pool_reward()` is true, you don't ask `is_farmer_reward()`, so `tx_type` becomes `TransactionType.COINBASE_REWARD`.

If a coin could be both, this would be a change in behavior

chia/wallet/wallet_state_manager.py
```diff
@@ -1540,15 +1537,15 @@ async def coin_added(
                     wallet_id=wallet_id,
                     sent_to=[],
                     trade_id=None,
-                    type=uint32(TransactionType.INCOMING_TX.value),
+                    type=uint32(tx_type),
```
arvidn: It still looks like `tx_type` changed *type* in this patch. it used to be `int` and now it's `TransactionType`. Casting this enum to `uint32`, is that really the same thing as asking for `.value`? I would expect that the enum itself might give you the *rank*, not the value assigned to the enum.

chia/wallet/wallet_state_manager.py
```diff
@@ -1472,26 +1472,23 @@ async def coin_added(
...
             wallet_type,
         )
-        farmer_reward = False
-        pool_reward = False
-        if self.is_farmer_reward(height, coin):
-            farmer_reward = True
-        elif self.is_pool_reward(height, coin):
-            pool_reward = True
-
-        farm_reward = False
```
arvidn, line 1482: I appreciate removing the two variables `farmer_reward` and `farm_reward`

chia/wallet/wallet_state_manager.py
```diff
@@ -1472,26 +1472,23 @@ async def coin_added(
...
+
+        if self.is_pool_reward(height, coin):
+            tx_type = TransactionType.COINBASE_REWARD
+        elif self.is_farmer_reward(height, coin):
+            tx_type = TransactionType.FEE_REWARD
+        else:
+            tx_type = TransactionType.INCOMING_TX
+
+        coinbase = tx_type != TransactionType.INCOMING_TX
```
arvidn, line 1483: it seems a bit risky to rely on no other enum values being added to this enum (and becoming possibilities here).
It's not a big deal, but I think it would be slightly easier to read and slightly more robust to say:
```
coinbase = tx_type in [TransactionType.FEE_REWARD, TransactionType.COINBASE_REWARD]
```

chia/wallet/wallet_state_manager.py
```diff
@@ -1515,30 +1514,8 @@ async def coin_added(
...
-                    wallet_id=wallet_id,
-                    sent_to=[],
-                    trade_id=None,
-                    type=uint32(tx_type),
-                    name=coin_name,
-                    memos=[],
-                )
-                if coin.amount > 0:
-                    await self.tx_store.add_transaction_record(tx_record)
+            if tx_record.amount > 0:
```
arvidn, line 1514: if `coinbase`, we call `add_transaction_record` unconditionally. I suppose it's safe to assume that coinbase coins always have amounts >0

arvidn (commented):
looks good! just a few comments

chia/wallet/wallet_state_manager.py
xdustinface: I mean, how can a coin be both?

chia/wallet/wallet_state_manager.py
xdustinface: Why don't you just try it in the console if you are concerned about it? You will see its working. I guess thats the magic of `IntEnum` what `TransactionType` is like i said here https://github.com/Chia-Network/chia-blockchain/pull/14852#discussion_r1143699640 :) 

Also, while `mypy` accepts both, PyCharm even complains about using `.value` here.

chia/wallet/wallet_state_manager.py
xdustinface: Good point, added 3822fc51b97594afc4dc7dfd1906761b6a5cced4.

chia/wallet/wallet_state_manager.py
arvidn: I'm not arguing it can. I'm arguing that this transformation is not mechanical, but requires this prerequisite knowledge

arvidn (approved):

PR #14853, opened 2023-03-20
Give a type to  tuple representing peak (height and block hash) of peer

aqk:
### Purpose:

Code cleanup
Give a type to  tuple representing peak (height and block hash) of peer

No Behavior Changes


chia/wallet/wallet_node.py
```diff
@@ -86,6 +86,12 @@ def get_wallet_db_path(root_path: Path, config: Dict[str, Any], key_fingerprint:
     return path
 
 
+@dataclasses.dataclass
```
altendky, line 89: I'd default to this at least when there are immutable attribute types and we don't need to mutate.

```suggestion
@dataclasses.dataclass(frozen=True)
```

altendky (changes requested):
Pending CI and the light tweak, I like it.  Thanks.

altendky (dismissed):

altendky (approved):

emlowe (approved):

PR #14854, opened 2023-03-20
Ak.exception during get timestamp for height.with exceptions

aqk:
DO NOT MERGE.

This PR is for discussing changes to `WalletNode.get_timestamp_for_height`

When we introduced the "Light Wallet" mode, it became possible to for `get_timestamp_for_height` to fail, because we might try to process a coin that has a spend height higher than any untrusted peer we are connected to has.

Quex made the issue more visible in this commit:
https://github.com/Chia-Network/chia-blockchain/commit/76a78ed180

This PR is not a "fix" for that issue, but makes clear what is happening. Once we reach consensus on the fix, I'll make a new PR.

I think the version of `get_timestamp_for_height` returning `Optional` is better, for two reasons:

One, it shows clearly that the call can fail.

Two, mypy will force the user to deal with the `None` case.

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -252,17 +252,18 @@ class DataLayerWallet:
         singleton_record: Optional[SingletonRecord] = await self.wallet_state_manager.dl_store.get_latest_singleton(
             launcher_id
         )
+        timestamp = await self.wallet_state_manager.wallet_node.get_timestamp_for_height(height)
+        if timestamp is None:
+            raise ValueError("Error fetching timestamp for height '%s'.", height)
         if singleton_record is not None:
             if (  # This is an unconfirmed singleton that we know about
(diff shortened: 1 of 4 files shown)


altendky:
The peer could go away.  I'm not sure about the timing as to how big a race window there is.

Quexington:
> The peer could go away. I'm not sure about the timing as to how big a race window there is.

If this is the entire risk then I'm not sure that not just adding it to the retry store is not a fine enough solution.  Wouldn't seem like it would justify the none handling everywhere.

Quexington:
I didn't see at first that this was removing the throwing of the `PeerRequestException` from the function as well.  I think this is a bad idea, as this is a call we expect to be somewhat flaky and throwing that error ensures that any processing of a coin state that fails due to flakiness will be retried later.  I do however see that not having peers of that height should be another trigger to throw the error.

aqk:
See https://github.com/Chia-Network/chia-blockchain/pull/14857

PR #14857, opened 2023-03-20
Check if any peers have a peak before making network calls to ask for block timestamps

aqk:
Check if any peers have a peak before making network calls to ask for block timestamps

### Purpose:

Optimization

### Current Behavior:

We currently will make an unnecessary wallet protocol call to every peer if none of them have the block we are interested in, in `get_timestamp_for_height`

### New Behavior:

We look at the cached peak height for each peer to determine whether to make the network call.

### Testing Notes:

If our cached block heights are not up to date, we could fail where we would have succeeded.

This issue should be resolved in an updated node <-> wallet protocol update.

Needs tests.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -1006,8 +1006,11 @@ class WalletNode:
 
     async def get_timestamp_for_height(self, height: uint32) -> uint64:
         """
-        Returns the timestamp for transaction block at h=height, if not transaction block, backtracks until it finds
-        a transaction block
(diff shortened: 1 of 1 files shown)


chia/wallet/wallet_node.py
```diff
@@ -1017,8 +1020,11 @@ async def get_timestamp_for_height(self, height: uint32) -> uint64:
             if cache_ts is not None:
                 return cache_ts
 
+        if all([peak[0] < height for peak in self.node_peaks.values()]):
```
aqk, line 1023: This check would be better done in the loop below.

PR #14863, opened 2023-03-21
wallet: Some tweaks and rename of `WalletStateManager.new_coin_state`

xdustinface:
### Purpose:

See individual commit messages.

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1013,7 +1013,7 @@ class WalletStateManager:
             wallet_type = WalletType.NFT
         return wallet_id, wallet_type
 
-    async def new_coin_state(
+    async def add_coin_states(
         self,
         coin_states: List[CoinState],
         peer: WSChiaConnection,
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1116,7 +1116,7 @@ class WalletStateManager:
                     # if the coin has been spent
                     elif coin_state.created_height is not None and coin_state.spent_height is not None:
                         self.log.debug("Coin Removed: %s", coin_state)
-                        children: Optional[List[CoinState]] = None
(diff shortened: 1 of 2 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -1116,7 +1116,7 @@ async def add_coin_states(
                     # if the coin has been spent
                     elif coin_state.created_height is not None and coin_state.spent_height is not None:
                         self.log.debug("Coin Removed: %s", coin_state)
-                        children: Optional[List[CoinState]] = None
```
arvidn, line 1119: this is a very long function (a separate issue). It's not trivial to tell that the patch has identical behavior. It looks like `fork_height`,  `peer` and `coin_name` is not mutated, which is a good start. Now, is `children` used unconditionally in this scope?

It's used if `record is None` and it looks like the other call is at the same indentation level as the new call. So this might save us a call. Unless there are any early exits. It looks like the only possibility of an early exit would be an exception.

chia/wallet/wallet_state_manager.py
```diff
@@ -1045,7 +1045,6 @@ async def add_coin_states(
                     # This only succeeds if we don't raise out of the transaction
                     await self.retry_store.remove_state(coin_state)
 
-                    existing: Optional[WalletCoinRecord]
```
arvidn, line 1047: it might be nice to preserve the hint though, down where we assign to this variable

arvidn (approved):

chia/wallet/wallet_state_manager.py
xdustinface: > this is a very long function (a separate issue)

Yes im working towards fixing, this PR here is also kind of preparation for this and https://github.com/Chia-Network/chia-blockchain/pull/14870 is also meant to be a step in that direction. 

> Now, is `children` used unconditionally in this scope?

From what i see when searching for it in this function, yes.

> It looks like the only possibility of an early exit would be an exception.

Yeah and for this case it probably doesn't make sense to delay it.

PR #14866, opened 2023-03-21
Correct SQLite logging when using db_wrapper.manage_connection()

altendky:
### Purpose:

Previously the log file was closed prior to even yielding the connection so none of the actual use of the connection would be expected to write to the file.


arvidn (dismissed):

arvidn (dismissed):

wallentx (dismissed):
Mypy failing 😭

arvidn (approved):

altendky:
> Mypy failing sob

Could you dismiss the requested change?

PR #14871, opened 2023-03-21
Increase NFT wallet test timeout to 55m

emlowe:
While further research is ongoing, raise NFT wallet timeout to 55m


emlowe:
Closing

PR #14878, opened 2023-03-22
data_layer: Simplify wallet get and create code 

xdustinface:
### Purpose:

Just some cleanup, see individual commits.

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -137,28 +137,22 @@ class DataLayerWallet:
         return self.wallet_info.id
 
     @classmethod
-    async def create_new_dl_wallet(
-        cls,
-        wallet_state_manager: WalletStateManager,
-        wallet: Wallet,
-        name: Optional[str] = "DataLayer Wallet",
-    ) -> DataLayerWallet:
+    async def create_new_dl_wallet(cls, wallet_state_manager: WalletStateManager) -> DataLayerWallet:
         """
         This must be called under the wallet state manager lock
         """
 
         self = cls()
         self.wallet_state_manager = wallet_state_manager
-        self.log = logging.getLogger(name if name else __name__)
-        self.standard_wallet = wallet
+        self.log = logging.getLogger(__name__)
+        self.standard_wallet = wallet_state_manager.main_wallet
 
         for _, w in self.wallet_state_manager.wallets.items():
             if w.type() == WalletType.DATA_LAYER:
                 raise ValueError("DataLayer Wallet already exists for this key")
 
-        assert name is not None
         self.wallet_info = await wallet_state_manager.user_store.create_wallet(
-            name,
+            "DataLayer Wallet",
             WalletType.DATA_LAYER.value,
             "",
         )
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -4,10 +4,11 @@ import dataclasses
 import logging
 import time
 from operator import attrgetter
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Type, TypeVar
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple
 
 from blspy import G1Element, G2Element
 from clvm.EvalError import EvalError
+from typing_extensions import final
 
 from chia.consensus.block_record import BlockRecord
 from chia.data_layer.data_layer_errors import OfferIntegrityError
@@ -100,9 +101,7 @@ class Mirror:
         )
 
 
-_T_DataLayerWallet = TypeVar("_T_DataLayerWallet", bound="DataLayerWallet")
-
-
+@final
 class DataLayerWallet:
     wallet_state_manager: WalletStateManager
     log: logging.Logger
@@ -115,12 +114,12 @@ class DataLayerWallet:
 
     @classmethod
     async def create(
-        cls: Type[_T_DataLayerWallet],
+        cls,
         wallet_state_manager: WalletStateManager,
         wallet: Wallet,
         wallet_info: WalletInfo,
         name: Optional[str] = None,
-    ) -> _T_DataLayerWallet:
+    ) -> DataLayerWallet:
         self = cls()
         self.wallet_state_manager = wallet_state_manager
         self.log = logging.getLogger(name if name else __name__)
@@ -139,11 +138,11 @@ class DataLayerWallet:
 
     @classmethod
     async def create_new_dl_wallet(
-        cls: Type[_T_DataLayerWallet],
+        cls,
         wallet_state_manager: WalletStateManager,
         wallet: Wallet,
         name: Optional[str] = "DataLayer Wallet",
-    ) -> _T_DataLayerWallet:
+    ) -> DataLayerWallet:
         """
         This must be called under the wallet state manager lock
         """
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -116,14 +116,12 @@ class DataLayerWallet:
     async def create(
         cls,
         wallet_state_manager: WalletStateManager,
-        wallet: Wallet,
         wallet_info: WalletInfo,
-        name: Optional[str] = None,
     ) -> DataLayerWallet:
         self = cls()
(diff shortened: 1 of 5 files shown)


chia/data_layer/data_layer_wallet.py
```diff
@@ -138,34 +135,26 @@ def id(self) -> uint32:
...
-
-        await self.wallet_state_manager.add_new_wallet(self, self.wallet_info.id)
+        for wallet in wallet_state_manager.wallets.values():
+            if wallet.type() == uint8(WalletType.DATA_LAYER):
+                assert isinstance(
+                    wallet, DataLayerWallet
+                ), f"WalletType.DATA_LAYER should be a DataLayerWallet instance got: {type(wallet).__name__}"
+                return wallet
+
+        async with wallet_state_manager.lock:
```
arvidn, line 150: the lock is new, right? and looks like it's fixing possible races. If any `async` call is introduced above though, we'd have a problem. It looks like it would be safer to acquire the lock before looking for an existing wallet.

e.g. if any call is made to yield while looking for the existing wallet, another task can create it without us noticing, and we then create it again

arvidn (dismissed):
looks good. I think it would make it a bit more robust if you acquire the mutex before looking for an existing wallet though

arvidn (dismissed):

xdustinface:
@arvidn Changed it again to lock only if its not already locked and release only if it wasn't locked because it is in some cases also called within existing lock contexts.

xdustinface:
@arvidn Okay so i dropped the idea of merging the get/create code into `DataLayerWallet.get_or_create` because we not necessarily need to lock the wallet state manager in most cases where we are only interested in finding the wallet in `WalletStateManager.wallets` imo. And merging the code would add those cases under the lock which would change behaviour of data layer RPC endpoints and one case during state processing. So, instead i now reverted those commits and added [cd895df](https://github.com/Chia-Network/chia-blockchain/pull/14878/commits/cd895df700368c163665594f30d905cf9d671e4a) and  [23ec318](https://github.com/Chia-Network/chia-blockchain/pull/14878/commits/23ec318c760c1a4e32df53df33043d2a4e8a092e).

chia/rpc/wallet_rpc_api.py
```diff
@@ -3120,18 +3120,11 @@ async def create_new_dl(self, request) -> Dict:
...
-        dl_wallet: DataLayerWallet
-        for _, wallet in self.service.wallet_state_manager.wallets.items():
-            if WalletType(wallet.type()) == WalletType.DATA_LAYER:
-                assert isinstance(wallet, DataLayerWallet)
-                dl_wallet = wallet
-                break
-        else:
+        try:
+            dl_wallet = self.service.wallet_state_manager.get_dl_wallet()
+        except ValueError:
```
altendky, line 3125: I think this could be in the lock?  Or am I misunderstanding the existing comments about locking etc.  At which point basically just the whole function could be in a single acquisition of the lock.

chia/wallet/wallet_state_manager.py
```diff
@@ -1745,9 +1739,11 @@ async def convert_puzzle_hash(self, wallet_id: uint32, puzzle_hash: bytes32) ->
...
 
-    def get_dl_wallet(self) -> Optional[DataLayerWallet]:
-        for _, wallet in self.wallets.items():
+    def get_dl_wallet(self) -> DataLayerWallet:
+        for wallet in self.wallets.values():
             if wallet.type() == WalletType.DATA_LAYER.value:
-                assert isinstance(wallet, DataLayerWallet)
+                assert isinstance(
+                    wallet, DataLayerWallet
+                ), f"WalletType.DATA_LAYER should be a DataLayerWallet instance got: {type(wallet).__name__}"
```
altendky, line 1747: thanks!

chia/wallet/wallet_state_manager.py
```diff
@@ -1745,9 +1739,11 @@ async def convert_puzzle_hash(self, wallet_id: uint32, puzzle_hash: bytes32) ->
...
                 return wallet
-        return None
+        raise ValueError("DataLayerWallet not available")
```
altendky, line 1749: Could afford to be it's own error, but sure, this fits with existing patterns.

chia/wallet/wallet_state_manager.py
```diff
@@ -1303,15 +1301,11 @@ async def add_coin_states(
...
-                                    for _, wallet in self.wallets.items():
-                                        if wallet.type() == WalletType.DATA_LAYER.value:
-                                            assert isinstance(wallet, DataLayerWallet)
-                                            dl_wallet = wallet
-                                            break
-                                    else:  # No DL wallet exists yet
+                                    try:
+                                        dl_wallet = self.get_dl_wallet()
+                                    except ValueError:
                                         dl_wallet = await DataLayerWallet.create_new_dl_wallet(
```
altendky, line 1307: Isn't this a call that should be made under a wallet state manager lock but is not?

chia/wallet/wallet_state_manager.py
```diff
@@ -1047,14 +1043,16 @@ async def add_coin_states(
...
+                            dl_wallet = self.get_dl_wallet()
                             if (
-                                await potential_dl.get_singleton_record(coin_name) is not None
+                                await dl_wallet.get_singleton_record(coin_name) is not None
                                 or coin_state.coin.puzzle_hash == MIRROR_PUZZLE_HASH
                             ):
-                                wallet_id = potential_dl.id()
-                                wallet_type = potential_dl.type()
+                                wallet_id = dl_wallet.id()
+                                wallet_type = dl_wallet.type()
```
altendky, line 1053: This probably belongs in the `else:` or such so you don't start consuming `ValueError` from it.  `try:` blocks should be minimal.

altendky (changes requested):
The locking comments can be ignored for this PR.  We should handle them at some point though so I at least made notes.  There are a couple other points to be addressed here.

chia/wallet/wallet_state_manager.py
xdustinface: Yeah i just moved it into there from all the outer places, we use the `ValueError` in a bunch of other related places so thats a separate PR.

chia/wallet/wallet_state_manager.py
xdustinface: See 471aedc9778bff68ee48e11ac23fb000742d4e95.

chia/rpc/wallet_rpc_api.py
xdustinface: Kind of yes, for the very first time its called there is a race and but after i guess it doesn't matter because there always will be only DL wallet. But still it seems there might be some cases where we could get "size changed during iteration" issues... However, this PR just retains behaviour while cleaning up some duplicated code now, moving it into the lock means those endpoints will block also for wallet fetching so i think instead of just locking there, we should cache the DL wallet separately in the wallet state manager since it can only be one and just access the cached wallet.

altendky (dismissed):

altendky (approved):

arvidn (approved):

chia/rpc/wallet_rpc_api.py
```diff
@@ -3191,119 +3176,93 @@ async def dl_latest_singleton(self, request) -> Dict:
...
-                only_confirmed = request.get("only_confirmed")
-                if only_confirmed is None:
-                    only_confirmed = False
-                record = await wallet.get_latest_singleton(bytes32.from_hexstr(request["launcher_id"]), only_confirmed)
-                return {"singleton": None if record is None else record.to_json_dict()}
-
-        raise ValueError("No DataLayer wallet has been initialized")
+        only_confirmed = request.get("only_confirmed")
+        if only_confirmed is None:
+            only_confirmed = False
```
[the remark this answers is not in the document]
xdustinface: Hm, how is `only_confirmed` related here?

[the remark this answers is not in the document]
altendky: A couple points.  `request` is unhinted and thus either this method is either not analyzed or `request` is `Any`.  I think the latter since there is a `Dict` return type hint.  Even if `request` were hinted, we hint as `Dict[str, Any]` so `only_confirmed` would still be `Any`.

This is the mess that will remain even if we ever get through the existing mypy loose-checking module list.

PR #14883, opened 2023-03-22
Dl plugin service

almogdepaz:
### Purpose:
allow usage of custom downloaders and uploaders for DL files

### Current Behavior:
we only use http for download and the filesystem for writing files

### New Behavior:
allow definition of new download and upload service


chia/data_layer/s3_plugin_service.py
```diff
@@ -0,0 +1,164 @@
...
+from urllib.parse import urlparse
+
+import boto3 as boto3
+import yaml
+from aiohttp import web
+from botocore.exceptions import ClientError
+
+from chia.types.blockchain_format.sized_bytes import bytes32
+
+log = logging.getLogger(__name__)
```
altendky, line 19: As it is now, this looks to be unused.  Maybe we should code it to log to stdout by default instead of using `print()`?  This could be a separate future change.

chia/data_layer/s3_plugin_service.py
```diff
@@ -0,0 +1,164 @@
...
+        self.urls = urls
+        self.instance_name = instance_name
+
+    async def check_store_id(self, request: web.Request) -> web.Response:
+        self.update_instance_from_config()
+        try:
+            data = await request.json()
+        except Exception as e:
+            print(f"failed parsing request {request} {e}")
+            return web.json_response({"handles_url": False})
```
altendky, line 60: Seems like this lacks some clarity since we don't actually know the answer to whether we handle the URL that was intended to be asked about.  This could be considered as a later debuggability refinement task.

chia/data_layer/s3_plugin_service.py
```diff
@@ -0,0 +1,164 @@
...
+        store_id = bytes32.from_hexstr(data["id"])
+        if store_id.hex() in self.store_ids:
```
altendky, line 62: I would rather `bytes32.from_hexstr()` the config data than `.hex()` this.  Make all the in-memory operations use non-serialized forms.

chia/data_layer/s3_plugin_service.py
```diff
@@ -0,0 +1,164 @@
...
+
+    async def upload(self, request: web.Request) -> web.Response:
+        try:
+            data = await request.json()
+            store_id = bytes32.from_hexstr(data["id"])
+            bucket = self.get_bucket(store_id)
+            try:
+                full_tree_path = Path(data["full_tree_path"])
+                diff_path = Path(data["diff_path"])
+                self.boto_client.upload_file(full_tree_path, bucket, full_tree_path.name)
```
altendky, line 74: I definitely hesitate to provide sync I/O calls from an async function as a "reference".  Hopefully in most other cases folks will be able to find a good project in async form, but providing them with good patterns as references in the sync cases would be good.  This can presumably be considered for a later task.

But, it looks like we already use a `ThreadPoolExecutor` below for downloading?

chia/data_layer/s3_plugin_service.py
```diff
@@ -0,0 +1,164 @@
...
+        return web.json_response({"handles_store": False})
+
+    async def upload(self, request: web.Request) -> web.Response:
+        try:
+            data = await request.json()
+            store_id = bytes32.from_hexstr(data["id"])
+            bucket = self.get_bucket(store_id)
+            try:
+                full_tree_path = Path(data["full_tree_path"])
+                diff_path = Path(data["diff_path"])
```
altendky, line 73: These aren't intended to be part of the exception handling at this level.

```suggestion
            full_tree_path = Path(data["full_tree_path"])
            diff_path = Path(data["diff_path"])
            try:
```

chia/data_layer/s3_plugin_service.py
```diff
@@ -0,0 +1,164 @@
...
+        bucket = parse_result.netloc
+        target_filename = client_folder.joinpath(filename)
+        # Create folder for parent directory
+        target_filename.parent.mkdir(parents=True, exist_ok=True)
+        with concurrent.futures.ThreadPoolExecutor() as pool:
+            await asyncio.get_running_loop().run_in_executor(
+                pool, functools.partial(self.boto_client.download_file, bucket, filename, str(target_filename))
+            )
+
+        return web.json_response({"downloaded": True})
```
altendky, line 115: I'm not sure what pattern we want to put in a reference, but the lack of error handling with a controlled response here is inconsistent.

altendky (changes requested):
@fizpawiz and @ccornick, a lot of the comments here should probably be discussed as tasks separate from merging this PR.  Feel free to look over if you want and we can also setup a call to go through them.

---

I think it would be helpful to log at least what uploader is used.  Perhaps the plugins should have an identification endpoint or field in existing calls where they can state how they should be identified in logs, in addition to their URL.

---

This could be addressed as a future change, but I think the pattern of only processing for the first plugin that is willing has a couple more things that make it awkward, beyond those previously discussed.  The order of the URLs in the config can affect where the data goes.  Which I guess could be confusing in some cases and a feature in others.  Also, a given plugin failing can result in the uploads suddenly going somewhere else.  Or if the plugin still reports it can handle the data but then fails while trying to process it, the data ends up not making it anywhere except the filesystem.

chia/server/start_data_layer.py
```diff
@@ -86,7 +99,9 @@ async def async_main() -> int:
         overwrite=False,
     )
 
-    service = create_data_layer_service(DEFAULT_ROOT_PATH, config)
+    uploaders: List[str] = config["data_layer"]["uploaders"]
+    downloaders: List[str] = config["data_layer"]["downloaders"]
+    service = create_data_layer_service(DEFAULT_ROOT_PATH, config, downloaders, uploaders)
```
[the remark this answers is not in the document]
almogdepaz: why ?

chia/data_layer/s3_plugin_service.py
almogdepaz: what if parsing fails ?

chia/data_layer/data_layer.py
```diff
@@ -404,7 +411,21 @@ async def fetch_and_validate(self, tree_id: bytes32) -> None:
             except Exception as e:
                 self.log.warning(f"Exception while downloading files for {tree_id}: {e} {traceback.format_exc()}.")
 
+    async def get_downloader(self, url: str) -> Optional[str]:
+        request_json = {"url": url}
+        for d in self.downloaders:
+            async with aiohttp.ClientSession() as session:
+                try:
+                    async with session.post("http://" + d + "/check_url", json=request_json) as response:
```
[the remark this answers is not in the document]
almogdepaz: did we decide to allow multiple schemes here ? i dont remember we said that...

chia/data_layer/data_layer.py
```diff
@@ -404,7 +411,21 @@ async def fetch_and_validate(self, tree_id: bytes32) -> None:
             except Exception as e:
                 self.log.warning(f"Exception while downloading files for {tree_id}: {e} {traceback.format_exc()}.")
 
+    async def get_downloader(self, url: str) -> Optional[str]:
+        request_json = {"url": url}
+        for d in self.downloaders:
+            async with aiohttp.ClientSession() as session:
+                try:
```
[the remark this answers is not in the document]
almogdepaz: but there is nothing that can throw there as far as i can tell, i prefer not including lines that dont throw in the try

chia/data_layer/s3_plugin_service.py
almogdepaz: i dont understand what you are trying to say here

chia/data_layer/s3_plugin_service.py
almogdepaz: what do you mean by "in mempry operations ?
i dont want to not use the conversions so we can throw in case we gat invalid input

chia/data_layer/data_layer.py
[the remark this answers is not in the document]
emlowe: The only other scheme would be `https` which might require other configurable items (certificates, etc) - that said, I'm inclined to put the scheme in the config as well.

chia/data_layer/data_layer.py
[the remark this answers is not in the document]
altendky: Certainly a focused `try:` block is important in many cases.  In my mind the intent here was that no matter what, we log an error and try the next downloader.  I think that having the `try:` block cover the entirety of the loop both better expresses this and encourages retention of that intent during future modifications and maintenance.  I also hesitate to assume that aiohttp will never add any possible cases of raising to https://github.com/aio-libs/aiohttp/blob/v3.8.4/aiohttp/client.py#L197 and https://github.com/aio-libs/aiohttp/blob/v3.8.4/aiohttp/client.py#L1105 (beyond the ones already there).

I did want to explain, but I don't see an error here so this conversation can be closed once you have read this.

chia/data_layer/s3_plugin_service.py
altendky: Failure here to parse the request results in the same response to the client as if we successfully parse the request and decide that we do not handle the URL.  I think it would be useful to respond with an indication of the error so that the two cases can be distinguished.

chia/data_layer/s3_plugin_service.py
altendky: A hexadecimal encoded string of bytes is a serialization of the data.  I think this should be used when serialization over a network connection or to a file is needed (and in this case that those mechanisms are using something like JSON that can't handle bytes directly).  Other things like storage in dicts and comparisons etc would be "in memory operations".

I like that the request `data["id"]` is processed as you have it, `store_id = bytes32.from_hexstr(data["id"])`.  I am suggesting that the `self.store_ids` be populated in the same way instead of storing the hexadecimal serialization.  Then we would just have `if store_id in self.store_ids:`

I probably should have raised this issue at the code populating `self.store_ids` or such.

chia/data_layer/s3_plugin_service.py
altendky: With my suggestion it would be categorized as `failed handling request` instead of `failed uploading file to aws` as it is now.  That seems more representative to me.

chia/server/start_data_layer.py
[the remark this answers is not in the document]
almogdepaz: seems not important for this pr if its consistency only, how critical is it to you ?

chia/data_layer/data_layer.py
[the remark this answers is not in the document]
almogdepaz: ok ill add that to the config

chia/data_layer/s3_plugin_service.py
almogdepaz: data layer currently doesn't handle those cases differently so i think its redundant if we want to handle them then we should and we should add different responses until then i dont think its called for, just confusing

chia/data_layer/s3_plugin_service.py
emlowe: Logging to a file is one of the items on Ken's list, so I think we leave this as is and make some followup PRs to use it

chia/data_layer/s3_plugin_config.yml
```diff
@@ -0,0 +1,28 @@
...
+    access_key_id: "xxx"
+    secret_access_key: "xxx"
+    region: "xxx"
+
+  store_ids: ["xxx"]
+
+  urls: [ "xxx"]
+
+  buckets:
+    chia-datalayer-test-bucket-2: ["xxx"]
```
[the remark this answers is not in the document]
emlowe: No, just test names - no rationale to them afaict

chia/data_layer/download_data.py
```diff
@@ -92,7 +92,7 @@ async def write_files_for_root(
     root: Root,
     foldername: Path,
     overwrite: bool = False,
-) -> bool:
+) -> Tuple[bool, Path, Path]:
```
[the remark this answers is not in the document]
emlowe: Added

emlowe (dismissed):

altendky (dismissed):

emlowe (dismissed):

emlowe (approved):

justinengland:
"the plugins are trusted services (you have to configure them in config.yaml and run them) - so IMO it's ignorable for now"- @emlowe , regarding the one failure on this run. They will follow up with resolution of this issue in a later revision.

PR #14884, opened 2023-03-22
change self peer connection error to info, remove traceback

altendky:
### Purpose:

Self connection errors don't need to be `ERROR` level in the log nor do they need to have tracebacks.  This will mostly happen when a node is unaware of one or more of its own public IPs being used by the network.  Could be IPv4 vs. IPv6, or multiple IPs, or just being behind a NAT.

### Current Behavior:

```python-traceback
2023-03-22T14:00:46.076 full_node full_node_server        : ERROR    Exception Trying to connect to a peer (PeerInfo(_ip=IPv4Address('<REDACTED>'), _port=58445)) with the same peer_id: <REDACTED>, exception Stack: Traceback (most recent call last):
  File "/home/ubuntu/chia-blockchain/chia/server/server.py", line 525, in start_client
    raise RuntimeError(f"Trying to connect to a peer ({target_node}) with the same peer_id: {peer_id}")
RuntimeError: Trying to connect to a peer (PeerInfo(_ip=IPv4Address('<REDACTED>'), _port=58445)) with the same peer_id: <REDACTED>
```

### New Behavior:

`INFO` level logging without a traceback


chia/server/server.py
```diff
@@ -42,6 +42,13 @@
 max_message_size = 50 * 1024 * 1024  # 50MB
 
 
+class SamePeerIdError(Exception):
```
xdustinface, line 45: Maybe `class SamePeerIdError(ServerException)` in `chia.utils.errors`?

altendky: Something in that direction sounds reasonable.  I don't see a `ServerException` there though.  Maybe still in an unmerged PR?

xdustinface: >I don't see a `ServerException` there though. Maybe still in an unmerged PR?

I thought i added it at some point. But yes, then maybe just add it and if i have a PR where i add it i can remove it when i revisit it :)

chia/util/errors.py
```diff
@@ -309,3 +312,14 @@ class InvalidPathError(Exception):
...
         self.path = path
+
+
+class ServerError(Exception):
+    pass
+
+
+class SamePeerIdError(ServerError):
+    def __init__(self, peer: PeerInfo, node_id: bytes32) -> None:
+        super().__init__(f"Trying to connect to a ourselves: {peer} {node_id}")
```
xdustinface, line 323: I guess it would make more sense to not talk about connection in here since `SamePeerIdError` sounds more generic than just "we connect to ourself". Also storing `peer` and `node_id` without any usage seems pointless? :) Not sure what else i would do but this seems not well thought out so far? Maybe just don't add a `__init__` for `SamePeerIdError` and instead pass in the message in where you raise it?

altendky: The wording of the message is wrong.  The rest seems useful?  Taking parameters and formatting the message brings consistency.  Storing the parameters makes them available when someone wants to catch the exception and be aware of the relevant information.  Yes, people could just add it when they need it.  Makes for bigger diffs and this also leads people to have exceptions which can be handled better I think.

I'll revisit the message and see what you think of the rest.

xdustinface: > Storing the parameters makes them available when someone wants to catch the exception and be aware of the relevant information. 

_When_ someone wants to use it. Can you imagine a case where its useful for this exception? I just don't think its very helpful to add not used things all over the place only because they _might_ be used in the future but most likely will just stay there unused forever.. 

> Yes, people could just add it when they need it. Makes for bigger diffs and this also leads people to have exceptions which can be handled better I think.

Either now or then the diff is bigger. If it gets added when its actually used you can at least better understand why it was introduced when checking the history, i think.

xdustinface (approved):

emlowe (approved):

PR #14887, opened 2023-03-23
daemon: Fix removal while iterating over connections set

xdustinface:
### Purpose:

Fix the issue below which was introduced in #14662.

```
 2023-03-23T01:07:51.905 daemon aiohttp.server             : ERROR    Error handling request
   File "/Users/dustinface/GIT/chia-blockchain/venv/lib/python3.11/site-packages/aiohttp/web_protocol.py", line 433, in _handle_request
     resp = await request_handler(request)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "/Users/dustinface/GIT/chia-blockchain/venv/lib/python3.11/site-packages/aiohttp/web_app.py", line 504, in _handle
  resp = await handler(request)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dustinface/GIT/chia-blockchain/chia/daemon/server.py", line 262, in incoming_connection
     await self.send_all_responses(connections, response)
   File "/Users/dustinface/GIT/chia-blockchain/chia/daemon/server.py", line 282, in send_all_responses
     for connection in connections:
 RuntimeError: Set changed size during iteration
```

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -279,7 +279,7 @@ class WebSocketServer:
         return ws
 
     async def send_all_responses(self, connections: Set[WebSocketResponse], response: str) -> None:
-        for connection in connections:
+        for connection in connections.copy():
             try:
                 await connection.send_str(response)
             except Exception as e:


chia/daemon/server.py
```diff
@@ -340,8 +340,7 @@ async def handle_message(
         destination = message["destination"]
         if destination != "daemon":
             if destination in self.connections:
-                sockets = self.connections[destination]
-                return dict_to_json_str(message), sockets
+                return dict_to_json_str(message), self.connections[destination].copy()
```
arvidn, line 343: It's not visible in this patch, but your change suggests that there is at least one caller of `handle_message()` that will iterate over the returned set with `async` calls interleaved. Is that right?
I think it would make sense to add a comment here to remind future readers of this code why we need to copy this set. We don't have a general policy to always copy containers or objects we returns (even though it would probably be more robust), so this is a bit special in that way.

arvidn (dismissed):

chia/daemon/server.py
xdustinface: Yes, see the stack trace in the PR description. There you see `send_all_responses` gets passed in the `connections` received from `handle_message` and in there we iterate over it and then remove the connection if it fails. I changed the place where we copy now with the force push. Instead of copying in `handle_message` it now copies when we actually iterate/remove which means we copy unnecessarily in few cases but only single connection sets, so i think its worth the more obvious way of copying.

emlowe:
TBH - I was a little concerned about this when writing the code

emlowe (approved):

chia/daemon/server.py
wallentx: This appears to be resolved

PR #14896, opened 2023-03-24
wallet: Some cleanup of `WalletNode.fetch_puzzle_solution`

xdustinface:
### Purpose:

Cleanup - working towards hinted `WalletStateManager`.

1. Move it out of `WalletNode` and make it a free function in `wallet_sync_util.py`
2. Rename it to `fetch_coin_spend`. Even though we request the puzzle solution via `request_puzzle_solution`, `CoinSpend` is what we get in return and to me it seems clearer when reading code this way.
3. Adjust the retry-store test have the flakiness for coin spend fetching on the server side (`FullNodeAPI`), not on the client side (`WalletNode`).
4. Introduce `fetch_coin_spend_from_coin_state` to make `mypy` happy in some cases where we fetch the `CoinSpend` for an existing `CoinState` which has a `Optional[int]` spent height.

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -217,7 +217,7 @@ class DataLayerWallet:
             await self.new_launcher_spend(spend, peer, height)
         else:
             launcher_state: CoinState = await self.get_launcher_coin_state(launcher_id, peer)
-            launcher_spend = await self.wallet_state_manager.wallet_node.get_coin_spend_for_coin_state(
+            launcher_spend = await self.wallet_state_manager.wallet_node.fetch_coin_spend_for_coin_state(
                 launcher_state, peer
             )
             await self.new_launcher_spend(launcher_spend, peer)
@@ -279,7 +279,7 @@ class DataLayerWallet:
         ] = await self.wallet_state_manager.coin_store.get_coin_record(new_singleton.name())
         while new_singleton_coin_record is not None and new_singleton_coin_record.spent_block_height > 0:
             # We've already synced this before, so we need to sort of force a resync
-            parent_spend: CoinSpend = await self.wallet_state_manager.wallet_node.fetch_puzzle_solution(
+            parent_spend: CoinSpend = await self.wallet_state_manager.wallet_node.fetch_coin_spend(
                 new_singleton_coin_record.spent_block_height, new_singleton, peer
             )
             await self.singleton_removed(parent_spend, new_singleton_coin_record.spent_block_height)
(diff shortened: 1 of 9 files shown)


emlowe:
Any thought to just changing `fetch_puzzle_solution` instead?

xdustinface:
> Any thought to just changing `fetch_puzzle_solution` instead?

@emlowe Indeed i initially thought about it too, but making it accepting an `Optional[int]` only to then assert the optional to be `not None` seems odd, no? I reworked the PR a bit and made the functions both free functions. Let me know what you think.

emlowe (dismissed):
looks ok

chia/data_layer/data_layer_wallet.py
```diff
@@ -226,9 +227,7 @@ async def track_new_launcher_id(
             await self.new_launcher_spend(spend, peer, height)
         else:
             launcher_state: CoinState = await self.get_launcher_coin_state(launcher_id, peer)
-            launcher_spend: CoinSpend = await self.wallet_state_manager.wallet_node.fetch_puzzle_solution(
```
arvidn, line 229: do we still need this function? `fetch_puzzle_solution()`?

arvidn (dismissed):

chia/data_layer/data_layer_wallet.py
xdustinface: It was renamed to `fetch_coin_spend` and we still use it yes. But there is additionally also `fetch_coin_spend_for_coin_state` now. See the PR description :)

justinengland:
@arvidn it appears we now have merge conflicts here. Sorry!

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -1184,11 +1185,17 @@ async def new_func(*args, **kwargs):
 
             return new_func
 
-        def flaky_fetch_coin_spend(node, func):
+        request_puzzle_solution_failure_tested = False
+
+        def flaky_request_puzzle_solution(func):
```
arvidn, line 1190: renaming this test doesn't look aligned with the other renames of `puzzle_solution` -> `coin_spend` renames

arvidn (approved):

tests/wallet/sync/test_wallet_sync.py
xdustinface: Its not really renamed in the final diff unless i miss something? And it tests the flakiness of `request_puzzle_solution` on the fullnode side so i think the naming is fine?

PR #14906, opened 2023-03-24
Adding py version exclusion

wallentx:
### Purpose:
To reduce CI overhead and wait time, only running our tests against ALL python versions on release/** branches

### Current Behavior:
Runs a version of our test for python 3.7, 3.8, 3.9, 3.10, and 3.11, for all OSes, every time test CI is triggered.

### New Behavior:
Only does the above for release/** branches, else it runs 3.9

### DIFF
--- a/.github/workflows/test-single.yml
+++ b/.github/workflows/test-single.yml
@@ -63,6 +63,7 @@ jobs:
             matrix: '3.7'
             exclude_from:
               limited: True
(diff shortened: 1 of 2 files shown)


arvidn (approved):

trepca (approved):

wjblanke (approved):
aok

PR #14911, opened 2023-03-27
wallet: Improve coin state retry wait logic and retry store test

xdustinface:
### Purpose:

- Simplify the retry logic
- Remove an indent for all the retry code
- Simplify the test
- Speedup the test. Times on my system:
    - `main`: 51s
    - This PR: 225s

### Current Behavior:

The logic sleeps two/three times 5 seconds before retrying the coin states.

### New Behavior:

The code sleeps one time for `coin_state_retry_seconds` (10s) seconds before retrying the coin states.

### Testing Notes:

Covered by tests.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -117,6 +117,7 @@ class WalletNode:
     LONG_SYNC_THRESHOLD: int = 300
     last_wallet_tx_resend_time: int = 0
     # Duration in seconds
+    coin_state_retry_seconds: int = 10
     wallet_tx_resend_timeout_secs: int = 1800
     _new_peak_queue: Optional[NewPeakQueue] = None
 
@@ -362,7 +363,6 @@ class WalletNode:
             self.wallet_state_manager.set_callback(self.state_changed_callback)
 
         self.last_wallet_tx_resend_time = int(time.time())
-        self.last_state_retry_time = int(time.time())
         self.wallet_tx_resend_timeout_secs = self.config.get("tx_resend_timeout_secs", 60 * 60)
         self.wallet_state_manager.set_pending_callback(self._pending_tx_handler)
         self._shut_down = False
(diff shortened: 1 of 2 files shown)


chia/wallet/wallet_node.py
```diff
@@ -466,41 +466,40 @@ async def _messages_to_resend(self) -> List[Tuple[Message, Set[bytes32]]]:
...
+                        except Exception as e:
+                            self.log.exception(f"Exception while adding states.. : {e}")
+                        else:
+                            await self.wallet_state_manager.blockchain.clean_block_records()
             except asyncio.CancelledError:
                 self.log.info("Retry task cancelled, exiting.")
                 raise
+            finally:
+                if not self._shut_down:
+                    await asyncio.sleep(self.coin_state_retry_seconds)
```
arvidn, line 502: this block looks weird to me. I would expect this `sleep()` as the last statement inside the try-catch block. Putting it in a finally block means we'll sleep when any exception is raised, before re-raising. Even `CancelledError` will have to sleep. The only exception is if `self._shut_down` is `True`.

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -1251,36 +1251,30 @@ async def new_func(*args, **kwargs):
             await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph))
             await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(bytes32([0] * 32)))
 
-            async def len_gt_0(func, *args):
-                return len((await func(*args))) > 0
```
arvidn, line 1255: thanks. This is quite a cryptic name

chia/wallet/wallet_node.py
```diff
@@ -466,41 +466,40 @@ async def _messages_to_resend(self) -> List[Tuple[Message, Set[bytes32]]]:
     async def _retry_failed_states(self):
         while not self._shut_down:
             try:
-                await asyncio.sleep(5)
-                current_time = time.time()
-                if self.last_state_retry_time < current_time - 10:
```
arvidn, line 471: can `self.last_state_retry_time` change from another task/coroutine? If so, it might not be safe to assume that sleeping 10 seconds is the same as this loop, checking the retry time regularly.

arvidn (changes requested):
I think the `finally` block may be a mistake

chia/wallet/wallet_node.py
xdustinface: No it can't this PR removes all occurrences of it.

chia/wallet/wallet_node.py
xdustinface: > this block looks weird to me. I would expect this `sleep()` as the last statement inside the try-catch block. Putting it in a finally block means we'll sleep when any exception is raised, before re-raising. Even `CancelledError` will have to sleep. The only exception is if `self._shut_down` is `True`.

Hm true, it will also sleep for other exceptions but i think in canceled case we will skip it because it only is canceled if `self._shut_down` is `True`, at least thats why i added it :) Initially i checked if `self._retry_failed_states_task.canceled()` is `True` to skip the sleep. But anyway, your right something needs to be changed here!

chia/wallet/wallet_node.py
```diff
@@ -466,38 +466,37 @@ async def _messages_to_resend(self) -> List[Tuple[Message, Set[bytes32]]]:
...
-                            peer = self.get_full_node_peer()
-                            if peer is None:
-                                self.log.info(f"disconnected from all peers, cannot retry state: {state}")
-                                continue
-                            else:
-                                self.log.info(
-                                    f"disconnected from peer {peer_id}, state will retry with {peer.peer_node_id}"
-                                )
+                if self.wallet_state_manager is None:
+                    await asyncio.sleep(self.coin_state_retry_seconds)
```
arvidn, line 470: so, every `continue` in this loop is expected to `sleep()` first, is that right?

arvidn (dismissed):
I think it would make sense for @trepca to take a look at this too. I believe he's been working with logic adjacent to this

chia/wallet/wallet_node.py
xdustinface: Yes otherwise we might end up in a not sleeping, blocking endless loop.

xdustinface: Okay well, now that we have the configurable timeout which is set to `1` for the test it doesn't really matter i guess, i added [00ea944](https://github.com/Chia-Network/chia-blockchain/pull/14911/commits/00ea944306865e94b3094eaa4ec65e686d2c7210).

arvidn (approved):

PR #14921, opened 2023-03-27
activate the soft-fork and soft-fork2 earlier on testnet10

arvidn:
To facilitate testing soft-fork2 features (like, `ASSERT_BEFORE_*` and other new condition codes), this patch makes the soft-fork activation heights lower when running on testnet10.

The current block height on testnet10 is about 2408931. The activation height proposed here is 5 weeks out.

I stepped through this in `pdb`, when running testnet, to ensure the constants were updated correctly.


trepca (approved):

wjblanke (approved):
aok

PR #14922, opened 2023-03-28
cleanup unused peer config

altendky:
### Purpose:

Provide clarity around which services use which ports by removing the unused port config entries.

### Current Behavior:

There are unused port config entries.

### New Behavior:

The default config file is cleaner and the configure command doesn't update the unused port entries.

### Draft For:
- [x] figure out why 8448 looks to be used but `nmap` claims it isn't open
- [x] break outs
  - [x] https://github.com/Chia-Network/chia-blockchain/pull/14928
  - [x] https://github.com/Chia-Network/chia-blockchain/pull/14932
  - [x] https://github.com/Chia-Network/chia-blockchain/pull/15024
- [x] review ui port


chia/util/initial-config.yaml
```diff
@@ -164,6 +164,7 @@ seeder:
 
 harvester:
   # The harvester server (if run) will run on this port
+  # TODO: this seems to be used, but i don't have the port open.  not sure what's going on
```
xdustinface, line 167: I think this is just used because we create a `ChiaServer` for the harvester but we don't connect to it iirc.

altendky: Thanks.  That's definitely backwards and should be a parameter or handled entirely outside.  The generic server should not be what is special casing the various usages of it.

altendky: And all those port configs should be fully removed.

emlowe (approved):
Really nice cleanup work!

PR #14926, opened 2023-03-28
Fix the mempool fee rate calculation

AmineKhaldi:
It's currently performing integer division instead of float division, resulting in incorrect sorting of mempool items by fee rate. Items with `x.y` fee rate get all treated as items of `x.0` fee rate.

### DIFF
--- a/chia/full_node/mempool.py
+++ b/chia/full_node/mempool.py
@@ -54,7 +54,7 @@ class Mempool:
             # assert_height may be NIL
             generated = ""
             if not SQLITE_NO_GENERATED_COLUMNS:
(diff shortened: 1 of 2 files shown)


trepca (commented):
should this also have a test case given it didn't fail before?

aqk (dismissed):
Thanks!

aqk:
> should this also have a test case given it didn't fail before?

Yes

AmineKhaldi:
> > should this also have a test case given it didn't fail before?
> 
> Yes

Done.

arvidn:
this is a problem in the current 1.7.1 release. does it warrant a hot-fix?

arvidn (approved):

wjblanke (approved):
aok

PR #14927, opened 2023-03-28
Adding daily full matrix test

wallentx:
### Purpose:
This adds a daily trigger of the test.yml workflow_dispatch at 8am CT, which will execute tests against all versions of python for `main` branch.

### Current Behavior:
`main` only does tests of python3.9 for PRs

### New Behavior:
We get a daily full dose of testing


altendky:
Was putting the schedule in the existing workflow already considered?

Also, thanks!

wallentx:
> Was putting the schedule in the existing workflow already considered?
> 
> Also, thanks!

I thought about this, but then I imagined that in the future that we may want to have some sort of landing space for notifications or something if this failed, and I thought that it'd be cleaner to have this as an individual item, rather than tied into the existing workflow that is a bit of a beast as-is.

altendky:
Does `gh workflow run` let us know what the run id is so we can monitor it or such?  Notifications are certainly valuable around this.  For now we'll just have to manually be checking every day and rerunning to see if they are flakes or what.  (yeah, dealing with that is separate scope from this PR)

wallentx:
> Does `gh workflow run` let us know what the run id is so we can monitor it or such? Notifications are certainly valuable around this. For now we'll just have to manually be checking every day and rerunning to see if they are flakes or what. (yeah, dealing with that is separate scope from this PR)

I've not had the time to explore this much more than just the initial thoughts. I just figured that we'd prob want this functionality in before we cut the release branch. The means of notification I don't think I have any strong opinions on yet, so whatever everyone thinks is best, I'm down with.

altendky:
I thought the release branch got full builds every time which would make this getting merged more or less unrelated to cutting the release branch.  I mean, there's no reason for it to not be in the release branch, but also no reason for it to be?

altendky (approved):

emlowe (approved):

wallentx:
> I thought the release branch got full builds every time which would make this getting merged more or less unrelated to cutting the release branch. I mean, there's no reason for it to not be in the release branch, but also no reason for it to be?

You are correct. My mind is kinda mush today

PR #14932, opened 2023-03-29
add explicit listen option for Service() and ChiaServer.start_server()

altendky:

### DIFF
--- a/chia/server/server.py
+++ b/chia/server/server.py
@@ -264,7 +264,7 @@ class ChiaServer:
             for peer_ip in to_remove_ban:
                 del self.banned_peers[peer_ip]
 
-    async def start_server(
+    async def start(
         self,
         listen: bool,
         prefer_ipv6: bool,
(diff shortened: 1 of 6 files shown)


xdustinface (changes requested):
Calling `ChiaServer.start_server` although we don't want to run a server seems strange to begin with, no? Wouldn't it make more sense to not call `ChiaServer.start_server` at all if we don't want to start a listening server instead of returning earlier based on a `listen` parameter? I guess we could just create the garbage collection task (which is the only thing left in there for non listening services) in maybe `ChiaServer.start_client` also or just in `ChiaServer.create`.

altendky:
This was indeed a tightly focused PR intended just to get the hard coded list of non-listening node types converted to parameters.  I agree with the oddity around the method name and the behavior, both existing and new.  Moving the garbage collection task creation to the `.create()` method would require making it async.  It is presently called from `Service.__init__()` which is called from `create_*_service()` which are also sync so that would be another bulk change.  Though, those are all called from their corresponding `async_main()` functions so not entirely implausible.

xdustinface (approved):

emlowe (approved):

PR #14933, opened 2023-03-29
seeder: case-insensitive DNS label matching to support bit 0x20 use

ctrlaltdel:
Some DNS resolvers (notably 8.8.8.8) have implemented the use of bit 0x20 in DNS labels to make cache poisoning attacks more difficult.

See https://datatracker.ietf.org/doc/html/draft-vixie-dnsext-dns0x20-00 for more details.

Without this patch, resolution though Google public DNS fails:

```
$ dig dns-introducer.chia.net @8.8.8.8 +short | wc -l
       0
```

Whereas it works as expected on a seeder which already deployed this patch:

```
$ dig chia.ctrlaltdel.ch @8.8.8.8 +short | wc -l
      32
```

### DIFF
--- a/chia/seeder/dns_server.py
+++ b/chia/seeder/dns_server.py
@@ -208,7 +208,9 @@ class DNSServer:
             }
 
             qname = request.q.qname
-            qn = str(qname)
+            # DNS labels are mixed case with DNS resolvers that implement the use of bit 0x20 to improve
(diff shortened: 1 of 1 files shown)


hoffmang9 (dismissed):
lgtm

hoffmang9 (approved):
refreshing my lgtm

emlowe:
Hey thanks, I was much confused why google DNS didn't work with the DNS seeders!

emlowe (approved):

PR #14940, opened 2023-03-30
Typecheck node_height.py

aqk:
Typecheck node_height.py


tests/core/node_height.py
```diff
@@ -1,20 +1,28 @@
...
+from chia.full_node.full_node_api import FullNodeAPI
+from chia.util.ints import uint32
 
-def node_height_at_least(node, h):
+
+def node_height_at_least(node: FullNodeAPI, h: uint32) -> bool:
     if node.full_node.blockchain.get_peak() is not None:
-        return node.full_node.blockchain.get_peak().height >= h
+        peak = node.full_node.blockchain.get_peak()
+        assert peak is not None
```
altendky, line 10: Since I'm not a big fan of asserts I'll at least bring up the idea of just returning `False`?  I think `None` here indicates that there have been no blocks so the height is not at least whatever was asked for.  Unless zero, I'm not sure if a height of 0 is 1 block or if that's a height of 1.

(and yes, I do assert sometimes myself...  `:|`)

aqk: @arvidn We are discussing whether it is okay to allow these tests to run in the case where `node.full_node.blockchain.get_peak()` returns `None`, and I wanted to alert you that we may have this scenario in production code - getting a (perhaps) unexpected `None` from some full node methods.

arvidn: afaict, these functions are used as predicates to `time_out_assert()`, so returning `False` seems quite reasonable if we don't have a peak yet

emlowe:

altendky (approved):

emlowe (approved):

PR #14941, opened 2023-03-30
Typecheck log_exceptions.py

aqk:
Typecheck log_exceptions.py


emlowe:

emlowe (approved):

PR #14942, opened 2023-03-30
Typecheck chia/util/hash.py

aqk:
Typecheck chia/util/hash.py


chia/util/hash.py
```diff
@@ -5,7 +5,7 @@
 from chia.types.blockchain_format.sized_bytes import bytes32
 
 
-def std_hash(b, skip_bytes_conversion: bool = False) -> bytes32:
+def std_hash(b: bytes, skip_bytes_conversion: bool = False) -> bytes32:
```
almogdepaz, line 8: looking around in the codebase we dong always pass bytes and there is a byte conversion inside std_hash, why do we want this change ?

if there is a good reason i found more calls that need refactoring

altendky: I think we expect https://github.com/Chia-Network/chia-blockchain/pull/14950 to be more likely where we go.  But, it is a first draft that I didn't expect to be complete.  Not surprisingly, some tests are failing.

PR #14946, opened 2023-03-30
Typecheck tests/util/db_connection.py

aqk:
Typecheck tests/util/db_connection.py


tests/util/db_connection.py
```diff
@@ -18,6 +19,11 @@ async def __aenter__(self) -> DBWrapper2:
 
         return self._db_wrapper
 
-    async def __aexit__(self, exc_t, exc_v, exc_tb) -> None:
+    async def __aexit__(
+        self,
+        exc_t: Callable[..., Awaitable[None]],
+        exc_v: Callable[..., Awaitable[None]],
+        exc_tb: Callable[..., Awaitable[None]],
+    ) -> None:
```
altendky, line 27: https://github.com/python/typeshed/blob/619125b72282128d3c2bccbd38518fad0671fae9/stdlib/contextlib.pyi#L52-L54

```suggestion
    async def __aexit__(
        self,
        __exc_type: Optional[Type[BaseException]],
        __exc_value: Optional[BaseException],
        __traceback: TracebackType,
    ) -> Optional[bool]:
```

There would also be a `from types import TracebackType` at the top.  https://docs.python.org/3.11/library/types.html#types.TracebackType

altendky (changes requested):

emlowe:

altendky (dismissed):

altendky (dismissed):

tests/util/db_connection.py
aqk: Thanks! I'll put that in now.

emlowe:
Closing Old PR

PR #14950, opened 2023-03-30
drop std_hash(skip_bytes_conversion=) parameter and use explicit bytes() calls

altendky:
### Purpose:

This is an exploration of improved hinting that is an alternative to https://github.com/Chia-Network/chia-blockchain/pull/14942.

### Draft For:
- [ ] more detailed checking of uses


aqk:
I agree with this. All the places that we feed the output of `std_hash` into clvm, we want to make sure the user has carefully considered the form of the byte string being fed into std_hash. I believe the uniqueness property of the CLVM number encoding conversion makes all int -> bytes conversions safe, but it's way better to be explicit, and we might catch something we didn't intend to be an int, or that unwittingly underwent an unintended conversion to int before `std_hash` was called.

PR #14954, opened 2023-03-30
expose NFT coin info's latest_height to RPCs requesting NFT info

paninaro:
### Purpose:
The GUI would like to have access to a sort key for NFTs that represents the order in which the NFTs are added into the user's wallet.


paninaro:
I was attempting to find a property that would match the order in which `nft_get_nfts` returns NFT info. This `latest_height` property doesn't seem to match in my testing however... Please advise if there's a better solution.

trepca (approved):

PR #14969, opened 2023-04-03
wallet: Reduce DB lookups in `handle_did` and `handle_nft`

xdustinface:
### Purpose:

Avoid some DB lookups during sync/coin state processing.

### Current Behavior:

`handle_did` and `handle_nft` use `WalletStateManager.get_all_wallet_info_entries` to read all wallets from the DB and then parses the extra data to find the related wallet.

### New Behavior:

`handle_did` and `handle_nft` just iterate over the cached wallets to find the related wallet.

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-import json
 import logging
 import multiprocessing.context
 import time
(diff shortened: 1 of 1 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -923,38 +921,37 @@ async def handle_nft(
                 uncurried_nft.singleton_launcher_id.hex(),
             )
             return wallet_identifier
-        for wallet_info in await self.get_all_wallet_info_entries(wallet_type=WalletType.NFT):
-            nft_wallet_info: NFTWalletInfo = NFTWalletInfo.from_json_dict(json.loads(wallet_info.data))
-            if nft_wallet_info.did_id == old_did_id and old_derivation_record is not None:
+        for nft_wallet in self.wallets.values():
+            if not isinstance(nft_wallet, NFTWallet):
+                continue
```
arvidn, line 926: our wallet protocol has a `type()` functions, doesn't it?
Maybe it won't work as well with mypy, since it won't "cast" `nft_wallet` to the right type

arvidn (approved):

chia/wallet/wallet_state_manager.py
xdustinface: Yes the `isinstance` check makes `mypy` happy.

PR #14971, opened 2023-04-03
Mempool insertion order

arvidn:
When transitioning the mempool to use sqlite for its data structure, the iteration order of mempool items whose fee-rate is the same is no longer defined. The insertion order, into the mempool, used to be the tie-breaker.

This patch makes insertion order a tie-breaker when iterating mempool items (by fee rate) and adds test coverage.

The way it's achieved is by making the rowid explicit and marked with `AUTOINCREMENT`. Every time a new item is added to the mempool, the sequence number will be incremented. Lower numbers indicate an earlier insertion time, so when iterating over fee rate in *decending* order, the tie-break sequence number is *ascending*, and vice versa.

### DIFF
--- a/chia/full_node/mempool.py
+++ b/chia/full_node/mempool.py
@@ -55,18 +55,22 @@ class Mempool:
             generated = ""
             if not SQLITE_NO_GENERATED_COLUMNS:
                 generated = " GENERATED ALWAYS AS (CAST(fee AS REAL) / cost) VIRTUAL"
-
+            # the seq field indicates the order of items being added to the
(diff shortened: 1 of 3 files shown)


AmineKhaldi (approved):

aqk (approved):
The comment for the SELECT statements at lines 181 could say

`# Iterates through all spends in increasing fee per cost, with insertion order as the tiebreaker`

https://github.com/Chia-Network/chia-blockchain/pull/14971/files#diff-1cd73731eeaeb06087e341a00a41d5a6e819772688cb56434f6687895da82dffR181

trepca (approved):

PR #14972, opened 2023-04-03
Check hint length in notficication manager to avoid raise

Quexington:


wjblanke (approved):
aok fixes tian's exception

arvidn:
looks reasonable. if we want to avoid re-introducing problems like this again in the future, we should probably have tests though

PR #14973, opened 2023-04-03
skip empty directories pre-commit

Quexington:
Sometimes empty directories are left in working copies when switching between branches and pre-commit forces some files to be created for these empty directories.  I don't see any need for this and it results in PRs like this: https://github.com/Chia-Network/chia-blockchain/pull/14858

### DIFF
--- a/tests/build-init-files.py
+++ b/tests/build-init-files.py
@@ -12,6 +12,7 @@
 from __future__ import annotations
 
 import logging
+import os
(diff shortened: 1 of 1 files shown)


aqk (approved):
Nice job tracking that down!

trepca (approved):

arvidn (approved):

PR #14974, opened 2023-04-03
remove synced status check from wallet rpc `/push_transactions` endpoint

altendky:
### Purpose:

This is causing trouble around creating transactions and then submitting them as the creation triggers the wallet to consider itself unsynced when creating a new address for change while building a transaction.  This creates a race between syncing and the calling code submitting the generated transaction.

### Current Behavior:

Wallet must be synced for the `/push_transactions` endpoint, though it is not required for the singular `/push_tx`.

### New Behavior:

Synced status is not required for the plural transaction push endpoint.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -538,9 +538,6 @@ class WalletRpcApi:
         return {}
 
     async def push_transactions(self, request: Dict) -> EndpointResult:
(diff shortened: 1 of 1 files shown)


emlowe (approved):
This looks fine to me - I don't see where any downstream code requires a synced wallet - in fact all this does is push the txs to the DB for eventual submission to the node

wjblanke (commented):
aok

PR #14983, opened 2023-04-04
Rename singleton helper functions

aqk:
Give the singleton helper functions more specific names.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -61,7 +61,7 @@ from chia.wallet.notification_store import Notification
 from chia.wallet.outer_puzzles import AssetType
 from chia.wallet.puzzle_drivers import PuzzleInfo, Solver
 from chia.wallet.puzzles.p2_delegated_puzzle_or_hidden_puzzle import puzzle_hash_for_synthetic_public_key
-from chia.wallet.singleton import create_singleton_fullpuz
+from chia.wallet.singleton import create_singleton_puzzle
 from chia.wallet.trade_record import TradeRecord
 from chia.wallet.trading.offer import Offer
 from chia.wallet.transaction_record import TransactionRecord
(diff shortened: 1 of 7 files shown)


geoffwalmsley (dismissed):

xdustinface (commented):
What about `create_singleton_puzzle` or at least `create_singleton_full_puzzle` if the `full` is important in here? The `puz` or `puzz` like we have in bunch of places as abbreviation for `puzzle` (like `puz_hash` or `puzz_hash`or `puzhash` or `puzzhash`) looks rather like a typo to me and i don't understand whats wrong with just typing out `puzzle` everywhere to make it obvious and consistent, in general. Im just curious what you think, not complaining about you keeping it :)

Quexington (dismissed):
I agree with @xdustinface but I don't see that this would exacerbate any issues so I don't care either way

aqk:
> What about `create_singleton_puzzle` or at least `create_singleton_full_puzzle` if the `full` is important in here? The `puz` or `puzz` like we have in bunch of places as abbreviation for `puzzle` (like `puz_hash` or `puzz_hash`or `puzhash` or `puzzhash`) looks rather like a typo to me and i don't understand whats wrong with just typing out `puzzle` everywhere to make it obvious and consistent, in general. Im just curious what you think, not complaining about you keeping it :)

I agree - I'll update the name :)

Quexington (approved):

xdustinface (approved):

chia/rpc/wallet_rpc_api.py
```diff
@@ -61,7 +61,7 @@
 from chia.wallet.outer_puzzles import AssetType
 from chia.wallet.puzzle_drivers import PuzzleInfo, Solver
 from chia.wallet.puzzles.p2_delegated_puzzle_or_hidden_puzzle import puzzle_hash_for_synthetic_public_key
-from chia.wallet.singleton import create_fullpuz
+from chia.wallet.singleton import create_singleton_puzzle
```
trepca, line 64: why is this better than `from chia.wallet import singleton` and then use `singleton.create_puzzle`?

Quexington: I think that that's a bad precedent to set unless you're importing something like a `Singleton` class.  Having a bunch of files all define `create_puzzle` seems like asking for confusing.

aqk: It's not necessarily better. However, since the programmer using the singleton module can choose between
`import singleton` and `from singleton import foo`, you can have a lot of `foo` floating around.

And then you'll be in deep foo foo.

emlowe (approved):

altendky:
@trepca, conversations need resolved for this to be merged.  Or, go ahead and raise a concern and work through the rest of whatever discussion is needed.

I personally prefer to use the namespacing the language gives us, but we've very much gone down the other path of throwing that out and putting the names in misc order in the individual variable names.

PR #14988, opened 2023-04-05
optimize Mempool.add_to_pool()

arvidn:
Instead of computing the total cost of items in the pool, compute it once and subtract the cost of the items being removed as we go. Also remove all items in a single call.

This is not a material improvement, but I'm working on a patch where I will add another similar check for items with assert-before conditions. Having this code a bit simpler may be more important once it's duplicated.

@trepca suggested further improvements by moving more logic into sqlite.

# current version

| --- | --- | --- | --- |
| add_spend_bundle() | 6.5029s | 6.3214s | 97.21% |
| add_spend_bundle() with replace-by-fee | 6.4152s | 6.3519s | 99.01% |

# first version

| --- | --- | --- | --- |
| add_spend_bundle() | 6.5029s | 6.4207s | 98.74% |
| add_spend_bundle() with replace-by-fee | 6.4152s | 6.3652s | 99.22% |

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -470,6 +470,9 @@ class MempoolManager:
         if cost == 0:
             return Err.UNKNOWN, None, []
 
+        if cost > self.max_block_clvm_cost:
+            return Err.BLOCK_COST_EXCEEDS_MAX, None, []
+
         # this is not very likely to happen, but it's here to ensure SQLite
         # never runs out of precision in its computation of fees.
         # sqlite's integers are signed int64, so the max value they can
--- a/tests/core/mempool/test_mempool.py
+++ b/tests/core/mempool/test_mempool.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import logging
+import random
 from typing import Callable, Dict, List, Optional, Tuple
 
 import pytest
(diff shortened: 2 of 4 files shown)


AmineKhaldi (dismissed):

arvidn:
I realized I should add a unit test covering this. stay tuned

AmineKhaldi (dismissed):

chia/full_node/mempool_manager.py
```diff
@@ -470,6 +470,9 @@ async def validate_spend_bundle(
         if cost == 0:
             return Err.UNKNOWN, None, []
 
+        if cost > self.max_block_clvm_cost:
+            return Err.BLOCK_COST_EXCEEDS_MAX, None, []
```
arvidn, line 474: I snuck this check in here. maybe I should have separated this out. In practice, we already have this limit, since it's the max block cost. In tests however, it's really convenient to set the block size smaller

tests/core/mempool/test_mempool.py
```diff
@@ -2722,3 +2723,57 @@ def test_spends_by_feerate(items: List[MempoolItem], expected: List[Coin]) -> No
...
+    ret = bytearray(32)
+    for i in range(32):
+        ret[i] = rng.getrandbits(8)
+    return bytes32(ret)
+
+
+def item_cost(cost: int, fee_rate: float) -> MempoolItem:
+    fee = cost * fee_rate
+    amount = int(fee + 100)
+    coin = Coin(rand_hash(), rand_hash(), amount)
```
trepca, line 2739: why not use `secrets.token_bytes(32)`?

arvidn: I copy pasted this function from a place where I wanted deterministic values. I suspect this is a little bit faster too.

AmineKhaldi (approved):

trepca (approved):

tests/fee_estimation/test_fee_estimation_integration.py
```diff
@@ -38,7 +38,7 @@ def make_mempoolitem() -> MempoolItem:
     ph = wallet_tool.get_new_puzzlehash()
     coin = Coin(ph, ph, uint64(10000))
     spend_bundle = wallet_tool.generate_signed_transaction(uint64(10000), ph, coin)
-    cost = uint64(5000000)
+    cost = uint64(1000000)
```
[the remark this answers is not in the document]
wallentx: Marking as resolved

chia/full_node/mempool_manager.py
wallentx: marking as resolved

PR #14990, opened 2023-04-05
Check for directory existence before creating offer

Quexington:
See title


trepca:
What kind of error will this throw?

trepca (approved):

PR #14992, opened 2023-04-06
Add flexibility to `log_exceptions()`

altendky:
### Purpose:

Allows configuration of the exception or exceptions to be caught, whether or not to show the traceback, the log level, and also adds the exception type to aid readability of errors like `KeyError` where the only text is the value of the key requested.

### Draft For:
- [x] add tests


cmmarslender (dismissed):

xdustinface (dismissed):
Looks good but i don't understand why you add this fixtures in tests instead of just making the tests parameterized via `pytest.mark.parametrize`. Maybe im missing some fancy python thing here but the parameterization next to the test seems much simpler and clearer to me. It also doesn't require all this extra methods, ignores and issue links.

xdustinface (approved):

emlowe (approved):

PR #14993, opened 2023-04-06
limit expiring transactions

arvidn:
With the forthcoming `ASSERT_BEFORE_*` conditions, there's a risk that maintaining the mempool becomes expensive, if the whole mempool is allowed to be filled with transactions that will all expire really soon. In this case, only one block worth of transaction will make it in, and the rest expires. In the worst case scenario, this means the whole mempool (50 blocks worth of cost) needs to be re-filled in between every block. Currently only one block worth of transactions need to be re-validated to fill it up (plus more depending on how active the fee-market is).

This change mitigates this problem by introducing a new rule saying that the total cost of all transactions that expires in 15 minutes or less may not exceed one block worth of cost. Within these transactions the ones paying the lowest fee-per-cost will be evicted first. This essentially creates two separate fee markets. One for transactions that expire soon, and one for transactions that do not expire.

### DIFF
--- a/chia/full_node/mempool.py
+++ b/chia/full_node/mempool.py
@@ -19,6 +19,7 @@ from chia.types.mempool_item import MempoolItem
 from chia.types.spend_bundle import SpendBundle
 from chia.util.chunks import chunks
 from chia.util.db_wrapper import SQLITE_MAX_VARIABLE_NUMBER
+from chia.util.errors import Err
 from chia.util.ints import uint32, uint64
 
 log = logging.getLogger(__name__)
@@ -51,9 +52,15 @@ class Mempool:
     # this separate dictionary
     _items: Dict[bytes32, InternalMempoolItem]
 
+    # the most recent block height and timestamp that we know of
+    _block_height: uint32
+    _timestamp: uint64
+
     def __init__(self, mempool_info: MempoolInfo, fee_estimator: FeeEstimatorInterface):
         self._db_conn = sqlite3.connect(":memory:")
         self._items = {}
+        self._block_height = uint32(0)
+        self._timestamp = uint64(0)
 
         with self._db_conn:
             # name means SpendBundle hash
@@ -81,10 +88,8 @@ class Mempool:
             self._db_conn.execute("CREATE INDEX cost_sum ON tx(cost)")
             self._db_conn.execute("CREATE INDEX feerate ON tx(fee_per_cost)")
             self._db_conn.execute(
-                "CREATE INDEX assert_before_height ON tx(assert_before_height) WHERE assert_before_height != NULL"
-            )
-            self._db_conn.execute(
-                "CREATE INDEX assert_before_seconds ON tx(assert_before_seconds) WHERE assert_before_seconds != NULL"
+                "CREATE INDEX assert_before ON tx(assert_before_height, assert_before_seconds) "
+                "WHERE assert_before_height IS NOT NULL OR assert_before_seconds IS NOT NULL"
             )
 
             # This table maps coin IDs to spend bundles hashes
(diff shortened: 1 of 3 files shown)


chia/full_node/mempool.py
```diff
@@ -248,23 +257,68 @@ def remove_from_pool(self, items: List[bytes32], reason: MempoolRemoveReason) ->
...
+                    # there's space for us, stop pruning
+                    if cumulative_cost + item.cost <= self.mempool_info.max_block_clvm_cost:
+                        break
+
+                    # we can't evict any more transactions, abort (and don't
+                    # evict what we put aside in "to_remove".
+                    if fee_per_cost > item.fee_per_cost:
+                        return Err.INVALID_FEE_LOW_FEE
+                    to_remove.append(name)
+                self.remove_from_pool(to_remove, MempoolRemoveReason.POOL_FULL)
```
arvidn, line 303: @aqk does this removal reason make sense? the mempool isn't full, it's just using a stricter limit for a subset of transactions

chia/full_node/mempool.py
```diff
@@ -264,6 +273,44 @@ def add_to_pool(self, item: MempoolItem) -> None:
...
+                cursor = self._db_conn.execute(
+                    """
+                    SELECT name,
+                        fee_per_cost,
+                        SUM(cost) OVER (ORDER BY fee_per_cost DESC, seq ASC) AS cumulative_cost
+                    FROM tx
+                    WHERE assert_before_seconds IS NOT NULL AND assert_before_seconds < ?
+                        OR assert_before_height IS NOT NULL AND assert_before_height < ?
+                    ORDER BY cumulative_cost DESC
+                    """,
```
arvidn, line 295: @trepca please review this query with some care. As far as I can tell (from the tests) the `SUM(cost) OVER ...` field is still constrained by the outer `WHERE` clause, is that right?

trepca: lgtm

trepca (dismissed):

chia/full_node/mempool.py
```diff
@@ -51,11 +52,19 @@ class Mempool:
...
+    _timestamp: uint64
+
     def __init__(self, mempool_info: MempoolInfo, fee_estimator: FeeEstimatorInterface):
         self._db_conn = sqlite3.connect(":memory:")
         self._items = {}
+        self._block_height = uint32(0)
+        self._timestamp = uint64(0)
 
         with self._db_conn:
+            self._db_conn.execute("PRAGMA strict=ON")
```
arvidn, line 66: this pragma doesn't appear to actually exist

chia/full_node/mempool.py
```diff
@@ -81,10 +90,8 @@ def __init__(self, mempool_info: MempoolInfo, fee_estimator: FeeEstimatorInterfa
             self._db_conn.execute("CREATE INDEX cost_sum ON tx(cost)")
             self._db_conn.execute("CREATE INDEX feerate ON tx(fee_per_cost)")
             self._db_conn.execute(
-                "CREATE INDEX assert_before_height ON tx(assert_before_height) WHERE assert_before_height != NULL"
-            )
-            self._db_conn.execute(
-                "CREATE INDEX assert_before_seconds ON tx(assert_before_seconds) WHERE assert_before_seconds != NULL"
+                "CREATE INDEX assert_before ON tx(assert_before_height) "
+                "WHERE assert_before_height IS NOT NULL OR assert_before_seconds IS NOT NULL"
```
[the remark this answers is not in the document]
AmineKhaldi: We query with both `assert_before_height` and `assert_before_seconds` but this seems to be set only on `tx(assert_before_height)`.

justinengland:
please resolve conversation so this can be merged.

arvidn:
I intend to address the comments from @AmineKhaldi

AmineKhaldi (dismissed):

trepca (approved):

AmineKhaldi (approved):

Issue #14957, opened 2023-03-31
[Bug] wallet state manager not assigned


gades:
### What happened?

I can't run wallet on Mac OS  anymore
<img width="1791" alt="Screenshot 2023-03-31 at 6 48 11 PM" src="https://user-images.githubusercontent.com/4200730/229168728-65979072-9ba6-4419-8643-a9a2b16eb8f0.png">

### Version

1.7.1

### What platform are you using?

macOS

### What ui mode are you using?

GUI

### Relevant log output

```shell
tail -f ~/.chia/mainnet/log/debug.log

2023-03-31T18:47:02.705 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia/rpc/rpc_server.py", line 340, in safe_handle
  File "chia/rpc/rpc_server.py", line 331, in ws_api
  File "chia/rpc/wallet_rpc_api.py", line 519, in get_sync_status
  File "chia/wallet/wallet_node.py", line 153, in wallet_state_manager
RuntimeError: wallet state manager not assigned

2023-03-31T18:47:12.715 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia/rpc/rpc_server.py", line 340, in safe_handle
  File "chia/rpc/rpc_server.py", line 331, in ws_api
  File "chia/rpc/wallet_rpc_api.py", line 519, in get_sync_status
  File "chia/wallet/wallet_node.py", line 153, in wallet_state_manager
RuntimeError: wallet state manager not assigned

2023-03-31T18:47:22.728 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "chia/rpc/rpc_server.py", line 340, in safe_handle
  File "chia/rpc/rpc_server.py", line 331, in ws_api
  File "chia/rpc/wallet_rpc_api.py", line 519, in get_sync_status
  File "chia/wallet/wallet_node.py", line 153, in wallet_state_manager
RuntimeError: wallet state manager not assigned
```

gades:
Looks like it's the same issue with https://github.com/Chia-Network/chia-blockchain/issues/14595

MumfMeisterT:
Are there any more errors or warnings in your logs? Could you provide full logs for us? thx

gades:
it's duplicate of https://github.com/Chia-Network/chia-blockchain/issues/14595

PR #14997, opened 2023-04-07
server: Rework config peer resolving and connection handling

### LINKED ISSUE #7648: [BUG] Harvester looks up 'host' just once.
**Describe the bug**

Harvester looks up host just once. If the dynamic IP @ home changes harvester does not reconnect to new IP Address.

```
harvester:
  chia_ssl_ca:
    crt: config/ssl/ca/chia_ca.crt
    key: config/ssl/ca/chia_ca.key
  farmer_peer:
    host: [remote DNS Name with changing IP]
    port: 8447
```

**To Reproduce**

1. Setup Full node with Internet @ home where the IP changes after each reconnect.
2. Setup remote Farmer node to use an external DNS (dyndns) service for changing IPs ex. noip.com
3. Change IP
4. Harvester will still use old IP after hours.

**Expected behavior**
Harvester looks up IP every x Minutes or use ex. gethostbyaddr and notices the change and use the new IP address.

**Desktop**

- OS: Linux
- OS Version/Flavor: Debian
- CPU: AMD Ryzen 9 5950X

xdustinface:
### Purpose:

This PR changes the way we currently handle reconnecting to default peers from the config:
- Don't try to resolve the config peers only once on startup and badly fail if the resolving fails but instead store `UnresolvedPeerInfo` objects in `Service._connect_peers` and have the reconnect task handle the resolving at runtime. This fixes issues where any service currently just fails or gets stuck to start if on startup there is no way to resolve the peers from the config due to missing network or just whatever DNS lookup failure.
- Drop `chia.server.reconnect_task.py` and instead add `Service. _connect_peers_task_handler` which does not longer create 1 task per peer but instead loops over the newly introduced `Service._connect_peers` set which contains the peers, thereby it does the following:
    - Check if the peer entry was resolved already, if not try to resolve it
    - Check if there is a connection to the resolved peer info
    - If there is on connection, try to establish it
    - If the connection attempt fails, try to re-resolve the initial host - This fixes #7648.

Maybe fixes issues like #14957/#14907 but only assumed there is an issue with the wallet peer lookup on startup, its not so clear to me there because some logs seems to be missing.

### Current Behavior:

- Only resolve config peers once on startup and fail if the lookup fails or get stuck if the lookup blocks
- Have 1 reconnect task for each config peer

### New Behavior:

- Resolve and even re-resolve at runtime in the long running tasks which handles reconnecting to config peers.
- Have 1 task which handles all config peers

### Testing 

- Just test that config peers automatically get connected to and reconnected after a disconnect.
- To test the re-resolve case you can just assign some domain to a random node, put it in the wallet config as full node peer, let the service connect to it, assign the domain to some other node, disconnect from the first node and wait until it re-resolves and connects to the new node.


chia/server/start_service.py
```diff
@@ -76,6 +76,7 @@ def __init__(
         self._rpc_close_task: Optional[asyncio.Task[None]] = None
         self._network_id: str = network_id
         self.max_request_body_size = max_request_body_size
+        self.reconnect_retry_seconds: int = 10
```
xchdata1, line 79: Maybe just keep the original 3 seconds `reconnect_task.py` used?

xchdata1:
Strong +1 from me, this change would be very welcome! Not only would this help with dyndns (which is not _that_ uncommon for fullnode/farmer), but it would also allow DNS-based failover and would make round-robin DNS finally work properly. RRDNS is especially useful to have a simple redundant/HA node setup: if one node goes down, farmer/peers reconnect quickly to the other node. (Working RRDNS is a missing piece in an actual HA node setup we run.)

chia/server/start_service.py
```diff
@@ -129,9 +130,44 @@ def __init__(
...
+                        resolved_new = PeerInfo(
+                            get_host_addr(unresolved.host, prefer_ipv6=prefer_ipv6), unresolved.port
+                        )
+                    except Exception as e:
+                        self._log.warning(f"Failed to resolve after connection failure {unresolved.host}: {e}")
+                        continue
+                    if resolved_new != resolved:
+                        self._log.info(f"Host {unresolved.host} changed from {resolved} to {resolved_new}")
+                        resolved_peers[unresolved] = resolved_new
+            await asyncio.sleep(self.reconnect_retry_seconds)
```
arvidn, line 169: if DNS resolution is stalling for some reason, running the loop above might take a lot longer than `reconnect_retry_seconds`. Perhaps it would make sense to subtract the time we spent in the loop from this sleep.

It probably doesn't matter much either way though

chia/server/start_service.py
```diff
@@ -129,9 +130,44 @@ def __init__(
 
         self._on_connect_callback = on_connect_callback
         self._advertised_port = advertised_port
-        self._reconnect_tasks: Dict[PeerInfo, Optional[asyncio.Task[None]]] = {peer: None for peer in connect_peers}
+        self._connect_peers = connect_peers
+        self._connect_peers_task: Optional[asyncio.Task[None]] = None
         self.upnp: UPnP = UPnP()
 
+    async def _connect_peers_task_handler(self) -> None:
+        resolved_peers: Dict[UnresolvedPeerInfo, PeerInfo] = {}
```
arvidn, line 138: this dict seems to grow unbounded. is that OK?
I'm assuming `self._connect_peers` may have churn on peers. but perhaps not.

chia/server/start_service.py
```diff
@@ -129,9 +130,44 @@ def __init__(
...
+        prefer_ipv6 = self.config.get("prefer_ipv6", False)
+        while True:
+            for unresolved in self._connect_peers:
+                resolved = resolved_peers.get(unresolved)
+                if resolved is None:
+                    try:
+                        resolved = PeerInfo(get_host_addr(unresolved.host, prefer_ipv6=prefer_ipv6), unresolved.port)
```
arvidn, line 145: I would have thought `get_host_addr()` would need to be async. Won't this stall the whole process?

chia/server/start_service.py
```diff
@@ -129,9 +130,44 @@ def __init__(
...
+                if any(connection.peer_info == resolved for connection in self._server.all_connections.values()):
+                    continue
+
+                if not await self._server.start_client(resolved, None):
+                    self._log.info(f"Failed to connect to {resolved}")
+                    # Re-resolve to make sure the IP didn't change, this helps for example to keep dyndns hostnames
+                    # up to date.
+                    try:
+                        resolved_new = PeerInfo(
+                            get_host_addr(unresolved.host, prefer_ipv6=prefer_ipv6), unresolved.port
```
arvidn, line 161: also here, I would have expected `await`

arvidn (changes requested):
I agree with this direction. I think the blocking DNS lookups are still a problem though

chia/server/start_service.py
xdustinface: It will only grow as big as `_connect_peers` which will not grow unbounded.

chia/server/start_service.py
xdustinface: Yeah its not about sleeping exactly this seconds. Waiting a bit longer is not a problem.

arvidn (approved):

Issue #14888, opened 2023-03-23
[Bug] wallet startup failure 


0x5BC2E7:
### What happened?

```
2023-03-23T11:52:48.117 wallet wallet                     : INFO     Starting service wallet ...
2023-03-23T11:52:48.118 wallet wallet                     : INFO     chia-blockchain version: 1.7.1
2023-03-23T11:52:48.118 wallet wallet_server              : INFO     Service capabilities: [(1, '1'), (2, '1'), (3, '1')]
2023-03-23T11:55:17.604 daemon chia.daemon.server         : INFO     ConnectionClosed. Closing websocket with ['Unknown']
2023-03-23T11:55:18.608 wallet wallet                     : INFO     Starting service wallet ...
2023-03-23T11:55:18.609 wallet wallet                     : INFO     chia-blockchain version: 1.7.1
2023-03-23T11:55:18.609 wallet wallet_server              : INFO     Service capabilities: [(1, '1'), (2, '1'), (3, '1')]
2023-03-23T11:58:23.987 daemon chia.daemon.server         : INFO     ConnectionClosed. Closing websocket with ['Unknown']
2023-03-23T11:58:24.858 harvester chia.harvester.harvester: INFO     `harvester.plot_loading_frequency_seconds` is deprecated. Consider replacing it with the new section `harvester.plots_refresh_parameter`. See `initial-config.yaml`.
2023-03-23T11:58:24.859 harvester chia.harvester.harvester: INFO     Using plots_refresh_parameter: {'batch_size': 300, 'batch_sleep_milliseconds': 1, 'interval_seconds': 120, 'retry_invalid_seconds': 1200}
2023-03-23T11:58:24.859 harvester harvester               : INFO     Starting service harvester ...
2023-03-23T11:58:24.859 harvester harvester               : INFO     chia-blockchain version: 1.7.1
2023-03-23T11:58:24.860 harvester harvester_server        : INFO     Service capabilities: [(1, '1'), (2, '1'), (3, '1')]
2023-03-23T11:58:24.865 harvester harvester               : INFO     Started harvester service on network_id: mainnet at port 8448
2023-03-23T11:58:24.868 harvester harvester               : INFO     Reconnecting to peer PeerInfo(_ip=IPv4Address('127.0.0.1'), _port=8447)
[...]

PR #14998, opened 2023-04-07
server: Resolve introducer right before the connection attempt

xdustinface:
### Purpose:

Fix issues like #14888 by delaying the introducer lookup.

### Current Behavior:

Lookup the introducer IP on startup and fail the service startup hard due to uncatched exception.

### New Behavior:

Lookup the introducer IP right before the connection attempt and just continue the alternating between dns introducer and introducer connections if it fails because the exception gets handled in the new place where the lookup happens.

### DIFF
--- a/chia/server/node_discovery.py
+++ b/chia/server/node_discovery.py
@@ -22,7 +22,7 @@ from chia.server.outbound_message import Message, NodeType, make_msg
 from chia.server.peer_store_resolver import PeerStoreResolver
 from chia.server.server import ChiaServer
 from chia.server.ws_connection import WSChiaConnection
-from chia.types.peer_info import PeerInfo, TimestampedPeerInfo
+from chia.types.peer_info import PeerInfo, TimestampedPeerInfo, UnresolvedPeerInfo
 from chia.util.hash import std_hash
 from chia.util.ints import uint16, uint64
 from chia.util.network import IPAddress, get_host_addr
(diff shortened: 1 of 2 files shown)


chia/server/node_discovery.py
```diff
@@ -212,7 +206,9 @@ async def on_connect(peer: WSChiaConnection) -> None:
             msg = make_msg(ProtocolMessageTypes.request_peers_introducer, RequestPeersIntroducer())
             await peer.send_message(msg)
 
-        await self.server.start_client(self.introducer_info, on_connect)
+        await self.server.start_client(
+            PeerInfo(get_host_addr(self.introducer_info.host, prefer_ipv6=False), self.introducer_info.port), on_connect
```
arvidn, line 210: this is still a blocking call to `get_host_addr()`. It looks like this is just shifting the problem around a bit without solving the underlying issue. I would expect this call to be async.

xdustinface: See https://github.com/Chia-Network/chia-blockchain/pull/14999

arvidn (approved):

MumfMeisterT:
PR needs a test for a bogus hostname the concern is if an unhandled exception from failed DNS lookup

MumfMeisterT (dismissed):
This needs a test per @wjblanke

xdustinface:
> PR needs a test for a bogus hostname the concern is if an unhandled exception from failed DNS lookup

@wjblanke `FullNodeDiscovery._introducer_client` is called within a by a `try`/`except` so there can't be an unhandled exception. See https://github.com/Chia-Network/chia-blockchain/blob/main/chia/server/node_discovery.py#L290-L437.

wjblanke (approved):
ok for the release branch

PR #14999, opened 2023-04-07
util: Async DNS lookups

xdustinface:
### Purpose:

- Rename `get_host_addr` to `resolve`
- Make it a non-blocking `async` function by leveraging `asyncio.loop.getaddrinfo`

### Current Behavior:

- `get_host_addr` blocks the main thread while doing DNS lookups.

### New Behavior:

The `resolve` lookup can now be `await` ed to let other tasks run in the meantime.

### Testing notes:

The lookups are tested in `test_network.py`.

### Based on:

- #14997 
- #14998

### DIFF
--- a/chia/seeder/crawler.py
+++ b/chia/seeder/crawler.py
@@ -127,7 +127,9 @@ class Crawler:
 
         try:
             connected = await self.create_client(
-                PeerInfo(str(resolve(peer.ip_address, prefer_ipv6=self.config.get("prefer_ipv6", False))), peer.port),
+                PeerInfo(
+                    str(await resolve(peer.ip_address, prefer_ipv6=self.config.get("prefer_ipv6", False))), peer.port
+                ),
                 peer_action,
             )
             if not connected:
--- a/chia/seeder/crawler.py
+++ b/chia/seeder/crawler.py
@@ -23,7 +23,7 @@ from chia.server.server import ChiaServer
 from chia.server.ws_connection import WSChiaConnection
 from chia.types.peer_info import PeerInfo
 from chia.util.ints import uint32, uint64
-from chia.util.network import get_host_addr
+from chia.util.network import resolve
 from chia.util.path import path_from_root
 
 log = logging.getLogger(__name__)
@@ -127,9 +127,7 @@ class Crawler:
 
         try:
             connected = await self.create_client(
-                PeerInfo(
-                    str(get_host_addr(peer.ip_address, prefer_ipv6=self.config.get("prefer_ipv6", False))), peer.port
-                ),
+                PeerInfo(str(resolve(peer.ip_address, prefer_ipv6=self.config.get("prefer_ipv6", False))), peer.port),
                 peer_action,
             )
             if not connected:
--- a/chia/seeder/crawler.py
+++ b/chia/seeder/crawler.py
@@ -127,9 +127,7 @@ class Crawler:
 
         try:
             connected = await self.create_client(
-                PeerInfo(
-                    str(await resolve(peer.ip_address, prefer_ipv6=self.config.get("prefer_ipv6", False))), peer.port
-                ),
+                PeerInfo(await resolve(peer.ip_address, prefer_ipv6=self.config.get("prefer_ipv6", False)), peer.port),
                 peer_action,
             )
             if not connected:
--- a/chia/util/network.py
+++ b/chia/util/network.py
@@ -145,14 +145,14 @@ def class_for_type(type: NodeType) -> Any:
     raise ValueError("No class for type")
 
 
-def resolve(host: str, *, prefer_ipv6: bool = False) -> IPAddress:
+async def resolve(host: str, *, prefer_ipv6: bool = False) -> IPAddress:
     try:
         return IPAddress.create(host)
     except ValueError:
         pass
     addrset: List[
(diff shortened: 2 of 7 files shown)


emlowe:
To me, renaming `get_host_addr` isn't particularly helpful, but I understand making it async would have meant adding `awaits` where it was used anyway. but I would have been happy with keeping the function the same name.

Note that the IPv6 tests always worked for me locally on my Mac - we never did figure out why in CI it didn't work, but in any event it's nice that it does work now.

chia/util/network.py
```diff
@@ -145,30 +147,40 @@ def class_for_type(type: NodeType) -> Any:
...
+    if answer.rrset is None or len(answer.rrset) == 0:
+        raise NetworkError(f"Empty {rd_type.name} answer for {host}")
+    try:
+        return IPAddress.create(str(answer.rrset[0]))
+    except Exception as e:
+        raise NetworkError(f"Invalid {rd_type.name} response for {host}: {answer.rrset[0]}") from e
+
+
+async def resolve(host: str, *, prefer_ipv6: bool = False) -> IPAddress:
+    if host == "localhost":
```
emlowe, line 164: might be useful to add in `ip6-localhost` which is linux (possibly Ubuntu) specific, but is common on those systems. Also on Ubuntu, `localhost` will always resolve to the IPv4 loopback afaict, so this may be a not so subtle change in behaviour

xdustinface:
> To me, renaming `get_host_addr` isn't particularly helpful, but I understand making it async would have meant adding `awaits` where it was used anyway. but I would have been happy with keeping the function the same name.

I don't know, for me `resolve` just sounds more reasonable because it implies what it does behind the scenes - resolving the host you pass in while `get_host_addr` could do something else, like reading it from some database or return it from some map, whatever..  Anyway, if the renaming blocks this PR at the end, i can as well just revert it.

> Note that the IPv6 tests always worked for me locally on my Mac - we never did figure out why in CI it didn't work, but in any event it's nice that it does work now.

It doesn't work for me locally on macOS on current `main`. Not sure why it works for you but anyway, it works with this PR also locally here for me.

emlowe:
I think there will be a change in behaviour on linux systems when setting `prefer_ipv6` but it's likely a positive change. Previously, all the internal services would listen on `localhost` as that is the default config.yaml setting. On Ubuntu, regardless of prefer_ipv6, this would always resolve to the IPv4 loopback. With this change, it will use the IPv6 loopback. Of course, this is what the user likely wants with `prefer_ipv6` set to true, so this seems reasonable

chia/util/network.py
emlowe: Worth testing if this limitation of asyncresolver is specific to localhost , but doesn't impact any other items in the `/etc/hosts` file?

arvidn (dismissed):
there are two places where you convert the `IPAddress` returned by `resolve()` to a `str` before passing it to the `PeerInfo` constructor. That looks suspicious.

tests/util/test_network.py
```diff
@@ -1,42 +1,36 @@
...
+        assert await resolve("localhost", prefer_ipv6=prefer_ipv6) == IPAddress.create("127.0.0.1")
+        assert await resolve("example.net", prefer_ipv6=prefer_ipv6) == IPAddress.create("93.184.216.34")
 
     @pytest.mark.asyncio
-    @pytest.mark.skipif(
-        condition=("GITHUB_ACTIONS" in os.environ) and (sys.platform in {"darwin", "win32"}),
-        reason="macOS and Windows runners in GitHub Actions do not seem to support IPv6",
-    )
-    async def test_get_host_addr6(self):
+    async def test_resolve6(self):
```
emlowe, line 22: Doesn't this need the skipif ?

emlowe:
The description is also incorrect now, since this isn't using `asyncresolver` anymore

tests/util/test_network.py
xdustinface: Yes of course, thanks! I thought i reverted it. But now i did, really :)

chia/seeder/crawler.py
```diff
@@ -127,9 +127,7 @@ async def peer_action(peer: WSChiaConnection):
 
         try:
             connected = await self.create_client(
-                PeerInfo(
-                    str(get_host_addr(peer.ip_address, prefer_ipv6=self.config.get("prefer_ipv6", False))), peer.port
-                ),
+                PeerInfo(str(resolve(peer.ip_address, prefer_ipv6=self.config.get("prefer_ipv6", False))), peer.port),
```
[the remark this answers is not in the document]
xdustinface: I just retained the behaviour but yeah, its not needed and there was one more place.. i changed it, see [1b18653](https://github.com/Chia-Network/chia-blockchain/pull/14999/commits/1b18653b14dd389971173e5f0c4b0b09c92d3e8d).

xdustinface:
> I'm not sure this is really important for 1.8.0?

@emlowe I think we should add it yes because we now have blocking lookups in long running tasks due to https://github.com/Chia-Network/chia-blockchain/pull/14997 and https://github.com/Chia-Network/chia-blockchain/pull/14998 being in the release, which might be problematic in some situations.

arvidn (approved):

wjblanke (approved):
aok

PR #15001, opened 2023-04-07
correct broken exit codes from `manage_clvm.py`

altendky:

### DIFF
--- a/tools/manage_clvm.py
+++ b/tools/manage_clvm.py
@@ -296,7 +296,7 @@ def check(use_cache: bool) -> int:
         with cache_path.open(mode="w") as file:
             dump_cache(cache=cache, file=file)
 
-    return 1 if overall_fail else 0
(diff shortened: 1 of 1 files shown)


Quexington (dismissed):

tools/manage_clvm.py
```diff
@@ -344,7 +344,7 @@ def build() -> int:
         if file_fail:
             overall_fail = True
 
-    return 1 if overall_fail else 0
+    sys.exit(1 if overall_fail else 0)
 
 
 sys.exit(main(auto_envvar_prefix="CHIA_MANAGE_CLVM"))
```
Quexington, line 350:
```suggestion
main(auto_envvar_prefix="CHIA_MANAGE_CLVM")
```
Do we want to remove this also?

Quexington (approved):

emlowe (approved):

PR #15005, opened 2023-04-09
Add Chia Simulator tools directly Into Chia Blockchain 

jack60612:
add simulator tools natively + tests

### Purpose:
We need this for improved testing

### Current Behavior:
No Simulator tools in chia blockchain repo

### New Behavior:
Simulator Tools natively in chia blockchain repo.

### Testing Notes:
Full Simulator based cli tests are added, the majority of all code changes were previously approved in various dev tools prs.

### DIFF
--- a/chia/cmds/cmds_util.py
+++ b/chia/cmds/cmds_util.py
@@ -15,6 +15,7 @@ from chia.rpc.full_node_rpc_client import FullNodeRpcClient
 from chia.rpc.harvester_rpc_client import HarvesterRpcClient
 from chia.rpc.rpc_client import RpcClient
 from chia.rpc.wallet_rpc_client import WalletRpcClient
+from chia.simulator.simulator_full_node_rpc_client import SimulatorFullNodeRpcClient
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.mempool_submission_status import MempoolSubmissionStatus
 from chia.util.config import load_config
@@ -29,6 +30,7 @@ NODE_TYPES: Dict[str, Type[RpcClient]] = {
     "full_node": FullNodeRpcClient,
     "harvester": HarvesterRpcClient,
     "data_layer": DataLayerRpcClient,
+    "simulator": SimulatorFullNodeRpcClient,
 }
 
 node_config_section_names: Dict[Type[RpcClient], str] = {
@@ -37,6 +39,7 @@ node_config_section_names: Dict[Type[RpcClient], str] = {
     FullNodeRpcClient: "full_node",
     HarvesterRpcClient: "harvester",
     DataLayerRpcClient: "data_layer",
+    SimulatorFullNodeRpcClient: "full_node",
 }
 
 
@@ -59,6 +62,7 @@ async def validate_client_connection(
     root_path: Path,
     fingerprint: Optional[int],
     login_to_wallet: bool,
+    catch_errors: bool = True,
 ) -> Optional[int]:
     try:
         await rpc_client.healthz()
@@ -67,9 +71,12 @@ async def validate_client_connection(
             if fingerprint is None:
                 rpc_client.close()
     except ClientConnectorError:
+        rpc_client.close()
+        if not catch_errors:
+            await rpc_client.await_closed()
+            raise
         print(f"Connection error. Check if {node_type.replace('_', ' ')} rpc is running at {rpc_port}")
         print(f"This is normal if {node_type.replace('_', ' ')} is still starting up")
-        rpc_client.close()
     await rpc_client.await_closed()  # if close is not already called this does nothing
     return fingerprint
 
@@ -81,6 +88,7 @@ async def get_any_service_client(
     root_path: Path = DEFAULT_ROOT_PATH,
     fingerprint: Optional[int] = None,
     login_to_wallet: bool = True,
+    catch_errors: bool = True,
 ) -> AsyncIterator[Tuple[Optional[_T_RpcClient], Dict[str, Any], Optional[int]]]:
     """
     Yields a tuple with a RpcClient for the applicable node type a dictionary of the node's configuration,
@@ -103,13 +111,15 @@ async def get_any_service_client(
         # check if we can connect to node, and if we can then validate
         # fingerprint access, otherwise return fingerprint and shutdown client
         fingerprint = await validate_client_connection(
-            node_client, node_type, rpc_port, root_path, fingerprint, login_to_wallet
+            node_client, node_type, rpc_port, root_path, fingerprint, login_to_wallet, catch_errors
         )
         if node_client.session.closed:
             yield None, config, fingerprint
         else:
             yield node_client, config, fingerprint
     except Exception as e:  # this is only here to make the errors more user-friendly.
+        if not catch_errors:
+            raise
         print(f"Exception from '{node_type}' {e}:\n{traceback.format_exc()}")
 
     finally:
--- a/chia/cmds/cmds_util.py
+++ b/chia/cmds/cmds_util.py
@@ -62,7 +62,7 @@ async def validate_client_connection(
     root_path: Path,
     fingerprint: Optional[int],
     login_to_wallet: bool,
-    catch_errors: bool = True,
+    consume_errors: bool = True,
 ) -> Optional[int]:
     try:
         await rpc_client.healthz()
@@ -72,7 +72,7 @@ async def validate_client_connection(
                 rpc_client.close()
     except ClientConnectorError:
         rpc_client.close()
-        if not catch_errors:
+        if not consume_errors:
             await rpc_client.await_closed()
             raise
         print(f"Connection error. Check if {node_type.replace('_', ' ')} rpc is running at {rpc_port}")
@@ -88,7 +88,7 @@ async def get_any_service_client(
     root_path: Path = DEFAULT_ROOT_PATH,
     fingerprint: Optional[int] = None,
     login_to_wallet: bool = True,
-    catch_errors: bool = True,
+    consume_errors: bool = True,
 ) -> AsyncIterator[Tuple[Optional[_T_RpcClient], Dict[str, Any], Optional[int]]]:
     """
     Yields a tuple with a RpcClient for the applicable node type a dictionary of the node's configuration,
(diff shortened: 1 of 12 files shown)


Quexington (dismissed):
Not an expert on this simulator but it seems fine

altendky (changes requested):
Do the test plots belong here or in https://github.com/Chia-Network/test-cache?  Or maybe what's already there can just be used?

chia/cmds/start_funcs.py
```diff
@@ -59,7 +59,7 @@ async def create_start_daemon_connection(root_path: Path, config: Dict[str, Any]
     return None
 
 
-async def async_start(root_path: Path, config: Dict[str, Any], group: str, restart: bool) -> None:
+async def async_start(root_path: Path, config: Dict[str, Any], restart: bool, group: tuple) -> None:
```
altendky, line 62: Is the re-order important?  I wonder if the hinting corrections should be submitted separately.

.github/test-plots/plot-k19-0edb21491fbd8c79aadb5364df04bbb2858d9bfa2ee22c0c2e7da1473db47352.plot
arvidn, line 1: it doesn't look right that we would check-in the test plots. The unit tests generate these on-demand, right?

chia/cmds/cmds_util.py
```diff
@@ -59,6 +62,7 @@ async def validate_client_connection(
     root_path: Path,
     fingerprint: Optional[int],
     login_to_wallet: bool,
+    catch_errors: bool = True,
```
arvidn, line 65: this parameter name is not intuitive to me. `True` means that we'll *ignore* errors and just log them (and not shut down gracefully). Note that we *catch* the error either way.

It seems the default should be to shut down gracefully (i.e. call `await rpc_client.await_closed()`).

I'm guessing there's some subtle context I'm unfamiliar with, making ungraceful shutdown making sense.

.github/workflows/test-single.yml
```diff
@@ -210,6 +210,12 @@ jobs:
           mkdir ${{ github.workspace }}/.chia
           mv ${{ github.workspace }}/test-cache-${{ env.BLOCKS_AND_PLOTS_VERSION }}/* ${{ github.workspace }}/.chia
 
+      - name: Copy pre-generated plots for CLI / Simulator Tests
+        run: |
+          mkdir ${{ github.workspace }}/.chia/simulator
+          mkdir ${{ github.workspace }}/.chia/simulator/plots
+          cp ${{ github.workspace }}/.github/test-plots/* ${{ github.workspace }}/.chia/simulator/plots
```
arvidn, line 217: I see, so it looks like it's deliberate. I think it's risky to check-in large files in git. They'll stick around in the repo forever. I think we should look for an alternative.

I would propose a new command that creates the test plots and puts them in the right place. That way the simulator could be used by people installing the chia package as well, and not just ones using it from the git repo.

chia/cmds/start.py
```diff
@@ -20,4 +20,4 @@ def start_cmd(ctx: click.Context, restart: bool, group: str) -> None:
     root_path = ctx.obj["root_path"]
     config = load_config(root_path, "config.yaml")
     warn_if_beta_enabled(config)
-    asyncio.run(async_start(root_path, config, group, restart))
+    asyncio.run(async_start(root_path, config, restart, group))
```
arvidn, line 23: what happened here? this looks like an unrelated bugfix

chia/cmds/sim_funcs.py
```diff
@@ -0,0 +1,516 @@
...
+        automated_testing=False,
+        plot_dir=config["simulator"].get("plot_directory", "plots"),
+        keychain=Keychain(),
+    )
+    await bt.setup_keys(fingerprint=fingerprint, reward_ph=farming_puzzle_hash)
+    starting_time = time()
+    await bt.setup_plots(
+        num_og_plots=PLOTS, num_pool_plots=0, num_non_keychain_plots=0, plot_size=PLOT_SIZE, bitfield=bitfield
+    )
+    print(f"{'New plots generated.' if time() - starting_time > 5 else 'Using Existing Plots'}\n")
```
arvidn, line 255: heh, if I have a machine so fast it can generate test plots in under 5 seconds, it will look like I'm "Using existing plots" :)

Anyway, I see we have a function to generate these plots already. I think the CI should use this as well. Since we pick up existing blocks that may exist already, I believe the `.chia/test_plots` directory is already stored and restored on CI so we wouldn't have to generate them every run

chia/cmds/start_funcs.py
arvidn: I agree that we need a strong justification for reordering these. It looks unncecessary.
Also, I seem to recall the `group` parameter had a more precise tuple type in a different place. Like `Tuple[str, ...]`

jack60612: > Is the re-order important? I wonder if the hinting corrections should be submitted separately.

yeah i can undo the re-order

chia/cmds/start.py
jack60612: unrelated, ill undo.

chia/cmds/sim.py
```diff
@@ -0,0 +1,218 @@
...
+    config = load_config(ctx.obj["root_path"], "config.yaml")
+    group: Any = ("simulator",)
+    if wallet:
+        group += ("wallet",)
+    sys.exit(asyncio.run(async_stop(ctx.obj["root_path"], config, group, daemon)))
+
+
+@sim_cmd.command("status", short_help="Get information about the state of the simulator.")
+@click.option("-f", "--fingerprint", type=int, help="Get detailed information on this fingerprint.")
+@click.option("-k", "--show_key", is_flag=True, help="Show detailed key information.")
```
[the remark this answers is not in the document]
jack60612: did not realize i used an underscore, ill make those changes too.

.github/test-plots/plot-k19-0edb21491fbd8c79aadb5364df04bbb2858d9bfa2ee22c0c2e7da1473db47352.plot
arvidn: it seems like the `test-cache` repo is a better place for them

altendky:
Are the plots used for testing or at runtime when a user is making use of the simulator CLI?

jack60612:
> Are the plots used for testing or at runtime when a user is making use of the simulator CLI?

only testing.

jack60612:
The ci run takes only a minute longer without the plots, so we might as well test that too.

.github/workflows/test-single.yml
jack60612: test plots removed, might as well test the creation. (ci 1 min longer)

chia/cmds/sim_funcs.py
jack60612: fixed

chia/cmds/cmds_util.py
jack60612: good suggestion, ill do that.

chia/cmds/cmds_util.py
```diff
@@ -103,13 +111,15 @@ async def get_any_service_client(
...
-            node_client, node_type, rpc_port, root_path, fingerprint, login_to_wallet
+            node_client, node_type, rpc_port, root_path, fingerprint, login_to_wallet, consume_errors
         )
         if node_client.session.closed:
             yield None, config, fingerprint
         else:
             yield node_client, config, fingerprint
     except Exception as e:  # this is only here to make the errors more user-friendly.
+        if not consume_errors:
+            raise
```
altendky, line 122: Should this be after the exception?  I'm not entirely sure, but logging here does provide the extra context of the node type.

chia/cmds/dev.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import annotations
+
+import click
+
+from chia.cmds.sim import sim_cmd
+
+
+@click.group("dev", short_help="Developer commands and tools")
```
altendky, line 8: Yeah, what you've got here is consistent, but I'm in the middle of redefining the consistency.  https://github.com/Chia-Network/chia-blockchain/pull/15076  This applies to any other places you don't specify a more complete `help=`.

```suggestion
@click.group("dev", help="Developer commands and tools")
```

chia/cmds/sim.py
```diff
@@ -0,0 +1,216 @@
...
+    ),
+    type=int,
+    default=None,
+)
+@click.option(
+    "--root-path", default=SIMULATOR_ROOT_PATH, help="Simulator root folder.", type=click.Path(), show_default=True
+)
+@click.option(
+    "-n",
+    "--simulator_name",
```
altendky, line 29: Did we want to clean up all the `_` or no?  (various other cases)

chia/simulator/block_tools.py
```diff
@@ -358,26 +358,34 @@ async def setup_plots(
...
-    ):
+    ) -> bool:
         self.add_plot_directory(self.plot_dir)
         assert self.created_plots == 0
+        existing_plots: bool = True
         # OG Plots
         for i in range(num_og_plots):
-            await self.new_plot(plot_size=plot_size, bitfield=bitfield)
+            _, plot_exists = await self.new_plot(plot_size=plot_size, bitfield=bitfield)
+            if existing_plots and not plot_exists:
```
altendky, line 368: I think resetting to `False` again is fine and simplifies the pattern.  A couple repetitions below.

```suggestion
            if not plot_exists:
```

wallentx:
I think we'd like to get this in for release

chia/simulator/block_tools.py
```diff
@@ -388,7 +396,7 @@ async def new_plot(
         exclude_plots: bool = False,
         plot_size: int = 20,
         bitfield: bool = True,
-    ) -> Optional[bytes32]:
+    ) -> tuple[bytes32, bool]:
```
altendky, line 399: Add a dataclass instead of introducing tupleness.

chia/cmds/cmds_util.py
```diff
@@ -67,9 +71,12 @@ async def validate_client_connection(
             if fingerprint is None:
                 rpc_client.close()
     except ClientConnectorError:
+        rpc_client.close()
+        if not consume_errors:
+            await rpc_client.await_closed()
```
altendky, line 76: This feels awkward, even though I can read around and figure out the pattern.  What about a `finally:` clause and just a single line calling this?  And do we want to close it for all exceptions?

altendky (changes requested):

chia/cmds/cmds_util.py
jack60612: maybe i can remove this and only have the one finally

jack60612: I am going to still close it on execpt

chia/cmds/cmds_util.py
jack60612: nope, we skip this as its for the cli.

chia/cmds/sim.py
jack60612: meant to, srry

chia/simulator/block_tools.py
jack60612: fair point, not sure why i added that.

chia/simulator/block_tools.py
jack60612: its only used in 2 places, is this necessary?

jack60612: i mean i might as well, not worth the discussion.

jack60612:
Force Pushed to remove plots from history as requested.

altendky (approved):

emlowe (approved):

PR #15008, opened 2023-04-10
Add Chialisp pretty printer to pre-commit

Quexington:
The pretty printer is very un-opinionated, mostly fixing tabs and spaces to be uniform across all files.

Also, reviewers should not that no hex files or clvm hashes changed, proof that the compilations for all programs remain the same.  If you hide whitespace in reviewing, you'll see all of the relevant changes.


prozacchiwawa (approved):
neat, tried it out.

trepca (approved):
💯

PR #15012, opened 2023-04-10
Clean create tandem xch tx

ytx1991:
### Purpose:
It's unnecessary to have the same create_tandem_xch_tx method in different wallets.

### Current Behavior:
Each wallet its own create_tandem_xch_tx

### New Behavior:
Share the create_tandem_xch_tx in the standard wallet, unless the wallet needs to override, like CAT wallet.


xdustinface (approved):
There is more to cleanup to be done here, we can also combine this with the CAT/DL wallets but given the clawback wants to go in, lets keep this as future improvement :)

trepca (approved):

PR #15013, opened 2023-04-10
Add new coin_type & metadata to coin_records

ytx1991:
### Purpose:
To support new types of wallet features such as Clawback, CRCAT, etc. We need to save additional data for a single coin.

### Current Behavior:
None

### New Behavior:
Save metadata and coin type for coins

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -70,7 +70,7 @@ from chia.wallet.util.address_type import AddressType, is_valid_address
 from chia.wallet.util.compute_hints import compute_coin_hints
 from chia.wallet.util.compute_memos import compute_memos
 from chia.wallet.util.transaction_type import TransactionType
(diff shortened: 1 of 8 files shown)


xdustinface (dismissed):
Looks good, there are some nits you maybe still want to address and also maybe adding the default values to the new `WalletCoinRecord` fields would make sense too but i'll leave it up to you :)

emlowe:
Any benefit to storing CoinType.CLAWBACK in their own table instead of adding a column to the existing table?

ytx1991:
> Any benefit to storing CoinType.CLAWBACK in their own table instead of adding a column to the existing table?

It was in a new table but there are many redundant codes so we just merge the tables. The main reason is we will have more coin types in the future and we want a unified mechanism to handle the detection and spend update.
More details in https://github.com/Chia-Network/chia-blockchain/pull/14348

xdustinface (approved):

trepca (approved):

PR #15020, opened 2023-04-11
Return fee estimates as integer mojos

aqk:
### Current Behavior:

The "estimates" field in the return of the `get_fee_estimate` API contains an array of mojos. Currently, they can be fractional.

### New Behavior:

This change ensures that the estimates are integers, not floats.


paninaro (approved):
lgtm

PR #15024, opened 2023-04-12
merge `listen` into `advertised_port` as `None`

altendky:
### Purpose:

When creating a `ChiaServer` you specify a port and then when you `.start()` it you also tell it whether to listen or not.  This seems redundant.  If there's a port to listen on, listen.  If not, don't.

This is in support of https://github.com/Chia-Network/chia-blockchain/pull/14922 which removes unused ports from the config file for the timelord, wallet, and harvester which do not actually listen on those ports.


tests/timelord/config.py
```diff
@@ -0,0 +1,4 @@
+# flake8: noqa: E501
```
xdustinface, line 1:
```suggestion
```

Why?

xdustinface (changes requested):

xdustinface (dismissed):

altendky:
Missing coverage is in benchmark and skipped tests.

- https://coveralls.io/builds/61199986
- https://coveralls.io/builds/61199986/source?filename=tests%2Fcore%2Ffull_node%2Ffull_sync%2Ftest_full_sync.py#L392
- https://coveralls.io/builds/61199986/source?filename=tests%2Fcore%2Fmempool%2Ftest_mempool_performance.py#L60
- https://coveralls.io/builds/61199986/source?filename=tests%2Fwallet%2Fdb_wallet%2Ftest_dl_wallet.py#L343

chia/server/start_service.py
```diff
@@ -190,11 +188,13 @@ async def start(self) -> None:
...
         await self._server.start(
-            listen=self._listen,
             prefer_ipv6=self.config.get("prefer_ipv6", False),
             on_connect=self._on_connect_callback,
         )
-        self._advertised_port = self._server.get_port()
+        try:
+            self._advertised_port = self._server.get_port()
+        except ValueError:
+            pass
```
arvidn, line 197: in this case, we'll just leave the class not having the `_advertised_port` field at all? I would have expected we'd want to at least set it to `None`

arvidn (dismissed):

chia/server/start_service.py
altendky: It looks like it is set unconditionally in `.__init__()`.  I think this is just handling the case where we had requested `0` and now we ought to be filling in the actual port we opened.  I think...

wallentx: Ping me if it is decided that no action is needed on this

altendky:
coverage diff failure on a skipped test and a benchmark

chia/server/server.py
```diff
@@ -467,11 +466,17 @@ async def start_client(
                 self.log.info(f"Connected to a node with the same peer ID, disconnecting: {target_node} {peer_id}")
                 return False
 
+            server_port: uint16
+            try:
+                server_port = self.get_port()
+            except ValueError:
+                server_port = uint16(0)
```
emlowe, line 473: thinking about this. I guess in the code today, I will perform the handshake and say I'm on port XYZ, but I may not be actually listening at all on the port, or using it in at all (is listen is false when the server started).

emlowe (approved):
I think this makes sense

PR #15029, opened 2023-04-12
New endpoints /add_store_id and /remove_store_id

emlowe:
Add new endpoints `/add_store_id` and `/remove_store_id` that add and remove stores for this plugin

Modifies the configuration yaml file as needed


emlowe:
opps messed this up

PR #15030, opened 2023-04-12
New S3 plugin endpoints for add_store_id and remove_store_id

emlowe:
`/add_store_id` endpoint. Takes "store_id" and either "bucket" or "urls" or both `{"id": "7acfcbd1ed73bfe2b698508f4ea5ed353c60ace154360272ce91f9ab0c8423c3", "bucket": "foo", "urls":["s3://one", "s3://two"]}`

either a bucket or a list of urls is required.

calling `/add_store_id` for a store_id already in the config returns an error - there is no update option, only add and delete.

`/remove_store_id` endpoint. Takes "store_id" input and removes this store from the DL plugin
`{"store_id": "7acfcbd1ed73bfe2b698508f4ea5ed353c60ace154360272ce91f9ab0c8423c3"}`

calling `/remove_store_id` for a store_id that is not in the config returns success.


chia/data_layer/s3_plugin_service.py
```diff
@@ -17,6 +20,8 @@
 from chia.types.blockchain_format.sized_bytes import bytes32
 
 log = logging.getLogger(__name__)
+plugin_id = "2bdf9ed6-f3bc-44f6-a69d-06602c0688d9"
```
almogdepaz, line 23: if we want an id + a version we might want to set the id in the config on first instantiation or something so different instances would have different ids

altendky (approved):
plenty to clean up, i don't see actual errors here though

paninaro (approved):

PR #15035, opened 2023-04-13
lower MAX_FUTURE_TIME to 2 minutes with the 1.8.0 soft-fork activation

arvidn:
To reduce the ability for farmers to manipulate block timestamps. To be conservative, the new `MAX_FUTURE_TIME` takes effect at the 1.8 soft fork activation.

Based on a ~50 hour sample of mainnet, **this would have rejected ~0.32% of unfinished blocks** because they had a timestamp too far into the future.
The maximum positive offset measured during this time (the unfinished block with a timestamp the farthest into the future) was 264 seconds.
94.84% of all samples were negative, i.e. the timestamp of the unfinished blocks were in the past. This is expected for farmers with correct clocks, since the propagation and validation delay will take some time.

The distribution of time offsets looked like this:
blockchain-timestamp-offsets-cdf

To zoom in on the interesting window, where the time offsets are higher than expected:
blockchain-timestamp-offsets-cdf-zoom

### DIFF
--- a/chia/consensus/block_header_validation.py
+++ b/chia/consensus/block_header_validation.py
@@ -815,7 +815,11 @@ def validate_unfinished_header_block(
                 return None, ValidationError(Err.INVALID_TRANSACTIONS_FILTER_HASH)
 
         # 26a. The timestamp in Foliage Block must not be over 5 minutes in the future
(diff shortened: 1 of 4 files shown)


bramcohen (dismissed):
Policy of 2 minutes time kicking it at a fixed height seems reasonable. I haven't reviewed the code

emlowe (approved):
Reads good to me

PR #15039, opened 2023-04-14
Introduce Verifiable Credentials Primitive

Quexington:
This PR introduces Chialisp for the Verifiable Credentials primitive and the relevant drivers, wallet code, and RPCs for it.

It also introduces the Chialisp and drivers for the Credential Restricted Assets primitive, but no wallet code or RPCs.  The CR-CATs serve as a useful tool for testing the full lifecycle of VCs which is why they are also included here.  Wallet support will come in a separate PR.


Quexington:
Closing in favor of #15059

PR #15040, opened 2023-04-14
Fix wallet type in sign_message_by_id

ytx1991:
### Purpose:

Fix the issue that cannot find the existing DID wallet because of unmatch of WalletType and its value.
Notice: There are many places that have the same issue. We need to change all of them. There are majorly in the wallet_rpc_api and wallet_state_manager.

### Current Behavior:
wallet.type() == WalletType.DECENTRALIZED_ID.value

### New Behavior:
wallet.type() == WalletType.DECENTRALIZED_ID

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1312,7 +1312,7 @@ class WalletRpcApi:
             is_hex = bool(is_hex)
         if is_valid_address(request["id"], {AddressType.DID}, self.service.config):
             for wallet in self.service.wallet_state_manager.wallets.values():
-                if wallet.type() == WalletType.DECENTRALIZED_ID.value:
+                if wallet.type() == WalletType.DECENTRALIZED_ID:
                     assert isinstance(wallet, DIDWallet)
                     assert wallet.did_info.origin_coin is not None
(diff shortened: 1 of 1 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1329,7 +1329,7 @@ async def sign_message_by_id(self, request) -> EndpointResult:
         elif is_valid_address(request["id"], {AddressType.NFT}, self.service.config):
             target_nft: Optional[NFTCoinInfo] = None
             for wallet in self.service.wallet_state_manager.wallets.values():
-                if wallet.type() == WalletType.NFT.value:
+                if wallet.type() == WalletType.NFT:
```
trepca, line 1332: we should add a test for this

trepca (changes requested):

xdustinface (changes requested):
How does this fix something? It doesn't matter if you compare `.value` here or not since its a `IntEnum` and `WalletProtocol.type` returns a `WalletType`.

```python
WalletType.DECENTRALIZED_ID == WalletType.DECENTRALIZED_ID.value
Out[19]: True
WalletType.DECENTRALIZED_ID == WalletType.DECENTRALIZED_ID
Out[20]: True
```

PR #15046, opened 2023-04-16
server: Try to convert in `known_active_capabilities`

xdustinface:
### Purpose:

Cleanup.

### New Behavior:

No change in behaviour.


altendky (approved):
Both options are bad.  I don't super care which we use in this case since it doesn't need to be high performance.

arvidn (approved):

PR #15051, opened 2023-04-17
wallet: Select enough XCH to cover the fee when creating an offer

xdustinface:
### Purpose:

Just cleanup / remove some extra logic from `Wallet._generate_unsigned_transaction` which was added in #14812 to get enough XCH amount to also pay the fee when creating the XCH side of an offer.  Just select enough coins during offer creation to also cover the fee instead of selecting extra coins for the fee in `Wallet._generate_unsigned_transaction`.

I think we should either pass in coins into this method or not. If not, we select what we need and if we pass coins in it should be enough to cover everything.

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/wallet/trade_manager.py
+++ b/chia/wallet/trade_manager.py
@@ -551,9 +551,8 @@ class TradeManager:
 
             all_transactions: List[TransactionRecord] = []
             fee_left_to_pay: uint64 = fee
-            # TODO: Create the transaction in place above when we select the coins? The access of the sorted keys here
-            #       makes sure we create the XCH transaction first to make sure we pay fee with the XCH side of the
-            #       offer and don't create an extra fee transaction in other wallets.
+            # The access of the sorted keys here makes sure we create the XCH transaction first to make sure we pay fee
+            # with the XCH side of the offer and don't create an extra fee transaction in other wallets.
             for id in sorted(list(coins_to_offer.keys())):
                 selected_coins = coins_to_offer[id]
                 if isinstance(id, int):
--- a/chia/wallet/trade_manager.py
+++ b/chia/wallet/trade_manager.py
@@ -553,7 +553,7 @@ class TradeManager:
             fee_left_to_pay: uint64 = fee
             # The access of the sorted keys here makes sure we create the XCH transaction first to make sure we pay fee
             # with the XCH side of the offer and don't create an extra fee transaction in other wallets.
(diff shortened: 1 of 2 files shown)


chia/wallet/trade_manager.py
```diff
@@ -507,8 +507,12 @@ async def _create_offer_for_ids(
                         wallet = await self.wallet_state_manager.get_wallet_for_asset_id(asset_id.hex())
                     if not callable(getattr(wallet, "get_coins_to_offer", None)):  # ATTENTION: new wallets
                         raise ValueError(f"Cannot offer coins from wallet id {wallet.id()}")
+                    # For the XCH wallet also include the fee amount to the coins we use to pay this offer
+                    get_coin_amount = abs(amount)
+                    if wallet.type() == WalletType.STANDARD_WALLET:
+                        get_coin_amount += fee
                     coins_to_offer[id] = await wallet.get_coins_to_offer(
-                        asset_id, uint64(abs(amount)), min_coin_amount, max_coin_amount
+                        asset_id, uint64(get_coin_amount), min_coin_amount, max_coin_amount
```
trepca, line 515: `get_coin_amount` sounds like a function, maybe rename to `total_coin_amount_with_fee` or something similar?

xdustinface: What about 72d90e4a5411e8e8383e8149d97c7634d9c74315?

chia/wallet/trade_manager.py
```diff
@@ -547,7 +551,11 @@ async def _create_offer_for_ids(
 
             all_transactions: List[TransactionRecord] = []
             fee_left_to_pay: uint64 = fee
-            for id, selected_coins in coins_to_offer.items():
+            # TODO: Create the transaction in place above when we select the coins? The access of the sorted keys here
+            #       makes sure we create the XCH transaction first to make sure we pay fee with the XCH side of the
+            #       offer and don't create an extra fee transaction in other wallets.
```
Quexington, line 556: You're not going to be able to create the transaction above.  Each transaction needs to assert an announcement that contains a hash of all the coins used in the entire offer.  This is why we must select the coins first -> calculate the bundle announcement -> create the transactions.  That part of this comment (assuming I read it correctly) should be deleted.

Quexington (changes requested):

Quexington (approved):

trepca (approved):
thanks!

PR #15053, opened 2023-04-17
util: Streamable enums based on sized ints and `Enum`

xdustinface:
### Purpose:

Make it possible to use enums in streamable classes. The way it works is that inheriting from the actual sized int adds `__bytes__` and `from_bytes` and with this, the enums are already supported in streamable while inheriting `Enum` adds enum features and forces the entries to be from the actual sized int.

There is only the move of the `int` conversion in `recurse_jsonify` required (see 1f2f5145f01dc7b2e2950621766855a931344114 ) to fully support it because we there currently return the name for normal `Enum` types. I checked the places where we call `recurse_jsonify` and it should not have an unintended impact on other JSON conversions. 

**Note:** This can not be just be swapped now retrospectively without further changes in all places and requires some more adjustments and considerations in some cases because swapping the type to an enum leads to deserialisation failure instead of the conversion failures we currently have when trying to convert the already deserialised value to the actual enum type.

### Example

```python
class TheEnum(UInt8Enum):
    a = 0
    b = 1
```

in the class

```python
@streamale
@dataclass(frozen=True)
class Example(Streamable)
    member: TheEnum
```

which had to look like 

```python
@streamale
@dataclass(frozen=True)
class Example(Streamable)
    member: uint8  # TheEnum
```

before this PR which leads to a bunch of conversations all over the place and just ambiguous usages of class members/values.

### New Behavior:

No change in behaviour.  It just adds support for it.

### Testing Notes:

I added unit tests in this PR.

### DIFF
--- a/.github/workflows/test-single.yml
+++ b/.github/workflows/test-single.yml
@@ -63,7 +63,6 @@ jobs:
             matrix: '3.7'
             exclude_from:
               limited: True
-              main: True
           - name: '3.8'
             file_name: '3.8'
             action: '3.8'
@@ -72,7 +71,6 @@ jobs:
             matrix: '3.8'
             exclude_from:
               limited: True
-              main: True
           - name: '3.9'
             file_name: '3.9'
             action: '3.9'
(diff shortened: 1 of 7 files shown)


xdustinface:
Back to draft..  `ReprEnum` was introduced in 3.11 :) 

https://docs.python.org/3/library/enum.html#enum.ReprEnum

PR #15057, opened 2023-04-17
Added some sanity checks to upload and download for S3 plugin

emlowe:
Add required config for `server_files_location` where the plugin will always download and upload to. This will need to be set to the same localtion as the `server_files_location` for datalayer in the chia config. At this time, this is a manual configuration

Added checks for the S3 plugin checks that all the files follow the Datalayer naming convention

Datalayer now only sends the filenames themselves for `/upload` and the plugin expects to find these files at the configured `server_files_location`

### DIFF
--- a/chia/data_layer/s3_plugin_service.py
+++ b/chia/data_layer/s3_plugin_service.py
@@ -149,7 +149,7 @@ class S3Plugin:
             parse_result = urlparse(url)
             should_download = False
             for store in self.stores:
-                if store.id == filename_tree_id and parse_result.scheme == "s3" and data["url"] in store.urls:
+                if store.id == filename_tree_id and parse_result.scheme == "s3" and url in store.urls:
                     should_download = True
 
             # filename must follow the DataLayer naming convention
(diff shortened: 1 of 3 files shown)


chia/data_layer/s3_plugin_service.py
```diff
@@ -123,13 +142,25 @@ async def download(self, request: web.Request) -> web.Response:
...
-            client_folder = Path(data["client_folder"])
             filename = data["filename"]
+            # Pull the store_id from the filename to make sure we only download for configured stores
+            filename_tree_id = bytes32(bytes.fromhex(filename.split("-")[0]))
+
             parse_result = urlparse(url)
+            should_download = False
+            for store in self.stores:
+                if store.id == filename_tree_id and parse_result.scheme == "s3" and url in store.urls:
+                    should_download = True
```
altendky, line 153:
```suggestion
                    should_download = True
                    break
```

altendky (changes requested):

altendky (approved):

paninaro (approved):

PR #15058, opened 2023-04-17
Simplify MempoolItem's cost property

AmineKhaldi:


chia/types/mempool_item.py
```diff
@@ -44,8 +44,7 @@ def name(self) -> bytes32:
 
     @property
     def cost(self) -> uint64:
-        assert self.npc_result.conds is not None
-        return uint64(self.npc_result.conds.cost)
```
arvidn, line 48: I'm hoping we can remove `NPCResult` entirely, long term. I think `SpendBundleConditions` covers all the non-failure cases right now.

AmineKhaldi: Until then, we're constantly constructing a uint64 from an int when we have this value currently as a ready to use uint64. This property is accessed quite frequently.

arvidn (approved):

PR #12917, opened 2022-08-11
cmds: Key label support for CLI - `chia keys label`

xdustinface:
This PR implements key label support for the CLI. For that it adds a new `chia keys label` subcommand: 

```
chia keys label -h
Usage: chia keys label [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --help  Show this message and exit.

Commands:
  delete  Delete the label of a key
  set     Set the label of a key
  show    Show the labels of all available keys
```

And it also adds a new parameter to `chia keys add` and `chia keys generate`:

```
-l, --label TEXT      Enter the label for the key
```

Based on:
- #12883

PR #15059, opened 2023-04-17
Introduce Verifiable Credentials Primitive

Quexington:
Recreation of #15039

This PR introduces Chialisp for the Verifiable Credentials primitive and the relevant drivers, wallet code, and RPCs for it.

Verifiable Credentials are singletons that can be imbued with metadata from a DID which is verifiable by other coins as having been put there by the DID.  The most common use case of this would be for a KYC provider to put in proofs like whether or not someone has a valid driver's license etc.  Another coin like a CR-CAT (chialisp for which is also included in this PR) can assert an announcement from a VC with that example proof in order to check that the person trading it does in fact have a valid driver's license.

It also introduces the Chialisp and drivers for the Credential Restricted Assets primitive, but no wallet code or RPCs. The CR-CATs serve as a useful tool for testing the full lifecycle of VCs which is why they are also included here. Wallet support will come in a separate PR.


chia/rpc/wallet_rpc_api.py
```diff
@@ -3254,3 +3264,174 @@ async def dl_delete_mirror(self, request) -> Dict:
...
             "transactions": [tx.to_json_dict_convenience(self.service.config) for tx in txs],
         }
+
+    ##########################################################################################
+    # Verified Credential
+    ##########################################################################################
+    async def vc_mint_vc(self, request) -> Dict:
+        """
+        Mint a verified credential using the assigned DID
+        :param request:
```
trepca, line 3274: lets document what parameters we require here

chia/rpc/wallet_rpc_api.py
```diff
@@ -3254,3 +3264,174 @@ async def dl_delete_mirror(self, request) -> Dict:
...
+            await self.service.wallet_state_manager.add_pending_transaction(tx)
+        return {
+            "vc_record": vc_record.to_json_dict(),
+            "transactions": [tx.to_json_dict_convenience(self.service.config) for tx in tx_list],
+        }
+
+    async def vc_get_vc(self, request) -> Dict:
+        """
+        Given a launcher ID get the verified credential
+        :param request:
```
trepca, line 3304: param docs

chia/rpc/wallet_rpc_api.py
```diff
@@ -3254,3 +3264,174 @@ async def dl_delete_mirror(self, request) -> Dict:
...
+        """
+        vc_record = await self.service.wallet_state_manager.vc_store.get_vc_record(
+            bytes32.from_hexstr(request["vc_id"])
+        )
+        return {"vc_record": vc_record}
+
+    async def vc_get_vc_list(self, request) -> Dict:
+        """
+        Get a list of verified credentials
+        :param request:
```
trepca, line 3315: param docs

chia/rpc/wallet_rpc_api.py
```diff
@@ -3254,3 +3264,174 @@ async def dl_delete_mirror(self, request) -> Dict:
...
+                for fetched_proof in (
+                    await self.service.wallet_state_manager.vc_store.get_proofs_for_root(rec.vc.proof_hash),
+                )
+            },
+        }
+
+    async def vc_spend_vc(self, request) -> Dict:
+        """
+        Spend a verified credential
+        :param request:
```
trepca, line 3336: param docs

chia/rpc/wallet_rpc_api.py
```diff
@@ -3254,3 +3264,174 @@ async def dl_delete_mirror(self, request) -> Dict:
...
+        )
+        for tx in txs:
+            await self.service.wallet_state_manager.add_pending_transaction(tx)
+
+        return {
+            "transactions": [tx.to_json_dict_convenience(self.service.config) for tx in txs],
+        }
+
+    async def add_vc_proofs(self, request) -> Dict:
+        # get VC wallet
```
trepca, line 3379: docs

chia/rpc/wallet_rpc_api.py
```diff
@@ -3254,3 +3264,174 @@ async def dl_delete_mirror(self, request) -> Dict:
...
+            vc_wallet = await VCWallet.create_new_vc_wallet(
+                self.service.wallet_state_manager, self.service.wallet_state_manager.main_wallet
+            )
+
+        await vc_wallet.store.add_vc_proofs(VCProofs(request["proofs"]))
+
+        return {}
+
+    async def get_proofs_for_root(self, request) -> Dict:
+        # get VC wallet
```
trepca, line 3396: docs

chia/rpc/wallet_rpc_api.py
```diff
@@ -3254,3 +3264,174 @@ async def dl_delete_mirror(self, request) -> Dict:
...
+            vc_wallet = await VCWallet.create_new_vc_wallet(
+                self.service.wallet_state_manager, self.service.wallet_state_manager.main_wallet
+            )
+
+        vc_proofs: Optional[VCProofs] = await vc_wallet.store.get_proofs_for_root(bytes32.from_hexstr(request["root"]))
+        if vc_proofs is None:
+            raise ValueError("no proofs found for specified root")
+        return {"proofs": vc_proofs.key_value_pairs}
+
+    async def vc_revoke_vc(self, request) -> Dict:
```
trepca, line 3413: docs

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -794,6 +796,12 @@ async def create_message_spend(
             ],
             puzzle_announcements=puzzle_announcements,
             coin_announcements=coin_announcements,
+            coin_announcements_to_assert={a.name() for a in coin_announcements_to_assert}
```
trepca, line 799: why do we require announcement object if we only need .name()? why not just bytes? same for puzzle ann

chia/wallet/vc_wallet/cr_cat_drivers.py
```diff
@@ -0,0 +1,591 @@
+from __future__ import annotations
```
trepca, line 1: I thought CR-CATs are coming in a separate PR?

chia/wallet/vc_wallet/vc_wallet.py
```diff
@@ -0,0 +1,460 @@
...
+                    found_did = True
+                    break
+        if not found_did:
+            raise ValueError(f"You don't own the DID {provider_did.hex()}")
+        # Mint VC
+        coins = await self.standard_wallet.select_coins(uint64(2 + fee), min_coin_amount=uint64(2 + fee))
+        if len(coins) == 0:
+            raise ValueError("Cannot find a coin to mint the verified credential.")
+        if inner_puzzle_hash is None:
+            inner_puzzle_hash = await self.standard_wallet.get_puzzle_hash(new=False)
```
trepca, line 150: does it always need to use the old puzzle hash?

chia/wallet/vc_wallet/vc_wallet.py
```diff
@@ -0,0 +1,460 @@
...
+        original_coin = coins.copy().pop()
```
trepca, line 151: why not `coins[0]`?

chia/wallet/vc_wallet/vc_wallet.py
```diff
@@ -0,0 +1,460 @@
...
+        for _, wallet in self.wallet_state_manager.wallets.items():
+            if wallet.type() == WalletType.DECENTRALIZED_ID:
+                assert isinstance(wallet, DIDWallet)
+                if bytes32.fromhex(wallet.get_my_DID()) == provider_did:
+                    found_did = True
+                    break
+        if not found_did:
+            raise ValueError(f"You don't own the DID {provider_did.hex()}")
+        # Mint VC
+        coins = await self.standard_wallet.select_coins(uint64(2 + fee), min_coin_amount=uint64(2 + fee))
```
trepca, line 146: why 2?

chia/wallet/vc_wallet/vc_wallet.py
```diff
@@ -0,0 +1,460 @@
...
+                fee, Announcement(vc_record.vc.coin.name(), announcement_to_make), reuse_puzhash=reuse_puzhash
+            )
+            if coin_announcements is None:
+                coin_announcements = set((announcement_to_make,))
+            else:
+                coin_announcements.add(announcement_to_make)
+        else:
+            chia_tx = None
+        if new_proof_hash is not None:
+            if provider_inner_puzhash is None:
```
trepca, line 247: can the provider update proofs even if they don't control the inner puzzle?

chia/wallet/vc_wallet/vc_wallet.py
```diff
@@ -0,0 +1,460 @@
...
+
+        # Generate spend specific nonce
+        coins = await did_wallet.select_coins(uint64(1))
+        assert coins is not None
+        coins.add(vc.coin)
+        if fee > 0:
+            coins.update(await self.standard_wallet.select_coins(fee))
+        sorted_coins: List[Coin] = sorted(coins, key=Coin.name)
+        sorted_coin_list: List[List[Union[bytes32, uint64]]] = [coin_as_list(c) for c in sorted_coins]
+        nonce: bytes32 = Program.to(sorted_coin_list).get_tree_hash()
```
trepca, line 354: why are we generating nonce via coin selection? also, why do we run separate select coin for a fee? given we're not ignoring previously selected coins, won't we get the same coins back potentially?

chia/wallet/vc_wallet/vc_wallet.py
```diff
@@ -0,0 +1,460 @@
...
+
+    async def select_coins(
+        self,
+        amount: uint64,
+        exclude: Optional[List[Coin]] = None,
+        min_coin_amount: Optional[uint64] = None,
+        max_coin_amount: Optional[uint64] = None,
+        excluded_coin_amounts: Optional[List[uint64]] = None,
+    ) -> Set[Coin]:
+        raise RuntimeError("NFTWallet does not support select_coins()")
```
trepca, line 426: VCWallet

tests/wallet/vc_wallet/test_vc_wallet.py
```diff
@@ -0,0 +1,104 @@
...
+from chia.rpc.wallet_rpc_client import WalletRpcClient
+from chia.simulator.full_node_simulator import FullNodeSimulator
+from chia.simulator.time_out_assert import time_out_assert_not_none
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.types.peer_info import PeerInfo
+from chia.util.ints import uint16, uint64
+from chia.wallet.did_wallet.did_wallet import DIDWallet
+from chia.wallet.util.wallet_types import WalletType
+from chia.wallet.vc_wallet.vc_store import VCProofs, VCRecord
+
```
trepca, line 16: we need negative tests, here are some:
 - try to mint a VC with wrong DID 
 - try to revoke a VC with wrong DID 
 - try to update proofs on a VC after it was transferred to the user 
 - try to prove proofs that are not in the VC 
 - try to update proofs without having a valid DID 
 - try to spend a VC's backdoor with invalid DID 
 - try to prove a tx with someone else's VC

chia/wallet/wallet_state_manager.py
```diff
@@ -1655,9 +1700,10 @@ async def create_wallet_for_puzzle_info(self, puzzle_driver: PuzzleInfo, name=No
                 name,
             )
 
-    async def add_new_wallet(self, wallet: WalletProtocol) -> None:
+    async def add_new_wallet(self, wallet: WalletProtocol, create_puzhash: bool = True) -> None:
         self.wallets[wallet.id()] = wallet
-        await self.create_more_puzzle_hashes()
+        if create_puzhash:
```
xdustinface, line 1705: This is already handled via `WalletProtocol.require_derivation_paths`

chia/wallet/vc_wallet/vc_store.py
```diff
@@ -0,0 +1,238 @@
...
+
+    db_wrapper: DBWrapper2
+
+    @classmethod
+    async def create(cls: Type[_T_VCStore], db_wrapper: DBWrapper2) -> _T_VCStore:
+        self = cls()
+
+        self.db_wrapper = db_wrapper
+
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
```
xdustinface, line 96: Do you think we could store all this new information in `coin_records.metadata` and add `CoinType.VERIFIABLE_CREDENTIAL` (#15013) to treat them as normal coins not add another redundant copy of the coins to the database and keep the mappings we want to query for or whatever is required in cache?

chia/rpc/wallet_rpc_api.py
```diff
@@ -3254,3 +3264,174 @@ async def dl_delete_mirror(self, request) -> Dict:
...
+        if "new_puzhash" in request and request["new_puzhash"] is not None:
+            new_puzhash = bytes32.from_hexstr(request["new_puzhash"])
+
+        new_proof_hash: Optional[bytes32] = None
+        if "new_proof_hash" in request and request["new_proof_hash"] is not None:
+            new_proof_hash = bytes32.from_hexstr(request["new_proof_hash"])
+
+        provider_inner_puzhash: Optional[bytes32] = None
+        if "provider_inner_puzhash" in request and request["provider_inner_puzhash"] is not None:
+            provider_inner_puzhash = bytes32.from_hexstr(request["provider_inner_puzhash"])
```
xdustinface, line 3350: This applies to all RPC endpoints.. 

- Don't you think a `vc` prefix is enough and we don't need also a postfix? Like `vc_get_vc` -> `vc_get` or `vc_get_vc_list` -> `vc_get_list` and `vc_spend_vc` -> `vc_spend`
- If you specify the request as streamable schema like below you can avoid repeated checks and manual parsing
```python
@streamable
class VCSpend(Streamable):
    vc_id: Optional[bytes32] = None
    new_proof_hash: Optional[bytes32] = None
    provider_inner_puzhash: Optional[bytes32] = None
    fee: uint64 = uint64(0)
    reuse_puzhash: bool = False
```

and you can just do 

```python
vs_spend = VCSpend.from_json_dict(request)
```

chia/wallet/vc_wallet/vc_wallet.py
```diff
@@ -0,0 +1,460 @@
...
+            chia_tx: TransactionRecord = await self.create_tandem_xch_tx(fee, vc_announcement, reuse_puzhash)
+            assert tx.spend_bundle is not None
+            assert chia_tx.spend_bundle is not None
+            tx = dataclasses.replace(tx, spend_bundle=SpendBundle.aggregate([chia_tx.spend_bundle, tx.spend_bundle]))
+            chia_tx = dataclasses.replace(chia_tx, spend_bundle=None)
+            return [tx, chia_tx]
+        else:
+            return [tx]
+
+    async def create_tandem_xch_tx(
```
xdustinface, line 392: This is probably not needed after #15012 ?

chia/wallet/vc_wallet/vc_wallet.py
```diff
@@ -0,0 +1,460 @@
...
+from chia.wallet.wallet_coin_record import WalletCoinRecord
+from chia.wallet.wallet_info import WalletInfo
+
+_T_VCWallet = TypeVar("_T_VCWallet", bound="VCWallet")
+
+
+class VCWallet:
+    # WalletStateManager is only imported for type hinting thus leaving pylint
+    # unable to process this
+    wallet_state_manager: Any  # pylint: disable=used-before-assignment
```
xdustinface, line 39: The comment doesn't seem to be correct and the `Any` can probably be replaced after #14900 if its not possible already?

xdustinface (dismissed):
There should be more tests overall, the CLI/RPC interfaces aren't tested at all and you can probably factor out some changes here into separate PRs. I didn't look at it in detail but you i guess you can at least think about it. Also maybe things like RPC/CLI should also just come in separate PRs... so you could add the feature's support along with its tests and puzzles and in follow up PRs add the (tested) RPC/CLI.

Also left some comments in the code and i think in general for PRs like this it also would make sense to squash everything and at least spit it up a bit into separate more focused commits before PRing it because having 116 commits involving your progress to come to the final result and a bunch of catchup merged in here is not really helping in any way because (imo) no one will go though all of them while having few commits like:

- add puzzles
- add vc primitives and basics
- vc integration
- add rpc
- add cli

would also make reviewing things easier which then comes back again to maybe just having separate PRs for each of those commits :) 

Also it would be nice if you could add some words about this feature to the PR description or maybe link to some docs where its written down..

chia/wallet/vc_wallet/cr_cat_drivers.py
Quexington: In the PR description I explained that the CR-CAT Chialisp is coming in at the same time as the VC Chialisp because they are an important case to test for VCs.

chia/rpc/wallet_rpc_api.py
Quexington: added

chia/rpc/wallet_rpc_api.py
Quexington: added

chia/rpc/wallet_rpc_api.py
Quexington: added

chia/wallet/did_wallet/did_wallet.py
Quexington: I agree, I think this is something introduced a while ago, but it happens to be the standard for the API and I think we should prioritize consistency with the other transaction APIs to be fixed in a later refactor.

chia/wallet/vc_wallet/vc_wallet.py
Quexington: I don't think there's much of an argument for privacy here, it's very clear that what you're doing is launching a VC and very likely to yourself.  Not generating a new puzzle hash here just keeps the puzzle DB a little slimmer.

chia/wallet/vc_wallet/vc_wallet.py
Quexington: coins is a set so that doesn't work

chia/wallet/vc_wallet/vc_wallet.py
Quexington: Not sure, fixed

chia/wallet/vc_wallet/vc_wallet.py
Quexington: > why are we generating nonce via coin selection
This is an anti replay measure.  We want the announcement that we assert for bundle cohesion to have a nonce specific to the spend we are creating so that if another overlapping spend is created they cannot both be pushed just due to a single announcement.  Rather, only one spend can go through because they are both depending on a unique announcement and the VC can only make one of those two.

> also, why do we run separate select coin for a fee
These coins are just being selected for nonce calculation, so they're not being used in any way.  That being said, this should be the same exact call that `create_tandem_xch_tx` makes internally and will thus select the coins that we used to calculate the nonce for use.

chia/wallet/vc_wallet/vc_wallet.py
Quexington: Whoops fixed

chia/wallet/wallet_state_manager.py
Quexington: noted, removed

chia/wallet/vc_wallet/vc_store.py
Quexington: I'm not a huge fan of this idea.  We've had problems in the past with trying to store a bunch of information concatenated together as a blob.  I'd be worried about performance, _and_ we also lose the flexibility for calls like `get_vcs_by_providers` because you'd have to deserialize the metadata to look at that information.

chia/wallet/vc_wallet/vc_wallet.py
Quexington: fixed

chia/wallet/vc_wallet/vc_wallet.py
Quexington: fixed

Quexington:
> the CLI/RPC interfaces aren't tested at all

CLIs are never tested (don't necessarily agree with this of course, but this PR is not the place for setting up the infrastructure to do that), but the RPCs are tested through the wallet test.  It's a bit of a different pattern than we normally use, but the wallet test is not calling wallet functions directly, but instead testing them through the RPC which adds coverage to the RPC as well.

> Also maybe things like RPC/CLI should also just come in separate PRs

I disagree with this because adding a bunch of wallet code with no way to use it is quite useless. We would never accept the wallet code unless we were planning on adding endpoints to use it (this would just be code bloat), so these concepts are inseparably linked.  Wallet makes no sense without RPCs and of course RPCs need a wallet to function. (Of course this is all a mater of opinion)

chia/wallet/wallet_state_manager.py
```diff
@@ -1657,7 +1702,6 @@ async def create_wallet_for_puzzle_info(self, puzzle_driver: PuzzleInfo, name=No
 
     async def add_new_wallet(self, wallet: WalletProtocol) -> None:
         self.wallets[wallet.id()] = wallet
-        await self.create_more_puzzle_hashes()
```
xdustinface, line 1660: This shouldn't be removed, the method still needs to be called. The comment was only about the new parameter to skip calling this method.

xdustinface (commented):
> CLIs are never tested

This is not true. There are not a lot yet, but there are some CLI tests, see:
- https://github.com/Chia-Network/chia-blockchain/tree/main/tests/core/cmds
- https://github.com/Chia-Network/chia-blockchain/tree/main/tests/cmds

Sorry that this ranting happens in your PR now but the fact that we in general accept the statement "CLIs are never tested so lets continue to not test them" to be ok is..  sad. So yeah, the reasons why CLIs are not tested in general are probably us wanting to get features released yesterday rather than properly tested, carelessness, ignorance or laziness. I really don't want to offend you personally or be super annoying here now and im definitely not perfect in this either although i indeed added tests for the last few CLI additions i brought in ;) It's more generally speaking because i think at some point we need to wake up and stop just closing our eyes for it (or maybe just get kyle to fix the coverage forcing haha). Anyway, its just sad that its obviously _only_ CLI which at the end is still what the community is confronted with and ends up eating their time to report and our time to investigate and fix the next GitHub issues.

> (don't necessarily agree with this of course, but this PR is not the place for setting up the infrastructure to do that) 

So, since you don't agree and there is no need to setup any infrastructure for this, how do you feel about removing CLI entirely from this PR, because its imo not required to be in here, and bring it up as follow up PR including tests? :)
 
> but the RPCs are tested through the wallet test. It's a bit of a different pattern than we normally use, but the wallet test is not calling wallet functions directly, but instead testing them through the RPC which adds coverage to the RPC as well.

Its not a good pattern because you don't really test any limits/edge cases i guess and it makes it very hard to get confidence when reviewing that the API is working properly because its usage is scattered around tests. In my mind there should ideally always be independent tests for the API endpoints success and failure cases.

> I disagree with this because adding a bunch of wallet code with no way to use it is quite useless. We would never accept the wallet code unless we were planning on adding endpoints to use it (this would just be code bloat), so these concepts are inseparably linked. Wallet makes no sense without RPCs and of course RPCs need a wallet to function.

Sure this all is linked together but it doesn't need to go in together. A wallet makes not a lot sense without RPC/CLI, true - in production. But it still can exist, be supported in the code and work properly while it's even used, without RPC - by tests.

By splitting it up you are forced to not leverage RPCs to tests the new feature but instead write proper, independent tests. Like i said, having the tested basic feature go in first, then independently the tested RPC / CLI is possible and at least in my opinion the much better way than pushing it through in one huge PR, because by the nature of the process you are required to think more about it, add more tests and it makes reviewing the code and tests easier which helps a lot to uncover more oddities which all yes, takes you more time during development but at the end it helps everyone.. our developers, other developers, the community, the world! :)

Its a smaller feature of course but one example where i did the same is https://github.com/Chia-Network/chia-blockchain/pull/12843, https://github.com/Chia-Network/chia-blockchain/pull/12883, https://github.com/Chia-Network/chia-blockchain/pull/12917.

I didn't look at the tests a lot yet really and i might be wrong but from only looking at the added LOC production/test ratio this seems not so much tested to me. This PR adds 2800 lines of production code, 800 lines of chia lisp, and 850 lines in `tests`. Its certainly tested in some ways but yeah, there are most likely bugs which would be uncovered by more tests.

>  (Of course this is all a mater of opinion)

And i know i already said it few times, here and there, and i might be annoying with my poking, sorry for that :) I don't think its only a matter of opinion only though.. reviewing so much code in one PR is just not effective at all and it:
- Takes the reviewer a **lot** time to understand and properly review it with each iteration and given the size of the PR one iteration will not be enough. Its much more efficient and easier for the reviewers to review this in smaller parts and focused topics.
- Tempts the reviewer to be more sloppy about the review.
- Makes it hard to even find reviewer who want to review it.
- Makes the PR confusing and unclear with tons of comments and conversations after two or more peopled reviewed few iterations because its a lot to address for a PR size like this usually.
  
And again, yes, bringing up smaller, focused and independent PRs takes you as developer more time in the first place but it certainly helps to get your code merged faster and at the end you are the one who wants to get the code in and in my mind for developer it should be a goal to make it as easy as possible for your reviewers to review your code and don't go the easy way and throw it all on the reviewers instead.

chia/wallet/vc_wallet/vc_wallet.py
trepca: ok, but still no need for `copy`

chia/rpc/wallet_rpc_api.py
Quexington: suggestion taken

chia/wallet/wallet_state_manager.py
Quexington: whoops, misunderstood the diff, fixing.

chia/wallet/vc_wallet/vc_wallet.py
Quexington: true, fixed

tests/wallet/vc_wallet/test_vc_wallet.py
Quexington: I believe I have addressed all of these now (I implemented most of the new ones in the chialisp lifecycle tests because it is more flexible with passing in wrong values etc.)

The exception is "try to update proofs on a VC after it was transferred to the user" because this is not something we support and I would expect such a test to fail.

trepca (approved):

chia/wallet/vc_wallet/vc_store.py
```diff
@@ -0,0 +1,239 @@
...
+            bytes32.from_hexstr(row[0]),
+            bytes32.from_hexstr(row[7]),
+            bytes32.from_hexstr(row[8]),
+            None if row[9] is None else bytes32.from_hexstr(row[9]),
+        ),
+        uint32(row[10]),
+    )
+
+
+class VCStore:
```
trepca, line 83: @arvidn noticed we're missing unit tests for these which I missed.
We need unit tests for:
 - VCStore
 - VCProofs
 - VerifiedCredential

PR #15062, opened 2023-04-17
DL plugin Add missing files

almogdepaz:
### Purpose:
add_missing_files endpoint for plugin


chia/data_layer/s3_plugin_service.py
```diff
@@ -160,20 +162,48 @@ async def download(self, request: web.Request) -> web.Response:
...
+                        with concurrent.futures.ThreadPoolExecutor() as pool:
+                            await asyncio.get_running_loop().run_in_executor(
+                                pool,
+                                functools.partial(my_bucket.upload_file, file_path, file_path.name),
+                            )
+            except ClientError as e:
+                log.error(f"failed uploading file to aws {e}")
+                return web.json_response({"uploaded": False})
+        except Exception as e:
+            log.error(f"failed handling request {request} {e}")
```
emlowe, line 203: `{type(e).__name__} {e}` is a better pattern since it prints out the name of the Exception (like KeyError, ValueError) etc. TILFK (Things I Learned From Kyle)

emlowe (dismissed):

chia/data_layer/data_layer.py
```diff
@@ -482,9 +483,24 @@ async def add_missing_files(self, store_id: bytes32, overwrite: bool, foldername
...
+        if uploaders is not None and len(uploaders) > 0:
+            request_json = {"store_id": store_id.hex(), "files": json.dumps(files)}
+            for uploader in uploaders:
+                async with aiohttp.ClientSession() as session:
+                    async with session.post(uploader + "/add_missing_files", json=request_json) as response:
+                        res_json = await response.json()
+                        if not res_json["uploaded"]:
+                            self.log.error(f"failed to upload to uploader {uploader}")
+                        else:
+                            self.log.debug(f"uploaded to uploader {uploader}")
```
altendky, line 503: If the intent is to hit all the uploaders, should we handle errors and continue?

chia/data_layer/s3_plugin_service.py
```diff
@@ -160,20 +163,65 @@ async def download(self, request: web.Request) -> web.Response:
...
+    async def add_missing_files(self, request: web.Request) -> web.Response:
+        try:
+            data = await request.json()
+            store_id = bytes32.from_hexstr(data["store_id"])
+            bucket_str = self.get_bucket(store_id)
+            files = json.loads(data["files"])
+            my_bucket = self.boto_resource.Bucket(bucket_str)
+            existing_file_list = []
+            for my_bucket_object in my_bucket.objects.all():
+                existing_file_list.append(my_bucket_object.key)
```
altendky, line 190: I personally like the comprehension better, but using a set is the functional change.

```suggestion
            existing_files = {my_bucket_object.key for my_bucket_object in my_bucket.objects.all()}
```

almogdepaz: i think its less readable

chia/data_layer/data_layer.py
almogdepaz: how do you suggest we do that besides just logging it and letting the user run this again ?

i think a retry mechanism is a bad idea

chia/data_layer/s3_plugin_service.py
almogdepaz: isn't the name of the exception printed anyway when you dump e ?

i dont mind changing if you insist but theres specific log lines for each exception case i dont think this information is interesting to add

altendky: The string conversion of the exception is just the message, not the type.  In many cases this is sufficient but in other cases you get silly stuff like `3`, such as for a `KeyError`.

This case is specifically `except Exception as e:` so I'm not sure we can say there is an exception for each case.

As a detail, I suggested retaining the `:` but Earle overlooked that.  `:]`  `f"our whatever message: {type(e).__name__}: {e}`.  But also, this topic needs general discussion resulting in a generally encouraged form and if we don't just use `logging.exception()` then we probably ought to make our own helper to aid in clarity, consistency, and any future changes.  Certainly, doing this is a future task, not something for this PR to wait for.

chia/data_layer/data_layer.py
emlowe: Not retry, I think Kyle is asking about handling exceptions. Right now if an uploader fails to upload, the error is logged and the for loop will go to the next one. However, if the uploader is not running, the post will raise ClientError exception and the next uploader not be tried

emlowe: `upload_files` and `add_missing_files` are consistent in this, exceptions break the loop while other errors do not.
I would say the most common exception is an uploader not running so this will raise a ClientError and break out of the loop.

emlowe: Per internal discussion, created GH issue to track this: https://github.com/Chia-Network/chia-blockchain/issues/15081
and resolving conversation

emlowe (approved):

PR #15069, opened 2023-04-18
Implement get_items_by_coin_ids() for the mempool

AmineKhaldi:


chia/full_node/mempool.py
```diff
@@ -374,6 +374,17 @@ def at_full_capacity(self, cost: int) -> bool:
 
         return self.total_mempool_cost() + cost > self.mempool_info.max_size_in_cost
 
+    def get_items_spending_coin_ids(self, spent_coin_ids: List[bytes32]) -> List[MempoolItem]:
```
arvidn, line 377: this is a bulk version of the existing `get_spend_by_coin_id()`. I think it would be good if this function was located adjacent to that one, and if the naming would be more consistent. Perhaps: `get_spends_by_coin_ids()`, or something like that

AmineKhaldi: I went for #15103 and `get_items_by_coin_ids`.

arvidn (approved):

PR #15080, opened 2023-04-19
fail for under 100% diff coverage

altendky:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -147,7 +147,7 @@ jobs:
         env:
           compare-branch: ${{ github.base_ref == '' && github.event.before || format('origin/{0}', github.base_ref) }}
         run: |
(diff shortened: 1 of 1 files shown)


xdustinface (approved):
Let's try! :)

AmineKhaldi (approved):
How come CI isn't failing on this?

arvidn:
I take it it only counts coverage of new or changed lines of code

arvidn (approved):

arvidn:
how would one access the test coverage report?

altendky:
Yes, this should fail if any of the lines changed by the PR are not covered in the test runs.

The diff coverage output is available at the end of the test workflow run summary page.  Scroll down at https://github.com/Chia-Network/chia-blockchain/actions/runs/4745370430 for example.  Yes, it's not great.  The HTML output is better than the markdown, but then I need a place to present the HTML.  If you go to the coverage job there https://github.com/Chia-Network/chia-blockchain/actions/runs/4745370430/jobs/8428586350 you can look at the coverage report steps for the console output.

altendky:
We could certainly require 100% coverage overall, but then we wouldn't get a PR merged for a couple months maybe.  And...  it would be impossibly huge.

PR #15094, opened 2023-04-20
Add timelord_launcher to the pyinstaller.spec

cmmarslender:
Adds the timelord launcher to the installers. Still requires vdf_client to be provided via some other means, but allows more flexibility in running the timelord


wjblanke (approved):
aok w/caveat that vdf_client still must be compiled

PR #15099, opened 2023-04-21
wallet: Improve `WalletCoinStore.get_coin_records`

xdustinface:
### Purpose:

Make it more generalised and give it more parameters to allow fetching coin records in various ways. This:

- Is used in #15100 to integrate it into an RPC command `/get_coin_records` which allows for variable queries so that we don't need to add yet another `get_coin_by_*` RPC command if we want to fetch coins by another attribute. Instead with this its possible already get better filtered results with one command.
- Will allow for a bunch of cleanups in the `WalletCoinStore` and the wallet codebase overall which i will do in a follow up PR.  

The parameter set of the method looks like below now:

```python
async def get_coin_records(
    *,
    offset: uint32 = uint32(0),
    limit: uint32 = uint32(uint32.MAXIMUM_EXCLUSIVE - 1),
    wallet_id: Optional[uint32] = None,
    wallet_type: Optional[WalletType] = None,
    coin_type: Optional[CoinType] = None,
    coin_id_filter: Optional[HashFilter] = None,
    puzzle_hash_filter: Optional[HashFilter] = None,
    parent_coin_id_filter: Optional[HashFilter] = None,
    amount_filter: Optional[AmountFilter] = None,
    amount_range: Optional[UInt64Range] = None,
    confirmed_range: Optional[UInt32Range] = None,
    spent_range: Optional[UInt32Range] = None,
    order: CoinRecordOrder = CoinRecordOrder.confirmed_height,
    reverse: bool = False,
    include_total_count: bool = False,
) -> GetCoinRecordsResult:
```

Where `HashFilter` and `AmountFilter` are there to either include or exclude a given list of hashes/amounts and `UInt32Range` and `UInt64Range` range to query entries between two values.

Actually, the method could look like:

```python
async def get_coin_records(request: GetCoinRecords) -> GetCoinRecordsResult:
```

also but this would require to wrap the arguments in `GetCoinRecords` for every call site so i decided to just duplicate the parameters.

### New Behavior:

No change in behaviour.

### Testing Notes:

I added tests with test cases for every parameter, some for mixed parameter sets and tests for the introduced caching for the total amount queries. 

The tests are split up by parameter because i'ts easier to understand which one failed if one fails. Im open for suggestions to maybe group them by name in one paramaterization without extra indentation or so, if possible. The parameterizations are in separate variables because it will also be used for the tests of the related RPC command in the follow up PR.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -28,10 +28,11 @@ from chia.types.signing_mode import SigningMode
 from chia.types.spend_bundle import SpendBundle
 from chia.util.bech32m import decode_puzzle_hash, encode_puzzle_hash
 from chia.util.byte_types import hexstr_to_bytes
-from chia.util.config import load_config
+from chia.util.config import load_config, str2bool
 from chia.util.errors import KeychainIsLocked
 from chia.util.ints import uint16, uint32, uint64
 from chia.util.keychain import bytes_to_mnemonic, generate_mnemonic
+from chia.util.misc import UInt32Range
 from chia.util.path import path_from_root
 from chia.util.ws_message import WsRpcMessage, create_payload_dict
 from chia.wallet.cat_wallet.cat_constants import DEFAULT_CATS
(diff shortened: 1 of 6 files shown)


trepca (dismissed):
lgtm

trepca (approved):

PR #15100, opened 2023-04-21
wallet: Implement `/get_coin_records` RPC command

xdustinface:
### Purpose:

This introduces a new RPC command `/get_coin_records` to fetch coin records based on the following request schema:

 ```python
@streamable
@dataclass(frozen=True)
class GetCoinRecords(Streamable):
    offset: uint32 = uint32(0)
    limit: uint32 = uint32(uint32.MAXIMUM_EXCLUSIVE - 1)
    wallet_id: Optional[uint32] = None
    wallet_type: Optional[uint8] = None  # WalletType
    coin_type: Optional[uint8] = None  # CoinType
    coin_id_filter: Optional[HashFilter] = None
    puzzle_hash_filter: Optional[HashFilter] = None
    parent_coin_id_filter: Optional[HashFilter] = None
    amount_filter: Optional[AmountFilter] = None
    amount_range: Optional[UInt64Range] = None
    confirmed_range: Optional[UInt32Range] = None
    spent_range: Optional[UInt32Range] = None
    order: uint8 = uint8(CoinRecordOrder.confirmed_height)
    reverse: bool = False
    include_total_count: bool = False  # Include the total number of entries for the query without applying offset/limit
```

### Testing Notes:

I added tests the `WalletCoinRecord.to_json_parsed_metadata` and also for the new RPC command with all test cases defined for `WalletCoinStore.get_coin_records` and i also added some failure and limit tests. 

### Based on:

- #15099 
- #15108

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -4,7 +4,7 @@ import dataclasses
 import json
 import logging
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set, Tuple, Union
+from typing import Any, ClassVar, Dict, List, Optional, Set, Tuple, Union
 
 from blspy import AugSchemeMPL, G1Element, G2Element, PrivateKey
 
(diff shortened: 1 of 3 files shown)


arvidn:
@xdustinface what's the use case for this RPC?

xdustinface:
> @xdustinface what's the use case for this RPC?

I don't have a use case by myself, it was meant to be used by #14348 which was supposed to add `get_coins_by_coin_type` initially. Also we have `get_coin_by_name` and `get_spendable_coin` and whatnot. I just thought it might be good to have some more generic, well tested RPC which covers all kinds of coin requests instead of having people adding new commands with different interfaces and responses here and there if they need some new request for some reason.

arvidn (approved):
looks good. I would think that the loops over test cases could be parametrized tests instead, and save a few lines of code

xdustinface:
> I would think that the loops over test cases could be parametrized tests instead, and save a few lines of code

Yeah agree, i wanted to do that also rather but i ended up with the loop over the tests cases to avoid creating so many extra test processes/wallet/node/instances because its so slow.

wallentx:
The snyk job shows as completed, but their status isn't updating - https://app.snyk.io/org/chia-network/pr-checks/9d1ba8f2-24fb-4cbe-9268-73d8a1b2b7dd

PR #15128, opened 2023-04-26
wallet: Implement `WalletNFTStore.delete_wallet`

xdustinface:
### Purpose:

Adds `WalletNFTStore.delete_wallet ` which allows to remove all DB entries for a given `wallet_id`.

### New Behavior:

No change in behaviour. Will be used in a follow up PR.

### Testing Notes:

Added tests for the new method.

### DIFF
--- a/chia/wallet/wallet_nft_store.py
+++ b/chia/wallet/wallet_nft_store.py
@@ -272,3 +272,8 @@ class WalletNftStore:
             if result.rowcount > 0:
                 return True
             return False
+
+    async def delete_wallet(self, wallet_id: uint32) -> None:
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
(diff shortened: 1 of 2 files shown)


trepca (approved):

chia/wallet/wallet_nft_store.py
```diff
@@ -272,3 +272,8 @@ async def rollback_to_block(self, height: int) -> bool:
             if result.rowcount > 0:
                 return True
             return False
+
+    async def delete_wallet(self, wallet_id: uint32) -> None:
```
trepca, line 276: why not put this in `nft_wallet`, wallet_nft_store is store for NFTs?

xdustinface: Its like with all other "stores" we have. When something gets changed inside the store, the method is inside since it accessed the DB of the store and this method is supposed to remove entries from the store so i guess it should be in there rather than `NFTWallet`, no?

PR #15130, opened 2023-04-26
add support for ubuntu 23 (and python 3.11) to install.sh

arvidn:
Ubuntu 23.04 defaults to python3.11 and the current `install.sh` script assumes `python3.10-venv` exists (which it doesn't appear to do for me).

This adds Ubuntu 23 to the distro detection and installs `python3.11-venv` in that case

### DIFF
--- a/.github/workflows/test-install-scripts.yml
+++ b/.github/workflows/test-install-scripts.yml
@@ -116,6 +116,10 @@ jobs:
           type: ubuntu
           # https://packages.ubuntu.com/jammy/python3 (22.04, 3.10)
           url: "docker://ubuntu:jammy"
+        - name: ubuntu:lunar (23.04)
+          type: ubuntu
+          # https://packages.ubuntu.com/lunar/python3 (23.04, 3.11)
+          url: "docker://ubuntu:lunar"
 
     steps:
     - name: Prepare Amazon Linux
--- a/install.sh
+++ b/install.sh
@@ -70,6 +70,7 @@ UBUNTU_PRE_20=0
 UBUNTU_20=0
 UBUNTU_21=0
 UBUNTU_22=0
+UBUNTU_23=0
 
 if $UBUNTU; then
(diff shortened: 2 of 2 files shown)


altendky (changes requested):
Looks like it fits the pattern.  I was going to very simply say that it should add to the pattern at https://github.com/Chia-Network/chia-blockchain/blob/4c7f7e7d23f437cd3fdc32ac19f5e20b6b8ffc9d/.github/workflows/test-install-scripts.yml#L107-L118 as well...  but then I noticed we don't cover the odd years.  Looks like they drop off the provided Docker image list sooner https://hub.docker.com/_/ubuntu and so maybe that's the issue.  Could we add 22.10 kinetic and 23.04 lunar to the list linked above to make sure we're covering this properly?

emlowe:
I think testing and supporting the LTS releases *only*, 18.04, 20.04, and 22.04 is fairly reasonable - the regular none-LTS releases EOL in one year (e.g 22.10 is EOL in July 2023).

altendky:
Then we can close this PR since 23.04 isn't LTS and we won't support it?

altendky:
To document my requested changes, there is ongoing discussion elsewhere around this area to identify what should and should not be handled.

arvidn:
I don't think status quo is acceptable. we either take this patch (or one like it) to know which package to install, or we expect the user to install the package manually.
Right now this failure aborts the whole script, including setting up the venv.

wallentx:
can we just have it install `python3-venv` for all of them

altendky:
https://github.com/Chia-Network/chia-blockchain/blob/4b3f02c3b270f19e2b38a5dc41aef0cb2b453034/install.sh#L196-L203

We're dropping 18.04 with v2.0, as far as I know, but I'm not sure we really gain a whole lot by condensing this.  I mean, it might be a good change regardless, but the testing getting added was I think maybe more of the complaint here?

PR #15131, opened 2023-04-26
Simplify handling additions and removals in validate_spend_bundle()

AmineKhaldi:


arvidn (commented):
looks reasonable. ~~In fact, it looks like `validate_spend_bundle()` really should be a free function. The only thing it does with `self` is to check that we have a peak (but then it never looks at the peak). I think it would be worth exploring making this a free function, and then ensure we have unit tests for it (which should be simpler when broken off from `MempoolManager`).~~

Never mind, I failed to expand the diff to see the whole function

arvidn (approved):

PR #15133, opened 2023-04-26
Set null default for metadata column

### LINKED ISSUE #15132: [Bug] 1.8.0 introduces a DB downgrade issue
### What happened?

1.8.0 (and main) introduce 2 new columns to the `coin_record` store here: https://github.com/Chia-Network/chia-blockchain/pull/15013

When downgrading to 1.7.1, I received many of the following in the log:

```
2023-04-25T10:48:14.156 wallet chia.wallet.wallet_state_manager: ERROR    Error adding state... table coin_record has 12 columns but 10 values were supplied
  File "chia/wallet/wallet_state_manager.py", line 1146, in new_coin_state
  File "chia/wallet/wallet_coin_store.py", line 86, in add_coin_record
  File "aiosqlite/core.py", line 194, in execute_insert
  File "aiosqlite/core.py", line 129, in _execute
  File "aiosqlite/core.py", line 102, in run
  File "aiosqlite/core.py", line 74, in _execute_insert
sqlite3.OperationalError: table coin_record has 12 columns but 10 values were supplied
```

### Version

1.8.0RC3

Windows

GUI

### Relevant log output

_No response_

emlowe:
Fixes #15132 
DB downgrade bug

### DIFF
--- a/chia/wallet/wallet_coin_store.py
+++ b/chia/wallet/wallet_coin_store.py
@@ -57,7 +57,7 @@ class WalletCoinStore:
 
             try:
                 await conn.execute("ALTER TABLE coin_record ADD COLUMN coin_type int DEFAULT 0")
-                await conn.execute("ALTER TABLE coin_record ADD COLUMN metadata blob")
+                await conn.execute("ALTER TABLE coin_record ADD COLUMN metadata blob DEFAULT null")
                 await conn.execute("CREATE INDEX IF NOT EXISTS coin_record_coin_type on coin_record(coin_type)")
             except sqlite3.OperationalError:
                 pass


arvidn (approved):

xdustinface (changes requested):
Hm, sure this works for you? I tried here and im still getting the failure.

arvidn:
@xdustinface did you delete your wallet DB and re-create it with this patch?

arvidn:
my understanding is that if you have added those columns already, without the defaults, your DB won't support downgrading. We won't update them if they exist. I expect we would just get an exception and exit the try-block

xdustinface:
> @xdustinface did you delete your wallet DB and re-create it with this patch?

I manually altered the DB back and forth running 1.7.1 to test it but it failed always. Did it work for you @arvidn?

arvidn:
> I manually altered the DB back and forth running 1.7.1 to test it but it failed always.

You mean you simulated the behavior of this PR by manually executing the command in `sqlite3`?

> Did it work for you @arvidn?

I have not tested it

xdustinface:
> > I manually altered the DB back and forth running 1.7.1 to test it but it failed always.
> 
> You mean you simulated the behavior of this PR by manually executing the command in `sqlite3`?

Yes basically i just run the below in the `sqlite3` CLI

```sql
ALTER TABLE coin_record ADD COLUMN coin_type int DEFAULT 0;
ALTER TABLE coin_record ADD COLUMN metadata int DEFAULT null;
```
then i verified the schema with `.schema coin_record`, closed the DB, started the wallet, waited until it tried to add a coin state and it failed.

Then stop, execute

```sql
ALTER TABLE coin_record DROP COLUMN coin_type;
ALTER TABLE coin_record DROP COLUMN metadata;
```

and it works again. I tried few times.. means, either this PR doesn't help or maybe i did some strange mistake so i think someone else should confirm this too :)

emlowe:
Ok confirmed this does NOT work.

when I initially tested it, I was led wrong by some proper "coin state added" log entries, however, those were for coins that were ALREADY in the table, so the log entry suggests those were added when in fact they were already present.

I tested this again using Kevin's idea and deleted all the rows:
```
ALTER TABLE coin_record ADD COLUMN coin_type int DEFAULT 0;
ALTER TABLE coin_record ADD COLUMN metadata blob DEFAULT null;
DELETE from coin_record;
```

and started the wallet. Now the coin states cannot be added with the same error.

Thanks Kevin for double-checking this change!

emlowe:
Closing PR, as it doesn't seem to fix the issue

PR #15134, opened 2023-04-26
rework class protocol check pattern

altendky:

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -4,7 +4,7 @@ import dataclasses
 import logging
 import time
 from operator import attrgetter
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple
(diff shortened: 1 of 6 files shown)


arvidn:
what's the behavior of moving those checks into the class body itself?

altendky:
It makes them located closer to the `class` code where I suspect people would expect it to be.  This should make the pattern more apparent to people (sorta...) and avoid stuff like https://github.com/Chia-Network/chia-blockchain/pull/14991.  In terms of functional change, you get a class var instead of a global (not a big deal) and the `cast()` pattern avoids 'instantiating' the class which makes it more flexible for use with classes that require parameters to be instantiated.

xdustinface (approved):
> you get a class var instead of a global 

Do we actually _get_ the class var in there since it's only happening `if TYPE_CHECKING`?

altendky:
If you're type checking you do.  `;]`  Which may include IDEs, I'm not sure.

arvidn (approved):

PR #15144, opened 2023-04-27
tests: Drop `run-py-tests.sh`

xdustinface:
### Purpose:

Doesn't look very helpful to me? It's also not used or mentioned anywhere.


emlowe (approved):
Yea, I always thought this should be removed

PR #15149, opened 2023-04-28
wallet: Lock `WalletStateManager.lock` while populating balances initially

### LINKED ISSUE #15142: [Bug] dictionary changed size during iteration
### What happened?

when switching between wallets 
wallet_change_error_GUI
occured.
after clicking [OK] and clicking on the desired wallet, it will load without issue

### Version

1.8.0rc3

Windows

GUI

_No response_

xdustinface:
### Purpose:

Actually this should not be an issue anymore after #15148 and will also fix #15142, but this PR is to be on the safe side. Basically it just makes sure to lock the `WalletStateManager.lock` before iterating `WalletStateManager.wallets` to populate the balances which will avoid changes during iteration because we only change the size within lock contexts. While we could also just copy the dict before iterating i think this makes more sense. 

### Current Behavior:

The lock gets locked for each individual balance update call.

### New Behavior:

The lock gets locked once for all balance update calls.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -385,8 +385,9 @@ class WalletNode:
         self.wallet_state_manager.state_changed("sync_changed")
 
         # Populate the balance caches for all wallets
-        for wallet_id in self.wallet_state_manager.wallets:
-            await self.get_balance(wallet_id)
+        async with self.wallet_state_manager.lock:
+            for wallet_id in self.wallet_state_manager.wallets:
+                await self._update_balance_cache(wallet_id)
 
         async with self.wallet_state_manager.puzzle_store.lock:
             index = await self.wallet_state_manager.puzzle_store.get_last_derivation_path()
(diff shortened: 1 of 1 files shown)


chia/wallet/wallet_node.py
```diff
@@ -385,8 +385,9 @@ async def _start_with_fingerprint(
         self.wallet_state_manager.state_changed("sync_changed")
 
         # Populate the balance caches for all wallets
-        for wallet_id in self.wallet_state_manager.wallets:
-            await self.get_balance(wallet_id)
+        async with self.wallet_state_manager.lock:
+            for wallet_id in self.wallet_state_manager.wallets:
+                await self._update_balance_cache(wallet_id)
```
arvidn, line 390: this seems to change more than what you describe in the PR. You no longer call `self._balance_cache.get(wallet_id, Balance())` from here now. You also don't take `self.wallet_state_manager.sync_mode` into account.

xdustinface: > this seems to change more than what you describe in the PR.

Okay yeah i slightly changed the "Purpose" section.

 > You no longer call `self._balance_cache.get(wallet_id, Balance())` from here now. 

Its not relevant since this is only to populate the caches on startup. 

> You also don't take `self.wallet_state_manager.sync_mode` into account.

This is also not relevant for startup since its only to make the RPC calls use the cache while we are syncing and update the balance if not.

arvidn (approved):

PR #15150, opened 2023-04-28
wallet: Move `WalletPeers` init to the end of `_start_with_fingerprint`

### LINKED ISSUE #15142: [Bug] dictionary changed size during iteration

xdustinface:
### Purpose:

Moves the init of `WalletPeers` to the end of the `WalletNode`'s start method to make sure we only connect to peers after we initialised and set up everything properly. This should like #15149 and #15148 also fix #15142 because we should not come into the situation where we modify `WalletStateManager.wallets` without connecting to peers, but yeah, i think it makes sense take them all into the release :)

### Current Behavior:

The wallet starts to connect to peers before everything is setup properly in `WalletNode`.

### New Behavior:

The wallet starts to connect to peers when everything is setup and initialised.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -377,9 +377,6 @@ class WalletNode:
             self,
         )
 
-        if self.wallet_peers is None:
-            self.initialize_wallet_peers()
-
         if self.state_changed_callback is not None:
             self.wallet_state_manager.set_callback(self.state_changed_callback)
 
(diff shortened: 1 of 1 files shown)


wallentx:
Targeting this to main, as per discussed in LR

emlowe (changes requested):
Looks like some things got a little mixed in during the base branch switch
-gui change
block_header_validation
etc,

xdustinface:
> Looks like some things got a little mixed in during the base branch switch

Oh yeah, rebased on `main`!

emlowe (approved):

PR #15152, opened 2023-04-28
add generator2 to the repository

arvidn:
This is intended to be used in the 2.0 hard fork. The difference from the current generator ROM is that it does not pass in the CLVM parser as the first argument (but instead passes in `NULL`. This makes it compatible with all existing blocks, but it is technically a hard-fork because a potential generator could depend on it.


richardkiss (approved):
Have we tested that this works on old block? My main concern is that the cost would be different (although when I think about it, it seems like it would not be).

arvidn:
I have tested it on some of the past blocks (but it takes 24 - 48 hours to run through the whole chain) and I got a power-outage somewhere before completion :)

I definitely intend to test all historical blocks, probably by running them with both old and new generator and ensuring the output is identical.

That said, this *is* a hard fork and is intended to be activated along with it.
The reason to land just the generator program, without hooking it up to anything, is that there's a bit of a chicken-and-egg issue in the order various patches are introduced. The generator that's used by consensus lives in `chia_rs` for instance, and without having this here, adding it there doesn't have anything to reference in here. And without adding the new generator to chia_rs, we can't run tests against both in here.

emlowe (approved):

PR #15154, opened 2023-04-28
wallet: Revert "Add new coin_type & metadata to coin_records (#15013)"

xdustinface:
### Purpose:

This reverts #15013 because clawback was postponed and we don't need this two extra columns for this release. 

### Current Behavior:

The wallet adds two new columns to the DB on first start of 1.8.0.  

### New Behavior:

The wallet does not add the two new columns to the DB anymore.

### DIFF
--- a/chia/wallet/util/wallet_types.py
+++ b/chia/wallet/util/wallet_types.py
@@ -29,11 +29,6 @@ class WalletType(IntEnum):
     DATA_LAYER_OFFER = 12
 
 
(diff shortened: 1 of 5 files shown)


arvidn (approved):
is this a verbatim revert of the original patch? (I assume yes)
did you have any conflicts? (I assume no)

xdustinface:
Yes i just reverted it and it was without conflicts.

PR #15163, opened 2023-04-29
Make Simulator CLI Tests Less Flaky

jack60612:
### Purpose:
Cleans up the code for cleaning up the simulator cli tests

### Current Behavior:
Not all errors are catched and only 1 dir is used in the tests

### New Behavior:
2 dirs and all errors are catched

### DIFF
--- a/tests/cmds/test_sim.py
+++ b/tests/cmds/test_sim.py
@@ -19,23 +19,25 @@ burn_address = "txch1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqm6ksh7qddh"
 def test_every_simulator_command() -> None:
     runner: CliRunner = CliRunner()
     address = std_farming_address
-    start_result: Result = runner.invoke(cli, ["dev", "sim", "-n", "ci_test", "create", "-bm", mnemonic])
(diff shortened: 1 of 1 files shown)


xdustinface (approved):
Looks like this should improve the situation yes.

It also seems like the code in `test_every_simulator_command` should maybe just be a fixture or contextmanager (not a test) with an optional parameter `address` which leads to setting `--reward-address address` if its not `None` and which yields `runner, address` and gets passed into the other tests where you then could remove the prefix `_` to make them separate tests.

- `test_sim_status` (which you could parameterize the fixture then with both addresses)
- `test_farm_and_revert_block`

jack60612:
> Looks like this should improve the situation yes.
> 
> It also seems like the code in `test_every_simulator_command` should maybe just be a fixture or contextmanager (not a test) with an optional parameter `address` which leads to setting `--reward-address address` if its not `None` and which yields `runner, address` and gets passed into the other tests where you then could remove the prefix `_` to make them separate tests.
> 
> * `test_sim_status` (which you could parameterize the fixture then with both addresses)
> * `test_farm_and_revert_block`

i'll do this later, I just want the flake to go away.

emlowe (approved):

PR #15165, opened 2023-05-01
full_node: Drop `peers_with_peak` removals

xdustinface:
### Purpose:

Drop the removals from `peers_with_peak` since it doesn't make sense. We initially copy it into `new_peers_with_peak` which:

- Is the list we exclusively work with later
- Gets anyway replaced every time `SyncStore.peers_changed` gets triggered, which happens on a disconnect.

Alternative to #14803.

### Current Behavior:

There are cases where we try to remove peers which not longer exist in the original `peers_with_peak` list because we not work of this list anymore after the initial copying. 

### New Behavior:

The invalid removals are gone but the general behaviour is the same.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -1078,12 +1078,10 @@ class FullNode:
                     fetched = False
                     for peer in random.sample(new_peers_with_peak, len(new_peers_with_peak)):
                         if peer.closed:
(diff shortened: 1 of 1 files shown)


altendky (approved):
Thanks, and sorry I didn't get back to this enough to get my head on straight before you went and PR'ed yourself.

emlowe (approved):

PR #15171, opened 2023-05-02
wallet: Explicit hinting for transaction generation

xdustinface:
### Purpose:

This removes the implicit hinting from `NFTWallet.generate_signed_transaction` and `CATWallet.generate_signed_transaction`, just to be explicit about the hinting.. There are at least the three cases below where we currently double hint NFT transactions, unless i miss the point of it. Maybe there are other places, might be also related to CATs. Alternative to this PR is to drop the redundant hints but i feel like this PR also helps to avoid double hinting in the future.

https://github.com/Chia-Network/chia-blockchain/blob/dcebabab303507a238ba575c8a8e7de727a89995/chia/wallet/nft_wallet/nft_wallet.py#L446

https://github.com/Chia-Network/chia-blockchain/blob/dcebabab303507a238ba575c8a8e7de727a89995/chia/wallet/nft_wallet/nft_wallet.py#L1415

https://github.com/Chia-Network/chia-blockchain/blob/dcebabab303507a238ba575c8a8e7de727a89995/chia/wallet/nft_wallet/nft_wallet.py#L1652

### Current Behavior:

The `generate_signed_transaction` methods implicitly hint each payment with its puzzle hash.

### New Behavior:

The `generate_signed_transaction` doesn't implicitly add hints, instead all call sites add them explicitly.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1454,7 +1454,7 @@ class WalletRpcApi:
                     fee,
                     cat_discrepancy=cat_discrepancy,
                     coins=coins,
-                    memos=memos if memos else None,
+                    memos=memos if memos else [[puzzle_hash] for puzzle_hash in puzzle_hashes],
                     min_coin_amount=min_coin_amount,
                     max_coin_amount=max_coin_amount,
                     exclude_coin_amounts=exclude_coin_amounts,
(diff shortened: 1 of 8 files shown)


Quexington (commented):
I don't like this and prefer the other fix that you proposed.  As you can see from the diff, there's a lot of places we were leaning on this.  Double hints are relatively harmless and if they're identified we can take them out without any trouble.  I think it's much worse to forget a hint (even if you catch it in test it would probably be a time sink for a dumb mistake) than to accidentally double hint so that's what I would prefer we guard against.

emlowe:
Close unfinished PR

PR #15179, opened 2023-05-02
Hint to CAT change

Quexington:
This PR would start making the CAT wallet hint to its change outputs which it does not do currently.  This should not fix anything necessarily right now, but is a good idea for the future as we try to make the wallet sync more robust.


xdustinface (changes requested):
Looks good but the tests are failing, looks like you need to adjust some more asserts.

emlowe (approved):

xdustinface (approved):

PR #15182, opened 2023-05-03
Fix chia wallet make_offer short-option collision on -r

### LINKED ISSUE #14874: [Bug] duplicated short options, `-r`, in `make_offer` CLI causing missing option error.
### What happened?

`-r` is already used for `--request` but it's now also used for `--reuse`, `Reuse existing address for the change.`, and this causes `make_offer` error, `Error: Missing option '-r' / '--request'.`

https://github.com/Chia-Network/chia-blockchain/blob/9a4b42b022a31357e36fd92900cf3efc3be8a02d/chia/cmds/wallet.py#L411-L425

### Version

1.7.1-rc3

macOS

CLI

### Relevant log output

```shell
❯ chia wallet make_offer -f $maker_fp -r 1:0.08 -o 2:15.898 -p offer1.offer
Usage: chia wallet make_offer [OPTIONS]
Try 'chia wallet make_offer -h' for help.

Error: Missing option '-r' / '--request'.
```

yyolk:
### Purpose:
Closes #14874

Previously I opted to change the short option for `--request` to `-R` and for consistency give `-O` as the short option for `--offer`. Currently, after agreeing with https://github.com/Chia-Network/chia-blockchain/pull/15182#issuecomment-1533650447 - the path of dropping the short option, `-r` for setting the flag `reuse` was decided. 

I dropped everywhere the short option `-r` is set for the `reuse` flag on `chia wallet` commands.

An associated update to Chia-Network/chia-docs: https://github.com/yyolk/chia-docs/tree/chia-blockchain-15182-chia-wallet-cli-options-reference-update

I'll open PR for the docs if this change will be accepted.

### Current Behavior:

- `-r` is ignored for a `chia wallet make_offer` since a later `@click.option(...)` for `--reuse` gets assigned to it 
- `chia wallet make_offer` requires `-r` to be set. Eg. If supplied `-r 1:10 -r 2:32` it will error with: 
   
         Error: Missing option '-r' / '--request'

   so the user must supply `--request`,  for all request args given to `chia wallet make_offer`
- `-o` is the short option for `--offer` and doesn't have any conflicts for its command
- the short option for `reuse`, `-r`, is able to be used and doesn't error the command when there's not a collision with another required option.

A command would look like this when given the only option to use `--request`:

    chia wallet make_offer -o 1:15.5333 -o 3:3.789898 --request 4:11108.32 ...

### New Behavior:
- `-r` works as the short option for `--request` in `chia wallet make_offer` command, either may be used
- the long option `--reuse` is the only way to set the flag `reuse` to a command

A command will look like this when short options for `--offer` and `--request` may be used:

    chia wallet make_offer -o 1:15.5333 -o 3:3.789898 -r 4:11108.32 ...

### Testing Notes:
I don't believe there are any integration tests that go as far as to test for the options supplied to a command. 

The commands invoked as python methods in any unit tests will work the same way as before.


paninaro:
@yyolk Thank you for the contribution! The fix and detailed commit PR description are greatly appreciated.

There was a brief internal discussion about how best to proceed with this fix. Since the `--offer` and `--request` options pre-date the newer `--reuse` option, it would be preferable to keep `-o` and `-r` as the short form options for offer and request. We can remove the `-r` short option for `--reuse` entirely (leaving just the long form), or find a new unique short option character (I would probably just keep the long option).

Would you mind making those changes in this PR? Admittedly, there are 7 locations that I see in `chia/cmds/wallet.py` that specify the `--reuse` option, so it would be a slightly larger patch. Thank you!

yyolk:
Certainly! 

The suggested path forward also simplifies things in regards to the needing to update the docs, since the short form is in the examples and it doesn't have any touch with `--reuse`. I also appreciate the brevity of `-o` & `-r` when making offers so I prefer this path as well. 😎

I'll update this PR shortly to: 

- drop the short option on `reuse` at a minimum for this `make_offer_cmd` leaving `request` and `offer`.
- For consistency I'll also drop the short option for `reuse` and leave it only as long option everywhere else it's used, which as you already mentioned is at least 7 locations. 

I'll make an effort to double check if `-r` is used elsewhere in the docs when it will need to be changed to the long option form to flag `reuse` to a command. I'll update my staging branch on my chia-docs fork to reflect those changes if they are needed so the docs do not suggest an option that won't work in its examples.

yyolk:
I've updated the PR with the changes. Let me know if there's more I should do. 
Note - the compare for the force push is also seeing pulling in changes from main, since I rebased and squashed to leave it as one commit :) Looking at Files Changed will show the appropriate delta. With the first commit to this PR being reverted immediately, I decided to drop those messages from the commit log as well during the squash.

IRT chia-docs: a quick look at chia-docs, there are some places where `-r` or `--reuse` is mentioned for the wallet cli reference command reference. I created a new staging branch to more directly associate to this PR and to similarly reflect the changes of dropping the short option of `-r` wherever `reuse` is used. I'll open a PR for that branch if this PR will be accepted. That branch is here, and is also in the updated OP: https://github.com/yyolk/chia-docs/tree/chia-blockchain-15182-chia-wallet-cli-options-reference-update.

I've also updated the OP to reflect these changes.

yyolk:
I read in CONTRIBUTING that all PRs are squashed, which I didn't see before. I'll leave the latest merge commit for resolving the merge conflict instead of rebasing from main@HEAD, since it will be squashed on merge.

paninaro (approved):
LGTM. Thanks for the fix!

PR #15183, opened 2023-05-03
Proposals for harvesting

ChiaMineJP:
Making a proposal of new RPC/state_change APIs from GUI point of view. This is just a draft and will vary in the future.


xdustinface (changes requested):
"Proposal" sounds like this is more like an idea for now than a concrete i want to get this merged thing but i still left some comments. Also want to note that this breaks:

- The harvester protocol and the plot sync which is probably ok to do and something we did before but we need to communicate it and be aware of it. 
- The caching which is imo not ok. This should do an update of the cache and maybe even store the new cache in a separate file to keep the cache for downgrades since this changes the protocol and if there are downgrades for some reasons for big farms its just nice to have the cache available.

chia/rpc/harvester_rpc_api.py
```diff
@@ -35,13 +36,17 @@ async def _state_changed(self, change: str, change_data: Dict[str, Any] = None)
 
         if change == "farming_info":
             payloads.append(create_payload_dict("farming_info", change_data, self.service_name, "metrics"))
+            payloads.append(create_payload_dict("farming_info", change_data, self.service_name, "wallet_ui"))
```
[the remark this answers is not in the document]
ChiaMineJP: Harvester's `farming_info` event is required for the GUI to make local harvester latency graph. (This is requested by https://github.com/Chia-Network/issue-tracker/issues/374)

Farmer's `new_farming_info` event looks like harvester's `farming_info` event but slightly different in that the latter has harvester latency data while the former does not. Farmer's `new_farming_info` summarizes data from connected harvester and it's difficult to let it have each harvesters' time which it takes to calculate proof of space from disks.

chia/rpc/harvester_rpc_api.py
```diff
@@ -35,13 +36,17 @@ async def _state_changed(self, change: str, change_data: Dict[str, Any] = None)
...
 
         if change == "add_connection":
             payloads.append(create_payload_dict("add_connection", change_data, self.service_name, "metrics"))
 
         if change == "close_connection":
             payloads.append(create_payload_dict("close_connection", change_data, self.service_name, "metrics"))
 
+        if change == "harvesting_mode_changed":
```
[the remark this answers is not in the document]
ChiaMineJP: Currently only "local" harvester's harvesting mode (CPU/GPU) is interested by the GUI. Besides, communication between farmer <-> harvester is done via Peer protocol (protocols defined under `chia/protocols`) and rarely use RPC API from my quick look. I think updating peer protocol should be avoided as long as possible because it is part of consensus.

chia/rpc/harvester_rpc_api.py
```diff
@@ -77,3 +82,16 @@ async def remove_plot_directory(self, request: Dict) -> EndpointResult:
         if await self.service.remove_plot_directory(directory_name):
             return {}
         raise ValueError(f"Did not remove plot directory {directory_name}")
+
+    async def change_harvesting_mode(self, request: Dict):
+        if request["mode"] not in ("GPU", "CPU"):
```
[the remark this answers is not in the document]
ChiaMineJP: This is my weak suggestion so just a temporal placeholder for testing GUI.
From GUI point of view, making this as `IntEnum` requires prior knowledge of constants for GUI developers so explicit string looks better for me.
(When code is only refered in Python code, `IntEnum` should be better)

chia/harvester/harvester.py
```diff
@@ -178,5 +190,15 @@ async def remove_plot_directory(self, str_path: str) -> bool:
         self.plot_manager.trigger_refresh()
         return True
 
+    async def change_harvesting_mode(self, mode: HarvestingMode):
+        # Check whether the harvester is eligible for GPU harvesting if it is requested
+        if mode == HarvestingMode.GPU and not self._support_gpu:
+            err_msg = "GPU harvesting is not available on this system"
+            return False, err_msg
```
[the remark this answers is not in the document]
ChiaMineJP: Fixed as suggested

chia/rpc/harvester_rpc_api.py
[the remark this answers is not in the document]
xdustinface: > From GUI point of view, making this as `IntEnum` requires prior knowledge of constants for GUI developers so explicit string looks better for me.

Doesn't a string parameter also require prior knowledge? I mean, in any case you need to look into the docs and understand what value you can pass to which parameter. And i think usually when doing this you anyway create an abstraction layer for the endpoints/parameters in whatever language you use to access the API. So at least i try to use the more efficient and less error prone way always which is imo just a int mapping and not some string value but sure, there are other opinions and its not the only way to go.

xdustinface (changes requested):
I think besides all this V2 stuff is not really fully implemented yet (new protocol message tests are missing and plot sync and few other tests aren't passing). We should not do this V2 messages thing for a few reasons imo:

1. We anyway already break the harvester protocol afaik with the plot filter changes and people need to upgrade all their components. I think this goes into the same release?
2. This adds not necessarily needed complexity and basically duplicated structures which will need to stay in here for a long time if we want to remove it silently.
3. Right now there are not really any tests added for all this V2 stuff and existing tests use the non-V2 stuff. There are even a bunch of places where you now pass non existing arguments to data structures in tests which leads to test failures. So yeah, this is all is not ready anyway at this point.
4. We waste `ProtocolMessageTypes` entries for this now were we currently only use `uint8` in a `Message` and we already almost at half of the possibilities :) 

Just my opinion, but if we for some reason want to continue this V2 thing it should at least be integrated fully/correct and be better tested.

chia/plotting/manager.py
```diff
@@ -262,18 +267,18 @@ def _refresh_task(self, sleep_interval_ms: int):
...
+                self.log.error(f"_refresh_callback raised: {e} with the traceback: {traceback.format_exc()}")
                 self.reset()
 
     def refresh_batch(self, plot_paths: List[Path], plot_directories: Set[Path]) -> PlotRefreshResult:
         start_time: float = time.time()
         result: PlotRefreshResult = PlotRefreshResult(processed=len(plot_paths))
         counter_lock = threading.Lock()
 
-        log.debug(f"refresh_batch: {len(plot_paths)} files in directories {plot_directories}")
+        self.log.debug(f"refresh_batch: {len(plot_paths)} files in directories {plot_directories}")
```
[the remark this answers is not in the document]
ChiaMineJP: Without adding `self.`, messages wouldn't be written in a log file. This is the reason I added `self.` around.
Could you check in your `debug.log` whether "refresh_batch: " or other log entires are appeared?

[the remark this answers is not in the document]
xdustinface: Works fine here for me on `main` yeah. You sure the harvester was connected to a farmer and did run this code at all when you didn't see logs? And just a general thing: Even if it would not work it would be good to move something like this into a separate PR into `main` instead of sneaking it into a PR against the compression branch.

chia/plotting/manager.py
```diff
@@ -56,7 +56,11 @@ def __init__(
         self.no_key_filenames = set()
         self.farmer_public_keys = []
         self.pool_public_keys = []
-        self.cache = Cache(self.root_path.resolve() / "cache" / "plot_manager.dat")
+        # Since `compression_level` property was added to Cache structure,
+        # previous cache file formats needs to be reset
+        # When user downgrades harvester, it looks 'plot_manager.dat` while
+        # latest harvester reads/writes 'plot_manager_v2.dat`
+        self.cache = Cache(self.root_path.resolve() / "cache" / "plot_manager_v2.dat")
```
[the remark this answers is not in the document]
ChiaMineJP: I investigated diff between `chiapos@1.0.11` (pinned in `chia-blockchain@main`) and `chiapos@1.0.12b12` (pinned in `chia-blockchain@long_lived/compression`).
https://github.com/Chia-Network/chiapos/compare/1.0.11...1.0.12b13

image

It seems new `chiapos` can serialize/deserialize both old and new versions of `DiskProver` binary.
But it doesn't mean old `chiapos` can deserialize binary created by new `chiapos`.


>The caching which is imo not ok. This should do an update of the cache and maybe even store the new cache in a separate file to keep the cache for downgrades since this changes the protocol and if there are downgrades for some reasons for big farms its just nice to have the cache available.

I feel your past suggestion above makes sense. When a user downgrade `chia-blockchain`, harvester would keep crashing because of cache deserialization failure.
I think `"plot_manager_v2.dat"` is nice and not messing up `cache` folder so much.

However, as you suggested, `compression_level` is not need in `CacheEntry` and `DiskCacheEntry`. I'll remove the property from both classes.

chia/util/misc.py
```diff
@@ -18,19 +18,20 @@ class VersionedBlob(Streamable):
     blob: bytes
 
 
-def format_bytes(bytes: int) -> str:
-    if not isinstance(bytes, int) or bytes < 0:
+def format_bytes(bytes_value: int, effective: bool = False) -> str:
```
xdustinface, line 21: Seems strange to add this `effective` parameter where the only difference is that it adds an `e` to the label? Why not just add the `e` in the places where you use it with `effective=True` because it seems in those places you anyway also add `(effective)` to the string.

chia/rpc/wallet_rpc_api.py
```diff
@@ -2987,13 +2989,17 @@ async def get_farmed_amount(self, request) -> EndpointResult:
             # .get_farming_rewards() above queries for only confirmed records.  This
             # could be hinted by making TransactionRecord generic but streamable can't
             # handle that presently.  Existing code would have raised an exception
-            # anyways if this were to fail and we already have an assert below.
+            # anyway if this were to fail and we already have an assert below.
```
xdustinface, line 2992: Unrelated.

chia/farmer/farmer.py
xdustinface, line 1: Looks like the changes in here should be PRed separately into `main` because its not related to plot compression?

chia/daemon/server.py
xdustinface, line 1: The changes in here don't seem to be related to harvesting. Maybe you should have a separate PR into the compression branch for plotting changes?

chia/farmer/farmer_api.py
xdustinface, line 1: Same here, looks like the changes in here should be PRed separately into `main` because its not related to plot compression?

chia/harvester/harvester_api.py
xdustinface, line 1: Also here all the changes besides the part with the plot compression in `request_plots` should just go into `main` it seems?

chia/plot_sync/receiver.py
```diff
@@ -114,6 +120,8 @@ def reset(self) -> None:
         self._keys_missing.clear()
         self._duplicates.clear()
         self._total_plot_size = 0
+        self._total_effective_plot_size = float(0)
```
xdustinface, line 123: Why do we need to add this to the receiver since its the same as `_total_plot_size` just times a constant factor which is only relevant for display purpose?

chia/plot_sync/receiver.py
```diff
@@ -357,6 +375,8 @@ def to_dict(self, counts_only: bool = False) -> Dict[str, Any]:
             "no_key_filenames": get_list_or_len(self._keys_missing, counts_only),
             "duplicates": get_list_or_len(self._duplicates, counts_only),
             "total_plot_size": self._total_plot_size,
+            "total_effective_plot_size": int(self._total_effective_plot_size),
```
xdustinface, line 378: Why do we need to add this here to the JSON which makes it part of the RPC? If we for some reason want to add it to RPC i think we should just apply `UI_ACTUAL_SPACE_CONSTANT_FACTOR` here instead of storing it also in `Receiver. _total_effective_plot_size`.

chia/plotters/bladebit.py
xdustinface, line 1: All this are plotting related changes and should maybe be a separate PR instead like mentioned in the other file already.

chia/plotters/plotters.py
xdustinface, line 1: All this are plotting related changes and should maybe be a separate PR instead like mentioned in the other file already.

chia/rpc/full_node_rpc_api.py
xdustinface, line 1: This is also unrelated to plot compression and should be a PR into `main` imo.

chia/rpc/wallet_rpc_api.py
xdustinface, line 1: Same here, unrelated to plot compression and can go separately into `main`.

xdustinface (changes requested):
I didn't really review yet but when i started to review i realized this evolved from a PR which wants to add some new messages and supported for compressed plots in the harvester to something pretty big which tries to implement, change and fix various things which makes it really hard to review. I think there are a bunch of things which can just go into `main` and don't need to go into some long lived branch. Looks like it should be split up into maybe

- Harvester support for compression into the compression branch
- Plotting changes into the compression branch
- Farmer pooling statistics changes into `main`
- The FullNode `get_blockchain_state ` RPC change into `main`
- `FarmingInfo.lookup_time` change into `main`

chia/consensus/pos_quality.py
```diff
@@ -2,12 +2,12 @@
 
 from chia.util.ints import uint64
 
-# The actual space in bytes of a plot, is _expected_plot_size(k) * UI_ACTUAL_SPACE_CONSTANT_FACTO
+# The actual space in bytes of a plot, is expected_plot_size(k) * UI_ACTUAL_SPACE_CONSTANT_FACTO
 # This is not used in consensus, only for display purposes
-UI_ACTUAL_SPACE_CONSTANT_FACTOR = 0.762
+UI_ACTUAL_SPACE_CONSTANT_FACTOR = 0.78
```
arvidn, line 7: I think this change warrants a good explanation

tests/plot_sync/test_delta.py
```diff
@@ -15,7 +15,7 @@
 
 
 def dummy_plot(path: str) -> Plot:
-    return Plot(path, uint8(32), bytes32(b"\00" * 32), G1Element(), None, G1Element(), uint64(0), uint64(0))
+    return Plot(path, uint8(32), bytes32(b"\00" * 32), G1Element(), None, G1Element(), uint64(0), uint64(0), uint8(0))
```
arvidn, line 18: I really think you need to force these arguments to be named.

tests/util/test_misc.py
```diff
@@ -27,6 +27,24 @@ async def test_format_bytes(self):
...
+        assert format_bytes(1024**2 - 1000, True) == "0.999 MiBe"
+        assert format_bytes(1024**2, True) == "1.000 MiBe"
+        assert format_bytes(1024**3, True) == "1.000 GiBe"
+        assert format_bytes(1024**4, True) == "1.000 TiBe"
+        assert format_bytes(1024**5, True) == "1.000 PiBe"
+        assert format_bytes(1024**6, True) == "1.000 EiBe"
+        assert format_bytes(1024**7, True) == "1.000 ZiBe"
+        assert format_bytes(1024**8, True) == "1.000 YiBe"
+        assert format_bytes(1024**9, True) == "1024.000 YiBe"
+        assert format_bytes(1024**10, True) == "1048576.000 YiBe"
```
arvidn, line 45: excuse my ignorance. what does the "e" suffix mean?

wallentx: "Effective", I think

PR #15184, opened 2023-05-03
fix GENEARATOR_MOD2 to have the same cost as GENERATOR_MOD

arvidn:
There was an issue with the first version of the `GENERATOR_MOD2` causing it to have lower cost than the current `GENERATOR_MOD`. This was because the *reference* into the environment to pick out the deserializer was replaced by an inline null, which is cheaper.

This makes the `GENERATOR_MOD2` more similar to the existing one by putting the null in the environment and referencing it, just like the existing one does.


richardkiss (approved):

richardkiss:
I can't remember, what's the motivation in changing this? I know it's ugly... is there anything beyond that?

arvidn:
that is the main motivation. I'm hoping to port this to rust at some point in the future, and this would make that code a bit cleaner too

trepca (approved):

PR #15186, opened 2023-05-03
protocols: Give rust wallet types an alias, don't use `__all__`

xdustinface:
### Purpose:

It's not "all". There are other classes defined in the file and PyCharm highlights other imports from that file as missing. 

### New Behavior:

No change in behaviour.


arvidn (dismissed):
I assume mypy is happy about this. It was surprisingly hard to find any documentation on how to explicitly re-export names

arvidn (approved):

PR #15196, opened 2023-05-03
Add type filter for get_transactions

ytx1991:
### Purpose:

UI team needs to filter out Clawback transactions

### Current Behavior:

Doesn't support filter by tx type.

### New Behavior:
Add type_filter parameter for the get_transactions API. Example: {type_filter: {include: True, types:[1,2,3]}}

### Testing Notes:
Unit tests

### DIFF
--- a/chia/wallet/util/query_filter.py
+++ b/chia/wallet/util/query_filter.py
@@ -15,11 +15,6 @@ class FilterMode(IntEnum):
     exclude = 2
 
 
-class CoinRecordOrder(IntEnum):
-    confirmed_height = 1
-    spent_height = 2
-
-
 @streamable
 @dataclass(frozen=True)
 class TransactionTypeFilter(Streamable):
--- a/chia/wallet/util/query_filter.py
+++ b/chia/wallet/util/query_filter.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import IntEnum
+from typing import List
+
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.ints import uint8, uint64
+from chia.util.streamable import Streamable, streamable
+from chia.wallet.util.transaction_type import TransactionType
+
+
+class FilterMode(IntEnum):
+    include = 1
+    exclude = 2
+
+
+class CoinRecordOrder(IntEnum):
+    confirmed_height = 1
+    spent_height = 2
+
+
+@streamable
+@dataclass(frozen=True)
+class TransactionTypeFilter(Streamable):
+    values: List[uint8]
+    mode: uint8  # FilterMode
+
+    @classmethod
+    def include(cls, values: List[TransactionType]) -> TransactionTypeFilter:
+        return cls([uint8(t.value) for t in values], uint8(FilterMode.include))
+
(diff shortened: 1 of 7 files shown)


xdustinface (changes requested):
Left few comments.

chia/wallet/wallet_transaction_store.py
```diff
@@ -3,20 +3,42 @@
...
 from chia.util.ints import uint8, uint32
+from chia.util.streamable import Streamable, streamable
 from chia.wallet.transaction_record import TransactionRecord
 from chia.wallet.transaction_sorting import SortKey
 from chia.wallet.util.transaction_type import TransactionType
 
 log = logging.getLogger(__name__)
 
 
+class FilterMode(IntEnum):
```
xdustinface, line 22: You should imo reuse `chia.wallet.wallet_coin_store.FilterMode` and not just duplicate it. Not sure if it adds some circular dependency somewhere, but if so you should probably move it to some other place, maybe `chia.wallet.wallet_types` or so.

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -637,6 +639,13 @@ async def test_get_transactions(wallet_rpc_environment: WalletRpcTestEnvironment
     assert len(tx_for_address) == 1
     assert tx_for_address[0].to_puzzle_hash == ph_by_addr
 
+    # Test type filter
+    all_transactions = await client.get_transactions(
+        1, type_filter=TypeFilter.include([TransactionType.COINBASE_REWARD])
+    )
+    assert len(all_transactions) == 5
+    assert all_transactions[0].type == TransactionType.COINBASE_REWARD
```
xdustinface, line 647:
```suggestion
    assert all(transaction.type == TransactionType.COINBASE_REWARD for transaction in all_transactions)
```

chia/wallet/wallet_transaction_store.py
```diff
@@ -278,10 +307,18 @@ async def get_transactions_between(
         else:
             query_str = SortKey[sort_key].ascending()
 
+        if type_filter is None or len(type_filter.values) == 0:
```
xdustinface, line 310:
```suggestion
        if type_filter is None:
```

Adding the test for the empty `values` array showed that at least in my opinion this is wrong to skip this for empty lists because you then have the same behaviour for adding all types in there which seems not so intuitive, i think? Needs test adjustment also of course if you change it.

xdustinface (changes requested):

chia/wallet/util/query_filter.py
```diff
@@ -0,0 +1,65 @@
...
+from chia.util.streamable import Streamable, streamable
+from chia.wallet.util.transaction_type import TransactionType
+
+
+class FilterMode(IntEnum):
+    include = 1
+    exclude = 2
+
+
+class CoinRecordOrder(IntEnum):
```
xdustinface, line 18: Why did you also move this one into here? It's not really related to the filtering, is it?

xdustinface (changes requested):

xdustinface (approved):

trepca (approved):

PR #15199, opened 2023-05-04
Make conflicting_items a list instead of a set as we're only iterating over it, we're not leveraging set ownership

AmineKhaldi:
This simplifies using `can_replace()` and also prepares ground to simplify `check_removals()` next.

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -516,7 +516,7 @@ class MempoolManager:
 
         return None, potential, [item.name for item in conflicts]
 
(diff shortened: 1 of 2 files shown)


arvidn:
I don't think this is an obvious improvement. `check_removals()`, the function creating the set, relies on it being a set, right? it needs to de-duplicate conflicting items. It seems quite reasonable to return this set without conversion. All you do is to allocate a new list and copying all the members in there. There's no obvious gain, you can iterate over the set just as well, without copying it into a new list first.

emlowe:
closing old PR - can_replace now take a list

PR #15200, opened 2023-05-04
rpc: Hint `wallet_rpc_api.py`

xdustinface:
### Purpose:

More type hints.

### New Behavior:

No change in behaviour.

### Based on:

- #15362

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -260,7 +260,7 @@ class WalletRpcApi:
 
         return payloads
 
-    async def _stop_wallet(self):
+    async def _stop_wallet(self) -> None:
         """
         Stops a currently running wallet/key, which allows starting the wallet with a new key.
         Each key has it's own wallet database.
@@ -315,7 +315,7 @@ class WalletRpcApi:
     # Key management
     ##########################################################################################
 
-    async def log_in(self, request) -> EndpointResult:
+    async def log_in(self, request: Dict[str, Any]) -> EndpointResult:
         """
         Logs in the wallet with a specific key.
         """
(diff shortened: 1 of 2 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -2441,8 +2439,9 @@ async def nft_set_did_bulk(self, request):
 
             for tx in refined_tx_list:
                 await self.service.wallet_state_manager.add_pending_transaction(tx)
-            for coin in coin_ids:
-                await nft_wallet.update_coin_status(coin, True)
+            for coin_id in coin_ids:
+                # Ignore update_coin_status is not included in WalletProtocol
```
altendky, line 2443: I think this, and similar below, could be avoided with a `nft_wallet: NFTWallet` near the top and then using `WalletStateManager.get_wallet()`.  Other than the uncertainty around mypy not complaining that this may not even be assigned...

altendky (dismissed):

chia/rpc/wallet_rpc_api.py
altendky:
```suggestion
```

chia/rpc/wallet_rpc_api.py
```diff
@@ -2473,19 +2472,19 @@ async def nft_count_nfts(self, request) -> EndpointResult:
...
         try:
-            start_index = int(request.get("start_index"))
+            start_index = int(request.get("start_index", 0))
         except (TypeError, ValueError):
             start_index = 0
         try:
-            count = int(request.get("num"))
+            count = int(request.get("num", 50))
         except (TypeError, ValueError):
             count = 50
```
arvidn, line 2489: the try-catch around these seem redundant now. If they are, I think they should be removed. if they aren't, it might warrant a comment, since it really *looks* suspicious.

I think this change is for the better, because it allows the removal of the try-catch blocks. But it also doesn't seem to be a typing change

arvidn (commented):
looks good to me. one of Kyles comment still seems to be relevant, but afaict the remaining ones are addressed

chia/rpc/wallet_rpc_api.py
xdustinface: The statement will still fail if the value can't be converted to `int` so both, `TypeError` and `ValueError` are still possible? What is sad though is that we now have two places which define the default value but this is the least invasive change to not require coverage and there anyway should be a request schema which then can define the default in one place, which should be a different PR and be done basically for all RPC endpoints.

arvidn: if the value exists but cannot be interpreted as an `int`, that would seem to warrant a fatal error.

arvidn: but I suppose it was like this before you got here

arvidn (approved):

chia/rpc/wallet_rpc_api.py
xdustinface: > but I suppose it was like this before you got here

Yes, basically nothing changed here in this regard.

PR #15204, opened 2023-05-04
Softfork2 testing

arvidn:
* update tests to use the new softfork2 activation height.
* activate both soft forks immediately on testnet10


wjblanke (approved):
aok

PR #15208, opened 2023-05-04
Add basic test for `get_farmed_amount` wallet RPC

emlowe:
Adds the basic test.

### DIFF
--- a/tests/wallet/rpc/test_wallet_rpc.py
+++ b/tests/wallet/rpc/test_wallet_rpc.py
@@ -350,21 +350,21 @@ async def test_get_farmed_amount(wallet_rpc_environment: WalletRpcTestEnvironmen
     result = await wallet_rpc_client.get_farmed_amount()
 
     expected_result = {
-        "farmed_amount": 4000000000000,
-        "farmer_reward_amount": 500000000000,
+        "farmed_amount": 4_000_000_000_000,
+        "farmer_reward_amount": 500_000_000_000,
         "fee_amount": 0,
         "last_height_farmed": 2,
-        "pool_reward_amount": 3500000000000,
+        "pool_reward_amount": 3_500_000_000_000,
         "success": True,
     }
 
     assert result == expected_result
 
-    our_ph = await wallet.get_new_puzzlehash()
+    fee_amount = 100
     tx = await wallet.generate_signed_transaction(
         amount=uint64(5),
-        puzzle_hash=our_ph,
-        fee=uint64(100),
+        puzzle_hash=bytes32([0] * 32),
+        fee=uint64(fee_amount),
     )
     await wallet.push_transaction(tx)
 
@@ -373,7 +373,7 @@ async def test_get_farmed_amount(wallet_rpc_environment: WalletRpcTestEnvironmen
     await full_node_api.wait_for_wallet_synced(wallet_node=wallet_node, timeout=20)
 
     result = await wallet_rpc_client.get_farmed_amount()
-    assert result["fee_amount"] == 100
+    assert result["fee_amount"] == fee_amount
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -338,6 +338,44 @@ async def test_get_balance(wallet_rpc_environment: WalletRpcTestEnvironment):
...
+    our_ph = await wallet.get_new_puzzlehash()
+    tx = await wallet.generate_signed_transaction(
+        amount=uint64(5),
+        puzzle_hash=our_ph,
+        fee=uint64(100),
+    )
+    await wallet.push_transaction(tx)
+
+    await full_node_api.wait_transaction_records_entered_mempool(records=[tx])
+    await full_node_api.farm_blocks_to_puzzlehash(count=2, farm_to=our_ph, guarantee_transaction_blocks=True)
```
altendky, line 372: Ok, so I see why `.process_transaction_records()` won't work here.  Seems like a shortcoming though that it doesn't pass through the farming target puzzlehash.  That's for a separate PR, but would save the `count=2` kinda just-hav'ta-know-the-implementation-details stuff I think.

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -338,6 +338,44 @@ async def test_get_balance(wallet_rpc_environment: WalletRpcTestEnvironment):
...
+
+    assert result == expected_result
+
+    our_ph = await wallet.get_new_puzzlehash()
+    tx = await wallet.generate_signed_transaction(
+        amount=uint64(5),
+        puzzle_hash=our_ph,
+        fee=uint64(100),
+    )
+    await wallet.push_transaction(tx)
```
altendky, line 369: I don't think we care where this send goes?

Using `fee_amount` explicitly documents the relationship between the fee here and the fee amount below.

```suggestion
    fee_amount = 100
    tx = await wallet.generate_signed_transaction(
        amount=uint64(5),
        puzzle_hash=bytes32([0] * 32),
        fee=uint64(fee_amount),
    )
    await wallet.push_transaction(tx)
```

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -338,6 +338,44 @@ async def test_get_balance(wallet_rpc_environment: WalletRpcTestEnvironment):
...
+
+    await full_node_api.wait_transaction_records_entered_mempool(records=[tx])
+    await full_node_api.farm_blocks_to_puzzlehash(count=2, farm_to=our_ph, guarantee_transaction_blocks=True)
+    await full_node_api.wait_for_wallet_synced(wallet_node=wallet_node, timeout=20)
+
+    result = await wallet_rpc_client.get_farmed_amount()
+    assert result["fee_amount"] == 100
```
altendky, line 376:
```suggestion
    assert result["fee_amount"] == fee_amount
```

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -338,6 +338,44 @@ async def test_get_balance(wallet_rpc_environment: WalletRpcTestEnvironment):
...
+        "farmer_reward_amount": 500000000000,
+        "fee_amount": 0,
+        "last_height_farmed": 2,
+        "pool_reward_amount": 3500000000000,
+        "success": True,
+    }
+
+    assert result == expected_result
+
+    our_ph = await wallet.get_new_puzzlehash()
```
altendky, line 363: Is it important to not have this be two test runs with two independent results?  Since the fee amount is reported separately I think the fee case could actually not test the amounts at all and not be codependent.

altendky (changes requested):

tests/wallet/rpc/test_wallet_rpc.py
emlowe: Yea, makes some sense, so to clarify, the suggestion is an entirely second test

emlowe:
Close and reopen for benchmarks fixes

tests/wallet/rpc/test_wallet_rpc.py
altendky: I think so here, yeah.  At first I thought it could be parametrized, but I think it can be more isolated than that without drastic code duplication.  See what it looks like.

altendky (approved):

paninaro (approved):

PR #15212, opened 2023-05-05
data_layer: Avoid redundant coin state fetching

xdustinface:
### Purpose:

The first commit makes it always pass the height to `DataLayerWallet.new_launcher_spend` because we currently don't pass it after fetching the coin state and coin spend here https://github.com/Chia-Network/chia-blockchain/blob/16a204ccd0f7d792b4c1f8395f6c2b78cdb406e1/chia/data_layer/data_layer_wallet.py#L226-L228

which leads to fetching the coin state again here https://github.com/Chia-Network/chia-blockchain/blob/16a204ccd0f7d792b4c1f8395f6c2b78cdb406e1/chia/data_layer/data_layer_wallet.py#L237-L240

which is not required because we already know the height. 

The second commit merges `new_launcher_spend` it into `track_new_launcher_id` since its the only place where it's needed and there is not so much happening in both of them but i may drop this one if there are arguments for it.

### Current Behavior:

The coin state for the launcher id gets fetched twice.

### New Behavior:

Just one API call less for the same behaviour.

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -220,24 +220,23 @@ class DataLayerWallet:
         if await self.wallet_state_manager.dl_store.get_launcher(launcher_id) is not None:
             self.log.info(f"Spend of launcher {launcher_id} has already been processed")
             return None
-        if spend is not None and spend.coin.name() == launcher_id:  # spend.coin.name() == launcher_id is a sanity check
-            await self.new_launcher_spend(spend, peer, height)
-        else:
+        if spend is None or height is None:
             launcher_state: CoinState = await self.get_launcher_coin_state(launcher_id, peer)
-            launcher_spend = await fetch_coin_spend_for_coin_state(launcher_state, peer)
-            await self.new_launcher_spend(launcher_spend, peer)
+            spend = await fetch_coin_spend_for_coin_state(launcher_state, peer)
+            assert launcher_state.spent_height is not None
+            height = uint32(launcher_state.spent_height)
+
+        assert spend.coin.name() == launcher_id, "coin_id should always match the launcher_id here"
+
+        await self.new_launcher_spend(spend, peer, height)
 
     async def new_launcher_spend(
         self,
         launcher_spend: CoinSpend,
         peer: WSChiaConnection,
-        height: Optional[uint32] = None,
+        height: uint32,
     ) -> None:
         launcher_id: bytes32 = launcher_spend.coin.name()
-        if height is None:
-            coin_state = await self.get_launcher_coin_state(launcher_id, peer)
-            height = None if coin_state.spent_height is None else uint32(coin_state.spent_height)
-            assert height is not None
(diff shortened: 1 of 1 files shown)


arvidn (approved):

chia/data_layer/data_layer_wallet.py
```diff
@@ -220,27 +220,15 @@ async def track_new_launcher_id(
...
             return None
-        if spend is not None and spend.coin.name() == launcher_id:  # spend.coin.name() == launcher_id is a sanity check
-            await self.new_launcher_spend(spend, peer, height)
-        else:
+        if spend is None or height is None:
             launcher_state: CoinState = await self.get_launcher_coin_state(launcher_id, peer)
-            launcher_spend = await fetch_coin_spend_for_coin_state(launcher_state, peer)
-            await self.new_launcher_spend(launcher_spend, peer)
+            spend = await fetch_coin_spend_for_coin_state(launcher_state, peer)
+            assert launcher_state.spent_height is not None
```
emlowe, line 226: any thought to raising here?

chia/data_layer/data_layer_wallet.py
```diff
@@ -220,27 +220,15 @@ async def track_new_launcher_id(
...
-    ) -> None:
-        launcher_id: bytes32 = launcher_spend.coin.name()
-        if height is None:
-            coin_state = await self.get_launcher_coin_state(launcher_id, peer)
-            height = None if coin_state.spent_height is None else uint32(coin_state.spent_height)
-            assert height is not None
-        full_puzhash, amount, root, inner_puzhash = launch_solution_to_singleton_info(
-            launcher_spend.solution.to_program()
-        )
+        assert spend.coin.name() == launcher_id, "coin_id should always match the launcher_id here"
```
emlowe, line 229: or here?

emlowe (changes requested):

chia/data_layer/data_layer_wallet.py
arvidn: `assert` will throw an `AssertionFailed` exception. are you suggesting that this is an inappropriate exception because it might fail in some legitimate case?

emlowe: It doesn't matter much. I just think if something might happen we might as well raise something specific - I'm not much a fan of just "assert".

emlowe (approved):

chia/data_layer/data_layer_wallet.py
xdustinface: In this case `fetch_coin_spend_for_coin_state` would anyway already raise if its `None`. Its to make mypy happy only.

chia/data_layer/data_layer_wallet.py
xdustinface: Here `fetch_coin_spend_for_coin_state` already asserts this when we get the coin spends in here on in the call sites and the explicit exception raising implies to me more like its expected to trigger in some situations while this is not expected to trigger actually.

PR #15225, opened 2023-05-08
checkpoint: into main from release/1.8.0  @ 138154d5cd14c612036502bb2a7127fcdb2fe8c5

AmineKhaldi:
Source hash: 138154d5cd14c612036502bb2a7127fcdb2fe8c5
Remaining commits: 4


xdustinface (commented):
This should not be empty actually the change is wanted in `main` too imo. It will conflict, yes, but we can just add the additional columns which are in `main`  but not in 1.8.0 in here to the query i guess? 

If for some reason not, and we end up merging this as empty PR i can separately PR it into `main` tomorrow.

xdustinface (approved):

emlowe (approved):

PR #15240, opened 2023-05-09
Add **kwargs to all generate_signed_transaction definitions

Quexington:
The `generate_signed_transaction` method is _supposed_ to be a general function that you can call on any kind of wallet.  Ideally, all definitions of this function would have the same API, but in practice, it hasn't really shaken out that way. However, there are some instances (for example in the trade manager) where this method is called in a general way.

This PR adds some flexibility to the general call by defining `**kwargs` for every definition of the function.  This means that in places where a wallet is called without necessarily knowing the `generate_signed_transaction` API, the caller can pass in wallet specific kwargs that will be ignored by any other wallets that don't use those kwargs.  Right now, other wallets would thrown an `unknown keyword argument` error when they shouldn't care that they're getting an argument they are not going to use.  I'm working on a PR right now that would require this, but I've broken this change off into its own PR because it seems more generally applicable.


emlowe (dismissed):
Seems a generic function that isn't particularly generic

xdustinface (changes requested):
I don't think this is a good idea because it makes `mypy` not longer detect any issues about arguments passed into this method by name. Basically with this change you can pass in whatever you like and it doesn't complain.

xdustinface:
> > I don't think this is a good idea because it makes `mypy` not longer detect any issues about arguments passed into this method by name. Basically with this change you can pass in whatever you like and it doesn't complain.
> 
> It will still throw an error if you try to pass something bad into one of the arguments that _is_ defined, just not if you specify a kwarg that it's not going to use right?

Yeah right but it means it does not longer detect typos for existing arguments? In my opinion we should in general just don't use this `**kwargs: Any` pattern at all if its possible to avoid it _somehow_? Usually when i see it somewhere in the code it hits my eyes and i start trying to find ways to refactor it away :)

Quexington:
> > > I don't think this is a good idea because it makes `mypy` not longer detect any issues about arguments passed into this method by name. Basically with this change you can pass in whatever you like and it doesn't complain.
> > 
> > 
> > It will still throw an error if you try to pass something bad into one of the arguments that _is_ defined, just not if you specify a kwarg that it's not going to use right?
> 
> Yeah right but it means it does not longer detect typos for existing arguments? In my opinion we should in general just don't use this `**kwargs: Any` pattern at all if its possible to avoid it _somehow_? Usually when i see it somewhere in the code it hits my eyes and i start trying to find ways to refactor it away :)

Okay I haven't done this yet, but in the trade manager I'm going to eventually have to pass `dont_approve_crcats` or something similar to `generate_signed_transaction` and not have it fail on every other wallet trying to call it, so how do you suggest I resolve this?

xdustinface:
> Okay I haven't done this yet, but in the trade manager I'm going to eventually have to pass `dont_approve_crcats` or something similar to `generate_signed_transaction` and not have it fail on every other wallet trying to call it, so how do you suggest I resolve this?

Im not sure to be honest what to suggest here since i don't know the context or what the intention of this parameter is. The first i would try is to move this functionality somehow out of `generate_signed_transaction`.

Quexington:
> > Okay I haven't done this yet, but in the trade manager I'm going to eventually have to pass `dont_approve_crcats` or something similar to `generate_signed_transaction` and not have it fail on every other wallet trying to call it, so how do you suggest I resolve this?
> 
> Im not sure to be honest what to suggest here since i don't know the context or what the intention of this parameter is. The first i would try is to move this functionality somehow out of `generate_signed_transaction`.

This is the point, I would want to do _all_ of the things in `generate_signed_transaction` with a slight tweak in one specific case.  So I could remake `generate_signed_transaction` and duplicate all of that code, or I could be more reasonable and add an extra `if wallet_type == WalletType.blah` and duplicate the _call_ to `generate_signed_transaction` but that's going in the opposite direction that I want to go with _more_ code duplication and _more_ logical code paths where I'd like to move closer to being able to put `generate_signed_transaction` on `WalletProtocol` and eliminating different calls to different versions of it.

Quexington:
@xdustinface I went with a different approach that hopefully improves the type safety to a level that you're comfortable with.

xdustinface (commented):
Not sure what happens if this gets merged with the `mypy.ini` even though its gitignored but lets maybe not find out and just remove it from here? :) 

And, still not really a fan of this approach of "generalising" `generate_signed_transaction` and also not of the `kwargs` thing because we still pull out the extra args like `kwargs.get("origin_id", None)` which has no checks for the name and just return the default value you provide there for whatever input key.

I will approve if the `mypy.ini` was dropped because the parameter hinting at least seems to work, thanks! I can't judge about this `Unpack` thing though since i never saw it before so i guess it would be good to hear from @altendky here also.

Quexington:
> Not sure what happens if this gets merged with the `mypy.ini` even though its gitignored but lets maybe not find out and just remove it from here? :)

Yeah hadn't gotten around to it yet, but it should be fixed now :+1:

> And, still not really a fan of this approach of "generalising" `generate_signed_transaction` and also not of the `kwargs` thing because we still pull out the extra args like `kwargs.get("origin_id", None)` which has no checks for the name and just return the default value you provide there for whatever input key.

I'm not 100% positive, but I think even the `kwargs.get` is type safe because `kwargs` is a `TypedDict` so I would imagine mypy will throw an error if you're trying to access an element that doesn't exist within that typed dict.

altendky:
The `.gitignore` is only relevant to doing stuff like `git status` and `git add .`.  If the file is tracked then it being in `.gitignore` is (mostly?) irrelevant.  So yes, it should be removed.  Any intended associated changes should be moved to the 'new' location(s).

At first glance, this seems like a really complicated way to hint that we don't know what is going where.

> Right now, other wallets would thrown an `unknown keyword argument` error when they shouldn't care that they're getting an argument they are not going to use.

I'm not clear why writing code that thinks it is calling, say, a pool wallet method and passing a pool wallet specific parameter should be let to pass hint checks when it is actually calling, say, a CAT wallet.  That seems like a bad thing, not a good thing.

Maybe transaction signing needs to be broken down into smaller pieces that actually are common so they can meet a common interface?

If I'm going to be involved here I think a call might be worthwhile to get me ramped up on the situation.  For example, looking over the call site that will be able to always pass a, say, CAT wallet parameter and know what it should be, but would be unable to make a CAT wallet specific function call.

Quexington:
> I'm not clear why writing code that thinks it is calling, say, a pool wallet method and passing a pool wallet specific parameter should be let to pass hint checks when it is actually calling, say, a CAT wallet. That seems like a bad thing, not a good thing

Agreed, but you can opt in to preventing this on the call-side with an `assert isinstance(wallet, PoolWallet)` for example.  The problem I'm trying to solve here is when the call-side actually doesn't care what the type of the thing it's calling is (aside from being a `WalletProtocol` object which hopefully we can get GST to be a part of one day), but it _does_ care that _if_ it takes a certain argument, that argument should be set to X.

You can work around this by a bunch of `if/else this_type_of_wallet:` but I want there to be less of that in these general call sites (like the trade manager) and not more.

> Maybe transaction signing needs to be broken down into smaller pieces that actually are common so they can meet a common interface?

I'm not sure how much it can be broken down to accommodate all of the optional configurations that any wallet might take.  If you were to do this, I think you'd end up with like 10 tx functions like `modify payments -> select coins -> calculate change -> ...` etc. that you would have to call everywhere and risk missing one and ... I don't think that's a direction we want to go.

xdustinface (dismissed):
> I'm not 100% positive, but I think even the `kwargs.get` is type safe because `kwargs` is a `TypedDict` so I would imagine mypy will throw an error if you're trying to access an element that doesn't exist within that typed dict.

It doesn't. See  https://mypy-play.net/?mypy=latest&python=3.11&gist=98c5445070ab5bd885b572fb48674e99. It works only if you access via `["key"]`, not via `get("key")`.

Approved but still not happy with the direction we are going here with this. Might be worthwhile to dig into kyle's suggestion or really just try to stop squeezing even more logic into `get_signed_transaction` but instead start to factor things out there somehow / find other way to generalise the signing.

altendky:
More things in an interface isn't a good goal itself.  More _common_ things is.  As is we are describing that it isn't common but doing so in a complicated and misleading way.  I would be interested in seeing this drafted until we can have a call so I can better understand the needs and we can see what other ideas we can come up with and consider the data class idea I shared.

Quexington:
> > I'm not 100% positive, but I think even the `kwargs.get` is type safe because `kwargs` is a `TypedDict` so I would imagine mypy will throw an error if you're trying to access an element that doesn't exist within that typed dict.
> 
> It doesn't. See https://mypy-play.net/?mypy=latest&python=3.11&gist=98c5445070ab5bd885b572fb48674e99. It works only if you access via `["key"]`, not via `get("key")`.

If you hint `b_passing` with `: int` then it actually does fail, so it's not perfect but it's not entirely dumb either.

Quexington:
> And once I submitted that, I had an idea. So my basic concern here is that we're trying to hint inconsistency. What if instead we hinted the consistent part in a reusable fashion? Make a dataclass with an attribute for every consistent parameter and have the different methods accept an instance of that class. This avoids duplication in the function definitions and also avoids duplication at a call site that calls multiple methods with different parameters. Create the instance first, then pass that one thing in each different call. If there are other groups of parameters that are common to a couple methods, but not all, they could get a dataclass for just those as well to clarify that consistency.
> 
> Again, I know I don't have the full context here but figured I'd toss this out there.

The problem I have with this idea is that it still requires you to go make a change to the general code whenever you add a specific use case.  I'd like (for example) the TradeManager to ideally reach a state where it can just sit there and handle new wallets that conform with its protocol easily.  Part of the generic API that the trade manager supplies is that it may pass you extra arguments and you need to be able to ignore those.  I think you potentially have the roles backwards, probably because I explained it in a poor/misleading way.  What I'm suggesting in this PR I guess is a change to the protocol by which you ask a wallet for signed transaction where you can specify a bunch of configuration options that maybe only a few will understand.  The changes in this PR are changing all of the existing wallets to conform to this new protocol.

trepca (approved):

PR #15241, opened 2023-05-09
tweak to diff-cover for chia production code only

emlowe:
Use `--include 'chia/**'` for the `diff-cover` check 

This keeps the checking of coverage diffs limited to production code

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,18 +3,18 @@ name: 🧪 test
 on:
   push:
     paths-ignore:
-    - '**.md'
+      - "**.md"
     branches:
-    - long_lived/**
-    - main
-    - release/**
+      - long_lived/**
+      - main
+      - release/**
     tags:
-    - '**'
+      - "**"
   pull_request:
     paths-ignore:
-    - '**.md'
+      - "**.md"
     branches:
-    - '**'
+      - "**"
   workflow_dispatch: null
 
 concurrency:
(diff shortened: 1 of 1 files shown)


altendky (dismissed):
this seems bad, it at least needs discussed

arvidn:
my impression is that Kyle and I are fundamentally interested in different metrics, and perhaps we've both conflated them, assuming we're talking about the same metric.

I think the main objective of the test coverage effort should be (1) to report what production logic we have tests for, in order to highlight production logic that we're missing tests for.

My understanding of Kyle's objective is that (2) we should have as much as possibly of our python code be executed by an interpreter, in order to highlight code that is not used and can be removed, and also to highlight tests that may accidentally not be running.

I think both of those are worthy goals, perhaps we just disagree on their priority. I probably consider (2) lower priority because of my background in statically typed languages. I get the impression that, sort of, the interpreter executing a line of code is similar to it compiling.

As a separate dimension of what we include (or how we present) code coverage (which I mean as a generalization of test coverage) we also want to have CI enforce some rules on PRs.

The issue that sparked this PR was that we don't collect test coverage for benchmarks. If the objective is (2), we should to instrument coverage data for benchmarks (but perhaps by running them twice, to measure the timing of the uninstrumented run).

If the objective is (1), we definitely should not collect coverage data from benchmarks, because code executing as part of a benchmark does not indicate that we have a test for that code.

This PR is assuming that our objective is (1) by only considering coverage of production code and only by tests (not benchmarks).
I don't think (1) and (2) are necessarily mutually exclusive, but implementing both may require some more sophisticated coverage instrumentation.

PR #15243, opened 2023-05-09
Don't leverage SQLite's generated column feature for fee_per_cost. We'll hopefully reintroduce it again in the future

AmineKhaldi:
The goal is to have 100% coverage for `chia/full_node/mempool.py`.


altendky (commented):
I approve, but would like to make sure @arvidn gets a chance to look too.

arvidn (dismissed):

arvidn (approved):

PR #1487, opened 2021-03-24
V1.0.x

justinengland:
PR prior to release changes merge back to main

PR #15249, opened 2023-05-09
Remove unused alert server test code and support files

emlowe:
Remove unused support files `create_alert_file.py` and `validate_alert.py`, along with unused test code.

Half of the test code for the alerts was removed back in #1487, this removes the rest.

All the code that actually used these alerts in the daemon was also removed in #1487, so even if we wanted to re-use this, the existing versions of Chia in the wild will never use the alerts. So I think this is all dead code we might as well just remove.

Also removed relevant files from mypy.ini


hoffmang9 (dismissed):
Ah the flag drop...

xdustinface (changes requested):
It's also one of the excluded files in `mypy.ini`, lets drop it there too then :)

hoffmang9 (approved):
lgtm

emlowe:
close and reopen for CI

xdustinface (approved):

wjblanke (approved):
aok

wjblanke (approved):
aok

PR #15256, opened 2023-05-10
CHIP13 draft

fchirica:
Draft implementation for CHIP-0013: https://github.com/Chia-Network/chips/blob/6a6351d9625f10ee8306bb0af98bd94ab0072d9b/CHIPs/chip-0013.md

### DIFF
--- a/chia/consensus/multiprocess_validation.py
+++ b/chia/consensus/multiprocess_validation.py
@@ -21,7 +21,7 @@ from chia.consensus.pot_iterations import calculate_iterations_quality, is_overf
 from chia.full_node.mempool_check_conditions import get_name_puzzle_conditions
 from chia.types.block_protocol import BlockInfo
 from chia.types.blockchain_format.coin import Coin
-from chia.types.blockchain_format.proof_of_space import verify_and_get_quality_string
+from chia.types.blockchain_format.proof_of_space import get_plot_id, verify_and_get_quality_string
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.blockchain_format.sub_epoch_summary import SubEpochSummary
 from chia.types.full_block import FullBlock
@@ -201,6 +201,7 @@ async def pre_validate_blocks_multiprocessing(
     recent_blocks_compressed: Dict[bytes32, BlockRecord] = {}
     num_sub_slots_found = 0
     num_blocks_seen = 0
+    recent_plot_ids: List[bytes32] = []
     if blocks[0].height > 0:
         if not block_records.contains_block(blocks[0].prev_header_hash):
             return [PreValidationResult(uint16(Err.INVALID_PREV_BLOCK_HASH.value), None, None, False)]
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -138,7 +138,7 @@ def softfork_height(request) -> int:
     return request.param
 
 
-saved_blocks_version = "rc5"
+saved_blocks_version = "2.0"
```
arvidn, line 141: are we changing the format of something here? like FullBlocks?

chia/consensus/multiprocess_validation.py
```diff
@@ -220,13 +223,19 @@ async def pre_validate_blocks_multiprocessing(
             recent_blocks[curr.header_hash] = curr
             if curr.is_transaction_block:
                 num_blocks_seen += 1
+            if len(recent_plot_ids) < 4:
+                # TODO: Add `get_proof_of_space()` to BlockStore.
+                full_block = await block_store.get_full_block(curr.header_hash)
+                assert full_block is not None
+                recent_plot_ids.append(get_plot_id(full_block.reward_chain_block.proof_of_space))
```
arvidn, line 230: this is not sensitive to reorgs, right? we just look up the previous blocks every time we validate one, yeah?

chia/consensus/multiprocess_validation.py
```diff
@@ -220,13 +223,19 @@ async def pre_validate_blocks_multiprocessing(
             recent_blocks[curr.header_hash] = curr
             if curr.is_transaction_block:
                 num_blocks_seen += 1
+            if len(recent_plot_ids) < 4:
+                # TODO: Add `get_proof_of_space()` to BlockStore.
+                full_block = await block_store.get_full_block(curr.header_hash)
```
arvidn, line 228: this lookup has the potential of being quite expensive. I think it would make sense to benchmark syncing with this patch

arvidn (commented):
you need to only enable this checking past the soft-fork activation height also

arvidn:
don't forget to add tests exercising this also

tests/conftest.py
fchirica: `BlockTools` produce with high probability blocks failing the new plot_id condition (hence making tests fail). Working to rebuild the block cache with the new `BlockTools` from this PR.

chia/consensus/multiprocess_validation.py
fchirica: `BlockTools` stores orphans too. So if the reorg is based off an orphan, we should still find the hash.

chia/consensus/multiprocess_validation.py
fchirica: Thinking to extract here just the `ProofOfSpace`, not the full block (and add to `BlockStore` a new SQL query that selects only that). But this is still WIP, trying to get all tests pass first.

almogdepaz: if you only need the plot_id we can create a cache for that instead of going to the disk 4 times on each block validation

fchirica:
closing since I'll create a new one redoing this in `block_header_validation `

PR #15257, opened 2023-05-10
Changed: output plotid when we find an invalid proof of space

wjblanke:
### Purpose:

Let user know which plot has issues

### Current Behavior:

Bad proof of space doesn't give any hint about which plot

### New Behavior:

User knows plotid of bad plot

### Testing Notes:

n/a


hoffmang9 (dismissed):
Nice

emlowe:
Code coverage failure is because we don't have tests that exercise this error path - we only have tests designed for network serialization of this protocol message

emlowe (approved):

PR #15264, opened 2023-05-12
don't add transactions to the mempool before it has a valid peak

### LINKED ISSUE #15217: [Bug] ERROR   Error in _handle_one_transaction, closing
### What happened?

If a node needs to sync it falls in a loop:
1) it connects to a remote node to request blocks, 
2) it starts receiving transactions to mempool from the remote node and calls `_handle_one_transaction`
3) there is no peak yet because syncing is in progress, it triggers assert in `pre_validate_spendbundle`,
4) assert exception is caught and remote node is banned before it is able to send any block.

Relevant #14236 #14274

### Version

1.7.0+

Linux

CLI

### Relevant log output

```shell
File "/chia-blockchain/chia/full_node/full_node.py", in _handle_one_transaction
File "/chia-blockchain/chia/full_node/full_node.py", in add_transaction
  cost_result = await self.mempool_manager.pre_validate_spendbundle(...)
File "/chia-blockchain/chia/full_node/mempool_manager.py", in pre_validate_spendbundle
  assert self.peak is not None
AssertionError
```

arvidn:
As part of the assert-before feature, a new precondition was added to the mempool transaction ingestion. It now needs a peak block height to know whether to activate the soft-fork logic or not.

Sometimes new transactions are received *before* we have a peak block, which now trows an assertion failure exception.

This patch treats this case the same as if we're in sync-mode, and won't forward transactions to the mempool.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -2211,8 +2211,9 @@ class FullNode:
             return MempoolInclusionStatus.FAILED, Err.ALREADY_INCLUDING_TRANSACTION
         self.mempool_manager.add_and_maybe_pop_seen(spend_name)
         self.log.debug(f"Processing transaction: {spend_name}")
-        # Ignore if syncing
-        if self.sync_store.get_sync_mode():
+        # Ignore if syncing or if we have not yet received a block
+        # the mempool must have a peak to validate transactions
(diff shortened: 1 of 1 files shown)


emlowe (approved):

wjblanke (approved):
aok

PR #15274, opened 2023-05-13
wallet: Fix and improve hint parsing

xdustinface:
### Purpose:

- Add  `hinted_coin_from_condition` to get a `HintedCoin` from a `Program` and a parent_id if the conditions match.
- Add `hinted_coins_in_coin_spend` which yields a `HintedCoin` after each processed condition of the passed `CoinSpend`.
- Add `hinted_coin_with_coin_id` which returns a `HintedCoin` for a given coin_id if it exists in the passed `CoinSpend` and raises if not.

with `HintedCoin` being: 
```python
@dataclass(frozen=True)
class HintedCoin:
    coin: Coin
    hint: bytes32
```
Alternative for #15259.

### Testing Notes:

Added unit tests for the new/refactored methods.

### DIFF
--- a/chia/wallet/util/compute_hints.py
+++ b/chia/wallet/util/compute_hints.py
@@ -1,22 +1,54 @@
 from __future__ import annotations
 
-from typing import List
+from dataclasses import dataclass
+from typing import Iterator
 
-from chia.types.blockchain_format.program import INFINITE_COST
+from clvm.casts import int_from_bytes
+
+from chia.types.blockchain_format.coin import Coin
+from chia.types.blockchain_format.program import INFINITE_COST, Program
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.coin_spend import CoinSpend
 from chia.types.condition_opcodes import ConditionOpcode
+from chia.util.ints import uint64
 
 
-def compute_coin_hints(cs: CoinSpend) -> List[bytes32]:
-    _, result_program = cs.puzzle_reveal.run_with_cost(INFINITE_COST, cs.solution)
+@dataclass(frozen=True)
+class HintedCoin:
+    coin: Coin
+    hint: bytes32
 
-    h_list: List[bytes32] = []
-    for condition_data in result_program.as_python():
-        condition = condition_data[0]
-        args = condition_data[1:]
-        if condition == ConditionOpcode.CREATE_COIN and len(args) > 2:
-            if isinstance(args[2], list):
-                if isinstance(args[2][0], bytes):
-                    h_list.append(bytes32(args[2][0]))
-    return h_list
+
+def hinted_coin_from_condition(condition: Program, parent_coin_id: bytes32) -> HintedCoin:
+    data = condition.as_python()
+    if len(data) != 4 or data[0] != ConditionOpcode.CREATE_COIN:
+        raise ValueError(f"Not a hinted coin: {data}")
+
+    try:
+        coin = Coin(parent_coin_id, bytes32(data[1]), uint64(int_from_bytes(data[2])))
+    except Exception as e:
+        raise ValueError(f"Invalid coin: {e}, data: {data[1:2]}") from e
+
+    try:
+        hint = bytes32(data[3][0])
+    except Exception as e:
+        raise ValueError(f"Invalid hint: {e}, data: {data[3:]}") from e
+
+    return HintedCoin(coin, hint)
+
+
+def hinted_coins_in_coin_spend(coin_spend: CoinSpend) -> Iterator[HintedCoin]:
+    _, result_program = coin_spend.puzzle_reveal.run_with_cost(INFINITE_COST, coin_spend.solution)
+    parent_coin_id = coin_spend.coin.name()
+    for condition in result_program.as_iter():
+        try:
+            yield hinted_coin_from_condition(condition, parent_coin_id)
+        except Exception:
+            continue
+
+
+def hinted_coin_with_coin_id(coin_spend: CoinSpend, coin_id: bytes32) -> HintedCoin:
+    for hinted_coin in hinted_coins_in_coin_spend(coin_spend):
+        if hinted_coin.coin.name() == coin_id:
+            return hinted_coin
+    raise ValueError(f"coin_id {coin_id} not found in coin_spend: {coin_spend}")
--- a/chia/wallet/util/compute_hints.py
+++ b/chia/wallet/util/compute_hints.py
@@ -10,6 +10,7 @@ from chia.types.blockchain_format.program import INFINITE_COST, Program
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.coin_spend import CoinSpend
 from chia.types.condition_opcodes import ConditionOpcode
+from chia.util.errors import ParseError
 from chia.util.ints import uint64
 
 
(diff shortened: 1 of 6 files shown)


chia/wallet/util/compute_hints.py
```diff
@@ -1,22 +1,54 @@
...
+
+
+@dataclass(frozen=True)
+class HintedCoin:
+    coin: Coin
+    hint: bytes32
+
+
+def hinted_coin_from_condition(condition: Program, parent_coin_id: bytes32) -> HintedCoin:
+    data = condition.as_python()
```
xdustinface, line 23: @Quexington In your PR you moved away from using `Program.as_python` towards using `Program.at`. Im not sure if there was a deeper meaning to this and im not very familiar with all this `Program` stuff.. so, let me know.

xdustinface:
Tests failing because we don't hint newly created CAT coins. Back to draft for now.

tests/wallet/test_util.py
```diff
@@ -0,0 +1,116 @@
...
+from chia.wallet.util.compute_hints import (
+    HintedCoin,
+    hinted_coin_from_condition,
+    hinted_coin_with_coin_id,
+    hinted_coins_in_coin_spend,
+)
+
+
+def random_coin() -> Coin:
+    return Coin(bytes32(token_bytes(32)), token_bytes(32), uint64.from_bytes(token_bytes(8)))
```
Quexington, line 25: I think randomness in tests is a bad idea, even if the idea is that everything _should_ pass. We used to have randomness in the wallet tests and it resulted in so much flakiness that wasn't really bugs so much as bad compounding randomness.  Also if we decide to test a specific edge case, this makes it harder and likely has to be done anyways.

chia/wallet/wallet_state_manager.py
```diff
@@ -742,12 +742,8 @@ async def handle_cat(
         """
         mod_hash, tail_hash, inner_puzzle = curried_args
 
-        hint_list = compute_coin_hints(coin_spend)
-        derivation_record = None
-        for hint in hint_list:
-            derivation_record = await self.puzzle_store.get_derivation_record_for_puzzle_hash(bytes32(hint))
-            if derivation_record is not None:
-                break
+        hinted_coin = hinted_coin_with_coin_id(coin_spend, coin_state.coin.name())
```
Quexington, line 745: As I think you've commented and figured out, there's a chance that there's no hint for the coin in here in which case we want to set the derivation record to None and continue on.

chia/wallet/util/compute_hints.py
```diff
@@ -1,22 +1,54 @@
...
+    return HintedCoin(coin, hint)
+
 
+def hinted_coins_in_coin_spend(coin_spend: CoinSpend) -> Iterator[HintedCoin]:
+    _, result_program = coin_spend.puzzle_reveal.run_with_cost(INFINITE_COST, coin_spend.solution)
+    parent_coin_id = coin_spend.coin.name()
+    for condition in result_program.as_iter():
+        try:
+            yield hinted_coin_from_condition(condition, parent_coin_id)
+        except Exception:
```
Quexington, line 46: It seems like you only throw value error from this function so maybe just catch that.  Also, perhaps it would be good to throw a specific exception because I imagine the caller would not like to accidentally consume an exception from an unexpectedly malformatted condition of some sort.

Quexington (changes requested):

tests/wallet/test_util.py
xdustinface: Added 2c249db828f225680769f4e738ebf6769644c2fd.

chia/wallet/wallet_state_manager.py
Quexington: We don't need to do any cleanup for unhinted coins for now.  We're picking up both hinted and unhinted just fine for a while.  That doesn't matter for this function though.  This function will not be hit except in the case where we do not already have a CAT wallet for an existing CAT in which case a hint is necessary to get our attention.  The segment of code below here does the proper thing, if it sees an unhinted CAT, it likely was not supposed to come here, so it's logging the abnormality as such and then returning None, as you would expect it to do.

chia/wallet/util/compute_hints.py
Quexington: I liked the behavior before because it's sort of a sanity check on the data we're processing.  This method is for processing coin spends, coins spends should only output valid conditions as part of their definition, so running this on something that doesn't should fail.  At the very least, as this PR sits, it would make the code less sane without adding anything to replace the functionality of checking that we're only processing valid coin spends.

chia/wallet/wallet_state_manager.py
xdustinface: > That doesn't matter for this function though. This function will not be hit except in the case where we do not already have a CAT wallet for an existing CAT in which case a hint is necessary to get our attention.

Yeah but that case will most likely exist out there and we should take care of it, or do i miss something now? To me it seems like there should be a PR which:

- Adds the missing hinting to the CAT creation code
- Somehow fixes and tests picking up those unhinted coins if there is not yet a wallet 

 The segment of code below here does the proper thing, if it sees an unhinted CAT, it likely was not supposed to come here, so it's logging the abnormality as such and then returning None, as you would expect it to do.

If it sees an unhinted CAT in this place it's definitely an issue, isn't it? Like, a case where we missed to hint the coin.

Quexington: > Adds the missing hinting to the CAT creation code

This PR exists https://github.com/Chia-Network/chia-blockchain/pull/14356, but should not matter for this PR

> Somehow fixes and tests picking up those unhinted coins if there is not yet a wallet

We don't _want_ unhinted coins to trigger wallet creation.  We only want wallet creation to occur upon receiving a hinted coin.  This is how it has always worked (in production, the function you're referring to is only used for tests), so there's no need for backwards compatibility with unhinted stuff.

> If it sees an unhinted CAT in this place it's definitely an issue, isn't it? Like, a case where we missed to hint the coin.

No, exactly not, we only want to pick up hinted coins here.

xdustinface: > > Adds the missing hinting to the CAT creation code
> 
> This PR exists #14356, but should not matter for this PR
 
There is fix for the issue in that PR (which still not need to be in there imo and could already be merged independently) yes but nothing which makes sure the coins out in the wild will be picked up properly. Im not sure if either you didn't yet follow what i mean so i will try to express it again:

Imagine you created a CAT in the past which sits in an unhinted coin now, syncing the related wallet will not be able to process the coin because it will just bail out with a "doesn't belong to our wallet" message. How can a coin like this become processed? From my understanding it would never be able to create a wallet. Maybe there is some way to work around this by manually creating the wallet manually or so, i don't know. But this is what i mean with "we hide the actual issue". 

> No, exactly not, we only want to pick up hinted coins here.

And of course, we don't *want* to pick up unhinted coins here.. usually.. but if we had an issue in earlier versions of the code and there actually are valid but unhinted coins like this in existence we can't just ignore them, can we?

And im not necessarily saying we should process or pick up the unhinted coin here in this place although it might be an option to do some checks for it.. but once we come here we already passed `match_cat_puzzle` and we know this coin is a CAT, right? So.. if also the hint is missing, while its supposed to be there if we come here, it means, from my understanding, this case is an issue and should not just silently fail with "not our coin".

Quexington: This is what I'm saying. There was never an issue in an earlier version of the code where _minted_ CATs would not have a hint.  The supported ways to mint have always had a hint.  The function that is failing in the test is a function whose use is limited to tests.  It is not a supported way to mint real CATs.

chia/wallet/util/compute_hints.py
xdustinface: Okay agree, probably better to let other stuff still raise, added [25185d9](https://github.com/Chia-Network/chia-blockchain/pull/15274/commits/25185d93a7762ed7c6308379d6e1d24babf0038e).

PR #15286, opened 2023-05-15
Pass VC fee parameter all the way through and check balances in tests

Quexington:
The fee for VC minting wasn't being properly passed through to the drivers from the wallet.  This PR fixes that as well as adds better balance checks throughout the vc wallet tests.


trepca (approved):

Quexington:
> this looks like it might be a typo: https://github.com/Chia-Network/chia-blockchain/actions/runs/4982911345/jobs/8922520892?pr=15286#step:17:55

That error actually shows up on pretty much every wallet test, I've kinda just forgotten it.  The line below it is actually evidence of an error though which I will fix in a separate PR (link soon) making the test flaky.

PR #15289, opened 2023-05-15
Fix flaky VC test

Quexington:
Occasionally, the revocation was happening before the next VC was confirmed on chain.


wallentx:
Yes yes yes yes

trepca (approved):

PR #15312, opened 2023-05-17
Create WalletSingletonStore for storing singleton data across wallets

geoffwalmsley:
### Purpose:
- We have several wallets that track singletons:  Pool, NFT, DAO, DL, and more will probably be added in future. Currently each wallet has its own local store for the singletons it tracks. 
- This PR is a starting point to have singleton wallets use one store for tracking coins. 
- Initially this will be used for DAO wallets, followed by the Pool wallet, and then others can be migrated.


chia/wallet/singleton.py
```diff
@@ -30,6 +30,22 @@ def get_inner_puzzle_from_singleton(puzzle: Program) -> Optional[Program]:
     return Program(INNER_PUZZLE)
 
 
+def get_singleton_id_from_singleton(puzzle: Program) -> Optional[bytes32]:
```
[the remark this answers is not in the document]
aqk: I am pleased that you are using types to represent more than the structure of data. I've updated this branch to use `uncurry_puzzle`.

chia/wallet/singleton.py
```diff
@@ -30,6 +30,22 @@ def get_inner_puzzle_from_singleton(puzzle: Program) -> Optional[Program]:
...
+    :param puzzle: Singleton puzzle
+    :return: Singleton ID
+    """
+    r = puzzle.uncurry()
+    if r is None:
+        return None
+    inner_f, args = r
+    if not is_singleton(inner_f):
+        return None
+    SINGLETON_STRUCT, INNER_PUZZLE = list(args.as_iter())
```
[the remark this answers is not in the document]
aqk: done.

chia/wallet/singleton_coin_record.py
```diff
@@ -0,0 +1,33 @@
...
+@dataclass(frozen=True)
+class SingletonCoinRecord:
+    """
+    These are values that correspond to a singleton in the WalletSingletonStore
+    """
+
+    coin: Coin
+    singleton_id: bytes32
+    wallet_id: uint32
+    parent_coin_spend: CoinSpend
```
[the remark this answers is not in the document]
aqk: mmm, good point. Would there be value in storing the full chain of singleton states with the SINGLETON_MOD wrapper taken off? I can't think of a case where we need to get at old states quickly currently.

chia/wallet/singleton_coin_record.py
```diff
@@ -0,0 +1,33 @@
...
+    inner_puzzle_hash: Optional[bytes32]
+    confirmed: bool
+    confirmed_at_height: uint32
+    spent_height: uint32
+    lineage_proof: LineageProof
+    custom_data: Optional[Any]
```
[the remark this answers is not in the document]
aqk: I agree that we should lock down this type.

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,235 @@
...
+    async def create(cls: Type[_T_WalletSingletonStore], wrapper: DBWrapper2) -> _T_WalletSingletonStore:
+        self = cls()
+        self.db_wrapper = wrapper
+
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
+            await conn.execute(
+                (
+                    "CREATE TABLE IF NOT EXISTS singletons("
+                    "coin_id blob PRIMARY KEY,"
+                    " coin text,"
```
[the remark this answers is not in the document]
aqk: Yes, let's break coin out into its parts.

chia/wallet/singleton_coin_record.py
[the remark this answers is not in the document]
Quexington: I could get behind storing only the `INNER_PUZZLE` actually yeah.  It's still probably redundant in most cases but that does give us some options moving forward.

chia/wallet/wallet_singleton_store.py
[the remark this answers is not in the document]
Quexington: I realized `coin_id` isn't actually redundant with `parent_coin_info` like I thought, but you'd still like to be able to search on those properties so I'm still in favor of splitting those out.

chia/wallet/singleton_coin_record.py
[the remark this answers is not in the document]
aqk: I changed it to `Dict[str, bytes]` because it is more general. We can always deserialize a `List` from bytes.

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,235 @@
...
+        Store SingletonCoinRecord in DB
+        """
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
+            columns = (
+                "coin_id, coin, singleton_id, wallet_id, parent_coin_spend, inner_puzzle_hash, "
+                "confirmed, confirmed_at_height, spent_height, lineage_proof, custom_data, "
+                "generation, timestamp"
+            )
+            await conn.execute(
+                f"INSERT or REPLACE INTO singletons ({columns}) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)",
```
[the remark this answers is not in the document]
aqk: Good catch. I think we can move this to "column name" syntax. That will be even better.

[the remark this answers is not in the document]
aqk: done.

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,235 @@
...
+                    record.timestamp,
+                ),
+            )
+
+    async def get_records_by_singleton_id(
+        self,
+        singleton_id: bytes32,
+        min_generation: Optional[uint32] = None,
+        max_generation: Optional[uint32] = None,
+        num_results: Optional[uint32] = None,
```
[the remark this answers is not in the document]
aqk: It's a good idea, but the newly introduced problem of possibly introducing a bug because of unexpected behavior (truncated results) seems worse to me.

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,235 @@
...
+                    " coin text,"
+                    " singleton_id blob,"
+                    " wallet_id int,"
+                    " parent_coin_spend blob,"
+                    " inner_puzzle_hash blob,"
+                    " confirmed tinyint,"
+                    " confirmed_at_height int,"
+                    " spent_height int,"
+                    " lineage_proof blob,"
+                    " custom_data blob,"
```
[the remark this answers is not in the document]
aqk: That's a really good idea.

chia/wallet/singleton_coin_record.py
```diff
@@ -0,0 +1,33 @@
...
+    confirmed: bool
+    confirmed_at_height: uint32
+    spent_height: uint32
+    lineage_proof: LineageProof
+    custom_data: Optional[Dict[str, Any]]
+    generation: uint32
+    timestamp: uint64
+
+    def name(self) -> bytes32:
+        return self.coin.name()
```
Quexington, line 33: I'm not sure how much this saves in terms of code space, but it does add a little bit of confusion.  When I first read `record.name()` below, I assumed it was like a hash of all of the pieces of data.  Maybe it doesn't need to change but that's my opinion at least.

chia/wallet/singleton_coin_record.py
```diff
@@ -0,0 +1,33 @@
...
+    coin: Coin
+    singleton_id: bytes32
+    wallet_id: uint32
+    inner_puzzle: Program
+    inner_puzzle_hash: bytes32
+    confirmed: bool
+    confirmed_at_height: uint32
+    spent_height: uint32
+    lineage_proof: LineageProof
+    custom_data: Optional[Dict[str, Any]]
```
Quexington, line 28: The values here are going to be stored in the DB as blobs anyway, so can we make this `Dict[str, bytes]`?

Quexington (commented):
just one or two more comments

chia/wallet/singleton_coin_record.py
geoffwalmsley: I'll make it `record.coin_id()` so there's no confusion

Quexington (approved):
Looks good to me, I might ping @arvidn for an opinion on the relatively complicated sql stuff we're doing here and if it could be done better, but it looks well tested to me so even if it's not perfectly optimal (which I have no reason to believe it isn't), it should work fine.

aqk:
Hey, what's holding up this PR?

@trepca @arvidn

arvidn:
is test coverage measured for this PR?

arvidn (commented):
It seems we already store duplicate data in the wallet database, and this causes more duplication. If we're serious about transitioning existing tables over to this, combined, table, we would need logic to actually perform the transition. Without that, this patch (alone) makes things worse.

Hence my comment about this should be a commit in a patch-set that actually improves the situation.

I also have some doubts that the unit tests actually cover all the interesting edge cases. I would expect the test coverage metrics would expose some of that, but I'm not sure if coverage was run on this PR

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,281 @@
...
+    async def add_singleton_record(self, record: SingletonCoinRecord) -> None:
+        """
+        Store SingletonCoinRecord in DB
+        """
+        async with self.db_wrapper.writer_maybe_transaction() as conn:
+            columns = (
+                "coin_id, parent_coin_info, puzzle_hash, amount, singleton_id, wallet_id, "
+                "inner_puzzle, inner_puzzle_hash, confirmed, confirmed_at_height, spent_height, "
+                "lineage_proof, generation, timestamp"
+            )
```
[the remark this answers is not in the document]
aqk: Doesn't make a difference in this case, but in general it's better to use tuples when there is a choice - they are immutable.

aqk:
> is test coverage measured for this PR?

Yes - 100%

aqk:
> I think it's very likely that this table will need tweaks, and improvements, once it's actually used. I think it would be a good idea to make this part of a patch-set where you also actually use it for DAO and Pool wallets. Until you actually have a concrete use case, you're likely to not quite get the features right

You are totally correct - that's why we developed it in the `dao_cats_rebased` branch. It is used there if you want to take a look.

We broke this PR out at my suggestion, because:

* Merging in the whole DAO Wallet change at once seem untenable
* I can't see a place to make the cut where we are both using this from the DAO wallet and not bringing in the whole DAO Wallet code

aqk:
> It seems we already store duplicate data in the wallet database, and this causes more duplication. If we're serious about transitioning existing tables over to this, combined, table, we would need logic to actually perform the transition. Without that, this patch (alone) makes things worse.

It does make it worse, in the sense that there will be yet another Singleton-wallet specific table for this new Singleton wallet. This one makes an attempt at making it easy for the NEXT Singleton wallet to reuse this code.

> Hence my comment about this should be a commit in a patch-set that actually improves the situation.

I apologise - we may not have explained the history of this patch thoroughly.

The purpose of the creation of WalletSingletonStore is to introduce a general Singleton Wallet table, but the purpose of this PR is to start to bring the DAO Wallet code into main in a controlled, sensible way.

If the goal of unifying the Singleton wallet tables were more important than shipping the DAO wallet, I would agree with you, but I don't think that is the case.

Some form of this table is going to be in a PR into main, either in this form, or as yet another Singleton Wallet specific table.

> I also have some doubts that the unit tests actually cover all the interesting edge cases. I would expect the test coverage metrics would expose some of that, but I'm not sure if coverage was run on this PR

Now we're talking. This is some feedback we can look into. I wonder why coverage didn't run on this.
@geoffwalmsley can you look into the coverage issue?

geoffwalmsley:
I think this is the coverage report:
https://github.com/Chia-Network/chia-blockchain/actions/runs/5108358972/jobs/9182458063?pr=15312#step:11:17

aqk:
> is test coverage measured for this PR?


@arvidn Geoff found it here:
https://github.com/Chia-Network/chia-blockchain/actions/runs/5108358972/jobs/9182458063?pr=15312#step:11:17

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,281 @@
...
+                "SELECT * from singletons WHERE singleton_id=? "
+                f"{'AND generation >=? ' if min_generation is not None else ''}"
+                f"{'AND generation <=? ' if max_generation is not None else ''}"
+                "ORDER BY generation DESC"
+                f"{' LIMIT ?' if num_results is not None else ''}",
+                tuple(query_params),
+            )
+        records = []
+        for row in rows:
+            custom_data = await self.get_custom_data_by_coin_id(row[0])
```
trepca, line 179: why not join these tables vs executing a separate SQL for each row?

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,281 @@
...
+        await self.add_singleton_record(
+            dataclasses.replace(current, confirmed=True, confirmed_at_height=confirmed_height, timestamp=timestamp)
+        )
+
+        # update the parent record spent_height
+        parent_id = current.coin.parent_coin_info
+        parent: Optional[SingletonCoinRecord] = await self.get_record_by_coin_id(parent_id)
+        if parent is None:
+            return
+        await self.add_singleton_record(dataclasses.replace(parent, spent_height=confirmed_height))
```
trepca, line 255: why do we update the singleton twice and not just one time setting both fields?
why do we use INSERT/REPLACE and not UPDATE so only those fields can be set, not overriding all each time?

arvidn:
> It does make it worse, in the sense that there will be yet another Singleton-wallet specific table for this new Singleton wallet. This one makes an attempt at making it easy for the NEXT Singleton wallet to reuse this code.

I was implying that this patch only makes sense once we want to land the next Singleton wallet (or when we want to refactor an existing singleton wallet). With the idea that `main` is ready to be released at any time, this PR should only land in `main` once it has a purpose. I think there are reasons to have some wiggle room with this rule of thumb, but when it actually comes with a cost, it seems like we should wait until it is combined with some savings.

Anyway, that's how I look at things.

arvidn:
I found this: https://coveralls.io/builds/60372484
and it was surprisingly hard.

tests/wallet/test_singleton_store.py
```diff
@@ -0,0 +1,279 @@
...
+        assert recs == []
+        record = get_new_singleton_record()
+        recs = await db.get_records_by_singleton_id(record.singleton_id)
+        assert len(recs) == 0
+        recs = await db.get_records_by_singleton_id(record.singleton_id, one, one, one)
+        assert len(recs) == 0
+        await db.add_singleton_record(record)
+        recs = await db.get_records_by_singleton_id(record.singleton_id)
+        assert len(recs) == 1
+        recs = await db.get_records_by_singleton_id(record.singleton_id, one, one, one)
```
arvidn, line 154: this function has 3 optional constraints.
```
        min_generation: Optional[uint32] = None,
        max_generation: Optional[uint32] = None,
        num_results: Optional[uint32] = None,
```

as far as I can tell, there are no tests ensuring these work as intended.

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,281 @@
...
+
+    async def set_spent(self, coin_id: bytes32, spent_height: uint32, timestamp: uint64) -> None:
+        """
+        Updates a confirmed record with a spent_height
+        """
+        current: Optional[SingletonCoinRecord] = await self.get_record_by_coin_id(coin_id)
```
arvidn, line 261: it's *very* slow to get the record, update it and then store it back. Just deserializing and serializing the class is probably a material cost.

I would think we would want a single SQL statement that updates these fields, specifically and then is done.

The same comment applies to `set_confirmedI()`

tests/wallet/test_singleton_store.py
```diff
@@ -0,0 +1,279 @@
...
+        assert unconfirmed[0] == recs[0]
+
+        # confirm the new record
+        await db.set_confirmed(next_record.coin_id(), uint32(20), uint64(time.time()))
+        recs = await db.get_records_by_singleton_id(record.singleton_id)
+        assert recs[0].confirmed
+        assert recs[1].spent_height == recs[0].confirmed_at_height == uint32(20)
+
+        # Delete the last record
+        await db.delete_record_by_coin_id(next_record.coin_id())
```
arvidn, line 127: there is no test case that ensures the coin record is gone after calling `delete_record_by_coin_id`.
There's also no test attempting to remove a record that doesn't exist by coin id

tests/wallet/test_singleton_store.py
```diff
@@ -0,0 +1,279 @@
...
+        recs = await db.get_records_by_singleton_id(singleton_id)
+        assert len(recs) == 1
+        assert recs[0].coin_id() == record.coin_id()
+
+        # Delete all records by id
+        await db.delete_records_by_singleton_id(singleton_id)
+        recs = await db.get_records_by_singleton_id(singleton_id)
```
arvidn, line 134: There's no test trying to remove a record that doesn't exist by singleton ID.

tests/wallet/test_singleton_store.py
```diff
@@ -0,0 +1,279 @@
...
+        timestamp = uint64(time.time())
+        await db.set_confirmed(record.coin_id(), confirmed_height, timestamp)
+        rec = await db.get_record_by_coin_id(record.coin_id())
+        assert isinstance(rec, SingletonCoinRecord)
+        assert rec.confirmed
+        assert rec.confirmed_at_height == confirmed_height
+
+        # get the next singleton record and add it to DB
+        next_record = get_next_singleton_record(record)
+        await db.add_singleton_record(next_record)
```
arvidn, line 99: I don't see a test attempting to add the same record twice

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            row = await execute_fetchone(
+                conn,
+                "SELECT * FROM singletons WHERE coin_id = ?",
+                (coin_id,),
+            )
+        if row is not None:
+            custom_data = await self.get_custom_data_by_coin_id(row[0])
```
arvidn, line 196: the same comment as @trepca made applies here too, about fetching all this in a single query. Although, it probably doesn't have the same impact here as in the loop above

tests/wallet/test_singleton_store.py
```diff
@@ -0,0 +1,279 @@
...
+        db = await WalletSingletonStore.create(wrapper)
+        custom_data = {"field_1": 100, "field_2": b"some text"}
+        record = get_new_singleton_record(custom_data)
+        await db.add_singleton_record(record)
+        res = await db.get_records_by_singleton_id(record.singleton_id)
+        assert res[0].custom_data == custom_data
+
+        # delete the singleton and check it's removed from custom data table
+        await db.delete_records_by_singleton_id(record.singleton_id)
+        custom_res = await db.get_custom_data_by_coin_id(record.coin_id())
```
arvidn, line 278: the only test of `get_custom_data_by_coin_id()` is that it can return `None`. More cases should be covered, especially a coin that exists, but also a coin that never existed.

chia/wallet/wallet_singleton_store.py
geoffwalmsley: I couldn't find a way to use JOIN that would return a serialized custom_data, so if we did use a join we'd have to go through each row and serialize it anyway. Not sure whether this is considerably less efficient?

arvidn: the round-trip to the sqlite thread is expensive, so the fewer requests you make, the faster (typically).

arvidn: I think it would look something like this:
```
SELECT field_name, value FROM singleton_data WHERE coin_id in (
   SELECT coin_id from singletons
   WHERE singleton_id=?
   ORDER BY generation DESC
   LIMIT ?
)
```

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,281 @@
...
+                    " confirmed_at_height int,"
+                    " spent_height int,"
+                    " lineage_proof blob,"
+                    " generation int,"
+                    " timestamp int)"
+                )
+            )
+
+            await conn.execute("CREATE INDEX IF NOT EXISTS singleton_id_index ON singletons(singleton_id)")
+            await conn.execute("CREATE INDEX IF NOT EXISTS wallet_id_index ON singletons(wallet_id)")
```
arvidn, line 78: I think you need an index for `generation` as well. You filter and order by it.

tests/wallet/test_singleton_store.py
```diff
@@ -0,0 +1,279 @@
...
+        recs = await db.get_records_by_singleton_id(record.singleton_id)
+        assert len(recs) == 0
+        recs = await db.get_records_by_singleton_id(record.singleton_id, one, one, one)
+        assert len(recs) == 0
+        await db.add_singleton_record(record)
+        recs = await db.get_records_by_singleton_id(record.singleton_id)
+        assert len(recs) == 1
+        recs = await db.get_records_by_singleton_id(record.singleton_id, one, one, one)
+        assert len(recs) == 0
+        recs = await db.get_records_by_singleton_id(record.singleton_id, one, uint32(20), one)
```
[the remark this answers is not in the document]
aqk: Agreed.

chia/wallet/wallet_singleton_store.py
aqk: Thanks, I'll take this.

chia/wallet/wallet_singleton_store.py
aqk: Thanks, I'll take this.

aqk:
I'm going to keep this PR open, address all commented issues, and merge this back into the DAO Wallet feature branch.

I'll leave this open until it's time to review the DAO PR.

chia/wallet/wallet_singleton_store.py
aqk: Thanks!

chia/wallet/wallet_singleton_store.py
```diff
@@ -0,0 +1,281 @@
...
+                "lineage_proof, generation, timestamp"
+            )
+            value_placeholders = ",".join(["?"] * len(columns.split(",")))
+            await conn.execute(
+                f"INSERT or REPLACE INTO singletons ({columns}) VALUES({value_placeholders})",
+                (
+                    record.coin_id(),
+                    record.coin.parent_coin_info,
+                    record.coin.puzzle_hash,
+                    record.coin.amount,
```
arvidn, line 117: `amount` is a `uint64` and has to be stored as a binary blob (of 8 bytes), because the internal integer type in sqlite is `int64`. I'm pretty sure you have to convert it to bytes here

Quexington:
I think this is a dead effort at this point and will take the initiative to close it.

PR #15316, opened 2023-05-17
Weight proof typing

arvidn:
Ahead of making changes to the plot filter logic, I would like to have full mypy coverage of these files.

There were some noteworthy changes I had to make to `weight_proof.py` that deserve careful review.


chia/full_node/weight_proof.py
```diff
@@ -1160,9 +1167,11 @@ def sub_slot_data_vdf_input(
                     if ssd.cc_slot_end_info is not None:
                         ssd = sub_slots[ssd_idx + 1]
                         break
+                    assert ssd.total_iters is not None
                     if not (ssd.total_iters > sp_total_iters):
```
arvidn, line 1171: all of these comparisons with `ssd.total_iters` seem to assume `ssd.total_iters` is never `None`, even though it can be.
I'm unsure whether this is a reasonable assumption, or whether the logic should be changed here to handle this case in some special way.
adding the asserts here makes the current assumptions explicit.

arvidn:
@wallentx 
it turns out that we don't have full test coverage of `weight_proof.py`. I still think this is an improvement, even though it doesn't add a unit test.

chia/full_node/weight_proof.py
altendky: Maybe `SubSlotData` could be generic against this?  It's frozen, so it's either `None` or not forever.

arvidn: what do you mean?
these asserts only hold in the specific places where they are placed, without the surrounding checks  we cannot make the assumption that `total_iters` is not `None`

chia/full_node/weight_proof.py
```diff
@@ -1606,41 +1628,41 @@ def map_segments_by_sub_epoch(sub_epoch_segments) -> Dict[int, List[SubEpochChal
...
+#        # Add other empty slots
+#        total_iters = uint128(total_iters + (expected_sub_slot_iters * (finished_sub_slots_since_prev - 1)))
+#    else:
+#        # Slot iters is guaranteed to be the same for header_block and prev_b
+#        # This takes the beginning of the slot, and adds ip_iters
+#        assert prev_b.cc_ip_vdf_info
+#        assert prev_b.total_iters
+#        total_iters = uint128(prev_b.total_iters - prev_b.cc_ip_vdf_info.number_of_iterations)
+#    total_iters = uint128(total_iters + sub_slot_data.cc_ip_vdf_info.number_of_iterations)
+#    return total_iters == sub_slot_data.total_iters
```
arvidn, line 1658: should I delete these functions instead of commenting them out?

almogdepaz: you can remove them

arvidn: done

chia/full_node/weight_proof.py
```diff
@@ -1480,19 +1473,21 @@ def __get_cc_sub_slot(sub_slots: List[SubSlotData], idx, ses: Optional[SubEpochS
...
 
-def _get_curr_diff_ssi(constants: ConsensusConstants, idx, summaries):
+def _get_curr_diff_ssi(
+    constants: ConsensusConstants, idx: int, summaries: List[SubEpochSummary]
+) -> Tuple[uint64, uint64]:
     curr_difficulty = constants.DIFFICULTY_STARTING
     curr_ssi = constants.SUB_SLOT_ITERS_STARTING
     for ses in reversed(summaries[0:idx]):
-        if ses.new_sub_slot_iters is not None:
+        if ses.new_sub_slot_iters is not None and ses.new_difficulty is not None:
```
arvidn, line 1482: is it reasonable to check for this here, or should this be an assert below the if-block instead?

chia/full_node/weight_proof.py
```diff
@@ -593,7 +595,7 @@ def get_fork_point_no_validations(self, weight_proof: WeightProof) -> Tuple[bool
         if len(weight_proof.sub_epochs) == 0:
             return False, uint32(0)
         summaries, sub_epoch_weight_list = _validate_sub_epoch_summaries(self.constants, weight_proof)
-        if summaries is None:
+        if summaries is None or sub_epoch_weight_list is None:
```
arvidn, line 598: here and on line 571, these used to be assumptions. Would it make more sense to have asserts instead of checks like this?

chia/full_node/weight_proof.py
almogdepaz: i think it would be easier to understand whats going on if we add an assert

chia/full_node/weight_proof.py
almogdepaz: yes asserting would be better

almogdepaz (approved):

chia/full_node/weight_proof.py
```diff
@@ -831,27 +835,6 @@ def handle_end_of_slot(
     )
 
 
-def compress_segments(full_segment_index, segments: List[SubEpochChallengeSegment]) -> List[SubEpochChallengeSegment]:
```
xdustinface, line 834: Seems like you drop this because its unused (and not longer needed?) code, so imo this should be a separate PR rather instead of dropping it here to avoid the need of hinting it?

xdustinface (approved):

chia/full_node/weight_proof.py
arvidn: yeah, probably. However, they were exposed as unused by the hinting, so in some sense it is "fixing" the hints

trepca (approved):

PR #15319, opened 2023-05-17
wallet: Introduce `CHIA_DEV_COMPILE_CLVM_DISABLED` environment variable

xdustinface:
### Purpose:

Make it possible to disable CLVM compiling until there is a proper fix for the situation. 

### Current Behavior:

All CLVM files are recompiled for each test process on startup which makes starting tests pretty slow. 

### New Behavior:

When setting `CHIA_DEV_COMPILE_CLVM_DISABLED` environment variable to something, compilation will always be skipped.

### Testing Notes:

Just play with setting/unsetting `CHIA_DEV_COMPILE_CLVM_DISABLED` and running tests.

### DIFF
--- a/chia/wallet/puzzles/load_clvm.py
+++ b/chia/wallet/puzzles/load_clvm.py
@@ -17,7 +17,9 @@ from chia.util.lock import Lockfile
 
 compile_clvm_py = None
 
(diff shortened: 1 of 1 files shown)


arvidn (approved):
I'm a bit ignorant on this setup, but it seems questionable to compile clvm (at all, default or by request) at import time. We pre-compile, and check-in, the compiled version of our CLVM. I would expect use to just load that at import-time. Ensuring that the pre-compiled programs are up-to-date with the clvm source is the job of CI

PR #15330, opened 2023-05-18
server: Store `Version` objects instead of `str` in `WSChiaConnection`

xdustinface:
### Purpose:

This is just cleaner imo. Now with this PR the default version is `0`. Other way of solving this would be to make `_version: Optional[Version]` and add a 
```
@property
def version(self) -> Version:
   assert self._version is not None
   return self._version
```
and the same for `protocol_version` but i think the 0 version is good enough and comes with less overhead. 

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -13,6 +13,7 @@ from typing import Any, Awaitable, Callable, Dict, List, Optional, Set, Tuple, U
 from aiohttp import ClientSession, WSCloseCode, WSMessage, WSMsgType
 from aiohttp.client import ClientWebSocketResponse
 from aiohttp.web import WebSocketResponse
+from packaging.version import Version
 from typing_extensions import Protocol, final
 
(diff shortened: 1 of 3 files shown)


emlowe:
Added a test for `WSChiaConnection:get_version` would eliminate the CI diff failure, as it would test the modified line.

wallentx:
Failure. Coverage is below 100%.
-------------
Diff Coverage
-------------
chia/full_node/weight_proof.py (92.5%): Missing lines 598,979,992
tests/core/custom_types/test_proof_of_space.py (100%)
-------------
Total:   45 lines
Missing: 3 lines
Coverage: 93%
-------------

xdustinface:
@wallentx I think you confused yourself somehow here :)

wallentx:
> 

Lol sorry, I'm testing a bot

emlowe:
Open and Close PR to deal with Snyk error

emlowe (approved):

PR #15333, opened 2023-05-18
Chip13 old

fchirica:


madMAx43v3r:
Maybe I'm wrong here, but CHIP-13 is about changing the plot filter to not allow a plot id to win if it passes more than X filters in a row. This branch only avoids a single plot _winning_ multiple plots in a row, which is very very unlikely and doesn't prevent the gain in efficiency for plot grinding.

fchirica:
> Maybe I'm wrong here, but CHIP-13 is about changing the plot filter to not allow a plot id to win if it passes multiple filters in short succession. This branch only avoids a single plot _winning_ multiple blocks in a row, which is very very unlikely and doesn't prevent the gain in efficiency for plot grinding.

Thanks for the comment!

chia/consensus/block_header_validation.py
```diff
@@ -491,6 +491,32 @@ def validate_unfinished_header_block(
...
+        seen_signage_points: Set[bytes32] = {cc_sp_hash}
+
+        while (
+            curr_optional_block_record is not None
+            and len(seen_signage_points) < constants.NUM_PLOT_FILTERS_DISALLOWED_TO_PASS
+        ):
+            if passes_plot_filter(
+                constants,
+                plot_id,
+                curr_optional_block_record.pos_ss_cc_challenge_hash,
```
almogdepaz, line 508: iirc the attack we where talking about is inside of one slot so the challenge chain hash (pos_ss_cc_challenge_hash) should be the same, we dont have to save it inside BlockRecord

chia/simulator/block_tools.py
```diff
@@ -536,6 +538,24 @@ def get_farmer_wallet_tool(self) -> WalletTool:
...
+
+            challenge = block.reward_chain_block.pos_ss_cc_challenge_hash
+            if block.reward_chain_block.challenge_chain_sp_vdf is None:
+                # Edge case of first sp (start of slot), where sp_iters == 0
+                cc_sp_hash = challenge
+            else:
+                cc_sp_hash = block.reward_chain_block.challenge_chain_sp_vdf.output.get_hash()
+            if passes_plot_filter(self.constants, plot_id, challenge, cc_sp_hash):
+                return True
+            seen_signage_points.add(cc_sp_hash)
```
almogdepaz, line 555: why not a counter ?

chia/consensus/block_header_validation.py
```diff
@@ -491,6 +491,32 @@ def validate_unfinished_header_block(
     if q_str is None:
         return None, ValidationError(Err.INVALID_POSPACE)
 
+    # 5c. Check plot id is not present within last `NUM_DISTINCT_CONSECUTIVE_PLOT_IDS` blocks.
+    if height >= constants.SOFT_FORK3_HEIGHT:
```
almogdepaz, line 495: do we need to add this to initial-config.py ?

chia/simulator/block_tools.py
```diff
@@ -692,6 +712,10 @@ def get_consecutive_blocks(
                                 if required_iters <= latest_block.required_iters:
                                     continue
                         assert latest_block.header_hash in blocks
+                        plot_id = get_plot_id(proof_of_space)
+                        if latest_block.height + 1 >= constants.SOFT_FORK3_HEIGHT:
+                            if self.plot_id_passes_previous_filters(plot_id, cc_sp_output_hash, block_list):
```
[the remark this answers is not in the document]
almogdepaz: if you need that for tests then maybe

tests/conftest.py
```diff
@@ -101,6 +102,8 @@ def consensus_mode(request):
 def blockchain_constants(consensus_mode) -> ConsensusConstants:
     if consensus_mode == Mode.PLAIN:
         return test_constants
+    if consensus_mode == Mode.SOFT_FORK3:
+        return test_constants.replace(**{"SOFT_FORK3_HEIGHT": 0})
```
arvidn, line 106:
```suggestion
        return test_constants.replace(SOFT_FORK3_HEIGHT=0)
```

arvidn (changes requested):

tests/blockchain/test_blockchain.py
```diff
@@ -3611,3 +3611,50 @@ async def test_reorg_flip_flop(empty_blockchain, bt):
...
+                SOFT_FORK3_HEIGHT=(0 if bt_respects_soft_fork3 else 10000),
+                NUM_PLOT_FILTERS_DISALLOWED_TO_PASS=num_plot_filters_disallowed_to_pass,
+            ),
+            keychain=keychain,
+        )
+        blockchain_constants = bt.constants.replace(
+            **{
+                "SOFT_FORK3_HEIGHT": soft_fork3_height,
+            }
+        )
```
arvidn, line 3635:
```suggestion
        blockchain_constants = bt.constants.replace(SOFT_FORK3_HEIGHT=soft_fork3_height)
```

tests/blockchain/test_blockchain.py
```diff
@@ -3611,3 +3611,50 @@ async def test_reorg_flip_flop(empty_blockchain, bt):
...
     for block in chain_b[40:]:
         await _validate_and_add_block(b, block)
+
+
+@pytest.mark.parametrize("num_plot_filters_disallowed_to_pass", [1, 2])
+@pytest.mark.parametrize("bt_respects_soft_fork3", [True, False])
+@pytest.mark.parametrize("soft_fork3_height", [0, 10, 10000])
+@pytest.mark.asyncio
+async def test_soft_fork3_activation(
+    blockchain_constants, bt_respects_soft_fork3, soft_fork3_height, db_version, num_plot_filters_disallowed_to_pass
```
arvidn, line 3621: it looks a bit odd to use `blockchain_constants` in addition to parameterizing on `soft_fork3_height`. The constants are already parameterized on the `Mode`, and you have `Mode.SOFT_FORK3` already

tests/blockchain/blockchain_test_utils.py
```diff
@@ -120,6 +120,18 @@ async def _validate_and_add_block_multi_error(
...
+async def _validate_and_add_block_multi_error_or_pass(
+    blockchain: Blockchain, block: FullBlock, expected_errors: List[Err], skip_prevalidation: bool = False
+) -> None:
+    # Checks that the blockchain returns one of the expected errors, also allows block to be added.
+    try:
+        await _validate_and_add_block(blockchain, block, skip_prevalidation=skip_prevalidation)
+    except Exception as e:
+        assert isinstance(e, AssertionError)
+        assert e.args[0] in expected_errors
+        return
```
arvidn, line 132: this `return` is redundant

chia/cmds/db_validate_func.py
```diff
@@ -102,8 +103,15 @@ def validate_v2(in_path: Path, *, validate_blocks: bool) -> None:
...
-                    block_record = BlockRecord.from_bytes(row[5])
+                    block_bytes = zstd.decompress(row[4])
+                    block = FullBlock.from_bytes(block_bytes)
+                    block_record_db: BlockRecordDB = BlockRecordDB.from_bytes(row[5])
+                    plot_filter_info: PlotFilterInfo = plot_filter_info_from_block(block_bytes)
+                    block_record = BlockRecord.from_block_record_db(
+                        block_record_db,
+                        plot_filter_info.pos_ss_cc_challenge_hash,
+                        plot_filter_info.cc_sp_hash,
+                    )
```
arvidn, line 114: renaming `block_record` to `block_record_db` and constructing this `block_record` object seems unnecessary, and just increases complexity. Unless I'm missing something, I would suggest you preserve the name of `block_record` (but obviously you need to change its type to `BlockRecordDB`.

In the code below where you check the `pos_cc_challenge_hash` and `pos_ss_cc_challenge_hash` you can just compare them against `plot_filter_info`.

But, come to think of it, all of this new code seems unnecessary. This is not validating the integrity of the database. You effectively parse out these fields using two separate functions and then ensure they are the same. It looks like a poor-man's test of your new parser function. I don't think that belongs in here. Add it to the unit test instead

chia/cmds/db_validate_func.py
arvidn, line 1: the only thing I think you should change in this file is to update the `BlockRecord` type to match what's stored in the DB, nothing else

chia/consensus/block_record.py
```diff
@@ -41,6 +41,41 @@ def is_transaction_block(self) -> bool:
...
+    overflow: bool
+    prev_transaction_block_height: uint32
+    timestamp: Optional[uint64]
+    prev_transaction_block_hash: Optional[bytes32]
+    fees: Optional[uint64]
+    reward_claims_incorporated: Optional[List[Coin]]
+    finished_challenge_slot_hashes: Optional[List[bytes32]]
+    finished_infused_challenge_slot_hashes: Optional[List[bytes32]]
+    finished_reward_slot_hashes: Optional[List[bytes32]]
+    sub_epoch_summary_included: Optional[SubEpochSummary]
```
arvidn, line 76: I think you should factor out this change into a separate PR. the change where you introduce `BlockRecordDB`. I think it's a good, independent, change.

Also, I think this type belongs in `block_store.py` and should be private in there

chia/consensus/block_record.py
```diff
@@ -117,3 +154,66 @@ def ip_iters(self, constants: ConsensusConstants) -> uint64:
 
     def sp_total_iters(self, constants: ConsensusConstants) -> uint128:
         return uint128(self.sp_sub_slot_total_iters(constants) + self.sp_iters(constants))
+
+    def to_block_record_db(self) -> BlockRecordDB:
```
arvidn, line 158: I think `BlockRecordDB` should be private in `block_store.py` and any conversion functions should also live in that file.
Keeping it private makes the surface area to the block store smaller and makes it easier to change its implementation without affecting the rest of the code base.

e.g. it opens the door for a v3 database format where we don't store all the duplicate fields in both the `BlockRecord` and `FullNode`

chia/full_node/block_store.py
```diff
@@ -350,6 +357,36 @@ async def get_block_info(self, header_hash: bytes32) -> Optional[GeneratorBlockI
                     b.foliage.prev_block_hash, b.transactions_generator, b.transactions_generator_ref_list
                 )
 
+    async def get_plot_filter_info(self, header_hash: bytes32) -> Optional[PlotFilterInfo]:
```
arvidn, line 360: I really think it's a mistake to add this function. The block store is a serious bottleneck in the full node, please avoid more round-trips to the database. In the places where you need these fields out of the full block, add it to the existing query.

chia/util/full_block_utils.py
```diff
@@ -322,3 +323,26 @@ def header_block_from_block(
...
+    buf = skip_uint128(buf)  # weight
+    buf = skip_uint32(buf)  # height
+    buf = skip_uint128(buf)  # total_iters
+    buf = skip_uint8(buf)  # signage_point_index
+    pos_ss_cc_challenge_hash = bytes32(buf[:32])
+    buf = skip_bytes32(buf)  # pos_ss_cc_challenge_hash
+    buf = skip_proof_of_space(buf)  # proof_of_space
+    # Optional[challenge_chain_sp_vdf]
+    if buf[0] == 0:
+        return PlotFilterInfo(pos_ss_cc_challenge_hash, pos_ss_cc_challenge_hash)
```
arvidn, line 345: this looks like a typo. I assume it's not, but I think it would help future readers of this code to have a comment explaining why the `pos_ss_cc_challange_hash` and `cc_sp_hash` are supposed to be the same in this case

arvidn (changes requested):
my main concern is the changes to `block_store.py`, I think they need to preserve the number of DB round-trips we have. But I also think you should pull out the `BlockRecordDB` patch into a separate PR. I think it's complex enough to not be mixed up with CHIP-13

chia/full_node/block_store.py
```diff
@@ -553,16 +522,18 @@ async def get_block_records_by_hash(self, header_hashes: List[bytes32]) -> List[
...
                 async with conn.execute(
-                    "SELECT header_hash,block_record FROM full_blocks "
+                    "SELECT header_hash,block_record,block FROM full_blocks "
                     f'WHERE header_hash in ({"?," * (len(header_hashes) - 1)}?)',
                     header_hashes,
                 ) as cursor:
                     for row in await cursor.fetchall():
                         header_hash = bytes32(row[0])
                         all_blocks[header_hash] = BlockRecordDB.from_bytes(row[1])
+                        all_block_bytes[header_hash] = zstd.decompress(row[2])
```
arvidn, line 536: I think it would be better to just save off the plot filter info here, and maybe call the dictionary `all_plot_filter_info`
```
all_plot_filter_info[header_hash] = plot_filter_info_from_block(zstd.decompress(row[2]))
```
Also for the v1 case on line 552

chia/full_node/block_store.py
```diff
@@ -695,15 +685,17 @@ async def get_block_records_in_range(
...
             async with self.db_wrapper.reader_no_transaction() as conn:
                 async with conn.execute(
-                    "SELECT header_hash, block_record FROM full_blocks WHERE height >= ? AND height <= ?",
+                    "SELECT header_hash, block_record,block FROM full_blocks WHERE height >= ? AND height <= ?",
                     (start, stop),
                 ) as cursor:
                     for row in await cursor.fetchall():
                         header_hash = bytes32(row[0])
                         ret[header_hash] = BlockRecordDB.from_bytes(row[1])
+                        all_block_bytes[header_hash] = zstd.decompress(row[2])
```
arvidn, line 698: same comment here. I think you should throw away the full block bytes as early as possible by just picking out the plot filter info you need inside the loop

chia/full_node/block_store.py
```diff
@@ -781,15 +781,17 @@ async def get_block_records_close_to_peak(
...
             async with self.db_wrapper.reader_no_transaction() as conn:
                 async with conn.execute(
-                    "SELECT header_hash, block_record FROM full_blocks WHERE height >= ?",
+                    "SELECT header_hash, block_record,block FROM full_blocks WHERE height >= ?",
                     (peak[1] - blocks_n,),
                 ) as cursor:
                     for row in await cursor.fetchall():
                         header_hash = bytes32(row[0])
                         ret[header_hash] = BlockRecordDB.from_bytes(row[1])
+                        all_block_bytes[header_hash] = zstd.decompress(row[2])
```
arvidn, line 794: and here too

chia/cmds/db_validate_func.py
```diff
@@ -102,8 +102,9 @@ def validate_v2(in_path: Path, *, validate_blocks: bool) -> None:
                     continue
 
                 if validate_blocks:
-                    block = FullBlock.from_bytes(zstd.decompress(row[4]))
-                    block_record = BlockRecord.from_bytes(row[5])
+                    block_bytes = zstd.decompress(row[4])
+                    block = FullBlock.from_bytes(block_bytes)
```
arvidn, line 106: this change should be reverted

chia/cmds/db_validate_func.py
almogdepaz: +1

PR #15337, opened 2023-05-19
catchup: into long_lived/compression from main  @ 1777ecc2e8b46a4c1a42466e5be55e87db7b4fa3

wallentx:
Source hash: 1777ecc2e8b46a4c1a42466e5be55e87db7b4fa3
Remaining commits: 0


altendky (approved):
It might be worth fixing CI up a bit in the target before catching it up, but I guess that kind of depends on what issues there are.

wallentx:
> It might be worth fixing CI up a bit in the target before catching it up, but I guess that kind of depends on what issues there are.

Thx, failures are known/expected since we don't have chiapos wheels for all the things yet

PR #15338, opened 2023-05-20
add test coverage for starting a node with an existing db

altendky:
### Purpose:

Add coverage for `FullNode._start()` in the case where the node is started with a pre-existing database.  This coverage is needed in particular for https://github.com/Chia-Network/chia-blockchain/pull/14920.


chia/simulator/setup_services.py
```diff
@@ -98,9 +98,10 @@ async def setup_full_node(
     connect_to_daemon: bool = False,
     db_version: int = 1,
     disable_capabilities: Optional[List[Capability]] = None,
+    reuse_db: bool = False,
```
arvidn, line 101: I would have preferred adding `*,` here, to force this bool to be named. but it's not a big deal

arvidn (dismissed):

chia/simulator/setup_services.py
altendky:
```suggestion
    *,
    reuse_db: bool = False,
```

emlowe (approved):

PR #15340, opened 2023-05-20
have pyinstaller include all chia modules and .hex files

altendky:

### DIFF
--- a/chia/pyinstaller.spec
+++ b/chia/pyinstaller.spec
@@ -65,9 +65,16 @@ SERVERS = [
     "timelord",
 ]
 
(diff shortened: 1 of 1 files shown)


wallentx (dismissed):

wallentx (approved):
Perhaps as a separate TODO issue, we should introduce CI that can detect the presence of new items, and verify that they are updated in the installer. I also assume using poetry might somehow provide a way to solve for this.

Quexington (approved):
Looks good to me :+1:

trepca (approved):

PR #15345, opened 2023-05-22
update build_network_protocol_files.py

arvidn:
`tests/util/test_network_protocol_files.py` (along with two other files) are generated by a script. These files were updated to pass mypy (specifically, to use different names for variables of different types). But the script generating the files was not updated.

This patch updates the script to generate the same update as was applied earlier, so the diff is small.

This is a prerequisite for adding new fields to protocol messages and regenerate the tests.


arvidn:
I don't think it's useful to block this PR because of this script not having test coverage. It's not a test nor production code.

.github/workflows/upload-pypi-source.yml
```diff
@@ -129,6 +129,10 @@ jobs:
             command: flake8 benchmarks build_scripts chia tests tools *.py
           - name: pylint
             command: pylint benchmarks build_scripts chia tests tools *.py
+          - name: generated protocol tests
```
xdustinface, line 132: Thoughts about just making this a pre-commit step?

arvidn: not really. the main purpose of these tests are to make sure we don't accidentally change a network protocol message, which would make it incompatible with previous versions. I think it's good to have an extra manual step to ensure you really mean it in this case.

xdustinface: I don't really understand what you want to tell me with this know :) I got the purpose of those tests, my question was actually about running this CI check you introduce here in `pre-commit` instead of `upload-pypi-source.yml`. Maybe i miss some difference here but in my mind running it in `pre-commit` has the same outcome on CI than what you do now + it already fails locally if you miss to rebuild the files (which should be done fast enough to not bother on `pre-commit` locally?). Or what is the purpose of this new step here?

altendky: It got a new workflow, I suggested this instead.  Could go in pre-commit, but we all know I'm not a huge fan.  Not that I need to argue against it going there.

xdustinface: > It got a new workflow, I suggested this instead.

Which new workflow? 

> Could go in pre-commit, but we all know I'm not a huge fan. Not that I need to argue against it going there.

Your not really a fan of pre-commit for _some_ reason yeah, but now this introduces the inconsistency to run our "CI check tools" in two different places while we get one more distinctions between local and CI linting which is not critical but also not heavy and could just run locally also.

xdustinface (approved):
Still not a fan of putting this into the source code upload action but i also don't want to block it.

.github/workflows/upload-pypi-source.yml
```diff
@@ -129,6 +129,10 @@ jobs:
...
+            command: |
+              python3 -m tests.util.build_network_protocol_files
+              git diff --exit-code
```
altendky, line 135: Note that this will at least not catch any new files if they are ever created.

.github/workflows/upload-pypi-source.yml
altendky: I'm fine with it going in pre-commit.  Had a longer keybase chat with Kevin as well.

.github/workflows/upload-pypi-source.yml
arvidn: I think that's ok for now

arvidn: actually. it won't catch new files that failed to be added to the repo. If the files *were* added, and the differ from the generated ones, it would be caught

trepca (approved):

emlowe:
Coverage exception is ok for this to merge.

PR #15357, opened 2023-05-22
Type hint

aqk:
Type hint code cleanup


altendky (approved):
Not sure what's happening with the unrelated CI failures but I think those are being looked at elsewhere.

emlowe (approved):

PR #15361, opened 2023-05-23
wallet: Refactor `WalletStateManager.get_keys`

xdustinface:
### Purpose:

Implicitly fixes hinting because the function didn't return `Optional` in preparation for a "Fix `Wallet` mypy" PR and get rid of the tuple return which isn't really needed because we only use the public key in one place where we can just calculate it instead (which happens currently even in two places in the method itself). 

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -8,20 +8,7 @@ import traceback
 from contextlib import asynccontextmanager
 from pathlib import Path
 from secrets import token_bytes
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    AsyncIterator,
-    Callable,
-    Dict,
-    Iterator,
-    List,
-    Optional,
-    Set,
-    Tuple,
-    Type,
-    TypeVar,
-)
+from typing import TYPE_CHECKING, Any, AsyncIterator, Callable, Dict, Iterator, List, Optional, Set, Type, TypeVar
 
 import aiosqlite
 from blspy import G1Element, PrivateKey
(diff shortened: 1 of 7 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -275,17 +262,13 @@ async def create(
...
         record = await self.puzzle_store.record_for_puzzle_hash(puzzle_hash)
         if record is None:
-            raise ValueError(f"No key for this puzzlehash {puzzle_hash})")
+            raise ValueError(f"No key for puzzle hash: {puzzle_hash.hex()}")
         if record.hardened:
-            private = master_sk_to_wallet_sk(self.private_key, record.index)
-            pubkey = private.get_g1()
-            return pubkey, private
-        private = master_sk_to_wallet_sk_unhardened(self.private_key, record.index)
-        pubkey = private.get_g1()
```
arvidn, line 287: this patch is an optimization even, since this function used to compute the public key, just to be thrown away in most cases

chia/wallet/wallet_state_manager.py
```diff
@@ -275,17 +262,13 @@ async def create(
     def get_public_key_unhardened(self, index: uint32) -> G1Element:
         return master_sk_to_wallet_sk_unhardened(self.private_key, index).get_g1()
 
-    async def get_keys(self, puzzle_hash: bytes32) -> Optional[Tuple[G1Element, PrivateKey]]:
+    async def get_private_key(self, puzzle_hash: bytes32) -> PrivateKey:
         record = await self.puzzle_store.record_for_puzzle_hash(puzzle_hash)
         if record is None:
-            raise ValueError(f"No key for this puzzlehash {puzzle_hash})")
```
arvidn, line 281: so, this never returned `None`, so all the code you removed checking for none was dead code all along. Are you sure we don't do anything important in those snippets?

chia/wallet/wallet_state_manager.py
```diff
@@ -275,17 +262,13 @@ async def create(
...
+            raise ValueError(f"No key for puzzle hash: {puzzle_hash.hex()}")
```
arvidn, line 268: we have a custom `__str__` for `bytes32` that always prints it as hex

arvidn (approved):

chia/wallet/wallet_state_manager.py
xdustinface: We just raise (and log in one case which is not critical to drop the log imo) which this method does already.

chia/wallet/wallet_state_manager.py
xdustinface: Yeah right but after one of your rust optimizations we have `bytes32` objects which are not really `bytes32` objects, right? Thats why i recently started to just `hex` them because the byte representation looks pretty ugly in logs :)

PR #15362, opened 2023-05-23
wallet: Hint `wallet.py`

xdustinface:
### Purpose:

More type hints.

Note that `chia.wallet.cat_wallet.cat_wallet` and `chia.data_layer.data_layer_wallet` are also dropped from the exclusions because they are implicitly fixed by this PR too.

### New Behavior:

No change in behaviour.

### Based on:

- #15361 
- #15422

### DIFF
--- a/chia/wallet/wallet.py
+++ b/chia/wallet/wallet.py
@@ -48,6 +48,7 @@ from chia.wallet.wallet_info import WalletInfo
 
 if TYPE_CHECKING:
     from chia.server.ws_connection import WSChiaConnection
+    from chia.wallet.wallet_state_manager import WalletStateManager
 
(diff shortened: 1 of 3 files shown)


chia/wallet/wallet.py
```diff
@@ -48,6 +48,7 @@
 
 if TYPE_CHECKING:
     from chia.server.ws_connection import WSChiaConnection
+    from chia.wallet.wallet_state_manager import WalletStateManager
```
xdustinface, line 51: @wallentx Coverage reports this line here which, this can be ignored.
@altendky Any way to not fail for type checking stuff like this? :)

arvidn (approved):

chia/wallet/wallet.py
altendky: https://github.com/Chia-Network/chia-blockchain/pull/15473

PR #15367, opened 2023-05-23
get_wallet_balances 

trepca:
### Current Behavior:
Clients would have to first fetch wallet ids and then call `get_wallet_balance` for each id. 

### New Behavior:
Added a `get_wallet_balances` that can either take wallet_ids or empty param to return balances for all wallets.


paninaro (dismissed):
Love it, thanks for adding this!

I did some manual testing of the new and existing RPCs and found them to be working as expected.

wallentx:
These test failures need attention @trepca

paninaro (approved):

PR #15374, opened 2023-05-24
Add confirmed filter for get_transactions

ytx1991:
### Purpose:
We need to get unconfirmed txs for the Clawback

### Current Behavior:
None

### New Behavior:
When confirmed=True only return confirmed txs
When confirmed=False only return unconfirmed txs
When confirmed=None return all txs

### Testing Notes:
Unit test

### DIFF
--- a/chia/wallet/wallet_transaction_store.py
+++ b/chia/wallet/wallet_transaction_store.py
@@ -263,6 +263,7 @@ class WalletTransactionStore:
         end,
         sort_key=None,
         reverse=False,
+        confirmed: Optional[bool] = None,
         to_puzzle_hash: Optional[bytes32] = None,
         type_filter: Optional[TransactionTypeFilter] = None,
     ) -> List[TransactionRecord]:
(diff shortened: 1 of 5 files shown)


chia/wallet/wallet_transaction_store.py
```diff
@@ -286,6 +287,11 @@ async def get_transactions_between(
         else:
             query_str = SortKey[sort_key].ascending()
 
+        if confirmed is None:
+            confirmed_str = ""
+        else:
+            confirmed_str = "AND confirmed=" + str(int(confirmed))
+
```
xdustinface, line 294:
```suggestion
        confirmed_str = ""
        if confirmed is not None:
            confirmed_str = f"AND confirmed={int(confirmed)}"
         
```

nit, feel free to just close this :)

xdustinface (dismissed):

xdustinface (dismissed):

trepca (dismissed):

trepca (approved):

PR #15376, opened 2023-05-24
use a more aggresive activation schedule on testnet10

arvidn:
As the comments suggests, this activates the:

* 2.0 *softfork* immediately
* 2.0 *hardfork* 1 weeks after (planned) release (which also reduces the plot filter by one bit)
* the next plot filter reduction 2 weeks after that
* the next plot filter reduction 3 years after that
* the last plot filter reduction 3 years after that
Note that these are all `testnet10` block heights, which is about a million behind `mainnet`

### DIFF
--- a/chia/server/start_full_node.py
+++ b/chia/server/start_full_node.py
@@ -67,16 +67,32 @@ def create_full_node_service(
     )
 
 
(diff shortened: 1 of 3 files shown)


trepca (approved):
exciting!

PR #15384, opened 2023-05-25
Convert blockchain_constants's consensus_mode param into a fixture and use it to skip some simulation related tests in HARD_FORK_2_0 mode

AmineKhaldi:


tests/conftest.py
```diff
@@ -614,7 +619,9 @@ async def farmer_three_harvester_not_started(
 # because of a hack in shutting down the full node, which means you cannot run
 # more than one simulations per process.
 @pytest_asyncio.fixture(scope="function")
-async def daemon_simulation(bt, get_b_tools, get_b_tools_1):
+async def daemon_simulation(consensus_mode, bt, get_b_tools, get_b_tools_1):
+    if consensus_mode == Mode.HARD_FORK_2_0:
+        pytest.skip("Skipping test for Mode.HARD_FORK_2_0")
```
arvidn, line 624: I think it would make sense to explain that this test only supports one running at a time.
And the condition should probably be:
```
if consensus_mode != Mode.PLAIN:
```

arvidn:
@fchirica does this look reasonable to you?

tests/conftest.py
```diff
@@ -838,3 +846,11 @@ def cost_logger_fixture() -> Iterator[CostLogger]:
...
     print()
     print(cost_logger.log_cost_statistics())
+
+
+@pytest_asyncio.fixture(scope="function")
+async def simulation(consensus_mode, bt):
+    if consensus_mode != Mode.PLAIN:
+        pytest.skip("Skipping this run. This test only supports one running at a time.")
+    async for _ in setup_full_system(test_constants_modified, bt, db_version=1):
+        yield _
```
arvidn, line 856: why did you move this fixture from `test_simulation.py`?
and does `test_simulation_1()` also have the issue of only working with a single instance at a time?

AmineKhaldi: > why did you move this fixture from `test_simulation.py`?
Importing `Mode` from `conftest` would result in circular dependency.

> does `test_simulation_1()` also have the issue of only working with a single instance at a time?
Yes, I noticed that from the test runs.

arvidn: > Importing Mode from conftest would result in circular dependency.
it seems odd that it would be any different from importing `bt` from `conftest.py`
You typically don't have to explicitly import fixtures in order to use them

AmineKhaldi: You do when you `if consensus_mode != Mode.PLAIN:` though.

arvidn: I see

arvidn:
test coverage suggests that `test_siumulation,py` isn't run at all now: https://github.com/Chia-Network/chia-blockchain/actions/runs/5083882773/jobs/9136673710?pr=15384#step:11:21

arvidn:
oh, right. that's the coverage, not the missed %

arvidn (approved):
the two lines not being executed in `conftest.py` is fine and expected. They will be once we extend the tests with more "modes"

PR #15387, opened 2023-05-26
Cache filter prefix bits by challenge chain signage point hash to allow plot filter validation

AmineKhaldi:


chia/farmer/farmer_api.py
```diff
@@ -517,14 +520,19 @@ async def request_signed_values(self, full_node_request: farmer_protocol.Request
...
-        (plot_identifier, challenge_hash, sp_hash, node_id) = self.farmer.quality_str_to_identifiers[
-            full_node_request.quality_string
-        ]
+        (
+            plot_identifier,
+            challenge_hash,
+            sp_hash,
+            node_id,
+            filter_prefix_bits,
+        ) = self.farmer.quality_str_to_identifiers[full_node_request.quality_string]
```
arvidn, line 529: if you already store the `filter_prefix_bits` here, in the `self.farmer.quality_str_to_identifiers` dictionary. Do you really need to bounce it against the harvester at all?

chia/farmer/farmer_api.py
```diff
@@ -321,13 +324,13 @@ async def respond_signatures(self, response: harvester_protocol.RespondSignature
...
         include_taproot: bool = pospace.pool_contract_puzzle_hash is not None
 
-        # we skip plot filter validation here, we trust the harvester
         computed_quality_string = verify_and_get_quality_string(
             pospace,
             self.farmer.constants,
             response.challenge_hash,
             response.sp_hash,
             height=None,
+            filter_prefix_bits=response.filter_prefix_bits,
```
arvidn, line 333: would it not be possible to look up the filter size from `self.farmer.quality_str_to_identifiers` instead?
I suppose we don't technically have the quality string before this call, but it seems we already store the filter size

chia/farmer/farmer.py
```diff
@@ -699,6 +702,7 @@ async def _periodically_clear_cache_and_refresh_task(self) -> None:
                     for key, add_time in self.cache_add_time.items():
                         if now - float(add_time) > self.constants.SUB_SLOT_TIME_TARGET * 3:
                             self.sps.pop(key, None)
+                            self.sp_hash_to_filter_size.pop(key, None)
```
arvidn, line 705: it looks like we could save a fair amount of space by combining all these dicts into one. but that's for a future improvement

chia/farmer/farmer_api.py
```diff
@@ -321,13 +326,13 @@ async def respond_signatures(self, response: harvester_protocol.RespondSignature
         assert pospace is not None
         include_taproot: bool = pospace.pool_contract_puzzle_hash is not None
 
-        # we skip plot filter validation here, we trust the harvester
         computed_quality_string = verify_and_get_quality_string(
```
arvidn, line 329: can you add an assert in `verify_and_get_quality_string()` to ensure we always pass in either `height` or `filter_prefix_bits` too?

arvidn (approved):

arvidn:
looks good. Do you want me to land this first or do you want to update the patch first?

arvidn (approved):

PR #15407, opened 2023-05-31
copy connection set to iterate while pinging

altendky:
### Purpose:

Failed pings would result in removing the connection from the connection set which causes the next for loop iteration to raise an exception.


emlowe (approved):

wjblanke (approved):
aok

PR #15415, opened 2023-05-31
util: Improve `list_to_batches`

xdustinface:
### Purpose:

We have a few places where we copy the set into a list before batching/chunking it, this PR intends to get rid of this. The set batching code is slightly slower than the list batching code but its negligible and only starts to make a real difference with ~5 million entries.

- Introduce generic `Batch` class.
- Make it `to_batches` and allow to split `set` or `list`
- Move it into `chia.util.misc`

### Current Behavior:

`list_to_batches` only allows splitting a `list`.

### New Behavior:

`to_batches` allows to split a `set` or a `list`.


arvidn (dismissed):
we also have a similar function called `chunks()`

xdustinface:
> we also have a similar function called `chunks()`

Yeah i already have a branch where this all gets replaced by `to_batches`.

arvidn (approved):

xdustinface:
@wallentx Coverage failures about plot refreshing here are false negatives due to some issue in multiprocessing coverage reporting. Running locally:

```
chia/plotting/manager.py                                        257     15    105     11  91.7%   96, 99-100, 131->exit, 166->165, 186-187, 201->205, 240, 244, 247, 267, 284-288, 298->301, 302-305
```

PR #15418, opened 2023-06-01
util: Drop `chia.util.chunks` and use `chia.util.to_batches` instead

xdustinface:
### Purpose:

- Replace all usage of `chunks` with `to_batches` which has the same purpose.
- Drop `chunks.py` and `test_chunks.py`. 

### New Behavior:

No change in behaviour.

### Based on:

- #15415


chia/full_node/coin_store.py
arvidn, line 1: this looks to be in conflict with https://github.com/Chia-Network/chia-blockchain/pull/15420

arvidn (commented):
looks good. But I think it will have to be rebased on top of #15420

arvidn (approved):

PR #15421, opened 2023-06-01
wallet: Drop unused conditions helper/parameter

xdustinface:
### Purpose:

Drop some unused/untested code.

### New Behavior:

No change in behaviour.


trepca (approved):

Quexington (approved):
I sort of like being able to use these should we need, but it's not like it's a hassle to add them back, so I'm on board.

PR #15424, opened 2023-06-01
wallet: Fix missing hint in `GenesisById.generate_issuance_bundle`

xdustinface:
### Purpose:

Avoid creation of unhinted CATs. Fixes a test failure in #15274.

### Current Behavior:

CATs created via `GenesisById.generate_issuance_bundle` are unhinted.

### New Behavior:

CATs created via `GenesisById.generate_issuance_bundle` are hinted.

### DIFF
--- a/chia/wallet/puzzles/tails.py
+++ b/chia/wallet/puzzles/tails.py
@@ -92,7 +92,9 @@ class GenesisById(LimitationsProgram):
 
         inner_solution = wallet.standard_wallet.add_condition_to_solution(
             Program.to([51, 0, -113, tail, []]),
-            wallet.standard_wallet.make_solution(primaries=[Payment(cat_inner.get_tree_hash(), amount)]),
+            wallet.standard_wallet.make_solution(
+                primaries=[Payment(cat_inner.get_tree_hash(), amount, [cat_inner.get_tree_hash()])]
+            ),
         )
         eve_spend = unsigned_spend_bundle_for_spendable_cats(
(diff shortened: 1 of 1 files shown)


Quexington (dismissed):

chia/wallet/puzzles/tails.py
```diff
@@ -92,7 +92,9 @@ async def generate_issuance_bundle(cls, wallet, _: Dict, amount: uint64) -> Tupl
 
         inner_solution = wallet.standard_wallet.add_condition_to_solution(
             Program.to([51, 0, -113, tail, []]),
-            wallet.standard_wallet.make_solution(primaries=[Payment(cat_inner.get_tree_hash(), amount)]),
+            wallet.standard_wallet.make_solution(
+                primaries=[Payment(cat_inner.get_tree_hash(), amount, [cat_inner.get_tree_hash()])]
```
arvidn, line 96: `get_tree_hash()` isn't all that cheap. I think it would make sense to just call it once

arvidn (dismissed):
looks

chia/wallet/puzzles/tails.py
xdustinface: Its probably negligible here because it only happens when creating a CAT but i added 92d3dec25d4797b4eb133a3ed8f73beb5b3c47f5.

arvidn (approved):

Quexington (approved):

PR #15437, opened 2023-06-03
remove unnecessary filtering of state_changed messages for the UI

paninaro:
### Purpose:
Too many times now the UI has been missing out on state_changed messages from the wallet backend because this silly hardcoded list doesn't get updated. The UI is interested in all state_changed messages from the wallet, so we can make life easier and just send all such messages. 

### Current Behavior:
vc_coin_added and vc_coin_removed state_changed messages are not sent to the UI. Others may not be sent as well...

### New Behavior:
All state_changed messages from the wallet backend are sent to the UI.

### Testing Notes:
Manually tested with the UI logging the previously unsent messages

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -246,26 +246,7 @@ class WalletRpcApi:
             # Metrics is the only current consumer for this event
             payloads.append(create_payload_dict(change, change_data, self.service_name, "metrics"))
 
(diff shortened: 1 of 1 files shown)


ytx1991 (approved):
I'm not sure the reason why we need this filter. A safer way is just to add vc notification to the list. But I don't have any concerns to remove it.

trepca (approved):

PR #15444, opened 2023-06-05
Add support for multi node farmers

felixbrucker:
### Purpose:
This PR allows the farmer to be configured with multiple full nodes to be able to continue farming while n-1 nodes are unavailable/down. This can be useful when currently using a single remote full node hosted locally or in the cloud, where having another full node helps to reduce farm downtime and making maintenance easier.

### Current Behavior:

Currently when configuring a farmer to use multiple full nodes it processes the new signage points multiple times, once for each full node. Additionally the signed values required in the block forging process are sent to all nodes, not just the node which originally requested them, resulting in n*n `SignedValues` messages

### New Behavior:

This PR fixes those two issues:
- The farmer now checks for duplicate SPs before processing the new SP to avoid duplicate lookups
- The `request_signed_values` api method of the farmer now handles its state by using `call_api` to keep track of the harvester response and returns the appropriate `SignedValues` message to the full node itself

### Testing Notes:

I added tests regarding the duplicate SP processing, returning the signed values from the farmer and the `call_api_of_specific` helper.


jack60612:
any way for you to make the diff on farmer_api,py not look so messy?

felixbrucker:
@jack60612 i have split up the commit into smaller chunks, the first commit only moves logic into a private method, but git still makes a mess there, don't think that can be avoided, but the change is much smaller now. All other commits should be easier to read now 👍

jack60612:
> @jack60612 i have split up the commit into smaller chunks, the first commit only moves logic into a private method, but git still makes a mess there, don't think that can be avoided, but the change is much smaller now. All other commits should be easier to read now 👍

Sounds good & sorry for the delay and feel free to bug me on discord / gh in the future.

jack60612:
I'll review this in the morning.

chia/farmer/farmer_api.py
```diff
@@ -443,139 +582,10 @@ async def respond_signatures(self, response: harvester_protocol.RespondSignature
...
-        await self.farmer.plot_sync_receivers[peer.peer_node_id].process_keys_missing(message)
-
-    @api_request(peer_required=True)
-    async def plot_sync_duplicates(self, message: PlotSyncPathList, peer: WSChiaConnection) -> None:
-        await self.farmer.plot_sync_receivers[peer.peer_node_id].process_duplicates(message)
-
-    @api_request(peer_required=True)
-    async def plot_sync_done(self, message: PlotSyncDone, peer: WSChiaConnection) -> None:
-        await self.farmer.plot_sync_receivers[peer.peer_node_id].sync_done(message)
+        return None
```
jack60612, line 607: not sure if this None is needed.

tests/connection_utils.py
```diff
@@ -39,16 +40,34 @@ async def disconnect_all_and_reconnect(server: ChiaServer, reconnect_to: ChiaSer
...
+    server: ChiaServer, self_hostname: str, dummy_port: int, type: NodeType = NodeType.FULL_NODE
+) -> Tuple[WSChiaConnection, bytes32]:
     timeout = aiohttp.ClientTimeout(total=10)
     session = aiohttp.ClientSession(timeout=timeout)
     config = load_config(server.root_path, "config.yaml")
-    chia_ca_crt_path, chia_ca_key_path = chia_ssl_ca_paths(server.root_path, config)
+
+    ca_crt_path: Path
+    ca_key_path: Path
+    authenticated_client_types = {NodeType.HARVESTER}
```
jack60612, line 57: could you more clearly define this is a set

chia/harvester/harvester_api.py
```diff
@@ -266,7 +266,7 @@ async def lookup_challenge(
             },
         )
 
-    @api_request()
+    @api_request(reply_types=[ProtocolMessageTypes.respond_signatures])
```
jack60612, line 269: good catch (feel free to dismiss this)

chia/farmer/farmer_api.py
```diff
@@ -354,8 +363,11 @@ async def new_signage_point(self, new_signage_point: farmer_protocol.NewSignageP
 
             msg = make_msg(ProtocolMessageTypes.new_signage_point_harvester, message)
             await self.farmer.server.send_to_all([msg], NodeType.HARVESTER)
```
jack60612, line 365: How about simply adding it to the list right here, and then not having to worry about an exception?
I'm not too worried about something getting added twice as the processing should be pretty fast, if you disagree let me know.

chia/farmer/farmer_api.py
```diff
@@ -399,8 +399,17 @@ async def request_signed_values(self, full_node_request: farmer_protocol.Request
...
         )
 
-        msg = make_msg(ProtocolMessageTypes.request_signatures, request)
-        await self.farmer.server.send_to_specific([msg], node_id)
+        response = await self.farmer.server.call_api_of_specific(HarvesterAPI.request_signatures, request, node_id)
+        if response is None or not isinstance(response, harvester_protocol.RespondSignatures):
+            self.farmer.log.error(f"Invalid response from harvester {node_id} for request_signatures: {response}")
+            return None
+
+        signed_values = self._process_respond_signatures(response)
```
jack60612, line 407: maybe adding an additional comment explaining why you separated this function would be nice.

jack60612 (changes requested):
feel free to do the merge now.

jack60612:
oh I just realized all those comments weren't one review, sorry about that.

chia/farmer/farmer_api.py
felixbrucker: It's not needed, but recommended for functions which actually return stuff (pep8), if desired i can remove it ofc

jack60612: should be fine then, was not sure if mypy cared or not.

chia/farmer/farmer_api.py
felixbrucker: Duplicate SP processing by the harvesters within a very short timeframe will cause multiple reads and processing in case of compressed plots, which is undesired

Imho ensuring we do not do unnecessary reads and compute is important, is it not?

jack60612: fair point with the compressed plots.

jack60612:
is the ci good?

felixbrucker:
It should go through except for the coverage one which complains about the extracted processing, as the farmer_api didn't have any tests yet

jack60612 (dismissed):
lgtm, I'm going to get @Chia-Network/required-reviewers approval now.

jack60612:
@felixbrucker we are going to review this after the 19th.

jack60612:
@felixbrucker don't worry about resolving the conflicts, ill let you know when we are going to review it.

jack60612:
@felixbrucker you are good to resolve now

jack60612 (approved):
lgtm

wjblanke (approved):
aok

PR #15455, opened 2023-06-06
Repin clvm_tools_rs==0.1.34

Quexington:
This should mainly fix the problem with CLVM constantly recompiling during pytest runs.  However, in an update to the compiler, hex files are now emitted with a new line (just a `\n` not a `\r\n` on Windows).  This means we also had to update all of those, but if you hide whitespace in the diff you can verify that this is the only thing that has changed.  However this will result in a bit of tumult for people pulling main after this change because they will have to rebuild their venv in order to pass the pre-commit check.


arvidn (approved):
it was awkward to not have a newline at the end of those files anyway

altendky (approved):
thanks!

PR #15460, opened 2023-06-06
rename exclude_coin_* -> excluded_coin_* for consistency

Quexington:
This PR does mostly what the title says, we had two names for the same concept.  I chose the one that seemed to make the most sense in the code (often you will assign exclude_coin_amounts to a list that you will call excluded_coin_amounts).  _both_ were used in the RPCs so I left backwards compatibility in place by falling back to the previous key if the first one did not exist.


trepca (changes requested):
thanks, ltgm, just need to resolve merge

trepca (approved):

PR #15465, opened 2023-06-07
server: Fix versions for incoming connections

xdustinface:
### Purpose:

Store `protocol_version` and `software_version` reported in the handshake of incoming connections.

### Current Behavior:

Incoming connections have a `0` version for `protocol_version` and `version`.

### New Behavior:

We store the actual versions reported by the handshake.

### Testing Notes:

Adjusted the related test.


arvidn (approved):
I would have expected the crawler to use this class too. It manages to record the correct version, afaik.

xdustinface:
> It manages to record the correct version, afaik.

For outgoing connections, i guess?

hoffmang9:
Crawler is a false node and only connects outbound.

PR #15466, opened 2023-06-07
server: Introduce `ApiProtocol`

xdustinface:
### Purpose:

Add basic `ApiProtocol` class which can be used to enforce attributes/methods on API classes.

### Current Behavior:

- APIs are hinted as `Any`
- APIs use mixed logging, some from the node, some from a separate
- `FarmerAPI` always accepts requests.

### New Behavior:

- APIs are hinted as `ApiProtocol`.
- APIs have consistently separate loggers
- `FarmerAPI` ignores requests until the keys are set up / its started.

### DIFF
--- a/chia/farmer/farmer_api.py
+++ b/chia/farmer/farmer_api.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import logging
 import time
 from typing import Any, Dict, List, Optional, Tuple
 
@@ -53,9 +54,11 @@ def strip_old_entries(pairs: List[Tuple[float, Any]], before: float) -> List[Tup
 
 
 class FarmerAPI:
+    log: logging.Logger
     farmer: Farmer
 
     def __init__(self, farmer: Farmer) -> None:
+        self.log = logging.getLogger(__name__)
         self.farmer = farmer
 
     def ready(self) -> bool:
--- a/chia/farmer/farmer_api.py
+++ b/chia/farmer/farmer_api.py
@@ -58,8 +58,7 @@ class FarmerAPI:
     def __init__(self, farmer: Farmer) -> None:
         self.farmer = farmer
 
-    @property
-    def api_ready(self) -> bool:
+    def ready(self) -> bool:
         return self.farmer.started
 
     @api_request(peer_required=True)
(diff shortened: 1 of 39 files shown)


arvidn (commented):
This looks good to me. I think it would be good for someone else to also have a look. perhaps @altendky 

is there a way to assert that the various classes actually do implement the `ApiProtocol`?
I seem to recall there is, in which case it would be nice to ensure we get an error if we ever add another function to the protocol, but fail to add it to one of the classes meant to implement it

xdustinface:
> perhaps @altendky

Yeah that review is requested already :)

> is there a way to assert that the various classes actually do implement the `ApiProtocol`?

They are checked when they get passed into `Service` which takes a `ApiProtocol` so when something is missing you get errors like: 

```
chia/server/start_farmer.py:54: error: Argument "peer_api" to "Service" has incompatible type "FarmerAPI"; expected "ApiProtocol"  [arg-type]
chia/server/start_farmer.py:54: note: "FarmerAPI" is missing following "ApiProtocol" protocol member:
chia/server/start_farmer.py:54: note:     api_ready
Found 1 error in 1 file (checked 773 source files)
```

chia/farmer/farmer_api.py
```diff
@@ -58,6 +58,10 @@ class FarmerAPI:
     def __init__(self, farmer: Farmer) -> None:
         self.farmer = farmer
 
+    @property
```
altendky, line 61: Are these used enough that we can't just make it a regular function?  I'd consider a property to be a tansitional tool when trying to retain back-compat, but it seems that we have just the single use that we're already updating.

I'm also vaguely itchy about the repetition of `api`, but not entirely sure I should be.  As in, `whatever_api.api_ready` is repetitive vs. just naming this `.ready()`.

altendky (changes requested):
I made https://github.com/xdustinface/chia-blockchain/pull/5 to make `Server` (more) generic and avoid the casts.  It could be added here or as a separate PR.

chia/farmer/farmer_api.py
xdustinface: I was tempted to remove the `api_` prefix also but i didn't want to make this a change PR actually, just introduce a protocol about what exists. Since its a bigger change due to your PR anyway now, i added d64dbf89243ab64716fa18857a6c11a6a68c6bd5 :)

altendky (approved):

emlowe (approved):

PR #15467, opened 2023-06-07
full_node: Add `max_height` to `CoinStore.get_coin_states_by_ids`

xdustinface:
### Purpose:

Allow to fetch `CoinState`'s based on min/max height instead of only min height as preparation for follow up PR to introduce new message `get_coin_infos`.

### New Behavior:

No change in behaviour.

### Based on:

- #15420

### DIFF
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -432,6 +432,7 @@ class CoinStore:
         coin_ids: Set[bytes32],
         min_height: uint32 = uint32(0),
         *,
+        max_height: uint32 = uint32(uint32.MAXIMUM_EXCLUSIVE - 1),
         max_items: int = 50000,
     ) -> List[CoinState]:
         if len(coin_ids) == 0:
@@ -445,10 +446,15 @@ class CoinStore:
                     coin_ids_db = tuple(batch.entries)
                 else:
                     coin_ids_db = tuple([pid.hex() for pid in batch.entries])
+
+                max_height_sql = ""
+                if max_height != uint32.MAXIMUM_EXCLUSIVE - 1:
+                    max_height_sql = f"AND confirmed_index<={max_height} AND spent_index<={max_height}"
+
                 async with conn.execute(
                     f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, coin_parent, amount, timestamp "
                     f'FROM coin_record WHERE coin_name in ({"?," * (len(batch.entries) - 1)}?) '
-                    f"AND (confirmed_index>=? OR spent_index>=?)"
+                    f"AND (confirmed_index>=? OR spent_index>=?) {max_height_sql}"
                     f"{'' if include_spent_coins else 'AND spent_index=0'}"
                     " LIMIT ?",
                     coin_ids_db + (min_height, min_height, max_items - len(coins)),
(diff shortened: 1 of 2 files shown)


chia/full_node/coin_store.py
```diff
@@ -432,6 +432,7 @@ async def get_coin_states_by_ids(
         include_spent_coins: bool,
         coin_ids: Set[bytes32],
         min_height: uint32 = uint32(0),
+        max_height: uint32 = uint32(uint32.MAXIMUM_EXCLUSIVE - 1),
```
arvidn, line 435: I think it would be good to add this after the `*`

chia/full_node/coin_store.py
```diff
@@ -449,10 +450,10 @@ async def get_coin_states_by_ids(
                 async with conn.execute(
                     f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, coin_parent, amount, timestamp "
                     f'FROM coin_record WHERE coin_name in ({"?," * (len(batch.entries) - 1)}?) '
-                    f"AND (confirmed_index>=? OR spent_index>=?)"
+                    f"AND (confirmed_index>=? OR spent_index>=?) AND (confirmed_index<=? AND spent_index<=?)"
```
arvidn, line 453: it's risky to make changes like these. the query planner might pick the wrong index for instance.
I would feel a bit better about it if this additional constraint would go away entirely when not used. Similar to `AND spent_index=0` below.
I suppose the parameters becomes a bit more complicated that way though.
maybe you can run the benchmark before and after this change to ensure it still performs well

chia/full_node/coin_store.py
xdustinface: Changed

chia/full_node/coin_store.py
xdustinface: > I suppose the parameters becomes a bit more complicated that way though.

Changed but not sure what you mean with ^ ? What makes it more complicated? And which benchmark do you mean?

arvidn: I mean that, if you make `AND (confirmed_index<=? AND spent_index<=?)` conditional, you would also have to make it conditional whether to pass in the two arguments for the `?`.

chia/full_node/coin_store.py
```diff
@@ -445,10 +446,15 @@ async def get_coin_states_by_ids(
                     coin_ids_db = tuple(batch.entries)
                 else:
                     coin_ids_db = tuple([pid.hex() for pid in batch.entries])
+
+                max_height_sql = ""
+                if max_height != uint32.MAXIMUM_EXCLUSIVE - 1:
+                    max_height_sql = f"AND confirmed_index<={max_height} AND spent_index<={max_height}"
```
arvidn, line 452: I understand that this might be the most practical solution to this. I just want to note that "burning in" arguments into the string like this will defeat python's query cache, requiring it to be recompiled every time (as long as the arguments differ). Perhaps not a big deal, especially not if this is only used in this one query you're adding.

The other obvious risk with doing things like this is that type-checking is weak in python, and if `max_height` somehow, somewhere, can be a string, you have sql injection attacks.

arvidn (approved):

chia/full_node/coin_store.py
xdustinface: > I understand that this might be the most practical solution to this. I just want to note that "burning in" arguments into the string like this will defeat python's query cache, requiring it to be recompiled every time (as long as the arguments differ). Perhaps not a big deal, especially not if this is only used in this one query you're adding.

Yeah probably not very relevant here since we anyway probably don't end up with the same queries, usually. 

> The other obvious risk with doing things like this is that type-checking is weak in python, and if `max_height` somehow, somewhere, can be a string, you have sql injection attacks.

Thanks, good thoughts to keep in mind! Generally i think the ways the `max_height` comes to us, via the peer API is safe due to the deserialisation of the messages and in case thats something which gets exposed to RPC we generally also convert to int and if not in some case its probably also not so critical since the RPC interface is not public.

PR #15469, opened 2023-06-08
Clawback handle refine

ytx1991:
### Purpose:
Moving the Clawback coin handling logic to the wallet level coin_added

### Current Behavior:
The coin record and tx are added in the wallet state manager.

### New Behavior:
The coin record and tx are added in the wallet coin_added.

### DIFF
--- a/chia/wallet/wallet.py
+++ b/chia/wallet/wallet.py
@@ -5,6 +5,7 @@ import time
 from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Optional, Set, Tuple, cast
 
 from blspy import AugSchemeMPL, G1Element, G2Element
+from chia_rs.chia_rs import CoinState
 
 from chia.consensus.cost_calculator import NPCResult
 from chia.full_node.bundle_tools import simple_solution_generator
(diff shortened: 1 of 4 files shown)


trepca (approved):
lgtm and much cleaner integration into standard wallet vs wsm

chia/wallet/wallet.py
```diff
@@ -623,10 +629,70 @@ async def get_coins_to_offer(
         return await self.select_coins(amount, min_coin_amount=min_coin_amount, max_coin_amount=max_coin_amount)
 
     # WSChiaConnection is only imported for type checking
-    async def coin_added(
-        self, coin: Coin, height: uint32, peer: WSChiaConnection
-    ) -> None:  # pylint: disable=used-before-assignment
-        pass
+    async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection) -> None:
+        wallet_node = self.wallet_state_manager.wallet_node
+        coin_states: Optional[List[CoinState]] = await wallet_node.get_coin_state([coin.parent_coin_info], peer=peer)
```
xdustinface, line 634: This duplicated fetching of spend info and metadata matching/parsing can be avoided with some refactoring.

xdustinface (changes requested):
In this state we request parent coin state + spend info twice for clawback coins which is not really an improvement. Also since from my understanding the clawback is meant to be used by any wallet i think the place where should integrate it is rather `WalletStateManager.coin_added`? There we can also reuse the code to add the `WalletCoinRecord` and `TransactionRecord` if we generalise it a bit it seems.

PR #15472, opened 2023-06-08
Optimize and simplify farmer RPC get_signage_point()

AmineKhaldi:


chia/rpc/farmer_rpc_api.py
```diff
@@ -218,23 +218,23 @@ async def _state_changed(self, change: str, change_data: Optional[Dict[str, Any]
...
-                            "challenge_chain_sp": sp.challenge_chain_sp,
-                            "reward_chain_sp": sp.reward_chain_sp,
-                            "difficulty": sp.difficulty,
-                            "sub_slot_iters": sp.sub_slot_iters,
-                            "signage_point_index": sp.signage_point_index,
-                        },
-                        "proofs": pospaces,
-                    }
-        raise ValueError(f"Signage point {sp_hash.hex()} not found")
+        sp_hash = bytes32.from_hexstr(request["sp_hash"])
```
arvidn, line 221: `hexstr_to_bytes()` supports stripping an optional `0x` prefix from the string as well. This change removes that feature

chia/rpc/farmer_rpc_api.py
```diff
@@ -218,23 +218,23 @@ async def _state_changed(self, change: str, change_data: Optional[Dict[str, Any]
         return payloads
 
     async def get_signage_point(self, request: Dict[str, Any]) -> EndpointResult:
-        sp_hash = hexstr_to_bytes(request["sp_hash"])
-        for _, sps in self.service.sps.items():
-            for sp in sps:
-                if sp.challenge_chain_sp == sp_hash:
```
arvidn, line 224: this code seems to suggest that there may be signage points recorded under the wrong hash. Is that just someone not understanding how the dictionary works, or is this a real concern?
Did you dig into the history of where this comes from?

arvidn (changes requested):

chia/rpc/farmer_rpc_api.py
AmineKhaldi: Hm? `from_hexstr()` also supports stripping the optional prefixes.

arvidn: so this code is pointless then? https://github.com/Chia-Network/chia-blockchain/blob/main/chia/util/byte_types.py#L14

altendky: For `bytes` you use `hexstr_to_bytes()`.  When I worked on the fixed width bytes I made a method for those types, `.from_hexstr()`.

ChiaMineJP (commented):
I noticed that `self.service.sps` is volatile...
Meaning the lifetime of it is only about 30 minutes and never picks signage points before the launches of farmer/harvester...
image

I think `get_signage_point` RPC API should search for the signage point in the database and return it if it exists.

chia/rpc/farmer_rpc_api.py
arvidn: you make the assumption that this linear scan of all signage points, looking for `sp_hash` behaves the same as your version, where you only pick the *first* signage point. Is this list guaranteed to only contain 1 item?
Did you make sure that we always insert SPs under the right key in this dict?

chia/rpc/farmer_rpc_api.py
```diff
@@ -237,23 +237,24 @@ async def _state_changed(self, change: str, change_data: Optional[Dict[str, Any]
...
-                        },
-                        "proofs": pospaces,
-                    }
-        raise ValueError(f"Signage point {sp_hash.hex()} not found")
+        sp_hash = bytes32.from_hexstr(request["sp_hash"])
+        sps = self.service.sps.get(sp_hash)
+        if sps is None or len(sps) < 1:
+            raise ValueError(f"Signage point {sp_hash.hex()} not found")
+        sp = sps[0]
+        assert sp_hash == sp.challenge_chain_sp
```
arvidn, line 245: this assert is a good start. But it only helps if we have thorough testing of this, which I wouldn't assume we do

AmineKhaldi: For instance, this function is covered through `test_farmer_signage_point_endpoints`

arvidn (approved):

PR #15481, opened 2023-06-09
bump blspy==2.0.0 (blst)

wallentx:
We can use this PR to verify the functioning state of CI when using the beta wheels of blspy


wjblanke (dismissed):
aok

emlowe (approved):

PR #15496, opened 2023-06-13
Clawback resync

ytx1991:


chia/wallet/wallet_state_manager.py
```diff
@@ -1130,6 +1130,8 @@ async def handle_clawback(
         # Record metadata
         assert coin_state.created_height is not None
         is_recipient: Optional[bool] = None
+        # For resync correctly, we need to fetch the coin state again
+        latest_coin_state = (await self.wallet_node.get_coin_state([coin_state.coin.name()], peer=peer))[0]
```
xdustinface, line 1134: Re-fetching the coin we are currently processing doesn't look right. We would we need to do that?

ytx1991: This is for fixing the resync issue of clawback. After the resync, the wallet may not update the clawback coin to spent, which causes the related clawback tx also be stuck in pending status. The possible root cause is the concurrent syncing. We may process or fetch the higher blocks before the block which confirms the clawback coin. To solve this, I added an additional check to fetch the latest coin state. The passed-in coin state is for a certain height.
This fix is verified on the testnet.

xdustinface: But why do you need to fetch the coins state again here? Wouldn't it be enough to just do what you do below with `coin_state` instead of `latest_coin_state`? If not, i think it would be good to investigate what is really happening here in the failure case and maybe try to reproduce this issue in a test?

xdustinface:
Reopening because CI obviously only run conflicts checks here.

chia/wallet/wallet_state_manager.py
ytx1991: Since the coin_state is always unspent, the coin may already have been spent long time ago. This is a general issue for syncing and caused many weird issues, not just for Clawback. We should think about if it is feasible to resync synchronously. I mean only process the higher blocks before all previous blocks are processed.

ytx1991: I just added a log. Seems the coin_state has a chance to be set as spent. So the question is why sometimes the coin_State is spent but sometimes it doesn't.

chia/wallet/wallet_state_manager.py
```diff
@@ -1149,11 +1151,14 @@ async def handle_clawback(
         if is_recipient is not None:
             spend_bundle = SpendBundle([coin_spend], G2Element())
             memos = compute_memos(spend_bundle)
+            spent_height: uint32 = (
+                uint32(0) if latest_coin_state.spent_height is None else uint32(latest_coin_state.spent_height)
+            )
```
altendky, line 1156: This does not address the coverage complaint.  https://github.com/Chia-Network/chia-blockchain/pull/15496/commits/9a2ed50a50e0724768cba358ad26c91c6902bdca  Probably either go back to the original, or take this slightly more explicit option.  Or maybe there's a good third option, I haven't looked too much.

```suggestion
            spent_height: uint32
            if latest_coin_state.spent_height is None:
                spent_height = uint32(0)
            else:
                spent_height = uint32(latest_coin_state.spent_height)
            )
```

altendky (dismissed):

chia/wallet/wallet_state_manager.py
altendky: If they fix https://github.com/nedbat/coveragepy/issues/509 it would complain regardless.

chia/wallet/wallet_state_manager.py
```diff
@@ -1149,11 +1149,14 @@ async def handle_clawback(
         if is_recipient is not None:
             spend_bundle = SpendBundle([coin_spend], G2Element())
             memos = compute_memos(spend_bundle)
+            spent_height: uint32 = uint32(0)
+            if coin_state.spent_height is not None:  # pragma: no cover
```
altendky, line 1153: Why should we not test this?

chia/wallet/wallet_state_manager.py
altendky: Continued at https://github.com/Chia-Network/chia-blockchain/pull/15496#pullrequestreview-1481700019.

chia/wallet/wallet_state_manager.py
ytx1991: I don't know how to simulate the nuke wallet DB. I extended the resync unit test but it doesn't trigger the change.

ytx1991:
This fix is not necessary for the release, close for now

chia/wallet/wallet_state_manager.py
ytx1991: This is removed

xdustinface:
@ytx1991 The test you added in 91ac933 doesn't seem to pass. It still running on CI but times out here locally with the following:

```
21:02:20 chia.wallet.wallet_node: INFO Selected multiprocessing start method: default
21:02:20 chia.wallet.wallet_node: INFO Resetting wallet sync data...
21:02:20 chia.wallet.wallet_node: INFO Reset wallet sync data completed.
21:02:20 chia.wallet.wallet_node: DEBUG Trying to disable resync: 2999502625 [2999502625]
21:02:20 chia.wallet.wallet_node: INFO Disabled resync for wallet fingerprint: 2999502625
21:02:20 chia.wallet.wallet_state_manager: DEBUG Starting in db path: /private/var/folders/6b/sb101ygd0_xfxmbc_ljxybr80000gn/T/tmprryrmaxf/test-wallet-db-None-2999502625.sqlite
21:02:20 chia.wallet.wallet_node: INFO Wallet is logged in using key with fingerprint: 2999502625
21:02:20 chia.wallet.wallet_node: INFO Updated last used fingerprint: 2999502625
21:02:20 chia.wallet.did_wallet.did_wallet: INFO Confirmed balance for did wallet is 0
21:02:20 chia.wallet.did_wallet.did_wallet: INFO Confirmed balance for did wallet is 0
FAILED
tests/wallet/rpc/test_wallet_rpc.py:1907 (test_fully_resync[True])
tests/wallet/rpc/test_wallet_rpc.py:1980: in test_fully_resync
    await time_out_assert(20, wc2.get_synced)
E   AssertionError: Timed assertion timed out after 22 seconds: expected True, got False
```

chia/wallet/wallet_state_manager.py
ytx1991: added test

trepca: can we remove the pragma now?

chia/wallet/wallet_state_manager.py
```diff
@@ -1431,7 +1473,7 @@ async def _add_coin_states(
                                         sent_to=[],
                                         trade_id=None,
                                         type=uint32(TransactionType.OUTGOING_TX.value),
-                                        name=bytes32(token_bytes()),
+                                        name=tx_name,
```
xdustinface, line 1476: Seems like a good idea to not use random bytes for a tx hash yeah but its probably not related to the actual problem here?

xdustinface (changes requested):

chia/wallet/wallet_state_manager.py
altendky: Looks like there are still lines that need covered, but this particular case seems addressed.  Feel free to close this discussion.  I'll dismiss my requested changes.  Thanks.

chia/wallet/wallet_state_manager.py
```diff
@@ -1149,11 +1158,41 @@ async def handle_clawback(
         if is_recipient is not None:
             spend_bundle = SpendBundle([coin_spend], G2Element())
             memos = compute_memos(spend_bundle)
+            spent_height: uint32 = uint32(0)
+            if coin_state.spent_height is not None:
+                self.log.debug("Resync clawback coin: %s", coin_state.coin.name().hex())
```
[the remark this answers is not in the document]
ytx1991: In the normal case, if the clawback coin is spent, this handle_clawback will not be executed. It will only be spent here when the wallet first time saw the coin, and it is spent. The tx record shouldn't exist at this point since if it exists the wallet should also have a related coin record. If there is a coin record, this handle_clawback will not be executed.

chia/wallet/wallet_state_manager.py
```diff
@@ -754,15 +754,24 @@ async def spend_clawback_coins(self, clawback_coins: Dict[Coin, ClawbackMetadata
...
+                if (
+                    len(memos) > 0
+                    and not is_recipient
+                    and await self.puzzle_store.puzzle_hash_exists(recipient_puzhash)
+                ):
+                    # Sender and recipient belong to the same wallet
+                    # Clean hint in this case to prevent duplicate processing
+                    memos[0] = b""
+                if len(memos) == 1 and memos[0] == b"":
+                    memos = []
```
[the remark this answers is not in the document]
ytx1991: We can't just drop the first element here since the second one will be the hint which is the actual memo. GUI will hide the first element.

chia/wallet/wallet_state_manager.py
```diff
@@ -1164,14 +1203,13 @@ async def handle_clawback(
...
             created_timestamp = await self.wallet_node.get_timestamp_for_height(uint32(coin_state.created_height))
-            spend_bundle = SpendBundle([coin_spend], G2Element())
             tx_record = TransactionRecord(
                 confirmed_at_height=uint32(coin_state.created_height),
                 created_at_time=uint64(created_timestamp),
                 to_puzzle_hash=metadata.recipient_puzzle_hash,
                 amount=uint64(coin_state.coin.amount),
                 fee_amount=uint64(0),
-                confirmed=False,
+                confirmed=False if spent_height == 0 else True,
```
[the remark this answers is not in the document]
ytx1991: No, the confirmed has a different meaning here. GUI uses this to know if one clawback tx is claimable.

chia/wallet/wallet_state_manager.py
```diff
@@ -1149,11 +1158,41 @@ async def handle_clawback(
...
+            spent_height: uint32 = uint32(0)
+            if coin_state.spent_height is not None:
+                self.log.debug("Resync clawback coin: %s", coin_state.coin.name().hex())
+                # Resync case
+                spent_height = uint32(coin_state.spent_height)
+                # Create Clawback outgoing transaction
+                created_timestamp = await self.wallet_node.get_timestamp_for_height(uint32(coin_state.spent_height))
+                clawback_coin_spend: CoinSpend = await fetch_coin_spend_for_coin_state(coin_state, peer)
+                clawback_spend_bundle: SpendBundle = SpendBundle([clawback_coin_spend], G2Element())
+                tx_record = TransactionRecord(
```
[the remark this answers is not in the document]
ytx1991: They are different. This is an outgoing tx. We user resync wallet, we need to recover it.

chia/wallet/wallet_state_manager.py
```diff
@@ -1149,11 +1158,41 @@ async def handle_clawback(
...
+                    additions=clawback_spend_bundle.additions(),
+                    removals=clawback_spend_bundle.removals(),
+                    wallet_id=uint32(1),
+                    sent_to=[],
+                    trade_id=None,
+                    type=uint32(TransactionType.OUTGOING_CLAWBACK),
+                    name=clawback_spend_bundle.name(),
+                    memos=list(compute_memos(clawback_spend_bundle).items()),
+                )
+                await self.add_pending_transaction(tx_record)
```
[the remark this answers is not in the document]
ytx1991: If the confirmed = true, then the wallet will not send the tx again. But I agree to change this.

chia/wallet/wallet_state_manager.py
ytx1991: This is the original fix for the duplicate issue since we saw duplicate tx records. It seems unrelated now, but good to have. Let me know if you think it's necessary to create another PR for this.

tests/wallet/clawback/test_clawback_lifecycle.py
```diff
@@ -195,9 +195,10 @@ async def test_clawback_spends(self, cost_logger: CostLogger) -> None:
                 cost_logger=cost_logger,
                 cost_log_msg="Create Second Clawback",
             )
-
-            new_cb_coin = (await sim_client.get_coin_records_by_puzzle_hash(cb_puz_hash))[1].coin
-
+            cb_coins = await sim_client.get_coin_records_by_puzzle_hash(cb_puz_hash)
+            new_cb_coin = cb_coins[0].coin
+            if cb_coins[0].spent:
+                new_cb_coin = cb_coins[1].coin
```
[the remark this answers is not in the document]
ytx1991: We observed a weird transient double-spend error in this test, trying to fix it.

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
xdustinface: Why would the GUI hide the first element? Did we introduce this for this clawback case now or is there some general rule we established, that we include empty bytes for the first memo if its not a hint? Im not aware of this. Maybe the GUI should be adjusted instead if this is the only case where this is relevant?

chia/wallet/wallet_state_manager.py
xdustinface: Hm i will defer to @trepca here but for me it seems like we don't need this in the release then and we can PR it against main instead.

tests/wallet/clawback/test_clawback_lifecycle.py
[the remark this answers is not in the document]
xdustinface: Seems like #15619 intends to fix this, right?

chia/wallet/wallet_state_manager.py
```diff
@@ -1164,14 +1198,13 @@ async def handle_clawback(
             await self.coin_store.add_coin_record(coin_record)
             # Add tx record
             created_timestamp = await self.wallet_node.get_timestamp_for_height(uint32(coin_state.created_height))
-            spend_bundle = SpendBundle([coin_spend], G2Element())
```
xdustinface, line 1167: I take it was unused but was it maybe supposed to be passed into the transaction record below rather than being dropped like you did with the other record above?

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
ytx1991: We set the first element as b"" majorly because we don't want to mess up others' wallets. If the second element is someone's puzzle hash, the wallet will process this coin but it will throw error since it doesn't own the puzzle hash. The GUI thing exists since the first day I think. Since the first element is set by the wallet, so we don't want to show the wallet generated memo unless it is the only element.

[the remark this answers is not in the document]
xdustinface: > We set the first element as b"" majorly because we don't want to mess up others' wallets.

Can you link me to some other place where we do this? Maybe i just always missed it but i never saw it before.

> If the second element is someone's puzzle hash, the wallet will process this coin but it will throw error since it doesn't own the puzzle hash. 

Why/How would this be the case?

tests/wallet/clawback/test_clawback_lifecycle.py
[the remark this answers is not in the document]
ytx1991: removed

chia/wallet/wallet_state_manager.py
ytx1991: The problem is this tx can be unconfirmed, and it will resend the spend bundle if it is None.  We will see double spend error in the log.

chia/wallet/util/transaction_type.py
```diff
@@ -15,7 +15,7 @@ class TransactionType(IntEnum):
     OUTGOING_CLAWBACK = 8
 
 
-CLAWBACK_TRANSACTION_TYPES = {
+CLAWBACK_INCOMING_TRANSACTION_TYPES = {
```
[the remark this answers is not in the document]
ytx1991: It may be because I merged the recent release/1.8.2 and reverted some changes mistakenly. Do you think I should revert this change?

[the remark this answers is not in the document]
xdustinface: Im fine either way, leaving it in here or dropping since its introduced in the release anyway.

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
ytx1991: Everywhere we use hint will have this issues, for examples
For CAT: https://github.com/Chia-Network/chia-blockchain/blob/release/1.8.2/chia/wallet/wallet_state_manager.py#L862
For DID: https://github.com/Chia-Network/chia-blockchain/blob/release/1.8.2/chia/wallet/wallet_state_manager.py#L917
It doesn't mean we add coins to wrong wallets, it just generate some unnecessary error logs in these wallets.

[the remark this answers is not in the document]
xdustinface: My point is, this is not something new and we already now on main in every other case don't know if the first entry, if its a puzzle hash, belongs to us or not so i wonder why should we now in case of clawback start to change the way we handle things usually and introduce a empty first entry? Its seems very unlikely to me that someone puts someone else's puzzle hash into the memo or do i miss something here?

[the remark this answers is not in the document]
ytx1991: For The GUI question, you can try it by sending a CAT with a memo. Every CAT has a hint as the first element and your memo will be the second element. On GUI you will only see the second element.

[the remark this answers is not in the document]
xdustinface: > For The GUI question, you can try it by sending a CAT or Clawback tx with a memo. Every CAT has a hint as the first element and your memo will be the second element. On GUI you will only see the second element.

Okay yeah but a CAT is supposed to have a hint in any case, a clawed/claimed XCH coin not?

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
ytx1991: The confirmed means different here. In general, it means "confirmed on the blockchain". but here it means "confirmed the ownership". An alternative solution is adding a new column and index in the tx table to indicate if a tx is claimable.

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
ytx1991: As I mentioned above, the empty byte is not for GUI majorly. Assume the memo list is [My_Puzzle_Hash, Your_Puzzle_HASH]. In the common case, your wallet will not process this coin state because the hint is my puzzle hash. But if you drop the first element, then the hint will be your puzzle hash and your wallet will try to process this coin states. But since it doesn't match the actual inner p2 puzzle so it will fail and log an error. In the worst case, it may trigger some unknown bug. So if we want to prevent this unnecessary processing, we need to set the first element to empty or a byte32 that doesn't belong to any wallet.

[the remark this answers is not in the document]
xdustinface: Did you read this https://github.com/Chia-Network/chia-blockchain/pull/15496#discussion_r1242716607?

chia/wallet/wallet_state_manager.py
trepca: @ytx1991  lets PR against main

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
trepca: a comment explaining this would be good

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
ytx1991: I agree it's a rare case, I will change this. We'd better merge this before the release.

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
ytx1991: Added

.github/workflows/test.yml
```diff
@@ -41,7 +41,7 @@ jobs:
...
+          echo matrix_mode=${{ ( github.event_name == 'workflow_dispatch' ) && 'all' || ( github.repository_owner == 'Chia-Network' && github.repository != 'Chia-Network/chia-blockchain' ) && 'limited' || ( github.repository_owner == 'Chia-Network' && github.repository == 'Chia-Network/chia-blockchain' && github.ref == 'refs/heads/main' ) && 'main' || ( github.repository_owner == 'Chia-Network' && github.repository == 'Chia-Network/chia-blockchain' && startsWith(github.ref, 'refs/heads/release/') ) && 'all' || ( github.repository_owner == 'Chia-Network' && github.repository == 'Chia-Network/chia-blockchain' && startsWith(github.base_ref, 'release/') ) && 'all' || 'main' }} >> $GITHUB_OUTPUT
```
xdustinface, line 44: This shouldn't be here.

chia/wallet/wallet_state_manager.py
```diff
@@ -1177,15 +1208,16 @@ async def handle_clawback(
             # Add merkle coin
             await self.coin_store.add_coin_record(coin_record)
             # Add tx record
+            # We use confirmed to indicate if a Clawback transaction is claimable
```
xdustinface, line 1211: Seems like this comment should sit next to the `TransactionRecord` creation and not above the timestamp.
```suggestion
            # We use TransactionRecord.confirmed to indicate if a Clawback transaction is claimable
```

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -1994,6 +1994,8 @@ async def test_set_wallet_resync_on_startup(wallet_rpc_environment: WalletRpcTes
     env: WalletRpcTestEnvironment = wallet_rpc_environment
     full_node_api: FullNodeSimulator = env.full_node.api
     client: WalletRpcClient = env.wallet_1.rpc_client
+    wallet_node: WalletNode = env.wallet_1.node
+    wallet_node_2: WalletNode = env.wallet_2.node
```
xdustinface, line 1998: Why was this moved?

xdustinface (changes requested):
There is still a leftover from the retargeting to `main` and two nits.

tests/wallet/rpc/test_wallet_rpc.py
ytx1991: I didn't change this. it should be because the base branch changed. I can change it back

chia/wallet/wallet_state_manager.py
```diff
@@ -761,15 +761,15 @@ async def spend_clawback_coins(self, clawback_coins: Dict[Coin, ClawbackMetadata
                 if self.main_wallet.secret_key_store.secret_key_for_public_key(derivation_record.pubkey) is None:
                     await self.main_wallet.hack_populate_secret_key_for_puzzle_hash(derivation_record.puzzle_hash)
                 amount = uint64(amount + coin.amount)
+                # Remove the clawback hint since it is unnecessary for the XCH coin
+                memos: List[bytes] = [] if len(incoming_tx.memos) == 0 else incoming_tx.memos[0][1][1:]
```
xdustinface, line 765:
```suggestion
                memos: List[bytes] = incoming_tx.memos[0][1][1:]
```

Is there any scenario how the memos could be empty here?

xdustinface (dismissed):

danieljperry (commented):
I cloned `main` and added this PR. I'm having trouble getting auto claim to work. This video shows when it fails (I did cut out a lot where nothing was happening, but I didn't cut the auto claim part):


https://github.com/Chia-Network/chia-blockchain/assets/69756004/f1a33a3e-0367-4aa4-a429-08e443ae7145

Here's my log from when the claim happened:

[auto_claim.txt](https://github.com/Chia-Network/chia-blockchain/files/11908780/auto_claim.txt)

danieljperry:
Here's another issue I saw while attempting to claw back a transaction. Note that this problem might not be related to this PR:


https://github.com/Chia-Network/chia-blockchain/assets/69756004/58132ca6-e735-47b8-b11b-069a81a4fd3f

Log snippet:

```2023-06-29T11:34:13.924 wallet chia.rpc.wallet_rpc_api    : ERROR    Failed to get transaction 1b332cee62521ffe34d04712db291eeb3681e2453ff50811d65d0b87c9d470c3.
  File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\wallet_rpc_api.py", line 912, in get_transactions
    assert record is not None, f"Cannot find coin record for clawback transaction {tx['name']}"
AssertionError: Cannot find coin record for clawback transaction 0x1b332cee62521ffe34d04712db291eeb3681e2453ff50811d65d0b87c9d470c3
2023-06-29T11:34:22.696 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
  File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\rpc_server.py", line 340, in safe_handle
    response = await self.ws_api(message)
  File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\rpc_server.py", line 331, in ws_api
    return await f_rpc_api(data)
  File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\wallet_rpc_api.py", line 1062, in spend_clawback_coins
    coin_ids: List[bytes32] = [bytes32.from_hexstr(coin) for coin in request["coin_ids"]]
  File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\wallet_rpc_api.py", line 1062, in <listcomp>
    coin_ids: List[bytes32] = [bytes32.from_hexstr(coin) for coin in request["coin_ids"]]
  File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\util\byte_types.py", line 51, in from_hexstr
    if input_str.startswith("0x") or input_str.startswith("0X"):
AttributeError: 'NoneType' object has no attribute 'startswith'```

chia/wallet/wallet_state_manager.py
ytx1991: It shouldn't be empty, just in case of unexpected things happen.

ytx1991:
> Here's another issue I saw while attempting to claw back a transaction. Note that this problem might not be related to this PR:
> 
>  nonetype.mp4 
> Log snippet:
> 
> ```
> Traceback (most recent call last):
>   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\wallet_rpc_api.py", line 912, in get_transactions
>     assert record is not None, f"Cannot find coin record for clawback transaction {tx['name']}"
> AssertionError: Cannot find coin record for clawback transaction 0x1b332cee62521ffe34d04712db291eeb3681e2453ff50811d65d0b87c9d470c3
> 2023-06-29T11:34:22.696 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
>   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\rpc_server.py", line 340, in safe_handle
>     response = await self.ws_api(message)
>   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\rpc_server.py", line 331, in ws_api
>     return await f_rpc_api(data)
>   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\wallet_rpc_api.py", line 1062, in spend_clawback_coins
>     coin_ids: List[bytes32] = [bytes32.from_hexstr(coin) for coin in request["coin_ids"]]
>   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\wallet_rpc_api.py", line 1062, in <listcomp>
>     coin_ids: List[bytes32] = [bytes32.from_hexstr(coin) for coin in request["coin_ids"]]
>   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\util\byte_types.py", line 51, in from_hexstr
>     if input_str.startswith("0x") or input_str.startswith("0X"):
> AttributeError: 'NoneType' object has no attribute 'startswith'```
> ```

It seems like the clawback coin is missing for some reason. I assume it is related to wallet switching. For the auto claim, did you wait several minutes after the timelock expired?

danieljperry:
> > Here's another issue I saw while attempting to claw back a transaction. Note that this problem might not be related to this PR:
> > nonetype.mp4
> > Log snippet:
> > ```
> > Traceback (most recent call last):
> >   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\wallet_rpc_api.py", line 912, in get_transactions
> >     assert record is not None, f"Cannot find coin record for clawback transaction {tx['name']}"
> > AssertionError: Cannot find coin record for clawback transaction 0x1b332cee62521ffe34d04712db291eeb3681e2453ff50811d65d0b87c9d470c3
> > 2023-06-29T11:34:22.696 wallet chia.rpc.rpc_server        : WARNING  Error while handling message: Traceback (most recent call last):
> >   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\rpc_server.py", line 340, in safe_handle
> >     response = await self.ws_api(message)
> >   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\rpc_server.py", line 331, in ws_api
> >     return await f_rpc_api(data)
> >   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\wallet_rpc_api.py", line 1062, in spend_clawback_coins
> >     coin_ids: List[bytes32] = [bytes32.from_hexstr(coin) for coin in request["coin_ids"]]
> >   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\rpc\wallet_rpc_api.py", line 1062, in <listcomp>
> >     coin_ids: List[bytes32] = [bytes32.from_hexstr(coin) for coin in request["coin_ids"]]
> >   File "C:\Users\User\Chia\main+15496\chia-blockchain\chia\util\byte_types.py", line 51, in from_hexstr
> >     if input_str.startswith("0x") or input_str.startswith("0X"):
> > AttributeError: 'NoneType' object has no attribute 'startswith'```
> > ```
> 
> It seems like the clawback coin is missing for some reason. I assume it is related to wallet switching. For the auto claim, did you wait several minutes after the timelock expired?

In this case for autoclaim, I switched to the receiver's wallet before the timelock expired. The autoclaim was then attempted in that wallet automatically as soon as the timelock expired.

wjblanke (approved):
aok

PR #15499, opened 2023-06-13
fix flaky `MAX_FUTURE_TIME` test

arvidn:
It appears that if the test ran too fast, and the (seconds) timestamp did not progress between starting the test and arriving to the block validation, it would not fail as expected.


xdustinface (approved):
Thanks!

trepca (approved):

PR #15510, opened 2023-06-14
tests: Use separate `BlockTools` in farmer/harvester fixtures

xdustinface:
### Purpose:

Not exactly sure how the failure gets triggered to begin with but using a separate blocktools/keychain for the fixtures seems to fix this frequent (where i still wonder why they came up recently) flaky runs like https://github.com/Chia-Network/chia-blockchain/actions/runs/5265202833/jobs/9517534934?pr=15509. 

### Current Behavior:

Farmer/Harvester fixtures use the shared `BlockTools`.

### New Behavior:

Farmer/Harvester fixtures use a separate `BlockTools`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -585,32 +585,32 @@ async def two_nodes_one_block():
 
 
 @pytest_asyncio.fixture(scope="function")
-async def farmer_one_harvester(tmp_path: Path, bt: BlockTools) -> AsyncIterator[Tuple[List[Service], Service]]:
-    async for _ in setup_farmer_multi_harvester(bt, 1, tmp_path, bt.constants, start_services=True):
+async def farmer_one_harvester(tmp_path: Path, get_b_tools: BlockTools) -> AsyncIterator[Tuple[List[Service], Service]]:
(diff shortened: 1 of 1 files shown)


arvidn:
would it be prohibitively expensive to make the "bt" fixture scoped as "function" instead of "session"?

arvidn:
since this touches the keyring, I think @paninaro should take a look.

arvidn:
if this change actually fixes something, I think it would be really good to leave a comment in the code. It's not obvious to a reader why we use multiple copies of this fixture

xdustinface:
There is certainly work to be done around all this block tools fixture stuff because we have a bunch of different ways where and how they are instantiated globally with session scope or other with function scope.

But at the end, this PR is here to fix a frequently failing flaky test and not to fix our block tools situation.

The issue is obviously somehow is related to using the global, shared block tools but using a separate one here in this fixtures is nothing bad i think? And the tests are also not really slower with this so it seems its just a win to merge it.

paninaro:
The changes look good to me from a keyring perspective, assuming that these fixtures aren't intending to share keys. The names of the fixtures lead me to believe that they are intended to be used standalone.

arvidn (approved):

PR #15511, opened 2023-06-14
tests: Run `build_network_protocol_files` as test

xdustinface:
### Purpose:

Even though its already run in the source code uploader to verify nothing changes, this get coverage for it. Could probably be improved to return the change status from `main` and assert nothing changed but this PR is only here to at least run it under tests.

### DIFF
--- a/tests/util/build_network_protocol_files.py
+++ b/tests/util/build_network_protocol_files.py
@@ -255,7 +255,7 @@ def test_protocol_json() -> None:
     return result
 
 
-if __name__ == "__main__":
+def main() -> None:
     get_network_protocol_filename().write_bytes(get_protocol_bytes())
 
     name_to_function = {
@@ -275,3 +275,7 @@ if __name__ == "__main__":
             check=True,
             cwd=tests_dir,
         )
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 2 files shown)


arvidn:
this doesn't seem right. You end up executing this code just to satisfy the test coverage metric, without actually testing anything.

It could make sene to run this as a test if it also ensured nothing changed. but that might be difficult when running the tests during developments, since you can't rely on `git diff` being empty.

I think it would make more sense to not measure coverage of this file. it would probably also make sense to move it to some other directory, to not make it look like it is a test.

xdustinface:
> this doesn't seem right. You end up executing this code just to satisfy the test coverage metric, without actually testing anything.
> 
> It could make sene to run this as a test if it also ensured nothing changed. but that might be difficult when running the tests during developments, since you can't rely on `git diff` being empty.
> 
> I think it would make more sense to not measure coverage of this file. it would probably also make sense to move it to some other directory, to not make it look like it is a test.

Did you read the PR description? :) At the end yes, for now this simply runs the code of that file to get coverage but it still tests at least that there are no issues in the file and the code runs without exception or failure.

Not sure why just ignoring coverage for that file is better than simply running it but i can do that too if you persist on it, sure.

xdustinface:
> Because this change makes it look and behave even more like a test

Actually the build file is not modified to look like a test. It just got a main function which can be imported by a test file but its still the same file as before.

> It's not testing anything.

Well it doesn't assert anything at this point, yes. But it still tests the run-ability of the code :) If there are issues in the code which lead to exceptions the test will fail.

Anyway, i don't really get what your concerns are about this, can you explain?

arvidn (approved):

PR #15520, opened 2023-06-14
farmer: Always signal API readiness

xdustinface:
### Purpose:

This was changed in #15466, explore if this is the reason for the mysterious farmer/harvester RPC test failures.


xdustinface:
Its not :)

PR #15537, opened 2023-06-15
Require caller to specify optional arguments to generate_signed_transaction

aqk:
### Purpose:

Code cleanup

No behaviour changes.

### Testing Notes:

No test changes

### DIFF
--- a/chia/pools/pool_wallet.py
+++ b/chia/pools/pool_wallet.py
@@ -681,10 +681,10 @@ class PoolWallet:
             amount,
             genesis_launcher_puz.get_tree_hash(),
             fee,
-            coins,
-            None,
-            False,
-            announcement_set,
+            coins=coins,
+            primaries=None,
+            ignore_max_send_amount=False,
+            coin_announcements_to_consume=announcement_set,
             origin_id=launcher_parent.name(),
         )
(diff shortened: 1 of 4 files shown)


Quexington:
I'm cautiously pro-this as a step in the right direction towards just making everything be specified, but what's the reason for doing this in the immediate term rather than taking the time to force all of the arguments to be passed via name?

Quexington:
Oh, also if we're going to do this, we should do it to all GSTs yeah?

aqk:
> Oh, also if we're going to do this, we should do it to all GSTs yeah?

GST?

altendky:
`GST` -> `generate_signed_transaction`.  It's a method on many (all?) of the wallet classes.

Quexington:
> > I'm cautiously pro-this as a step in the right direction towards just making everything be specified, but what's the reason for doing this in the immediate term rather than taking the time to force all of the arguments to be passed via name?
> 
> I don't mind adding it to all the arguments - do you think that would help?
> 
> By the way, that would mean changing 231 additional call sites.

Yeah it would not be a small task but if there's no burning need, I'd prefer to see the full step rather than the half step.

PR #15538, opened 2023-06-15
Rephrase async contextmanager DBConnection to use asynccontextmanager

aqk:
https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager

### Purpose:

Code improvement suggested from https://github.com/Chia-Network/chia-blockchain/pull/14946 a PR meant to improve type-checking.

This PR updates DBConnection to use [asynccontextmanager](https://docs.python.org/3/library/contextlib.html#contextlib.asynccontextmanager)

No new behaviour.


altendky (approved):
context managers are the way, and this is the way to write them...

arvidn (approved):

PR #15540, opened 2023-06-15
[DO NOT MERGE] Review DAO chialisp

Quexington:
I classified my review comments for easy searchable-ness.
```
[REVIEW - architecture]
[REVIEW - bugs]
[REVIEW - comments]
[REVIEW - optimization]
[REVIEW - misc]
```
The vast majority are optimization, I think there is a _lot_ of cost to be saved here with an optimization pass.

Also one big comment more generally is that I think the test coverage could be vastly improved.  There's a lot of test cases but this infrastructure is so complex that I'm not sure all of the edge cases (and some not-so-edge cases) are tested very well.  I would really like to see `SpendSim` style tests that actually validate this code in the mempool.


chia/wallet/puzzles/dao_finished_state.clsp
```diff
@@ -26,9 +28,14 @@
...
     )
     (list
+      ; [REVIEW - optimization] What's the point of this assertion.  This coin creates its own inner puzhash, it's not like it could lie, right?
+      ; [^] Perhaps the purpose is to make sure that this puzzle can't exist outside of a singleton, but anything that is asserting an announcement
+      ; [^] from this puzzle is also going to have to calculate that it is in fact inside of a singleton.
       (list ASSERT_MY_PUZZLEHASH (wrap_in_singleton SINGLETON_STRUCT my_inner_puzhash))
+      ; [REVIEW - optimization] This is never going to be able to pay out right?  Why not just force this to 0 and save a solution argument?
       (list ASSERT_MY_AMOUNT my_amount)
       (list CREATE_COIN my_inner_puzhash my_amount)
+      ; [REVIEW - optimization] This is the only announcement that this puzzle could create.  Why the special 'f'?  It should probably just be ().
```
matt-o-how, line 38: because we assert that the singleton is making an `f` announcement elsewhere

chia/wallet/puzzles/dao_finished_state.clsp
```diff
@@ -26,9 +28,14 @@
       (my_inner_puzhash (recreate_self SINGLETON_STRUCT DAO_FINISHED_STATE_MOD_HASH))
     )
     (list
+      ; [REVIEW - optimization] What's the point of this assertion.  This coin creates its own inner puzhash, it's not like it could lie, right?
+      ; [^] Perhaps the purpose is to make sure that this puzzle can't exist outside of a singleton, but anything that is asserting an announcement
+      ; [^] from this puzzle is also going to have to calculate that it is in fact inside of a singleton.
       (list ASSERT_MY_PUZZLEHASH (wrap_in_singleton SINGLETON_STRUCT my_inner_puzhash))
+      ; [REVIEW - optimization] This is never going to be able to pay out right?  Why not just force this to 0 and save a solution argument?
```
matt-o-how, line 35: this is inside a singleton

chia/wallet/puzzles/dao_finished_state.clsp
Quexington: But that thing could assert `()` as well right? What's special about `f`?

chia/wallet/puzzles/dao_treasury.clsp
```diff
@@ -85,6 +98,7 @@
                     PASS_MARGIN
                     proposal_announcement
                     proposal_validator_solution
+                    ; [REVIEW - architecture] does the proposal validator really need the conditions here?  I would assume it's checking the proposal's puzzle which means that it knows the likely set of conditions it could possibly output.
```
matt-o-how, line 101: We might want validators that prevent certain conditions or certain types of announcements

chia/wallet/puzzles/p2_singleton_via_delegated_puzzle.clsp
```diff
@@ -32,10 +40,15 @@
...
         (cons_announcements_to_output
+          ; [REVIEW - optimization] why do we care if they have the same puzzle hash as us.
+          ; [^] _They_ should only care that _we_ are the same puzhash as them (so they don't get spent by some rando).
+          ; [^] This as it sits blocks potentially folding things that could be meltable in.
+          ; [^] The alternative would be to pass in a list of coin_id/amount pairs and have the asserting p2_singletons assert that their amount is accurate.
           (calculate_coin_id (f (f list_of_parent_amounts)) my_puzhash (f (r (f list_of_parent_amounts))))  ; asserts that all the coins have the same puzzle as me
           (for_parent_amounts my_puzhash (r list_of_parent_amounts) (+ total (f (r (f list_of_parent_amounts)))))
         )
         (list
+          ; [REVIEW - comments] I think higher would be fine, there's probably no burning need for aggregation on the part of an honest party
```
matt-o-how, line 51: The worry is about dust storms from dishonest parties, but we also want to prevent DoS via aggregation too.

chia/wallet/puzzles/dao_update_proposal.clsp
```diff
@@ -1,3 +1,5 @@
+; [REVIEW - misc] I think these files should be grouped into a folder or have a prefix or something to classify them as potential proposals
+; [REVIEW - optimization] That said, I don't think this needs to be a file at all.  Just `(1, ((CREATE_COIN puzzle_hash_that_can_be_calculated_and_verified_off_chain ONE)))`.
```
matt-o-how, line 2: as you've identified elsewhere in your review, we sacrifice optimisation for the sake of letting non-technical users have a reasonably safe integrated experience. Having known proposal types that can be parsed by the wallet that are revealed on chain is the main one of these

chia/wallet/puzzles/dao_proposal_timer.clsp
```diff
@@ -40,6 +40,11 @@
   ; proposal_yes_votes
   ; proposal_total_votes
   ; proposal_innerpuzhash
+  ; [REVIEW - optimization] We're passing in a _lot_ of stuff here to simply validate the proposal timelock is correct
+  ; [^] Could we either a) curry in the timelock to this puzzle
+  ; [^] b) have the proposal announce the timelock to us and verify it that way?
```
matt-o-how, line 45: this is what's happening

chia/wallet/puzzles/dao_proposal_timer.clsp
```diff
@@ -40,6 +40,11 @@
...
+  ; [REVIEW - optimization] Similar vein as above, but couldn't the proposal also verify our parent ID since it already has access to it's own singleton info?
```
matt-o-how, line 46: Our parent is the proposal when it had 0 votes. We assert this so we know that the timer was not cheated by being made earlier. We don't know what the timelock amount is at this time though.

chia/wallet/puzzles/dao_treasury.clsp
Quexington: Hmm yeah, I was thinking that the proposal could just check that, but I guess the proposal doesn't normally run the delegated puzzle it's announcing.

Quexington: Well actually there's still some argument to omit this.  The security of the delegated puzzle is that it gets voted on.  If there are certain conditions that are banned, that could be a layer in the dpuz, and the proposal could check that the dpuz is of that format.  All you'd have to change in the treasury for this is the PROPOSAL_MOD instead of the whole PROPOSAL_VALIDATOR which is potentially more dangerous.  Additionally, it optimizes most of the other spend cases where a dpuz is allowed to output whatever it wants.  I don't feel super strongly about this though, I just think it would be a little bit better.

chia/wallet/puzzles/p2_singleton_via_delegated_puzzle.clsp
Quexington: Right that's what I'm saying, dust is usually coming from a dishonest party so honest parties only need to do a couple of aggregations at most.  Trying to pick up the dust would be annoying, but how likely are you really to need to do that?

chia/wallet/puzzles/dao_update_proposal.clsp
Quexington: So as a user, what am I going to be getting/checking here?  My understanding is this:
1) I detect a proposal to vote on, which has a curried in dpuz hash
2) I get some information separately that the mod hash is this file and what the curried args are
3) I validate the args are good to me
4) I curry the args into the mod hash and check that it matches the proposed dpuz

But for some reason after all of that validation, we make full nodes do the tree hash of the new DAO for us.  What I'm proposing is:
1) I detect a proposal to vote on, which has a curried in dpuz hash
2) I get some information separately that it's an update proposal with the known form  `(1, ((CREATE_COIN new_dao_hash ONE)))` and what the curried args are
3) I validate the args are good to me
4) I curry the args into the treasury mod hash, stick it in the format above, and make sure that matches the proposed dpuz

It's not anymore complex, but you don't have to do the `puzzle-hash-of-a-curried-function`.

chia/wallet/puzzles/dao_proposal_timer.clsp
Quexington: Er, perhaps I meant that we can just get away with currying in the proposal's singleton information since none of the inner puzzle stuff matters to us, so you can just pass in the whole hash of that.

chia/wallet/puzzles/dao_proposal_timer.clsp
Quexington: Right but the proposal also knows what it's ID was when it had zero votes so it could do ASSERT_COIN_ANNOUNCEMENT and check our parent for us and then we don't need to curry in any of the inner puzzle stuff, this puzzle becomes "I self destruct w/ a relative timelock when a curried singleton tells me to".

chia/wallet/puzzles/dao_treasury.clsp
```diff
@@ -42,12 +51,15 @@
...
   (defun calculate_singleton_puzzle_hash (SINGLETON_STRUCT inner_puzzle_hash)
-    (puzzle-hash-of-curried-function (f SINGLETON_STRUCT)
+    (puzzle-hash-of-curried-function (f SINGLETON_STRUCT)  ; [REVIEW - misc]  I think curry_hashes or curry_hashes_inline here will be less expensive/have a more sensible API.  I think we're trying to deprecate this function.
       inner_puzzle_hash
       (sha256tree SINGLETON_STRUCT)
     )
   )
 
+  ; [REVIEW - comment] I'm not sure why this is a thing.  A comment here would be good.
+  ; [REVIEW - optimization] I could be confused (see comment above) but this seems to be a hack in place for the sake of wallet complexity when spending the thing for the first time after launch
```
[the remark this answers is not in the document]
matt-o-how: How would you create the eve spend when you can't send it at the same time as the creation and there is no wallet to track if the creation step has succeeded?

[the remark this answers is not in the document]
Quexington: You can subscribe to the eve coin's ID and create the wallet at the same time as that (or do the final transaction and then create the wallet once everything has completed) yeah?

PR #15541, opened 2023-06-15
Fix get transactions CLI for Clawback

ytx1991:
### Purpose:
Fix the get transactions CLI while the wallet is not fully synced and the coin record for Clawback transactions is missing.

### Current Behavior:
Throw exception

### New Behavior:
Skip the clawback transactions

### Testing Notes:
CLI


trepca (approved):
not seeing the error anymore

PR #15543, opened 2023-06-16
Add config option to accept private network addresses from introducers

cmmarslender:
### Purpose:
In some cases, we have test networks that run on private networks and we need to be able to automatically discover peers on private IP addresses

### Current Behavior:
Nodes wont connect to private IPs gossiped from introducers/peers

### New Behavior:
Nodes will connect to private IPs gossiped from introducers/peers IF the option is set to True in the config. Defaults to False in config and if the config entry is missing, to retain the old behavior


hoffmang9 (dismissed):
lgtm

wallentx:
Nice!

emlowe (approved):

PR #15546, opened 2023-06-16
Add more cases to verify_and_get_quality_string() unit tests

AmineKhaldi:


tests/core/custom_types/test_proof_of_space.py
```diff
@@ -68,7 +70,33 @@ class ProofOfSpaceCase:
         expected_error="Calculated pos challenge doesn't match the provided one",
     ),
     ProofOfSpaceCase(
-        id="Not passing the plot filter",
+        id="Neither height nor filter prefix bits",
+        pos_challenge=bytes32.from_hexstr("3e46235f4b78237763f4ac2ed59515e7aae9c5b6e40c5c9ccec406bfbc30ed2e"),
+        plot_size=uint8(42),
+        pool_public_key=G1Element(),
+        plot_public_key=G1Element(),
+        expected_error="Expected either height or filter_prefix_bits",
```
arvidn, line 78: this test case is a little bit odd I think. it's suggesting that throwing an `AssertionError` exception is part of its interface. I consider assertion failures rather like a contract violation (of the precondition). If you violate the contract, anything can happen, and you're not necessarily guaranteed you can recover from it.

But perhaps the python-world is different, since assertions *are* exceptions that you can recover from.

arvidn (approved):

PR #15547, opened 2023-06-16
wallet: Some suggestions from #15274 for #15259

xdustinface:
### Purpose:

- Introduce `HintedCoin`
- Introduce `compute_hint_for_coin` (which now makes it fail with an `ValueError` if there is no hint for the requested coin instead of silently moving on)
- Add some unit test

### DIFF
--- a/chia/wallet/util/compute_hints.py
+++ b/chia/wallet/util/compute_hints.py
@@ -35,13 +35,3 @@ def compute_spend_hints_and_additions(cs: CoinSpend) -> Dict[bytes32, HintedCoin
             hinted_coins[bytes32(coin.name())] = HintedCoin(coin, hint)
 
     return hinted_coins
-
-
-def compute_hint_for_coin(coin_name: bytes32, coin_spend: CoinSpend) -> bytes32:
-    try:
-        hinted_coin = compute_spend_hints_and_additions(coin_spend)[coin_name]
-    except KeyError:
-        raise ValueError(f"{coin_name} not found in {coin_spend}")
-    if hinted_coin.hint is None:
-        raise ValueError(f"No hint found for {coin_name} in {coin_spend}")
-    return hinted_coin.hint
--- a/chia/wallet/util/compute_hints.py
+++ b/chia/wallet/util/compute_hints.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import Dict, Tuple
+from dataclasses import dataclass
+from typing import List, Optional
 
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.program import INFINITE_COST, Program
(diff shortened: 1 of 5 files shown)


chia/wallet/util/compute_hints.py
```diff
@@ -10,22 +11,36 @@
...
+
 
-    return hint_dict, coin_dict
+def compute_hint_for_coin(coin_name: bytes32, coin_spend: CoinSpend) -> bytes32:
+    for hinted_coin in compute_spend_hints_and_additions(coin_spend):
+        if hinted_coin.coin.name() == coin_name:
+            if hinted_coin.hint is None:
+                raise ValueError(f"No hint found for {coin_name} in {coin_spend}")
+            return hinted_coin.hint
+    raise ValueError(f"{coin_name} not found in {coin_spend}")
```
Quexington, line 46: I don't like this function existing.  It might tempt someone to run it in a loop and end up running the spend a bunch of times.

Quexington (changes requested):
I'm still not super convinced this is the right way to go and am not a fan of having to import `HintedCoin` everywhere, but I'd be willing to compromise if the function returns a `Dict[bytes32, HintedCoin]` (where the bytes32 is the coin's name) just so that you can easily look up the coin you wanted if you'd like.

chia/wallet/util/compute_hints.py
xdustinface: I think the `compute_` prefix signaled this might be a heavy thing and im not sure how i feel about avoiding helpers because they could be misused which can be avoided via proper reviews also. Anyway, i dropped and in the relevant cases on the call site i think we can already assume the coin is in the spend and just access it and also we can assert the hint is there and just have a assertion message for the case there isn't? See  [b9186ed](https://github.com/Chia-Network/chia-blockchain/pull/15547/commits/b9186ed2ba3c8b6116557c065248669b1bd77afb).

chia/wallet/wallet_state_manager.py
```diff
@@ -855,12 +855,9 @@ async def handle_cat(
...
         mod_hash, tail_hash, inner_puzzle = curried_args
 
-        hint_dict, _ = compute_spend_hints_and_additions(coin_spend)
-        coin_name: bytes32 = bytes32(coin_state.coin.name())
-        if coin_name in hint_dict:
-            derivation_record = await self.puzzle_store.get_derivation_record_for_puzzle_hash(hint_dict[coin_name])
-        else:
-            derivation_record = None
+        hinted_coin = compute_spend_hints_and_additions(coin_spend)[coin_state.coin.name()]
+        assert hinted_coin.hint is not None, f"hint missing for coin {hinted_coin.coin}"
```
Quexington, line 859: Sorry, one last comment here, I believe I've commented this on some PR at some point, but I think raising here is not advisable because we might try to sync some unhinted change or something and this is a perfectly normal failure and we want to handle it gracefully rather than flood a user's logs with this as a warning.

Quexington: The DID one below might be fine, but for the sake of consistency maybe it's better to fix that one too.

xdustinface: > this is a perfectly normal failure

Why is this perfectly normal? From my understanding not hinting CATs is and will ever be an issue and if we again due to some bug somewhere start to not hint CATs or some other wallet misses it we at least will be alerted about it this way.

And why do you think its better/less confusing for users to log a bunch of `Received state for the coin that doesn't belong to us` messages (which happens if we go your way) which basically is not true here instead of logging the actual issue?

xdustinface: Which brings me to the question, how do we even sync unhinted CATs? Yes, we have a test which subscribes to them so that we at least get them synced but if there is no hint we can not figure out the derivation record here anyway and we just step out of it and assume it doesn't belong to us, no?

Quexington: Yeah my bad, I didn't have it fully swapped into my brain, but you're right we pick up change via the puzhash in the DB, we shouldn't be going through this function.

I do still feel that this should not be a raise because you're basically short circuiting any potential logic for "we couldn't determine the coin type here".  Looking at the existing logic though, this seems fine currently, I just think it's a bad choice going forward were someone to add something that we are unaware of currently (maybe a like "we don't know what this is" store or something).

Since it works in the existing code though and I'd really just like to get this merged, you can resolve if you disagree.

Quexington (approved):

chia/wallet/wallet_state_manager.py
xdustinface: > we pick up change via the puzhash in the DB, we shouldn't be going through this function.

Oh right so this works because we still derive/store the puzzle hashes and then we get the wallet identifier from looking up the puzzle store. But we still run into `handle_cat` if the CAT wasn't added yet? So wouldn't syncing change also fail if we sync the change coin as first coin of the given CAT, i.e. out of order?

> I do still feel that this should not be a raise because you're basically short circuiting any potential logic for "we couldn't determine the coin type here".

Not sure i get what you mean here.

> And why do you think its better/less confusing for users to log a bunch of Received state for the coin that doesn't belong to us messages (which happens if we go your way) which basically is not true here instead of logging the actual issue?

Did you read my comment ^ about logging the actual issue vs logging "this doesn't belong to us" what you propose?

Quexington: > Oh right so this works because we still derive/store the puzzle hashes and then we get the wallet identifier from looking up the puzzle store. But we still run into handle_cat if the CAT wasn't added yet? So wouldn't syncing change also fail if we sync the change coin as first coin of the given CAT, i.e. out of order?

Right, we need at least one hinted CAT, but the thing is that it's actually impossible for change to come out of order now because the only way we pick it up is when the parent gets spent.  Even if it did come out of order, I think the new subscription when the parent gets spent will pick it up again (it's a coin ID subscription).

> Not sure i get what you mean here.

Right now, we log something then return None from the `determine_coin_type` function and then the loop continues.  Right now almost nothing happens and we just continue to the next state, but it's possible that in the future someone might want to do something different and they'll have to undo your change here in order to do so.

> Did you read my comment ^ about logging the actual issue vs logging "this doesn't belong to us" what you propose?
Well if it's not change like we've identified, then that is the correct error message.  "I got this state, but I can't verify it belongs to me (because it wasn't hinted and I don't know what the inner hash is)".

xdustinface: > Right now, we log something then return None from the `determine_coin_type` function and then the loop continues. Right now almost nothing happens and we just continue to the next state

Now we properly log an error, which is what we should do in a case a not hinted CAT comes in imo, and then we continue with the loop just like we would in your case.

The difference is in your case we would log something which isn't helpful in any way and silently step over it. It doesn't allow to distinguish between unhinted CATs and CATs which really don't belong to our wallet, it would both look the same and we wouldn't become noticed about coin states we can't process because of this issue but instead it would like they just don't belong to us.

> but it's possible that in the future someone might want to do something different and they'll have to undo your change here in order to do so.

Maybe i miss something here but isn't the only thing which would make sense to do different in this case trying to fetch the spend of this coin and see if we can get the puzzle to identify the CAT which should happen in `handle_cat` anyway? 

> I got this state, but I can't verify it belongs to me (because it wasn't hinted and I don't know what the inner hash is)"

Thats not what the log would say though. I just logs an `INFO` with `Received state for the coin that doesn't belong to us`.

xdustinface: Btw, the failing CI is, i think, only because your branch doesn't contain the mypy 1.4 fixes which recently came up so if you merge it in it should be fine because your branch goes into `main`.

PR #15550, opened 2023-06-16
simplify db connection helper

altendky:


arvidn:
it doesn't look like a clear cut simplification. there's more code and the user of the class now need to provide a path (which I don't believe the caller really cares about)

altendky:
The simplification is proper usage of temporary files without the multiple prior race conditions.  The passing in of the temporary directory is so we can get it from pytest and let it exist where pytest temporary stuff ought to.  I looked around for a way to get the pytest dir from the ether, but failed.  Maybe I just missed it.

But sure, there are costs.

PR #15555, opened 2023-06-20
Raise full node stores tests timeout

AmineKhaldi:


arvidn:
I don't believe these tests timed-out. I believe the benchmark test timed out and cancelled these tests as the whole CI run failed.
The benchmark was parameterized by the consensus mode parameter we introduced. I added a commit to run the benchmark just once.

arvidn:
I think we can close this

PR #15559, opened 2023-06-20
Dao wallet

matt-o-how:
### Purpose:
Merge DAO Wallet into main

### Testing Notes:
tests/wallet/dao_wallet


chia/cmds/dao_funcs.py
```diff
@@ -0,0 +1,456 @@
...
+    print("############################")
+    for prop in proposals:
+        print("Proposal ID: {proposal_id}".format(**prop))
+        prop_status = "CLOSED" if prop["closed"] else "OPEN"
+        print(f"Status: {prop_status}")
+        print("Votes for: {yes_votes}".format(**prop))
+        votes_against = prop["amount_voted"] - prop["yes_votes"]
+        print(f"Votes against: {votes_against}")
+        print("------------------------")
+    print(f"Proposals have {soft_close_length} blocks of soft close time.")
```
trepca, line 161: what would `soft_close_length` be if proposal is already closed?

chia/cmds/dao_funcs.py
```diff
@@ -0,0 +1,456 @@
...
+        fee=final_fee,
+        self_destruct=self_destruct,
+        reuse_puzhash=reuse_puzhash,
+    )
+    # dao_close_proposal(self, wallet_id: int, proposal_id: str, fee: uint64 = uint64(0))
+    if res["success"]:
+        name = res["tx_id"]
+        print(f"Submitted proposal close transaction with name: {name}")
+    else:
+        print("Unable to generate close transaction.")
```
trepca, line 272: could we provide more info here on why it's not able to close?

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -176,6 +176,7 @@ async def create_new_cat_wallet(
         chia_tx = dataclasses.replace(chia_tx, spend_bundle=spend_bundle)
         await self.standard_wallet.push_transaction(chia_tx)
         await self.standard_wallet.push_transaction(cat_record)
+        # breakpoint()
```
trepca, line 179: let remove this

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -806,7 +807,7 @@ async def generate_unsigned_spendbundle(
             chia_tx,
         )
 
-    async def generate_signed_transaction(
+    async def generate_signed_transactions(
```
trepca, line 810: @Quexington is this ok? to rename this

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -820,6 +821,7 @@ async def generate_signed_transaction(
         max_coin_amount: Optional[uint64] = None,
         excluded_coin_amounts: Optional[List[uint64]] = None,
         reuse_puzhash: Optional[bool] = None,
+        override_memos: Optional[bool] = None,
```
trepca, line 824: why do we need this? this functions already accepts memos?

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+        PROPOSAL_MINIMUM_AMOUNT.as_int(),
+    )
+    return dao_rules
+
+
+def get_spend_p2_singleton_puzzle(
+    treasury_id: bytes32, xch_conditions: Optional[List], asset_conditions: Optional[List[Tuple]]  # type: ignore
+) -> Program:
+    # TODO: typecheck get_spend_p2_singleton_puzzle arguments
+    # TODO: add tests for get_spend_p2_singleton_puzzle: pass xch_conditions as Puzzle, List and ConditionWithArgs
```
trepca, line 183: has this been done? (the todos)

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+    )
+    puzzle = self_hash.curry(
+        self_hash.get_tree_hash(),
+        previous_votes_list,  # TODO: maybe format check this in this function
+        innerpuz,
+    )
+    return puzzle
+
+
+def add_proposal_to_active_list(
```
trepca, line 239: what's an active list? can we document this function?

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+    *,
+    proposal_id: bytes32,
+    cat_tail_hash: bytes32,
+    treasury_id: bytes32,
+    votes_sum: uint64,
+    total_votes: uint64,
+    proposed_puzzle_hash: bytes32,
+) -> Program:
+    """
+    spend_or_update_flag can take on the following values, ranked from safest to most dangerous:
```
trepca, line 305: spend_or_update_flag? where is that?

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+    ) = curried_args
+    curried_args = uncurry_proposal_validator(proposal_validator)
+    (
+        SINGLETON_STRUCT,
+        PROPOSAL_SELF_HASH,
+        PROPOSAL_MINIMUM_AMOUNT,
+        PAYOUT_PUZHASH,
+    ) = curried_args.as_iter()
+    return DAORules(
+        uint64(int_from_bytes(proposal_timelock.as_atom())),
```
trepca, line 404: use `as_int()` instead of `as_atom` here

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+        uint64(int_from_bytes(soft_close_length.as_atom())),
+        uint64(int_from_bytes(attendance_required.as_atom())),
+        uint64(int_from_bytes(pass_percentage.as_atom())),
+        uint64(int_from_bytes(self_destruct_length.as_atom())),
+        uint64(int_from_bytes(oracle_spend_delay.as_atom())),
+        uint64(PROPOSAL_MINIMUM_AMOUNT.as_int()),
+    )
+
+
+# This takes the treasury puzzle and treasury solution, not the full puzzle and full solution
```
trepca, line 414: lets make this as docstring so can be discoverable

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+    #     DAO_TREASURY_MOD_HASH,
+    #     DAO_PROPOSAL_VALIDATOR_MOD,
+    #     proposal_timelock,
+    #     soft_close_length,
+    #     attendance_required_percentage,
+    #     proposal_pass_percentage,
+    #     proposal_self_destruct_length,
+    #     oracle_spend_delay,
+    # ) = curried_args
+    if solution.rest().rest().first() != Program.to(0):
```
trepca, line 428: `solution.at("rrf")`

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+    else:
+        # Oracle Spend - treasury is unchanged
+        return puzzle_reveal
+
+
+# This takes the proposal puzzle and proposal solution, not the full puzzle and full solution
+# This also returns the proposal puzzle and not the full puzzle
+def get_new_puzzle_from_proposal_solution(puzzle_reveal: Program, solution: Program) -> Optional[Program]:
+    # Check if soft_close_length is in solution. If not, then add votes, otherwise close proposal
+    if len(solution.as_python()) == 1:
```
trepca, line 463: `.as_python` will be deprecated as can cause DOS attack, can we use something else here?

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+        raise ValueError("Unrecognised proposal type")
+
+
+def generate_cat_tail(genesis_coin_id: bytes32, treasury_id: bytes32) -> Program:
+    dao_cat_launcher = create_cat_launcher_for_singleton_id(treasury_id).get_tree_hash()
+    puzzle = DAO_CAT_TAIL.curry(genesis_coin_id, dao_cat_launcher)
+    return puzzle
+
+
+# TODO: move curry_singleton to chia.wallet.singleton
```
trepca, line 641: have these TODOs been done?

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+    """
+    try:
+        if mod == SINGLETON_MOD:
+            mod, curried_args = curried_args.rest().first().uncurry()
+            if mod == DAO_TREASURY_MOD:
+                return curried_args.first().as_iter()  # type: ignore[no-any-return]
+    except ValueError:
+        import traceback
+
+        print(f"exception: {traceback.format_exc()}")
```
trepca, line 686: why do we print here?

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+
+    return None
+
+
+# This is for use in the WalletStateManager to determine the type of coin received
```
trepca, line 691: docstring

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+            if mod == DAO_FINISHED_STATE:
+                return curried_args.as_iter()  # type: ignore[no-any-return]
+    except ValueError:
+        import traceback
+
+        print(f"exception: {traceback.format_exc()}")
+    return None
+
+
+# This is used in WSM to determine whether we have a dao funding spend
```
trepca, line 729: docstring

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+def match_funding_puzzle(uncurried: UncurriedPuzzle, solution: Program) -> Optional[bool]:
+    # TODO: handle case where solution is for existing p2_singleton
```
trepca, line 731: completed?

chia/wallet/dao_wallet/dao_utils.py
```diff
@@ -0,0 +1,763 @@
...
+    try:
+        if match_cat_puzzle(uncurried):
```
trepca, line 733: can we comment what's happening in this if ?

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+            current_height=uint64(0),
+        )
+        self.dao_rules = dao_rules
+        info_as_string = json.dumps(self.dao_info.to_json_dict())
+        self.wallet_info = await wallet_state_manager.user_store.create_wallet(
+            name, WalletType.DAO.value, info_as_string
+        )
+        self.wallet_id = self.wallet_info.id
+        std_wallet_id = self.standard_wallet.wallet_id
+        bal = await wallet_state_manager.get_confirmed_balance_for_wallet(std_wallet_id)
```
trepca, line 180: why do we fetch balance again here?

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+        self = DAOWallet()
+        self.wallet_state_manager = wallet_state_manager
+        if name is None:
+            name = self.generate_wallet_name()
+
+        self.standard_wallet = wallet
+        self.log = logging.getLogger(name if name else __name__)
+        std_wallet_id = self.standard_wallet.wallet_id
+        bal = await wallet_state_manager.get_confirmed_balance_for_wallet(std_wallet_id)
+        if amount_of_cats > bal:
```
trepca, line 157: 1 cat = 1 mojo in this case?

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+        )
+        info_as_string = json.dumps(self.dao_info.to_json_dict())
+        self.wallet_info = await wallet_state_manager.user_store.create_wallet(
+            name, WalletType.DAO.value, info_as_string
+        )
+        await self.resync_treasury_state()
+        await self.wallet_state_manager.add_new_wallet(self)
+        await self.save_info(self.dao_info)
+        if self.wallet_info is None:
+            raise ValueError("Internal Error")
```
trepca, line 252: why would this happen?

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+        return self
+
+    @classmethod
+    def type(cls) -> WalletType:
+        return WalletType.DAO
+
+    def id(self) -> uint32:
+        return self.wallet_info.id
+
+    def puzzle_for_pk(self, pubkey: G1Element) -> Program:
```
trepca, line 302: why do we need this method since it's a constant?

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+        parent_info = None
+        for name, ccparent in self.dao_info.parent_info:
+            if name == coin.parent_coin_info:
+                parent_info = ccparent
+        return parent_info
+
+    async def get_frozen_amount(self) -> uint64:
+        return uint64(0)
+
+    async def get_max_send_amount(self, records: Optional[Set[WalletCoinRecord]] = None) -> uint128:
```
trepca, line 339: so DAO can't be transferred to a different wallet?

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+            raise ValueError(f"Not enough of that asset_id: {asset_id}")
+        return coins
+
+    async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection) -> None:
+        """
+        Notification from wallet state manager that a coin has been received.
+        This can be either a treasury coin update or funds added to the treasury
+        """
+        self.log.info(f"DAOWallet.coin_added() called with the coin: {coin.name()}:{coin}.")
+        wallet_node: Any = self.wallet_state_manager.wallet_node
```
trepca, line 444: why any and not WalletNode?

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+            else:
+                # funding coin
+                asset_id = get_asset_id_from_puzzle(parent_spend.puzzle_reveal.to_program())
+                if asset_id not in self.dao_info.assets:
+                    new_asset_list = self.dao_info.assets.copy()
+                    new_asset_list.append(asset_id)
+                    dao_info = dataclasses.replace(self.dao_info, assets=new_asset_list)
+                    await self.save_info(dao_info)
+        except Exception as e:
+            self.log.error(f"Error occurred during dao wallet coin addition: {e}")
```
trepca, line 466: log.exception

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+        return
+
+    async def resync_treasury_state(self) -> None:
+        parent_coin_id: bytes32 = self.dao_info.treasury_id
+        wallet_node: Any = self.wallet_state_manager.wallet_node
+        peer: WSChiaConnection = wallet_node.get_full_node_peer()
+        if peer is None:
+            raise ValueError("Could not find any peers to request puzzle and solution from")
+
+        children = await wallet_node.fetch_children(parent_coin_id, peer)
```
trepca, line 496: why do you fetch children here?
you could just have `while True` loop and break as you are breaking now in loop below?

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+            prop_info for prop_info in self.dao_info.proposals_list
+            if not prop_info.closed
+            or prop_info.closed is None
+            or any(prop_info.proposal_id in lci.active_votes for lci in dao_cat_wallet.dao_cat_info.locked_coins)
+        ]
+        dao_info = dataclasses.replace(self.dao_info, proposals_list=new_list)
+        await self.save_info(dao_info)
+        return
+
+    async def resync_treasury_state(self) -> None:
```
trepca, line 489: not sure what this does, please add a docstring

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+        return eve_spend_bundle
+
+    # This has to be in the wallet because we are taking an ID and then searching our stored proposals for that ID
+    def get_proposal_curry_values(self, proposal_id: bytes32) -> Tuple[Program, Program, Program]:
+        for prop in self.dao_info.proposals_list:
+            if prop.proposal_id == proposal_id:
+                return get_curry_vals_from_proposal_puzzle(prop.inner_puzzle)
+        raise ValueError("proposal not found")
+
+    # This has to be in the wallet because we are taking an ID and then searching our stored proposals for that ID
```
trepca, line 863: duplicate of the above, not sure why it's needed

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+        xch_conditions = [
+            [
+                51,
+                cat_launcher.get_tree_hash(),
+                uint64(amount_of_cats_to_create),
+                [cats_new_innerpuzhash],
+            ],  # create cat_launcher coin
+            [
+                60,
+                Program.to(["m", full_puz.get_tree_hash()]).get_tree_hash(),
```
trepca, line 914: I assume `m` is mint? can we make it into an enum so it's clear?

tests/wallet/dao_wallet/test_dao_clvm.py
```diff
@@ -0,0 +1,973 @@
...
+            200,
+            Program.to("parent_parent").get_tree_hash(),
+        ]
+    )
+    conds: Program = proposal_timer_full.run(solution)
+    assert len(conds.as_python()) == 4
+
+
+def test_validator() -> None:
+    # This test covers proposal_validator and spend_p2_singleton
```
trepca, line 381: not sure what this means, what test cases and scenarios does this cover?

tests/wallet/dao_wallet/test_dao_clvm.py
```diff
@@ -0,0 +1,973 @@
...
+    merged_coin_id = Coin(fake_parent_id, my_puzhash, 200).name()
+    merge_sol = Program.to([[my_id, my_puzhash, 100, [[fake_parent_id, my_puzhash, 200]]]])
+    conds = conditions_dict_for_solution(p2_singleton, merge_sol, INFINITE_COST)
+    assert len(conds) == 5
+    assert conds[ConditionOpcode.ASSERT_COIN_ANNOUNCEMENT][0].vars[0] == Program.to([merged_coin_id, 0]).get_tree_hash()
+    assert conds[ConditionOpcode.CREATE_COIN][0].vars[1] == int_to_bytes(300)
+    return
+
+
+def test_treasury() -> None:
```
trepca, line 543: hard to see which scenarios we're testing here, can you describe here please or create separate tests for each

tests/wallet/dao_wallet/test_dao_wallets.py
```diff
@@ -0,0 +1,2067 @@
...
+    await full_node_api.process_transaction_records(records=[tx_record])
+    await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hash_0))
+
+    # Check the spend was successful
+    treasury_id = dao_wallet_0.dao_info.treasury_id
+    # Farm enough blocks to pass the oracle_spend_delay and then complete the treasury eve spend
+    for i in range(1, 11):
+        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hash_0))
+    if not trusted:
+        await asyncio.sleep(1)
```
trepca, line 157: ?

tests/wallet/dao_wallet/test_dao_wallets.py
```diff
@@ -0,0 +1,2067 @@
...
+    dao_cat_amt = uint64(100)
+    async with wallet_node_0.wallet_state_manager.lock:
+        txs, new_coins = await dao_wallet_0.create_new_dao_cats(dao_cat_amt, push=True)
+    sb = txs[0].spend_bundle
+    await time_out_assert_not_none(5, full_node_api.full_node.mempool_manager.get_spendbundle, sb.name())
+    await full_node_api.process_transaction_records(records=txs)
+
+    # Give the full node a moment to catch up if there are no trusted peers
+    if not trusted:
+        await asyncio.sleep(1)
```
trepca, line 201: this will make this test flaky, wait for something instead of doing sleep

tests/wallet/dao_wallet/test_dao_wallets.py
```diff
@@ -0,0 +1,2067 @@
...
+    assert dao_wallet_0.puzzle_hash_for_pk(G1Element()) == Program.to(0).get_tree_hash()
+    assert (await dao_wallet_0.get_new_puzzle()) == Program.to(0)
+
+    await dao_wallet_0.set_name("Renamed Wallet")
+    assert dao_wallet_0.get_name() == "Renamed Wallet"
+
+    new_inner_puzhash = await dao_wallet_0.get_new_p2_inner_hash()
+    assert isinstance(new_inner_puzhash, bytes32)
+
+    # # generate new dao spends with bad cat details
```
trepca, line 244: why is this commented out?

tests/wallet/dao_wallet/test_dao_wallets.py
```diff
@@ -0,0 +1,2067 @@
...
+    cat_wallet_0 = dao_wallet_0.wallet_state_manager.wallets[dao_wallet_0.dao_info.cat_wallet_id]
+    await time_out_assert(10, cat_wallet_0.get_confirmed_balance, cat_amt)
+
+    # Create funding spends for xch and cat
+    xch_funds = uint64(500000)
+    cat_funds = uint64(100000)
+    funding_tx = await dao_wallet_0.create_add_money_to_treasury_spend(xch_funds)
+    funding_sb = funding_tx.spend_bundle
+    await time_out_assert_not_none(5, full_node_api.full_node.mempool_manager.get_spendbundle, funding_sb.name())
+    await full_node_api.process_transaction_records(records=[funding_tx])
```
trepca, line 365: why not process cat and xch funding in one block here?

tests/wallet/dao_wallet/test_dao_wallets.py
```diff
@@ -0,0 +1,2067 @@
...
+    tx_queue: List[TransactionRecord] = await wallet_node_0.wallet_state_manager.tx_store.get_not_sent()
+    tx_record = tx_queue[0]
+    await full_node_api.process_transaction_records(records=[tx_record])
+    await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hash_0))
+
+    # Farm enough blocks to pass the oracle_spend_delay and then complete the treasury eve spend
+    for i in range(1, 11):
+        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hash_0))
+
+    # async with wallet_node_0.wallet_state_manager.lock:
```
trepca, line 497: can we remove this?

tests/wallet/dao_wallet/test_dao_wallets.py
```diff
@@ -0,0 +1,2067 @@
...
+    )
+    update_inner = await dao_wallet_0.generate_update_proposal_innerpuz(new_dao_rules)
+    dao_cat_0_bal = await dao_cat_wallet_0.get_votable_balance()
+    proposal_tx = await dao_wallet_0.generate_new_proposal(update_inner, dao_cat_0_bal, uint64(1000))
+    proposal_sb = proposal_tx.spend_bundle
+    await time_out_assert_not_none(5, full_node_api.full_node.mempool_manager.get_spendbundle, proposal_sb.name())
+    await full_node_api.process_spend_bundles(bundles=[proposal_sb])
+
+    # Give the wallet nodes a second
+    await asyncio.sleep(1)
```
trepca, line 668: why a second exactly? isn't there any other way to detecting that nodes have been updated?

tests/wallet/dao_wallet/test_dao_wallets.py
```diff
@@ -0,0 +1,2067 @@
...
+    await dao_wallet_0.clear_finished_proposals_from_memory()
+    await time_out_assert(20, len, 4, dao_wallet_0.dao_info.proposals_list)  # we still have coins locked up
+
+    fs = await dao_wallet_0.free_coins_from_finished_proposals()
+    assert fs is not None
+
+    for i in range(1, num_blocks):
+        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hash_0))
+    # Wait a second for the dao cat wallet to update
+    await asyncio.sleep(2)
```
trepca, line 856: 2 seconds here?

tests/wallet/dao_wallet/test_dao_wallets.py
```diff
@@ -0,0 +1,2067 @@
...
+    tx_queue: List[TransactionRecord] = await wallet_node_0.wallet_state_manager.tx_store.get_not_sent()
+    tx_record = tx_queue[0]
+    await full_node_api.process_transaction_records(records=[tx_record])
+    await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hash_0))
+
+    # Farm enough blocks to pass the oracle_spend_delay and then complete the treasury eve spend
+    for i in range(1, 11):
+        await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(puzzle_hash_0))
+
+    # async with wallet_node_0.wallet_state_manager.lock:
```
trepca, line 945: can we remove this commented block?

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: yes, amount_of_cats is in mojos.

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: good point. The create_wallet method will raise a ValueError if wallet_info is none anyway, so this check is redundant.

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: I thought these were necessary for wallet integration, but seems like they aren't so have removed `puzzle_for_pk`, `puzzle_hash_for_pk`, and `get_new_puzzle`

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: Importing WalletNode in wallets creates a circular dependency via WSM. So has to be either untyped or Any

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+
+        # get existing xch funds for treasury
+        await self.wallet_state_manager.add_interested_puzzle_hashes([self.dao_info.treasury_id], [self.id()])
+        await self.wallet_state_manager.add_interested_puzzle_hashes(
+            [self.dao_info.current_treasury_coin.puzzle_hash], [self.id()]
+        )
+
+        # Resync the wallet from when the treasury was created to get the existing funds
+        # TODO: Maybe split this out as an option for users since it may be slow?
+        if not wallet_node.is_trusted(peer):
```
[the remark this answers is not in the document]
geoffwalmsley: Untrusted nodes won't automatically send us the history of all the treasury and proposal coins, so we have to request all the updates between the treasury being created and current peak.

tests/wallet/dao_wallet/test_dao_wallets.py
geoffwalmsley: We removed the functions for creating DAO's for existing CATs and creating DAOs 'offline' via spendbundle, since they were out of scope for the beta. These tests were relating to those functions, so I've just removed it for now. We've stored the unused code in a WIP branch so it can be picked up again later.

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2240 @@
...
+            if singleton_id:
+                await self.wallet_state_manager.singleton_store.add_spend(self.id(), parent_spend, height)
+            else:
+                # funding coin
+                asset_id = get_asset_id_from_puzzle(parent_spend.puzzle_reveal.to_program())
+                if asset_id not in self.dao_info.assets:
+                    new_asset_list = self.dao_info.assets.copy()
+                    new_asset_list.append(asset_id)
+                    dao_info = dataclasses.replace(self.dao_info, assets=new_asset_list)
+                    await self.save_info(dao_info)
```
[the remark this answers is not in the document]
geoffwalmsley: agreed that this could be slow - we will look to optimise the resync during the next iteration

chia/wallet/puzzles/dao_proposal_validator.clsp
```diff
@@ -74,11 +74,13 @@
...
-        (list CREATE_PUZZLE_ANNOUNCEMENT proposal_id)  ; specify the proposal we're talking about
-        conditions
-      )
+      (list CREATE_PUZZLE_ANNOUNCEMENT proposal_id)  ; specify the proposal we're talking about
+      (if (> (- coin_amount 1) 0)
+          (c
+            (list CREATE_COIN PROPOSAL_EXCESS_PAYOUT_PUZ_HASH (- coin_amount 1) (list (f (r SINGLETON_STRUCT))))
+            conditions
+          )
+      conditions)
```
matt-o-how, line 83: two thoughts:
- could we have the wallets ignore 0 coins? this would reduce on-chain code. not urgent but just a thought
- please can you fix the indentation here either way?

chia/wallet/puzzles/dao_spend_p2_singleton_v2.clsp
```diff
@@ -51,9 +51,12 @@
               (+ total (f (r (f p2_singleton_list))))
               output
             )
-            (c
-              (list CREATE_COIN P2_SINGLETON_PUZHASH (- total SPEND_AMOUNT) (list (f (r TREASURY_SINGLETON_STRUCT))))
-              output
+            (if (> (- total SPEND_AMOUNT) 0)
```
matt-o-how, line 54: same comment as before about wallet maybe ignoring this

chia/wallet/puzzles/dao_spend_p2_singleton_v2.clsp
```diff
@@ -119,9 +122,12 @@
             (r parent_amount_list)  ; this is the delegated_puzhash
             (sha256tree 0)  ; most coins destroy themselves
             (calculate_delegated_puzzlehash  ; the last coin creates the conditions
-              (c
-                (list CREATE_COIN P2_SINGLETON_PUZHASH (- (+ total (f (r (f parent_amount_list)))) create_coin_sum) (list (f (r TREASURY_SINGLETON_STRUCT))))
-                CONDITIONS_FOR_THIS_ASSET_TYPE
+              (if (> (- (+ total (f (r (f parent_amount_list)))) create_coin_sum) 0)
```
matt-o-how, line 125: Same comment as before with regards to 0 value change. Make it the wallet's problem saves on chain space.
But if you are going to keep this here then consider making `(- (+ total (f (r (f parent_amount_list)))) create_coin_sum)` a stored value instead of calculating it twice. Both for readability and for optimisation.

chia/wallet/puzzles/dao_spend_p2_singleton_v2.clsp
```diff
@@ -186,21 +192,31 @@
...
+            (for_each_asset_type
+              TREASURY_SINGLETON_STRUCT
+              CAT_MOD_HASH
+              P2_SINGLETON_VIA_DELEGATED_PUZZLE_PUZHASH
+              LIST_OF_TAILHASH_CONDITIONS
+              p2_singleton_tailhash_parent_amount_list  ; ((tailhash ((parent amount) (parent_amount)... ) (tailhash (parent amount))..)
+              CONDITIONS
+            )
+          )
+          (for_each_asset_type
```
matt-o-how, line 212: should we disallow spending coins if there are no `CONDITIONS`?

chia/wallet/puzzles/dao_proposal_validator.clsp
geoffwalmsley: - the main reasoning for doing it in the chialisp is to prevent proposals creating multiple zero coins which potentially create duplicate condition errors invalidating the whole proposal (particularly for cases where the proposal is spending multiple coins). The other trade off is that while it's more on chain code, it saves the CREATE_COIN cost. I'm not sure which way is more efficient?
- fixed

chia/wallet/puzzles/dao_spend_p2_singleton_v2.clsp
geoffwalmsley: fixed so it's only calculating once

chia/wallet/puzzles/dao_spend_p2_singleton_v2.clsp
geoffwalmsley: We need to capture the case where there are TAILHASH_CONDITIONS but not [xch] CONDITIONS. If neither are present then it will just return nil

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2268 @@
...
+
+    def get_parent_for_coin(self, coin: Coin) -> Optional[LineageProof]:
+        parent_info = None
+        for name, ccparent in self.dao_info.parent_info:
+            if name == coin.parent_coin_info:
+                parent_info = ccparent
+        return parent_info
+
+    async def get_frozen_amount(self) -> uint64:
+        return uint64(0)  # pragma: no cover
```
Quexington, line 332: Does this function need to exist?  It's not part of the wallet protocol and seems like it's unused/pointless

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2268 @@
...
+        """
+        self.log.info(f"DAOWallet.coin_added() called with the coin: {coin.name().hex()}:{coin}.")
+        wallet_node: Any = self.wallet_state_manager.wallet_node
+        peer = wallet_node.get_full_node_peer()
+        if peer is None:  # pragma: no cover
+            raise ValueError("Could not find any peers to request puzzle and solution from")
+        try:
+            # Get the parent coin spend
+            cs = (await wallet_node.get_coin_state([coin.parent_coin_info], peer, height))[0]
+            parent_spend = await fetch_coin_spend(cs.spent_height, cs.coin, peer)
```
Quexington, line 436: We have a utility function for this probably better to use it: https://github.com/Chia-Network/chia-blockchain/blob/d966f3f9e63aed52dbd73544164202a9f11ce3d2/chia/wallet/util/wallet_sync_utils.py#L428

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2268 @@
...
+        if not wallet_node.is_trusted(peer):
+            # Untrusted nodes won't automatically send us the history of all the treasury and proposal coins,
+            # so we have to request them via sync_from_untrusted_close_to_peak
+            request = RequestBlockHeader(children_state.created_height)
+            response: Optional[RespondBlockHeader] = await peer.call_api(FullNodeAPI.request_block_header, request)
+            await wallet_node.sync_from_untrusted_close_to_peak(response.header_block, peer)
+
+        return
+
+    async def create_tandem_xch_tx(
```
Quexington, line 615: there's a `Wallet.create_tandem_xch_tx` function now that you could probably use.

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2268 @@
...
+        next_coin = Coin(eve_coin.name(), eve_coin.puzzle_hash, eve_coin.amount)
+        await self.add_parent(eve_coin.name(), next_proof)
+        await self.wallet_state_manager.add_interested_coin_ids([next_coin.name()], [self.wallet_id])
+
+        dao_info = dataclasses.replace(self.dao_info, current_treasury_coin=next_coin)
+        await self.save_info(dao_info)
+        await self.wallet_state_manager.singleton_store.add_spend(self.id(), eve_coin_spend)
+        return eve_spend_bundle
+
+    def generate_simple_proposal_innerpuz(
```
Quexington, line 870: These `generate_` functions seem like maybe they belong in the dao_utils file

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2268 @@
...
+
+    async def generate_new_proposal(
+        self,
+        proposed_puzzle: Program,
+        tx_config: TXConfig,
+        vote_amount: Optional[uint64] = None,
+        fee: uint64 = uint64(0),
+        push: bool = True,
+        extra_conditions: Tuple[Condition, ...] = tuple(),
+    ) -> Union[SpendBundle, TransactionRecord]:
```
Quexington, line 936: Same comments as in the dao_cat_wallet file about transaction methods only returning `List[TransactionRecord]` and not pushing them automatically.

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2268 @@
...
+    async def generate_proposal_eve_spend(
+        self,
+        *,
+        eve_coin: Coin,
+        full_proposal_puzzle: Program,
+        dao_proposal_puzzle: Program,
+        proposed_puzzle_reveal: Program,
+        launcher_coin: Coin,
+        vote_amount: uint64,
+    ) -> SpendBundle:
```
Quexington, line 1054: Same comment about transaction methods as above.

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2268 @@
...
+    async def generate_proposal_vote_spend(
+        self,
+        proposal_id: bytes32,
+        vote_amount: Optional[uint64],
+        is_yes_vote: bool,
+        tx_config: TXConfig,
+        fee: uint64 = uint64(0),
+        push: bool = True,
+        extra_conditions: Tuple[Condition, ...] = tuple(),
+    ) -> SpendBundle:
```
Quexington, line 1118: Same comment about transaction methods as above

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2268 @@
...
+    async def create_proposal_close_spend(
+        self,
+        proposal_id: bytes32,
+        tx_config: TXConfig,
+        genesis_id: Optional[bytes32] = None,
+        fee: uint64 = uint64(0),
+        push: bool = True,
+        self_destruct: bool = False,
+        extra_conditions: Tuple[Condition, ...] = tuple(),
+    ) -> SpendBundle:
```
Quexington, line 1238: transaction method comment again

chia/wallet/dao_wallet/dao_wallet.py
```diff
@@ -0,0 +1,2268 @@
...
+                    [
+                        xch_parent_amount_list,
+                        tailhash_parent_amount_list,
+                        treasury_inner_puzhash,
+                    ]
+                )
+
+            elif proposal_type == ProposalType.UPDATE:
+                (
+                    _,
```
Quexington, line 1508: These underscores come out of nowhere and I can't find where they're defined.  It'd be good to give this constant value a name.

chia/wallet/puzzles/p2_conditions_curryable.clsp
```diff
@@ -0,0 +1,3 @@
+(mod (CONDITIONS_LIST)
```
Quexington, line 1: I think it's slightly dump optimization to just make this `(mod CONDITIONS_LIST` so the compiled output is just `01` but obviously that's a very standard puzzle so this file is certainly not waranted.  I'm not sure why this is necessary, I think the currying is just overhead here instead of just `(q . conditions)`

chia/wallet/singleton.py
```diff
@@ -14,7 +17,7 @@
 SINGLETON_LAUNCHER_PUZZLE_HASH = SINGLETON_LAUNCHER_PUZZLE.get_tree_hash()
 
 
-def get_inner_puzzle_from_singleton(puzzle: Program) -> Optional[Program]:
+def get_inner_puzzle_from_singleton(puzzle: Union[Program, SerializedProgram]) -> Optional[Program]:
```
Quexington, line 20: I think I would prefer `Program.to`ing the call sites instead but idc that much

pytest.ini
```diff
@@ -2,7 +2,7 @@
 ; logging options
 log_cli = False
 addopts = --verbose --tb=short -n auto -p no:monitor
-log_level = WARNING
+log_level = INFO
```
Quexington, line 5: Undo this probably

tools/manage_clvm.py
```diff
@@ -155,6 +155,9 @@ def from_hex_bytes(cls, hex_bytes: bytes) -> ClvmBytes:
             hash=generate_hash_bytes(hex_bytes=hex_bytes),
         )
 
+    def strip(self) -> ClvmBytes:
+        return ClvmBytes(hex=self.hex.strip(b"\n"), hash=self.hash.strip(b"\n"))
+
```
Quexington, line 160: What's the purpose of this change?  The neline is something we're trying very specifically to commit to so this change is "dangerous" to that idea I think.

chia/cmds/dao.py
```diff
@@ -0,0 +1,958 @@
...
+@dao_cmd.command("get_id", short_help="Get the Treasury ID of a DAO", no_args_is_help=True)
+@click.option(
+    "-wp",
+    "--wallet-rpc-port",
+    help="Set the port where the Wallet is hosting the RPC interface. See the rpc_port under wallet in config.yaml",
+    type=int,
+    default=None,
+)
+@click.option("-f", "--fingerprint", help="Set the fingerprint to specify which key to use", type=int)
+@click.option("-i", "--wallet-id", help="ID of the DAO wallet which will receive the funds", type=int, required=True)
```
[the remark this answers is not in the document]
geoffwalmsley: this was a copy/paste error. Fixed

chia/cmds/dao.py
```diff
@@ -0,0 +1,958 @@
...
+@click.option(
+    "-m",
+    "--fee",
+    help="Set the fees per transaction, in XCH.",
+    type=str,
+    default="0",
+    show_default=True,
+    callback=validate_fee,
+)
+@click.option(
```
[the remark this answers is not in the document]
geoffwalmsley: added

chia/cmds/dao.py
```diff
@@ -0,0 +1,958 @@
...
+dao_cmd.add_command(dao_show_proposal_cmd)
+dao_cmd.add_command(dao_vote_cmd)
+dao_cmd.add_command(dao_close_proposal_cmd)
+dao_cmd.add_command(dao_lockup_coins_cmd)
+dao_cmd.add_command(dao_exit_lockup_cmd)
+dao_cmd.add_command(dao_release_coins_cmd)
+dao_cmd.add_command(dao_proposal)
+
+
+# TODO: status: how many of your voting coins are locked away vs. spendable, etc.
```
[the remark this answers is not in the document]
geoffwalmsley: Correct - we don't need to do this, removed the comment

chia/cmds/dao_funcs.py
```diff
@@ -0,0 +1,466 @@
...
+
+    print(f"Adding wallet for DAO: {treasury_id}")
+    print("This may take awhile.")
+
+    async with get_wallet_client(wallet_rpc_port, fp) as (wallet_client, fingerprint, config):
+        res = await wallet_client.create_new_dao_wallet(
+            mode="existing",
+            tx_config=CMDTXConfigLoader(
+                reuse_puzhash=True,
+            ).to_tx_config(units["chia"], config, fingerprint),
```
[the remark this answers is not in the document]
geoffwalmsley: fixed

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: fixed

chia/wallet/puzzles/p2_conditions_curryable.clsp
geoffwalmsley: This file wasn't even being used for anything, so have deleted it

pytest.ini
geoffwalmsley: Done

chia/wallet/cat_wallet/cat_utils.py
```diff
@@ -74,6 +74,12 @@ def construct_cat_puzzle(
     return mod_code.curry(mod_code_hash, limitations_program_hash, inner_puzzle)
 
 
+def construct_cat_puzzlehash_for_inner_puzzlehash(
+    tail_hash: bytes32, inner_puzzle_hash: bytes32
+) -> bytes32:  # pragma: no cover
```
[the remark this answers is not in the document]
geoffwalmsley: removed

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+            self.lineage_store = await CATLineageStore.create(self.wallet_state_manager.db_wrapper, self.get_asset_id())
+        except AssertionError as e:  # pragma: no cover
+            self.log.error(f"Error creating DAO CAT wallet: {e}")
+            # Do a migration of the lineage proofs
+            # cat_info = LegacyCATInfo.from_bytes(hexstr_to_bytes(self.wallet_info.data))
+            # self.cat_info = DAOCATInfo(cat_info.limitations_program_hash, cat_info.my_tail)
+            # self.lineage_store = await CATLineageStore.create(self.wallet_state_manager.db_wrapper, self.get_asset_id())
+            # for coin_id, lineage in cat_info.lineage_proofs:
+            #     await self.add_lineage(coin_id, lineage)
+            # await self.save_info(self.cat_info)
```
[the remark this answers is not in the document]
geoffwalmsley: removed

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+
+if TYPE_CHECKING:
+    from chia.wallet.wallet_state_manager import WalletStateManager
+
+CAT_MOD_HASH = CAT_MOD.get_tree_hash()
+CAT_MOD_HASH_HASH = Program.to(CAT_MOD_HASH).get_tree_hash()
+QUOTED_MOD_HASH = calculate_hash_of_quoted_mod_hash(CAT_MOD_HASH)
+
+
+class DAOCATWallet:
```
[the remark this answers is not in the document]
geoffwalmsley: This was something we considered early on, but I think we had to override a lot of the methods anyway and it became more complex using a subclass than not. @matt-o-how could you fill in the gaps on why we elected not to take the subclass route?

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+
+    async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection) -> None:
+        """Notification from wallet state manager that wallet has been received."""
+        self.log.info(f"DAO CAT wallet has been notified that {coin} was added")
+        # We can't get the inner puzzle for this coin's puzhash because it has the lockup layer.
+        # So look for it's parent coin, and get the inner puzzle for it, which should be the same as
+        # the one contained in the lockup.
+        wallet_node: Any = self.wallet_state_manager.wallet_node
+        parent_coin = (await wallet_node.get_coin_state([coin.parent_coin_info], peer, height))[0]
+        parent_spend = await fetch_coin_spend(height, parent_coin.coin, peer)
```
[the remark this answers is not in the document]
geoffwalmsley: not sure what you meant here - are we not using that utility function already?

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+        else:
+            inner_puzzle = get_innerpuz_from_lockup_puzzle(cat_inner)
+            active_votes_list = get_active_votes_from_lockup_puzzle(cat_inner)
+            active_votes_list = [x.as_atom() for x in active_votes_list.as_iter()]
+
+        if parent_spend.coin.puzzle_hash == coin.puzzle_hash:
+            # shortcut, works for change
+            lockup_puz = cat_inner
+        else:
+            # TODO: Move this section to dao_utils once we've got the close spend sorted
```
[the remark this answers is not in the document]
geoffwalmsley: it is sorted. Comment removed

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+        fee: uint64 = uint64(0),
+        extra_conditions: Tuple[Condition, ...] = tuple(),
+    ) -> Tuple[List[TransactionRecord], Optional[List[Coin]]]:
+        # check there are enough cats to convert
+        cat_wallet = self.wallet_state_manager.wallets[self.dao_cat_info.free_cat_wallet_id]
+        cat_balance = await cat_wallet.get_spendable_balance()
+        if cat_balance < amount:  # pragma: no cover
+            raise ValueError(f"Insufficient CAT balance. Requested: {amount} Available: {cat_balance}")
+        # get the lockup puzzle hash
+        lockup_puzzle = await self.get_new_puzzle()
```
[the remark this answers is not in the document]
geoffwalmsley: fixed

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+        ).get_tree_hash()
+        if push:
+            for tx in txs:
+                await self.wallet_state_manager.add_pending_transaction(tx)
+                for coin in tx.additions:
+                    if coin.puzzle_hash == cat_puzzle_hash:
+                        new_cats.append(coin)
+        await self.wallet_state_manager.add_interested_puzzle_hashes([cat_puzzle_hash], [self.id()])
+
+        return txs, new_cats
```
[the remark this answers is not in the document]
geoffwalmsley: yeah, I agree that the mix of spend bundles and tx records is not ideal. Have made all the wallet functions return a tx record, and moved add_pending_tx to the rpc (except for dao creation which needs to push the tx itself)

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+            solution = Program.to(
+                [
+                    0,
+                    inner_solution,
+                    coin.amount,
+                    0,
+                    0,
+                    0,
+                    0,
+                    0,
```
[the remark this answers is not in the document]
geoffwalmsley: done

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+                fee,
+                tx_config,
+                extra_conditions=extra_conditions,
+            )
+            assert chia_tx.spend_bundle is not None
+            full_spend = SpendBundle.aggregate([spend_bundle, chia_tx.spend_bundle])
+        else:
+            full_spend = spend_bundle
+
+        if push:
```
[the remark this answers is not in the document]
geoffwalmsley: done (see previous comment)

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+        return txs, new_cats
+
+    async def exit_vote_state(
+        self,
+        coins: List[LockedCoinInfo],
+        tx_config: TXConfig,
+        fee: uint64 = uint64(0),
+        push: bool = True,
+        extra_conditions: Tuple[Condition, ...] = tuple(),
+    ) -> SpendBundle:
```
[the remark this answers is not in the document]
geoffwalmsley: done

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+                self.dao_cat_info.my_tail,
+                new_locked_coins,
+            )
+            await self.save_info(dao_cat_info)
+
+        return spend_bundle
+
+    async def remove_active_proposal(
+        self, proposal_id_list: List[bytes32], fee: uint64 = uint64(0), push: bool = True
+    ) -> SpendBundle:
```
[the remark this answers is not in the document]
geoffwalmsley: done

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+                    balance += coin.coin.amount
+            else:
+                balance += coin.coin.amount
+        if include_free_cats:
+            cat_wallet = self.wallet_state_manager.wallets[self.dao_cat_info.free_cat_wallet_id]
+            cat_balance = await cat_wallet.get_spendable_balance()
+            balance += cat_balance
+        return uint64(balance)
+
+    async def sign(self, spend_bundle: SpendBundle) -> SpendBundle:
```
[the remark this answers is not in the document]
geoffwalmsley: done

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: removed

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: same as previous comment - it looks to me like we're already using the `fetch_coin_spend` utility?

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: replaced the old tandem_tx with this one

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: moved to utils

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: done

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: done

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: done

chia/wallet/dao_wallet/dao_wallet.py
geoffwalmsley: done

tools/manage_clvm.py
geoffwalmsley: I don't know where this came from - have reverted the file to match main

chia/wallet/singleton.py
geoffwalmsley: left as-is for now

chia/wallet/cat_wallet/dao_cat_wallet.py
```diff
@@ -0,0 +1,753 @@
...
+        self.lineage_store = await CATLineageStore.create(self.wallet_state_manager.db_wrapper, self.get_asset_id())
+        await self.wallet_state_manager.add_new_wallet(self)
+        return self
+
+    async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection) -> None:
+        """Notification from wallet state manager that wallet has been received."""
+        self.log.info(f"DAO CAT wallet has been notified that {coin} was added")
+        # We can't get the inner puzzle for this coin's puzhash because it has the lockup layer.
+        # So look for it's parent coin, and get the inner puzzle for it, which should be the same as
+        # the one contained in the lockup.
```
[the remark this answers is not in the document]
geoffwalmsley: The approach here is to reconstruct the inner puzzle by taking the list of active votes in the parent's puzzle, and adding the votes found in the solution. We then check that inner puzzle's hash against the incoming coin's puzhash to make sure we've created the puzzle correctly.

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -783,6 +783,7 @@ async def generate_signed_transaction(
     ) -> List[TransactionRecord]:
         # (extra_delta, tail_reveal, tail_solution)
         cat_discrepancy: Optional[Tuple[int, Program, Program]] = kwargs.get("cat_discrepancy", None)
+        override_memos: Optional[bool] = kwargs.get("override_memos", None)
```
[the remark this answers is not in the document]
Quexington: I just want to make sure I have this straight in my head:
* You use this override_memos argument when you are sending a CAT to a DAO
* The format of this CAT is CAT(inner_puzzle)
* The DAO is recieving this CAT at CAT(p2_singleton_of_dao)

Is that right?  Because it seems to me that the DAO wallet could subscribe to its p2_singleton address and then that would be what the CAT normally hints to as well right?

chia/wallet/cat_wallet/dao_cat_wallet.py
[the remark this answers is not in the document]
Quexington: Okay that makes sense to me, but this comment seems to tell me you're just assuming the parent's inner puzzle is the child's inner puzzle as well which is what I was nervous about so maybe edit/delete this comment.

chia/wallet/cat_wallet/dao_cat_wallet.py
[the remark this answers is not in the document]
Quexington: Hmmmmmmmmmmmmmmmmmmmmm I remember otherwise but github would seem to be gaslighting me into believing this is how it always was and I guess I'm going to let.

chia/wallet/dao_wallet/dao_wallet.py
Quexington: I don't know what you did but this is NOT how it was >:(

chia/wallet/cat_wallet/cat_wallet.py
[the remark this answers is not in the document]
geoffwalmsley: Yes that's right...the trouble is that p2_singleton_of_dao is unique for each CAT's asset id so everyone's DAO wallet would have to subscribe to all the possible p2_singleton addresses for every CAT in existence.

If you're the one sending the funds, you can subscribe to the p2_singleton puzhash, but then nobody else's DAO wallet would know to do that. The only way we have (currently) to make sure everyone can see incoming funds is to use the dao ID as the hint. The only alternative I can see would be for people to manually create a list of CAT's and have their wallet subscribe to just those assets.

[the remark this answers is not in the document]
Quexington: I'm not quite sure I'm following.  p2_singleton is parameterized by only the DAO's singleton ID.  Sure each cat will have a different FULL puzzle but the inner puzzle will be the same across all.  It's actually less permutations that our existing system because the standard puzzle can have derivations but p2_singleton is always p2_singleton.

chia/wallet/cat_wallet/dao_cat_wallet.py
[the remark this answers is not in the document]
geoffwalmsley: comment deleted

chia/wallet/cat_wallet/cat_wallet.py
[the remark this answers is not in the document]
geoffwalmsley: yeah this makes sense. I think in an earlier the version the asset ID was being used somewhere that made this not possible, but you're absolutely right and your suggested way works better. Have removed the cat wallet override, and adapted the dao chialisp, drivers and wallet to hint on the p2_singleton. Thanks for your persistence :)

chia/cmds/dao.py
```diff
@@ -0,0 +1,958 @@
...
+@click.option("-n", "--name", help="Set the DAO wallet name", type=str)
+@click.option(
+    "-t",
+    "--treasury-id",
+    help="The Treasury ID of the DAO you want to track",
+    type=str,
+    required=True,
+)
+@click.option(
+    "-fa",
```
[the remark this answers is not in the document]
geoffwalmsley: I've removed the multi-letter options. For DAO creation and rule updates I've left only the long-form options available since having a bunch of random letters is more confusing. And since these commands aren't going to be used terribly frequently I feel like this isn't a bad trade off

chia/wallet/cat_wallet/dao_cat_wallet.py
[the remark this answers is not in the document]
Quexington: I'll resolve this for now, it's something we could look into as an upgrade in the future for better type safety.

Quexington (dismissed):

Quexington (dismissed):

wjblanke (approved):
aok

PR #15561, opened 2023-06-20
mypy: Fix issues introduced by 1.4.0 release

xdustinface:
### Purpose:

`mypy` version `1.4.0` was obviously just released and leads to `main` failure.

PR #15562, opened 2023-06-20
mypy: Fix issues introduced by 1.4.0 release

xdustinface:
### Purpose:

Same like #15561 but for `main` instead of `release/1.8.2` (which maybe doesn't get merged anyway) but still i opened this to speed up getting `main` green.


wjblanke (approved):
aok

PR #15563, opened 2023-06-20
setup: Pin `mypy==1.3.0`

xdustinface:
### Purpose:

Pin `mypy` to avoid running into the issues fixed in #15561.


wjblanke (approved):
aok

PR #15567, opened 2023-06-20
fix python matrix for release

wallentx:
### Purpose:
This fixes the test matrix not running on all python versions for PRs targeting `release/**` branches

### Current Behavior:
PRs targeting `release/**` branches currently run the "limited" matrix

### New Behavior:
PRs targeting `release/**` branches will run the "all" matrix


altendky (approved):
Looks ok.  I think...  GitHub Actions expressions isn't a great language for this.  Maybe at some point (unrelated to this PR) we should consider if these conditions would fit well in the Python.

Thanks.

I would suggest that this could be merged through the failures and they can be addressed separately now that they are exposed.

emlowe (approved):

PR #15580, opened 2023-06-21
2.0 updates

arvidn:
add python constants for new error codes and condition codes, introduced by the new version of `chia_rs`

### DIFF
--- a/chia/util/errors.py
+++ b/chia/util/errors.py
@@ -175,6 +175,15 @@ class Err(Enum):
 
     ASSERT_EPHEMERAL_FAILED = 140
     EPHEMERAL_RELATIVE_CONDITION = 141
+    # raised if a SOFTFORK condition invokes an unknown extension in mempool
+    # mode
+    INVALID_SOFTFORK_CONDITION = 142
+    # raised if the first argument to the SOFTFORK condition is not a valid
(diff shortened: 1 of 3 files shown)


xdustinface (dismissed):
Couldn't we just use `INVALID_CONDITION` and `INVALID_COST_RESULT` instead of introducing this specifically for `SOFTFORK`?

arvidn:
it would be possibly of course, but `INVALID_CONDITION` is already used in a lot of places for lower-level errors (like, invalid formatting of a condition). It's quite helpful to be able to identify which condition there's an issue with. `INVALID_COST_RESULT` does not appear to be used anywhere, so perhaps it could be removed. The new cost-related error is if the cost argument to the `SOFTFORK` condition is invalid, either too small or too large. It's not really a result.

chia/util/errors.py
```diff
@@ -175,6 +175,9 @@ class Err(Enum):
 
     ASSERT_EPHEMERAL_FAILED = 140
     EPHEMERAL_RELATIVE_CONDITION = 141
+    INVALID_SOFTFORK_CONDITION = 142
```
trepca, line 178: can we add a comment when this error is usually raised?

trepca (approved):

xdustinface (approved):

PR #15588, opened 2023-06-22
don't parameterize the blockchain tests on DB version

arvidn:
We're about to need a lot more CI time. The first commit saves CI time by not running the blockchain tests with both v1 and v2 database schemas, but just v2. These are primarily system tests anyway, and the underlying `CoinStore` and `BlockStore` classes are still unit tested using both v1 and v2 schemas.

Another aspect of this patch is to use the `db_version` fixture rather than (unnamed) parameters directly.

The second commit makes tests of `CoinStore`, `BlockStore` and `FullNodeStore` only run in `PLAIN` consensus mode. This is currently the only mode we have, but we're about to add two new modes (for the hard fork and soft fork). This avoids running these unit tests 3 times instead of 1.

This PR brings the blockchain CI times from 24 minutes to 15 minutes.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -131,14 +131,14 @@ def self_hostname():
 #       the fixtures below. Just be aware of the filesystem modification during bt fixture creation
 
 
-@pytest_asyncio.fixture(scope="function", params=[1, 2])
-async def empty_blockchain(request, blockchain_constants):
+@pytest_asyncio.fixture(scope="function")
+async def empty_blockchain(latest_db_version, blockchain_constants):
     """
     Provides a list of 10 valid blocks, as well as a blockchain with 9 blocks added to it.
     """
     from tests.util.blockchain import create_blockchain
 
-    bc1, db_wrapper, db_path = await create_blockchain(blockchain_constants, request.param)
+    bc1, db_wrapper, db_path = await create_blockchain(blockchain_constants, latest_db_version)
     yield bc1
 
     await db_wrapper.close()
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -131,14 +131,14 @@ def self_hostname():
 #       the fixtures below. Just be aware of the filesystem modification during bt fixture creation
 
 
-@pytest_asyncio.fixture(scope="function", params=[1, 2])
-async def empty_blockchain(request, blockchain_constants):
+@pytest_asyncio.fixture(scope="function")
+async def empty_blockchain(latest_db_version, blockchain_constants):
```
arvidn, line 135: this is the main change. the `empty_blockchain` fixture is only run with the latest db schema

arvidn:
The coverage report says:

```
tests/blockchain/test_blockchain.py (87.5%): Missing lines 147
tests/conftest.py (100%)
tests/core/full_node/stores/test_block_store.py (67.9%): Missing lines 35,94,124,171,194,215,263,305,345
tests/core/full_node/stores/test_coin_store.py (68.2%): Missing lines 60,170,216,251,339,395,427
tests/core/full_node/stores/test_full_node_store.py (81.8%): Missing lines 77,801
tests/core/mempool/test_mempool_performance.py (25.0%): Missing lines 43-44,69
```

The cases that will cause tests to be skipped are never exercised in this PR. They are in 3 PRs that are in the pipeline right now. (by me, florin and amine). Having these conditions in `main` helps to keep our PRs more independent.
test_mempool_performance is a benchmark, and we don't measure coverage for those.

I think we should disregard these missed lines.

fchirica (dismissed):

fchirica (approved):

emlowe:
Approved with coverage diff exception per Arvid's explanation

emlowe (approved):

PR #15590, opened 2023-06-22
Move harvester_farmer_environment to conftest.py

AmineKhaldi:
This allows the fixture to be available to more than one test file.


arvidn:
is this moving the fixture verbatim?

AmineKhaldi:
> is this moving the fixture verbatim?

Yes.

arvidn (approved):

PR #15593, opened 2023-06-22
deduplicate coin states

ytx1991:
### Purpose:
In some situations, one coin can be processed multiple times. To prevent that, we need to add deduplication in add_coin_states.

### Current Behavior:
No deduplication

### New Behavior:
Deduplicate  coin states

### DIFF
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -87,6 +87,7 @@ from chia.wallet.util.wallet_sync_utils import (
     PeerRequestException,
     fetch_coin_spend_for_coin_state,
     last_change_height_cs,
+    sort_coin_states,
 )
 from chia.wallet.util.wallet_types import CoinType, WalletIdentifier, WalletType
 from chia.wallet.vc_wallet.vc_drivers import VerifiedCredential
@@ -1602,7 +1603,8 @@ class WalletStateManager:
         fork_height: Optional[uint32],
     ) -> bool:
         try:
-            await self._add_coin_states(list(set(coin_states)), peer, fork_height)
+            unique_coin_states = sort_coin_states(list(set(coin_states)))
+            await self._add_coin_states(unique_coin_states, peer, fork_height)
         except Exception as e:
             log_level = logging.DEBUG if peer.closed else logging.ERROR
             self.log.log(log_level, f"add_coin_states failed - exception {e}, traceback: {traceback.format_exc()}")
--- a/chia/wallet/wallet_state_manager.py
+++ b/chia/wallet/wallet_state_manager.py
@@ -1602,7 +1602,7 @@ class WalletStateManager:
         fork_height: Optional[uint32],
     ) -> bool:
         try:
-            await self._add_coin_states(coin_states, peer, fork_height)
+            await self._add_coin_states(list(set(coin_states)), peer, fork_height)
         except Exception as e:
             log_level = logging.DEBUG if peer.closed else logging.ERROR
             self.log.log(log_level, f"add_coin_states failed - exception {e}, traceback: {traceback.format_exc()}")
--- a/tests/wallet/clawback/test_clawback_lifecycle.py
+++ b/tests/wallet/clawback/test_clawback_lifecycle.py
@@ -195,9 +195,10 @@ class TestClawbackLifecycle:
                 cost_logger=cost_logger,
                 cost_log_msg="Create Second Clawback",
             )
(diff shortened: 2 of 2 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -1602,7 +1603,8 @@ async def add_coin_states(
         fork_height: Optional[uint32],
     ) -> bool:
         try:
-            await self._add_coin_states(list(set(coin_states)), peer, fork_height)
+            unique_coin_states = sort_coin_states(list(set(coin_states)))
```
arvidn, line 1606: both before and after does this:
`list(set(coin_states))`
which is deduplicating coins. Can you explain how this change makes a difference in deduplication?

arvidn (commented):
it looks like you just order the coin states, they were already deduplicated

chia/wallet/wallet_state_manager.py
ytx1991: I can't find the deduplicating code you are referring to, can you add a link?

arvidn: putting items in a `set` deduplicates them. We do that on the line I put this comment on.

tests/wallet/clawback/test_clawback_lifecycle.py
```diff
@@ -195,9 +195,10 @@ async def test_clawback_spends(self, cost_logger: CostLogger) -> None:
                 cost_logger=cost_logger,
                 cost_log_msg="Create Second Clawback",
             )
-
-            new_cb_coin = (await sim_client.get_coin_records_by_puzzle_hash(cb_puz_hash))[1].coin
-
+            cb_coins = await sim_client.get_coin_records_by_puzzle_hash(cb_puz_hash)
+            new_cb_coin = cb_coins[0].coin
+            if cb_coins[0].spent:
+                new_cb_coin = cb_coins[1].coin
```
arvidn, line 201: please add a comment to what this logic intends to do, and maybe why it's safe to assume that the second coin in this list is the one you want

chia/wallet/wallet_state_manager.py
ytx1991: Oh I see, you need to check the diff with the release/1.8.2. It doesn't have List(set()) thing there.

xdustinface: > putting items in a `set` deduplicates them. We do that on the line I put this comment on.

You only look at the second commit, i guess?

chia/wallet/wallet_state_manager.py
```diff
@@ -1602,7 +1603,8 @@ async def add_coin_states(
         fork_height: Optional[uint32],
     ) -> bool:
         try:
-            await self._add_coin_states(coin_states, peer, fork_height)
+            unique_coin_states = sort_coin_states(list(set(coin_states)))
+            await self._add_coin_states(unique_coin_states, peer, fork_height)
```
xdustinface, line 1607: This isn't the right place to deduplicate the coin states imo because this is after we already sorted all coin states here https://github.com/Chia-Network/chia-blockchain/blob/b8a1515d7350f92e246a369b2685fafdc7864d5c/chia/wallet/wallet_node.py#L870 before we split them up into batches to process in tasks which call `add_coin_states` here https://github.com/Chia-Network/chia-blockchain/blob/b8a1515d7350f92e246a369b2685fafdc7864d5c/chia/wallet/wallet_node.py#L905 

So basically you now only deduplicate the separate batches instead the whole list coming from a peer and we would still validate duplicates and also this adds a second round of sorting.

xdustinface (changes requested):

chia/wallet/wallet_state_manager.py
xdustinface: I just created https://github.com/Chia-Network/chia-blockchain/pull/15598.

tests/wallet/clawback/test_clawback_lifecycle.py
xdustinface: I also wonder, whats the purpose of this change?

chia/wallet/wallet_state_manager.py
arvidn: yes. I clicked on the link in the email and apparently only saw a single commit

ytx1991:
Close this PR since it is solved in https://github.com/Chia-Network/chia-blockchain/pull/15598

PR #15594, opened 2023-06-22
Use PKCS#8 format for SSL private keys

Rigidity:
### Purpose:

PKCS#8 is a more generally supported and recommended private key format.

### Current Behavior:

We use PKCS#1 for the CA private key as well as all private keys for certificates signed with it.

### New Behavior:

Switch to PKCS#8 format for our SSL keys by default.

### Testing Notes:

Tested key generation and running the full node with the new key format. Everything loaded the same way.


hoffmang9 (approved):
lgtm

arvidn (approved):
are these all the places where we write out SSL keys?

Rigidity:
> are these all the places where we write out SSL keys?

I searched for `TraditionalOpenSSL` and the method used to write out the PEM files, and only found those two lines. I also made sure the wallet private key file was in the new format. I think that's the only place, but definitely need a second pair of eyes to double check and test this, just in case.

PR #15598, opened 2023-06-23
wallet: Deduplicate coin states from peers

xdustinface:
### Purpose:

The full node might send duplicated coin states which are currently getting processed if they are passed into `add_coin_states` in the same batch because we only avoid re-processing duplicated states which are already in the DB the moment we step into `add_coin_states` here https://github.com/Chia-Network/chia-blockchain/blob/b8a1515d7350f92e246a369b2685fafdc7864d5c/chia/wallet/wallet_state_manager.py#L1264-L1273 which isn't the case if there are two of the same in the input to that method because then the duplicated state will not be in `local_records` after first processing. We could add a set there to track what we processed in this round but deduplicating at the higher level here probably makes more sense. 

Alternative to #15593.

### Current Behavior:

Duplicated sates from the full node will be processed if they come in the same batch into `add_coin_states`.

### New Behavior:

Duplicated states from the full node are removed.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -865,9 +865,9 @@ class WalletNode:
         target_concurrent_tasks: int = 30
 
         # Ensure the list is sorted
-
-        before = len(items_input)
-        items = await self.wallet_state_manager.filter_spam(sort_coin_states(items_input))
+        unique_items = set(items_input)
+        before = len(unique_items)
+        items = await self.wallet_state_manager.filter_spam(sort_coin_states(unique_items))
         num_filtered = before - len(items)
         if num_filtered > 0:
             self.log.info(f"Filtered {num_filtered} spam transactions")
--- a/tests/wallet/test_wallet_utils.py
+++ b/tests/wallet/test_wallet_utils.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
-from random import shuffle
-from typing import List, Optional, Tuple
+from typing import Collection, List, Optional, Tuple
 
 from chia_rs import Coin, CoinState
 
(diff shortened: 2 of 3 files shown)


xdustinface:
I targeted `release/1.8.2` because #15593 does, but i think we don't really need this in the release because the duplication only happened due to coins mistakenly being hinted with their target puzzle hash which will be fixed before the release? So im retargeting to `main`.

arvidn:
it would also make sense for the full node to not sending duplicates, right?

chia/wallet/wallet_node.py
```diff
@@ -865,9 +865,9 @@ async def add_states_from_peer(
         target_concurrent_tasks: int = 30
 
         # Ensure the list is sorted
-
-        before = len(items_input)
-        items = await self.wallet_state_manager.filter_spam(sort_coin_states(items_input))
+        unique_items = set(items_input)
+        before = len(unique_items)
+        items = await self.wallet_state_manager.filter_spam(sort_coin_states(unique_items))
```
arvidn, line 870: this is an unrelated comment, but it seems like we should probably filter *before* we sort. There's no point in sorting items that we'll drop later.

tests/wallet/test_wallet_utils.py
```diff
@@ -12,19 +11,17 @@ def dummy_coin_state(*, created_height: Optional[int], spent_height: Optional[in
     return CoinState(Coin(bytes(b"0" * 32), bytes(b"0" * 32), 0), spent_height, created_height)
 
 
-def heights(coin_states: List[CoinState]) -> List[Tuple[Optional[int], Optional[int]]]:
+def heights(coin_states: Collection[CoinState]) -> List[Tuple[Optional[int], Optional[int]]]:
```
arvidn, line 14: does `Collection` imply that it's ordered? I think this function requires the input to be ordered, and not a `set` for example.

chia/wallet/wallet_node.py
```diff
@@ -865,9 +865,9 @@ async def add_states_from_peer(
         target_concurrent_tasks: int = 30
 
         # Ensure the list is sorted
-
-        before = len(items_input)
-        items = await self.wallet_state_manager.filter_spam(sort_coin_states(items_input))
+        unique_items = set(items_input)
+        before = len(unique_items)
```
arvidn, line 869: would it make sense to have an `INFO` log here if `len(unique_items) < len(items_input)`?

arvidn (approved):

xdustinface:
> it would also make sense for the full node to not sending duplicates, right?

Thats already merged here #15560.

tests/wallet/test_wallet_utils.py
xdustinface: I changed it to `Collection` to allow a `Set` which makes testing this line https://github.com/Chia-Network/chia-blockchain/pull/15598/files#diff-0d32919512429d54cb0a91510a71245a1e1b32f461e4358fe1e24a872e41e12aR36 possible. Could have probably also just converted it back to a list before passing into `heights` instead.

chia/wallet/wallet_node.py
xdustinface: Hm, maybe, but it would also require a test to get a green CI or a coverage exception :)

PR #15601, opened 2023-06-23
Add some tests for Farmer's respond_signatures

AmineKhaldi:


arvidn:
it's a bit hard to tell what's being tested exactly. would you mind adding some comments describing that?
Also, I would expect to have a case where you don't get an error also. or is that not interesting to test?

AmineKhaldi:
> it's a bit hard to tell what's being tested exactly. would you mind adding some comments describing that? Also, I would expect to have a case where you don't get an error also. or is that not interesting to test?

Done.

arvidn (approved):

PR #15612, opened 2023-06-23
clvm test fixes

geoffwalmsley:
- Updated clvm tests for the latest chialisp 
- Removed test_spendbundles as no longer needed


chia/wallet/puzzles/p2_singleton_aggregator.clsp
arvidn, line 1: since this is new chialisp, shouldn't it be using the new compiler?

matt-o-how (approved):
in the future we might want to add in the ability to support multiple aggregator programs but for now, only supporting one gives us some security and stability

PR #15613, opened 2023-06-23
pass DataLayer root and generation instead of requerying

emlowe:
Explorations in reducing SQL select calls during insert_batch operations

Results in a reasonable speedup of about 2-3x.

Added a test that serves as a possible benchmark test - needs some further work to be proper.

Extracted:
- https://github.com/Chia-Network/chia-blockchain/pull/15645
- https://github.com/Chia-Network/chia-blockchain/pull/15690
- https://github.com/Chia-Network/chia-blockchain/pull/15691
- https://github.com/Chia-Network/chia-blockchain/pull/15728

Draft for:
 - [x] consider helpers for the repeated 'if not root get it' pattern
 - [x] consider removal of `DataStore.table_is_empty()`
 - [x] should `DataStore.insert()` take a root or a generation and hash?
 - [x] avoid returning tuples
 - [x] should `DataStore.delete()` have an `Optional` return?  or should the no-key-to-delete case return `None`?
 - [x] move `DataStore.insert_batch()` optimization to `DataStore.get_keys_values_dict()`
 - [x] `iter_root` -> `intermediate_root`


chia/data_layer/data_store.py
```diff
@@ -809,7 +834,7 @@ async def insert(
                 if side is not None:
                     raise Exception("Tree was empty so side must be unspecified, got: {side!r}")
 
-                await self._insert_root(
+                new_root = await self._insert_root(
```
almogdepaz, line 837: looks like we are never using this value i would remove

altendky: Isn't this used for the `InsertResult` at the end in the case of inserting a first leaf into an empty tree?

paninaro (approved):

PR #15614, opened 2023-06-23
stop using pip cache

altendky:

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -49,20 +49,6 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Get pip cache dir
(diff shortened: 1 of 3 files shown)


hoffmang9:
Purely only curiosity - what is the motivation to drop the caches?

hoffmang9:
Makes sense - will be curious about the performance hit - if any.

altendky:
Minimal checking looks like 15'ish seconds slower.  That's looking at 10 runs or so on macOS from this vs. the latest `main` build.

altendky:
handled slightly better now over in https://github.com/Chia-Network/actions/blob/1d39c9f963acf2cb1b41ea527149ed9bd00b0140/cache-pip/action.yml, so at least further corrections would be applied there and not here in blockchain

PR #15621, opened 2023-06-26
Apply Kevin's suggestions to simplify some test checks and fix a couple typos

AmineKhaldi:


tests/farmer_harvester/test_filter_prefix_bits.py
```diff
@@ -69,15 +69,10 @@ def state_has_changed() -> bool:
...
     # We're intercepting the harvester's state changes as we're expecting
-    # a farming_info one. eligible_plots are what passed the plot filter
-    if (
-        state_change == "farming_info"
-        and state_change_data is not None
-        and state_change_data.get("eligible_plots") == 1
-    ):
-        passed = True
-    assert passed == should_pass
+    # a farming_info one.
```
xdustinface, line 75:
```suggestion
    # a farming_info.
```

I think the `one` doesn't really make sense there either?

AmineKhaldi: "We're intercepting the harvester's state changes as we're expecting a farming_info one." one here is referring to one state change from many that can theoretically get intercepted from the harvester (and that one state change is of type farming_info). Does it make sense?

xdustinface: Seems rather confusing then just saying "we expect a farming_info" but with the explanation it makes more sense now :)

arvidn (approved):

wallentx:
Just to verify- you guys aren't waiting on me to merge this, right?

arvidn:
correct

Issue #11757, opened 2022-06-02
[Bug] Processing the same signage point multiple times


Jacek-ghub:
### What happened?

From time to time my logs show that the node processed the same signage point, e.g.,:
```
2022-06-01T23:17:09.473 full_node chia.full_node.full_node: INFO       Finished signage point 39/64: CC: 609df8fce29565da1ccbb7257669d09a49b5e0a8bd76532126677fb3e21fc726 RC: 384b2bdf32ca312ddcf773d754206a5644ff4df68891333be039af73dfd6e94d
2022-06-01T23:17:09.489 full_node chia.full_node.full_node: INFO       Finished signage point 39/64: CC: 609df8fce29565da1ccbb7257669d09a49b5e0a8bd76532126677fb3e21fc726 RC: 384b2bdf32ca312ddcf773d754206a5644ff4df68891333be039af73dfd6e94d
```

Should those multiple processing of the same thing be suppressed by the node?

Are those duplicate signage points due to some peers sending the same info, as such rather node specific (as such, the node not filtering those dups).

Or rather the network decided that there was some change in a given signage point, and every node on the network needs to reprocess it?  If this is the case, can the protocol be a bit extended to add a version to the signage point (so, the second time it needs to be processed, it will be ...:v2 mark on that signage point).

The problem with those dups comes mostly from pools marking those as potential duplicate plots on the farm, thus triggering people to constantly search for duplicate plots.

### Version
1.3.5

### What platform are you using?
Windows

### What ui mode are you using?
GUI

william-gr:
Would appreciate some insight here. On the pool side as he mentions it can be quite annoying.
We do not want to ignore double signage points partials because there is a chance they are real, like duplicated plots.

Jacek-ghub:
If we assume that (some of) those duplicates represent duplicate plots, isn't this the internal problem for the farmer, not really an error that should be propagated to the network (to the pool)?  As such, this should be addressed as an additional bug (unrelated to duplicates coming from the network).

Having the node at least flag those dups with ERROR level, and identifying which was the offending plot would be a great improvement already.

PR #15624, opened 2023-06-26
process new signage points no more than once

arvidn:
there are two places we call `signage_point_post_processing()` from.

1. when we receive a new signage point from a peer, and we don't already know about it. We add it to our full node store and call the post processing function. We do all of this while holding the `timelord_lock`. This ensures we don't receive the same SP multiple times and end up adding them multiple times. The checking if we have it, adding it and post processing all happens atomically, by holding the `timelord_lock`.
2. when receiving a new peak, we iterate over the signage points of the block, and we call the post processing function for them. We do not currently hold the lock or check to see if we have already seen the signage point.

This patch acquires the `timelord_lock` while iterating over the signage points in a new peak, to ensure the iteration does not interleave with receiving any new signage points. It also checks to see if we already have the signage point, and if so, skips calling the post processing function for it.

This is believed to fix an issue where we sometimes would send the same signage point multiple times to the farmer.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -1639,16 +1639,23 @@ class FullNode:
         )
 
         if fns_peak_result.new_signage_points is not None and peer is not None:
-            for index, sp in fns_peak_result.new_signage_points:
-                assert (
-                    sp.cc_vdf is not None
-                    and sp.cc_proof is not None
-                    and sp.rc_vdf is not None
-                    and sp.rc_proof is not None
-                )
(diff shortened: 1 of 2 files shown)


arvidn:
I will not have time to develop a test to exercise this. we may have to punt or resort to manual testing

altendky (changes requested):
just putting a hold on this while i run it for a bit and explore

arvidn:
this change might fix https://github.com/Chia-Network/chia-blockchain/issues/11757

Issue #14810, opened 2023-03-13
Full Node RPC: get_mempool_item_by_coin_name endpoint


Yakuhito:
### Discussed in https://github.com/Chia-Network/chia-blockchain/discussions/14809

<div type='discussions-op-text'>

<sup>Originally posted by **Yakuhito** March 13, 2023</sup>
Hello,

The way I see it, a fully 'decentralized' spend aggregation feature (i.e., no central sequencer) would require searching the mempool for spends & building your transaction on top of them. I've tried doing something similar and was surprised to find out that the full node RPC only allows you to get mempool items either by the transaction id (which is the sha256 of the spend bundle) or all at once. I'm writing this message to request a new endpoint that would take a coin name (id) as an argument and return the mempool item that spends that coin, if it exists.

The current workaround is getting all items at once and filtering them, which is very inefficient. I've seen that, for the current ~1700 items on testnet10, my python script took more than 30 seconds to decode the response (that's useless, given that the block time is lower than 60s).

This is not urgent, but I'm sure it would be a great addition to the Full Node RPC API.

Thank you,
yakuhito </div>

Yakuhito:
Update: I don't know what my computer was doing last night, but the JSON response is now decoding in a few seconds. However, this endpoint would still be nice to have.

Yakuhito:
Hey there, GitHub Actions Bot! Thanks for the heads up about my issue being flagged as stale. I know it's been a while since I've updated it, but trust me, it's like a fine wine that only gets better with age. The issue already contains all the relevant details, so there's no need for me to add more grapes to the mix. Let's keep this issue open and fermenting like a good bottle of cabernet sauvignon! 🍷

kimsk:
I have a [PR](https://github.com/Chia-Network/chia-blockchain/pull/15625) ready.

PR #15625, opened 2023-06-26
Add a full node RPC endpoint, get_mempool_item_by_coin_name

kimsk:
### Purpose:
Implement a new endpoint to get mempool item by a coin name mentioned in https://github.com/Chia-Network/chia-blockchain/issues/14810.

#### pytest
tests/core/test_full_node_rpc/test_coin_name_not_found_in_mempool
tests/core/test_full_node_rpc/test_coin_name_found_in_mempool

##### found
image

##### not found
image


Yakuhito:
Wow - thank you!

emlowe:
Error: The following new issues have been introduced:
chia/rpc/full_node_rpc_api.py:745: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]

`request: Dict[str,Any]` should fix that

Yakuhito:
Hey there! I know I was the one who originally opened an issue, but it seems this endpoint needs to be changed based on #13799. From my understanding, multiple mempool items can now exist that contain the same CoinSpend. I would make the endpoint return a list of those, and name it  `get_mempool_items_by_coin_name`. This could also allow you to delete the `if len(mempool_items) == 0` condition (just return an empty list).

kimsk:
> Easy enough to add a test to test the Exception case on line 747. I suspect the test code can't trigger the if is_overflow_block condition ( I might be inclined to just remove that if check from the test as it doesn't seem relevant to the specific test.)
@emlowe good suggestions. I update the tests and add the exception test (i.e., line 747).

kimsk:
> I would make the endpoint return a list of those, and name it get_mempool_items_by_coin_name. This could also allow you to delete the if len(mempool_items) == 0 condition (just return an empty list).

@Yakuhito  That was my original implementation 😄. Since 1.8.2 allows multiple mempool items with the same coin id now, I agree that we should update the RPC to return a list as well as its name to `get_mempool_items_by_coin_name`.

@emlowe is it fine to just use this PR or you prefer closing this one and re-open the new one?

kimsk:
Use this #16019 instead.

PR #15627, opened 2023-06-26
Handle VC syncing exceptions better

Quexington:
This just makes it so that standard non-debug logs don't get filled up with an exception that we expect to see when a VC is revoked.  It has the downside that if something genuinely goes wrong during syncing from the parent coin spend then we need to enable debug logs to see it, but this is probably fine.

### DIFF
--- a/chia/wallet/vc_wallet/vc_wallet.py
+++ b/chia/wallet/vc_wallet/vc_wallet.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import dataclasses
 import logging
 import time
+import traceback
 from typing import TYPE_CHECKING, List, Optional, Set, Tuple, Type, TypeVar, Union
 
 from blspy import G1Element, G2Element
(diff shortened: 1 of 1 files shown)


chia/wallet/vc_wallet/vc_wallet.py
```diff
@@ -103,7 +104,12 @@ async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection) -
                 f"Cannot get verified credential coin: {coin.name().hex()} puzzle and solution"
             )  # pragma: no cover
             return  # pragma: no cover
-        vc = VerifiedCredential.get_next_from_coin_spend(cs)
+        try:
+            vc = VerifiedCredential.get_next_from_coin_spend(cs)
+        except Exception as e:  # pragma: no cover
+            self.log.debug(
+                f"Syncing VC from coin spend failed (likely means it was revoked): {e}\n{traceback.format_exc()}"
+            )
```
trepca, line 112: shouldn't this return? otherwise it will run into and error in the next statement

Quexington: oops yeah

Quexington: fixed

trepca (approved):

PR #15628, opened 2023-06-26
Convert All CLI Commands to using async generators

jack60612:
### Purpose:
Cleaned up cli code, switched to an async generator

### Current Behavior:
No breaking changes, only internal cli changes

### Testing Notes:
Testing coming in next pr


jack60612:
i checked a couple commands, seems like its working well.

chia/cmds/coin_funcs.py
```diff
@@ -2,79 +2,88 @@
...
-    if show_unconfirmed:
-        print("\nUnconfirmed Removals:")
-        print_coins(
-            "\tPrevious Address: {} Amount: {}, Confirmed in block: {}\n",
-            [(cr.coin, str(cr.confirmed_block_index)) for cr in unconfirmed_removals],
-            mojo_per_unit,
-            addr_prefix,
-            paginate,
+async def async_list(
+    wallet_rpc_port: Optional[int],
```
arvidn, line 21: when you have this many parameters, some of them having the same type, I think it might be appropriate to add `*,` at the beginning of the parameter list, to require callers to name the arguments.

I think this would make sense for the other functions too

chia/cmds/cmds_util.py
```diff
@@ -207,18 +206,18 @@ async def get_wallet(root_path: Path, wallet_client: WalletRpcClient, fingerprin
...
-            return
-
-        await function(extra_params, wallet_client, new_fp)
+            yield None, 0, config
+        else:
+            new_fp = await get_wallet(root_path, wallet_client, fingerprint)
+            if new_fp is None:
+                yield None, 0, config
+            else:
+                yield wallet_client, new_fp, config
```
arvidn, line 223: this change (`execute_with_wallet` -> `get_wallet_client()` context manager) seems to be independent from the main change in this PR. It's also not obvious that it's an improvement. You gain an extra level of indentation in the commands for instance.
It seems reasonable to make this a separate PR (or drop it if it's not important).

Or is the issue that the previous function would have to pass through variadic argument list? That's simple in python though I would think. Perhaps mypy doesn't support figuring out the types when passing through a generic function like this

chia/cmds/plotnft.py
```diff
@@ -130,8 +122,7 @@ def join_cmd(wallet_rpc_port: Optional[int], fingerprint: int, id: int, fee: int
 
     from .plotnft_funcs import join_pool
 
-    extra_params = {"pool_url": pool_url, "id": id, "fee": fee, "yes": yes}
-    asyncio.run(execute_with_wallet(wallet_rpc_port, fingerprint, extra_params, join_pool))
+    asyncio.run(join_pool(wallet_rpc_port, fingerprint, pool_url, Decimal(fee), id, yes))
```
arvidn, line 125: `yes` is a pretty cryptic name

arvidn (commented):
I think this is a good direction. Do you have a sense of how far away this is from being able to remove the files from mypy exemptions? (doing so would increase confidence in the correctness of this change I think).

chia/cmds/cmds_util.py
jack60612: yeah mypy does not support figuring out the types, so that's why we do it this way,

jack60612: I could remove this change but it would make every other change more lines and just a mess.

jack60612:
> I think this is a good direction. Do you have a sense of how far away this is from being able to remove the files from mypy exemptions? (doing so would increase confidence in the correctness of this change I think).

I think i can probably remove the mypy exemptions from the command files not the func ones, but i'll try doing both.

chia/cmds/plotnft.py
jack60612: i kept the existing name but i could go around and change them.

jack60612:
> > I think this is a good direction. Do you have a sense of how far away this is from being able to remove the files from mypy exemptions? (doing so would increase confidence in the correctness of this change I think).
> 
> I think i can probably remove the mypy exemptions from the command files not the func ones, but i'll try doing both.

i had to reinstall mypy but i think i it.

chia/cmds/coin_funcs.py
jack60612: Just finished changing all the long ones.

chia/cmds/cmds_util.py
arvidn: right. if mypy can't see through it anyway, it kind of defeats the purpose of the patch

chia/cmds/cmds_util.py
```diff
@@ -207,18 +206,18 @@ async def get_wallet(root_path: Path, wallet_client: WalletRpcClient, fingerprin
     return selected_fingerprint
 
 
-async def execute_with_wallet(
+@asynccontextmanager
+async def get_wallet_client(
     wallet_rpc_port: Optional[int],
```
arvidn, line 211: it's surprising to me to have this parameter be optional, but not default it to `None`

chia/cmds/coin_funcs.py
```diff
@@ -105,120 +115,140 @@ def print_coins(
...
-            print("No coins to combine.")
+async def async_combine(
+    *,
+    wallet_rpc_port: Optional[int],
+    fingerprint: Optional[int],
+    wallet_id: int,
+    fee: Decimal,
+    max_coin_amount: Decimal,
+    min_coin_amount: Decimal,
+    excluded_amounts: list[str],
```
arvidn, line 126: I don't have a strong opinion on this, but the types in this call seem geared towards passing in strings that are then converted in this function itself. e.g. this could be `List[Decimal]` (btw, why is it lower case `list` right now?).

and `target_coin_ids_str` could be `List[Bytes32]`. In a way, it would seem appropriate to perform the conversion from string -> true type in the caller.

chia/cmds/plotnft_funcs.py
```diff
@@ -262,86 +274,98 @@ async def submit_tx_with_confirmation(
...
-    fee_mojos = uint64(int(fee * units["chia"]))
-    msg = f"\nWill claim rewards for wallet ID: {wallet_id}."
-    func = functools.partial(
-        wallet_client.pw_absorb_rewards,
-        wallet_id,
-        fee_mojos,
-    )
-    await submit_tx_with_confirmation(msg, False, func, wallet_client, fingerprint, wallet_id)
+async def self_pool(
+    wallet_rpc_port: Optional[int], fingerprint: int, fee: Decimal, wallet_id: int, prompt: bool
```
arvidn, line 331: I think it would be good to force callers to name the parameters here too, with the `*,`

chia/cmds/plotnft_funcs.py
```diff
@@ -262,86 +274,98 @@ async def submit_tx_with_confirmation(
...
+            {
+                "pool_wallet_info": pool_wallet_info,
+                "unconfirmed_transactions": [
+                    {"sent_to": tx.sent_to, "transaction_id": tx.name.hex()} for tx in unconfirmed_transactions
+                ],
+            }
+        )
+
+
+async def claim_cmd(wallet_rpc_port: Optional[int], fingerprint: int, fee: Decimal, wallet_id: int) -> None:
```
arvidn, line 357: and `*,` here

chia/cmds/wallet.py
```diff
@@ -316,13 +308,12 @@ def get_address_cmd(wallet_rpc_port: Optional[int], id, fingerprint: int, new_ad
 @click.option(
     "-m", "--fee", help="A fee to add to the offer when it gets taken, in XCH", default="0", show_default=True
 )
-def clawback(wallet_rpc_port: Optional[int], id, fingerprint: int, tx_ids: str, fee: str) -> None:  # pragma: no cover
-    extra_params = {"id": id, "tx_ids": tx_ids, "fee": fee}
-    import asyncio
-
+def clawback(
```
arvidn, line 311: maybe I'm going overkill with suggesting `*,` on these functions. if you think it makes sense, feel free to add them

chia/cmds/wallet.py
```diff
@@ -1406,26 +1326,17 @@ def vcs_cmd():  # pragma: no cover
...
-@click.option("-m", "--fee", help="Blockchain fee for mint transaction, in XCH", type=str, required=False)
-def _mint_vc(
+@click.option("-m", "--fee", help="Blockchain fee for mint transaction, in XCH", type=str, required=False, default="0")
+def mint_vc_cmd(
     wallet_rpc_port: Optional[int],
     fingerprint: int,
     did: str,
     target_address: Optional[str],
-    fee: Optional[str],
+    fee: str,
```
arvidn, line 1335: for future PRs, I think it's best to leave unrelated changes, like this one, for separate PRs. including the renaming of functions. it makes reviewing simpler.

chia/cmds/wallet_funcs.py
```diff
@@ -132,415 +131,454 @@ async def get_unit_name_for_wallet_id(
...
-
-    try:
-        wallet_type = await get_wallet_type(wallet_id=tx.wallet_id, wallet_client=wallet_client)
-        mojo_per_unit = get_mojo_per_unit(wallet_type=wallet_type)
-        name = await get_unit_name_for_wallet_id(
-            config=config,
-            wallet_type=wallet_type,
-            wallet_id=tx.wallet_id,
-            wallet_client=wallet_client,
+async def get_transaction(wallet_rpc_port: Optional[int], fingerprint: Optional[int], tx_id: str, verbose: int) -> None:
```
arvidn, line 134: I think this would benefit from `*,` as well

chia/cmds/wallet_funcs.py
```diff
@@ -132,415 +131,454 @@ async def get_unit_name_for_wallet_id(
...
+    async with get_wallet_client(wallet_rpc_port, fingerprint) as (wallet_client, fingerprint, config):
+        if wallet_client is None:
+            return
```
arvidn, line 137: instead of checking this condition everywhere you use this context manager, what would you think of throwing an exception instead? I think it would simplify using the context manager, and avoid the risk of forgetting this check

chia/cmds/wallet_funcs.py
```diff
@@ -132,415 +131,454 @@ async def get_unit_name_for_wallet_id(
...
-                result = await wallet_client.cat_asset_id_to_name(b32_id)
-                if result is not None:
-                    name = result[1]
-                else:
-                    name = "Unknown CAT"
-                unit = units["cat"]
-                if item in offers:
-                    fungible_asset_dict[name] = uint64(abs(int(Decimal(amount) * unit)))
-            except ValueError:
+        if len(offers) == 0 or len(requests) == 0:
```
arvidn, line 424: this is also an unrelated change (and I think a nicer form is `if offers == [] or requests == []`)

chia/cmds/wallet_funcs.py
```diff
@@ -791,297 +847,367 @@ def print_balance(amount: int, scale: int, address_prefix: str, *, decimal_only:
...
+                reuse_puzhash,
+            )
+            print(f"Successfully transferred DID to {target_address}")
+            print(f"Transaction ID: {response['transaction_id']}")
+            print(f"Transaction: {response['transaction']}")
+        except Exception as e:
+            print(f"Failed to transfer DID: {e}")
+
+
+async def find_lost_did(
```
arvidn, line 1039: I think this would benefit from `*,` too

arvidn (commented):
I think this looks good. I had a few comments. Especially avoiding checking to see of the context manager gave you `None` or not would be an important improvement.

I think it would be good if someone else could review this too

chia/cmds/wallet_funcs.py
jack60612: mypy was mad and made be change it, but ill change it to that form.

chia/cmds/wallet_funcs.py
jack60612: we have the weird none thing so we can cleanly output a pretty version of the traceback

jack60612: > we have the weird none thing so we can cleanly output a pretty version of the traceback

I would happily get rid of it but i think product would be mad.

chia/cmds/wallet.py
jack60612: it cant hurt

chia/cmds/wallet.py
jack60612: mypy also made me do this

chia/cmds/cmds_util.py
jack60612: fixed

chia/cmds/wallet.py
jack60612: added

arvidn:
I think this looks good. we should find one more person to review.
This is a step towards adding test coverage to the CLI, so I think it's OK to ignore the missing coverage for this PR.

altendky (changes requested):
i didn't do a detailed review of every change, but scanned through every line for smells anyways.  here are a few comments.

note that i didn't repeat every comment at every relevant site.  look around for repeats if making a change.

thanks for the progress!

chia/cmds/coins.py
```diff
@@ -42,12 +40,14 @@ def coins_cmd(ctx: click.Context) -> None:
     "--exclude-coin",
     "coins_to_exclude",
     multiple=True,
+    default=[],
```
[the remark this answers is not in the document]
jack60612: this allows it to be a list by default, as when it was a tuple we immediately converted it anyway

chia/cmds/wallet.py
```diff
@@ -33,12 +34,13 @@ def wallet_cmd(ctx: click.Context) -> None:
...
 @click.option("--verbose", "-v", count=True, type=int)
 def get_transaction_cmd(wallet_rpc_port: Optional[int], fingerprint: int, id: int, tx_id: str, verbose: int) -> None:
-    extra_params = {"id": id, "tx_id": tx_id, "verbose": verbose}
-    import asyncio
-
     from .wallet_funcs import get_transaction
 
-    asyncio.run(execute_with_wallet(wallet_rpc_port, fingerprint, extra_params, get_transaction))
+    asyncio.run(
+        get_transaction(  # Wallet ID is not used.
```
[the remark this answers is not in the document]
jack60612: changed

chia/cmds/wallet.py
```diff
@@ -316,13 +312,16 @@ def get_address_cmd(wallet_rpc_port: Optional[int], id, fingerprint: int, new_ad
...
-    import asyncio
-
+def clawback(
+    wallet_rpc_port: Optional[int], id: int, fingerprint: int, tx_ids: str, fee: str
+) -> None:  # pragma: no cover
     from .wallet_funcs import spend_clawback
 
-    asyncio.run(execute_with_wallet(wallet_rpc_port, fingerprint, extra_params, spend_clawback))
+    asyncio.run(
+        spend_clawback(  # Wallet ID is not Used.
```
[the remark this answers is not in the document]
jack60612: changed

chia/cmds/coins.py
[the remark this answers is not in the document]
altendky: Ah, `Tuple`, gotcha.  Could we just hint `Sequence` and not specify the default?  Related to our other keybase chat on hinting `Sequence` for parameters.

chia/cmds/wallet_funcs.py
```diff
@@ -132,415 +131,456 @@ async def get_unit_name_for_wallet_id(
...
+                raise
+
+
+async def make_offer(
+    *,
+    wallet_rpc_port: Optional[int],
+    fp: Optional[int],
+    d_fee: Decimal,
+    offers: list[str],
+    requests: list[str],
```
altendky, line 417:
```suggestion
    offers: List[str],
    requests: List[str],
```

chia/cmds/wallet_funcs.py
```diff
@@ -791,297 +849,368 @@ def print_balance(amount: int, scale: int, address_prefix: str, *, decimal_only:
...
+        except Exception as e:
+            print(f"Failed to update DID metadata: {e}")
+
+
+async def did_message_spend(
+    wallet_rpc_port: Optional[int],
+    fp: Optional[int],
+    did_wallet_id: int,
+    puzzle_announcements: list[str],
+    coin_announcements: list[str],
```
altendky, line 1002:
```suggestion
    puzzle_announcements: List[str],
    coin_announcements: List[str],
```

chia/cmds/wallet_funcs.py
```diff
@@ -791,297 +849,368 @@ def print_balance(amount: int, scale: int, address_prefix: str, *, decimal_only:
...
+    wallet_id: int,
+    royalty_address: Optional[str],
+    target_address: Optional[str],
+    no_did_ownership: bool,
+    hash: str,
+    uris: list[str],
+    metadata_hash: Optional[str],
+    metadata_uris: list[str],
+    license_hash: Optional[str],
+    license_uris: list[str],
```
altendky, line 1095:
```suggestion
    uris: List[str],
    metadata_hash: Optional[str],
    metadata_uris: List[str],
    license_hash: Optional[str],
    license_uris: List[str],
```

chia/cmds/wallet_funcs.py
```diff
@@ -1220,191 +1358,275 @@ def fungible_assets_from_offer(offer: Offer) -> List[Optional[bytes32]]:
...
+        print("Notification sent successfully.")
+        print(f"To get status, use command: chia wallet get_transaction -f {fingerprint} -tx 0x{tx.name}")
 
 
-async def get_notifications(args: Dict, wallet_client: WalletRpcClient, fingerprint: int) -> None:
-    ids: Optional[List[bytes32]] = [bytes32.from_hexstr(id) for id in args["ids"]]
-    if ids is not None and len(ids) == 0:
-        ids = None
+async def get_notifications(
+    wallet_rpc_port: Optional[int], fp: Optional[int], str_ids: list[str], start: Optional[int], end: Optional[int]
```
altendky, line 1384:
```suggestion
    wallet_rpc_port: Optional[int], fp: Optional[int], str_ids: List[str], start: Optional[int], end: Optional[int]
```

chia/cmds/wallet_funcs.py
```diff
@@ -1220,191 +1358,275 @@ def fungible_assets_from_offer(offer: Offer) -> List[Optional[bytes32]]:
...
+        notifications = await wallet_client.get_notifications(ids=ids, pagination=(start, end))
+        for notification in notifications:
+            print("")
+            print(f"ID: {notification.coin_id.hex()}")
+            print(f"message: {notification.message.decode('utf-8')}")
+            print(f"amount: {notification.amount}")
+
+
+async def delete_notifications(
+    wallet_rpc_port: Optional[int], fp: Optional[int], str_ids: list[str], delete_all: bool
```
altendky, line 1402:
```suggestion
    wallet_rpc_port: Optional[int], fp: Optional[int], str_ids: List[str], delete_all: bool
```

chia/cmds/wallet_funcs.py
```diff
@@ -1220,191 +1358,275 @@ def fungible_assets_from_offer(offer: Offer) -> List[Optional[bytes32]]:
...
-    if args["root_only"]:
-        print(f"Proof Hash: {VCProofs(proof_dict).root()}")
-        return
-    else:
-        await wallet_client.vc_add_proofs(proof_dict)
-        print("Proofs added to DB successfully!")
-        return
 
+async def add_proof_reveal(
+    wallet_rpc_port: Optional[int], fp: Optional[int], proofs: list[str], root_only: bool
```
altendky, line 1563:
```suggestion
    wallet_rpc_port: Optional[int], fp: Optional[int], proofs: List[str], root_only: bool
```

chia/cmds/coin_funcs.py
```diff
@@ -2,79 +2,89 @@
...
-            addr_prefix,
-            paginate,
+async def async_list(
+    *,
+    wallet_rpc_port: Optional[int],
+    fingerprint: Optional[int],
+    wallet_id: int,
+    max_coin_amount: Decimal,
+    min_coin_amount: Decimal,
+    excluded_amounts: list[int],
```
altendky, line 27:
```suggestion
    excluded_amounts: List[int],
```

chia/cmds/coin_funcs.py
altendky, line 126:
```suggestion
    excluded_amounts: List[str],
```

altendky: @arvidn, do we want to enforce the `str` -> `Decimal` conversion location in this PR?

altendky (changes requested):

chia/cmds/coin_funcs.py
altendky: We'll take this up later.

arvidn: I think it would be fine to punt that to a second PR. @jack60612 would you mind doing that?

altendky (approved):

arvidn (approved):

PR #15629, opened 2023-06-26
Fix fork height constant for simulator / block tools

jack60612:
### Purpose:
Allows simulator / tests to generate blocks fast to allow for testing.

### Current Behavior:
Does not work when fork is activated

### New Behavior:
Works when fork is activated

### Testing Notes:
N/A

### DIFF
--- a/tests/blockchain/test_blockchain.py
+++ b/tests/blockchain/test_blockchain.py
@@ -1510,7 +1510,7 @@ class TestBlockHeaderValidation:
             # enable softfork2 at height 0, to make it apply to this test
             # the test constants set MAX_FUTURE_TIME to 10 days, restore it to
             # default for this test
-            constants = bt.constants.replace(SOFT_FORK2_HEIGHT=0, MAX_FUTURE_TIME=5 * 60)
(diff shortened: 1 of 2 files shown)


tests/blockchain/test_blockchain.py
```diff
@@ -1510,7 +1510,7 @@ async def test_bad_timestamp(self, bt, with_softfork2):
             # enable softfork2 at height 0, to make it apply to this test
             # the test constants set MAX_FUTURE_TIME to 10 days, restore it to
             # default for this test
-            constants = bt.constants.replace(SOFT_FORK2_HEIGHT=0, MAX_FUTURE_TIME=5 * 60)
+            constants = bt.constants.replace(SOFT_FORK2_HEIGHT=0, MAX_FUTURE_TIME=5 * 60, MAX_FUTURE_TIME2=2 * 60)
```
arvidn, line 1513: this is already the default. I don't understand in what way this makes a difference

jack60612: @arvidn  its using the test constants which have a higher timestamp limit (10 days), I am setting it to the mainnet constants, because the test needs to use the real constants in order to test bad timestamps.

arvidn (approved):

PR #15634, opened 2023-06-27
Apply Kevin's suggestion to send peak height instead of filter prefix bits as part of NewSignagePoint

AmineKhaldi:


xdustinface (changes requested):
You can now also remove `filter_prefix_bits` parameter from `verify_and_get_quality_string`, the assertion in there and the related testing around it and make `height` non-optional.

AmineKhaldi:
> You can now also remove `filter_prefix_bits` parameter from `verify_and_get_quality_string`, the assertion in there and the related testing around it and make `height` non-optional.

That's planned for the next PR where we get rid several things filter prefix bits related (including the one you mentioned and the sp hash to filter prefix bits, in favor of sps[0])

xdustinface (dismissed):

tests/farmer_harvester/test_filter_prefix_bits.py
```diff
@@ -50,29 +55,58 @@ def state_changed_callback(change: str, change_data: Optional[Dict[str, Any]]) -
...
+        await time_out_assert(5, state_has_changed, True)
+        # We're intercepting the harvester's state changes as we're expecting
+        # a farming_info one.
+        assert state_change == "farming_info"
+        assert state_change_data is not None
+        assert state_change_data.get("eligible_plots") == eligible_plots
+        farmer_rpc_cl.close()
+        harvester_rpc_cl.close()
+        await farmer_rpc_cl.await_closed()
+        await harvester_rpc_cl.await_closed()
```
arvidn, line 112: I think this needs to live in a `finally` block, no?

tests/farmer_harvester/test_filter_prefix_bits.py
```diff
@@ -50,29 +55,58 @@ def state_changed_callback(change: str, change_data: Optional[Dict[str, Any]]) -
...
-    await farmer_api.new_signage_point(sp)
-    await time_out_assert(5, state_has_changed, True)
-    # We're intercepting the harvester's state changes as we're expecting
-    # a farming_info one.
-    assert state_change == "farming_info"
-    assert state_change_data is not None
-    assert state_change_data.get("eligible_plots") == eligible_plots
+    new_config = local_b_tools._config
+    local_b_tools.change_config(new_config)
+    async for harvester_services, farmer_service, _ in setup_farmer_multi_harvester(
```
arvidn, line 68: I think `async with` is a lot more readable than `async for`

arvidn: I understand now that you can't use `async with`

arvidn (commented):
I think this looks reasonable. @xdustinface what do you think?

tests/farmer_harvester/test_filter_prefix_bits.py
xdustinface: I think so, thats also why i think using fixtures makes more sense, because _i think_ pytest already takes care of it? And if not we can out the finally there without having to indent tests.

xdustinface (commented):
Im ok with but i would still prefer to use the fixture and adjust the consensus constants of the farmer/harvester in place (inside the test) over this last change, i will defer to you here @arvidn. Even if replacing them is a little hacky we at least don't reimplement the fixture in this test.

arvidn:
swapping out the constants mid-test seems really sketchy to me. in that, different parts of the test runs with different constants, is asking for future trouble.

I don't have a strong opinion on leaving this as-is (without `finally`) or making a custom fixture for just this test (that would live just above the test function, *not* in `conftest.py`). My understanding is that in the future these fixtures will be proper context managers, and we can use them in `async with` blocks, and they will be able to take care of their own cleanup.

arvidn (approved):

tests/farmer_harvester/test_filter_prefix_bits.py
AmineKhaldi: I made a local custom fixture for this test.

arvidn (approved):

xdustinface (approved):

PR #15666, opened 2023-06-29
Introduce TXConfig and CoinSelectionConfig

Quexington:
This PR introduces the concept of `TXConfig` and `CoinSelectionConfig` which are mandatory parameters to transaction functions and coin selection functions respectively.  The idea here is to consolidate a lot of general configuration options that apply to all transactions into a single dataclass so it's not easy to forget them.

There exists another problem currently which is that every level of the function stack is required to decide whether or not it cares to set a default value for these parameters, and if so, it has to re implement the default setting behavior.  You see lots of duplicated blocks like this: https://github.com/Chia-Network/chia-blockchain/pull/15666/files#diff-1cebb819eec76392877939438d8c2f55d405f65d98c4948608e2569dbf6b2a81L584-L591 .  By making these `-Config` classes manadatory, it forces the default setting to the very top of the function stack.  1) take user input 2) autofill any defaults 3) pass the config object down.  It's for this reason as well that the `-Config` dataclasses do not have any `Optional` parameters.  All defaults must be set before proceeding down the function stack.

To facilitate uniform user input parsing and default setting, there are two `-Loader` classes for each `-Config` class.  These loader classes are streamable and allow for optional parameters.  The idea is that you can parse user input with these classes using a method like `from_json_dict` then call their `autofill` method to generate a `-Config` object with the proper defaults set.  `-Loader` objects ideally should never be passed to another function, only the results from calls to `autofill`.

A final motivation behind this PR is to make it so that new configurations are introduced with ease, adding to these dataclasses will propagate the change to every relevant RPC, wallet function, etc.


chia-blockchain-gui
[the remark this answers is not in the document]
Quexington: Nope

tests/cmds/test_wallet_funcs.py
```diff
@@ -0,0 +1,307 @@
...
+import contextlib
+import io
+import sys
+from decimal import Decimal
+
+import click.testing
+import pytest
+
+from chia.cmds.wallet_funcs import (
+    add_uri_to_nft,
```
geoffwalmsley, line 12: I'm not sure whether this should also include tests for vanilla spends, and VC spends?

tests/wallet/nft_wallet/test_nft_bulk_mint.py
geoffwalmsley, line 1: I think the bulk minting repo will need updating to include this change (which is a job for me)

chia/rpc/wallet_rpc_api.py
geoffwalmsley, line 1: Will these RPC changes have any impact on GUI? Or on third-party apps that plugin to the rpc?

Quexington: The RPC changes should only be additive.  I tried to leave all of the old names in there (still able to be parsed) for backwards compatibility and all of the new options are optional and can be in the request or not.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -717,24 +715,33 @@ async def transfer_did(
...
-        uncurried = did_wallet_puzzles.uncurry_innerpuz(new_innerpuzzle)
-        assert uncurried is not None
-        p2_puzzle = uncurried[0]
+        if tx_config.reuse_puzhash:
+            new_innerpuzzle_hash = innerpuz.get_tree_hash()
+            uncurried = did_wallet_puzzles.uncurry_innerpuz(innerpuz)
+            assert uncurried is not None
+            p2_ph = uncurried[0].get_tree_hash()
+        else:
+            p2_ph = await self.standard_wallet.get_puzzle_hash(new=False)
```
geoffwalmsley, line 735: should this be `new=True` to return a new puzzle hash?

geoffwalmsley (approved):

wjblanke (approved):
aok

altendky (dismissed):
just holding this while i check with quex on something

PR #15668, opened 2023-06-30
add get_wallet_addresses RPC for deriving wallet addresses

paninaro:
### Purpose:
Adds a new daemon RPC for deriving wallet addresses for one or more keys specified by their fingerprints. The functionality is equivalent to the `chia keys derive wallet-address` CLI.

RPC params include:
* (optional) `fingerprints`: an array of fingerprints identifying wallet keys used for derivation. If omitted, all keys from the keyring are used.
* (optional) `count`: a number specifying how many wallet addresses will be derived. If omitted, defaults to `1`
* (optional) `index`: a number specifying the wallet key index to begin derivations from. If omitted, defaults to `0`
* (optional) `non_observer_derivation`: a boolean specifying whether to perform observer/non-observer key derivations. If omitted, defaults to 'false'

Example usage:
```
$ chia rpc daemon get_wallet_addresses '{"fingerprints": [123456789, 998877], "count": 2, "index": 5, "non_observer_derivation": false}'
{
    "success": true,
    "wallet_addresses": {
        "123456789": [
            {
                "address": "txch1chf8g4fefmygermfxs6252yxddnxr2geqw6fr2834d7wpkhljd5q98ten7",
                "hd_path": "m/12381/8444/2/5"
            },
            {
                "address": "txch1g4n09emepmxt5840ajcquxnrquym28cn6tkc5chpxqk3ujg5x46qs8ep4g",
                "hd_path": "m/12381/8444/2/6"
            }
        ],
        "998877": [
            {
                "address": "txch1xvqzxn5ndhfwn70raqqwzwzjn5pndftsumh3t4xnt5tukeeg046sz6etqq",
                "hd_path": "m/12381/8444/2/5"
            },
            {
                "address": "txch1hugjzz3pwcfg6sw22mq4xakm4ne8alj53w0q4r7ewpw9lxhxp2hsz583xj",
                "hd_path": "m/12381/8444/2/6"
            }
        ]
    }
}
```

### Current Behavior:
This is a new RPC. Equivalent functionality is available from the CLI.

### New Behavior:
Exposes a new `get_wallet_addresses` RPC from the daemon service.

### Testing Notes:
Tests included. The change has also been manually tested.

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -23,15 +23,18 @@ from typing_extensions import Protocol
 from chia import __version__
 from chia.cmds.init_funcs import check_keys, chia_full_version_str, chia_init
 from chia.cmds.passphrase_funcs import default_passphrase, using_default_passphrase
+from chia.consensus.coinbase import create_puzzlehash_for_pk
 from chia.daemon.keychain_server import KeychainServer, keychain_commands
 from chia.daemon.windows_signal import kill
 from chia.plotters.plotters import get_available_plotters
 from chia.plotting.util import add_plot_directory
 from chia.server.server import ssl_context_for_server
+from chia.util.bech32m import encode_puzzle_hash
 from chia.util.beta_metrics import BetaMetricsLogger
 from chia.util.chia_logging import initialize_service_logging
 from chia.util.config import load_config
 from chia.util.errors import KeychainCurrentPassphraseIsInvalid
+from chia.util.ints import uint32
 from chia.util.json_util import dict_to_json_str
 from chia.util.keychain import Keychain, passphrase_requirements, supports_os_passphrase_storage
 from chia.util.lock import Lockfile, LockfileError
@@ -39,6 +42,7 @@ from chia.util.network import WebServer
 from chia.util.service_groups import validate_service
 from chia.util.setproctitle import setproctitle
 from chia.util.ws_message import WsRpcMessage, create_payload, format_response
+from chia.wallet.derive_keys import master_sk_to_wallet_sk, master_sk_to_wallet_sk_unhardened
 
 io_pool_exc = ThreadPoolExecutor()
 
@@ -389,6 +393,7 @@ class WebSocketServer:
             "get_version": self.get_version,
             "get_plotters": self.get_plotters,
             "get_routes": self.get_routes,
+            "get_wallet_addresses": self.get_wallet_addresses,
         }
 
     async def is_keyring_locked(self, websocket: WebSocketResponse, request: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


tests/core/daemon/test_daemon.py
```diff
@@ -60,6 +62,9 @@ class Daemon:
     services: Dict[str, Union[List[Service], Service]]
     connections: Dict[str, Optional[List[Any]]]
 
+    # Instance variables used by WebSocketServer.get_wallet_addresses()
+    net_config: Dict[str, Any]
```
altendky, line 66: Maybe this would be nice?

```suggestion
    net_config: Dict[str, Any] =  field(default_factory=dict)
```

tests/core/daemon/test_daemon.py
```diff
@@ -438,6 +476,123 @@ async def test_get_routes(mock_lonely_daemon):
     }
 
 
+@pytest.mark.parametrize(
```
altendky, line 479: Do a search for `@datacases`.  Earle and I put it together a few weeks ago.  It is in some ways more verbose.  In others, more clear, better type checked, and has one less layer of indentation iirc.  In some places we want a calculated `.id` attribute and use `@property`.  But that can be a regular attribute as well.

altendky (changes requested):

tests/core/daemon/test_daemon.py
```diff
@@ -438,6 +476,123 @@ async def test_get_routes(mock_lonely_daemon):
...
+@pytest.mark.asyncio
+async def test_get_wallet_addresses(
+    mock_daemon_with_config_and_keys, monkeypatch, rpc_request, pubkeys_only, expected_result
+):
+    daemon = mock_daemon_with_config_and_keys
+
+    original_get_keys = Keychain.get_keys
+
+    # monkeypatch Keychain.get_keys() to always call get_keys() with include_secrets=False
+    def mock_get_keys(self, include_secrets=False):
```
altendky, line 584: I think this layer ended up not needed?

tests/core/daemon/test_daemon.py
paninaro: Replaced with datacases

tests/core/daemon/test_daemon.py
paninaro: Updated

tests/core/daemon/test_daemon.py
paninaro: Perhaps there's another way to write it, but this code is used and works.

altendky (approved):

arvidn (approved):

chia/daemon/server.py
```diff
@@ -561,6 +566,58 @@ async def get_routes(self, websocket: WebSocketResponse, request: Dict[str, Any]
...
+                return {"success": False, "error": f"key(s) not found for fingerprint(s) {missing_fingerprints}"}
+
+        selected = self.net_config["selected_network"]
+        prefix = self.net_config["network_overrides"]["config"][selected]["address_prefix"]
+
+        wallet_addresses_by_fingerprint = {}
+        for key in keys:
+            address_entries = []
+
+            # we require access to the private key to generate wallet addresses
```
wjblanke, line 599: this isnt true for unhardened tho is it?

paninaro: Well yes and no... If given the wallet's root public key, you're right, we should be able to derive the unhardened addresses from that wallet root public key. But we don't store public keys today, and the `master_sk_to_wallet_sk_unhardened()` function currently assumes that the master private key is provided, so we'd have to make more sweeping changes to support observer mode.

wjblanke (approved):
aok

PR #15675, opened 2023-07-01
Fix DID resync

ytx1991:
### Purpose:
During the resync, it may create DID wallets that don't belong to the current key. These wallets will not be removed after fully synced. It confuses users.

### New Behavior:
Only create DID wallet when we see the DID coin the first time, and it is unspent.


chia/wallet/wallet_state_manager.py
```diff
@@ -985,6 +985,11 @@ async def handle_did(
                     assert wallet.did_info.origin_coin is not None
                     if origin_coin.name() == wallet.did_info.origin_coin.name():
                         return WalletIdentifier.create(wallet)
+            if coin_state.spent_height is not None:
+                # The first coin we received for DID wallet is spent.
+                # This means the wallet is in a resync process, skip the coin
+                print("test")
```
xdustinface, line 991:
```suggestion
```

xdustinface (changes requested):
There is a `print` which should be dropped. And do you understand why this even happens during a sync because actually the coin states should be sorted so that unspent coin states come before spent ones? I know there are cases where we due to batching/task processing of the coin states might still process them out of order but here this should probably not be this case.. ? After quick check this feels like this might become flaky because you not really force the coin states to be out of order in the added test.

ytx1991:
I have added a check while fixing Clawback resync. It will log something if the coin state in add_coin_states is unspent but in the latest block the coin state is spent. I resynced several times but never see that log printed so I assume during the resync, the wallet will not receive an unspent coin state for a spent coin. I thought there should be two coin states for a spent coin, but seems like it only has one coin state.

xdustinface (changes requested):
> so I assume during the resync, the wallet will not receive an unspent coin state for a spent coin.

Okay yeah thats right. I digged more into this and its indeed because we alter the states in the full node and of course then only send the spent states after.

I think it would be good to add the the proposed renaming for the test, it makes it easier to read/understand it. 

Thinking about this test maybe you should just resync both wallets and also assert at the end that the DID wallets (not) exists for the particular wallet instead of only checking the tx record and unspent coin count?

I also think, now you only test the resync after the coin was spent but not if the coin is still unspent. Do you think we should test all cases?

tests/wallet/did_wallet/test_did.py
```diff
@@ -1219,3 +1220,103 @@ async def test_create_did_with_recovery_list(self, self_hostname, two_nodes_two_
...
+        wallet_node_2._close()
+        await wallet_node_2._await_closed()
+        new_config_1 = wallet_node_1.config.copy()
+        new_config_1["reset_sync_for_fingerprint"] = fingerprint_1
+        new_config_1["database_path"] = "wallet/db/blockchain_wallet_v2_test_1_CHALLENGE_KEY.sqlite"
+        new_config_2 = wallet_node_2.config.copy()
+        new_config_2["reset_sync_for_fingerprint"] = fingerprint_2
+        new_config_2["database_path"] = "wallet/db/blockchain_wallet_v2_test_2_CHALLENGE_KEY.sqlite"
+        wallet_node_1.config = new_config_1
+        wallet_node_2.config = new_config_2
```
xdustinface, line 1307:
```suggestion
        wallet_node_1.config["reset_sync_for_fingerprint"] = fingerprint_1
        wallet_node_1.config["database_path"] = "wallet/db/blockchain_wallet_v2_test_1_CHALLENGE_KEY.sqlite"
        wallet_node_2.config["reset_sync_for_fingerprint"] = fingerprint_2
        wallet_node_2.config["database_path"] = "wallet/db/blockchain_wallet_v2_test_2_CHALLENGE_KEY.sqlite"
```

Why this copying?

xdustinface (changes requested):

xdustinface (approved):

wjblanke (dismissed):
aok

wjblanke (approved):
aok thanks kyle

PR #15679, opened 2023-07-03
Pass full version in `Handshake`

felixbrucker:
### Purpose:

Currently the `Handshake` contains a cut down version of the package version (`__version__`). As it is desired to get the exact version of peers for a future PR related to farming/pooling and there is no downside in having more/exact info on the version of peers this PR changes this to now include the full `__version__`. The `version` of a `WSChiaConnection` is currently only used by the seeder via `get_version`, which was refactored to return the desired cut down version. Additionally legacy code was removed in favor of using `packaging.version.Version`, same as we already do in the `WSChiaConnection` for parsing the Handshake version.

### Current Behavior:

A `Handshake` contains a cut down version, for example: `__version__` is `1.8.2rc3.dev116`, the handshake contains `1.8.2rc3` instead. Another example: `__version__` is `2.0.0b2.dev0+og-1.4.0`, the handshake contains `2.0.0b2`.

### New Behavior:

The `Handshake` now contains `__version__` as is, so from the examples above: `1.8.2rc3.dev116` or `2.0.0b2.dev0+og-1.4.0` respectively.
The `get_chia_version_str()` method on a `WSChiaConnection` returns the cut down version, same as before.

### Testing Notes:

I added tests for the three "new" functions which are responsible for retrieving the "chia version" (cut down version) of the package or a passed `Version` or `str` instance, these basically replace the `chia_full_version_str` from `init_funcs.py`.


hoffmang9:
This probably has unintended consequences for the crawler/seeder as it will significantly change all downstream version reporting on the dashboards. This is may drive inconsistency between versions inside the windows installer as the windows installer has to be int.int.int.

felixbrucker:
> This probably has unintended consequences for the crawler/seeder as it will significantly change all downstream version reporting on the dashboards.

The version used by the seeder/crawler remains unchanged (cut down version), or are you referring to something else?

> This is may drive inconsistency between versions inside the windows installer as the windows installer has to be int.int.int.

The `chia_full_version_str()` usages remained unchanged (using the cut down version)


The only place which actually changed is the version in the handshake message

felixbrucker:
Waiting for CNI re versioning schema decisions

felixbrucker:
Since i opened this PR about 3 months ago the `version` on the connection is now a `str` again, i have updated the code accordingly.

With this being said i don't actually think we need a versioning discussion for this PR anymore, as the field in the protocol is a string and as indicated by others is already being used with values other than `x.y.z`. As such the result of `get_version` on the connection is not guaranteed to yield a `x.y.z` version string, it can contain any string value instead.

This PR will try to transform the version string from the handshake into the `x.y.z` format, but fallback to the string value (as was returned before) for its currently only use: the seeder.

I don't think the fact that new clients now send a more detailed/exact version in case they indeed are different from the official releases is a breaking change, instead it is how other custom nodes already behave.

chia/server/ws_connection.py
```diff
@@ -694,8 +695,8 @@ async def _read_one_message(self) -> Optional[Message]:
         return None
 
     # Used by the Chia Seeder.
-    def get_version(self) -> str:
-        return self.version
+    def get_chia_version_str(self) -> str:
+        return chia_version_str_from_str(self.version) or self.version
```
emlowe, line 699: I suggest keeping the existing get_version call as is -  and not adding in the `get_chia_version_str` call
This can be adjusted on the crawler side (regardless the crawler needs to be updated anyway).

chia/seeder/crawler.py
```diff
@@ -140,7 +140,7 @@ async def connect_task(self, peer: PeerRecord) -> None:
 
         async def peer_action(peer: WSChiaConnection) -> None:
             peer_info = peer.get_peer_info()
-            version = peer.get_version()
+            version = peer.get_chia_version_str()
```
emlowe, line 143: I would call get_version here (as mentioned earlier) and then use the standalone `chia_version_str_from_str`

chia/util/chia_version.py
```diff
@@ -0,0 +1,24 @@
...
+    release_version_str = f"{version.major}.{version.minor}.{version.micro}"
+
+    return release_version_str if version.pre is None else release_version_str + "".join(map(str, version.pre))
+
+
+def chia_version_str_from_str(version: str) -> Optional[str]:
+    try:
+        return chia_version_str_from_version(Version(version))
+    except InvalidVersion:
+        return None
```
emlowe, line 24: I might suggest returning the OG version as passed in if it fails.

chia/util/chia_version.py
```diff
@@ -0,0 +1,24 @@
...
+from packaging.version import InvalidVersion, Version
+
+from chia import __version__
+
+
+def chia_version_str() -> str:
+    return chia_version_str_from_version(Version(__version__))
+
+
+def chia_version_str_from_version(version: Version) -> str:
```
emlowe, line 14: One could overload this function - I might also suggest chia_short_version() - I think something like this would work and is correct

@overload
def chia_short_version(version: Version) -> str:

@overload
def chia_short_version(version: str) -> str:

emlowe (changes requested):
Suggested a few changes

emlowe:
After taking some time to study the code, I think some minor adjustments to the proposed code changes and this should be ok. The Handshake version string isn't really used anywhere other than the crawler

Note, this does require the crawler to need an update such that it does some post-processing on the version in the Handshake into the form it is expecting. But this is a reasonable lift.

An alternative approach would be to change the code such that only the harvester uses a modified version string, which would have modestly less implications, at the cost of some muckity muck.

emlowe:
I'm going to go back a bit - at least specifically for 3rd party harvesters

For 3rd Party Harvesters, this isn't needed at all. That harvester can put whatever they want into the Handshake version string. and the farmer can use it from the connection `get_version()` call

chia/util/chia_version.py
felixbrucker: As python doesn't have real overload support and we need a function which handles all cases anyway i just opted for an optional union type. Let me know if you prefer this over the previous approach

felixbrucker:
From discord:

> While that is true, it's not true for otherwise unchanged chia code, where a cut down version is sent
Sending the full version is the right thing to do
A future pr will send the harvester peer version as a separate header, similarly to your example
It's just that having a potentially incorrect version is not desired

emlowe (commented):
Approved

emlowe (dismissed):

felixbrucker:
I have made one more commit to ensure we extract major.minor.patch from versions which are invalid pep440 but contain a major.minor.patch version. This is now the same behavior as `chia_version_number` previously did with manually splitting the string etc.

emlowe (approved):

emlowe:
coverage exemption for a couple of lines not previously covered

PR #15680, opened 2023-07-03
Fix variable name typo

xearl4:
Just fixing a minor variable name typo correcting "smc" -> "scm".


jack60612 (approved):
LGTM, Thanks for your contribution!
We should probably clean up this file too, I don't think we need some of these edge cases anymore.

arvidn (approved):

wjblanke (approved):
aok

PR #15688, opened 2023-07-06
Chip13

fchirica:


chia/consensus/block_header_validation.py
```diff
@@ -491,6 +491,27 @@ def validate_unfinished_header_block(
     if q_str is None:
         return None, ValidationError(Err.INVALID_POSPACE)
 
+    # 5c. Check plot id is not present within last `NUM_DISTINCT_CONSECUTIVE_PLOT_IDS` blocks.
+    if height >= constants.SOFT_FORK3_HEIGHT:
+        curr_optional_block_record: Optional[BlockRecord] = prev_b
```
arvidn, line 496: @fchirica if you look up `FullBlock`s here, rather than `BlockRecord`, there's a chance this patch could be a lot simpler. That way you wouldn't need to add any fields to `BlockRecord`.

arvidn: the `BlockchainInterface` is tightly coupled with `BlockRecord`. To add these two fields *without* adding them to the `BlockRecord` class would be quite messy. It's not just about querying block records, it's about adding them as well

tests/blockchain/config.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import annotations
 
 parallel = True
-job_timeout = 60
+job_timeout = 90
```
arvidn, line 4: this is a pretty significant increase. I would hope we would try to avoid tests taking this long. Do you have a sense of what's pushing the timings this high?

chia/consensus/block_header_validation.py
```diff
@@ -491,6 +496,28 @@ def validate_unfinished_header_block(
...
+                plot_id,
+                curr_optional_block_record.pos_ss_cc_challenge_hash,
+                curr_optional_block_record.cc_sp_hash,
+            ):
+                return None, ValidationError(Err.INVALID_POSPACE)
+
+            if curr_optional_block_record.cc_sp_hash != curr_sp:
+                sp_count += 1
+                curr_sp = curr_optional_block_record.cc_sp_hash
+            curr_optional_block_record = blocks.try_block_record(curr_optional_block_record.prev_hash)
```
arvidn, line 519: this block will have a different height, and so you need to recompute `prefix_bits`. You should probably do this at the top of the loop, where you've already checked that `curr_optional_block_recrod` is not `None`.
```
prefix_bits = calculate_prefix_bits(constants, curr_optional_block_record.height)
```

chia/simulator/block_tools.py
```diff
@@ -540,6 +542,31 @@ def get_farmer_wallet_tool(self) -> WalletTool:
     def get_pool_wallet_tool(self) -> WalletTool:
         return WalletTool(self.constants, self.pool_master_sk)
 
+    # Verifies if the given plot passed any of the previous `UNIQUE_PLOTS_WINDOW` plot filters.
+    def plot_id_passed_previous_filters(
+        self, height: uint32, plot_id: bytes32, cc_sp_hash: bytes32, blocks: List[FullBlock]
```
arvidn, line 547: it looks suspicious that `height` is pass in here, rather than using the height from the `FullBlock`

chia/simulator/block_tools.py
```diff
@@ -540,6 +542,31 @@ def get_farmer_wallet_tool(self) -> WalletTool:
...
+            if sp_count >= self.constants.UNIQUE_PLOTS_WINDOW:
+                return False
+
+            challenge = block.reward_chain_block.pos_ss_cc_challenge_hash
+            if block.reward_chain_block.challenge_chain_sp_vdf is None:
+                # Edge case of first sp (start of slot), where sp_iters == 0
+                cc_sp_hash = challenge
+            else:
+                cc_sp_hash = block.reward_chain_block.challenge_chain_sp_vdf.output.get_hash()
+            prefix_bits = calculate_prefix_bits(self.constants, height)
```
arvidn, line 561:
```suggestion
            prefix_bits = calculate_prefix_bits(self.constants, block.height)
```

tests/blockchain/test_blockchain.py
```diff
@@ -3634,3 +3634,46 @@ async def test_reorg_flip_flop(empty_blockchain, bt):
...
+            peak = b.get_peak()
+            assert peak is not None
+            if peak.height != height:
+                break
+
+        peak = b.get_peak()
+        assert peak is not None
+
+        if bt_respects_soft_fork3 or unique_plots_window == 1:
+            assert peak.height == 24
```
arvidn, line 3673: I think it would be good to add some comments here, to explain these values we expect. why is this 24?

tests/blockchain/test_blockchain.py
```diff
@@ -3634,3 +3634,46 @@ async def test_reorg_flip_flop(empty_blockchain, bt):
...
+        else:
+            if soft_fork3_height == 0:
+                assert peak.height < 24
+            elif soft_fork3_height == 10:
+                assert peak.height < 24 and peak.height >= 9
```
arvidn, line 3678: why do we not know what height to expect in this case? where does 9 come from?

tests/conftest.py
```diff
@@ -112,6 +113,8 @@ def consensus_mode(request):
 def blockchain_constants(consensus_mode) -> ConsensusConstants:
     if consensus_mode == Mode.PLAIN:
         return test_constants
+    if consensus_mode == Mode.SOFT_FORK3:
+        return test_constants.replace(SOFT_FORK3_HEIGHT=0)
```
arvidn, line 117: setting this to 2 might expose more interesting conditions

tests/conftest.py
```diff
@@ -177,25 +180,37 @@ def softfork_height(request) -> int:
 
 @pytest.fixture(scope="session")
 def default_400_blocks(bt):
+    version = ""
+    if bt.constants.SOFT_FORK3_HEIGHT == 0:
```
arvidn, line 184: I think it would be clearer to depend on the `consensus_mode` fixture here, and check if you're in `Mode.SOFT_FORK3`. There are a few similar cases below as well.

tests/conftest.py
```diff
@@ -257,16 +290,19 @@ def default_2000_blocks_compact(bt):
...
-        bt,
-        normalized_to_identity_cc_eos=True,
-        normalized_to_identity_icc_eos=True,
-        normalized_to_identity_cc_ip=True,
-        normalized_to_identity_cc_sp=True,
-        seed=b"1000_compact",
-    )
+    if bt.constants.SOFT_FORK3_HEIGHT == 0:
+        pytest.skip("Test cache not available yet")
+    else:
```
arvidn, line 295: I don't think you need to put this in an else-block. you can preserve the indentation level

tests/conftest.py
```diff
@@ -327,8 +363,13 @@ async def five_nodes(db_version: int, self_hostname, blockchain_constants):
...
 
 @pytest_asyncio.fixture(scope="function")
-async def wallet_nodes():
-    async_gen = setup_simulators_and_wallets(2, 1, {"MEMPOOL_BLOCK_BUFFER": 1, "MAX_BLOCK_COST_CLVM": 400000000})
+async def wallet_nodes(blockchain_constants):
+    constants = blockchain_constants
+    async_gen = setup_simulators_and_wallets(
+        2,
+        1,
+        {"MEMPOOL_BLOCK_BUFFER": 1, "MAX_BLOCK_COST_CLVM": 400000000, "SOFT_FORK3_HEIGHT": constants.SOFT_FORK3_HEIGHT},
```
arvidn, line 371: if you make this fixture parameterized by `blockchain_constants`, it doesn't look right to *only* forward the soft-fork3 height. there's no parameter to pass in the whole constants object?

tests/conftest.py
```diff
@@ -355,8 +396,10 @@ async def two_nodes_sim_and_wallets():
 
 
 @pytest_asyncio.fixture(scope="function")
-async def two_nodes_sim_and_wallets_services():
-    async for _ in setup_simulators_and_wallets_service(2, 0, {}):
+async def two_nodes_sim_and_wallets_services(blockchain_constants):
+    async for _ in setup_simulators_and_wallets_service(
+        2, 0, {"SOFT_FORK3_HEIGHT": blockchain_constants.SOFT_FORK3_HEIGHT}
```
arvidn, line 401: same comment here. doing it like this will run identical tests multiple times. e.g. `Mode.PLAIN` and `Mode.HARD_FORK2_0` will cause the exact same setup, and run any tests that depend on this twice with identical constants.

tests/core/full_node/stores/test_full_node_store.py
```diff
@@ -950,6 +952,8 @@ async def test_long_chain_slots(
         store = FullNodeStore(blockchain.constants)
         peak = None
         peak_full_block = None
+        if blockchain.constants.SOFT_FORK3_HEIGHT == 0:
```
arvidn, line 955: I think you should look at `consensus_mode == Mode.SOFT_FORK3` instead

arvidn (dismissed):

chia/consensus/block_header_validation.py
```diff
@@ -491,6 +496,28 @@ def validate_unfinished_header_block(
...
+            if passes_plot_filter(
+                prefix_bits,
+                plot_id,
+                curr_optional_block_record.pos_ss_cc_challenge_hash,
+                curr_optional_block_record.cc_sp_hash,
+            ):
+                return None, ValidationError(Err.INVALID_POSPACE)
+
+            if curr_optional_block_record.cc_sp_hash != curr_sp:
+                sp_count += 1
```
arvidn, line 517: maybe this is a nit, but this is where `sp_count` will be `>= constants.UNIQUE_PLOTS_WINDOW`. In which case it's time to end the loop. Yet, we still look up one more block record before we exit.

chia/consensus/block_header_validation.py
```diff
@@ -491,6 +496,28 @@ def validate_unfinished_header_block(
...
+                curr_sp = curr_optional_block_record.cc_sp_hash
+            curr_optional_block_record = blocks.try_block_record(curr_optional_block_record.prev_hash)
```
arvidn, line 519: is it safe to use `try_block_record()`? that will return `None` if we don't have the block in the cache. So, if the blocks aren't in the cache, we just won't enforce the CHIP-13 rules

chia/wallet/wallet_blockchain.py
```diff
@@ -226,3 +226,6 @@ async def clean_block_records(self) -> None:
 
         for header_hash in to_remove:
             del self._block_records[header_hash]
+
+    async def get_block_record_from_db(self, header_hash: bytes32) -> Optional[BlockRecord]:
```
arvidn, line 230: if you're adding this function, I think you want to add it to `BlockchainInterface` as well.

arvidn: oh, it seems it's already a part of `BlockchainInterface`. never mind

chia/simulator/block_tools.py
```diff
@@ -540,6 +542,29 @@ def get_farmer_wallet_tool(self) -> WalletTool:
     def get_pool_wallet_tool(self) -> WalletTool:
         return WalletTool(self.constants, self.pool_master_sk)
 
+    # Verifies if the given plot passed any of the previous `UNIQUE_PLOTS_WINDOW` plot filters.
+    def plot_id_passed_previous_filters(self, plot_id: bytes32, cc_sp_hash: bytes32, blocks: List[FullBlock]) -> bool:
+        curr_sp_hash = cc_sp_hash
+        sp_count = 1
+        for block in reversed(blocks):
```
wjblanke, line 549: is this cheap to do?

wjblanke (approved):
aok

chia/simulator/block_tools.py
fchirica: I _think_ it's a reverse iterator - so it won't do any reversing. also, the loop breaks really fast (when we hit 4 distinct SP hashes), so we won't iterate almost at all through blocks.

PR #15691, opened 2023-07-06
avoid DataLayer insert precheck

altendky:
### Purpose:

The expectation is that we will not trigger collisions usually so the except block won't be triggered and we'll save an sql query.

extracted from https://github.com/Chia-Network/chia-blockchain/pull/15613


almogdepaz:
https://github.com/Chia-Network/chia-blockchain/pull/15728#issuecomment-1637940907 same question

altendky:
https://github.com/Chia-Network/chia-blockchain/pull/15728#issuecomment-1638227287
> Yes, this change is intended to boost performance by reducing the number of queries. I can't say that a detailed analysis has been done but evidence so far is that we tend to insert new rows, not re-insert the exact same info, so cutting the query count in half (for this function) when doing that expected normal case is an improvement.
>
> The check is just in case we end up inserting the same keys (SQLite checked) with different values in other columns. It is a double check for errors elsewhere, but I wanted to get the optimization in place without working through whether the double check was critical or important or slightly useful or.... We think this does that.

almogdepaz (approved):

emlowe (approved):

PR #15692, opened 2023-07-06
CLI fixes for DAO wallets

geoffwalmsley:


chia/cmds/wallet_funcs.py
```diff
@@ -90,6 +90,7 @@ def get_mojo_per_unit(wallet_type: WalletType) -> int:
         WalletType.POOLING_WALLET,
         WalletType.DATA_LAYER,
         WalletType.VC,
+        WalletType.DAO,
     }:  # pragma: no cover
         mojo_per_unit = units["chia"]
     elif wallet_type == WalletType.CAT:
```
matt-o-how, line 96: Do we need to add DAO_CAT here too?

chia/cmds/dao.py
```diff
@@ -172,7 +181,10 @@ def dao_create_cmd(
 
     from .dao_funcs import create_dao_wallet
 
-    print("Creating new DAO")
+    if proposal_minimum % 2 == 0:
+        raise ValueError("Please use an odd mojo amount for proposal minimum amount")
+    else:
+        print("Creating new DAO")
```
matt-o-how, line 187: this doesn't need to be in an `else` - the raise will hit before the print

chia/cmds/wallet_funcs.py
geoffwalmsley: Only DAO type is needed since it's just converting XCH to mojo for treasury balances and spend proposals

PR #15693, opened 2023-07-06
multi threaded `chia plots check`

arvidn:
fix some log messages in harvester
running `chia plots check` in multiple threads. 5 by default but configurable via `harvester/parallel_decompressers_count`.
a failure in `check plots` is not fatal, but will log the error and keep going

### DIFF
--- a/chia/plotting/check_plots.py
+++ b/chia/plotting/check_plots.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import concurrent.futures
 import logging
 from collections import Counter
 from pathlib import Path
@@ -11,6 +12,7 @@ from chiapos import Verifier
 
 from chia.plotting.manager import PlotManager
 from chia.plotting.util import (
+    PlotInfo,
     PlotRefreshEvents,
     PlotRefreshResult,
     PlotsRefreshParameter,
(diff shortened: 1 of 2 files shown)


chia/plotting/check_plots.py
```diff
@@ -153,35 +164,52 @@ def check_plots(
...
+                    return
                 except SystemExit:
                     log.warning("System is shutting down.")
-                    return None
+                    return
                 except Exception as e:
                     log.error(f"{type(e)}: {e} error in getting challenge qualities for plot {plot_path}")
                     caught_exception = True
-                if caught_exception is True:
-                    break
```
arvidn, line 177: removing this means we'll continue checking more plots, after logging an error. note that we still treat it as a bad plot below, since `caught_exception` is `True`

fchirica (dismissed):

paninaro (dismissed):
lgtm

wjblanke (dismissed):
aok

chia/plotting/check_plots.py
arvidn: I will restore this. talking to Harold, the expected case where we're currently throwing should probably not throw in the final version

fchirica (approved):

wjblanke (approved):
aok

arvidn:
`check_plots.py` does not have any tests. we should add some at some point, but I don't think it's critical for the 2.0 release

PR #15694, opened 2023-07-06
Remove references to old offer mod

Quexington:
We previously deprecated the original offer mod in favor of a new one that checks create coin amounts.  However, we left in some forwards compatibility code so that new clients could still take old offers, they would just never make any new ones.  This PR deletes that code as it's been enough time to fully drop support for the old offers.

The only remaining evidence that the original existed is in the puzzle compression dictionary and in some temporary warnings to have a graceful error if someone attempts to take an old style offer.


geoffwalmsley (dismissed):

chia/rpc/wallet_rpc_api.py
```diff
@@ -1714,6 +1724,26 @@ async def get_offer_summary(self, request) -> EndpointResult:
 
     async def check_offer_validity(self, request) -> EndpointResult:
         offer_hex: str = request["offer"]
+
+        ###
```
trepca, line 1728: so we're double parsing offers just to provide the correct error message?

why not expand existing error message to `Invalid offer data or pre 1.7 offer format` or similar?

Quexington: What existing error message are you referring to?  I think if we don't explicitly check here, the offer will just get parsed as nothing offered, nothing requested.

trepca (approved):

PR #15704, opened 2023-07-07
Cleanup CLI RPC connection error handling 

jack60612:
### Purpose:
This reduces complexity by catching a `generator didn't yield error` for the `get_any_service_client` async generator which gives clients a connected instance of a client for whatever service they need. This is instead of passing none and every function having to use it. 

This PR is part of a series to cleanup the cli.
Todo:
- [x] PR: #15628 Merged
- [x] This branch rebased on top of main
- [x] Edit all wallet commands, `cmds_util.py` and `plotnft_funcs.py`

### Current Behavior:
Error caught In generator, function expected to exit if given `None` as a value for the rpc client.

### New Behavior:
Error still caught, but nothing instead of `None` is yielded, and the yield error is caught by a wrapper around the base click command group.

### Testing Notes:
Coverage Coming in next pr of series

### DIFF
--- a/chia/cmds/chia.py
+++ b/chia/cmds/chia.py
@@ -27,7 +27,7 @@ from chia.cmds.start import start_cmd
 from chia.cmds.stop import stop_cmd
 from chia.cmds.wallet import wallet_cmd
 from chia.util.default_root import DEFAULT_KEYS_ROOT_PATH, DEFAULT_ROOT_PATH
-from chia.util.errors import KeychainCurrentPassphraseIsInvalid
+from chia.util.errors import CliRpcConnectionError, KeychainCurrentPassphraseIsInvalid
 from chia.util.keychain import Keychain, set_keys_root_path
 from chia.util.ssl_check import check_ssl
 
@@ -132,7 +132,10 @@ cli.add_command(dev_cmd)
 
 
 def main() -> None:
-    cli()  # pylint: disable=no-value-for-parameter
+    try:
+        cli()  # pylint: disable=no-value-for-parameter
+    except CliRpcConnectionError:  # this happens when we cant connect to a client and is only comes from the cli code
+        exit(1)
 
 
 if __name__ == "__main__":
--- a/chia/cmds/cmds_util.py
+++ b/chia/cmds/cmds_util.py
@@ -100,10 +100,9 @@ async def get_any_service_client(
     try:
         # check if we can connect to node
         connected = await validate_client_connection(node_client, node_type, rpc_port, consume_errors)
-        if connected:
-            yield node_client, config
-        else:
+        if not connected:
             raise CliRpcConnectionError
+        yield node_client, config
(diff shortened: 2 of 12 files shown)


jack60612:
I also noticed this fixes a bug with `chia plotnft show` where nothing is outputted if it cant connect to a farmer.

chia/cmds/chia.py
```diff
@@ -132,7 +132,10 @@ def run_daemon_cmd(ctx: click.Context, wait_for_unlock: bool) -> None:
 
 
 def main() -> None:
-    cli()  # pylint: disable=no-value-for-parameter
+    try:
+        cli()  # pylint: disable=no-value-for-parameter
+    except CliRpcConnectionError:  # this happens when we cant connect to a client and is only comes from the cli code
+        exit(1)
```
arvidn, line 138: we don't have any error message here?

chia/cmds/cmds_util.py
```diff
@@ -103,9 +103,9 @@ async def get_any_service_client(
         if connected:
             yield node_client, config
         else:
-            yield None, config
+            raise CliRpcConnectionError
     except Exception as e:  # this is only here to make the errors more user-friendly.
-        if not consume_errors:
+        if not consume_errors or isinstance(e, CliRpcConnectionError):
```
arvidn, line 108: this is a bit hacky, but I can't think of an obvious improvement. I imagine `validate_client_connection()` itself can throw other exceptions too, that we want to print out. Otherwise the scope of the `try` block could just be made smaller

chia/cmds/cmds_util.py
```diff
@@ -212,13 +212,10 @@ async def get_wallet_client(
...
+        new_fp = await get_wallet(root_path, wallet_client, fingerprint)
+        if new_fp is not None:
+            yield wallet_client, new_fp, config
         else:
-            new_fp = await get_wallet(root_path, wallet_client, fingerprint)
-            if new_fp is None:
-                yield None, 0, config
-            else:
-                yield wallet_client, new_fp, config
+            raise CliRpcConnectionError  # this is caught by the main cli function
```
arvidn, line 221: generally I think it would be easier to follow to say:
```
if new_fp is None:
    raise CliRpcConnectionError # this is caught by the main cli function
yield wallet_client, new_fp, config
```

arvidn (dismissed):
lgtm

chia/cmds/chia.py
jack60612: error message is already printed out by the generator.

chia/cmds/cmds_util.py
jack60612: yeah we only really want to catch a timeout error but throw anything else

chia/cmds/cmds_util.py
jack60612: makes sense, ill change it.

arvidn (approved):

PR #15706, opened 2023-07-07
Fix max-line-lengh for pylint

aisk:
The current codes are using max line length for 120, and which is specified in black and flake8, but the pylint config is outdated.


emlowe (approved):

emlowe:
Close and reopen to run latest CI tests

PR #15708, opened 2023-07-07
forward `flags` in `get_puzzle_and_solution_for_coin()`

arvidn:
### Purpose:

This patch adds a new `flags`-parameter to `get_puzzle_and_solution_for_coin()` which is forwarded to the underlying call to run the CLVM. The main (future) use case for this is to be able to pass in `ALLOW_BACKREFS`, once we're ready to support that.

### Current Behavior:

`get_puzzle_and_solution_for_coin()` does not support any configuration of the underlying interpreter.

### New Behavior:

`get_puzzle_and_solution_for_coin()` supports configuration of the underlying interpreter, most notably `ALLOW_BACKREFS`.

### Testing Notes:

This function already has existing test coverage. Once we start making use of this flag we need coverage for those cases also.

### DIFF
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -59,7 +59,7 @@ def get_name_puzzle_conditions(
         return NPCResult(uint16(Err.GENERATOR_RUNTIME_ERROR.value), None, uint64(0))
 
 
-def get_puzzle_and_solution_for_coin(generator: BlockGenerator, coin: Coin) -> SpendInfo:
+def get_puzzle_and_solution_for_coin(generator: BlockGenerator, coin: Coin, flags: int) -> SpendInfo:
     try:
         args = bytearray(b"\xff")
         args += bytes(DESERIALIZE_MOD)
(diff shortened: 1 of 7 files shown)


chia/full_node/mempool_check_conditions.py
```diff
@@ -59,7 +59,7 @@ def get_name_puzzle_conditions(
         return NPCResult(uint16(Err.GENERATOR_RUNTIME_ERROR.value), None, uint64(0))
 
 
-def get_puzzle_and_solution_for_coin(generator: BlockGenerator, coin: Coin) -> SpendInfo:
+def get_puzzle_and_solution_for_coin(generator: BlockGenerator, coin: Coin, flags: int) -> SpendInfo:
```
arvidn, line 62: this is the main change. the rest are just fixing up the call sites

AmineKhaldi (approved):

arvidn:
```
Diff Coverage
-------------
chia/clvm/spend_sim.py (100%)
chia/full_node/mempool_check_conditions.py (100%)
chia/rpc/full_node_rpc_api.py (100%)
tests/core/mempool/test_mempool.py (100%)
tests/core/test_cost_calculation.py (66.7%): Missing lines 276
tests/generator/test_compression.py (100%)
-------------
Total:   9 lines
Missing: 1 line
Coverage: 88%
```

paninaro (approved):

PR #15715, opened 2023-07-07
Some NFT RPC mypy fixes

Quexington:
I was working in a different PR and these fixes were necessary when I changed some code in this area.  Mypy wasn't tracking it before but due to changes to the function, it started picking up errors around this.


ytx1991 (approved):
lgtm

wjblanke (approved):
seems ok to me

PR #15718, opened 2023-07-09
Refactor CLI by adding add new click types for fees, amounts, addresses and bytes32

### LINKED ISSUE #16802: [Bug] Combining coins with fee
### What happened?

The user wants to combine coins using

1. Run CLI combine
`chia wallet coins combine -f 1269832460 --number-of-coins 100 --id 25 --fee 0.0000001`

2. Response:
```
Combining 7 coins.
Would you like to Continue? (y/n): y
```

3. Response:
```
Exception from 'wallet' {'error': 'Transaction for 100000 is greater than spendable balance of 0. There may be other transactions pending or our minimum coin amount is too 
high.', 'success': False, 'traceback': 'Traceback (most recent call last):\n  File "chia\\rpc\\util.py", line 21, in inner\n  File "chia\\rpc\\wallet_rpc_api.py", line 1072, in send_transaction_multi\n  File "chia\\rpc\\util.py", line 82, in rpc_endpoint\n  File "chia\\rpc\\wallet_rpc_api.py", line 1616, in cat_spend\n  File "chia\\wallet\\cat_wallet\\cat_wallet.py", line 818, in generate_signed_transaction\n  File "chia\\wallet\\cat_wallet\\cat_wallet.py", line 717, in generate_unsigned_spendbundle\n  File "chia\\wallet\\cat_wallet\\cat_wallet.py", line 581, in create_tandem_xch_tx\n  File "chia\\wallet\\wallet.py", line 256, in select_coins\n  File "chia\\wallet\\coin_selection.py", line 57, in select_coins\nValueError: Transaction for 100000 is greater than spendable balance of 0. There may be other transactions pending or our minimum coin amount is too high.\n'}
```

However, standard wallet does have enough spendable balance:
```
Chia Wallet:
   -Total Balance:         5.345662453651 xch (5345662453651 mojo)
   -Pending Total Balance: 5.345662453651 xch (5345662453651 mojo)
   -Spendable:             2.311613686008 xch (2311613686008 mojo)
   -Type:                  STANDARD_WALLET
   -Wallet ID:             1
```

**Running the ``coins split`` command including same fee size works without problems**

### Version

2.1.1

Windows

CLI

### Relevant log output

_No response_

jack60612:
### Purpose:

- Adds new `TransactionFeeParamType` for automatic fee conversion 
- Adds new `CliAmount` class and `AmountParamType` for easy conversion between units.
- Adds new `CliAddress` class and `AddressParamType` for easy bech32m address conversion and wallet type tracking 
- Adds new `Bytes32ParamType` for automatic str to bytes32 conversion.
- Lastly this pr does more standardization too.

Merge requirements:

- [x] #15704 Needs to be Merged
- [x] PR needs to be rebased on main after merge
- [x] ~~Some light testing of the --mojos argument~~ (Moved to future PR)
- [x] Add Unit Tests
- [x] Cli tests added by #15886 Merged

### Current Behavior:
No Breaking Changes, just lots of backend changes.

### New Behavior:
N/A

### Testing Notes:
This PR has unit tests for the new types, and actual cli tests are coming in #15718


chia/cmds/param_types.py
```diff
@@ -0,0 +1,164 @@
...
+                self.fail("Fee must be decimal dotted value in XCH (e.g. 0.00005)", param, ctx)
+
+
+@dataclass(frozen=True)
+class CliAmount:
+    """
+    A dataclass for TX / wallet amounts for both XCH and CAT, and of course mojos.
+    """
+
+    mojos: bool
```
arvidn, line 49: you already have this bool encoded in the `Optional` members below. except there are two of them and both being `None` or both being set violates the invariant

arvidn (commented):
I appreciate this integration with Click. I think the new classes should have unit tests though.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,164 @@
...
+    mojo_amount: Optional[uint64] = None
+    decimal_amount: Optional[Decimal] = None
```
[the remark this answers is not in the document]
jack60612: let me clean it up

wallentx:
Failure. Coverage is below 100%.
-------------
Diff Coverage
-------------
chia/cmds/coin_funcs.py (0.0%): Missing lines 7,44-45,109,111,143-144,148,184,188-189,220,223
chia/cmds/coins.py (100%)
chia/cmds/data.py (100%)
chia/cmds/data_funcs.py (23.5%): Missing lines 17,24,35,45,55,64,74,83,93,104,113,153,174
chia/cmds/param_types.py (92.9%): Missing lines 30-33,78-81
chia/cmds/plotnft.py (85.7%): Missing lines 84
chia/cmds/plotnft_funcs.py (0.0%): Missing lines 19,29,36,327,344,355,358
chia/cmds/wallet.py (84.8%): Missing lines 585,612,665,1303-1304
chia/cmds/wallet_funcs.py (16.7%): Missing lines 374-379,381-383,652,970,1015,1043-1044,1129,1286,1288,1337,1344,1349
chia/rpc/wallet_rpc_client.py (100%)
tests/cmds/test_types.py (100%)
tests/pools/test_pool_cmdline.py (100%)
-------------
Total:   318 lines
Missing: 67 lines
Coverage: 78%
-------------

jack60612:
> ## Failure. Coverage is below 100%.
> ## Diff Coverage
> Diff: origin/main...HEAD, staged and unstaged changes
> ## chia/cmds/coin_funcs.py (0.0%): Missing lines 7,44-45,109,111,143-144,148,184,188-189,220,223
> chia/cmds/coins.py (100%)
> chia/cmds/data.py (100%)
> chia/cmds/data_funcs.py (23.5%): Missing lines 17,24,35,45,55,64,74,83,93,104,113,153,174
> chia/cmds/param_types.py (92.9%): Missing lines 30-33,78-81
> chia/cmds/plotnft.py (85.7%): Missing lines 84
> chia/cmds/plotnft_funcs.py (0.0%): Missing lines 19,29,36,327,344,355,358
> chia/cmds/wallet.py (84.8%): Missing lines 585,612,665,1303-1304
> chia/cmds/wallet_funcs.py (16.7%): Missing lines 374-379,381-383,652,970,1015,1043-1044,1129,1286,1288,1337,1344,1349
> chia/rpc/wallet_rpc_client.py (100%)
> tests/cmds/test_types.py (100%)
> tests/pools/test_pool_cmdline.py (100%)
> ## Total: 318 lines
> Missing: 67 lines
> Coverage: 78%

this only tests the new types class, and the missing lines in in param funcs are expected, as the mojos part isn't user available yet.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+            return bytes32.from_hexstr(value)
+        except ValueError:
+            self.fail("Value must be a valid bytes32 hex string like a coin id or puzzle hash", param, ctx)
+
+
+# These are what we use in click decorators
+TRANSACTION_FEE = TransactionFeeParamType()
+AMOUNT_TYPE = AmountParamType()
+ADDRESS_TYPE = AddressParamType()
+BYTES32_TYPE = Bytes32ParamType()
```
arvidn, line 166: is there a convention to use all-caps type aliases for these? (other types usually use CamelCase)

arvidn (commented):
looks good. Could you expand a bit on why you need the `isinstance` in the convert functions? it looks risky. If it's just for the default value, would it work yo change the type to the default value to `str` to look just like what would be passed on the command line?

chia/cmds/coin_funcs.py
```diff
@@ -201,13 +197,12 @@ async def async_split(
...
     wallet_id: int,
-    fee: Decimal,
+    fee: uint64,
     number_of_coins: int,
-    amount_per_coin: Decimal,
+    amount_per_coin: CliAmount,
     target_coin_id_str: str,
 ) -> None:
     async with get_wallet_client(wallet_rpc_port, fingerprint) as (wallet_client, fingerprint, config):
-        final_fee = uint64(int(fee * units["chia"]))
```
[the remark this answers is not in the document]
jack60612: we are not changing the unit

chia/cmds/param_types.py
jack60612: the click docs showed it using all caps, ill double check why

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    A Click parameter type for transaction fees, which can be specified in XCH or mojos.
+    """
+
+    name: str = "uint64"  # output type
+    value_limit: Decimal = Decimal(0.5)
+
+    def convert(self, value: Any, param: Optional[click.Parameter], ctx: Optional[click.Context]) -> uint64:
+        if isinstance(value, uint64):  # required by click
+            return value
```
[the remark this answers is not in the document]
jack60612: the click docs say you always have to do the isinstance check.

chia/cmds/param_types.py
jack60612: I think I could use the python_case for them too

chia/cmds/param_types.py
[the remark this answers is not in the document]
arvidn: it makes me a bit uneasy, but I suppose that's how it's meant to be used. Please double-check that you have test coverage of all cases where one may mix up the units for this

arvidn:
It would make sense to hold off landing this until we have test coverage, right?

```
-------------
chia/cmds/coin_funcs.py (0.0%): Missing lines 7,44-45,109,111,143-144,148,[18](https://github.com/Chia-Network/chia-blockchain/actions/runs/5556353350/job/15055427445?pr=15718#step:11:19)4,188-189,220,223
chia/cmds/coins.py (100%)
chia/cmds/data.py (100%)
chia/cmds/data_funcs.py (23.5%): Missing lines 17,24,35,45,55,64,74,83,93,104,113,153,174
chia/cmds/param_types.py (92.9%): Missing lines 30-33,78-81
chia/cmds/plotnft.py (85.7%): Missing lines 84
chia/cmds/plotnft_funcs.py (0.0%): Missing lines [19](https://github.com/Chia-Network/chia-blockchain/actions/runs/5556353350/job/15055427445?pr=15718#step:11:20),29,36,327,344,355,358
chia/cmds/wallet.py (84.8%): Missing lines 585,612,665,1303-1304
chia/cmds/wallet_funcs.py (16.7%): Missing lines 374-379,381-383,652,970,1015,1043-1044,1129,1286,1288,1337,1344,1349
chia/rpc/wallet_rpc_client.py (100%)
tests/cmds/test_types.py (100%)
tests/pools/test_pool_cmdline.py (100%)
```

jack60612:
i removed a bunch of `# pragma: no cover`s so the reported coverage diff is probably fine. Also for the new `param_types.py` file the missing coverage can be ignored as its for a future mojo units mode.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+                # now that we have the expected prefix, we can validate the address
+                if hrp == expected_prefix:
+                    addr_type = AddressType.XCH
+                else:
+                    self.fail(f"Unexpected Address Prefix: {hrp}, are you sure its for the right network?", param, ctx)
+            else:
+                addr_type = AddressType(hrp)
+            return CliAddress(puzzle_hash=decode_puzzle_hash(value), address_type=addr_type, original_address=value)
+        except ValueError:
+            self.fail("Address must be a valid bech32m address", param, ctx)
```
arvidn, line 143: don't you need to re-throw the exception here?

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+            try:
+                return CliAmount(mojos=True, amount=uint64(int(value)))
+            except ValueError:
+                self.fail("Amount must be positive integer value in mojos", param, ctx)
+        try:
+            if Decimal(value) <= 0:
+                self.fail("Amount must be greater than 0", param, ctx)
+            return CliAmount(mojos=False, amount=Decimal(value))
+        except InvalidOperation:
+            self.fail("Amount must be a decimal dotted value in XCH / CAT (e.g. 0.00005)", param, ctx)
```
arvidn, line 87: I would expect you to re-throw the exception here too

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+
+    name: str = "bytes32"  # output type
+
+    def convert(self, value: Any, param: Optional[click.Parameter], ctx: Optional[click.Context]) -> bytes32:
+        if isinstance(value, bytes32):  # required by click
+            return value
+        try:
+            return bytes32.from_hexstr(value)
+        except ValueError:
+            self.fail("Value must be a valid bytes32 hex string like a coin id or puzzle hash", param, ctx)
```
arvidn, line 159: and re-throw here too

chia/cmds/param_types.py
jack60612: `ParamType.fail` is what the docs said to use for an invalid value message: [https://click.palletsprojects.com/en/8.0.x/api/?highlight=type#click.ParamType.fail](https://click.palletsprojects.com/en/8.0.x/api/?highlight=type#click.ParamType.fail)

chia/cmds/param_types.py
jack60612: view response above

chia/cmds/param_types.py
jack60612: resp above

tests/cmds/wallet/test_notifications.py
```diff
@@ -87,7 +87,7 @@ async def get_notifications(
         "amount: 1000000000",
     ]
     run_cli_command_and_assert(capsys, root_dir, command_args, assert_list)
-    expected_calls: logType = {"get_notifications": [([get_bytes32(1)], (10, 10))]}
+    expected_calls: logType = {"get_notifications": [((get_bytes32(1),), (10, 10))]}
```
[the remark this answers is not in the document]
jack60612: Kyle suggested that i leave most cli input as a tuple (default), because almost every rpc simply uses an iterable.

arvidn (dismissed):

arvidn:
it might be good to have at least one more person review this as well.

emlowe (dismissed):
Requesting changes only because I'd like Kyle to review prior to merge

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+@dataclass(frozen=True)
+class CliAmount:
+    """
+    A dataclass for TX / wallet amounts for both XCH and CAT, and of course mojos.
+    """
+
+    mojos: bool
+    amount: Union[uint64, Decimal]  # uint64 if mojos, Decimal if not
```
altendky, line 52: Why do we need uncertainty here?  Can't we eliminate that at the point where we parse the user input?  At which point, perhaps this class goes away and we just have `uint64`.

chia/cmds/param_types.py
altendky: There is a general Python convention, that I don't prefer, to `ALL_CAPS` 'constants'.  So I guess that's where they got it from?  https://click.palletsprojects.com/en/8.1.x/parameters/#implementing-custom-types  I'm not sure why having this reused instance is a good thing to begin with though.  Seems more like an extra layer that, as Arvid noted, looks a bit weird.  I guess it avoids folks forgetting to `()`.

chia/cmds/param_types.py
altendky: Should we consider (here and below and wherever) to include the failure?  The user getting a message pointing out a length, or invalid character, etc would help them identify their error.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+        if mojos:
+            try:
+                return uint64(int(value))
+            except ValueError:
+                self.fail("Fee must be positive integer value in mojos", param, ctx)
+        try:
+            d_value = Decimal(value)
+            if 0 > d_value:
+                self.fail("Fee can not be negative", param, ctx)
+            if self.value_limit != Decimal(0) and d_value > self.value_limit:
```
altendky, line 38: I don't think we need a magic value of `0`.  If we want no limit, set `decimal.Decimal("Infinity")` and then it is pretty apparent when you read the limit.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+                self.fail(f"Fee must be in the range 0 to {self.value_limit}", param, ctx)
+            return uint64(int(d_value * units["chia"]))
```
altendky, line 40: Some reason this doesn't work?  We used to have hinting issues around this but they are (mostly) fixed.

```suggestion
            return uint64(d_value * units["chia"])
```

chia/cmds/param_types.py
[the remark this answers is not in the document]
altendky: I see the doc comments about this being the thing to do.  https://click.palletsprojects.com/en/8.1.x/api/#click.ParamType.convert  I'm not sure it is actually mandatory vs. being a practice to help deal with letting folks write defaults like `default=123` instead of `default="123"`.  I suspect that there are other corner cases where it matters as well.  Maybe when passing something like `123` to `CliRunner.invoke()` instead of `"123"`?

I agree with there being some risk here which makes me curious to know if it works, in our specific uses, without this type check.  Or perhaps replaced with a requirement that the input be a string.  Then we'll get notified if we hit some corner case and we can then consider how to handle it instead of it 'just working' possibly in some unexpected way.

It appears that perhaps since this comment there has been a change below to put off implementing support for mojos, maybe.  I think the general question of how we should handle this remains.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+from chia.util.ints import uint64
+from chia.wallet.util.address_type import AddressType
+
+
+class TransactionFeeParamType(click.ParamType):
+    """
+    A Click parameter type for transaction fees, which can be specified in XCH or mojos.
+    """
+
+    name: str = "uint64"  # output type
```
altendky, line 22: This might be a nice way to write this so that there's an actual code reference to the type.  Though, depending what the type is and what Click means when they say this is for documentation, other text may be preferred.

```suggestion
    name: str = uint64.__name__  # output type
```

altendky (changes requested):
Let's start the conversation with this and see how you two feel about my thoughts.  Once we settle this I can do a pass over the rest.

chia/cmds/param_types.py
jack60612: if we are in mojos mode we just take that, otherwise we don't know the conversion ratio so we cant convert it to uint64

chia/cmds/param_types.py
jack60612: not sure why this infinity value would apply, if its zero we just skip the check.

chia/cmds/param_types.py
jack60612: i think click intends it to just be "uint64"

jack60612:
> Should we consider (here and below and wherever) to include the failure? The user getting a message pointing out a length, or invalid character, etc would help them identify their error.
I don't think that's nessasary @altendky

jack60612:
> I see the doc comments about this being the thing to do. [click.palletsprojects.com/en/8.1.x/api/#click.ParamType.convert](https://click.palletsprojects.com/en/8.1.x/api/#click.ParamType.convert) I'm not sure it is actually mandatory vs. being a practice to help deal with letting folks write defaults like default=123 instead of default="123". I suspect that there are other corner cases where it matters as well. Maybe when passing something like 123 to CliRunner.invoke() instead of "123"?

> I agree with there being some risk here which makes me curious to know if it works, in our specific uses, without this type check. Or perhaps replaced with a requirement that the input be a string. Then we'll get notified if we hit some corner case and we can then consider how to handle it instead of it 'just working' possibly in some unexpected way.

> It appears that perhaps since this comment there has been a change below to put off implementing support for mojos, maybe. I think the general question of how we should handle this remains.
@altendky arguments from the cli are always strings, while the defaults are generally not.

chia/cmds/param_types.py
jack60612: So what should I do

chia/cmds/param_types.py
altendky: Yes, then you have to code the magic number 0 check.  If you used infinity to describe infinity, then the regular check would just do the right thing.

chia/cmds/param_types.py
altendky: Isn't that what `uint64.__name__` is?  Except in that case we are referencing a type which makes it more checkable and harder to have a typo like `unt64` or whatever.

altendky:
> > Should we consider (here and below and wherever) to include the failure? The user getting a message pointing out a length, or invalid character, etc would help them identify their error.
> 
> I don't think that's necessary (most of these errors wont give us detailed errors anyway) @altendky

Expressing wrong length, invalid character, checksum failure, etc can be quite valuable to the user.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+    value_limit: Decimal = Decimal(0.5)
+
+    def convert(self, value: Any, param: Optional[click.Parameter], ctx: Optional[click.Context]) -> uint64:
+        if isinstance(value, uint64):  # required by click
+            return value
+        mojos = False  # TODO: Add unit logic
+        if mojos:
+            try:
+                return uint64(int(value))
```
altendky, line 31: Check for any other redundant conversions.

```suggestion
                return uint64(value)
```

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+            except ValueError:
+                self.fail("Fee must be positive integer value in mojos", param, ctx)
+        try:
+            d_value = Decimal(value)
+            if 0 > d_value:
+                self.fail("Fee can not be negative", param, ctx)
+            if self.value_limit != Decimal(0) and d_value > self.value_limit:
+                self.fail(f"Fee must be in the range 0 to {self.value_limit}", param, ctx)
+            return uint64(d_value * units["chia"])
```
altendky, line 40: It looks like we don't have any check for excess decimal places.  We should check that here and for any other amounts.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,166 @@
...
+from chia.wallet.util.address_type import AddressType
+
+
+class TransactionFeeParamType(click.ParamType):
+    """
+    A Click parameter type for transaction fees, which can be specified in XCH or mojos.
+    """
+
+    name: str = "uint64"  # output type
+    value_limit: Decimal = Decimal(0.5)
```
altendky, line 23: We should have a habit of never feeding floats to `Decimal`.  Check for this elsewhere as well.

```pycon
>>> decimal.Decimal(0.1)
Decimal('0.1000000000000000055511151231257827021181583404541015625')
>>> decimal.Decimal("0.1")
Decimal('0.1')
```

```suggestion
    value_limit: Decimal = Decimal("0.5")
```

altendky:
> > I see the doc comments about this being the thing to do. [click.palletsprojects.com/en/8.1.x/api/#click.ParamType.convert](https://click.palletsprojects.com/en/8.1.x/api/#click.ParamType.convert) I'm not sure it is actually mandatory vs. being a practice to help deal with letting folks write defaults like default=123 instead of default="123". I suspect that there are other corner cases where it matters as well. Maybe when passing something like 123 to CliRunner.invoke() instead of "123"?
> 
> > I agree with there being some risk here which makes me curious to know if it works, in our specific uses, without this type check. Or perhaps replaced with a requirement that the input be a string. Then we'll get notified if we hit some corner case and we can then consider how to handle it instead of it 'just working' possibly in some unexpected way.
> 
> > It appears that perhaps since this comment there has been a change below to put off implementing support for mojos, maybe. I think the general question of how we should handle this remains.
> 
> @altendky arguments from the cli are always strings, while the defaults are generally not.

Been awhile, sorry, but I think we didn't actually leverage the type check and Arvid pointed out room for confusion with trivially different values.

chia/cmds/param_types.py
jack60612: removed extra conversions

chia/cmds/param_types.py
jack60612: checked and changed

tests/core/services/config.py
```diff
@@ -1,5 +1,3 @@
 from __future__ import annotations
 
 parallel = False
-install_timelord = False
-checkout_blocks_and_plots = False
```
arvidn: these are the default values, right? so removing them shouldn't have an effect iiuc.

arvidn (dismissed):

chia/cmds/param_types.py
jack60612: i got rid of the check

chia/cmds/param_types.py
jack60612: can you clarify what you mean?

altendky: `0.0000000000000000000000000000000000000001` isn't a valid XCH amount for chia.

tests/cmds/test_click_types.py
```diff
@@ -0,0 +1,138 @@
...
+    with pytest.raises(BadParameter):
+        TRANSACTION_FEE.convert("test", None, None)
+    with pytest.raises(BadParameter):
+        TRANSACTION_FEE.convert("0.6", None, None)
+    with pytest.raises(BadParameter):
+        TRANSACTION_FEE.convert("-0.6", None, None)
+
+
+def test_click_amount_type() -> None:
+    decimal_cli_amount = CliAmount(mojos=False, amount=Decimal(5.25))
```
altendky, line 55:
```suggestion
    decimal_cli_amount = CliAmount(mojos=False, amount=Decimal("5.25"))
```

tests/cmds/test_click_types.py
```diff
@@ -0,0 +1,138 @@
...
+    with pytest.raises(BadParameter):
+        AMOUNT_TYPE.convert("-999999", None, None)
+    with pytest.raises(BadParameter):
+        AMOUNT_TYPE.convert("-0.6", None, None)
+
+    # Test CliAmount Class
+    assert decimal_cli_amount.convert_amount(units["chia"]) == uint64(int(5.25 * units["chia"]))
+    assert mojos_cli_amount.convert_amount(units["chia"]) == uint64(100000)
+    with pytest.raises(ValueError):  # incorrect arg
+        CliAmount(mojos=True, amount=Decimal(5.25)).convert_amount(units["chia"])
```
altendky, line 81:
```suggestion
        CliAmount(mojos=True, amount=Decimal("5.25")).convert_amount(units["chia"])
```

chia/cmds/param_types.py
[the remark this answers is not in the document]
altendky: Should we try just failing on it not being a string instance and see what we learn?

chia/cmds/param_types.py
altendky: We can have a group discussion about this, but it seems this topic remains open.

altendky (changes requested):
I didn't repeat a detailed review yet.  Let's address the existing open issues then go back over the whole thing.

tests/cmds/test_click_types.py
jack60612: oh i didn't check the tests, my bad.

chia/cmds/param_types.py
jack60612: gotcha, thanks!

jack60612: added

chia/cmds/param_types.py
```diff
@@ -14,6 +14,48 @@
...
+    ctx: Optional[click.Context],
+) -> uint64:
+    try:
+        d_value = Decimal(value)
+    except InvalidOperation:
+        fail_func("Value must be a valid uint64 number", param, ctx)
+    if d_value.is_signed():
+        fail_func("Value must be a positive integer", param, ctx)
+    if d_value != d_value.to_integral():
+        fail_func("Value must be an integer", param, ctx)
```
arvidn, line 30: why do we care about `d_value` here? afaict, we only return `u_value`

arvidn (commented):
looks reasonable. except it looks like that block of code in `validate_uint64()` could be removed

chia/cmds/param_types.py
jack60612: for clearer error codes

chia/cmds/param_types.py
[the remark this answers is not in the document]
altendky: This looks unchanged.  It's been awhile since I had all the nuances here in my head.  Do we want to try to have a call about this with the three of us to refresh etc?

chia/cmds/param_types.py
altendky: This still looks untouched.  I don't necessarily have to get my way here but I think either a code change or another comment here is needed.

altendky (dismissed):
Looks like a couple still open conversations to finish with.

altendky (changes requested):
previous approval was meant to be a comment

chia/cmds/param_types.py
jack60612: removed as discussed

chia/cmds/param_types.py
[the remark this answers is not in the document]
jack60612: removed when appropriate as discussed

arvidn:
the cmd test on windows failed twice in a row (with what looks like a time out). I would expect all of these tests to be 100% deterministic, and not be flaky.

arvidn:
https://github.com/Chia-Network/chia-blockchain/actions/runs/7414705354/job/20221244941?pr=15718

altendky (changes requested):
alrighty, this covers the param types and the test file for them.  i'll try to get to the rest of the files, but expect them to be a lot less interesting.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,225 @@
...
+def validate_uint64(
+    value: str,
+    fail_func: Callable[[str, Optional[click.Parameter], Optional[click.Context]], None],
+    param: Optional[click.Parameter],
+    ctx: Optional[click.Context],
+) -> uint64:
+    try:
+        d_value = Decimal(value)
+    except InvalidOperation:
+        fail_func("Value must be a valid uint64 number", param, ctx)
```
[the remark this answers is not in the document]
jack60612: sounds good, changing now

chia/cmds/param_types.py
```diff
@@ -0,0 +1,225 @@
...
+
+def validate_decimal_xch(
+    value: str,
+    fail_func: Callable[[str, Optional[click.Parameter], Optional[click.Context]], None],
+    param: Optional[click.Parameter],
+    ctx: Optional[click.Context],
+) -> Decimal:
+    try:
+        d_value = Decimal(value)
+    except InvalidOperation:  # won't ever be value error because of the fee limit check
```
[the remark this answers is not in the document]
jack60612: good catch

chia/cmds/param_types.py
```diff
@@ -0,0 +1,225 @@
...
+        fail_func("Value must be decimal dotted value in XCH (e.g. 0.00005)", param, ctx)
+    if d_value.is_signed():
+        fail_func("Value can not be negative", param, ctx)
+    if not d_value.is_zero() and d_value < Decimal("0.000000000001"):
```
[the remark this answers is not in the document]
jack60612: maybe if i divide by the min value, if there is a remainder its wrong

chia/cmds/param_types.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+    return d_value
+
+
+class TransactionFeeParamType(click.ParamType):
+    """
+    A Click parameter type for transaction fees, which can be specified in XCH or mojos.
+    """
+
+    name: str = uint64.__name__  # output type
```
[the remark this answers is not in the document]
jack60612: this is used for the name in the cli, so TransactionFeeParamType is called unit64, what would you want to rename it to?

chia/cmds/param_types.py
```diff
@@ -0,0 +1,225 @@
...
+        #    return value
+        if not isinstance(value, str):
+            self.fail("Invalid Type, address must be string.", param, ctx)
+        try:
+            hrp, b32data = bech32_decode(value)
+            if hrp in ["xch", "txch"]:  # I hate having to load the config here
+                expected_prefix = ctx.obj.get("expected_prefix") if ctx else None
+                if expected_prefix is None:
+                    from chia.util.config import load_config
+                    from chia.util.default_root import DEFAULT_ROOT_PATH
```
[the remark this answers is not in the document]
jack60612: i just did that to reduce load time if it isnt needed but im not sure if that matters.

chia/cmds/param_types.py
```diff
@@ -0,0 +1,225 @@
...
+                    expected_prefix = selected_network_address_prefix(config)
+                    if ctx is not None:
+                        ctx.obj["expected_prefix"] = expected_prefix
+                # now that we have the expected prefix, we can validate the address
+                if hrp == expected_prefix:
+                    addr_type = AddressType.XCH
+                else:
+                    self.fail(f"Unexpected Address Prefix: {hrp}, are you sure its for the right network?", param, ctx)
+            else:
+                addr_type = AddressType(hrp)
```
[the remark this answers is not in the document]
jack60612: good point i think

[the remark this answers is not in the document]
jack60612: by options do you mean like override to allowing a txch addr?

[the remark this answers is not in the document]
jack60612: ok so i looked back at it and i feel like its more painful to make this part of the addresstype class, as then to keep the same pattern as the hrp function in address type i feel like i would have to cache the whole config making everything more complicated.

tests/cmds/test_click_types.py
```diff
@@ -0,0 +1,193 @@
...
+        AmountParamType().convert("test", None, None)
+    with pytest.raises(BadParameter):
+        AmountParamType().convert("0.0000000000001", None, None)  # 0.1 mojos
+    with pytest.raises(BadParameter):
+        AmountParamType().convert("-999999", None, None)
+    with pytest.raises(BadParameter):
+        AmountParamType().convert("-0.6", None, None)
+    # Test Type Failures
+    with pytest.raises(BadParameter):
+        AmountParamType().convert(float(0.01), None, None)
```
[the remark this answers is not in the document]
jack60612: lol i think i did that because i wanted it to be explicltly a float but it is anyway so who knows, updated

tests/cmds/test_click_types.py
```diff
@@ -0,0 +1,193 @@
...
+
+    # Test CliAmount Class
+    assert decimal_cli_amount.convert_amount(units["chia"]) == uint64(int(5.25 * units["chia"]))
```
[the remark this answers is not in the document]
jack60612: left int on outside because sometimes the linters care, but i added the decimal around the string.

chia/cmds/param_types.py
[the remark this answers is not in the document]
jack60612: chia\cmds\param_types.py:219: error: Cannot override instance variable (previously declared on base class "ParamType") with class variable  [misc]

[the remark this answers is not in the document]
jack60612: putting ignore for that message

jack60612:
might have to manually test plotnft cmds

chia/cmds/param_types.py
```diff
@@ -0,0 +1,229 @@
...
+import click
+
+from chia.cmds.units import units
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.bech32m import bech32_decode, decode_puzzle_hash
+from chia.util.config import selected_network_address_prefix
+from chia.util.ints import uint64
+from chia.wallet.util.address_type import AddressType
+
+one_decimal_mojo = Decimal("0.000000000001")
```
altendky, line 16: Just helps out with readability.
```suggestion
one_decimal_mojo = Decimal("0.000_000_000_001")
```

or...

```python-console
>>> Decimal("0.000_000_000_001") == Decimal("1e-12")
True
```

```suggestion
one_decimal_mojo = Decimal("1e-12")
```

chia/_tests/cmds/test_click_types.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def test_click_tx_fee_type() -> None:
+    # Test uint64 (only used as default)
+    # assert TransactionFeeParamType().convert(uint64(10000), None, None) == uint64(10000)
+
+    # TODO: Test MOJO Logic When Implemented
+
+    # Test Decimal / XCH
+    assert TransactionFeeParamType().convert("0.5", None, None) == uint64(int(0.5 * units["chia"]))
```
altendky, line 54:
```suggestion
    assert TransactionFeeParamType().convert("0.5", None, None) == uint64(Decimal("0.5") * units["chia"])
```

chia/cmds/param_types.py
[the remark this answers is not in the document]
altendky: Sounds good to me.  Makes sense but I also checked...

```python-console
>>> one_mojo = Decimal("0.000_000_000_001")
>>> remainder = Decimal("0.000_000_000_001") % one_mojo; print(remainder, remainder == 0)
0E-12 True
>>> remainder = Decimal("0.000_000_000_000_1") % one_mojo; print(remainder, remainder == 0)
1E-13 False
>>> remainder = Decimal("0.000_000_000_001_1") % one_mojo; print(remainder, remainder == 0)
1E-13 False
>>> remainder = Decimal("1.000_000_000_000_1") % one_mojo; print(remainder, remainder == 0)
1E-13 False
```

Not a problem, but the `0E-12` does raise an eyebrow or two.  `:]`  I guess that's decimal retaining significant figures perhaps?

chia/cmds/param_types.py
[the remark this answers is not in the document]
altendky: So I was bad and didn't provide 'evidence' for my statement.  Turns out I still think it's right, at least in the way we assign it, but it took me a bit to remember this isn't a dataclass so that's not indicating an instance attribute with a default...  Anyways, looking into the error you got I see that upstream indeed did not define this as a class var.  https://github.com/pallets/click/blame/923d197b56caa9ffea21edeef5baf1816585b099/src/click/types.py#L49  So, per them, it is indeed 'supposed' to be an instance var.  One option to avoid the ignore would be an `.__init__()` with a `super()` call to set the instance attribute.

Sorry I didn't look to the next level when I initially commented on this.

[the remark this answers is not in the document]
altendky: I think the likes of 'xch' and 'mojos' would be the user facing units?  Since we don't support mojos yet (there's the todo) I guess xch.  Or XCH.  Not sure which would be consistent with other ux.

chia/cmds/param_types.py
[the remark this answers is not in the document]
altendky: The intent is certainly valid and I think it would be good for us to improve that.  In practice, I think we already import every everything already and I would rather deal with this consistently with a plan than ad-hoc here and there sometimes.  For example, we could look at configuring a lazy import strategy that doesn't even involve changing each import site.

tests/cmds/test_click_types.py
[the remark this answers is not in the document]
altendky: mypy used to care due to incomplete hinting of our fixed width integers.  I have since improved them (though still incomplete iirc) so it should be ok now.  Unless you have had issues 'recently'?

chia/wallet/trading/offer.py
```diff
@@ -72,7 +72,7 @@ def from_condition_and_nonce(cls, condition: Program, nonce: bytes32) -> Notariz
         return cls(puzzle_hash, amount, memos, nonce)
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True, eq=False)
```
arvidn, line 75: all changes in this file appear unrelated. Are they a merge mistake?

jack60612: let me look back at commit history

chia/rpc/wallet_rpc_client.py
```diff
@@ -1262,7 +1262,7 @@ async def dl_verify_proof(self, request: DLProof) -> VerifyProofResponse:
     async def get_notifications(self, request: GetNotifications) -> GetNotificationsResponse:
         return GetNotificationsResponse.from_json_dict(await self.fetch("get_notifications", request.to_json_dict()))
 
-    async def delete_notifications(self, ids: Optional[List[bytes32]] = None) -> bool:
+    async def delete_notifications(self, ids: Optional[Sequence[bytes32]] = None) -> bool:
```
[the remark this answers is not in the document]
jack60612: i think it was due to a mypy error

chia/cmds/param_types.py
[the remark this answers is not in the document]
jack60612: lol thats funky, i have no clue.

chia/cmds/param_types.py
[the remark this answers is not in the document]
jack60612: ok so i should revert it?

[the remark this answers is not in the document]
jack60612: fair point, what should i put

chia/cmds/param_types.py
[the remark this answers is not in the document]
jack60612: ok so like move it, got it

tests/cmds/test_click_types.py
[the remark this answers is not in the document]
jack60612: oh ok, ill remove that.

chia/rpc/wallet_rpc_client.py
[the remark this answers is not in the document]
jack60612: i can leave it or do whatever with it, let me know

chia/wallet/trading/offer.py
jack60612: it was to fix a comparison bug between offers for the test: https://github.com/Chia-Network/chia-blockchain/commit/229934fe2ebf2f03462848b1d824dac9feb40fd6

chia/cmds/param_types.py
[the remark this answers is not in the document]
jack60612: reveerting change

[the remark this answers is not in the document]
jack60612: ill put xch or mojos

chia/cmds/param_types.py
```diff
@@ -0,0 +1,227 @@
...
+    except InvalidOperation:
+        fail_func("Value must be decimal dotted value in XCH (e.g. 0.00005)", param, ctx)
+    if d_value.is_signed():
+        fail_func("Value can not be negative", param, ctx)
+    if not d_value.is_zero() and d_value < one_decimal_mojo:
+        fail_func(
+            "Invalid amount of mojos, too many zeros. Either give zero or at least 1 mojo (0.000000000001 xch).",
+            param,
+            ctx,
+        )
```
altendky, line 58: I think the mod check covers what this does, plus all other fractional mojo cases.

```suggestion
```

chia/rpc/wallet_rpc_client.py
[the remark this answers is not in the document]
altendky: I'm guessing it is related to `multiple=True` options sometimes giving you lists and sometimes tuples.  I forget the details.  But, a quick look didn't make it obvious why it is needed right now.  Just a quick look though.

chia/wallet/trading/offer.py
arvidn: so we still have this problem in `main`? It would be great to land this as its own PR, ideally with a test demonstrating the problem

jack60612: i looked into it, its so that the cache isnt printed when comparing as sometimes the offer is just printed, and the order could be different in the cache so it would make the test fail

arvidn (approved):

altendky (approved):

Starttoaster:
Closing and re-opening to update CI

Starttoaster:
Arvid overrid coverage-diff earlier, and there have been no changes since then. Going to override it again since I re-ran CI, adding the label back, but it should mostly be taken as an extension of his override from a couple of days ago.

chia/cmds/param_types.py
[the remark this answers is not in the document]
altendky: i may revisit this later, let's move forward

PR #15725, opened 2023-07-10
server: Store `Handshake.software_version` in connections as `str`

xdustinface:
### Purpose:

Parsing it to `Version` here (introduced by #15330) enforces a certain client version format which leads to disconnecting from clients running other implementations which not follow the required format. While this currently makes sense for `protocol_version` we don't really depend on the client version for now and can just store it as it comes in to avoid those disconnects.

### DIFF
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -111,7 +111,7 @@ class WSChiaConnection:
     request_nonce: uint16 = uint16(0)
     peer_capabilities: List[Capability] = field(default_factory=list)
     # Used by the Chia Seeder.
-    version: Version = field(default_factory=lambda: Version("0"))
+    version: str = field(default_factory=str)
     protocol_version: Version = field(default_factory=lambda: Version("0"))
 
     log_rate_limit_last_time: Dict[ProtocolMessageTypes, float] = field(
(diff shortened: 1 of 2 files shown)


altendky:
Just to put it in the record here, I think there should be a discussion around this software version.  I think it would be good to consider prescribing a format, maybe semver, that allows for customized notes while still being parseable.  I think relatedly it might be good to have a client identifier to separate different forks and clients etc separate from the version.  It may be that we can't do much with this other than log it sometimes but I think even that has use when diagnosing issues as more clients come about.

arvidn:
this seems like a reasonable change to make while having that discussion though

arvidn (approved):

felixbrucker:
As this is related to https://github.com/Chia-Network/chia-blockchain/pull/15679 ill add that pep440 already allows custom modifiers for versions via post and local identifiers

Whether other software uses pep440 or a new/different standard is to be used is a different story

hoffmang9:
One pool is already using a non PEP440 version string for their custom node and that's how I noticed the problem here.

altendky:
I tend towards semver as a language neutral standard as opposed to the Python specific PEP440.  Though they are quite close.

PR #15728, opened 2023-07-10
avoid DataLayer insert ancestors precheck

altendky:
### Purpose:

The expectation is that we will not trigger collisions usually so the except block won't be triggered and we'll save an sql query.

extracted from https://github.com/Chia-Network/chia-blockchain/pull/15613


almogdepaz:
does this improve performance ?

before we used to read and write if we read none
now we always write and always read

chia/data_layer/data_store.py
```diff
@@ -289,26 +289,42 @@ async def _insert_ancestor_table(
...
                         values,
                     )
-                else:
+                except aiosqlite.IntegrityError as e:
+                    if not e.args[0].startswith("UNIQUE constraint"):
+                        # UNIQUE constraint failed: ancestors.hash, ancestors.tree_id, ancestors.generation
+                        raise
+
+                    async with writer.execute(
+                        """
```
almogdepaz, line 306: do we really need this ? looks like we already checked if we got the expected error, at this point are we validating sqlite is doing its work ?

altendky:
Yes, this change is intended to boost performance by reducing the number of queries.  I can't say that a detailed analysis has been done but evidence so far is that we tend to insert new rows, not re-insert the exact same info, so cutting the query count in half (for this function) when doing that expected normal case is an improvement.

The check is just in case we end up inserting the same keys (SQLite checked) with different values in other columns.  It is a double check for errors elsewhere, but I wanted to get the optimization in place without working through whether the double check was critical or important or slightly useful or....  We think this does that.

almogdepaz (dismissed):

emlowe (dismissed):

emlowe (approved):

PR #15730, opened 2023-07-10
Add DIDWallet.get_coin() to simplify DID wallet

Quexington:
Right now we use DIDWallet.select_coins when we just want to get the current DID coin.  The parameters to this function are almost always exactly `1`. This adds a method to just do that as a simplification that way any changes to the `select_coins` protocol doesn't balloon quite so large to other wallets that use the DID wallet.

I did decide to leave `select_coins` functional in some respect because it seems like that's the only way if you're using the RPC to get the current DID coin info.  This is probably not good, but is a more controversial and thoughtful change that should go in a separate PR.


ytx1991 (approved):
looks good

wjblanke (approved):
aok

PR #15735, opened 2023-07-11
Prevent redundant peer calls in coin_added

ytx1991:
### Purpose:
For most wallet types, we need to fetch the coin state and coin spend again in the wallet coin_added method. These are expensive calls that slow down the wallet syncing. The data we need are already fetched and extracted in the wallet state manager.

**This is the main PR of a series PRs. We need to modify each wallet's coin_added method to use the passed-in coin_data.** 

### Current Behavior:

Redundant coin spend fetch

### New Behavior:

Use the data in the wallet state manager

### Testing Notes:
Tested performance improvement of time spent since the _add_coin_states() to the end of coin_added() in wallets.
DID & untrusted peer
previous 165ms / after 64ms, 60% less time
DID & trusted peer
previous 158ms / after 85ms, 50% less time
NFT & Untrusted
previous  50ms (13ms for peer call)  /after 20-30ms, about 50% less time
NFT & Trusted (I don't know why trusted is much slower than untrusted)
previous 170ms (143ms for peer call) / after <20ms, about 85% less time


chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -166,7 +168,7 @@ def uncurry(cls: Type[_T_UncurriedNFT], mod: Program, curried_args: Program) ->
             return None
 
         return cls(
-            nft_mod_hash=nft_mod_hash,
+            nft_mod_hash=bytes32.from_bytes(bytes(nft_mod_hash)[1:]),
```
Quexington, line 171: You don't want to turn Program objects into bytes32 like this even though it will work.  The `a0` length prefix that you're stripping off here would be different for other, longer blobs.  You should instead do `bytes32(nft_mod_hash.atom)`.

chia/wallet/wallet_state_manager.py
```diff
@@ -641,68 +644,85 @@ async def unconfirmed_removals_for_wallet(self, wallet_id: int) -> Dict[bytes32,
...
 
+        puzzle = coin_spend.puzzle_reveal.to_program()
         uncurried = uncurry_puzzle(puzzle)
 
         # Check if the coin is a CAT
         cat_curried_args = match_cat_puzzle(uncurried)
         if cat_curried_args is not None:
+            cat_mod_hash, tail_program_hash, cat_inner_puzzle = cat_curried_args
+            cat_data: CATCoinData = CATCoinData(
+                bytes32(bytes(cat_mod_hash)[1:]), bytes32(bytes(tail_program_hash)[1:]), cat_inner_puzzle
```
Quexington, line 675: Similar to the comment above, you don't want to slice the bytes manually here.  Instead, do `<program>.atom`.

chia/wallet/wallet_state_manager.py
```diff
@@ -848,23 +868,21 @@ async def is_standard_wallet_tx(self, coin_state: CoinState) -> bool:
 
     async def handle_cat(
         self,
-        curried_args: Iterator[Program],
+        cat_data: CATCoinData,
```
Quexington, line 871: The names of these parameters might be better expressed as `parent_spend_data` or something like that, you don't want to give the impression that this is the current inner_puzzle etc.

chia/wallet/wallet_state_manager.py
```diff
@@ -641,68 +644,85 @@ async def unconfirmed_removals_for_wallet(self, wallet_id: int) -> Dict[bytes32,
 
     async def determine_coin_type(
         self, peer: WSChiaConnection, coin_state: CoinState, fork_height: Optional[uint32]
-    ) -> Optional[WalletIdentifier]:
+    ) -> Tuple[Optional[WalletIdentifier], Optional[VersionedBlob]]:
```
Quexington, line 647: Should we be using `VersionedBlob`?  It's not like this is going into a DB or anything, and you're forcing wallets to deserialize this when they get passed.  It might be better to define a `Protocol` or something for all of the `...CoinData` types and just return the object and have the wallet method assert that it's the right type of object.

ytx1991: These types of data are quite different and don't have any common method, so I'm not sure a Protocol is necessary. Actually, I'm planning to save them in the database since we added a new column in the coin_store for metadata. It can help us prevent creating more coin-level tables, display history,  and debugging. But obviously, it will increase the db size.

Quexington: Right, maybe not a `Protocol` just like a superclass for all of them.  Serializing it here, then immediately de serializing it in the `coin_added` methods seems wasteful, if you want to store them in the DB, there's no harm in serializing them before you do that if/when you decide to.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1976,7 +1976,13 @@ async def did_find_lost_did(self, request) -> EndpointResult:
         if curried_args is None:
             return {"success": False, "error": "The coin is not a DID."}
         p2_puzzle, recovery_list_hash, num_verification, singleton_struct, metadata = curried_args
-
+        did_data: DIDCoinData = DIDCoinData(
+            p2_puzzle,
+            bytes32.from_bytes(bytes(recovery_list_hash)[1:]),
```
Quexington, line 1981: This is a spot where you should use `.atom` instead of the slicing method.

chia/wallet/wallet_state_manager.py
```diff
@@ -903,29 +911,29 @@ async def handle_cat(
...
+
+        inner_puzzle_hash = parent_data.p2_puzzle.get_tree_hash()
         self.log.info(f"parent: {parent_coin_state.coin.name()} inner_puzzle_hash for parent is {inner_puzzle_hash}")
 
         hinted_coin = compute_spend_hints_and_additions(coin_spend)[coin_state.coin.name()]
         assert hinted_coin.hint is not None, f"hint missing for coin {hinted_coin.coin}"
         derivation_record = await self.puzzle_store.get_derivation_record_for_puzzle_hash(hinted_coin.hint)
 
-        launch_id: bytes32 = bytes32(bytes(singleton_struct.rest().first())[1:])
+        launch_id: bytes32 = bytes32(bytes(parent_data.singleton_struct.rest().first())[1:])
```
Quexington, line 936: Another spot to use `.atom`

chia/rpc/wallet_rpc_api.py
```diff
@@ -2134,7 +2140,12 @@ async def did_find_lost_did(self, request) -> EndpointResult:
                 coin_state.coin, uint32(coin_state.created_height), uint32(0), False, False, wallet_type, wallet_id
             )
             await self.service.wallet_state_manager.coin_store.add_coin_record(coin_record, coin_state.coin.name())
-            await did_wallet.coin_added(coin_state.coin, uint32(coin_state.created_height), peer)
+            await did_wallet.coin_added(
+                coin_state.coin,
+                uint32(coin_state.created_height),
+                peer,
+                VersionedBlob(DIDVersion.V1.value, bytes(did_data)),
```
Quexington, line 2147: Shouldn't you just pass in `did_data` here now?

Quexington (dismissed):

chia/wallet/wallet_state_manager.py
```diff
@@ -641,68 +644,79 @@ async def unconfirmed_removals_for_wallet(self, wallet_id: int) -> Dict[bytes32,
...
-                cat_curried_args,
-                parent_coin_state,
-                coin_state,
-                coin_spend,
-                peer,
-                fork_height,
+            cat_mod_hash, tail_program_hash, cat_inner_puzzle = cat_curried_args
+            cat_data: CATCoinData = CATCoinData(cat_mod_hash.atom, tail_program_hash.atom, cat_inner_puzzle)
+            return (
+                await self.handle_cat(
```
trepca, line 675: could we extract `handle_*` methods from WSM to the wallet?

chia/rpc/wallet_rpc_api.py
```diff
@@ -1976,7 +1976,13 @@ async def did_find_lost_did(self, request) -> EndpointResult:
         if curried_args is None:
             return {"success": False, "error": "The coin is not a DID."}
         p2_puzzle, recovery_list_hash, num_verification, singleton_struct, metadata = curried_args
-
+        did_data: DIDCoinData = DIDCoinData(
+            p2_puzzle,
+            recovery_list_hash.atom,
```
xdustinface, line 1981:
```suggestion
            bytes32(recovery_list_hash.atom),
```

Same as the others.

chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -166,7 +168,7 @@ def uncurry(cls: Type[_T_UncurriedNFT], mod: Program, curried_args: Program) ->
             return None
 
         return cls(
-            nft_mod_hash=nft_mod_hash,
+            nft_mod_hash=nft_mod_hash.atom,
```
xdustinface, line 171:
```suggestion
            nft_mod_hash=bytes32(nft_mod_hash.atom),
```

Same as the others.

xdustinface (changes requested):
There is some unused code, looks like leftover from earlier iterations of this PR which definitely should be dropped and i commented on some other things you might want to change.

altendky (dismissed):
just requesting a change to make sure my conversation is completed

chia/wallet/wallet_state_manager.py
ytx1991: Sure, I can do this in another PR.

chia/wallet/wallet_state_manager.py
```diff
@@ -641,68 +644,79 @@ async def unconfirmed_removals_for_wallet(self, wallet_id: int) -> Dict[bytes32,
...
-            return None
+            return None, None
 
         response: List[CoinState] = await self.wallet_node.get_coin_state(
             [coin_state.coin.parent_coin_info], peer=peer, fork_height=fork_height
         )
-        if len(response) == 0:
-            self.log.warning(f"Could not find a parent coin with ID: {coin_state.coin.parent_coin_info}")
-            return None
+        assert len(response) != 0, f"Could not find a parent coin with ID: {coin_state.coin.parent_coin_info.hex()}"
```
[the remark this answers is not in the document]
ytx1991: It blocked the coverage test and I cannot find a good way to cover this case.

chia/wallet/wallet_state_manager.py
```diff
@@ -641,68 +644,79 @@ async def unconfirmed_removals_for_wallet(self, wallet_id: int) -> Dict[bytes32,
...
         coin_spend = await fetch_coin_spend_for_coin_state(parent_coin_state, peer)
-        if coin_spend is None:
-            return None
-
-        puzzle = Program.from_bytes(bytes(coin_spend.puzzle_reveal))
+        assert (
+            coin_spend is not None
+        ), f"Could not find the coin spent  with ID: {coin_state.coin.parent_coin_info.hex()}"
 
+        puzzle = coin_spend.puzzle_reveal.to_program()
```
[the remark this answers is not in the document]
ytx1991: It's unrelated. Just simplify the code.

chia/wallet/wallet_protocol.py
```diff
@@ -28,7 +29,9 @@ def type(cls) -> WalletType:
     def id(self) -> uint32:
         ...
 
-    async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection) -> None:
+    async def coin_added(
+        self, coin: Coin, height: uint32, peer: WSChiaConnection, coin_data: Optional[Streamable]
```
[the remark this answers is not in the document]
ytx1991: Yes, the idea is to reuse the data we got in the WSM. I split PRs, so you will not see changes for specific wallets in this PR. So far, we know we can reuse it in Clawback/CAT/NFT/DID/VC. Some wallets may still need to fetch data on the fly in some cases because the coin_added is not called after the determine_coin_type().

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
ytx1991: For assert, it same, bypass coverage

chia/wallet/wallet_protocol.py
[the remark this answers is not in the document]
altendky: The `coin_data` parameter will be of a specific type for each class that implements `WalletProtocol`?  If so, then i think that making the protocol generic would fit.  I think that 'fake implementing' the protocol, sorta, would deserve a `# TODO:` at least at each site that still needs updated, or at least reviewed.  Rather than just an unused parameter which leaves it unclear to readers whether it is intentionally unused or not.

chia/wallet/wallet_state_manager.py
```diff
@@ -874,16 +886,12 @@ async def handle_cat(
...
             if cat_puzzle.get_tree_hash() != coin_state.coin.puzzle_hash:
                 return None
-            if bytes(tail_hash).hex()[2:] in self.default_cats or self.config.get(
-                "automatically_add_unknown_cats", False
-            ):
-                cat_wallet = await CATWallet.get_or_create_wallet_for_cat(
-                    self, self.main_wallet, bytes(tail_hash).hex()[2:]
-                )
+            if asset_id.hex() in self.default_cats or self.config.get("automatically_add_unknown_cats", False):
+                cat_wallet = await CATWallet.get_or_create_wallet_for_cat(self, self.main_wallet, asset_id.hex())
```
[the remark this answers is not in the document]
ytx1991: It doesn't change anything. Since bytes(tail_hash).hex() will return a0+actual hex, so we want to truncate the first 2 chars. But parent_data.tail_program_hash already truncated "a0" by using .atom, so we don't need [2:] any more.

chia/wallet/cat_wallet/cat_constants.py
```diff
@@ -30,3 +34,8 @@
     STABLY_USDS["asset_id"]: STABLY_USDS,
     CHIA_HOLIDAY_TOKEN["asset_id"]: CHIA_HOLIDAY_TOKEN,
 }
+
+
+class CATVersion(IntEnum):
+    V1 = uint16(1)
+    V2 = uint16(2)
```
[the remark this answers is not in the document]
ytx1991: removed

chia/wallet/did_wallet/did_info.py
```diff
@@ -24,3 +24,13 @@ class DIDInfo(Streamable):
     temp_pubkey: Optional[bytes]
     sent_recovery_transaction: bool
     metadata: str  # JSON of the user defined metadata
+
+
+@streamable
+@dataclass(frozen=True)
+class DIDCoinData(Streamable):
```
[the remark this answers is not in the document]
ytx1991: The next step is storing the coin data in the coin_store table metadata column. Therefore we can get coin data easily from everywhere without a peer call.

chia/wallet/wallet_state_manager.py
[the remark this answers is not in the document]
xdustinface: Hm, i don't think you should start to put in unrelated changes and get used to refactor code like this to bypass coverage.

chia/wallet/wallet_protocol.py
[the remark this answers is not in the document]
ytx1991: For generic, do you mean something like T = TypeVar("T") or Streamable is okay?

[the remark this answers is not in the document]
altendky: `Streamable` expresses nearly nothing of interest around the type you intend here.  https://docs.python.org/3.11/library/typing.html#user-defined-generic-types We have other `Generic` protocols as well.

chia/data_layer/data_layer_wallet.py
```diff
@@ -107,7 +107,8 @@ def from_json_dict(cls, json_dict: Dict[str, Any]) -> "Mirror":
 @final
 class DataLayerWallet:
     if TYPE_CHECKING:
-        _protocol_check: ClassVar[WalletProtocol] = cast("DataLayerWallet", None)
+        # TODO Create DataLayer coin data model if necessary
+        _protocol_check: ClassVar[WalletProtocol[Any]] = cast("DataLayerWallet", None)
```
altendky, line 111: `object` or `None` instead of `Any` might be better.  Either one would keep us from doing stuff with the parameter whereas `Any` lets us pretend it is anything we want without any hinting errors.

chia/data_layer/data_layer_wallet.py
```diff
@@ -814,7 +815,7 @@ async def delete_mirror(
     # SYNCING #
     ###########
 
-    async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection) -> None:
+    async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection, coin_data: Optional[Any]) -> None:
```
altendky, line 818: As above around the `Any`.

chia/pools/pool_wallet.py
```diff
@@ -966,7 +967,9 @@ async def get_pending_change_balance(self) -> uint64:
     async def get_max_send_amount(self, records: Optional[Set[WalletCoinRecord]] = None) -> uint128:
         return uint128(0)
 
-    async def coin_added(self, coin: Coin, height: uint32, peer: WSChiaConnection) -> None:
+    async def coin_added(
+        self, coin: Coin, height: uint32, peer: WSChiaConnection, coin_data: Optional[Streamable]
```
altendky, line 971: As above.  Also the protocol check at the beginning of the class needs a parameter for the generic like the others.

chia/wallet/wallet_protocol.py
```diff
@@ -19,16 +19,18 @@
 if TYPE_CHECKING:
     from chia.wallet.wallet_state_manager import WalletStateManager
 
+T = TypeVar("T", contravariant=True)
```
altendky, line 22: This is where you can require that the types are `Streamable`.

```suggestion
T = TypeVar("T", bound=Streamable, contravariant=True)
```

chia/wallet/wallet_state_manager.py
```diff
@@ -105,7 +108,7 @@
 from chia.wallet.wallet_transaction_store import WalletTransactionStore
 from chia.wallet.wallet_user_store import WalletUserStore
 
-TWalletType = TypeVar("TWalletType", bound=WalletProtocol)
+TWalletType = TypeVar("TWalletType", bound=WalletProtocol[Any])
```
altendky, line 111: Probably `object` here as well.

chia/wallet/wallet_state_manager.py
```diff
@@ -138,7 +141,7 @@ class WalletStateManager:
     db_wrapper: DBWrapper2
 
     main_wallet: Wallet
-    wallets: Dict[uint32, WalletProtocol]
+    wallets: Dict[uint32, WalletProtocol[Any]]
```
altendky, line 144: `object`

chia/wallet/wallet_state_manager.py
```diff
@@ -230,7 +233,7 @@ async def create(
             AssetType.CAT: CATWallet,
         }
 
-        wallet: Optional[WalletProtocol] = None
+        wallet: Optional[WalletProtocol[Any]] = None
```
altendky, line 236: `object`

chia/wallet/wallet_state_manager.py
```diff
@@ -641,68 +644,82 @@ async def unconfirmed_removals_for_wallet(self, wallet_id: int) -> Dict[bytes32,
 
     async def determine_coin_type(
         self, peer: WSChiaConnection, coin_state: CoinState, fork_height: Optional[uint32]
-    ) -> Optional[WalletIdentifier]:
+    ) -> Tuple[Optional[WalletIdentifier], Optional[Streamable]]:
```
altendky, line 647: Let's not introduce `Tuple` returns.  Make a dataclass so the elements have names.

chia/wallet/wallet_state_manager.py
```diff
@@ -1879,7 +1902,7 @@ async def get_coin_records_by_coin_ids(self, **kwargs: Any) -> List[CoinRecord]:
         result = await self.coin_store.get_coin_records(**kwargs)
         return [await self.get_coin_record_by_wallet_record(record) for record in result.records]
 
-    async def get_wallet_for_coin(self, coin_id: bytes32) -> Optional[WalletProtocol]:
+    async def get_wallet_for_coin(self, coin_id: bytes32) -> Optional[WalletProtocol[Any]]:
```
altendky, line 1905: `object`.  Also more below.

altendky (commented):
There are many different suffixes for the names of the classes.  `CoinData`, `Data`, `Metadata`, nothing...  I think some consistency would be good.

Some of the cases where I am suggesting `object` may cause issues.  Let me know which ones do and we can look at them again.

wjblanke (dismissed):
aok

Quexington (dismissed):

xdustinface (dismissed):

emlowe (changes requested):
I'm not a fan of the precedent of changing code to a naked assert to avoid coverage checks. For many reasons. (1) I don't like using assert() at all. and (2) I'd rather acknowledge the edge cases aren't tested then hiding them.

xdustinface:
> I'm not a fan of the precedent of changing code to a naked assert to avoid coverage checks.

@emlowe Are you referring to https://github.com/Chia-Network/chia-blockchain/pull/15735#discussion_r1268348000 and https://github.com/Chia-Network/chia-blockchain/pull/15735#discussion_r1266803608 or are there others that were missed? Because for that two i just forgot to resolve the conversation after https://github.com/Chia-Network/chia-blockchain/pull/15735/commits/677fe998cf9e49ba4bcecbbee1dfa4a3a673949b. Did that now.

xdustinface (dismissed):

xdustinface (approved):

emlowe (approved):

PR #15746, opened 2023-07-11
Add Click CLI Unit Tests (Proof of Concept)

jack60612:
### Purpose:
Adds CLI tests & does bug fixes

### Current Behavior:
n/a

### New Behavior:
n/a

### Testing Notes:
Adds Unit tests and fixes issues found


tests/cmds/cmd_test_utils.py
```diff
@@ -0,0 +1,208 @@
...
+            wallet_client.fingerprint = fingerprint  # type: ignore
+            assert fingerprint is not None
+            yield wallet_client, fingerprint, config
+
+    # override the functions
+    chia.cmds.cmds_util.get_any_service_client = test_get_any_service_client
+    chia.cmds.wallet_funcs.get_wallet_client = test_get_wallet_client  # type: ignore[attr-defined]
+
+
+def run_cli_command(capsys: Any, chia_root: Path, command_list: List[str]) -> Tuple[bool, str]:
```
arvidn, line 185: is `capsys` really `Any`? If we truly don't know what the type is, I'm under the impression `object` is better

tests/cmds/test_show.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_chia_show(capsys: Any, get_test_cli_clients: Tuple[TestRpcClients, Path]) -> None:
+    test_rpc_clients, root_dir = get_test_cli_clients
+    full_block = RPC_CLIENT_TO_USE.full_block  # a generated full block
+    # set RPC Client
+    test_rpc_clients.full_node_rpc_client = RPC_CLIENT_TO_USE
+    # get output with all options
+    command_args = ["-s", "-f", "-bh 1", "-b 0000000000000000000000000000000000000000000000000000000000000002"]
+    success, output = run_cli_command(capsys, root_dir, BASE_LIST + command_args)
```
arvidn, line 67: I don't think it's helpful to make the reader go find `BASE_LIST` to understand this test

tests/cmds/test_show.py
```diff
@@ -0,0 +1,88 @@
...
+    full_block: FullBlock = field(init=False)
+
+    def __post_init__(self) -> None:
+        full_block: Optional[FullBlock] = None
+        for f_b in get_full_blocks(tx_block=True):
+            if f_b is not None and f_b.foliage_transaction_block is not None:
+                full_block = f_b
+                break
+        assert full_block is not None
+        self.full_block = full_block
```
arvidn, line 27: this is just generating a random FullBlock, right? I think this would be clearer as a free function

tests/cmds/test_show.py
```diff
@@ -0,0 +1,88 @@
...
+from tests.cmds.cmd_test_utils import TestFullNodeRpcClient, TestRpcClients, cli_assert_shortcut, run_cli_command
+from tests.cmds.testing_classes import height_hash
+from tests.util.test_full_block_utils import get_full_blocks
+
+BASE_LIST = ["show"]
+
+
+@dataclass
+class ShowFullNodeRpcClient(TestFullNodeRpcClient):
+    full_block: FullBlock = field(init=False)
```
arvidn, line 18: you can specify a factory function here, where you could generate the FullBlock

tests/cmds/test_show.py
```diff
@@ -0,0 +1,88 @@
...
+    assert success
+    tx_filter_hash: str = "Not a transaction block"
+    if full_block.foliage_transaction_block:
+        tx_filter_hash = full_block.foliage_transaction_block.filter_hash.hex()
+    # these are various things that should be in the output
+    assert_list = [
+        "Current Blockchain Status: Full Node Synced",
+        "Estimated network space: 25.647 EiB",
+        "Block fees: 500000000000 mojos",
+        "Fee rate:    3.077e+04 mojos per CLVM cost",
```
arvidn, line 77: this test is fragile. You generate a random FullBlock, tweaking the function generating that block, even in some small subtle way, will break this test. If many tests are like this, all tests would break.

It would be much better if this test was self-contained in that it specified the fields of the full block and then expected to get back those values here

tests/cmds/cmd_test_utils.py
jack60612: capsys is a defined in the _pytest module so i assumed it was not supposed to be imported, i can change it to that spooky type or object, what would you perfer?

tests/cmds/test_show.py
jack60612: its more for when you have 30 tests in one file, as a kind of shortcut, but i can get rid of it.

tests/cmds/cmd_test_utils.py
jack60612: changed to object

tests/cmds/test_show.py
arvidn: `BASE_LIST` is not shorter than `"show"` though

tests/cmds/cmd_test_utils.py
```diff
@@ -0,0 +1,213 @@
...
+
+    farmer_rpc_client: TestFarmerRpcClient = TestFarmerRpcClient()
+    wallet_rpc_client: TestWalletRpcClient = TestWalletRpcClient()
+    full_node_rpc_client: TestFullNodeRpcClient = TestFullNodeRpcClient()
+    data_layer_rpc_client: TestDataLayerRpcClient = TestDataLayerRpcClient()
+    simulator_full_node_rpc_client: TestSimulatorFullNodeRpcClient = TestSimulatorFullNodeRpcClient()
+
+    def get_client(self, client_type: Type[_T_RpcClient]) -> _T_RpcClient:
+        if client_type == FarmerRpcClient:
+            return cast(FarmerRpcClient, self.farmer_rpc_client)  # type: ignore[return-value]
```
arvidn, line 133: it looks like you've gone to some length to get the correct type here, with `_T_RpcClient`, why do you still need to suppress this type warning?

jack60612: because its returning a child class, its mypy being stupid.

tests/cmds/cmd_test_utils.py
```diff
@@ -0,0 +1,213 @@
...
+        yield test_rpc_client, config
+
+    @asynccontextmanager
+    async def test_get_wallet_client(
+        wallet_rpc_port: Optional[int] = None,
+        fingerprint: Optional[int] = None,
+        root_path: Path = default_root,
+    ) -> AsyncIterator[Tuple[WalletRpcClient, int, Dict[str, Any]]]:
+        async with test_get_any_service_client(WalletRpcClient, wallet_rpc_port, root_path) as (wallet_client, config):
+            wallet_client.fingerprint = fingerprint  # type: ignore
```
arvidn, line 181: is `wallet_client.fingerprint` not an `Optional[int]`? why do you need to ignore the type warning here?

jack60612: the fingerprint arg is only part of the fake class

tests/cmds/cmd_test_utils.py
arvidn: I see it's changed to `object` in one place, but not here

tests/cmds/cmd_test_utils.py
```diff
@@ -0,0 +1,213 @@
...
+            assert fingerprint is not None
+            yield wallet_client, fingerprint, config
+
+    # override the functions
```
arvidn, line 185: I think this might warrant a bit more detailed explanation of what's going on. Imagine this is the first time you're reading this test for the first time, it would be really helpful to get a big-picture understanding of what's being mocked and what's being tested

jack60612: i added a fair amount of comments, let me know if we need more.

arvidn (dismissed):

arvidn (approved):

PR #15758, opened 2023-07-13
Add RPC client and tests for cancel_offers endpoint

Quexington:
see title

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1846,7 +1846,7 @@ class WalletRpcApi:
                     continue
                 if trade.offer and trade.offer != b"":
                     offer = Offer.from_bytes(trade.offer)
-                    if key in offer.driver_dict:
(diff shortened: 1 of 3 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1846,7 +1846,7 @@ async def cancel_offers(self, request: Dict) -> EndpointResult:
                     continue
                 if trade.offer and trade.offer != b"":
                     offer = Offer.from_bytes(trade.offer)
-                    if key in offer.driver_dict:
+                    if key in offer.arbitrage():
```
Quexington, line 1849: This was broken prior to this PR.  The driver_dict only contains non-XCH assets so if you don't specify the cancel_all option and try to specify XCH as the one to cancel, it would not work (though it would say it did).  The arbitrage dictionary always contains all assets involved including XCH.

geoffwalmsley (approved):

trepca (approved):

PR #15759, opened 2023-07-13
Fix flaky CAT test

Quexington:
see title


trepca:
thanks, lint fix needed though

trepca (approved):

PR #15760, opened 2023-07-13
bump chia_rs dependency to 0.2.9

arvidn:
this is blocked on `chia_rs` 0.2.9 being released.

This just bumps the dependency and updates the `Spend` constructor calls in tests. In production the `Spend` objects are all created on the rust side.


setup.py
```diff
@@ -15,7 +15,7 @@
     "chiapos==2.0.0b3",  # proof of space
     "clvm==0.9.7",
     "clvm_tools==0.4.6",  # Currying, Program.to, other conveniences
-    "chia_rs==0.2.8",
+    "chia_rs==0.2.9",
```
arvidn, line 18: this is the main change, which causes the `Spend` class to be updated

tests/core/mempool/test_mempool_manager.py
```diff
@@ -159,6 +159,8 @@ def make_test_conds(
             Spend(
                 spend_id,
                 IDENTITY_PUZZLE_HASH,
+                IDENTITY_PUZZLE_HASH,
```
jack60612, line 162: does it matter that we use the same for both?

tests/generator/test_rom.py
```diff
@@ -142,6 +144,12 @@ def test_get_name_puzzle_conditions(self, softfork_height: int) -> None:
             birth_seconds=None,
             create_coin=[(bytes([0] * 31 + [1]), 500, None)],
             agg_sig_me=[],
+            agg_sig_parent=[],
```
jack60612, line 147: we should probably make all of these keyworded at some point.

jack60612 (approved):
a bit messy but looks good.

tests/core/mempool/test_mempool_manager.py
arvidn: no, it's just any value

tests/generator/test_rom.py
arvidn: every time I have to update this, I think to myself: "this should be factored out". Maybe the next time it happens it will be painful enough to actually fix it

wjblanke (approved):
aok

PR #15762, opened 2023-07-13
Fix Cli error when exiting on some builds of python.

jack60612:
### Purpose:
originally made sure the catch runs sys.exit not exit() which is not in some builds of python.
It  also changed the error to a click exception to remove the need for the handling at all.

### Current Behavior:
Changes to use click errors instead of std errors.

### New Behavior:
N/A

### Testing Notes:
Kinda hard to replicate, but this is a fix

### DIFF
--- a/chia/cmds/cmds_util.py
+++ b/chia/cmds/cmds_util.py
@@ -111,11 +111,8 @@ async def get_any_service_client(
         await node_client.await_closed()
 
 
-async def get_wallet(
-    root_path: Path, wallet_client: WalletRpcClient, fingerprint: Optional[int]
-) -> Tuple[Optional[int], str]:
-    selected_fingerprint: Optional[int] = None
-    error: str = "No Fingerprint Selected"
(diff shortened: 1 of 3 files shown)


chia/cmds/cmds_util.py
```diff
@@ -113,8 +111,11 @@ async def get_any_service_client(
         await node_client.await_closed()
 
 
-async def get_wallet(root_path: Path, wallet_client: WalletRpcClient, fingerprint: Optional[int]) -> Optional[int]:
+async def get_wallet(
+    root_path: Path, wallet_client: WalletRpcClient, fingerprint: Optional[int]
+) -> Tuple[Optional[int], str]:
```
arvidn, line 116: you return a string on errors?

arvidn (dismissed):

chia/cmds/cmds_util.py
jack60612: let me fix this

jack60612: > let me fix this

like its kinda a mess

arvidn (approved):

PR #15763, opened 2023-07-14
Updated `executable_for_service`

ChiaMineJP:
This PR is a partial PR cut from `cmj.plot2_0` branch.


arvidn:
why is this change necessary?
Is it OK to not have test coverage? (it looks like it would be easy to test)

ChiaMineJP:
This change is ignorable. Priority is low. I don't know whether these kind of small code change requires dedicated PR but will add test coverage after high priority PRs are done.

ChiaMineJP:
This PR only tries to reduce 3 lines for the file.
So closing this PR...

PR #15764, opened 2023-07-14
Improved pool stats handling

ChiaMineJP:
This PR is a partial PR cut from `cmj.plot2_0` branch.

### DIFF
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -6,7 +6,7 @@ import logging
 import time
 import traceback
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
 import aiohttp
 from blspy import AugSchemeMPL, G1Element, G2Element, PrivateKey
@@ -70,6 +70,32 @@ def strip_old_entries(pairs: List[Tuple[float, Any]], before: float) -> List[Tup
             if index > 0:
                 return pairs[index:]
     return []
+
+
+def increment_pool_stats(
+    pool_states: Dict[bytes32, Any],
+    p2_singleton_puzzlehash: bytes32,
+    name: str,
+    count: int = 1,
+    value: Optional[Union[int, Dict[str, Any]]] = None,
+) -> None:
+    if p2_singleton_puzzlehash not in pool_states:
+        return
+    pool_state = pool_states[p2_singleton_puzzlehash]
+    if f"{name}_since_start" in pool_state:
+        pool_state[f"{name}_since_start"] += count
+    if f"{name}_24h" in pool_state:
+        if value is None:
+            pool_state[f"{name}_24h"].append((uint32(time.time()), pool_state["current_difficulty"]))
+        else:
+            pool_state[f"{name}_24h"].append((uint32(time.time()), value))
+
+        # Age out old 24h information for every signage point regardless
+        # of any failures.  Note that this still lets old data remain if
+        # the client isn't receiving signage points.
+        cutoff_24h = time.time() - (24 * 60 * 60)
+        pool_state[f"{name}_24h"] = strip_old_entries(pairs=pool_state[f"{name}_24h"], before=cutoff_24h)
+    return
 
 
 """
@@ -267,7 +293,8 @@ class Farmer:
 
     def handle_failed_pool_response(self, p2_singleton_puzzle_hash: bytes32, error_message: str) -> None:
         self.log.error(error_message)
-        self.increment_pool_stats(
+        increment_pool_stats(
+            self.pool_state,
             p2_singleton_puzzle_hash,
             "pool_errors",
             value=ErrorResponse(uint16(PoolErrorCode.REQUEST_FAILED.value), error_message).to_json_dict(),
@@ -337,8 +364,11 @@ class Farmer:
                         log_level = logging.INFO
                         if "error_code" in response:
                             log_level = logging.WARNING
-                            self.increment_pool_stats(
-                                pool_config.p2_singleton_puzzle_hash, "pool_errors", value=response
+                            increment_pool_stats(
+                                self.pool_state,
+                                pool_config.p2_singleton_puzzle_hash,
+                                "pool_errors",
+                                value=response,
                             )
                         self.log.log(log_level, f"GET /farmer response: {response}")
                         return response
@@ -381,8 +411,11 @@ class Farmer:
                         log_level = logging.INFO
                         if "error_code" in response:
                             log_level = logging.WARNING
-                            self.increment_pool_stats(
-                                pool_config.p2_singleton_puzzle_hash, "pool_errors", value=response
+                            increment_pool_stats(
+                                self.pool_state,
+                                pool_config.p2_singleton_puzzle_hash,
+                                "pool_errors",
+                                value=response,
                             )
                         self.log.log(log_level, f"POST /farmer response: {response}")
                         return response
@@ -425,8 +458,11 @@ class Farmer:
                         log_level = logging.INFO
                         if "error_code" in response:
                             log_level = logging.WARNING
-                            self.increment_pool_stats(
-                                pool_config.p2_singleton_puzzle_hash, "pool_errors", value=response
+                            increment_pool_stats(
+                                self.pool_state,
+                                pool_config.p2_singleton_puzzle_hash,
+                                "pool_errors",
+                                value=response,
                             )
                         self.log.log(log_level, f"PUT /farmer response: {response}")
                     else:
@@ -701,32 +737,6 @@ class Farmer:
         if receiver is None:
             raise KeyError(f"Receiver missing for {node_id}")
         return receiver
-
-    def increment_pool_stats(
-        self,
-        p2_singleton_puzzle_hash: bytes32,
-        name: str,
-        count: int = 1,
-        value: Any = None,
-    ) -> None:
-        pool_state = self.pool_state[p2_singleton_puzzle_hash]
-        if pool_state is None:
-            return
-        if f"{name}_since_start" in pool_state:
-            pool_state[f"{name}_since_start"] += count
-        if f"{name}_24h" in pool_state:
-            if value is None:
-                pool_state[f"{name}_24h"].append((uint32(time.time()), pool_state["current_difficulty"]))
-            else:
-                pool_state[f"{name}_24h"].append((uint32(time.time()), value))
-
-            # Age out old 24h information for every signage point regardless
-            # of any failures.  Note that this still lets old data remain if
-            # the client isn't receiving signage points.
-            cutoff_24h = time.time() - (24 * 60 * 60)
-            self.pool_state[p2_singleton_puzzle_hash][f"{name}_24h"] = strip_old_entries(
-                pairs=pool_state[f"{name}_24h"], before=cutoff_24h
-            )
 
     async def _periodically_update_pool_state_task(self) -> None:
         time_slept = 0
--- a/chia/farmer/farmer.py
+++ b/chia/farmer/farmer.py
@@ -76,6 +76,7 @@ def increment_pool_stats(
     pool_states: Dict[bytes32, Any],
     p2_singleton_puzzlehash: bytes32,
     name: str,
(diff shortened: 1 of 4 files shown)


ChiaMineJP:
I moved `strip_old_entries` to `chia/farmer/farmer.py` in order to prevent circular dependency.

arvidn:
can you describe what this patch does (more specific than "improve"), and do you intend to add test coverage? If not, is there a good reason to skip test coverage of this code?

ChiaMineJP:
The primary change of this PR is adding pool stats items like `valid_partials_since_start`, `valid_partials_24h`, etc.
There are several similar stat items and the mechanism to maintain 24h life time cycle for that items is almost the same.
So I added `increment_pool_stat` method.

chia/farmer/farmer.py
```diff
@@ -670,6 +702,32 @@ def get_receiver(self, node_id: bytes32) -> Receiver:
             raise KeyError(f"Receiver missing for {node_id}")
         return receiver
 
+    def increment_pool_stats(
+        self,
+        p2_singleton_puzzle_hash: bytes32,
+        name: str,
+        count: int = 1,
+        value: Any = None,
```
arvidn, line 710: it's a bit unfortunate that we use a dictionary instead of a class for `pool_state`, so we lose type information. but it looks like this `Any` could be `int`, `ErrorResponse` and `Dict[str, Any]`. I would think we want to make it `object`

chia/farmer/farmer.py
```diff
@@ -61,6 +61,17 @@
...
 
+
+def strip_old_entries(pairs: List[Tuple[float, Any]], before: float) -> List[Tuple[float, Any]]:
+    for index, [timestamp, points] in enumerate(pairs):
+        if timestamp >= before:
+            if index == 0:
+                return pairs
+            if index > 0:
+                return pairs[index:]
+    return []
```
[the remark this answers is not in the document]
ChiaMineJP: I'll add a test for this function

chia/farmer/farmer.py
```diff
@@ -670,6 +702,32 @@ def get_receiver(self, node_id: bytes32) -> Receiver:
...
         return receiver
 
+    def increment_pool_stats(
+        self,
+        p2_singleton_puzzle_hash: bytes32,
+        name: str,
+        count: int = 1,
+        value: Any = None,
+    ) -> None:
+        pool_state = self.pool_state[p2_singleton_puzzle_hash]
```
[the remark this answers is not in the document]
ChiaMineJP: Made it as a free function

chia/farmer/farmer.py
```diff
@@ -670,6 +702,32 @@ def get_receiver(self, node_id: bytes32) -> Receiver:
...
+        if f"{name}_24h" in pool_state:
+            if value is None:
+                pool_state[f"{name}_24h"].append((uint32(time.time()), pool_state["current_difficulty"]))
+            else:
+                pool_state[f"{name}_24h"].append((uint32(time.time()), value))
+
+            # Age out old 24h information for every signage point regardless
+            # of any failures.  Note that this still lets old data remain if
+            # the client isn't receiving signage points.
+            cutoff_24h = time.time() - (24 * 60 * 60)
```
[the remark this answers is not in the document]
ChiaMineJP: Done

chia/farmer/farmer.py
```diff
@@ -670,6 +702,32 @@ def get_receiver(self, node_id: bytes32) -> Receiver:
...
+        value: Any = None,
+    ) -> None:
+        pool_state = self.pool_state[p2_singleton_puzzle_hash]
+        if pool_state is None:
+            return
+        if f"{name}_since_start" in pool_state:
+            pool_state[f"{name}_since_start"] += count
+        if f"{name}_24h" in pool_state:
+            if value is None:
+                pool_state[f"{name}_24h"].append((uint32(time.time()), pool_state["current_difficulty"]))
```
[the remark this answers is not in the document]
ChiaMineJP: It is significant for the GUI since `uint64` can be either `BigInt` or `number` in JSON.
`uint32` gurantees to be `number` and not `BigInt`.

chia/farmer/farmer.py
ChiaMineJP: Done

chia/farmer/farmer.py
```diff
@@ -61,6 +61,44 @@
...
+    for index, [timestamp, points] in enumerate(pairs):
+        if timestamp >= before:
+            if index == 0:
+                return pairs
+            if index > 0:
+                return pairs[index:]
+    return []
+
+
+def increment_pool_stats(
```
arvidn, line 75: do you intend to add a test for this now too?

ChiaMineJP: Done.

tests/farmer_harvester/test_farmer.py
```diff
@@ -0,0 +1,519 @@
...
+@pytest.mark.parametrize(
+    argnames="case",
+    argvalues=[
+        pytest.param(StripOldEntriesCase([], 0, []), id="no_params"),
+        pytest.param(StripOldEntriesCase([(1689491043.3493967, 1)], 1689491044, []), id="stripped"),
+        pytest.param(
+            StripOldEntriesCase([(1689491043.3493967, 1)], 1689491043, [(1689491043.3493967, 1)]),
+            id="not_stripped",
+        ),
+    ],
```
arvidn, line 224: I think you're making this a lot more complicated than it needs to be by introducing `StripOldEntriesCase`. you can just put tuples of values in `argvalues`

tests/farmer_harvester/test_farmer.py
```diff
@@ -0,0 +1,519 @@
...
+        )
+
+
+@pytest.mark.parametrize(
+    argnames="case",
+    argvalues=[
+        pytest.param(StripOldEntriesCase([], 0, []), id="no_params"),
+        pytest.param(StripOldEntriesCase([(1689491043.3493967, 1)], 1689491044, []), id="stripped"),
+        pytest.param(
+            StripOldEntriesCase([(1689491043.3493967, 1)], 1689491043, [(1689491043.3493967, 1)]),
```
arvidn, line 221: I would recommend using smaller values to demonstrate this behavior. say, single digit timestamps. it would be a lot easier to read and be confident that the behavior is correct.

tests/farmer_harvester/test_farmer.py
```diff
@@ -0,0 +1,519 @@
...
+        assert case.p2_singleton_puzzle_hash not in case.pool_states
+    else:
+        assert case.pool_states[case.p2_singleton_puzzle_hash] == case.expected_result
+
+
+@pytest.mark.parametrize(
+    argnames="case",
+    argvalues=[
+        pytest.param(
+            NewProofOfSpaceCase.create_verified_quality_case(
```
arvidn, line 324: I think you should create a helper function to construct `NewProofOfSpaceCase` where you can omit some values (in these cases, you omit most) and just leave in the ones relevant for the test. But, given that so many of these parameters stay the same, do you really need to parameterize the test by them?

tests/farmer_harvester/test_farmer.py
```diff
@@ -0,0 +1,519 @@
...
+            id="invalid_peer_response",
+        ),
+    ],
+)
+@pytest.mark.asyncio
+async def test_farmer_new_proof_of_space_for_pool_stats(
+    harvester_farmer_environment: HarvesterFarmerEnvironment,
+    case: NewProofOfSpaceCase,
+) -> None:
+    farmer_service, farmer_rpc_client, _, _, _ = harvester_farmer_environment
```
arvidn, line 533: this looks like a pretty fragile test, but it's hard to tell. This is such a large test that it's hard to read it and feel confident that it's correct.

Now that you have it, it probably makes sense to keep it and simplify it to make it helpful.

However, the new function you added, `increment_pool_stats()` is very simple function, and it can (most likely) have a very simple unit test. By unit test I mean *just* setting up the values to pass in to that function and test that it does the thing you would expect.

tests/farmer_harvester/test_farmer.py
```diff
@@ -0,0 +1,519 @@
...
+    pool_config: PoolWalletConfig
+    pool_difficulty: Optional[uint64]
+    authentication_token_timeout: Optional[uint8]
+    use_invalid_peer_response: bool
+    expected_pool_state: Dict[str, Any]
+    marks: Marks = ()
+
+    # This creates a test case whose proof of space passes plot filter and quality check
+    @classmethod
+    def create_verified_quality_case(
```
arvidn, line 139: if you make this a free function with a shorter name and give the parameters names (and ideally add the `*` parameter first) I think you could improve readability of the test quite a bit.

tests/farmer_harvester/test_farmer.py
```diff
@@ -0,0 +1,519 @@
...
+        cls,
+        difficulty: uint64,
+        sub_slot_iters: uint64,
+        pool_url: str,
+        pool_difficulty: Optional[uint64],
+        authentication_token_timeout: Optional[uint8],
+        use_invalid_peer_response: bool,
+        expected_pool_stats: Dict[str, Any],
+    ):
+        p2_singleton_puzzle_hash = bytes32.fromhex("302e05a1e6af431c22043ae2a9a8f71148c955c372697cb8ab348160976283df")
```
arvidn, line 150: There are a lot of magic numbers here. I think it would be good to document what they are and why you chose these numbers.

Also, it seems like it would be appropriate the move these constants from the test parameter to the test itself

tests/farmer_harvester/test_farmer.py
```diff
@@ -0,0 +1,519 @@
...
+                1,
+                None,
+            ),
+            id="p2_singleton_puzzle_hash_not_exist",
+        ),
+        pytest.param(
+            IncrementPoolStatsCase(
+                std_hash(b"11223344"),
+                "yyy",
+                1689491043.3493967,
```
arvidn, line 248: I think it would make it easier to read to use smaller values for timestamps

tests/farmer_harvester/test_farmer.py
```diff
@@ -0,0 +1,519 @@
...
+                1,
+                1,
+                None,
+            ),
+            id="p2_singleton_puzzle_hash_not_exist",
+        ),
+        pytest.param(
+            IncrementPoolStatsCase(
+                std_hash(b"11223344"),
```
arvidn, line 246: this hash is the same in every test case. it would probably make sense to move this into the test itself, and out of the test parameters

tests/farmer_harvester/test_farmer.py
```diff
@@ -0,0 +1,519 @@
...
+                "yyy",
+                1689491043.3493967,
+                1,
+                1,
```
arvidn, line 250: I think these test cases would be easier to understand if you would make a utility function that constructs `IncrementPoolStatsCase` with named parameters, possibly with default values. these `1`s are not so easy to know what they mean for example

emlowe (approved):

tests/farmer_harvester/test_farmer.py
emlowe: Resolving comments in test code - we should iterate and simplfy post release

tests/farmer_harvester/test_farmer.py
emlowe: Resolving comments in test code - we should iterate and simplfy post release

tests/farmer_harvester/test_farmer.py
emlowe: Resolving comments in test code - we should iterate and simplfy post release

tests/farmer_harvester/test_farmer.py
emlowe: Resolving comments in test code - we should iterate and simplfy post release

tests/farmer_harvester/test_farmer.py
emlowe: Resolving comments in test code - we should iterate and simplfy post release

emlowe:
Coverage exemption accepted - there are long standing gaps in testing pool protocol errors and in `FarmerRpcApi`. This might be a good target for code coverage tech debt

PR #15765, opened 2023-07-14
Farmer now returns missing signage points

ChiaMineJP:


chia/farmer/farmer.py
```diff
@@ -670,6 +674,36 @@ def get_receiver(self, node_id: bytes32) -> Receiver:
             raise KeyError(f"Receiver missing for {node_id}")
         return receiver
 
+    def check_missing_signage_points(
```
arvidn, line 677: is there no way to test this new function?

chia/farmer/farmer.py
```diff
@@ -670,6 +674,36 @@ def get_receiver(self, node_id: bytes32) -> Receiver:
...
+        actual_sp_interval_seconds = float(timestamp - prev_time)
+        if actual_sp_interval_seconds <= 0:
+            return None
+
+        expected_sp_interval_seconds = self.constants.SUB_SLOT_TIME_TARGET / self.constants.NUM_SPS_SUB_SLOT
+        allowance = 1.6  # Should be chosen from the range (1 <= allowance < 2)
+        if actual_sp_interval_seconds < expected_sp_interval_seconds * allowance:
+            return None
+
+        skipped_sps = uint32(floor(actual_sp_interval_seconds / expected_sp_interval_seconds))
```
arvidn, line 702: this looks like a suspicious way of counting the number of signage points that have been skipped. Is there no index of signage points you can subtract?

ChiaMineJP: I think index of signage points is less accurate to count missing sps.
If you observe sp#2 after 60 seconds from sp#1 is observed, you definitely missed sp#2.

chia/farmer/farmer.py
ChiaMineJP: I'll add test case.

chia/farmer/farmer.py
arvidn: that would be a great comment to have here.

emlowe (changes requested):
Needs test for `check_missing_signage_points`

emlowe (approved):

PR #15769, opened 2023-07-14
New aggregate signature conditions

AmineKhaldi:


chia/util/initial-config.yaml
```diff
@@ -62,6 +62,12 @@ network_overrides: &network_overrides
       EPOCH_BLOCKS: 768
       DIFFICULTY_STARTING: 30
     testnet10:
+      AGG_SIG_PARENT_ADDITIONAL_DATA: e63a48559e680467b17a5a8cee0b5c759cf71b6924cee3eee72a72104eb46757
+      AGG_SIG_PUZZLE_ADDITIONAL_DATA: fe5898612c20bd1f25dda962591d61192878eae6c9f7fbe542c3dda2af233446
+      AGG_SIG_AMOUNT_ADDITIONAL_DATA: 1a47f5a6ac95683193d07ee83a3d9de685a92f4aa10e674984d4d87b2a858ba2
+      AGG_SIG_PUZZLE_AMOUNT_ADDITIONAL_DATA: 8366e4e93c96ccdad697828c49313ef4b98d810e555d4b0c30d5b05e24e6be62
+      AGG_SIG_PARENT_AMOUNT_ADDITIONAL_DATA: b14ae650b70f97386389a23fc834f0d7360c564878f79cf41c0478d894ced040
+      AGG_SIG_PARENT_PUZZLE_ADDITIONAL_DATA: 158493ea272d6714de8bc3dc50d252a8648a1a874316d0e80a8ed2d5d825f8f6
```
arvidn, line 70: this seems reasonable to just keep as constants

chia/util/condition_tools.py
```diff
@@ -55,41 +52,206 @@ def parse_sexp_to_conditions(sexp: Program) -> List[ConditionWithArgs]:
...
-        if cwa.vars[1].endswith(additional_data):
+        if (
+            len(cwa.vars) != 2
+            or len(cwa.vars[0]) != 48
+            or len(cwa.vars[1]) > 1024
+            or cwa.vars[0] is None
+            or cwa.vars[1] is None
+        ):
             raise ConsensusError(Err.INVALID_CONDITION)
+        for disallowed in additional_data.values():
```
arvidn, line 132: we should make sure that there's a test for all of these strings

tests/clvm/test_puzzles.py
```diff
@@ -89,7 +89,7 @@ def do_test_spend(
     # make sure we can actually sign the solution
     signatures = []
     for coin_spend in spend_bundle.coin_spends:
-        signature = key_lookup.signature_for_solution(coin_spend, bytes([2] * 32))
+        signature = key_lookup.signature_for_solution(coin_spend, {})
```
arvidn, line 92: I don't understand why it's OK to pass in an empty dict as additional_data here

chia/util/condition_tools.py
```diff
@@ -55,41 +52,206 @@ def parse_sexp_to_conditions(sexp: Program) -> List[ConditionWithArgs]:
...
 
+    for cwa in conditions_dict.get(ConditionOpcode.AGG_SIG_PARENT, []):
+        if (
+            len(cwa.vars) != 2
+            or len(cwa.vars[0]) != 48
+            or len(cwa.vars[1]) > 1024
+            or cwa.vars[0] is None
+            or cwa.vars[1] is None
+        ):
+            raise ConsensusError(Err.INVALID_CONDITION)
```
arvidn, line 145: this check should probably be factored out to a function

chia/util/condition_tools.py
```diff
@@ -55,41 +52,206 @@ def parse_sexp_to_conditions(sexp: Program) -> List[ConditionWithArgs]:
...
-        if msg.endswith(additional_data):
-            raise ConsensusError(Err.INVALID_CONDITION)
+        for dissallowed in additional_data.values():
+            if msg.endswith(dissallowed):
+                raise ConsensusError(Err.INVALID_CONDITION)
 
     for spend in conditions.spends:
+        for pk, msg in spend.agg_sig_parent:
+            ret[0].append(bytes48(pk))
+            ret[1].append(msg + spend.parent_id + additional_data.get(ConditionOpcode.AGG_SIG_PARENT, b""))
```
arvidn, line 70:
```suggestion
            ret[1].append(msg + spend.parent_id + additional_data[ConditionOpcode.AGG_SIG_PARENT])
```

arvidn: the domain string missing here should be considered a fatal error

tests/util/test_condition_tools.py
arvidn, line 1: I think it would make sense to have tests to ensure `pkm_pairs` and `pkm_pairs_for_conditions_dict` behave identically as well, for all combinations of inputs

tests/util/test_condition_tools.py
```diff
@@ -188,19 +191,7 @@ def test_agg_sig_unsafe_restriction(self, opcode: ConditionOpcode) -> None:
...
-        [
-            ConditionOpcode.AGG_SIG_PARENT,
-            ConditionOpcode.AGG_SIG_PUZZLE,
-            ConditionOpcode.AGG_SIG_AMOUNT,
-            ConditionOpcode.AGG_SIG_PUZZLE_AMOUNT,
-            ConditionOpcode.AGG_SIG_PARENT_AMOUNT,
-            ConditionOpcode.AGG_SIG_PARENT_PUZZLE,
-            ConditionOpcode.AGG_SIG_ME,
-        ],
-    )
```
arvidn: don't you need to parameterize on all conditions still?

AmineKhaldi: There is a similarly named function with the parameters, this one just covers the invalid condition case for `pkm_pairs_for_conditions_dict()` and the only use for the opcodes was in constructing the additional data dictionary.

chia/util/condition_tools.py
```diff
@@ -55,41 +54,157 @@ def parse_sexp_to_conditions(sexp: Program) -> List[ConditionWithArgs]:
...
 
+    data = agg_sig_additional_data(additional_data)
+
     for pk, msg in conditions.agg_sig_unsafe:
         ret[0].append(bytes48(pk))
         ret[1].append(msg)
-        if msg.endswith(additional_data):
-            raise ConsensusError(Err.INVALID_CONDITION)
+        for dissallowed in data.values():
+            if msg.endswith(dissallowed):
```
arvidn, line 83:
```suggestion
        for disallowed in data.values():
            if msg.endswith(disallowed):
```

chia/util/condition_tools.py
```diff
@@ -55,41 +54,157 @@ def parse_sexp_to_conditions(sexp: Program) -> List[ConditionWithArgs]:
...
+    ret: Dict[ConditionOpcode, bytes] = {}
+    for code in [
+        ConditionOpcode.AGG_SIG_PARENT,
+        ConditionOpcode.AGG_SIG_PUZZLE,
+        ConditionOpcode.AGG_SIG_AMOUNT,
+        ConditionOpcode.AGG_SIG_PUZZLE_AMOUNT,
+        ConditionOpcode.AGG_SIG_PARENT_AMOUNT,
+        ConditionOpcode.AGG_SIG_PARENT_PUZZLE,
+    ]:
+        ret[code] = std_hash(agg_sig_data + code)
```
arvidn, line 68: does this need to use `code.value`? I'm unsure if this ends up being the byte representation of the condition code

AmineKhaldi: I hashed with both at that time to compare, so with them ending up with the same hash I picked the more readable version (without `.value`).

chia/util/condition_tools.py
AmineKhaldi: Thanks, done.

arvidn:
lgtm

arvidn:
```
Diff Coverage
-------------
chia/simulator/block_tools.py (50.0%): Missing lines 1796
chia/simulator/wallet_tools.py (33.3%): Missing lines 194-196,199-201,204-206,209,215-216,219,225-226,229,235-236
chia/types/coin_spend.py (100%)
chia/types/condition_opcodes.py (100%)
chia/util/condition_tools.py (98.4%): Missing lines 130
chia/wallet/sign_coin_spends.py (100%)
chia/wallet/util/debug_spend_bundle.py (100%)
tests/core/mempool/test_mempool.py (100%)
tests/util/key_tool.py (100%)
tests/util/test_condition_tools.py (100%)
```

chia/simulator/block_tools.py
```diff
@@ -1785,6 +1785,15 @@ def compute_cost_test(
...
                     condition_cost += arg * 10000
+                elif hard_fork and condition in [
+                    ConditionOpcode.AGG_SIG_PARENT,
+                    ConditionOpcode.AGG_SIG_PUZZLE,
+                    ConditionOpcode.AGG_SIG_AMOUNT,
+                    ConditionOpcode.AGG_SIG_PUZZLE_AMOUNT,
+                    ConditionOpcode.AGG_SIG_PARENT_AMOUNT,
+                    ConditionOpcode.AGG_SIG_PARENT_PUZZLE,
+                ]:
+                    condition_cost += ConditionCost.AGG_SIG.value
```
arvidn, line 1796: there's no test coverage of this yet

arvidn:
I think this is a good addition to this PR: https://github.com/AmineKhaldi/chia-blockchain/pull/3

arvidn (dismissed):
I think it would be good if @richardkiss or @bramcohen would take a look at this as well, as it's a sensitive patch

wallentx:
Holding on merge until one of the above mentioned approve

chia/util/condition_tools.py
```diff
@@ -55,41 +54,157 @@ def parse_sexp_to_conditions(sexp: Program) -> List[ConditionWithArgs]:
...
+            )
+        )
+
+    for cwa in conditions_dict.get(ConditionOpcode.AGG_SIG_PUZZLE_AMOUNT, []):
+        validate_cwa(cwa)
+        ret.append(
+            (
+                bytes48(cwa.vars[0]),
+                cwa.vars[1]
+                + coin.puzzle_hash
```
richardkiss, line 177: Would you consider moving all six of the new `AGG_SIG` types into a loop and using a function parameter to determine how to turn the `coin` into a blob with a `serialized_coin_data_for_aggsig_type` sort of function? It seems a bit curious to have so many loops that are so similar. Very vulnerable to typos.

richardkiss: something like

```python
COIN_TO_BLOB_F_LOOKUP: Callable[[Coin], bytes] = {
   ConditionOpcode.AGG_SIG_PARENT: lambda coin: coin.parent_info,
    ConditionOpcode.AGG_SIG_PUZZLE_AMOUNT: lambda coin: coin.puzzle_hash + int_to_bytes(coin.amount),
    # etc
}
```

chia/util/condition_tools.py
```diff
@@ -55,41 +54,157 @@ def parse_sexp_to_conditions(sexp: Program) -> List[ConditionWithArgs]:
...
+            ret[0].append(bytes48(pk))
+            ret[1].append(msg + spend.parent_id + data[ConditionOpcode.AGG_SIG_PARENT])
+
+        for pk, msg in spend.agg_sig_puzzle:
+            ret[0].append(bytes48(pk))
+            ret[1].append(msg + spend.puzzle_hash + data[ConditionOpcode.AGG_SIG_PUZZLE])
+
+        for pk, msg in spend.agg_sig_amount:
+            ret[0].append(bytes48(pk))
+            ret[1].append(msg + int_to_bytes(spend.coin_amount) + data[ConditionOpcode.AGG_SIG_AMOUNT])
```
richardkiss, line 97: If `spend` is a `chia_rs.Spend`, I don't believe it has a `.coin_amount`. Does this have test coverage?

chia/util/condition_tools.py
```diff
@@ -55,41 +54,157 @@ def parse_sexp_to_conditions(sexp: Program) -> List[ConditionWithArgs]:
...
         ret[1].append(msg)
-        if msg.endswith(additional_data):
-            raise ConsensusError(Err.INVALID_CONDITION)
+        for disallowed in data.values():
+            if msg.endswith(disallowed):
+                raise ConsensusError(Err.INVALID_CONDITION)
 
     for spend in conditions.spends:
+        for pk, msg in spend.agg_sig_parent:
+            ret[0].append(bytes48(pk))
```
richardkiss, line 88: This is another loop that could simply use `COIN_TO_BLOB_F_LOOKUP` (see below) and iterate over `ConditionOpcode`.

Essentially, each `ConditionOpcode` has a corresponding function that takes the message from the condition, the coin, and the network identifier, and it produces a final message. Maybe we even make a function like that and use it here and [below](https://github.com/Chia-Network/chia-blockchain/pull/15769/files#diff-c9e77e30ba4a90734cb2edfde146fe37f02fa71593bf2f0b62bf045b2d68a57bR178).

```
def final_message(opcode: ConditionOpcode, spend: Coin | Spend, network_identifier: bytes32) -> bytes:
    # etc.
```

It's a bit ugly because in some places you want to pass in a `Coin` and in others a `Spend`.

Oh wait; does `Spend` even have `coin_amount`?

chia/util/condition_tools.py
AmineKhaldi: Sure, I sent ac43b34 to that effect.

chia/util/condition_tools.py
AmineKhaldi: Yep. We added coin amount and parent id just recently so your recollection is correct.

chia/util/condition_tools.py
AmineKhaldi: I sent ac43b34 to that effect.

chia/util/condition_tools.py
```diff
@@ -86,35 +125,31 @@ def pkm_pairs(conditions: SpendBundleConditions, additional_data: bytes) -> Tupl
     for spend in conditions.spends:
```
richardkiss, line 125: Here you might consider more meta-programming vs boilerplate with something like:

```python
condition_spend_pairs = [(ConditionOpcode.AGG_SIG_PARENT, spend.agg_sig_parent), (ConditionOpcode.AGG_SIG_PUZZLE, spend.agg_sig_puzzle), ...]
for condition, spends in condition_spend_pairs:
    for pk, msg in spends:
        ret[0].append(bytes48(pk))
        ret[1].append(make_aggsig_final_message(condition, msg, spend, data))
```

It would really be better if we could go `spend[condition]` so we wouldn't need pairs, but whatevs.

The names are pretty uniform, so I suppose we could go

```python
condition_spend_pairs = [
    (getattr(ConditionOpcode, condition_name.upper()), getattr(spend, condition_name)
    for condition_name in "agg_sig_parent agg_sig_puzzle ...".split())
   
```

chia/util/condition_tools.py
```diff
@@ -71,6 +71,45 @@ def agg_sig_additional_data(agg_sig_data: bytes) -> Dict[ConditionOpcode, bytes]
     return ret
 
 
+def make_aggsig_final_message(
+    opcode: ConditionOpcode,
+    msg: bytes,
+    spend: Union[Coin, Spend],
```
richardkiss, line 77: At some point, we should give `Spend` a `.coin` method so we don't need this weird `Union` type here. But this is fine for now.

chia/util/condition_tools.py
```diff
@@ -71,6 +71,45 @@ def agg_sig_additional_data(agg_sig_data: bytes) -> Dict[ConditionOpcode, bytes]
...
+    agg_sig_additional_data: Dict[ConditionOpcode, bytes],
+) -> bytes:
+    if isinstance(spend, Coin):
+        parent_coin_info = bytes(spend.parent_coin_info)
+        puzzle_hash = bytes(spend.puzzle_hash)
+        amount = int_to_bytes(spend.amount)
+        coin_id = bytes(spend.name())
+    elif isinstance(spend, Spend):
```
richardkiss, line 85: You could simply build a `Coin` out of the `Spend` here and then your logic could always assume `Coin` type.

chia/util/condition_tools.py
```diff
@@ -71,6 +71,45 @@ def agg_sig_additional_data(agg_sig_data: bytes) -> Dict[ConditionOpcode, bytes]
...
+        parent_coin_info = spend.parent_id
+        puzzle_hash = spend.puzzle_hash
+        amount = int_to_bytes(spend.coin_amount)
+        coin_id = spend.coin_id
+    else:
+        raise ValueError(f"Expected Coin or Spend, got {type(spend)}")
+
+    if opcode == ConditionOpcode.AGG_SIG_PARENT:
```
richardkiss, line 93: Python 3.10 supports `match` but a python trick I like is to use a `dict` rather than a large if/else branch.

```
addendum = dict(
   ConditionOpcode.AGG_SIG_PARENT = lambda coin: coin.parent_info,
   ...
)[opcode](coin)
```

It's probably not as readable. Don't know offhand if it's faster or slower. But it's definitely more concise and symmetric.

richardkiss (dismissed):
I have a couple more annoying suggestions, but they're totally optional.

chia/util/condition_tools.py
AmineKhaldi: Thanks, done.

chia/util/condition_tools.py
AmineKhaldi: Thanks, done.

chia/util/condition_tools.py
AmineKhaldi: Thanks, done.

chia/types/coin_spend.py
```diff
@@ -50,7 +50,16 @@ def compute_additions_with_cost(
             raise ValidationError(Err.BLOCK_COST_EXCEEDS_MAX, "compute_additions() for CoinSpend")
         atoms = cond.as_iter()
         op = next(atoms).atom
-        if op in [ConditionOpcode.AGG_SIG_ME, ConditionOpcode.AGG_SIG_UNSAFE]:
+        if op in [
```
richardkiss, line 53: This doesn't seem to be gated on the hard fork taking effect. Maybe it doesn't matter since these costs are just supposed to be estimates anyway.

richardkiss (approved):

paninaro (approved):

PR #15771, opened 2023-07-14
Updated `UI_ACTUAL_SPACE_CONSTANT_FACTOR` to 0.78

ChiaMineJP:


chia/consensus/pos_quality.py
```diff
@@ -4,7 +4,7 @@
 
 # The actual space in bytes of a plot, is _expected_plot_size(k) * UI_ACTUAL_SPACE_CONSTANT_FACTO
 # This is not used in consensus, only for display purposes
```
arvidn, line 6: it would be great to extend this comment to explain what this constant means and why this number was chosen

arvidn: oh, this is the average number of matches that are pruned from plots, to make them smaller than the theoretical size

arvidn (approved):

PR #15772, opened 2023-07-14
Added `get_keys_for_plot` daemon server API

ChiaMineJP:

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -130,13 +130,15 @@ def _get_keys_by_fingerprints(fingerprints: Optional[List[uint32]]) -> Tuple[Lis
     if fingerprints is None:
         keys = all_keys
     else:
+        assert isinstance(fingerprints, list)
         keys_by_fingerprint = {key.fingerprint: key for key in all_keys}
         keys = []
         for fingerprint in fingerprints:
-            if fingerprint not in keys_by_fingerprint:
-                missing_fingerprints.add(fingerprint)
+            f = uint32(fingerprint)
+            if f not in keys_by_fingerprint:
+                missing_fingerprints.add(f)
             else:
-                keys.append(keys_by_fingerprint[fingerprint])
+                keys.append(keys_by_fingerprint[f])
     return keys, missing_fingerprints
 
 
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -18,6 +18,7 @@ from enum import Enum
 from pathlib import Path
 from typing import Any, AsyncIterator, Dict, List, Optional, Set, TextIO, Tuple
 
+from blspy import G1Element
 from typing_extensions import Protocol
 
 from chia import __version__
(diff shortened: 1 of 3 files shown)


chia/daemon/server.py
```diff
@@ -567,28 +592,15 @@ async def get_routes(self, websocket: WebSocketResponse, request: Dict[str, Any]
         return response
 
     async def get_wallet_addresses(self, websocket: WebSocketResponse, request: Dict[str, Any]) -> Dict[str, Any]:
-        all_keys = Keychain().get_keys(include_secrets=True)
         fingerprints = request.get("fingerprints", None)
```
arvidn, line 595: I suppose if the "fingerprints" value is not a `List[uin32]`, you get a poor error message. We assume that's the type in the call to `_get_keys_by_fingerprints()`

arvidn (commented):
looks reasonable. do you think it's realistic to have tests? I don't think we have any infrastructure for it

chia/daemon/server.py
ChiaMineJP: Added assertion.

tests/core/daemon/test_daemon.py
```diff
@@ -596,6 +609,74 @@ def get_keys_no_secrets(self, include_secrets):
...
+    ),
+)
+@pytest.mark.asyncio
+async def test_get_keys_for_plot_error(
+    mock_daemon_with_config_and_keys,
+    monkeypatch,
+    case: KeysForPlotCase,
+):
+    daemon = mock_daemon_with_config_and_keys
+    with pytest.raises(Exception):
```
arvidn, line 682: what's the error we expect here? maybe you can add a `match="..."` to that `raises()` call

tests/core/daemon/test_daemon.py
```diff
@@ -617,12 +617,16 @@ def get_keys_no_secrets(self, include_secrets):
...
+                    "pool_public_key": "b1b0d8e777e3b1e79f33bc8707c74c8e4907fd8c3da870d7"
+                    "81803f647b17f3cc6250f09362f25699eb0ee385c57a2e01",
                 },
                 test_key_data_2.fingerprint: {
-                    "farmer_public_key": "a96a194fe76043b4867fa154b1fcda6709257699a720cc10af68a8414d9bf08cee0945becf9b36a1c76d5b04d379bc7f",
-                    "pool_public_key": "837646f1378b1194aa1054b8869dcea5006cb173730ceeb90e9b9bb7f5d73e481775d20a1726fec3ef140c5b9cdbd4b3",
+                    "farmer_public_key": "a96a194fe76043b4867fa154b1fcda6709257699a720cc10"
+                    "af68a8414d9bf08cee0945becf9b36a1c76d5b04d379bc7f",
+                    "pool_public_key": "837646f1378b1194aa1054b8869dcea5006cb173730ceeb9"
+                    "0e9b9bb7f5d73e481775d20a1726fec3ef140c5b9cdbd4b3",
```
arvidn, line 629: it's not easy to tell that these numbers make sense and that the test is correct. can you add some comments (or compute these in some way that makes it clear), to explain to the reader of what these test cases exercise any where these numbers come from?

ChiaMineJP: I changed the code to derive keys from the original test private keys.

tests/core/daemon/test_daemon.py
ChiaMineJP: Done

emlowe:
This seems fine, but I don't like the name that much `get_keys_for_plot` would suggest to me that I provide a PLOT as a parameter to the call and get back the keys used for the plot.

However, I believe this call is returning the farmer and pool public keys for the given fingerprint. So it returns the public keys one would use for plotting (it is the same farmer and pool pk as `chia keys show`)

Some alternative names:
get_plotting_public_keys
get_plotting_pks
get_farmer_pool_public_keys
get_farmer_pool_pks
get_pks_for_plotting
get_keys_for_plotting

tests/core/daemon/test_daemon.py
emlowe: Resoving comment per code change

ChiaMineJP:
@emlowe I picked up `get_keys_for_plotting`.

emlowe (approved):

Issue #15752, opened 2023-07-13
[Bug] `get_farmed_amount` rpc api of wallet is extremely slow


felixbrucker:
### What happened?

https://github.com/Chia-Network/chia-blockchain/commit/1e976464a26fcb6ca620820be95475fcfe965815 introduced the usage of `get_timestamp_for_height` for every reward coin record, which is super slow

for about 600 coin records this takes just over 2 minutes

### Version

compression alpha 4.6

### What platform are you using?

Windows

### What ui mode are you using?

CLI

felixbrucker:
Fixed by https://github.com/Chia-Network/chia-blockchain/pull/15778

PR #15778, opened 2023-07-14
Added `blocks_won` and `last_time_farmed` to `get_farmed_amount` Wall…

ChiaMineJP:

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -2965,7 +2965,9 @@ class WalletRpcApi:
         pool_reward_amount = 0
         farmer_reward_amount = 0
         fee_amount = 0
+        blocks_won = 0
         last_height_farmed = 0
+        last_time_farmed = 0
         for record in tx_records:
             if record.wallet_id not in self.service.wallet_state_manager.wallets:
                 continue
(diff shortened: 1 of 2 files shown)


tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -388,10 +388,12 @@ async def test_get_farmed_amount(wallet_rpc_environment: WalletRpcTestEnvironmen
     result = await wallet_rpc_client.get_farmed_amount()
 
     expected_result = {
+        "blocks_won": 2,
         "farmed_amount": 4_000_000_000_000,
         "farmer_reward_amount": 500_000_000_000,
         "fee_amount": 0,
         "last_height_farmed": 2,
+        "last_time_farmed": result["last_time_farmed"],  # time cannot be predicted so skipping test here
```
arvidn, line 396: can it not be predicted because we use wall clock time in the test, and not simulated time?

chia/rpc/wallet_rpc_api.py
```diff
@@ -2978,13 +2980,17 @@ async def get_farmed_amount(self, request) -> EndpointResult:
             # .get_farming_rewards() above queries for only confirmed records.  This
             # could be hinted by making TransactionRecord generic but streamable can't
             # handle that presently.  Existing code would have raised an exception
-            # anyways if this were to fail and we already have an assert below.
+            # anyway if this were to fail and we already have an assert below.
             assert height is not None
+            timestamp = await self.service.get_timestamp_for_height(height)
```
arvidn, line 2985: you only need the timestamp on line 2993. it would be more efficient to only ask for it in that case

arvidn (commented):
looks good

felixbrucker:
Please see https://github.com/Chia-Network/chia-blockchain/issues/15752

arvidn (dismissed):

chia/rpc/wallet_rpc_api.py
```diff
@@ -2978,13 +2980,16 @@ async def get_farmed_amount(self, request) -> EndpointResult:
...
             if record.type == TransactionType.FEE_REWARD:
-                fee_amount += record.amount - calculate_base_farmer_reward(height)
-                farmer_reward_amount += calculate_base_farmer_reward(height)
+                base_farmer_reward = calculate_base_farmer_reward(height)
+                fee_amount += record.amount - base_farmer_reward
+                farmer_reward_amount += base_farmer_reward
+                blocks_won += 1
             if height > last_height_farmed:
                 last_height_farmed = height
+                last_time_farmed = await self.service.get_timestamp_for_height(height)
```
arvidn, line 2992: we could save even more time by moving this call out of the loop too, and just call it once in `last_height_farmed`

ChiaMineJP: Thanks, this is really important comment.
@felixbrucker I've fixed the issue adderssed by this comment.
Performance degradation should be resolved after this PR is merged.

arvidn (approved):

PR #15790, opened 2023-07-16
Announce conditions limit

arvidn:
add test for soft-fork3 limit on number of announcements. This limit applies both to coin and puzzle assertions and creations. It also applies to `ASSERT_CONCURRENT_SPEND` and `ASSERT_CONCURRENT_PUZZLE`.

This is testing the `LIMIT_ANNOUNCES` flag, introduced in chia_rs here: https://github.com/Chia-Network/chia_rs/pull/175
The limit was later updated from 1000 -> 1024 here: https://github.com/Chia-Network/chia_rs/pull/185

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -101,9 +101,10 @@ def get_keychain():
 class Mode(Enum):
     PLAIN = 0
     HARD_FORK_2_0 = 1
+    SOFT_FORK3 = 2
 
 
-@pytest.fixture(scope="session", params=[Mode.PLAIN, Mode.HARD_FORK_2_0])
+@pytest.fixture(scope="session", params=[Mode.PLAIN, Mode.HARD_FORK_2_0, Mode.SOFT_FORK3])
 def consensus_mode(request):
     return request.param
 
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -101,9 +101,10 @@ def get_keychain():
 class Mode(Enum):
```
trepca, line 101: a bit of a nit, but shouldn't this be called `ConsensusMode`?

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -53,7 +54,10 @@ async def get_nft_count(wallet: NFTWallet) -> int:
 
 class TestWalletSync:
     @pytest.mark.asyncio
-    async def test_request_block_headers(self, simulator_and_wallet, default_1000_blocks):
+    async def test_request_block_headers(self, simulator_and_wallet, default_1000_blocks, consensus_mode: Mode):
+        if consensus_mode != Mode.PLAIN:
```
trepca, line 58: this will still wait for all fixtures to be initialized which takes most time, no?
can you access mode in fixures and skip test that way?

tests/conftest.py
AmineKhaldi: That's indeed clearer, thanks. I sent #15794 if we're going with it.

arvidn: yeah, this interacts with Florin's chip-13 patch as well, so we might want to wait with renaming it until after his PR has landed too

tests/wallet/sync/test_wallet_sync.py
arvidn: ideally, it would be possible to use `@pytest.mark.skipif()` and have it depend on one of the fixtures (`consensus_mode`). However, this appears to be an open issue with pytest: https://github.com/pytest-dev/pytest/issues/7395

trepca (approved):

AmineKhaldi (approved):

PR #15793, opened 2023-07-17
remove softfork2 logic

arvidn:
Since soft-fork2 has activated on mainnet, we no longer need to preserve the special cases from before the soft-fork. This reduces the number of tests we run on CI and simplifies the consensus logic.

This change intends to make the consensus logic behave as if soft-fork2 activated at block 0. i.e. it was always active.

To ensure the mainnet chain did not violate any soft-fork2 rules prior to activation, I have successfully synced a node from scratch with this patch.

### DIFF
--- a/tests/blockchain/test_blockchain_transactions.py
+++ b/tests/blockchain/test_blockchain_transactions.py
@@ -890,6 +890,19 @@ class TestBlockchainTransactions:
             full_node_1.blockchain, invalid_new_blocks[-1], expected_error=Err.ASSERT_SECONDS_RELATIVE_FAILED
         )
 
+        # we compare the timestamp against the previous transaction block, so in
+        # order to progress the timestamp, we need to farm one more block
+        blocks.extend(
+            bt.get_consecutive_blocks(
+                1,
+                blocks,
+                farmer_reward_puzzle_hash=coinbase_puzzlehash,
+                guarantee_transaction_block=True,
+                time_per_block=301,
+            )
+        )
+        await full_node_api_1.full_node.add_block(blocks[-1])
+
         valid_new_blocks = bt.get_consecutive_blocks(
             1,
             blocks,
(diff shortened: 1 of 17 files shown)


tests/blockchain/test_blockchain_transactions.py
```diff
@@ -890,6 +890,19 @@ async def test_assert_seconds_relative(
             full_node_1.blockchain, invalid_new_blocks[-1], expected_error=Err.ASSERT_SECONDS_RELATIVE_FAILED
         )
 
+        # we compare the timestamp against the previous transaction block, so in
```
richardkiss, line 893: Is this really part of the patch? I don't immediately see the connection.

tests/blockchain/test_blockchain_transactions.py
```diff
@@ -951,6 +964,19 @@ async def test_assert_seconds_absolute(
             full_node_1.blockchain, invalid_new_blocks[-1], expected_error=Err.ASSERT_SECONDS_ABSOLUTE_FAILED
         )
 
+        # we compare the timestamp against the previous transaction block, so in
```
richardkiss, line 967: Ditto this

richardkiss (approved):
Looks good.

tests/blockchain/test_blockchain_transactions.py
arvidn: yes, this is in fact required for the new consensus rules. This test used to only run pre soft-fork2.
One change in soft-fork2 was that relative time conditions now compare against the timestamp of the previous transaction block (as opposed to its own timestamp). This made it aligned with the relative height conditions, which compared against the height of the previous transaction block.

This test is specifically trying to trigger a relative time condition failure by adjusting the timestamp of the block that also has the condition. The previous way this test did this doesn't work anymore because the *current* block's timestamp is no longer used, but the previous transaction block's.

This fix inserts one more block that allows use to bump the timestamp of the block that's used for the conditions

wjblanke (approved):
aok

PR #15804, opened 2023-07-18
Dedup offer cancellation logic

Quexington:
Right now, there's three functions for cancelling an offer, one for a single offer + insecure, one for a single offer + secure, and one for multiple offers secure/insecurely.  This PR collapses the three functions into the last one as well as gives the last one a bit of a cleanup.

Since two of these functions were next to each other, the diff is a little confusing and it's probably better if you just look at the function in trade manager just by itself (by clicking "View File").

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1856,10 +1856,7 @@ class WalletRpcApi:
         trade_id = bytes32.from_hexstr(request["trade_id"])
         fee: uint64 = uint64(request.get("fee", 0))
         async with self.service.wallet_state_manager.lock:
-            if secure:
-                await wsm.trade_manager.cancel_pending_offer_safely(bytes32(trade_id), fee=fee)
-            else:
-                await wsm.trade_manager.cancel_pending_offer(bytes32(trade_id))
+            await wsm.trade_manager.cancel_pending_offers([bytes32(trade_id)], fee=fee, secure=secure)
         return {}
 
     async def cancel_offers(self, request: Dict) -> EndpointResult:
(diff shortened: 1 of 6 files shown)


chia/wallet/trade_manager.py
```diff
@@ -212,114 +212,49 @@ async def get_trade_by_id(self, trade_id: bytes32) -> Optional[TradeRecord]:
...
-    async def cancel_pending_offer_safely(
-        self, trade_id: bytes32, fee: uint64 = uint64(0)
+    async def cancel_pending_offers(
+        self,
+        trades: List[bytes32],
+        fee: uint64 = uint64(0),
+        secure: bool = True,
+        trade_cache: Dict[bytes32, TradeRecord] = {},
     ) -> Optional[List[TransactionRecord]]:
         """This will create a transaction that includes coins that were offered"""
```
trepca, line 226: can you document what added params are here

chia/rpc/wallet_rpc_api.py
```diff
@@ -1881,7 +1878,7 @@ async def cancel_offers(self, request: Dict) -> EndpointResult:
         if asset_id is not None and asset_id != "xch":
             key = bytes32.from_hexstr(asset_id)
         while True:
-            records: List[TradeRecord] = []
+            records: Dict[bytes32, TradeRecord] = {}
```
trepca, line 1881: why not use a set?

Quexington: Because downstream in `cancel_pending_offers` we're going to want to look up a trade record by its ID and instead of making a redundant DB call, it has provided the option to pass in a dictionary for easy lookup.

trepca (approved):

PR #15812, opened 2023-07-19
Remove CAT1 UX guards

Quexington:
We originally put in some guards to our offer APIs to give users good errors if they are attempting to take offers with CAT1s included.  I think we have enough distance from the CAT2 upgrade that these legacy code blocks can be disappeared.


paninaro:
@Quexington are you ok with this getting merged after the release branch is created?

Quexington:
Yeah no rush on this

trepca (approved):

trepca:
lets wait for after release branch as per comments above to merge

PR #15821, opened 2023-07-20
Update macos build scripts lerna to 7.1.3 and nx to 16.3.2 

emlowe:
Update macos build scripts lerna to 7.1.3 and nx to 16.3.2 for macos 10.14 build support

### DIFF
--- a/.github/workflows/build-macos-installers.yml
+++ b/.github/workflows/build-macos-installers.yml
@@ -33,7 +33,7 @@ jobs:
       matrix:
         python-version: [3.9]
         os:
-          - runs-on: macos-11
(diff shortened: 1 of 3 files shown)


paninaro (dismissed):

wallentx (changes requested):
https://github.com/Chia-Network/chia-blockchain/blob/0d210a424ebfc5391a4487c11c28796177c7a95b/.github/workflows/build-macos-installers.yml#L36

This should be set back to `macos-10.14`

wallentx (approved):

paninaro (approved):

PR #15822, opened 2023-07-20
Move some puzzles out of chia.wallet.puzzles

Quexington:
This tries to tackle the massive chia/wallet/puzzles folder by moving chialisp files for specific purposes into more descriptive places in chia-blockchain.  The folder still remains a little large because it contains all of the general puzzles, but it's certainly more manageable and more likely to find what your looking for.


richardkiss:
It's possible that the two puzzles you moved to in `consensus/puzzles` are not used by anything anymore, since they are in `chia_rs` now.

(They might be used by the API provider stuff though.)

richardkiss (approved):
Looks good to me.

trepca (approved):

PR #15824, opened 2023-07-20
Remove some old chialisp files

Quexington:


richardkiss (approved):
I'm okay with this, and I'm always in favor of deleting things.

Note that the `calculate_synthetic_public_key` puzzle was used at some point but has likely been re-implemented in python. It's a bit tricky in python because of typecasts, and we may need to resurrect the code if we decide to rewrite drivers in clvm. But it's not exactly rocket science.

Quexington:
> I'm okay with this, and I'm always in favor of deleting things.
> 
> Note that the `calculate_synthetic_public_key` puzzle was used at some point but has likely been re-implemented in python. It's a bit tricky in python because of typecasts, and we may need to resurrect the code if we decide to rewrite drivers in clvm. But it's not exactly rocket science.

(A note on this is that the concept also lives on still in our standard puzzle so even if we somehow forgot how to do it, it's part of that puzzle)

trepca (approved):

PR #15831, opened 2023-07-21
Update testnet height for SOFT_FORK3 and fix Chip-13 logic

emlowe:
Update testnet height for SOFT_FORK3 and fix some logic issues in the Chip-13 implementation


fchirica (approved):

wjblanke (approved):
aok

PR #15846, opened 2023-07-24
Remove hack_populate'ing secret keys

Quexington:
This hack has been around for a while because while we derive standard secret keys, we sign using synthetic secret keys.  When we examine spends, we see synthetic pubkeys but don't have an easy way to get the corresponding secret keys without populating an in memory mapping.

I think perhaps the better fix here would be to actually store derived synthetic secret keys instead of the basic secret keys we store now.  However, due to this being a breaking change to the DB, we'd need a migration and it would be hard to test and potentially sketchy.  Also this would cause us problems if we ever wanted to sign with the basic keys for some reason.

Instead, I've come up with a sort of middle ground fix.  When we see a pubkey in a spend the logic becomes:
1) Check to see if we have any straight up private keys for this public key (this allows signing of AGG_SIG_UNSAFE conditions which was previously impossible)
2) If we don't, try to derive puzzle hashes using the pubkey and check to see if we have a private key associated with one of those puzzle hashes
3) Sign the spend if we have found a key, fail if not


chia/wallet/sign_coin_spends.py
```diff
@@ -49,10 +52,17 @@ async def sign_coin_spends(
...
+                for derive in potential_derivation_functions:
+                    if inspect.iscoroutinefunction(secret_key_for_puzzle_hash):
+                        secret_key = await secret_key_for_puzzle_hash(derive(pk))
+                    else:
+                        secret_key = secret_key_for_puzzle_hash(derive(pk))
+                    if secret_key is not None and secret_key.get_g1() == pk:
+                        break
+                else:
+                    e_msg = f"no secret key for {pk}"
+                    raise ValueError(e_msg)
```
trepca, line 65: since we don't use e_msg for anything else might as well put that string directly into ValueError

chia/wallet/wallet_state_manager.py
```diff
@@ -2055,3 +2078,13 @@ async def get_or_create_vc_wallet(self) -> VCWallet:
             vc_wallet = await VCWallet.create_new_vc_wallet(self, self.main_wallet)
 
         return vc_wallet
+
+    async def sign_transaction(self, coin_spends: List[CoinSpend]) -> SpendBundle:
```
trepca, line 2082: why not keep this in the wallet?

Quexington: Because it has nothing to do with the standard wallet, not really.  Every wallet uses it, and it's just generic "find the right private key and sign this spend".

trepca (approved):
ideally it'd be great to remove code from WSM, not add, but I don't really have a better idea either where to put that code

PR #15847, opened 2023-07-24
Reduce test runs for test_clear_pending_roots

emlowe:
As a workaround for test_clear_pending_roots flakes, reduce the number of test runs by only running the tests for the PLAIN consensus mode. The changes to consensus should have no impact on this test, and by reducing the number of times this test runs, it should help reduce the number of strange failures seen on Windows that seem entirely unrelated to the test itself.


arvidn (approved):

emlowe:
Closing PR

PR #15853, opened 2023-07-25
Fix clawback sender resync issue

ytx1991:
### Purpose:
This PR fixes two issues:
1. Clawback sender cannot claim their fund after the resync.
2. Add missing INCOMING_TX for the claim side and ensure the tx counts are the same after the resync.

### Current Behavior:
1. After resync, the interested_coins table will not recovery as expected. The Clawback sender will not receive the coin state of the clawback coin.
2. After the claim and before the resync, the wallet will not add an INCOMING_TX for the claimed fund. But after the resync, this tx will be added

### New Behavior:
1. For resync coins only, check if the coin is a clawback coin. If it is, then add it to the interested_coins.
2. If it is a clawback's child, add the tx no matter if there is a pending tx related to it.

### Testing Notes:
Unit test

### DIFF
--- a/chia/wallet/puzzles/clawback/drivers.py
+++ b/chia/wallet/puzzles/clawback/drivers.py
@@ -3,22 +3,33 @@ from __future__ import annotations
 import logging
 from typing import Any, List, Optional, Set, Union
 
+from blspy import G2Element
+from chia_rs.chia_rs import CoinState
+
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
+from chia.server.ws_connection import WSChiaConnection
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.program import Program
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.coin_spend import CoinSpend
 from chia.types.condition_opcodes import ConditionOpcode
+from chia.types.spend_bundle import SpendBundle
 from chia.util.condition_tools import conditions_for_solution
-from chia.util.ints import uint64
+from chia.util.ints import uint32, uint64
 from chia.util.misc import VersionedBlob
 from chia.wallet.puzzles.clawback.metadata import ClawbackMetadata
 from chia.wallet.puzzles.load_clvm import load_clvm_maybe_recompile
 from chia.wallet.puzzles.p2_delegated_puzzle_or_hidden_puzzle import MOD
+from chia.wallet.transaction_record import TransactionRecord
 from chia.wallet.uncurried_puzzle import UncurriedPuzzle
+from chia.wallet.util.compute_memos import compute_memos
 from chia.wallet.util.curry_and_treehash import calculate_hash_of_quoted_mod_hash, curry_and_treehash
 from chia.wallet.util.merkle_tree import MerkleTree
+from chia.wallet.util.transaction_type import TransactionType
+from chia.wallet.util.wallet_sync_utils import fetch_coin_spend_for_coin_state
 from chia.wallet.util.wallet_types import RemarkDataType
+from chia.wallet.wallet_puzzle_store import WalletPuzzleStore
+from chia.wallet.wallet_transaction_store import WalletTransactionStore
 
 P2_1_OF_N = load_clvm_maybe_recompile("p2_1_of_n.clsp")
 P2_CURRIED_PUZZLE_MOD = load_clvm_maybe_recompile("p2_puzzle_hash.clsp")
@@ -179,3 +190,38 @@ def generate_clawback_spend_bundle(
         time_lock, metadata.sender_puzzle_hash, metadata.recipient_puzzle_hash, inner_puzzle, inner_solution
     )
     return CoinSpend(coin, puzzle, solution)
+
+
+async def add_clawback_outgoing_tx(
+    created_timestamp: uint64,
+    puzzle_store: WalletPuzzleStore,
+    tx_store: WalletTransactionStore,
+    coin_state: CoinState,
+    peer: WSChiaConnection,
+    metadata: ClawbackMetadata,
+) -> None:
+    # Create Clawback outgoing transaction
+    clawback_coin_spend: CoinSpend = await fetch_coin_spend_for_coin_state(coin_state, peer)
+    clawback_spend_bundle: SpendBundle = SpendBundle([clawback_coin_spend], G2Element())
+    if await puzzle_store.puzzle_hash_exists(clawback_spend_bundle.additions()[0].puzzle_hash):
+        tx_record = TransactionRecord(
+            confirmed_at_height=uint32(coin_state.spent_height),
+            created_at_time=created_timestamp,
+            to_puzzle_hash=metadata.sender_puzzle_hash
+            if clawback_spend_bundle.additions()[0].puzzle_hash == metadata.sender_puzzle_hash
+            else metadata.recipient_puzzle_hash,
+            amount=uint64(coin_state.coin.amount),
+            fee_amount=uint64(0),
+            confirmed=True,
+            sent=uint32(0),
+            spend_bundle=clawback_spend_bundle,
+            additions=clawback_spend_bundle.additions(),
+            removals=clawback_spend_bundle.removals(),
+            wallet_id=uint32(1),
+            sent_to=[],
+            trade_id=None,
+            type=uint32(TransactionType.OUTGOING_CLAWBACK),
+            name=clawback_spend_bundle.name(),
+            memos=list(compute_memos(clawback_spend_bundle).items()),
+        )
+        await tx_store.add_transaction_record(tx_record)
--- a/chia/wallet/puzzles/clawback/drivers.py
+++ b/chia/wallet/puzzles/clawback/drivers.py
@@ -3,33 +3,22 @@ from __future__ import annotations
 import logging
 from typing import Any, List, Optional, Set, Union
 
-from blspy import G2Element
-
 from chia.consensus.default_constants import DEFAULT_CONSTANTS
-from chia.protocols.wallet_protocol import CoinState
-from chia.server.ws_connection import WSChiaConnection
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.program import Program
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.types.coin_spend import CoinSpend
 from chia.types.condition_opcodes import ConditionOpcode
-from chia.types.spend_bundle import SpendBundle
 from chia.util.condition_tools import conditions_for_solution
-from chia.util.ints import uint32, uint64
+from chia.util.ints import uint64
 from chia.util.misc import VersionedBlob
 from chia.wallet.puzzles.clawback.metadata import ClawbackMetadata
 from chia.wallet.puzzles.load_clvm import load_clvm_maybe_recompile
 from chia.wallet.puzzles.p2_delegated_puzzle_or_hidden_puzzle import MOD
-from chia.wallet.transaction_record import TransactionRecord
 from chia.wallet.uncurried_puzzle import UncurriedPuzzle
-from chia.wallet.util.compute_memos import compute_memos
 from chia.wallet.util.curry_and_treehash import calculate_hash_of_quoted_mod_hash, curry_and_treehash
 from chia.wallet.util.merkle_tree import MerkleTree
-from chia.wallet.util.transaction_type import TransactionType
-from chia.wallet.util.wallet_sync_utils import fetch_coin_spend_for_coin_state
 from chia.wallet.util.wallet_types import RemarkDataType
-from chia.wallet.wallet_puzzle_store import WalletPuzzleStore
-from chia.wallet.wallet_transaction_store import WalletTransactionStore
 
 P2_1_OF_N = load_clvm_maybe_recompile("p2_1_of_n.clsp")
 P2_CURRIED_PUZZLE_MOD = load_clvm_maybe_recompile("p2_puzzle_hash.clsp")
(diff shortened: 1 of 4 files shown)


chia/wallet/wallet_state_manager.py
```diff
@@ -1419,6 +1419,7 @@ async def _add_coin_states(
                                 fee = 0
 
                                 to_puzzle_hash = None
+                                coin_spend = await fetch_coin_spend_for_coin_state(coin_state, peer)
```
trepca, line 1422: shouldn't this be under `if derivation_record is None` ?

chia/wallet/wallet_state_manager.py
```diff
@@ -1428,6 +1429,15 @@ async def _add_coin_states(
...
                                         amount += coin.amount
+                                        if coin_spend is not None:
+                                            # Check if this is a Clawback coin
+                                            puzzle: Program = coin_spend.puzzle_reveal.to_program()
+                                            solution: Program = coin_spend.solution.to_program()
+                                            uncurried = uncurry_puzzle(puzzle)
+                                            clawback_metadata = match_clawback_puzzle(uncurried, puzzle, solution)
+                                            if clawback_metadata is not None:
+                                                # Add the Clawback coin as the interested coin for the sender
+                                                await self.add_interested_coin_ids([coin.name()])
```
trepca, line 1440: since we already received all the data for this coin, do we really need to add it to interested coin ids here? can we reuse data we fetched to add it to the wallet?

tests/wallet/test_wallet.py
```diff
@@ -777,7 +777,7 @@ async def test_clawback_resync(
         await server_2.start_client(PeerInfo(self_hostname, uint16(server_1._port)), None)
         await server_3.start_client(PeerInfo(self_hostname, uint16(server_1._port)), None)
         expected_confirmed_balance = await full_node_api.farm_blocks_to_wallet(count=num_blocks, wallet=wallet)
```
trepca, line 779: does this test fail without this resync fix?

tests/wallet/test_wallet.py
```diff
@@ -777,7 +777,7 @@ async def test_clawback_resync(
...
-        normal_puzhash = await wallet.get_new_puzzlehash()
+        normal_puzhash = bytes32(b"\00" * 32)
```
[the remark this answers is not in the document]
ytx1991: If the sender sends clawback to itself, the bug will not be triggered. We need to send it to an address not owned by the current wallet.

chia/wallet/wallet_state_manager.py
ytx1991: We just need to fetch the parent coin spend once. for coin in additions is child coin level.

ytx1991: No, for here we are already sure we know the coin puzzle hash so derivation_record cannot be none.

tests/wallet/test_wallet.py
ytx1991: No, but it doesn't cover the issue. That's why we didn't find this bug.

chia/wallet/wallet_state_manager.py
ytx1991: I thought about that. If it is an unspent coin, then we still need to add it to the table and that will trigger another coin state processing for the clawback coin. It doesn't compatible with the coin_added change we will do for the clawback.

tests/wallet/test_wallet.py
trepca: lets add a test that covers this issue

ytx1991: This change will cover the case

chia/wallet/wallet_state_manager.py
```diff
@@ -1493,9 +1488,20 @@ async def _add_coin_states(
                                     for rem_coin in tx_record.removals:
                                         if rem_coin == coin_state.coin:
                                             confirmed_tx_records.append(tx_record)
-
                             for tx_record in confirmed_tx_records:
                                 await self.tx_store.set_confirmed(tx_record.name, uint32(coin_state.spent_height))
+                            if record.coin_type == CoinType.CLAWBACK:
+                                await self.interested_store.remove_interested_coin_id(coin_state.coin.name())
+                                if len(confirmed_tx_records) == 1 and record.metadata is not None:
```
trepca, line 1495: can you explain what's happening here

ytx1991: This is the weird issue I mentioned in the engineering channel. After the user sends clawback to another wallet and claims/clawback then starts the resync, the wallet will receive 1 coin creation event and 1 coin spent event. Usually, during the resync, we will only receive 1 coin spent event. This weird behavior will cause we cannot recover the Clawback_Outgoing tx.
To fix this issue, I added this part. For a spent clawback coin which we already saved it in the DB, we expect there are two TXs need to be updated. One for CLAWBACK_INCOMING and another for CLAWBACK_OUTGOING. If we only find 1 TX then it means we don't have the outgoing TX. So in this case, we want to add a CLAWBACK_OUTGOING TX.
Actually, this is a general issue for other assets too. It means the wallet is not able to recovery 100% of outgoing TXs.

trepca: can you please add this comment as a comment in the code

chia/wallet/puzzles/clawback/drivers.py
```diff
@@ -179,3 +190,39 @@ def generate_clawback_spend_bundle(
         time_lock, metadata.sender_puzzle_hash, metadata.recipient_puzzle_hash, inner_puzzle, inner_solution
     )
     return CoinSpend(coin, puzzle, solution)
+
+
+async def add_clawback_outgoing_tx(
```
xdustinface, line 195: Assumed we need to keep this because we really need both places to add it, is `chia.wallet.puzzles` really the right place for this?

chia/wallet/wallet_state_manager.py
```diff
@@ -1417,17 +1399,32 @@ async def _add_coin_states(
...
+                                for child_state in children:
+                                    coin: Coin = child_state.coin
                                     derivation_record = await self.puzzle_store.get_derivation_record_for_puzzle_hash(
                                         coin.puzzle_hash
                                     )
                                     if derivation_record is None:  # not change
                                         to_puzzle_hash = coin.puzzle_hash
                                         amount += coin.amount
+                                        if coin_spend is None:
+                                            coin_spend = await fetch_coin_spend_for_coin_state(coin_state, peer)
```
xdustinface, line 1418: Why is all this code related to the spend processing here in the loop?

chia/wallet/wallet_state_manager.py
```diff
@@ -1417,17 +1399,32 @@ async def _add_coin_states(
...
-
+                                known_puzzle_hash: bool = await self.puzzle_store.puzzle_hash_exists(
+                                    coin_state.coin.puzzle_hash
+                                )
                                 to_puzzle_hash = None
+                                coin_spend: Optional[CoinSpend] = None
                                 # Find coin that doesn't belong to us
                                 amount = 0
-                                for coin in additions:
+                                for child_state in children:
```
xdustinface, line 1409:
```suggestion
                                for child_state in children:
```

Why did you change this loop to use `children` is you still only access the `coin` which is what `additions` already contains?

chia/wallet/wallet_state_manager.py
```diff
@@ -1417,17 +1399,32 @@ async def _add_coin_states(
...
+                                            coin_spend = await fetch_coin_spend_for_coin_state(coin_state, peer)
+
+                                        # Check if this is a Clawback coin
+                                        puzzle: Program = coin_spend.puzzle_reveal.to_program()
+                                        solution: Program = coin_spend.solution.to_program()
+                                        uncurried = uncurry_puzzle(puzzle)
+                                        clawback_metadata = match_clawback_puzzle(uncurried, puzzle, solution)
+                                        if clawback_metadata is not None and known_puzzle_hash:
+                                            # Add the Clawback coin as the interested coin for the sender
+                                            await self.add_interested_coin_ids([coin.name()])
```
xdustinface, line 1427: Can you explain why we need this here?

chia/wallet/wallet_state_manager.py
xdustinface: I spent few hours now looking into this and it feels now more like hack than a fix. Since the trusted test passes without this block we should imo rather better understand the issue why the untrusted doesn't behave the same than just adding this block here. What i know so far is the difference is that we don't long sync in this case for untrusted but instead end up doing a short sync via `sync_from_untrusted_close_to_peak` and we even end up with this edge case here https://github.com/Chia-Network/chia-blockchain/blob/88a692666d737f56a74b8241a36e89d508bd3343/chia/wallet/wallet_node.py#L1174-L1190 because the chain isn't long enough for a weight proof.

xdustinface (changes requested):

chia/wallet/puzzles/clawback/drivers.py
ytx1991: We don't have a wallet for the Clawback. The basic idea is to put this in the Clawback related files.

chia/wallet/wallet_state_manager.py
ytx1991: This code path is for the coin that is already spent the first time we saw it. This usually means we are doing a resync. But since we are adding the clawback coin to interested_coin manually for the sender, it cannot recover during the resync. This means during the resync, the sender will not receive any coin state about the Clawback coin. So we want to get the coin spends for these coins and check if they are Clawback coins owned by the current wallet.

chia/wallet/wallet_state_manager.py
ytx1991: As I mentioned in the comment above, if we want to recover the clawback for the sender, we need to manually add it to the interested_coin table since we don't own the puzzle hash and hint. The basic idea is we check the parent coin of the clawback coin. If it has a clawback coin owned by the current wallet, then we add it to the interested_coin so that we can receive the clawback coin state.

chia/wallet/wallet_state_manager.py
```diff
@@ -1275,6 +1275,7 @@ async def _add_coin_states(
         for coin_name, coin_state in zip(coin_names, coin_states):
             if peer.closed:
                 raise ConnectionError("Connection closed")
+            print(f"{self.wallet_node.logged_in_fingerprint} {coin_state}")
```
xdustinface, line 1278:
```suggestion
```

tests/wallet/test_wallet.py
```diff
@@ -762,95 +761,194 @@ async def test_clawback_resync(
...
-        assert config["wallet"]["reset_sync_for_fingerprint"] == fingerprint
-        new_config = wallet_node.config.copy()
-        new_config["reset_sync_for_fingerprint"] = config["wallet"]["reset_sync_for_fingerprint"]
-        new_config["database_path"] = "wallet/db/blockchain_wallet_v2_test_CHALLENGE_KEY.sqlite"
-        wallet_node_2.config = new_config
-        wallet_node_2.root_path = wallet_node.root_path
-        wallet_node_2.local_keychain = wallet_node.local_keychain
+        wallet_node_1.config["database_path"] = "wallet/db/blockchain_wallet_v2_test1_CHALLENGE_KEY.sqlite"
+        wallet_node_2.config["database_path"] = "wallet/db/blockchain_wallet_v2_test2_CHALLENGE_KEY.sqlite"
+        print("Resync")
```
xdustinface, line 885:
```suggestion
```

tests/wallet/test_wallet.py
```diff
@@ -762,95 +761,194 @@ async def test_clawback_resync(
...
+        outgoing_clawback_txs = await wallet_node_1.wallet_state_manager.tx_store.get_transactions_between(
             1, 0, 100, type_filter=TransactionTypeFilter.include([TransactionType.OUTGOING_CLAWBACK])
         )
-        assert len(outgoing_clawback_txs) == 1
+        assert len(outgoing_clawback_txs) == 2
         assert outgoing_clawback_txs[0].confirmed
+        assert outgoing_clawback_txs[1].confirmed
+
+        # transactions should be the same
+        print(before_txs)
```
xdustinface, line 942:
```suggestion
```

xdustinface (changes requested):

xdustinface (dismissed):

chia/wallet/wallet_state_manager.py
ytx1991: Unless we change the hint mechanism fundamentally, there isn't another way. Currently, we only check the first element in the memo and see if it is a known puzzle hash. We need to check all elements in the hint. But this may cause other issues.

chia/wallet/wallet_state_manager.py
ytx1991: Ideally, we want to hint both the sender and recipient puzzle hash. But we can only hint one puzzle hash and we are pointing to the recipient. That's why the recipient doesn't have this issue. As I mentioned above, unless we change how the hinting works, I don't think there is another way.

chia/wallet/wallet_state_manager.py
```diff
@@ -1417,8 +1417,12 @@ async def _add_coin_states(
                             additions = [state.coin for state in children]
                             if len(children) > 0:
                                 fee = 0
-
+                                known_puzzle_hash: bool = await self.puzzle_store.puzzle_hash_exists(
```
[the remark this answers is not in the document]
ytx1991: Yes, that is the assumption. This is a double-check since we want to do this expensive check only if it is necessary.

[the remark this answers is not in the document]
xdustinface: Right now it doesn't help to avoid the expensive check though because you only check against `known_puzzle_hash` after you already fetched and processed the coin spend, right?

If you really want to make this potentially avoid the expensive check it should rather be like below i guess?

```python
if known_puzzle_hash and coin_spend is None:
``` 

But again, can you think about any situation where we would end up here without already knowing this coin_state's puzzle hash belongs to us? Just want to make sure we don't add another DB lookup for every coin only to "double check" if its at the end not really required.

tests/wallet/test_wallet.py
```diff
@@ -762,95 +761,194 @@ async def test_clawback_resync(
...
+        expected_confirmed_balance += await full_node_api.farm_blocks_to_wallet(count=num_blocks, wallet=wallet_1)
         await time_out_assert(
-            20, wallet_node.wallet_state_manager.coin_store.count_small_unspent, 0, 1000, CoinType.CLAWBACK
+            20, wallet_node_1.wallet_state_manager.coin_store.count_small_unspent, 0, 1000, CoinType.CLAWBACK
+        )
+        await time_out_assert(
+            20, wallet_node_2.wallet_state_manager.coin_store.count_small_unspent, 0, 1000, CoinType.CLAWBACK
         )
+        assert wallet_node_1.wallet_state_manager is not None
+        assert wallet_node_2.wallet_state_manager is not None
```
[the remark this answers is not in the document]
ytx1991: It is for mypy check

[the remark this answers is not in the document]
xdustinface: Like i said, it will never be `None` so this "mypy check" makes no sense and you can just drop those lines. 

https://github.com/Chia-Network/chia-blockchain/blob/226c6017649a0efc713bb6f390cddb8c085f530f/chia/wallet/wallet_node.py#L167-L173

tests/wallet/test_wallet.py
```diff
@@ -762,95 +761,190 @@ async def test_clawback_resync(
...
-        wallet_node_2.local_keychain = wallet_node.local_keychain
+        wallet_node_1.config["database_path"] = "wallet/db/blockchain_wallet_v2_test1_CHALLENGE_KEY.sqlite"
+        wallet_node_2.config["database_path"] = "wallet/db/blockchain_wallet_v2_test2_CHALLENGE_KEY.sqlite"
+
         # use second node to start the same wallet, reusing config and db
-        await wallet_node_2._start_with_fingerprint(fingerprint)
-        assert wallet_node_2._wallet_state_manager
-        await server_3.start_client(PeerInfo(self_hostname, uint16(server_1._port)), None)
+        await wallet_node_1._start()
+        assert wallet_node_1.wallet_state_manager is not None
```
xdustinface, line 884:
```suggestion
```

tests/wallet/test_wallet.py
```diff
@@ -762,95 +761,190 @@ async def test_clawback_resync(
...
+        await wallet_server_1.start_client(PeerInfo(self_hostname, uint16(full_node_server._port)), None)
+        await wallet_node_2._start()
+        assert wallet_node_2.wallet_state_manager is not None
```
xdustinface, line 887:
```suggestion
```

xdustinface (commented):
There are still two open conversations and you obviously added another two of those pointless `not None` assertions with the last commit.

emlowe:
In talking with @xdustinface , he wanted a final check on the resyncing of the wallet tests. @ytx1991 Has confirmed that he is resyncing by using a new wallet DB entirely instead of using the resync feature, so this code is correct. Marking conversations resolved.

tests/wallet/test_wallet.py
emlowe: Resolving per discussion with @xdustinface and @ytx1991

tests/wallet/test_wallet.py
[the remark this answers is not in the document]
emlowe: Resolving per discussion with @xdustinface and @ytx1991

emlowe:
Close and Reopen for CI check

arvidn:
@emlowe it looks like @xdustinface had approved this PR, but it was dismissed. I'm not sure how that happened.
Is it meant to be approved?

emlowe (approved):
Approved

PR #15855, opened 2023-07-25
Mojave is dead

wallentx:
Long live Big Sur

### Purpose:
Nobody should be running Mojave anymore 

### Current Behavior:
Our CI fails to build installers for MacOS 

### New Behavior:
Our CI works, and we no longer support Mojave


hoffmang9 (dismissed):
lgtm

PR #15858, opened 2023-07-25
make ConditionOpcodes not streamable

arvidn:
`ConditionOpcode`s can be more than 1 byte now (as we introduced multi-byte, unknown, condition codes with cost, in the hard-fork).

Leaving them streamable in their current form (as a single byte) seems risky. It turns out that we no longer need them to be streamable anyway. This is a hold-over from the times when the multi-threaded block validation passed `ConditionWithArgs` across the process boundary.

### DIFF
--- a/chia/types/condition_opcodes.py
+++ b/chia/types/condition_opcodes.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import enum
-from typing import Any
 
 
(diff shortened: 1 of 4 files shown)


richardkiss (approved):
Seems reasonable. If the tests pass, then I think all is good.

emlowe (approved):

PR #15859, opened 2023-07-25
Simplify get_max_send_amount for XCH and CATs

Quexington:
### Purpose:
Simplify the get_max_send_amount method in the fungible wallets to not actually sign a transaction to measure the cost of a single transaction.  The cost of transactions in these wallets doesn't fluctuate almost at all, so reasonable estimations should result in the desired behavior.

### Current Behavior:
Sign a transaction when trying to figure out the cost of a single transaction.

### New Behavior:
Use a hard coded estimate instead.


chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -297,31 +289,15 @@ async def get_confirmed_balance(self, record_list: Optional[Set[WalletCoinRecord
     async def get_unconfirmed_balance(self, unspent_records: Optional[Set[WalletCoinRecord]] = None) -> uint128:
         return await self.wallet_state_manager.get_unconfirmed_balance(self.id(), unspent_records)
 
+    @property
+    def cost_of_single_tx(self) -> int:
+        return 30000000  # Estimate
```
trepca, line 294: we could consider memos here too?

Quexington: The behavior before didn't consider anything like that and memos could drive up the cost of a transaction, but it's very unlikely to include so many memos that we end up failing on fewer coins.

geoffwalmsley (approved):
Looks good.

wjblanke (approved):
aok

PR #15860, opened 2023-07-25
Change test to run only once per consensus mode as a flake workaround

emlowe:
As a workaround for test_clear_pending_roots flakes, reduce the number of test runs by only running the tests for the PLAIN consensus mode. The changes to consensus should have no impact on this test, and by reducing the number of times this test runs, it should help reduce the number of strange failures seen on Windows that seem entirely unrelated to the test itself.


wallentx (approved):

arvidn (approved):

arvidn:
we only made it to test `1301/1450` still. maybe there are more tests that need to be limited to just `Mode.PLAIN`

PR #15862, opened 2023-07-25
Drop macos-10.14 support

wallentx:
### Purpose:
Dropping Mojave after 2.0.0

### Current Behavior:
currently uses macos-10.14

### New Behavior:
macos-11

### DIFF
--- a/.github/workflows/build-macos-installers.yml
+++ b/.github/workflows/build-macos-installers.yml
@@ -33,7 +33,7 @@ jobs:
       matrix:
         python-version: [3.9]
         os:
-          - runs-on: macos-10.14
+          - runs-on: macos-11
             name: intel
             file-suffix: ""
             mac-package-name: "Chia-darwin-x64"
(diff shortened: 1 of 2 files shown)


emlowe:
We should also adjust the macos target in the Wheel repos to be macos 11 such that the binary wheels specify they don't support 10.14 or 10.15 anymore.

emlowe:
We may also want to set the envvar MACOSX_DEPLOYMENT_TARGET=11 when building the macos installer - I think this will automatically prevent the app from being able to be launched on previous versions

emlowe:
Unfortunately setting the deployment target envvar didn't seem to do what I wanted, so that might need to get done in a different way. The Chia application doesn't "advertise" itself as not supported 10.14, so you can try to run it, and then it gets stuck loading the services. Ideally uses wouldn't be able to launch the app at all.

But this build properly, so approving so we can move forward in main

emlowe (approved):

PR #15870, opened 2023-07-26
fix initial-config for testnet fork points heights

arvidn:
fix typo in `initial-config.yaml` and bump soft-fork 3 activation height on testnet


wjblanke (approved):
aok

emlowe:
I guess we need the fix for the `"` in release. Do we just allow this to checkpoint into main, or do you want to make a separate PR?

PR #15886, opened 2023-07-28
Add Wallet CLI Unit tests

jack60612:
### Purpose:
This PR adds wallet cli unit tests, using the same framework as the poc in #15746 
This also adds coverage for #15718 

### Current Behavior:
N/A (No tests)

### New Behavior:
N/A (Tests, Kyle is happy!!! 🥳)
Some commands now have a `--no-prompt` argument to skip any interactions.

### Testing Notes:
This pr adds coverage to the entire wallet cli

This PR is being split up for ease of reviewing:


jack60612:
the missing lines are a result of the new --no-confirm argument being added and can be ignored.

chia/cmds/wallet.py
```diff
@@ -479,9 +497,14 @@ def make_offer_cmd(
     filepath: str,
     fee: str,
     reuse: bool,
+    no_confirm: bool,
+    no_file: bool,
 ) -> None:
     from .wallet_funcs import make_offer
 
+    if filepath == "" and not no_file:
```
arvidn, line 505: is the offer printed to stdout when `--no-file` is passed?
There's a convention among some unix tools to use `-` as the filename to indicate `stdout` or `stdin`

chia/cmds/wallet.py
```diff
@@ -590,10 +624,18 @@ def take_offer_cmd(
...
+    is_flag=True,
+    default=False,
+)
+def cancel_offer_cmd(
+    wallet_rpc_port: Optional[int], fingerprint: int, id: str, insecure: bool, fee: str, no_confirm: bool
+) -> None:
     from .wallet_funcs import cancel_offer
 
-    asyncio.run(cancel_offer(wallet_rpc_port, fingerprint, Decimal(fee), id, insecure))
+    asyncio.run(cancel_offer(wallet_rpc_port, fingerprint, Decimal(fee), id, not insecure, no_confirm))
```
arvidn, line 638: this `not` here looks like an important change in behavior. it looks like something that shouldn't be mixed up in a PR adding unit tests, and like something that should have a clear rationale

chia/cmds/wallet_funcs.py
```diff
@@ -528,32 +530,34 @@ async def make_offer(
...
 
-                    print()
-                    nft_confirmation = input(
-                        "Offers for NFTs will have royalties automatically added. "
-                        + "Are you sure you would like to continue? (y/n): "
-                    )
-                    if nft_confirmation not in ["y", "yes"]:
+                    if not no_confirm:
+                        print()
+                        nft_confirmation = input(
```
arvidn, line 535: the problem with solving this by introducing `no_confirm` is that you kind of just punt on the core of it. This is still not covered by tests (afaiu). Is there no way in python to write to `stdin`?

One way to make this a bit more palatable would be to wrap `input()` in our own function that we use to ask for confirmation, and it always return `True` if `no_confirm` is set. That way only that function is lacking unit tests. and maybe we can even figure out some way to test it in isolation.

Maybe en even better API is to throw if the user denies. That way, callers don't need to check the return value and exit.

chia/cmds/wallet_funcs.py
```diff
@@ -528,32 +530,34 @@ async def make_offer(
...
-                    with open(pathlib.Path(filepath), "w") as file:
-                        offer, trade_record = await wallet_client.create_offer_for_ids(
-                            offer_dict, driver_dict=driver_dict, fee=fee, reuse_puzhash=reuse_puzhash
-                        )
-                        if offer is not None:
+                offer, trade_record = await wallet_client.create_offer_for_ids(
+                    offer_dict, driver_dict=driver_dict, fee=fee, reuse_puzhash=reuse_puzhash
+                )
+                if offer is not None:
+                    if not no_file:
```
arvidn, line 552: is the `--no-file` feature important for the tests?
I would think you could just ask for a temporary directory and write the file there

tests/cmds/test_wallet.py
```diff
@@ -160,3 +188,854 @@ async def get_coin_records(self, request: GetCoinRecords) -> Dict[str, Any]:
...
+
+        async def cancel_offer(self, trade_id: bytes32, fee: uint64 = uint64(0), secure: bool = True) -> None:
+            self.add_to_log("cancel_offer", (trade_id, fee, secure))
+            return None
+
+    inst_rpc_client = CancelOfferRpcClient()  # pylint: disable=no-value-for-parameter
+    test_rpc_clients.wallet_rpc_client = inst_rpc_client
+    command_args = ["wallet", "cancel_offer", FINGERPRINT_ARG, "--no-confirm", "-m1", "--id", test_offer_id]
+    success, output = run_cli_command(capsys, root_dir, command_args)
+    assert success
```
arvidn: if every invocation of `run_cli_command()` asserts `success`, you could probably just move that into the function and not return it

tests/cmds/test_wallet.py
```diff
@@ -160,3 +188,854 @@ async def get_coin_records(self, request: GetCoinRecords) -> Dict[str, Any]:
...
+        "set_cat_name": [(2, "examplecat"), (3, "examplecat")],
+    }
+    test_rpc_clients.wallet_rpc_client.check_log(expected_calls)
+
+
+def test_make_offer(capsys: object, get_test_cli_clients: Tuple[TestRpcClients, Path]) -> None:
+    test_rpc_clients, root_dir = get_test_cli_clients
+
+    # set RPC Client
+    class MakeOfferRpcClient(TestWalletRpcClient):
```
arvidn: I would have expected that some tests could share the mocked client class. Is there no opportunity for reusing this?

tests/cmds/cmd_test_utils.py
```diff
@@ -407,26 +407,24 @@ async def test_get_wallet_client(
...
+    except SystemExit as e:
+        if e.code != 0:
+            exited_cleanly = False
     finally:  # always reset sys.argv
         sys.argv = argv_temp
-    if not exited_cleanly:  # so we can look at what went wrong
-        print(f"\n{output.out}\n{output.err}")
-    return exited_cleanly, output.out
+    output = capsys.readouterr()  # type: ignore[attr-defined]
+    assert exited_cleanly, f"\n{output.out}\n{output.err}"
```
arvidn, line 426: I see you made this change later

tests/cmds/test_wallet.py
arvidn, line 1: I take it these tests were moved verbatim, right? no other changes?

chia/cmds/wallet.py
```diff
@@ -101,6 +101,7 @@ def get_transaction_cmd(wallet_rpc_port: Optional[int], fingerprint: int, id: in
     default=False,
     help="Only show clawback transactions",
 )
+@click.option("--testing", is_flag=True, default=False, hidden=True, help="Used by cli tests")
```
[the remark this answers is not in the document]
jack60612: yeah

chia/cmds/wallet.py
```diff
@@ -130,13 +132,14 @@ def get_transactions_cmd(
...
-    sys.stdout.flush()
-    sys.stdout.close()
+    if not testing:
+        # The flush/close avoids output like below when piping through `head -n 1`
+        # which will close stdout.
+        #
+        # Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
+        # BrokenPipeError: [Errno 32] Broken pipe
+        sys.stdout.flush()
+        sys.stdout.close()
```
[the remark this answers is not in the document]
jack60612: if it flushes the stdout i cant even read the result, so not really

chia/cmds/wallet.py
jack60612: yeah whoever made the test seemed to forget it, the command would do the wrong thing without it.

tests/cmds/test_wallet.py
jack60612: the only change was switching from `bytes([number here]*32)` to `get_bytes32(number here)` to make the code easier to read.

tests/cmds/test_wallet.py
jack60612: yeah i ended up moving it

tests/cmds/test_wallet.py
jack60612: If the rpc call / function is used in multiple tests it is added to  TestWalletRpcClient directly, so some tests just directly use it instead of making a new class.

chia/cmds/wallet_funcs.py
jack60612: yeah i could do that.

chia/cmds/wallet.py
jack60612: no its just saved, but not outputted at all.

jack60612: i could remove that arg i guess.

chia/cmds/wallet_funcs.py
jack60612: oh wait i remember why, by using the --no-file argument i would not have to fake the part that compresses the offer and turns it into bytes.

chia/cmds/wallet_funcs.py
jack60612: i could add this.

jack60612: added

chia/cmds/wallet.py
arvidn: so `--no-file` is also a pure test option then, right? It would never be useful in production. If so, it would seem like a better solution to have the tests output to temporary directories

chia/cmds/wallet.py
arvidn: this isn't a test though. It looks like a bug in production code. should we fix this bug in the 2.0 branch as well?

arvidn: this change looks correct though. I just think it's buried in mostly unit test code. I think it should be pulled out and maybe fixed in 2.0 if it applies as well

chia/cmds/wallet_funcs.py
arvidn: unless it would be insurmountable to fake, I think that would be a better approach. You wouldn't have to add test code mixed in to the production code (and you would get better test coverage metric).

I would expect that you don't actually need to save the offer in any correct format, or anything but just an empty buffer perhaps

arvidn (commented):
it sounds like `--no-file` could be dropped at least. I'm hopeful `--testing` can be removed too. Worst case, it would be named something that indicates what behavior it has, but if we *really* need to flush and close `stdout` like that, I'd like to understand why. If it works without it, I'd say we should just remove it

chia/cmds/wallet.py
arvidn: https://github.com/Chia-Network/chia-blockchain/pull/16012

chia/cmds/wallet_funcs.py
jack60612: yeah i went back with a fresh mind and it was easy, fixed,

chia/cmds/wallet.py
jack60612: removed

chia/cmds/wallet.py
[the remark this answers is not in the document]
jack60612: for clarity, closing it broke pytest,  but i'm not sure why only this command has it as its pretty unnecessary.

arvidn (approved):
I think this looks good. Have you done some manual smoke testing to make sure there aren't any serious mistakes?

jack60612:
> I think this looks good. Have you done some manual smoke testing to make sure there aren't any serious mistakes?

yeah i tested out the input part, `chia show -s` and `chia wallet show`

PR #15888, opened 2023-07-28
Add the option for arbitrary conditions to make_solution

Quexington:
### Purpose:

Right now `make_solution` doesn't takes magic conditions, but they are hinted as `Any`.  This solidifies that definition a bit to use the new `Condition` subclasses instead.  It also edits a few spots that had commented that they were doing a hack because `make_solution` didn't take magic conditions.


matt-o-how (approved):
This is a great change

wjblanke (approved):
aok

PR #15889, opened 2023-07-28
Add condition opcodes for agg sigs to condition_codes.clib

Quexington:


arvidn (dismissed):
nice test!
since these conditions won't be available until after the hard-fork, it might make sense to make a comment about that. it's a year out after all

arvidn (approved):

PR #15892, opened 2023-07-29
cudaplot fix

ChiaMineJP:
Fixed an issue where `chia plotters bladebit cudaplot ...` command fails with an error.
This is caused by:
- `--device` option was set as an `int` where it should be `str`
- Lack of  `-t` and `-2` flags support

You can test the behavior by running
`chia plotters bladebit cudaplot -n 1 -d <Final dir> -r 0 --compress 0 --device 0 -t <Tmp dir>`
with bladebit 3 installed. (If you install from source, put the bladebit_cuda(.exe) binary under `venv/bin/`)
Before this PR, the above command fails. After this PR, plotting will begin.

### DIFF
--- a/chia/plotters/bladebit.py
+++ b/chia/plotters/bladebit.py
@@ -365,7 +365,7 @@ def plot_bladebit(args, chia_root_path, root_path):
         call_args.append("--no-t2-direct")
     if "device" in args and str(args.device).isdigit():
         call_args.append("--device")
-        call_args.append(args.device)
+        call_args.append(str(args.device))
     if "no_direct_downloads" in args and args.no_direct_downloads is not None:
         call_args.append("--no-direct-downloads")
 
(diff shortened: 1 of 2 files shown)


wallentx (approved):

ChiaMineJP:
To test the changed line for coverage, it needs to invoke a subprocess which creates a plot by bladebit but currently that's difficult due to the limited resources.

arvidn:
```
chia/plotters/bladebit.py (0.0%): Missing lines 368
```

arvidn:
> To test the changed line for coverage, it needs to invoke a subprocess which creates a plot by bladebit but currently that's difficult due to the limited resources (CUDA capable CI instance).

this could be tested by sticking a test executable in the `PATH`, or something that ensures it's invoked with the right command line. i.e. you don't need to actually run the production bladebit (and probably shouldn't).

I'm fine with making a test-coverage exception for now though.

arvidn (approved):

PR #15899, opened 2023-07-30
various typing fixes

arvidn:
This PR adds type stubs for `blspy` and addresses mypy errors that resulted from it.

There are 3 lines missing coverage:

```
chia/full_node/full_node.py (0.0%): Missing lines 1297
```
This is a trivially correct change (I think). It's encoding an invariant we have in FullBlock as an `assert`.
This function seems simple to break out and unit test, which we should do in the future.

```
chia/rpc/wallet_rpc_api.py (0.0%): Missing lines 2257
```
This is a typo in the DID recovery code. It should have a test. Matt reviewed the change and said it looks good. I think it's a fairly trivially correct change.

```
tests/util/benchmark_cost.py (0.0%): Missing lines 116
```
We don't instrument benchmarks for coverage. And we probably don't even run this benchmark in CI, since this was another trivial typo.


chia/cmds/keys_funcs.py
```diff
@@ -285,9 +285,9 @@ def sign(message: str, private_key: PrivateKey, hd_path: str, as_bytes: bool, js
 
 def verify(message: str, public_key: str, signature: str, as_bytes: bool) -> None:
     data = bytes.fromhex(message) if as_bytes else bytes(message, "utf-8")
-    public_key = G1Element.from_bytes(bytes.fromhex(public_key))
-    signature = G2Element.from_bytes(bytes.fromhex(signature))
-    print(AugSchemeMPL.verify(public_key, data, signature))
+    pk: G1Element = G1Element.from_bytes(bytes.fromhex(public_key))
+    sig: G2Element = G2Element.from_bytes(bytes.fromhex(signature))
+    print(AugSchemeMPL.verify(pk, data, sig))
```
arvidn, line 290: the variables `public_key` and `signature` already have types, mypy doesn't allow shadowing the previous variables with new ones of different types

chia/util/cached_bls.py
```diff
@@ -42,12 +42,12 @@ def get_pairings(
                 pk_parsed = G1Element.from_bytes(pks[i])
                 pk_bytes_to_g1[pks[i]] = pk_parsed
 
-            pairing = pk_parsed.pair(aug_hash)
+            pairing = aug_hash.pair(pk_parsed)
```
arvidn, line 45: `pair()` is commutative. this unifies the member function we use to be called on `G2Element`

chia/util/cached_bls.py
```diff
@@ -42,12 +42,12 @@ def get_pairings(
...
 
             h = std_hash(aug_msg)
             cache.put(h, pairing)
             pairings[i] = pairing
-    return pairings
+    return cast(List[G2Element], pairings)
```
arvidn, line 50: `pairings` is a list of *optional* `G2Element`s, but this function ensures they're all set, so we cast it to the correct return value

tests/blockchain/test_blockchain.py
```diff
@@ -2074,7 +2074,7 @@ async def test_aggsig_garbage(self, empty_blockchain, opcode, with_garbage, expe
         synthetic_secret_key = calculate_synthetic_secret_key(secret_key, DEFAULT_HIDDEN_PUZZLE_HASH)
         public_key = synthetic_secret_key.get_g1()
 
-        args = [public_key, b"msg"] + ([b"garbage"] if with_garbage else [])
+        args = [bytes(public_key), b"msg"] + ([b"garbage"] if with_garbage else [])
```
arvidn, line 2077: args is `List[bytes]`

tests/clvm/test_puzzles.py
```diff
@@ -34,7 +34,7 @@
 
 def secret_exponent_for_index(index: int) -> int:
     blob = index.to_bytes(32, "big")
-    hashed_blob = BasicSchemeMPL.key_gen(std_hash(b"foo" + blob))
+    hashed_blob = AugSchemeMPL.key_gen(std_hash(b"foo" + blob))
```
arvidn, line 37: `key_gen` does not depend on the scheme. this is the only place in out code base where we refer to `BasicSchemeMPL`. Everywhere else we use `AugSchemeMPL`.

tests/pools/test_pool_rpc.py
```diff
@@ -402,7 +402,7 @@ def mempool_empty() -> bool:
                                 [some_wallet.wallet_state_manager.private_key], status.current.owner_pubkey
                             )
                             assert owner_sk is not None
-                            assert owner_sk != auth_sk
+                            assert owner_sk[0] != auth_sk
```
arvidn, line 405: `owner_sk` is a pair of (secret_key, index). This test was meant to compare the secret keys

tests/util/benchmark_cost.py
```diff
@@ -113,7 +113,7 @@ def benchmark_all_operators():
 
     for i in range(0, 1000):
         private_key: PrivateKey = master_sk_to_wallet_sk(secret_key, uint32(i))
-        public_key = private_key.public_key()
+        public_key = private_key.get_g1()
```
arvidn, line 116: this seems like an old typo, it was introduced in the first commit of this file, Tue Mar 24 23:03:31 2020

tests/util/network_protocol_data.py
```diff
@@ -60,13 +60,13 @@
 
 proof_of_space = ProofOfSpace(
     bytes32(bytes.fromhex("1fb331df88bc142e70c110e21620374118fb220ccc3ef621378197e850882ec9")),
-    G1Element(
+    G1Element.from_bytes(
```
arvidn, line 63: `blspy` does provide a constructor for `G1Element` and `G2Element` that takes bytes, I think it's a bit clearer to use the `from_bytes()` function

arvidn:
```
tests/util/benchmark_cost.py (0.0%): Missing lines 116
```
This is a benchmark. We probably don't run it on CI, but even if we did, we don't instrument benchmarks for coverage

chia/cmds/keys_funcs.py
altendky: Do the `.from_bytes()` methods not hint the proper return type?  The `: G1Element` doesn't seem needed for clarity etc.

chia/util/cached_bls.py
altendky: Maybe it would be better to have a separate list hinted properly such that mypy can help us enforce this?

tests/blockchain/test_blockchain.py
altendky: Did mypy catch this?  Or should be explicitly `: List[bytes]` to make sure it is enforced.

tests/pools/test_pool_rpc.py
altendky: Wouldn't have caught the error here, but a not-tuple return would be nice.  Not for this PR though.

chia/cmds/keys_funcs.py
arvidn: good point. They are not necessary, I added them to explain to myself how these are new variables with different types

tests/pools/test_pool_rpc.py
arvidn: yeah, I'm not entirely confident how this is related to `blspy` types, but it seems that untyped variables are "viral", so even comparing `Tuple[PrivateKey, int]` to `PrivateKey` doesn't trigger mypy, because `PrivateKey` is untyped

altendky: We have strict equality enabled for the strict files, but not for the files in the mypy exceptions list.  But yes, if one thing is an `Any` then...  almost anything you do with it is `Any` as well and mypy won't complain much.

https://github.com/Chia-Network/chia-blockchain/blob/f0cb47baefbab6f0e94bb60dd8df9985bf4079dd/mypy.ini.template#L17

arvidn:
@altendky when I run this locally, I get the following errors (and I don't understand why):
```
benchmarks/utils.py:76: error: Returning Any from function declared to return "G2Element"  [no-any-return]
chia/simulator/block_tools.py:545: error: Returning Any from function declared to return "G2Element | None"  [no-any-return]
Found 2 errors in 2 files (checked 811 source files)
```
but it seems to pass on CI

chia/rpc/wallet_rpc_api.py
```diff
@@ -2254,7 +2254,7 @@ async def did_recovery_spend(self, request) -> EndpointResult:
                 pubkey = G1Element.from_bytes(hexstr_to_bytes(request["pubkey"]))
             else:
                 assert wallet.did_info.temp_pubkey is not None
-                pubkey = wallet.did_info.temp_pubkey
```
arvidn, line 2257: @matt-o-how This line is not exercised by any test, and has a (fatal) typo. I don't think it's reasonable to hold up this fix by adding a test. Do you want to vouch for this change?

chia/full_node/full_node.py
```diff
@@ -1294,6 +1294,7 @@ def has_valid_pool_sig(self, block: Union[UnfinishedBlock, FullBlock]) -> bool:
             and block.foliage.prev_block_hash != self.constants.GENESIS_CHALLENGE
             and block.reward_chain_block.proof_of_space.pool_public_key is not None
         ):
+            assert block.foliage.foliage_block_data.pool_signature is not None
```
arvidn, line 1297: apparently there's no test covering this line. I think it's trivially correct and warrants an exception. In the long run this function should probably be pulled out as a free function and unit tested

chia/rpc/wallet_rpc_api.py
arvidn: matt approves

install.sh
```diff
@@ -334,6 +335,10 @@ python -m pip install wheel
 python -m pip install --extra-index-url https://pypi.chia.net/simple/ miniupnpc==2.2.2
 python -m pip install ${EDITABLE} ."${EXTRAS}" --extra-index-url https://pypi.chia.net/simple/
 
+if [ -n "$BLSPY_STUBS" ]; then
+python -m pip install ${EDITABLE} ./blspy-stubs
+fi
+
```
altendky, line 341: If this is truly temporary then I guess we can let it stay separate, but the fewer pip calls the better it can handle itself.  I know we already have two and that's not good.

chia/util/cached_bls.py
```diff
@@ -42,12 +42,12 @@ def get_pairings(
...
                 pk_bytes_to_g1[pks[i]] = pk_parsed
 
-            pairing = pk_parsed.pair(aug_hash)
+            pairing = aug_hash.pair(pk_parsed)
 
             h = std_hash(aug_msg)
             cache.put(h, pairing)
             pairings[i] = pairing
-    return pairings
+    return cast(List[GTElement], pairings)
```
altendky, line 50: Maybe it would be better to have a separate list hinted properly such that mypy can help us enforce this?

chia/util/cached_bls.py
altendky: reraised after modification

altendky (commented):
I'm not excited about adding type stubs here for other packages, but I'm also not driving the future of these libs.  The asserts in the runtime code aren't great, but are also in keeping with our practices.

Other than above and my couple inline comments I'm ok with this.  Someone familiar with the actual code changes should review them specifically.

install.sh
arvidn: I have a PR in the works that replaces blspy, but it's not ready yet.

chia/util/cached_bls.py
arvidn: I went for `cast` to avoid making changes to the original intention. But I'll add a separate list to make the types more explicit

rostislav (approved):
Looks good

paninaro (approved):
lgtm

PR #15900, opened 2023-07-30
Arvid's fix flake8 warnings (PR #15897)

AmineKhaldi:


wallentx (approved):

arvidn (approved):

arvidn:
I think, technically, it would be better to pin our dependencies, so this doesn't happen in release branches (or any older branches either)

PR #15903, opened 2023-07-31
Add coverage failure as comment to PR

wallentx:
### Purpose:
Adds coverage failure to the PR as a comment

### Current Behavior:
When coverage fails, you have to go to the job summary to see the details 

### New Behavior:
Adds those details as a comment to the PR

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,8 +40,8 @@ jobs:
         run: |
           python tests/build-job-matrix.py --per directory --verbose > matrix.json
           cat matrix.json
-          echo configuration=$(cat matrix.json) >> $GITHUB_OUTPUT
-          echo matrix_mode=${{ ( github.event_name == 'workflow_dispatch' ) && 'all' || ( github.repository_owner == 'Chia-Network' && github.repository != 'Chia-Network/chia-blockchain' ) && 'limited' || ( github.repository_owner == 'Chia-Network' && github.repository == 'Chia-Network/chia-blockchain' && github.ref == 'refs/heads/main' ) && 'main' || ( github.repository_owner == 'Chia-Network' && github.repository == 'Chia-Network/chia-blockchain' && startsWith(github.ref, 'refs/heads/release/') ) && 'all' || ( github.repository_owner == 'Chia-Network' && github.repository == 'Chia-Network/chia-blockchain' && startsWith(github.base_ref, 'release/') ) && 'all' || 'main' }} >> $GITHUB_OUTPUT
+          echo configuration=$(cat matrix.json) >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -40,8 +40,8 @@ jobs:
...
+          echo configuration=$(cat matrix.json) >> "$GITHUB_OUTPUT"
+          echo matrix_mode=${{ ( github.event_name == 'workflow_dispatch' ) && 'all' || ( github.repository_owner == 'Chia-Network' && github.repository != 'Chia-Network/chia-blockchain' ) && 'limited' || ( github.repository_owner == 'Chia-Network' && github.repository == 'Chia-Network/chia-blockchain' && github.ref == 'refs/heads/main' ) && 'main' || ( github.repository_owner == 'Chia-Network' && github.repository == 'Chia-Network/chia-blockchain' && startsWith(github.ref, 'refs/heads/release/') ) && 'all' || ( github.repository_owner == 'Chia-Network' && github.repository == 'Chia-Network/chia-blockchain' && startsWith(github.base_ref, 'release/') ) && 'all' || 'main' }} >> "$GITHUB_OUTPUT"
```
arvidn, line 44: does `GITHUB_OUTPUT` sometimes have spaces in it?
If this used to fail, are we not using this file anywhere currently?

.github/workflows/test.yml
```diff
@@ -159,11 +159,36 @@ jobs:
...
+                  print "| " a[1] " | " a[2] " | None |"
+                } else {
+                  split(a[2],b,":");
+                  print "| " a[1] " | " b[1] " | " substr(b[2], 15) " |"
+                }
+              }' coverage-reports/diff-cover-stdout > coverage-reports/diff-cover-md-table
+          cat coverage-reports/diff-cover-md-table >> "$GITHUB_STEP_SUMMARY"
+          # Add comment to PR
+          PR_NUM=$(jq -r '.number' "$GITHUB_EVENT_PATH")
+          gh pr edit "$PR_NUM" --add-label "coverage-diff"
```
arvidn, line 185: it would be nice to automatically remove this label when there is no missing lines in the diff too

.github/workflows/test.yml
```diff
@@ -159,11 +159,36 @@ jobs:
...
+          gh pr comment "$PR_NUM" -F coverage-reports/diff-cover-md-table
```
arvidn, line 186: perhaps it's asking for a lot, but it would be nice to update the existing comment, if there is one

.github/workflows/test.yml
wallentx: I ran `actionlint` on this file to make sure I didn't introduce any white spaces or anything, and it picked up on these existing variables not being quoted, and was simply doing some cleanup.

.github/workflows/test.yml
wallentx: Done

.github/workflows/test.yml
wallentx: Done. It actually embeds a hidden comment ID when it adds the comment, and removes that comment before posting a new one, or removes it up on success change.

wallentx:
@arvidn feel free to make a commit to this PR that would trigger a coverage failure, and then undo that change to see that it operates as desired

arvidn (approved):

Issue #15637, opened 2023-06-27
[Bug]  Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x000001B85BA47B80>


azrielitay:
### What happened?

When closing chia GUI app the bellow errors show in log. 

### Version

1.8.1

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output

```shell
(node:7468) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 state_changed listeners added to [J]. Use emitter.setMaxListeners() to increase limit
(Use `Chia --trace-warnings ...` to show where the warning was created)
stderr: Exception ignored in: <function _ProactorBasePipeTransport.__del__ at 0x000001B85BA47B80>
Traceback (most recent call last):
  File "asyncio\proactor_events.py", line 116, in __del__
stderr:   File "asyncio\proactor_events.py", line 108, in close
  File "asyncio\base_events.py", line 751, in call_soon
  File "asyncio\base_events.py", line 515, in _check_closed
stderr: RuntimeError: Event loop is closed
```

azrielitay:
Open

בתאריך יום ד׳, 12 ביולי 2023 ב-14:05 מאת github-actions[bot] <
***@***.***>:

Issue #12009, opened 2022-06-22
Error after minting NFT - RuntimeError: Event loop is closed


wallentx:
### What happened?

On winders, I minted an NFT and got an error output, but everything seemed to have worked just fine(?)
https://test.skynft.org/item.php?launcher_id=bb6fc35e177beaa24e93d1b5a95cf325ac90856184311a12ebb2f4c3356a2853

@wjblanke found the following:
https://github.com/aio-libs/aiohttp/issues/6635
> recommends sleep(.25)?

https://docs.aiohttp.org/en/stable/client_advanced.html#graceful-shutdown
> Note that the appropriate amount of time to wait will vary from application to application.

### Version

1.3.6671

### What platform are you using?

Windows

### What ui mode are you using?

CLI

### Relevant log output

```shell
.\chia rpc wallet nft_mint_nft '{
  \"wallet_id\": 7,
  \"uris\": [
    \"https://lh3.googleusercontent.com/59H4vAxqBvyD5BKXCxHVMrZzWSS0BG5izhAms0FFsq0Hy71YgaiuRBGES7KJJY0oT5TkrNPIj3lVLUMG2nyljWu3I1jRgvr45NYm=s0\"
  ],
  \"series_total\": 12,
  \"series_number\": 1,
  \"hash\": \"24e04e8a50fbba39dfdef8759362236fb4ca8b2d4a02391a19bf4b7375ea97bb\",
  \"artist_percentage\": 1500,
  \"artist_address\": \"txch1yxpslrx30k7lnngpfczr3ltrge0ap25f4739jet5lz069lhn5szsu49uyh\"

}'
{
    "spend_bundle": {
        "aggregated_signature": "0x834a3be681a38f7cc10b338644619c35a057c0f89bbebee7ae8abf46d38d08a6c9d21740b8b6b239376c7bc7a9843c85123d0044ae40e0fe9115ebda24a5eb9d69d47fefe154cdc966cdb4c6253d619c834b53266eb4a2ffc97d6f7124c8cdf5",
        "coin_solutions": [
            {
                "coin": {
                    "amount": 999999999999,
                    "parent_coin_info": "0xd9ad384705f0013c1e69433bbacc2ff2e1e0c7b0791622e78be9e05297fd179b",
                    "puzzle_hash": "0x26fbaa149a5a1dd9cd30463cc6f49502c63f0dded3cce47e3a1d018c8447a8b5"
                },
[...]

PR #15908, opened 2023-08-01
allow `_ProactorBasePipeTransport.__del__` failures in tests

altendky:
### Purpose:

Ignores the issues such as reported in https://github.com/Chia-Network/chia-blockchain/issues/12009 and https://github.com/Chia-Network/chia-blockchain/issues/15637 that are showing up in the DataLayer CLI testing.  I believe these are being caused by https://github.com/python/cpython/issues/92841 which was fixed in 3.10.6 and not backported to 3.8 or 3.9.  This does not fix the issue during runtime.


emlowe (approved):

altendky:
Coverage failure is due to not running the tests on 3.10.6+ on `main`.

PR #15911, opened 2023-08-01
Add `extra_conditions` as an option to transaction endpoints

Quexington:
Most transactions have the ability to include any of our consensus conditions alongside the ones returned from their standard spend.  We currently don't expose that functionality but this PR adds it as an option to every transaction endpoint (similar to TXConfig).  This implicitly creates support for things like expiring offers and pre-signed transactions.  Further work is needed to make this information displayed in the wallet state manager somehow (PRs in the pipeline)


geoffwalmsley (approved):

wjblanke (approved):
aok

PR #15915, opened 2023-08-01
Add timelock information to Trades and Transactions

Quexington:
A previous PR added the ability to throw in extra conditions to all of the transaction endpoints.  This includes timelock conditions that could potentially invalidate a transaction after it has already successfully made it into the mempool.  To help with UX in this scenario, this PR adds an extra field to `TransactionRecord` and `TradeRecord` that describes all of the various timelock restrictions on the transaction.  This will help a UI for example describe to the user that if their transaction is not confirmed, it is expired.

(I also removed some unused TradeStore functions for test coverage purposes)

### DIFF
--- a/chia/wallet/trade_record.py
+++ b/chia/wallet/trade_record.py
@@ -1,19 +1,22 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Type, TypeVar
 
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.ints import uint8, uint32, uint64
 from chia.util.streamable import Streamable, streamable
+from chia.wallet.conditions import ConditionValidTimes
 from chia.wallet.trading.offer import Offer
 from chia.wallet.trading.trade_status import TradeStatus
+
+_T_TradeRecord = TypeVar("_T_TradeRecord", bound="TradeRecordOld")
 
 
 @streamable
 @dataclass(frozen=True)
-class TradeRecord(Streamable):
+class TradeRecordOld(Streamable):
     """
     Used for storing transaction data and status in wallets.
     """
@@ -47,10 +50,18 @@ class TradeRecord(Streamable):
         return formatted
 
     @classmethod
-    def from_json_dict_convenience(cls, record: Dict[str, Any], offer: str = "") -> "TradeRecord":
+    def from_json_dict_convenience(
+        cls: Type[_T_TradeRecord], record: Dict[str, Any], offer: str = ""
+    ) -> _T_TradeRecord:
         new_record = record.copy()
         new_record["status"] = TradeStatus[record["status"]].value
         del new_record["summary"]
         del new_record["pending"]
         new_record["offer"] = offer
         return cls.from_json_dict(new_record)
+
+
+@streamable
+@dataclass(frozen=True)
+class TradeRecord(TradeRecordOld):
+    valid_times: ConditionValidTimes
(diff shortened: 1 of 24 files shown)


chia/wallet/trade_record.py
```diff
@@ -47,10 +50,18 @@ def to_json_dict_convenience(self) -> Dict[str, Any]:
...
         new_record["status"] = TradeStatus[record["status"]].value
         del new_record["summary"]
         del new_record["pending"]
         new_record["offer"] = offer
         return cls.from_json_dict(new_record)
+
+
+@streamable
+@dataclass(frozen=True)
+class TradeRecord(TradeRecordOld):
```
ytx1991, line 66: Why not just add a new field in the old class?

chia/wallet/trading/trade_store.py
```diff
@@ -140,12 +172,21 @@ async def create(
             except aiosqlite.OperationalError:
                 pass  # ignore what is likely Duplicate column error
 
+            cursor = await conn.execute(
+                "SELECT name FROM sqlite_master WHERE type='table' AND name='MIGRATED_VALID_TIMES_TRADES'"
+            )
+            if await cursor.fetchone() is None:
+                needs_valid_times_migration = True
+                await conn.execute("CREATE TABLE MIGRATED_VALID_TIMES_TRADES(_ blob)")
```
ytx1991, line 180: Why not just alter the table and add a new column with a default value? Therefore you don't have to all these migration things.

chia/wallet/trade_record.py
Quexington: In the DB migration code, I need to be able to de-serialize the old blobs and instead of hand rolling the de-serialization code in that block of code, I figured I would just make a new streamable class and leave the old one available for use during migration purposes.

chia/wallet/trading/trade_store.py
Quexington: This table being created is really just the equivalent of a boolean, unfortunately sql doesn't allow you to just store table metadata so I'm left with this hack.

The reason I have to do the migration is because we store the entire trade record in the DB as a blob under `trade_record`.  In the various DB methods, we will deserialize this blob using the `TradeRecord` streamable class and streamable is very specific that it can only deserialize things in exactly one way.  Therefore, I have to go into the DB and replace all of the blobs with blobs that are deserializable by the new `TradeRecord` class.

ytx1991 (dismissed):

wjblanke:
Just need to see what happens with the wallet db on downgrade to 2.0.1 with all the migrations

chia/wallet/trading/trade_store.py
```diff
@@ -140,6 +142,17 @@ async def create(
             except aiosqlite.OperationalError:
                 pass  # ignore what is likely Duplicate column error
 
+            try:
+                await conn.execute("CREATE TABLE trade_record_times(trade_id blob PRIMARY KEY, valid_times blob)")
```
ytx1991, line 146: I prefer not to create a new table just for this column. It is also easier when you read the data, you don't have to read data from two tables and ensure they are synced. You can take a look of how we add new column in other tables like coin_store.

ytx1991: You can just add a line after ln140
await conn.execute("ALTER TABLE trade_records ADD COLUMN valid_times blob DEFAULT EMPTY_BLOB()")

To make this backward compatible, the key is the Alter SQL must be the last Alter sentences. This means you are not inserting a column between existing columns. So when the user rollback to the old version, the index of row[] will not change.

Quexington: That was my instinct too but @emlowe raised an issue where when you have statements like this: https://github.com/Chia-Network/chia-blockchain/blob/e7778c77496fdb8ab822ef60f2722c44510245de/chia/wallet/wallet_transaction_store.py#L94 then altering the table actually is bad for forwards compatibility because then sqlite will error with something like `table transaction_records got N+1 values, was expecting N`

ytx1991 (approved):
Creating a new table looks messy but it should be the safest compatible solution we have. We should think about how to clean these tables in the future.

wjblanke (approved):
aok

PR #15916, opened 2023-08-01
Add new option to set sqlite soft_heap_limit in config

emlowe:
Add an option to set the SQLite `soft_heap_limit` in config.

The default value for this is 0 which means no set limit

This may be helpful on lower RAM systems (eg 4GB ) where SQLite seems to be using more memory due to fetching full blocks from the DB for Chip-13 verification. It's unknown what impact on performance this may have.

For 4GB systems, a limit of either 1GB or 1GiB does reduce overall memory usage of the full node in empirical testing.

### DIFF
--- a/chia/util/db_wrapper.py
+++ b/chia/util/db_wrapper.py
@@ -145,6 +145,7 @@ class DBWrapper2:
         synchronous: Optional[str] = None,
         foreign_keys: bool = False,
         row_factory: Optional[Type[aiosqlite.Row]] = None,
+        soft_heap_limit: Optional[int] = None,
     ) -> DBWrapper2:
         if log_path is None:
             log_file = None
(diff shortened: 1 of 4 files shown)


arvidn (dismissed):

chia/util/db_wrapper.py
```diff
@@ -145,6 +145,7 @@ async def create(
         synchronous: Optional[str] = None,
         foreign_keys: bool = False,
         row_factory: Optional[Type[aiosqlite.Row]] = None,
+        soft_heap_limit: Optional[int] = None,
```
altendky, line 148: If this is going to be optional then perhaps we shouldn't default to 0 when pulling out of the config?  And `null` should be encouraged, and used, in the config?

altendky (dismissed):

chia/util/db_wrapper.py
emlowe: RIght now, I'm thinking 0 might be better. although I'm a little confused in that sqlite seems to remember this pragma across test runs in some way. And 0 resets the limit back to no-limit, so 0 I think might be better.

arvidn (approved):

emlowe:
Further testing suggests this doesn't really work as initially thought during preliminary testing. Closing PR for now - possibly could explore in the future, but for now I don't understand enough to be confident this is helpful.

PR #15918, opened 2023-08-02
Fixed `chia plotnft show` command

ChiaMineJP:
Fixed an issue where `chia plotnft show` command showed nothing if `-i/--id` option is not provided.
I believe that `chia plotnft show` changes introduced by this PR(#15704 ) is not fully expected so my PR is reverting behavior before the #15704 was merged.

### DIFF
--- a/chia/cmds/plotnft_funcs.py
+++ b/chia/cmds/plotnft_funcs.py
@@ -181,6 +181,29 @@ async def pprint_pool_wallet_state(
             print(f"Expected to leave after block height: {expected_leave_height}")
 
 
+async def pprint_all_pool_wallet_state(
+    wallet_client: WalletRpcClient,
+    get_wallets_response: List[Dict[str, Any]],
+    address_prefix: str,
+    pool_state_dict: Dict[bytes32, Dict[str, Any]],
+) -> None:
+    print(f"Wallet height: {await wallet_client.get_height_info()}")
+    print(f"Sync status: {'Synced' if (await wallet_client.get_synced()) else 'Not synced'}")
+    for wallet_info in get_wallets_response:
+        pool_wallet_id = wallet_info["id"]
+        typ = WalletType(int(wallet_info["type"]))
+        if typ == WalletType.POOLING_WALLET:
+            pool_wallet_info, _ = await wallet_client.pw_status(pool_wallet_id)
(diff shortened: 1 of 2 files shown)


arvidn (dismissed):

chia/cmds/plotnft_funcs.py
```diff
@@ -181,6 +181,29 @@ async def pprint_pool_wallet_state(
             print(f"Expected to leave after block height: {expected_leave_height}")
 
 
+async def pprint_all_pool_wallet_state(
```
jack60612, line 184: Extra p in print

jack60612: Wait isn't this function above?

jack60612 (changes requested):
I think this merge / revert got messed up

chia/cmds/plotnft_funcs.py
ChiaMineJP: `pprint` means pretty-print.
There is alerady `pprint_pool_wallet_state` function around line 122.
`pprint_all_pool_wallet_state` just runs `pprint_pool_wallet_state` through all pool wallet.

jack60612: pprint is still kinda stupid but sounds good.

jack60612 (dismissed):
lgtm, last review sucked because i was on mobile.

jack60612 (approved):

jack60612 (approved):
lgtm, i cant see any new changes except the catch up with main

ChiaMineJP:
@arvidn @jack60612 
Sorry I accidentally added commits which should belong to another PR.
I ran `git reset HEAD^` to dismiss the wrong commit and force-pushed the original commit-history which doesn't include the wrong commit.
The state of this PR is the same to the one both of you reviewed and approved.
Sorry for letting you review the same code twice!

jack60612:
> @arvidn @jack60612 Sorry I accidentally added commits which should belong to another PR. I ran `git reset HEAD^` to dismiss the wrong commit and force-pushed the original commit-history which doesn't include the wrong commit. The state of this PR is the same to the one both of you reviewed and approved. Sorry for letting you review the same code twice!

all good, approved.

paninaro (approved):

PR #15921, opened 2023-08-03
Fixed an issue where insufficient block interval was used to calculate average block time

### LINKED ISSUE #15904: [Bug] 2.0 RC2 Estimated Time to Win Inconsistencies
### What happened?

When users farm compressed plots their estimated time to win will become inconsistent and change constantly (from a few seconds to multiple days).

It seems the client has difficulty in determining true farming space and calculating the estimated time to win from that.

Example images are from a few seconds after one another:
image

image

side note- this user has 1 harvester that reports to the full node from these images

Reported in Discord (https://discord.com/channels/1034523881404370984/1102690350218354920/1135625524509675711) by dachshundad.

### Version

2.0 RC2

Linux

GUI

### Relevant log output

```shell
no errors or warnings present in the logs
```

ChiaMineJP:
Before this PR
```
prev_height = uint32(max(newer_block.height - 1, newer_block.height - height_distance))
```
In most cases, `height_distance` was not used because `newer_block.height - 1` usually wins.

### DIFF
--- a/chia/rpc/full_node_rpc_api.py
+++ b/chia/rpc/full_node_rpc_api.py
@@ -56,7 +56,7 @@ async def get_average_block_time(
     if newer_block.height < 1:
         return None
 
-    prev_height = uint32(max(newer_block.height - 1, newer_block.height - height_distance))
+    prev_height = uint32(max(1, newer_block.height - height_distance))
     prev_hash = blockchain.height_to_hash(prev_height)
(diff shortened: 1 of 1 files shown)


wjblanke (approved):
aok

arvidn (commented):
it would be great to have a test case covering this too

PR #15922, opened 2023-08-03
Set parallel to 1

wallentx:
Attempt to prevent datalayer tests flakes


arvidn:
this won't fix errors running this locally (because that would be in a single `pytest` invocation).
I have a feeling that this test that cannot be run in parallel should just be skipped for any `consensus_mode != Mode.PLAIN` instead, so we only run it once

PR #15937, opened 2023-08-04
Add ALLOW_BACKREFS as a flag to get_puzzle_and_solution_for_coin in full_node_api.py

Quexington:

### DIFF
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -12,6 +12,7 @@ from secrets import token_bytes
 from typing import TYPE_CHECKING, Dict, List, Optional, Set, Tuple
 
 from blspy import AugSchemeMPL, G1Element, G2Element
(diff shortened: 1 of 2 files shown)


arvidn (commented):
this looks correct. It would be great to paramaterize our test for this function on whether the hard fork has activated or not as well (maybe `conftest.Mode`, to cover soft forks as well)

arvidn:
I think we might want to consider this fix for the 2.0 release branch as well

arvidn:
one issue is that passing in `ALLOW_BACKREFS` not only affects the parsing of the input, it also implies that the *output* is allowed to use backrefs. Which is problematic, because then the full node's response uses this encoding, and our wallet isn't prepared for that. (I was thinking I would have a year about to fix that).

emlowe:
Yea, I think this should be re-targeted to `release/2.0.0`

emlowe (approved):

PR #15938, opened 2023-08-04
bump soft-fork3 activation height

arvidn:
since release is slipping


emlowe (approved):

emlowe:
Does this need to change in initial-config?

hoffmang9 (approved):
lgtm

wallentx:
> Does this need to change in initial-config?

https://github.com/Chia-Network/chia-blockchain/blob/a5a988c32098d8a3f56d4c284680f20aeed2fccb/chia/util/initial-config.yaml#L74-L78

These look like they are prob not right.

arvidn:
the config file contains overrides for testnet. but testnet has already activated soft-fork3, so we can't change those.

arvidn:
I think we have to accept the benchmark workflow an exception, since we don't have those runners back up yet

PR #15940, opened 2023-08-05
Must force a flush here so reloaded cache doesn't have bogus values

wjblanke:
### Purpose:

Ensure cache coherency of height to hash

### Current Behavior:

Node could return orphaned blocks due to reorgs

### New Behavior:

Reorgs flush out height to hash cache

### Testing Notes:

Must force reorg, restart, then query block

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -286,8 +286,6 @@ class Blockchain(BlockchainInterface):
                 self.add_block_record(block_record)
                 if state_change_summary is not None:
                     self.__height_map.rollback(state_change_summary.fork_height)
-                    # Must force a flush here so reloaded cache doesn't have bogus values
-                    await self.__height_map.maybe_flush(True)
 
                 for fetched_block_record in records:
                     self.__height_map.update_height(
@@ -309,7 +307,7 @@ class Blockchain(BlockchainInterface):
             self._peak_height = block_record.height
 
         # This is done outside the try-except in case it fails, since we do not want to revert anything if it does
-        await self.__height_map.maybe_flush()
+        await self.__height_map.maybe_flush(True)
 
         if state_change_summary is not None:
             # new coin records added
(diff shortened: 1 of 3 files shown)


arvidn (commented):
when the node starts up, this cache is verified against the blockchain database and patched up in case of any discrepancy. I don't see how this could make any real difference

wjblanke:
arvid, on startup the cache is fixed up by taking the database and going down until you hit the top of the cache right?
if the top of the cache is bogus, due to an unflushed reorg, those blocks are orphans and not correct

arvidn:
> arvid, on startup the cache is fixed up by taking the database and going down until you hit the top of the cache right?

Oh, not quite. it starts at the peak and moves down the chain until the cache matches the database. i.e. after the fixup, the cache is expected to match the DB. Since the DB is the authoritative source of our state, that should be sufficient.

> if the top of the cache is bogus, due to an unflushed reorg, those blocks are orphans and not correct

If the top of the chain in the DB are the same orphaned blocks, that's OK. Then we'll reorg once we find a better peak.
If the DB peak have different blocks than the cache, the cache will be updated to match.

wjblanke:
> > arvid, on startup the cache is fixed up by taking the database and going down until you hit the top of the cache right?
> 
> Oh, not quite. it starts at the peak and moves down the chain until the cache matches the database. i.e. after the fixup, the cache is expected to match the DB. Since the DB is the authoritative source of our state, that should be sufficient.
> 
> > if the top of the cache is bogus, due to an unflushed reorg, those blocks are orphans and not correct
> 
> If the top of the chain in the DB are the same orphaned blocks, that's OK. Then we'll reorg once we find a better peak. If the DB peak have different blocks than the cache, the cache will be updated to match.

Ya i see that now
https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/block_height_map.py#L175

ok looking at other areas

PR #15943, opened 2023-08-05
Bladebit3 as the default plotter to install

ChiaMineJP:
Updated the default bladebit version to install.
How to test:
```
.\Install-plotter.ps1 bladebit
```

### DIFF
--- a/Install-plotter.ps1
+++ b/Install-plotter.ps1
@@ -16,7 +16,7 @@ param(
 
 $ErrorActionPreference = "Stop"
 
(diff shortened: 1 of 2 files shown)


wallentx (changes requested):
We need to test the .sh on MacOS, as bumping this might cause it to fail. Or perhaps leave the default as v2.0.1 for MacOS. We are currently no longer building bladebit for MacOS, but that could possibly change in the future.

wallentx (approved):

arvidn (approved):

PR #15946, opened 2023-08-06
fixup block store and coin store benchmarks

arvidn:
* Fix relative imports
* remove reference to v1 database
* additional benchmarks for `BlockStore`

### DIFF
--- a/benchmarks/block_store.py
+++ b/benchmarks/block_store.py
@@ -6,8 +6,10 @@ import random
 import sys
 from pathlib import Path
 from time import monotonic
+from typing import List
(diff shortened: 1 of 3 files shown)


altendky (approved):
I just glanced over the new tests.  We discussed some further refactors but I'm good with this change as a useful and more contained step.

It looks like there is a timing and printing pattern that could go into a context manager or somesuch along with potentially breaking each test piece into separate functions.  But yes, that doesn't belong here.

wjblanke (approved):
aok

Issue #15836, opened 2023-07-22
[Bug]  Increased CPU usage and Higher Partial Times


PhillWalker:
### What happened?

Since Chia Compression 4.3 alpha through to the current 2.0 RC1, partial times have increased noticeably, causing lots of lates on lower end equipment.

LebanonJon has been able to verify that the CPU usage has increased since this time, and WallentX has also acknowledged this increased resourced usage for non compression, and wonders if this has also impacted compression.

### Version

2.0 RC1

### What platform are you using?

Linux

### What ui mode are you using?

CLI

PhillWalker:
<img width="1538" alt="Screenshot 2023-07-22 at 12 24 22" src="https://github.com/Chia-Network/chia-blockchain/assets/72089507/03dba11a-0f0e-4b39-ac70-67f179f3a8d3">
You can see the sudden rise in partial times.

Previous version running in the screenshot was the last compression_fc branch build.
And the version where the jump happened was 2.0 Beta5 from main branch.

Issue still persists from early data on 2.0 RC1

Coyote-UK:
I can confirm I have seen this same issue using my Windows GUI harvester.
Windows 10, 2,453 C5 K32 plots. i7-4500U processor, 16GB DDR3 memory, SH920 mSATA HDD for mainnet db.
Logfile for 2.0.0b5 attached covering startup of node/farmer/harvester & 6 hours of farming, including several lates.
[debug.log](https://github.com/Chia-Network/chia-blockchain/files/12136938/debug.log)

PhillWalker:
It is worth noting, this is not the config.yaml spelling issue.

These are freshly generated config.yaml files for RC1, using the same values as before

PhillWalker:
<img width="857" alt="Screenshot 2023-07-24 at 12 36 28" src="https://github.com/Chia-Network/chia-blockchain/assets/72089507/3921c260-7b4c-4c6f-96ea-a22ce395947c">

Lates being caused since the upgrade.  These were all 25-30 seconds previously with no lates.

Coyote-UK:
I verified, in the settings tab on the GUI, that the correct settings for decompress**O**rs was being read.

PhillWalker:
I concur, my settings are correct too, and the issue is far from resolved

harold-b:
Apologies I did not see this sooner. My github I have my github notifications messed up due to a mail filter.

[...]

PR #15947, opened 2023-08-06
Revert "Add fields to BlockRecord. (#15695)"

arvidn:
This reverts commit e84bdb0b3056c150193e36ed2c00855c3ba08834.

To address https://github.com/Chia-Network/chia-blockchain/issues/15836

This change was introduced to support CHIP-13 anti-grind feature. But that was pulled from the 2.0 release already, so it's safe to pull this `BlockStore` change as well. We can fix the performance of this in `main` for the 2.1 release (which will include CHIP-13)

# not to be checkpointed back into main


arvidn:
```
chia/full_node/block_store.py (82.4%): Missing lines 529,630,635
```

fchirica (approved):

emlowe (approved):

PR #15963, opened 2023-08-07
Fixed an issue where `-t` option was required in cudaplot

ChiaMineJP:
`cudaplot` does not take the `-t/--tmp-dir` option however, it is required for `diskplot`
Update code in chia to correctly set the values based on invocation.


wallentx (approved):
:shipit:

emlowe (approved):

PR #15967, opened 2023-08-08
fix testnet10 sync-from-scratch

arvidn:
soft-fork2 was not a smooth softfork on testnet (but it was on mainnet).
This re-introduces the minimum logic for supporting syncing on testnet10.

The height 3000000 was picked arbitrarily assuming that there would be no violations of soft-fork2 past that height.

I would prefer to scrap testnet10 and start testnet11 instead of landing this PR.

I successfully synced testnet10 from scratch (up to block 3035527)

### DIFF
--- a/chia/consensus/block_body_validation.py
+++ b/chia/consensus/block_body_validation.py
@@ -495,11 +495,18 @@ async def validate_block_body(
     if npc_result is not None:
         assert npc_result.conds is not None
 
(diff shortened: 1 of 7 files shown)


arvidn:
the test coverage failure is part of the testnet10 specific logic

emlowe (approved):

PR #15970, opened 2023-08-08
Permissions test

ericaltendorf:
Do not submit.  Testing PR permissions

For the record, @altendky requested that Eric create this test.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Generate matrix configuration
         id: configure
         run: |
-          python tests/build-job-matrix.py --per directory --verbose > matrix.json
+          python tests/build-job-matrix.py --per directory --verbose --only clvm > matrix.json
           cat matrix.json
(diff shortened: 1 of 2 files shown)


altendky:
I think this run is not valid since it was triggered by a privileged user.  I'll check in with Eric in a bit.

altendky (changes requested):
this is not to be merged

altendky:
failed for eric's push-triggered (and me approved) run https://github.com/Chia-Network/chia-blockchain/actions/runs/5814492701/job/15764465113?pr=15970#step:13:39

failed when i reran that job https://github.com/Chia-Network/chia-blockchain/actions/runs/5814492701/job/15765985053?pr=15970#step:13:39

now let's see what happens when i close and reopen this as then it will be an external pr but _i_ will have triggered the run.

altendky:
still fails even when i initiated the run https://github.com/Chia-Network/chia-blockchain/actions/runs/5815436110/job/15767084192?pr=15970#step:13:39

altendky:
i think we got the data points here.  thanks for the help @ericaltendorf.  we can always reopen if useful.

PR #15971, opened 2023-08-08
fix DataLayer deadlock

### LINKED ISSUE #15955: [Bug] Datalayer get_value JSON RPC ceases functioning after multiple calls, requires Chia restart
### What happened?

**Description:**

When making multiple sequential calls to the `get_value` JSON RPC function, the DataLayer service becomes unresponsive and fails to process new requests. The issue continues to persist even after the script execution has been stopped. The only way to restore functionality is by restarting the Chia service. 

**Steps to reproduce:**

1. Create a store and add approximately 1000 keys/values. (The role of data size in this issue is currently unclear)
2. Write a script that iterates through an array of all keys in the store and calls the `get_value` JSON RPC for each key, one at a time.
3. Run the script.

**Expected result:**

The script should successfully retrieve the values of all keys in the store without causing the DataLayer service to become unresponsive.

**Actual result:**

The script ceases to process new keys part way through its execution. The exact point of failure varies with each run. Upon reaching this state, all JSON RPC functions become unresponsive, failing to return a response even after the script execution is halted.

**Workaround:**

The only known workaround at this time is to restart the Chia service.

**Severity:**

High. This issue completely halts the functionality of all JSON RPC calls and requires a service restart to resolve.

**Additional Notes:**

This issue is consistently reproducible on my machine. Further investigation is needed to understand whether the size or the number of the data keys have any impact on triggering this issue.

### Version

1.8.2

Windows

CLI

### Relevant log output

```shell
I could not detect any log output related to this issue.
```

altendky:
### Purpose:

https://github.com/Chia-Network/chia-blockchain/issues/15955

There were two locks acquired in opposite orders between some RPC calls and the periodic data management task via the call to `DataLayer.fetch_and_validate()`.  They are the data store db wrapper lock for the writer connection and the `DataLayer.lock` lock.  All uses of the `DataLayer.lock` were surrounding the update check prior to handling the RPC request.  In that update check it immediately acquired the db writer connection.  As such it was already exclusive since the db wrapper only allows a single write connection.  This PR removes the `DataLayer.lock`.

### Current Behavior:

Making lots of queries can easily deadlock the service.  Though, in case you care, it could still process a shutdown request normally.  `:]`

### New Behavior:

This deadlock scenario is resolved.

### Testing Notes:

I initially tested by triggering the deadlock via `venv/bin/chia rpc data_layer get_keys '{"id": "b9312cc24584fa34fc69f29fdbdd625e6c4f670b30e266c4032127d0a0fe1491"}' | jq -r '.keys[]' | xargs --max-procs=10 --replace={} venv/bin/chia data get_value --id b9312cc24584fa34fc69f29fdbdd625e6c4f670b30e266c4032127d0a0fe1491 --key "{}"`.

### Draft For:
- [x] description
- [x] tests
- [x] `TODO:`s
- [x] self consideration

### Followups For `main`:
- [x] consistently update outside the existing transaction https://github.com/Chia-Network/chia-blockchain/pull/16094
- [ ] review for potential use of read-only transactions

### DIFF
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -667,7 +667,6 @@ class DataLayer:
                 )
 
                 if len(changelist) > 0:
-                    # TODO: calls update too...  inside the above transaction...
                     new_root_hash = await self.batch_insert(
                         tree_id=offer_store.store_id,
                         changelist=changelist,
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -181,10 +181,9 @@ class DataLayer:
         tree_id: bytes32,
         changelist: List[Dict[str, Any]],
     ) -> bytes32:
-        # Make sure we update based on the latest confirmed root.
-        await self._update_confirmation_status(tree_id=tree_id)
-
         async with self.data_store.transaction():
+            # Make sure we update based on the latest confirmed root.
+            await self._update_confirmation_status(tree_id=tree_id)
             pending_root: Optional[Root] = await self.data_store.get_pending_root(tree_id=tree_id)
             if pending_root is not None:
                 raise Exception("Already have a pending root waiting for confirmation.")
@@ -213,7 +212,6 @@ class DataLayer:
     ) -> TransactionRecord:
         # Make sure we update based on the latest confirmed root.
         await self._update_confirmation_status(tree_id=tree_id)
-
         pending_root: Optional[Root] = await self.data_store.get_pending_root(tree_id=tree_id)
         if pending_root is None:
             raise Exception("Latest root is already confirmed.")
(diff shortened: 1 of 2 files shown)


arvidn (commented):
It looks like the existing code expects that `_update_confirmation_status()` make multiple async calls internally and require no state changes in between. Is that not the case? Was this mutex always redundant?

Is there a risk that `_update_confirmation_status()` may be changed in the future where it *would* require the state to be locked down for the duration of that call?

arvidn:
I see you kind of answer my question in:

> In that update check it immediately acquired the db writer connection. As such it was already exclusive since the db wrapper only allows a single write connection. 

This is assuming that it's mostly concerned about the state in the database, and not potential other state in the program

emlowe:
The code is making an "optimistic" call to `_update_confirmation_status()` to update the root - but it currently doesn't use any return values from this call, so the non-DB state isn't important.
If we started to return the root here and use it, then there might be something we need to lock.
But we don't think so in the current code base

altendky:
I chatted with Arvid directly and I believe the concern is less about the change and more about just the general hazard of non-db data potentially not being protected.  I agree that a review of this could be useful.  Very generally, I have wondered about making more granular data holding classes with the db wrapper pattern of having to request a thing that gives you access to use the data and that access is controlled by the class owning the data, not just by hoping you use its lock.  Rollback is a related relevant feature.

fchirica (approved):
LGTM

emlowe (approved):

PR #15972, opened 2023-08-08
send --fix-ssl-permissions to stderr

altendky:
### Purpose:

Avoid corrupting JSON outputs and such while still presenting the message to the user.


arvidn (approved):
it would also be nice to not print this message when running:
```
chia init --fix-ssl-permissions
```
:)

altendky:
It would...  someday...

PR #15977, opened 2023-08-08
extend tests being parametrised by soft-fork

arvidn:

Currently, only a few of our test fixtures are parameterized by soft-fork and hard-fork activation. They primarily test the full node. However, the fixtures we use to test wallets, and multiple nodes, are not parameterized. This makes for a blind-spot in our testing (missing a bug that was instead found by manual testing ahead of the 2.0 release).

In order to parameterize these fixtures, the underlying functions used to set up the simulation environment also need to be changed to take a full `ConsensusConstants` object, indicating the mode we're testing. Specifically, these functions:
* `setup_simulators_and_wallets()`
* `setup_simulators_and_wallets_service()`
* `setup_simulators_and_wallets_inner()`

Instead of taking an "overrides" dictionary, the caller can override the constants before passing it in. However, there's only one instance where a caller actually does override constants (and it's not a particularly convenient interface).

A consequence of parameterizing is that we end up running a lot more tests in CI. This needs to be mitigated by disabling the parameterization for some tests and also to bump the timeouts.

### DIFF
--- a/tests/core/daemon/test_daemon.py
+++ b/tests/core/daemon/test_daemon.py
@@ -382,7 +382,7 @@ async def test_daemon_simulation(self_hostname, daemon_simulation):
     data = {"service": service_name}
     payload = create_payload("register_service", data, service_name, "daemon")
     await ws.send_str(payload)
-    message_queue = asyncio.Queue()
+    message_queue: asyncio.Queue = asyncio.Queue()
 
     async def reader(ws, queue):
         while True:
--- a/tests/core/full_node/test_full_node.py
+++ b/tests/core/full_node/test_full_node.py
@@ -55,6 +55,7 @@ from chia.util.vdf_prover import get_vdf_info_and_proof
 from chia.wallet.transaction_record import TransactionRecord
 from chia.wallet.util.tx_config import DEFAULT_TX_CONFIG
 from tests.blockchain.blockchain_test_utils import _validate_and_add_block, _validate_and_add_block_no_error
+from tests.conftest import ConsensusMode
 from tests.connection_utils import add_dummy_connection, connect_and_get_peer
 from tests.core.full_node.stores.test_coin_store import get_future_reward_coins
 from tests.core.make_block_generator import make_spend_bundle
@@ -105,7 +106,9 @@ async def get_block_path(full_node: FullNodeAPI):
 class TestFullNodeBlockCompression:
     @pytest.mark.asyncio
     @pytest.mark.parametrize("tx_size", [3000000000000])
-    async def test_block_compression(self, setup_two_nodes_and_wallet, empty_blockchain, tx_size, self_hostname):
+    async def test_block_compression(
+        self, setup_two_nodes_and_wallet, empty_blockchain, tx_size, self_hostname, consensus_mode
+    ):
         nodes, wallets, bt = setup_two_nodes_and_wallet
         server_1 = nodes[0].full_node.server
         server_2 = nodes[1].full_node.server
@@ -165,7 +168,12 @@ class TestFullNodeBlockCompression:
         # Confirm generator is not compressed
         program: Optional[SerializedProgram] = (await full_node_1.get_all_full_blocks())[-1].transactions_generator
         assert program is not None
-        assert detect_potential_template_generator(uint32(5), program) is not None
+        if consensus_mode == ConsensusMode.HARD_FORK_2_0:
+            # after the hard fork we don't use this compression mechanism
+            # anymore, we use CLVM backrefs in the encoding instead
+            assert detect_potential_template_generator(uint32(5), program) is None
+        else:
+            assert detect_potential_template_generator(uint32(5), program) is not None
         assert len((await full_node_1.get_all_full_blocks())[-1].transactions_generator_ref_list) == 0
 
         # Send another tx
@@ -195,7 +203,12 @@ class TestFullNodeBlockCompression:
         program: Optional[SerializedProgram] = (await full_node_1.get_all_full_blocks())[-1].transactions_generator
         assert program is not None
         assert detect_potential_template_generator(uint32(6), program) is None
-        assert len((await full_node_1.get_all_full_blocks())[-1].transactions_generator_ref_list) > 0
+        if consensus_mode == ConsensusMode.HARD_FORK_2_0:
+            # after the hard fork we don't use this compression mechanism
+            # anymore, we use CLVM backrefs in the encoding instead
+            assert len((await full_node_1.get_all_full_blocks())[-1].transactions_generator_ref_list) == 0
+        else:
+            assert len((await full_node_1.get_all_full_blocks())[-1].transactions_generator_ref_list) > 0
 
         # Farm two empty blocks
         await full_node_1.farm_new_transaction_block(FarmNewBlockProtocol(ph))
@@ -270,7 +283,12 @@ class TestFullNodeBlockCompression:
         program: Optional[SerializedProgram] = (await full_node_1.get_all_full_blocks())[-1].transactions_generator
         assert program is not None
         assert detect_potential_template_generator(uint32(9), program) is None
-        assert len((await full_node_1.get_all_full_blocks())[-1].transactions_generator_ref_list) > 0
+        if consensus_mode == ConsensusMode.HARD_FORK_2_0:
+            # after the hard fork we don't use this compression mechanism
+            # anymore, we use CLVM backrefs in the encoding instead
+            assert len((await full_node_1.get_all_full_blocks())[-1].transactions_generator_ref_list) == 0
+        else:
+            assert len((await full_node_1.get_all_full_blocks())[-1].transactions_generator_ref_list) > 0
 
         # Creates a standard_transaction and an anyone-can-spend tx
         tr: TransactionRecord = await wallet.generate_signed_transaction(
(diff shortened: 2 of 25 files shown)


Quexington (commented):
Seems fine but 30-45 minute jump in CI runtime is really unfortunate.  We really gotta prioritize speeding these up now because we're only going to have more.  I think the main speedup (based on some experimentation I did some time ago) would be to have the DBs operate in memory instead of in a temp file.  Anyways, none of this is relevant directly here, just leaving this as a comment for the record.  (Not approving because PR is in draft but consider it soft-approved.)

arvidn:
I removed the `coverage-diff` label. the 3 missing lines are in test fixtures used in benchmarks, which we don't instrument

tests/core/full_node/test_full_node.py
```diff
@@ -165,7 +168,12 @@ async def check_transaction_confirmed(transaction) -> bool:
         # Confirm generator is not compressed
         program: Optional[SerializedProgram] = (await full_node_1.get_all_full_blocks())[-1].transactions_generator
         assert program is not None
-        assert detect_potential_template_generator(uint32(5), program) is not None
+        if consensus_mode == ConsensusMode.HARD_FORK_2_0:
+            # after the hard fork we don't use this compression mechanism
+            # anymore, we use CLVM backrefs in the encoding instead
+            assert detect_potential_template_generator(uint32(5), program) is None
+        else:
+            assert detect_potential_template_generator(uint32(5), program) is not None
```
altendky, line 176: I think that here and below it would be worth not copying the value and instead assigning it to a variable that is then used in the two assert conditions.

altendky (changes requested):

tests/pools/test_pool_rpc.py
```diff
@@ -486,12 +488,13 @@ async def test_absorb_self(
...
+        allowed=[ConsensusMode.PLAIN, ConsensusMode.HARD_FORK_2_0, ConsensusMode.SOFT_FORK3],
+        reason="the pooling test only creates one plot, and cannot pass CHIP-13 rules",
+    )
     @pytest.mark.asyncio
     async def test_absorb_self_multiple_coins(
-        self,
-        one_wallet_node_and_rpc: OneWalletNodeAndRpc,
-        fee: uint64,
-        self_hostname: str,
+        self, one_wallet_node_and_rpc: OneWalletNodeAndRpc, fee: uint64, self_hostname: str
```
xdustinface, line 497: This looks like some unrelated formatting?

tests/core/daemon/test_daemon.py
```diff
@@ -382,7 +382,7 @@ async def num_connections():
     data = {"service": service_name}
     payload = create_payload("register_service", data, service_name, "daemon")
     await ws.send_str(payload)
-    message_queue = asyncio.Queue()
+    message_queue: asyncio.Queue = asyncio.Queue()
```
xdustinface, line 385: Was this required for some reason?

tests/core/full_node/test_full_node.py
```diff
@@ -165,7 +168,16 @@ async def check_transaction_confirmed(transaction) -> bool:
...
         assert program is not None
-        assert detect_potential_template_generator(uint32(5), program) is not None
+        if consensus_mode == ConsensusMode.HARD_FORK_2_0:
+            # after the hard fork we don't use this compression mechanism
+            # anymore, we use CLVM backrefs in the encoding instead
+            check = lambda x: x is None  # noqa: 731
+            check_num_blocks = lambda x: x == 0  # noqa: 731
+        else:
+            check = lambda x: x is not None  # noqa: 731
+            check_num_blocks = lambda x: x > 0  # noqa: 731
```
xdustinface, line 178: Not so excited about this lambda blocks and the "noqa"s  but i guess its fine :) Maybe rename `check` to `check_none`.

tests/pools/test_pool_rpc.py
```diff
@@ -405,12 +406,13 @@ def mempool_empty() -> bool:
...
+        allowed=[ConsensusMode.PLAIN, ConsensusMode.HARD_FORK_2_0, ConsensusMode.SOFT_FORK3],
+        reason="the pooling test only creates one plot, and cannot pass CHIP-13 rules",
+    )
     @pytest.mark.asyncio
     async def test_absorb_self(
-        self,
-        one_wallet_node_and_rpc: OneWalletNodeAndRpc,
-        fee: uint64,
-        self_hostname: str,
+        self, one_wallet_node_and_rpc: OneWalletNodeAndRpc, fee: uint64, self_hostname: str
```
xdustinface, line 415: Looks like some unrelated formatting.

tests/pools/test_pool_rpc.py
```diff
@@ -558,12 +561,13 @@ async def test_absorb_self_multiple_coins(
...
+        allowed=[ConsensusMode.PLAIN, ConsensusMode.HARD_FORK_2_0, ConsensusMode.SOFT_FORK3],
+        reason="the pooling test only creates one plot, and cannot pass CHIP-13 rules",
+    )
     @pytest.mark.asyncio
     async def test_absorb_pooling(
-        self,
-        one_wallet_node_and_rpc: OneWalletNodeAndRpc,
-        fee: uint64,
-        self_hostname: str,
+        self, one_wallet_node_and_rpc: OneWalletNodeAndRpc, fee: uint64, self_hostname: str
```
xdustinface, line 570: Same.

tests/wallet/cat_wallet/config.py
```diff
@@ -1,5 +1,4 @@
 from __future__ import annotations
 
-parallel = False
```
xdustinface, line 3: Do you remember why you set it to `False` when you introduced it?

xdustinface (dismissed):

altendky (changes requested):

tests/wallet/cat_wallet/config.py
altendky: FYI, we did run a 25x set of flake check runs on the edited test directories (admittedly, fixture changes could affect more than those) and got a single timeout that was 3x out from normal runtimes and a single test failure.  Just evidence that we aren't making things catastrophically worse in terms of flakes, at least.

tests/core/full_node/test_full_node.py
altendky: Continued at https://github.com/Chia-Network/chia-blockchain/pull/15977#discussion_r1318547690

tests/wallet/cat_wallet/config.py
arvidn: @xdustinface I set it to false when I changed the default to be `true`, in order to preserve the existing behavior

tests/core/full_node/test_full_node.py
```diff
@@ -171,9 +171,13 @@ async def check_transaction_confirmed(transaction) -> bool:
...
             # after the hard fork we don't use this compression mechanism
             # anymore, we use CLVM backrefs in the encoding instead
-            assert detect_potential_template_generator(uint32(5), program) is None
+            check = lambda x: x is None  # noqa: 731
+            check_num_blocks = lambda x: x == 0  # noqa: 731
         else:
-            assert detect_potential_template_generator(uint32(5), program) is not None
+            check = lambda x: x is not None  # noqa: 731
+            check_num_blocks = lambda x: x > 0  # noqa: 731
+        assert check(detect_potential_template_generator(uint32(5), program))
```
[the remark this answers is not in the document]
arvidn: Yeah, I did that, and then I kept going because it still seemed quite verbose. I was surprised to learn that you can't have `assert` in a lambda, that was weird.
But sure, I'll undo the lambdas

tests/core/full_node/test_full_node.py
arvidn: flake8 complains about assigning lambdas to variables and wants you to use `def` instead. mypy doesn't like that you have conditional and different `def`s (and defs take a lot more space)

tests/core/daemon/test_daemon.py
arvidn: while working on the patch, mypy complained about not knowing the type of this variable. I'm not sure the thing that triggered it is still in the patch. sometimes mypy work in mysterious ways

tests/core/full_node/test_full_node.py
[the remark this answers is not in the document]
altendky: Lambda is just syntax and that syntax only allows a single expression, so no statements like `assert`.  I mean you can cheat and do stuff like `lambda x: [print(x), 2 * x][1]` to 'do multiple things and return one value' but...

tests/core/daemon/test_daemon.py
altendky: This test function isn't hinted and is also in the exclusion list so mypy wouldn't be looking inside anyways.  Maybe you temporarily hinted the function def at some point?

altendky (dismissed):

tests/core/daemon/test_daemon.py
arvidn: mypy will enable additional hinting if you introduce hinting for added code. e.g. if you add a hinted parameter to a function where nothing is hinted, some type checking is enabled

arvidn:
I agree the formatting changes were a bit unfortunate.

altendky (dismissed):

emlowe (dismissed):

emlowe (dismissed):

emlowe (dismissed):

emlowe (dismissed):

emlowe (dismissed):

emlowe (dismissed):

emlowe (approved):

PR #15986, opened 2023-08-09
Remove pipefail

wallentx:
The pipefail prevents the environment from getting set


altendky (changes requested):
this was intended to fail

altendky (approved):

paninaro (approved):

PR #15992, opened 2023-08-09
Fix timelord-install.sh for CentOS\RHEL

LeroyINC:
### Purpose:

This pull request fixes the issue were the install-timelord.sh fails on CentOS and RHEL

### Current Behavior:

When running the install-timelord.sh on CentOS\RHEL the install fails because the name of the python development package that the script tries to install is wrong. For CentOS\RHEL the name is "python3-devel"

If you chose to not to install the python3 development package using the "-n" option  --> "install-timelord.sh -n"
This causes the yum command to have a double space in the command and then yum throws an error and the script aborts.

### New Behavior:

The package name is updated only for CentOS/RHEL, the Package name for other OSes is not changed. The double space issues in the yum command has been fixed.

This results in the expected behavior.

### Testing Notes:

tested by running install-timelord.sh on CentOS 9 system with and without the -n switch

### DIFF
--- a/install-timelord.sh
+++ b/install-timelord.sh
@@ -82,7 +82,9 @@ elif [ "$(uname)" = "Linux" ] && type dnf || yum; then
 	echo "Found RedHat."
 
 	if [ "$INSTALL_PYTHON_DEV" ]; then
-		PYTHON_DEV_DEPENDENCY="$PYTHON_VERSION"-devel
+		yumcmd="sudo yum install "$PYTHON_VERSION"-devel gcc gcc-c++ gmp-devel libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y"
+  	else
+		yumcmd="sudo yum install gcc gcc-c++ gmp-devel libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y"
 	fi
 
 elif [ "$(uname)" = "Darwin" ]; then
@@ -109,8 +111,8 @@ else
 	elif [ -e venv/bin/python ] && test "$RHEL_BASED"; then
 		echo "Installing chiavdf dependencies on RedHat/CentOS/Fedora"
 		# Install remaining needed development tools - assumes venv and prior run of install.sh
-		echo "yum install gcc gcc-c++ gmp-devel $PYTHON_DEV_DEPENDENCY libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y" | tr -s ' '
-		sudo yum install gcc gcc-c++ gmp-devel "$PYTHON_DEV_DEPENDENCY" libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y | tr -s ' '
+		echo "$yumcmd"
+		${yumcmd}
 		echo "Installing chiavdf from source on RedHat/CentOS/Fedora"
 		echo venv/bin/python -m pip install --force --no-binary chiavdf "$CHIAVDF_VERSION"
 		venv/bin/python -m pip install --force --no-binary chiavdf "$CHIAVDF_VERSION"
--- a/install-timelord.sh
+++ b/install-timelord.sh
@@ -76,13 +76,20 @@ symlink_vdf_bench() {
 if [ "$(uname)" = "Linux" ] && type apt-get; then
 	UBUNTU_DEBIAN=true
 	echo "Found Ubuntu/Debian."
+
 elif [ "$(uname)" = "Linux" ] && type dnf || yum; then
 	RHEL_BASED=true
 	echo "Found RedHat."
+
+	if [ "$INSTALL_PYTHON_DEV" ]; then
+		PYTHON_DEV_DEPENDENCY="$PYTHON_VERSION"-devel
+	fi
+
 elif [ "$(uname)" = "Darwin" ]; then
 	MACOS=true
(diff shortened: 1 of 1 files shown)


install-timelord.sh
```diff
@@ -102,8 +109,8 @@ else
 	elif [ -e venv/bin/python ] && test "$RHEL_BASED"; then
 		echo "Installing chiavdf dependencies on RedHat/CentOS/Fedora"
 		# Install remaining needed development tools - assumes venv and prior run of install.sh
-		echo "yum install gcc gcc-c++ gmp-devel $PYTHON_DEV_DEPENDENCY libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y"
-		sudo yum install gcc gcc-c++ gmp-devel "$PYTHON_DEV_DEPENDENCY" libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y
+		echo "yum install gcc gcc-c++ gmp-devel $PYTHON_DEV_DEPENDENCY libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y" | tr -s ' '
+		sudo yum install gcc gcc-c++ gmp-devel "$PYTHON_DEV_DEPENDENCY" libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y | tr -s ' '
```
Starttoaster, line 113: I don't have a centos system to test with, but I ran this in a rockylinux container, and I don't think this has the desired effect. It would seem that the `sudo yum` command gets ran as-is before the output of which gets piped to `tr`

Starttoaster: Perhaps something like this would do it?

```bash
yumcmd=$(echo "sudo yum install gcc gcc-c++ gmp-devel $PYTHON_DEV_DEPENDENCY libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y" | tr -s ' ')
echo "$yumcmd"
${yumcmd}
```

What do you think @LeroyINC ?

LeroyINC: good catch... and that makes sense.

how about this to make it more simple

if [ "$(uname)" = "Linux" ] && type apt-get; then
	UBUNTU_DEBIAN=true
	echo "Found Ubuntu/Debian."
elif [ "$(uname)" = "Linux" ] && type dnf || yum; then
	RHEL_BASED=true
	echo "Found RedHat."
	
	if [ "$INSTALL_PYTHON_DEV" ]; then
		yumcmd="sudo yum install "$PYTHON_VERSION"-devel gcc gcc-c++ gmp-devel libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y"
	else
		yumcmd="sudo yum install gcc gcc-c++ gmp-devel libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y"
	fi
	
elif [ "$(uname)" = "Darwin" ]; then
	MACOS=true
	echo "Found MacOS."
fi


then lower down... 

echo "$yumcmd"
${yumcmd}

LeroyINC: updated the code... check it out and let me know

install-timelord.sh
```diff
@@ -76,14 +76,23 @@ symlink_vdf_bench() {
 if [ "$(uname)" = "Linux" ] && type apt-get; then
 	UBUNTU_DEBIAN=true
 	echo "Found Ubuntu/Debian."
+
 elif [ "$(uname)" = "Linux" ] && type dnf || yum; then
 	RHEL_BASED=true
 	echo "Found RedHat."
+
+	if [ "$INSTALL_PYTHON_DEV" ]; then
+		yumcmd="sudo yum install "$PYTHON_VERSION"-devel gcc gcc-c++ gmp-devel libtool make autoconf automake openssl-devel libevent-devel boost-devel python3 cmake -y"
```
Starttoaster, line 85: The use of quotes around `$PYTHON_VERSION` actually unquotes this variable. If you wouldn't mind removing the quotes, I think that's my last piece of feedback before I hit the approve button :) Mostly to appease shellcheck but having it in the quotes is technically correct

install-timelord.sh
Starttoaster: Thanks for the update! Left one more comment, and I'll resolve this one.

install-timelord.sh
LeroyINC: i have fixed this as requested.

Starttoaster: Thank you!

Starttoaster (dismissed):
LGTM

Starttoaster:
Hi @LeroyINC sorry for the late call out on this. I just noticed because of the CI workflow that there are unsigned commits in this PR. Unfortunately, commit signatures are required for contributions, which can be seen in the contributing.md file/policy here https://github.com/Chia-Network/chia-blockchain/blob/main/CONTRIBUTING.md#creating-signed-commits

LeroyINC:
I will see if I can sign this.. not sure how, since it has some other commits\edits after the unsigned commit.

Starttoaster:
If git isn't working with you, feel free to close the PR and make a new one from a new branch and just link it to me in here and we can go from there. :)

LeroyINC:
ok i have signed my commits...

LeroyINC (commented):
signed my commits

Starttoaster (approved):
LGTM

cmmarslender (approved):

PR #15996, opened 2023-08-10
flush only the updated parts of the height-to-hash cache file

arvidn:
This makes flushing scale better as the chain grows.

Amine wrote additional tests for this in a separate PR which has landed (and found a few issues in the first iteration of this patch).

https://github.com/Chia-Network/chia-blockchain/pull/16147
https://github.com/Chia-Network/chia-blockchain/pull/16181


AmineKhaldi (approved):

wjblanke (approved):
aok!

PR #15920, opened 2023-08-02
Fixed python3-venv into install.sh

d1m1trus:
**Ubuntu 23.04**
Fixed Installation aborted due to not being able to install an older version of python3.10-venv

PR #16000, opened 2023-08-10
Remove all install.sh code that installs python and leave it to the user

emlowe:
One approach to keeping the installer working on newer Ubuntu versions and whatnot is to add in the code to handle this. (see #15920 or #15130)

However, a simpler and more maintainable approach is to make the user responsible for installing the proper version of python, sqlite, and openssl

This PR here removes all the code that installs python, sqlite, and openssl and just checks for the existence of versions and exits out if the versions aren't present.

This is easier to manage and allows the installer to work on a variety of platforms including new Ubuntu versions and other more esoteric versions of Linux

This may break some AWS deployment scripts as the code to install and compile python for amazon linux is removed


cmmarslender (dismissed):
Generally in favor of these changes. In addition to simplifying, I think it also gets rid of all the cases of needing `sudo` in the install script, which is a problem for some environments we try to install this. 

There are a couple more things that can be cleaned up if we aren't going to be installing system level packages:

```
PACMAN_AUTOMATED
SKIP_PACKAGE_INSTALL
```
These vars at the top don't end up used anymore


Additionally, there is a whole section that checks and defines `UBUNTU` and `DEBIAN` vars that are no longer used as well with these changes.

d1m1trus:
Using this option will break the entire auto-install mechanism for those who use it.
One of the following options should be implemented:
* Keep the original option that automatically installs all dependencies, and include a new install file that does not install dependencies. For example, install.sh and full-install.sh.
* Remove the dependency installation from the install.sh file and place it in a separate file that can be used before installation to automatically install all dependencies.

This will include a basic installer that does not require sudo, and an advanced version that installs all dependencies for automation purposes.

wallentx:
https://github.com/Chia-Network/chia-blockchain/blob/main/install-gui.sh#L106-L178
It'd be nice to remove all this `sudo` usage in `install-gui.sh`
Given the popularity of `nvm`, and due to the fact that we are installing EOL `nodejs:12` on some distros, it'd be much better if we simply did a min version check, and exit 1 stating that the user needs `nodejs > $MIN_VERSION` installed.

altendky:
I know this is a bit self-centric, but I think it also is born out in the packages that get shipped.  The system generally provides an easily accessible Python that will satisfy our requirements.  The GUI requirements though are generally not provided in remotely recent enough versions.  They are accessible more in the way that pyenv might provide Python where you have to go figure out what tool to acquire, how to use it, how to then create an isolated environment to work in and work in it and...  That is functionality that this script retains for the Python side.  So that's how I convinced myself that not installing Python was reasonable despite suggesting keeping the JS helpers stuff.

Maybe @ChiaMineJP can weigh in on how the GUI script works etc and how nvm compares to npm etc.

cmmarslender (commented):
Looks good to me. Just leaving "comment" for now so others have a chance to review and provide feedback without this getting merged, since I would count for required reviewers

.github/workflows/test-install-scripts.yml
```diff
@@ -141,7 +141,7 @@ jobs:
     - name: Prepare Rocky
       if: ${{ matrix.distribution.type == 'rocky' }}
       run: |
-        yum install --assumeyes git sudo
+        yum install --assumeyes git sudo python39
```
altendky, line 144: I see that `python3` gets you 3.6...  So that doesn't work.  Should we pick 3.8 as our oldest supported version?

install.sh
```diff
@@ -292,6 +150,7 @@ find_sqlite
 echo "SQLite version for Python is ${SQLITE_VERSION}"
 if [ "$SQLITE_MAJOR_VER" -lt "3" ] || [ "$SQLITE_MAJOR_VER" = "3" ] && [ "$SQLITE_MINOR_VER" -lt "8" ]; then
   echo "Only sqlite>=3.8 is supported"
+  echo "Please install sqlite3 per your OS instructions."
```
altendky, line 153: Do CPython packages generally use the system SQLite

install.sh
```diff
@@ -302,6 +161,7 @@ echo "OpenSSL version for Python is ${OPENSSL_VERSION_STRING}"
 if [ "$OPENSSL_VERSION_INT" -lt "269488367" ]; then
   echo "WARNING: OpenSSL versions before 3.0.2, 1.1.1n, or 1.0.2zd are vulnerable to CVE-2022-0778"
   echo "Your OS may have patched OpenSSL and not updated the version to 1.1.1n"
+  echo "We recommend updating to the latest version of OpenSSL available for your OS"
```
altendky, line 164: As above.  I just don't remember this.

altendky (changes requested):

install.sh
hoffmang9: I think this was an old mistake of mine during the emergency SQLite vulnerability release. We should be seeing what SQLite is compiled into the system Python instead of testing the system SQLite - though we don't really need to even do that anymore

altendky: I think we are getting the Python SQLite.  I was questioning the guidance of using OS instructions to address an issue with the version.

`SQLITE_VERSION=$($INSTALL_PYTHON_PATH -c 'import sqlite3; print(sqlite3.sqlite_version)')`

hoffmang9: Yep. The Python query is at least getting the correct data. However we may not even need that data anymore - but probably better to clean that up in a subsequent PR?

altendky: Yes, I wasn't concerned about the check itself.  Just the wording of the message being added here.  I'm chatting with Earle to resolve this.

emlowe: Typically on unix platforms that I'm aware of, the sqlite3 cpython library is dynamically linked to the system sqlite3 library. If you compile python yourself then you could use any version you wanted of course - but typically you will use the system one as well (e.g., pyenv uses the system library unless you do something really custom)
I think in most cases the instruction is ok.

.github/workflows/test-install-scripts.yml
emlowe: I picked 3.9 just so I wouldn't have to come back in a few months and remove 3.8 when it's deprecated...

altendky: The nominal search we already should do ought to catch this.  `3\.?8` or somesuch.

altendky (approved):

arvidn (approved):

PR #16003, opened 2023-08-10
fix mypy warnings from latest version

arvidn:
The main theme of these changes are that types are validated in `dataclasses.replace()` calls.

I think it's reasonable to land this despite the parts of the patch missing test coverage. Its main objective is to correct mypy rules.

### DIFF
--- a/chia/consensus/block_header_validation.py
+++ b/chia/consensus/block_header_validation.py
@@ -198,6 +198,8 @@ def validate_unfinished_header_block(
                         icc_iters_proof,
                         sub_slot.infused_challenge_chain.infused_challenge_chain_end_of_slot_vdf.output,
                     )
+                    if icc_iters_committed is None:
+                        return None, ValidationError(Err.INVALID_ICC_EOS_VDF)
                     if sub_slot.infused_challenge_chain.infused_challenge_chain_end_of_slot_vdf != dataclasses.replace(
                         target_vdf_info,
                         number_of_iterations=icc_iters_committed,
--- a/chia/consensus/constants.py
+++ b/chia/consensus/constants.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import logging
-from typing import Any
+from typing import Any, Dict
 
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.byte_types import hexstr_to_bytes
@@ -80,7 +80,7 @@ class ConsensusConstants:
     # number of consecutive plot ids required to be distinct
     UNIQUE_PLOTS_WINDOW: uint8
 
-    def replace(self, **changes: object) -> "ConsensusConstants":
+    def replace(self, **changes: Any) -> "ConsensusConstants":
         return dataclasses.replace(self, **changes)
 
     def replace_str_to_bytes(self, **changes: Any) -> "ConsensusConstants":
@@ -88,7 +88,7 @@ class ConsensusConstants:
         Overrides str (hex) values with bytes.
         """
 
-        filtered_changes = {}
+        filtered_changes: Dict[str, Any] = {}
         for k, v in changes.items():
             if not hasattr(self, k):
                 # NETWORK_TYPE used to be present in default config, but has been removed
--- a/tests/core/test_farmer_harvester_rpc.py
+++ b/tests/core/test_farmer_harvester_rpc.py
@@ -440,10 +440,12 @@ async def test_farmer_get_harvester_plots_endpoints(
     await wait_for_plot_sync(receiver, last_sync_id)
 
     for page_size in [1, int(total_count / 2), total_count - 1, total_count, total_count + 1, 100]:
-        request = dataclasses.replace(request, page_size=uint32(page_size))
+        # we know that the concrete type of request has a page_size field, but
+        # the type is a Protocol that doesn't, so mypy is right in complaining
+        request = dataclasses.replace(request, page_size=uint32(page_size))  # type: ignore[call-arg]
         expected_page_count = ceil(total_count / page_size)
         for page in range(expected_page_count):
-            request = dataclasses.replace(request, page=uint32(page))
+            request = dataclasses.replace(request, page=uint32(page))  # type: ignore[call-arg]
             await wait_for_synced_receiver(farmer_service._api.farmer, harvester_id)
             page_result = await endpoint(farmer_rpc_client, request)
             offset = page * page_size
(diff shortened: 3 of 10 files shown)


chia/consensus/block_header_validation.py
```diff
@@ -198,6 +198,8 @@ def validate_unfinished_header_block(
                         icc_iters_proof,
                         sub_slot.infused_challenge_chain.infused_challenge_chain_end_of_slot_vdf.output,
                     )
+                    if icc_iters_committed is None:
+                        return None, ValidationError(Err.INVALID_ICC_EOS_VDF)
```
arvidn, line 202: `icc_iters_committed` may not be `None` in the `replace()` call below

chia/consensus/constants.py
```diff
@@ -80,15 +80,15 @@ class ConsensusConstants:
...
+    def replace(self, **changes: Any) -> "ConsensusConstants":
         return dataclasses.replace(self, **changes)
 
     def replace_str_to_bytes(self, **changes: Any) -> "ConsensusConstants":
         """
         Overrides str (hex) values with bytes.
         """
 
-        filtered_changes = {}
+        filtered_changes: Any = {}
```
altendky, line 91: On my phone now but I'll explore this later today.  Not sure why it needs to be quite this unknown what this is.

altendky (changes requested):

chia/consensus/constants.py
arvidn: without this, the call to `replace()` below fails with a type error

arvidn:
```
chia/consensus/constants.py:103: error: Argument 2 to "replace" of "ConsensusConstants" has incompatible type "**dict[str, bytes]"; expected "uint32"  [arg-type]
chia/consensus/constants.py:103: error: Argument 2 to "replace" of "ConsensusConstants" has incompatible type "**dict[str, bytes]"; expected "uint8"  [arg-type]
chia/consensus/constants.py:103: error: Argument 2 to "replace" of "ConsensusConstants" has incompatible type "**dict[str, bytes]"; expected "uint64"  [arg-type]
chia/consensus/constants.py:103: error: Argument 2 to "replace" of "ConsensusConstants" has incompatible type "**dict[str, bytes]"; expected "uint128"  [arg-type]
chia/consensus/constants.py:103: error: Argument 2 to "replace" of "ConsensusConstants" has incompatible type "**dict[str, bytes]"; expected "int"  [arg-type]
chia/consensus/constants.py:103: error: Argument 2 to "replace" of "ConsensusConstants" has incompatible type "**dict[str, bytes]"; expected "bytes32"  [arg-type]
```

chia/consensus/block_header_validation.py
arvidn: this line is not covered by tests. It could probably have been an `assert`, but it seemed a bit more conservative to treat it as a consensus failure

tests/core/test_farmer_harvester_rpc.py
```diff
@@ -440,10 +440,12 @@ def add_plot_directories(prefix: str, count: int) -> List[Path]:
     await wait_for_plot_sync(receiver, last_sync_id)
 
     for page_size in [1, int(total_count / 2), total_count - 1, total_count, total_count + 1, 100]:
-        request = dataclasses.replace(request, page_size=uint32(page_size))
+        # we know that the concrete type of request has a page_size field, but
+        # the type is a Protocol that doesn't, so mypy is right in complaining
```
altendky, line 444: I think this is an issue with our protocol and that should be worked on.  I'll see what I can figure out.

tests/wallet/test_transaction_store.py
```diff
@@ -17,14 +17,19 @@
 from chia.wallet.wallet_transaction_store import WalletTransactionStore, filter_ok_mempool_status
 from tests.util.db_connection import DBConnection
 
-coin_1 = Coin(token_bytes(32), token_bytes(32), uint64(12312))
-coin_2 = Coin(token_bytes(32), token_bytes(32), uint64(1234))
-coin_3 = Coin(token_bytes(32), token_bytes(32), uint64(12312 - 1234))
+
+def rand_hash() -> bytes32:
+    return bytes32(token_bytes(32))
```
altendky, line 22: What was the hinting issue here?

tests/wallet/test_transaction_store.py
```diff
@@ -17,14 +17,19 @@
...
+
+coin_1 = Coin(rand_hash(), rand_hash(), uint64(12312))
+coin_2 = Coin(rand_hash(), rand_hash(), uint64(1234))
+coin_3 = Coin(rand_hash(), rand_hash(), uint64(12312 - 1234))
 
 tr1 = TransactionRecord(
     uint32(0),  # confirmed height
     uint64(1000),  # created_at_time
-    bytes32(token_bytes(32)),  # to_puzzle_hash
+    bytes32(rand_hash()),  # to_puzzle_hash
```
altendky, line 32: There are at least a few `bytes32()` left here that are unneeded.

altendky (commented):
There are a lot of new conversions to fixed width integers.  They seem good offhand but I should at least raise the question of whether or not we respond as intended (or as before at least) if they end up raising exceptions for out of bounds inputs.

chia/consensus/block_header_validation.py
altendky: Yeah, I noticed you didn't pile it into the existing asserts and I appreciate the consideration of what you did add.  The structure just feels like we're missing something else to identify what case we're in etc.  But, I expect that even if I am correct that addressing that situation would be entirely separate from this PR.  Thanks.

chia/consensus/constants.py
arvidn: I took this change (which also required adding `Dict` to the import statement)

chia/consensus/constants.py
```diff
@@ -80,15 +80,15 @@ class ConsensusConstants:
     # number of consecutive plot ids required to be distinct
     UNIQUE_PLOTS_WINDOW: uint8
 
-    def replace(self, **changes: object) -> "ConsensusConstants":
+    def replace(self, **changes: Any) -> "ConsensusConstants":
```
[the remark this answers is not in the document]
arvidn: I'm not opposed to this. But I would punt on that

tests/wallet/test_transaction_store.py
arvidn: `bytes` vs `bytes32`. Every call to `token_bytes(32)` needed to pass a `bytes32` object

tests/core/test_farmer_harvester_rpc.py
arvidn: I don't know how to express those `dataclass` properties as a Protocol. We already declare that we have *read only* fields with those names, but I suppose `replace()` will also need to be able to construct a new object with new values somehow

chia/consensus/block_header_validation.py
arvidn: we have some non-trivial invariants of the optionals in our FulBlock

PR #16004, opened 2023-08-10
Don't farm compressed plots if context count is 0

### LINKED ISSUE #15949: [Bug] 2.0.0-rc3 silently stops sending partials to a pool after a few minutes if a compressed plot is present
### What happened?

Running 2.0.0-rc3 on Debian 12. The client starts normally (though it does complain it can't initialise GPU compression), it starts sending partials to the pool as normal. If any compressed plots are present in any of the locations, it will silently stop sending partials to the pool after about 10 minutes, and doesn't recover. Externally the processes seem fine, all of them respond, blockchain and wallet stay synced. If I remove the compressed plots and restart, it continues on normally without issue. There's no hardware issues, no kernel log errors, Flexfarmer can continue without issue, albeit by ignoring the compressed plots.

The compressed plots were made with Bladebit Cuda v3-rc1, also on Debian 12.

This can be reproduced without fail every time. If I start it with uncompressed only plots, and then add the compressed ones location, it will silently fail, again after about 10 minutes.

This occurred on rc2 as well. 

The logs don't show anything immediately obviously of use.

I've done completely fresh installs of the client from source and .deb package, both show the same, as well as starting with a completely fresh config.yaml and freshly added keys.

### Version

2.0.0-rc3

Linux

CLI

### Relevant log output

```shell
The logs don't show anything immediately obviously of use.
```

fchirica:

### DIFF
--- a/chia/plotting/manager.py
+++ b/chia/plotting/manager.py
@@ -46,6 +46,7 @@ class PlotManager:
     _refresh_callback: Callable
     _initial: bool
     max_compression_level_allowed: int
+    context_count: int
 
     def __init__(
         self,
@@ -79,6 +80,7 @@ class PlotManager:
         self._refresh_callback = refresh_callback
         self._initial = True
         self.max_compression_level_allowed = 0
+        self.context_count = 0
 
     def __enter__(self):
         self._lock.acquire()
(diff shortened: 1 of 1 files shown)


emlowe (approved):

emlowe (changes requested):
Doesn't seem to work in my testing

emlowe (approved):
Works as expected for me - loads of complaints on every refresh interval

PR #15948, opened 2023-08-06
optimize blockchain DB for Chip-13

arvidn:
This patch is meant to improve performance pulling `BlockRecord`s from the `BlockStore` database table.
Currently, sqlite appears to allocate a very large buffer. Over time this memory usage may cause swapping and stalled disk I/O, delaying proof lookups.

Another issue with the current code is that we look up blocks from the database even when we have them in the cache.

This memory usage of sqlite isn't measured by the benchmark, but there is a slight time improvement of not having to load the full blocks into memory as well.

| function | before | after | after / before |
| --- | --- | --- | --- |
| get_block_records_by_hash | 2.6470s | 2.4786s | 93.6 % |
| get_block_record | 3.4475s | 2.4806s | 72 % |
| get_block_records_in_range | 0.5901s | 0.3550s | 60 % |

# This patch
```
19.5477s, add_full_block
2.0765s, get_block_info
1.1192s, get_generator
5.5172s, get_full_block
2.0444s, get_full_block_bytes
0.9421s, get_generators_at
5.5845s, get_full_blocks_at
2.4786s, get_block_records_by_hash
1.7986s, get_block_bytes_by_hash
5.5716s, get_blocks_by_hash
2.4806s, get_block_record
0.3550s, get_block_records_in_range
0.0034s, get_block_records_close_to_peak
4.4870s, is_fully_compactified
13.2819s, get_random_not_compactified
all tests completed in 67.2883s
database size: 96.518 MB
```

# main

```
14.9223s, add_full_block
2.0637s, get_block_info
1.0842s, get_generator
5.4009s, get_full_block
1.6925s, get_full_block_bytes
0.3226s, get_generators_at
5.0399s, get_full_blocks_at
2.6470s, get_block_records_by_hash
1.8506s, get_block_bytes_by_hash
5.5557s, get_blocks_by_hash
3.4475s, get_block_record
0.5901s, get_block_records_in_range
0.0091s, get_block_records_close_to_peak
3.5308s, is_fully_compactified
13.8417s, get_random_not_compactified
all tests completed in 61.9986s
database size: 93.503 MB
```

PR #16007, opened 2023-08-10
fix bug in recent BlockStore optimization patch

arvidn:
and fix the issue with the test that allowed it through.

The problem of using a regular `JOIN` is that the `plot_info` table is populated lazily, and starts out empty. With an `INNER JOIN`, no rows will be returned.

The test would have caught this if I had deleted the rows correctly.


hoffmang9 (approved):
Tested and running as a fix to the issue I reported from https://github.com/Chia-Network/chia-blockchain/pull/15948

emlowe (approved):

PR #16017, opened 2023-08-11
Deprecate minting CATs via RPC

Quexington:
### Purpose:

This RPC was never really meant to be official.  It's a holdover from the first iteration of CATs that is no longer supported.  This PR deprecates the RPC path that would have triggered it, and adds a new path that is clearly marked for testing (because people have found use for it in that respect).  The supported way to mint CATs now is via the [CAT Admin Tool](https://github.com/Chia-Network/CAT-admin-tool).

### Current Behavior:

This RPC works, and mints a single use CAT.

### New Behavior:

The old call to this RPC will return an error notifying of the deprecation.  A new call under the `"mode": "new test"` has been added.


emlowe:
Not sure I follow this.
You "removed" (not deprecated) passing in "new" - but the exact same code runs if I pass in "new test".
I think what you are trying to do is discourage people from making CATs using this RPC while at the same time allowing people to continue to make CATs with this RPC.

Quexington:
> Not sure I follow this. You "removed" (not deprecated) passing in "new" - but the exact same code runs if I pass in "new test". I think what you are trying to do is discourage people from making CATs using this RPC while at the same time allowing people to continue to make CATs with this RPC.

Yeah that's the idea.  People have expressed before that they like to use this RPC to mint CATs instead of the CAT admin tool when they're testing and just need some balance of CATs.  We also use this RPC a lot in the RPC tests. That's why we didn't remove this a long time ago, so now the point is that you have to acknowledge that you're doing testing related stuff when you use this RPC (people shouldn't build on it).

> You "removed" (not deprecated) passing in "new"
Not "removed", it still exists, it just throws a deprecation error.  (Is that not what deprecation means?)

paninaro:
Often, marking an API as deprecated doesn't actually break the functionality, but signals to the developer that the API _will_ be removed in the future. I don't think we really have a mechanism for supporting that (warning in the IDE for instance), so I'm ok with this approach. I'd would probably say that we're making this API obsolete, but that's just being pedantic.

The GUI does expose this mint CAT functionality through WalletConnect, so we'll need to make a change there, but this change seems fine to me (though I agree with Earle that the mode renaming is a little weird)

chia/rpc/wallet_rpc_api.py
```diff
@@ -636,7 +636,7 @@ async def create_new_wallet(self, request: Dict) -> EndpointResult:
         if request["wallet_type"] == "cat_wallet":
             # If not provided, the name will be autogenerated based on the tail hash.
             name = request.get("name", None)
-            if request["mode"] == "new":
+            if request["mode"] == "new test":
```
paninaro, line 639: My recommendation would be to avoid overloading the `"mode"` by defining a new value, and instead, add a new `"test": true` key/value. Then the `request["mode"] == "new"` branch can remain the same but be dependent on the test key.

If that's not workable, I would at least change the mode to "new_test" instead of starting a pattern of including spaces in param values.

paninaro (changes requested):

emlowe:
So it's even stranger now - we expose this via WalletConnect? I mean this doesn't seem workable - one can't use the catadmin tool with WalletConnect ? So that seems odd the recommended way forward doesn't work with WC.

We could add into the response JSON a `deprecated: true` and `message: "blah blah"` And NOT block the RPC from working. Until some future release - but see WC above.

But I don't see the point of making someone add "magic foo" to the RPC call which then does the same thing it used to do without the "magic foo".

Quexington:
> So it's even stranger now - we expose this via WalletConnect? I mean this doesn't seem workable - one can't use the catadmin tool with WalletConnect ? So that seems odd the recommended way forward doesn't work with WC.

The point is that they should not have been using this RPC via WC either, it's _been_ a test RPC, we just haven't made that apparent really at all.

> We could add into the response JSON a `deprecated: true` and `message: "blah blah"` And NOT block the RPC from working. Until some future release - but see WC above.

I mean the CAT admin tool has been around for a while and has been touted as the official way since CAT launch, this RPC just still hangs around out of convenience for testing.  I'm not sure the amount/kind of use it's getting is worth following this whole process for.

> But I don't see the point of making someone add "magic foo" to the RPC call which then does the same thing it used to do without the "magic foo".

Because, again, the RPC _is_ useful but only for testing.  It makes sense to leave it available, but it needs positive affirmation from the person calling it that they're not using it for production code.

chia/rpc/wallet_rpc_api.py
```diff
@@ -637,17 +637,23 @@ async def create_new_wallet(self, request: Dict) -> EndpointResult:
...
             name = request.get("name", None)
             if request["mode"] == "new":
-                async with self.service.wallet_state_manager.lock:
-                    cat_wallet: CATWallet = await CATWallet.create_new_cat_wallet(
-                        wallet_state_manager,
-                        main_wallet,
-                        {"identifier": "genesis_by_id"},
-                        uint64(request["amount"]),
-                        name,
+                if request["test"]:
```
paninaro, line 640: Maybe make this optional? `request.get("test", False)`

Quexington: Ah yeah good call out (fixed)

emlowe:
So my position is that we have neither dropped nor deprecated this RPC - it continues to work and will continue to work going forward

So if we want to tell people something, I think we should tell them expressively:

```    
raise ValueError(
    "This is a test-only RPC - please include "test": true in your request. "
    "Do not use for production CATs. "
    "Use the CAT Admin Tool @ https://github.com/Chia-Network/CAT-admin-tool."
)
```

I'd be ok with that kind of language

paninaro (approved):

PR #16029, opened 2023-08-14
Improve testnet connectivity.

fchirica:
Improve network connectivity by restricting peers running default network port of another network (i.e. testnet nodes running 8444). additionally, we don't propagate those peers either.

### DIFF
--- a/chia/server/node_discovery.py
+++ b/chia/server/node_discovery.py
@@ -469,6 +469,10 @@ class FullNodeDiscovery:
                 async with self.address_manager.lock:
                     self.address_manager.cleanup(max_timestamp_difference, max_consecutive_failures)
 
+    def _peer_has_wrong_network_port(self, port: uint16):
+        # Check if the peer is having the default port of a network different than ours.
+        return port in NETWORK_ID_DEFAULT_PORTS.values() and port != self.default_port
+
     async def _add_peers_common(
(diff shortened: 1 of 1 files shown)


emlowe:
Just to check, this doesn't require me to run mainnet on 8444 right? I can still run mainnet or testnet on any port I want with the exception of the default port for another network.
That is, I can run mainnet on anything OTHER THAN 58444
And I can run testnet on anything OTHER THAN 8444

hoffmang9:
It looks like if you set your mainnet port to 8446 then you wouldn't gossip and 8444 addresses - but maybe I'm missing something

fchirica:
> It looks like if you set your mainnet port to 8446 then you wouldn't gossip and 8444 addresses - but maybe I'm missing something

in case the port is 8446, it'll miss from `NETWORK_ID_DEFAULT_PORTS`, making `_peer_has_wrong_network_port` return False. so the code will work as before in that case.

emlowe:
The idea is that every port should work for `mainnet` EXCEPT 58444 which is the default `testnet` port. And every port should work for `testnet` EXCEPT 8444 which is the default `mainnet` port.

But at first read, I had the same question.

emlowe:
Should I expect my node to continue to try to connect to these peers? I am testing this branch and I still see my `testnet` 
node trying to connect to other peers on 8448

emlowe (approved):
Approved, but this will take some time before it actually helps anyone

PR #16037, opened 2023-08-15
wallet: Fix comment in `WalletNode.validate_received_state_from_peer`

xdustinface:
### Purpose:

Just fix an outdated comment (`return_old_states` doesn't longer exist).

### New Behavior:

No change in behaviour.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -1302,8 +1302,7 @@ class WalletNode:
         fork_height: Optional[uint32],
     ) -> bool:
         """
-        Returns all state that is valid and included in the blockchain proved by the weight proof. If return_old_states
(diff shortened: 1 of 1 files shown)


chia/wallet/wallet_node.py
```diff
@@ -1302,8 +1302,7 @@ async def validate_received_state_from_peer(
         fork_height: Optional[uint32],
     ) -> bool:
         """
-        Returns all state that is valid and included in the blockchain proved by the weight proof. If return_old_states
-        is False, only new states that are not in the coin_store are returned.
+        Returns all states that are valid and included in the blockchain proved by the weight proof.
```
arvidn, line 1305: the function returns a `bool`, not "all states". It's not obvious that the updated version of the comment is correct

arvidn (approved):

PR #16054, opened 2023-08-15
Add ergonomic timelock parsing to RPCs

Quexington:
This adds a convenience to the RPC API/Client for specifying expirations/delays more ergonomically than directly specifying the conditions.

### DIFF
--- a/chia/rpc/util.py
+++ b/chia/rpc/util.py
@@ -8,7 +8,7 @@ import aiohttp
 
 from chia.types.blockchain_format.coin import Coin
 from chia.util.json_util import obj_to_response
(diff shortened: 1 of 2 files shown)


geoffwalmsley (approved):

arvidn:
I don't see any tests. do we not have tests for the RPC clients?

wjblanke (approved):
aok

Quexington:
> I don't see any tests. do we not have tests for the RPC clients?

We do, but this kind of automatically gets test coverage due to the fact that it's a default argument, streamable, and since the behavior in the client is to send the dictionary with a bunch of `None` values across no matter what, the epi end gets tested as well.

PR #16055, opened 2023-08-15
Fix Core Test Flake

jack60612:
### Purpose:
fixes test flake by changing to the formerly racey random port finder

### Current Behavior:
N/A

### New Behavior:
N/A

### Testing Notes:
Fixes race condition by making global fixture, and adding option to check ipv6

### DIFF
--- a/chia/seeder/dns_server.py
+++ b/chia/seeder/dns_server.py
@@ -276,7 +276,8 @@ class DNSServer:
     # TODO: After 3.10 is dropped change to asyncio.Server
     tcp_server: Optional[asyncio.base_events.Server] = field(init=False, default=None)
     # these are all set in __post_init__
-    dns_port: int = field(init=False)
+    tcp_dns_port: int = field(init=False)
+    udp_dns_port: int = field(init=False)
     db_path: Path = field(init=False)
     domain: DomainName = field(init=False)
     ns1: DomainName = field(init=False)
@@ -292,13 +293,15 @@ class DNSServer:
         """
         We initialize all the variables set to field(init=False) here.
         """
-        # From Config
-        self.dns_port: int = self.config.get("dns_port", 53)
-        # DB Path
+        # From Config:
+        # The dns ports should only really be different if testing.
+        self.tcp_dns_port: int = self.config.get("dns_port", 53)
+        self.udp_dns_port: int = self.config.get("dns_port", 53)
+        # DB Path:
         crawler_db_path: str = self.config.get("crawler_db_path", "crawler.db")
         self.db_path: Path = path_from_root(self.root_path, crawler_db_path)
         self.db_path.parent.mkdir(parents=True, exist_ok=True)
-        # DNS info
+        # DNS info:
         self.domain: DomainName = DomainName(self.config["domain_name"])
         if not self.domain.endswith("."):
             self.domain = DomainName(self.domain + ".")  # Make sure the domain ends with a period, as per RFC 1035.
@@ -330,23 +333,23 @@ class DNSServer:
 
         # One protocol instance will be created for each udp transport, so that we can accept ipv4 and ipv6
         self.udp_transport_ipv6, self.udp_protocol_ipv6 = await loop.create_datagram_endpoint(
-            lambda: UDPDNSServerProtocol(self.dns_response), local_addr=("::0", self.dns_port)
+            lambda: UDPDNSServerProtocol(self.dns_response), local_addr=("::0", self.udp_dns_port)
         )
         self.udp_protocol_ipv6.start()  # start ipv6 udp transmit task
 
-        # in case the port is 0 we need all protocols on the same port.
-        self.dns_port = self.udp_transport_ipv6.get_extra_info("sockname")[1]  # get the actual port we are listening to
+        # in case the port is 0, we get the real port
+        self.udp_dns_port = self.udp_transport_ipv6.get_extra_info("sockname")[1]  # get the port we binded to
 
         if sys.platform.startswith("win32") or sys.platform.startswith("cygwin"):
             # Windows does not support dual stack sockets, so we need to create a new socket for ipv4.
             self.udp_transport_ipv4, self.udp_protocol_ipv4 = await loop.create_datagram_endpoint(
-                lambda: UDPDNSServerProtocol(self.dns_response), local_addr=("0.0.0.0", self.dns_port)
+                lambda: UDPDNSServerProtocol(self.dns_response), local_addr=("0.0.0.0", self.udp_dns_port)
             )
             self.udp_protocol_ipv4.start()  # start ipv4 udp transmit task
 
         # One tcp server will handle both ipv4 and ipv6 on both linux and windows.
         self.tcp_server = await loop.create_server(
-            lambda: TCPDNSServerProtocol(self.dns_response), ["::0", "0.0.0.0"], self.dns_port
+            lambda: TCPDNSServerProtocol(self.dns_response), ["::0", "0.0.0.0"], self.tcp_dns_port
         )
 
         log.info("DNS server started.")
--- a/chia/seeder/dns_server.py
+++ b/chia/seeder/dns_server.py
@@ -338,7 +338,7 @@ class DNSServer:
         self.udp_protocol_ipv6.start()  # start ipv6 udp transmit task
 
         # in case the port is 0, we get the real port
-        self.udp_dns_port = self.udp_transport_ipv6.get_extra_info("sockname")[1]  # get the port we binded to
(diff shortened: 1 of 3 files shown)


wallentx (dismissed):

chia/simulator/socket.py
```diff
@@ -7,17 +7,28 @@
...
         if port in recent_ports:
             continue
 
-        with socket.socket() as s:
-            try:
+        try:
+            with socket.socket(type=socket.SOCK_STREAM) as s:  # TCP
+                s.bind(("127.0.0.1", port))
+            with socket.socket(type=socket.SOCK_DGRAM) as s:  # UDP
                 s.bind(("127.0.0.1", port))
```
altendky, line 22: We're forcing all existing uses to open UDP ports as well?

altendky (dismissed):

chia/simulator/socket.py
jack60612: yeah, i can make that separate.

tests/conftest.py
```diff
@@ -995,3 +1000,13 @@ async def have_connections() -> bool:
 @pytest.fixture(name="database_uri")
 def database_uri_fixture() -> str:
     return f"file:db_{random.randint(0, 99999999)}?mode=memory&cache=shared"
+
+
+@pytest.fixture(name="available_listen_port", scope="session")
+def available_listen_port() -> int:
```
altendky, line 1006: The idea with the explicit name specification is so that the function can have a different name.  This makes it harder to accidentally reference the fixture function without having depended on it as a fixture.  This is generally just an issue within the same file.

```suggestion
def available_listen_port_fixture() -> int:
```

tests/conftest.py
```diff
@@ -995,3 +1000,13 @@ async def have_connections() -> bool:
...
+    return find_available_listen_port()
+
+
+@pytest.fixture(name="available_ipv6_listen_port", scope="session")
+def available_ipv6_listen_port() -> int:
```
altendky, line 1011:
```suggestion
def available_ipv6_listen_port_fixture() -> int:
```

tests/conftest.py
jack60612: ah makes sense.

tests/conftest.py
```diff
@@ -995,3 +1000,13 @@ async def have_connections() -> bool:
 @pytest.fixture(name="database_uri")
 def database_uri_fixture() -> str:
     return f"file:db_{random.randint(0, 99999999)}?mode=memory&cache=shared"
+
+
+@pytest.fixture(name="available_listen_port", scope="session")
```
altendky, line 1005: `scope="session"` means it will be run once and that same result will be used in every test that needs it.  This doesn't seem good.  Am I missing something?

altendky: I suppose I should caveat that I'm not sure with pytest-xdist if this ends up being once per session or once per process.  Either way doesn't seem great.

jack60612: oh i forgot about the caching, i was trying to get the recent_ports thing to be shared per session...

altendky: `recent_ports` is already a global (within the process).

jack60612: > `recent_ports` is already a global (within the process).

i assumed pytest would somehow mess that up but ill set the scope to per function.

chia/simulator/socket.py
```diff
@@ -7,20 +7,29 @@
 recent_ports: Set[int] = set()
 
 
-def find_available_listen_port(name: str = "free") -> int:
+def find_available_listen_port(name: str = "free", ipv6: bool = False, udp: bool = False) -> int:
     global recent_ports
 
     while True:
```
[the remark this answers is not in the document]
jack60612: completely removed

chia/simulator/setup_services.py
```diff
@@ -215,14 +215,16 @@ async def setup_crawler(
...
+    root_path_populated_with_config: Path, database_uri: str, dns_port: int
+) -> AsyncGenerator[DNSServer, None]:
     config = load_config(root_path_populated_with_config, "config.yaml")
     service_config = config["seeder"]
 
     service_config["selected_network"] = "testnet0"
     if service_config["domain_name"].endswith("."):  # remove the trailing . so that we can test that logic.
         service_config["domain_name"] = service_config["domain_name"][:-1]
-    service_config["dns_port"] = 0
+    service_config["dns_port"] = dns_port
```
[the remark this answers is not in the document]
jack60612: completely removed

tests/conftest.py
```diff
@@ -995,3 +1000,23 @@ async def have_connections() -> bool:
...
+
+
+@pytest.fixture(name="tcp_dualstack_listen_port", scope="function")
+def tcp_dualstack_listen_port_fixture() -> int:
+    return find_available_listen_port(ipv6=True)
+
+
+@pytest.fixture(name="tcp_udp_ipv4_listen_port", scope="function")
+def tcp_udp_ipv4_listen_port_fixture() -> int:
+    return find_available_listen_port(udp=True)
```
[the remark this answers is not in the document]
jack60612: completely removed

chia/seeder/dns_server.py
```diff
@@ -330,23 +333,23 @@ async def run(self) -> AsyncIterator[None]:
...
         self.udp_transport_ipv6, self.udp_protocol_ipv6 = await loop.create_datagram_endpoint(
-            lambda: UDPDNSServerProtocol(self.dns_response), local_addr=("::0", self.dns_port)
+            lambda: UDPDNSServerProtocol(self.dns_response), local_addr=("::0", self.udp_dns_port)
         )
         self.udp_protocol_ipv6.start()  # start ipv6 udp transmit task
 
-        # in case the port is 0 we need all protocols on the same port.
-        self.dns_port = self.udp_transport_ipv6.get_extra_info("sockname")[1]  # get the actual port we are listening to
+        # in case the port is 0, we get the real port
+        self.udp_dns_port = self.udp_transport_ipv6.get_extra_info("sockname")[1]  # get the port we binded to
```
arvidn, line 341:
```suggestion
        self.udp_dns_port = self.udp_transport_ipv6.get_extra_info("sockname")[1]  # get the port we bound to
```

tests/core/test_seeder.py
```diff
@@ -33,6 +33,17 @@ def generate_test_combs() -> List[Tuple[bool, str, dns.rdatatype.RdataType]]:
...
 
+def get_dns_port(dns_server: DNSServer, use_tcp: bool, target_addr: str) -> int:
+    if use_tcp:
+        assert dns_server.tcp_server is not None
+        tcp_sockets = dns_server.tcp_server.sockets
+        wanted_addr = "::" if target_addr == "::1" else "0.0.0.0"
+        if tcp_sockets[0].getsockname()[0] == wanted_addr:
+            return int(tcp_sockets[0].getsockname()[1])
+        return int(tcp_sockets[1].getsockname()[1])
+    return dns_server.udp_dns_port
```
arvidn, line 44: this function looks suspicious. why is the `use_tcp` path not looking at teh `tcp_dns_port` field?
it would also seem simpler to just have this function always return both ports instead of having the caller pick one

tests/core/test_seeder.py
```diff
@@ -269,6 +281,7 @@ async def test_db_processing(seeder_service: DNSServer) -> None:
 
     # now we check all the combinations once (not a stupid amount of times)
     for use_tcp, target_address, request_type in all_test_combinations:
```
arvidn, line 283: this looks a bit controversial, instead of using `pytest` parametrization

tests/core/test_seeder.py
arvidn: and UDP always gets the same port for v4 and v6?

jack60612: Yep

tests/core/test_seeder.py
altendky: That's what module scope fixtures or such are for?

altendky: If there's no updating then maybe a module scoped fixture used by a function scoped fixture that creates a transaction and then makes sure to revert it?  This would help isolate tests just in case while retaining an optimization.

jack60612: > If there's no updating then maybe a module scoped fixture used by a function scoped fixture that creates a transaction and then makes sure to revert it? This would help isolate tests just in case while retaining an optimization.

the one issue with making a module scope fixture is that it will depend on a function scoped fixture, as i don't think it will reuse the same file.

jack60612: i changed it to mostly use cached data but the tests still take about 2 seconds each.

arvidn (approved):

PR #16061, opened 2023-08-15
Remove extra check in send_transaction_multi for chia coins split

### LINKED ISSUE #15527: When running Wallet only mode, unable to run Split coins command[Bug]
### What happened?

When trying to run the `Split` coins command when running in Wallet only mode, it throws an error that the wallet needs to be in sync, when switching to farming mode, it works as intended, reported by user in discord https://discord.com/channels/1034523881404370984/1034870571864948777/1118370242066923652

### Version

1.8.1

Linux

GUI

### Relevant log output

```shell
chia wallet coins split -n 1 -a .01 -t 0x2f663a3298190844cfef29c86a0c82e096a72bcec3cc1e90fe705d1d13351e1f  Exception from 'wallet' {'error': 'Wallet needs to be fully synced before sending transactions', 'success': False}:
  File "chia/cmds/cmds_util.py", line 116, in get_any_service_client
  File "chia/cmds/cmds_util.py", line 235, in execute_with_wallet
  File "chia/cmds/coin_funcs.py", line 219, in async_split
  File "chia/rpc/wallet_rpc_client.py", line 214, in send_transaction_multi
  File "chia/rpc/rpc_client.py", line 61, in fetch
ValueError: {'error': 'Wallet needs to be fully synced before sending transactions', 'success': False}
```

jack60612:
### Purpose:
This removes a sync check from the wallet rpc `send_transaction_multi`

### Current Behavior:
N/A

### New Behavior:
N/A

### Testing Notes:
Applicable test coverage exists, however this issue is hard to replicate and just happens, and this is the best way to fix it

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1059,9 +1059,6 @@ class WalletRpcApi:
         }
 
     async def send_transaction_multi(self, request) -> EndpointResult:
-        if await self.service.wallet_state_manager.synced() is False:
-            raise ValueError("Wallet needs to be fully synced before sending transactions")
-
         wallet_id = uint32(request["wallet_id"])
         wallet = self.service.wallet_state_manager.wallets[wallet_id]
 


xdustinface (changes requested):
This doesn't feel like the fix for the problem and it anyway will still run into the same issue in `cat_spend` here 

https://github.com/Chia-Network/chia-blockchain/blob/7557b6cbdf52b81fc6d2d6e84716fdc95ec5b4c9/chia/rpc/wallet_rpc_api.py#L1071

which does https://github.com/Chia-Network/chia-blockchain/blob/7557b6cbdf52b81fc6d2d6e84716fdc95ec5b4c9/chia/rpc/wallet_rpc_api.py#L1539-L1540

I think we should instead make sure `WalletStateManager.synced` reliably knows when to return `True` and my guess is that the issue is related to the queue not being empty when the requests are triggered. Maybe because it in wallet only mode in those cases it uses an untrusted node which does the secondary sync always and keeps items in the queue or so. https://github.com/Chia-Network/chia-blockchain/blob/7557b6cbdf52b81fc6d2d6e84716fdc95ec5b4c9/chia/wallet/wallet_state_manager.py#L576-L579 It feels wrong to make the sync state also dependent on the queue which repeatedly might get filled even if we are "synced" because that probably also will lead to the sync state flipping back and forth, i guess. Maybe we just check the number of queue entries or so instead to make it at least more reliable while not flipping to synced too early but it feels like something which needs some more investigation.

emlowe:
It does look like one could also remove the check from `cat_spend` however as well.  Neither of these calls depend on anything in the wallet needing to be synced. They are simple constructing the bundle from the supplied options.

So I think removing this check is ok, but you do need to remove it from `cat_spend` as well

It is a much deeper question as to understanding how the wallet is synced or not

PR #16062, opened 2023-08-16
full_node: Stop updating wallets during long sync

xdustinface:
### Purpose:

I think its not necessary and only leads to pointless DB lookups and API calls, also from the wallet back to the node in few places like here https://github.com/Chia-Network/chia-blockchain/blob/7b9801fa7f8be84f528f8aa0d09eeb685676053a/chia/wallet/wallet_node.py#L1059 or inside `WalletNode.is_peer_synced` here https://github.com/Chia-Network/chia-blockchain/blob/7b9801fa7f8be84f528f8aa0d09eeb685676053a/chia/wallet/wallet_node.py#L1079

only to throw it away here (after even disconnecting in untrusted case) https://github.com/Chia-Network/chia-blockchain/blob/7b9801fa7f8be84f528f8aa0d09eeb685676053a/chia/wallet/wallet_node.py#L1086 because the node isn't synced.

Note that this doesn't have an impact on short syncs and we also still send a `new_peak_wallet` after connecting and there will be one right after the long sync was done in here https://github.com/Chia-Network/chia-blockchain/blob/7b9801fa7f8be84f528f8aa0d09eeb685676053a/chia/full_node/full_node.py#L1021

### Current Behavior:

The node sends new peak messages / wallet updates during long sync.

### New Behavior:

The only sends one peak update right after the long sync is done which then can trigger the wallet to start the sync.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -1091,14 +1091,12 @@ class FullNode:
                     advanced_peak = True
                     assert peak is not None
                     # Hints must be added to the DB. The other post-processing tasks are not required when syncing
-                    hints_to_add, lookup_coin_ids = get_hints_and_subscription_coin_ids(
+                    hints_to_add, _ = get_hints_and_subscription_coin_ids(
                         state_change_summary,
                         self.subscriptions.has_coin_subscription,
                         self.subscriptions.has_ph_subscription,
                     )
                     await self.hint_store.add_hints(hints_to_add)
-                    await self.update_wallets(state_change_summary, hints_to_add, lookup_coin_ids)
-                await self.send_peak_to_wallets()
                 self.blockchain.clean_block_record(end_height - self.constants.BLOCKS_CACHE_SIZE)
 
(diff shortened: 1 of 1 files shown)


arvidn:
does the wallet sync in parallel with the full node right now?
If so, it would seem this change would prevent that, forcing them to sync one at a time, in serial

arvidn:
does the wallet sync in parallel with the full node right now?
If so, it would seem this change would prevent that, forcing them to sync one at a time, in serial.

Did you manually test syncing the wallet and node at the same time with this patch, to make sure it still works?

xdustinface:
> does the wallet sync in parallel with the full node right now? If so, it would seem this change would prevent that, forcing them to sync one at a time, in serial.
> 
> Did you manually test syncing the wallet and node at the same time with this patch, to make sure it still works?

Like states in the PR description the wallet throws away the new peaks if the full node isn't synced.

arvidn (approved):
I suppose if we ever want to support syncing the wallet in parallel with the node we'd have to add something like this back.

PR #16071, opened 2023-08-16
Parallel CI

arvidn:
### Purpose:

Speed up CI and catch tests that don't support being run in parallel, before landing them.

### Current Behavior:

Most tests run with `-n 0`

### New Behavior:

Most test run with `-n 2`

### DIFF
--- a/tests/core/cmds/config.py
+++ b/tests/core/cmds/config.py
@@ -1,3 +1,0 @@
-from __future__ import annotations
-
-parallel = True
--- a/tests/core/consensus/config.py
+++ b/tests/core/consensus/config.py
@@ -1,3 +1,0 @@
-from __future__ import annotations
-
-parallel = True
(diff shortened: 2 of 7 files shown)


tests/build-job-matrix.py
```diff
@@ -102,9 +102,9 @@ def update_config(parent: Dict[str, Any], child: Dict[str, Any]) -> Dict[str, An
 
     # TODO: design a configurable system for this
     process_count = {
-        "macos": {False: 0, True: 4}.get(conf["parallel"], conf["parallel"]),
-        "ubuntu": {False: 0, True: 4}.get(conf["parallel"], conf["parallel"]),
-        "windows": {False: 0, True: 2}.get(conf["parallel"], conf["parallel"]),
+        "macos": {False: 0, True: 4}.get(conf["parallel"], 4),
```
emlowe, line 105: It seems to me that if `parallel` is missing from `conf` - I don't see how this works either old or new?
If it's there, it should be either true or false, so I'm confused how this changes anything?

emlowe (approved):

Issue #16052, opened 2023-08-15
[Bug] VCs - vcs get doesn't show proof reveals available in the database


esaung:
### What happened?

running vcs get doesn't show proofs.

trying to add an proof reveal that's already in the database returns a code error

looking up proof reveals by root hash shows the values are there.

I'm expecting to see the proofs and the root hash when running the vcs get command. I'm also expecting to see a more user friendly error in the CLI when trying to add proof_reveals that are already in the proofs database.

### Version

2.0.0 rc4

### What platform are you using?

macOS

### What ui mode are you using?

CLI

### Relevant log output

```shell
(venv) esaung@Chia-Macbook chia-blockchain-gui % chia wallet vcs get

Active Wallet Key (*):
   -Fingerprint:           87074341
   -Label:                 Issuer (Issuer)
   -Sync Status:           Synced

Wallet Keys:
1)    3830434144   main farm
2)    1645933707   esaung hot wallet
3)    2920272449   TN10 - Custody
4)    3049838316   Documentation demo wallet tn10
5)    1964611303   pawket tn10 wallet
6)    233773231    NFT wallet
7)    2618175902   Background wallet
8)    2780699030   test walleth tn10
9)    1783862898   mainnet exp
10)   2803029382   TN10 main wallet
11)   1498393629   TN10 CF bag wallet
12)   965505910    tn10 chia nft use case test wallet
13)   852356298    nucle wallet
14)   107923803    mainnet test wallet
15)   2691838111   Big Big of NFts
16)   2867337490   5k tn10 NFT test wallet
17)   2752540257   Frodo wallet testnet
18)   3208516197   Jeff's CB test wallet
19)   1622336804   jeff cb test wallet 2 untrusted opera
20)   3554663204   County Demo wallet (Issuer)
21) * 87074341     Issuer (Issuer)
Choose a wallet key [1-21] ('q' to quit, or Enter to use 87074341): 
Proofs:

Launcher ID: 165e3cc5789923c0de4731f4a6b76320c2e39ea175d64e19c8a6b17c183ec798
Coin ID: bc15d1c6d0ed191fa08c0618179fb66e4646fcdc4f844171c8a90c24b9104af2
Inner Address: txch1z7dfrxx2qfd2lmxrl2fq7vdvgmk9ac6mnffv6tu0whnpx7y3eynql320y2
[...]

PR #16077, opened 2023-08-16
Don't raise on duplicate VC proof insertion

Quexington:
Fixes (one half of) https://github.com/Chia-Network/chia-blockchain/issues/16052

### DIFF
--- a/chia/wallet/vc_wallet/vc_store.py
+++ b/chia/wallet/vc_wallet/vc_store.py
@@ -251,7 +251,7 @@ class VCStore:
     async def add_vc_proofs(self, vc_proofs: VCProofs) -> None:
         async with self.db_wrapper.writer_maybe_transaction() as conn:
             await conn.execute(
-                "INSERT INTO vc_proofs VALUES(?, ?)", (vc_proofs.root().hex(), bytes(vc_proofs.as_program()))
+                "INSERT OR IGNORE INTO vc_proofs VALUES(?, ?)", (vc_proofs.root().hex(), bytes(vc_proofs.as_program()))
             )
 
     async def get_proofs_for_root(self, root: bytes32) -> Optional[VCProofs]:
--- a/tests/wallet/vc_wallet/test_vc_wallet.py
+++ b/tests/wallet/vc_wallet/test_vc_wallet.py
@@ -228,6 +228,7 @@ async def test_vc_lifecycle(self_hostname: str, two_wallet_nodes_services: Any, 
 
     # Add proofs to DB
     await client_0.vc_add_proofs(proofs.key_value_pairs)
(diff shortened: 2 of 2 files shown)


chia/wallet/vc_wallet/vc_store.py
```diff
@@ -250,9 +250,16 @@ async def get_vc_record_by_coin_id(self, coin_id: bytes32) -> Optional[VCRecord]
 
     async def add_vc_proofs(self, vc_proofs: VCProofs) -> None:
         async with self.db_wrapper.writer_maybe_transaction() as conn:
-            await conn.execute(
-                "INSERT INTO vc_proofs VALUES(?, ?)", (vc_proofs.root().hex(), bytes(vc_proofs.as_program()))
-            )
+            root: bytes32 = vc_proofs.root()
+            exists_cursor = await conn.execute("SELECT EXISTS(SELECT 1 FROM vc_proofs WHERE root=?)", (root.hex(),))
```
arvidn, line 254: I think it's more reliable to use a context manager for `execute()` calls that returns a cursor. That way the cursor will be closed implicitly. Something like `async with conn.execute("...") as cursor:`

However, performing a lookup first and then decide whether to insert looks both expensive and risky, especially since this is just inside a *maybe*_transaction(). I would think it would be faster and more robust to just say:

```
INSERT OR IGNORE
```

or perhaps:

```
INSERT OR REPLACE
```

instead. I suspect that if the item exists, it's guaranteed to be identical to the one we're trying to insert. is that right? So `INSERT OR IGNORE` might be the way to go.

chia/wallet/vc_wallet/vc_store.py
```diff
@@ -250,9 +250,16 @@ async def get_vc_record_by_coin_id(self, coin_id: bytes32) -> Optional[VCRecord]
...
+            root_exists = await exists_cursor.fetchone()
+            await exists_cursor.close()
+            if root_exists is not None and root_exists[0] != 0:
+                existing_proof = await self.get_proofs_for_root(root)
+                assert existing_proof is not None
+                if existing_proof == vc_proofs:
```
arvidn, line 260: as far as I can tell, the UNIQUE constraint only applies to the `root` (since it's the `PRIMARY KEY`). So if this comparison fails, you'll still try to insert the new proof and get the DB constraint violation.

If this is deliberate, it would be good to have a test for it.

Or perhaps the mistake is in the table constraints, perhaps the *pair* of `root` and `proof` is supposed to be unique.

arvidn (changes requested):

chia/wallet/vc_wallet/vc_store.py
Quexington: You were correct INSERT OR IGNORE was just fine.  The change is much smaller now.

arvidn (approved):

PR #16078, opened 2023-08-16
Added wallet id showing when using the 'chia plotnft show'.

d1m1trus:


paninaro (approved):
Looks reasonable to me. Thanks for the enhancement!

PR #16079, opened 2023-08-16
Do not enable seeder signal handlers in tests

altendky:
### Purpose:

There are two reasons for this.  One relates to `RuntimeError: set_wakeup_fd only works in main thread of the main interpreter` which is triggered because tooling under pytest-xdist uses threads (sometimes) as I believe is discussed in https://github.com/pytest-dev/execnet/issues/96.  The other is that it messes up pytest's own handling of signals which means `ctrl+c` and such don't work well.

It would be great at some point to bring consistency to our signal handling.  I started that awhile back in https://github.com/Chia-Network/chia-blockchain/pull/13568 and will add a note there about taking care of the seeder as well.

### Testing Notes:

Daft For:
- [x] probably going to consider this after https://github.com/Chia-Network/chia-blockchain/pull/16055

### DIFF
--- a/chia/seeder/dns_server.py
+++ b/chia/seeder/dns_server.py
@@ -5,7 +5,6 @@ import logging
 import signal
 import sys
 import traceback
-from asyncio import AbstractEventLoop
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


jack60612 (approved):

emlowe (approved):
Approved with coverage diff exception

PR #16081, opened 2023-08-16
Refactor daemon simulation test

emlowe:
The test `test_daemon_simulation` would bring up an entire full stack including a timelord and then only tests that you can send a specific message to the node thru the daemon. This test doesn't seem particularly useful and has been consolidated with the original "full_system" test that was `test_simulation_1`

It was noticed that the test `test_daemon_simulation` and the test `test_simulation_1` are essentially the same test. They both bring up the entire full stack of multiple nodes, farmer, harvest, timelord, vdf_clients, and the daemon. There was only one minor difference in that the `test_simulation_1` didn't connect any nodes to the daemon. Since these tests are heavy and very slow to bring up everything, they have been consolidated into a single test `test_full_system` that brings up the entire system, runs the original tests, then connects a node to the daemon and runs a few additional daemon specific tests.

The `timelord_launcher.py` code needed some refactor in order to allow the `setup_vdf_clients` code to work properly multiple times in the same test run. Previously this would have issues as the multiple runs would stomp on a global variable in `timelord_launcher.py`

This now allows multiple "simulation" tests to be run in the same test run, and we can remove the check in the `simulation` fixture.

A new simple daemon test was added that brings up the daemon and a node - it makes sure the node connects to the daemon and then runs a singular test that the daemon can be used as a communication hub and send messages to/from the node.

Todo:
- [x] new daemon test should verify the payload of the return message and not just that "something" was returned.


emlowe:
Coverage complaints are due to missing existing tests around the signal handling for vdf client teardown. Can be added later.

arvidn (dismissed):
lgtm. maybe at least one other person should review too, before landing

tests/conftest.py
```diff
@@ -951,8 +949,8 @@ def cost_logger_fixture() -> Iterator[CostLogger]:
 
 @pytest_asyncio.fixture(scope="function")
 async def simulation(consensus_mode, bt):
-    if consensus_mode != Mode.PLAIN:
-        pytest.skip("Skipping this run. This test only supports one running at a time.")
+    #    if consensus_mode != Mode.PLAIN:
```
almogdepaz, line 952: i think we can delete this

almogdepaz (dismissed):

tests/conftest.py
altendky: I'm going to go ahead and make this a raised concern since we at least shouldn't leave it sitting commented out.

altendky (dismissed):

tests/conftest.py
emlowe: Removed

emlowe:
Close and reopen for CI

chia/timelord/timelord_launcher.py
```diff
@@ -18,22 +19,25 @@
...
-        stopped = True
-        for process in active_processes:
+@dataclass
+class VDFClientProcessMgr:
+    lock: asyncio.Lock
+    stopped: bool
+    active_processes: List[asyncio.subprocess.Process] = field(default_factory=list)
+
+
+async def kill_processes(process_mgr: VDFClientProcessMgr):
```
altendky, line 32: Make this a method?  Seems pretty closely related.

chia/timelord/timelord_launcher.py
```diff
@@ -74,14 +84,15 @@ async def spawn_process(host: str, port: int, counter: int, lock: asyncio.Lock,
...
                 log.error(f"VDF client {counter}: {stderr.decode().rstrip()}")
-        log.info(f"Process number {counter} ended.")
-        async with lock:
-            if proc in active_processes:
-                active_processes.remove(proc)
+
+        async with process_mgr.lock:
+            if proc in process_mgr.active_processes:
+                process_mgr.active_processes.remove(proc)
+
```
altendky, line 91: It feels like this and other stuff should be wrapped up by the class.  Basically, instead of providing a locking mechanism and hoping folks use it, apply the locking mechanism yourself with helpful abstractions that can be called instead.

Or maybe this whole function is a method maybe?

chia/timelord/timelord_launcher.py
```diff
@@ -18,22 +19,25 @@
...
-async def kill_processes(lock: asyncio.Lock):
-    global stopped
-    global active_processes
-    async with lock:
-        stopped = True
-        for process in active_processes:
+@dataclass
+class VDFClientProcessMgr:
+    lock: asyncio.Lock
+    stopped: bool
```
altendky, line 28: And associated call-site simplifications.

```suggestion
    lock: asyncio.Lock = field(default_factory=asyncio.Lock)
    stopped: bool = False
```

altendky (changes requested):
Mostly it seems like the new abstraction `VDFClientProcessMgr` could pull in some more methods and code.

chia/timelord/timelord_launcher.py
emlowe: Added

chia/simulator/setup_services.py
```diff
@@ -436,53 +436,84 @@ async def setup_introducer(bt: BlockTools, port: int) -> AsyncGenerator[Service[
 
 @asynccontextmanager
 async def setup_vdf_client(bt: BlockTools, self_hostname: str, port: int) -> AsyncGenerator[asyncio.Task[Any], None]:
-    lock = asyncio.Lock()
+    process_mgr = VDFClientProcessMgr(asyncio.Lock(), False, [])
```
altendky, line 439:
```suggestion
    process_mgr = VDFClientProcessMgr()
```

chia/simulator/setup_services.py
```diff
@@ -436,53 +436,84 @@ async def setup_introducer(bt: BlockTools, port: int) -> AsyncGenerator[Service[
...
+    vdf_task_1.cancel()
+    await process_mgr.kill_processes()
 
 
 @asynccontextmanager
 async def setup_vdf_clients(
     bt: BlockTools, self_hostname: str, port: int
 ) -> AsyncGenerator[Tuple[asyncio.Task[Any], asyncio.Task[Any], asyncio.Task[Any]], None]:
-    lock = asyncio.Lock()
+    process_mgr = VDFClientProcessMgr(asyncio.Lock(), False, [])
```
altendky, line 470:
```suggestion
    process_mgr = VDFClientProcessMgr()
```

chia/timelord/timelord_launcher.py
```diff
@@ -90,27 +107,33 @@ async def spawn_all_processes(config: Dict, net_config: Dict, lock: asyncio.Lock
...
+            prefer_ipv6=net_config.get("prefer_ipv6", False),
+        )
         for i in range(process_count)
     ]
     await asyncio.gather(*awaitables)
 
 
 async def async_main(config: Dict[str, Any], net_config: Dict[str, Any]) -> None:
-    lock = asyncio.Lock()
+    process_mgr = VDFClientProcessMgr(asyncio.Lock(), False, [])
```
altendky, line 123:
```suggestion
    process_mgr = VDFClientProcessMgr()
```

altendky (changes requested):

emlowe:
Set to draft while debugging test problem

emlowe:
Finally got this passing 100/100 in a repeated test run.

chia/simulator/setup_nodes.py
```diff
@@ -349,77 +332,94 @@ async def setup_full_system_inner(
...
+        )
+
         farmer_service = await async_exit_stack.enter_async_context(
             setup_farmer(
                 shared_b_tools,
                 shared_b_tools.root_path / "harvester",
-                shared_b_tools.config["self_hostname"],
-                consensus_constants,
-                full_node_0_port,
+                self_hostname=self_hostname,
```
altendky, line 395: Does `shared_b_tools.config["self_hostname"]` have the wrong value?

chia/simulator/setup_nodes.py
```diff
@@ -433,25 +433,15 @@ async def num_connections() -> int:
...
             farmer_service._node,
             introducer,
             timelord,
-            vdf_clients,
-            vdf_bluebox_clients,
-            timelord_bluebox,
-            timelord_bluebox_server,
+            timelord_bluebox_service,
+            daemon_ws,
         )
```
altendky, line 446: Pass by name?

chia/simulator/setup_services.py
```diff
@@ -441,57 +441,67 @@ async def setup_introducer(bt: BlockTools, port: int) -> AsyncGenerator[Service[
...
 
 @asynccontextmanager
-async def setup_vdf_clients(
-    bt: BlockTools, self_hostname: str, port: int
-) -> AsyncGenerator[Tuple[asyncio.Task[Any], asyncio.Task[Any], asyncio.Task[Any]], None]:
-    lock = asyncio.Lock()
+async def setup_vdf_clients(bt: BlockTools, self_hostname: str, port: int) -> AsyncIterator[None]:
+    find_vdf_client()  # raises FileNotFoundError if not found
+
+    process_mgr = VDFClientProcessMgr(asyncio.Lock(), False, [])
```
altendky, line 473:
```suggestion
    process_mgr = VDFClientProcessMgr()
```

chia/timelord/timelord_launcher.py
```diff
@@ -90,27 +114,33 @@ async def spawn_all_processes(config: Dict, net_config: Dict, lock: asyncio.Lock
...
+            i,
+            process_mgr,
+            prefer_ipv6=net_config.get("prefer_ipv6", False),
+        )
         for i in range(process_count)
     ]
     await asyncio.gather(*awaitables)
 
 
 async def async_main(config: Dict[str, Any], net_config: Dict[str, Any]) -> None:
```
altendky, line 129: Given that we're changing runtime timelord stuff, let's create an ephemeral testnet for this when we get close to the end.  We can leave this conversation open until we take care of that.

tests/simulation/config.py
```diff
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
 job_timeout = 60
-parallel = False
 install_timelord = True
+parallel = False
```
altendky, line 5: Diff noise?

altendky (changes requested):

chia/simulator/setup_nodes.py
emlowe: No I just used a variable at the top `self_hostname = shared_b_tools.config["self_hostname"]`

chia/simulator/setup_services.py
```diff
@@ -441,57 +441,67 @@ async def setup_introducer(bt: BlockTools, port: int) -> AsyncGenerator[Service[
...
     async with SignalHandlers.manage() as signal_handlers:
         signal_handlers.setup_async_signal_handler(handler=stop)
         try:
-            yield vdf_task_1
+            yield
         finally:
-            await kill_processes(lock)
+            await process_mgr.kill_processes()
+            vdf_task_1.cancel()
+            await asyncio.gather(vdf_task_1, return_exceptions=True)
```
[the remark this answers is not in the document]
emlowe: Yes, I originally just did `await` and while this works on macos fine, it did not on linux. It resulted in a hang as the await never returned - however, it's possible it would work now that I properly await the process ending after the kill. So I then tried a `cancel` and `await` and this ended up raising CancelledError in the await call - but I know it's Cancelled as I cancelled it so this just seemed less code than another try handler.
Note, I also just tried `cancel` but sometimes the process ends before the cancel is processed resulting in teardown errors about pending tasks. Since this is all teardown code, it seems consuming any exceptions while waiting for the task to finish after cancelling seemed reasonable to me

chia/simulator/setup_nodes.py
```diff
@@ -48,6 +50,18 @@
 ]
 
 
+@dataclass(frozen=True)
+class FullSystem:
+    node_1: Service[FullNode, FullNodeAPI] | Service[FullNode, FullNodeSimulator]
+    node_2: Service[FullNode, FullNodeAPI] | Service[FullNode, FullNodeSimulator]
```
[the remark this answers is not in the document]
emlowe: Changed

chia/timelord/timelord_launcher.py
```diff
@@ -18,38 +20,62 @@
...
+    *,
+    prefer_ipv6: bool,
+    task_status: Optional[anyio.abc.TaskStatus[None]] = None,
+):
     path_to_vdf_client = find_vdf_client()
     first_10_seconds = True
     start_time = time.time()
-    while not stopped:
+    if task_status is not None:
+        task_status.started()
```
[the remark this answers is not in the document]
emlowe: Removed

chia/timelord/timelord_launcher.py
```diff
@@ -18,38 +20,62 @@
...
+            self.stopped = True
+            for process in self.active_processes:
+                try:
+                    process.kill()
+                    await process.wait()
+                    # hack to avoid `Event loop is closed` errors (fixed in python 3.11.1)
+                    process._transport.close()  # type: ignore
+                except (ProcessLookupError, AttributeError):
+                    pass
+        self.active_processes.clear()
```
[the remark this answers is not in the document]
emlowe: Fixed to inside lock (nice catch)

tests/simulation/config.py
emlowe: yup, fixed

chia/simulator/setup_nodes.py
```diff
@@ -349,77 +332,94 @@ async def setup_full_system_inner(
...
     if b_tools is None:
         b_tools = await create_block_tools_async(constants=consensus_constants, keychain=keychain1)
     if b_tools_1 is None:
         b_tools_1 = await create_block_tools_async(constants=consensus_constants, keychain=keychain2)
+
+    self_hostname = shared_b_tools.config["self_hostname"]
+
     async with AsyncExitStack() as async_exit_stack:
+        vdf1_port = uint16(find_available_listen_port("vdf1"))
+        vdf2_port = uint16(find_available_listen_port("vdf2"))
```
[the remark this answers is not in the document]
emlowe: Digging into this seems outside the scope and I really didn't want get caught up in a big mess here

chia/simulator/setup_services.py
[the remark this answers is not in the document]
altendky: 1) I think most exceptions should be reported, we don't want silent failures we don't know about on shutdown.
2) The cancellation might not be yours...  but we won't deal with that in asyncio, at least not for now.
3) (actually actionable) I don't think that `asyncio.gather()` is a good way to express 'consume exceptions from a single call'.  I'd definitely rather see a `try`/`except`.  And I don't think using `asyncio.gather()` should have fixed anything vs. a direct `await`.

chia/timelord/timelord_launcher.py
```diff
@@ -18,38 +19,60 @@
...
+                    pass
+            self.active_processes.clear()
+
 
 def find_vdf_client() -> pathlib.Path:
     p = pathlib.Path(pkg_resources.get_distribution("chiavdf").location) / "vdf_client"
     if p.is_file():
         return p
-    raise FileNotFoundError("can't find vdf_client binary")
+    raise FileNotFoundError("Cannot find vdf_client binary. Is Timelord installed?")
```
altendky, line 61:
```suggestion
    raise FileNotFoundError("Cannot find vdf_client binary. Is Timelord installed?  See install-timelord.sh")
```

emlowe: fixed

chia/timelord/timelord_launcher.py
```diff
@@ -74,14 +100,12 @@ async def spawn_process(host: str, port: int, counter: int, lock: asyncio.Lock,
                     first_10_seconds = False
             else:
                 log.error(f"VDF client {counter}: {stderr.decode().rstrip()}")
-        log.info(f"Process number {counter} ended.")
-        async with lock:
-            if proc in active_processes:
-                active_processes.remove(proc)
+
+        await process_mgr.remove_process(proc)
```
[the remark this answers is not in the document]
emlowe: This loops around and just starts up new processes as the old ones exit (for any reason) - so I'm not sure it helps much to add that in?

altendky (approved):

chia/simulator/setup_nodes.py
```diff
@@ -124,7 +138,7 @@ async def setup_simulators_and_wallets(
     *,
     key_seed: Optional[bytes32] = None,
     initial_num_public_keys: int = 5,
-    db_version: int = 1,
+    db_version: int = 2,
```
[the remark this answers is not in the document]
arvidn: what's even more suspicious is that support for v1 was dropped. It's possible that this parameter doesn't do anything anymore. There is a fixture still, that's always 2 now.

arvidn (approved):

PR #16101, opened 2023-08-17
demote log level for TIMESTAMP_TOO_FAR_IN_FUTURE errors

arvidn:
### Purpose:

This patch aims to match the log level with the attention needed from the user.

### Current Behavior:

When a node receives an unfinished block with a timestamp that's more than 2 minutes *ahead* of its own clock, it's rejected with a `ConsensusError` exception, printing a very serious looking error to the log. We expect to receive more of these invalid blocks while transitioning to more 1.8 full nodes.

Prior to version 1.8.0 the upper timestamp limit was 5 minutes, now it's 2. This means that a sizeable part of the network still forwards blocks more than 2 minutes ahead of wall-clock.

### New Behavior:

Instead of printing the stack trace at `ERROR` log level, print a message at `INFO` log level describing the validation failure.

### Testing Notes:

I'm running this locally and have two instances printed to my log so far:
```
2023-08-17T22:09:41.847 full_node full_node_server        : INFO     Received block with timestamp too far into the future
2023-08-17T23:11:12.479 full_node full_node_server        : INFO     Received block with timestamp too far into the future
```

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -70,7 +70,7 @@ from chia.util.config import PEER_DB_PATH_KEY_DEPRECATED, process_config_start_m
 from chia.util.db_synchronous import db_synchronous_on
 from chia.util.db_version import lookup_db_version, set_db_version_async
 from chia.util.db_wrapper import DBWrapper2, manage_connection
-from chia.util.errors import ConsensusError, Err, ValidationError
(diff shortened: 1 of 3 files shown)


danieljperry (approved):
Agreed with changing the severity of this message to `INFO`.

emlowe (approved):

PR #16113, opened 2023-08-18
Add the basics for a more ergonomic wallet test framework

Quexington:
This PR aims to set up a more ergonomic fixture for writing wallet tests.

- With the existing fixtures, our wallet tests need a considerable amount of boilerplate before you can even start running legitimate test code.  This fixture will do all of that before it is imported into your test.
- Right now you must remember to parametrize your wallet function by trusted/untrusted syncing mode.  This fixture automatically creates that parametrization.
- To override config values with the existing fixtures is imperfect.  This fixture provides an API to specify configuration overrides that will automatically be applied to the full node/wallet node/wallet state manager.

The idea in the future is to add more capabilities to `WalletTestFramework` that make tracking things certain things easier, like:
- Whether or not TXs are pending/confirmed
- Balance numbers
- Checking for wallet creations

One test has been modified to use this framework as a proof of concept.  Some of the changes are related to the fact that `WalletTestFramework` actually exposes the types of things like `WalletNode` and `WalletStateManager` which were being used as `Any`s before.


tests/wallet/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+
+@pytest.fixture(scope="function", params=[True, False])
+def trusted_full_node(request: Any) -> bool:
+    trusted: bool = request.param
+    return trusted
+
+
+@pytest_asyncio.fixture(scope="function")
+async def wallet_environments(
+    trusted_full_node: bool, request: Any, blockchain_constants: ConsensusConstants
```
matt-o-how, line 43: request should be type `Dict`?

matt-o-how (dismissed):
seems useful to reduce overhead

wjblanke (approved):
aok

PR #16122, opened 2023-08-20
Notify glue of completed prerelease, and fix permissions 

wallentx:
### Purpose:
Send notification to glue of pre-release events, and also fix token permissions 

### Current Behavior:
Only notifies on release

### New Behavior:
Includes pre-release

### DIFF
--- a/.github/workflows/build-linux-arm64-installer.yml
+++ b/.github/workflows/build-linux-arm64-installer.yml
@@ -271,8 +271,13 @@ jobs:
           build_scripts/final_installer/chia-blockchain_${CHIA_INSTALLER_VERSION}_arm64.deb \
           build_scripts/final_installer/chia-blockchain-cli_${CHIA_INSTALLER_VERSION}-1_arm64.deb
 
-    - name: Mark installer complete
-      if: steps.check_secrets.outputs.HAS_GLUE_SECRET && env.RELEASE == 'true'
+    - name: Mark pre-release installer complete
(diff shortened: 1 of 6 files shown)


wallentx:
Will need to update glue to only execute release workflow on full release

.github/workflows/build-linux-arm64-installer.yml
```diff
@@ -272,7 +272,7 @@ jobs:
           build_scripts/final_installer/chia-blockchain-cli_${CHIA_INSTALLER_VERSION}-1_arm64.deb
 
     - name: Mark installer complete
-      if: steps.check_secrets.outputs.HAS_GLUE_SECRET && env.FULL_RELEASE == 'true'
+      if: steps.check_secrets.outputs.HAS_GLUE_SECRET && env.RELEASE == 'true'
```
cmmarslender, line 275: Glue is configured with two different jobs: `chia-blockchain` and `chia-blockchain-prerelease`
We need to make one call for FULL_RELEASE and a separate call for prerelease, because the url is different.

For full release, its what you have here already. For the prerelease call, its:
`${{ secrets.GLUE_API_URL }}/api/v1/$RFC_REPO-prerelease/$RELEASE_TAG/success/build-arm`

The same pattern applies for all the prerelease calls - after `v1`, it needs to be `chia-blockchain-prerelease`

cmmarslender (changes requested):

.github/workflows/build-linux-arm64-installer.yml
wallentx: Updated

cmmarslender (approved):

PR #16126, opened 2023-08-21
Refactor simulator setups from async generators to context managers

Quexington:
This is a behind the scenes fix to some test fixtures that shouldn't effect the normal operation of tests but makes the writing of future fixtures much easier and less error prone.

Currently, most of our service setup functions act _as if_ they're context managers but actually they are implemented as async generators.  It's possible this was done due to lack of knowledge about context managers or perhaps exit stacks which can manage a variable number of contexts.  The crucial difference between context managers and generators for test fixtures is that if an exception occurs while under the "context" of a certain generator value, the cleanup code that tears down the services will not get called.  This means all test fixtures utilizing these setup functions need to be written in a way that they will not raise which is difficult with complex fixtures that take input from the developer writing the test.  Under a context manager, the tear down _does_ happen even if the fixture raises for some reason.

(needs a coverage exemption for lines 489, 528, 541, 548-552, 900 - all except 489 & 885 are benchmark tests and both 489 & 885 is in use in some currently skipped tests)


altendky:
Me too, just hadn't gotten to a detailed review yet.  It's long been on 'the list'.  If someone else makes it through a detailed review before me, all the better.

Quexington:
Rerunning CI because it's been a while

chia/simulator/setup_nodes.py
```diff
@@ -409,103 +358,99 @@ async def setup_full_system_inner(
...
+                                        farmer_service._node,
+                                        introducer,
+                                        timelord,
+                                        vdf_clients,
+                                        vdf_bluebox_clients,
+                                        timelord_bluebox,
+                                        timelord_bluebox_server,
+                                    )
+                                    async with setup_daemon(btools=b_tools) as daemon_ws:
+                                        yield daemon_ws, ret
```
arvidn, line 456: it seems to me that you may be able to avoid this deep nesting with the exit stack, no?

arvidn (commented):
looks good! This is a great improvement to the tests.
It will need to be fixed-up once https://github.com/Chia-Network/chia-blockchain/pull/15977 lands

chia/simulator/setup_nodes.py
```diff
@@ -409,103 +358,99 @@ async def setup_full_system_inner(
...
-        vdf_clients,
-        vdf_bluebox_clients,
-        timelord_bluebox,
-        timelord_bluebox_server,
-    )
-    return daemon_ws, node_iters, ret
+    async with setup_introducer(shared_b_tools, uint16(0)) as introducer_service:
+        introducer = introducer_service._api
+        introducer_server = introducer_service._node.server
+        async with AsyncExitStack() as astack:
```
altendky, line 375: I'd say that `exit_stack` would be more clear.  The exit'ness of the stack seems more descriptive than the async'ness.  As in, it isn't an async call stack, for example.

chia/simulator/setup_nodes.py
altendky: In general, I wouldn't suggest using an exit stack just to eliminate nesting.  In this case...  maybe it's special enough?  They all exit at the same time...  It's really deep...

chia/simulator/setup_nodes.py
Quexington: This is a slight modification of the way they do it in the docs here: https://docs.python.org/3/library/contextlib.html#contextlib.AsyncExitStack (I only added the `a` because of a conflict in some section of the code with a non-async exit stack)

chia/simulator/setup_nodes.py
Quexington: Yeah I agree with Arvid, I think the intention here if written again fresh would have been to just have them all part of one context

chia/simulator/setup_nodes.py
altendky: image

Our existing uses.  `:]`

But...  I didn't actually raise a concern so...  I just prefer non-abbreviations.  And sure, sometimes you do want to clarify async vs. not and maybe whatever the pattern is should be consistent instead of conditional on whether they are used in the same scope or not.

Quexington: Ah I didn't know we were using this at all, so yeah I'd rather be consistent across our own repo

chia/simulator/setup_nodes.py
```diff
@@ -167,22 +146,20 @@ async def setup_simulators_and_wallets(
...
-
-        bt_tools, node_iters, simulators, wallets_services = res
-        wallets = []
-        for wallets_servic in wallets_services:
-            wallets.append((wallets_servic._node, wallets_servic._node.server))
-
-        nodes = []
-        for nodes_service in simulators:
-            nodes.append(nodes_service._api)
+        ) as res:
```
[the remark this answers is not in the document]
Quexington: Done

chia/simulator/setup_nodes.py
```diff
@@ -212,14 +189,12 @@ async def setup_simulators_and_wallets_service(
...
             config_overrides,
             disable_capabilities,
-        )
-
-        bt_tools, node_iters, simulators, wallets_services = res
-        yield simulators, wallets_services, bt_tools[0]
-
-        await _teardown_nodes(node_iters)
+        ) as res:
+            bt_tools, simulators, wallets_services = res
```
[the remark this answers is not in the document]
Quexington: done

chia/simulator/setup_nodes.py
Quexington: The nesting has been eliminated

chia/simulator/setup_nodes.py
Quexington: Renamed to async_exit_stack

arvidn:
looks good to me. I think we should wait to land this until after we cut the 2.1 release branch though.
Also, this looks like a legitimate error:

https://github.com/Chia-Network/chia-blockchain/actions/runs/6124774267/job/16630973767?pr=16126#step:10:3762

altendky:
Though it is all test code and avoiding divergence and conflicts would be nice.

Quexington:
I agree with @altendky I don't see how this matters whether or not this is in the release branch

altendky:
I think it matters because of the risk of extra confusion if there are unexpected issues.  But, I think that is less likely than the odds of continued hassle with merging both with `main` and then any divergence risk with `release/` after that.

arvidn (approved):

PR #16130, opened 2023-08-22
Use pytest's monkeypatch fixture instead of unittest's MagicMock

AmineKhaldi:
This allows us to mock the `_sync` function using the proper mechanism for the test framework we use, and has the side effect of allowing us to fully annotate `test_full_sync.py` in an upcoming PR.


altendky (approved):

emlowe (approved):

wallentx:
The subject of this PR made me think I ate the wrong bag of treats 😵‍💫

PR #16131, opened 2023-08-22
Add flags to CR-CAT offer summary

Quexington:
This was a request from the UI side to make viewing information about CR-CATs in offers a little more ergonomic than having to manually uncurry a clvm function.  It feels a bit hacky, but I'm not sure what the more general solution is as of now.  Maybe a separate RPC to return information about a proofs checker?  That feels a little bit off as well, so I'm inclined to go with this right now as it's the simplest solution that unblocks the UI.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -1712,7 +1712,7 @@ class WalletRpcApi:
             }
 
         # This is a bit of a hack in favor of returning some more manageable information about CR-CATs
-        # A more general solution surely exists, but I'm sure what it is right now
+        # A more general solution surely exists, but I'm not sure what it is right now
         return {
             **response,
             "summary": {
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Any, ClassVar, Dict, List, Optional, Set, Tuple, Union
 
 from blspy import AugSchemeMPL, G1Element, G2Element, PrivateKey
+from clvm_tools.binutils import assemble
 
 from chia.consensus.block_rewards import calculate_base_farmer_reward
 from chia.data_layer.data_layer_errors import LauncherCoinNotFoundError
@@ -82,6 +83,7 @@ from chia.wallet.util.transaction_type import CLAWBACK_INCOMING_TRANSACTION_TYPE
 from chia.wallet.util.tx_config import DEFAULT_TX_CONFIG, TXConfig
 from chia.wallet.util.wallet_sync_utils import fetch_coin_spend_for_coin_state
 from chia.wallet.util.wallet_types import CoinType, WalletType
+from chia.wallet.vc_wallet.cr_cat_drivers import ProofsChecker
 from chia.wallet.vc_wallet.cr_cat_wallet import CRCATWallet
 from chia.wallet.vc_wallet.vc_store import VCProofs
 from chia.wallet.vc_wallet.vc_wallet import VCWallet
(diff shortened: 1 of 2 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -1699,16 +1701,41 @@ async def get_offer_summary(self, request: Dict[str, Any]) -> EndpointResult:
...
             }
         else:
-            return {
+            response = {
                 "summary": await self.service.wallet_state_manager.trade_manager.get_offer_summary(offer),
                 "id": offer.name(),
             }
 
+        # This is a bit of a hack in favor of returning some more manageable information about CR-CATs
+        # A more general solution surely exists, but I'm sure what it is right now
```
ytx1991, line 1715: "I'm not sure"

chia/rpc/wallet_rpc_api.py
```diff
@@ -1699,16 +1701,41 @@ async def get_offer_summary(self, request: Dict[str, Any]) -> EndpointResult:
...
+        return {
```
ytx1991, line 1716: If it is not a CRCAT offer, will the response be exactly the same as the previous response? I'm worried that this change will cause issues on DEX or NFT exchange.

chia/rpc/wallet_rpc_api.py
```diff
@@ -1699,16 +1701,41 @@ async def get_offer_summary(self, request: Dict[str, Any]) -> EndpointResult:
...
                 "id": offer.name(),
             }
         else:
-            return {
+            response = {
                 "summary": await self.service.wallet_state_manager.trade_manager.get_offer_summary(offer),
                 "id": offer.name(),
             }
 
+        # This is a bit of a hack in favor of returning some more manageable information about CR-CATs
```
ytx1991, line 1714: I think we should do this in a general way since it is expected there will be more special offers with extra info. What is the disadvantage of using plain text data in the offer file? Is there a privacy concern or file size concern? If we can save raw data in the offer file, it will be easier for everyone to read.

chia/rpc/wallet_rpc_api.py
Quexington: Yeah the hack only triggers if there's exactly a key called "proofs_checker" that is under exactly a key called "also" in the driver dict of the offer being summarized.  In addition, we're only adding a key, which should break any existing parsing anyways.

chia/rpc/wallet_rpc_api.py
Quexington: The offer file format is pretty specifically a spendbundle and changing it now would be a pretty breaking change.  People can make formats on top of the existing format with more metadata, and I have no problem with that.

The problem here is that this information is in the offer file already and it's just a matter of displaying it to the user of the RPC.  I think it's potentially the front-end's responsibility to interpret the proofs checker but that's a big lift which is why we're doing this as a convenience on the backend. 

If more cases like this come up, it will inform me of the more general solution but like I said in the comment, I'm not sure what it is.

chia/rpc/wallet_rpc_api.py
Quexington: Fixed

chia/rpc/wallet_rpc_api.py
ytx1991: If we can remove the "infos" when it is not a CRCAT it will be better.

Quexington: Infos is something that's already in the request.  It gives information about the assets involved in the trade.  I'm just adding one extra key to the info for a CR-CAT.

chia/rpc/wallet_rpc_api.py
Quexington: Yeah that's basically it, the front end needs a much more sophisticated understanding of CLVM than it currently operates with.  Right now it kind of ignores that altogether.  I'm not saying this is good, I think understanding CLVM would be good, but it's a paradigm shift.

ytx1991 (approved):
lgtm

paninaro (approved):
I don't love the way destructuring with overrides is utilized here as it makes it harder to understand what's being returned, but I also don't know of a better alternative without breaking things down into smaller functions. Good enough for me.

PR #16132, opened 2023-08-22
DataLayer fingerprint control

altendky:
### Draft For:
- [x] tests
- [x] wallet selection prompt for the CLI


altendky:
For the missing coverage, it is true that we mostly don't test the CLI, though the new one is tested.  There are a couple other error cases I could get exercised if someone wants me to.

emlowe (approved):
Approved with coverage diff exception

PR #16135, opened 2023-08-22
Configure number of stored full files in DL.

fchirica:
Introduces configuration option `full_file_storage_window` (default: 1) that controls the number of full files store in the local DAT store for Datalayer.

The new default means that we will only store the most recent full tree file and delete the others on an upgrade.


altendky (changes requested):
I think a few bits could read a bit more simply.  No major issues.  I'll come back to this in more detail later.

chia/data_layer/data_layer.py
```diff
@@ -441,22 +451,33 @@ async def upload_files(self, tree_id: bytes32) -> None:
...
                 break
             try:
                 if uploaders is not None and len(uploaders) > 0:
                     request_json = {
                         "store_id": tree_id.hex(),
-                        "full_tree_filename": write_file_result.full_tree.name,
                         "diff_filename": write_file_result.diff_tree.name,
                     }
+                    if write_file_result.full_tree is not None:
+                        request_json["full_tree_filename"] = write_file_result.full_tree.name
```
altendky, line 479: https://github.com/Chia-Network/chia-blockchain/blob/2126598941f9fab021929e58a2ec9fb276f1f547/chia/data_layer/s3_plugin_service.py#L145

The plugin requires that the `full_tree_filename` key be present and it assumes the value will be a string.  Seems like we're looking at a breaking API change here.  I'll ask around about this.

This also seems to imply we are not testing this case.

altendky (dismissed):

chia/data_layer/data_layer.py
emlowe: I created PR https://github.com/Chia-Network/chia-blockchain/pull/16135 to update the plugin.
still nothing tested in the plugin though.

emlowe: Florin made some small updates to the plugin to not need the full file name

emlowe:
Yea, I think something like `maximum_full_file_count` makes more sense than `full_file_storage_window`

emlowe (approved):

emlowe:
coverage diff exception approved

PR #16136, opened 2023-08-23
Code Cleanup: Fix overly broad exception handler

emlowe:
Previously code was a little strange IMO - but primarily, it was eating and ignoring all exceptions, which generally is a bad idea.

And since the idea here was if there is some problem getting the extra clawback info we would just continue, using exceptions here seems the wrong idea.

I think this PR represents better code flow. Any real exceptions will get raised as expected, while typical parsing errors for clawback items are ignored and not displayed

There is a test for this code that is flaky and it is difficult to determine why because this hides and ignores all exceptions

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -948,10 +948,11 @@ class WalletRpcApi:
             if record is None:
                 log.error(f"Cannot find coin record for type {tx['type']} transaction {tx['name']}")
                 continue
-            if record.metadata is None:
-                log.error(f"Cannot find metadata for coin {coin.name().hex()}")
+            try:
+                tx["metadata"] = record.parsed_metadata().to_json_dict()
+            except ValueError as e:
+                log.exception(f"Could not parse coin record metadata: {type(e).__name__} {e}")
                 continue
-            tx["metadata"] = record.parsed_metadata().to_json_dict()
             tx["metadata"]["coin_id"] = coin.name().hex()
             tx["metadata"]["spent"] = record.spent
         return {
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -937,21 +937,23 @@ class WalletRpcApi:
         tx_list = []
         # Format for clawback transactions
         for tr in transactions:
-            try:
-                tx = (await self._convert_tx_puzzle_hash(tr)).to_json_dict_convenience(self.service.config)
-                tx_list.append(tx)
-                if tx["type"] not in CLAWBACK_INCOMING_TRANSACTION_TYPES:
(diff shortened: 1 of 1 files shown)


chia/rpc/wallet_rpc_api.py
```diff
@@ -937,21 +937,23 @@ async def get_transactions(self, request: Dict[str, Any]) -> EndpointResult:
...
+            if tx["type"] not in CLAWBACK_INCOMING_TRANSACTION_TYPES:
+                continue
+            coin: Coin = tr.additions[0]
+            record: Optional[WalletCoinRecord] = await self.service.wallet_state_manager.coin_store.get_coin_record(
+                coin.name()
+            )
+            if record is None:
+                log.error(f"Cannot find coin record for type {tx['type']} transaction {tx['name']}")
+                continue
+            if record.metadata is None:
```
altendky, line 951: I hesitate a bit with this precheck.  Can we not leave this detail to `.parsed_metadata()` and handle when it fails?

altendky (changes requested):
Overall, I agree.  Thanks.  Unless...  we want to still log remaining unexpected exceptions and continue with them?  Aside from coding practices, I think the lack of exposure of the exception information, such as traceback and full exception info, was the issue here?

chia/rpc/wallet_rpc_api.py
emlowe: We could use, it raises so I suppose it would be pythonic to let it do so inside a try/except

I think letting the standard RPC handling for exceptions is good for anything outside the typical flow - so those would be all reported in the log and back to the caller

altendky: The main point of not pre-checking is to avoid writing the checks twice.  The second place will almost certainly not get maintained and then still in some case result in an exception anyways.

Also, note my general comment.  Previous code continued the loop on 'unexpected' exceptions, but was silent.  This code is noisy (in a good way), but doesn't continue.  Maybe we want noisy and also continuing?  Don't let one failure cost you from finding all other valid records.

emlowe: No, I want to short-circuit on things very strange and only continue for specific cases around clawback handling. See if  the latest commit works for you

altendky (approved):

arvidn (approved):

Starttoaster:
It's been 5 months since this PR saw an Actions workflow. This should probably see a workflow run again prior to merge

arvidn:
I think it's fine to not cover the exception case here (at least as part of this PR)

altendky:
sorry...  `:[`

Starttoaster:
^ removed coverage-diff because it was reviewed and removed earlier. The latest workflow run just added it back.

PR #16138, opened 2023-08-23
Improve wallet rpc test consistency

emlowe:
Improve run consistency by adding in some wait for wallet sync whenever we get new puz hashes.

Sometimes on very slow runs, getting a new puz hash will cause the wallet to lose sync long enough for subsequent calls to fail.


altendky (approved):
thanks

paninaro (approved):

PR #16139, opened 2023-08-23
increase windows pytest threads to 3

emlowe:
Experiment in increasing windows pytest thread count to 3


emlowe:
This seems to work pretty well. However, it's difficult to tell if tests pass any faster.

paninaro (approved):

PR #16144, opened 2023-08-24
wallet: Fix leaking `WalletNode.race_cache`

xdustinface:
### Purpose:

This fixes the leaking `WalletNode.race_cache` which is used for untrusted nodes because we send coin state updates before the blocks which should be vice versa for the validation in untrusted mode. 

There is some cleanup code in there but it would never cleanup anything the way it is right now because it depends on entries being added to `WalletNode.race_cache_hashes` which we don't do and never did, i checked back the PR where it was introduced. 

### Current Behavior:

There is no cache cleanup happening because we never add any entries to `WalletNode.race_cache_hashes` which is required for the existing cleanup logic to work. 

### New Behavior:

Outdates cache entries will be removed.

### Testing Notes:

This adds some unit tests for the helpers of the cache.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -114,6 +114,12 @@ class PeerPeak:
     hash: bytes32
 
 
+@dataclasses.dataclass(frozen=True)
+class RaceCacheEntry:
+    height: uint32
+    coin_states: Set[CoinState]
+
+
 @dataclasses.dataclass
 class WalletNode:
     config: Dict[str, Any]
@@ -138,8 +144,7 @@ class WalletNode:
     wallet_peers: Optional[WalletPeers] = None
     peer_caches: Dict[bytes32, PeerRequestCache] = dataclasses.field(default_factory=dict)
     # in Untrusted mode wallet might get the state update before receiving the block
-    race_cache: Dict[bytes32, Set[CoinState]] = dataclasses.field(default_factory=dict)
-    race_cache_hashes: List[Tuple[uint32, bytes32]] = dataclasses.field(default_factory=list)
+    race_cache: Dict[bytes32, RaceCacheEntry] = dataclasses.field(default_factory=dict)
     node_peaks: Dict[bytes32, PeerPeak] = dataclasses.field(default_factory=dict)
     validation_semaphore: Optional[asyncio.Semaphore] = None
     local_node_synced: bool = False
@@ -459,7 +464,6 @@ class WalletNode:
             await asyncio.sleep(0.5)  # https://docs.aiohttp.org/en/stable/client_advanced.html#graceful-shutdown
         self.wallet_peers = None
         self.race_cache = {}
-        self.race_cache_hashes = []
         self._balance_cache = {}
 
     def _set_state_changed_callback(self, callback: StateChangedProtocol) -> None:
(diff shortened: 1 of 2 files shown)


chia/wallet/wallet_node.py
```diff
@@ -1199,9 +1198,14 @@ async def sync_from_untrusted_close_to_peak(self, new_peak_hb: HeaderBlock, peer
...
-                    self.log.info(f"Receiving race state: {self.race_cache[header_hash]}")
-                    await self.add_states_from_peer(list(self.race_cache[header_hash]), peer)
+                race_cache = self.race_cache.get(header_hash)
+                if race_cache is not None:
+                    self.log.info(f"Receiving race cache: {race_cache}")
+                    if await self.add_states_from_peer(list(race_cache.coin_states), peer):
+                        self.race_cache.pop(header_hash)
+
+            # Clear old entries that are no longer relevant
+            self.cleanup_race_cache(min_height=new_peak_hb.height - 100)
```
arvidn, line 1208: do do we need the race cache for reorgs? I would have expected the height to not subtract 100 here

tests/wallet/test_wallet_node.py
```diff
@@ -384,3 +388,65 @@ async def restart_with_fingerprint(fingerprint: Optional[int]) -> None:
     # Restart one more time and make sure the balance is still correct after start
     await restart_with_fingerprint(initial_fingerprint)
     assert await wallet_node.get_balance(wallet_id) == expected_more_balance
+
+
+@pytest.mark.asyncio
+async def test_race_cache_helper(root_path_populated_with_config: Path, get_temp_keyring: Keychain) -> None:
+    config = load_config(root_path_populated_with_config, "config.yaml", "wallet")
+    wallet_node = WalletNode(config, root_path_populated_with_config, test_constants, get_temp_keyring)
```
arvidn, line 396: in an ideal world the race cache would be a separate class that could be tested without constructing the full `WalletNode`. But this is fine for now I think

arvidn (commented):
did you test this manually? I'm not sure we have very good automated test coverage of this logic

chia/wallet/wallet_node.py
xdustinface: No we don't need it for reorgs, the reorgs come in as new updates anyway and the 100 is just inherited from what it was before. I moved the cache cleanup out of ` add_state_to_race_cache`. See few lines above.

tests/wallet/test_wallet_node.py
xdustinface: I think we should get rid of this race cache anyway with the new protocol. Its just here because the messages come in the wrong order at the moment but we can't just change it.

xdustinface:
> did you test this manually? I'm not sure we have very good automated test coverage of this logic

No i didn't but i added tests for the add/clearup logic and the code will certainly run for each peak. Do you have some specific concern?

xdustinface:
@arvidn Dropped the first commit (https://github.com/Chia-Network/chia-blockchain/compare/b4302f7b59479386dff589e37ae6f990eaf6d066..f93e619c0febef8019075fd4cf88f7a19349452f#diff-8387032f76cb8397134e314bda2f2ebbe6dd1c4aa473feac200bdbd4f27051b4L1204-R1204) which was about dropping race cache entries which were added "successfully", i.e. if `add_states_from_peer` returned `True` which doesn't imply that the coin states were added and valid, i got that wrong before.

xdustinface:
Closing in favour of #16239.

PR #16147, opened 2023-08-24
Add some tests related to the state of the height-to-hash cache file

AmineKhaldi:


tests/core/full_node/test_block_height_map.py
```diff
@@ -220,8 +243,21 @@ async def test_restore_extend(self, tmp_dir, db_version):
...
+                assert len(heights) == (2000 + 1) * 32
+
             height_map = await BlockHeightMap.create(tmp_dir, db_wrapper)
 
+            # Make sure we properly wrote the additional data to the cache
+            async with aiofiles.open(tmp_dir / "height-to-hash", "rb") as f:
+                new_heights = bytearray(await f.read())
+                assert len(new_heights) == (4000 + 1) * 32
+                for i in range(0, len(heights), 32):
+                    assert new_heights[i : i + 32] == heights[i : i + 32]
```
arvidn, line 259: is `len(new_heights)` allowed to be different than `len(heights)`?
it looks like this might be able to be replaced by:
```
assert heights == await f.read()
```

tests/core/full_node/test_block_height_map.py
```diff
@@ -148,8 +149,21 @@ async def test_save_restore(self, tmp_dir, db_version):
...
+                assert len(heights) == (10000 + 1) * 32
+
             height_map = await BlockHeightMap.create(tmp_dir, db_wrapper)
 
+            # Make sure we didn't alter the cache (nothing to write)
+            async with aiofiles.open(tmp_dir / "height-to-hash", "rb") as f:
+                new_heights = bytearray(await f.read())
+                assert len(new_heights) == len(heights)
+                for i in range(0, len(heights), 32):
+                    assert new_heights[i : i + 32] == heights[i : i + 32]
```
arvidn, line 165:
```suggestion
                assert bytearray(await f.read()) == heights
```

tests/core/full_node/test_block_height_map.py
```diff
@@ -148,8 +149,21 @@ async def test_save_restore(self, tmp_dir, db_version):
                 await conn.execute("DROP TABLE full_blocks")
             await setup_db(db_wrapper)
             await setup_chain(db_wrapper, 10000, ses_every=20, start_height=9970)
+
+            # At this point we should have a proper cache file
+            async with aiofiles.open(tmp_dir / "height-to-hash", "rb") as f:
+                heights = bytearray(await f.read())
+                assert len(heights) == (10000 + 1) * 32
```
arvidn, line 156: this also looks like you could just use `Path.getsize()`

arvidn (commented):
there's no test to actually ensure the optimization works. maybe that's a bit cubersome, it might involve monkey-patching `aiofiles` to see what's being written to disk, and where

tests/core/full_node/test_block_height_map.py
```diff
@@ -183,6 +197,15 @@ async def test_restore_entire_chain(self, tmp_dir, db_version):
...
 
+            # We replaced the whole cache at this point so all values should be different
+            async with aiofiles.open(tmp_dir / "height-to-hash", "rb") as f:
+                new_heights = bytearray(await f.read())
+                # Make sure we wrote the whole file
+                assert len(new_heights) == (10000 + 1) * 32
+                # Make sure none of the old values remain
+                for i in range(0, len(heights), 32):
+                    assert new_heights[i : i + 32] != heights[i : i + 32]
+
```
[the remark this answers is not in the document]
AmineKhaldi: It's different because in this test we initially set this up with 900 items and then expose that to 10000 DB items that are completely different. We're restoring the entire chain from DB, rendering the whole cache file invalid.

tests/core/full_node/test_block_height_map.py
AmineKhaldi: Here it's different because we're extending the initial 2000 DB items with 2000 more. The file cache gets updated accordingly.

tests/core/full_node/test_block_height_map.py
```diff
@@ -220,8 +243,21 @@ async def test_restore_extend(self, tmp_dir, db_version):
             await setup_db(db_wrapper)
             # add 2000 blocks to the chain
             await setup_chain(db_wrapper, 4000, ses_every=20)
+
+            # At this point we should have a proper cache with the old chain data
+            async with aiofiles.open(tmp_dir / "height-to-hash", "rb") as f:
+                heights = bytearray(await f.read())
+                assert len(heights) == (2000 + 1) * 32
```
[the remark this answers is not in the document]
AmineKhaldi: `object has no attribute 'getsize'` ?

tests/core/full_node/test_block_height_map.py
AmineKhaldi: `object has no attribute 'getsize'`

tests/core/full_node/test_block_height_map.py
AmineKhaldi: Thanks, done.

tests/core/full_node/test_block_height_map.py
altendky: https://docs.python.org/3.11/library/pathlib.html#pathlib.Path.stat?

AmineKhaldi: Thanks.

tests/core/full_node/test_block_height_map.py
```diff
@@ -374,6 +375,76 @@ async def test_rollback2(self, tmp_dir: Path, db_version: int) -> None:
...
+        # the cache file and there is nothing to write in that file as a result.
+        async with DBConnection(db_version) as db_wrapper:
+            await setup_db(db_wrapper)
+            await setup_chain(db_wrapper, 10000, ses_every=20)
+            await BlockHeightMap.create(tmp_dir, db_wrapper)
+            # To ensure we're actually loading from cache, and not the DB, clear
+            # the table.
+            async with db_wrapper.writer_maybe_transaction() as conn:
+                await conn.execute("DROP TABLE full_blocks")
+            await setup_db(db_wrapper)
```
arvidn, line 390: you drop the table, but then you immediately create it again.
The intention might be a bit clearer if you, instead of `DROP TABLE`, you do `DELETE FROM full_blocks`.
That way you don't have to call `setup_db()` to re-create the table

tests/core/full_node/test_block_height_map.py
```diff
@@ -374,6 +375,76 @@ async def test_rollback2(self, tmp_dir: Path, db_version: int) -> None:
...
+            await setup_chain(db_wrapper, 10000, ses_every=20, start_height=9970)
+            # At this point we should have a proper cache file
+            heights_stat = (tmp_dir / "height-to-hash").stat()
+            assert heights_stat.st_size == (10000 + 1) * 32
+            await BlockHeightMap.create(tmp_dir, db_wrapper)
+            # Make sure we didn't alter the cache (nothing to write)
+            new_heights_stat = (tmp_dir / "height-to-hash").stat()
+            assert new_heights_stat.st_mtime == heights_stat.st_mtime
+            assert new_heights_stat.st_size == heights_stat.st_size
```
arvidn, line 399: if the timestamp only has seconds resolution, this wouldn't tell you if the file was updated. You would a sleep somewhere probably

tests/core/full_node/test_block_height_map.py
```diff
@@ -374,6 +375,76 @@ async def test_rollback2(self, tmp_dir: Path, db_version: int) -> None:
...
+            # the table.
+            async with db_wrapper.writer_maybe_transaction() as conn:
+                await conn.execute("DROP TABLE full_blocks")
+            await setup_db(db_wrapper)
+            await setup_chain(db_wrapper, 10000, ses_every=20, start_height=9970)
+            # At this point we should have a proper cache file
+            heights_stat = (tmp_dir / "height-to-hash").stat()
+            assert heights_stat.st_size == (10000 + 1) * 32
+            await BlockHeightMap.create(tmp_dir, db_wrapper)
+            # Make sure we didn't alter the cache (nothing to write)
```
arvidn, line 396: it's not entirely obvious why we expect the file not to be written. doesn't `setup_chain()` produce a random chain? So when you clear the first one (that one the cache was built from) with a new (random) chain, won't the cache be outdated, and recreated?

tests/core/full_node/test_block_height_map.py
```diff
@@ -374,6 +375,76 @@ async def test_rollback2(self, tmp_dir: Path, db_version: int) -> None:
...
+            heights = bytearray(900 * 32)
+            for i in range(900):
+                idx = i * 32
+                heights[idx : idx + 32] = bytes([i % 256] * 32)
+            await write_file_async(tmp_dir / "height-to-hash", heights)
+            await setup_db(db_wrapper)
+            await setup_chain(db_wrapper, 10000, ses_every=20)
+            await BlockHeightMap.create(tmp_dir, db_wrapper)
+            # We replaced the whole cache at this point so all values should be different
+            async with aiofiles.open(tmp_dir / "height-to-hash", "rb") as f:
```
arvidn, line 415: I would recommend you just use normal `open()` here, instead of `aiofiles`

tests/core/full_node/test_block_height_map.py
```diff
@@ -374,6 +375,76 @@ async def test_rollback2(self, tmp_dir: Path, db_version: int) -> None:
...
+                heights = bytearray(await f.read())
+                assert len(heights) == (2000 + 1) * 32
+            await BlockHeightMap.create(tmp_dir, db_wrapper)
+            # Make sure we properly wrote the additional data to the cache
+            async with aiofiles.open(tmp_dir / "height-to-hash", "rb") as f:
+                new_heights = bytearray(await f.read())
+                assert len(new_heights) == (4000 + 1) * 32
+                for i in range(0, len(heights), 32):
+                    assert new_heights[i : i + 32] == heights[i : i + 32]
```
arvidn, line 446: I think the loop is redundant here. if `new_heights` is guaranteed to have the same length as `heights`, you can just:
```
assert new_heights == heights
```
Otherwise, you can just compare a slice:
```
assert new_heights[:len(heights)] == heights
```

tests/core/full_node/test_block_height_map.py
```diff
@@ -374,6 +375,76 @@ async def test_rollback2(self, tmp_dir: Path, db_version: int) -> None:
...
+            await setup_db(db_wrapper)
+            # Add 2000 blocks to the chain
+            await setup_chain(db_wrapper, 4000, ses_every=20)
+            # At this point we should have a proper cache with the old chain data
+            async with aiofiles.open(tmp_dir / "height-to-hash", "rb") as f:
+                heights = bytearray(await f.read())
+                assert len(heights) == (2000 + 1) * 32
+            await BlockHeightMap.create(tmp_dir, db_wrapper)
+            # Make sure we properly wrote the additional data to the cache
+            async with aiofiles.open(tmp_dir / "height-to-hash", "rb") as f:
```
arvidn, line 442: here too, I would suggest using `open` instead of aiofiles in the test

tests/core/full_node/test_block_height_map.py
AmineKhaldi: Thanks, done.

tests/core/full_node/test_block_height_map.py
AmineKhaldi: As the intent here is to check the size (and content) match, I removed this additional time check.

tests/core/full_node/test_block_height_map.py
AmineKhaldi: There is nothing random about the chain creation here so the intent is to check that nothing new was written and that the size and content match expectations.

tests/core/full_node/test_block_height_map.py
AmineKhaldi: Thanks, done.

tests/core/full_node/test_block_height_map.py
AmineKhaldi: As the lengths do not match, I took the second option of comparing the first part of `new_heights`, thanks.

tests/core/full_node/test_block_height_map.py
AmineKhaldi: Thanks, done.

tests/core/full_node/test_block_height_map.py
arvidn: you don't check the content now though. did you intend to?

arvidn (approved):

PR #16148, opened 2023-08-24
`@pytest.mark.limit_consensus_modes`

altendky:
### Purpose:

- reduce the copy paste code around limiting otherwise parametrized consensus modes
- avoid actually executing the test cases just to skip them
- clean up these extraneous skip statuses from the test results, build logs, and local developer test runs
- _maybe_ also avoid the fixture setups for these skipped tests (i haven't checked into this in detail yet) (a quick check seems to suggest that fixture setups are being avoided)

### Current Behavior:

```python
consensus_mode: Mode
```

and

```python
if consensus_mode != Mode.PLAIN:
    pytest.skip("only run in PLAIN mode to save time")
```

### New Behavior:

```python
@pytest.mark.limit_consensus_modes(reason="save time")
```

### Testing Notes:

`main`:
> `<testsuites tests="16538" failures="0" errors="0" skipped="991" time="134501.239">`

this:
> `<testsuites tests="15874" failures="0" errors="0" skipped="327" time="126181.596">`

Non-skipped test counts match.

```
>>> (16538 - 991) == (15874 - 327)
True
```

Regarding the missed coverage, all three lines are in test framework code intended to notify us in the case that this mark is applied without the `consensus_mode` fixture being in use by the test.  I did exercise that at some point, but we often don't have test code for our test frameworks.

https://github.com/Chia-Network/chia-blockchain/pull/16148/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R342

https://github.com/Chia-Network/chia-blockchain/pull/16148/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R354-R356

### Draft For:
- [x] review total test counts to make sure this doesn't lose any non-skipped tests https://github.com/Chia-Network/chia-blockchain/pull/16149
- [ ] should we auto-apply to benchmarks?

### DIFF
--- a/tests/simulation/test_simulation.py
+++ b/tests/simulation/test_simulation.py
@@ -60,7 +60,7 @@ async def extra_node(self_hostname):
 
 
 class TestSimulation:
-    @pytest.mark.plain_consensus_only(reason="This test only supports one running at a time.")
+    @pytest.mark.limit_consensus_modes(reason="This test only supports one running at a time.")
     @pytest.mark.asyncio
     async def test_simulation_1(self, simulation, extra_node, self_hostname):
         node1, node2, _, _, _, _, _, _, _, sanitizer_server = simulation
--- a/tests/simulation/test_simulation.py
+++ b/tests/simulation/test_simulation.py
@@ -60,6 +60,7 @@ async def extra_node(self_hostname):
 
 
 class TestSimulation:
+    @pytest.mark.plain_consensus_only(reason="This test only supports one running at a time.")
     @pytest.mark.asyncio
     async def test_simulation_1(self, simulation, extra_node, self_hostname):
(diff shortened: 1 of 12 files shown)


tests/simulation/test_simulation.py
```diff
@@ -60,6 +60,7 @@ async def extra_node(self_hostname):
 
 
 class TestSimulation:
+    @pytest.mark.limit_consensus_modes(reason="This test only supports one running at a time.")
```
altendky, line 63: This relates to https://github.com/Chia-Network/chia-blockchain/pull/16081/files#r1306648251.

arvidn (approved):

PR #16157, opened 2023-08-24
Add `add_coin_records` to WalletCoinStore to speed up batch inserting

emlowe:
WalletCoinStore has the function `add_coin_record` to add a single record at a time, but it doesn't have any batch update call that takes a list of records.

The test `test_get_coin_records_rpc_limits` was creating 10k records and inserting them one at a time for the test code. This is very slow, particularly on slow CI runners, or a pi4.

Using the new batch insert function results in a 10x speed improvement in this test.

Also, changed some class variables in `WalletRpcApi` that control the limit to be instance variables. This allows the test to override the limit with a much smaller number.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -5,7 +5,7 @@ import json
 import logging
 import zlib
 from pathlib import Path
-from typing import Any, ClassVar, Dict, List, Optional, Set, Tuple, Union
+from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
 from blspy import AugSchemeMPL, G1Element, G2Element, PrivateKey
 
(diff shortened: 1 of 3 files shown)


xdustinface (changes requested):
You can assign the class vars via the class itself so i guess the below diff should good enough? Works fine for me locally and i guess adding 50 entries without the extra batch inserting should be fine so we could avoid adding the test-only changes in the coin store?

```python
diff --git a/tests/wallet/rpc/test_wallet_rpc.py b/tests/wallet/rpc/test_wallet_rpc.py
index be9548379..775778c0a 100644
--- a/tests/wallet/rpc/test_wallet_rpc.py
+++ b/tests/wallet/rpc/test_wallet_rpc.py
@@ -1863,6 +1863,10 @@ async def test_get_coin_records_rpc_limits(
     api: WalletRpcApi = cast(WalletRpcApi, rpc_server.rpc_api)
     store = wallet_node.wallet_state_manager.coin_store
 
+    # Adjust the limits for faster testing
+    WalletRpcApi.max_get_coin_records_limit = 5
+    WalletRpcApi.max_get_coin_records_filter_items = 5
+
     max_coins = api.max_get_coin_records_limit * 10
     coin_records = [
         WalletCoinRecord(
```

emlowe:
Hmm. I thought I tried that, but it is definitely a smaller patch. I didn't know if a batch insert would be useful going forward. Ok. I think I'll close this PR and make a new one with your suggestion.

emlowe:
Closing PR

PR #16159, opened 2023-08-25
type mismatch for last_time_farmed

dkackman:
### Purpose:

There is a type mismatch when setting `last_height_farmed`:

```
last_time_farmed = uint32(
    await self.service.get_timestamp_for_height(last_height_farmed) if last_height_farmed > 0 else 0
)
```

`get_timestamp_for_height` returns uint64, so not sure when, but at some epoch-seconds that will overflow.

### Current Behavior:

non-breaking since using wider data type

### Testing Notes:

None - visual inspection and correctness

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -3027,7 +3027,7 @@ class WalletRpcApi:
                 last_height_farmed = height
             amount += record.amount
 
-        last_time_farmed = uint32(
+        last_time_farmed = uint64(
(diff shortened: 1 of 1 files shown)


hoffmang9:
We will all long be dead when that overflows, but uint32 doesn't have that luxury so good catch!

dkackman:
> We will all long be dead when that overflows, but uint32 doesn't have that luxury so good catch!

2/7/2106 6:28:16 AM

emlowe (approved):

PR #16175, opened 2023-08-26
Fix comment typo

xearl4:
This fixes a typo in the comment describing the computation of the recently added `lookup_time` field in the harvester->farmer protocol `FarmingInfo` message. `time_taken` is computed using `time.time()`. `time.time()` returns a timestamp in (fractional) seconds. Multiplying seconds by 1_000_000 results in microseconds, not nanoseconds.

### DIFF
--- a/chia/harvester/harvester_api.py
+++ b/chia/harvester/harvester_api.py
@@ -272,7 +272,7 @@ class HarvesterAPI:
             uint32(passed),
             uint32(total_proofs_found),
             uint32(total),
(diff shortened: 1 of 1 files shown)


paninaro (approved):
Thanks for the correction

paninaro:
@ChiaMineJP added you in case there are other locations you want to check.

ChiaMineJP (approved):
lgtm
Thanks for the correction!

Issue #11980, opened 2022-06-19
[Bug] Errors in Log file from latest Beta


Drhicom:
### What happened?

Errors in log file with latest betas

### Version

1.3.6871

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output

```shell
2022-06-19T15:24:43.048 wallet chia.wallet.wallet_node    : ERROR    Exception handling None, Task <Task pending name='Task-17' coro=<WalletNode._process_new_subscriptions() running at chia\wallet\wallet_node.py:346>> got Future <Future pending> attached to a different loop Traceback (most recent call last):
  File "chia\wallet\wallet_node.py", line 346, in _process_new_subscriptions
  File "chia\wallet\util\new_peak_queue.py", line 70, in get
  File "asyncio\queues.py", line 166, in get
RuntimeError: Task <Task pending name='Task-17' coro=<WalletNode._process_new_subscriptions() running at chia\wallet\wallet_node.py:346>> got Future <Future pending> attached to a different loop

2022-06-19T15:24:43.054 wallet chia.wallet.wallet_node    : ERROR    Exception handling None, Task <Task pending name='Task-17' coro=<WalletNode._process_new_subscriptions() running at chia\wallet\wallet_node.py:346>> got Future <Future pending> attached to a different loop Traceback (most recent call last):
  File "chia\wallet\wallet_node.py", line 346, in _process_new_subscriptions
  File "chia\wallet\util\new_peak_queue.py", line 70, in get
  File "asyncio\queues.py", line 166, in get
RuntimeError: Task <Task pending name='Task-17' coro=<WalletNode._process_new_subscriptions() running at chia\wallet\wallet_node.py:346>> got Future <Future pending> attached to a different loop

2022-06-19T15:24:43.058 wallet chia.wallet.wallet_node    : ERROR    Exception handling None, Task <Task pending name='Task-17' coro=<WalletNode._process_new_subscriptions() running at chia\wallet\wallet_node.py:346>> got Future <Future pending> attached to a different loop Traceback (most recent call last):
  File "chia\wallet\wallet_node.py", line 346, in _process_new_subscriptions
  File "chia\wallet\util\new_peak_queue.py", line 70, in get
  File "asyncio\queues.py", line 166, in get
RuntimeError: Task <Task pending name='Task-17' coro=<WalletNode._process_new_subscriptions() running at chia\wallet\wallet_node.py:346>> got Future <Future pending> attached to a different loop

[...]

PR #16204, opened 2023-08-31
`@dataclass` for `DataLayer`

altendky:

### DIFF
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -71,10 +71,10 @@ class DataLayer:
     root_path: Path
     log: logging.Logger
     wallet_rpc_init: Awaitable[WalletRpcClient]
-    _server: Optional[ChiaServer]
     downloaders: List[str]
     uploaders: List[str]
     server_files_location: Path
+    _server: Optional[ChiaServer] = None
     none_bytes: bytes32 = bytes32([0] * 32)
     initialized: bool = False
     _data_store: Optional[DataStore] = None
@@ -148,7 +148,6 @@ class DataLayer:
             log=logging.getLogger(name if name is None else __name__),
             db_path=path_from_root(root_path, db_path_replaced),
             server_files_location=path_from_root(root_path, server_files_replaced),
-            _server=None,
             downloaders=downloaders,
             uploaders=uploaders,
         )
--- a/chia/data_layer/data_layer.py
+++ b/chia/data_layer/data_layer.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import dataclasses
 import json
 import logging
 import os
@@ -8,7 +9,7 @@ import random
 import time
 import traceback
 from pathlib import Path
-from typing import Any, Awaitable, Dict, List, Optional, Set, Tuple, Union, cast
+from typing import Any, Awaitable, Dict, List, Optional, Set, Tuple, Union, cast, final
 
 import aiohttp
 
@@ -62,20 +63,26 @@ async def get_plugin_info(url: str) -> Tuple[str, Dict[str, Any]]:
         return url, {"error": f"ClientError: {e}"}
 
 
+@final
+@dataclasses.dataclass
 class DataLayer:
-    data_store: DataStore
     db_path: Path
     config: Dict[str, Any]
     root_path: Path
     log: logging.Logger
     wallet_rpc_init: Awaitable[WalletRpcClient]
-    state_changed_callback: Optional[StateChangedProtocol] = None
-    wallet_id: uint64
-    initialized: bool
     none_bytes: bytes32
     _server: Optional[ChiaServer]
     downloaders: List[str]
     uploaders: List[str]
+    server_files_location: Path
+    initialized: bool = False
+    _data_store: Optional[DataStore] = None
+    state_changed_callback: Optional[StateChangedProtocol] = None
+    _shut_down: bool = False
+    periodically_manage_data_task: Optional[asyncio.Task[None]] = None
+    _wallet_rpc: Optional[WalletRpcClient] = None
+    _subscription_lock: Optional[asyncio.Lock] = None
 
     @property
     def server(self) -> ChiaServer:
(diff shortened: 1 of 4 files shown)


altendky:
Coverage issues are all guardrail error conditions.

tests/core/data_layer/test_data_rpc.py
```diff
@@ -803,7 +803,11 @@ async def offer_setup_fixture(
 
 async def populate_offer_setup(offer_setup: OfferSetup, count: int) -> OfferSetup:
     if count > 0:
-        for store_setup, value_prefix in {offer_setup.maker: b"\x01", offer_setup.taker: b"\x02"}.items():
+        setups: Tuple[Tuple[StoreSetup, bytes], Tuple[StoreSetup, bytes]] = (
```
xdustinface, line 806: This doesn't look related here?

chia/data_layer/data_layer.py
```diff
@@ -86,38 +93,72 @@ def server(self) -> ChiaServer:
...
-        self.server_files_location = path_from_root(self.root_path, server_files_replaced)
+
+        db_path_replaced: str = config["database_path"].replace("CHALLENGE", config["selected_network"])
+
+        self = cls(
+            config=config,
+            root_path=root_path,
+            wallet_rpc_init=wallet_rpc_init,
+            log=logging.getLogger(name if name is None else __name__),
+            _shut_down=False,
```
xdustinface, line 149:
```suggestion
```

chia/data_layer/data_layer.py
```diff
@@ -86,38 +93,72 @@ def server(self) -> ChiaServer:
...
+            db_path=path_from_root(root_path, db_path_replaced),
+            server_files_location=path_from_root(root_path, server_files_replaced),
+            none_bytes=bytes32([0] * 32),
+            _server=None,
```
xdustinface, line 153:
```suggestion
```

There is no default yet but seems like you could just assign it and drop this.

chia/data_layer/data_layer.py
```diff
@@ -86,38 +93,72 @@ def server(self) -> ChiaServer:
...
+
+        self = cls(
+            config=config,
+            root_path=root_path,
+            wallet_rpc_init=wallet_rpc_init,
+            log=logging.getLogger(name if name is None else __name__),
+            _shut_down=False,
+            db_path=path_from_root(root_path, db_path_replaced),
+            server_files_location=path_from_root(root_path, server_files_replaced),
+            none_bytes=bytes32([0] * 32),
```
xdustinface, line 152: This _could_ be a default also, maybe even not a member at all.

chia/data_layer/data_layer.py
```diff
@@ -62,20 +63,26 @@ async def get_plugin_info(url: str) -> Tuple[str, Dict[str, Any]]:
...
     downloaders: List[str]
     uploaders: List[str]
+    server_files_location: Path
+    initialized: bool = False
+    _data_store: Optional[DataStore] = None
+    state_changed_callback: Optional[StateChangedProtocol] = None
+    _shut_down: bool = False
+    periodically_manage_data_task: Optional[asyncio.Task[None]] = None
+    _wallet_rpc: Optional[WalletRpcClient] = None
+    _subscription_lock: Optional[asyncio.Lock] = None
```
xdustinface, line 85:
```suggestion
    _subscription_lock: asyncio.Lock = field(default_factory=asyncio.Lock)
```

Could be ^ and you could drop the related property?

xdustinface (changes requested):

chia/data_layer/data_layer.py
altendky: Could we leave this?  It looks to be pointless but also looks to be part of a common pattern that we should maybe analyse.  For example, the wallet probably shouldn't need it either.  Also the timelord etc as related to my PRs around removing the unused peer ports in the config.

tests/core/data_layer/test_data_rpc.py
altendky: Hum...  I thought mypy was complaining or something.  But maybe I was just annoyed it wasn't detecting the type or...  Yep, I'll make a separate pr for this.

https://github.com/chia-network/chia-blockchain/commit/680d2833e4965968c17be673f2d7d67fc0ea676d
https://github.com/Chia-Network/chia-blockchain/pull/16257

altendky (commented):
thanks

tests/core/data_layer/test_data_rpc.py
altendky: Oh yeah, this is because `DataLayer` is no longer hashable so we need to avoid hashing it such as as a dict key.

chia/data_layer/data_layer.py
xdustinface: Not sure what the common pattern is you are referring to here or what your plans are with keeping this, but im ok if you see any sense in it. So feel free to close it but i want to mention, there are other classes where we already assign a default value via the dataclass to the optional server like proposed.

chia/data_layer/data_layer.py
xdustinface: I think its just you every time this kind of thing comes up feeling like it needs to be done in a async call for some reason because its from the async package :) But we have bunch of places where we instantiate async members (`Queue`, `Event` and even also `Lock` already) like this and there doesn't seem to be an issue with it. I don't really see a reason to add this extra optional/property thing but feel free to close this.

xdustinface (dismissed):

chia/data_layer/data_layer.py
altendky: https://github.com/chia-network/chia-blockchain/commit/0cc9aa655d2eaf77e033d817aa2b20107d00dd18

I...  think I wasn't thinking straight...  sorry.

xdustinface (dismissed):

chia/data_layer/data_layer.py
altendky: https://github.com/chia-network/chia-blockchain/commit/96f7bbca1238ec3e0dbc1d5dfe6532b755646626

altendky: The issue is with `asyncio.Queue` being created outside the loop.

https://github.com/Chia-Network/chia-blockchain/pull/12041/files

https://github.com/Chia-Network/chia-blockchain/issues/11980

arvidn (approved):

PR #16206, opened 2023-08-31
Warn if running `install-plotter.sh` as root

ChiaMineJP:
Confirmed to work on
- Fedora 35
- Ubuntu 22.04
- MacOS 13

### DIFF
--- a/install-plotter.sh
+++ b/install-plotter.sh
@@ -166,11 +166,6 @@ if [ "$VIRTUAL_ENV" = "" ]; then
   exit 1
 fi
 
-if [ "$(id -u)" = 0 ]; then
-  echo "ERROR: Plotter can not be installed or run by the root user."
-  exit 1
-fi
-
 OS=""
 ARCH="x86-64"
 if [ "$(uname)" = "Linux" ]; then
--- a/install-plotter.sh
+++ b/install-plotter.sh
@@ -166,6 +166,10 @@ if [ "$VIRTUAL_ENV" = "" ]; then
   exit 1
 fi
 
+if [ "$(id -u)" = 0 ]; then
+  echo "WARN: Plotter should not be installed or run by the root user."
+fi
+
 OS=""
 ARCH="x86-64"
 if [ "$(uname)" = "Linux" ]; then


arvidn:
your change to the code appear to be the opposite of the title of the PR.

The check you're removing (appears to) make the script fail if you run it as root. This seems like a quite reasonable check.
Why do you think it should be removed?

ChiaMineJP:
@arvidn Sorry the title was wrong and I updated it.
I was asked if the non-root requirement is truly necessary.
There are some cases where users want to be root when running `install-plotter.sh` and I wanted to hear the comments from reviewers if there are background requirement which definitely needs to be a non-root user.

Starttoaster:
> your change to the code appear to be the opposite of the title of the PR.
> 
> The check you're removing (appears to) make the script fail if you run it as root. This seems like a quite reasonable check. Why do you think it should be removed?

Changes to chia-docker are being considered to allow the script to work as-is. However, that will be a breaking change to existing chia-docker users. This check in the install-plotter script causes it to exit if the user running the script is root. The script works (with this check removed) if the user is running as root. 

Understanding that it's not ideal from a systems administration perspective to advise people to run as root. But, to me, the flip side of this is that it does come across as unnecessary prescription of ideals to explicitly require that the user be not-root for no reason other than general advisement. Should we play the systems administrator role for people running our scripts? That's a philosophical question maybe, I don't know the answer to it.

arvidn:
do you think it would make sense to just demote this to a warning message? But still allow it to start

Starttoaster:
> do you think it would make sense to just demote this to a warning message? But still allow it to start

I'm fine with a warning message if that makes the change more palatable. I just don't personally see the point in forcing the script to exit, outside of an actual technical challenge running as non-root solves. Anyone running as root on a linux server understands, or should understand by themselves, why running as root is not ideal. I wouldn't generally consider it the job of the software I run to instill better sysadmin hygiene habits in me :smile:

ChiaMineJP:
@arvidn @Starttoaster 
I've updated to just warn and not exit.

wallentx:
I think it might be somewhat bad etiquette to force root/sudo within scripts, especially when it comes to a sensitive environment.

I know for any other project I use, I tend to frown when I see that their installer uses `sudo` to execute several things that I might not particularly agree with (i.e. putting binaries in `/sbin`, or force installing packages).

If the user has the appropriate permissions, then they should be able to execute the script via `sudo`, if appropriate for their environment.

Elevated script execution should be the responsibility of the user, since the user should be ultimately play an involved role in executing scripts with elevated permissions, with an understanding of the consequences of doing so on their particular system.

As to completely _disallowing_ this to be executed by root... I'm not sure why a user would want to do anything related to Chia as root, but I don't think that there's a technical reason that would prevent anything from working. I would predict that there are more things that could go wrong by doing this, but I think that's between the user, and their gods.

wallentx (approved):

arvidn (approved):

Starttoaster (approved):

PR #16210, opened 2023-08-31
Fix mypy in main

Quexington:


wallentx:
I don't believe this is needed

PR #16216, opened 2023-09-01
server: Drop `ChiaServer.validate_broadcast_message_type`

xdustinface:
### Purpose:

I might miss the point but this validation there seems strange to me and even more strange that we disconnect/ban from the peers because we tried to initiate sending the "wrong" messages. For me it looks like even sending a message which requires a response doesn't lead to actual issues, does it? 

If there is a point in keeping the `message_requires_reply` checks i might just put it directly into `send_to_all` but seems to me like just dropping it is fine? 

### Current Behavior:

The method validates static messages to peers and disconnects from the peers if the code tries to send a message which expects a response. 

### New Behavior:

It just sends the messages.

### DIFF
--- a/chia/protocols/protocol_state_machine.py
+++ b/chia/protocols/protocol_state_machine.py
@@ -56,12 +56,6 @@ def static_check_sent_message_response() -> None:
         raise AssertionError(f"Overlapping NO_REPLY_EXPECTED and VALID_REPLY_MESSAGE_MAP values: {overlap}")
 
 
-def message_requires_reply(sent: ProtocolMessageTypes) -> bool:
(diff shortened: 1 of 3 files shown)


arvidn:
did you run into a problem with this check?

arvidn:
as far as I can tell, this patch removes the check entirely. I agree that it seems reasonable to just drop the message that we attempt to send erroneously, rather than dropping both the peer connection and the message. However, since this would only happen if we have a bug, dropping the peer connection (and the associated co-routines) might make it more likely we chop off the damaged part of the program.

I also would think we really want an error message in the log if this happens.

PR #16217, opened 2023-09-01
tests: Fix a `BlockCache` instance in `test_weight_proof_extend_new_ses`

xdustinface:
### Purpose:

While this doesn't impact the test currently i noticed this while working on another PR where it became an issue.


arvidn:
if we truly just passed those arguments in the wrong order, you would think mypy would have told us.
Is mypy disabled for this type or file?

arvidn (approved):

xdustinface:
> if we truly just passed those arguments in the wrong order, you would think mypy would have told us.
> Is mypy disabled for this type or file?

Yes its excluded and it would complain if not.

PR #16223, opened 2023-09-01
Run daemon tests in parallel

AmineKhaldi:


arvidn:
I thought I made the default be parrallel. you only need to specify this field for tests that cannot be run in parallel

AmineKhaldi:
Parallel is default now, this is a noop as a result.

PR #16225, opened 2023-09-01
Avoid setting test config items to their default values

AmineKhaldi:


arvidn (dismissed):

altendky (changes requested):
If we want this, shouldn't we have the matrix generation enforce it?

altendky (approved):

arvidn (approved):

PR #16226, opened 2023-09-02
Optimize CAT coin_added

ytx1991:
### Purpose:
This is the actual optimization for the CAT wallet based on the https://github.com/Chia-Network/chia-blockchain/pull/15735.

### Current Behavior:
The CAT will call peer API again to get the coin state and coin spend.

### New Behavior:
The CAT will not call peer API again to get the coin state and coin spend when the CATCoinData is provided.

### Testing Notes:
unit test

### DIFF
--- a/chia/wallet/cat_wallet/cat_wallet.py
+++ b/chia/wallet/cat_wallet/cat_wallet.py
@@ -15,7 +15,6 @@ from chia.types.announcement import Announcement
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.program import Program
 from chia.types.blockchain_format.sized_bytes import bytes32
-from chia.types.coin_spend import CoinSpend
 from chia.types.condition_opcodes import ConditionOpcode
 from chia.types.spend_bundle import SpendBundle
 from chia.util.byte_types import hexstr_to_bytes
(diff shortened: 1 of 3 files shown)


xdustinface (changes requested):

chia/wallet/cat_wallet/cat_wallet.py
```diff
@@ -364,26 +363,37 @@ async def coin_added(
...
                 self.log.debug(f"Exception: {e}, traceback: {traceback.format_exc()}")
 
-    async def puzzle_solution_received(self, coin_spend: CoinSpend, parent_coin: Coin) -> None:
-        coin_name = coin_spend.coin.name()
-        puzzle: Program = Program.from_bytes(bytes(coin_spend.puzzle_reveal))
-        args = match_cat_puzzle(uncurry_puzzle(puzzle))
-        if args is not None:
-            mod_hash, genesis_coin_checker_hash, inner_puzzle = args
-            self.log.info(f"parent: {coin_name.hex()} inner_puzzle for parent is {inner_puzzle}")
+    async def puzzle_solution_received(self, coin: Coin, parent_coin_data: Optional[CATCoinData]) -> None:
```
Quexington, line 387: The `CATCoinData` is not optional anymore right?  I think maybe this function was improperly ported to the function above because the spend path where all of the children are scrubbed from the DB doesn't seem like it would trigger anymore (although I thought there was a test for this so I'm not 100% sure.

xdustinface: I was thinking the same but it is indeed still `Optional` as far as i can tell. Its because it gets assigned inside the nested `if cat_curried_args is not None:` block so it might still be `None` when its passed to `puzzle_solution_received` because the input to `coin_added` is `Optional`, right?

Quexington: Ohhh yeah I see, yes you're right my bad.

Quexington (dismissed):
Good to me but Kevin's comment about a comment should be resolved as well.

xdustinface (approved):

wjblanke (approved):
aok

PR #16227, opened 2023-09-02
Optimize NFT coin_added

ytx1991:
### Purpose:
This is the actual optimization for the NFT wallet based on the https://github.com/Chia-Network/chia-blockchain/pull/15735.

### Current Behavior:
The NFT wallet will call peer API again to get the coin state and coin spend.

### New Behavior:
The CAT will not call peer API again to get the coin state.

### Testing Notes:
unit tests

### DIFF
--- a/chia/wallet/nft_wallet/nft_wallet.py
+++ b/chia/wallet/nft_wallet/nft_wallet.py
@@ -31,7 +31,7 @@ from chia.wallet.lineage_proof import LineageProof
 from chia.wallet.nft_wallet import nft_puzzles
 from chia.wallet.nft_wallet.nft_info import NFTCoinInfo, NFTWalletInfo
 from chia.wallet.nft_wallet.nft_puzzles import NFT_METADATA_UPDATER, create_ownership_layer_puzzle, get_metadata_and_phs
-from chia.wallet.nft_wallet.uncurry_nft import UncurriedNFT
+from chia.wallet.nft_wallet.uncurry_nft import NFTCoinData, UncurriedNFT
 from chia.wallet.outer_puzzles import AssetType, construct_puzzle, match_puzzle, solve_puzzle
 from chia.wallet.payment import Payment
 from chia.wallet.puzzle_drivers import PuzzleInfo, Solver
@@ -58,7 +58,7 @@ _T_NFTWallet = TypeVar("_T_NFTWallet", bound="NFTWallet")
 
 class NFTWallet:
     if TYPE_CHECKING:
-        _protocol_check: ClassVar[WalletProtocol[UncurriedNFT]] = cast("NFTWallet", None)
+        _protocol_check: ClassVar[WalletProtocol[NFTCoinData]] = cast("NFTWallet", None)
 
     wallet_state_manager: Any
     log: logging.Logger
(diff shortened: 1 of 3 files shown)


chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -87,9 +89,17 @@ class UncurriedNFT(Streamable):
     """
     royalty_address: Optional[bytes32]
     trade_price_percentage: Optional[uint16]
+    coin_state: Optional[CoinState] = None
```
xdustinface, line 92: Not necessarily requesting this but any thoughts about using some extra class here like in the CAT case instead of starting to mess with `UncurriedNFT`? 

```python
@streamable
@dataclass(frozen=True)
class NFTCoinData(Streamable):
    uncurried_nft: UncurriedNFT
    parent_coin_state: CoinState
    parent_coin_spend: CoinSpend
```

Then you could drop the streamable additions again from `UncurriedNFT` and the state/spend would not need to be optional.

xdustinface (changes requested):

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -158,44 +157,35 @@ async def get_nft_coin_by_id(self, nft_coin_id: bytes32) -> NFTCoinInfo:
...
         self.log.debug("Puzzle solution received to wallet: %s", self.wallet_info)
-        coin_name = coin_spend.coin.name()
-        puzzle: Program = Program.from_bytes(bytes(coin_spend.puzzle_reveal))
         # At this point, the puzzle must be a NFT puzzle.
         # This method will be called only when the wallet state manager uncurried this coin as a NFT puzzle.
 
-        uncurried_nft = UncurriedNFT.uncurry(*puzzle.uncurry())
         assert uncurried_nft is not None
+        assert uncurried_nft.coin_spend is not None
+        assert uncurried_nft.coin_state is not None
```
Quexington, line 179: Why were these not necessary before? I don't mind it obviously, just curious.

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -247,16 +239,11 @@ async def puzzle_solution_received(self, coin_spend: CoinSpend, peer: WSChiaConn
...
+        parent_coin = uncurried_nft.coin_state.coin
+        confirmed_height = (
+            None if uncurried_nft.coin_state.spent_height is None else uint32(uncurried_nft.coin_state.spent_height)
         )
 
-        if coin_states is not None:
-            parent_coin = coin_states[0].coin
-            confirmed_height = None if coin_states[0].spent_height is None else uint32(coin_states[0].spent_height)
-
         if parent_coin is None or confirmed_height is None:
```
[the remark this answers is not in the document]
ytx1991: Yes, since the NFT wallet doesn't create any derived puzzle hash, the only code path to the coin_added is the determine_coin_type in the wallet manager. So we expect there must be a parent coin data passed in.

chia/wallet/nft_wallet/uncurry_nft.py
ytx1991: I think we still need the Streamable annotation since the members of a Streamable have to be Streamable, too, right? I think use a unified naming convention is a good idea.

xdustinface: > I think we still need the Streamable annotation since the members of a Streamable have to be Streamable, too, right? I think using a unified naming convention is a good idea. I will change this.

Oh yeah your right, the streamable is still required then!

chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -158,44 +157,36 @@ async def get_nft_coin_by_id(self, nft_coin_id: bytes32) -> NFTCoinInfo:
...
-        puzzle: Program = Program.from_bytes(bytes(coin_spend.puzzle_reveal))
         # At this point, the puzzle must be a NFT puzzle.
         # This method will be called only when the wallet state manager uncurried this coin as a NFT puzzle.
 
-        uncurried_nft = UncurriedNFT.uncurry(*puzzle.uncurry())
-        assert uncurried_nft is not None
+        assert data is not None
+        assert data.uncurried_nft is not None
+        assert data.parent_coin_state is not None
+        assert data.parent_coin_spend is not None
```
xdustinface, line 178:
```suggestion
```

This all seems to be not needed?

chia/wallet/nft_wallet/uncurry_nft.py
```diff
@@ -89,7 +91,11 @@ class UncurriedNFT(Streamable):
     trade_price_percentage: Optional[uint16]
 
     @classmethod
-    def uncurry(cls: Type[_T_UncurriedNFT], mod: Program, curried_args: Program) -> Optional[_T_UncurriedNFT]:
+    def uncurry(
+        cls: Type[_T_UncurriedNFT],
+        mod: Program,
+        curried_args: Program,
```
xdustinface, line 97: This reformatting could be reverted now also.

xdustinface (approved):

wjblanke (approved):
aok

PR #16232, opened 2023-09-02
Fix issue with GUI and `chia plotters` when making compressed plots

emlowe:
Don't use `no-direct-downloads` at all for `cudaplot`.

Will not set it when using `chia plotters` and will also not set it from the GUI despite the checkbox
Fixes https://github.com/Chia-Network/chia-blockchain/issues/16233

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -865,15 +865,12 @@ class WebSocketServer:
         # Options only applicable for cudaplot
         if plot_type == "cudaplot":
             device_index = request.get("device", None)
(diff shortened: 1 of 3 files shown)


ChiaMineJP (dismissed):
Almost identical to my PR (https://github.com/Chia-Network/chia-blockchain/pull/16236)

Please check the PR for the GUI for this issue too:
https://github.com/Chia-Network/chia-blockchain-gui/pull/2061

wjblanke (approved):
always nice to delete things

PR #16238, opened 2023-09-04
full_node: Move wallet updates into a separate task

xdustinface:
### Purpose:

Move wallet updates into a separate task which works them off from a queue. This is as preparation for an updated wallet sync protocol.

Requires #16239 to pass the all the tests.

### Current Behavior:

Wallet updates are processed in peak post processing.

### New Behavior:

Wallet updates are processed in a separate task. 

### Based on:
- #16062

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -35,7 +35,7 @@ from chia.full_node.hint_store import HintStore
 from chia.full_node.mempool_manager import MempoolManager
 from chia.full_node.signage_point import SignagePoint
 from chia.full_node.subscriptions import PeerSubscriptions
-from chia.full_node.sync_store import SyncStore
+from chia.full_node.sync_store import Peak, SyncStore
 from chia.full_node.tx_processing_queue import TransactionQueue
 from chia.full_node.weight_proof import WeightProofHandler
 from chia.protocols import farmer_protocol, full_node_protocol, timelord_protocol, wallet_protocol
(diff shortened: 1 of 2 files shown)


chia/full_node/full_node.py
```diff
@@ -89,6 +89,14 @@ class PeakPostProcessingResult:
     lookup_coin_ids: List[bytes32]  # The coin IDs that we need to look up to notify wallets of changes
 
 
+@dataclasses.dataclass(frozen=True)
+class WalletUpdate:
```
arvidn, line 93: I think it would be good to add a comment describing what this class is used for. It's not obvious from its name.

does `fork_height` refer to a reorg? The name suggests it does, but I wouldn't expect to find that field in here.

chia/full_node/full_node.py
```diff
@@ -1122,52 +1138,52 @@ def get_peers_with_peak(self, peak_hash: bytes32) -> List[WSChiaConnection]:
...
+                state = CoinStateUpdate(
+                    wallet_update.peak.height,
+                    wallet_update.fork_height,
+                    wallet_update.peak.header_hash,
+                    list(changes),
+                )
+                await connection.send_message(make_msg(ProtocolMessageTypes.coin_state_update, state))
+
+        # Tell wallets about the new peak
+        new_peak_message = make_msg(
```
arvidn, line 1177: what would happen if we told wallets about the new peak first, and then about the coin state updates?
wouldn't that let us remove the race cache?

arvidn (approved):

chia/full_node/full_node.py
xdustinface: No we discussed this already. Its also about a race between subscribing which returns the current state and the first actual state update which happens in peak processing and also we would need to cut off old nodes which send in the wrong direction if we get rid of the race cache.

chia/full_node/full_node.py
xdustinface: If there was a reorg, yes, it tells the wallet about the reorg but usually its just from `peak_height - 1` to `peak_height`. Its how the protocol works, so why wouldn't you expect it to be in here? The `WalletUpdate` class is simply a dataclass container to bundle some items to put them into a queue.

PR #16239, opened 2023-09-05
wallet: Fix and improve untrusted race caching

xdustinface:
### Purpose:

- Fixes the same memory leak as in #16144 
- Adjusts how the race cache works to make it height based with proper reorgs/cleanups. 
- Makes the race cache peer based instead of sharing it among all untrusted peers.
- Processes reorged coin states upfront to avoid adding them to the race cache first for no reason. 
- Fixes a bunch of different failures happening in tests of #16238 because `WalletNode.add_states_from_peer` gets called with invalid height/hash due to races during short syncs.

### Testing Notes:

Added unit tests for the race cache helpers and i spent a lot time debugging and understanding all the related places here while running tests.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -108,12 +108,6 @@ class Balance(Streamable):
     pending_coin_removal_count: uint32 = uint32(0)
 
 
-@dataclasses.dataclass(frozen=True)
-class PeerPeak:
-    height: uint32
-    hash: bytes32
-
-
 @dataclasses.dataclass
 class WalletNode:
     config: Dict[str, Any]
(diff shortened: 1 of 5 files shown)


jack60612 (approved):
lgtm

chia/wallet/wallet_node.py
```diff
@@ -863,6 +848,9 @@ async def add_states_from_peer(
                 # only one peer told us to rollback so only clear for that peer
                 cache.clear_after_height(fork_height)
                 self.log.info(f"clear_after_height {fork_height} for peer {peer}")
+                if not trusted:
+                    # Rollback race_cache not in clear_after_height to avoid applying rollbacks from new peak processing
+                    cache.rollback_race_cache(fork_height=fork_height)
```
arvidn, line 853: I take it we did not used to clear the cache at all, right?

chia/wallet/wallet_node.py
```diff
@@ -1174,6 +1164,7 @@ async def sync_from_untrusted_close_to_peak(self, new_peak_hb: HeaderBlock, peer
             else:
                 backtrack_fork_height = new_peak_hb.height - 1
 
+            cache = self.get_cache_for_peer(peer)
```
arvidn, line 1167: what does this return? it can't be a `Dict[uint32, Set[CoinState]]`, because you call `get_race_cache()` on it below

xdustinface: The `PeerRequestCache` where i added `get_race_cache` in this patch here
https://github.com/Chia-Network/chia-blockchain/blob/208485075c335cfe8a715fad10192b68a0afca25/chia/wallet/wallet_node.py#L197-L200

arvidn (approved):

PR #16241, opened 2023-09-05
run_block_generator2()

arvidn:

This updates the hard-fork to use `run_block_generator2()` as intended. this alters the cost block generators as the cost of the ROM itself is not charged (once the hard fork activates)

The new `run_block_generator2()` can be reviewed here: https://github.com/Chia-Network/chia_rs/blob/main/src/gen/run_block_generator.rs#L111-L191

Since `testnet10` has already activated the hard-fork, this fix needs a separate activation height there. In mainnet, the activation stays the same as the original hard fork.

There are new pre-generated test blockchains in the `test-cache` repository, with the `_harfork` suffix.


arvidn:
I'm removing `coverage-diff` label. the one missing line points into a benchmark, which we don't instrument for coverage

chia/simulator/block_tools.py
```diff
@@ -130,6 +130,10 @@
     "rom_bootstrap_generator.clsp", package_or_requirement="chia.consensus.puzzles"
 )
 
+DESERIALIZE_MOD = load_serialized_clvm_maybe_recompile(
+    "chialisp_deserialisation.clsp", package_or_requirement="chia.consensus.puzzles"
+)
+
```
richardkiss, line 136: Might be worth adding a `assert DESERIALIZE_MOD.get_tree_hash() == "xxx"` (for some `xxx`)  here. The notion that we "maybe recompile" the clsp file is probably not a good idea.

richardkiss: I realize this is just a test though, so maybe it doesn't matter too much.

chia/simulator/block_tools.py
```diff
@@ -1844,17 +1848,31 @@ def compute_cost_test(generator: BlockGenerator, cost_per_byte: int, hard_fork:
...
+        cost, result = generator.program._run(INFINITE_COST, MEMPOOL_MODE | ALLOW_BACKREFS, DESERIALIZE_MOD, blocks)
+        clvm_cost += cost
+
+        for spend in result.first().as_iter():
+            spend = spend.rest()  # skip parent coin id
+            puzzle = spend.first()
+            spend = spend.rest()  # skip puzzle
+            spend = spend.rest()  # skip amount
+            solution = spend.first()
+
```
richardkiss, line 1866: Might consider

```
puzzle = spend.at("rf")
solution = spend.at("rrrf")
```

chia/simulator/block_tools.py
```diff
@@ -1844,17 +1848,31 @@ def compute_cost_test(generator: BlockGenerator, cost_per_byte: int, hard_fork:
...
+            cost, result = puzzle._run(INFINITE_COST, MEMPOOL_MODE, solution)
+            clvm_cost += cost
+            condition_cost += conditions_cost(result, hard_fork)
+
+    else:
+        block_program_args = Program.to([[bytes(g) for g in generator.generator_refs]])
+        clvm_cost, result = GENERATOR_MOD._run(INFINITE_COST, MEMPOOL_MODE, generator.program, block_program_args)
+
+        for res in result.first().as_iter():
+            res = res.rest()  # skip parent coin id
```
richardkiss, line 1876: Ditto here: `res = res.at("rrrf")`

chia/util/initial-config.yaml
```diff
@@ -77,6 +77,8 @@ network_overrides: &network_overrides
       # planned 2.0 release is July 26, height 2965036 on testnet
       # 1 week later
       HARD_FORK_HEIGHT: 2997292
+      # November 2023
```
richardkiss, line 80: 2023?

richardkiss (dismissed):
I don't quite understand the intricacies of all the new test branches, although I do have the idea that they have to do with different costs for the new `run_block_generator2`. I assume the new costs are always lower than the old costs, which will give us a slightly simpler hard fork discontinuity (non-smoothness) of expecting the committed costs to be below the actual block_generator2 costs (maybe even by a constant?).

chia/util/initial-config.yaml
arvidn: yes, these are the overrides for `testnet10`. The hard fork and chip-13 have already activated on testnet

chia/simulator/block_tools.py
arvidn: yeah, good point. This is true for most of our puzzles I think.

tests/conftest.py
```diff
@@ -290,7 +306,7 @@ def default_2000_blocks_compact(bt, consensus_mode):
 def default_10000_blocks_compact(bt, consensus_mode):
     from tests.util.blockchain import persistent_blocks
 
-    if consensus_mode == ConsensusMode.SOFT_FORK4:
+    if consensus_mode == ConsensusMode.SOFT_FORK4 or consensus_mode == ConsensusMode.HARD_FORK_2_0:
```
arvidn, line 309: I'm not sure why I had to disable these tests. perhaps it takes a long time to generate these chains. I'll try to generate these over the weekend. This and `default_10000_blocks`

arvidn:
> I assume the new costs are always lower than the old costs, which will give us a slightly simpler hard fork discontinuity (non-smoothness) of expecting the committed costs to be below the actual block_generator2 costs

correct.

> (maybe even by a constant?)

it won't be constant. the most expensive thing the generator ROM does is to compute the puzzle hashes of puzzles. This will be free after the hard-fork, so the "savings" depend on how expensive the tree-hashes are, essentially.

richardkiss (approved):

wjblanke (approved):
aok thanks guys!

PR #16243, opened 2023-09-05
Annotate test_weight_proof.py

AmineKhaldi:


tests/weight_proof/test_weight_proof.py
```diff
@@ -485,58 +470,53 @@ async def test_weight_proof_extend_multiple_ses(self, default_1000_blocks, block
...
+    #         curr = blocks[curr.prev_hash]
+    #     assert len(sub_height_to_hash) == peak_height + 1
+    #     block_cache = BlockCache(blocks, headers, sub_height_to_hash, sub_epoch_summaries)
+    #     wpf = WeightProofHandler(DEFAULT_CONSTANTS, block_cache)
+    #     wp = await wpf._create_proof_of_weight(sub_height_to_hash[peak_height - 50])
+    #     valid, _ = wpf.validate_weight_proof_single_proc(wp)
+
+    #     await connection.close()
+    #     assert valid
+    #     print(f"size of proof is {get_size(wp)}")
```
arvidn, line 515: instead of commenting this out, could it be moved into its own file under `tools`?
There are some other debugging utilities in there too

arvidn (commented):
looks good, but I'm a bit hesitant to just comment that test out. The comment suggests that it's not actually a test, but a debugging utility that was marked as a test.

Was it @almogdepaz that made it by chance?

arvidn (approved):

PR #16255, opened 2023-09-05
Add valid_times to Offer object

Quexington:
This PR adds a property to the `Offer` driver to get the information about what times the offer will be valid for.  It also adds this information to the summary returned by the `.summary()` method and the `get_offer_summary` RPC.

### DIFF
--- a/chia/wallet/trading/offer.py
+++ b/chia/wallet/trading/offer.py
@@ -152,7 +152,7 @@ class Offer:
         object.__setattr__(self, "_additions", adds)
         object.__setattr__(self, "_conditions", None)
 
-    def conditions(self) -> List[Condition]:
+    def conditions(self) -> Dict[Coin, List[Condition]]:
         if self._conditions is None:
             conditions: Dict[Coin, List[Condition]] = {}
             max_cost = DEFAULT_CONSTANTS.MAX_BLOCK_COST_CLVM
@@ -167,10 +167,23 @@ class Offer:
                     raise ValidationError(Err.BLOCK_COST_EXCEEDS_MAX, "computing conditions for CoinSpend")
             object.__setattr__(self, "_conditions", conditions)
         assert self._conditions is not None, "self._conditions is None"
-        return [c for conditions in self._conditions.values() for c in conditions]
+        return self._conditions
 
-    def valid_times(self) -> ConditionValidTimes:
-        return parse_timelock_info(self.conditions())
+    def valid_times(self) -> Dict[Coin, ConditionValidTimes]:
+        return {coin: parse_timelock_info(conditions) for coin, conditions in self.conditions().items()}
+
+    def absolute_valid_times_ban_relatives(self) -> ConditionValidTimes:
+        valid_times: ConditionValidTimes = parse_timelock_info(
+            [c for conditions in self.conditions().values() for c in conditions]
+        )
+        if (
+            valid_times.max_secs_after_created is not None
+            or valid_times.min_secs_since_created is not None
+            or valid_times.max_blocks_after_created is not None
+            or valid_times.min_blocks_since_created is not None
+        ):
+            raise ValueError("Offers with relative timelocks are not currently supported")
+        return valid_times
 
     def additions(self) -> List[Coin]:
         return [c for additions in self._additions.values() for c in additions]
@@ -310,7 +323,12 @@ class Offer:
         for key, value in self.driver_dict.items():
             driver_dict[key.hex()] = value.info
 
-        return keys_to_strings(offered_amounts), keys_to_strings(requested_amounts), driver_dict, self.valid_times()
+        return (
+            keys_to_strings(offered_amounts),
+            keys_to_strings(requested_amounts),
+            driver_dict,
+            self.absolute_valid_times_ban_relatives(),
+        )
 
     # Also mostly for the UI, returns a dictionary of assets and how much of them is pended for this offer
     # This method is also imperfect for sufficiently complex spends
--- a/chia/wallet/trading/offer.py
+++ b/chia/wallet/trading/offer.py
@@ -150,6 +150,7 @@ class Offer:
             if max_cost < 0:
                 raise ValidationError(Err.BLOCK_COST_EXCEEDS_MAX, "compute_additions for CoinSpend")
         object.__setattr__(self, "_additions", adds)
+        object.__setattr__(self, "_conditions", None)
 
     def conditions(self) -> List[Condition]:
         if self._conditions is None:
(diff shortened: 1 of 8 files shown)


wjblanke (dismissed):
aok

tests/wallet/rpc/test_wallet_rpc.py
```diff
@@ -1148,7 +1148,22 @@ async def test_offer_endpoints(wallet_rpc_environment: WalletRpcTestEnvironment)
...
     id, advanced_summary = await wallet_1_rpc.get_offer_summary(offer, advanced=True)
     assert id == offer.name()
-    assert summary == {"offered": {"xch": 5}, "requested": {cat_asset_id.hex(): 1}, "infos": driver_dict, "fees": 1}
+    assert summary == {
+        "offered": {"xch": 5},
+        "requested": {cat_asset_id.hex(): 1},
+        "infos": driver_dict,
+        "fees": 1,
+        "valid_times": {
+            "max_blocks_after_created": None,
```
arvidn, line 1157: these relative time locks; they are relative a specific coin. I assume an offer can contain multiple coins with (potentially) separate relative time locks.

At this point, when we have the offer file, offering concrete coins, we (can) know their creation times and heights. I think these fields could be boiled down to only the most restrictive *absolute* fields.

i.e. "max_height", "min_height", "max_seconds", "min_seconds"

I don't think you have enough information to interpret the relative timestamps in this RPC

chia/wallet/trading/offer.py
```diff
@@ -148,6 +150,27 @@ def __post_init__(self) -> None:
...
+
+    def conditions(self) -> List[Condition]:
+        if self._conditions is None:
+            conditions: Dict[Coin, List[Condition]] = {}
+            max_cost = DEFAULT_CONSTANTS.MAX_BLOCK_COST_CLVM
+            for cs in self._bundle.coin_spends:
+                try:
+                    cost, conds = cs.puzzle_reveal.run_with_cost(max_cost, cs.solution)
+                    max_cost -= cost
+                    conditions[cs.coin] = parse_conditions_non_consensus(conds.as_iter())
```
arvidn, line 163: at this point, you need to capture the coin's timestamp and creation height and "resolve" the relative time locks based on it.

Maybe we don't have access to the coin database here though.

chia/wallet/trading/offer.py
```diff
@@ -148,6 +150,27 @@ def __post_init__(self) -> None:
...
+                except Exception:  # pragma: no cover
+                    continue
+                if max_cost < 0:  # pragma: no cover
+                    raise ValidationError(Err.BLOCK_COST_EXCEEDS_MAX, "computing conditions for CoinSpend")
+            object.__setattr__(self, "_conditions", conditions)
```
arvidn, line 168: this looks like some trickery to fool `mypy`.. if `self._conditions` is an `Optional[]`, why not just use normal assignment?

chia/wallet/trading/offer.py
```diff
@@ -148,6 +150,27 @@ def __post_init__(self) -> None:
...
+        assert self._conditions is not None, "self._conditions is None"
+        return [c for conditions in self._conditions.values() for c in conditions]
```
arvidn, line 170: at this point you "launder" away the coins, making all relative time locks meaningless

arvidn (changes requested):

chia/wallet/trading/offer.py
Quexington: We don't necessarily need to do this computation for every `Offer` that we deal with so I'm basically making sure it only happens when we need it.  We set it to `None` during the `__post_init__` and only if we try to access `.conditions` do we do the computation.

Quexington:
@arvidn To address your concerns with the relative timelocks generally, I agree with you that they are mostly meaningless because they are being talked about at the level of the spend bundle. They mostly exist as part of this class for completeness and when presented in this context don't make a ton of sense but they are easily ignored.

They potentially take on more meaning when we talk about offers involving a revolving door of singletons because you can put a relative timelock of 100 blocks or so to make sure you have ample opportunity to change things in between offer takes.  We can also add better infrastructure to make sure that the relative timelocks always get placed on the newest coin or something like that.  There's a _path_ to them being relevant although right now you are correct, they are a bit silly right now.

chia/wallet/trading/offer.py
arvidn: is it different from:
```
self._conditions = conditions
```
?

arvidn:
They are misleading though. I think it would be better to remove them from the RPC response until we support them. And I would expect the way we will support them in the future is to turn the relative conditions into absolute ones.
Right now, it's *impossible* to do anything with them. They are meaningless without knowing which coin (and specifically its creation height and time).

I suppose the only use they have now is to indicate that there's a time lock that we don't support present in the offer. But in that case, it would probably be better to just not allow any interaction with it.

chia/wallet/trading/offer.py
```diff
@@ -148,6 +150,27 @@ def __post_init__(self) -> None:
...
+
+    def valid_times(self) -> ConditionValidTimes:
+        return parse_timelock_info(self.conditions())
```
arvidn, line 173: this function will take the *min* of all relative valid-before conditions. this is nonsensical across different coins. The relative conditions need to be resolved into absolute ones (given that we know the coin's creation height and time) *before* we take the *min* of them. The same thing applies to the valid-after conditions.

chia/wallet/trading/offer.py
Quexington: Oh I see I thought you were talking about the spot in `__post_init__`.  I think we need to use this because the Offer class is a frozen dataclass.  This mechanism is how the `_additions` property etc. work as well.

arvidn: I see

chia/wallet/trading/offer.py
```diff
@@ -148,6 +150,40 @@ def __post_init__(self) -> None:
...
+        if self._conditions is None:
+            conditions: Dict[Coin, List[Condition]] = {}
+            max_cost = DEFAULT_CONSTANTS.MAX_BLOCK_COST_CLVM
+            for cs in self._bundle.coin_spends:
+                try:
+                    cost, conds = cs.puzzle_reveal.run_with_cost(max_cost, cs.solution)
+                    max_cost -= cost
+                    conditions[cs.coin] = parse_conditions_non_consensus(conds.as_iter())
+                except Exception:  # pragma: no cover
+                    continue
```
arvidn, line 165: I suppose getting these conditions is best-effort. I can imagine spending a long time trouble-shooting an issue, to then find that this throws without leaving a trace

arvidn (dismissed):

chia/wallet/trading/offer.py
Quexington: Yeah this mimics the additions logic above.  Some offers raise (which is potentially not great and will be fixed in the future) and we need to ignore those ones.

wjblanke (dismissed):
aok

wjblanke (approved):
aok

PR #16256, opened 2023-09-05
DID wallet coin_added

ytx1991:
### Purpose:
Prevent redundant peer call in the DID wallet.

### Current Behavior:
DID will call peer API again to get coin state and coin spend.

### New Behavior:
Reuse the coin state and coin spend got in the wallet manager.


tests/wallet/did_wallet/test_did.py
```diff
@@ -793,10 +793,8 @@ async def test_did_transfer(self, self_hostname, two_wallet_nodes, with_recovery
         assert metadata["GitHub"] == "测试"
 
         # Test match_hinted_coin
-        assert await did_wallet_2.match_hinted_coin(
-            await did_wallet_2.get_coin(),
-            new_puzhash,
-        )
+        if with_recovery:
+            assert await did_wallet_2.match_hinted_coin(await did_wallet_2.get_coin(), new_puzhash)
```
Quexington, line 797: Why this change?

ytx1991: In the without recovery info mode, the recovery setting will be reset so the match will fail.

Quexington: But why did it break in this PR?  This is supposed to be an optimization right so why did the behavior change?

ytx1991: I thought there is a bug, revert the change.

chia/wallet/did_wallet/did_wallet.py
```diff
@@ -376,7 +394,7 @@ async def coin_added(self, coin: Coin, _: uint32, peer: WSChiaConnection, coin_d
             assert full_puzzle.get_tree_hash() == coin.puzzle_hash
         if self.did_info.temp_coin is not None:
             self.wallet_state_manager.state_changed("did_coin_added", self.wallet_info.id)
-
+        print(f"wallet {self.get_name()} {did_data.num_verification} {self.did_info.num_of_backup_ids_needed}")
```
Quexington, line 397: Is this a leftover debug line?

Quexington (dismissed):

xdustinface (changes requested):
Seems like the last catch up merge messed something up here and there is one unrelated formatting change.

xdustinface (commented):
pre-commit (`isort`) is still broken.

xdustinface (approved):

wjblanke (approved):
aok

PR #16259, opened 2023-09-05
setup_simulators_and_wallets() with constants

arvidn:
### Purpose:

simplify setup_simulators_and_wallets() and setup_simulators_and_wallets_service() to pass in the consensus constants directly.

The underlying functions that our test fixtures use, `setup_simulators_and_wallets()` and `setup_simulators_and_wallets_service()` implicitly use `test_constants` as their consensus constants. They take a dictionary that allows a caller to patch up specific constants for specific tests (which we only use in one place, and it's not convenient).

Long term, all our fixtures that set up nodes and wallets need to be parameterized by `ConsensusMode`, which means the fixture need to be able to specify which constants to construct the nodes and wallets with. This patch is a step along that way, to make it possible to specify exactly which constants to use from the caller.

There is no (intentional) change in behavior. This prepares for a gradual transition to increased parameterization of our test fixtures.

### DIFF
--- a/chia/simulator/setup_nodes.py
+++ b/chia/simulator/setup_nodes.py
@@ -16,7 +16,7 @@ from chia.harvester.harvester_api import HarvesterAPI
 from chia.protocols.shared_protocol import Capability
 from chia.server.server import ChiaServer
 from chia.server.start_service import Service
-from chia.simulator.block_tools import BlockTools, create_block_tools_async, test_constants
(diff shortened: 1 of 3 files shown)


AmineKhaldi (approved):

arvidn:
This PR is superseded by https://github.com/Chia-Network/chia-blockchain/pull/15977

PR #16261, opened 2023-09-06
wallet: Drop `is_peer_synced` / More cache usage

xdustinface:
### Purpose:

Drop `is_peer_synced` and use the here introduced `WalletNode.get_timestamp_for_height_from_peer` to fetch the timestamp during peak processing for more usage of the timestamp caching and also i wanted to split out `WalletNode.timstamp_in_sync` logic for the wallet protocol update which already contains the last transaction block timestamp in the peak message.

### Current Behavior:

The wallet fetches the block/timestamp twice from the full node for each peak.

### New Behavior:

The wallet fetches the block/timestamp only once and uses the cache after. 

### Testing Notes:

I added a test for `WalletNode.is_timestamp_in_sync` and `WalletNode.get_timestamp_for_height_from_peer`.

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -940,22 +940,8 @@ class WalletNode:
         await self.update_ui()
         return still_connected and self._server is not None and peer.peer_node_id in self.server.all_connections
 
-    async def is_peer_synced(self, peer: WSChiaConnection, height: uint32) -> Optional[uint64]:
-        # Get last timestamp
-        last_tx: Optional[HeaderBlock] = await fetch_last_tx_from_peer(height, peer)
-        latest_timestamp: Optional[uint64] = None
-        if last_tx is not None:
-            assert last_tx.foliage_transaction_block is not None
-            latest_timestamp = last_tx.foliage_transaction_block.timestamp
-
-        # Return None if not synced
-        if (
-            latest_timestamp is None
-            or self.config.get("testing", False) is False
-            and latest_timestamp < uint64(time.time()) - 600
-        ):
-            return None
-        return latest_timestamp
+    def is_timestamp_in_sync(self, timestamp: uint64) -> bool:
+        return self.config.get("testing", False) or uint64(time.time()) - timestamp < 600
 
     def is_trusted(self, peer: WSChiaConnection) -> bool:
         return self.server.is_trusted_peer(peer, self.config.get("trusted_peers", {}))
@@ -1012,25 +998,29 @@ class WalletNode:
                 neither.append(node)
         return synced_and_trusted + synced + trusted + neither
 
-    async def get_timestamp_for_height(self, height: uint32) -> uint64:
+    async def get_timestamp_for_height_from_peer(self, height: uint32, peer: WSChiaConnection) -> Optional[uint64]:
         """
         Returns the timestamp for transaction block at h=height, if not transaction block, backtracks until it finds
         a transaction block
         """
-        for cache in self.peer_caches.values():
-            cache_ts: Optional[uint64] = cache.get_height_timestamp(height)
-            if cache_ts is not None:
-                return cache_ts
+        cache = self.get_cache_for_peer(peer)
+        cached_timestamp = cache.get_height_timestamp(height)
+        if cached_timestamp is not None:
+            return cached_timestamp
 
+        last_tx_block = await fetch_last_tx_from_peer(height, peer)
+        if last_tx_block is None:
+            return None
+
+        assert last_tx_block.foliage_transaction_block is not None
+        self.get_cache_for_peer(peer).add_to_blocks(last_tx_block)
+        return last_tx_block.foliage_transaction_block.timestamp
+
+    async def get_timestamp_for_height(self, height: uint32) -> uint64:
         for peer in self.get_full_node_peers_in_order():
-            last_tx_block = await fetch_last_tx_from_peer(height, peer)
-            if last_tx_block is None:
-                continue
-
-            assert last_tx_block.foliage_transaction_block is not None
-            self.get_cache_for_peer(peer).add_to_blocks(last_tx_block)
-            return last_tx_block.foliage_transaction_block.timestamp
-
+            timestamp = await self.get_timestamp_for_height_from_peer(height, peer)
+            if timestamp is not None:
+                return timestamp
         raise PeerRequestException("Error fetching timestamp from all peers")
 
     async def new_peak_wallet(self, new_peak: NewPeakWallet, peer: WSChiaConnection) -> None:
@@ -1066,8 +1056,8 @@ class WalletNode:
             # dont disconnect from peer, this might be a reorg
             return
 
-        latest_timestamp: Optional[uint64] = await self.is_peer_synced(peer, new_peak_hb.height)
-        if latest_timestamp is None:
+        latest_timestamp = await self.get_timestamp_for_height_from_peer(new_peak_hb.height, peer)
+        if latest_timestamp is None or not self.is_timestamp_in_sync(latest_timestamp):
             if trusted:
                 self.log.debug(f"Trusted peer {peer.get_peer_info()} is not synced.")
             else:
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -1003,13 +1003,12 @@ class WalletNode:
         a transaction block
         """
         cache = self.get_cache_for_peer(peer)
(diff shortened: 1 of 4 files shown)


chia/wallet/wallet_node.py
```diff
@@ -1022,25 +1008,29 @@ def get_full_node_peers_in_order(self) -> List[WSChiaConnection]:
...
+        cache = self.get_cache_for_peer(peer)
+        cached_timestamp = cache.get_height_timestamp(height)
+        if cached_timestamp is not None:
+            return cached_timestamp
 
-        for peer in self.get_full_node_peers_in_order():
-            last_tx_block = await fetch_last_tx_from_peer(height, peer)
-            if last_tx_block is None:
-                continue
+        last_tx_block = await fetch_last_tx_from_peer(height, peer)
```
arvidn, line 1021: "last_tx" sounds like the most recent, but you specify a height, which suggests that you get the block at the specified height. Which one is it?

chia/wallet/wallet_node.py
```diff
@@ -1022,25 +1008,29 @@ def get_full_node_peers_in_order(self) -> List[WSChiaConnection]:
...
-            assert last_tx_block.foliage_transaction_block is not None
-            self.get_cache_for_peer(peer).add_to_blocks(last_tx_block)
-            return last_tx_block.foliage_transaction_block.timestamp
+        assert last_tx_block.foliage_transaction_block is not None
+        self.get_cache_for_peer(peer).add_to_blocks(last_tx_block)
+        return last_tx_block.foliage_transaction_block.timestamp
 
+    async def get_timestamp_for_height(self, height: uint32) -> uint64:
+        for peer in self.get_full_node_peers_in_order():
+            timestamp = await self.get_timestamp_for_height_from_peer(height, peer)
```
arvidn, line 1031: it's not obvious that picking the first hit is a good strategy. What if you have a trusted peer with this information, shouldn't you prefer that?

chia/wallet/wallet_node_api.py
```diff
@@ -53,8 +53,12 @@ async def new_peak_wallet(self, peak: wallet_protocol.NewPeakWallet, peer: WSChi
                 peer for peer in full_node_connections if not self.wallet_node.is_trusted(peer) and not peer.closed
             ]
 
-            # Check for untrusted peers first to avoid calling is_peer_synced if not required
-            if len(untrusted_peers) > 0 and await self.wallet_node.is_peer_synced(peer, peak.height):
+            # Check for untrusted peers to avoid fetching the timestamp if not required
+            if len(untrusted_peers) > 0:
+                timestamp = await self.wallet_node.get_timestamp_for_height_from_peer(peak.height, peer)
```
arvidn, line 58: if we have a trusted node, do we always have access to the timestamp then?

arvidn (commented):
looks good. jus a few questions

chia/wallet/wallet_node.py
xdustinface: There is some heavy walk back logic backed into `fetch_last_tx_from_peer`. See https://github.com/Chia-Network/chia-blockchain/blob/208485075c335cfe8a715fad10192b68a0afca25/chia/wallet/util/wallet_sync_utils.py#L51C11-L66

Thats what i want to get rid of with having the last timestamp included in the initial message with the new protocol and also why #15500 returns them.

chia/wallet/wallet_node.py
xdustinface: The first might be a trusted one, see https://github.com/Chia-Network/chia-blockchain/blob/208485075c335cfe8a715fad10192b68a0afca25/chia/wallet/wallet_node.py#L988-L1013

chia/wallet/wallet_node_api.py
xdustinface: This logic explicitly request the timestamp from the trusted peer to check if its synced. If its synced we disconnect from untrusted peers. This `if len(untrusted_peers) > 0:` is only to avoid requesting the timestamp if there are anyway no untrusted peers to disconnect from.

chia/wallet/wallet_node.py
arvidn: I think the function you link to also has a confusing name. what does "last_tx" refer to exactly? It doesn't return a transaction, which typically is what "tx" refers to. It returns a `HeaderBlock`, but it's not even the "last" header block, it's the header block at the specified height.

perhaps I'm misunderstanding something, but this function should be called something like; `fetch_header_at_height()`.

I suppose `TX-block` refers to transaction block, and "last" refers to the most recent TX-block before or at `height`

chia/wallet/wallet_node.py
arvidn: I see. that's quite subtle.

arvidn (dismissed):

chia/wallet/wallet_node.py
```diff
@@ -1022,25 +1008,29 @@ def get_full_node_peers_in_order(self) -> List[WSChiaConnection]:
...
-                continue
+        last_tx_block = await fetch_last_tx_from_peer(height, peer)
+        if last_tx_block is None:
+            return None
 
-            assert last_tx_block.foliage_transaction_block is not None
-            self.get_cache_for_peer(peer).add_to_blocks(last_tx_block)
-            return last_tx_block.foliage_transaction_block.timestamp
+        assert last_tx_block.foliage_transaction_block is not None
+        self.get_cache_for_peer(peer).add_to_blocks(last_tx_block)
```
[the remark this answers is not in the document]
xdustinface: Good point! I integrated `fetch_last_tx_from_peer` into `get_timestamp_for_height_from_peer` to cache the blocks in 187f64e69579d2ecd968e1b695b41c241940a3ed and i also moved the timestamp cache check into the loop in 52403dfa83f9b6e9004db3655237a996d32ac1ef.

chia/wallet/wallet_node.py
```diff
@@ -1003,13 +1003,12 @@ async def get_timestamp_for_height_from_peer(self, height: uint32, peer: WSChiaC
...
-        cached_timestamp = cache.get_height_timestamp(height)
-        if cached_timestamp is not None:
-            return cached_timestamp
-
         last_tx_block = None
         request_height: int = height
         while request_height >= 0:
+            cached_timestamp = cache.get_height_timestamp(uint32(request_height))
+            if cached_timestamp is not None:
+                return cached_timestamp
```
arvidn, line 1011: I'm not sure this will have the desired effect. I believe this will only give you a timestamp if `request_height` refers to a tx-block, right?

Which means you'll still keep requesting blocks from the peer until you find the tx-block. You'll perform the same search next time, because the non-tx-blocks still won't have timestamps

chia/wallet/wallet_node.py
```diff
@@ -1008,12 +1007,26 @@ async def get_timestamp_for_height_from_peer(self, height: uint32, peer: WSChiaC
...
 
-        last_tx_block = await fetch_last_tx_from_peer(height, peer)
+        last_tx_block = None
+        request_height: int = height
+        while request_height >= 0:
+            response: Optional[List[HeaderBlock]] = await request_header_blocks(
+                peer, uint32(request_height), uint32(request_height)
+            )
+            if response is not None and len(response) > 0:
+                cache.add_to_blocks(response[0])
```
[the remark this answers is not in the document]
xdustinface: The method takes a range so it might return more than but we request start == end.

chia/wallet/wallet_node.py
xdustinface: Yeah right! I added 0aa3dea76a3a733f30d50eee500278b763d915d6 to make use of the cached blocks.

chia/wallet/wallet_node.py
```diff
@@ -1003,30 +1003,33 @@ async def get_timestamp_for_height_from_peer(self, height: uint32, peer: WSChiaC
...
-                    last_tx_block = response[0]
+            block = cache.get_block(uint32(request_height))
+            if block is None:
+                self.log.debug(f"get_timestamp_for_height_from_peer cache miss for height {request_height}")
+                response: Optional[List[HeaderBlock]] = await request_header_blocks(
+                    peer, uint32(request_height), uint32(request_height)
+                )
+                if response is not None and len(response) > 0:
+                    self.log.debug(f"get_timestamp_for_height_from_peer add to cache for height {request_height}")
+                    cache.add_to_blocks(response[0])
```
arvidn, line 1019: if the only thing we use this cache for is to get timestamps, we don't technically need to cache non-tx blocks. But it looks like it works

arvidn (approved):

PR #16263, opened 2023-09-06
Invoke python instead of python3 consistently in pre-commit flows

AmineKhaldi:
On Windows, `python.exe` is safely assumed to exist in venv's `Scripts` folder.


altendky (changes requested):
i don't activate and i think this should continue to handle that as needed.

altendky (approved):
This isn't a beautiful situation.  This is a small fix to accept reality.

emlowe (approved):

PR #16264, opened 2023-09-06
Updating changelog

wallentx:
### Purpose:
Changelog update


hoffmang9:
We should specify that it's only relevant to Linux

emlowe (approved):

PR #16269, opened 2023-09-06
reference SHA in coverage diff comment

altendky:
### Current Behavior:

Links are created referencing `main` which is unlikely to be correct.

### New Behavior:

References the commit SHA that was used in the build so the link will be to an accurate line and continue being accurate later such as if included in another comment or a DM.


arvidn (approved):

wallentx (approved):
Thanks for making this! It was on my todo list (@xdustinface pointed this out to me yesterday)

PR #16272, opened 2023-09-06
Handle upload request with only one file for plugin

emlowe:
Due to https://github.com/Chia-Network/chia-blockchain/pull/16135 there will be cases now where the full file is not present.

Support upload request to the 3s plugin where only one of the files to upload (full or diff) is required in the request.


emlowe:
Oh I see Florin made some changes in 16135 that make this unneeded

PR #16276, opened 2023-09-07
add `SizedBytes.random()`/`.secret()`, replace `token_bytes()` usage

altendky:
### Purpose:

There are two basic common patterns around generating 'random' 32 byte sequences.  At runtime, we often do something like `bytes32(token_bytes(32))`.  In tests we either do the same or something more like `bytes32(random.randrange(256) for _ in range(32))`.  There are a few issues with this.

- The `32` byte count parameter is often left off and the call is just `token_bytes()` which happens to presently default to 32 bytes though that is not guaranteed.  "If `nbytes` is `None` or not supplied, a reasonable default is used." (https://docs.python.org/3.11/library/secrets.html#secrets.token_bytes)
- Just calling `token_bytes(32)` is often insufficient for type hinting since we are passing the result to places that expect `bytes32` resulting in changes to `bytes32(token_bytes(32))` This ends up duplicating the desire for 32 bytes in a way that is entirely uncheckable with standard tooling.
- Using `token_bytes()` in the tests is slower than needed (though sure, probably not _that_ slow) since we don't need secure tokens and also is non-deterministic which is not a plus for tests.  Admittedly, using deterministic options will likely only make it somewhat more deterministic, not entirely so, but it is a step in the right direction.
- `random.randbytes()` is only available as of 3.9+ (https://docs.python.org/3.11/library/random.html#random.randbytes) while we still support 3.8.  This means that deterministic options are more involved such as `(random.randrange(256) for _ in range(32)`, noted above.

All this lead me to proposing offering `bytes32.random()` which accepts an optional random source such as a generator reseeded consistently for each test along with a `bytes32.secret()` for secure generation in runtime code where that is critical.  These are, of course, provided on all the fixed width bytes types, not just `bytes32`.

I do see the obfuscation aspect of this but given the list of issues above I think this is a reasonable option.

There do remain a few uses of `token_bytes()` which would not make sense to force to fixed width bytes types.

### Draft For:
- [x] https://github.com/Chia-Network/chia-blockchain/pull/16296


arvidn:
I think this is a reasonable approach. +1 for deterministic values. I'm not super excited about putting test code in `bytes32`, but I can live with that.

altendky:
I suspect we could apply it in runtime as well.  Not with a seeded `Random`, but a `random.SystemRandom`.   I think that is the same as `secrets.SystemRandom`?  Should double check.  Or perhaps a separate `.secret()` to use the secure random generator.  I hadn't thought through those details yet.

arvidn:
one distincion between `random` and `secrets` has to do when forking `pytest` instances. `random` will seed once and then clone the process, making all processes use the same seed. But which tests are run in which process and in which order is still not deterministic afaik. `secrets`, as I understand it, provides fresh entropy in every call.

altendky:
Yes, secrets is for runtime when we do cryptography.  The expectation is that each test will pass in it's own `random.Random` instance, presently created by a fixture all with the same seed.  If you modify the test, or if the test is not deterministic such as can happen inherently with threads and tasks, then yeah, still not deterministic results across those conditions.  If we wanted to  avoid the same seed everywhere, we could make the seed depend on the 'node name' (test name plus parametrizations, roughly).  Not sure that's valuable, but an option.

arvidn (dismissed):
lgtm!

arvidn (approved):

PR #16280, opened 2023-09-07
Add traceback to front-end error responses

Quexington:
We often get issue submissions where users capture some relevant log output from the front end error message, but not the log message from the backend containing the backend stack trace.  This change would make it much more likely that while debugging we get all of the relevant info in a single log entry.

### DIFF
--- a/chia/rpc/util.py
+++ b/chia/rpc/util.py
@@ -27,7 +27,7 @@ def wrap_http_handler(f) -> Callable:
             tb = traceback.format_exc()
             log.warning(f"Error while handling message: {tb}")
             if len(e.args) > 0:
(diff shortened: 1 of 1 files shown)


altendky (dismissed):
I was trying to do this and more in https://github.com/Chia-Network/chia-blockchain/pull/11489/files but failed to actually finish it.  This is a useful step in my opinion that covers at least some cases.

altendky (approved):
or do it in all the places...  which quickly leads back to my pr

wjblanke (commented):
aok thanks!

wjblanke (approved):
aok

PR #16285, opened 2023-09-08
add fee for cat creation

geoffwalmsley:
When creating new CATs there is no option to set a fee for the transaction. This PR adds the option to specify a fee for the transaction


Quexington (approved):

wjblanke (approved):
aok

PR #16288, opened 2023-09-08
Fix test_dusted_wallet flakiness

Quexington:
@arvidn reported this test as flaky.  I'm not sure what the exact issue is but from my experience this test wasn't very good at handling the standard race conditions we have in wallet tests.  This PR should fix that.


arvidn (approved):
nice!

PR #16290, opened 2023-09-08
Return TransactionRecord from create_message_spend

Quexington:
`create_message_spend` unlike most other transaction methods, only returns a spendbundle rather than an entire transaction.  For the VC Wallet this means it is unable to properly pend the DID while the transaction is pending.  This PR adds a TX record returned from that function.


emlowe (dismissed):

ytx1991 (approved):
synced offline, lgtm

emlowe (approved):

PR #16291, opened 2023-09-08
Enhancements to WalletTestFramework

Quexington:
This PR introduces extensions to the WalletTestFramework (introduced in #16113) to make it more useful for common scenarios seen in wallet tests.  The framework attempts to provide the following flow as a way of writing tests:

1) Do some wallet actions that generate and submit a transaction
2) Run `WalletTestFramework.process_pending_states` specifying all of the state changes you expect
3) Do any checks outside of the framework that everything happened correctly
4) Repeat

This removes a lot of the burden from the dev writing the test to know about common simulator methods and race conditions.  It also encourages checking all state completely, making it more difficult for a dev to forget to test certain edge cases.

To see an example test written with this framework, see an upstream PR: https://github.com/Chia-Network/chia-blockchain/blob/a32c21a57e4d72ca8b32cf704e2399c75faf8a48/tests/wallet/vc_wallet/test_vc_wallet.py#L122


matt-o-how (dismissed):
seems good to me

wjblanke (approved):
aok

PR #16295, opened 2023-09-08
Rename USDS --> USDSC

paninaro:
### Purpose:
Rename USDS to USDSC for CAT asset ID `6d95dae356e32a71db5ddcb42224754a02524c615c5fc35f568c2af04774e589`

Note that this will not rename existing USDS wallets that have been added to the wallet DB.


hoffmang9 (approved):
lgtm

arvidn:
is the asset ID for the new USDS not known yet? I suppose the new asset ID needs a different name either way, to not collide with USDS assets in wallets that haven't renamed them

arvidn (approved):

PR #16297, opened 2023-09-09
BlockTools difficulty fix

arvidn:
### Purpose:

in BlockTools (used in simulations), increase the simulated time-per-block as the iterations increases and lower it as iterations decreases. This will prevent the ever-increasing difficulty in simulations of long chains

### Current Behavior:

Generating the long test chains take a long time, and the block rate slows down over time as the difficulty increases indefinitely.

### New Behavior:

Generating the long test chains is fast and the block rate stays relatively stable, as the difficulty converge.

### DIFF
--- a/chia/simulator/block_tools.py
+++ b/chia/simulator/block_tools.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import copy
+import functools
 import logging
 import math
 import os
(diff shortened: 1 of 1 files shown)


wallentx (dismissed):

chia/simulator/block_tools.py
```diff
@@ -146,7 +147,7 @@
         "BLOCKS_CACHE_SIZE": 340 + 3 * 50,  # Coordinate with the above values
         "SUB_SLOT_TIME_TARGET": 600,  # The target number of seconds per slot, mainnet 600
         "SUB_SLOT_ITERS_STARTING": 2**10,  # Must be a multiple of 64
-        "NUMBER_ZERO_BITS_PLOT_FILTER": 1,  # H(plot signature of the challenge) must start with these many zeroes
+        "NUMBER_ZERO_BITS_PLOT_FILTER": 2,  # H(plot signature of the challenge) must start with these many zeroes
```
arvidn, line 150: without this change, too many plots pass the plot filter and CHIP-13 is likely to reject all test plots, preventing us from generating long test chains

PR #16303, opened 2023-09-11
max_coin_amount should default to None in walllet send command

geoffwalmsley:
Using CLI to send a transaction is failing because max_coin_amount is defaulting to 0. This means `select_coins` produces no spendable coins since it is looking for coins with value between min_amount=0, max_amount=0.

Update the wallet's `send_cmd` to use `Optional[str]` as max_coin_amount's type, so if it is unspecified then we default to None and TXConfig will use the calculated max amount correctly.


Quexington (approved):

aqk (approved):

wjblanke (approved):
aok

PR #16314, opened 2023-09-12
Don't autopush txs across RPC

Quexington:
This PR is a step in the direction of a more observer-style wallet.

Currently, most of our wallet functions just push directly to the network as a matter of course when creating a transaction.  While this is convenient (especially in tests) it will be impossible should we get to a wallet design that does not contain the private key as a key component to wallet operations.

For now, the wallet still signs the transaction and has it ready to be pushed to the network should the user specify that.  But all of the pushing logic has been migrated to the top-most layer available (usually the RPC) so that the user always has control over whether or not they would like to immediately push the transaction.

In addition, any transaction method in the RPC now returns a `"transactions"` key so that it is easy for a front-end client to have everything they need to push the transaction as if the wallet was doing it (using the `push_transactions` endpoint).


geoffwalmsley (dismissed):

geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16320, opened 2023-09-13
remove CHIP-13 and SOFT_FORK4

arvidn:
The existing implementation of CHIP-13 only looks at signage points in winning blocks, if we want a correct CHIP-13 implementation, we need to look at the complete sequence of signage points, including ones that have no winning proof of space

CHIP-13 that only looks at signage points that has made it into blocks is problematic, since withholding a block or not affect consensus for future blocks.

If we decide to land this, testnet10 will need to be considered. Soft-fork4 already activated in testnet10, but most space appear to be on a fork that has not activated chip-13, so we could probably just move back to that chain.

### DIFF
--- a/tests/core/full_node/test_full_node.py
+++ b/tests/core/full_node/test_full_node.py
@@ -947,8 +947,8 @@ class TestFullNodeProtocol:
                 if force_high_fee:
                     successful_bundle = spend_bundle
             else:
-                assert full_node_1.full_node.mempool_manager.mempool.at_full_capacity(5000000 * group_size)
-                assert full_node_1.full_node.mempool_manager.mempool.get_min_fee_rate(5000000 * group_size) > 0
+                assert full_node_1.full_node.mempool_manager.mempool.at_full_capacity(10500000 * group_size)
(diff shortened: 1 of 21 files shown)


fchirica (dismissed):

tests/core/full_node/test_full_node.py
```diff
@@ -947,8 +947,8 @@ async def test_new_transaction_and_mempool(self, wallet_nodes, self_hostname):
                 if force_high_fee:
                     successful_bundle = spend_bundle
             else:
-                assert full_node_1.full_node.mempool_manager.mempool.at_full_capacity(5000000 * group_size)
-                assert full_node_1.full_node.mempool_manager.mempool.get_min_fee_rate(5000000 * group_size) > 0
```
arvidn, line 951: this change may be a bit mysterious. This test did not used to run in `HARD_FORK_2_0` mode, but since I removed some restrictions (along with `SOFT_FORK4`) this started running (and failing) in hard fork mode.

I tracked it down to the cost of the transaction we're trying to add to the mempool cost more than `5000000 * group_size`. So, just because we can't fit that TX in the mempool doesn't mean it's at capacity for this size.

fchirica (approved):

wjblanke (approved):
aok

PR #16321, opened 2023-09-13
Ban relative timelocks from RPC

Quexington:
Relative timelocks across an entire spendbundle don't make much sense so we'll throw in a little hack here to ban them from being used in this way.


arvidn (approved):
it would be nice to have a test exercising this check as well

Quexington:
> it would be nice to have a test exercising this check as well

A test in the PR right before this will actually start checking this (this is why we won't get a coverage exception I don't think)

wjblanke (approved):
aok

PR #16323, opened 2023-09-13
ban peers for 10 minutes when violating consensus rules

arvidn:
### Purpose:

Avoid peers that may be on a separate fork stronger, than other protocol errors.

### Current Behavior:

A peer violating consensus rules is banned for 10 seconds.

### New Behavior:

A peer violating consensus rules is banned for 10 minutes.


wjblanke (approved):
aok

PR #16325, opened 2023-09-13
Remove tx_endpoint from select_coins

Quexington:
This decorator doesn't really make sense for this RPC, it only needs the `CoinSelectionArgs`.


geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16326, opened 2023-09-13
Better timestamps on simulated chains

arvidn:

### Purpose:

This updates the `BlockTools` logic to generate more authentic simulated chains, and also lets us generate the long chains faster.

The highlights are:

* The (simulated) timestamps on blocks are proportional to the number of iterations between the blocks
* The logic to support tracking of sub-second simulated time is simplified to just use `float`, instead of integer seconds + residual

### Current Behavior:

The timestamp between two transaction blocks is always 18 seconds, regardless of there being non-tx blocks in between
The timestamp between blocks is irrespective of the number of actual VDF iterations between the blocks. This effectively simulates a timelord that speeds up over time, making it increasingly expensive to generate the chain.

### New Behavior:

The timestamp between two transaction blocks is (by default) 18.75 seconds for each block (including non-TX blocks). This time delta is scaled by the number of iterations.
The simulated chain does not simulate a timelord that speeds up over time. The cost of generating the chain is linear to the number of blocks.

### Testing Notes:

I generated the chains, including the 10000 block one with the debug output to verify.

### DIFF
--- a/chia/simulator/block_tools.py
+++ b/chia/simulator/block_tools.py
@@ -793,6 +793,9 @@ class BlockTools:
                             assert full_block.foliage_transaction_block is not None
                         elif guarantee_transaction_block:
                             continue
+                        # print(f"{full_block.height}: difficulty {difficulty} "
+                        #     f"time: {new_timestamp - last_timestamp:0.2f} "
+                        #     f"tx: {block_record.is_transaction_block}")
                         last_timestamp = new_timestamp
                         block_list.append(full_block)
                         if full_block.transactions_generator is not None:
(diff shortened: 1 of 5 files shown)


chia/simulator/block_tools.py
```diff
@@ -757,7 +757,12 @@ def get_consecutive_blocks(
                         if not use_timestamp_residual:
                             self._block_time_residual = 0.0
 
-                        full_block, block_record, self._block_time_residual = get_full_block_and_block_record(
+                        (
+                            full_block,
+                            block_record,
+                            self._block_time_residual,
+                            new_timestamp,
+                        ) = get_full_block_and_block_record(
```
arvidn, line 760: every new block returns its timestamp. This way, we increment `last_timestamp` for *every* block. Previously we would set it to the most recent tx-block.

fchirica (dismissed):

fchirica (approved):

emlowe (approved):
Do we need new plots checked into the plot cache repo?

arvidn:
> Do we need new plots checked into the plot cache repo?

Yes, see: https://github.com/Chia-Network/test-cache/pull/7

PR #16327, opened 2023-09-13
Extend sign APIs for customized puzzle

ytx1991:
### Purpose:
In order to allow DApps use WalletConnect to sign customized puzzles, we need to extend existing sign APIs.

### Current Behavior:
Cannot sign customized message

### New Behavior:
Allow DApp sign any message by setting safe_mode = false

### Testing Notes:
unit test


Quexington (approved):
This seems fine to me when guarded behind a `safe mode` parameter

wjblanke (approved):
aok

greimela:
@ytx1991 Are you planning to add the safe_mode parameter to WalletConnect too, so we can call this from a dApp?

PR #16328, opened 2023-09-14
server: Add `WSChiaConnection.__hash__`

xdustinface:
### Purpose:

Makes it for example possible to use `WSChiaConnection` in a `set` or as key in a `dict` which is something i want to do in a upcoming PR.

### DIFF
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -169,6 +169,9 @@ class WSChiaConnection:
             session=session,
         )
 
+    def __hash__(self) -> int:
+        return hash(self.peer_node_id)
+
     def _get_extra_info(self, name: str) -> Optional[Any]:
         writer = self.ws._writer
         assert writer is not None, "websocket's ._writer is None, was .prepare() called?"
--- a/tests/core/server/test_server.py
+++ b/tests/core/server/test_server.py
@@ -200,3 +200,19 @@ async def test_call_api_of_specific_for_missing_peer(
     )
 
     assert message is None
+
+
+@pytest.mark.limit_consensus_modes(reason="unnecessary")
+@pytest.mark.asyncio
+async def test_connection_hash(
+    self_hostname: str,
+    one_wallet_and_one_simulator_services: SimulatorsAndWalletsServices,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    [full_node_service], [wallet_service], _ = one_wallet_and_one_simulator_services
+    wallet_node = wallet_service._node
+    await wallet_node.server.start_client(
+        PeerInfo(self_hostname, uint16(cast(FullNodeAPI, full_node_service._api).server._port)), None
+    )
+    connection = full_node_service._node.server.all_connections[wallet_node.server.node_id]
+    assert hash(connection) == hash(connection.peer_node_id)


arvidn (approved):

chia/server/ws_connection.py
```diff
@@ -169,6 +169,9 @@ def create(
             session=session,
         )
 
+    def __hash__(self) -> int:
+        return hash(self.peer_node_id)
```
altendky, line 173: Are we really sure this is always unique at all times?  We've definitely had issues with mixing up old and reconnected connections.

altendky (changes requested):

chia/server/ws_connection.py
altendky: I'm trying to remember the details around this and collisions and checking equality and...  I'll have to think this over a bit later.

altendky: https://docs.python.org/3.11/reference/datamodel.html#object.__hash__
> The only required property is that objects which compare equal have the same hash value

Also read the rest about mutable attributes etc.  I think we can do this if we really want to, but have to appropriately handle both `.__hash__()` and `.__eq__()` in compatible ways.  `dataclasses(unsafe_hash=)` also has commentary.

https://docs.python.org/3.11/library/dataclasses.html#dataclasses.dataclass

I'm guessing that we'll end up delegating to `object.__hash__()` and `object.__eq__()`, but think it over if we really want to make this mutable class hashable.

xdustinface: > Are we really sure this is always unique at all times? We've definitely had issues with mixing up old and reconnected connections.

Yeah they are unique as they are calculated from the client certs. Sure those can be reused on different clients to force a misuse and duplication but this shouldn't affect using it as identifier. We already use the peer id as identifier for connections spread all over the codebase so i don't see an issue with doing this. 


> I think we can do this if we really want to, but have to appropriately handle both .__hash__() and .__eq__() in compatible ways.  dataclasses(unsafe_hash=) also has commentary.

What if we don't handle `__eq__`? I don't care about which way we do it so just let me know what you prefer. 

> I'm guessing that we'll end up delegating to object.__hash__() and object.__eq__(), but think it over if we really want to make this mutable class hashable.

Why not? Whats the downside of doing it? I feel like its nice to be able to just put connections into sets or dicts instead of putting their ids in there and then wherever you want to use them fetch them again from the server based on the id.

altendky: > > Are we really sure this is always unique at all times? We've definitely had issues with mixing up old and reconnected connections.
> 
> Yeah they are unique as they are calculated from the client certs. Sure those can be reused on different clients to force a misuse and duplication but this shouldn't affect using it as identifier. We already use the peer id as identifier for connections spread all over the codebase so i don't see an issue with doing this.

What about when we initiate a new connection to a node we are already connected to?  We've already had bugs around this.  It is actually extremely common for this to happen on small testnets since you never have as many peers as you want even when connected to every peer on the network so you are always trying to connect to every single peer even though already connected to them.  Oh yeah, with IPv4 and IPv6...

> > I think we can do this if we really want to, but have to appropriately handle both .**hash**() and .**eq**() in compatible ways.  dataclasses(unsafe_hash=) also has commentary.
> 
> What if we don't handle `__eq__`? I don't care about which way we do it so just let me know what you prefer.

The `dict` and `set` etc assume that you follow the rules laid out for `.__hash__()` and `.__eq__()`.  I could look further into their exact failure mode, but I think we just want to avoid these basic things working incorrectly.  Due to hash collisions, equality checks are part of lookups.

> > I'm guessing that we'll end up delegating to object.**hash**() and object.**eq**(), but think it over if we really want to make this mutable class hashable.
> 
> Why not? Whats the downside of doing it? I feel like its nice to be able to just put connections into sets or dicts instead of putting their ids in there and then wherever you want to use them fetch them again from the server based on the id.

The downside is that we exercise the misassumptions and incorrect implementations already laid out here `:]` resulting in bugs lots of us wouldn't be aware are possible.  I'm not saying a hard no to this pr, but I am suggesting we are careful about both the intent and the implementation.

Note that the `object` methods are written around using identity (`is`) for equality checks so they kind of side step the issues.  Not good when you want to be able to check if two objects are, well, equal...  but useful when you want to consider every instance to be unique even with the same attributes.

PR #16333, opened 2023-09-14
fix recovery of invalid sub epoch summaries cache

arvidn:
in case the height-to-hash is fully valid and complete but the ses file is invalid we would previously skip recovery entirely.

The new logic runs a recovery of the top 400 blocks unconditionally. If we don't find the "reconciliation point" (where the cache starts to agree with the DB), then we'll keep recovery with 5000 blocks at a time (for throughput).

### Purpose:

Recover from a corrupt sub epoch summary file, on startup.

### Current Behavior:

If the `height-to-hash` file is complete and valid, we won't recover a corrupt ses cache file.

### New Behavior:

Always recover a corrupt ses cache file and height-to-hash file on startup.

### DIFF
--- a/chia/full_node/block_height_map.py
+++ b/chia/full_node/block_height_map.py
@@ -112,18 +112,18 @@ class BlockHeightMap:
 
         self.__first_dirty = height + 1
 
-        # if the peak hash is already in the height-to-hash map, we don't need
-        # to load anything more from the DB
         if self.get_hash(height) != peak:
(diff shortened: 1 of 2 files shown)


emlowe (dismissed):

arvidn:
I intend to improve the test to cover the missing line, and also rebase it on top of the release branch targeting 2.1.1

emlowe (approved):

PR #16335, opened 2023-09-14
refactor get block generator

almogdepaz:


chia/consensus/blockchain.py
```diff
@@ -905,6 +912,7 @@ async def get_block_generator(
             # (as long as we're in the main chain)
             result = await self.block_store.get_generators_at(block.transactions_generator_ref_list)
         else:
+            assert fork is not None
```
arvidn, line 915: the if-statement on line 905-908 could probably be simplified to:
```
if fork is None:
```

and maybe the current condition could be an `assert`

chia/consensus/blockchain.py
```diff
@@ -917,44 +925,28 @@ async def get_block_generator(
...
+                    ref_block = reorg_chain[ref_height]
+                    assert ref_block is not None
+                    if ref_block.transactions_generator is None:
+                        raise ValueError(Err.GENERATOR_REF_HAS_NO_GENERATOR)
+                    result.append(ref_block.transactions_generator)
+                else:
+                    header_hash = self.height_to_hash(ref_height)
+                    if header_hash is None:
+                        raise ValueError(Err.GENERATOR_REF_HAS_NO_GENERATOR)
+                    gen = await self.block_store.get_generator(header_hash)
```
arvidn, line 946:
```suggestion
                    [gen] = await self.block_store.get_generators_at([ref_height])
```

arvidn: It could be improved by pulling all generators in one database query, but it wouldn't be so simple to interleave with the generators that don't come from the database

tests/blockchain/blockchain_test_utils.py
```diff
@@ -59,7 +59,10 @@ async def _validate_and_add_block(
     else:
         # Do not change this, validate_signatures must be False
         pre_validation_results: List[PreValidationResult] = await blockchain.pre_validate_blocks_multiprocessing(
-            [block], {}, validate_signatures=False
+            [block],
+            {},
+            validate_signatures=False,
+            fork_height=uint32(0) if fork_point_with_peak is None else fork_point_with_peak,
```
arvidn, line 65: `fork_height` is `Optional`. it looks suspicious to convert a `None` into `uint32(0)` here. Why not just forward `fork_point_with_peak` as is?

chia/consensus/blockchain.py
almogdepaz: i almost changed it but realised that would change the logic a bit to decide if this is a reorg by the caller and not in get_block_generator

tests/core/full_node/stores/test_full_node_store.py
```diff
@@ -246,6 +246,7 @@ async def test_basic_store(
         # Test adding non genesis peak directly
```
arvidn, line 246: do you know what this comment means?

almogdepaz: no idea im assuming this was supposed to be a peak from another chain that is not a fork but we should fail in that case not just handle it...

chia/consensus/blockchain.py
```diff
@@ -940,10 +943,11 @@ async def get_block_generator(
...
                 else:
-                    [gen] = await self.block_store.get_generators_at([ref_height])
-                    if gen is None:
-                        raise ValueError(Err.GENERATOR_REF_HAS_NO_GENERATOR)
-                    result.append(gen)
+                    heights_from_db.append(ref_height)
+            refs_from_db = await self.block_store.get_generators_at(heights_from_db)
+            if refs_from_db is None or len(refs_from_db) != len(heights_from_db):
+                raise ValueError(Err.GENERATOR_REF_HAS_NO_GENERATOR)
+            result.extend(refs_from_db)
```
arvidn, line 950: I don't think this is correct. The block-references is a list, not a set. The order we pass those generators to the block generator has to match that order.

arvidn: I think it's fine to not implement this optimization. In practice, our blocks have 0 or 1 block reference anyway

tests/wallet/sync/test_wallet_sync.py
```diff
@@ -312,8 +312,9 @@ async def test_long_sync_wallet(self, two_wallet_nodes, default_1000_blocks, def
         for wallet_node, wallet_server in wallets:
             await time_out_assert(600, wallet_height_at_least, True, wallet_node, len(default_400_blocks) - 1)
 
+        reorg_blocks = bt.get_consecutive_blocks(num_blocks=400,block_list_input=default_400_blocks[:10])
```
arvidn, line 315: generating this many block is non-trival work. we have some test fixtures (that are pre-calculated) to form a reorg-chain that you might be able to use instead. It would speed up the test.

see `test_long_reorg_blocks`, which shared the first 320 blocks with `default_1500_blocks`

chia/consensus/blockchain.py
almogdepaz: are they just sorted by height ?

tests/wallet/sync/test_wallet_sync.py
almogdepaz: cool thanks, i thought i would need to write one

chia/consensus/blockchain.py
arvidn: I don't believe there are any constraints on them. you may have duplicates. It's just a list.

if the question is how *our* farmer will create blocks, we just have 0 or 1 block reference, but that doesn't define the protocol

chia/consensus/blockchain.py
```diff
@@ -917,44 +925,24 @@ async def get_block_generator(
...
-                assert prev_block is not None
-                assert prev_block_record is not None
-                fork = find_fork_point_in_chain(self, peak, prev_block_record)
-                curr_2: Optional[FullBlock] = prev_block
-                assert curr_2 is not None and isinstance(curr_2, FullBlock)
-                reorg_chain[curr_2.height] = curr_2
-                while curr_2.height > fork and curr_2.height > 0:
-                    curr_2 = await self.block_store.get_full_block(curr_2.prev_header_hash)
-                    assert curr_2 is not None
-                    reorg_chain[curr_2.height] = curr_2
```
arvidn, line 935: we still need, at least, the height-to-hash mapping that `reorg_chain` provides. I don't think you can remove this whole seciont

almogdepaz: ok i get what wrong here.

we traverse all the heights in the end anyway and fetch the missing blocks from db but we do that by height so might get the wrong one, so i just need the mapping to get the hashes this code still redundant, we dont need all the full blocks up to the fork we only need the referenced ones

arvidn (commented):
we use two different representations for fork-height in the code base. one is `Optional[uint32]`, where `None` either means the chains don't share any blocks, or it means we don't know what the fork height is. The other is `int`, where -1 means the chains don't share any blocks.

In the case of `Optional`, I think there's a risk of confusion what `None` really means. The case of `int` makes arithmetic simpler, because -1 naturally fits in to comparing heights.

I think it would be good to get some clarity on exactly what semantics the code ascribes to the different values of fork height, and maybe try to unify them.

chia/consensus/block_body_validation.py
```diff
@@ -42,7 +42,7 @@ async def validate_block_body(
     height: uint32,
     npc_result: Optional[NPCResult],
     fork_point_with_peak: int,
-    get_block_generator: Callable[[BlockInfo, uint32], Awaitable[Optional[BlockGenerator]]],
+    get_block_generator: Callable[[BlockInfo, int, Dict[uint32, FullBlock]], Awaitable[Optional[BlockGenerator]]],
```
arvidn, line 45: I think we should bundle the `fork_height` and this height->block map into a class. We'll need to extend this with passing in the unspent coin set of the fork as well.

Also, I would worry about storing all the `FullNode` objects in memory. We're not likely to actually need all of that, and it can take up a lot of space

chia/consensus/blockchain.py
arvidn: by `block_records`, you mean the `Blockchain` object, right? The `self`. It has access to all blocks, including orphaned blocks and other forks. If we're in a reorg, those blocks are in the `Blockchain` object as well.

I think the intention of that if-statement, `if self.contains.block(curr.prev_header_hash)` was to ask if we know of that block, not if we happen to have it in the cache right now. If we assume the intention was the latter, there's another block of code missing here to do the same thing but pull the block hashes from the database.

the `additional_blocks` is just the current *batch* of block's we're validating. It does not contain the whole fork. The fork is still only stored in the database at this point, that's where we need to get the generators from.

chia/consensus/block_body_validation.py
almogdepaz: what do you mean by "FullNode" objects?

chia/consensus/blockchain.py
almogdepaz: then i dont get the point of this check, im pretty sure the initial intention was to see if its in the chain

otherwise we are just checking the store to contain something that we expect to be there, in that case i would have expected an assert or some exception being thrown, what is the case then when the block not in blockchain.blockrecords ?

PR #16340, opened 2023-09-15
Add extra_conditions to special offer making

Quexington:
Missed one area in which extra_conditions could have been piped through but wasn't.  This resulted in NFT and DL offers telling you they would expire when, in fact, they wouldn't.  This PR fixes that.


paninaro (approved):
Tested and works for me.

PR #16346, opened 2023-09-15
Unify TX pushing

Quexington:
The main section of this change is here: https://github.com/Chia-Network/chia-blockchain/pull/16346/files#diff-a367f650aed3b60eba2c253ce15852630c1baf5e9891223d1557ac9cb3333d20L1995

Right now, there exists two different ways of pushing transactions and wallets seem to use either almost arbitrarily.  This change combines the `Wallet.push_transaction` and `WalletStateManager.add_pending_transaction` into a single method that includes the union of their two functionalities.

In addition, this method now takes a list of transactions by default and will consolidate the spend bundles of all of transactions onto the first transaction in the list to ensure that all transactions make it into the mempool at the same time.  Most wallets already do this manually and I have not refactored them not to do so (yet) but this gives the option for wallets going forward to not have to worry about that particular aspect.


geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16354, opened 2023-09-18
Use TXConfig instead of pending_transactions for coin exclusion in spend_clawback_coins

Quexington:
Previously, this logic depended on pushing clawback transactions to the wallet DB to be submitted so that on further transaction generations, the same coins would not be selected.  With the change to make all transaction pushing optional, this logic no longer functioned correctly.  This change leverages the `excluded_coin_ids` key in the `TXConfig` to exclude coins between transaction batches instead, resulting in a stateless generation of these transactions (fitting the new pattern).


geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16358, opened 2023-09-18
RPM: Change the --depends to a Recommends: tag on the specific compat package

cmmarslender:
Removes dependency on a specific lib/file, which is apparently not actually supported, and we just got lucky with it working before. Instead, adding a recommended dependency on the package that provides that library on newer OSes (older ones have it by default). When the package exists, the package is installed, if not (older systems) its just skipped.

Tested installation on:
* Rocky 8
* Rocky 9
* Fedora 37
* Fedora 38

Fixes the CLI rpm. The UI rpm requires a fix in the GUI repo, and then an updated pin in this repo (see https://github.com/Chia-Network/chia-blockchain-gui/pull/2124)


wjblanke (approved):
aok

PR #16360, opened 2023-09-18
Make the standard wallet return `List[TransactionRecord]`

Quexington:
This seems like a potentially silly change on first glance, but the idea is to better conform with our other `generate_signed_transaction` methods which return a `List[TransactionRecord]` like I have changed this to.  The API still needs some work to fully conform but given how many places this function is used it seems wise to move slowly in that direction.

(Needs a coverage exception for the change to the benchmark test)


geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16349, opened 2023-09-15
run `chia version` from packaging artifacts to make sure they pass the simple smoke test

altendky:
### Draft For:
- [x] catch up with changes in https://github.com/Chia-Network/chia-blockchain/pull/16357
- [x] test gui packages as well

PR #16374, opened 2023-09-19
run `chia version` from packaging artifacts

altendky:
### Purpose:

Replacing https://github.com/Chia-Network/chia-blockchain/pull/16349 to provide this directly to the release branch.

### DIFF
--- a/.github/workflows/build-linux-installer-deb.yml
+++ b/.github/workflows/build-linux-installer-deb.yml
@@ -39,7 +39,7 @@ permissions:
 
 jobs:
   build:
-    name: Linux ${{ matrix.os.arch }} DEB Installer
+    name: Build ${{ matrix.os.arch }}
     runs-on: ${{ matrix.os.runs-on }}
     container: chianetwork/ubuntu-20.04-builder:latest
     timeout-minutes: 120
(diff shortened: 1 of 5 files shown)


.github/workflows/build-linux-installer-deb.yml
```diff
@@ -295,3 +295,71 @@ jobs:
...
+
+  test:
+    name: Test ${{ matrix.distribution.name }} ${{ matrix.mode.name }} ${{ matrix.arch.name }}
+    runs-on: ${{ matrix.os.runs-on[matrix.arch.matrix] }}
+    needs:
+    - build
+    container: ${{ matrix.distribution.url }}
+    strategy:
+      fail-fast: false
+      max-parallel: 4
```
cmmarslender, line 307: Only question is if this is really necessary. If we have the spare runners, seems like we'd want to just run as many of these at once as possible to reduce the overall time this matrix takes to run. If we're low on runners, jobs are going to be waiting either way, so might as well still just get these in the queue?

cmmarslender, line 307: Only question is if this is really necessary. If we have the spare runners, seems like we'd want to just run as many of these at once as possible to reduce the overall time this matrix takes to run. If we're low on runners, jobs are going to be waiting either way, so might as well still just get these in the queue?

cmmarslender (approved):

PR #16381, opened 2023-09-20
correct backtick usage in coverage comment

altendky:
### Purpose:

unbreak the links since they should not include the backtick, only the link text.


wallentx (approved):
Thanks!

cmmarslender (approved):

PR #16384, opened 2023-09-20
Remove Announcement class

Quexington:
With the introduction of https://github.com/Chia-Network/chia-blockchain/pull/15883 we introduced some more ergonomic classes for conditions in general including announcements.  The current driver is a bit flawed in that there is no way to determine if an announcement is intended to be asserted/created, whether it's for a coin/puzzle, and you can't simply create one with only a message, you must have both components.  Due to these weaknesses, this PR aims to remove that class entirely from the codebase and replace it with the more ergonomic ones from the predecessor PR.


geoffwalmsley (approved):

wjblanke (approved):
aok thanks

PR #16389, opened 2023-09-21
Fix condition serialization in RPC client

Quexington:
This is a bug fix from some introduced code in #15911 which serialized the `extra_conditions` improperly in the wallet RPC client causing it to fail to parse in the API.  This has been broken since it's introduction and this PR fixes it.


ytx1991 (approved):
lgtm

wjblanke (approved):
aok

PR #16393, opened 2023-09-22
add `--diff` for isort

altendky:
### Purpose:

https://github.com/Chia-Network/chia-blockchain/actions/runs/6274090643/job/17038962616?pr=16387#step:9:136
```
isort....................................................................Failed
- hook id: isort
- duration: 1.55s
- files were modified by this hook

Fixing /home/runner/work/chia-blockchain/chia-blockchain/chia/cmds/farm_funcs.py
```

Let's make this output a little more useful so we can tell what it wanted changed.

### Draft For:
- [ ] but this needs to be ci only... maybe?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,7 +65,7 @@ repos:
     hooks:
     -   id: isort
         name: isort
-        entry: ./activated.py isort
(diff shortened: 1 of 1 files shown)


emlowe:
Guessing we don't need this anymore with ruff replacing isort

PR #16395, opened 2023-09-22
Remove `ignore_max_send_amount`

Quexington:
This is a very old parameter that we've been propogating through new version of generate signed transaction for (it seems to me) no reason.  The only instances in which this parameter is actually set to `True` we manually specify the set of coins anyway.  Perhaps the idea is to prevent choosing coins that won't fit in a block, but such an unlikely error will still get caught by the mempool and there's a million other similar errors that would behave that way as well.  I think more likely the intention was simply for testing in tests that have since been re-written or deleted.


geoffwalmsley (dismissed):

emlowe (approved):
Approved

PR #16397, opened 2023-09-22
update to anyio4. may help with timeouts as per anyio/issues/589

wjblanke:
### Purpose:

Update Anyio to 4.0.0 from 3.7.1

https://github.com/agronholm/anyio/issues/589

### Current Behavior:

May timeout even under successful runs

### New Behavior:

Should no longer timeout due to anyio issues

### Testing Notes:

Check CI runs if there are more reported failures


emlowe:
closing old PR

PR #16399, opened 2023-09-22
Fix add_vc_proofs over daemon/walletconnect

freddiecoleman:
### Purpose:

Bugfix of addVCProofs WalletConnect call. Currently it doesn't work because an error is thrown.

### Current Behavior:

- When calling addVCProofs using the HTTP RPC it parses the proofs into a Python dictionary which gets passed into a VCProofs class.
- When calling addVCProofs using WalletConnect it parses the proofs into a string which throws an error when passed into a VCProofs class.

### New Behavior:

- When calling addVCProofs using the HTTP RPC it parses the proofs into a Python dictionary which gets passed into a VCProofs class.
- When calling addVCProofs using WalletConnect it parses the proofs into a Python dictionary which gets passed into a VCProofs class.

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -3869,7 +3869,8 @@ class WalletRpcApi:
         """
         vc_wallet: VCWallet = await self.service.wallet_state_manager.get_or_create_vc_wallet()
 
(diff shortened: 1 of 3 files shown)


paninaro:
@freddiecoleman the fix should likely be applied in chia-blockchain-gui instead of here. I'll follow up on Keybase

paninaro:
Fixed in chia-blockhain-gui

PR #16417, opened 2023-09-25
fix typo in PendingTxCache

arvidn:
when evicting the last MempoolItem of an assert-height level. add test coverage, we `pop()` from the `SortedDict`. This fails unconditionally because we're not passing in the required `key` argument.

The intention was to call `popitem()`, which pops the last item.

Kyle discovered this while experimenting with new linting.


aqk (approved):

wjblanke (approved):
aok!

PR #16424, opened 2023-09-26
Update install-gui.sh to use check Node 18 and npm 9

paninaro:
### Purpose:
The GUI has been updated to use Node 18 and npm 9. This change updates the install-gui.sh to verify that Node 18 and npm 9 are installed.

### DIFF
--- a/install-gui.sh
+++ b/install-gui.sh
@@ -37,17 +37,17 @@ do_install_npm_locally(){
   NODEJS_VERSION="$(node -v | cut -d'.' -f 1 | sed -e 's/^v//')"
   NPM_VERSION="$(npm -v | cut -d'.' -f 1)"
 
-  if [ "$NODEJS_VERSION" -lt "16" ] || [ "$NPM_VERSION" -lt "7" ]; then
-    if [ "$NODEJS_VERSION" -lt "16" ]; then
-      echo "Current NodeJS version($(node -v)) is less than 16. GUI app requires NodeJS>=16."
+  if [ "$NODEJS_VERSION" -lt "18" ] || [ "$NPM_VERSION" -lt "9" ]; then
+    if [ "$NODEJS_VERSION" -lt "18" ]; then
+      echo "Current NodeJS version($(node -v)) is less than 18. GUI app requires NodeJS>=18."
     fi
-    if [ "$NPM_VERSION" -lt "7" ]; then
-      echo "Current npm version($(npm -v)) is less than 7. GUI app requires npm>=7."
+    if [ "$NPM_VERSION" -lt "9" ]; then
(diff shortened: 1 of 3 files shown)


hoffmang9 (dismissed):
Lgtm

install-gui.sh
```diff
@@ -67,17 +67,17 @@ do_install_npm_locally(){
     export N_PREFIX=${SCRIPT_DIR}/.n
     PATH="${N_PREFIX}/bin:$(npm bin):${PATH}"
```
wallentx, line 68: Is `npm bin` still valid in node18?

arvidn (dismissed):

install-gui.sh
wallentx: @paninaro I verified `npm bin` was removed in v9.0.0. It is suggested to use `npm exec foo` or `npx --yes foo`.

paninaro: @ChiaMineJP Can you take a look at the issue @wallentx mentioned? Feel free to make whatever changes you feel are necessary.

ChiaMineJP: I've replaced `$(npm bin)` with `$(npm prefix)/node_modules/.bin`.
Reference:
https://docs.npmjs.com/cli/v8/commands/npm-prefix
https://docs.npmjs.com/cli/v9/configuring-npm/folders?v=true

hoffmang9 (dismissed):
lgtm

install-gui.sh
wallentx: This actually may be proper, since the only way to reach this part of the code is if you have an `npm` version less than `v9.x.x`

wallentx (dismissed):

ChiaMineJP:
@hoffmang9 @wallentx 
Sorry everyone, I need to push one more commit to suppress this warning:
image

ChiaMineJP:
Also needed to update this lines of code:
image

ChiaMineJP:
Tested both `install-gui.sh` and `start-gui.sh` in Ubuntu 22.04 and confirmed it works.

hoffmang9 (dismissed):
lgtm

ChiaMineJP:
Test on very old NodeJS/npm failed with the newer format of `package-lock.json`. (Thank you William)
Execuse me, but one-last commit will be pushed.

wallentx (approved):
Tested on Arch:
- With a good version of node/npm installed ✅
- With an old version of node/npm installed ✅
- Without node/npm installed ✅

hoffmang9 (approved):

cmmarslender (approved):

wjblanke (approved):
aok

PR #16430, opened 2023-09-26
Optimize CRCAT trades

Quexington:
There was some low hanging optimizations that could be made to CRCATs and offers more generally.  This PR combines those changes which should result in a ~35% speed up to trade tests involving CRCATs.


chia/wallet/vc_wallet/cr_cat_drivers.py
```diff
@@ -49,6 +49,49 @@
...
+        (
+            Program.to(EXTIGENT_METADATA_LAYER_HASH)
+            .curry(
+                Program.to(EXTIGENT_METADATA_LAYER_HASH).get_tree_hash(),
+                Program.to(None),
+                GUARANTEED_NIL_TP,
+                GUARANTEED_NIL_TP.get_tree_hash(),
+                P2_ANNOUNCED_DELEGATED_PUZZLE,
+            )
+            .get_tree_hash_precalc(
```
arvidn, line 73: we have a function to compute the hash of a curried puzzle without actually produce the curried puzzle first, which is a lot faster. `curry_and_treehash()`.
I suppose this cost has been shifted to import-time, so it might not be as critical anymore. I would expect tests to benefit though, since they may import this module more times than the actually use this data structure.

chia/wallet/vc_wallet/cr_cat_drivers.py
```diff
@@ -58,55 +101,38 @@ def construct_cr_layer(
...
+        .get_tree_hash_precalc(
+            CREDENTIAL_RESTRICTION_HASH, CREDENTIAL_STRUCT_HASH, authorized_providers_hash, proofs_checker_hash
+        )
+    )
+    first_curry_hash_hash: bytes32 = Program.to(first_curry_hash).get_tree_hash()
+    final_hash: bytes32 = (
+        Program.to(first_curry_hash)
+        .curry(first_curry_hash_hash, inner_puzzle_hash)
+        .get_tree_hash_precalc(first_curry_hash, first_curry_hash_hash, inner_puzzle_hash)
+    )
```
arvidn, line 132: I think these computations would be smaller, simpler and much faster if you would use `curry_and_treehash()` instead. chia/wallet/util/curry_and_treehash.py

chia/wallet/vc_wallet/cr_cat_wallet.py
```diff
@@ -867,28 +869,41 @@ async def match_hinted_coin(self, coin: Coin, hint: bytes32) -> bool:
...
-                CAT_MOD,
+                Program.to(CAT_MOD_HASH),
                 self.info.limitations_program_hash,
-                construct_cr_layer(
-                    self.info.authorized_providers,
-                    self.info.proofs_checker.as_program(),
-                    hint,  # type: ignore
-                ),
-            ).get_tree_hash_precalc(hint)
+                hint_inner_hash,  # type: ignore
```
arvidn, line 883: is `bytes` or `bytes32` not an acceptable type for the hint?
It looks like this would be exactly the right type here, why does the type need to be ignored here?

chia/wallet/vc_wallet/cr_cat_wallet.py
```diff
@@ -867,28 +869,41 @@ async def match_hinted_coin(self, coin: Coin, hint: bytes32) -> bool:
...
                 self.info.limitations_program_hash,
-                construct_cr_layer(
-                    self.info.authorized_providers,
-                    self.info.proofs_checker.as_program(),
-                    construct_pending_approval_state(hint, uint64(coin.amount)),
-                ),
-            ).get_tree_hash()
+                pending_approval_inner_hash,  # type: ignore
+                mod_code_hash=CAT_MOD_HASH_HASH,
+            ).get_tree_hash_precalc(CAT_MOD_HASH, CAT_MOD_HASH_HASH, pending_approval_inner_hash)
```
arvidn, line 901: this looks like another example of where `curry_and_treehash()` would be a lot faster. You don't actually need to construct the puzzle if all you need is the tree hash

chia/wallet/vc_wallet/cr_cat_wallet.py
Quexington: We're calling the function that constructs the _puzzle_ which expects an inner _puzzle_ rather than puzzle hash. The `precalc` is making sure the hash comes out properly, but the arguments we're passing in are hashes rather than programs as it expects.

chia/wallet/vc_wallet/cr_cat_wallet.py
Quexington: These ones I would like to leave because they're using the existing `construct_cat_puzzle` driver.  I've passed in all hashes here so the program that's being hashed is just:
```
(a (q . CAT_MOD_HASH) (c CAT_MOD_HASH_HASH (c TAIL_HASH (c INNER_PUZZLE_HASH 1))))
```
and therefore should not be very slow

chia/wallet/vc_wallet/cr_cat_drivers.py
Quexington: changed

chia/wallet/vc_wallet/cr_cat_drivers.py
Quexington: changed

chia/wallet/util/compute_hints.py
```diff
@@ -17,21 +20,44 @@ class HintedCoin:
...
+            ConditionOpcode.AGG_SIG_PARENT,
+            ConditionOpcode.AGG_SIG_PUZZLE,
+            ConditionOpcode.AGG_SIG_AMOUNT,
+            ConditionOpcode.AGG_SIG_PUZZLE_AMOUNT,
+            ConditionOpcode.AGG_SIG_PARENT_AMOUNT,
+            ConditionOpcode.AGG_SIG_PARENT_PUZZLE,
+            ConditionOpcode.AGG_SIG_UNSAFE,
+            ConditionOpcode.AGG_SIG_ME,
+        ]:
+            cost += ConditionCost.AGG_SIG.value
```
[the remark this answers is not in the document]
arvidn: I see that this code seems copied from `compute_additions_with_cost()`, which also suffers from these same problems (no unit tests and no support for the hard fork). In the long run, I think it would be good if these rules (especially the cost of conditions) would be concentrated in one place, maybe even in `chia_rs`.
Perhaps `chia_rs` could provide a lower-level function that both of these can depend on.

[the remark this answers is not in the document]
arvidn: I experimented with this here: https://github.com/Chia-Network/chia-blockchain/pull/16661
There are some hurdles

chia/wallet/vc_wallet/cr_cat_wallet.py
```diff
@@ -871,28 +873,41 @@ async def match_hinted_coin(self, coin: Coin, hint: bytes32) -> bool:
...
+        proofs_checker_hash: bytes32 = self.info.proofs_checker.as_program().get_tree_hash()
+        hint_inner_hash: bytes32 = construct_cr_layer_hash(
+            authorized_providers_hash,
+            proofs_checker_hash,
+            hint,
+        )
+        if (
             construct_cat_puzzle(
-                CAT_MOD,
+                Program.to(CAT_MOD_HASH),
```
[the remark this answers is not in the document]
Quexington: The change further below changes the call to `.get_tree_hash_precalc` to escape this value from the hash calculation.  So instead of hashing `CAT_MOD` every time, we now tell the tree hash function that we've already done it and it can skip doing that hash when it sees it.

chia/wallet/util/compute_hints.py
```diff
@@ -17,21 +20,44 @@ class HintedCoin:
...
-                condition.at("rrr") != Program.to(None)  # There's more than two arguments
-                and condition.at("rrrf").atom is None  # The 3rd argument is a cons
-            ):
-                potential_hint: bytes = condition.at("rrrff").atom
-                if len(potential_hint) == 32:
-                    hint = bytes32(potential_hint)
-            hinted_coins[bytes32(coin.name())] = HintedCoin(coin, hint)
-
-    return hinted_coins
+        if cost > max_cost:  # pragma: no cover
```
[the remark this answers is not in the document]
Quexington: Gotcha, updated the unit test to cover it

chia/wallet/util/compute_hints.py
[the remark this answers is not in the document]
Quexington: There's a unit test here: https://github.com/Chia-Network/chia-blockchain/blob/main/tests/wallet/test_util.py#L17 which I've updated to cover some of your other comments.

This function is a _wallet_ utility and we really just calculate/restrict the cost out of an abundance of caution.  We maybe don't even need the condition costs but in the case of offers it's nice to know it's invalid before it hits the mempool.  I don't think we want it to behave as if it knows about the hard fork because that would defeat the purpose in  offers of validating that it's mempool ready (ish).  If you have proposed simplifications I'm happy to review them in other PRs, but I don't think it really belongs here in this one.

chia/wallet/util/compute_hints.py
```diff
@@ -17,21 +20,44 @@ class HintedCoin:
...
+        cost += ConditionCost.CREATE_COIN.value
+
+        coin: Coin = Coin(cs.coin.name(), bytes32(condition.at("rf").atom), uint64(condition.at("rrf").as_int()))
+        hint: Optional[bytes32] = None
+        if (
+            condition.at("rrr") != Program.to(None)  # There's more than two arguments
+            and condition.at("rrrf").atom is None  # The 3rd argument is a cons
+        ):
+            potential_hint: bytes = condition.at("rrrff").atom
+            if len(potential_hint) == 32:
```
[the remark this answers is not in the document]
Quexington: You're right - fixed - and updated with a test for this case

arvidn (approved):

PR #16434, opened 2023-09-26
Fix glitch NFT wallet test

ytx1991:
### Purpose:
test_nft_transfer_nft_with_did may fail because the spend bundle is not in the mempool before farming a new block.

### Current Behavior:
No check of the spend bundle

### New Behavior:

Add a check to ensure the spend bundle is in the mempool.


Quexington (changes requested):
You also want a `wait_for_wallet_synced` call both before and after farming a block

ytx1991:
I think the asserts around do the same thing.

Quexington (approved):

wjblanke (approved):
aok!

PR #16435, opened 2023-09-26
Add new RPC to get block spends with conditions

freddiecoleman:
### Purpose:

Support for conditions can change at different block heights. Calculating this is complex and easily forgotten so this RPC gets the spends with conditions that are valid for the height of the spend. This encapsulates the logic in the full node and the caller doesn't need to think about it.

### DIFF
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -32,7 +32,6 @@ from chia.types.coin_record import CoinRecord
 from chia.types.coin_spend import CoinSpend, SpendInfo
 from chia.types.generator_types import BlockGenerator
 from chia.types.spend_bundle_conditions import SpendBundleConditions
-from chia.util.condition_tools import parse_sexp_to_conditions
 from chia.util.errors import Err
 from chia.util.ints import uint16, uint32, uint64
 from chia.wallet.puzzles.load_clvm import load_serialized_clvm_maybe_recompile
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -32,6 +32,7 @@ from chia.types.coin_record import CoinRecord
 from chia.types.coin_spend import CoinSpend, CoinSpendWithConditions, SpendInfo
 from chia.types.generator_types import BlockGenerator
 from chia.types.spend_bundle_conditions import SpendBundleConditions
+from chia.util.condition_tools import conditions_for_solution
 from chia.util.errors import Err
 from chia.util.ints import uint16, uint32, uint64
 from chia.wallet.puzzles.load_clvm import load_serialized_clvm_maybe_recompile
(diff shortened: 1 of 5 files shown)


chia/full_node/mempool_check_conditions.py
```diff
@@ -156,6 +158,45 @@ def get_spends_for_block(generator: BlockGenerator) -> List[CoinSpend]:
...
+        )
+        spends.append({
+            "coin": {
+                "amount": int_from_bytes(amount.atom),
+                "parent_coin_info": parent.atom,
+                "puzzle_hash": puzzle_hash,
+            },
+            "puzzle": puzzle,
+            "solution": solution,
+            "conditions": Program.to(r),
```
arvidn, line 196: this call is potentially *very* expensive. Do we always trust callers of this function?
would it not be acceptable to return a condensed version of the conditions here? e.g. Something like what's returned by `run_block_generator2()`.

if you call this function on a malicious block, your node will run out of memory.

chia/full_node/mempool_check_conditions.py
```diff
@@ -43,22 +43,8 @@
...
-    height: uint32,
-    constants: ConsensusConstants,
-) -> NPCResult:
-    run_block = run_block_generator
-
-    flags = 0
-    if mempool_mode:
-        flags = flags | MEMPOOL_MODE
-
-    if height >= constants.SOFT_FORK2_HEIGHT:
```
arvidn, line 60: dropping this condition will make testnet10 fail to sync

chia/full_node/mempool_check_conditions.py
```diff
@@ -156,6 +158,45 @@ def get_spends_for_block(generator: BlockGenerator) -> List[CoinSpend]:
 
     return spends
 
+def get_spends_for_block_with_conditions(generator: BlockGenerator, height: int, constants: ConsensusConstants) -> List[Dict[str, Any]]:
```
arvidn, line 161: I'm not super excited about the type-unsafety bleed into this file. If at all possible, I think it would be preferable to return a `SpendBundleConditions` object. It won't be a verbatim list of all conditions though, ones that are "shadowed" by others will be dropped, and coin/puzzle announcements and assertions will too.

tests/core/test_full_node_rpc.py
```diff
@@ -229,6 +230,14 @@ async def test1(self, two_nodes_sim_and_wallets_services, self_hostname):
             assert len(block_spends) == 3
             assert sorted(block_spends, key=lambda x: str(x)) == sorted(coin_spends, key=lambda x: str(x))
 
+            block_spends_with_conditions = await client.get_block_spends_with_conditions(block.header_hash)
+
+            assert len(block_spends_with_conditions) == 3
+            assert sorted(block_spends_with_conditions, key=lambda x: str(x)) == sorted(coin_spends, key=lambda x: str(x))
+
+            for coin_spend in block_spends_with_conditions:
+                assert isinstance(coin_spend.conditions, Program) == True
```
arvidn, line 239: I think this test should validate all fields it expects to see in the response

arvidn (changes requested):

chia/full_node/mempool_check_conditions.py
freddiecoleman: The program is run with `MAX_BLOCK_COST_CLVM` and the code is coming from a full block in the block store, so hasn't it already been run at least once before anyway?

freddiecoleman: If there are duplicate conditions we may want to see/display that.

chia/full_node/mempool_check_conditions.py
```diff
@@ -188,19 +192,14 @@ def get_spends_for_block_with_conditions(generator: BlockGenerator, height: int,
...
-                "parent_coin_info": parent.atom,
-                "puzzle_hash": puzzle_hash,
-            },
-            "puzzle": puzzle,
-            "solution": solution,
-            "conditions": Program.to(r),
-        })
+        coin = Coin(parent.atom, puzzle_hash, int_from_bytes(amount.atom))
+        coin_spend = CoinSpend(coin, puzzle, solution)
+        conditions = Program.to(r)
```
arvidn, line 197: If you, instead of `Program` make this variable a `List[ConditionWithArg]`, you can use the existing function: `parse_sexp_to_conditions()`. It does some sanitization

arvidn (approved):

PR #16439, opened 2023-09-27
Add sign_message_by_secp

ytx1991:
### Purpose:
For testing the vault, we need a tool to sign messages with SECP keys.

### Current Behavior:
None

### New Behavior:

Added new API

### Testing Notes:

Unit Test

### DIFF
--- a/chia/rpc/wallet_rpc_api.py
+++ b/chia/rpc/wallet_rpc_api.py
@@ -168,7 +168,7 @@ class WalletRpcApi:
             "/send_notification": self.send_notification,
             "/sign_message_by_address": self.sign_message_by_address,
             "/sign_message_by_id": self.sign_message_by_id,
-            "/sign_message_by_secp": self.sign_message_by_secp,
+            "/beta_sign_message_by_secp": self.beta_sign_message_by_secp(),
             "/verify_signature": self.verify_signature,
             "/get_transaction_memo": self.get_transaction_memo,
             # CATs and trading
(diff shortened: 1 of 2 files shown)


Quexington (changes requested):
Can we make sure to add like a `test_` prefix here, I don't want people thinking it's some new supported behavior.  Otherwise this seems okay (not an expert on whether this is the way signer apps work, maybe @paninaro can weigh in too).

ytx1991:
test_ looks line a unit test method. I don't think users will use this API if we don't mention it in the blog. For others who check the code, they know what is this API does.

Quexington:
> test_ looks like a unit test method. I don't think users will use this API if we don't mention it in the blog. For others who check the code, they know what this API does.

I think that's what we want, we don't want users using this, at least not yet.  We can remove the test_ prefix when we want to support it but for now it's just a helpful utility.

paninaro:
> > test_ looks like a unit test method. I don't think users will use this API if we don't mention it in the blog. For others who check the code, they know what this API does.
> 
> I think that's what we want, we don't want users using this, at least not yet. We can remove the test_ prefix when we want to support it but for now it's just a helpful utility.

Another option could be to _require_ the request to contain a param like `"FOR_TESTING_PURPOSES_ONLY": true` and error out if not present (or set to false).

paninaro:
But I also agree that there's a risk that some community members may see the new RPC and think that it should be added to a library like https://github.com/Chia-Mine/chia-agent or https://github.com/dkackman/chia-dotnet, maybe using some automated tool.

PR #16441, opened 2023-09-27
encapsulate `SyncStore._backtrack_syncing` dict

altendky:


arvidn (approved):
it's not obvious why this counter can't just be a member of the peer. it looks like in all three places we call `is_backtrack_syncing()`, `decrement_backtrack_syncinng()` and `increment_backtrac_syncing()`, we have access to the peer object

altendky:
I chatted with Arvid and we'll leave consideration of that for later.

PR #16451, opened 2023-09-28
Get block spends

arvidn:
### Purpose:

The `get_block_spends` RPC will not currently work after the hard fork. This can be demonstrated on testnet10 where the hard fork has already activated.

This patch fixes `get_block_spends` and also improves the reliability of the `get_puzzle_and_solution` RPC, which currently does take the hard fork serialization format into account, but no other hard- or soft fork rule changes.

This change incorporates part of a patch from @freddiecoleman .

### Current Behavior:

`get_block_spends` fail after the hard fork.
`get_puzzle_and_solution` may fail if certain new soft-fork features are used on the block in question.

### New Behavior:

`get_block_spends` have the intended behavior after the hard fork.
`get_puzzle_and_solution` have the intended behavior regardless of which block it's called on.

### Testing Notes:

The test for `get_block_spends` was extended to be run for each soft-fork version and the hard fork.

### DIFF
--- a/chia/clvm/spend_sim.py
+++ b/chia/clvm/spend_sim.py
@@ -422,7 +422,7 @@ class SimClient:
         generator: BlockGenerator = filtered_generators[0].transactions_generator  # type: ignore[assignment]
         coin_record = await self.service.coin_store.get_coin_record(coin_id)
         assert coin_record is not None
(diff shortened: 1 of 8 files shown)


arvidn:
the missing test coverage is in a benchmark, which we don't instrument

cameroncooper (dismissed):

emlowe (approved):

PR #16453, opened 2023-09-28
remove test classes from `test/core/full_node/stores/`

altendky:


arvidn (approved):

arvidn:
it looks like the existing test has covers a few cases that aren't exercised. I'm removing the coverage diff label as these tests are unchanged by this patch

PR #16457, opened 2023-09-29
improvements to deep re-org logic

arvidn:

### Purpose:

The node is currently having a hard time with deep reorgs. Some parts of block validation require some previous blocks to be loaded in the block record cache in the blockchain object. When performing a reorg deeper than the cache size, those block records may not be in the cache and fail.

Furthermore, when validating blocks on the fork of the chain (before accepting it and updating our peak) we need to be able to pick up block records (and generators) from blocks that are not in the main chain. This requires looking them up from the database. One case of interest, that has previously not been covered by a test, is when a block reference points into the forked chain (i.e. a block not in the main chain).

The important change in this PR are the last 3 commits, (8, 9 and 10 in the list below).

The other commits are tests and minor optimizations.

### Scope

This PR introduces tests covering cases we've seen fail on testnet recently. Additionally it fixes the logic to work as intended. Now, the way it was intended to work was very inefficient, so there are still serious efficiency problems with deep reorgs. Since this patch is already pretty large, I think it makes sense to split up the work and address the performance issues in separate PRs.

### Test chains

Since this patch alters the test chains (specifically, `default_10000_blocks` and `test_long_reorg_blocks`) CI will fail some tests that use them. CI requires test chains to have been committed to the `test-cache` repository and been published in a release for it to be able to pick them up.

If the general approach in this PR is approved, I will push the updated test chains and update this PR to use them, to get CI green. Here's the PR for the new test chains: https://github.com/Chia-Network/test-cache/pull/8

### Summary of changes:

In order to test long reorgs with block references, `BlockTools` need to be updated to support creating chains with block references. To avoid this change clashing with `main` later, the first commit is pulled from `main`, which also updates `BlockTools`. For details on this change, see https://github.com/Chia-Network/chia-blockchain/pull/16326 .

commits:

1. back-port of `BlockTools` change from `main`. 

2. Introduces a feature `dummy_block_references` and updates the two test blockchains `default_10000_blocks` and `test_long_reorg_blocks` to use them. These two chains will form the foundation of the new tests exercising the long reorg logic (with block references). This commit also introduces a second `long_reorg` test chain with the same weight as the main chain it's forking from (as opposed to higher weigh).

3. Updates `test_long_reorg` in `test_blockchain.py` to make the reorg *longer* than the size of the block record cache. It also makes it use the new test chains which include block references. It also clears the block reference cache to more accurately match a real-life scenario.

4. Add a new test to `test_full_node.py` to do the same thing but at the `full_node` level, making for a more authentic test (and slower).

5. Add a new test to `test_full_node.py` which sets up two full nodes with different forks, connect them and wait for them to have synchronized. This test exposed an issue where we end up requesting a block the other peer does not have. This issue need more investigation. 

6. There was a test in `test_full_node.py` that used to be parametrized, and would only run its reorg test in one of the test runs. Since then, it appears changes to the parameters have caused the reorg test to never run. The next commit runs it unconditionally.

7. Some minor optimizations to avoid duplicate dictionary lookups and multiple computations of the header hash of blocks.

8. This makes the parts of block validation that currently *only* look in the cache, also able to fall back to pull blocks from the database. Using `async` functions instead of immediate ones.

9. This makes the `find_fork_poin_in_chain()` `async` and also able to find blocks in the database that aren't in the cache.

10. Update the version of `block-cache` blocks we pull on CI, to use the new test chains.

11. Optimize `find_fork_point_in_chain()` to mitigate some of the issue of CI taking too ling.

12. Before initiating a sync, we collect peaks from at least 3 peers, unless that takes more than 30 seconds, in which case we start syncing anyway. This commit makes this timeout configurable, and configures it to 1 second in the tests. This shaves off a 30 second wait in the tests introduced in (5). 

### Current Behavior:

In a long/deep reorg, the node may fail with `KeyError` while looking up block records that are not in the cache.

### New Behavior:

Long/deep reorgs work as expected.

### Testing Notes:

There are two new relevant tests added, one in `test_blockchain.py` and one in `test_full_node.py`. These fail without the fix.

### DIFF
--- a/chia/consensus/blockchain_interface.py
+++ b/chia/consensus/blockchain_interface.py
@@ -64,6 +64,9 @@ class BlockchainInterface:
         # ignoring hinting error until we handle our interfaces more formally
         return  # type: ignore[return-value]
 
+    async def prev_block_hash(self, header_hashes: List[bytes32]) -> List[bytes32]:
+        return  # type: ignore[return-value]
+
     async def get_header_blocks_in_range(
         self, start: int, stop: int, tx_filter: bool = True
     ) -> Dict[bytes32, HeaderBlock]:
--- a/chia/consensus/blockchain_interface.py
+++ b/chia/consensus/blockchain_interface.py
@@ -41,6 +41,9 @@ class BlockchainInterface:
         # ignoring hinting error until we handle our interfaces more formally
         return  # type: ignore[return-value]
 
+    async def contains_block_from_db(self, header_hash: bytes32) -> bool:
+        return  # type: ignore[return-value]
+
     def remove_block_record(self, header_hash: bytes32) -> None:
         pass
 
--- a/chia/consensus/multiprocess_validation.py
+++ b/chia/consensus/multiprocess_validation.py
@@ -197,9 +197,9 @@ async def pre_validate_blocks_multiprocessing(
     num_sub_slots_found = 0
     num_blocks_seen = 0
     if blocks[0].height > 0:
-        if not block_records.contains_block(blocks[0].prev_header_hash):
+        curr = await block_records.get_block_record_from_db(blocks[0].prev_header_hash)
+        if curr is None:
             return [PreValidationResult(uint16(Err.INVALID_PREV_BLOCK_HASH.value), None, None, False)]
-        curr = block_records.block_record(blocks[0].prev_header_hash)
         num_sub_slots_to_look_for = 3 if curr.overflow else 2
         while (
             curr.sub_epoch_summary_included is None
@@ -212,7 +212,8 @@ async def pre_validate_blocks_multiprocessing(
             recent_blocks[curr.header_hash] = curr
             if curr.is_transaction_block:
                 num_blocks_seen += 1
-            curr = block_records.block_record(curr.prev_hash)
+            curr = await block_records.get_block_record_from_db(curr.prev_hash)
+            assert curr is not None
         recent_blocks[curr.header_hash] = curr
     block_record_was_present = []
     for block in blocks:
@@ -221,10 +222,18 @@ async def pre_validate_blocks_multiprocessing(
     diff_ssis: List[Tuple[uint64, uint64]] = []
     for block in blocks:
         if block.height != 0:
-            assert block_records.contains_block(block.prev_header_hash)
             if prev_b is None:
-                prev_b = block_records.block_record(block.prev_header_hash)
+                prev_b = await block_records.get_block_record_from_db(block.prev_header_hash)
+            assert prev_b is not None
 
+        # get_next_sub_slot_iters_and_difficulty() requires prev_b.prev_hash to
+        # be in the block record cache.
+        if prev_b is not None and prev_b.height != 0 and not block_records.contains_block(prev_b.prev_hash):
+            return [PreValidationResult(uint16(Err.INVALID_PREV_BLOCK_HASH.value), None, None, False)]
+        if prev_b is not None and not block_records.contains_block(block.prev_header_hash):
+            # the call to block_to_block_record() requires the previous
+            # block is in the cache
+            block_records.add_block_record(prev_b)
         sub_slot_iters, difficulty = get_next_sub_slot_iters_and_difficulty(
             constants, len(block.finished_sub_slots) > 0, prev_b, block_records
         )
@@ -253,6 +262,10 @@ async def pre_validate_blocks_multiprocessing(
         )
 
         try:
+            if prev_b is not None and not block_records.contains_block(block.prev_header_hash):
+                # the call to block_to_block_record() requires the previous
+                # block is in the cache
+                block_records.add_block_record(prev_b)
             block_rec = block_to_block_record(
                 constants,
                 block_records,
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -546,7 +546,8 @@ class FullNode:
             )
             if first is None or not isinstance(first, full_node_protocol.RespondBlock):
                 self.sync_store.batch_syncing.remove(peer.peer_node_id)
-                raise ValueError(f"Error short batch syncing, could not fetch block at height {start_height}")
+                self.log.error(f"Error short batch syncing, could not fetch block at height {start_height}")
+                return False
             if not self.blockchain.contains_block(first.block.prev_header_hash):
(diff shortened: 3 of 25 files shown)


tests/blockchain/test_blockchain.py
```diff
@@ -3131,18 +3131,20 @@ async def test_basic_reorg(self, empty_blockchain, bt):
...
 
     @pytest.mark.asyncio
-    async def test_long_reorg(self, empty_blockchain, default_1500_blocks, test_long_reorg_blocks, bt):
+    async def test_long_reorg(self, empty_blockchain, default_10000_blocks, test_long_reorg_blocks, bt):
         # Reorg longer than a difficulty adjustment
         # Also tests higher weight chain but lower height
         b = empty_blockchain
-        num_blocks_chain_1 = 3 * bt.constants.EPOCH_BLOCKS + bt.constants.MAX_SUB_SLOT_BLOCKS + 10
-        num_blocks_chain_2_start = bt.constants.EPOCH_BLOCKS - 20
+        num_blocks_chain_1 = 2000
```
almogdepaz, line 3138: looks like this could change the test since the relation to the constants is not enforced anymore

chia/wallet/wallet_blockchain.py
```diff
@@ -193,20 +193,24 @@ def get_latest_timestamp(self) -> uint64:
...
-            return self.block_record(header_hash)
-        return None
+        return self._block_records.get(header_hash)
 
     def block_record(self, header_hash: bytes32) -> BlockRecord:
         return self._block_records[header_hash]
 
+    async def get_block_record_from_db(self, header_hash: bytes32) -> Optional[BlockRecord]:
+        return self._block_records.get(header_hash)
+
```
almogdepaz, line 213: not sure i get the naming, where is the db access ?

chia/full_node/full_node.py
```diff
@@ -1424,7 +1425,7 @@ async def peak_post_processing(
             # This is a reorg
             fork_hash: Optional[bytes32] = self.blockchain.height_to_hash(state_change_summary.fork_height)
             assert fork_hash is not None
-            fork_block = self.blockchain.block_record(fork_hash)
+            fork_block = await self.blockchain.get_block_record_from_db(fork_hash)
```
almogdepaz, line 1428: the record should be in the cache, im afraid this can hide/open some problems

chia/full_node/full_node.py
```diff
@@ -1253,7 +1253,8 @@ async def add_block_batch(
                 if error is not None:
                     self.log.error(f"Error: {error}, Invalid block from peer: {peer.get_peer_logging()} ")
                 return False, agg_state_change_summary, error
-            block_record = self.blockchain.block_record(block.header_hash)
+            block_record = await self.blockchain.get_block_record_from_db(block.header_hash)
```
almogdepaz, line 1256: the record should be in the cache, im afraid this can hide/open some problems

chia/consensus/multiprocess_validation.py
```diff
@@ -253,6 +257,10 @@ async def pre_validate_blocks_multiprocessing(
         )
 
         try:
+            if prev_b is not None and not block_records.contains_block(block.prev_header_hash):
+                # the call to block_to_block_record() requires the previous
+                # block is in the cache
+                block_records.add_block_record(prev_b)
             block_rec = block_to_block_record(
```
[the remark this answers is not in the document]
almogdepaz: also look inside get_next_sub_slot_iters_and_difficulty we use the cache there as well and we traverse to different points that i think this breaks

chia/consensus/multiprocess_validation.py
```diff
@@ -221,10 +222,13 @@ async def pre_validate_blocks_multiprocessing(
...
-            assert block_records.contains_block(block.prev_header_hash)
             if prev_b is None:
-                prev_b = block_records.block_record(block.prev_header_hash)
+                prev_b = await block_records.get_block_record_from_db(block.prev_header_hash)
 
+        # get_next_sub_slot_iters_and_difficulty() requires prev_b.prev_hash to
+        # be in the block record cache.
+        if prev_b is not None and prev_b.height != 0 and not block_records.contains_block(prev_b.prev_hash):
+            return [PreValidationResult(uint16(Err.INVALID_PREV_BLOCK_HASH.value), None, None, False)]
         sub_slot_iters, difficulty = get_next_sub_slot_iters_and_difficulty(
```
almogdepaz, line 232: we use the cache here as well and we traverse to different points that i think this breaks, for instance we need  the previous previous block and we need the last transaction block before this block's signage point

chia/consensus/blockchain_interface.py
```diff
@@ -41,6 +41,9 @@ def contains_block(self, header_hash: bytes32) -> bool:
         # ignoring hinting error until we handle our interfaces more formally
         return  # type: ignore[return-value]
 
+    async def contains_block_from_db(self, header_hash: bytes32) -> bool:
```
almogdepaz, line 44: need to update BlockCache as well so it will still implement the interface

chia/full_node/full_node.py
arvidn: Do you mean that we should *add* code to populate the cache before getting here, or do you mean that there is already code to do this? (the tests would fail without being able to fall back to the DB here, so if there is logic that's meant to populate the cache, I don't think it's working right).

I haven't found code that looks like it attempts to make sure this block is in the cache.

chia/wallet/wallet_blockchain.py
arvidn: I'll add a comment. This is implementing the `BlockchainInterface` which is kind-of like an abstract base class

tests/blockchain/test_blockchain.py
arvidn: My understanding is that the test wants to make sure there are at least 3 epochs before the fork. I don't think the exact block height matters so much other than that. Do you think there's something subtle being tested by this exact fork height?

chia/consensus/multiprocess_validation.py
arvidn: maybe it passes the tests by "accident". This is definitely the part of the patch that I'm the least confident in.
We have a test that ensures this function fail with this error under certain circumstances, but the new reorg tests also need to be able to validate blocks whose immediate history is not in the cache. It was tempting to make all of this logic fall back to the DB if a block isn't in the cache, but that breaks another, existing, test.

PR #16459, opened 2023-09-29
remove synchronized block height based clearing of `SyncStore.peak_to_peer` cache

altendky:
### Purpose:

As is, the `SyncStore.peak_to_peer` cache is cleared periodically in a synchronized fashion across the network.  There is also another clearing mechanism that limits the cache to active peers and the last 256 heights.  This leaves us spending approximately 25% of the time ramping up to the 256 block limit and 75% of the time limited by it.  I don't think the periodic clearing is doing anything useful anymore and that avoiding the sawtooth that is synchronized across the network would be just as well.

image

If we do want to limit this cache more we could consider a lower limit than 256 blocks or a limit of the total number of peer entries for all peaks or...


altendky:
Yes, we limit to the last 256 blocks whenever we add a new height.

https://github.com/Chia-Network/chia-blockchain/blob/abbf29b9745d970d9cd5dcb86712482a7b4b5c31/chia/full_node/sync_store.py#L72-L78

AmineKhaldi (approved):

emlowe (approved):

PR #16471, opened 2023-10-02
improve performance of find-fork-point

arvidn:
This PR sits on top of https://github.com/Chia-Network/chia-blockchain/pull/16457 which should land before this one.

### Purpose:

The current implementation of `find_fork_point_in_chain()` traverses two chains from their peaks backwards to find the most recent common block. As long as the block records it traverses are in the block record cache, that's not a problem. However, when it looks up block records from the database, it needs to parse and instantiate a `BlockRecord` object, just to get to the `prev_hash` field. The header hash of the previous block already has its own column in the database and can be looked up from there much more efficiently.

### Current Behavior:

When traversing blocks not in the cache, `BlockRecord` objects are parsed and instantiated.

### New Behavior:

When traversing blocks not in the cache, the header hash of the previous block is pulled directly from the database. This improves performance. 

### Testing Notes:

`find_fork_point_in_chain()` is exercised by our reorg tests.

### Benchmarks

| test | before | after | after/before |
| --- | --- | --- | --- |
| test_long_reorg | 663.8 s | 538.8 s | 81.2 % |

### profile

before:
baseline-profile

The time spend parsing `BlockRecord` is saved.

after:
optimized

### DIFF
--- a/chia/consensus/blockchain_interface.py
+++ b/chia/consensus/blockchain_interface.py
@@ -64,6 +64,9 @@ class BlockchainInterface:
         # ignoring hinting error until we handle our interfaces more formally
         return  # type: ignore[return-value]
 
+    async def prev_block_hash(self, header_hashes: List[bytes32]) -> List[bytes32]:
+        return  # type: ignore[return-value]
+
     async def get_header_blocks_in_range(
         self, start: int, stop: int, tx_filter: bool = True
     ) -> Dict[bytes32, HeaderBlock]:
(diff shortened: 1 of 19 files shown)


chia/consensus/blockchain_interface.py
```diff
@@ -64,6 +64,9 @@ async def get_block_records_in_range(self, start: int, stop: int) -> Dict[bytes3
         # ignoring hinting error until we handle our interfaces more formally
         return  # type: ignore[return-value]
 
+    async def prev_block_hash(self, header_hashes: List[bytes32]) -> List[bytes32]:
+        return  # type: ignore[return-value]
```
arvidn, line 68: this needs to be added to `BlockCache` as well

PR #16525, opened 2023-10-04
Use network overrides for default port for WalletPeers

emlowe:
I believe this fixes issues with wallet throwing `Network id not supported in NETWORK_ID_DEFAULT_PORTS` errors in ephemeral or alternative testnets

The full node code in `full_node.py` uses similar logic to find the default port from the network overrides. It seems the wallet should do the same thing. As long as the config.yaml is setup with a default port for the ephemeral testnet, this should prevent the wallet from throwing this error

### DIFF
--- a/chia/wallet/wallet_node.py
+++ b/chia/wallet/wallet_node.py
@@ -640,7 +640,7 @@ class WalletNode:
         network_name = self.config["selected_network"]
         try:
             default_port = self.config["network_overrides"]["config"][network_name]["default_full_node_port"]
-        except Exception:
+        except KeyError:
             self.log.info("Default port field not found in config.")
             default_port = None
         connect_to_unknown_peers = self.config.get("connect_to_unknown_peers", True)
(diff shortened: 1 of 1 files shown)


arvidn:
looks reasonable. @fchirica should probably confirm

chia/wallet/wallet_node.py
```diff
@@ -638,7 +638,11 @@ def set_server(self, server: ChiaServer) -> None:
     def initialize_wallet_peers(self) -> None:
         self.server.on_connect = self.on_connect
         network_name = self.config["selected_network"]
-
+        try:
+            default_port = self.config["network_overrides"]["config"][network_name]["default_full_node_port"]
+        except Exception:
```
altendky, line 643: `KeyError`?

emlowe: Probably, but I copied it from the full node - https://github.com/Chia-Network/chia-blockchain/blob/f8849e81824e88fd121996df8ac9a3ae89a39b2f/chia/full_node/full_node.py#L452C9-L452C9

emlowe: I changed it - don't really like blanket catchs

fchirica (approved):

emlowe:
I tested this with an ephemeral testnet and it definitely squelches the error. The wallet appears to work as expected and connects to the nodes at the proper port.

arvidn (approved):

emlowe:
Coverage diff exemption

PR #16549, opened 2023-10-07
shutdown on startup failure and log to the log if possible

altendky:


altendky:
see https://github.com/Chia-Network/chia-blockchain/pull/16593

PR #16552, opened 2023-10-07
separate installer build and publish to separate jobs

altendky:
### Draft For:
- [x] review timeouts on publish jobs
- [x] ~review `setjobenv` usage for what is needed in which jobs~
- [x] ~can we use a more, or entirely, common image for the publish portion?~
  - great idea!  for later...
- [x] can the test jobs depend directly on the build job instead of publish?

### DIFF
--- a/.github/workflows/build-linux-installer-deb.yml
+++ b/.github/workflows/build-linux-installer-deb.yml
@@ -239,7 +239,7 @@ jobs:
       uses: actions/download-artifact@v3
       with:
         name: chia-installers-linux-deb-${{ matrix.os.arch }}
-        path: packages
+        path: build_scripts/final_installer/
 
     - name: Set Env
       uses: Chia-Network/actions/setjobenv@main
--- a/.github/workflows/build-linux-installer-deb.yml
+++ b/.github/workflows/build-linux-installer-deb.yml
@@ -207,7 +207,7 @@ jobs:
     needs:
     - build
     container: chianetwork/ubuntu-20.04-builder:latest
-    timeout-minutes: 120
+    timeout-minutes: 5
     strategy:
       fail-fast: false
       max-parallel: 4
--- a/.github/workflows/build-linux-installer-deb.yml
+++ b/.github/workflows/build-linux-installer-deb.yml
@@ -346,7 +346,7 @@ jobs:
     name: Test ${{ matrix.distribution.name }} ${{ matrix.mode.name }} ${{ matrix.arch.name }}
     runs-on: ${{ matrix.os.runs-on[matrix.arch.matrix] }}
     needs:
-    - publish
+    - build
     container: ${{ matrix.distribution.url }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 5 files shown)


.github/workflows/build-linux-installer-deb.yml
```diff
@@ -212,6 +197,80 @@ jobs:
...
+  publish:
+    name: Publish ${{ matrix.os.arch }}
+    runs-on: ${{ matrix.os.runs-on }}
+    needs:
+    - build
+    container: chianetwork/ubuntu-20.04-builder:latest
+    timeout-minutes: 5
+    strategy:
+      fail-fast: false
+      max-parallel: 4
```
cmmarslender, line 213:
```suggestion
```
Assume this was a copy/paste, but generally we dont need to limit matrices in cases like this (Looking earlier in the file, I now see this was already there for the other matrix.. might make sense to just get rid of the max-parallel for all the installer workflows)

.github/workflows/build-linux-installer-rpm.yml
```diff
@@ -203,6 +190,74 @@ jobs:
...
+    name: Publish amd64 RPM
+    runs-on: ubuntu-latest
+    needs:
+    - build
+    container:
+      image: chianetwork/rocky8-builder:latest
+    timeout-minutes: 5
+    strategy:
+      fail-fast: false
+      max-parallel: 4
```
cmmarslender, line 207:
```suggestion
```
Same comment as before.. We should probably remove the other max-parallel in this file as well.

.github/workflows/build-macos-installers.yml
```diff
@@ -250,6 +237,88 @@ jobs:
...
+
+  publish:
+    name: Publish ${{ matrix.os.name }} DMG
+    runs-on: ${{ matrix.os.runs-on }}
+    needs:
+    - build
+    timeout-minutes: 5
+    strategy:
+      fail-fast: false
+      max-parallel: 4
```
cmmarslender, line 252:
```suggestion
```
Same comment as before.. We should probably remove the other max-parallel in this file as well.

.github/workflows/build-macos-installers.yml
```diff
@@ -250,6 +237,88 @@ jobs:
...
+        uses: actions/download-artifact@v3
+        with:
+          name: chia-installers-macos-dmg-${{ matrix.os.name }}
+          path: build_scripts/final_installer/
+
+      - name: Set Env
+        uses: Chia-Network/actions/setjobenv@main
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          MACOSX_DEPLOYMENT_TARGET: 11
```
cmmarslender, line 304: This is a weird thing (that I see also already existed) that.. I dont think does anything in the context of this action?

https://github.com/Chia-Network/actions/blob/main/setjobenv/action.yml

cmmarslender (changes requested):
Left a few requests for minor changes

.github/workflows/build-macos-installers.yml
altendky:
```suggestion
```

cmmarslender (approved):

PR #16556, opened 2023-10-09
`chia show keys --show-mnemonic-seed`: Also show farmer private key

xearl4:
Technically, the farmer process alone doesn't need the master secret key, but only the farmer secret key. This change therefore makes it easier to retrieve this derived secret key.

### DIFF
--- a/chia/cmds/keys_funcs.py
+++ b/chia/cmds/keys_funcs.py
@@ -174,6 +174,7 @@ def show_keys(
 
         if show_mnemonic:
             key["master_sk"] = bytes(sk).hex()
+            key["farmer_sk"] = bytes(master_sk_to_farmer_sk(sk)).hex()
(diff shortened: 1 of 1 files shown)


xchdata1:
Additional thoughts:
- We could show the pool secret key as well, if we think that helps OP plot farmers in some way.
- I'd like to harmonise naming between "master private key" and "wallet secret key" and use "secret key" everywhere. But I'm not sure what stability requirements you have on the CLI output.

paninaro (dismissed):

paninaro:
@xchdata1 Thanks for the change. Would you be able to update the PR to target the `release/2.1.2` branch?

xchdata1:
> @xchdata1 Thanks for the change. Would you be able to update the PR to target the `release/2.1.2` branch?

Sure, done!

paninaro (approved):

PR #16562, opened 2023-10-10
enable reorg tests in TestFullNodeBlockCompression

arvidn:
### Purpose:

It appears to have been disabled by mistake, as test parameters were altered. This restores some test coverage.

### Current Behavior:

Some reorg tests are not running

### New Behavior:

Enable those reorg tests

### DIFF
--- a/tests/core/full_node/test_full_node.py
+++ b/tests/core/full_node/test_full_node.py
@@ -124,12 +124,7 @@ class TestFullNodeBlockCompression:
         wallet = wallet_node_1.wallet_state_manager.main_wallet
 
         # Avoid retesting the slow reorg portion, not necessary more than once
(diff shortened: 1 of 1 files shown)


tests/core/full_node/test_full_node.py
```diff
@@ -124,12 +124,7 @@ async def test_block_compression(
         wallet = wallet_node_1.wallet_state_manager.main_wallet
 
         # Avoid retesting the slow reorg portion, not necessary more than once
-        test_reorgs = (
-            tx_size == 10000
```
arvidn, line 128: as you can see above, `tx_size` is never 10000

aqk (approved):

emlowe (approved):

PR #16583, opened 2023-10-11
drop ipython dev dependency

altendky:


arvidn:
what's the rationale for this change?
do we no longer need to "debug asyncio"?

altendky:
Does anyone use this to "debug asyncio"?

arvidn (approved):

PR #16587, opened 2023-10-11
Remove commented out prints in get_consecutive_blocks

AmineKhaldi:


arvidn:
I would like to keep these. they are very handy when working on `BlockTools`

PR #16604, opened 2023-10-12
tidy publish steps for consistency

altendky:
### Draft For:
- [ ] refer to gui not main
- [ ] name the 'regular' (not dev or latest) file type


altendky:
super old and it seems something else maybe kinda similar was done

PR #16618, opened 2023-10-13
fix issue with syncing testnet, introduced in: Get block spends (#16451)

arvidn:
### Purpose:

When soft fork 2 activated on mainnet, most of the pre-softfork special cases were removed from the code. However, testnet still violated some of the new soft-fork rules prior to activation, so some soft fork 2 special cases need to remain in order to support testnet.

Specifically, the `NO_RELATIVE_CONDITIONS_ON_EPHEMERAL` need to be disabled on lower block heights. On mainnet, this is always active now.

There was a bug in #16451 which enabled `NO_RELATIVE_CONDITIONS_ON_EPHEMERAL` unconditionally, which is fine on mainnet, but not on testnet. This restores the previous behavior.

### Current Behavior:

testnet10 can not sync from scratch

### New Behavior:

testnet10 *can* sync from scratch (although, I have not tested this)

### Testing Notes:

I'm hoping @cmmarslender can test the sync-from-scratch on testnet10.

### DIFF
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -44,10 +44,10 @@ log = logging.getLogger(__name__)
 
 
 def get_flags_for_height_and_constants(height: int, constants: ConsensusConstants) -> int:
-    flags = ENABLE_ASSERT_BEFORE | NO_RELATIVE_CONDITIONS_ON_EPHEMERAL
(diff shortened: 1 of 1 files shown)


arvidn:
The sync test appears to have succeeded

wjblanke (approved):
aok

PR #16623, opened 2023-10-13
Fixed an issue where plot filter reduction was not applied on testnet10

ChiaMineJP:
There are testnet10 config adjustment function for full_node as attached below.
Since farmer does not have such kind of testnet10 config adjustment, if testnet10 user does not update `config.yaml` to set `PLOT_FILTER_XXX_HEIGHT` to appropriate value, farmer misses a lot of sps for farming.

I expect that many testnet users don't update their `config.yaml` for plot filter reduction height.
In this case, while FullNode is aware of plot filter reduction on testnet10, Farmer isn't.

image

### DIFF
--- a/chia/consensus/default_constants.py
+++ b/chia/consensus/default_constants.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from typing import Any, Dict
+
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.ints import uint8, uint32, uint64, uint128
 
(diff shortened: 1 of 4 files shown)


chia/consensus/default_constants.py
```diff
@@ -69,3 +71,24 @@
     # June 2033
     PLOT_FILTER_32_HEIGHT=uint32(20643000),
 )
+
+
+def update_testnet_overrides(network_id: str, overrides: Dict[str, Any]) -> None:
+    if network_id != "testnet10":
+        return
+    # activate softforks immediately on testnet
+    # these numbers are supposed to match initial-config.yaml
```
altendky, line 80: This would be for a separate PR, but we could load the initial config file here and load out the values instead of duplicating them.  Feel free to close this conversation after reading it.

altendky (approved):
I feel like we're patching patched patches here in terms of the config...  But applying this to the farmer does make sense to me.

arvidn (approved):
at some point we should probably start a testnet11, and remove all of the run-time overrides

chia/consensus/default_constants.py
ChiaMineJP: I;ve read it. Thanks for the comment!

PR #16625, opened 2023-10-14
also do peak post processing for cancellation in `FullNode.short_sync_batch()`

altendky:
### Purpose:

Based on the comment `# Still do post processing after cancel (or exception)` the intent was to, well, do this on cancellation.  In Python 3.7 and before `asyncio.CancelledError` inherited from `Exception` but in 3.8 and later it inherits from `BaseException`.  To get the same behavior in both cases we need to explicitly call out `asyncio.CancelledError`.

Do note that at least the incoming new peak message path here does not get cancelled when the peer is disconnected.

### Draft for:
- [ ] Should we be doing this this way or just shielding from cancellation?
- [ ] Review all calls to post processing as some do this, some only handled exceptions, some only handle cancellation, and some don't do anything.


arvidn:
I agree that this does seem reasonable

PR #16632, opened 2023-10-16
add NFT, DID, DAO_CAT to wallets denominated in mojos

geoffwalmsley:
running `chia wallet coins list` does not work for NFT or DID wallets because the `get_mojo_per_unit` function does not handle those wallet types.

Adding them in means `coins list` will now return the coins in NFT, DID, DAO_CAT wallets.


Quexington (approved):

wjblanke (approved):
aok

PR #16633, opened 2023-10-16
move vdf validation to a free free function

arvidn:
### Purpose:

This loosens the coupling between the `VDFInfo` type and the chiavdf module (which is a native implementation). Specifically, it opens up the possibility for the `VDFInfo` type to be implemented in `chia_rs`.

That's currently not possible without also make `chia_rs` depend on `chiavdf`, which would be messy.

### Current Behavior:

`VDFInfo` depends on `chiavdf`

### New Behavior:

`VDFInfo` does not depend on `chiavdf`


arvidn:
as far as I can tell, the missing test coverage was here before this change

fchirica (approved):

emlowe (approved):
seems reasonable

PR #16637, opened 2023-10-16
Speed up test runs by leveraging in-memory databases within reason

AmineKhaldi:


chia/simulator/setup_services.py
```diff
@@ -119,17 +119,18 @@ async def setup_full_node(
...
-        # TODO: remove (maybe) when fixed https://github.com/python/cpython/issues/97641
-        gc.collect()
-        db_path.unlink()
-
-        if db_version > 1:
-            with sqlite3.connect(db_path) as connection:
-                connection.execute("CREATE TABLE database_version(version int)")
-                connection.execute("INSERT INTO database_version VALUES (?)", (db_version,))
-                connection.commit()
+    if reuse_db:
```
arvidn, line 122: does `reuse_db` mean that we want to be able to close the connection, and then open up a new connection?
if it just means we want multiple connections to it, you can do that with an in-memory DB as well, you just need to add `&cache=shared` to the URI

tests/util/db_connection.py
```diff
@@ -1,19 +1,32 @@
...
-        db_path = Path(directory).joinpath("db.sqlite")
-        _db_wrapper = await DBWrapper2.create(database=db_path, reader_count=4, db_version=db_version)
-        try:
-            yield _db_wrapper
-        finally:
-            await _db_wrapper.close()
+async def DBConnection(db_version: int, in_memory_db: bool = True) -> AsyncIterator[DBWrapper2]:
+    if in_memory_db:
+        directory = None
+        db_path: Union[str, Path] = gen_in_memory_db_uri("db_cnx_")
```
arvidn, line 22: if multiple tests open a DB connection to the same URI, won't they access the same underlying DB, and possibly interfere with each other?

tests/util/db_connection.py
```diff
@@ -1,19 +1,32 @@
...
+    else:
+        directory = tempfile.TemporaryDirectory()
```
arvidn, line 24: I think an important benefit of using the context manager here is that the temporary directory *will* be cleaned up. in your version, if `_db_wrapper.close()` throws, it won't be

tests/util/db_connection.py
```diff
@@ -1,19 +1,32 @@
...
 from pathlib import Path
-from typing import AsyncIterator
+from typing import AsyncIterator, Union
 
 from chia.util.db_wrapper import DBWrapper2
 
 
+def gen_in_memory_db_uri(prefix: str) -> str:
+    random_string = "".join(random.choice(string.ascii_letters) for _ in range(8))
+    return f"file:{prefix}{random_string}?mode=memory&cache=shared"
```
arvidn, line 15: I see you added `cache=shared` here, to make sure multiple DB connections will share the same underlying database. Is that the intention?

chia/simulator/setup_services.py
```diff
@@ -119,17 +119,18 @@ async def setup_full_node(
...
+        db_path: Union[str, Path] = local_bt.root_path / f"{db_name}"
+        uri = False
+    else:
+        db_path = f"file:db_{db_name}?mode=memory"
```
arvidn, line 126: is there a reason to not include `&cache=shared` here?

arvidn:
How much faster are the tests with this change?

tests/util/db_connection.py
```diff
@@ -1,19 +1,32 @@
...
 from contextlib import asynccontextmanager
 from pathlib import Path
-from typing import AsyncIterator
+from typing import AsyncIterator, Union
 
 from chia.util.db_wrapper import DBWrapper2
 
 
+def gen_in_memory_db_uri(prefix: str) -> str:
+    random_string = "".join(random.choice(string.ascii_letters) for _ in range(8))
```
arvidn, line 14: given that you use a random string here, is there really a need to pass in the prefix?
this won't show up on a filesystem anyway, I don't think a user or developer will really encounter this string anyway

AmineKhaldi:
> How much faster are the tests with this change?

I used `tests/wallet/test_wallet_trade_store.py::test_large_trade_record_query` locally as an example:

main: 1 passed in 141.55s
this: 1 passed in 124.67s

arvidn (approved):

trepca:
@arvidn @AmineKhaldi  This introduces a dependency in `setup_simulators_and_wallets` to `tests.util`. Problem is that `tests.util` is not included in distro package, meaning my tests that are using simulators are not working anymore.

AmineKhaldi:
> @arvidn @AmineKhaldi This introduces a dependency in `setup_simulators_and_wallets` to `tests.util`. Problem is that `tests.util` is not included in distro package, meaning my tests that are using simulators are not working anymore.

Yep, this PR got merged before adding this change to it, so I sent an addendum PR #16752 shortly after.
Good to see you seb!

Issue #16589, opened 2023-10-11
[Bug] Fail to resync the wallet


YeungTing:
### What happened?

Fail to resync the wallet.

if old_records is too long, the sql error

    async def _get_new_tx_records_from_old(self, old_records: List[TransactionRecordOld]) -> List[TransactionRecord]:
        async with self.db_wrapper.reader_no_transaction() as conn:
            cursor = await conn.execute(
                f"SELECT txid, valid_times from tx_times WHERE txid IN ({','.join('?' *  len(old_records))})",
                tuple(tx.name for tx in old_records),
            )
            valid_times: Dict[bytes32, ConditionValidTimes] = {
                bytes32(res[0]): ConditionValidTimes.from_bytes(res[1]) for res in await cursor.fetchall()
            }
            await cursor.close()
        return [
            TransactionRecord(
                valid_times=valid_times[record.name] if record.name in valid_times else ConditionValidTimes(),
                spend_bundle=record.spend_bundle,
                **{k: v for k, v in dataclasses.asdict(record).items() if k != "spend_bundle"},
            )
            for record in old_records
        ]

### Version

2.1.0

### What platform are you using?

Windows

### What ui mode are you using?

GUI

### Relevant log output

```shell
2023-10-11T21:58:19.839 wallet chia.wallet.wallet_node    : ERROR    Exception while perform_atomic_rollback: too many SQL variables Traceback (most recent call last):
  File "C:\Chia\chia-blockchain-2.1.0.Official\chia-blockchain\chia\wallet\wallet_node.py", line 708, in perform_atomic_rollback
    removed_wallet_ids = await self.wallet_state_manager.reorg_rollback(fork_height)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Chia\chia-blockchain-2.1.0.Official\chia-blockchain\chia\wallet\wallet_state_manager.py", line 2099, in reorg_rollback
    reorged: List[TransactionRecord] = await self.tx_store.get_transaction_above(height)
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Chia\chia-blockchain-2.1.0.Official\chia-blockchain\chia\wallet\wallet_transaction_store.py", line 402, in get_transaction_above
[...]

PR #16639, opened 2023-10-16
Chunk SQLite query for old TR/TX conversion

Quexington:
For users with large amounts of transactions or trades, querying the new valid_times table with `?` variables can exceed a local limit.  This PR fixes this issue.

https://github.com/Chia-Network/chia-blockchain/issues/16589


wjblanke (approved):
aok

PR #16644, opened 2023-10-17
Simplify test_mempool_performance.py

AmineKhaldi:
* Remove flake8 exclusions.
* Cleanup imports.
* Remove unused logger.
* Simplify helper functions.
* Remove unneeded test class.


arvidn (approved):

arvidn:
the missing lines are in benchmarks, so we wouldn't expect them to be instrumented by coverage. There's some issue with our coverage workflow where it fails to add the `coverage-diff` tag as well as posting the comment

PR #16649, opened 2023-10-18
Fix benchmark flakiness

fchirica:


arvidn:
https://github.com/Chia-Network/chia-blockchain/actions/runs/6563702905/job/17828538516?pr=16649#step:13:1314
```
=================================== FAILURES ===================================
____________ test_benchmark_batch_insert_speed[pre=1000,count=1000] ____________
tests/core/data_layer/test_data_store.py:1431: in test_benchmark_batch_insert_speed
    with benchmark_runner.assert_runtime(seconds=case.limit):
E   AssertionError: 26.133334149999996 seconds not less than 24 seconds ( 109 % )
=========================== short test summary info ============================
FAILED tests/core/data_layer/test_data_store.py::test_benchmark_batch_insert_speed[pre=1000,count=1000] - AssertionError: 26.133334149999996 seconds not less than 24 seconds ( 109 % )
```

.github/workflows/benchmarks.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     runs-on: benchmark
     container:
       image: ubuntu:22.04
-    timeout-minutes: 30
+    timeout-minutes: 60
```
arvidn, line 31: is this really necessary?
I thought the benchmarks typically only takes about 10 minutes

tests/core/full_node/test_performance.py
```diff
@@ -119,7 +119,7 @@ async def test_full_block_performance(
         pr = cProfile.Profile()
         pr.enable()
```
arvidn, line 120: I would think the profilers slow this down a bit. We should probably remove them.

.github/workflows/benchmarks.yml
altendky: The changes this week dropped 'normal' runtimes from 28 minutes to 9.

fchirica: ok, given this i'll reduce to 30 again.

altendky (approved):

emlowe (approved):

PR #16660, opened 2023-10-19
add junit output and summaries for benchmarks

altendky:
### New Behavior:

https://github.com/Chia-Network/chia-blockchain/actions/runs/6616802037?pr=16660#summary-17971786470
image


aqk:
`--benchmark-repeats` is a tiny bit confusing (eg. does it mean `num_benchmark_repeats`, or `do_benchmark_repeats`)
https://github.com/Chia-Network/chia-blockchain/pull/16660/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R378

aqk:
You could be a little less fancy here - keep the regex match, then just `str.replace()` `[` and `]`
https://github.com/Chia-Network/chia-blockchain/pull/16660/files#diff-4d8b94eb59960fd8e48cf3187b1c63427acef34c56b07ee835ec665288a66096R122

altendky:
> `--benchmark-repeats` is a tiny bit confusing (eg. does it mean `num_benchmark_repeats`, or `do_benchmark_repeats`) [#16660 (files)](https://github.com/Chia-Network/chia-blockchain/pull/16660/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R378)

Oy, I did _not_ add help text for the `--benchmark-repeats` option.  Would that be good?  Or do you want a different name.

altendky:
> You could be a little less fancy here - keep the regex match, then just `str.replace()` `[` and `]` [#16660 (files)](https://github.com/Chia-Network/chia-blockchain/pull/16660/files#diff-4d8b94eb59960fd8e48cf3187b1c63427acef34c56b07ee835ec665288a66096R122)

I don't quite follow, but also this code is...  not exactly clear about the purpose.  It is meant to take the parametrization of the form `[a-b-c]` for the three parameter ids `a`, `b`, and `c` and remove a parametrization id of the form `benchmark_repeat000` regardless of it being the only item, first item, middle item, or last item.  Basically to get back to the parametrization as it would be presented if there were no benchmark repeat fixture parametrization in use.  All without ending up with `--` or `[-` or `-]`.  But maybe that's what you're saying.  Let those happen and then just do three replaces to catch those cases?

I probably ought to just do this on the test run side of things and then record that as another property in the results.  I expect I would be able to do it there without any string manipulation.

aqk:
Ya, I was just saying you didn't have to do the match and sub at the same time, and that might help sidestep the issue you were commenting on.

aqk:
Adding a docstring `--benchmark-repeats` should be plenty!

emlowe (approved):

PR #16661, opened 2023-10-19
update compute_additions_with_cost() to use run_puzzle()

arvidn:
### Purpose:

This should make it more accurate with regards to the cost of conditions and possibly a little bit faster. It will now assume that the hard fork has activated, for purposes of validating the puzzle+solution and computing the cost of conditions

### Current Behavior:

new conditions like `SOFTFORK` and the unknown condition codes with cost (introduced with the hard fork) are not taken into account when computing the cost.

### New Behavior:

The new conditions are taken into account (unconditionally) when computing the cost, as if the hard fork has activated.

### DIFF
--- a/chia/types/coin_spend.py
+++ b/chia/types/coin_spend.py
@@ -3,16 +3,25 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Any, Dict, List, Tuple
 
-from clvm.casts import int_from_bytes
+from chia_rs import (
+    AGG_SIG_ARGS,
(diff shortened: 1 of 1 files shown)


arvidn:
it seems we have a lot of tests exercising this function with invalid puzzles, and still expect to get the created coins returned..

arvidn:
the main challenge with this change is that we run unmatched offers to find their output coins, even though they are invalid. This change will just fail the whole puzzle rather than making a best-effort guess at what the output coins of a matched offer would be

PR #16663, opened 2023-10-19
Swap the benchmark job to the same container we use for installers with some things preinstalled

cmmarslender:
All dependencies that were required for benchmark are now in the builder image, so this will help with run times a bit not having to install this stuff as part of each benchmark run.


altendky (approved):

wjblanke (approved):
aok

PR #16674, opened 2023-10-23
require SerializedProgram.from_bytes() is passed a valid clvm structure

arvidn:
### Purpose:

The *Streamable* protocol (normally) requires input passed to `parse()` and `from_bytes()` be equally valid input. The distinction is that `parse()` consumes input from a stream, whereas `from_bytes()` expects the complete byte buffer to represent the complete value to be consumed for the streamable type.

`SerializedProgram` is currently an exception to this, where there is no validation at all of the input passed to `from_bytes()`, whereas `parse()` *does* validate the input. In most normal production use, `parse()` is how `SerializedProgram`s are initialized (such as parsing a `FullBlock` from the network or database).

There are however a few tests where we initialize `SerializedProgram` with invalid input. This is a small tweak to correct this quirk, and fix the tests.

### Current Behavior:

`SerializedProgram.from_bytes(b"aaa")` is valid.

### New Behavior:

`SerializedProgram.from_bytes(b"aaa")` is *not* valid.

### DIFF
--- a/tests/core/full_node/stores/test_block_store.py
+++ b/tests/core/full_node/stores/test_block_store.py
@@ -300,7 +300,7 @@ async def test_get_generator(bt: BlockTools, db_version: int, use_cache: bool) -
     blocks = bt.get_consecutive_blocks(10)
 
     def generator(i: int) -> SerializedProgram:
-        return SerializedProgram.from_bytes(int_to_bytes(i))
+        return SerializedProgram.from_bytes(int_to_bytes(i + 1))
 
     async with DBConnection(db_version) as db_wrapper:
         store = await BlockStore.create(db_wrapper, use_cache=use_cache)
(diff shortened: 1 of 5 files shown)


tests/core/full_node/stores/test_block_store.py
```diff
@@ -300,7 +300,7 @@ async def test_get_generator(bt: BlockTools, db_version: int, use_cache: bool) -
     blocks = bt.get_consecutive_blocks(10)
 
     def generator(i: int) -> SerializedProgram:
-        return SerializedProgram.from_bytes(int_to_bytes(i))
+        return SerializedProgram.from_bytes(int_to_bytes(i + 1))
```
arvidn, line 303: `int_to_bytes(0)` is an empty buffer, which is not a valid CLVM program. adding `1` avoids this issue.

tests/generator/test_compression.py
```diff
@@ -51,11 +51,17 @@
     "ff01ffffffa00000000000000000000000000000000000000000000000000000000000000000ff830186a080ffffff02ffff01ff02ffff01ff02ffff03ff0bffff01ff02ffff03ffff09ff05ffff1dff0bffff1effff0bff0bffff02ff06ffff04ff02ffff04ff17ff8080808080808080ffff01ff02ff17ff2f80ffff01ff088080ff0180ffff01ff04ffff04ff04ffff04ff05ffff04ffff02ff06ffff04ff02ffff04ff17ff80808080ff80808080ffff02ff17ff2f808080ff0180ffff04ffff01ff32ff02ffff03ffff07ff0580ffff01ff0bffff0102ffff02ff06ffff04ff02ffff04ff09ff80808080ffff02ff06ffff04ff02ffff04ff0dff8080808080ffff01ff0bffff0101ff058080ff0180ff018080ffff04ffff01b081963921826355dcb6c355ccf9c2637c18adf7d38ee44d803ea9ca41587e48c913d8d46896eb830aeadfc13144a8eac3ff018080ffff80ffff01ffff33ffa06b7a83babea1eec790c947db4464ab657dbe9b887fe9acc247062847b8c2a8a9ff830186a08080ff8080808080"
 )  # noqa
 
-gen1 = b"aaaaaaaaaa" + original_generator
-gen2 = b"bb" + original_generator
+gen1 = b"\xff\x01" + original_generator
+gen2 = b"\xff\x01\xff\x01" + original_generator
```
arvidn, line 55: these tests just need some bytes to offset the original generator. This changes the invalid prefix to valid ones (still not a valid *program*, just a valid CLVM structure).

altendky (approved):

emlowe (approved):

PR #16680, opened 2023-10-24
return List[TransactionRecord] from nft bulk mint functions

geoffwalmsley:
Bringing the `mint_from_did` and `mint_from_xch` functions in line with the standard practice of returning TransactionRecords instead of SpendBundles

### DIFF
--- a/chia/wallet/nft_wallet/nft_wallet.py
+++ b/chia/wallet/nft_wallet/nft_wallet.py
@@ -1259,7 +1259,7 @@ class NFTWallet:
         did_lineage_parent: Optional[bytes32] = None,
         fee: Optional[uint64] = uint64(0),
         extra_conditions: Tuple[Condition, ...] = tuple(),
-    ) -> SpendBundle:
+    ) -> List[TransactionRecord]:
         """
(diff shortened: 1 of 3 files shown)


chia/wallet/nft_wallet/nft_wallet.py
```diff
@@ -1731,7 +1733,8 @@ async def mint_from_xch(
 
         # Aggregate everything into a single spend bundle
         total_spend = SpendBundle.aggregate([signed_spend_bundle, xch_spend, *eve_spends])
-        return total_spend
+        tx_record: TransactionRecord = dataclasses.replace(eve_txs[0], spend_bundle=total_spend)
```
Quexington, line 1736: I haven't looked at the rest of this function, but are the rest of `eve_txs` worth returning too?  It's fine if they don't have a spend bundle attached, just everything that a wallet should show to a user is important.

geoffwalmsley: eve_txs comes from calling `generate_signed_transaction` so it's just one tx (in a list)

Quexington (approved):

emlowe (approved):

PR #16693, opened 2023-10-25
Annotate test_mempool_performance.py

AmineKhaldi:


arvidn (approved):

arvidn:
the lines lacking coverage are in benchmarks, which we don't instrument

PR #16698, opened 2023-10-26
Sync no farmer

arvidn:
### Purpose:

As part of the patch to fix deep-reorgs, it was discovered that one of the tests would not trigger a reorg sync unless a new block was farmed. That's not the intention, so I added this test to ensure a normal, catch-up, sync can be triggered without a new farm being farmed.

In order to make the tests not take too long, I cherry-picked one of the commits from the deep-reorg PR to make the delay configurable, of how long to wait to receive 3 peers with the same peak. This is set to 1 seconds in the tests now.

There is no change in behavior, except the delay is configurable, still defaults to 30 seconds.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -897,8 +897,9 @@ class FullNode:
             self.log.info("Waiting to receive peaks from peers.")
 
             # Wait until we have 3 peaks or up to a max of 30 seconds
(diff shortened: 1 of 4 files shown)


AmineKhaldi (approved):

wjblanke (approved):
aok

PR #16701, opened 2023-10-26
transition away from `__bytes__` conversion for fixed-size integers

arvidn:
### Purpose:

Transition away from `__bytes__` conversion for fixed-size integers, use an explicit conversion function instead (`stream_to_bytes()`). The intention is to avoid accidental mixups with constructing bytes of a specified size (which is what happens with `int`).

This is something that has bothered me for a while; `bytes(value)` has vastly different behavior based on whether `value` is a plain `int` or a fixed-size integer, e.g. `uint32`. The former will allocate a buffer of the specified size, the latter will serialize the integer into a byte representation.

Ensuring that the expression `bytes(value)` does the right thing hinges entirely on mypy checks, and we don't have 100% coverage of those checks. If anyone, accidentally, passes an `int` instead of `uint32` to a function that then attempts to serialize it, may even trigger OOM.

My first attempt was to simply call the explicit streaming function `to_bytes()` (as a counterpart to `from_bytes()`), but that clashes with the built-in member function on `int` with the same name. The fixed-sized integers derive from `int`. The name `stream_to_bytes()` is unique and less likely to clash with any unintended functions.

### Current Behavior:

```
bytes(value)
```

### New Behavior:

```
value.stream_to_bytes()
```


Quexington:
Would it make sense here you think to just straight up ban `__bytes__` on these classes as well?  I can see some potential whiplash effects here for a developer who thinks they can call bytes on these and doesn't know about this change and ends up with an empty buffer instead.  If you want that functionality you can always cast to a plain int.

chia/util/struct_stream.py
```diff
@@ -78,13 +78,19 @@ def parse(cls: Type[_T_StructStream], f: BinaryIO) -> _T_StructStream:
...
 
-    def __bytes__(self) -> bytes:
+    def stream_to_bytes(self) -> bytes:
         return super().to_bytes(length=self.SIZE, byteorder="big", signed=self.SIGNED)
+
+    # this is meant to avoid mixing up construcing a bytes object of a specific
+    # size (i.e. bytes(int)) vs. serializing the integer to bytes (i.e. bytes(uint32))
+    def __bytes__(self) -> bytes:
+        assert False
+        return b""
```
altendky, line 96: https://docs.python.org/3.11/library/exceptions.html#NotImplementedError
> It should not be used to indicate that an operator or method is not meant to be supported at all – in that case either leave the operator / method undefined or, if a subclass, set it to [None](https://docs.python.org/3.11/library/constants.html#None).

So maybe we are supposed to do this?  I'm not really sure, haven't gone down this path much before.

```suggestion
    __bytes__ = None
```

arvidn: I'm a bit ignorant of when definitions end up in the class object and the instance object. I think methods always end up in the class object, but does `__bytes__ = None` also end up in the class object? It seems like it could also become a member of the instance object.

altendky: It would be a class attribute and would also be accessible via instances since they fall back on the class if there's no instance attribute.

Given their recommendation, I was expecting `bytes()` do given a sensible error rather than just a generic type error.

```pycon
>>> class C(int):
...   __bytes__ = None
... 
>>> bytes(C())
TypeError: 'NoneType' object is not callable
```

I think something other than assert would be good though, and to provide a description and the intended replacement.

altendky: https://mypy-play.net/?mypy=latest&python=3.11&gist=fff4a0a0cdbe9b7fdc4e00174443a5c5
```python
class C(int):
    __bytes__ = None

bytes(C())
```

This gives no error, so that isn't great.  Would be nice to find a way to catch this with mypy instead of at runtime.

Quexington (approved):

emlowe (approved):

PR #16703, opened 2023-10-26
Swap some info logs in seeder to warning

cmmarslender:
INFO logs when running a crawler are very noisy, so this log level makes it easier to see important information a crawler/seeder operator needs to see


hoffmang9 (approved):
The ongoing bouncing log tags...

arvidn (approved):

wjblanke (approved):
aok

PR #16704, opened 2023-10-27
Update to support looking up mnemonic by just the first 4 letters of each word

cmmarslender:
Since the first 4 letters of each word is unique (or the full word, if less than 4 characters), and its common practice to only store the first 4 letters of each word in many offline storage solutions, this adds support to look up words by the first 4 characters


justinengland (dismissed):

chia/util/keychain.py
```diff
@@ -87,11 +87,32 @@ def bytes_to_mnemonic(mnemonic_bytes: bytes) -> str:
...
+    up words by the first 4 characters
+    """
     mnemonic: List[str] = mnemonic_str.split(" ")
     if len(mnemonic) not in [12, 15, 18, 21, 24]:
         raise ValueError("Invalid mnemonic length")
 
+    four_char_dict = {word[: min(4, len(word))]: word for word in bip39_word_list().splitlines()}
+    full_words: List[str] = []
+    for i in range(0, len(mnemonic)):
+        word = mnemonic[i]
```
altendky, line 103: I think?

```suggestion
    for word in mnemonic:
```

chia/util/keychain.py
```diff
@@ -87,11 +87,32 @@ def bytes_to_mnemonic(mnemonic_bytes: bytes) -> str:
...
+    """
+    Since the first 4 letters of each word is unique (or the full word, if less than 4 characters), and its common
+    practice to only store the first 4 letters of each word in many offline storage solutions, also support looking
+    up words by the first 4 characters
+    """
     mnemonic: List[str] = mnemonic_str.split(" ")
     if len(mnemonic) not in [12, 15, 18, 21, 24]:
         raise ValueError("Invalid mnemonic length")
 
+    four_char_dict = {word[: min(4, len(word))]: word for word in bip39_word_list().splitlines()}
```
altendky, line 100: Slicing already does this for you.  I forget, black may reformat this.

```suggestion
    four_char_dict = {word[:4]: word for word in bip39_word_list().splitlines()}
```

chia/util/keychain.py
```diff
@@ -87,11 +87,32 @@ def bytes_to_mnemonic(mnemonic_bytes: bytes) -> str:
...
+    full_words: List[str] = []
+    for i in range(0, len(mnemonic)):
+        word = mnemonic[i]
+        full_word = four_char_dict.get(word[: min(4, len(word))])
```
altendky, line 104:
```suggestion
        full_word = four_char_dict.get(word[:4])
```

chia/util/keychain.py
```diff
@@ -87,11 +87,32 @@ def bytes_to_mnemonic(mnemonic_bytes: bytes) -> str:
...
+        if not full_word:
```
altendky, line 105: Personal preference, I'm on the side of being more explicit.

```suggestion
        if full_word is None:
```

chia/util/keychain.py
```diff
@@ -87,11 +87,32 @@ def bytes_to_mnemonic(mnemonic_bytes: bytes) -> str:
...
+            raise ValueError(f"'{word}' is not in the mnemonic dictionary; may be misspelled")
```
altendky, line 106: Let python format it in a quoted form.  For example, if there's a `'` in there then the existing quoting would be odd.  The `r` is for `repr()` rather than `str()` which is used otherwise.

```suggestion
            raise ValueError(f"{word!r} is not in the mnemonic dictionary; may be misspelled")
```

tests/core/util/test_keychain.py
```diff
@@ -165,6 +166,28 @@ def test_bip39_test_vectors(self):
             assert bytes_to_mnemonic(entropy_bytes) == mnemonic
             assert mnemonic_to_seed(mnemonic) == seed
 
+    def test_bip39_test_vectors_short(self):
+        """
+        Tests that the first 4 letters of each mnemonic phrase matches as if it were the full phrase
+        """
+        with open("tests/util/bip39_test_vectors_short.json") as f:
+            all_vectors = json.loads(f.read())
```
altendky, line 174:
```suggestion
            all_vectors = json.load(f)
```

tests/core/util/test_keychain.py
```diff
@@ -165,6 +166,28 @@ def test_bip39_test_vectors(self):
             assert bytes_to_mnemonic(entropy_bytes) == mnemonic
             assert mnemonic_to_seed(mnemonic) == seed
 
+    def test_bip39_test_vectors_short(self):
+        """
+        Tests that the first 4 letters of each mnemonic phrase matches as if it were the full phrase
+        """
+        with open("tests/util/bip39_test_vectors_short.json") as f:
```
altendky, line 173: This should be loading a resource probably, or at least working relative to some known path like `tests.__file__`.  This would be the minimal.  `pkg_resources` or `importlib.metadata` would be the bigger thing, but...  I need to do some work around those.

```suggestion
        with pathlib.Path(tests.__file__).joinpath("util", "bip39_test_vectors_short.json") as f:
```

tests/core/util/test_keychain.py
```diff
@@ -165,6 +166,28 @@ def test_bip39_test_vectors(self):
...
+            all_vectors = json.loads(f.read())
+
+        with open("tests/util/bip39_test_vectors.json") as f:
+            full_vectors = json.loads(f.read())
```
altendky, line 177:
```suggestion
        with pathlib.Path(tests.__file__).joinpath("util", "bip39_test_vectors.json") as f:
            full_vectors = json.loads(f.read())
```

tests/core/util/test_keychain.py
```diff
@@ -165,6 +166,28 @@ def test_bip39_test_vectors(self):
...
+
+        for idx, vector_list in enumerate(all_vectors["english"]):
```
altendky, line 179: I would probably unpack here first and avoid 0, 1, 2...  but yeah, it is still index based access so...  not a biggy.

```suggestion
        for idx, [this, that, the other] in enumerate(all_vectors["english"]):
```

Parallel lists are also icky, but when you have them you can zip.

```suggestion
        for idx, [av, fv] in zip(all_vectors["english"], full_vectors["english"]):`
```

Where that can be combined with unpacking like `for index, [[e, m, s], [_, f, _]] in...`.  Or...  using good names of course.

tests/util/bip39_test_vectors_short.json
```diff
@@ -0,0 +1,124 @@
+{
+    "english": [
```
altendky, line 2: Maybe just add the single new entry to the existing file?  or even generate it on load.

altendky (changes requested):
Seems like a good feature, thanks.

tests/core/util/test_keychain.py
cmmarslender: Updated in last commit

altendky (approved):

emlowe (approved):

PR #16708, opened 2023-10-27
avoid recursive_replace() in `block_creation.py`

arvidn:
### Purpose:

It makes the code a bit more linear and avoids copying objects.
This is the only production use of `recursive_replace()`, and with this it could be turned into a test-only facility.

My long term intention is to make these types (`FullBlock` and all its members, transitively) not be dataclasses, but instead native types. This poses a challenge with all the places we use `replace()`. `recursive_replace()` is especially challenging to substitute.

This is a small step towards reducing the use of `replace()`. One possibility might be to have these objects be mutable, to cover some of the existing use of `replace()`, but not all.


altendky (approved):
Much better, thanks.

wjblanke (approved):
aok

PR #16717, opened 2023-10-27
Support for third-party, farmer-rewarded, Harvesters

harold-b:
This adds support for third-party Harvesters to be able to overwrite the farmer reward puzzle hash when it is time for them to take their reward, given a winning block.

- [x] Harvester overwrite address
  - [x] ~Test overwrite address is respected and the `UnifinishedBlock` includes it~
  - [x] Test correct fields for data source are included when the Harvester requests them.
- [x] introduce convention for agreement between Harvester & Farmer as to when the Harvester is allowed to take the farmer reward by overwriting the reward address.
  - [x] Add test for Harvester-Farmer reward convention

~- [x] `install_timelord = True` must be enabled for new tests~ Timelord test requirement removed.


madMAx43v3r:
Great stuff, but there's one issue, the harvester cannot blindly sign any message hash, it needs the original data for all messages that need to be signed:
- `FoliageBlockData` for the reward block, and the `RewardChainBlockUnfinished` structure (corresponding to the `unfinished_reward_block_hash`)
- optional `FoliageTransactionBlock` in case of transaction blocks
- `VDFInfo` for `NewSignagePoint.challenge_chain_sp` (from `full_node_protocol.RespondSignagePoint.challenge_chain_vdf.output`)
- `VDFInfo` for `NewSignagePoint.reward_chain_sp` (from `full_node_protocol.RespondSignagePoint.reward_chain_vdf.output`)
- `PostPartialPayload` for partials

If the harvester has an API to blindly sign message hashes, that API can be used to get a signature for a block header that does not give the dev fee.

harold-b:
Good to see you, here @madMAx43v3r. I was going to reach out to you this week, so I'm happy to see that you're already involved here.

I was assuming the validation of what was being signed would be determined by the harvester given the `plot_identifier` and `challenge_hash` used by the winning proof of space, but we can definitely send back the original payload for each hash being signed. I think that makes it simpler.

Please let me know if you have any more feedback.

madMAx43v3r:
Ok so let me simplify the problem. Right now we essentially have an API for signing a message hash (array index 0) and an API for signing a data structure that _can_ be validated (array index 1).

What stops someone from using the API that doesnt validate anything and just signs the message hash for the message that is supposed be validated? You can simply bypass the validation.

So yeah, unfortunately the harvester will need all the data structures to be signed in original form. Maybe they can be passed in serialized form, ie. byte array, to simplify the protocol.

Since the message type information is not hashed, we have to rely on the serialized byte count to determine what kind of message it is. Not prepending a type string for the message hashes is really dangerous, but there's nothing we can do about it now, it's part of the consensus already.

Even if you don't serialize the messages, you can still spoof the API into not validating the message that should be validated by using another message type and making it serialize to the same bytes, so the hash is what you need to be signed. This only works if another message has the same serialized size of course. Just saying, if the type information was included in the hash, this attack would not work. So now we have to be careful about message size collisions. If the message to be validated has a dynamic field that can take any size, like a string or byte array, then this whole idea will not work... Remember the CAT1 debacle was an attack of the same nature.

For the random dev fee lottery I would just hash the `ProofOfSpace.challenge` and `ProofOfSpace.proof` together and use the first or last byte as a random number. The harvester then also needs to send a 8-bit number to the farmer / node to specify the fee threshold. When the random 8-bit is <= the fee threshold given by the harvester the fee should be paid. Also need a boolean to disable dev fee of course.

madMAx43v3r:
Looking at the code again now, the message to be validated is `FoliageBlockData`, which has one optional field, so it can have two different sizes. Not ideal but we might get lucky that no other message possibly can have one of those sizes:

`FoliageBlockData` can be 133 or 165 bytes (with pool signature)
`FoliageTransactionBlock` is 168 bytes (close call with 165)
`VDFInfo` is 140 bytes (close again to 133)
`PostPartialPayload` is over 300 bytes, so not an issue

`RewardChainBlockUnfinished` is not relevant because it's not being signed, but is still needed to validate `FoliageBlockData`.

madMAx43v3r:
> Then rigidly reject signing anything that did not follow the expected signing [flow](https://camo.githubusercontent.com/13f127cd2f720424f9833ceabf5afeb3a3fd9b3da9fa9e882393d3931d2ebd70/687474703a2f2f7777772e706c616e74756d6c2e636f6d2f706c616e74756d6c2f70726f78793f63616368653d6e6f267372633d68747470733a2f2f7261772e67697468756275736572636f6e74656e742e636f6d2f656d6c6f77652f636869612d73747566662f6d61696e2f6661726d696e672e70756d6c) for that proof that it found.

I don't think that would work. For example I could replace the signature request for the partial with the `FoliageBlockData` for a block that does not pay the fee, and present the expected `FoliageBlockData` to the harvester. So the harvester will see what it expects for the `FoliageBlockData`, but I will use the other signature I got via the fake partial to publish a block that doesn't pay the fee.

As long as there exists any path where you can get a signature for any hash, it allows to bypass the fee.

> The Harvester just need to request the original data be included when it sends its proof of space to the Farmer.

So a harvester with dev fee would then always set that bool to true right?

madMAx43v3r:
Just saw your other commit. I missed `ChallengeChainSubSlot` and `RewardChainSubSlot` before, let's check their serialized message sizes.

`ChallengeChainSubSlot` can be 144, 152, 160, 176 bytes or more (not relevant)
`RewardChainSubSlot` can be 174 or 206 bytes

So got lucky again, no collision with `FoliageBlockData`.

harold-b:
> I don't think that would work.

I think you're right. The partial payload would make it troublesome.

> For the random dev fee lottery I would just hash the ProofOfSpace.challenge and ProofOfSpace.proof together and use the first or last byte as a random number...

That sounds good, is that what you currently do? We decided to let the Harvester itself take care of that without any convention agreement. We would simply display the stats to the user. The Harvester provider would not get locked-in into any convention this way and simplifies things from the Farmer side.

> So a harvester with dev fee would then always set that bool to true right?

Yes, I also changed it so that a bool field can be specified [here](https://github.com/Chia-Network/chia-blockchain/pull/16717/files#diff-b8692b7c49cc660b8d9ee43584d492ca8c90b449d0da609435fdbdbe4e32e428R58) so that the farmer would include signing data even if the reward address will not be overwritten. Though overwriting the reward address also implies sending back source data for signing.

Is there anything else that sticks out to you as troublesome from the current changes?

madMAx43v3r:
> That sounds good, is that what you currently do?

No

> We decided to let the Harvester itself take care of that without any convention agreement

That means an optional address field for the fee address in the proof response? I'm fine with that.

> Yes, I also changed it so that a bool field can be specified [here](https://github.com/Chia-Network/chia-blockchain/pull/16717/files#diff-b8692b7c49cc660b8d9ee43584d492ca8c90b449d0da609435fdbdbe4e32e428R58) so that the farmer would include signing data even if the reward address will not be overwritten.

That's needed for sure, I wont sign anything without hashing all the data myself. Doing some if-else magic there is just asking for a loophole.

chia/protocols/harvester_protocol.py
```diff
@@ -60,6 +79,7 @@ class RequestSignatures(Streamable):
     challenge_hash: bytes32
     sp_hash: bytes32
     messages: List[bytes32]
+    message_data: Optional[List[SignatureRequestSourceData]]
```
arvidn, line 82: would it not be simpler to make this field a `SignatureRequestSourceData` (instead of a list of them), and lift the restriction of just having one field set.

Don't we want to send all the source data anyway?

chia/protocols/harvester_protocol.py
```diff
@@ -71,6 +91,7 @@ class RespondSignatures(Streamable):
     local_pk: G1Element
     farmer_pk: G1Element
     message_signatures: List[Tuple[bytes32, G2Element]]
+    farmer_reward_address_overwrite: Optional[bytes32]
```
arvidn, line 94: not important; my model is that this is a "diverge" rather than an "overwrite". The latter suggests that we first put the default farmer reward address in, and then overwrite it. Perhaps we implement it like that, but it's not an important distinction.

chia/protocols/farmer_protocol.py
```diff
@@ -27,6 +50,7 @@ class NewSignagePoint(Streamable):
     sub_slot_iters: uint64
     signage_point_index: uint8
     peak_height: uint32
+    sp_source_data: SignagePointSourceData
```
arvidn, line 53: since these changes to the harvester protocol are not a backwards compatible, I think you should:

1. separate the "protocol version" constant we have right now, into one for each protocol (full node, farmer and harvester)
2. bump the harvester protocol version
3. make sure the handshake fails and prints a clear error if the harvester is handshaking with the wrong version

chia/protocols/farmer_protocol.py
```diff
@@ -42,6 +66,7 @@ class DeclareProofOfSpace(Streamable):
     farmer_puzzle_hash: bytes32
     pool_target: Optional[PoolTarget]
     pool_signature: Optional[G2Element]
+    include_signature_source_data: Optional[bool]
```
arvidn, line 69: I don't think this bool needs to be optional. Presumably `None and `False` are interpreted  the same.

chia/protocols/harvester_protocol.py
```diff
@@ -51,6 +55,21 @@ class NewProofOfSpace(Streamable):
     plot_identifier: str
     proof: ProofOfSpace
     signage_point_index: uint8
+    include_source_signature_data: Optional[bool]
```
arvidn, line 58: same here, unless I'm missing something subtle, I would expect this to just be a `bool`

chia/protocols/harvester_protocol.py
arvidn: excuse my ignorance, but couldn't this just be:
```
    foliage_block_data: Optional[FoliageBlockData]
    foliage_transaction_block_data: Optional[FoliageTransactionBlock]
```

as well? Or does the harvester sign more than those two blocks?

madMAx43v3r: In most cases the harvester does sign two messages at once.

chia/protocols/farmer_protocol.py
madMAx43v3r: Or use new messages ids? I think that's what has been done before.

harold-b: I had asked Earle about this before I made the changes. He mentioned we should just modify existing messages and bump the version number and that we ought to just let people know they have to upgrade.

Are you sure we want to include the scope of separating the protocol versions per node type as part of this? Not necessarily opposed, but...

chia/protocols/harvester_protocol.py
harold-b: That's precisely why I changed it to "overwrite" (originally named it override), since we are indeed overwriting the default one. I've no problem going back to override, as it seems more intuitive

chia/protocols/harvester_protocol.py
harold-b: The Harvester is able to sign hashes belonging to serialized versions of all of the fields listed in `SignatureRequestSourceData`. I added all the known types it signs in that list. First signature requested is for the signage points (which can be VDFInfos or one of the 2 SubSlot ones), then, potentially, a Partial, finally FoliageBlockData, which could or could not include the transaction block.
They come in pairs usually, like Max said, except for the Partial which is sent by itself. It's in a List so that the data corresponding to the particular hash being signed can easily be obtained via the same index.

The fields are also separate even with the same data type (ie VDFInfo) so that the original purpose of data can be identified via its field location

harold-b:
> That means an optional address field for the fee address in the proof response? I'm fine with that.

Yes. Probably also specified whenever `RespondSignatures` is sent back to continue the chain. This avoid having to add stateful data to the Farmer

harold-b:
FYI @madMAx43v3r, it's supposed to be `ClassgroupElement` for the signing data source instead of `VDFInfo` [here](https://github.com/Chia-Network/chia-blockchain/pull/16717/files#diff-b8692b7c49cc660b8d9ee43584d492ca8c90b449d0da609435fdbdbe4e32e428R69-R70). Updated in last commit

madMAx43v3r:
Yeah you right, those are 100 bytes, all good.

madMAx43v3r:
>  was assuming the validation of what was being signed would be determined by the harvester given the plot_identifier and challenge_hash used by the winning proof of space

There's a lot of challenge hashes and it's easy to get confused:
- the static challenge for 64 SPs in a row
- the challenge chain VDF output hash for each SP
- the individual challenge for each plot (hashed with plot id)

I'm not sure if you can simply change the `challenge_hash` in `RequestSignatures` and still get the needed signature while bypassing the fee. After all that field is not being hashed / signed directly.

chia/protocols/farmer_protocol.py
```diff
@@ -16,6 +20,26 @@
...
+class SPVDFSourceData(Streamable):
+    cc_vdf: ClassgroupElement
+    rc_vdf: ClassgroupElement
+
+
+@streamable
+@dataclass(frozen=True)
+class SignagePointSourceData(Streamable):
+    sub_slot_data: Optional[SPSubSlotSourceData]
+    vdf_data: Optional[SPVDFSourceData]
```
arvidn, line 41: this is not a strong opinion, but it seems like the existing protocol suffers from trying to over-generalize (by just sending an arbitrary number of un-typed hashes to be signed). Rather than doubling down on this generalization, I think there's an opportunity to simplify the protocol to more closely match what we actually use and need.

Rather than sending both the hashes (un-typed) and then a (relatively deep) structure with typed source-data, couldn't you just send the source data as optional fields? and whichever field is set indicates what we want to be signed. Maybe we can even make the hashes optional, and only send those when we use our harvester that doesn't verify what it's signing.

madMAx43v3r: I'm not an expert in python, but in C++ you would just have base class that every structure inherits from and then you can pass a list of pointers with the base type. The implementation can then check each instance what the actual type is if needed, but the hash function would be generic. There would not be any need to invent new types just for this purpose, or using a list of optional fields. If that is possible in python I have no idea...

.gitignore
```diff
@@ -305,3 +305,6 @@ tags
 # Ignore the binaries that are pulled for the installer
 /bladebit/
 /madmax/
+
+# Sandbox/playground
+.sandbox/
```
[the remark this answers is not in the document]
harold-b: I use this directory to experiment and to store the chia root dir when developing. Nothing to worry about as it will be removed before merge

chia/full_node/full_node_api.py
```diff
@@ -944,10 +945,19 @@ def get_pool_sig(_1: PoolTarget, _2: Optional[G1Element]) -> Optional[G2Element]
                 foliage_transaction_block_hash = bytes32([0] * 32)
             assert foliage_transaction_block_hash is not None
 
+            foliage_block_data: Optional[FoliageBlockData] = None
+            foliage_transaction_block_data: Optional[FoliageTransactionBlock] = None
+            if request.include_signature_source_data is True:
+                foliage_block_data = unfinished_block.foliage.foliage_block_data
+                if unfinished_block.is_transaction_block():
+                    foliage_transaction_block_data = unfinished_block.foliage_transaction_block
```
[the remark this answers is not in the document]
harold-b: I think it returned a defaulted transaction block, if I recall when debugging. I'd have to check again

chia/protocols/farmer_protocol.py
harold-b: I opted for a more conservative approach when modifying the protocol since I was not aware as to the reasoning behind taking a generalized approach when signing the messages. If I have your blessing to make it more specific and explicit, I am all for it.

nossdpool:
> Update on the "convention" mentioned above:
> No convention will be used. Instead, the Harvester would decide itself when to take its fee. The farmer will respect its decision without doing any checks on its side. Instead only stats would be presented to the user.

Why have you decided to simplify it this way? Instead of just gathering statistics there can be a defined algorithm how to determine which block reward is going to be claimed by a harvester. `FoliageBlockData` refer to `RewardChainBlockUnfinished` which in turn contains `ProofOfSpace` and `challenge_chain_sp` which hash can be used as a deterministic source of randomness for example. This way farmer and a harvester can use the same decision algorithm and fee percentage will be real and not just "statistics". Most users wouldn't even able to mine enough blocks to gather statistics and a harvester can cheat in many ways, on some machines it will take small percent on another machine it will take bigger percent and attribute it all to bad luck. Lets just use pseudo random hashes and make reward block claims deterministic.

harold-b:
> Why have you decided to simplify it this way? 

I'll tag-in @wjblanke and @arvidn for further discussion on that

madMAx43v3r:
I can confirm some people get paranoid when they are unlucky with the fee lottery.

chia/protocols/farmer_protocol.py
arvidn: I'm a proponent of "you-ain't-gonna-need-it". i.e. keep things as concrete (and simple) as possible, and refactor later if you need to.

arvidn: > in C++ you would just have base class that every structure inherits from and then you can pass a list of pointers with the base type

This type describes the network protocol, and our serialization is somewhat limited. We don't support a variant type for example. The alternative would be a series of optional fields, where one is set.

arvidn:
> Why have you decided to simplify it this way?

Requiring a specific mechanism to be used to determine which blocks have their reward diverted adds complexity and limits the freedom of the harvester. If one of you guys come up with some new clever scheme of picking blocks to divert, you would need an update to the protocol.

The point of doing this, as you point out, would be that both the farmer and the harvester could check each other. However, it's not obvious what the best action would be, if a discrepancy is discovered.

Refusing to sign, and squander the block-win doesn't help *anyone*. The farmer loses the fees and pool reward, the harvester loses its reward and the network loses netspace. It's kind of the "doomsday machine"-option; incentivize conformance through mutual destruction. I also think it has questionable value.
1. The farmer running a 3rd party harvester already implicitly trusts it.
2. I would think, by far, the most likely cause of a discrepancy would be a bug

nossdpool:
> come up with some new clever scheme of picking blocks

In current design the most obvious clever scheme is to cheat on user. For example a difference between 3% fee and a 5% fee is so tiny that only a really big miner can trace with a high confidence but for a harvester author it is a huge difference.

The algorithm can be very simple and straightforward. Get the lower 32 bit of an sp_hash and compare it with a value `fee * UINT_MAX`.

> The farmer running a 3rd party harvester already implicitly trusts it

Not having a strict and deterministic reward distribution algorithm opens a room for social engineering attacks on competitors. Fanboys of a harvester author A will spam all over the internet that a harvester B is cheating and claiming bigger fee than it declares. We want to focus on programming and not spend resources on SMM.

arvidn:
> Not having a strict and deterministic reward distribution algorithm opens a room for social engineering attacks on competitors.

I believe validating such algorithm could even be done on-chain, right? no need to extend the harvester protocol or build it into the node or farmer.

madMAx43v3r:
It's not that simple, GH has CPU / GPU fee for example, on-chain you never know if it was farmed with CPU or GPU. There's also uncompressed plots in the mix as well sometimes.

Since there's only two providers right now, it would be easy to agree on something. First 32-bit of `sp_hash` is fine by me.

I think this would be really helpful, to give users the peace of mind.

nossdpool:
Have thought a little more. `sp_hash` can be altered by a timelord, so probably a pos challenge is a better option, but be careful to choose bits from the middle of it, they shouldn't overlap with bits actually used for challenge.

harold-b:
Thanks for the feedback, all. Perhaps we can continue the discussion to implement the convention outside the scope of this PR and we can get this out first, then we can follow up with a convention that seems appropriate for all parties without having that hold the rollout of this one. Sound fair?

madMAx43v3r:
My first proposal was to hash plot challenge and proof together, which should be fool proof.

> then we can follow up with a convention that seems appropriate for all parties

We still need to incorporate it into the PR, for example the fee percentage needs to sent to the farmer / full node, in some format that is not floating point. Such as a 32-bit threshold.

arvidn:
> We still need to incorporate it into the PR, for example the fee percentage needs to be sent to the farmer / full node, in some format that is not floating point. Such as a 32-bit threshold.

I don't think we want the farmer to interpret or take any action on this fee rate auditing. Perhaps I've misunderstood some of your points though. If the main objective (as I believe it to be) is to be able to tell the user that the fee rate is exactly what it is supposed to be, and exactly the correct blocks are diverted, we don't need to involve the farmer.

The auditing can be done separately.

madMAx43v3r:
> The auditing can be done separately.

At least with GH the fee depends on each proof separately, if it was farmed with CPU or GPU or uncompressed. So the farmer / node need to check every block if the fee is correct, not some average over time.

In order to do this the farmer / node need to know the fee rate that applies to each proof / block. Without this information it's impossible to double check.

nossdpool:
`challenge_hash` is the same for 64 SPs, better use `sp_hash` or pos challenge (aka `sp_challenge_hash`, `calculate_pos_challenge`) which includes all of the above. Adding proof to the hash is also a good idea.

The fee threshold should be declared by the harvester at startup, not with each proof. Otherwise, it creates the potential for misleading fee reporting. For example, if you plan to take a 3% fee and the current proof falls well below 2%, you might report 2% to please the user. If you plan to take a 3% fee and the current proof is well above it, you might report 0% to please the user.

nossdpool:
You can collect CPU rewards in one wallet and GPU rewards in another to differentiate between the two

madMAx43v3r:
> challenge_hash is the same for 64 SPs

That cannot be, otherwise you would get the same proofs for every 64 SPs.

> The fee threshold should be declared by the harvester at startup

That wont work for GH, since I support uncompressed as well. Not to mention CPU / GPU mix, even for the same harvester you can mix CPU and GPU via multiple remote compute.

For GH the fee will be specific to each proof, there's no way around it.

nossdpool:
`challenge_hash` (aka `cc_hash`, "𝑠𝑢𝑏 𝑠𝑙𝑜𝑡 𝑐ℎ𝑎𝑙𝑙𝑒𝑛𝑔𝑒") is the same for 64 SPs
`calculate_pos_challenge` uses `plot_id`, `challenge_hash` and `sp_hash` to calculate the pos challenge.

Specifying a fee for each proof undermines the entire idea of farmer assisted fee verification. If the fee cannot be set at startup, then we are inclined to support the current protocol, which does not include verification by the farmer. We may also publish a tool that allows users to verify our reward claims on the blockchain, although this is not user friendly.

To achieve the same, you can send different types of rewards to separate wallets. For instance, CPU reward percentages could go into one wallet, GPU percentages into another, and so on. This method would allow you to make a posteriori verification.

madMAx43v3r:
> calculate_pos_challenge uses plot_id, challenge_hash and sp_hash to calculate the pos challenge

True, but then your previous statement about the timelord being able to control the `sp_hash` is misleading. That would be a serious flaw in the consensus, as it would allow to grind for better proofs.

In any case, we can use `challenge` in `ProofOfSpace` also, to hash together with `proof` bytes. Makes no difference, both are available in `RewardChainBlockUnfinished`.

> If the fee cannot be set at startup.

This is simply not possible with GH, as I support uncompressed plots as well, and you can mix CPU / GPU on the same harvester.

> Specifying a fee for each proof undermines the entire idea of farmer assisted fee verification

I don't see how, the farmer can double check every block if the fee was taken correctly or not. The logs can show the fee percentage like I already do it now in GH.

madMAx43v3r:
Verifying the fee on the blockchain is pretty much impossible, since you never know when somebody switched to a different harvester provider, or their mix of plot types over time, etc.

You can use the same farmer_key for uncompressed and compressed plots with GH and BB.

arvidn:
> Verifying the fee on the blockchain is pretty much impossible, since you never know when somebody switched to a different harvester provider, or their mix of plot types over time, etc.

If you are the person running the farm can know, right?
My understanding is that the use case we're interested in is allowing the farmer (the human, not the process) ensure that the harvester behaves the way it's expected to.

I don't see much value in involving the farmer (process) in this validation. That seems like unnecessary complexity and coupling.

madMAx43v3r:
> I don't see much value in involving the farmer (process) in this validation.

I see it as the only way to do it properly.

arvidn:
> I see it as the only way to do it properly.

> All I'm asking for is to send the fee threshold back to the farmer for verification, just one additional 32-bit integer. Actually implementing the verification is optional, but also kinda trivial.

What would you expect or want the farmer to do if it implements verification?
Would the farmer be able to know whether a plot was CPU or GPU compressed?
As I understand it, only the harvester would know this. The only benefit of having the farmer implement the verification would be that it already knows the compression type, as far as I can tell.
Are there other reasons to implement this in the farmer?

madMAx43v3r:
> What would you expect or want the farmer to do if it implements verification?

Log an error if fee is taken wrongly.

>  Would the farmer be able to know whether a plot was CPU or GPU compressed?

Yes, indirectly through the fee threshold (just a different representation of the fee percentage)

>  Are there other reasons to implement this in the farmer?

Like I said, it's the only way I can see to verify it properly. All other methods won't work in every use case.

harold-b:
I don't think anyone has mentioned anything about setting a threshold in `NewProofOfSpace`. It would obviously be done in the handshake.

harold-b:
Given that you both seem interested in having that convention in place, let's put it back in the table. I believe having @madMAx43v3r 's suggestion of hashing both challenge and proof as sources of entropy ought to be good enough.

I was made aware, though, that GH has varying fees depending on compression level. Would you, in this case send back the highest threshold on handshake? That is, the user would still have to trust that if they only have plots of a C-level that is of a lower fee %, that indeed that is what the harvester would use, even though the higher-end of the threshold is sent back initially. Would that be acceptable?

@nossdpool Do you guys have any objection to using `(hash(proof | challenge) mod 2^32) < fee_threshold` for fee conention?

madMAx43v3r:
> I don't think anyone has mentioned anything about setting a threshold in NewProofOfSpace. It would obviously be done in the handshake.

Sure ok, but then forget about verifying CPU GH fee. And I could take 3% fee on uncompressed plots also.

> What value should you now return as the threshold? 0.03 or 0.02?

3.125%, 1.625% and 0% are the only possible values for GH, anything other than those would indicate fraud. The average fee doesn't have any meaning, and would only show somebodies mix of CPU / GPU / uncompressed.

All a user needs to check (for GH) is that it's never higher than 3.125%, and somebody who only CPU farms would be alarmed if it ever showed more than 1.625%.

In the case of mixing CPU / GPU / uncompressed a user would have to check the plot id and see if that plot is computed according to the fee.

> GH has varying fees depending on compression level

No, it depends on if a proof was computed with CPU or GPU. Or if it's an uncompressed plot.

> Would you, in this case send back the highest threshold on handshake?

That would be the only way, because with remote compute you never know in advance if a harvester has GPU or not, and it can change at runtime also.

> Would that be acceptable?

Not really, because the GUI and logs would then display 3% fee even if someone only farms with CPU. My proposal would still work just fine.

Nobody expects us to cheat the fee, and 95% of people trust us. The remaining 5% will check the logs if they suspect something. So this verification doesn't need to be super user friendly, it just needs to be 100% accurate.

harold-b:
> Sure ok, but then forget about verifying CPU GH fee. And I could take 3% fee on uncompressed plots also.

I am not sure what you're responding to here since this was a response to nossd's last message.

> Not really, because the GUI and logs would then display 3% fee even if someone only farms with CPU. My proposal would still work just fine.

Perhaps there's a misunderstanding since you seem to agree just above that saying _"All a user needs to check (for GH) is that it's never higher than 3.125%"_. So you are in agreement then with setting the maximum case scenario fee during the handshake _(ie 3.125% in your case)_ , even though the appropriate amount will be taken given the method of decompression, which could be less, correct?

madMAx43v3r:
> So you are in agreement then with setting the maximum case scenario fee during the handshake.

No I'm not, since this only works for pure GPU farms.

CPU farmers and farmers with a mix of uncompressed plots would not have the peace of mind of having the open source code double check the closed source fee logic. And they would be confused / alarmed by seeing a 3% fee displayed for GH.

If you want to do the handshake method, there would need to be a config setting like "I'm CPU farming only with GH" to enable checking for a maximum fee of 1.625%. But then you still would not know if a proof was coming from an uncompressed plot or not, with my approach the GUI could display a 0% fee in this case.

Currently in GH if a proof is found a message like this is logged:
`INFO     Found proof: K = 32, C = 15, is_compressed = True, used_gpu = True, fee_rate = 3.125 %`

And when the fee is taken from a block win a message like this is logged:
`WARNING  Used farmer reward of block 2187769 as dev fee (3.125 % on average)`

With the handshake method the fee percentage information would be lost.

harold-b:
> And they would be confused / alarmed by seeing a 3%

They wouldn't see a 3% fee if what was taken is not based on 3% but instead of 1.625%. 3% would be the maximum allowed. Or do you mean that, even if the 3% would never be taken, they would be uneasy because, visually, a "maximum fee of 3%" might be displayed?


> ie. clearly I have mentioned multiple times the need of setting a threshold in NewProofOfSpace.

I see. This would be of no use in terms of "trust" if there's no initial agreement in the handshake. It is equivalent to what was mentioned before about simply showing the stats, instead of agreeing on a convention. Since a harvester could simply set an arbitrary fee whenever (that is, equivalent to just showing stats).



It seems like you both want different things from this convention, then.

harold-b:
> The lowest one is zero.

Mmm right, so I don't think this would work at all for a variable rate fee.

madMAx43v3r:
> What exactly do you mean by your method? Setting the fee % used in each proof of space found?

Yes, as I have stated over 5 times now. This is how it works in GH right now. With the only difference of no verification.

> Whatever you return, the point is it is agreed upon ahead of time, not at the time of a proof is found.

The point of this verification is not to avoid cheating, it's to _detect_ cheating. For detection no agreement ahead of time is needed.

harold-b:
How can we possibly detect cheating if all we're doing is trusting what the harvester tells us each proof of space if nothing is set during the handshake?

madMAx43v3r:
> How can we possibly detect cheating if all we're doing is trusting what the harvester tells us each proof of space if nothing is set during the handshake?

See my previous message, I just wrote it.

madMAx43v3r:
> @madMAx43v3r do you want the farmer to validate anything? Or do you want to simply display fee % as given by the harvester?

I want the farmer to display the fee percentage for every proof found _and_ validate that the fee was taken correctly according to the convention.

harold-b:
> It's easier if you tell me what the difference is between returning a constant fee percentage once, and a possibly variable fee for every proof?

For example, the use could have a config value set of acceptable fees, which they can lookup beforehand for what is advertised for the harvester they want to use, then both sides can validate that upon handshake

harold-b:
> Both approaches still require user checking of the fee percentage, either at every startup or when winning a block, the only thing that we can automate is the computation of the fee convention formula.

Yup, but ideally it would be at handshake as that is usually when the user is explicitly interacting with the software.

madMAx43v3r:
> For example, the use could have a config value set of acceptable fees, which they can lookup beforehand for what is advertised for the harvester they want to use, then both sides can validate that upon handshake

That doesn't help anything, I could still take a higher fee.

Like I said, all we can do is automate the validation of the fee convention. The percentage needs to be checked _by the user_ for every block win individually, there is no way around it.

> Yup, but ideally it would be at handshake as that is usually when the user is explicitly interacting with the software. After that, they shouldn't need to worry about it. The harvester accepted the fees, and so did the farmer.

What about command line usage? It seems you just want to display an information to the user so they know what the fee is, you can still do that with the handshake.

madMAx43v3r:
Alright, I feel like I'm talking to a wall. Just do the handshake, but know I will return 12.5%

I can make my own PR later to add proper validation for GH, and you can merge it or not.

harold-b:
There's no need to be rude. The whole point of this discussion is to try and come up with a convention where everyone is happy.

I see that you simply want to validate that the formula was applied correctly.

felixbrucker:
> I see that you simply want to validate that the formula was applied correctly.

I think that is the only thing that can be done by the farmer (in an automated way), given that the reward percentage is dynamic and can change at any time during runtime. The farmer does not know which fee percentage is valid, only if a fee block is valid for a given percentage, according to the agreed formula.

Exchanging a fee percentage on handshake is kinda pointless unless the fee percentage were static for the entire runtime of the harvester.

harold-b:
> Exchanging a fee percentage on handshake is kinda pointless unless the fee percentage were static for the entire runtime of the harvester.

This isn't being done for only *1* harvesting service provider, but as a convention for *all* current and future. We can't just take 1 party's concerns and ignore the rest.

We were talking about different concepts of "validation" here. I certainly fine with applying the formula by convention on what the harvester sends over given a fee % sent as part of the proof of space message. If @nossdpool is fine with it as well, I think we can move forward with that as the starting point.

felixbrucker:
> This isn't being done for only 1 harvesting service provider, but as a convention for all current and future. We can't just take 1 party's concerns and ignore the rest.

Exactly, and as this won't work for one of them (GH) it is not suitable as a general solution

> If @nossdpool is fine with it as well, I think we can move forward with that as the starting point.

I'd love for this to move forward as well 👍

madMAx43v3r:
Actually I will return 0% fee on the handshake, maybe that helps to understand.

nossdpool:
We don't insist on any particular implementation, actually, just pointing out potential flaws.

At the moment solution may look like this:

1) Harvester declares maximum fee in a handshake (it can be shown somewhere in Chia UI)
2) Dynamic per proof threshold is redundant. Farmer can just print an actual fee percent for the claimed reward block {hash[:4] / 2^32 * 100}. The message will be like "INFO: The harvester claimed the reward that accounted for the 2.17% fee", here 2.17 is an actual 32 bit value from the hash calculated from the challenge_hash, proof, sp_hash, whatever. Everything else, plot type etc, harvester can output to the log on it own.

We are ok with the current protocol implementation. If you want to add a redundant dynamic threshold field we are fine too, though users may not be aware that it's not enforced by the farmer and may be misled by it.

madMAx43v3r:
Given the lack of effort from CNI and incredible difficulty in communicating I'm calling it quits on this project.

harold-b:
@madMAx43v3r @nossdpool would you guys like to properly meet in real time? Perhaps we can get everything cleared up and remove any communication barriers from here. We can get Bram in there too. Communication seems to be breaking down here so I think it would be helpful.

nossdpool:
Our comments were suggestions on how to avoid creating a protocol that misleads users into believing that the fee is predefined and rigorously checked by a farmer, while it is actually dynamic and provided by a harvester for informational purposes only.

We don't insist on passing the maximum fee on handshake, although it's a good additional precautionary measure.

We plan to collect a fixed fee, as we are currently doing on the pool. Therefore, any scheme (with or without a per-proof fee) is acceptable to us. We may implement a tool to check the fee on the blockchain, as suggested by @arvidn. Since the current protocol implementation is sufficient for this purpose, we approve it also.

danieljperry:
@madMAx43v3r @nossdpool This PR is described in [CHIP-22](https://github.com/Chia-Network/chips/pull/88). We would like to have a public Zoom call to discuss these changes. Would either of you be interested in joining? It would be nice to have you there. Due to the US holiday this week, it would be ideal to meet during the week of Nov 27-Dec1

tests/farmer_harvester/third_party_harvesters/test_third_party_harvesters.py
```diff
@@ -0,0 +1,331 @@
...
+                and validated_sub_slot_rc
+            )
+
+            assert data is not None
+            data_hash = data.get_hash()
+            assert data_hash == hash
+
+    async def intercept_farmer_new_proof_of_space(*args: Any) -> None:
+        nonlocal farmer
+        nonlocal farmer_reward_address
```
arvidn, line 170: I don't understand the rationale behind declaring these as `nonlocal`. the outer scope here is the global scope, isn't it?

arvidn:
I would expect this test to fail when you change the protocol messages.
https://github.com/Chia-Network/chia-blockchain/blob/main/tests/util/test_network_protocol_files.py

chia/farmer/farmer_api.py
```diff
@@ -256,17 +270,20 @@ async def new_proof_of_space(
 
                 # The plot key is 2/2 so we need the harvester's half of the signature
                 m_to_sign = payload.get_hash()
-                m_src_data: Optional[List[SignatureRequestSourceData]] = None
+                m_src_data: Optional[List[Optional[SignatureRequestSourceData]]] = None
```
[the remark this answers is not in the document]
harold-b: I believe this is because there may not be a `FoliageTransactionBlock` sent over from the `FullNode` if the `UnfinishedBlock`'s `is_transaction_block()` returns `False`, so we have nothing to place there as the source data in that case.

chia/farmer/farmer.py
```diff
@@ -826,3 +826,29 @@ async def _periodically_clear_cache_and_refresh_task(self) -> None:
...
+
+        fee_quality = calculate_harvester_fee_quality(proof_of_space.proof.proof, sp.challenge_hash)
+
+        if proof_of_space.fee_info:
+            fee_threshold = proof_of_space.fee_info.applied_fee_threshold
+
+            if fee_quality <= fee_threshold:
+                self.log.info(f"Fee threshold passed for challenge '{challenge_str}': {fee_quality}/{fee_threshold}")
+            else:
+                self.log.warning(f"Fee threshold failed for challenge '{challenge_str}': {fee_quality}/{fee_threshold}")
```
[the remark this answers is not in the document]
arvidn: I still think this message is obscure to the average user

arvidn (commented):
looks good!

harold-b:
*Should be squashed upon merge*

tests/farmer_harvester/test_third_party_harvesters.py
```diff
@@ -0,0 +1,441 @@
...
+    await time_out_assert(60, log_has_new_text, True)
+
+    find_message = "Fee threshold passed for challenge"
+    find_index = 0
+    log_text = caplog.text
+    fail_count = 0
+
+    for _ in range(fee_count):
+        index = log_text.find(find_message, find_index) + len(find_message)
+        if index < 0:
```
arvidn, line 359: it seems `index` is never < 0 in the test. is that OK?

tests/farmer_harvester/test_third_party_harvesters.py
```diff
@@ -0,0 +1,441 @@
...
+    mocker.patch.object(farmer.server.api, "new_proof_of_space", side_effect=intercept_farmer_new_proof_of_space)
+
+    log_text_len = 0
+
+    def log_has_new_text() -> bool:
+        nonlocal caplog
+        nonlocal log_text_len
+
+        text_len = len(caplog.text)
+        if text_len > log_text_len:
```
arvidn, line 320: this is always true, in the test. I take it that's expected. Maybe you could mark the else-case as `no-cover`

tests/farmer_harvester/test_third_party_harvesters.py
```diff
@@ -0,0 +1,441 @@
...
+                data = ChallengeChainSubSlot.from_bytes(src.data)
+                validated_sub_slot_cc = True
+            elif src.kind == uint8(SigningDataKind.REWARD_CHAIN_SUB_SLOT):
+                data = RewardChainSubSlot.from_bytes(src.data)
+                validated_sub_slot_rc = True
+            elif src.kind == uint8(SigningDataKind.PARTIAL):
+                # #NOTE: This data type is difficult to trigger, so it is
+                #        not tested for the time being.
+                # data = PostPartialPayload.from_bytes(src.data)
+                # validated_partial = True
```
arvidn, line 176: maybe you could add a word on why it's hard to test, and add a `no-cover` annotation

chia/farmer/farmer_api.py
```diff
@@ -235,11 +270,20 @@ async def new_proof_of_space(
 
                 # The plot key is 2/2 so we need the harvester's half of the signature
                 m_to_sign = payload.get_hash()
+                m_src_data: Optional[List[Optional[SignatureRequestSourceData]]] = None
+
+                if (
+                    new_proof_of_space.include_source_signature_data
+                    or new_proof_of_space.farmer_reward_address_override is not None
+                ):
+                    m_src_data = [SignatureRequestSourceData(uint8(SigningDataKind.PARTIAL), bytes(payload))]
```
arvidn, line 279: this is not covered by any test. I imagine this is also hard to trigger

chia/farmer/farmer.py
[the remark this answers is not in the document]
arvidn: also, it seems this is not exercised by a test. Do you have any failure-path tests? Could the existing test be parameterized to run a few different versions, including one where the harvester takes a reward it shouldn't?

madMAx43v3r:
`FoliageBlockData` can be 133 or 229 bytes (with pool signature)

I made a mistake previously, counting `G2Element` as 32 bytes, but it's actually 96 bytes.

chia/protocols/harvester_protocol.py
```diff
@@ -71,6 +102,8 @@ class RespondSignatures(Streamable):
     local_pk: G1Element
     farmer_pk: G1Element
     message_signatures: List[Tuple[bytes32, G2Element]]
+    include_source_signature_data: bool = False
+    farmer_reward_address_override: Optional[bytes32] = None
```
arvidn, line 106: I think you should update `tests/util/network_protocol_data.py` to make these fields not be `None` in the tests. That would increase the coverage of the protocol stability unit tests.

Not just these fields, but all new fields you're adding to protocol messages that are `Optional` and defaults to `None`

tests/farmer_harvester/test_third_party_harvesters.py
haorldbchi: Will fix

danieljperry:
@nossdpool @madMAx43v3r and other interested parties: As evidenced by the new commits, we're nearly ready to merge this PR. Feel free to comment here and/or leave a review in the CHIP.
https://github.com/Chia-Network/chips/pull/88

felixbrucker:
Please increment the protocol version

haorldbchi:
> Please increment the protocol version

Thanks for the reminder, have not forgotten, though.

felixbrucker:
For gigahorse this addition is required: https://github.com/felixbrucker/chia-blockchain/commit/0d4e9b1fddd6b949f7ae7fa87c4ad4306f5c6d8f
Max can probably better explain why, just letting you know already 👍

chia/farmer/farmer.py
```diff
@@ -835,3 +835,59 @@ async def _periodically_clear_cache_and_refresh_task(self) -> None:
...
+        ph_prefix = self.config["network_overrides"]["config"][self.config["selected_network"]]["address_prefix"]
+        farmer_reward_puzzle_hash = encode_puzzle_hash(proof_of_space.farmer_reward_address_override, ph_prefix)
+
+        self.log.info(
+            f"Farmer reward for challenge '{challenge_str}' "
+            + f"taken by harvester for reward address '{farmer_reward_puzzle_hash}'"
+        )
+
+        fee_quality = calculate_harvester_fee_quality(proof_of_space.proof.proof, sp.challenge_hash)
+        fee_quality_rate = int(float(fee_quality) / float(0xFFFFFFFF) * 100.0)
```
arvidn, line 860: this metric seems a bit odd to me. it's not really a percent. it's just the "quality", which could be mapped onto a scale of 0-100

chia/farmer/farmer.py
```diff
@@ -835,3 +835,59 @@ async def _periodically_clear_cache_and_refresh_task(self) -> None:
...
+
+        if proof_of_space.fee_info is not None:
+            fee_threshold = proof_of_space.fee_info.applied_fee_threshold
+            fee_threshold_rate = int(float(fee_threshold) / float(0xFFFFFFFF) * 100.0)
+
+            if fee_quality <= fee_threshold:
+                self.log.info(
+                    f"Fee threshold passed for challenge '{challenge_str}': "
+                    + f"{fee_quality_rate}%/{fee_threshold_rate}% ({fee_quality}/{fee_threshold})"
```
arvidn, line 869: I was expecting the `fee_threshold_rate` to be formatted as `0.2f`. Also below, on line 874

chia/farmer/farmer.py
harold-b: That is what the line is doing, mapping it to 0-100

harold-b: (Or did I misunderstand something?)

chia/farmer/farmer.py
harold-b: Will do. Probably even `.3f` as I just checked some of the rates go specify 3 decimal places

chia/farmer/farmer.py
arvidn: it doesn't really matter. it just seems odd to present this value as a fraction, in percent. But the more I think about it, the less odd it seems

haorldbchi:
> For gigahorse this addition is required: [felixbrucker@0d4e9b1](https://github.com/felixbrucker/chia-blockchain/commit/0d4e9b1fddd6b949f7ae7fa87c4ad4306f5c6d8f) Max can probably better explain why, just letting you know already 👍

There's a few more steps needed to get the data all the way to the Harvester. It would be good to add a new field here, since this field is not the source data of one of the signing hashes:
https://github.com/Chia-Network/chia-blockchain/pull/16717/files#diff-b8692b7c49cc660b8d9ee43584d492ca8c90b449d0da609435fdbdbe4e32e428L63

Then you need to also add it to the message that goes from the farmer to the harvester:
https://github.com/Chia-Network/chia-blockchain/pull/16717/files#diff-ba35af4d14747a4b954e3cabb328c726cbfb883d5a2c217afb64c0ea65e31b83L528

Are you sure this is what Max needs for GH? Would be good to be certain so that we know we're implementing what's necessary.

Also needs to be rebased again to the latest commit. If this get's merged, I can tackle updating tests, etc.

arvidn (dismissed):

arvidn (approved):

madMAx43v3r:
> Are you sure this is what Max needs for GH? Would be good to be certain so that we know we're implementing what's necessary.

Anyone who wants to implement this CHIP will need it, since it's needed to verify what reward address has been set in the foliage block. It needs to be passed together with `FoliageBlockData` which contains the `unfinished_reward_block_hash` to the 3rd party harvester.

harold-b:
OK sorry I missed it, I'll implement it

danieljperry:
@madMAx43v3r @nossdpool 
This PR is included in our latest Release Candidate build:
https://github.com/Chia-Network/chia-blockchain/releases/tag/2.2.0-rc1
Feel free to test its functionality in your pools.

Additional work is being completed in [PR #17435](https://github.com/Chia-Network/chia-blockchain/pull/17435), which we intend to merge soon.

danieljperry:
@madMAx43v3r @nossdpool The code for CHIP-22 has all been merged. 2.2.0-rc3 is the first build to contain the complete CHIP. Please test it with your pools.
https://github.com/Chia-Network/chia-blockchain/releases/tag/2.2.0-rc3

PR #16732, opened 2023-10-30
Added Contributors section

Sentious:
### Purpose:
Add a contributors section displaying all the contributors for the project

Screenshot preview:
image


wjblanke:
We are a bit worried about the external reference img. reopening for another look. seems fun

PR #16739, opened 2023-10-30
use await for `create_block_tools()`

altendky:


arvidn:
this seems like a legitimate failure still: https://github.com/Chia-Network/chia-blockchain/actions/runs/6699170130/job/18202823729?pr=16739#step:17:311

PR #16740, opened 2023-10-31
fix flaky dao tests

geoffwalmsley:
Updating the dao tests to use the newer methods in full_node_sim, removing dependency on farmed blocks. This will improve the efficiency of the tests and prevent flaking especially for the rpc client tests

Also found a problem in proposals with multple cat spends which is fixed in the first commit here


matt-o-how (approved):

wjblanke (approved):
aok

PR #16767, opened 2023-11-03
use separate protocol versions for the different services

arvidn:
### Purpose:

When updating the harvester protocol, we should bump its version number. Currently all our protocols use a single protocol version constant, which means we would also have to bump the version number of the full node- and wallet protocol as well (even though they haven't changed).

This patch separates the constant into one version number per protocol.

### DIFF
--- a/chia/server/ws_connection.py
+++ b/chia/server/ws_connection.py
@@ -223,6 +223,9 @@ class WSChiaConnection:
             if inbound_handshake.network_id != network_id:
                 raise ProtocolError(Err.INCOMPATIBLE_NETWORK_ID)
 
+            if inbound_handshake.protocol_version != protocol_version:
+                raise ProtocolError(Err.INCOMPATIBLE_PROTOCOL_VERSION)
+
             self.version = inbound_handshake.software_version
             self.protocol_version = Version(inbound_handshake.protocol_version)
             self.peer_server_port = inbound_handshake.server_port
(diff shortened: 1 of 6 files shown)


aqk (dismissed):

harold-b:
I might be mistaken, but I don't see a place where the protocol version is validated on either end of the connection.`ws_connection.py.perform_handshake()` seems like it would be the ideal location for this check

harold-b:
I pushed a commit to perform the check. Let me know if that works.
There was already an error message for it `INCOMPATIBLE_PROTOCOL_VERSION`, which did not seem to be used in the codebase.

chia/server/ws_connection.py
```diff
@@ -223,6 +223,9 @@ async def perform_handshake(
             if inbound_handshake.network_id != network_id:
                 raise ProtocolError(Err.INCOMPATIBLE_NETWORK_ID)
 
+            if inbound_handshake.protocol_version != protocol_version:
+                raise ProtocolError(Err.INCOMPATIBLE_PROTOCOL_VERSION)
```
arvidn, line 228: both the full node- and wallet protocols need to be backwards compatible, so I don't think it's safe to be this strict for all protocols.

For the full node- and wallet protocols, we use the version to know whether new messages are supported or not.

harold-b: OK understood. Is it fine if we enforce this then only for the Harvester and Farmer protocols?

harold-b:
Also, are handshakes legal bi-directionally? That is, can a `Harvester` initiate a connection to a `Farmer` and vice-versa or is it only ever uni-directional?

PR #16771, opened 2023-11-03
make sure we set max_sync_wait to 0 in all tests

arvidn:
### Purpose:

Some tests may still wait the full 30 seconds before starting sync.


almogdepaz (approved):

wjblanke (approved):
aok!

PR #16774, opened 2023-11-04
add test performing a deep reorg with 3 nodes

arvidn:
### Purpose:

The main point of this PR is to add a test exercising deep reorgs. It's using 3 nodes and two forks of the chain. the first node first reorgs to a fork, and then back to the original chain again.

This test exposes 2 issues:

1. There is a failure in correctly requesting blocks from the node whose chain we are re-orging onto. This issue is not addressed by this patch, but the failure is mitigated by allowing it to fall back to a sync-from-scratch.
2. The management of the block record cache clears blocks it shouldn't clear, especially during a reorg. It seems our logic to trigger pruning the cache makes an incorrect assumption that the chain we're validating is the main chain. We also validate blocks on the reorg chain, before it becomes the main chain.

The fix to (2) may not be the ideal fix. Currently, I made the clear cache function simply refuse to clear blocks close to the peak (of the main chain). So, even though we may tell it to clear more while validating a reorg chain, it won't. ~~Additionally, it won't clear blocks with sub-epoch summaries. We need those blocks~~ **EDIT: I increased the cache size for simulations instead**. Otherwise we fail like this:
```
22:14:00 chia.full_node.full_node: ERROR sync from fork point failed: KeyError: <bytes32: 65b34e3f8ca0be2905109f2b23b172e0af192730fa6f88307eb0273b1bd0379b>
  File "/home/arvid/dev/3/chia-blockchain/chia/util/log_exceptions.py", line 20, in log_exceptions
    yield
  File "/home/arvid/dev/3/chia-blockchain/chia/full_node/full_node.py", line 1163, in sync_from_fork_point
    await asyncio.gather(fetch_task, validate_task)
  File "/home/arvid/dev/3/chia-blockchain/chia/full_node/full_node.py", line 1125, in validate_block_batches
    success, state_change_summary, err = await self.add_block_batch(
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arvid/dev/3/chia-blockchain/chia/full_node/full_node.py", line 1320, in add_block_batch
    await self.weight_proof_handler.create_prev_sub_epoch_segments()
  File "/home/arvid/dev/3/chia-blockchain/chia/full_node/weight_proof.py", line 234, in create_prev_sub_epoch_segments
    prev_ses_sub_block = self.blockchain.height_to_block_record(heights[-3])
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arvid/dev/3/chia-blockchain/chia/consensus/blockchain.py", line 876, in height_to_block_record
    return self.block_record(header_hash)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arvid/dev/3/chia-blockchain/chia/consensus/blockchain.py", line 869, in block_record
    return self.__block_records[header_hash]
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: <bytes32: 65b34e3f8ca0be2905109f2b23b172e0af192730fa6f88307eb0273b1bd0379b>
```

Let's discuss whether we need this or whether there's a more elegant solution by updating the call-site.

### Current Behavior:

Deep reorgs fail.

### New Behavior:

Deep reorgs pass.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -735,6 +735,8 @@ class Blockchain(BlockchainInterface):
         Args:
             height: Minimum height that we need to keep in the cache
         """
+        if self._peak_height is not None and height > self._peak_height - self.constants.BLOCKS_CACHE_SIZE:
+            height = self._peak_height - self.constants.BLOCKS_CACHE_SIZE
         if height < 0:
             return None
         blocks_to_remove = self.__heights_in_cache.get(uint32(height), None)
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -538,7 +538,8 @@ class FullNode:
             )
             if first is None or not isinstance(first, full_node_protocol.RespondBlock):
                 self.sync_store.batch_syncing.remove(peer.peer_node_id)
-                raise ValueError(f"Error short batch syncing, could not fetch block at height {start_height}")
+                self.log.error(f"Error short batch syncing, could not fetch block at height {start_height}")
+                return False
             if not self.blockchain.contains_block(first.block.prev_header_hash):
                 self.log.info("Batch syncing stopped, this is a deep chain")
                 self.sync_store.batch_syncing.remove(peer.peer_node_id)
@@ -716,6 +717,10 @@ class FullNode:
                 return None
 
             if request.height < curr_peak_height + self.config["sync_blocks_behind_threshold"]:
+                # TODO: We get here if we encountered a heavier peak with a
+                # lower height than ours. We don't seem to handle this case
+                # right now. This ends up requesting the block at *our* peak
+                # height.
                 # This case of being behind but not by so much
                 if await self.short_sync_batch(peer, uint32(max(curr_peak_height - 6, 0)), request.height):
                     return None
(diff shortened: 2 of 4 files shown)


chia/consensus/blockchain.py
```diff
@@ -735,13 +735,19 @@ def clean_block_record(self, height: int) -> None:
         Args:
             height: Minimum height that we need to keep in the cache
         """
+        if self._peak_height is not None and height > self._peak_height - self.constants.BLOCKS_CACHE_SIZE:
+            height = self._peak_height - self.constants.BLOCKS_CACHE_SIZE
         if height < 0:
             return None
+        ses_heights = self.get_ses_heights()[-4:]
```
almogdepaz, line 742: your intention here is just to leave the records last blocks that correspond with the last four sub epoch summaries ?

that seems a bit weird since the cache under the peak is larger then 4 sub epochs and these should always be in the cache anyway 


is this just for tests ?

arvidn: fundamentally, the intention is to allow `create_prev_sub_epoch_segments()` to not fail with the `KeyError` that I mention in the description. These blocks are required to be in the cache for that, so, yes.

> that seems a bit weird since the cache under the peak is larger then 4 sub epochs and these should always be in the cache anyway

Are you sure? As you can see, I also make sure we keep `self.constants.BLOCKS_CACHE_SIZE` block records below the current peak. That was not enough to avoid the error.

> is this just for tests ?

It's possible that there's a quirk with the test constants that makes this problem only appear in tests.

almogdepaz: i dont like that we add production code specifically so tests would work :\
if you really insist i guess we can but i feel like a better approach is fixing the test quirk

arvidn: I added a commit to increase test block cache size in simulations, and remove the special case for blocks with sub-epoch-summaries

almogdepaz (approved):

wjblanke (approved):
aok

PR #16776, opened 2023-11-05
Fix peak height race

arvidn:
### Purpose:

When we add a block, there's a subtle co-routine suspension point in between updating the height-to-hash map and the peak height. There's a comment in the code to highlight this now.

This creates a window where a call to `get_peak()` fails because the peak height doesn't exist in height-to-hash map.

This is demonstrated in https://github.com/Chia-Network/chia-blockchain/pull/16774 .

```
  File "/home/arvid/dev/3/chia-blockchain/tests/core/full_node/test_full_node.py", line 2195, in check_nodes_in_sync
    p2 = full_node_1.full_node.blockchain.get_peak()
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arvid/dev/3/chia-blockchain/chia/consensus/blockchain.py", line 191, in get_peak
    return self.height_to_block_record(self._peak_height)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arvid/dev/3/chia-blockchain/chia/consensus/blockchain.py", line 697, in height_to_block_record
    if header_hash is None:
        ^^^^^^^^^^^^^^^^^^^^
ValueError: Height is not in blockchain: 1599
```

This patch moves the suspension point to make these updates to the height-to-hash map and `peak_height` atomic.
Then it's possible to remove the try-catch handler in the test.

### Current Behavior:

`get_peak()` may throw an exception if called at an unfortunate time.

### New Behavior:

`get_peak()` always work

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -271,9 +271,14 @@ class Blockchain(BlockchainInterface):
             block,
             None,
         )
-        # Always add the block to the database
-        async with self.block_store.db_wrapper.writer():
-            try:
+
+        # in case we fail and need to restore the blockchain state, remember the
+        # peak height
+        previous_peak_height = self._peak_height
+
+        try:
+            # Always add the block to the database
+            async with self.block_store.db_wrapper.writer():
                 header_hash: bytes32 = block.header_hash
                 # Perform the DB operations to update the state, and rollback if something goes wrong
                 await self.block_store.add_full_block(header_hash, block, block_record)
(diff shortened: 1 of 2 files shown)


chia/consensus/blockchain.py
```diff
@@ -271,9 +271,14 @@ async def add_block(
...
-        async with self.block_store.db_wrapper.writer():
-            try:
+
+        # in case we fail and need to restore the blockchain state, remember the
+        # peak height
+        previous_peak_height = self._peak_height
+
+        try:
+            # Always add the block to the database
+            async with self.block_store.db_wrapper.writer():
```
arvidn, line 281: the `try` was swapped with `async with`, that's why this looks a bit messy. Viewing this ignoring whitespace makes it a bit easier to see what happened.

almogdepaz (approved):

wjblanke (approved):
aok

PR #16781, opened 2023-11-06
Cache derivations in pytest runs, saving about 12% of test runtime

aqk:
### Purpose:

Make tests run faster.


chia/simulator/block_tools.py
```diff
@@ -292,6 +296,16 @@ def test_callback(event: PlotRefreshEvents, update_result: PlotRefreshResult) ->
             match_str=str(self.plot_dir.relative_to(DEFAULT_ROOT_PATH.parent)) if not automated_testing else None,
         )
 
+    def load_derivation_cache(self, path: Path) -> None:
+        try:
+            with path.open(encoding="UTF-8") as cache_file:
+                objects = json.load(cache_file)
+                for ob in objects:
+                    a, b, c, d = ob
+                    self.derivation_cache[(bytes.fromhex(a), b, c)] = bytes.fromhex(d)
```
arvidn, line 305: wouldn't it be simpler to use pickle here?

arvidn (commented):
it makes me a bit uneasy that the cache doesn't update by itself and that it's checked in. It seems like something we could keep in the `test-cache` repo instead, and I would expect it to have some check to ensure the master key matches the cache, so we at least are notified of the cache being outdated if we change the tests.

Also, it seems pretty intrusive. Would it not be possible to inject this at a lower level. It seems like a good use for a singleton.

prozacchiwawa:
A small comment i have is that at block_tools.py line 303, it'd be nice to name the members even if just for some future person.

chia/simulator/block_tools.py
aqk: Yes, definitely! I'll make that change.

aqk: Hmm, it might be better to have the cached derivations checked in as plaintext, though.

aqk:
> it makes me a bit uneasy that the cache doesn't update by itself and that it's checked in. It seems like something we could keep in the `test-cache` repo instead, and I would expect it to have some check to ensure the master key matches the cache, so we at least are notified of the cache being outdated if we change the tests.

I'll definitely add the master key check.

I agree, this data would be better living in the `test-cache` repo, but that does conflict a bit with the goal of auto updating.
Developers would have to have the `test-cache` repo checked out to add tests. That's ok if we have good git hook error messages.

I'll add auto-updating within this repo first, then we can move the data to `test-cache` if you think that's reasonable.

BTW, The way it works right now, the cache load can fail completely; it will fall back to the old behaviour and pass the tests.

> Also, it seems pretty intrusive. Would it not be possible to inject this at a lower level. It seems like a good use for a singleton.

Yes - Earle had a good idea - pre-populate the database with the equivalent entries. I think it's a good idea (trying it out now), but as this is a performance hack, I'm trying to keep this patch as small as possible, so it can be removed easily. Hopefully I can find a way to do it that doesn't touch all the functions in `derive_keys.py`

aqk:
Thank you both for the constructive comments; I'll ping you when it has all the updates.

aqk:
> A small comment i have is that at block_tools.py line 303, it'd be nice to name the members even if just for some future person.

Yes, that's better. Done.


> At line 202, how would some feel about using an environment variable (and disabling it if the env var isn't present) rather than specifying the name in block_tools, which is used by other stuff?

Good point. Come to think of it, there are still a bunch of other hardcoded things in `BlockTools.__init__()`


> At line 307: might want to silence this unless we really know something wants the key cache.
Ya, done.

> Since it's just a few methods (such as master_sk_to_farmer_sk etc), it might make the code nicer looking to have methods on this object with those names and call those:
> 
> Instead of:
> 
> ```
>         self.farmer_pk = master_sk_to_farmer_sk(self.farmer_master_sk, self.derivation_cache).get_g1()
> ```
> 
> Maybe?
> 
> ```
>         self.farmer_pk = self.master_sk_to_farmer_sk(self.farmer_master_sk).get_g1()
> ```

Yes, that does look nicer. Part of what I was going for here was to make it clear the only call sites using the cache were in `block_tools.py` - right now you can grep for `master_sk_to_farmer_sk` and be sure the cache hasn't leaked into other areas of the code.

BTW, I'm a little uncomfortable having BlockTools be under `chia.*` vs. `tests.*`. Let's talk more about that one.

> Otherwise neat overall imo, generally in favor of explicit DI for stuff like caches.

Thanks!

aqk:
> it makes me a bit uneasy that the cache doesn't update by itself and that it's checked in. It seems like something we could keep in the `test-cache` repo instead, and I would expect it to have some check to ensure the master key matches the cache, so we at least are notified of the cache being outdated if we change the tests.

Another issue with keeping this is the `test-cache` repo is that we don't get a guarantee that the cache is up to date. (Not an error, just a loos of performance). If I write the checked-in cache in place, it will be obvious that it should be added to any commit that updates it.

emlowe:
Rather than change `wallet/derive_keys.py` to add an Optional value, does it make sense to add wrappers to `simulator/block_tools` and `simulator/wallet_tools`? This keeps all the changes to "inside" the `simulator`

I guess this is somewhat contrary to Art's "generally in favor of explicit DI for stuff like caches" - since it wouldn't be very explicit

prozacchiwawa:
Definitely in favor of this pr.  Apologies for letting the review process get old.

emlowe (dismissed):

arvidn (commented):
`tests/derivation_cache.pickle` is an empty file. that's a mistake, right?

It makes me uneasy to make such intrusive changes in the production code for this. Would it not be possible to check the cache before calling into these key derivation functions instead?
Alternatively, could you monkey patch the global functions to essentially wrap them in a check against a global cache, just for the tests?

PR #16786, opened 2023-11-07
DL: remove data from DB on unsubscribe

fchirica:


chia/data_layer/data_store.py
```diff
@@ -1443,6 +1446,64 @@ async def remove_subscriptions(self, tree_id: bytes32, urls: List[str]) -> None:
...
+                )
+
+                SELECT hash, left, right
+                FROM all_nodes
+                WHERE hash NOT IN (SELECT hash FROM ancestors WHERE tree_id != ?)
+                """,
+                (
+                    tree_id,
+                    tree_id,
+                ),
```
altendky, line 1467: Note the use of `:tree_id` and `{"tree_id": tree_id}` so we are using names to map things.  The same could be applied below for consistency.

```suggestion
                """
                WITH all_nodes AS (
                    SELECT a.hash, n.left, n.right
                    FROM ancestors AS a
                    JOIN node AS n ON a.hash = n.hash
                    WHERE a.tree_id = :tree_id
                )

                SELECT hash, left, right
                FROM all_nodes
                WHERE hash NOT IN (SELECT hash FROM ancestors WHERE tree_id != :tree_id)
                """,
                {"tree_id": tree_id},
```

chia/data_layer/data_store.py
```diff
@@ -1443,6 +1446,64 @@ async def remove_subscriptions(self, tree_id: bytes32, urls: List[str]) -> None:
...
+                    assert prev_left == left
+                    assert prev_right == right
+                    continue
+                to_delete[hash] = (left, right)
+                if left is not None:
+                    ref_counts[left] = ref_counts.get(left, 0) + 1
+                if right is not None:
+                    ref_counts[right] = ref_counts.get(right, 0) + 1
+
+        async with self.db_wrapper.writer() as writer:
```
altendky, line 1486: With this being a separate db transactions, doesn't this leave open a race opportunity?  I think just getting the writer to begin with would be good even though we start with just reading.

chia/data_layer/data_store.py
```diff
@@ -1203,6 +1203,9 @@ async def build_ancestor_table_for_latest_root(self, tree_id: bytes32) -> None:
                 if node.hash not in known_hashes:
                     await self._insert_ancestor_table(node.left_hash, node.right_hash, tree_id, root.generation)
 
+            # Clean up intermediary hashes that are no longer needed, since every hash we need should be in ancestors.
+            await writer.execute("DELETE FROM node WHERE hash NOT IN (SELECT hash FROM ancestors)")
```
altendky, line 1207: Is this a scalable query?  I don't know how it gets executed but if we have lots of ancestors it might get big.  I'm wondering if there's some way to leverage foreign keys to tell SQLite what to delete maybe?

tests/core/data_layer/test_data_store.py
```diff
@@ -1433,3 +1433,129 @@ async def test_benchmark_batch_insert_speed(
...
+
+    await raw_data_store.insert_batch(tree_id_1, batch1, status=Status.COMMITTED)
+    await raw_data_store.insert_batch(tree_id_2, batch2, status=Status.COMMITTED)
+
+    await raw_data_store.delete_store_data(tree_id_1)
+    async with raw_data_store.db_wrapper.reader() as reader:
+        result = await reader.execute("SELECT * FROM node")
+        nodes = await result.fetchall()
+
+    keys = [node["key"] for node in nodes if node["key"] is not None]
```
altendky, line 1556: These collections could probably all be sets?  Then checks could be something like `assert keys.intersection(unique_keys_1) != set()`.  This would end up showing in the error all of the common keys.  I guess the second would use `.symmetric_difference()`

tests/core/data_layer/test_data_store.py
```diff
@@ -1433,3 +1433,129 @@ async def test_benchmark_batch_insert_speed(
...
+    common_keys = [key.to_bytes(4, byteorder="big") for key in range(common_keys_count)]
+    unique_keys_1 = [(key + 1000).to_bytes(4, byteorder="big") for key in range(500 - common_keys_count)]
+    unique_keys_2 = [(key + 2000).to_bytes(4, byteorder="big") for key in range(500 - common_keys_count)]
+
+    batch1 = []
+    batch2 = []
+    for key in common_keys + unique_keys_1:
+        batch1.append({"action": "insert", "key": key, "value": key})
+    for key in common_keys + unique_keys_2:
+        batch2.append({"action": "insert", "key": key, "value": key})
```
altendky, line 1546: (also maybe switch to a set as noted below)

```suggestion
    batch1 = [{"action": "insert", "key": key, "value": key} for key in common_keys + unique_keys_1]
    batch2 = [{"action": "insert", "key": key, "value": key} for key in common_keys + unique_keys_2]
```

tests/core/data_layer/test_data_store.py
```diff
@@ -1433,3 +1433,129 @@ async def test_benchmark_batch_insert_speed(
...
+async def test_delete_store_data_with_common_values(raw_data_store: DataStore, common_keys_count: int) -> None:
+    tree_id_1 = bytes32(b"\x00" * 31 + b"\x01")
+    tree_id_2 = bytes32(b"\x00" * 31 + b"\x02")
+
+    await raw_data_store.create_tree(tree_id=tree_id_1, status=Status.COMMITTED)
+    await raw_data_store.create_tree(tree_id=tree_id_2, status=Status.COMMITTED)
+
+    common_keys = [key.to_bytes(4, byteorder="big") for key in range(common_keys_count)]
+    unique_keys_1 = [(key + 1000).to_bytes(4, byteorder="big") for key in range(500 - common_keys_count)]
+    unique_keys_2 = [(key + 2000).to_bytes(4, byteorder="big") for key in range(500 - common_keys_count)]
```
altendky, line 1539: I think this expresses and enforces the offset intent while retaining the offset being constant across the parametrization?

```suggestion
    key_offset = 1000
    assert common_keys_count < key_offset
    common_keys = [key.to_bytes(4, byteorder="big") for key in range(common_keys_count)]
    unique_keys_1 = [(key + key_offset).to_bytes(4, byteorder="big") for key in range(500 - common_keys_count)]
    unique_keys_2 = [(key + (2 * key_offset)).to_bytes(4, byteorder="big") for key in range(500 - common_keys_count)]
```

tests/core/data_layer/test_data_store.py
```diff
@@ -1433,3 +1433,129 @@ async def test_benchmark_batch_insert_speed(
...
+    await raw_data_store.insert_batch(tree_id_1, batch1, status=Status.COMMITTED)
+    await raw_data_store.insert_batch(tree_id_2, batch2, status=Status.COMMITTED)
+
+    await raw_data_store.delete_store_data(tree_id_1)
+    async with raw_data_store.db_wrapper.reader() as reader:
+        result = await reader.execute("SELECT * FROM node")
+        nodes = await result.fetchall()
+
+    keys = [node["key"] for node in nodes if node["key"] is not None]
+    assert len(keys) == 500
```
altendky, line 1557: 500 should be set in a variable to relate all the uses of it.  Also could use an assertion that it is greater than the common keys count parametrization.

altendky (changes requested):
Thanks for implementing this.  Nothing stands out as 'broken' except for the dual reader/writer acquisition when deleting.  The rest is just cleanup I think.  Maybe apply the concepts from the comments on the last test to the other tests as well then I'll look them over in detail as well.

chia/data_layer/data_store.py
fchirica: I didn't get to test it for very large data sets (only what's included in the tests). but my reasoning is that this happens only once (when a new root gets confirmed on chain), so as long as it's not _completely_ terrible, we should still be fine given it's executed rarely.

fchirica: actually, now that I look at it, we might have to run it one more time at startup - to have backwards compatibility for stores created before this PR.

tests/core/data_layer/test_data_store.py
```diff
@@ -1433,3 +1433,137 @@ async def test_benchmark_batch_insert_speed(
...
+
+
+@pytest.mark.anyio
+async def test_delete_store_data_multiple_stores(raw_data_store: DataStore) -> None:
+    # Make sure inserting and deleting the same data works
+    for repetition in range(2):
+        num_stores = 50
+        total_keys = 150
+        keys_deleted_per_store = 3
+        tree_ids = [bytes32(bytes((0,) * 31 + (i,))) for i in range(num_stores)]
```
altendky, line 1500:
```suggestion
        tree_ids = [bytes32((0,) * 31 + (i,)) for i in range(num_stores)]
```

chia/data_layer/data_store.py
altendky: Ok, we can remember this anyways f we start hearing about relevant performance issues.  Feel free to resolve this conversation after reading.

tests/core/data_layer/test_data_store.py
```diff
@@ -1433,3 +1433,137 @@ async def test_benchmark_batch_insert_speed(
...
+    batch1 = [
+        {"action": "insert", "key": keys[0], "value": keys[0]},
+        {"action": "insert", "key": keys[1], "value": keys[1]},
+    ]
+    batch2 = batch1.copy()
+    batch1.append({"action": "insert", "key": keys[2], "value": keys[2]})
+    batch2.append({"action": "insert", "key": keys[3], "value": keys[3]})
+    await raw_data_store.insert_batch(tree_id, batch1, status=Status.COMMITTED)
+    await raw_data_store.insert_batch(tree_id_2, batch2, status=Status.COMMITTED)
+    assert batch1 != batch2
```
altendky, line 1455: Certainly a nitpick.  Slightly more clear though I think to have the assertion immediately adjacent to the variation.

```suggestion
    assert batch1 != batch2
    await raw_data_store.insert_batch(tree_id, batch1, status=Status.COMMITTED)
    await raw_data_store.insert_batch(tree_id_2, batch2, status=Status.COMMITTED)
```

tests/core/data_layer/test_data_store.py
```diff
@@ -1433,3 +1433,137 @@ async def test_benchmark_batch_insert_speed(
...
+    keys_values_before = await raw_data_store.get_keys_values(tree_id_2)
+    async with raw_data_store.db_wrapper.reader() as reader:
+        result = await reader.execute("SELECT * FROM node")
+        nodes = await result.fetchall()
+        kv_nodes_before = {}
+        for node in nodes:
+            if node["key"] is not None:
+                kv_nodes_before[node["key"]] = node["value"]
+    for i in range(total_keys):
+        assert kv_nodes_before[keys[i]] == keys[i]
```
altendky, line 1465: I think this is checking the same thing?

```suggestion
    assert [kv_nodes_before[key] for key in keys] == keys
```

altendky (changes requested):
Minor minor, then approval.  Thanks!

emlowe (approved):

altendky (approved):

PR #16793, opened 2023-11-08
update the `StateChangeSummary` to not keep full `NPCResult` objects

arvidn:
### Purpose:

This is a simplification of the state the `Blockchain` object returns in its `StateChangeSummary` class.
Currently, it includes the `NPCResult` object for the most recent block (and possibly all blocks along a reorg-chain).

The `NPCResult` is the conditions and created coins from each block. In the case of a reorg, we currently have to re-run all the blocks to produce this state, to be able to return it.

The only things we actually *need* out of the `NPCResult` object are additions and removals (and hints). This patch simplifies the `StateChangeSummary` class to only include the information we need.

The consumers of this information are the mempool (to determine which mempool items are no longer valid) and the hint store, to index hints and update subscriptions.

This is a step towards not having to re-run the blocks in the case of a reorg.

The main change is the members of `StateChangeSummary`, here: https://github.com/Chia-Network/chia-blockchain/pull/16793/files#diff-40409ea13958d9e8cec2be388c12e68da7c3179de55f01879713c5abdb1fed38R78-R81

### Current Behavior:

`StateChangeSummary` contains information that's expensive to compute.

### New Behavior:

`StateChangeSummary` contains information that's cheaper to compute.

## Profile

Profile of the cost of computing the `NPCResult` in `_reconsider_peak()`:

reorg-node-1-and-3

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -76,7 +76,10 @@ class StateChangeSummary:
     peak: BlockRecord
     fork_height: uint32
     rolled_back_records: List[CoinRecord]
-    new_npc_results: List[NPCResult]
+    # list of coin-id, puzzle-hash pairs
+    removals: List[Tuple[bytes32, bytes32]]
+    # new coin and hint
+    additions: List[Tuple[Coin, Optional[bytes]]]
     new_rewards: List[Coin]
 
 
@@ -539,7 +542,7 @@ class Blockchain(BlockchainInterface):
                 await self.block_store.set_in_chain([(block_record.header_hash,)])
                 await self.block_store.set_peak(block_record.header_hash)
                 return [block_record], StateChangeSummary(
-                    block_record, uint32(0), [], [], list(block.get_included_reward_coins())
+                    block_record, uint32(0), [], [], [], list(block.get_included_reward_coins())
                 )
             return [], None
 
@@ -569,7 +572,9 @@ class Blockchain(BlockchainInterface):
             curr = fetched_block_record.prev_hash
 
         records_to_add: List[BlockRecord] = []
-        npc_results: List[NPCResult] = []
+        # coin-id, puzzle-hash
+        removals: List[Tuple[bytes32, bytes32]] = []
+        additions: List[Tuple[Coin, Optional[bytes]]] = []
         reward_coins: List[Coin] = []
         for fetched_full_block, fetched_block_record in reversed(blocks_to_add):
             records_to_add.append(fetched_block_record)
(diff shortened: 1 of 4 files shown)


fchirica (dismissed):

chia/consensus/blockchain.py
```diff
@@ -76,7 +76,10 @@ class StateChangeSummary:
     peak: BlockRecord
     fork_height: uint32
     rolled_back_records: List[CoinRecord]
-    new_npc_results: List[NPCResult]
+    # list of coin-id, puzzle-hash pairs
+    removals: List[Tuple[bytes32, bytes32]]
+    # new coin and hint
+    additions: List[Tuple[Coin, Optional[bytes]]]
```
Quexington, line 82: A hint should be bytes32 yeah?

chia/full_node/full_node.py
```diff
@@ -1535,10 +1536,7 @@ async def peak_post_processing(
         )
 
         # Update the mempool (returns successful pending transactions added to the mempool)
-        spent_coins: Optional[List[bytes32]] = None
-        new_npc_results: List[NPCResult] = state_change_summary.new_npc_results
-        if len(new_npc_results) > 0 and new_npc_results[-1].conds is not None:
-            spent_coins = [bytes32(s.coin_id) for s in new_npc_results[-1].conds.spends]
+        spent_coins: List[bytes32] = [s[0] for s in state_change_summary.removals]
```
Quexington, line 1539: stylistic nitpick that I don't care about that much (feel free to resolve if you disagree) but maybe `id for id, _ in ...`

chia/full_node/hint_management.py
```diff
@@ -31,27 +29,23 @@ def add_if_ph_subscription(puzzle_hash: bytes32, coin_id: bytes32) -> None:
...
-
-                if len(hint) > 0:
-                    assert len(hint) <= 32
-                    hints_to_add.append((addition_coin_name, hint))
+        if len(hint) == 32:
+            add_if_ph_subscription(bytes32(hint), addition_coin_name)
+
+        if len(hint) > 0:
+            assert len(hint) <= 32
+            hints_to_add.append((addition_coin_name, hint))
```
Quexington, line 48: Things not exactly equal to length 32 are not considered hints right?

chia/consensus/blockchain.py
arvidn: actually, this is populated with any length bytes as long as it doesn't exceed 32 bytes

chia/full_node/hint_management.py
arvidn: they are, on the rust side: https://github.com/Chia-Network/chia_rs/blob/main/src/gen/conditions.rs#L346
since `bytes32` isn't available on the rust side, all are hinted as `bytes` anyway: https://github.com/Chia-Network/chia_rs/blob/main/wheel/chia_rs.pyi#L214

Quexington (approved):

emlowe (approved):

PR #16797, opened 2023-11-08
reduce redundant calls to compute the header hash

arvidn:
### Purpose:

In `multiprocess_validation`, we compute the header hash of `FullBlock`s multiple times. This is a quite expensive operation. This patch reduces the redundant calls to speed things up.

In profiling the setup in `test_long_reorg_nodes`, the time spent calling `header_hash` goes from 11.13% (163876 calls) to 3.47% (30379 calls).

### Current Behavior:

test-setup-2

### New Behavior:

test-setup-2

### DIFF
--- a/chia/consensus/multiprocess_validation.py
+++ b/chia/consensus/multiprocess_validation.py
@@ -201,6 +201,7 @@ async def pre_validate_blocks_multiprocessing(
             return [PreValidationResult(uint16(Err.INVALID_PREV_BLOCK_HASH.value), None, None, False)]
         curr = block_records.block_record(blocks[0].prev_header_hash)
         num_sub_slots_to_look_for = 3 if curr.overflow else 2
+        header_hash = curr.header_hash
         while (
             curr.sub_epoch_summary_included is None
             or num_blocks_seen < constants.NUMBER_OF_TIMESTAMPS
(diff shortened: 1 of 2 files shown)


hoffmang9:
Sexy.

chia/consensus/multiprocess_validation.py
```diff
@@ -209,14 +210,19 @@ async def pre_validate_blocks_multiprocessing(
             if curr.first_in_sub_slot:
                 assert curr.finished_challenge_slot_hashes is not None
                 num_sub_slots_found += len(curr.finished_challenge_slot_hashes)
-            recent_blocks[curr.header_hash] = curr
+            recent_blocks[header_hash] = curr
             if curr.is_transaction_block:
                 num_blocks_seen += 1
-            curr = block_records.block_record(curr.prev_hash)
-        recent_blocks[curr.header_hash] = curr
+            header_hash = curr.prev_hash
```
arvidn, line 216: once we follow the link to the previous block, we know the next header hash will be the same as `prev_hash`, so just save that instead of re-computing it every time

chia/consensus/multiprocess_validation.py
```diff
@@ -304,8 +310,9 @@ async def pre_validate_blocks_multiprocessing(
             curr_b: FullBlock = block
 
             while curr_b.prev_header_hash in block_dict:
+                header_hash = curr_b.prev_header_hash
```
harold-b, line 313: Only a minor consideration: Perhaps it would be clearer if the `header_hash` var in this section would be named `prev_header_hash` instead, and then used in all 3 subsequent reads (as opposed to alternated with `curr_b.prev_header_hash` in the following line:
```python
prev_header_hash = curr_b.prev_header_hash
curr_b = block_dict[prev_header_hash]
prev_blocks_dict[prev_header_hash] = curr_b
```

harold-b (approved):
Looks good

chia/consensus/multiprocess_validation.py
arvidn: I chose `header_hash` because it's really only the "previous" hash in the first two lines, and the "current" header hash on the 3rd line. `header_hash` is a bit neutral on which one it is

arvidn:
it's a bit odd that no test exercises the case where we add a block record to the cache that we remove after validation. I think it's a pretty safe change still

emlowe (approved):

PR #16805, opened 2023-11-09
Add a check to GitHub JWT step in Trigger Dev Docker Build flow

AmineKhaldi:

PR #16806, opened 2023-11-09
Add check to github jwt step in other workflows that call github-glue

Starttoaster:
### Purpose:
This fixes an issue in various jobs that call github-glue that run an unnecessary step on workflows ran from a fork. The step doesn't actually do anything from fork workflows because it doesn't have the necessary Actions permissions.

The JWT in question is used to authenticate to the github-glue service, but those job steps also get skipped if it fails this same secrets access check.

### Current Behavior:
This step fails on fork workflows.

### New Behavior:
This step is skipped on fork workflows.


hoffmang9:
Heads up on the dupe @AmineKhaldi https://github.com/Chia-Network/chia-blockchain/pull/16805

emlowe (approved):

AmineKhaldi:
> Heads up on the dupe @AmineKhaldi #16805

Thanks! Closing #16805 as this supersedes it.

PR #16807, opened 2023-11-10
Fix mypy types

arvidn:
### Purpose:

improve mypy coverage

### DIFF
--- a/mypy-exclusions.txt
+++ b/mypy-exclusions.txt
@@ -80,7 +80,6 @@ tests.core.mempool.test_mempool
 tests.core.server.test_dos
 tests.core.server.test_rate_limits
 tests.core.ssl.test_ssl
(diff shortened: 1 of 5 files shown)


arvidn:
@richardkiss do you know why mypy says `binutils.assemble` is an untyped call? It looks like we have type annotations for it now

richardkiss:
If the module isn't annotated with a `py.typed` file, the type hints are ignored. Seems dumb to me, but I suppose there's some reason. So we need a new version of clvm_tools with that file in the module, or type stubs for clvm_tools (which also seems dumb). See also https://peps.python.org/pep-0561/#packaging-type-information

AmineKhaldi (approved):

emlowe (approved):

PR #16819, opened 2023-11-13
remove special case logic for soft-fork 3 activation

arvidn:
### Purpose:

Apply the new rules to the entire chain, to simplify our code.

### Current Behavior:

Softfork-3 activates at height 4510000

### New Behavior:

Softfork-3 activates at height 0

### Test

[here](https://grafana.chiatechlab.com/d/oBgwkDI7z/blockchain-sync-tests?orgId=1&var-ref=enable-softfork3-unconditionally&var-repo=All&var-network=mainnet&var-network=testnet10&from=now-4d&to=now)

Main currently runs 6391 tests, with this patch we run 5227 tests (saving 1164 tests)

### DIFF
--- a/chia/clvm/spend_sim.py
+++ b/chia/clvm/spend_sim.py
@@ -68,7 +68,7 @@ class CostLogger:
             program,
             INFINITE_COST,
             mempool_mode=True,
-            height=DEFAULT_CONSTANTS.SOFT_FORK3_HEIGHT,
+            height=DEFAULT_CONSTANTS.HARD_FORK_HEIGHT,
             constants=DEFAULT_CONSTANTS,
         )
         self.cost_dict[descriptor] = npc_result.cost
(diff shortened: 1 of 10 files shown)


arvidn:
the missing test coverage is from a benchmark, which we don't instrument

aqk:
What do you think about leaving the enum entry for `SOFT_FORK3` at this line?

https://github.com/Chia-Network/chia-blockchain/pull/16819/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128L156

I know it's in git history; I just wonder if it's worth keeping a list of all the soft-forks somewhere visible.

aqk:
I wish I could reply to comments on this main page, since we're forced to use comments here for comments that won't block the PR.

aqk (approved):

emlowe (approved):

PR #16821, opened 2023-11-13
skip some bad fee cmdline tests and note these aren't currently run in CI

emlowe:
The tests in `tests/fee_estimation/cmdline_test.py` don't work if run manually due to several issues. These tests currently are not run in CI because the filename doesn't match the expected `test_*.py` pattern

For now, skipping these tests. This only impacts when you run the tests manually since as noted they aren't run in CI.


altendky (approved):
Note that this relates somewhat to https://github.com/Chia-Network/chia-blockchain/pull/16524.

emlowe:
close and reopen

PR #16826, opened 2023-11-13
BUG create_data_layer_service does not respect wallet_peer config

dkackman:
### Purpose:
Allow data_layer service to use a remote wallet. I don't think this is the only step as I haven't had success making it connect yet but this is an obvious oversight and a necessary first step. 

### Current Behavior:

The `start_data_layer` function uses the `self_hostname` config when configuring its wallet rpc client. 

### New Behavior:

The `start_data_layer.create_data_layer_service` function uses the `[data_layer][wallet_peer][host]` config when configuring its wallet rpc client. By default this is `localhost` so won't change anything in practice for current usage. The very next line uses [data_layer][wallet_peer][port] so it must be the intent and safe.

### Testing Notes:
- Works fine when `[data_layer][wallet_peer][host]` is `localhost` (the default)
- I am unable to connect to a remote wallet so there may be more to look at.

### DIFF
--- a/chia/server/start_data_layer.py
+++ b/chia/server/start_data_layer.py
@@ -42,7 +42,7 @@ def create_data_layer_service(
     if downloaders is None:
         downloaders = []
     service_config = config[SERVICE_NAME]
-    self_hostname = config["self_hostname"]
+    wallet_hostname = service_config["wallet_peer"]["host"]
     wallet_rpc_port = service_config["wallet_peer"]["port"]
     if wallet_service is None:
         wallet_root_path = root_path
@@ -50,7 +50,7 @@ def create_data_layer_service(
     else:
         wallet_root_path = wallet_service.root_path
         wallet_config = wallet_service.config
-    wallet_rpc_init = WalletRpcClient.create(self_hostname, uint16(wallet_rpc_port), wallet_root_path, wallet_config)
+    wallet_rpc_init = WalletRpcClient.create(wallet_hostname, uint16(wallet_rpc_port), wallet_root_path, wallet_config)
 
     data_layer = DataLayer.create(
         config=service_config,


emlowe:
DL calls the normally private wallet RPCs, so in order for this to work, you would have to 
a) expose the wallet RPCs (normally the listen on localhost and not publically)
b) do the "remote harvester" certificate dance to make sure the certs on both DL and wallet are generated from the same root

dkackman:
> DL calls the normally private wallet RPCs, so in order for this to work, you would have to
> 
> * expose the wallet RPCs (normally the listen on localhost and not publically)
> * do the "remote harvester" certificate dance to make sure the certs on both DL and wallet are generated from the same root

Understood. I haven't fully debugged connecting to a remote wallet peer, after including this code change, so it might be configuration on my side, but I think I have all the certs and ports lines up. I plan to do some more debugging this weekend. 

I am inferring from the fact that the `data_layer` config section has a `wallet_peer` entry that this was the intended design, even if it is a power user/advanced feature. If it is the intent, I don't think it is working right now (especially with the way the DL resolves the wallet client in the latest code). If it isn't the design intent, I'd suggest removing the `wallet_peer` config section altogether.

dkackman:
> DL calls the normally private wallet RPCs, so in order for this to work, you would have to
> 
> * expose the wallet RPCs (normally the listen on localhost and not publically)
> * do the "remote harvester" certificate dance to make sure the certs on both DL and wallet are generated from the same root

Understood. I am inferring from the fact that the `data_layer` config section has a `wallet_peer` entry that this was the intended design, even if it is a power user/advanced feature. If it is the intent, I don't think it is working right now (especially with the way the DL resolves the wallet client in the latest code). If it isn't the design intent, I'd suggest removing the `wallet_peer` config section altogether. 

I haven't fully debugged connecting to a remote wallet peer, after including this code change, so it might be configuration on my side, but I think I have all the certs and ports lines up. I plan on doing some more debugging this weekend.

emlowe:
This isn't wrong code per se, it makes code sense, however, given my earlier comment this is unlikely to benefit anyone as we wouldn't recommend to people that they expose their wallet RPCs as a general rule to outside localhost.

I'm going to close this PR - it could be argued the other way that DL should remove these config items and it shouldn't be configurable at all.

PR #16830, opened 2023-11-14
add measured sizes for plot check

wjblanke:
### Purpose:

Avoid flagging plots for bad file sizes

### Current Behavior:

Code uses estimates which are sometimes too low for large plots

### New Behavior:

Use measures sizes if larger than estimate

### Testing Notes:

Need to test with farm of varying plot k sizes. Should reduce the number of false positives but not flag additional plots


neurosis69:
Successfully tested this patch for my farm, all plots are being cached.

In the meantime I've also logged the prover size for a k39 plot, which is: 44367

emlowe (approved):

PR #16844, opened 2023-11-16
DL: reduce amount of tracebacks in DL.

fchirica:


altendky (changes requested):
Maybe make a context manager to deduplicate this change and have it raise out a `ClickException` instead of printing to achieve both the output and termination?

On the server side, I don't think the server's json response should be catering to a raw json dump being pretty on the cli side.  The cli should get the info and present it as we decide to.

emlowe:
I think we will go with https://github.com/Chia-Network/chia-blockchain/pull/17460. 
Closing this PR

PR #16845, opened 2023-11-16
Add support for generating and verifying DataLayer Proofs of Inclusions

emlowe:
Add in commands for `get_proof` and `verify_proof`. The proof is a proof of inclusion that a given key, value pair is in the specified datalayer store by chaining the Merkle hashes up to the published on-chain root hash

On verification, the verifier only needs to perform a single lookup of the on-chain root and doesn't need to have synced any of the data, or subscribed to the data store.

To keep the proofs smaller, only the `clvm` hash of the key and value are included in the proof and not the actual key or value. A `clvm` hash is just a sha256 hash of the data prepended with `0x01`. Datalayer uses CLVM hashes for ease of verification in CLVM, although for this specific use case, there is no on-chain validation happening.

A user can generate a proof for multiple k,v pairs in the same datastore

During verification, the user must pay attention to the `current_root` value in the returned JSON. If `current_root` is `True`, this data chains to the current published root, and so if you synced the data, you can be sure it would be there. However, if this value is `False`, the root has moved from the time the proof was generated. You cannot make any assumptions in this case about whether the data is in fact in the datastore or not since the root has changed, therefore the data *might have* changed. It is up to the caller to determine how to treat this case; one possible action would be to obtain a new proof.

Example proof:
```
{
    "proof": {
        "coin_id": "0xb4b534539c0b459225161c42cf9d5e70e8dea1c650ee668c289fb23a4c42ada0",
        "inner_puzzle_hash": "0x06159fb06beeab30e96bbe3d4f2bcc7376fbe0e73947a57a260ddac14aac7dc9",
        "store_proofs": {
            "proofs": [
                {
                    "key_clvm_hash": "0x6998f3a5de71a94f5d14b31ae0673d8d54b50c4e420310d969ae56ac8e11872c",
                    "layers": [
                        {
                            "combined_hash": "0xeef5dbcd18885d68ac5d16b05726c79b99297ddd50bdf22750d6e383e4e4ebbf",
                            "other_hash": "0x02021fce3c97de3b60afbaf8b95df4c5713889e6101b698d0383c0481929a14a",
                            "other_hash_side": 0
                        }
                    ],
                    "node_hash": "0x52716bc27d6c266d42f1fc59dc5ba2cf819587388b899d1c063a228b26fabbe0",
                    "value_clvm_hash": "0x43138432a85321ce4eb110af0ec388a4fabc1deb8e82242430579bd3e41abda2"
                },
                {
                    "key_clvm_hash": "0x2847213288f0988543a76512fab09684131809d98baa8b93c2cd8a03a0e31d41",
                    "layers": [
                        {
                            "combined_hash": "0xeef5dbcd18885d68ac5d16b05726c79b99297ddd50bdf22750d6e383e4e4ebbf",
                            "other_hash": "0x52716bc27d6c266d42f1fc59dc5ba2cf819587388b899d1c063a228b26fabbe0",
                            "other_hash_side": 1
                        }
                    ],
                    "node_hash": "0x02021fce3c97de3b60afbaf8b95df4c5713889e6101b698d0383c0481929a14a",
                    "value_clvm_hash": "0x0d32015d4a33eda5e78a527a17ce65310fe00e01151315b57504246e38858656"
                }
            ],
            "store_id": "0x12d89e41e763e6d10f7e5ed0c540eb5cd3bdd392466636c91bec1750873dfade"
        },
    }
    "success": true
}
```

Example verify response:
```
{
    "current_root": true,
    "success": true,
    "verified_clvm_hashes": {
        "inclusions": [
            {
                "key_clvm_hash": "0x6998f3a5de71a94f5d14b31ae0673d8d54b50c4e420310d969ae56ac8e11872c",
                "value_clvm_hash": "0x43138432a85321ce4eb110af0ec388a4fabc1deb8e82242430579bd3e41abda2"
            },
            {
                "key_clvm_hash": "0x2847213288f0988543a76512fab09684131809d98baa8b93c2cd8a03a0e31d41",
                "value_clvm_hash": "0x0d32015d4a33eda5e78a527a17ce65310fe00e01151315b57504246e38858656"
            }
        ],
        "store_id": "0x12d89e41e763e6d10f7e5ed0c540eb5cd3bdd392466636c91bec1750873dfade"
    }
}
```


altendky (changes requested):

chia/cmds/data.py
```diff
@@ -51,14 +51,15 @@ def create_changelist_option() -> Callable[[FC], FC]:
     )
 
 
-def create_key_option() -> Callable[[FC], FC]:
+def create_key_option(**kwargs: bool) -> Callable[[FC], FC]:
```
altendky, line 54: This encourages arbitrary variation which is counter to the purpose of this function.  Also, not everything else is `bool` I don't think.  I would rather see explicit parameters for the options that have been considered.  This also makes it clear here what the variations are when considering more.

As mentioned, perhaps the name should become plural when multiple is set.

chia/rpc/data_layer_rpc_api.py
```diff
@@ -458,3 +468,42 @@ async def clear_pending_roots(self, request: ClearPendingRootsRequest) -> ClearP
...
+                node_hash=pi.node_hash,
+                layers=[
+                    ProofLayer(
+                        other_hash_side=uint8(layer.other_hash_side),
+                        other_hash=layer.other_hash,
+                        combined_hash=layer.combined_hash,
+                    )
+                    for layer in pi.layers
+                ],
+            )
```
altendky, line 495: It feels like a helper classmethod on `HashOnlyProof` might be worthwhile.  It both avoids repetition of the conversion code and defines the expected hash pattern at the class.

tests/wallet/rpc/test_dl_wallet_rpc.py
```diff
@@ -222,3 +224,57 @@ async def is_singleton_generation(rpc_client: WalletRpcClient, lid: bytes32, gen
...
+    ) -> None:
+        [full_node_service], [wallet_service], bt = one_wallet_and_one_simulator_services
+        full_node_api = full_node_service._api
+        full_node_server = full_node_api.full_node.server
+        wallet_node = wallet_service._node
+
+        # Create fake proof
+        fakeproof = HashOnlyProof(
+            key_clvm_hash=std_hash(b"\1" + b"key"),
+            value_clvm_hash=std_hash(b"\1" + b"value"),
```
altendky, line 242: See above about a classmethod.  Regardless, this is still manually applying the hash pattern.

tests/core/data_layer/test_data_rpc.py
```diff
@@ -2328,3 +2349,292 @@ async def test_mirrors(
...
+        expected_value=value,
+        full_node_api=offer_setup.full_node_api,
+        data_layer=offer_setup.maker.data_layer,
+        store_id=offer_setup.maker.id,
+    )
+
+    root_changed = await offer_setup.taker.api.verify_proof(request=proof)
+    assert root_changed["current_root"] is False
+    # comparing everything except "current_root"
+    assert root_changed == {**verify, "current_root": root_changed["current_root"]}
```
altendky, line 2640: I guess this would be pretty similar.  It varies due to checking equality rather than identify of the `"current_root"` value.  Is it meaningful to have these checks separate?  Single asserts are general nice.

```suggestion
    assert root_changed == {**verify, "current_root": False}
```

altendky (changes requested):

tests/wallet/rpc/test_dl_wallet_rpc.py
emlowe: I did this specifically because I expect a user of this API set may have to "manually" verify hashes using plain SHA256 without the benefit of treehash and I wanted to test that you can verify things with just using SHA256 as long as you know the secret sauce of prepending 1

altendky: Thanks.  Maybe add an abbreviated version of that explanation as a comment?

chia/rpc/data_layer_rpc_client.py
```diff
@@ -131,3 +131,12 @@ async def clear_pending_roots(self, store_id: bytes32) -> Dict[str, Any]:
         request = ClearPendingRootsRequest(store_id=store_id)
         response = await self.fetch("clear_pending_roots", request.marshal())
         return response
+
+    async def get_proof(self, store_id: bytes32, keys: List[bytes]) -> Dict[str, Any]:
+        request: Dict[str, Any] = {"store_id": store_id.hex(), "keys": [key.hex() for key in keys]}
+        response = await self.fetch("get_proof", request)
+        return response
+
+    async def verify_proof(self, proof: Dict[str, Any]) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
emlowe: This matches many of the existing calls in here, but I see the argument.

tests/wallet/rpc/test_dl_wallet_rpc.py
emlowe: Oh, actually here this is just fake data - ok I misspoke, this could do whatever

altendky (dismissed):

arvidn (dismissed):

wjblanke (approved):
aok

PR #16849, opened 2023-11-16
vault puzzles

geoffwalmsley:
first crack at the vault puzzles:
- secp signing
- recovery spend path via BLS sigs


chia/wallet/puzzles/p2_delegated_secp.clsp
```diff
@@ -1,14 +1,15 @@
...
 ; this is the "standard puzzle" for spending coins with SECP keys (ie secure enclave)
 
-(mod (SECP_PK delegated_puzzle delegated_solution signature)
+(mod (SECP_PK delegated_puzzle delegated_solution signature coin_id genesis_challenge)
   (include *standard-cl-21*)
   (include condition_codes.clib)
   (include sha256tree.clib)
 
-  (if (secp256r1_verify SECP_PK (sha256 (sha256tree delegated_puzzle)) signature)
+  (if (secp256r1_verify SECP_PK (sha256 (sha256tree delegated_puzzle) coin_id genesis_challenge) signature)
```
Quexington, line 9: You've got to do `(ASSERT_MY_COIN_ID` or else this can still be replayed in future coins.  Also, you should curry the genesis challenge rather than pass it in as a solution value otherwise it is as pointless as the coin ID without the assertion.

chia/wallet/vault/vault_drivers.py
```diff
@@ -45,3 +46,8 @@ def get_vault_proof(merkle_tree: MerkleTree, puzzlehash: bytes32) -> Program:
     proof = merkle_tree.generate_proof(puzzlehash)
     vault_proof: Program = Program.to((proof[0], proof[1][0]))
     return vault_proof
+
+
+# SECP SIGNATURE
+def construct_secp_message(delegated_puzzlehash: bytes32, coin_id: bytes32) -> bytes:
+    return delegated_puzzlehash + coin_id + DEFAULT_CONSTANTS.GENESIS_CHALLENGE
```
Quexington, line 53: You shouldn't be using this default here.  At the very least make it a default kwarg, but I don't recommend that either.

chia/wallet/puzzles/p2_delegated_secp.clsp
Quexington: Also, to be perfectly aligned with AGG_SIG_ME, you don't need to hash this.  Just use `concat`.

Quexington (changes requested):

chia/wallet/puzzles/p2_delegated_secp.clsp
```diff
@@ -0,0 +1,14 @@
...
+(mod (SECP_PK delegated_puzzle delegated_solution signature)
+  (include *standard-cl-21*)
+  (include condition_codes.clib)
+  (include sha256tree.clib)
+
+  (if (secp256r1_verify SECP_PK (sha256 (sha256tree delegated_puzzle)) signature)
+      ; secp256r1_verify returns 0 if successful.
+      (x)  ; this doesn't actually run because secp256_verify will raise on failure
+      (a delegated_puzzle delegated_solution)
+  )
```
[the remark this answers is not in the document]
geoffwalmsley: We'd have to do `assert not` since (somewhat confusingly) the verify returns 0 for success and raises for failure. But I think having `assert not secp_verify` diminishes readability

chia/wallet/puzzles/p2_delegated_secp.clsp
geoffwalmsley: Thanks - I had taken out the coin_id assertion when something was failing and neglected to add it back. 
In terms of using `concat`, the function I'm using to generate the signature (`ecdsa.SigningKey.sign_deterministic`) seems to demand a hash of the input data. I'm not sure what the signing process is for the enclave, but can swap it out if there's another lib for this?

chia/wallet/vault/vault_drivers.py
geoffwalmsley: moved it out to the test

chia/wallet/puzzles/p2_delegated_secp.clsp
Quexington: That's weird, `sign_deterministic` doesn't let you sign a message, only its hash?

geoffwalmsley: `sign_deterministic` takes the raw message, and hashes it internally while creating the sig. So then the chialisp secp_verify needs to use the hash of the message to verify against the signature we're passing in.

Quexington: Do we know if this is a limitation of the algorithm?  It can only sign 32 bytes? If so, I'm fine with this, but if it's just the particular python library we're using, then I think we should defer to the cheaper chialisp (no hashing) and figure it out on the next layer up.

chia/wallet/puzzles/vault_p2_recovery.clsp
```diff
@@ -0,0 +1,68 @@
...
+; the escape puzzle sends funds back to the vault (my_puzzlehash)
+
+(mod
+  (
+    P2_1_OF_N_MOD_HASH
+    FINISH_RECOVERY_MOD_HASH
+    P2_SECP_PUZZLEHASH
+    BLS_PK
+    TIMELOCK
+    my_puzzlehash
```
Quexington, line 13: Just realized this parameter is unused

chia/wallet/puzzles/p2_delegated_or_hidden_secp.clsp
```diff
@@ -0,0 +1,19 @@
...
+  (if (= (sha256tree delegated_puzzle) HIDDEN_PUZZLE_HASH)
+      (a delegated_puzzle delegated_solution)
+      (if (secp256r1_verify SECP_PK (sha256 (sha256tree delegated_puzzle) coin_id GENESIS_CHALLENGE HIDDEN_PUZZLE_HASH) signature)
+          (x)  ; this doesn't actually run because secp256_verify will raise on failure
+          (c
+            (list ASSERT_MY_COIN_ID coin_id)
+            (a delegated_puzzle delegated_solution)
+          )
+      )
+  )
```
Quexington, line 18: I'm prett sure it will be cheaper here (in most cases at least) do make this a `defun` that you pass the tree hash into that way we don't have to repeat the calculation.

Quexington (approved):

wjblanke (approved):
aok

PR #16855, opened 2023-11-17
avoid redundant list->set and set->list conversions

arvidn:
### Purpose:

`get_included_reward_coins()` constructs a `set` to return. Most call sites then convert it back to a list (at least in tests). But even in the consensus code, we never take advantage of this being a set. We just iterate over it.

This patch makes `get_included_reward_coins()` just return the list of reward coins (which is a member of `FullBlock`), and removes conversions to `list` which are now redundant.

It also changes the parameter type of `CoinStore.new_block()` to take a list instead of a set.

### Current Behavior:

we create a (redundant) set every time we call `get_included_reward_coins()` and we mostly convert it back into a list immediately.

### New Behavior:

We avoid constructing a temporary set + list when we call `get_included_reward_coins()`

### DIFF
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -4,7 +4,7 @@ import dataclasses
 import logging
 import sqlite3
 import time
-from typing import Any, Dict, List, Optional, Set, Tuple
+from typing import Any, Collection, Dict, List, Optional, Set, Tuple
 
 import typing_extensions
 from aiosqlite import Cursor
@@ -81,8 +81,8 @@ class CoinStore:
         self,
         height: uint32,
         timestamp: uint64,
-        included_reward_coins: List[Coin],
-        tx_additions: List[Coin],
+        included_reward_coins: Collection[Coin],
+        tx_additions: Collection[Coin],
         tx_removals: List[bytes32],
     ) -> List[CoinRecord]:
         """
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -81,7 +81,7 @@ class CoinStore:
         self,
         height: uint32,
         timestamp: uint64,
-        included_reward_coins: Set[Coin],
+        included_reward_coins: List[Coin],
         tx_additions: List[Coin],
         tx_removals: List[bytes32],
     ) -> List[CoinRecord]:
--- a/chia/types/full_block.py
+++ b/chia/types/full_block.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import List, Optional, Set
+from typing import List, Optional
 
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.foliage import Foliage, FoliageTransactionBlock, TransactionsInfo
@@ -56,11 +56,11 @@ class FullBlock(Streamable):
     def is_transaction_block(self) -> bool:
         return self.foliage_transaction_block is not None
 
-    def get_included_reward_coins(self) -> Set[Coin]:
+    def get_included_reward_coins(self) -> List[Coin]:
         if not self.is_transaction_block():
-            return set()
+            return []
         assert self.transactions_info is not None
-        return set(self.transactions_info.reward_claims_incorporated)
+        return self.transactions_info.reward_claims_incorporated
 
     def is_fully_compactified(self) -> bool:
         for sub_slot in self.finished_sub_slots:
--- a/chia/util/generator_tools.py
+++ b/chia/util/generator_tools.py
@@ -16,7 +16,7 @@ from chia.util.ints import uint64
 def get_block_header(block: FullBlock, tx_addition_coins: List[Coin], removals_names: List[bytes32]) -> HeaderBlock:
     # Create filter
     byte_array_tx: List[bytearray] = []
-    addition_coins = tx_addition_coins + list(block.get_included_reward_coins())
+    addition_coins = tx_addition_coins + block.get_included_reward_coins()
     if block.is_transaction_block():
         for coin in addition_coins:
             byte_array_tx.append(bytearray(coin.puzzle_hash))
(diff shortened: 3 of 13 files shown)


chia/types/full_block.py
```diff
@@ -56,11 +56,11 @@ def header_hash(self) -> bytes32:
     def is_transaction_block(self) -> bool:
         return self.foliage_transaction_block is not None
 
-    def get_included_reward_coins(self) -> Set[Coin]:
+    def get_included_reward_coins(self) -> List[Coin]:
         if not self.is_transaction_block():
-            return set()
+            return []
         assert self.transactions_info is not None
-        return set(self.transactions_info.reward_claims_incorporated)
```
arvidn, line 63: this is the main change

chia/full_node/coin_store.py
```diff
@@ -81,7 +81,7 @@ async def new_block(
         self,
         height: uint32,
         timestamp: uint64,
-        included_reward_coins: Set[Coin],
+        included_reward_coins: List[Coin],
```
altendky, line 84:
```suggestion
        included_reward_coins: Collection[Coin],
```

chia/util/generator_tools.py
```diff
@@ -16,7 +16,7 @@
 def get_block_header(block: FullBlock, tx_addition_coins: List[Coin], removals_names: List[bytes32]) -> HeaderBlock:
     # Create filter
     byte_array_tx: List[bytearray] = []
-    addition_coins = tx_addition_coins + list(block.get_included_reward_coins())
+    addition_coins = tx_addition_coins + block.get_included_reward_coins()
```
altendky, line 19: This can be totally agnostic to the types it is combining.

if you want a list:
```suggestion
    addition_coins = [*tx_addition_coins, *block.get_included_reward_coins()]
```
or a set:
```suggestion
    addition_coins = {*tx_addition_coins, *block.get_included_reward_coins()}
```

tests/blockchain/test_blockchain.py
```diff
@@ -1816,7 +1816,7 @@ async def test_conditions(self, empty_blockchain, opcode, with_garbage, bt):
         wt: WalletTool = bt.get_pool_wallet_tool()
 
         tx1: SpendBundle = wt.generate_signed_transaction(
-            10, wt.get_new_puzzlehash(), list(blocks[-1].get_included_reward_coins())[0]
+            10, wt.get_new_puzzlehash(), blocks[-1].get_included_reward_coins()[0]
```
altendky, line 1819: This is less obviously a good answer, but I'll mention a couple ideas here and maybe we have a conversation.

This can get 'some' element from any iterable.

```suggestion
            10, wt.get_new_puzzlehash(), next(iter(blocks[-1].get_included_reward_coins()))
```

altendky (changes requested):

chia/util/generator_tools.py
arvidn: it turns out it's not necessary to combine these collections. and in fact, most of the time it's wasted work

tests/blockchain/test_blockchain.py
altendky: As in existing code before this PR depends on order in sets?  That can't be expected to be consistent.  Python hashing is seeded differently each run to avoid some attacks.  But maybe I'm misunderstanding.  I think that iterating a set twice for example would yield the same order each time.

tests/core/test_full_node_rpc.py
```diff
@@ -97,19 +97,19 @@ async def test1(two_nodes_sim_and_wallets_services, self_hostname, consensus_mod
 
         assert (await client.get_block_record_by_height(100)) is None
 
-        ph = list(blocks[-1].get_included_reward_coins())[0].puzzle_hash
+        ph = list(set(blocks[-1].get_included_reward_coins()))[0].puzzle_hash
```
arvidn, line 100: without the conversion to `set` and back to `list` in this test, it fails.

tests/blockchain/test_blockchain.py
arvidn: https://github.com/Chia-Network/chia-blockchain/pull/16855/files#r1398231751

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -105,7 +105,7 @@ async def test_basic_coin_store(db_version: int, softfork_height: uint32, bt: Bl
                 else:
                     tx_removals, tx_additions = [], []
 
-                assert block.get_included_reward_coins() == should_be_included_prev
+                assert set(block.get_included_reward_coins()) == should_be_included_prev
```
altendky, line 108: Do we want this to be order agnostic?

tests/core/test_full_node_rpc.py
altendky: It's like the included reward coins order is inconsistent but the coins hash consistently across Python invocations (not necessarily expected) and so the `set()` call is inducing order.  Exactly the opposite of how we generally view `set()`.

Let's at least get a comment on this.

```suggestion
        # TODO: Understand why the list(set()) is required to make this work and address it.  This shouldn't be needed.
        ph = list(set(blocks[-1].get_included_reward_coins()))[0].puzzle_hash
```

altendky (changes requested):

tests/core/full_node/stores/test_coin_store.py
arvidn: well, I don't think the intention of this test ever was to ensure it predicts the order of the reward coins in the list. Maybe we could make it do that to ensure it's consistently what we expect.

altendky (dismissed):
I'm not entirely on board with using lists for collections of coins, as opposed to sets, but I guess there's some order dependence of this particular data on chain so maybe...?

altendky (approved):

wjblanke (approved):
aok

PR #16859, opened 2023-11-17
Extract push logic from RPC API to `tx_endpoint`

Quexington:
All of the RPCs are currently responsible for pushing their own transactions if the user specifies that should happen.  This paradigm shifts the responsibility a layer higher.  So long as a `tx_endpoint` returns a `transactions` key, the wrapper can handle pushing the transaction.

This change is clearer if you hide whitespace


geoffwalmsley (approved):

wjblanke (approved):
aok

Issue #16817, opened 2023-11-13
[Bug] offer `confirmed_at_index` is always empty


KryptomineCH:
### What happened?

When checking the transaction history, the value of `confirmed_at_index` is always 0.

This leads to confusing replies by CLI and RPC:
```
chia wallet get_offers -id 0xff211040f9b946d1461d1631fcc0c812b7bc24735989cdf58598d9f6dd3b7bfe

Record with id: ff211040f9b946d1461d1631fcc0c812b7bc24735989cdf58598d9f6dd3b7bfe
---------------
Created at: 2023-10-13 23:22:42
Confirmed at: Not confirmed
Accepted at: 2023-10-13 23:22:42
Status: CONFIRMED
---------------
```

Relevant seems to be the status, but `confirmed_at_index` would be important / valuable information

Furthermore, the `accepted_at_time` is only beeing filled when `is_my_offer` = false. This might be a technical limitation.
 

### Version

2.1.1

### What platform are you using?

Linux

### What ui mode are you using?

CLI

wjblanke:
Quex where do we save this off in the DB?

PR #16860, opened 2023-11-17
Fix set_status accidental arg

Quexington:
A new kwarg got added and this broke

https://github.com/Chia-Network/chia-blockchain/issues/16817


wjblanke (approved):
aok!

PR #16863, opened 2023-11-17
Return TX records from offer creation endpoints

Quexington:
Even though the tx records that an offer generates can't be pushed to the network, it's worth including all of the information about their creation to endpoints that might need it.


geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16865, opened 2023-11-17
Add a marshaller for RPC APIs

Quexington:
The handling of RPC requests and responses is pretty scattered and prone to error right now.  This change proposes a way to specify concrete types for both the request and response of an RPC and have the parsing of dictionaries occur auto-magically due to a decorator.  Something similar exists for data layer RPCs but this proposal tries to focus heavily on the ergonomics of writing a new API and therefore handles more behavior behind the scenes and leans on the `Streamable` library for the business logic.

Writing a new RPC should now look like this:
```
<in some other file probably>
@streamable
@dataclass(frozen=True)
class MyNewRPCRequest(Streamable):
  class_var_1: SomeType
  
@streamable
@dataclass(frozen=True)
class MyNewRPCResponse(Streamable):
  class_var_2: SomeOtherType
  
<in *_rpc_api.py>
@marshal
async def new_rpc(self, request: MyNewRPCRequest) -> MyNewRPCResponse:
  ...
```


chia/rpc/util.py
```diff
@@ -2,18 +2,56 @@
...
 
+RpcEndpoint = Callable[..., Coroutine[Any, Any, Dict[str, Any]]]
+MarshallableRpcEndpoint = Callable[..., Coroutine[Any, Any, Streamable]]
+
+
+class RequestType(TypedDict):
+    pass
+
+
+_T_Streamable = TypeVar("_T_Streamable", bound="Streamable")
```
altendky, line 28:
```suggestion
_T_Streamable = TypeVar("_T_Streamable", bound=Streamable)
```

chia/rpc/util.py
```diff
@@ -2,18 +2,56 @@
...
+    return streamable(
+        dataclass(frozen=True)(type("_" + cls.__name__, (Streamable,), {"__annotations__": cls.__annotations__}))
+    )
+
+
+def marshall(func: MarshallableRpcEndpoint) -> RpcEndpoint:
+    hints = get_type_hints(func)
+    request_class: Type[RequestType] = hints["request"]
+
+    async def rpc_endpoint(self, request: Dict[str, Any], *args, **kwargs) -> Dict[str, Any]:
```
altendky, line 42: hints for args and kwargs.  maybe just object?

chia/rpc/util.py
```diff
@@ -2,18 +2,56 @@
...
+        response_obj: Streamable = await func(
+            self,
+            request_class(
+                get_streamable_from_request_type(request_class).from_json_dict(request).__dict__  # type: ignore
```
altendky, line 46: explain the ignore and make it specific by calling out the particular error.  maybe break the line down to isolate the not-quite-typed thing and avoid ignoring anything else accidentally.

altendky, line 46: is this basically for runtime type enforcement?  seems it should perhaps get remarshalled instead of just accessing `.__dict__`.

tests/core/test_rpc_util.py
```diff
@@ -0,0 +1,48 @@
...
+    foofoo: str
+    barbar: uint32
+    bat: bytes
+    bam: SubObject
+
+
+@streamable
+@dataclass(frozen=True)
+class TestResponseObject(Streamable):
+    qat: List[str]
```
altendky, line 29: i think a little more content here would be good to show it can handle bytes and nested types and whatever sensibly.  not 'everything', but more than just list and string which could as well be handled by a typed dict converted to json.

altendky (changes requested):

chia/rpc/util.py
Quexington: Ah, the reason I didn't is because addict recurses which I don't want

chia/rpc/util.py
Quexington: Deleted this, I wasn't using it correctly

chia/rpc/util.py
Quexington: hinted as such

chia/rpc/util.py
Quexington: I was able to get rid of the ignore actually it turns out I misunderstood what a typed dict was and I actually can't believe this ever worked

tests/core/test_rpc_util.py
altendky: Yeah, I wasn't looking to test streamable -> json, just enough to show that this does in fact use that path, kinda.  Thanks.

altendky: Feel free to close when you've read.

chia/rpc/util.py
```diff
@@ -2,18 +2,54 @@
...
 from chia.wallet.transaction_record import TransactionRecord
 from chia.wallet.util.tx_config import TXConfig, TXConfigLoader
 
 log = logging.getLogger(__name__)
 
+# TODO: consolidate this with chia.rpc.rpc_server.Endpoint
+# Not all endpoints only take a dictionary so that definition is imperfect
+# This definition is weaker than that one however because the arguments can be anything
+RpcEndpoint = Callable[..., Coroutine[Any, Any, Dict[str, Any]]]
+MarshallableRpcEndpoint = Callable[..., Coroutine[Any, Any, Streamable]]
```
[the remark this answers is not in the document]
Quexington: Probably, I'll check it out

chia/rpc/util.py
```diff
@@ -2,18 +2,41 @@
...
+# TODO: consolidate this with chia.rpc.rpc_server.Endpoint
+# Not all endpoints only take a dictionary so that definition is imperfect
+# This definition is weaker than that one however because the arguments can be anything
+RpcEndpoint = Callable[..., Awaitable[Dict[str, Any]]]
+MarshallableRpcEndpoint = Callable[..., Awaitable[Streamable]]
+
+
+def marshal(func: MarshallableRpcEndpoint) -> RpcEndpoint:
+    hints = get_type_hints(func)
+    request_class: Type[Streamable] = hints["request"]
```
altendky, line 27: Might be worth checking here that this is true?  Then we would get an import time exception for mistakes.

```suggestion
    request_hint = hints["request"]
    assert isinstance(request_hint, Streamable)
    request_class = request_hint
```

We _might_ also be able to get to a way to hint this requirement.  Maybe.

altendky (changes requested):

altendky (dismissed):

altendky (approved):

wjblanke (approved):
aok

PR #16866, opened 2023-11-17
Add systemd init files to CLI-only Linux packages

TheLastCicada:
### Purpose:
Add Systemd initialization files to the .deb and .rpm CLI-only installers for more friendly process management.  This specifically helps hobby and enterprise users running Chia on a headless Linux server, which a big improvement being the ability to easily start Chia services at boot.  This is currently a pain-point for the Carbon users attempting to install Chia and the carbon applications in their infrastructure.  

These bundled systemd files allow the user to pass in the Linux user that Chia should run as and assumes default ports and locations for the CHIA_ROOT directory.  This likely covers a very large percentage of users and those who need customization can modify their own systemd files.  By passing in the user to the systemd script, multiple versions of Chia could be run on a single system using different Linux user accounts for the CHIA_ROOT.  

### Current Behavior:
No bundled init files.

### New Behavior:
If installed with the CLI .rpm or .deb file, Linux admins will be able to start Chia services with

```
systemctl start chia-full-node@username
systemctl start chia-wallet@username
systemctl start chia-data-layer@username
```

These services will start at boot with the following commands:

```
systemctl enable chia-full-node@username
systemctl enable chia-wallet@username
systemctl enable chia-data-layer@username
```

### DIFF
--- a/build_scripts/assets/systemd/chia-crawler@.service
+++ b/build_scripts/assets/systemd/chia-crawler@.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Chia Crawler Service for %i
+Requires=chia-daemon@%i.service
+After=chia-daemon@%i.service
+
+[Service]
+Type=simple
+Environment=CHIA_ROOT=/home/%i/.chia/mainnet
+ExecStart=/opt/chia/start_crawler
+User=%i
+Group=%i
+LimitNOFILE=1048576
+LimitNPROC=1048576
+
+[Install]
+WantedBy=multi-user.target
--- a/build_scripts/assets/systemd/chia-daemon@.service
+++ b/build_scripts/assets/systemd/chia-daemon@.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Chia Daemon Service for %i
+StopWhenUnneeded=true
+
+[Service]
+Type=simple
+Environment=CHIA_ROOT=/home/%i/.chia/mainnet
+ExecStart=/opt/chia/daemon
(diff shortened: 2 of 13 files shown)


cmmarslender (dismissed):

cmmarslender (approved):

wjblanke (approved):
aok

TCB13:
I've a question about this PR.

Starting the entire thing required the user to enter a password to unlock the wallet, how will that be handled with this?

Thank you.

cmmarslender:
> I've a question about this PR.
> 
> Starting the entire thing required the user to enter a password to unlock the wallet, how will that be handled with this?
> 
> Thank you.

Currently, the systemd configs are most beneficial to those without password protected keys, or running services that aren't utilizing keys (full node). There are possibly some clever integrations you could come up with that notify you of a need to unlock your keys, but those are a bit out of scope for the current iteration here. 

What I personally do to work around this is and successfully farm on boot is to just leave the keys that my plots are tied to without a password, and have my payout addresses all set to pay to another set of keys that aren't online on the farming host.

TCB13:
@cmmarslender okay... but then this update once it gets released will probably break tons of setups. New systemd services will be installed and automatically try to start a full node / wallet and the process will stay in the background asking for password. Meanwhile a user might run the cli manually and have both instances opening the same database.

How are you planning to handle this?

cmmarslender:
> @cmmarslender ~okay... but then this update once it gets released will probably break tons of setups with a simple apt upgrade.~
> 
> ~New systemd services will be installed and automatically try to start a full node / wallet and the process will stay in the background asking for password. Meanwhile a user might run the cli manually and have both instances opening the same database.~
> 
> ~How are you planning to handle this?~
> 
> Okay I see, you're not enabling those by default.

Correct, not enabled by default. Just there as an option in case you would like to use them with your setup without having to manually create your own systemd configs.

PR #16868, opened 2023-11-18
remove singleton_store

geoffwalmsley:
There's a backwards compatibility issue with adding the singletons table to `required_tables` in `wallet_node.reset_sync_db`. The singletons table exists in the DB, but when users revert to an older release of chia-blockchain reset_sync_db raises an error since the singletons table doesn't exist in earlier releases.

This PR provides a temporary fix by removing the use of singleton_store in the DAO wallet and WSM. Once there's a fix for the required_tables issue we can revert these changes.


matt-o-how (approved):

wjblanke (approved):
aok

PR #16873, opened 2023-11-19
don't relay signage points that have aged more than 40 seconds

wjblanke:
### Purpose:

Avoid sending old signage points to other nodes

### Current Behavior:

forward all cached signage points for challenge

### New Behavior:

Ignore those we have had for more than 40 seconds

### DIFF
--- a/chia/full_node/full_node_store.py
+++ b/chia/full_node/full_node_store.py
@@ -799,7 +799,7 @@ class FullNodeStore:
             self.finished_sub_slots.append((ip_sub_slot, ip_sub_slot_sps, ip_sub_slot_total_iters))
 
         new_eos: Optional[EndOfSubSlotBundle] = None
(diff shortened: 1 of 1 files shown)


chia/full_node/full_node_store.py
```diff
@@ -218,7 +218,7 @@ def add_to_future_sp(self, signage_point: SignagePoint, index: uint8) -> None:
             return None
 
         self.future_cache_key_times[signage_point.rc_vdf.challenge] = int(time.time())
-        self.future_sp_cache[signage_point.rc_vdf.challenge].append((index, signage_point))
+        self.future_sp_cache[signage_point.rc_vdf.challenge].append((index, signage_point, int(time.time())))
```
arvidn, line 221: I think you should use `time.monotonic()` instead of `time.time()`.
Even though it only matters twice a year :)

PR #16874, opened 2023-11-19
cleanup assertions of `True` and `False`

altendky:
### Purpose:

just cleaning up

### DIFF
--- a/tests/wallet/test_singleton.py
+++ b/tests/wallet/test_singleton.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import pytest
 from clvm_tools import binutils
 
 from chia.types.announcement import Announcement
--- a/tests/wallet/test_singleton.py
+++ b/tests/wallet/test_singleton.py
@@ -50,12 +50,10 @@ def test_only_odd_coins():
             [],
         ]
     )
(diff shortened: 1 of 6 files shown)


arvidn (dismissed):

tests/wallet/test_singleton.py
```diff
@@ -50,12 +51,10 @@ def test_only_odd_coins():
             [],
         ]
     )
-    try:
+
+    with pytest.raises(Exception) as exception_info:
```
arvidn, line 55: instead of capturing the `exception_info` and checking it, you can add `match="clvm raise" to the call to `raises()`

tests/wallet/test_singleton.py
```diff
@@ -83,12 +79,11 @@ def test_only_one_odd_coin_created():
...
```
arvidn, line 83: and here too

arvidn (approved):

tests/wallet/test_singleton.py
altendky: I find this pattern with the values separate to be more explicit and less error prone than having the exception encode the 'name' and number to a string and using a regex to match that.

PR #16880, opened 2023-11-19
add `DBWrapper2.managed()`

altendky:
### Purpose:

Just adding a `DBWrapper2.managed()` class method to the db wrapper so it can easily be created directly for use with a context manager.  This means it can provide it's own cleanup code instead of needing the close call duplicated at every use.

There are a few remaining uses of the create/stop pattern that will need to be updated as well.

- https://github.com/Chia-Network/chia-blockchain/pull/16890
- https://github.com/Chia-Network/chia-blockchain/pull/16889
- and i think the wallet state manager which probably plays into https://github.com/Chia-Network/chia-blockchain/pull/16760

The anyio cancel scopes and shielding will be introduced to the team soon.  I am going ahead and including them myself both to get a start at filling them in and get some time thinking about them in actual code changes.

### Draft For:
- [x] [`# TODO: switch to log_file`](https://github.com/Chia-Network/chia-blockchain/pull/16880/files#diff-70ec43bc92826e3bab2e1ae6a4c59a21d96a369430ea600f94c903568d47841dR59) after https://github.com/Chia-Network/chia-blockchain/pull/16753

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -214,6 +214,7 @@ class FullNode:
         # These many respond_transaction tasks can be active at any point in time
         self._add_transaction_semaphore = asyncio.Semaphore(200)
 
+        sql_log_path: Optional[Path] = None
         with contextlib.ExitStack() as exit_stack:
             sql_log_file: Optional[TextIO] = None
             if self.config.get("log_sqlite_cmds", False):
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -22,6 +22,7 @@ from typing import (
     List,
     Optional,
     Set,
+    TextIO,
     Tuple,
     Union,
     cast,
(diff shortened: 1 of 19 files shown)


chia/full_node/full_node.py
```diff
@@ -214,160 +215,162 @@ async def manage(self) -> AsyncIterator[None]:
...
+            if self.config.get("log_sqlite_cmds", False):
+                sql_log_path = path_from_root(self.root_path, "log/sql.log")
+                self.log.info(f"logging SQL commands to {sql_log_path}")
+                sql_log_file = exit_stack.enter_context(sql_log_path.open("a", encoding="utf-8"))
+
+            # create the store (db) and full node instance
+            # TODO: is this standardized and thus able to be handled by DBWrapper2?
+            async with manage_connection(self.db_path, log_file=sql_log_file, name="version_check") as db_connection:
+                db_version = await lookup_db_version(db_connection)
+
```
arvidn, line 229: the exit stack will destruct here, and close the `sql_log_file`, right?
it's not obvious how this will work. By the time the DBWrapper logs to the file, won't it already be closed?

arvidn (approved):

chia/full_node/full_node.py
altendky: The log aspect of the interface isn't super great/consistent and we could think it through a bit.  Including this PR, the `manage_connection()` manager takes an open file.  Yes, that file will be closed at the end of this exit stack.  The `DBWrapper2.managed()` manager takes a path, not a file.  It then opens that file and uses it for all the connections.  So, we open the log file twice in this case.  Once when checking the db version, and again when actually setting up the wrapper.

arvidn: I see

PR #16881, opened 2023-11-20
refactor new_peak_timelord

almogdepaz:
### Purpose:

improving the flow that handles a new peak

### Current Behavior:
we would skip a new peak for the wrong condition in some edge cases
we would also  handle peaks we should skip

### New Behavior:
skip less heavy peaks even if we dont have unfinished blocks in the cache 

### Testing Notes:
added a test case for receiving a less heavy peak while we dont have unfinished blocks in the cache


fchirica (dismissed):

tests/blockchain/test_blockchain.py
```diff
@@ -3509,8 +3509,15 @@ async def test_reorg_new_ref(empty_blockchain, bt):
...
-        elif i < 20:
+        elif i < 19:
             expected = AddBlockResult.ADDED_AS_ORPHAN
+        elif i == 19:
+            # same height as peak decide by iterations
+            peak = b.get_peak()
+            if block.total_iters < peak.total_iters:
+                expected = AddBlockResult.NEW_PEAK
+            else:
+                expected = AddBlockResult.ADDED_AS_ORPHAN
```
fchirica, line 3520: Coverage says this code is not exercised, but I think we should. One way would be to parametrize the test with a boolean to decide in which case we're in (`NEW_PEAK` or `ORPHAN`), and depending on that, to change the order in which we add the chains (chain1-chain2 or chain2-chain1 in each test case, depending which chain has the heaviest 19th block).

chia/simulator/block_tools.py
```diff
@@ -1073,7 +1079,13 @@ def get_consecutive_blocks(
...
             # Handle overflows: No overflows on new epoch or sub-epoch
-            if new_sub_slot_iters is None and num_empty_slots_added >= skip_slots and not pending_ses:
+
+            if (
+                new_sub_slot_iters is None
+                and num_empty_slots_added >= skip_slots
+                and not pending_ses
+                and not skip_overflow
+            ):
                 for signage_point_index in range(
```
fchirica, line 1089: should we check `min_signage_point` here too? probably not relevant for this PR tests, but maybe for someone who wants to use this new feature for overflow blocks as well?

tests/timelord/test_new_peak.py
```diff
@@ -43,103 +46,344 @@ async def test_timelord_new_peak_basic(
...
+                await get_rc_prev(b1, block_1),
+            )
+            await timelord_api.new_unfinished_block_timelord(timelord_unf_block)
+
+            assert timelord_api.timelord.unfinished_blocks[-1].get_hash() == timelord_unf_block.get_hash()
+            new_peak = timelord_peak_from_block(b1, block_2)
+            assert timelord_unf_block.reward_chain_block.total_iters <= new_peak.reward_chain_block.total_iters
+            await timelord_api.new_peak_timelord(new_peak)
+
+            await time_out_assert(60, peak_new_peak_is_none, True, timelord_api)
```
fchirica, line 109: I would merge this assertion and the next one below, since the behavior is not intuitive at all. Something like this:

```
def last_state_matches_hash(timelord: TimelordAPI, expected_hash: bytes32) -> bool:
    if timelord.timelord.new_peak is not None:
        # `new_peak_timelord` was not processed yet.
        return False
    return timelord.timelord.last_state.peak.reward_chain_block.get_hash() == expected_hash
```

Similarly with the code below. Otherwise, add a comment explaining `last_state` is not calculated correctly until `new_peak` gets to `None`, because `_handle_new_peak` didn't execute yet.

tests/timelord/test_new_peak.py
```diff
@@ -43,103 +46,344 @@ async def test_timelord_new_peak_basic(
...
+
+            # make two new blocks on tip
+            blocks_1 = bt.get_consecutive_blocks(2, default_1000_blocks)
+            block_1 = blocks_1[-2]
+            block_2 = blocks_1[-1]
+            await _validate_and_add_block(b1, block_1)
+            await _validate_and_add_block(b1, block_2)
+
+            block_record = b1.block_record(block_2.header_hash)
+            sub_slot_iters, difficulty = get_next_sub_slot_iters_and_difficulty(
```
fchirica, line 90: I'd merge `get_next_sub_slot_iters_and_difficulty` and `timelord_protocol.NewUnfinishedBlockTimelord` into a single function (for example `get_unfinished_block_timelord`), taking the inputs `bt`, `b1`, `block_record` and `block_1` and returning `timelord_protocol.NewUnfinishedBlockTimelord`. I think this could simplify the test code, since we are repeating these complicated calculations 4 times across all tests.

tests/timelord/test_new_peak.py
```diff
@@ -43,103 +46,344 @@ async def test_timelord_new_peak_basic(
...
-                # # new unknown peak, weight less then curr peak
-                # blocks_2 = bt.get_consecutive_blocks(1, blocks)
-                # await _validate_and_add_block(b2, blocks_2[-1])
-                # await timelord_api.new_peak_timelord(timelord_peak_from_block(blocks_2[-1], b2, bt.constants))
-                # assert timelord_api.timelord.new_peak.reward_chain_block.height == blocks_1[-1].height
+                # new unknown peak, weight less then curr peak
+                blocks_2 = bt.get_consecutive_blocks(1, blocks)
+                await _validate_and_add_block(b2, blocks_2[-1])
+                await timelord_api.new_peak_timelord(timelord_peak_from_block(b2, blocks_2[-1]))
+                assert timelord_api.timelord.last_state.last_weight == blocks_1[-1].weight
```
fchirica, line 63: Since we use `last_state` here, shouldn't we do something similar to the other tests? (after `new_peak_timelord`, make sure `timelord.timelord.new_peak` is None)

fchirica (commented):
The logic looks good and the tests seem really strong. My comments are mostly stylistic for the tests, but otherwise this should be good to merge.

tests/timelord/test_new_peak.py
almogdepaz: get_next_sub_slot_iters_and_difficulty is not really related to NewUnfinishedBlockTimelord and is used all over the full_node without calling NewUnfinishedBlockTimelord so that wont make much sense

tests/timelord/test_new_peak.py
almogdepaz: we are checking last_state at the other tests as well, the weight is not what we want to assert in some of them

chia/simulator/block_tools.py
almogdepaz: yes ill add it that would also work for overflows

fchirica (dismissed):

emlowe (dismissed):

emlowe:
coverage exemption

Starttoaster:
Last CI run is stale, going to close and re-open to run again.

chia/simulator/block_tools.py
Starttoaster: @fchirica please check if this review comment is still relevant, and resolve it if not. Open threads block merge.

tests/blockchain/test_blockchain.py
Starttoaster: @fchirica please check if this review comment is still relevant, and resolve it if not. Open threads block merge.

Starttoaster:
Tests are failing with the current CI suite. Seems like it's not for flakes.

Starttoaster:
Please check the latest CI run. Some seemingly non-flakey failed jobs. As an aside, this PR was fairly clearly stale and imo should not have been approved without a new CI run. @almogdepaz @emlowe FYI

emlowe (approved):

PR #16882, opened 2023-11-20
Fix SimBlockRecord construction w.r.t. prev_transaction_block_hash

AmineKhaldi:


chia/clvm/spend_sim.py
```diff
@@ -109,14 +109,15 @@ class SimBlockRecord(Streamable):
...
         return cls(
             rci,
             height,
-            uint32(height - 1 if height > 0 else 0),
+            prev_transaction_block_height,
             timestamp,
             True,
             std_hash(height.stream_to_bytes()),
-            std_hash(std_hash(height.stream_to_bytes())),
+            std_hash(prev_transaction_block_height.stream_to_bytes()),
```
arvidn, line 120: so, previously these blocks would not form a valid chain, as the `prev_transaction_block_hash` would never point to a previously computed `header_hash`. yeah?

Did you encounter some issue leading to this fix?

arvidn (approved):

PR #16883, opened 2023-11-20
Wallet Signer Protocol

Quexington:
This PR introduces a couple of new things in service of a new protocol between a wallet implementation and a transaction signer like an hsms or mobile device:

* New `ClvmStreamable` parent class.  Contains all `Streamable` functionality with new methods `.as_program` and `.from_program`.  A child class inheriting from this parent class will also automatically become a dataclass (and automatically apply the `@streamable` wrapper).

* New `clvm_serialization_mode` context manager.  When code under this context manager attempts to access the streamable library, the bytes serialization is overridden to first convert the object to a `Program` and _then_ to bytes.  In addition, the JSON serialization is overridden so that instead of continuing to JSON serialize past `ClvmStreamable` objects, the serialization instead just terminates with the byte representation in place like so:

  _without context manager_
  ```
  {
    "some key on a regular streamable": {
      "key on clvm streamable": foo,
      "key 2 on clvm streamable": bar,
    }
  }
  ```
  _with context manager_
  ```
  {
    "some key on a regular streamable": "0xclvm_streamable_as_program_bytes"
  }
  ```

* Of course the actual protocol message definitions.  These should not be taken as official without a corresponding CHIP.


richardkiss:
I get that you're following the pattern that's been used with `Serializable`. However I don't think it's a good pattern. The `dataclass` decorator provides plenty enough metadata that we can build the serializers and deserializers from that, and it's much more standard, so requiring our serializable objects to inherit from a home-grown base class seems unnecessary. That's why I've not used this pattern in `chia_base` and the `clvm_serde` submodule in `hsms`.

The idea is, you define your data type with `dataclass`. Then you feed that class to code that auto-generates the serializer or deserializer. Rebinding that function into the original class becomes totally optional.

There's a popular project called [pydantic](https://docs.pydantic.dev/latest/) that does a lot of what we're doing. I haven't looked at it too closely, but I suppose I should to get a better indication of what pattern a popular project uses. Popular doesn't mean good of course, but they've definitely had to respond to a lot more use cases.

I guess what I'm saying is I need to think about this more but I'm not convinced continuing to follow the pattern we've used here is a fantastic idea. Definitely `Serializable` has some shortcomings.

Quexington:
> I get that you're following the pattern that's been used with `Serializable`. However I don't think it's a good pattern. The `dataclass` decorator provides plenty enough metadata that we can build the serializers and deserializers from that, and it's much more standard, so requiring our serializable objects to inherit from a home-grown base class seems unnecessary. That's why I've not used this pattern in `chia_base` and the `clvm_serde` submodule in `hsms`.
> 
> The idea is, you define your data type with `dataclass`. Then you feed that class to code that auto-generates the serializer or deserializer. Rebinding that function into the original class becomes totally optional.
> 
> There's a popular project called [pydantic](https://docs.pydantic.dev/latest/) that does a lot of what we're doing. I haven't looked at it too closely, but I suppose I should to get a better indication of what pattern a popular project uses. Popular doesn't mean good of course, but they've definitely had to respond to a lot more use cases.
> 
> I guess what I'm saying is I need to think about this more but I'm not convinced continuing to follow the pattern we've used here is a fantastic idea. Definitely `Serializable` has some shortcomings.

Sure, this is not an endorsement of `Streamable` but there is some need to extend it to fit into the current zen of how all of our wallet code currently works.  Instead of ripping out all of the infrastructure we have for streamable or pivoting to new infrastructure moving forward, it seemed prudent to me to instead add an extension that can be optionally activated with a context manager.  This minimizes impact on existing code and makes new code ergonomic to write without worrying about all of the underlying serialization logic.

> The `dataclass` decorator provides plenty enough metadata that we can build the serializers and deserializers from that, and it's much more standard, so requiring our serializable objects to inherit from a home-grown base class seems unnecessary

The base class does turn the child class into a dataclass.  I found a nifty way with metaclasses to make that the case.  Writing the following:
```
class SomeThing(ClvmStreamable):
  ...
```
is the same as writing the following:
```
@streamable
@dataclass(frozen=True)
class SomeThing(Streamable):
  ...
  
  def as_program(...):
    ...
    
  def from_program(...):
    ...
```

> The idea is, you define your data type with dataclass. Then you feed that class to code that auto-generates the serializer or deserializer. Rebinding that function into the original class becomes totally optional.

This is what the metaclass does behind the scenes: https://github.com/Chia-Network/chia-blockchain/pull/16883/files#diff-e4a332263bcd7419d078b7710b837926c9b08f1c5d4938f3463a72276ec76992R51-R57 .  Granted, I don't have a way to opt out of attaching the serialization functions, but I could, I just don't entirely see the point.

geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16895, opened 2023-11-20
Implement Wallet Signer Protocol in Chia Wallet

Quexington:
This PR implements the wallet signer protocol (CHIPs TBD) in the chia wallet.  It does not only implement the wallet role, it also implements the signer role for the spends that require signatures from BLS secret keys.


geoffwalmsley (dismissed):

wjblanke (approved):
aok

PR #16896, opened 2023-11-20
Remove signing from wallet

Quexington:
This change removes the default behavior of wallets to sign a transaction before returning it.  That responsibility has been moved to clients (usually the RPC handling it automatically in our case). As such, a lot of specific signing code has been deleted in favor of the more generic version introduced in #16895.


geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16897, opened 2023-11-20
Make WalletStateManager.private_key optional

Quexington:
This PR removes the need of any code to rely on `WalletStateManager.private_key`.  This has been possible since #16896 which was the main need for the private key in the wallet (besides finding non-observer addresses which is obviously impossible if you run without the private key now).

This does not however remove the _ability_ to have a private key in the wallet and operate as usual.


geoffwalmsley (approved):

wjblanke (approved):
woot

PR #16898, opened 2023-11-21
clvm mypie

altendky:
### Purpose:

https://github.com/Chia-Network/clvm/pull/125

### Draft For:
- [x] pin to a released clvm
- [x] figure out test failures
- [x] review `TODO: review`s


chia/rpc/wallet_rpc_api.py
```diff
@@ -2282,6 +2284,7 @@ async def did_find_lost_did(self, request: Dict[str, Any]) -> EndpointResult:
         if curried_args is None:
             return {"success": False, "error": "The coin is not a DID."}
         p2_puzzle, recovery_list_hash, num_verification, singleton_struct, metadata = curried_args
+        assert recovery_list_hash.atom is not None
```
Quexington, line 2287: Seems like below we're doing `.as_atom()` which I think is not `Optional` so do we forsure need this?

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -255,14 +254,18 @@ def get_metadata_and_phs(unft: UncurriedNFT, solution: SerializedProgram) -> Tup
...
-                # ignore duplicated create coin conditions
-                continue
-            memo = bytes32(condition.as_python()[-1][0])
-            puzhash_for_derivation = memo
-            log.debug("Got back puzhash from solution: %s", puzhash_for_derivation)
+        elif condition_code == 51:
+            atom = condition.rest().rest().first().atom
+            assert atom is not None
+
+            if int_from_bytes(atom) == 1:
```
Quexington, line 261: Could we do `as_int()` here, and that will throw a more helpful exception if necessary.

Quexington (commented):
I left a couple of comments to this effect, but I'm not sure I like/see the reason for the general trend here of separating stuff into `CastableType` definitions before calling `Program.to` on it.  In my mind, when I make a `Program`, I'm defining it all at once inside the call to `.to`.  If I could specify it as a clvm string I would absolutely do that, but the `.to` syntax is what we're working with.

Quexington (commented):
I left a couple of comments to this effect, but I'm not sure I like/see the reason for the general trend here of separating stuff into `CastableType` definitions before calling `Program.to` on it.  In my mind, when I make a `Program`, I'm defining it all at once inside the call to `.to`.  If I could specify it as a clvm string I would absolutely do that, but the `.to` syntax is what we're working with.

chia/_tests/core/data_layer/test_data_store.py
```diff
@@ -255,7 +255,7 @@ async def test_build_a_tree(
     example = await create_example(data_store, tree_id)
 
     await _debug_dump(db=data_store.db_wrapper, description="final")
-    actual = await data_store.get_tree_as_program(tree_id=tree_id)
+    actual = await data_store.get_tree_as_nodes(tree_id=tree_id)
```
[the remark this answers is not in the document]
altendky: In these tests I initially used `Program` ...  because...  seemed like it fit with our way.  And I guess early on I thought maybe we would treat the values as clvm etc.  But, we didn't go that route and nothing runtime actually gets this tree IIRC.  I decided to just remove this clvm dependency from DataLayer.  I think I could easily separate this into an isolated PR if you would like.

chia/_tests/wallet/nft_wallet/test_nft_lifecycle.py
```diff
@@ -318,9 +319,11 @@ async def test_default_transfer_program(cost_logger: CostLogger) -> None:
...
         )
 
-        expected_announcement_data = Program.to(
-            (FAKE_LAUNCHER_ID, [[ROYALTY_ADDRESS, 50, [ROYALTY_ADDRESS]]])
-        ).get_tree_hash()
+        expected_announcement_python: Tuple[bytes32, List[List[CastableType]]] = (
+            FAKE_LAUNCHER_ID,
+            [[ROYALTY_ADDRESS, 50, [ROYALTY_ADDRESS]]],
+        )
+        expected_announcement_data = Program.to(expected_announcement_python).get_tree_hash()
```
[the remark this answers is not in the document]
altendky:
```
chia/_tests/wallet/nft_wallet/test_nft_lifecycle.py:323: error: Argument 1 to "to" of "SExp" has incompatible type "tuple[bytes32, list[list[object]]]"; expected "CastableType"  [arg-type]
```
The issue is that we designed our Python representation of CLVM to be entirely untyped.  As opposed to having classes that clearly define the needed form of that CLVM for different uses in a human readable and computer readable form.  Specifically here we have `[ROYALTY_ADDRESS, 50` so mypy looks at this and identifies the common base class of `ROYALTY_ADDRESS` and `50` which is `object`, hence the inner `list[object]`.  While this data does satisfy the more strict hint `Tuple[bytes32, List[List[CastableType]]]`, mypy can't guess that for us so I had to separate a bunch of these out to deal with that.

Maybe as part of the typing work for CLVM we could grow some corresponding Python types to address this design error?

chia/_tests/core/mempool/test_mempool_manager.py
```diff
@@ -1844,7 +1845,7 @@ async def make_setup_and_coins(
         ),
     ],
 )
-async def test_mempool_timelocks(cond1: List[object], cond2: List[object], expected: Optional[Err]) -> None:
+async def test_mempool_timelocks(cond1: List[CastableType], cond2: List[CastableType], expected: Optional[Err]) -> None:
```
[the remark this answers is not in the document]
altendky: I don't know.  Do you want to explore rewriting all the unstructured Python basic types CLVM data in unstructured `Program` CLVM data?  I think that's what you are asking about.

chia/_tests/core/data_layer/test_data_store.py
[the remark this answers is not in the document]
Quexington: No its fine as I read more of the PR it became more clear why you were doing it :+1:

chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -255,14 +254,18 @@ def get_metadata_and_phs(unft: UncurriedNFT, solution: SerializedProgram) -> Tup
...
+                # destination puzhash
+                if puzhash_for_derivation is not None:
+                    # ignore duplicated create coin conditions
+                    continue
+                memo = bytes32(condition.as_python()[-1][0])
```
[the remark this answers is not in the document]
altendky:
```suggestion
                memo = bytes32(condition.at("rrrfrf").as_atom())
```

chia/_tests/core/mempool/test_mempool_manager.py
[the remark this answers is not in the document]
Quexington: Yeah I know it's not any stronger typing really but if the idea for the thing is that we're immediately going to call `Program.to` on it, we may as well just force the caller to do that, it makes more sense to me at least.

chia/wallet/nft_wallet/nft_puzzles.py
Quexington: Yeah we could also save an indentation layer by moving the inside if back to an `and` but I'll resolve and leave that optional for you.

chia/_tests/wallet/cat_wallet/test_offer_lifecycle.py
```diff
@@ -118,10 +117,10 @@ def generate_secure_bundle(
     offered_amount: uint64,
     tail_str: Optional[str] = None,
 ) -> SpendBundle:
-    announcement_assertions = [a.to_program() for a in announcements]
+    announcement_assertions: List[Any] = [a.to_program() for a in announcements]
     selected_coin_amount = sum([c.amount for c in selected_coins])
     non_primaries = [] if len(selected_coins) < 2 else selected_coins[1:]
-    inner_solution: List[CastableType] = [
+    inner_solution: List[List[Any]] = [
```
Quexington, line 123: mypy isn't smart enough to handle this as a `List[Program]` hint?

Quexington: (the announcement_assertions that is)

chia/types/blockchain_format/program.py
```diff
@@ -243,13 +241,13 @@ def _sexp_replace(
         if len(kwargs) > 1:
             raise ValueError("conflicting paths")
         # yes, the terminal case is hinted incorrectly for now
```
Quexington, line 243: this comment to delete also?

Quexington (commented):
Some nitpicks

chia/_tests/wallet/cat_wallet/test_offer_lifecycle.py
altendky: [`d79ac98` (#16898)](https://github.com/Chia-Network/chia-blockchain/pull/16898/commits/d79ac98c017af255c54e9e2af85e0d979bde3f19) And that pointed out an error in the `inner_solution` hint.

chia/types/blockchain_format/program.py
altendky: Are these properly hinted?  `kwargs` is hinted as `Any` not `Program` so there's nothing saying that this return path is in fact going to match the function return hint.  I could re-explore why we simplified this to `T_Program` instead of the more complicated type in the comment above the `def`, but I don't recall offhand.

altendky:
```suggestion
```

Quexington (dismissed):

Quexington (dismissed):

emlowe (dismissed):

emlowe (approved):

PR #16903, opened 2023-11-21
Update key functions to handle observer only functionality

Quexington:
This PR introduces the finishing touches on a pure "observer" mode.  This pipes in the ability to add public key only entries in the keyring to the CLI so users can add just public information and still sync their wallet and view balances etc.  It also updates the wallet signing function to work without the master private key.  This is crucial infrastructure for hypothetical multi-sig as there are operations that a final aggregator needs to take even if they are not an owner of any of the relevant secret keys.


emlowe (dismissed):

chia/cmds/keys_funcs.py
```diff
@@ -59,29 +58,34 @@ def generate_and_add(label: Optional[str]) -> None:
...
     """
-    Add a private key seed to the keyring, with the given mnemonic and an optional label.
+    Add a private key seed or public key to the keyring, with the given mnemonic and an optional label.
     """
     unlock_keyring()
     try:
-        sk = Keychain().add_key(mnemonic, label)
-        fingerprint = sk.get_g1().get_fingerprint()
-        print(f"Added private key with public key fingerprint {fingerprint}")
+        if mnemonic_or_pk.count(" ") == 23:
```
geoffwalmsley, line 81: if someone has an extra `" "` then this could break. Maybe `len(mnemonic_or_pk.strip().split()) == 24`?

Quexington: Fixed

geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16909, opened 2023-11-21
undo BlockRecord cache insert, when DB fails

arvidn:
We currently have two caches on top of the block store database.

1. The `BlockStore` itself keeps an LRU of recently requested `FullBlock` objects.
2. The `Blockchain` class keeps *recent* `BlockRecords` in a cache (recent, as in close to the peak)

When we fail to insert a block in the DB, we undo the insertion into the block cache in `BlockStore`, but we currently don't undo the insertion of the block record into the block-record cache in Blockchain.

This PR fixes that.

Additionally, this PR rolls back any additions and removals from `ForkInfo`, if they have been added while validating the block.

### DIFF
--- a/chia/consensus/block_body_validation.py
+++ b/chia/consensus/block_body_validation.py
@@ -118,6 +118,13 @@ class ForkInfo:
             assert coin.name() not in self.additions_since_fork
             self.additions_since_fork[coin.name()] = ForkAdd(coin, uint32(block.height), uint64(timestamp), None, True)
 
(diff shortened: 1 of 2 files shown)


arvidn:
this original patch causes some tests to fail, so I will have to take a deeper look

wjblanke (dismissed):
aok

wjblanke:
earle had issues so holding off for now

wjblanke (approved):
aok

PR #16915, opened 2023-11-22
reorg optimization

arvidn:

The main change is in the last commit.

### Purpose:

In `_reconsider_peak()` we check if the new fork is heavier and if so, decide to switch our peak to this new chain. We also return a `StateChangeSummary` indicating to the caller what (if anything) has changed. e.g. any coin spends that were reorged-out and any new additions and removals (to update the mempool) and hints (to update the hint store).

In order to compute the additions and removals, we fetch all the full blocks and block records (again) along the fork and execute all the block generators (again) (`get_tx_removals_and_additions()`). This is wasteful, and especially risky with expensive block generators.

This patch updates the `ForkInfo` object to keep a little bit more state, in order to be able to produce the `StateChangeSummary` without having to re-run the block generators. This saves looking up `FullBlock`s from the database in addition to the CPU of executing blocks.

### Current Behavior:

When switching to a new fork of the chain, we re-run all the block generators along the fork in order to compute additions and removals, even though we already have most of that information now (in `ForkInfo`).

### New Behavior:

While validating blocks on the fork, store confirmation height of additions and removals as well as whether additions are coinbase rewards or not. This lets us build the `StateChangeSummary` object without re-running block generators, and without fetching `FullBlock`s from the database. We still need to fetch the `BlockRecord`s though.

### Testing Notes:

There's a sync-from-scratch test [here](https://grafana.chiatechlab.com/d/oBgwkDI7z/blockchain-sync-tests?orgId=1&var-ref=reorg-optimization&var-repo=All&var-network=mainnet&var-network=testnet10&var-network=testneta&refresh=1h).

Profiling this on an M1 did not give the speed-up I was expecting. It saves about 9% (time) in `test_long_reorg_nodes` test.

### timing M1

| test | before | after | after / before |
| --- | --- | --- | --- |
| reorg1 | 6.47s | 6.02s | 93% |
| reorg2 | 27.13s | 24.77s | 91.3% |

### timing RPi

| test | before | after | after / before |
| --- | --- | --- | --- |
| reorg1 | 61.84s | 58.27s  | 94.38% |
| reorg2 | 274.26s | 251.79s | 91.80% |

### profile (RPi)

It appears a large part of the time is spent waiting for database I/O (i.e. time spent in `kqueue`). Pay attention to the `kqueue` and `_reconsider_peak` timings of the following profiles:

| test | function | before | after |
| --- | --- | --- | --- |
| reorg 1 | _reconsider_peak | 6.51% | 1.72% |
| reorg 1 | add_block | 14.02% | 9.81% |
| reorg 2 | _reconsider_peak | 9.18% | 3.32% |
| reorg 2 | add_block | 17.31% | 12.26% |

### main

test-reorg-1

test-reorg-2

### with this patch

test-reorg-1

test-reorg-2

### DIFF
--- a/chia/consensus/block_body_validation.py
+++ b/chia/consensus/block_body_validation.py
@@ -63,6 +63,13 @@ class ForkInfo:
     # coin-id
     removals_since_fork: Set[bytes32] = field(default_factory=set)
 
+    def reset(self, fork_height: int, header_hash: bytes32) -> None:
+        self.fork_height = fork_height
+        self.peak_height = fork_height
+        self.peak_hash = header_hash
+        self.additions_since_fork = {}
+        self.removals_since_fork = set()
+
     def include_spends(self, npc_result: Optional[NPCResult], block: FullBlock) -> None:
         height = block.height
         if npc_result is not None:
--- a/chia/consensus/block_body_validation.py
+++ b/chia/consensus/block_body_validation.py
@@ -76,6 +76,10 @@ class ForkInfo:
     additions_since_fork: Dict[bytes32, ForkAdd] = field(default_factory=dict)
     # coin-id, ForkRem
     removals_since_fork: Dict[bytes32, ForkRem] = field(default_factory=dict)
+    # the header hashes of the blocks, starting with the one-past fork_height
+    # i.e. the header hash of fork_height + 1 is stored in block_hashes[0]
+    # followed by fork_height + 2, and so on.
+    block_hashes: List[bytes32] = field(default_factory=list)
 
     def reset(self, fork_height: int, header_hash: bytes32) -> None:
         self.fork_height = fork_height
@@ -83,10 +87,16 @@ class ForkInfo:
         self.peak_hash = header_hash
         self.additions_since_fork = {}
         self.removals_since_fork = {}
+        self.block_hashes = []
 
     def include_spends(self, npc_result: Optional[NPCResult], block: FullBlock, header_hash: bytes32) -> None:
         height = block.height
+
         assert self.peak_height == height - 1
+
+        assert len(self.block_hashes) == self.peak_height - self.fork_height
+        assert block.height == self.fork_height + 1 + len(self.block_hashes)
+        self.block_hashes.append(header_hash)
 
         self.peak_height = int(block.height)
         self.peak_hash = header_hash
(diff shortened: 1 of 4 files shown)


chia/consensus/blockchain.py
```diff
@@ -511,93 +508,70 @@ async def _reconsider_peak(
...
-                        block.height,
-                        block.foliage_transaction_block.timestamp,
-                        block.get_included_reward_coins(),
-                        tx_additions,
-                        tx_removals,
-                    )
-                await self.block_store.set_in_chain([(block_record.header_hash,)])
-                await self.block_store.set_peak(block_record.header_hash)
-                return [block_record], StateChangeSummary(
-                    block_record, uint32(0), [], [], [], block.get_included_reward_coins()
```
arvidn, line 545: this section is duplicated at the end of this function. As part of this patch, I simplified this function to only have this logic once. This is just handling the (special case) of adding the genesis block, where we don't have a `peak`

chia/consensus/blockchain.py
```diff
@@ -511,93 +508,70 @@ async def _reconsider_peak(
...
-        if block_record.prev_hash != peak.header_hash:
-            for coin_record in await self.coin_store.rollback_to_block(fork_info.fork_height):
-                rolled_back_state[coin_record.name] = coin_record
+        if peak is not None:
+            if block_record.weight <= peak.weight:
+                # This is not a heavier block than the heaviest we have seen, so we don't change the coin set
+                return [], None
 
-        # Collects all blocks from fork point to new peak
-        blocks_to_add: List[Tuple[FullBlock, BlockRecord]] = []
```
arvidn, line 559: since we no longer need to fetch `FullBlock`s from the database, this becomes identical to `records_to_add`, and we can just have a single list of `BlockRecord`s

chia/consensus/block_body_validation.py
```diff
@@ -44,6 +44,21 @@
 #           :
 
 
+@dataclass(frozen=True)
```
almogdepaz, line 47: should we move this to types ?

almogdepaz (approved):

chia/consensus/block_body_validation.py
arvidn: maybe. I'm not sure what reasoning we use to decide whether a type belongs in `chia.types` or not. `BlockRecord` isn't in types for instance, which is a bit surprising.

emlowe (approved):

PR #16917, opened 2023-11-22
sync caches to db after rollback

wjblanke:
## Purpose:

If we rollback due to disk I/O errors, caches may not be in sync with the on disk db state

## Current Behavior:

Caches still have records of rolled back blocks which causes problems later on

## New Behavior:

Make sure caches don't have blocks that are newer than the peak on disk


wjblanke:
Arvid has a different solution for this. PR #16909

PR #16918, opened 2023-11-22
Fix propagating spent coins IDs to the mempool on successful generator bundle creation in SpendSim's farm_block

AmineKhaldi:


chia/clvm/spend_sim.py
```diff
@@ -263,9 +264,10 @@ async def farm_block(
                     generator_bundle = bundle
                     return_additions = additions
                     return_removals = bundle.removals()
+                    spent_coins_ids = [r.name() for r in return_removals]
 
                     await self.coin_store._add_coin_records([self.new_coin_record(addition) for addition in additions])
-                    await self.coin_store._set_spent([r.name() for r in return_removals], uint32(self.block_height + 1))
+                    await self.coin_store._set_spent(spent_coins_ids, uint32(self.block_height + 1))
```
arvidn, line 270: I've struggled with the fact that the simulator doesn't use our normal code paths for running the chain. This was a particularly painful part, having to reach into an internal function of the coin store.

arvidn (approved):

chia/clvm/spend_sim.py
```diff
@@ -253,6 +253,7 @@ async def farm_block(
         generator_bundle: Optional[SpendBundle] = None
         return_additions: List[Coin] = []
         return_removals: List[Coin] = []
+        spent_coins_ids = None
```
arvidn, line 256: this line seems unnecessary. Is there any scenario where you access this variable before assigning a value to it?

AmineKhaldi: An example code path is when `peak` is `None`.

arvidn: I see

PR #16919, opened 2023-11-22
Singleton fast forward

AmineKhaldi:


chia/full_node/coin_store.py
```diff
@@ -495,3 +496,28 @@ async def _set_spent(self, coin_names: List[bytes32], index: uint32) -> None:
...
+                "SELECT unspent.coin_name, "
+                "unspent.coin_parent, "
+                "parent.coin_parent "
+                "FROM coin_record AS unspent "
+                "LEFT JOIN coin_record AS parent ON unspent.coin_parent = parent.coin_name "
+                "WHERE unspent.spent_index = 0 "
+                "AND parent.spent_index > 0 "
+                "AND unspent.puzzle_hash = ? "
+                "AND parent.puzzle_hash = unspent.puzzle_hash "
+                "AND parent.amount = unspent.amount",
```
arvidn, line 513: this rule was relaxed. the amounts no longer need to match

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -498,3 +502,137 @@ async def test_unsupported_version() -> None:
...
+        items=[
+            UnspentLineageIdsTestItem(TEST_COIN_ID, TEST_PUZZLEHASH, TEST_AMOUNT, TEST_PARENT_ID),
+            UnspentLineageIdsTestItem(b"2" * 32, b"2" * 32, 2, b"1" * 32),
+            UnspentLineageIdsTestItem(b"3" * 32, b"3" * 32, 3, b"2" * 32),
+            UnspentLineageIdsTestItem(TEST_PARENT_ID, b"4" * 32, TEST_AMOUNT, TEST_PARENT_PARENT_ID, is_spent=True),
+        ],
+        expected_success=False,
+    ),
+    UnspentLineageIdsCase(
+        id="Unspent with parent that has same puzzlehash but different amount",
```
arvidn, line 544: I suppose this test case is flipped too then

chia/types/eligible_coin_spends.py
```diff
@@ -97,12 +211,138 @@ def get_deduplication_info(
...
-                new_eligible_spends[coin_id] = DedupCoinSpend(current_solution, duplicate_cost)
+                new_dedup_spends[coin_id] = DedupCoinSpend(current_solution, duplicate_cost)
             cost_saving += duplicate_cost
         # Update the eligible coin spends data
-        self.eligible_spends.update(new_eligible_spends)
+        self.deduplication_spends.update(new_dedup_spends)
         return unique_coin_spends, uint64(cost_saving), unique_additions
+
+    async def process_fast_forward_spends(
+        self,
```
arvidn, line 229: the only use of `self` here is to get to `self.fast_forward_spends`. I believe it would be easier to unit test this function if you would make this a free function, and passing in the `fast_forward_spends` as a parameter instead

chia/full_node/mempool.py
```diff
@@ -389,6 +398,12 @@ def create_bundle_from_mempool_items(
             if not item_inclusion_filter(name):
                 continue
             try:
+                await eligible_coin_spends.process_fast_forward_spends(
```
arvidn, line 401: if this throws, that means we should drop `item` from the mempool, right?
or maybe we don't do that here

tests/core/mempool/test_mempool_manager.py
```diff
@@ -351,7 +351,7 @@ def make_bundle_spends_map_and_fee(
     spend_bundle: SpendBundle, npc_result: NPCResult
 ) -> Tuple[Dict[bytes32, BundleCoinSpend], uint64]:
     bundle_coin_spends: Dict[bytes32, BundleCoinSpend] = {}
-    eligibility_and_additions: Dict[bytes32, Tuple[bool, List[Coin]]] = {}
+    eligibility_and_additions: Dict[bytes32, Tuple[bool, List[Coin], bool]] = {}
```
arvidn, line 354: this `Tuple` might be complex enough to warrant a named type now. I'm especially worried about two unnamed `bool` fields. they can easily be mixed up

chia/types/eligible_coin_spends.py
```diff
@@ -97,12 +211,138 @@ def get_deduplication_info(
...
+                    coin_spend=new_coin_spend,
+                    eligible_for_dedup=spend_data.eligible_for_dedup,
+                    eligible_for_fast_forward=spend_data.eligible_for_fast_forward,
+                    additions=patched_additions,
+                    cost=spend_data.cost,
+                )
+                # Update the list of coins spends that will make the new fast
+                # forward spend bundle
+                new_coin_spends.append(new_coin_spend)
+                # We're done here, moving on
```
arvidn, line 296: this looks very similar to the section below, outside of this if-block. does the code end up more complex if you try to join this code with the one below?

tests/core/mempool/test_singleton_fast_forward.py
```diff
@@ -0,0 +1,486 @@
...
+    test_parent_id = bytes32.from_hexstr("0x039759eda861cd44c0af6c9501300f66fe4f5de144b8ae4fc4e8da35701f38ac")
+    test_ph = bytes32.from_hexstr("0x9ae0917f3ca301f934468ec60412904c0a88b232aeabf220c01ef53054e0281a")
+    test_amount = 1337
+    test_coin = Coin(test_parent_id, test_ph, test_amount)
+    test_child_coin = Coin(test_coin.name(), test_ph, test_amount)
+    latest_unspent_coin = Coin(test_child_coin.name(), test_ph, test_amount)
+    # This spend setup makes us eligible for fast forward so that we perform a
+    # meaningful fast forward on the rust side
+    test_puzzle_reveal = SerializedProgram.fromhex(
+        "ff02ffff01ff02ffff01ff02ffff03ffff18ff2fff3480ffff01ff04ffff04ff20ffff04ff2fff808080ffff04ffff02ff3effff04ff0"
```
arvidn, line 173: this program is pretty opaque. Do we have the source code to it? I think some comment may be warranted

arvidn (commented):
looks good so far!

chia/full_node/coin_store.py
AmineKhaldi: I sent a second commit that adapts to this, now that we relaxed this rule in chia_rs 0.3.3.

tests/core/full_node/stores/test_coin_store.py
AmineKhaldi: Yep, I sent an adaptation as part of the second commit.

chia/types/eligible_coin_spends.py
AmineKhaldi: `fast_forward_spends` only makes sense as an internal state to `EligibleCoinSpends` and current tests already show how easy it is to unit test that. This is consistent with `deduplication_spends`.

chia/full_node/mempool.py
AmineKhaldi: Yes, we do so when we can't proceed with the fast forward.

chia/types/eligible_coin_spends.py
AmineKhaldi: Yep, I originally explored different styles, but this is the most beneficial in terms of readability and maintainability.

tests/core/mempool/test_singleton_fast_forward.py
AmineKhaldi: Yep, I added a comment to mention that we leveraged the end to end test so we generate a realistic test case and use it here.

chia/types/eligible_coin_spends.py
```diff
@@ -27,9 +47,114 @@ class DedupCoinSpend:
...
+    )
+    new_solution = SerializedProgram.from_bytes(
+        fast_forward_singleton(spend=rust_coin_spend, new_coin=new_coin, new_parent=new_parent)
+    )
+    singleton_child = None
+    patched_additions = []
+    for addition in spend_data.additions:
+        patched_addition = Coin(unspent_lineage_info.coin_id, addition.puzzle_hash, addition.amount)
+        patched_additions.append(patched_addition)
+        if addition.puzzle_hash == spend_data.coin_spend.coin.puzzle_hash:
```
arvidn, line 137: I think you need to check that the `amount` is odd as well

chia/types/eligible_coin_spends.py
```diff
@@ -56,19 +181,19 @@ def get_deduplication_info(
         cost_saving = 0
         unique_coin_spends: List[CoinSpend] = []
         unique_additions: List[Coin] = []
-        new_eligible_spends: Dict[bytes32, DedupCoinSpend] = {}
+        new_dedup_spends: Dict[bytes32, DedupCoinSpend] = {}
```
arvidn, line 184: can you add a comment to indicate what the key is in this dictionary?

chia/types/eligible_coin_spends.py
```diff
@@ -97,12 +222,138 @@ def get_deduplication_info(
...
+        *,
+        mempool_item: InternalMempoolItem,
+        get_unspent_lineage_info_for_puzzle_hash: Callable[[bytes32], Awaitable[Optional[UnspentLineageInfo]]],
+        height: uint32,
+        constants: ConsensusConstants,
+    ) -> None:
+        """
+        Provides the caller with an in-place internal mempool item that has a
+        proper state of fast forwarded coin spends and additions starting from
+        the most recent unspent versions of the related singleton spends.
```
arvidn, line 250: it sounds like `mempool_item` is an in-out-parameter. It might be good to mention that

tests/core/mempool/test_mempool_manager.py
```diff
@@ -351,7 +358,7 @@ def make_bundle_spends_map_and_fee(
     spend_bundle: SpendBundle, npc_result: NPCResult
 ) -> Tuple[Dict[bytes32, BundleCoinSpend], uint64]:
     bundle_coin_spends: Dict[bytes32, BundleCoinSpend] = {}
```
arvidn, line 360: what is a `BundleCoinSpend`?

chia/types/eligible_coin_spends.py
```diff
@@ -97,12 +222,138 @@ def get_deduplication_info(
...
+            If a fast forward cannot proceed, to prevent potential double spends
+        """
+        new_coin_spends = []
+        ff_bundle_coin_spends = {}
+        for coin_id, spend_data in mempool_item.bundle_coin_spends.items():
+            if not spend_data.eligible_for_fast_forward:
+                # Nothing to do for this spend, moving on
+                new_coin_spends.append(spend_data.coin_spend)
+                continue
+            # See if we processed a fast forward item with this puzzle hash before
```
arvidn, line 269: I think this comment should explain why we need to do this. I believe we need to support finding the most recent version of a singleton both in the coin database *and* in the current state of the block we're building (because maybe we've already spent the singleton. Without this information it's hard to read the code following this.

chia/types/eligible_coin_spends.py
```diff
@@ -97,12 +222,138 @@ def get_deduplication_info(
...
+            unspent_lineage_info = self.fast_forward_spends.get(spend_data.coin_spend.coin.puzzle_hash)
+            if unspent_lineage_info is None:
+                # We didn't, so let's lookup the most recent version from the DB
+                unspent_lineage_info = await get_unspent_lineage_info_for_puzzle_hash(
+                    spend_data.coin_spend.coin.puzzle_hash
+                )
+                if unspent_lineage_info is None:
+                    raise ValueError("Cannot proceed with singleton spend fast forward.")
```
arvidn, line 277: can't we get here just by the singleton being spent for the first time? The first spend looks a bit different, with the eve-proof. It's not obvious that throwing is the right thing here. If it is, maybe it can be explained in a comment

chia/types/eligible_coin_spends.py
```diff
@@ -97,12 +222,138 @@ def get_deduplication_info(
...
+                    additions=patched_additions,
+                    cost=spend_data.cost,
+                )
+                # Update the list of coins spends that will make the new fast
+                # forward spend bundle
+                new_coin_spends.append(new_coin_spend)
+                # We're done here, moving on
+                continue
+            # We processed this puzzle hash before, so build on that
+            # See first if we're the most recent version
```
arvidn, line 310: I'd prefer using more concrete language. when you say "processed" you mean "added a spend of it" to the block we're building, right?

How would it be possible for our coin to be the most recent version?
I don't think you can submit a transaction to the mempool that spends the output of another transaction already in the pool.

chia/types/eligible_coin_spends.py
```diff
@@ -97,12 +222,138 @@ def get_deduplication_info(
...
+        generator = simple_solution_generator(new_sb)
+        new_npc_result = get_name_puzzle_conditions(
+            generator=generator,
+            max_cost=mempool_item.npc_result.cost,
+            mempool_mode=True,
+            height=height,
+            constants=constants,
+        )
+        if new_npc_result.error is not None:
+            raise ValueError("Mempool item became invalid after singleton fast forward.")
```
arvidn, line 355: does any exception here lead to eviction from the mempool?

chia/types/eligible_coin_spends.py
```diff
@@ -97,12 +222,138 @@ def get_deduplication_info(
...
+        # Update bundle_coin_spends using the collected data
+        mempool_item.bundle_coin_spends.update(ff_bundle_coin_spends)
+        # Update the mempool item with the new spend bundle related data
+        dataclasses.replace(mempool_item, spend_bundle=new_sb, npc_result=new_npc_result)
```
arvidn, line 359: `replace()` returns a new copy of the object with the replaced fields. It's *not* changing them in-place.
I take it there's no test to catch this

arvidn (dismissed):
If the exact same ff-eligible spend is added to the mempool twice, will the second one be rejected? If not, we may end up applying the same update many times, which would be redundant and expensive.

Perhaps they would trigger the identical-spend deduplication logic.

chia/types/eligible_coin_spends.py
```diff
@@ -27,9 +47,114 @@ class DedupCoinSpend:
...
+
+    Raises:
+        ValueError if none of the additions are considered to be the singleton's
+        next iteration
+    """
+    singleton_child = next(
+        (addition for addition in singleton_additions if addition.puzzle_hash == current_singleton.puzzle_hash), None
+    )
+    if singleton_child is None:
+        raise ValueError("Could not find fast forward child singleton.")
```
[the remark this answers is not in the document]
AmineKhaldi: We do these validations already on the rust side.

chia/types/eligible_coin_spends.py
```diff
@@ -27,9 +47,114 @@ class DedupCoinSpend:
...
+@dataclasses.dataclass(frozen=True)
+class UnspentLineageInfo:
+    coin_id: bytes32
+    coin_amount: int
+    parent_id: bytes32
+    parent_amount: int
+    parent_parent_id: bytes32
+
+
+def set_next_singleton_version(
```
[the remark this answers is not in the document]
AmineKhaldi: It's covered one layer up, through the `test_process_fast_forward_spends_*` tests.

chia/types/eligible_coin_spends.py
AmineKhaldi: We do these validations already on the rust side.

tests/core/mempool/test_mempool_manager.py
AmineKhaldi: It's a data structure that bridges between `CoinSpend` and `Spend` so that a `coin_id` can give us cheap access to both `coin_spend` and `additions` (as well as eligibility info). It was introduced with the deduplication feature.

chia/types/eligible_coin_spends.py
AmineKhaldi: The requirement indicated that we only consider post-eve singletons for this eligibility, so we should check for the eve case on the rust side and mark it as non eligible if so (if we're not doing that already).

chia/types/eligible_coin_spends.py
AmineKhaldi: Yes. That's visible upstream from here (`create_bundle_from_mempool_items`).

chia/types/eligible_coin_spends.py
AmineKhaldi: We just update those as good form, we don't need that new data in the next steps technically (we use spend bundle data right above it). I added a comment explaining this.

chia/full_node/mempool_manager.py
```diff
@@ -583,7 +621,9 @@ def check_removals(
         # 1. Checks if it's been spent already
         for record in removals.values():
             if record.spent:
-                return Err.DOUBLE_SPEND, []
+                # Only consider it a double spend if this is not a fast forward
+                if record.name not in fast_forward_coin_ids:
```
arvidn, line 625: do you remember why we don't need to handle duplicate spends here too? If you have two spends that can be deduplicated, presumably they would  be considered a double-spend under normal circumstances too

arvidn (commented):
mostly looks good. just some comments. The most important one being handling `skipped_items` >= `PRIORITY_TX_THRESHOLD`

chia/full_node/mempool.py
AmineKhaldi: That was the plan, but I didn't realize when I updated this that the mempool change landed into main already.

chia/full_node/mempool_manager.py
AmineKhaldi: Yes: we don't allow any spend (even if it turns out to be eligible for deduplication) to spend an already spent coin. If we get two similar spends, that's a conflict, and we do handle the deduplication case at that level (see code below this line). The exception now is these specific singleton spends, so now we allow them to technically be in a temporary double spend situation here.

chia/types/eligible_coin_spends.py
```diff
@@ -97,12 +219,145 @@ def get_deduplication_info(
...
+        for coin_id in replaced_coin_ids:
+            mempool_item.bundle_coin_spends.pop(coin_id, None)
+        mempool_item.bundle_coin_spends.update(ff_bundle_coin_spends)
+        # Update the mempool item with the new spend bundle related information
+        # NOTE: From this point on, in `create_bundle_from_mempool_items`, we rely
+        # on `bundle_coin_spends` and we don't use this updated spend bundle
+        # information, as we'll only need `aggregated_signature` which doesn't
+        # change. Still, it's good form to update the spend bundle with the
+        # new coin spends
+        mempool_item = dataclasses.replace(mempool_item, spend_bundle=new_sb, npc_result=new_npc_result)
```
arvidn, line 363: are you sure this works? my understanding of python is that `mempool_item` is a local variable that points to the object that was passed in. on this line, all you do is changing what the *local* variable `mempool_item` points to, you don't alter the object passed in.

AmineKhaldi: The code now explicitly mutates the internal mempool item.

arvidn:
missing test coverage:

```
chia/types/eligible_coin_spends.py (96.7%): Missing lines 75,137,352
tests/core/mempool/test_mempool.py (85.7%): Missing lines 2917
tests/core/mempool/test_mempool_manager.py (83.3%): Missing lines 972,1013
tests/core/mempool/test_singleton_fast_forward.py (98.7%): Missing lines 48,80,123
```

tests/core/mempool/test_mempool.py
```diff
@@ -2908,10 +2908,14 @@ def test_get_items_by_coin_ids(items: List[MempoolItem], coin_ids: List[bytes32]
...
 
 
-def test_aggregating_on_a_solution_then_a_more_cost_saving_one_appears() -> None:
+@pytest.mark.anyio
+async def test_aggregating_on_a_solution_then_a_more_cost_saving_one_appears() -> None:
     def always(_: bytes32) -> bool:
         return True
 
+    async def get_unspent_lineage_info_for_puzzle_hash(_: bytes32) -> Optional[UnspentLineageInfo]:
+        assert False
```
arvidn, line 2917: I think this deserves a `pragma: no cover` since it's part of the failing test path

tests/core/mempool/test_mempool_manager.py
```diff
@@ -951,6 +968,9 @@ async def get_coin_records(coin_ids: Collection[bytes32]) -> List[CoinRecord]:
 @pytest.mark.parametrize("reverse_tx_order", [True, False])
 @pytest.mark.anyio
 async def test_create_bundle_from_mempool(reverse_tx_order: bool) -> None:
+    async def get_unspent_lineage_info_for_puzzle_hash(_: bytes32) -> Optional[UnspentLineageInfo]:
+        assert False
```
arvidn, line 972: I think this also deserves a `pragma: no cover`

tests/core/mempool/test_mempool_manager.py
```diff
@@ -987,6 +1009,9 @@ async def send_spends_to_mempool(coin_spends: List[CoinSpend]) -> None:
 @pytest.mark.parametrize("num_skipped_items", [PRIORITY_TX_THRESHOLD, MAX_SKIPPED_ITEMS])
 @pytest.mark.anyio
 async def test_create_bundle_from_mempool_on_max_cost(num_skipped_items: int, caplog: pytest.LogCaptureFixture) -> None:
+    async def get_unspent_lineage_info_for_puzzle_hash(_: bytes32) -> Optional[UnspentLineageInfo]:
+        assert False
```
arvidn, line 1013: and this too, `pragma: no cover`

tests/core/mempool/test_singleton_fast_forward.py
```diff
@@ -0,0 +1,554 @@
...
+
+@pytest.mark.anyio
+async def test_process_fast_forward_spends_nothing_to_do() -> None:
+    """
+    This tests the case when we don't have an eligible coin, so there is
+    nothing to fast forward and the item remains unchanged
+    """
+
+    async def get_unspent_lineage_info_for_puzzle_hash(_: bytes32) -> Optional[UnspentLineageInfo]:
+        assert False
```
arvidn, line 48: this deserves a `pragma: no cover`

tests/core/mempool/test_singleton_fast_forward.py
```diff
@@ -0,0 +1,554 @@
...
+async def test_process_fast_forward_spends_unknown_ff() -> None:
+    """
+    This tests the case when we process for the first time but we are unable
+    to lookup the latest version from the DB
+    """
+
+    async def get_unspent_lineage_info_for_puzzle_hash(puzzle_hash: bytes32) -> Optional[UnspentLineageInfo]:
+        if puzzle_hash == IDENTITY_PUZZLE_HASH:
+            return None
+        assert False
```
arvidn, line 80: `pragma: no cover`

tests/core/mempool/test_singleton_fast_forward.py
```diff
@@ -0,0 +1,554 @@
...
+        coin_amount=test_coin.amount,
+        parent_id=test_coin.parent_coin_info,
+        parent_amount=test_coin.amount,
+        parent_parent_id=TEST_COIN_ID,
+    )
+
+    async def get_unspent_lineage_info_for_puzzle_hash(puzzle_hash: bytes32) -> Optional[UnspentLineageInfo]:
+        if puzzle_hash == IDENTITY_PUZZLE_HASH:
+            return test_unspent_lineage_info
+        assert False
```
arvidn, line 123: `pragma: no cover`

arvidn (approved):

PR #16926, opened 2023-11-26
`SerializedProgram` updates

arvidn:
### Purpose:

Makes `SerializedProgram` not default constructible. When constructing a `Streamable` type, we ensure all the members have the correct type in `__post_init__`, by calling each member's constructor with whatever value was passed in to the constructor of the aggregate type. Since `Program` is considered `Any`, mypy will not complain when constructing a `SerializedProgram` with a `Program`. adding an explicit `__init__` function that ensures it's only constructed by `bytes`, these kinds of type mismatches are caught.

The main causes of these mismatches are constructing `CoinSpend`, which has two `SerializedProgram` members, but are often constructed by `Program`. To address these cases, a new function, `make_spend()` is introduced which converts whatever arguments were passed in to `SerializedProgram`.

This is a step towards addressing all type mismatches involving `Program` and `SerializedProgram`.

### Current Behavior:

Some places construct an invalid `SerializedProgram` with a `Program` instead of a buffer.

### New Behavior:

These places now construct a `SerializedProgram` with a buffer, as is expected.

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -491,7 +491,7 @@ class DataLayerWallet:
             second_coin_spend = CoinSpend(
                 second_coin,
                 SerializedProgram.from_program(second_full_puz),
-                Program.to(
+                SerializedProgram.to(
                     [
                         LineageProof(
                             current_coin.parent_coin_info,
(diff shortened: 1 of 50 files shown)


arvidn:
apparently, these changes to `SerializedProgram` caused the `post_init` check in `Streamable` to fire type failures when constructing types with `Program` instead of `SerializedProgram`.. mypy sometimes works in mysterious ways.

So, this patch is snow balling into a much larger change, to also fix all the type issues.

arvidn:
as far as I can tell, we don't have any tests for `spend_prefarm`, hence the missing coverage.

altendky (approved):

emlowe (approved):

PR #16930, opened 2023-11-27
Type checked `Program`

arvidn:
### Purpose:

Improve type checking of `Program`.

Instead of `Program` deriving from `SExp` (which propagates the untyped nature of the type), contain an `SExp` object instead, and mimic the `SExp` interface.

### Current Behavior:

`Program` is treated as `Any` by mypy, making a lot of type errors not be reported.

### New Behavior:

`Program` is treated as a distinct type by mypy, making a lot of type errors reported and CI red (for now).


emlowe:
closing old PR

PR #16942, opened 2023-11-28
Revert "Refactor daemon simulation test (#16081)"

altendky:
This reverts commit 5ea14326c9b4a4bf1ca74ce910df07e4627f65a1.

### DIFF
--- a/chia/daemon/server.py
+++ b/chia/daemon/server.py
@@ -174,7 +174,7 @@ class WebSocketServer:
 
     @asynccontextmanager
     async def run(self) -> AsyncIterator[None]:
-        self.log.info(f"Starting Daemon Server ({self.self_hostname}:{self.daemon_port})")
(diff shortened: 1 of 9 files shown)


emlowe:
Seems reasonable to revert if causing issues - PR looks ok

arvidn:
What problems did this patch cause?
It fixed some issues (specifically, I could never run the daemon-simulation test locally). Was the new problem worse than the old one?

altendky:
This revert was diagnostic.  The 'fix' is over in https://github.com/Chia-Network/chia-blockchain/pull/16944.  From what I can tell, between 3.10.13 and 3.11.6 the shell call resulted in subprocesses that `.kill()` failed to kill.  exec doesn't have that issue and I would generally prefer it aside from this.  I commented about this in the test channel with links to changes in cpython that may be related.

PR #16944, opened 2023-11-28
`asyncio.create_subprocess_shell()` -> `asyncio.create_subprocess_exec()`

altendky:
### Purpose:

avoid a ci hang in 3.11 on linux only.

### DIFF
--- a/chia/timelord/timelord_launcher.py
+++ b/chia/timelord/timelord_launcher.py
@@ -88,8 +88,11 @@ async def spawn_process(
             dirname = path_to_vdf_client.parent
             basename = path_to_vdf_client.name
             resolved = await resolve(host, prefer_ipv6=prefer_ipv6)
-            proc = await asyncio.create_subprocess_shell(
-                f"{basename} {resolved} {port} {counter}",
+            proc = await asyncio.create_subprocess_exec(
+                os.fspath(basename),
+                str(resolved),
+                str(port),
+                str(counter),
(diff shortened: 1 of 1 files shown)


aqk:
I know dirname is where the binary is, but was dirname also the CWD of the previous code?

https://github.com/Chia-Network/chia-blockchain/pull/16944/files#diff-1054b958a0759914ed87f5b4bcdf6de32cecad90d3f380d79c1bad6ec4061f41R98

aqk (approved):

emlowe (approved):

chia/timelord/timelord_launcher.py
```diff
@@ -88,8 +88,11 @@ async def spawn_process(
             dirname = path_to_vdf_client.parent
             basename = path_to_vdf_client.name
             resolved = await resolve(host, prefer_ipv6=prefer_ipv6)
-            proc = await asyncio.create_subprocess_shell(
-                f"{basename} {resolved} {port} {counter}",
```
arvidn, line 92: yikes! you better hope none of those strings have spaces in them!

arvidn (approved):

arvidn:
feel free to resolve my comment. I got carried away, it should have been a non-blocking comment

chia/timelord/timelord_launcher.py
altendky: Quite so...  All the more reason to not shove a shell in the middle.

PR #16953, opened 2023-11-28
Broaden the `Keychain` definition of public keys to include more than BLS

Quexington:
This PR adds a new dimension to the keychain in which it can store more than one type of public information.  It converts most instances of `G1Element` to a new Protocol called `ObservationRoot` which must simply implement the ability to serialize to bytes and generate a fingerprint.  This is crucial for upcoming applications in which the public information used to sync a wallet is not a BLS public key but rather a launcher ID, or perhaps other key types like SECP, etc.

The private information is left constrained to BLS secret keys for now as it is less immediately useful.


chia/util/keychain.py
```diff
@@ -337,43 +360,58 @@ def _get_free_private_key_index(self) -> int:
...
     @overload
-    def add_key(self, mnemonic_or_pk: str, label: Optional[str], private: bool) -> Union[PrivateKey, G1Element]: ...
+    def add_key(
+        self, mnemonic_or_pk: str, label: Optional[str], private: bool
+    ) -> Tuple[Union[PrivateKey, ObservationRoot], KeyTypes]:
+        raise NotImplementedError()  # pragma: no cover
 
     def add_key(
         self, mnemonic_or_pk: str, label: Optional[str] = None, private: bool = True
-    ) -> Union[PrivateKey, G1Element]:
```
emlowe, line 357: I find it a little odd that add_key returns the type of key added, when I would have expected it to take the type of key as a parameter - doesn't the caller already know what it is asking to be added?

chia/util/keychain.py
```diff
@@ -337,43 +360,58 @@ def _get_free_private_key_index(self) -> int:
...
             assert isinstance(key, PrivateKey)
             pk = key.get_g1()
             key_data = bytes(pk).hex() + entropy.hex()
             fingerprint = pk.get_fingerprint()
         else:
             index = self._get_free_private_key_index()
             pk_bytes = hexstr_to_bytes(mnemonic_or_pk)
-            key = G1Element.from_bytes(pk_bytes)
-            assert isinstance(key, G1Element)
+            if len(pk_bytes) == 48:
```
emlowe, line 410: This kind of length check doesn't thrill me , I'm not sure it's even needed - I expect you could just call G1Element.from_byte in a try?

chia/util/keychain.py
```diff
@@ -337,43 +360,58 @@ def _get_free_private_key_index(self) -> int:
...
-    def add_key(self, mnemonic_or_pk: str, label: Optional[str], private: bool) -> Union[PrivateKey, G1Element]: ...
+    def add_key(
+        self, mnemonic_or_pk: str, label: Optional[str], private: bool
+    ) -> Tuple[Union[PrivateKey, ObservationRoot], KeyTypes]:
+        raise NotImplementedError()  # pragma: no cover
 
     def add_key(
         self, mnemonic_or_pk: str, label: Optional[str] = None, private: bool = True
-    ) -> Union[PrivateKey, G1Element]:
+    ) -> Tuple[Union[PrivateKey, ObservationRoot], KeyTypes]:
```
emlowe, line 390: Also not thrilled with this returning a tuple as it mucks up all the callers

chia/_tests/core/data_layer/test_data_rpc.py
```diff
@@ -2288,7 +2288,7 @@ async def test_wallet_log_in_changes_active_fingerprint(
 
     mnemonic = create_mnemonic()
     assert wallet_rpc_api.service.local_keychain is not None
-    private_key = wallet_rpc_api.service.local_keychain.add_key(mnemonic_or_pk=mnemonic)
+    private_key, _ = wallet_rpc_api.service.local_keychain.add_key(mnemonic_or_pk=mnemonic)
```
emlowe, line 2291: ugly tuple mucking.....

chia/simulator/block_tools.py
```diff
@@ -327,10 +327,14 @@ async def setup_keys(self, fingerprint: Optional[int] = None, reward_ph: Optiona
                 await keychain_proxy.delete_all_keys()
                 self.farmer_master_sk_entropy = std_hash(b"block_tools farmer key")  # both entropies are only used here
                 self.pool_master_sk_entropy = std_hash(b"block_tools pool key")
-                self.farmer_master_sk = await keychain_proxy.add_key(bytes_to_mnemonic(self.farmer_master_sk_entropy))
-                self.pool_master_sk = await keychain_proxy.add_key(
-                    bytes_to_mnemonic(self.pool_master_sk_entropy),
-                )
+                self.farmer_master_sk = (
+                    await keychain_proxy.add_key(bytes_to_mnemonic(self.farmer_master_sk_entropy))
+                )[0]
```
emlowe, line 332: ugly tuple mucking...

emlowe (changes requested):

chia/util/keychain.py
Quexington: As you can see, the thing passed into the `add_key` function is an assumed user-supplied string.  So the function as it is used is actually something more like `add_and_get_key` because you're taking the user input, saving it to the keychain, and bring it into a python type to further work with if so desired.

chia/util/keychain.py
Quexington: I was under the impression that using try/catch for control flow was bad practice.  And under the hood G1Element is going to do this anyways.  I'm not thrilled with the length check either but it seems to be the best way we have available to us with existing keyring format not having anticipated further types.

chia/_tests/core/data_layer/test_data_rpc.py
Quexington: Is this just pointing out that you think the function shouldn't return a tuple or would you rather `(...)[0]` or something?

emlowe: I'm not sure , but this change made everything that calls this now have to do `key, _ = ` - and rarely are you using the key type element from the tuple.
maybe it's a stepping stone to using a proper class here though

chia/util/keychain.py
```diff
@@ -219,31 +222,49 @@ def mnemonic_str(self) -> str:
         return " ".join(self.mnemonic)
 
 
+class KeyTypes(str, Enum):
+    G1_ELEMENT = "G1 Element"
```
emlowe, line 226: Do we want a type for private BLS key - although I guess we don't actually store the private key - it's the mnemonic. (although maybe we should also store the private key bits, I dunno)

So do we want a type for "mnemonic"?

chia/_tests/core/data_layer/test_data_rpc.py
Quexington: Most of the instances where we use the `, _` is in test code, there's three instances in production code and there could be an argument to use the types there.  Like "added key of type blah with fingerprint blah".  I'll wait until the discussion of whether or not this should be returning a tuple is resolved and then I will add that in if so.

emlowe (approved):

chia/_tests/core/data_layer/test_data_rpc.py
emlowe: alrighty

PR #16967, opened 2023-11-29
Allow public-key-only entries in keyring

Quexington:
This adds support for entries in the keyring that have no secret information associated with them.  This is perhaps a silly choice to keep in the keyring since it is public information but since the keyring is the standard area where we manage keys right now it seems like the logical choice at least in the short term.


chia/daemon/keychain_proxy.py
```diff
@@ -348,6 +373,42 @@ async def get_key_for_fingerprint(self, fingerprint: Optional[int]) -> Optional[
 
         return key
 
+    async def get_public_key_for_fingerprint(self, fingerprint: Optional[int]) -> Optional[G1Element]:
+        """
+        Locates and returns a private key matching the provided fingerprint
+        """
```
emlowe, line 379: Typo in docstring - this returns a public key

chia/daemon/keychain_server.py
```diff
@@ -234,6 +240,37 @@ async def add_private_key(self, request: Dict[str, Any]) -> Dict[str, Any]:
...
+                "error_details": {"message": "missing public_key"},
+            }
+
+        try:
+            pk = self.get_keychain_for_request(request).add_public_key(public_key, label)
+        except KeyError as e:  # pragma: no cover
+            return {
+                "success": False,
+                "error": KEYCHAIN_ERR_KEYERROR,
+                "error_details": {"message": f"The word '{e.args[0]}' is incorrect.'", "word": e.args[0]},
```
emlowe, line 263: this seems a curious error message - what `word` are you referring to here?

chia/util/keychain.py
```diff
@@ -362,6 +366,36 @@ def add_private_key(self, mnemonic: str, label: Optional[str] = None) -> Private
...
+        # This can probably just be moved into `FileKeyring.set_passphrase` after the legacy keyring stuff was dropped.
+        if label is not None:
+            self.keyring_wrapper.set_label(fingerprint, label)
+
+        try:
+            self.keyring_wrapper.set_passphrase(
+                self.service,
+                get_private_key_user(self.user, index),
+                bytes(key).hex(),
+            )
```
emlowe, line 391: I must admit to being a little confused what set_passphrase is doing here, but I note its used in add_private_key as well where it is equally confusing

chia/daemon/keychain_proxy.py
emlowe, line 379: docstring typo - "returns a public key"

chia/daemon/keychain_server.py
emlowe, line 263: what `word` is this referring to - seems a curious error message

emlowe (commented):
Can we do a little more to remove the `no cover` pragmas

chia/daemon/keychain_proxy.py
emlowe: dup

chia/daemon/keychain_server.py
Quexington: Not sure but I didn't write it and (now) didn't touch it so none of my business.

chia/util/keychain.py
emlowe: Ok, this is oldish code to maintain API consistency with the keyring python module. Could be refactored out since we don't put "keys as a passphrase" anymore into the OS keyring...

emlowe (dismissed):

emlowe (approved):

PR #16978, opened 2023-11-30
Wallet -> MainWalletProtocol

Quexington:
This PR replaces the assumption in the wallet as a whole that the wallet that controls the inner puzzle of a coin will be of type `Wallet` with a new assumption that that wallet will match the `MainWalletProtocol`.


geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16985, opened 2023-12-02
enable compression for cli rpm

altendky:
### Testing Notes:

https://github.com/Chia-Network/chia-blockchain/actions/runs/7730563646?pr=16985
image

### Draft For:
- [x] explicitly set for the gui as well?
  - how about after we get the config consolidated
- [x] is 9 a good level...?
- [x] latest build shows the cli larger again at about 180mb


emlowe:
I might consider using `6` which is the default XZ compression level. Memory use during decompression can be pretty high with `9` (well, "high" in relative terms, about 674 MiB) - the XV docs suggest don't use anything over 7 unless you really need to. `Unless you want to maximize the compression ratio, you probably don't want a higher preset level than -7 due to speed and memory usage.
`
https://linux.die.net/man/1/xz

cmmarslender (approved):

emlowe (approved):

PR #16988, opened 2023-12-03
dao vote amount checks

geoffwalmsley:
1. The proposal and validator puzzles check the total votes and yes vote amounts using `>`, changed to use `gte` so that we don't raise if the actuals are the same as target.
2. Added `gte` and `lte` into utility_macros.
3. Wallets joining the DAO need to know the cat origin details so they can close mint proposals. We can include the cat_origin and launcher IDs in the memos of the DAO creation spend. Syncing wallets can then grab this from the solution when they're setting up. 
4. Other minor fixes:
- prevent creating dao with the same timelock and self-destruct
- display correct asset id for DAO_CAT wallet
- only run exlit lockup if there are coins to spend
- prevent fake p2_singletons being added to treasury
- calculate the CAT tail from the memo data (added in extra fields to the memo to enable this)
- increment index in add_or_update_proposal loop


chia/wallet/puzzles/utility_macros.clib
```diff
@@ -37,4 +37,12 @@
     )
   )
 
+  (defun gte (atom_a atom_b)
+    (or (> atom_a atom_b) (= atom_a atom_b))
```
matt-o-how, line 41: this is incredibly pedantic but I believe `(not (< atom_a atom_b))` would be more efficient here

matt-o-how (dismissed):

matt-o-how (approved):

wjblanke (approved):
aok

PR #16993, opened 2023-12-04
[CHIA-434] Add translation layer for blind signing

Quexington:
This PR introduces a new `TranslationLayer` concept to the `@clvm_streamable` library enabling reconstruction from CHIP-27 APIs into more specific responses for different types of clients.  CHIP-28 is one such chip and support for it is enable as well.


geoffwalmsley (approved):

wjblanke (approved):
aok

PR #16997, opened 2023-12-05
Changed: clarified install prerequisite and added info on cmake error troubles…

Abakrombie:
Changed

### Purpose:

Improve ease and troubleshooting of timelord installation

### Current Behavior:

Current instructions result in failed install for many users and creation of new issues on this repo.

### New Behavior:

Higher install success rate, plus much quicker troubleshooting for a common error.

### Testing Notes:

Test by following instructions on a fresh install


Abachrombie:
bump

hoffmang9:
Two things

1. The sentence before says to be in the venv - which somewhat assumes you ran install.sh and created a venv in the first place. Then there is your change, then there is `activate` again. So we're doing poorly on internal consistency.
2. Tell me more about the cmake error you hit? Was it a version issue?

emlowe:
Closing PR for lack of response

PR #17007, opened 2023-12-06
move electron build config to `build_scripts/electron-builder.json`

altendky:
### Purpose:

https://github.com/Chia-Network/chia-blockchain-gui/pull/2240

### Testing Notes:

The spec files from main and here look unchanged except for the dev version numbers in files etc.

### Draft For:
- [x] after release so as to avoid diverging branches in this regard
- [x] general consideration, this was made as a quick first draft

### At Some Point:
- [x] remove the config from the gui repo https://github.com/Chia-Network/chia-blockchain-gui/pull/2240


paninaro:
Looks good to me. I assume we want to wait for https://github.com/Chia-Network/chia-blockchain-gui/pull/2240 to land in chia-blockchain-gui's `main` before merging this, right?

Only nit-picky thing that I noticed while reading the diff was that the `../../../build_scripts/electron-builder.json` literal is repeated 2+ times in each script -- maybe there's an additional refactor to move common bits into a shared file to reduce copypasta. I wouldn't hold up the review for that, however.

altendky:
This was intended as a minimal 'move the config' PR.  I certainly want to do more cleanup in a few different directions including less `../../../../../../../../`.

altendky:
Err, yeah, this shouldn't go in with the existing submodule reference.  Thanks for catching that.  I pointed directly at the GUI PR commit to get an actual test here.  This PR can go in without changing the submodule at all and then the removal of the then-dead config in the GUI can happen whenever.  I'll remove the submodule update here.

paninaro (approved):

PR #17011, opened 2023-12-06
DL upsert.

fchirica:
Refactors `insert` by breaking down its logic within functions `get_ancestors_common` and `update_ancestor_hashes_on_insert`. These functions are later used to implement `upsert` - we locate the terminal node for the key within the tree, replace it with the terminal node corresponding to the new value, then update the hashes of the ancestors similarly to when we insert a new terminal node.


tests/core/data_layer/test_data_store.py
```diff
@@ -405,10 +406,10 @@ async def test_batch_update(data_store: DataStore, tree_id: bytes32, use_optimiz
                         key=key, value=value, tree_id=tree_id, use_optimized=False, status=Status.COMMITTED
                     )
                 batch.append({"action": "insert", "key": key, "value": value})
-                keys.append(key)
-            else:
-                key = random.choice(keys)
-                keys.remove(key)
+                keys_values[key] = value
+            elif op_type == 0:
```
fchirica, line 410: This modified test should act as a good stress test - we do 60% of operations inserts, 20% upserts and 20% deletes. The whole test runs 1000 operations.

fchirica:
Removed coverage-diff since the 4 lines not covered correspond to exceptions thrown only if bugs are present.

chia/data_layer/data_store.py
```diff
@@ -1073,6 +1102,77 @@ async def delete(
...
+            if root is None:
+                root = await self.get_tree_root(tree_id=tree_id)
+
+            if hint_keys_values is None:
+                old_node = await self.get_node_by_key(key=key, tree_id=tree_id)
+                if old_node.value == new_value:
+                    raise ValueError(f"New value matches old value in upsert operation: {key.hex()}")
+                old_node_hash = old_node.hash
+            else:
+                if bytes(key) not in hint_keys_values:
```
altendky, line 1125: `key` is already `bytes`.  I did not point out all cases of this so look around a bit.

```suggestion
                if key not in hint_keys_values:
```

chia/data_layer/data_store.py
```diff
@@ -1073,6 +1102,77 @@ async def delete(
...
+                    raise KeyNotFoundError(key=key)
```
altendky, line 1126: Isn't this the case for the insert functionality of an upsert?

chia/data_layer/data_store.py
```diff
@@ -1073,6 +1102,77 @@ async def delete(
 
         return new_root
 
+    async def upsert(
```
altendky, line 1105: Do we need side and reference hash control for the insert case of upsert?

tests/core/data_layer/test_data_rpc.py
```diff
@@ -241,6 +241,15 @@ async def test_create_insert_get(
         with pytest.raises(ValueError, match="Changelist resulted in no change to tree data"):
             await data_rpc_api.batch_update({"id": store_id.hex(), "changelist": changelist})
 
+        # test upsert
```
altendky, line 244: Test both the insert and update cases of upsert.

tests/core/data_layer/test_data_store.py
```diff
@@ -386,10 +386,11 @@ async def test_batch_update(data_store: DataStore, tree_id: bytes32, use_optimiz
         random.seed(100, version=2)
 
         batch: List[Dict[str, Any]] = []
-        keys: List[bytes] = []
+        keys_values: Dict[bytes, bytes] = {}
         hint_keys_values: Dict[bytes, bytes] = {}
         for operation in range(num_batches * num_ops_per_batch):
-            if random.randint(0, 4) > 0 or len(keys) == 0:
+            op_type = random.randint(0, 5)
```
altendky, line 392: Maybe something like `[op_type] = random.choices(["insert", "upsert", "delete"], weights=[6, 2, 2], k=1)`?  (or whatever weights you intend)

Though in some way we need to test both insert and update cases of upsert.  So maybe `["insert", "upsert-insert", "upsert-update", "delete"]`.

altendky (changes requested):
Nice to see existing code cleaning up along the way.  Thanks!

chia/data_layer/data_store.py
fchirica: I think it should be equivalent to an autoinsert, otherwise we'll just complicate the interface (when proper use should be `insert` anyway...)

tests/core/data_layer/test_data_store.py
```diff
@@ -451,6 +480,70 @@ async def test_batch_update(data_store: DataStore, tree_id: bytes32, use_optimiz
...
+    all_kv = await data_store.get_keys_values(tree_id)
+    assert {node.key: node.value for node in all_kv} == keys_values
+
+
+@pytest.mark.anyio
+@pytest.mark.parametrize(
+    "use_optimized",
+    [True, False],
+)
+async def test_upsert_ignores_invalid_arguments(
```
altendky, line 492: Invalid?  Or just already existing.

arvidn:
with a type-checked `Program`, this PR still triggers these mypy issues:
```
chia/data_layer/data_layer_util.py:154: error: Incompatible return value type (got "tuple[Program, Program]", expected "tuple[bytes32, bytes32]")  [return-value]
tests/core/data_layer/util.py:193: error: Argument 1 to "get_tree_hash_precalc" of "Program" has incompatible type "bytes32 | None"; expected "bytes32"  [arg-type]
tests/core/data_layer/util.py:193: error: Argument 2 to "get_tree_hash_precalc" of "Program" has incompatible type "bytes32 | None"; expected "bytes32"  [arg-type]
chia/data_layer/data_layer_wallet.py:1269: error: Incompatible types in assignment (expression has type "Program", variable has type "Iterator[Program]")  [assignment]
chia/data_layer/data_layer_wallet.py:1271: error: "Iterator[Program]" has no attribute "as_iter"  [attr-defined]
chia/data_layer/data_layer_wallet.py:1283: error: Argument 1 to "from_hexstr" of "SizedBytes" has incompatible type "Program"; expected "str"  [arg-type]
chia/data_layer/data_layer_wallet.py:1288: error: Incompatible types in assignment (expression has type "Program", variable has type "str | None")  [assignment]
```

They are addressed by https://github.com/Chia-Network/chia-blockchain/pull/17062. But I believe these PRs conflict.

altendky (dismissed):

fchirica:
Draft until #17062 gets in.

altendky (approved):

emlowe (approved):

PR #17012, opened 2023-12-06
Add unregister subscription messages to full node

Rigidity:
### Purpose:

This is a work in progress PR for updating the wallet protocol.

Introduces the following changes to the wallet protocol:

- `UnregisterPhUpdates`, which removes subscriptions added with `RegisterForPhUpdates`.
- `UnregisterCoinUpdates`, which removes subscriptions added with `RegisterForCoinUpdates`.

### Current Behavior:

Doesn't support these new messages and changes to the wallet protocol.

### New Behavior:

Adds the new wallet protocol changes with backward compatibility with previous protocol versions.

### Testing Notes:

Not yet fully tested.

### DIFF
--- a/chia/full_node/subscriptions.py
+++ b/chia/full_node/subscriptions.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import random
 from dataclasses import dataclass
 from typing import List, Set, final
 
@@ -17,7 +18,10 @@ class PeerSubscriptions:
 
     @classmethod
     async def create(cls) -> PeerSubscriptions:
-        db_wrapper = await DBWrapper2.create(database=":memory:")
+        unique_database_uri = (
+            f"file:db_{cls.__module__}.{cls.__qualname__}_{random.randint(0, 99999999)}?mode=memory&cache=shared"
+        )
+        db_wrapper = await DBWrapper2.create(database=unique_database_uri)
         self = PeerSubscriptions(db_wrapper)
 
         async with self.db_wrapper.writer() as conn:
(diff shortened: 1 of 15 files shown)


chia/full_node/subscriptions.py
```diff
@@ -111,6 +111,54 @@ def add_coin_subscriptions(self, peer_id: bytes32, coin_ids: List[bytes32], max_
...
 
+    def remove_ph_subscriptions(self, peer_id: bytes32, phs: List[bytes32]) -> None:
+        puzzle_hash_peers = self._peer_puzzle_hash.get(peer_id)
+
+        # there are no subscriptions to remove
+        if puzzle_hash_peers is None:
+            return
+
+        for ph in phs:
+            if ph not in puzzle_hash_peers:
```
arvidn, line 122: you could probably save some time by removing the puzzle hash optimistically here, and if it fails, `continue`. You would avoid looking it up twice

chia/full_node/subscriptions.py
```diff
@@ -111,6 +111,54 @@ def add_coin_subscriptions(self, peer_id: bytes32, coin_ids: List[bytes32], max_
...
+    def remove_coin_subscriptions(self, peer_id: bytes32, coin_ids: List[bytes32]) -> None:
+        peer_coin_ids = self._peer_coin_ids.get(peer_id)
+
+        # there are no subscriptions to remove
+        if peer_coin_ids is None:
+            return
+
+        for coin_id in coin_ids:
+            if coin_id not in peer_coin_ids:
+                continue
```
arvidn, line 147: ditto here. removing optimistically and `continue` if that fails.

PR #17019, opened 2023-12-08
Move nightly schedule to the actual test workflow

cmmarslender:
Simplify the nightly test runs. No need to trigger a workflow in the same repo when we can just run the schedule directly on the job


hoffmang9 (approved):
lgtm

wjblanke (approved):
aok

PR #17021, opened 2023-12-08
Adjust ban times for DL dat files

emlowe:
Currently, if a file is downloaded, but fails to insert properly, we ban that server for 7 days. In this case, the files can fail to insert for a variety of reasons, such as a truncated download, or disk problem that is likely not a result of a malicious actor.

This ban is overly long and has caused some problems with both CADT and Labrador.

This PR uses the logic for missing files to set a progressive ban per attempt with a maximum of one hour. This is using an existing function for this and so has very little risk. We would like to get into 2.1.2 so we can update CADT with this change

Change made in agreement with Zach and Ken.


wjblanke (approved):
aok

emlowe:
coverage diff exemption

PR #17022, opened 2023-12-08
add SerializedProgram.to() to simplify some code

arvidn:
### Purpose:

This is a step towards making `SerializedProgram` have stronger type-checking. The cases where this new function is used end up constructing the correct type (`SerializedProgram`) rather than `Program` (which is incorrect, but not checked by mypy because it's treated as `Any`).

### Current Behavior:

`Program` is used in places that expect `SerializedProgram`

### New Behavior:

Some (but not all) type-errors are corrected to use the expected type, `SerializedProgram`.


Quexington (dismissed):
Can I approve something more than once :)

Quexington (approved):

emlowe (approved):

PR #17031, opened 2023-12-11
make spend_prefarm.py checked by mypy

arvidn:
### Purpose:

remove `spend_prefarm.py` from mypy exclusions.
All of `clvm` is untyped (currently), that's why we need the type ignore comments.

### Current Behavior:

spend_prefarm.py is not fully checked by mypy

### New Behavior:

spend_prefarm.py is checked by mypy

### Testing Notes:

There seem to not be any tests for `spend_prefarm.py`


arvidn:
We don't have any tests for spend_prefarm..

AmineKhaldi (approved):

wjblanke (approved):
aok

PR #17034, opened 2023-12-11
adapt `conditions_dict_for_solution()` to how it's used

arvidn:
### Purpose:

Most call-sites of `conditions_dict_for_solution()` pass it `Program` instead of `SerializedProgram`. This patch makes it take `Program` instead. Once we have type checking for `Program`, these errors would be detected.

### Current Behavior:

We pass `Program` instead of `SerializedProgram` to `conditions_dict_for_solution()`.

### New Behavior:

`conditions_dict_for_solution()` accepts `Program`, so there's no type error.

### Testing

The fact that both cases have test coverage is evidence of call sites exists for both types.


altendky (dismissed):
not sure this is the best place to end up, but i think it is more correct for a minimal change.

arvidn:
yeah, actually. I'm having second thoughts about this. I will investigate this a little bit more. It's possible the best approach is to just make this take `Program`. It's mostly used by the wallet and tests

altendky (approved):

wjblanke (approved):
aok

PR #17037, opened 2023-12-11
Rewrite peer subscription manager with an in-memory SQLite db

Rigidity:
### Purpose:

Switching from dictionaries to SQLite for PeerSubscriptions reduces the amount of duplicated logic and enables addition of new fields or functionality in the future without exponentially increasing the complexity.

### Current Behavior:

PeerSubscriptions is implemented as a set of dictionaries containing peer ids and subscription ids.

### New Behavior:

It uses an in-memory managed SQLite database to store and query the subscription data.

### Testing Notes:

All of the tests have been updated to use this new async PeerSubscription implementation.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -131,6 +131,7 @@ class FullNode:
     log: logging.Logger
     db_path: Path
     wallet_sync_queue: asyncio.Queue[WalletUpdate]
+    subscriptions: PeerSubscriptions
     _segment_task: Optional[asyncio.Task[None]] = None
     initialized: bool = False
     _server: Optional[ChiaServer] = None
@@ -145,7 +146,6 @@ class FullNode:
     #       config would end up on stdout if handled here.
     multiprocessing_context: Optional[BaseContext] = None
     _ui_tasks: Set[asyncio.Task[None]] = dataclasses.field(default_factory=set)
-    _subscriptions: Optional[PeerSubscriptions] = None
     _transaction_queue_task: Optional[asyncio.Task[None]] = None
     simulator_transaction_callback: Optional[Callable[[bytes32], Awaitable[None]]] = None
     _sync_task: Optional[asyncio.Task[None]] = None
@@ -200,6 +200,7 @@ class FullNode:
             log=logging.getLogger(name),
             db_path=db_path,
             wallet_sync_queue=asyncio.Queue(),
+            subscriptions=PeerSubscriptions(),
         )
 
     @contextlib.asynccontextmanager
@@ -238,7 +239,7 @@ class FullNode:
             reader_count=4,
             log_path=sql_log_path,
             synchronous=db_sync,
-        ) as self._db_wrapper, PeerSubscriptions.managed() as self._subscriptions:
+        ) as self._db_wrapper:
             if self.db_wrapper.db_version != 2:
                 async with self.db_wrapper.reader_no_transaction() as conn:
                     async with conn.execute(
@@ -416,11 +417,6 @@ class FullNode:
     def hint_store(self) -> HintStore:
         assert self._hint_store is not None
         return self._hint_store
-
-    @property
-    def subscriptions(self) -> PeerSubscriptions:
-        assert self._subscriptions is not None
-        return self._subscriptions
 
     @property
     def new_peak_sem(self) -> LimitedSemaphore:
@@ -907,7 +903,7 @@ class FullNode:
         if self.sync_store is not None:
             self.sync_store.peer_disconnected(connection.peer_node_id)
         # Remove all ph | coin id subscription for this peer
-        await self.subscriptions.remove_peer(connection.peer_node_id)
+        self.subscriptions.remove_peer(connection.peer_node_id)
 
     async def _sync(self) -> None:
         """
@@ -1155,7 +1151,7 @@ class FullNode:
                 if state_change_summary is not None:
                     assert peak is not None
                     # Hints must be added to the DB. The other post-processing tasks are not required when syncing
-                    hints_to_add, _ = await get_hints_and_subscription_coin_ids(
+                    hints_to_add, _ = get_hints_and_subscription_coin_ids(
                         state_change_summary,
                         self.subscriptions.is_coin_subscribed,
                         self.subscriptions.is_puzzle_subscribed,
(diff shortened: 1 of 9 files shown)


arvidn:
one concern with a sqlite DB is performance. when converting the mempool to use sqlite, performance was an issue, and some data had to be left out in python dictionaries. It would be good to get some sense of how this impacts performance, and maybe memory pressure as well.

Do you think it would make sense to make a benchmark for the subscriptions class?

The other aspect that's not entirely clear is what kinds of features you anticipate adding to the subscriptions class.

In the case of the mempool, the main rationale for using sqlite was that the number of indices we needed to look up mempool items by grew from 2 to 4. and the two new ones was ordered. I couldn't find any multi-index container for python that would work, so sqite it was.

In this case we still only need two bi-maps, right? puzzle-hash <-> peer-d and coin-id <-> peer-id.

chia/full_node/full_node.py
```diff
@@ -145,7 +145,7 @@ class FullNode:
     #       config would end up on stdout if handled here.
     multiprocessing_context: Optional[BaseContext] = None
     _ui_tasks: Set[asyncio.Task[None]] = dataclasses.field(default_factory=set)
-    subscriptions: PeerSubscriptions = dataclasses.field(default_factory=PeerSubscriptions)
+    _subscriptions: Optional[PeerSubscriptions] = None
```
arvidn, line 148: why are you making this field optional?
it looks like it would add additional complexity. Especially the pre-condition to `subscriptions()` is now much more complex, since it asserts that this field is not `None`.

arvidn: I see, this is the convention now that we use `managed()`

chia/full_node/subscriptions.py
```diff
@@ -1,144 +1,183 @@
...
+    db_wrapper: DBWrapper2
+
+    @classmethod
+    @contextlib.asynccontextmanager
+    async def managed(cls) -> AsyncIterator[PeerSubscriptions]:
+        unique_database_uri = (
+            f"file:db_{cls.__module__}_{cls.__qualname__}_{random.randint(0, 99999999)}?mode=memory&cache=shared"
+        )
+
+        async with DBWrapper2.managed(database=unique_database_uri) as db_wrapper:
```
arvidn, line 27: I would highly recommend against using `aiosqlite`, but instead use `sqlite3` directly (the syncronous API). The time you use to do lookups and updates would, most likely, be dominated by the communication with the DB thread.

I would also recommend against using `DBWrapper` (which you wouldn't be able to with the synchronous API anyway).

chia/full_node/subscriptions.py
```diff
@@ -1,144 +1,183 @@
...
-        * some puzzle hashes are already subscribed to
-        * the max_items limit is exceeded
-        """
+    db_wrapper: DBWrapper2
+
+    @classmethod
+    @contextlib.asynccontextmanager
+    async def managed(cls) -> AsyncIterator[PeerSubscriptions]:
+        unique_database_uri = (
+            f"file:db_{cls.__module__}_{cls.__qualname__}_{random.randint(0, 99999999)}?mode=memory&cache=shared"
```
arvidn, line 24: if you don't use `DBWrapper`, and use the synchronous API, you only need a single connection to the database, and you can just use `":memory:"`.

arvidn (commented):
by using the synchronous `sqlite3` API, you can:
* preserve the synchronous API of `PeerSubscriptions`
* have better performance (because there's no threads)
* have a simpler (and more reliable) database name

chia/full_node/subscriptions.py
```diff
@@ -1,144 +1,179 @@
...
+                    coin_id BLOB,
+                    UNIQUE (peer_id, coin_id)
+                )
+                """
+            )
 
-        puzzle_hash_peers = self._peer_puzzle_hash.setdefault(peer_id, set())
-        existing_sub_count = self._peer_sub_counter.setdefault(peer_id, 0)
+    def __del__(self) -> None:
+        self._db_conn.close()
```
arvidn, line 42: I suspect Kyle would prefer a `manage()` member that's a context manager and handles construction an destruction. But I haven't personally really gotten familiar with this pattern

chia/full_node/subscriptions.py
```diff
@@ -1,144 +1,179 @@
...
+            )
+            conn.execute(
+                """
+                CREATE TABLE coin_subscriptions (
+                    id INT PRIMARY KEY,
+                    peer_id BLOB,
+                    coin_id BLOB,
+                    UNIQUE (peer_id, coin_id)
+                )
+                """
```
arvidn, line 38: These tables look like they may actually store duplicates of coin-IDs and puzzle-hashes. Are you confident this is more memory efficient?

I would also expect that you would need an index for both `peer_id`, `coin_id` and `puzzle_hash` for both of these tables.

chia/full_node/subscriptions.py
```diff
@@ -1,144 +1,179 @@
...
+
+        with self._db_conn as conn:
+            # Decrement this counter as we go, to know if we've hit the subscription limit.
+            subscriptions_left = max_items - existing_sub_count
+
+            for item in items:
+                row = conn.execute(
+                    f"SELECT COUNT(*) FROM {table_name} WHERE peer_id=? AND {item_name}=?",
+                    (peer_id, item),
+                ).fetchone()
```
arvidn, line 84: it looks really odd to count the number of rows here just to decide whether we try to insert it or not.
I would expect it to be a lot faster to just insert it and handle the error if it's already added.

chia/full_node/subscriptions.py
```diff
@@ -1,144 +1,179 @@
...
+            max_items=max_items,
+        )
+
+    def peer_subscription_count(self, peer_id: bytes32) -> int:
+        with self._db_conn as conn:
+            row = conn.execute("SELECT COUNT(*) FROM puzzle_subscriptions WHERE peer_id=?", (peer_id,)).fetchone()
+            assert row is not None
+            puzzle_subscription_count = int(row[0])
+
+            row = conn.execute("SELECT COUNT(*) FROM coin_subscriptions WHERE peer_id=?", (peer_id,)).fetchone()
```
arvidn, line 140: I believe this could be a single query. https://stackoverflow.com/questions/7524612/how-to-count-rows-from-multiple-tables-in-sqlite

chia/full_node/subscriptions.py
```diff
@@ -1,144 +1,179 @@
...
+
+    def total_puzzle_subscriptions(self) -> int:
+        with self._db_conn as conn:
+            row = conn.execute("SELECT COUNT(*) FROM puzzle_subscriptions").fetchone()
+            assert row is not None
+            puzzle_subscription_count = int(row[0])
+
+        return puzzle_subscription_count
+
+    def total_coin_subscriptions(self) -> int:
```
arvidn, line 154: I believe `peer_subscription_count()`, `total_puzzle_subscriptions()` and `total_coin_subscriptions()` are missing unit tests

arvidn (commented):
some comments.

I also think it would be really good to ensure this is not significantly slower or more memory hungry than the existing implementation

PR #17030, opened 2023-12-11
fix type mismatch with Optional[bytes] and bytes in wallet/conditions.py

arvidn:
### Purpose:

This is a step towards improving type checking of `Program` objects.

In this PR, types that are `Optional[]` are not checked for `None` before being used. mypy doesn't complain because all `Program` members are untyped.

### Current Behavior:

`Optional[bytes]` are used as if they're `bytes`.

### New Behavior:

`Optional[bytes]` are checked for `None` before being used as `bytes`.

### mypy

When enabling mypy checking in `conditions.py`, it reports these issues:

```
chia/wallet/conditions.py:47: error: Argument 2 to "AggSigParent" has incompatible type "bytes | None"; expected "bytes"  [arg-type]
chia/wallet/conditions.py:68: error: Argument 2 to "AggSigPuzzle" has incompatible type "bytes | None"; expected "bytes"  [arg-type]
chia/wallet/conditions.py:89: error: Argument 2 to "AggSigAmount" has incompatible type "bytes | None"; expected "bytes"  [arg-type]
chia/wallet/conditions.py:116: error: Argument 2 to "AggSigPuzzleAmount" has incompatible type "bytes | None"; expected "bytes"  [arg-type]
chia/wallet/conditions.py:144: error: Argument 2 to "AggSigParentAmount" has incompatible type "bytes | None"; expected "bytes"  [arg-type]
chia/wallet/conditions.py:172: error: Argument 2 to "AggSigParentPuzzle" has incompatible type "bytes | None"; expected "bytes"  [arg-type]
chia/wallet/conditions.py:193: error: Argument 2 to "AggSigUnsafe" has incompatible type "bytes | None"; expected "bytes"  [arg-type]
chia/wallet/conditions.py:219: error: Argument 2 to "AggSigMe" has incompatible type "bytes | None"; expected "bytes"  [arg-type]
chia/wallet/conditions.py:244: error: Argument 1 to "bytes32" has incompatible type "bytes | None"; expected "Iterable[SupportsIndex] | SupportsBytes"  [arg-type]
chia/wallet/conditions.py:246: error: List comprehension has incompatible type List[bytes | None]; expected List[bytes]  [misc]
chia/wallet/conditions.py:305: error: Argument 1 to "bytes32" has incompatible type "bytes | None"; expected "Iterable[SupportsIndex] | SupportsBytes"  [arg-type]
[...]

PR #17040, opened 2023-12-12
Optional atom fixes

arvidn:
### Purpose:

This PR addresses more cases where `Optional[bytes]` is used as `bytes` directly. These type errors are not caught by mypy because `Program` and all `SExp` members are untyped.

This is a continuation of https://github.com/Chia-Network/chia-blockchain/pull/17030, where `as_atom()` was made to throw in case the `SExp` node isn't an atom, and otherwise return `bytes`. This is by far the most common use case of `.atom`.

Also note that `as_int()` also throws in case the node is not an atom.

### Current Behavior:

When enabling type checking, all these calls to `.atom` are flagged because they are used as if they are `bytes`, rather than `Optional[bytes]`.

### New Behavior:

When enabling type checking, these cases are not flagged.


arvidn:
I spent some time trying to add test coverage for eve-spends in `test_singleton_store.py`, but it's quite complicated. It would be better if someone more familiar with CATs would do that.

Quexington (approved):
Slam dunk

emlowe (approved):

PR #17043, opened 2023-12-12
Add a few missing type annotations

Rigidity:
### Purpose:

Adds type annotations that were missing in a few places, which mypy (configured more strictly for my editor, presumably) was complaining about.

### DIFF
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -63,7 +63,7 @@ class KeyringWrapper:
     """
 
     # Static members
-    __shared_instance = None
+    __shared_instance: Optional[KeyringWrapper] = None
     __keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH
 
     # Instance members
--- a/chia/util/keyring_wrapper.py
+++ b/chia/util/keyring_wrapper.py
@@ -63,8 +63,8 @@ class KeyringWrapper:
     """
 
     # Static members
-    __shared_instance: Optional[KeyringWrapper] = None
-    __keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH
+    __shared_instance: ClassVar[Optional[KeyringWrapper]] = None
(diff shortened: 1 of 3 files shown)


AmineKhaldi (dismissed):

chia/util/struct_stream.py
```diff
@@ -89,4 +89,4 @@ def stream_to_bytes(self) -> bytes:
 
     # this is meant to avoid mixing up construcing a bytes object of a specific
     # size (i.e. bytes(int)) vs. serializing the integer to bytes (i.e. bytes(uint32))
-    __bytes__ = None
+    __bytes__: None = None
```
altendky, line 92: This was changed in release and then checkpointed so presumably it is just conflicted.  It's a regular function def again (for now) so probably just revert this.

```suggestion
    __bytes__ = None
```

altendky (changes requested):
I would guess your IDE isn't using mypy.  If it were and was using the strict config or such it'd be complaining a lot more.  But, I'm just guessing.

chia/util/keyring_wrapper.py
```diff
@@ -63,7 +63,7 @@ class KeyringWrapper:
     """
 
     # Static members
-    __shared_instance = None
+    __shared_instance: Optional[KeyringWrapper] = None
     __keys_root_path: Path = DEFAULT_KEYS_ROOT_PATH
```
altendky, line 67:
```suggestion
    __shared_instance: ClassVar[Optional[KeyringWrapper]] = None
    __keys_root_path: ClassVar[Path] = DEFAULT_KEYS_ROOT_PATH
```

Rigidity:
> I would guess your IDE isn't using mypy. If it were and was using the strict config or such it'd be complaining a lot more. But, I'm just guessing.

It should be, don't know why it didn't catch this. Do I need to configure it to lint more strictly?

altendky (dismissed):

altendky (approved):

arvidn (approved):

PR #17054, opened 2023-12-13
Verify p2 delegated conditions signatures

MarvinQuevedo:
- Refactor the` /verify_signature `API endpoint to support the verification process for signatures generated with Tangem Cards. Tangem Cards utilize a distinct puzzle. When crafting a signature with this hardware wallet, the synthetic_public_key is not employed. Consequently, when a developer attempts to verify the signature, the initial BLS signature verification step is successful. However, challenges arise when attempting to recreate the puzzle for comparing the PuzzleHash, as they are not equal. Now, if it is a Tangem Signature, the code recreates the Puzzle with the PublicKey using the p2_delegated_conditions puzzle, which is equivalent to the TangemCard puzzle, It allow verify correctly the signature maked with a Tangem Card.

- Introduce a new SigningMode to identify signatures produced with Tangem Cards using CHIP_002:

`CHIP_0002_P2_DELEGATED_CONDITIONS = "BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_AUG:CHIP-0002_P2_DELEGATED_PUZZLE"`


paninaro:
Thanks @MarvinQuevedo, the change looks good to me. We'll need a test added for this case. I think `test_verify_signature` is the correct place for that: https://github.com/Chia-Network/chia-blockchain/blob/main/tests/wallet/rpc/test_wallet_rpc.py#L2193

Note that the test has simulated requests and responses starting at https://github.com/Chia-Network/chia-blockchain/blob/main/tests/wallet/rpc/test_wallet_rpc.py#L2097, so it should be fairly simple to add another request/response case for the p2 delegated conditions signing scheme.

Thanks for the addition!

MarvinQuevedo:
Hi  @paninaro I just added the test data, I test using the wallet UI and everything seems to be going well

MarvinQuevedo:
@paninaro I fixed the error message in the unit test for the invalid signature

MarvinQuevedo:
I just refactor the param type because it break the build process

MarvinQuevedo:
I just reformatter the files with back formatter, If It need any other change, please tell me about, thanks!

chia/rpc/wallet_rpc_api.py
```diff
@@ -9,7 +9,7 @@
 
 from chia_rs import AugSchemeMPL, G1Element, G2Element, PrivateKey
 from clvm_tools.binutils import assemble
-
+from chia.wallet.puzzles import p2_delegated_conditions
```
hoffmang9, line 12: isort is complaining that this is not in alphabetic order

MarvinQuevedo:
Imports sorted

MarvinQuevedo:
Hi, I tried to avoid the error, but for some reason all the test about the verify signatures, I think that are the old signatures, but I don't know whay it's happening

https://github.com/Chia-Network/chia-blockchain/actions/runs/7214785151/job/19657832866?pr=17054

MarvinQuevedo:
I will to continue checking why in some cases don't pass the test

MarvinQuevedo:
@paninaro I found the error, I fixed the test data and It would be ok now with the signatures test, thanks!

MarvinQuevedo:
@paninaro only has a error with the core test, It's ok or need to pull the main to try to obtain all test passed?

paninaro:
@MarvinQuevedo thanks for working through the test errors. This last issue doesn't appear to be related to your PR so I'll just re-run it to see if it passes. This looks good to me now.

I did spend a bit of time yesterday trying to work through the test failures. The challenge I ran into is that it's not trivial to create a signature using this scheme. It would be nice if `/sign_message_by_address` and `/sign_message_by_id` supported the p2 delegated conditions scheme, but there's also the challenge of creating an XCH address using that puzzle. I'll check with some folks to see if this is something we'd want to take on.

hoffmang9 (approved):
lgtm and thanks!

paninaro (approved):

MarvinQuevedo:
> @MarvinQuevedo thanks for working through the test errors. This last issue doesn't appear to be related to your PR so I'll just re-run it to see if it passes. This looks good to me now.
> 
> I did spend a bit of time yesterday trying to work through the test failures. The challenge I ran into is that it's not trivial to create a signature using this scheme. It would be nice if `/sign_message_by_address` and `/sign_message_by_id` supported the p2 delegated conditions scheme, but there's also the challenge of creating an XCH address using that puzzle. I'll check with some folks to see if this is something we'd want to take on.

Yes, the problem It's that for default ChiaWallet are not using this puzzle, the idea of have at least the verification  it's can have a oficial way to check this signatures, for these dAPP can verify the Tangem Signatures and with it the Tangem Card can join the most completely posible in the Chia Blockchain ecosystem an can verify theirs NFT, DID and new features that can develop later.

Thanks for all your time too!

PR #17056, opened 2023-12-13
fix type annotation for solve_did_tp

arvidn:
### Purpose:

The type passed in as 4th argument to `solve_did_tp()` is `bytes32` at all call sites. This updates the type annotation to match.

`match_tp_covenant_adapter()` is not used anywhere, but its declared return type does not match what's actually returned. That's addressed.

a value of type `Program` is assigned to a variable of type `bytes32`. This as addressed by calling `bytes32(x.as_atom())`.

This is part of an effort to improve type checking in `chia-blockchain` for `Program`.

It fixes these issues:
```
chia/wallet/vc_wallet/vc_drivers.py:148: error: Incompatible return value type (got "Program", expected "tuple[Program] | None")  [return-value]
chia/wallet/vc_wallet/vc_drivers.py:550: error: Incompatible types in assignment (expression has type "Program", variable has type "bytes32")  [assignment]
tests/wallet/vc_wallet/test_vc_lifecycle.py:228: error: Argument 4 to "solve_did_tp" has incompatible type "bytes32"; expected "Program"  [arg-type]
tests/wallet/vc_wallet/test_vc_lifecycle.py:263: error: Argument 4 to "solve_did_tp" has incompatible type "bytes32"; expected "Program"  [arg-type]
tests/wallet/vc_wallet/test_vc_lifecycle.py:289: error: Argument 4 to "solve_did_tp" has incompatible type "bytes32"; expected "Program"  [arg-type]
```

### Current Behavior:

`solve_did_tp()` has incorrect type hint for the 4th parameter.

### New Behavior:

`solve_did_tp()` has correct type hint for the 4th parameter.

### DIFF
--- a/chia/wallet/vc_wallet/vc_drivers.py
+++ b/chia/wallet/vc_wallet/vc_drivers.py
@@ -143,7 +143,7 @@ def create_tp_covenant_adapter(covenant_layer: Program) -> Program:
     return EML_TP_COVENANT_ADAPTER.curry(covenant_layer)
 
 
-def match_tp_covenant_adapter(uncurried_puzzle: UncurriedPuzzle) -> Optional[Tuple[Program]]:  # pragma: no cover
+def match_tp_covenant_adapter(uncurried_puzzle: UncurriedPuzzle) -> Optional[Program]:  # pragma: no cover
     if uncurried_puzzle.mod == EML_TP_COVENANT_ADAPTER:
         return uncurried_puzzle.args.at("f")
     else:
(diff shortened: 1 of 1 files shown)


Quexington (approved):
I feel like maybe the hint was `Tuple[Program]` for a reason and maybe the return should have been fixed but if that's the case it's not difficult to add back.  Just leaving this comment in case we need to look back on this.

arvidn:
> I feel like maybe the hint was Tuple[Program] for a reason

Yeah, I tried to figure out what the intention was, but this function is not called from anywhere. So, technically it doesn't matter (yet).

wjblanke (approved):
aok

PR #17062, opened 2023-12-14
fix type hints in data layer

arvidn:
### Purpose:

This is a step towards enabling type hints for `Program` and its member functions, as well as full mypy coverage of chia-blockchain.

### Current Behavior:

there are some invalid type hints and mypy violations in the data store code

### New Behavior:

there are fewer incorrect type hints and mypy violations in the data store code

### Testing Notes:

These are the mypy issues addressed by this PR:

```
chia/data_layer/data_layer_util.py:154: error: Incompatible return value type (got "tuple[Program, Program]", expected "tuple[bytes32, bytes32]")  [return-value]
tests/core/data_layer/util.py:193: error: Argument 1 to "get_tree_hash_precalc" of "Program" has incompatible type "bytes32 | None"; expected "bytes32"  [arg-type]
tests/core/data_layer/util.py:193: error: Argument 2 to "get_tree_hash_precalc" of "Program" has incompatible type "bytes32 | None"; expected "bytes32"  [arg-type]
tests/core/data_layer/test_data_store.py:684: error: Argument "key" to "delete" of "DataStore" has incompatible type "Program"; expected "bytes"  [arg-type]
tests/core/data_layer/test_data_store.py:723: error: Argument "key" to "delete" of "DataStore" has incompatible type "Program"; expected "bytes"  [arg-type]
tests/core/data_layer/test_data_store.py:726: error: Argument "key" to "delete" of "DataStore" has incompatible type "Program"; expected "bytes"  [arg-type]
tests/core/data_layer/test_data_store.py:768: error: Argument "key" to "delete" of "DataStore" has incompatible type "Program"; expected "bytes"  [arg-type]
tests/core/data_layer/test_data_store.py:807: error: Argument "key" to "delete" of "DataStore" has incompatible type "Program"; expected "bytes"  [arg-type]
tests/core/data_layer/test_data_store.py:810: error: Argument "key" to "delete" of "DataStore" has incompatible type "Program"; expected "bytes"  [arg-type]
chia/data_layer/data_layer_wallet.py:1269: error: Incompatible types in assignment (expression has type "Program", variable has type "Iterator[Program]")  [assignment]
chia/data_layer/data_layer_wallet.py:1271: error: "Iterator[Program]" has no attribute "as_iter"  [attr-defined]
chia/data_layer/data_layer_wallet.py:1283: error: Argument 1 to "from_hexstr" of "SizedBytes" has incompatible type "Program"; expected "str"  [arg-type]
chia/data_layer/data_layer_wallet.py:1288: error: Incompatible types in assignment (expression has type "Program", variable has type "str | None")  [assignment]
```

### DIFF
--- a/chia/data_layer/data_layer_wallet.py
+++ b/chia/data_layer/data_layer_wallet.py
@@ -1251,9 +1251,9 @@ class DataLayerWallet:
         for spend in offer.coin_spends():
             matched, curried_args = match_dl_singleton(spend.puzzle_reveal.to_program())
             if matched and spend.coin.name() not in all_parent_ids:
-                innerpuz, root, launcher_id = curried_args
+                innerpuz, root_prg, launcher_id = curried_args
                 singleton_struct = launcher_to_struct(bytes32(launcher_id.as_python())).get_tree_hash()
-                singleton_to_root[singleton_struct] = bytes32(root.as_python())
+                singleton_to_root[singleton_struct] = bytes32(root_prg.as_python())
                 singleton_to_innerpuzhash[singleton_struct] = innerpuz.get_tree_hash()
 
         # Create all of the new solutions
@@ -1266,9 +1266,9 @@ class DataLayerWallet:
                 except EvalError:
                     new_spends.append(spend)
                     continue
-                mod, curried_args = graftroot.uncurry()
+                mod, curried_args_prg = graftroot.uncurry()
                 if mod == GRAFTROOT_DL_OFFERS:
-                    _, singleton_structs, _, values_to_prove = curried_args.as_iter()
+                    _, singleton_structs, _, values_to_prove = curried_args_prg.as_iter()
                     all_proofs = []
                     roots = []
                     for singleton, values in zip(singleton_structs.as_iter(), values_to_prove.as_python()):
@@ -1276,7 +1276,7 @@ class DataLayerWallet:
                         proofs_of_inclusion = []
                         for value in values:
                             for proof_of_inclusion in solver["proofs_of_inclusion"]:
-                                root = proof_of_inclusion[0]
+                                root: str = proof_of_inclusion[0]
                                 proof: Tuple[int, List[bytes32]] = (proof_of_inclusion[1], proof_of_inclusion[2])
                                 calculated_root: bytes32 = _simplify_merkle_proof(value, proof)
                                 if (
--- a/chia/data_layer/data_store.py
+++ b/chia/data_layer/data_store.py
@@ -978,7 +978,7 @@ class DataStore:
                         await self._insert_ancestor_table(left_hash, right_hash, tree_id, new_generation)
 
         if hint_keys_values is not None:
-            hint_keys_values[bytes(key)] = value
+            hint_keys_values[key] = value
         return InsertResult(node_hash=new_terminal_node_hash, root=new_root)
 
     async def delete(
@@ -995,13 +995,13 @@ class DataStore:
             if hint_keys_values is None:
                 node = await self.get_node_by_key(key=key, tree_id=tree_id)
             else:
-                if bytes(key) not in hint_keys_values:
+                if key not in hint_keys_values:
                     log.debug(f"Request to delete an unknown key ignored: {key.hex()}")
                     return root
-                value = hint_keys_values[bytes(key)]
+                value = hint_keys_values[key]
                 node_hash = leaf_hash(key=key, value=value)
                 node = TerminalNode(node_hash, key, value)
-                del hint_keys_values[bytes(key)]
+                del hint_keys_values[key]
             if use_optimized:
                 ancestors: List[InternalNode] = await self.get_ancestors_optimized(
                     node_hash=node.hash, tree_id=tree_id, root_hash=root_hash
--- a/tests/core/data_layer/util.py
+++ b/tests/core/data_layer/util.py
@@ -189,6 +189,8 @@ def create_valid_node_values(
     right_hash: Optional[bytes32] = None,
 ) -> Dict[str, Any]:
     if node_type == NodeType.INTERNAL:
+        assert left_hash is not None
+        assert right_hash is not None
         return {
             "hash": Program.to((left_hash, right_hash)).get_tree_hash_precalc(left_hash, right_hash),
             "node_type": node_type,
--- a/tests/core/data_layer/util.py
+++ b/tests/core/data_layer/util.py
@@ -6,7 +6,7 @@ import os
 import pathlib
 import subprocess
 from dataclasses import dataclass
-from typing import IO, TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Union
+from typing import IO, TYPE_CHECKING, Any, Dict, Iterator, List, Literal, Optional, Union, overload
 
 from chia.data_layer.data_layer_util import NodeType, Side, Status
 from chia.data_layer.data_store import DataStore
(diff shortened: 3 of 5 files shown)


chia/data_layer/data_layer_util.py
```diff
@@ -150,7 +150,7 @@ class TerminalNode:
     atom: None = field(init=False, default=None)
 
     @property
-    def pair(self) -> Tuple[bytes32, bytes32]:
+    def pair(self) -> Tuple[Program, Program]:
```
arvidn, line 153: I wasn't sure about which way to go to address this. but as far as I can tell, this function isn't used anywhere (at least not triggered by tests, I believe).

Perhaps the mistake was to wrap `self.key` and `self.value` in `Program` and the return type was correct.

chia/data_layer/data_layer_wallet.py
```diff
@@ -1251,9 +1251,9 @@ async def finish_graftroot_solutions(offer: Offer, solver: Solver) -> Offer:
         for spend in offer.coin_spends():
             matched, curried_args = match_dl_singleton(spend.puzzle_reveal.to_program())
             if matched and spend.coin.name() not in all_parent_ids:
-                innerpuz, root, launcher_id = curried_args
+                innerpuz, root_prg, launcher_id = curried_args
                 singleton_struct = launcher_to_struct(bytes32(launcher_id.as_python())).get_tree_hash()
-                singleton_to_root[singleton_struct] = bytes32(root.as_python())
+                singleton_to_root[singleton_struct] = bytes32(root_prg.as_python())
```
arvidn, line 1256: this change warrants some scrutiny. the variable `root` is used below as well (but with the type `str`). As far as I can tell, this version of `root` was intended to be isolated to this scope.

chia/data_layer/data_layer_wallet.py
```diff
@@ -1266,17 +1266,17 @@ async def finish_graftroot_solutions(offer: Offer, solver: Solver) -> Offer:
...
+                    _, singleton_structs, _, values_to_prove = curried_args_prg.as_iter()
                     all_proofs = []
                     roots = []
                     for singleton, values in zip(singleton_structs.as_iter(), values_to_prove.as_python()):
                         asserted_root: Optional[str] = None
                         proofs_of_inclusion = []
                         for value in values:
                             for proof_of_inclusion in solver["proofs_of_inclusion"]:
-                                root = proof_of_inclusion[0]
+                                root: str = proof_of_inclusion[0]
```
arvidn, line 1279: this value's type is laundered through the `Solver`s dictionary of `Any`. That's why it needs to be hinted here

chia/data_layer/data_layer_wallet.py
```diff
@@ -1266,17 +1266,17 @@ async def finish_graftroot_solutions(offer: Offer, solver: Solver) -> Offer:
                 except EvalError:
                     new_spends.append(spend)
                     continue
-                mod, curried_args = graftroot.uncurry()
+                mod, curried_args_prg = graftroot.uncurry()
                 if mod == GRAFTROOT_DL_OFFERS:
-                    _, singleton_structs, _, values_to_prove = curried_args.as_iter()
+                    _, singleton_structs, _, values_to_prove = curried_args_prg.as_iter()
```
arvidn, line 1271: `curried_args` is a `Program` here, but in other places it's a `List[Program]` in other places, so it needs a different name here

chia/data_layer/data_store.py
```diff
@@ -978,7 +979,7 @@ async def insert(
                         await self._insert_ancestor_table(left_hash, right_hash, tree_id, new_generation)
 
         if hint_keys_values is not None:
-            hint_keys_values[bytes(key)] = value
```
arvidn, line 981: `key` is `bytes` already. This seems redundant (as long as we can trust mypy. I added the assert above for now, just to make sure we can trust it.

tests/core/data_layer/util.py
```diff
@@ -189,8 +189,11 @@ def create_valid_node_values(
     right_hash: Optional[bytes32] = None,
 ) -> Dict[str, Any]:
     if node_type == NodeType.INTERNAL:
+        treehash = Program.to((left_hash, right_hash)).get_tree_hash_precalc(
+            *[h for h in [left_hash, right_hash] if h is not None]
+        )
```
arvidn, line 194: `get_tree_hash_precalc()` takes varargs of `bytes32`, not `Optional[bytes32]`. This creates a `List[Optional[bytes32]]` the condenses it into a `List[bytes32]` and then expands it to a parameter pack.

altendky: This probably needs an overload.  If the node type is internal then the hashes must be specified, otherwise they must not be.  I believe the sql layer already enforces this, fwiw, but yeah, hinting was incomplete at this level.

chia/data_layer/data_store.py
```diff
@@ -990,18 +991,19 @@ async def delete(
         status: Status = Status.PENDING,
         root: Optional[Root] = None,
     ) -> Optional[Root]:
+        assert isinstance(key, bytes)
```
altendky, line 994: We don't generally runtime type check and if we want to I'd consider a general library for that purpose.  It's certainly useful for double checking but I would rather not leave this in.

chia/data_layer/data_layer_util.py
altendky: I think this atom and pair stuff is just garbage at this point.  We'll see what CI says in https://github.com/Chia-Network/chia-blockchain/pull/17063.

altendky (changes requested):

tests/core/data_layer/util.py
arvidn: you mean `get_tree_hash_precalc()` needs an overload?
Would you prefer I change this function to take a vararg of `Optional[bytes32]`?

chia/data_layer/data_layer_util.py
altendky: Looks more intertwined than I thought.  Nodes end up passed to `Program.to()`.  Maybe that shouldn't be the case, but it's at least not a quick delete to resolve this.

tests/core/data_layer/util.py
altendky: No, this function we're in `create_valid_node_values()`.  If called for an internal node the hashes should be specified.  If called for a terminal node, they should not be specified.  It might be that it should just be two functions.  The suggested overload would help call sites where it is called with a literal for the node type, not sure about call sites using a variable there.  Probably not those cases.  And, I don't think it would actually help anything at all inside the function where you are trying to fix something.  So...

Maybe just assert not `None` on the hashes here.  Maybe also assert `None` for them in the terminal node case below for completeness.  Not beautiful, but puts checks in according to the intended usage and also the hinted usage if the overload is done.

arvidn: in the `TERMINAL` node case, `left_hash` (and I assume `right_hash`) are not guaranteed to be `None` at least not in the tests. In the internal node case, the tests don't seem to trigger an assert however.

e.g.
```
[gw0] linux -- Python 3.11.6 /home/arvid/dev/2/chia-blockchain/venv/bin/python
venv/lib/python3.11/site-packages/anyio/pytest_plugin.py:72: in wrapper
    yield runner.run_fixture(func, kwargs)
venv/lib/python3.11/site-packages/anyio/_backends/_asyncio.py:1926: in run_fixture
    retval = self.get_loop().run_until_complete(
/usr/lib/python3.11/asyncio/base_events.py:653: in run_until_complete
    return future.result()
venv/lib/python3.11/site-packages/anyio/_backends/_asyncio.py:1896: in _call_in_runner_task
    return await future
venv/lib/python3.11/site-packages/anyio/_backends/_asyncio.py:1874: in _run_tests_and_fixtures
    retval = await coro
tests/core/data_layer/conftest.py:92: in valid_node_values_fixture
    return create_valid_node_values(node_type=node_type, left_hash=node_a.hash, right_hash=node_b.hash)
tests/core/data_layer/util.py:203: in create_valid_node_values
    assert left_hash is None
E   AssertionError
```

chia/data_layer/data_layer_util.py
arvidn: should I restore the original patch here, to make the return type `Program`?

fchirica (approved):

altendky (approved):

arvidn:
the line not covered by tests is:
```
raise Exception(f"invalid node type: {node_type!r}")
```
which is error handling that wasn't covered prior to this change either

wjblanke (approved):
aok

PR #17071, opened 2023-12-16
Fix compressed blocks

arvidn:
### Purpose:

`BlockTools` require the ability to generate invalid blocks, for negative tests. Originally, `BlockTools` used the production versions of `create_unfinished_block()` and `create_foliage()`. However, these functions are made to only generate valid blocks, so it was duplicated and modified in the `block_tools.py` file, to avoid mixing test code with production code. However, `BlockTools` was the only reason the production code had test coverage, so this change (a long time ago) removed test coverage for the production code generating blocks.

This PR restores `BlockTools` to use the production versions of `create_unfinished_block()` and `create_foliage()`. In order to still be able to customize these functions to accept invalid blocks, two customization points are introduced. `compute_cost`  and `compute_fees`.

1. `compute_cost()` uses the real block validation code in production, which won't allow invalid blocks. The test version attempts to compute the cost with as little validation as possible
2. `compute_fees()` does not allow negative fees in production (it's stored in an unsigned field) whereas the test version clamps negative fees to 0, thus allowing tests to create blocks that mint coins.

### Current Behavior:

`create_unfinished_block()` and `create_foliage()` do not have test coverage.

### New Behavior:

`create_unfinished_block()` and `create_foliage()` *do* have test coverage.

### DIFF
--- a/tests/core/consensus/test_block_creation.py
+++ b/tests/core/consensus/test_block_creation.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+
+from typing import List
+
+import pytest
+
+from chia.consensus.block_creation import compute_block_cost, compute_block_fee
+from chia.consensus.condition_costs import ConditionCost
+from chia.consensus.default_constants import DEFAULT_CONSTANTS
+from chia.types.blockchain_format.coin import Coin
+from chia.types.blockchain_format.sized_bytes import bytes32
+from chia.util.ints import uint32
+from tests.core.make_block_generator import make_block_generator
+
+
+@pytest.mark.parametrize("add_amount", [[0], [1, 2, 3], []])
+@pytest.mark.parametrize("rem_amount", [[0], [1, 2, 3], []])
+def test_compute_block_fee(add_amount: List[int], rem_amount: List[int]) -> None:
+    additions: List[Coin] = [Coin(bytes32.random(), bytes32.random(), amt) for amt in add_amount]
+    removals: List[Coin] = [Coin(bytes32.random(), bytes32.random(), amt) for amt in rem_amount]
+
+    # the fee is the left-overs from the removals (spent) coins after deducting
+    # the newly created coins (additions)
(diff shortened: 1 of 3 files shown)


tests/core/consensus/test_block_creation.py
```diff
@@ -0,0 +1,58 @@
...
+        assert compute_block_fee(additions, removals) == expected
+
+
+def test_compute_block_cost(softfork_height: uint32) -> None:
+    num_coins = 10
+    generator = make_block_generator(num_coins)
+    cost = int(compute_block_cost(generator, DEFAULT_CONSTANTS, softfork_height))
+
+    coin_cost = 1800000 * num_coins
+    agg_sig_cost = 1200000 * num_coins
```
altendky, line 48: Where do these literals come from?

tests/core/consensus/test_block_creation.py
```diff
@@ -0,0 +1,58 @@
...
+
+    cost -= coin_cost
+    cost -= agg_sig_cost
+    print(cost)
```
altendky, line 52: Maybe more than this would be good, but at least have some sort of text with the output.

```suggestion
    print(f"{cost=}")
```

tests/core/consensus/test_block_creation.py
```diff
@@ -0,0 +1,58 @@
...
+    if softfork_height >= DEFAULT_CONSTANTS.HARD_FORK_HEIGHT:
+        expected = 46560980
+    else:
+        expected = 50316699
```
altendky, line 56: I guess this will make sure it doesn't change, not clear about the literals again here though and what makes them correct.

altendky (changes requested):

tests/core/consensus/test_block_creation.py
arvidn: fixed

altendky (approved):

emlowe (approved):

PR #17089, opened 2023-12-19
initiate phasing out of the `coin_solutions` name in JSON structs

arvidn:
in favor of `coin_spends`.

### Purpose:

The end goal is to remove the special case JSON logic in `SpendBundle`.

### Current Behavior:

The JSON field is just called `coin_solutions`.

### New Behavior:

There are two fields in the JSON struct, both `coin_solutions` *and* `coin_spends`.


Quexington (approved):

wjblanke (approved):
aok

PR #17092, opened 2023-12-19
slight simplification to get_min_fee_rate()

arvidn:
### Purpose:

instead of throwing an exception from `get_min_fee_rate()`, return None.

### Current Behavior:

`get_min_fee_rate()` may throw an exception if the mempool limit is set small.

### New Behavior:

`get_min_fee_rate()`  does not throw an exception because the mempool limit is set small.

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -244,11 +244,12 @@ class MempoolManager:
         if cost == 0:
             return False
         fees_per_cost = fees / cost
-        if not self.mempool.at_full_capacity(cost) or (
-            fees_per_cost >= self.nonzero_fee_minimum_fpc and fees_per_cost > self.mempool.get_min_fee_rate(cost)
-        ):
+        if not self.mempool.at_full_capacity(cost):
             return True
-        return False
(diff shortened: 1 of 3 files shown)


chia/full_node/mempool_manager.py
```diff
@@ -244,10 +244,14 @@ def is_fee_enough(self, fees: uint64, cost: uint64) -> bool:
...
-        ):
+        if not self.mempool.at_full_capacity(cost):
             return True
+        if fees_per_cost >= self.nonzero_fee_minimum_fpc:
+            min_fee_rate = self.mempool.get_min_fee_rate(cost)
+            if min_fee_rate is None:
+                return False
+            if fees_per_cost > min_fee_rate:
+                return True
         return False
```
AmineKhaldi, line 255:
```suggestion
        if fees_per_cost < self.nonzero_fee_minimum_fpc:
            return False
        min_fee_rate = self.mempool.get_min_fee_rate(cost)
        return min_fee_rate is not None and fees_per_cost > min_fee_rate
```

arvidn: updated

AmineKhaldi (approved):

emlowe (approved):

PR #17093, opened 2023-12-19
Endpoint for vault singleton creation

geoffwalmsley:


Quexington (dismissed):

chia/rpc/wallet_rpc_api.py
```diff
@@ -4496,3 +4504,83 @@ class CRCATApprovePending(Streamable):
...
+        extra_conditions: Tuple[Condition, ...] = tuple(),
+    ) -> EndpointResult:
+        """
+        Create a new vault
+        """
+        assert self.service.wallet_state_manager
+        wallet = self.service.wallet_state_manager.main_wallet
+
+        secp_pk = bytes.fromhex(str(request.get("secp_pk")))
+        entropy = bytes.fromhex(str(request.get("entropy")))
```
Quexington, line 4525: Can we call this "hidden_puzzle_hash" now?

chia/rpc/wallet_rpc_api.py
```diff
@@ -4496,3 +4504,83 @@ class CRCATApprovePending(Streamable):
         return {
             "transactions": [tx.to_json_dict_convenience(self.service.config) for tx in txs],
         }
+
+    ##########################################################################################
+    # VAULT
+    ##########################################################################################
+    @tx_endpoint(push=True)
```
Quexington, line 4511: Let's start making these False by default, the ones with push=True currently are just for backwards compatbility

chia/rpc/wallet_rpc_api.py
```diff
@@ -4496,3 +4504,83 @@ class CRCATApprovePending(Streamable):
...
+            additions=full_spend.additions(),
+            removals=full_spend.removals(),
+            wallet_id=wallet.id(),
+            sent_to=[],
+            trade_id=None,
+            type=uint32(TransactionType.INCOMING_TX.value),
+            name=full_spend.name(),
+            memos=[],
+            valid_times=ConditionValidTimes(),
+        )
```
Quexington, line 4582: I'd like to keep the RPCs as pass-thru as possible so maybe this should all go into WSM

chia/rpc/wallet_rpc_api.py
```diff
@@ -4496,3 +4504,83 @@ class CRCATApprovePending(Streamable):
...
+        # Get xch coin
+        amount = uint64(1)
+        coins = await wallet.select_coins(uint64(amount + fee), tx_config.coin_selection_config)
+
+        # Create singleton launcher
+        origin = coins.copy().pop()
+        launcher_coin = Coin(origin.name(), SINGLETON_LAUNCHER_PUZZLE_HASH, amount)
+
+        vault_full_puzzlehash = create_singleton_puzzle_hash(vault_puzzlehash, launcher_coin.name())
+        announcement_message = Program.to([vault_puzzlehash, amount, bytes(0x80)]).get_tree_hash()
```
Quexington, line 4542: We should hint the secp_pk and hidden_puzzle_hash here so the information is not entirely lost to the chain

chia/rpc/wallet_rpc_api.py
Quexington: Actually, on further thought, let's take this out and come up with something canonical to use for the entropy of the hidden puzzle hash:
```
def calculate_hp_w_index_for_vault(index, hidden_puzzle = DEFAULT_HIDDEN_PUZZLE):
  return Program.to([6, (index, hidden_puzzle)])
```

Quexington: I think we may need to have some way to make the entropy secret but we can add that later.

chia/rpc/wallet_rpc_api.py
```diff
@@ -4496,3 +4504,83 @@ class CRCATApprovePending(Streamable):
...
+            fee,
+            coins,
+            None,
+            origin_id=origin.name(),
+            extra_conditions=(
+                AssertCoinAnnouncement(asserted_id=launcher_coin.name(), asserted_msg=announcement_message),
+            ),
+        )
+
+        genesis_launcher_solution = Program.to([vault_puzzlehash, amount, bytes(0x80)])
```
Quexington, line 4557: Use the calculation you already did above

chia/rpc/wallet_rpc_api.py
```diff
@@ -4496,3 +4504,83 @@ class CRCATApprovePending(Streamable):
...
+        launcher_cs = CoinSpend(launcher_coin, SINGLETON_LAUNCHER_PUZZLE, genesis_launcher_solution)
+        launcher_sb = SpendBundle([launcher_cs], AugSchemeMPL.aggregate([]))
+        assert tx_record.spend_bundle is not None
+        full_spend = SpendBundle.aggregate([tx_record.spend_bundle, launcher_sb])
+
+        vault_record = TransactionRecord(
+            confirmed_at_height=uint32(0),
+            created_at_time=uint64(int(time.time())),
+            amount=uint64(amount),
+            to_puzzle_hash=vault_full_puzzlehash,
```
Quexington, line 4568: Use the inner puzzle hash you sent to here

tests/wallet/vault/test_vault_wallet.py
```diff
@@ -0,0 +1,47 @@
...
+
+    timelock = uint64(1000)
+    entropy = int_to_bytes(101)
+
+    res = await client.vault_create(SECP_PK, entropy, bls_pk, timelock, tx_config=DEFAULT_TX_CONFIG, fee=uint64(10))
+    vault_tx = res[0]
+    assert vault_tx
+
+    eve_coin = [item for item in vault_tx.additions if item not in vault_tx.removals and item.amount == 1][0]
+    assert eve_coin
```
Quexington, line 47: Can we push the transaction and add a `.process_pending_states` call here to make sure the transaction goes through properly?

Quexington (changes requested):

chia/wallet/wallet_state_manager.py
```diff
@@ -2533,3 +2539,71 @@ async def sign_transaction(self, coin_spends: List[CoinSpend]) -> SpendBundle:
...
+
+        [tx_record] = await wallet.generate_signed_transaction(
+            amount,
+            SINGLETON_LAUNCHER_HASH,
+            tx_config,
+            fee,
+            coins,
+            None,
+            origin_id=origin.name(),
+            memos=[secp_pk, hidden_puzzle_hash],
```
Quexington, line 2579: One last nitpick, but these should go in the `genesis_launcher_solution` (where you currently have `None`) that way we don't have to fetch and run the spend before the launcher as well.

Quexington (approved):

wjblanke (approved):
aok

PR #17102, opened 2023-12-20
use rust types for VDFInfo, VDFProof and ClassgroupElement

arvidn:
### Purpose:

This is a step towards making `FullBlock` a rust type, to move enough of the block validation into rust to be able to use threads instead of processes.

The rust counterparts are defined here:
https://github.com/Chia-Network/chia_rs/blob/main/chia-protocol/src/vdf.rs
https://github.com/Chia-Network/chia_rs/blob/main/chia-protocol/src/classgroup.rs

Since these types aren't dataclasses, the `dataclass.replace()` is changed to method `replace()`.

### Current Behavior:

`VDFInfo`, `VDFProof` and `ClasgroupElement` are python types.

### New Behavior:

`VDFInfo`, `VDFProof` and `ClasgroupElement` are native (rust)  types.

### DIFF
--- a/chia/types/blockchain_format/vdf.py
+++ b/chia/types/blockchain_format/vdf.py
@@ -2,20 +2,21 @@ from __future__ import annotations
 
 import logging
 import traceback
-from dataclasses import dataclass
 from enum import IntEnum
 from functools import lru_cache
 from typing import Optional
 
+from chia_rs import VDFInfo, VDFProof
 from chiavdf import create_discriminant, verify_n_wesolowski
 
 from chia.consensus.constants import ConsensusConstants
 from chia.types.blockchain_format.classgroup import ClassgroupElement
 from chia.types.blockchain_format.sized_bytes import bytes32, bytes100
 from chia.util.ints import uint8, uint64
-from chia.util.streamable import Streamable, streamable
 
 log = logging.getLogger(__name__)
+
+__all__ = ["VDFInfo", "VDFProof"]
 
 
 @lru_cache(maxsize=200)
@@ -44,22 +45,6 @@ def verify_vdf(
         discriminant_size,
         witness_type,
     )
-
-
-@streamable
-@dataclass(frozen=True)
-class VDFInfo(Streamable):
-    challenge: bytes32  # Used to generate the discriminant (VDF group)
-    number_of_iterations: uint64
-    output: ClassgroupElement
-
-
-@streamable
-@dataclass(frozen=True)
-class VDFProof(Streamable):
-    witness_type: uint8
-    witness: bytes
-    normalized_to_identity: bool
 
 
 def validate_vdf(
--- a/chia/util/recursive_replace.py
+++ b/chia/util/recursive_replace.py
@@ -7,11 +7,15 @@ from typing import Any
 def recursive_replace(root_obj: Any, replace_str: str, replace_with: Any) -> Any:
     split_str = replace_str.split(".")
     if len(split_str) == 1:
+        # This check is here to support native types (implemented in Rust
+        # in chia_rs) that aren't dataclasses. They instead implement a
+        # replace() method in their python bindings.
         if hasattr(root_obj, "replace"):
             return root_obj.replace(**{split_str[0]: replace_with})
         else:
             return replace(root_obj, **{split_str[0]: replace_with})
     sub_obj = recursive_replace(getattr(root_obj, split_str[0]), ".".join(split_str[1:]), replace_with)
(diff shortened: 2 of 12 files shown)


chia/util/recursive_replace.py
```diff
@@ -1,12 +1,18 @@
...
 
 
 def recursive_replace(root_obj: Any, replace_str: str, replace_with: Any) -> Any:
     split_str = replace_str.split(".")
     if len(split_str) == 1:
-        return replace(root_obj, **{split_str[0]: replace_with})
+        if is_dataclass(root_obj):
+            return replace(root_obj, **{split_str[0]: replace_with})
+        else:
+            return root_obj.replace(**{split_str[0]: replace_with})
```
altendky, line 13: I'm not sure if it is overall better or not, but it would be more flexible to just use the `.replace()` method if present.  Then even dataclasses can choose to override it.  Might end up useful as a transitional step somewhere?  Though, that hazards accidentally adding one for another purpose and accidentally changing the behavior.  Could decorate them with a marker of some sort...  Maybe some part of this rambling is an interesting though...

tests/util/test_recursive_replace.py
```diff
@@ -0,0 +1,112 @@
...
+    a: int
+    b: str
+    c: Optional[TestA]
+
+    def __init__(self, a: int, b: str, c: Optional[TestA]):
+        self.a = a
+        self.b = b
+        self.c = c
+
+    def replace(self, **kwargs: Dict[str, Union[int, str, Optional[TestA]]]) -> TestB:
```
altendky, line 50: Since the type is forced to dict and the key is forced to str, you generally just hint the value.  Or maybe you _have_ to just hint the value.

```suggestion
    def replace(self, **kwargs: Union[int, str, Optional[TestA]]) -> TestB:
```

altendky, line 50: I get that this is kinda meant as a stand-in for rust implementations and as such the actual implementation doesn't matter.  Would probably be good to add a note warning about this fact and how this is not a reference implementation, just a test implementation.

If this isn't the case, I'll give a real review with various suggestions about patterns to use here.

chia/types/blockchain_format/vdf.py
```diff
@@ -2,21 +2,22 @@
...
 
 from chia.consensus.constants import ConsensusConstants
 from chia.types.blockchain_format.classgroup import ClassgroupElement
 from chia.types.blockchain_format.sized_bytes import bytes32, bytes100
 from chia.util.ints import uint8, uint64
-from chia.util.streamable import Streamable, streamable
 
 log = logging.getLogger(__name__)
 
+__all__ = ["VDFInfo", "VDFProof"]
```
altendky, line 19: Will we separately update the imports to get these from `chia_rs` directly?

altendky (changes requested):

chia/util/recursive_replace.py
arvidn: yeah, I did that in an earlier version of this patch. I don't have a strong opinion either way, I changed it to this after reading up on people on the internet expecting `dataclass` to add a member `replace()` and asking about why it doesn't.

tests/util/test_recursive_replace.py
arvidn: oh, I see. I suppose the `Dict` and `str` are implied

chia/types/blockchain_format/vdf.py
arvidn: maybe down the line. I'm trying to make minimal updates to have all types be exported from the same place, to retain some backwards compatibility

tests/util/test_recursive_replace.py
arvidn: it *is* just a way to test the non-dataclass path in `recursive_replace()`. As such, I think it makes sense to keep it minimal and not necessarily robust for real-world use.

chia/util/recursive_replace.py
altendky: I guess folks that want that could add `replace = dataclasses.replace` to the class definition.  I personally am glad that it doesn't take over lots of "regular" methods and sticks to helping out with dunder methods.  Makes it more like a nice way to write a "regular class" than a big ol' framework that takes control of your class interface.  Makes it easier to attach multiple sets of functionality to a single class without them colliding since it mostly exposes the metadata for access as needed.

altendky (approved):

wjblanke (approved):
aok! exciting

PR #17106, opened 2023-12-20
Updated wallet protocol

Rigidity:
### Purpose:

Adds the following features to the wallet protocol:

- Streaming of the initial coin state when syncing the wallet's puzzle hashes. This is to prevent the response exceeding the limit or using up too much memory at once.
- Messages to remove and clear subscriptions later.
- Messages to fetch state without subscribing, for example for getting parent coin information or checking the status of a spend.


benchmarks/coin_state_streaming.py
```diff
@@ -0,0 +1,57 @@
...
+from time import monotonic
+
+from benchmarks.utils import setup_db
+from chia.full_node.coin_store import CoinStore
+from chia.full_node.hint_store import HintStore
+from chia.util.ints import uint32
+from tests.core.full_node.stores.test_coin_store import generate_crs, generate_phs
+
+
+async def run_benchmark(version: int) -> None:
```
arvidn, line 13: would it be very inconvenient to move this into `test_coin_store.py` and mark it as a "benchmark"?
Currently we run tests with `-m "not benchmark"` in CI, and then we run (serially) only the benchmarks.

There is some precedence for adding benchmarks this way, especially in measuring the mempool.
If the test takes 10 minutes, it might not be great to run it in CI however.

I think @altendky may have some plans for the `benchmarks` directory.

benchmarks/subscriptions.py
```diff
@@ -0,0 +1,62 @@
...
+        for i in range(0, 32):
+            if next[i] != 255:
+                next = bytes32(next[:i] + bytes([next[i] + 1]) + next[i + 1 :])
+                break
+            else:
+                next = bytes32(next[:i] + bytes([0]) + next[i + 1 :])
+    return ids
+
+
+def run_subscriptions_benchmark() -> None:
```
arvidn, line 24: The comment above applies to this benchmark as well.

chia/full_node/coin_store.py
```diff
@@ -348,6 +348,98 @@ async def get_coin_states_by_puzzle_hashes(
...
         return coins
 
+    async def stream_coin_states_by_puzzle_hashes(
+        self,
+        puzzle_hashes: Set[bytes32],
+        min_height: uint32,
+        include_spent: bool,
+        include_unspent: bool,
+        include_hints: bool,
+        max_batch_size: int = 50000,
```
arvidn, line 358: when you have this many parameters of very similar type (especially the `bools`) I think it's really important to require these parameters be named. you can do that by adding `*` to the parameter list before the first names parameter. I would suggest before `min_height`

chia/full_node/coin_store.py
```diff
@@ -348,6 +348,98 @@ async def get_coin_states_by_puzzle_hashes(
 
         return coins
 
+    async def stream_coin_states_by_puzzle_hashes(
+        self,
+        puzzle_hashes: Set[bytes32],
+        min_height: uint32,
```
arvidn, line 354:
```suggestion
        puzzle_hashes: Set[bytes32],
        *,
        min_height: uint32,
```

chia/full_node/full_node_api.py
```diff
@@ -1650,5 +1642,157 @@ def get_fee_estimates(est: FeeEstimatorInterface, req_times: List[uint64]) -> Li
...
+        assert peak is not None
+
+        try:
+            async with self.full_node.stream_coin_state_sem.acquire():
+                stream = self.full_node.coin_store.stream_coin_states_by_puzzle_hashes(
+                    set(request.puzzle_hashes),
+                    request.min_height,
+                    request.filters.include_spent,
+                    request.filters.include_unspent,
+                    request.filters.include_hinted,
```
arvidn, line 1743: this won't work if you "blindly" accept my change above, so you should probably make the change yourself and make everything work before updating the PR

arvidn (commented):
looks good so far. We should probably involve a few other people to review the protocol before setting it in stone.

One more thing; we have a test that ensures that none of our types that are used to define any protocol are changed, breaking the protocol.

This test is updated by:

1. adding your new message types to: `tests/util/build_network_protocol_files.py`
2. Updating `tests/util/network_protocol_data.py` to include the new message objects with dummy values (but ideally random values that are unlikely to accidentally be encoded the same if a structure is changed)
3. run `tests/util/build_network_protocol_files.py` to update the auto-generated files

Rigidity:
Closing as this has once again grown too large in scope. Take 3 :smile:

PR #17115, opened 2023-12-20
Extract get_trade_and_status to be reused instead of reimplemented

AmineKhaldi:


arvidn:
that line missing coverage should probably be annotated, as it looks like it represents a test failure

altendky (approved):

emlowe (approved):

PR #17119, opened 2023-12-20
Explicitly name fields to improve clarity and avoid mistakes

AmineKhaldi:


altendky (dismissed):

emlowe (dismissed):

prozacchiwawa (dismissed):

altendky (approved):

arvidn (approved):
I take it we can't update the constructor signatures to require named variables, right?

altendky:
We could presumably.  Breaking changes and all though.  We still need to work on that pylint checker for this...

PR #17139, opened 2023-12-21
Mempool multifetch

arvidn:
This PR is best reviewed one commit at a time. But the second commit contains the bulk of the changes.

### checkpoint merge

Reverse this patch when merging to `main`:

```
diff --git a/tests/core/mempool/test_mempool_manager.py b/tests/core/mempool/test_mempool_manager.py
index ce187fe7f..7a7d32899 100644
--- a/tests/core/mempool/test_mempool_manager.py
+++ b/tests/core/mempool/test_mempool_manager.py
@@ -1753,8 +1753,8 @@ async def test_mempool_timelocks(cond1: List[object], cond2: List[object], expec
     )
 
     coin_spends = [
-        make_spend(coins[0], IDENTITY_PUZZLE, Program.to([cond1])),
-        make_spend(coins[1], IDENTITY_PUZZLE, Program.to([cond2])),
+        CoinSpend(coins[0], IDENTITY_PUZZLE, Program.to([cond1])),
+        CoinSpend(coins[1], IDENTITY_PUZZLE, Program.to([cond2])),
     ]
 
     bundle = SpendBundle(coin_spends, G2Element())
```

### Purpose:

This patch improves the disk I/O performance of the mempool by fetching coin records in batches instead of one at a time. It primarily reduces CPU usage when adding spend bundles to the mempool, which is especially critical when updating the mempool through the new-block slow-path.

There are two key changes in this PR:

#### get_coin_records

The `MempoolManager` and `Mempool` objects are passed a function from the `CoinStore`, `get_coin_record`, that looks up a single coin record. The first change is to instead of pass in `get_coin_records`. This function fetches multiple coin records in a single query.

This change requires updates a lot of tests, which ends up being the build of the changes. The core changes, however, are in `def validate_spend_bundle()` in `mempool_manager.py`.

#### new-peak slow-path

The second change (3rd commit) is to, when entering the slow-path to update the peak; fetch *all* spent coins in a single query, stick the `CoinRecords` in a dictionary. Pass in a local version of `get_coin_records` that only looks up coins from the local dictionary to `add_spend_bundle()`, when adding all the transactions into the new mempool.

This turns thousands of individual `CoinRecord` lookups into a single query.

### Current Behavior:

`Mempool` takes a function that can fetch one `CoinRecord`. Coin record are looked up one at a time when validating transactions.

### New Behavior:

`Mempool` takes a function that can fetch multiple `CoinRecord`s. Coin record are looked up in a single query.
When updating the peak through the slow-path, *all* coin spends are looked up in a single query, to speed up re-adding the transactions to the new mempool instance.

### Testing Notes:

This was tested on IPS machines. The wall-clock time was not improved as much as the CPU usage:

| patch | mempool size | wall-clock time | time per spend |
| --- | --- | --- | ---- |
| before | 3551 | 44.42s | 12.5ms |
| after | 7069 | 67.71 | 9.57ms |

#### Before

chia-hotspot-3532-3574

#### After

chia-hotspot-13804-13861

### DIFF
--- a/chia/full_node/mempool_manager.py
+++ b/chia/full_node/mempool_manager.py
@@ -638,9 +638,35 @@ class MempoolManager:
             old_pool = self.mempool
             self.mempool = Mempool(old_pool.mempool_info, old_pool.fee_estimator)
             self.seen_bundle_hashes = {}
+
+            # in order to make this a bit quicker, we look-up all the spends in
+            # a single query, rather than one at a time.
+            coin_records: Dict[bytes32, CoinRecord] = {}
+
+            removals: Set[bytes32] = set()
+            for item in old_pool.all_items():
+                for s in item.spend_bundle.coin_spends:
+                    removals.add(s.coin.name())
+
+            for record in await self.get_coin_records(removals):
+                name = record.coin.name()
+                coin_records[name] = record
+
+            async def local_get_coin_records(names: Collection[bytes32]) -> List[CoinRecord]:
+                ret: List[CoinRecord] = []
+                for name in names:
+                    r = coin_records.get(name)
+                    if r is not None:
+                        ret.append(r)
+                return ret
+
             for item in old_pool.all_items():
                 _, result, err = await self.add_spend_bundle(
-                    item.spend_bundle, item.npc_result, item.spend_bundle_name, item.height_added_to_mempool
+                    item.spend_bundle,
+                    item.npc_result,
+                    item.spend_bundle_name,
+                    item.height_added_to_mempool,
+                    local_get_coin_records,
                 )
                 # Only add to `seen` if inclusion worked, so it can be resubmitted in case of a reorg
                 if result == MempoolInclusionStatus.SUCCESS:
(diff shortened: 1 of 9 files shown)


chia/full_node/mempool_manager.py
```diff
@@ -447,10 +462,10 @@ async def validate_spend_bundle(
                     False,
                     self.peak.timestamp,
                 )
-
-            assert removal_record is not None
+                removal_record_dict[name] = removal_record
+            else:
```
wjblanke, line 466: Just curious what this new code path does. we didn't have it in the old code and it doesn't seem related to grouping the query

arvidn: this for-loop used to query one spend at a time (at the top of the loop). This was split into:

1. query all spends in one query
2. build a dict of the responses
3. this for-loop

So, some code moved around a bit.

Previously we would have the current `CoinRecord` here, named `removal_record`, because that's what we looked up (either from the DB or from `additions_dict`. We would then populate the `removal_record_dict` (line 453 in the previous version).

In the new version, we've already built the dictionary (in the first for-loop, line 443 in the new version), so now we need to go the other way, and set `removal_record` in order for the last line to update `removal_amount`.

wjblanke (dismissed):
aok

wjblanke (dismissed):
aok

emlowe (approved):

PR #17143, opened 2023-12-22
Don't verify right after signing, and as a result don't re-verify the aggregated version, in sign_coin_spends

AmineKhaldi:
This (and the branch name) is dedicated to @emlowe


emlowe (approved):
Much love to my good friend Amine ❤️

PR #17149, opened 2023-12-22
Annotate test_offer_lifecycle.py

AmineKhaldi:
Extracted the first commit into PR #17189 because we squash PRs and the resulting diff with both commits is not optimal.

### DIFF
--- a/tests/wallet/cat_wallet/test_offer_lifecycle.py
+++ b/tests/wallet/cat_wallet/test_offer_lifecycle.py
@@ -1,18 +1,18 @@
 from __future__ import annotations
 
 from dataclasses import replace
-from typing import Any, Dict, List, Optional
+from typing import Dict, List, Optional, cast
 
 import pytest
 from chia_rs import G2Element
 
-from chia.clvm.spend_sim import sim_and_client
+from chia.clvm.spend_sim import CostLogger, SimClient, SpendSim, sim_and_client
 from chia.types.announcement import Announcement
 from chia.types.blockchain_format.coin import Coin
 from chia.types.blockchain_format.program import Program
 from chia.types.blockchain_format.serialized_program import SerializedProgram
 from chia.types.blockchain_format.sized_bytes import bytes32
-from chia.types.coin_spend import CoinSpend, make_spend
+from chia.types.coin_spend import make_spend
 from chia.types.mempool_inclusion_status import MempoolInclusionStatus
 from chia.types.spend_bundle import SpendBundle
 from chia.util.ints import uint64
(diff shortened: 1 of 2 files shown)


tests/wallet/cat_wallet/test_offer_lifecycle.py
```diff
@@ -47,24 +48,22 @@ def str_to_cat_hash(tail_str: str) -> bytes32:
...
 ) -> Dict[Optional[str], List[Coin]]:
     await sim.farm_block(acs_ph)
-    parent_coin: Coin = [cr.coin for cr in await sim_client.get_coin_records_by_puzzle_hash(acs_ph)][0]
+    parent_coin = [cr.coin for cr in await sim_client.get_coin_records_by_puzzle_hash(acs_ph)][0]
 
     # We need to gather a list of initial coins to create as well as spends that do the eve spend for every CAT
-    payments: List[Payment] = []
-    cat_bundles: List[SpendBundle] = []
+    payments = []
+    cat_bundles = []
```
altendky, line 58: I think that being explicit in this context is often good and makes it easier to read since the append etc require scanning further down to figure out what this is going to become.

tests/wallet/cat_wallet/test_offer_lifecycle.py
```diff
@@ -297,15 +254,16 @@ async def test_complex_offer(cost_logger):
         assert new_offer.is_valid()
 
         # Test preventing TAIL from running during exchange
-        blue_cat_puz: Program = construct_cat_puzzle(CAT_MOD, str_to_tail_hash("blue"), OFFER_MOD)
-        blue_spend: CoinSpend = make_spend(
-            Coin(bytes32(32), blue_cat_puz.get_tree_hash(), uint64(0)),
+        blue_cat_puz = construct_cat_puzzle(CAT_MOD, str_to_tail_hash("blue"), OFFER_MOD)
+        random_hash = bytes32([0] * 32)
```
altendky, line 258: There's `bytes32.random()`, and if you use the `seeded_random: Random` fixture then you can `bytes32.random(r=seeded_random)` for deterministic random `bytes32`s.  But sure, `bytes32(32)` is also valid for just zeros.

altendky (dismissed):
Feel free to close discussions after reading them.

tests/wallet/cat_wallet/test_offer_lifecycle.py
AmineKhaldi: Yep, I tend to opt for this only when the current block itself makes it clear what we're appending (in this example `payments.append(Payment(...))`).

In general, I found so many instances where we declare the wrong types, or shadow values later on but with different types, so whenever the type is clear from its immediate use I favor that.

tests/wallet/cat_wallet/test_offer_lifecycle.py
AmineKhaldi: Thanks. `bytes32(32)` is currently problematic with `mypy` and so I went with this in consistency with that code.

altendky (approved):

emlowe (approved):

PR #17150, opened 2023-12-22
Remove redundant bytes32 import in test_nft_wallet_rpc_update_metadata

AmineKhaldi:


arvidn (approved):

arvidn:
normally our linters warn about things like this

altendky:
I'm guessing the difference is that here the import is inside a function rather than in global scope.

PR #17152, opened 2023-12-22
Remove debug prints from DID's test_creation_from_coin_spend and test_create_did_with_recovery_list

AmineKhaldi:


arvidn:
why should these be removed?

AmineKhaldi:
> why should these be removed?

When you run tests locally, these prints result in massive dumps that completely mess up your terminal output without providing anything informative whatsoever. They're clearly meant as temporary prints in order to create the nearby test asserts but were forgotten afterwards.

arvidn (approved):

PR #17153, opened 2023-12-22
Annotate did_wallet_puzzles.py

AmineKhaldi:


altendky (approved):

arvidn (approved):

Starttoaster:
The untested line that the coverage report appears to be arguing over just ignores an unused function return value, where the overall line existed already. Ignoring the untested line.

The coverage report didn't make it to this PR because of an authorization error as this is ran from a fork.

PR #17155, opened 2023-12-22
Performance tweaks

wjblanke:
### Purpose:

Performance tweaks

### Current Behavior:

config.yaml peer count init'd to 80
db_sync defaults to FULL

### New Behavior:

config.yaml peer count init'd to 40
db_sync defaults to NORMAL

### DIFF
--- a/chia/util/initial-config.yaml
+++ b/chia/util/initial-config.yaml
@@ -408,7 +408,7 @@ full_node:
   # How long to wait for a peer connection
   peer_connect_timeout: 30
   # Accept peers until this number of connections
-  target_peer_count: 80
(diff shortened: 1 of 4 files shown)


arvidn (approved):

chia/util/initial-config.yaml
```diff
@@ -408,7 +408,7 @@ full_node:
   # How long to wait for a peer connection
   peer_connect_timeout: 30
   # Accept peers until this number of connections
-  target_peer_count: 80
+  target_peer_count: 40
```
josinaldobarbosa, line 411: This should be changed automatically when updating, most users won't look at this

PR #17158, opened 2023-12-23
Check dependencies to avoid a useless sudo on Archlinux

Ealrann:
### Purpose:

Prevent a useless call to `sudo` during this installation of Chia on Archlinux, by checking if the dependencies are not already installed. 
The point is to make the installation easier: 
* `sudo` forces the user to type its password (event if there is nothing to do).
* Some people prefer to use `su` and don't setup `sudo`. In this case, the installation of chia is impossible without modifying the `instal.sh`. (it's what I do for years)

### Old Behavior:

Calling `sudo` every time to install the dependencies. 

### New Behavior:

Verify if `git` and `openssl` are installed, then call `sudo pacman -S git openssl` only if needed.

### Testing Notes:

I wanted to keep the existing code, so I simply added a if statement. But in fact, we can simplify the code for Archlinux: 
no dependencies needs to be installed (at least line 213), because: 
* openssl is a base package of archlinux  (base => coreutils => openssl)
* git is already called line 71 (so no point to install it line 213, it's too late).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -210,7 +210,9 @@ elif [ "$(uname)" = "Linux" ]; then
     echo "Installing on Arch Linux."
     case $(uname -m) in
       x86_64|aarch64)
-        sudo pacman ${PACMAN_AUTOMATED} -S --needed git openssl
+        if ! pacman -Qs "^git$" > /dev/null || ! pacman -Qs "^openssl$" > /dev/null ; then
+          sudo pacman ${PACMAN_AUTOMATED} -S --needed git openssl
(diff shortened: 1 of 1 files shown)


hoffmang9:
Thanks for this. Can you sign your commit or open a new PR with a signed commit to start it?

Ealrann:
Oh sorry, I've just corrected it

hoffmang9 (approved):
lgtm with thannks!

emlowe (approved):

PR #17166, opened 2023-12-26
Spin slower `send_transaction()`

arvidn:
### Purpose:

we currenly spin-wait for hearing back whether a transaction made it into the mempool or not (in FullNodeAPI.send_transaction()). We currently spin in a very tight loop, only sleeping 10ms, which appears to hog a significant amount of CPU. This patch increases the sleep to spin slower.

### Current Behavior:

Sleep 10ms in the spin-wait loop.

### New Behavior:

Sleep 500ms in the spin-wait loop.

### Testing Notes:

The main evidence suggesting that this is a problem is the following profile, collected while validating a block:


wjblanke (approved):
aok

PR #17167, opened 2023-12-26
remove redundant cost field in NPCResults

arvidn:
### Purpose:

Simplify our data structure to not contain the cost twice.

This makes `NPCResult` just an error or the value.

### Current Behavior:

the field `cost` in NPCResult is always the same as `NPCResult.conds.cost`

### New Behavior:

There is no duplication of the cost in two fields.

### DIFF
--- a/chia/full_node/mempool_check_conditions.py
+++ b/chia/full_node/mempool_check_conditions.py
@@ -93,13 +93,13 @@ def get_name_puzzle_conditions(
         err, result = run_block(bytes(generator.program), block_args, max_cost, flags)
         assert (err is None) != (result is None)
         if err is not None:
-            return NPCResult(uint16(err), None, uint64(0))
+            return NPCResult(uint16(err), None)
         else:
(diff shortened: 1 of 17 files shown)


chia/full_node/mempool_check_conditions.py
```diff
@@ -93,13 +93,13 @@ def get_name_puzzle_conditions(
         err, result = run_block(bytes(generator.program), block_args, max_cost, flags)
         assert (err is None) != (result is None)
         if err is not None:
-            return NPCResult(uint16(err), None, uint64(0))
+            return NPCResult(uint16(err), None)
         else:
             assert result is not None
-            return NPCResult(None, result, uint64(result.cost))
```
arvidn, line 99: as can be seen here, the `cost` field is set to `result.cost`, duplicating the value

AmineKhaldi (approved):

emlowe (approved):

PR #17168, opened 2023-12-26
add feature to profile just the block validation

arvidn:
This exposes what is actually going on while we validate blocks. Sometimes the block validation times are affected by other tasks that hog the CPU, blocking the main co-routine from being scheduled.

### Purpose:

When understanding what's causing spikes in block validation times, having a profile focused only on the time of validating the block can be quite helpful, especially when pre_validation nor post_processing takes very long.

It can help identify co-routines that run for too long (i.e. perform expensive work in the main thread), hogging the scheduler and delaying other tasks.

### Current Behavior:

The only available profiler is for the whole full node task.

### New Behavior:

There's now also a profiler for just block validation. If the validation takes longer than 2 seconds, the profile is saved for later analysis.

### Testing Notes:

Example profiles from `main`.

(parsing the `SubEpochChallengeSegments` is slow)

(serializing the weight proof is slow)

(parsing `BlockRecord` is slow)

(`sleep(0.01)` in a loop is slow)

(merkle root validation and BLS signatures can still be expensive)

(saving peers is slow. This caused an 11.9 seconds block validation)

(request_additions is slow)

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -94,7 +94,7 @@ from chia.util.ints import uint8, uint32, uint64, uint128
 from chia.util.limited_semaphore import LimitedSemaphore
 from chia.util.log_exceptions import log_exceptions
 from chia.util.path import path_from_root
-from chia.util.profiler import mem_profile_task, profile_task
+from chia.util.profiler import enable_profiler, mem_profile_task, profile_task
 from chia.util.safe_cancel_task import cancel_task_safe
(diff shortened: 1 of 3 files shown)


arvidn:
@altendky Here's an example of the `coverage-diff` tag not being removed automatically

AmineKhaldi (approved):

wjblanke (approved):
aok

PR #17189, opened 2024-01-02
Remove no longer needed TestOfferLifecycle class

AmineKhaldi:
This is better reviewed with `Ignore whitespace`.


arvidn (approved):
best reviewed with whitespaces ignored.

PR #17196, opened 2024-01-02
Add keychain support and placeholder for vault wallet

geoffwalmsley:
This PR adds the basic XCH functionality for vault wallets using SECP keys with BLS recovery, along with CLI and RPC endpoints.
Currently the vault secret key is passed to wallet via config setting, and this will be removed in a future PR.


chia/util/keychain.py
```diff
@@ -336,19 +345,28 @@ def _get_key_data(self, index: int, include_secrets: bool = True) -> KeyData:
         str_bytes = bytes.fromhex(read_str)
 
         pk_bytes: bytes = str_bytes[: G1Element.SIZE]
-        observation_root: ObservationRoot = G1Element.from_bytes(pk_bytes)
-        fingerprint = observation_root.get_fingerprint()
-        if len(str_bytes) == G1Element.SIZE + 32:
-            entropy = str_bytes[G1Element.SIZE : G1Element.SIZE + 32]
-        else:
+        if len(pk_bytes) == 32:
```
Quexington, line 348: since pk_bytes is generated from `str_bytes[: G1Element.SIZE]` isn't this `if` impossible to hit?

chia/wallet/puzzles/vault_p2_recovery.clsp
```diff
@@ -47,19 +47,22 @@
...
+          (create_recovery_puzzlehash
+            P2_1_OF_N_MOD_HASH
+            FINISH_RECOVERY_MOD_HASH
+            P2_SECP_PUZZLEHASH
+            TIMELOCK
+            recovery_conditions
+          )
+          my_amount
+        )
+        (list ASSERT_MY_AMOUNT my_amount)
```
Quexington, line 63: Not a fan of this change I don't think, if someone doesn't want this path to work, we can just not put this path as a viable one to execute within the puzzle.

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,317 @@
...
+        self = Vault()
+        self.wallet_state_manager = wallet_state_manager
+        self.wallet_info = info
+        self.wallet_id = info.id
+        self.log = logging.getLogger(name)
+        return self
+
+    async def get_new_puzzle(self) -> Program:
+        dr = await self.wallet_state_manager.get_unused_derivation_record(self.id())
+        puzzle = construct_p2_delegated_secp(dr.pubkey, DEFAULT_CONSTANTS.GENESIS_CHALLENGE, dr.puzzle_hash)
```
Quexington, line 71: Don't use `DEFAULT_CONSTANTS`, use `WalletStateManager.constants`

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,317 @@
...
+    async def match_hinted_coin(self, coin: Coin, hint: bytes32) -> bool:
+        raise NotImplementedError("vault wallet")
+
+    def handle_own_derivation(self) -> bool:
+        return True
+
+    def get_recovery_info(self) -> Tuple[Optional[G1Element], Optional[uint64]]:
+        if self.vault_info.is_recoverable:
+            return self.recovery_info.bls_pk, self.recovery_info.timelock
+        return None, None
```
Quexington, line 166: Are these two things ever in the state where one is option and one isn't?  If not, I think maybe make this `Optional[Tuple[G1Element, uint64]]` instead.

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,317 @@
...
+            self.vault_info.pubkey,
+            DEFAULT_CONSTANTS.GENESIS_CHALLENGE,
+            self.vault_info.hidden_puzzle_hash,
+        ).get_tree_hash()
+
+        recovery_puzzle = get_recovery_puzzle(
+            secp_puzzle_hash, self.recovery_info.bls_pk, self.recovery_info.timelock, amount
+        )
+        recovery_puzzle_hash = recovery_puzzle.get_tree_hash()
+        # TODO: Once we have get_wallet_balance we can use it for the amount field in the solution
```
Quexington, line 220: I feel like the amount of the vault will always be one in standard operation.  I'm not saying hardcode 1 here, but use `vault_info.coin` maybe and I think `get_wallet_balance` is irrelevant.

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,317 @@
...
+        recovery_finish_puzzle = get_recovery_finish_puzzle(
+            self.recovery_info.bls_pk, self.recovery_info.timelock, amount
+        )
+        recovery_finish_solution = Program.to([])
+        recovery_solution = get_vault_inner_solution(recovery_finish_puzzle, recovery_finish_solution, proof)
+        lineage = LineageProof(self.vault_info.coin.name(), inner_puzzle.get_tree_hash(), amount)
+        full_recovery_solution = get_vault_full_solution(lineage, uint64(1), recovery_solution)
+        finish_spend = CoinSpend(recovery_coin, full_recovery_puzzle, full_recovery_solution)
+
+        return recovery_spend, finish_spend
```
Quexington, line 249: Always return transaction records from endpoints that make transactions

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,317 @@
...
+        vault_info = dataclasses.replace(vault_info, coin=coin_state.coin)
+        await self.save_info(vault_info)
+        await self.wallet_state_manager.create_more_puzzle_hashes()
+
+    async def save_info(self, vault_info: VaultInfo) -> None:
+        self.vault_info = vault_info
+        current_info = self.wallet_info
+        data_str = json.dumps(vault_info.to_json_dict())
+        wallet_info = WalletInfo(current_info.id, current_info.name, current_info.type, data_str)
+        self.wallet_info = wallet_info
```
Quexington, line 315: Let's not use the wallet info store here.  Maybe the singleton table used for DAOs in combination with another store for the vault wallet.

tests/wallet/vault/test_vault_wallet.py
```diff
@@ -1,19 +1,82 @@
...
+    wallet_environments: WalletTestFramework, monkeypatch: pytest.MonkeyPatch, with_recovery: bool
+) -> None:
+    def get_main_wallet_driver(self: WalletStateManager, observation_root: ObservationRoot) -> Type[MainWalletProtocol]:
+        return Vault
+
+    monkeypatch.setattr(
+        WalletStateManager,
+        "get_main_wallet_driver",
+        types.MethodType(get_main_wallet_driver, WalletStateManager),
+    )
```
Quexington, line 34: Don't monkeypatch this like in the ACS test.  You can actually change the function in WSM to properly return Vault based on the observation_root.  I only monkeypatched it in the ACS test because that wallet shouldn't be part of the main codebase.

tests/wallet/vault/test_vault_wallet.py
```diff
@@ -1,19 +1,82 @@
...
+
+    for env in wallet_environments.environments:
+        SECP_SK = SigningKey.generate(curve=NIST256p, hashfunc=sha256)
```
Quexington, line 37: Random keys are a headache waiting to happen, you should use a deterministic key (is this deterministic? I assumed not but am unfamiliar with the library.)

chia/wallet/vault/vault_info.py
```diff
@@ -0,0 +1,30 @@
...
+from chia.util.ints import uint64
+from chia.util.streamable import Streamable, streamable
+from chia.wallet.lineage_proof import LineageProof
+
+
+@streamable
+@dataclass(frozen=True)
+class VaultInfo(Streamable):
+    coin: Coin
+    launcher_id: bytes32
```
Quexington, line 18: Our whole wallet service is parametrized by this value when running in vault mode. I don't think this property is necessary.

chia/wallet/vault/vault_info.py
```diff
@@ -0,0 +1,30 @@
...
+    pubkey: bytes
+    hidden_puzzle_hash: bytes32
+    inner_puzzle_hash: bytes32
+    is_recoverable: bool
+    launcher_coin_id: bytes32
```
Quexington, line 23: Isn't this a duplicate of above?

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,606 @@
...
+from chia.wallet.wallet_protocol import GSTOptionalArgs
+
+
+class Vault(Wallet):
+    @staticmethod
+    async def create(
+        wallet_state_manager: Any,
+        info: WalletInfo,
+        name: str = __name__,
+    ) -> Vault:
```
Quexington, line 70: I see `.vault_info` in other parts of the program, but I don't see it being set here.  Could we potentially make this wallet a dataclass so that all of the properties unique to it are at the top and are always set?

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,606 @@
...
+        tx_config: TXConfig,
+        fee: uint64 = uint64(0),
+        coins: Optional[Set[Coin]] = None,
+        primaries: Optional[List[Payment]] = None,
+        memos: Optional[List[bytes]] = None,
+        puzzle_decorator_override: Optional[List[Dict[str, Any]]] = None,
+        extra_conditions: Tuple[Condition, ...] = tuple(),
+        **kwargs: Unpack[GSTOptionalArgs],
+    ) -> List[TransactionRecord]:
+        raise NotImplementedError("vault wallet")
```
Quexington, line 107: I don't know if this was intentional but we definitely want to implement this even if it just calls `generate_signed_vault_spend` under the hood.

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,606 @@
...
+            self.vault_info.coin.name(),
+            self.wallet_state_manager.constants.GENESIS_CHALLENGE,
+            self.vault_info.hidden_puzzle_hash,
+        )
+
+        return message_to_sign, delegated_puzzle, delegated_solution
+
+    async def generate_signed_vault_spend(
+        self,
+        signed_message: bytes,
```
Quexington, line 203: We don't actually want the paradigm here of some special value passed to some special method.  The paradigm for all wallets now should be the following:
* Call `generate_signed_transaction` (poorly named now it doesn't actually sign anymore) to generate an unsigned transaction for the user
* User calls `gather_signing_info` RPC.  See the following wallet state manager methods to see what you have to implement: https://github.com/Chia-Network/chia-blockchain/blob/603cf479d2b5f14c60bff700694d052c3237142a/chia/wallet/wallet_state_manager.py#L2616-L2650 (it looks like you'll have to extract out the signing target gathering as well)
* User takes signing instructions to signer.  Our wallet is also a signer and in this context the user calls `execute_signing_instructions`.
* User takes signatures and unsigned transaction from previous steps and calls `apply_signatures` RPC
* User takes newly signed transaction and pushes to network

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,606 @@
...
+    async def sum_hint_for_pubkey(self, pk: bytes) -> Optional[SumHint]:
+        raise NotImplementedError("vault wallet")
+
+    def make_solution(
+        self,
+        primaries: List[Payment],
+        conditions: Tuple[Condition, ...] = tuple(),
+        fee: uint64 = uint64(0),
+    ) -> Program:
+        raise NotImplementedError("vault wallet")
```
Quexington, line 320: This will have to be implemented for the sake of other wallets that are building spends with this as an inner puzzle.

chia/wallet/wallet_state_manager.py
```diff
@@ -797,6 +805,14 @@ async def determine_coin_type(
 
         uncurried = uncurry_puzzle(puzzle)
 
+        from chia.wallet.vault.vault_drivers import match_vault_puzzle
+
+        vault_check = match_vault_puzzle(uncurried.mod, uncurried.args)
+        if vault_check:
+            await self.handle_vault(puzzle, coin_spend, coin_state)
+            # Return None since we don't want to add the vault singleton to the normal wallet coin store
+            return None, None
```
Quexington, line 814: What's the reason for this?  I wouldn't assume adding the seen coin as a general coin record would hurt much and a surprising amount of our sync relies on the state of that table.

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,780 @@
...
+        primaries: Optional[List[Payment]] = None,
+        memos: Optional[List[bytes]] = None,
+        puzzle_decorator_override: Optional[List[Dict[str, Any]]] = None,
+        extra_conditions: Tuple[Condition, ...] = tuple(),
+        **kwargs: Unpack[GSTOptionalArgs],
+    ) -> List[TransactionRecord]:
+        """
+        Creates Un-signed transactions to be passed into signer.
+        """
+        negative_change_allowed: bool = kwargs.get("negative_change_allowed", False)
```
Quexington, line 115: Seems like this is unused?

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,780 @@
...
+        coins = {spend.coin for spend in p2_singleton_spends}
+        spend_value = sum([coin.amount for coin in coins])
+        change = spend_value - total_amount
+        assert change >= 0
+        if change > 0:
+            change_puzzle_hash: bytes32 = get_p2_singleton_puzzle_hash(self.vault_info.launcher_id)
+            primaries.append(Payment(change_puzzle_hash, uint64(change)))
+
+        conditions = [primary.as_condition() for primary in primaries]
+        next_puzzle_hash = await self.get_new_puzzlehash()
```
Quexington, line 219: We need to take tx_config.reuse_puzhash into account here.

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,780 @@
...
+        # TODO: should the vault inner puz create this condition?
```
Quexington, line 220: Is this TODO still relevant?  It seems like it is?

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,780 @@
...
+
+    def puzzle_hash_for_pk(self, pubkey: G1Element) -> bytes32:
+        raise ValueError("This won't work")
+
+    def require_derivation_paths(self) -> bool:
+        if getattr(self, "vault_info", None):
+            return True
+        return False
+
+    async def match_hinted_coin(self, coin: Coin, hint: bytes32) -> bool:
```
Quexington, line 508: This should be implemented I think, it'll be important for offers I believe.

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,780 @@
...
+            timelock,
+        )
+        record = DerivationRecord(
+            uint32(index), inner_puzzle_hash, self.vault_info.pubkey, self.type(), self.id(), False
+        )
+        return [record]
+
+    async def create_recovery_spends(self) -> List[TransactionRecord]:
+        """
+        Returns two spendbundles
```
Quexington, line 558: ~spendbundles~ tx records

chia/wallet/wallet_state_manager.py
```diff
@@ -797,6 +803,12 @@ async def determine_coin_type(
 
         uncurried = uncurry_puzzle(puzzle)
 
+        from chia.wallet.vault.vault_drivers import match_vault_puzzle
```
Quexington, line 806: Can we import this at the top?

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,691 @@
...
+class Vault(Wallet):
+    _vault_info: Optional[VaultInfo] = None
+
+    @property
+    def vault_info(self) -> VaultInfo:
+        if self._vault_info is None:
+            raise ValueError("VaultInfo is not set")
+        return self._vault_info
+
+    @vault_info.setter
```
Quexington, line 79: I don't know if we want this ergonomic, I think the vault info should probably only be messed with by this class.

chia/util/keychain.py
geoffwalmsley: if str_bytes is the secp case then it's only 32 bytes and `str_bytes[: G1Element.size]` will just return those 32 bytes. So the `if` is reached as expected. Could also take len(str_bytes) if you prefer though?

chia/wallet/vault/vault_info.py
geoffwalmsley: I removed the launcher_coin_id field. It was a bit confusing because we'd termed the pk as the launcher early on in development. Have since removed this, so what was launcher_coin_id is now launcher_id. Perhaps it should have a different name though for clarity?

chia/wallet/vault/vault_wallet.py
geoffwalmsley: Do you mean to just set `self._vault_info` directly from `save_info`?

chia/wallet/vault/vault_wallet.py
geoffwalmsley: `negative_change_allowed` is a required field for `generate_unsigned_tx`, so I left this here. I can just make it `False` in the call to `generate_unsigned_tx`

Quexington: It's an optional kwargs right?  So you don't have to get it and store it in a variable no?

chia/wallet/vault/vault_info.py
Quexington: I'm still a bit confused, can anything in the vault wallet that does `self.vault_info.launcher_id` not equally do `self.wallet_state_manager.observation_root`?

geoffwalmsley: yeah, it's confusing... `vault_info.launcher_id` refers to the vault singleton's launcher coin id. But `vault_root.launcher_id` refers to the vault's pubkey. We need `vault_info.launcher_id` in a few places, and can use the observation root elsewhere. So might be worth renaming `vault_info.launcher_id` to `singleton_launcher_id` or something?

Quexington: Oh this is where I'm misunderstanding, I thought the observation root was the vault singleton ID, that's what I was expecting it to be.  We get the pubkey from the singleton after we have synced it from the users initial piece of information which is the launcher ID.

chia/util/keychain.py
Quexington: Ohhh I see, I didn't know it worked that way.  I think it would be more clear to just switch on the length of the str_bytes yeah.  If 32, it's a launcher ID, if 48, its a pubkey only, if 80, it's a pubkey+privatekey

chia/wallet/vault/vault_wallet.py
Quexington: Yeah exactly.  We don't want to give the impression that things external to the class can set that attribute.

chia/wallet/vault/vault_info.py
geoffwalmsley: oh yeah you're right, not sure what I was thinking there. `vault_root.launcher_id` is the launcher coin ID. I noticed a mistake in the test setup that it was taking the eve coin as the launcher, so fixed that. And added a property for launcher_id to get pulled from the observation root. Sorry for the confusion :)

chia/wallet/puzzles/vault_p2_recovery.clsp
```diff
@@ -62,5 +62,4 @@
     (list ASSERT_MY_AMOUNT my_amount)
```
freddiecoleman, line 62: The singleton top layer already puts an ASSERT_MY_AMOUNT condition so we end up with a redundant assertion.

Quexington: Yes, but unfortunately, we don't know from the inner puzzle here that the singleton has already done that.  We are being passed an amount in the solution and while someone may tell the truth at the singleton layer, they may lie to us at this layer, so we have to assert the amount all the same.

chia/wallet/puzzles/p2_singleton_via_delegated_puzzle_safe.clsp
```diff
@@ -27,7 +27,10 @@
     )
     (c
       (list ASSERT_MY_COIN_ID my_id)
-      (a delegated_puzzle delegated_solution)
+      (c
+        (list CREATE_COIN_ANNOUNCEMENT '$')
```
Quexington, line 31: The `'$'` is no longer necessary.  We used to use it because we didn't know how to use python very well but now we've learned better.

chia/wallet/vault/vault_wallet.py
```diff
@@ -0,0 +1,773 @@
...
+
+    async def update_vault_store(self, vault_info: VaultInfo, coin_spend: CoinSpend) -> None:
+        custom_data = bytes(
+            json.dumps(
+                {
+                    "vault_info": vault_info.to_json_dict(),
+                }
+            ),
+            "utf-8",
+        )
```
Quexington, line 772: Is there a reason we're doing this as JSON rather than as the streamable hex?

chia/wallet/wallet.py
```diff
@@ -568,30 +569,30 @@ async def path_hint_for_pubkey(self, pk: bytes) -> Optional[PathHint]:
...
+    #         ):
+    #             pk_bytes = bytes(pk)
+    #             pks.append(pk_bytes)
+    #             fingerprint: bytes = pk.get_fingerprint().to_bytes(4, "big")
+    #             signing_targets.append(SigningTarget(fingerprint, msg, std_hash(pk_bytes + msg)))
+
+    #     return SigningInstructions(
+    #         await self.wallet_state_manager.key_hints_for_pubkeys(pks),
+    #         signing_targets,
+    #     )
```
Quexington, line 595: ???

Quexington: Oh didn't scroll down.  Still, obviously delete this.

chia/wallet/puzzles/vault_p2_recovery.clsp
```diff
@@ -47,20 +47,23 @@
...
-  (list
-    (list AGG_SIG_ME BLS_PK (sha256tree recovery_conditions))
-    (list CREATE_COIN
-      (create_recovery_puzzlehash
-        P2_1_OF_N_MOD_HASH
-        FINISH_RECOVERY_MOD_HASH
-        P2_SECP_PUZZLEHASH
-        TIMELOCK
-        recovery_conditions
+  (let
```
Quexington, line 50: I'm double checking with Art but I'm 80% sure that `assign` is more appropriate here so you don't end up calculating the puzzle hash twice.

Quexington (approved):

wjblanke (approved):
aok

PR #17197, opened 2024-01-03
Update chia_rs to 0.2.15 for AMD K10 fix

### LINKED ISSUE #16386: (Bug) Unable to run Chia using Windows on AMD K10 architecture (AMD Phenom)
### What happened?

hello, I would like someone to help me. this is how the chia app works for me on 1.8.2. since I want a stronger version, when I install 2.0 or 2.0.1 I have a problem. the problem is the following. When entering the application, it always says Loading keyring status.... what should I do? thank you

### Version

2.0.0 - 2.0.1

Windows

GUI

### Relevant log output

```shell
Loading keyring status
```

emlowe:
Update chia_rs binary wheel to 0.2.15 to fix issue with Windows on AMD K10 architecture (AMD Phenom and other K10 based chips)

fixes #16386

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ dependencies = [
     "chiapos==2.0.3",  # proof of space
     "clvm==0.9.8",
     "clvm_tools==0.4.7",  # Currying, Program.to, other conveniences
-    "chia_rs==0.2.13",
+    "chia_rs==0.2.15",
     "clvm-tools-rs==0.1.39",  # Rust implementation of clvm_tools' compiler
     "aiohttp==3.9.1",  # HTTP server for full node rpc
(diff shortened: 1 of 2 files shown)


wjblanke (dismissed):
aok

arvidn (dismissed):
it seems a bit unlikely that the test failure is because it takes too long for the two NFTs to be transferred. I would guess it's rather a flaky test where they often just fail to be transferred at all

Starttoaster:
Seems like there is one untested line that it's also triggering on, besides the authorization error for adding the coverage report to the PR. But if you hide whitespace changes, the line it claims is untested is outside of the diff so I'm good with ignoring it.

Starttoaster (commented):
Approved :)

PR #17207, opened 2024-01-03
evict entries continuously from `seen_unfinished_blocks`

arvidn:
### Purpose:

Currently we empty the *whole* cache of recently seen unfinished blocks every 1000 blocks. This causes inefficiencies around those points, as we may process some unfinished blocks multiple times immediately after clearing.

### Current Behavior:

We drop the whole cache periodically.

### New Behavior:

We drop the *oldest* entry whenever we exceed 1000 entries.

### DIFF
--- a/chia/full_node/full_node_store.py
+++ b/chia/full_node/full_node_store.py
@@ -45,8 +45,11 @@ class FullNodeStore:
     candidate_blocks: Dict[bytes32, Tuple[uint32, UnfinishedBlock]]
     candidate_backup_blocks: Dict[bytes32, Tuple[uint32, UnfinishedBlock]]
 
-    # Header hashes of unfinished blocks that we have seen recently
-    seen_unfinished_blocks: Set[bytes32]
+    # Block hashes of unfinished blocks that we have seen recently. This is
+    # effectively a Set[bytes32] but in order to evict the oldest items first,
+    # we use a Dict that preserves insertion order, and remove from the
+    # beginning
+    seen_unfinished_blocks: Dict[bytes32, None]
 
     # Unfinished blocks, keyed from reward hash
     unfinished_blocks: Dict[bytes32, Tuple[uint32, UnfinishedBlock, PreValidationResult]]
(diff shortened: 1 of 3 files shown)


chia/full_node/full_node_store.py
```diff
@@ -45,8 +45,11 @@ class FullNodeStore:
     candidate_blocks: Dict[bytes32, Tuple[uint32, UnfinishedBlock]]
     candidate_backup_blocks: Dict[bytes32, Tuple[uint32, UnfinishedBlock]]
 
-    # Header hashes of unfinished blocks that we have seen recently
-    seen_unfinished_blocks: Set[bytes32]
+    # Block hashes of unfinished blocks that we have seen recently. This is
+    # effectively a Set[bytes32] but in order to evict the oldest items first,
+    # we use a Dict that preserves insertion order, and remove from the
+    # beginning
+    seen_unfinished_blocks: Dict[bytes32, bool]
```
altendky, line 52: Does this work?  Seems _slightly_ more explicit that the value is irrelevant.

```suggestion
    seen_unfinished_blocks: Dict[bytes32, None]
```

tests/core/full_node/stores/test_full_node_store.py
```diff
@@ -114,7 +114,9 @@ async def test_basic_store(
     h_hash_1 = bytes32.random(seeded_random)
     assert not store.seen_unfinished_block(h_hash_1)
     assert store.seen_unfinished_block(h_hash_1)
-    store.clear_seen_unfinished_blocks()
+    # this will crowd out h_hash_1
+    for _ in range(1000):
+        store.seen_unfinished_block(bytes32.random())
```
altendky, line 119:
```suggestion
        store.seen_unfinished_block(bytes32.random(r=seeded_random))
```

altendky (changes requested):

chia/full_node/full_node_store.py
arvidn: I got errors when using `None`. I think it has to do with being able to distinguish between a key existing or not, in the return value of `.get()`. But the error said something about `Sized`

altendky: mypy seems ok with it at this point and I don't see code using the value at all, as you would expect.

arvidn: I must have been mistaking this for some unrelated error. You're right, it seems to work

altendky (approved):

emlowe (approved):

PR #17209, opened 2024-01-04
move tests

altendky:
### Purpose:

This Proposal moves our existing `tests` into the public part of `chia-blockchain` - for example, if you `pip install chia-blockchain` you get all the `test` code as well.

This makes it quite a bit easier for a consumer of chia-blockchain to use test framework code - for example, things like CDV, nft minter, bridge - all that stuff would have access to testing tooling out of the box.

This involves physically moving the `test` directory to \ under `chia`.

The current proposal involves making two "test" directories. `chia/tests` consists of tooling and frameworks while `chia/_tests` consists of the specific chia-blockchain tests. This logical separation allows us to suggest that things in `_tests` are more private to chia-blockchain (or perhaps more specific) while things in `chia/tests` are more expected to be used publicly by other things.

Primarily this change was to eliminate issues with needing *some* parts of the test framework (the `simulator` needs some parts) but not others, and we were running into issues where some developers would move items from `tests` to `chia` and other developers would them move back from `chia` to `tests`. Making it all public at least solves any issues with which parts should be public.

A secondary benefit was that devs using the wheel have access to all the test framework for their purposes

And finally, a third benefit was the wheel users now can run the tests locally

### Draft For:
- [x] general consideration and a final decision
- [x] removing move script
- [x] perhaps moving the test utils to `chia.tests` or `chia.testing` or...
  - let's take that on separately


emlowe (dismissed):

emlowe:
coverage diff exemption

emlowe (approved):

PR #17212, opened 2024-01-04
Document the fact that we should be using tx_config and taking a fee in DID's create_attestment

AmineKhaldi:


arvidn (dismissed):
It might be good to double check with whoever added this parameter that there isn't a plan to use it for something in the near future

Quexington (approved):

arvidn (approved):

PR #17218, opened 2024-01-04
Rust proof-of-space, reward chain and foliage types

arvidn:
This is best reviewed one commit at a time.

### Purpose:

This is a step towards addressing performance issues in serializing and deserializing certain data structures, as highlighted by [this](https://github.com/Chia-Network/chia-blockchain/pull/17168) instrumentation. See "(parsing the SubEpochChallengeSegments is slow)", "(serializing the weight proof is slow)" and "(parsing BlockRecord is slow)".

Another, long term, goal these changes work towards is to move block pre-validation (CLVM and BLS signatures) into rust, allowing to run it in threads instead of processes.

### Rust counterparts

These commits each replace python implementation of classes with rust counterparts. The rust implementation of serialize and deserialize are faster as well having a more compact in-memory representation. The rust classes can be found [here](https://github.com/Chia-Network/chia_rs/blob/main/chia-protocol/src/weight_proof.rs), [here](https://github.com/Chia-Network/chia_rs/blob/main/chia-protocol/src/foliage.rs), [here](https://github.com/Chia-Network/chia_rs/blob/main/chia-protocol/src/pool_target.rs) and [here](https://github.com/Chia-Network/chia_rs/blob/main/chia-protocol/src/reward_chain_block.rs).

These types are mostly unrelated to each other, but share the trait that they are "leaf" types. Other types contain these as members, so these need to be transitioned to rust before the aggregate types can transition.

Note that the rust python bindings don't know about our `bytes32` type nor our fixed-width integer types (`uint8`, uint32` etc.). The bindings will expose plain `int` to python (but they are fixed-width under the hood).

The python types stubs accurately type these fields as `int`, which means mypy can warn in places where there's a cast required.

### Current Behavior:

`TransactionsInfo`, `FoliageTransactionBlock`, `FoliageBlockData`, `Foliage`, `PoolTarget`, `ProofOfSpace`, `RewardChainBlockUnfinished` and `RewardChainBlock` are python classes.

### New Behavior:

`TransactionsInfo`, `FoliageTransactionBlock`, `FoliageBlockData`, `Foliage`, `PoolTarget`, `ProofOfSpace`, `RewardChainBlockUnfinished` and `RewardChainBlock` are native classes, implemented in `chia_rs`.


chia/consensus/full_block_to_block_record.py
```diff
@@ -88,6 +88,13 @@ def block_to_block_record(
     )
 
 
+def maybe_uint64(val: Optional[int]) -> Optional[uint64]:
```
altendky, line 91: In case we use this somewhere we shouldn't (somewhere where the type is known) it can propagate the constrained type instead of widening it.  Or, maybe this is just a silly thing to bother with in this case.


```suggestion
@overload
def maybe_uint64(val: None) -> None:
    ...


@overload
def maybe_uint64(val: int) -> uint64:
    ...


def maybe_uint64(val: Optional[int]) -> Optional[uint64]:
```

altendky, line 91: Maybe add `StructStream.maybe()` instead.

chia/types/blockchain_format/foliage.py
```diff
@@ -1,61 +1,5 @@
...
-    # The entire foliage block, containing signature and the unsigned back pointer
-    # The hash of this is the "header hash". Note that for unfinished blocks, the prev_block_hash
-    # Is the prev from the signage point, and can be replaced with a more recent block
-    prev_block_hash: bytes32
-    reward_block_hash: bytes32
-    foliage_block_data: FoliageBlockData
-    foliage_block_data_signature: G2Element
-    foliage_transaction_block_hash: Optional[bytes32]
-    foliage_transaction_block_signature: Optional[G2Element]
+__all__ = ["TransactionsInfo", "FoliageTransactionBlock", "FoliageBlockData", "Foliage"]
```
altendky, line 5: Let's add at least a comment here.  Maybe at all such locations so we can readily search for them.  Maybe even have a PR deprecating this module so we get warnings and clean up the usages?

chia/types/blockchain_format/proof_of_space.py
```diff
@@ -1,31 +1,20 @@
...
 from chia.consensus.constants import ConsensusConstants
 from chia.types.blockchain_format.sized_bytes import bytes32
 from chia.util.hash import std_hash
-from chia.util.ints import uint8, uint32
-from chia.util.streamable import Streamable, streamable
-
-log = logging.getLogger(__name__)
+from chia.util.ints import uint32
 
+__all__ = ["ProofOfSpace"]
```
altendky, line 15: Note that adding `__all__` actually reduces the default effective `__all__` for `*` imports down to that one thing.  I understand that doing nothing leaves complaints about things not being explicitly reexported.  What about something more like `import chia_rs; ProofOfSpace = chia_rs.ProofOfSpace`?  Does that work?

altendky (changes requested):
Has anything been done around identifying typing issues etc in less strictly typed files?

chia/consensus/full_block_to_block_record.py
arvidn: this function is meant to *maybe* construct a `uint64` object, around an existing optional `int`.
It's used when the resulting parameter type or field type is `Optional[uint64]` and we don't really want to change it.

arvidn: how would that work?

chia/types/blockchain_format/foliage.py
arvidn: I would expect we'd want to keep these names around for at least a while. I suspect these are the kinds of types that are likely to be pulled out by a 3rd party. Deprecating them makes sense though, but only after we transition away from them ourselves

chia/types/blockchain_format/proof_of_space.py
arvidn: I haven't tried that. I googled a long time ago and people said to do it like this. I'll try

chia/consensus/full_block_to_block_record.py
altendky:
```python
@classmethod
def maybe(cls: Type[T], val: Optional[int]) -> Optional[T]:
    # regular logic here
```

chia/types/blockchain_format/foliage.py
altendky: A was figuring a separate PR (at some point...) which deprecated it and thus caught our own uses of it would be fine.

Feel free to close this conversation.

chia/consensus/full_block_to_block_record.py
arvidn: I think `uint64.maybe()` reads a bit unintuitively. I called it `construct_optional()`

altendky (approved):

emlowe (approved):
Looks good to me, although I find all the casts somewhat tedious

PR #17227, opened 2024-01-05
Fix bug trading NFT0+XCH for CAT

cameroncooper:
### Purpose:

This fixes a bug when trading NFT0+XCH for a CAT. The root cause is a sorting function that attempts to sort a list that contains ints (wallet ids) and bytes32 (asset ids). This does not affect NFT1 because NFT1 offers are handled in a separate code path.

```
in _create_offer_for_ids
    for id in sorted(coins_to_offer.keys()):
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'bytes32' and 'int'
```

The fix is to always use wallet ids.

### Current Behavior:

Exception when trying to accept an offer for NFT0+XCH.

### New Behavior:

No exception, works as expected.

### Testing Notes:

Added a new test trading NFT0+XCH for CAT.

### DIFF
--- a/chia/wallet/trade_manager.py
+++ b/chia/wallet/trade_manager.py
@@ -485,7 +485,7 @@ class TradeManager:
                     amount_to_select = abs(amount)
                     if wallet.type() == WalletType.STANDARD_WALLET:
                         amount_to_select += fee
-                    coins_to_offer[id] = await wallet.get_coins_to_offer(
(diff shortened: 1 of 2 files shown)


emlowe:
Should this be closed due to https://github.com/Chia-Network/chia-blockchain/pull/17270

PR #17241, opened 2024-01-08
Add ability to get the aggsig additional data from full node RPC

matt-o-how:
### Purpose:
This adds in the ability to request a full node to tell you what the additional data required for an aggsig_me is.
This makes it easier for people to create applications which support use in the full node and create signatures to work on testnet, simulator, mainnet, and even other chia forks.

### Current Behavior:
Function does not currently exist.

### New Behavior:
Added `get_aggsig_additional_data` to both RPC API and RPC Client.

### Testing Notes:
Added a line of test in to RPC tests.

### DIFF
--- a/chia/rpc/full_node_rpc_client.py
+++ b/chia/rpc/full_node_rpc_client.py
@@ -159,11 +159,9 @@ class FullNodeRpcClient(RpcClient):
         response = await self.fetch("get_coin_records_by_parent_ids", d)
         return [CoinRecord.from_json_dict(coin_record_dict_backwards_compat(coin)) for coin in response["coin_records"]]
 
-    async def get_aggsig_additional_data(self) -> str:
+    async def get_aggsig_additional_data(self) -> bytes32:
         result = await self.fetch("get_aggsig_additional_data", {})
-        add_data: str = result["additional_data"]
-        assert add_data is not None
-        return add_data
+        return bytes32.from_hexstr(result["additional_data"])
 
     async def get_coin_records_by_hint(
         self,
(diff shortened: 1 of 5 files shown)


matt-o-how:
Despite the branch name, this code no longer adds the ability to melt CATs

chia/rpc/full_node_rpc_client.py
```diff
@@ -159,6 +159,12 @@ async def get_coin_records_by_parent_ids(
         response = await self.fetch("get_coin_records_by_parent_ids", d)
         return [CoinRecord.from_json_dict(coin_record_dict_backwards_compat(coin)) for coin in response["coin_records"]]
 
+    async def get_aggsig_additional_data(self) -> str:
```
Quexington, line 162: Let's return bytes from this client endpoint.  `bytes32` if it's appropriate.

chia/rpc/full_node_rpc_client.py
```diff
@@ -159,6 +159,12 @@ async def get_coin_records_by_parent_ids(
...
+        result = await self.fetch("get_aggsig_additional_data", {})
+        add_data: str = result["additional_data"]
+        assert add_data is not None
```
Quexington, line 165: This is not necessary according to the hint above so one of them is wrong.  I don't think you need to assign this to a variable though, just return `hexstr_to_bytes` (or `bytes32.from_hexstr`) on `result["additional_data"]`.

Quexington (changes requested):

chia/rpc/full_node_rpc_client.py
matt-o-how: Done

chia/rpc/full_node_rpc_client.py
matt-o-how: Done

Quexington (approved):

arvidn (approved):

PR #17249, opened 2024-01-09
Refactor subscriptions and add remove functionality

Rigidity:
### Purpose:

Cleans up some of the code in the PeerSubscriptions class and adds support for removing individual subscriptions rather than the whole peer.

### Testing Notes:

All of the tests have been updated to match new method names, and new tests are added for removing subscriptions. There are no protocol changes to test in this PR.

### DIFF
--- a/chia/full_node/subscriptions.py
+++ b/chia/full_node/subscriptions.py
@@ -48,22 +48,16 @@ class SubscriptionSet:
         return item in self._peers_for_subscription
 
     def count_subscriptions(self, peer_id: bytes32) -> int:
-        if peer_id not in self._subscriptions_for_peer:
-            return 0
-
-        return len(self._subscriptions_for_peer[peer_id])
+        return len(self._subscriptions_for_peer.get(peer_id, {}))
 
     def remove_peer(self, peer_id: bytes32) -> None:
-        if peer_id not in self._subscriptions_for_peer:
-            return
-
-        for item in self._subscriptions_for_peer[peer_id]:
+        for item in self._subscriptions_for_peer.pop(peer_id, {}):
             self._peers_for_subscription[item].remove(peer_id)
 
             if len(self._peers_for_subscription[item]) == 0:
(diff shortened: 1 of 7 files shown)


chia/full_node/subscriptions.py
```diff
@@ -9,136 +9,204 @@
...
+        subscription_count = self.peer_subscription_count(peer_id)
+        added: Set[bytes32] = set()
 
-        # if we've reached the limit on number of subscriptions, just bail
-        if existing_sub_count >= max_items:
+        def limit_reached() -> Set[bytes32]:
             log.info(
-                "peer_id: %s reached max number of puzzle-hash subscriptions. "
-                "Not all its coin states will be reported",
+                "Peer %s reached the subscription limit.",
```
arvidn, line 104: since you have separate functions and log line, you might as well specify whether it's puzzle-hash or coin id subscriptions

arvidn (commented):
thanks for factoring out `SubscriptionSet`. You don't think it deserves a unit test as well? I suppose all of its functionality is tested via the subscriptions class

chia/full_node/subscriptions.py
```diff
@@ -48,22 +48,16 @@ def has_subscription(self, item: bytes32) -> bool:
...
-
-        for item in self._subscriptions_for_peer[peer_id]:
+        for item in self._subscriptions_for_peer.pop(peer_id, {}):
             self._peers_for_subscription[item].remove(peer_id)
 
             if len(self._peers_for_subscription[item]) == 0:
                 self._peers_for_subscription.pop(item)
 
-        self._subscriptions_for_peer.pop(peer_id)
+        self._subscriptions_for_peer.pop(peer_id, None)
```
arvidn, line 60: you don't need this line anymore, right?

Rigidity: Right, good point

arvidn (approved):

PR #17272, opened 2024-01-10
This should ensure that the wallets are fully synced before and after the reorg test in the nft wallet.

prozacchiwawa:
…  This should fully fix the test up.

### Purpose:

This ensures that the nft wallet is synced for some operations that heavily interact with the full node and the wallet and have somewhat unpredictable timing.

### Current Behavior:

The test_nft_wallet_creation_and_transfer is flaky and fails almost constantly on windows with python 3.8.

### New Behavior:

The test should work consistently.

### Testing Notes:

This is an update of a test.  It doesn't change the test's own checks or functionality but adds a few needed calls to ensure that the wallets update themselves to the current blockchain status.

### DIFF
--- a/tests/wallet/nft_wallet/test_nft_wallet.py
+++ b/tests/wallet/nft_wallet/test_nft_wallet.py
@@ -195,6 +195,10 @@ async def test_nft_wallet_creation_and_transfer(self_hostname: str, two_wallet_n
     ph = await wallet_0.get_new_puzzlehash()
     ph1 = await wallet_1.get_new_puzzlehash()
 
+    async def ensure_wallet_sync():
+        await full_node_api.wait_for_wallet_synced(wallet_node=wallet_node_0, timeout=20)
+        await full_node_api.wait_for_wallet_synced(wallet_node=wallet_node_1, timeout=20)
+
     if trusted:
         wallet_node_0.config["trusted_peers"] = {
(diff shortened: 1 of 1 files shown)


tests/wallet/nft_wallet/test_nft_wallet.py
```diff
@@ -320,17 +332,28 @@ async def test_nft_wallet_creation_and_transfer(self_hostname: str, two_wallet_n
...
     assert compute_memos(txs[0].spend_bundle)
 
+    await ensure_wallet_sync()
+
+    # We wish for the wallet to be in a specific state at this point after having
+    # become synced above.
     for i in range(1, num_blocks):
         await full_node_api.farm_new_transaction_block(FarmNewBlockProtocol(ph1))
 
-    await time_out_assert(30, wallet_node_0.wallet_state_manager.lock.locked, False)
```
arvidn, line 326: was this a misguided attempt at, essentially, waiting for the wallet to be synced?

arvidn (approved):
looks good! feel free to resolve my comment

tests/wallet/nft_wallet/test_nft_wallet.py
prozacchiwawa: my read is this: this lock is used when the wallet is in long sync mode, so i think this was trying to ensure that we don't go into the next phase if a long sync might be in progress and waiting at an io transaction.  I'm certain there's no reason to do this if wait_for_wallet_synced is used.

PR #17275, opened 2024-01-10
extend Block validation timing logs to measure just the CLVM and conditions

arvidn:
### Purpose:

This patch adds additional instrumentation to help pinpoint what's causing block validation delay. This is an example log. This is an outlier in that the overall time takes a lot longer than the inner work. However, it's representative in that the CLVM timing is always low.

```
2024-01-10T13:47:19.862 full_node chia.full_node.full_node: WARNING
   Block validation: 7.87s,
   pre_validation: 2.04s,
   CLVM: 0.17s,
   post-process: 0.04s,
   cost: None header_hash: 598741206f38f940a7a032730bc6890e2dc0f904a6eec524445496106ed69eed
   height: 4780875
```

Note that:
* there's no reorg, so the post-processing time is negligible.
* The CLVM time is also very small. This is the call to `get_name_puzzle_conditions()`, which parses and executes the block generator as well as parses and coalesces the resulting conditions. It also checks some conditions that can be checked without context.
* The `pre_validation` time is quite a lot higher (2 seconds). This includes the overhead of pulling blocks from the DB and post jobs to the worker processes.
* The total validation time is even higher (almost 8 seconds)

My theory is that this delay happens when there are many other asyncio tasks ready to run, and gets scheduled ahead of the block validation task when its results are done.

The above log line is getting longer, so I tried to shorten it a bit, by removing the word `time` (which was repeated) as well as turned `seconds` -> `s`.

The profile for this specific block looks like this:

### Current Behavior:

It's hard to tell whether running CLVM is taking a long time, or something else.

### New Behavior:

It's clearer that CLVM is not the bottleneck.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -728,7 +728,7 @@ class Blockchain(BlockchainInterface):
         required_iters, error = await self.validate_unfinished_block_header(block, skip_overflow_ss_validation)
 
         if error is not None:
-            return PreValidationResult(uint16(error.value), None, None, False)
+            return PreValidationResult(uint16(error.value), None, None, False, uint32(0))
 
         prev_height = (
             -1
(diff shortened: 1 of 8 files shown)


Rigidity (approved):

wjblanke (approved):
aok

PR #17276, opened 2024-01-10
Add fork height & rolled_back_records to block event for metrics

cmmarslender:


chia/full_node/full_node.py
```diff
@@ -1831,11 +1831,17 @@ async def add_block(
             "transaction_block": False,
             "k_size": block.reward_chain_block.proof_of_space.size,
             "header_hash": block.header_hash,
+            "fork_height": 0,
+            "rolled_back_records": 0,
```
arvidn, line 1835: it looks like we're fine with omitting fields that aren't relevant. for example: `block_cost` and `block_fees`.
including `rolled_back_records` as 0 makes a certain amount of sense, but I'm not sure about including `fork_height` as 0 by default. if that value actually ends up in the result, it will look like an extremely deep reorg

cmmarslender: Updated this to `None` as default values instead of zero (since its a bit easier to parse that out vs a sometimes there sometimes not field in chia-exporter)

arvidn (approved):

PR #17278, opened 2024-01-10
Tweak wallet RPC test that wasn't working

Quexington:
This test wasn't working due to a typo (`outputs` instead of `output`) and getting it to work required rewriting that section a bit.


AmineKhaldi (approved):

wjblanke (approved):
aok

PR #17285, opened 2024-01-10
Clarify to_program handling of args in UnknownCondition

AmineKhaldi:


arvidn (approved):
as far as I can tell this has identical behavior, it's just explicit about the types

PR #17290, opened 2024-01-10
Introduce and use LineageProof.from_program

AmineKhaldi:


chia/wallet/lineage_proof.py
```diff
@@ -1,21 +1,43 @@
...
 class LineageProof(Streamable):
     parent_name: Optional[bytes32] = None
     inner_puzzle_hash: Optional[bytes32] = None
     amount: Optional[uint64] = None
 
+    @classmethod
+    def from_program(cls, program: Program, fields: List[LineageProofField]) -> LineageProof:
+        program_values = list(program.as_iter())
+        if len(program_values) != len(fields):
+            raise ValueError("Mismatch between program data and fields information")
```
arvidn, line 30: for the record: this is the line that has no test coverage

chia/wallet/lineage_proof.py
```diff
@@ -1,21 +1,43 @@
...
 @streamable
 @dataclass(frozen=True)
 class LineageProof(Streamable):
     parent_name: Optional[bytes32] = None
     inner_puzzle_hash: Optional[bytes32] = None
     amount: Optional[uint64] = None
 
+    @classmethod
+    def from_program(cls, program: Program, fields: List[LineageProofField]) -> LineageProof:
+        program_values = list(program.as_iter())
```
arvidn, line 28: I believe `program` cannot be trusted here. But perhaps I'm wrong.
There's a subtle risk when you turn a CLVM structure into a concrete python list. It can be very very long. it would be safer to test the length of this list by iterating `len(fields)` times and ensure the list ends there.
I thought we had a utility function for this, but looking at `Program` and `SExp`, the best I can find is `SExp.list_len()`, which is a little bit better than converting to a `list`

arvidn (commented):
I think the new function should have a unit test

AmineKhaldi:
Superseded by PR #17301

PR #17291, opened 2024-01-10
construct_exigent_metadata_layer's metadata is an optional Program

AmineKhaldi:


arvidn:
a `Program` is a "wider" type than `bytes32`. Can you justify why metadata should be allowed to be any CLVM structure rather than just a hash?

Quexington:
> a `Program` is a "wider" type than `bytes32`. Can you justify why metadata should be allowed to be any CLVM structure rather than just a hash?

Because the chialisp allows things that are not a hash and many call sites pass in a program already.  The metadata is allowed to be any tree, not just a hash.  Type broadens to fit actual broad definition.

Quexington (approved):

arvidn:
thanks @Quexington . I wanted to make sure the semantics change was deliberate (as I think @AmineKhaldi primarily is fixing up typing issues)

arvidn (approved):

PR #17292, opened 2024-01-10
Compute authorized providers list and pass it to construct_cr_layer

AmineKhaldi:


chia/wallet/vc_wallet/cr_cat_drivers.py
```diff
@@ -369,14 +370,14 @@ def get_next_from_coin_spend(
             if conditions is None:
                 conditions = inner_puzzle.run(inner_solution)
             inner_puzzle_hash: bytes32 = inner_puzzle.get_tree_hash()
+            authorized_providers = [bytes32(p.as_atom()) for p in authorized_providers_as_prog.as_iter()]
             lineage_inner_puzhash = construct_cr_layer(
-                authorized_providers_as_prog,
```
arvidn, line 373: does `construct_cr_layer` expect a `Program` or a `List[bytes32]`?

arvidn (commented):
it's not obvious that it's correct to pass a `List[bytes32]` instead of a `Program`. can you explain why it is?

chia/wallet/vc_wallet/cr_cat_drivers.py
arvidn: It's typed to take a `List[bytes32]`, but the value is only passed on to `Program.to()`, so it doesn't really matter. Is it better to convert it to a list first, then pass it, and then convert it back?
An alternative approach would be to change the type of the parameter to be `Program`. I think @Quexington needs to have an opinion.

Quexington: I think this change is correct.  It's not captured in the diff, but the `List[bytes32]` is used elsewhere in the function to initialized the class and having the member of the class be `List[bytes32]` is definitely the ergonomics we want.  In addition, the driver function `construct_cr_layer` should take the bytes list because in most places we only have the bytes list and forcing conversion to program before using the driver function would be awkward and inconsistent with other drivers of its type.

This is representative of a sort of python/clvm tension that permeates our codebase though.  It's almost as if we would like our program construction files to take `Program.to()`-able python OR clvm OR prehashed subtrees of clvm in some scenarios.  There's probably a more elegant solution to stuff like this that we can find in the future but for now I think we're stuck with some inefficiencies like this.

Quexington (approved):

arvidn (approved):

PR #17295, opened 2024-01-10
Fix run_with_cost calls in benchmark_cost.py

AmineKhaldi:


arvidn (approved):

arvidn:
the test coverage failure is OK. We don't instrument benchmarks for coverage.

PR #17297, opened 2024-01-11
use rust implementation of SerializedProgram

arvidn:
### Purpose:

This replaces the `SerializedProgram` class with the counterpart from `chia_rs`. The long-term objective is to make `FullBlock` able to live entirely in rust, to move validation to threads. The `transactions_generator` field is a serialized CLVM program.

The main challenge in the python binding for this class, is how *python* arguments are converted into CLVM structures before invoking the program (e.g. `run_with_cost()`). This conversion is partly implemented in `_serialize()`, [here](https://github.com/Chia-Network/chia-blockchain/pull/17297/files#diff-61e718c007654ae217fed41b50187ce5c34f0bbbc973bb69e179e64650ef1513L15) as well as `SExp.to`, in the `clvm` wheel.

Note how there's a special case when passing a `SerializedProgram`, its bytes are used directly. i.e. a serialized program is passed as the CLVM structure it represents, not as a blob of bytes (i.e. a single atom). This special handling no longer exists when entering the `SExp.to()` path.

The rust implementation of this behavior can be found [here](https://github.com/Chia-Network/chia_rs/blob/main/chia-protocol/src/program.rs#L196-L239) (the `_serialize()` counterpart) and [here](https://github.com/Chia-Network/chia_rs/blob/main/chia-protocol/src/program.rs#L117-L193) (the `SExp.to()` counterpart).

### Current Behavior:

`SerializedProgram` is implemented in python (but CLVM is run by rust)

### New Behavior:

`SerializedProgram` is implemented in rust


Rigidity (approved):
Love to see it

emlowe (approved):

PR #17300, opened 2024-01-11
Add way to get coin states in batches

Rigidity:
### Purpose:

When added to the wallet protocol, this will allow wallets to sync their coin state efficiently in batches rather than downloading all of the state in a single response (which can fail if it's too large).

### Testing Notes:

There is a unit test for ensuring that it successfully fetches the correct number of coins with a high volume of puzzle hashes (including with filters enabled and disabled), and another for high coin record count.

### DIFF
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -420,9 +420,9 @@ class CoinStore:
         include_unspent: bool = True,
         include_hinted: bool = True,
         max_items: int = 50000,
-    ) -> Tuple[List[CoinState], uint32, bool]:
+    ) -> Tuple[List[CoinState], Optional[uint32]]:
         """
-        Returns the coin states, as well as the next block height and whether finished.
+        Returns the coin states, as well as the next block height (or `None` if finished).
         Note that the maximum number of puzzle hashes is currently set to 15000.
         """
 
@@ -448,8 +448,8 @@ class CoinStore:
             elif include_unspent:
                 height_filter = f"AND {require_unspent}"
             else:
-                # There are no coins which are both spent and unspent, so return an empty list.
-                return [], min_height, True
+                # There are no coins which are both spent and unspent, so we're finished.
+                return [], None
 
             if include_hinted:
                 cursor = await conn.execute(
@@ -479,26 +479,26 @@ class CoinStore:
             for row in await cursor.fetchall():
                 coin_states.append(self.row_to_coin_state(row))
 
-            is_finished: bool
+        # If there aren't too many coin states, we've finished syncing these hashes.
+        # There is no next height to start from, so return `None`.
+        if len(coin_states) <= max_items:
+            return coin_states, None
 
-            if len(coin_states) <= max_items:
-                next_height = min_height
-                is_finished = True
-            else:
-                next_coin_state = coin_states.pop()
-                next_height = uint32(max(next_coin_state.created_height or 0, next_coin_state.spent_height or 0))
+        # The last item is the start of the next batch of coin states.
+        next_coin_state = coin_states.pop()
+        next_height = uint32(max(next_coin_state.created_height or 0, next_coin_state.spent_height or 0))
 
-                while len(coin_states) > 0:
-                    last_coin_state = coin_states[-1]
-                    height = uint32(max(last_coin_state.created_height or 0, last_coin_state.spent_height or 0))
-                    if height == min_height:
-                        coin_states.pop()
-                    else:
-                        break
+        # In order to prevent blocks from being split up between batches, remove
+        # all coin states whose max height is the same as the last coin state's height.
+        while len(coin_states) > 0:
+            last_coin_state = coin_states[-1]
+            height = uint32(max(last_coin_state.created_height or 0, last_coin_state.spent_height or 0))
+            if height != next_height:
+                break
 
-                is_finished = False
+            coin_states.pop()
 
-        return coin_states, next_height, is_finished
+        return coin_states, next_height
 
     async def rollback_to_block(self, block_index: int) -> List[CoinRecord]:
         """
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -483,6 +483,15 @@ class CoinStore:
                 elif len(coin_states) > max_items:
                     next_coin_state = coin_states.pop()
                     start_height = uint32(max(next_coin_state.created_height or 0, next_coin_state.spent_height or 0))
+
+                    while len(coin_states) > 0:
+                        last_coin_state = coin_states[-1]
+                        height = uint32(max(last_coin_state.created_height or 0, last_coin_state.spent_height or 0))
+                        if height == start_height:
+                            coin_states.pop()
+                        else:
+                            break
+
                     break
                 else:
                     start_height = min_height
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -411,6 +411,85 @@ class CoinStore:
 
         return coins
 
+    async def batch_coin_states_by_puzzle_hashes(
+        self,
+        puzzle_hashes: List[bytes32],
+        min_height: uint32 = uint32(0),
+        *,
+        include_spent: bool = True,
+        include_unspent: bool = True,
+        include_hinted: bool = True,
+        max_items: int = 50000,
+    ) -> Tuple[List[CoinState], List[bytes32], uint32]:
+        """
+        Returns the coin states in the current batch, as well as the remaining puzzle hashes
+        to be searched by in the next batch and the next block height to start from.
+        """
+
+        # Calculate the total number of groups of puzzle hashes to search through.
+        # Each query is be for both puzzle hash and hint, in addition to some additional filters.
+        other_variable_count = 3
+
+        if include_hinted:
+            group_size = SQLITE_MAX_VARIABLE_NUMBER // 2 - other_variable_count
+        else:
+            group_size = SQLITE_MAX_VARIABLE_NUMBER - other_variable_count
+
+        start_height = min_height
+        synced_puzzle_hashes = 0
+
+        coin_states: List[CoinState] = []
+
+        async with self.db_wrapper.reader_no_transaction() as conn:
+            for select_batch in to_batches(puzzle_hashes, group_size):
+                select_batch_db = tuple(select_batch.entries)
+
+                if include_hinted:
+                    cursor = await conn.execute(
+                        f"SELECT cr.confirmed_index, cr.spent_index, cr.coinbase, cr.puzzle_hash, "
+                        f"cr.coin_parent, cr.amount, cr.timestamp FROM coin_record cr "
+                        f"LEFT JOIN hints h ON cr.coin_name = h.coin_id "
+                        f'WHERE (cr.puzzle_hash in ({"?," * (len(select_batch.entries) - 1)}?) '
+                        f'OR h.hint in ({"?," * (len(select_batch.entries) - 1)}?)) '
+                        f"AND (cr.confirmed_index>=? OR cr.spent_index>=?) "
+                        f"{'' if include_spent else 'AND cr.spent_index=0'} "
+                        f"{'' if include_unspent else 'AND cr.spent_index>0'} "
+                        f"ORDER BY MAX(cr.confirmed_index, cr.spent_index) ASC "
+                        f"LIMIT ?",
+                        select_batch_db
+                        + select_batch_db
+                        + (start_height, start_height, max_items - len(coin_states) + 1),
+                    )
+                else:
+                    cursor = await conn.execute(
+                        f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
+                        f"coin_parent, amount, timestamp FROM coin_record INDEXED BY coin_puzzle_hash "
+                        f'WHERE puzzle_hash in ({"?," * (len(select_batch.entries) - 1)}?) '
+                        f"AND (confirmed_index>=? OR spent_index>=?) "
+                        f"{'' if include_spent else 'AND spent_index=0'} "
+                        f"{'' if include_unspent else 'AND spent_index>0'} "
+                        f"ORDER BY MAX(confirmed_index, spent_index) ASC"
+                        f"LIMIT ?",
+                        select_batch_db + (start_height, start_height, max_items - len(coin_states) + 1),
+                    )
+
+                for row in await cursor.fetchall():
+                    coin_states.append(self.row_to_coin_state(row))
+
+                if len(coin_states) == max_items:
+                    start_height = min_height
+                    synced_puzzle_hashes += len(select_batch.entries)
+                    break
+                elif len(coin_states) > max_items:
+                    next_coin_state = coin_states.pop()
+                    start_height = uint32(max(next_coin_state.created_height or 0, next_coin_state.spent_height or 0))
+                    break
+                else:
+                    start_height = min_height
+                    synced_puzzle_hashes += len(select_batch.entries)
+
+        return (coin_states, puzzle_hashes[synced_puzzle_hashes:], start_height)
+
     async def rollback_to_block(self, block_index: int) -> List[CoinRecord]:
         """
         Note that block_index can be negative, in which case everything is rolled back
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -414,8 +414,9 @@ class CoinStore:
     async def batch_coin_states_by_puzzle_hashes(
         self,
         puzzle_hashes: List[bytes32],
-        min_height: uint32 = uint32(0),
         *,
+        start_height: uint32,
+        min_height: uint32 = uint32(0),
         include_spent: bool = True,
         include_unspent: bool = True,
         include_hinted: bool = True,
@@ -435,7 +436,6 @@ class CoinStore:
         else:
             group_size = SQLITE_MAX_VARIABLE_NUMBER - other_variable_count
 
-        start_height = min_height
         synced_puzzle_hashes = 0
 
         coin_states: List[CoinState] = []
@@ -468,7 +468,7 @@ class CoinStore:
                         f"AND (confirmed_index>=? OR spent_index>=?) "
                         f"{'' if include_spent else 'AND spent_index=0'} "
                         f"{'' if include_unspent else 'AND spent_index>0'} "
-                        f"ORDER BY MAX(confirmed_index, spent_index) ASC"
+                        f"ORDER BY MAX(confirmed_index, spent_index) ASC "
                         f"LIMIT ?",
                         select_batch_db + (start_height, start_height, max_items - len(coin_states) + 1),
                     )
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -466,10 +466,10 @@ class CoinStore:
 
             is_finished: bool
 
-            if len(coin_states) == max_items:
+            if len(coin_states) <= max_items:
                 next_height = min_height
                 is_finished = True
-            elif len(coin_states) > max_items:
+            else:
                 next_coin_state = coin_states.pop()
                 next_height = uint32(max(next_coin_state.created_height or 0, next_coin_state.spent_height or 0))
 
@@ -482,9 +482,6 @@ class CoinStore:
                         break
 
                 is_finished = False
-            else:
-                next_height = min_height
-                is_finished = True
 
         return coin_states, next_height, is_finished
 
--- a/chia/full_node/coin_store.py
+++ b/chia/full_node/coin_store.py
@@ -415,89 +415,78 @@ class CoinStore:
         self,
         puzzle_hashes: List[bytes32],
         *,
-        start_height: uint32,
         min_height: uint32 = uint32(0),
         include_spent: bool = True,
         include_unspent: bool = True,
         include_hinted: bool = True,
         max_items: int = 50000,
-    ) -> Tuple[List[CoinState], List[bytes32], uint32]:
+    ) -> Tuple[List[CoinState], uint32, bool]:
         """
-        Returns the coin states in the current batch, as well as the remaining puzzle hashes
-        to be searched by in the next batch and the next block height to start from.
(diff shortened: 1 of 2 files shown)


chia/full_node/coin_store.py
```diff
@@ -411,6 +411,83 @@ async def get_coin_states_by_ids(
...
+                    f"LIMIT ?",
+                    puzzle_hashes_db + (min_height, min_height, max_items + 1),
+                )
+
+            for row in await cursor.fetchall():
+                coin_states.append(self.row_to_coin_state(row))
+
+            is_finished: bool
+
+            if len(coin_states) == max_items:
```
arvidn, line 469: if you change this condition to `len(coin_states) <= max_items`, I think you can drop the `else`-block, right?

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,83 @@ async def get_coin_states_by_ids(
...
+                        coin_states.pop()
+                    else:
+                        break
+
+                is_finished = False
+            else:
+                next_height = min_height
+                is_finished = True
+
+        return (coin_states, next_height, is_finished)
```
arvidn, line 489: you don't technically need the parentheses in the `return` statement

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,83 @@ async def get_coin_states_by_ids(
...
+            if include_hinted:
+                cursor = await conn.execute(
+                    f"SELECT cr.confirmed_index, cr.spent_index, cr.coinbase, cr.puzzle_hash, "
+                    f"cr.coin_parent, cr.amount, cr.timestamp FROM coin_record cr "
+                    f"LEFT JOIN hints h ON cr.coin_name = h.coin_id "
+                    f'WHERE (cr.puzzle_hash in ({"?," * (puzzle_hash_count - 1)}?) '
+                    f'OR h.hint in ({"?," * (puzzle_hash_count - 1)}?)) '
+                    f"AND (cr.confirmed_index>=? OR cr.spent_index>=?) "
+                    f"{'' if include_spent else 'AND cr.spent_index=0'} "
+                    f"{'' if include_unspent else 'AND cr.spent_index>0'} "
```
arvidn, line 446: if both `include_spent` and `include_unspent` are `True`, you don't need either of these conditions. and I would expect sqlite to be a little bit happier without them

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,83 @@ async def get_coin_states_by_ids(
...
+                    puzzle_hashes_db + puzzle_hashes_db + (min_height, min_height, max_items + 1),
+                )
+            else:
+                cursor = await conn.execute(
+                    f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
+                    f"coin_parent, amount, timestamp FROM coin_record INDEXED BY coin_puzzle_hash "
+                    f'WHERE puzzle_hash in ({"?," * (puzzle_hash_count - 1)}?) '
+                    f"AND (confirmed_index>=? OR spent_index>=?) "
+                    f"{'' if include_spent else 'AND spent_index=0'} "
+                    f"{'' if include_unspent else 'AND spent_index>0'} "
```
arvidn, line 458: ditto

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,83 @@ async def get_coin_states_by_ids(
 
         return coins
 
+    async def batch_coin_states_by_puzzle_hashes(
+        self,
+        puzzle_hashes: List[bytes32],
+        *,
+        min_height: uint32 = uint32(0),
+        include_spent: bool = True,
```
arvidn, line 419: is this really supposed to default to true?
In the case of subscribing, I don't think we include spent coins in the initial update at least, but I could be wrong.

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,169 @@ async def test_get_coin_states(db_version: int) -> None:
...
+                    remaining_phs = remaining_phs[15000:]
+
+                    if len(remaining_phs) > 0:
+                        height = uint32(0)
+                        is_finished = False
+
+            return all_coin_states
+
+        # Make sure all of the coin states are found when batching.
+        all_coin_states = await sync_states(include_spent=True, include_unspent=True, include_hinted=True)
```
arvidn, line 574: it seems like these calls to `sync_states()` would be quite suitable to be test parameters. Perhaps the setup is a bit too expensive for that though. The main benefit would be that it would be easy to tell whether all combinations are covered

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,169 @@ async def test_get_coin_states(db_version: int) -> None:
...
+        all_coin_states.sort(key=lambda cs: cs.coin.amount)
+
+        assert len(all_coin_states) == len(coin_records)
+
+        for i in range(min(len(coin_records), len(all_coin_states))):
+            assert coin_records[i].coin.name().hex() == all_coin_states[i].coin.name().hex(), i
+
+        # Make sure you can filter out hints.
+        all_coin_states = await sync_states(include_spent=True, include_unspent=True, include_hinted=False)
+        assert len(all_coin_states) == len(coin_records) - len(hints)
```
arvidn, line 584: I think it would be good to also ensure you get the exact records you expect. It's probably impractical to list every coin (among 15000), but you could ensure that all coins satisfies certain properties (like, not being spent or being spent, or having a hint or not having a specific hint).
Maybe ensure that all puzzle hashes are unique

chia/full_node/coin_store.py
Rigidity: We do:
https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/full_node_api.py#L1508

You need it for syncing historical transactions. That said, this will be specified by the new wallet protocol message anyways using `InitialStateFilters`. All future coin state updates will be received regardless of whether they're spent or not, but for initial state you will have the option to filter out spent coins for performance.

chia/full_node/coin_store.py
Rigidity: The more you know :P

chia/full_node/coin_store.py
Rigidity: Yes, my bad - this code used to do something different but it no longer does now that I've changed how it does the querying.

chia/full_node/coin_store.py
altendky: The comma makes the tuple...  except in the case of an empty tuple where the parentheses do.  Or as a function parameter where the parentheses clarify the commas are not delimiting function arguments.  See, totally simple...  `:]`

chia/full_node/coin_store.py
Rigidity: I'm not sure what you mean? They aren't included in the query if both are True.

Rigidity: Though if both are False (for some reason), it's probably better to return an empty list early to save time.

arvidn: what I mean is that if your SQL query includes `AND cr.spent_index=0 AND cr.spent_index>0`, it seems a bit redundant. But I see that I misread this. I think the `include_spent` and `unclude_inspent` don't quite mean what their names suggest. you would think that both being true would mean that both spent and unspent are included

tests/core/full_node/stores/test_coin_store.py
Rigidity: Made it more explicitly check the contents of the sorted output

tests/core/full_node/stores/test_coin_store.py
Rigidity: Yes, I was worried about the setup time taking too long, but in practice it'll probably be at most a second or two extra per case, which isn't terrible I guess.

chia/full_node/coin_store.py
Rigidity: It's a bit confusing since it's backwards, but they do mean what they say. If `include_spent` is `False`, then you need to only select unspent coins. Otherwise you can include the spent ones too. And if `include_unspent` is `False`, then similarly you only select spent coins. If these are `True`, you don't need to add a filter since it's the default to include everything already. Hence the `'' if` (no additional filter). I think it works the way you want already - this is the same logic used above in other functions.

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,95 @@ async def get_coin_states_by_ids(
...
+                    puzzle_hashes_db + (min_height, min_height, max_items + 1),
+                )
+
+            for row in await cursor.fetchall():
+                coin_states.append(self.row_to_coin_state(row))
+
+            is_finished: bool
+
+            if len(coin_states) <= max_items:
+                next_height = min_height
```
arvidn, line 485: I suppose once `is_finished` is `True`, you don't really care about `next_height` anymore, but it seems misleading to set it to `min_height`. I'd think you should either set it to the height of `coin_states[-1]` or `None`

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,186 @@ async def test_get_coin_states(db_version: int) -> None:
...
+
+            all_coin_states.sort(key=lambda cs: cs.coin.amount)
+
+            assert len(all_coin_states) == len(expected_records)
+
+            for i in range(len(expected_records)):
+                assert expected_records[i].coin.name().hex() == all_coin_states[i].coin.name().hex(), i
+
+        # Make sure all of the coin states are found when batching.
+        await test_states(coin_records, include_spent=True, include_unspent=True, include_hinted=True)
```
arvidn, line 579: you have three binary options, I would expect (at least) 8 test cases to cover all combinations. I see 5 calls to `test_states()`. `include_hinted` = `False` has only one test case

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,186 @@ async def test_get_coin_states(db_version: int) -> None:
...
+                    remaining_phs[:15000],
+                    min_height=height,
+                    include_spent=include_spent,
+                    include_unspent=include_unspent,
+                    include_hinted=include_hinted,
+                )
+                all_coin_states += coin_states
+
+                if is_finished:
+                    remaining_phs = remaining_phs[15000:]
```
arvidn, line 565: what happens if the number of puzzle hashes exceed the limit? Is there an assert?
Now it looks like the caller can just assume that all puzzle hashes that were requested will be returned.

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,95 @@ async def get_coin_states_by_ids(
...
+                is_finished = True
+            else:
+                next_coin_state = coin_states.pop()
+                next_height = uint32(max(next_coin_state.created_height or 0, next_coin_state.spent_height or 0))
+
+                while len(coin_states) > 0:
```
arvidn, line 491: I think this warrants a comment. It might not be obvious why you remove all coin states matching the last height

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,95 @@ async def get_coin_states_by_ids(
...
+                )
+
+            for row in await cursor.fetchall():
+                coin_states.append(self.row_to_coin_state(row))
+
+            is_finished: bool
+
+            if len(coin_states) <= max_items:
+                next_height = min_height
+                is_finished = True
```
arvidn, line 486: If you make this case an early exit from the function, you no longer need the `is_finished` variable and you can de-dent the whole else-block

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,186 @@ async def test_get_coin_states(db_version: int) -> None:
...
+
+            while not is_finished:
+                (coin_states, height, is_finished) = await coin_store.batch_coin_states_by_puzzle_hashes(
+                    remaining_phs[:15000],
+                    min_height=height,
+                    include_spent=include_spent,
+                    include_unspent=include_unspent,
+                    include_hinted=include_hinted,
+                )
+                all_coin_states += coin_states
```
arvidn, line 562: I think it would make sense to assert that all coin states are returned in block-height order, before you sort them by `amount`

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,186 @@ async def test_get_coin_states(db_version: int) -> None:
...
+                    if len(remaining_phs) > 0:
+                        height = uint32(0)
+                        is_finished = False
+
+            all_coin_states.sort(key=lambda cs: cs.coin.amount)
+
+            assert len(all_coin_states) == len(expected_records)
+
+            for i in range(len(expected_records)):
+                assert expected_records[i].coin.name().hex() == all_coin_states[i].coin.name().hex(), i
```
arvidn, line 576: would it not work to check all aspects of the returned coin states, rather than just the names?

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,186 @@ async def test_get_coin_states(db_version: int) -> None:
...
+
+                    if len(remaining_phs) > 0:
+                        height = uint32(0)
+                        is_finished = False
+
+            all_coin_states.sort(key=lambda cs: cs.coin.amount)
+
+            assert len(all_coin_states) == len(expected_records)
+
+            for i in range(len(expected_records)):
```
arvidn, line 575: I don't see where `expected_records` is sorted by `amount`. How come this works?
Are they generated in the `amount` order?

arvidn (commented):
maybe it would make sense to generate the coins in height order instead, and compare them directly to what you get back from the coin store, without sorting the coin states by `amount`.

The main challenge would be the cut in the middle where you need to make another request batch

tests/core/full_node/stores/test_coin_store.py
Rigidity: Yeah, I'll change everything to be ordered by height for the tests though

tests/core/full_node/stores/test_coin_store.py
Rigidity: Yes, there's an assertion to prevent this. It's up to the caller of the batch coin states method to know this based on the doc string, or it will raise an error.

Rigidity: The wallet protocol will handle this gracefully though

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,95 @@ async def get_coin_states_by_ids(
...
+                cursor = await conn.execute(
+                    f"SELECT confirmed_index, spent_index, coinbase, puzzle_hash, "
+                    f"coin_parent, amount, timestamp FROM coin_record INDEXED BY coin_puzzle_hash "
+                    f'WHERE puzzle_hash in ({"?," * (puzzle_hash_count - 1)}?) '
+                    f"AND (confirmed_index>=? OR spent_index>=?) "
+                    f"{height_filter} "
+                    f"ORDER BY MAX(confirmed_index, spent_index) ASC "
+                    f"LIMIT ?",
+                    puzzle_hashes_db + (min_height, min_height, max_items + 1),
+                )
```
emlowe, line 477: be nice if we could avoid so much duplication here but I don't have any suggestions

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,95 @@ async def get_coin_states_by_ids(
...
+        include_unspent: bool = True,
+        include_hinted: bool = True,
+        max_items: int = 50000,
+    ) -> Tuple[List[CoinState], Optional[uint32]]:
+        """
+        Returns the coin states, as well as the next block height (or `None` if finished).
+        Note that the maximum number of puzzle hashes is currently set to 15000.
+        """
+
+        assert len(puzzle_hashes) <= 15000
```
emlowe, line 429: Should this be hard-coded like this? Perhaps a config item - or maybe even a part of the protocol version in some way? hard-coded numbers just tend to be problems eventually

emlowe:
Would also be useful to look into why the tests don't exercise line 495 as the coverage diff reports

chia/full_node/coin_store.py
Rigidity: Currently this code won't work if the limit is above half of the SQLite variable limit (which sometime in Python 3.8 the SQLite version was raised such that the limit is around 32,700, rather than 999). I haven't found a performant enough solution that doesn't involve using variables like this.

That's why this number is fixed to this value, since raising it could cause issues in the query (I believe it would just ignore the rest or something like that). I could make it a constant or something though, maybe.

I am implementing the protocol in such a way that this limit is mostly irrelevant and can change at any time without breaking the protocol (it will return the ignored puzzle hashes so you can do it in multiple passes on the wallet side if handled properly).

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,183 @@ async def test_get_coin_states(db_version: int) -> None:
...
+        coin_store = await CoinStore.create(db_wrapper)
+        hint_store = await HintStore.create(db_wrapper)
+
+        await coin_store._add_coin_records(batch_coin_records)
+        await hint_store.add_hints(batch_hints)
+
+        # Helper for syncing all of the coin states.
+        async def test_states(
+            expected_records: List[CoinRecord], *, include_spent: bool, include_unspent: bool, include_hinted: bool
+        ) -> None:
```
arvidn, line 555: it might be easier to read this code if you would inline this function. At least it doesn't look like it serves a purpose anymore, since you parameterized the test

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,183 @@ async def test_get_coin_states(db_version: int) -> None:
...
+
+        # Make sure all of the coin states are found when batching.
+        ph_set = set(batch_puzzle_hashes)
+        expected_crs = []
+        for cr in batch_coin_records:
+            if cr.spent_block_index == 0 and not include_unspent:
+                continue
+            elif cr.spent_block_index > 0 and not include_spent:
+                continue
+            elif cr.coin.puzzle_hash not in ph_set and not include_hinted:
```
arvidn, line 605: super nit-pick: I think it would be just a tiny bit easier to read if you turned these `elif` -> `if`. But maybe that's just how I model this loop in my mind

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,183 @@ async def test_get_coin_states(db_version: int) -> None:
...
+        await coin_store._add_coin_records(batch_coin_records)
+        await hint_store.add_hints(batch_hints)
+
+        # Helper for syncing all of the coin states.
+        async def test_states(
+            expected_records: List[CoinRecord], *, include_spent: bool, include_unspent: bool, include_hinted: bool
+        ) -> None:
+            height: Optional[uint32] = uint32(0)
+            all_coin_states: List[CoinState] = []
+            remaining_phs = batch_puzzle_hashes.copy()
```
arvidn, line 558: in order to cover the hint case, you need to include the `batch_hints` here too, don't you?

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -493,6 +495,183 @@ async def test_get_coin_states(db_version: int) -> None:
...
+    puzzle_hash = std_hash(i.to_bytes(4, byteorder="big"))
+
+    coin = Coin(
+        std_hash(b"Parent Coin Id " + i.to_bytes(4, byteorder="big")),
+        puzzle_hash,
+        uint64(1000),
+    )
+
+    if is_hinted:
+        hint = std_hash(b"Hinted " + puzzle_hash)
```
arvidn, line 518: unless I'm missing something, this hint won't ever match a puzzle hash that you're selecting for in your query. in this test, do you ever pick up *any* coin records because `include_hinted` is `True`?
See comment below

arvidn (commented):
I think the fixture is an important change to make.
I'm also not convinced you test the `include_hints` case. The SQL query is quite complex, I'm not confident that I can tell that it's correct (until we have a solid test)

tests/core/full_node/stores/test_coin_store.py
Rigidity: `batch_hints` is only used for inserting into the `HintStore`. The actual hint is added to `batch_puzzle_hashes` for querying.

tests/core/full_node/stores/test_coin_store.py
Rigidity: I believe it does

chia/full_node/coin_store.py
arvidn: I think it's hard to reduce duplication without obfuscating the query

chia/full_node/coin_store.py
```diff
@@ -411,6 +411,95 @@ async def get_coin_states_by_ids(
...
+                # There are no coins which are both spent and unspent, so we're finished.
+                return [], None
+
+            if include_hinted:
+                cursor = await conn.execute(
+                    f"SELECT cr.confirmed_index, cr.spent_index, cr.coinbase, cr.puzzle_hash, "
+                    f"cr.coin_parent, cr.amount, cr.timestamp FROM coin_record cr "
+                    f"LEFT JOIN hints h ON cr.coin_name = h.coin_id "
+                    f'WHERE (cr.puzzle_hash in ({"?," * (puzzle_hash_count - 1)}?) '
+                    f'OR h.hint in ({"?," * (puzzle_hash_count - 1)}?)) '
```
arvidn, line 460: a crazy idea for the future might be to create a temporary table with all the puzzle hashes and join with it, rather than passing them all (twice) as parameters.

We could even have separate paths for when we have few puzzle hashes and when we have many (say, more than 15000). See [sqlite temp database](https://www.sqlite.org/tempfiles.html#temp_databases).

arvidn (dismissed):
looks good! Feel free to resolve my comments. Maybe we should let Earle have another pass before we land as well

chia/full_node/coin_store.py
Rigidity: Interesting, yeah this might enable us to increase the limit later. Would just have to make sure it's performant enough.

chia/full_node/coin_store.py
emlowe: Perhaps some comment in the code that this limit is related to the sqlite parameter limit

arvidn (approved):

PR #17305, opened 2024-01-12
switch transaction queue completion to a new `ValuedEvent`

altendky:
### Purpose:

https://github.com/Chia-Network/chia-blockchain/pull/17171 was the introducing pr. 
 it seems this error would have been triggered when the api handler timed out waiting for the transaction to be processed and the transaction processing later completed.  i didn't feel great about using the future there and ended up getting bit by it.  when awaiting a future gets cancelled it marks the future itself as cancelled.  when something later attempts to set a result on it the invalid state error is raised.  in this case, i believe the only effect of this bug is the logging of the error.

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -473,7 +473,7 @@ class FullNode:
         peer = entry.peer
         try:
             inc_status, err = await self.add_transaction(entry.transaction, entry.spend_name, peer, entry.test)
-            entry.done.set_result((inc_status, err))
+            entry.done.set((inc_status, err))
         except asyncio.CancelledError:
             error_stack = traceback.format_exc()
             self.log.debug(f"Cancelling _handle_one_transaction, closing: {error_stack}")
--- a/chia/full_node/full_node_api.py
+++ b/chia/full_node/full_node_api.py
@@ -1272,7 +1272,7 @@ class FullNodeAPI:
         await self.full_node.transaction_queue.put(queue_entry, peer_id=None, high_priority=True)
         try:
             with anyio.fail_after(delay=45):
-                status, error = await queue_entry.done
+                status, error = await queue_entry.done.wait()
         except TimeoutError:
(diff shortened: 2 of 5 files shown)


chia/full_node/full_node_api.py
```diff
@@ -1272,7 +1272,7 @@ async def send_transaction(
         await self.full_node.transaction_queue.put(queue_entry, peer_id=None, high_priority=True)
         try:
             with anyio.fail_after(delay=45):
-                status, error = await queue_entry.done
+                status, error = await queue_entry.done.wait()
```
arvidn, line 1275: wouldn't it be simpler (and maybe more robust) to just use:
```
asyncio.wait_for(queue_entry, 45)
```

instead of the `fail_after()` context manager?

chia/full_node/full_node.py
```diff
@@ -473,7 +473,7 @@ async def _handle_one_transaction(self, entry: TransactionQueueEntry) -> None:
         peer = entry.peer
         try:
             inc_status, err = await self.add_transaction(entry.transaction, entry.spend_name, peer, entry.test)
-            entry.done.set_result((inc_status, err))
+            entry.done.set((inc_status, err))
```
arvidn, line 476: `done` is a `Future`, right? I can't see `set()` documented as a member function

chia/full_node/full_node_api.py
altendky: Using `asyncio.wait_for()` wouldn't be terrible, but I don't think it is better.  Using `anyio.fail_after()` has a few benefits in my mind.

1) one pattern that can handle more cases
2) no callbacks be passed around (they can be useful, but not my preferred pattern when easily avoided)
3) in general, i hope to shift away from asyncio and towards anyio for most things due to them leading towards better patterns in general.  if we can consistently use anyio that will make it more clear what people should use in new code.

Now sure, transitions take time and there may be cases where we continue to use asyncio directly, but I think they should have distinct valuable reasons.

arvidn (approved):

PR #17311, opened 2024-01-12
Use Monkeypatching for request_proof_of_weight and request_block_header in test_bad_peak_mismatch

AmineKhaldi:


tests/wallet/sync/test_wallet_sync.py
```diff
@@ -1283,7 +1285,7 @@ async def test_bad_peak_mismatch(two_wallet_nodes, default_1000_blocks, self_hos
     )
     f = asyncio.Future()
     f.set_result(wp_msg)
-    full_node_api.request_proof_of_weight = MagicMock(return_value=f)
+    monkeypatch.setattr(full_node_api, "request_proof_of_weight", MagicMock(return_value=f))
```
emlowe, line 1288: I like the pattern with using the context manager: `with monkeypatch.context() as m:` 
automatic cleanup FTW

emlowe (changes requested):

tests/wallet/sync/test_wallet_sync.py
AmineKhaldi: I considered it, but it has an indentation cost, and this patching is needed throughout the test run so it seemed fine to me. What do you think?

emlowe (approved):

PR #17316, opened 2024-01-12
Annotate wallet_rpc_client.py

AmineKhaldi:


chia/rpc/wallet_rpc_client.py
```diff
@@ -43,97 +43,117 @@ class WalletRpcClient(RpcClient):
     """
 
     # Key Management APIs
-    async def log_in(self, fingerprint: int) -> Dict:
+    async def log_in(self, fingerprint: int) -> Union[Dict[str, Any], Any]:
```
arvidn, line 46: `Union[..., Any]` seems a bit meaningless. that's really just `Any`, isn't it?
Is it the failure case that returns `Any`?

AmineKhaldi: We have an exception path that returns `e.args[0]` which is `Any`, and success/failure paths return `Dict[str, Any]`. What do you think?

arvidn (approved):

arvidn:
I suppose we don't have any test coverage for wallet_rpc_client

PR #17324, opened 2024-01-12
CHIA-69 remove task for long sync

wjblanke:
### Purpose:

remove task for long sync

### Current Behavior:

creates new task for long sync which could be double entered, killed, etc causing issues for sync primitives

### New Behavior:

just wait on the async function. new_peak already runs in its own task and allows 2 simultaneously.

### Testing Notes:

None

### DIFF
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -148,7 +148,7 @@ class FullNode:
     subscriptions: PeerSubscriptions = dataclasses.field(default_factory=PeerSubscriptions)
     _transaction_queue_task: Optional[asyncio.Task[None]] = None
     simulator_transaction_callback: Optional[Callable[[bytes32], Awaitable[None]]] = None
-    _sync_task: Optional[asyncio.Task[None]] = None
+    _sync_bool: bool = False
     _transaction_queue: Optional[TransactionQueue] = None
     _compact_vdf_sem: Optional[LimitedSemaphore] = None
     _new_peak_sem: Optional[LimitedSemaphore] = None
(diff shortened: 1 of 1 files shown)


altendky (changes requested):
I believe this will change to hold the new peak semaphore during the sync.  This should be discussed anyways.

chia/full_node/full_node.py
```diff
@@ -765,9 +761,16 @@ async def new_peak(self, request: full_node_protocol.NewPeak, peer: WSChiaConnec
                 if await self.short_sync_batch(peer, uint32(max(curr_peak_height - 6, 0)), request.height):
                     return None
 
+            if self._sync_bool:
+                return None
```
arvidn, line 765: iiuc, this is the key change, really. I expect you could just check to see if `self._sync_task is not None` here. except, I suppose we would need to check if the task is complete, and join it and clear the task handle somewhere as well.

I think kyle's comment seems relevant. It seems risky to perform the sync inside the call. Current and/or future callers of `new_peak()` may not expect it to be held up for the full sync duration.

arvidn:
I think this issue requires some more research. The fundamental problem is that we launch (and replace the handle) for sync tasks multiple times. I've confirmed that we do this, especially if you yank the network cord and put it back in.

Inside the task, we have a check to see if we're already syncing (by asking the `sync_store`). But at that time we've already overwritten the `self._sync_task` handle. It would be best to have this check outside the task itself.

I made an experiment where I would just not start another task if the handle was already set. We never clear the handle, so I also wrapped the task function to clear it whenever it exited. I ran this experiment by turning off internet while syncing, then turning it back on again. It seemed to recover for a little bit, but then stalled. It's hard to tell whether this caused it to stall or whether it would have stalled anyway. I was surprised to see how many times we attempted to create another sync task though.

altendky:
Setting outside and clearing inside will not be good.  If a task is cancelled before it runs then the body of the function is never reached so even `finally:` code won't get run.  And just checking outside isn't enough, it needs to be check-and-set as an atomic action (on the task reference, a boolean, a lock, this applies to them all).  So if we have the check-and-set outside of the task and the clear inside we still end up being able to miss the 'release' (setting a boolean to false, releasing a lock, setting a task reference to `None` all count as releasing in this pattern).

PR #17340, opened 2024-01-16
New subscription and wallet sync protocol

Rigidity:
### Purpose:

The wallet protocol is lacking some essential functionality for syncing wallets and requesting coin data. This improves it with new messages for fetching coin state and subscribing.

### Existing Limitations:

- You cannot unsubscribe from puzzle hashes or coin ids without disconnecting from the peer. Once you reach the limit you have to reset the connection or connect to additional peers.
- If a puzzle hash has too many coins to fit in one message, the response will be truncated. This is not a reasonable outcome for many use cases.
- You cannot request coin state states without subscribing for future updates, which is unnecessary and means the subscription limit will eventually be reached.

### New Features

- Removing subscriptions.
- Requesting coin state without subscribing.
- Filtering coin states by amount, hint, and spent status.
- Pagination when syncing coin state for puzzle hashes.

#### Remove Puzzle Subscriptions

```py
class RequestRemovePuzzleSubscriptions:
    puzzle_hashes: Optional[List[bytes32]]

class RespondRemovePuzzleSubscriptions:
    puzzle_hashes: List[bytes32]
```

Removes puzzle hashes from the subscription list (or all of them if `None`), returning the hashes that were actually removed.

#### Remove Coin Subscriptions

```py
class RequestRemoveCoinSubscriptions:
    coin_ids: Optional[List[bytes32]]

class RespondRemoveCoinSubscriptions:
    coin_ids: List[bytes32]
```

Removes coin ids from the subscription list (or all of them if `None`), returning the ids that were actually removed.

#### Request Puzzle State

```py
class RequestPuzzleState:
    puzzle_hashes: List[bytes32]
    previous_height: Optional[uint32]
    header_hash: bytes32
    filters: CoinStateFilters
    subscribe_when_finished: bool

class RespondPuzzleState:
    puzzle_hashes: List[bytes32]
    height: uint32
    header_hash: bytes32
    is_finished: bool
    coin_states: List[CoinState]

class RejectPuzzleState:
    reason: uint8 # RejectStateReason

class CoinStateFilters:
    include_spent: bool
    include_unspent: bool
    include_hinted: bool
    min_amount: uint64
```

Requests coin states that match the given puzzle hashes (or hints).

Unlike `RegisterForPhUpdates`, this does not add subscriptions for the puzzle hashes automatically.
When `subscribe_when_finished` is set to `True`, it will add subscriptions, but only once the last batch has been requested.

As well as this, previously it was impossible to get all coin records if the number of items exceeded the limit.
This implementation allows you to continue where you left off with `previous_height` and `header_hash`.

If a reorg of relevant blocks occurs while doing so, `previous_height` will no longer match `header_hash`.
This can be handled by a wallet by simply backtracking a bit, or restarting the sync from genesis. It could be inconvenient, but at least you can detect it.
In the event that a reorg is detected by a node, `RejectPuzzleState` will be returned. This is the only scenario it will be rejected directly like this.

Additionally, it is now possible to filter out spent, unspent, or hinted coins, as well as coins below a minimum amount.
This can reduce the risk of spamming or DoS of a wallet in some cases, and improve performance.

If `previous_height` is `None`, you are syncing from genesis. The `header_hash` should match the genesis challenge of the network you are connected to.

#### Request Coin State

```py
class RequestCoinState:
    coin_ids: List[bytes32]
    previous_height: Optional[uint32]
    header_hash: bytes32
    subscribe: bool

class RespondCoinState:
    coin_ids: List[bytes32]
    coin_states: List[CoinState]

class RejectCoinState:
    reason: uint8 # RejectStateReason
```

Request coin states that match the given coin ids.

Unlike `RegisterForCoinUpdates`, this does not add subscriptions for the coin ids automatically.
When `subscribe` is set to `True`, it will add and return as many coin ids to the subscriptions list as possible.

Unlike the new `RequestPuzzleState` message, this does not implement batching for simplicity. The order is also not guaranteed.
However, you can still specify the `previous_height` and `header_hash` to start from.

If a reorg of relevant blocks has occurred, `previous_height` will no longer match `header_hash`.
This can be handled by a wallet depending on the use case (for example by restarting from zero). It could be inconvenient, but at least you can detect it.
In the event that a reorg is detected by a node, `RejectCoinState` will be returned. This is the only scenario it will be rejected directly like this.

If `previous_height` is `None`, you are syncing from genesis. The `header_hash` should match the genesis challenge of the network you are connected to.

#### Reject State Reason

```py
class RejectStateReason(IntEnum):
    REORG = 0
    EXCEEDED_SUBSCRIPTION_LIMIT = 1
```

### Concerns

The data which a full node gives you cannot be trusted unless it is your own node.
These additions to the protocol make no effort to make untrusted data any easier to validate.
The best method I am aware of is still just to cross reference with other peers and ban outliers.

### Checklist

- [x] Waiting for https://github.com/Chia-Network/chia-blockchain/pull/17300 to be merged.
- [x] Design and document the new protocol messages.
- [x] Implement the protocol messages in the full node.
- [x] Write unit tests for these subscription and coin state protocol messages.
- [x] Update serialization and regression tests with these messages.
- [x] Write integration test for syncing a wallet from zero using this new sync protocol.
- [x] Write integration test for syncing from the last known height, and recovering when a reorg occurs.
- [x] Write integration test for a reorg occurring during a wallet's sync.
- [x] Ensure protocol version is correct, and doesn't need to be bumped again.
- [x] Implement protocol messages in Rust https://github.com/Chia-Network/chia_rs/pull/447 to keep parity.


richardkiss:
In general, there are opportunities to improve the protocol design as new messages are added. Subscriptions essentially open a "channel" multiplexed over the message stream, so we could introduce an explicit channel concept - just an incrementing nonce namespaced on each websocket. 

The request could allow specifying a start block height, so that we don't sync from genesis each time. Wallets can track a "current height" per puzzle hash.

Perhaps something like:

```
# non-message struct
class PuzzleSubscriptionRequest:  
  puzzle_hash: bytes32
  height: uint32

# message
class RequestAddPuzzleSubscriptions:
  requests: List[PuzzleSubscriptionRequest]

# non-message struct
class RespondSubscriptionRequest:
  channel: uint32
  puzzle_hash: bytes32  
  height: uint32
  header_hash: bytes32

# message
class RespondAddPuzzleSubscriptions:
  requests: List[RespondSubscriptionRequest]
```

This makes unsubscribe simpler:

```
class RequestRemoveSubscriptions:
  channel: List[uint32]

class RespondRemoveSubscriptions:
  channel: List[uint32]
```

Update messages might look something like:

```
class SubscriptionUpdate:
  channel: uint32
  updates: list[(info about coin spend)] 
  completed_updated_to_height: uint32
```

The `completed_updated_to_height` field is a height that we have completely exhausted all updates for. This might be much larger than the block of the last update (if we see no further spends for example) or ONE LESS than the block the last update occurs in (if there are further updates in the same block that will not fit into this message). Essentially, it's the value the wallet can later subscribe to and be assured to have not missed an update.

richardkiss:
Generally speaking, having multiple potential responses to a particular request is annoying. It means when a client makes a request, it has multiple message IDs it needs to wait on to find a response.

The serialization protocol doesn't support `enum` types, and there's no obvious way to see how to do it. However, we could simply have each different potential response be `Optional`. As an example, rather than having two messages each with their own id:

```
class RespondPuzzleState:
    puzzle_hashes: List[bytes32]
    next_height: Optional[uint32]
    next_header_hash: Optional[bytes32]
    coin_states: List[CoinState]

class RejectPuzzleState:
    header_hash: Optional[bytes32]
```

we could merge them:

```
# used in `RespondPuzzleState`
class ResponsePuzzleStateOK:
    puzzle_hashes: List[bytes32]
    next_height: Optional[uint32]
    next_header_hash: Optional[bytes32]
    coin_states: List[CoinState]

# used in `RespondPuzzleState`
class ResponsePuzzleStateFail:
    header_hash: Optional[bytes32]

# message
class RespondPuzzleState:
    ok: Optional[ResponsePuzzleStateOK]
    fail: Optional[ResponsePuzzleStateFail]
```

Here, exactly one of `ok` and `fail` should be not `None`. We can't enforce this through type-checking, but we can enforce at runtime, and simply fail on messages that don't comply like we would any other ill-formed message (for example, disconnect from or otherwise punish the remote).

Rigidity:
> In general, there are opportunities to improve the protocol design as new messages are added. Subscriptions essentially open a "channel" multiplexed over the message stream, so we could introduce an explicit channel concept - just an incrementing nonce namespaced on each websocket.
> 
> The request could allow specifying a start block height, so that we don't sync from genesis each time. Wallets can track a "current height" per puzzle hash.
> 
> Perhaps something like:
> 
> ```
> # non-message struct
> class PuzzleSubscriptionRequest:  
>   puzzle_hash: bytes32
>   height: uint32
> 
> # message
> class RequestAddPuzzleSubscriptions:
>   requests: List[PuzzleSubscriptionRequest]
> 
> # non-message struct
> class RespondSubscriptionRequest:
>   channel: uint32
>   puzzle_hash: bytes32  
>   height: uint32
>   header_hash: bytes32
> 
> # message
> class RespondAddPuzzleSubscriptions:
>   requests: List[RespondSubscriptionRequest]
> ```
> 
> This makes unsubscribe simpler:
> 
> ```
> class RequestRemoveSubscriptions:
>   channel: List[uint32]
> 
> class RespondRemoveSubscriptions:
>   channel: List[uint32]
> ```
> 
> Update messages might look something like:
> 
> ```
> class SubscriptionUpdate:
>   channel: uint32
>   updates: list[(info about coin spend)] 
>   completed_updated_to_height: uint32
> ```
> 
> The `completed_updated_to_height` field is a height that we have completely exhausted all updates for. This might be much larger than the block of the last update (if we see no further spends for example) or ONE LESS than the block the last update occurs in (if there are further updates in the same block that will not fit into this message). Essentially, it's the value the wallet can later subscribe to and be assured to have not missed an update.

I don't understand the benefit of this, it seems like a more complicated API to do roughly the same thing. Subscribing doesn't actually return any coin states, it just registers that you're interested in a given set of puzzle hashes going forward whenever coin states are updated.

The `RequestPuzzleState` message allows you to specify the height and header hash to start from, like you're describing, and you keep requesting more coin states until you've reached the peak or you decide to stop.

Rigidity:
> Generally speaking, having multiple potential responses to a particular request is annoying. It means when a client makes a request, it has multiple message IDs it needs to wait on to find a response.
> 
> The serialization protocol doesn't support `enum` types, and there's no obvious way to see how to do it. However, we could simply have each different potential response be `Optional`. As an example, rather than having two messages each with their own id:
> 
> ```
> class RespondPuzzleState:
>     puzzle_hashes: List[bytes32]
>     next_height: Optional[uint32]
>     next_header_hash: Optional[bytes32]
>     coin_states: List[CoinState]
> 
> class RejectPuzzleState:
>     header_hash: Optional[bytes32]
> ```
> 
> we could merge them:
> 
> ```
> # used in `RespondPuzzleState`
> class ResponsePuzzleStateOK:
>     puzzle_hashes: List[bytes32]
>     next_height: Optional[uint32]
>     next_header_hash: Optional[bytes32]
>     coin_states: List[CoinState]
> 
> # used in `RespondPuzzleState`
> class ResponsePuzzleStateFail:
>     header_hash: Optional[bytes32]
> 
> # message
> class RespondPuzzleState:
>     ok: Optional[ResponsePuzzleStateOK]
>     fail: Optional[ResponsePuzzleStateFail]
> ```
> 
> Here, exactly one of `ok` and `fail` should be not `None`. We can't enforce this through type-checking, but we can enforce at runtime, and simply fail on messages that don't comply like we would any other ill-formed message (for example, disconnect from or otherwise punish the remote).

We don't have any precedent for doing it this way, and this increases the size of the protocol messages and requires differentiating between the response types just the same. Message responses are identified by an id, so rather than listening for a given response type you just wait until you receive a message with the same nonce as was passed in the request.

It's pretty simple to do, here is an example I have (a bit outdated, but it shows the idea):
https://github.com/Chia-Network/chia_rs/pull/369/files#diff-7f65462d1a9e187413a5013428d70e00a201bec3b2d7e0b26ce7192cbc281462R275

Rigidity:
You will usually use `RequestPuzzleState` and `RequestCoinState` as an entrypoint for subscribing, since it's more resilient to reorgs and more concise. There's a case to be made for removing the `AddSubscriptions` messages entirely if they don't have a use-case.

chia/full_node/full_node.py
```diff
@@ -2351,26 +2343,71 @@ async def add_transaction(
...
+        wallet_update = wallet_protocol.TransactionAddedUpdate(
+            mempool_item.spend_bundle_name,
+            mempool_item.height_added_to_mempool,
+            mempool_item.cost,
+            mempool_item.additions,
+            mempool_item.removals,
+        )
+        wallet_update_msg = make_msg(ProtocolMessageTypes.transaction_added_update, wallet_update)
+
+        peers = await self.peers_for_mempool_item(mempool_item)
```
arvidn, line 2378: it would probably make sense to call this earlier and have an early exit in case `peers == []`

chia/full_node/full_node.py
```diff
@@ -2351,26 +2343,71 @@ async def add_transaction(
...
+
+        def new_clients(conn: WSChiaConnection) -> bool:
+            return conn.peer_node_id in peers and conn.protocol_version > Version("0.0.36")
+
+        await self.server.send_to_all_if(
+            [wallet_update_msg], NodeType.WALLET, new_clients, None if peer is None else peer.peer_node_id
```
arvidn, line 2384: it seems questionable that we send this new message to you unsolicited. At least you've only subscribed to blockchain updates, not explicitly to mempool updates. Although, arguably you would be interested in these updates too, most likely.

Did you consider a protocol where the wallet explicitly opts-in to receiving these updates?
If you do that, you also don't need to filter based on protocol version, only wallets supporting the new protocol will have opt-in.

Also, @almogdepaz prefers advertizing support for a new protocol using capabilities. I think this is a case where that might make sense, since it's not replacing an old protocol (that we would like to get rid of in the future).

chia/full_node/full_node.py
```diff
@@ -2351,26 +2343,71 @@ async def add_transaction(
...
+            peers_for_ph = self.subscriptions.peers_for_puzzle_hash(coin.puzzle_hash)
+            peers.update(peers_for_ph)
+
+            coin_id = coin.name()
+            peers_for_coin = self.subscriptions.peers_for_coin_id(coin_id)
+            peers.update(peers_for_coin)
+
+            coin_ids.append(coin_id)
+
+        hints = [bytes32(hint) for hint in await self.hint_store.get_hints_for_coins(coin_ids) if len(hint) == 32]
```
arvidn, line 2403: this is an expensive operation. Are there ways to avoid these lookups unconditionally, every time the mempool is updated?

arvidn (commented):
seems reasonable

chia/full_node/full_node.py
Rigidity: This would be for e.g. detecting that one of your NFTs have been spent or created. I don't know of a faster way to do this, at least it's one query

richardkiss:
> We don't have any precedent for doing it this way, and this increases the size of the protocol messages and requires differentiating between the response types just the same. Message responses are identified by an id, so rather than listening for a given response type you just wait until you receive a message with the same nonce as was passed in the request.

I did notice a `Optional[uint16]` in the wrapper `Message` type. I thought I tried using that and it was `None`, but I will try again. It's definitely the best way I can think of to route messages, by mapping the responses to their requests with a nonce.

richardkiss:
> I don't understand the benefit of this, it seems like a more complicated API to do roughly the same thing. But maybe I'm misunderstanding something? Subscribing doesn't actually return any coin states, it just registers that you're interested in a given set of puzzle hashes going forward whenever coin states are updated.

For request/responses, I see the signature conceptually as

```
class SomeRemoteObject:
    async def some_request(self, foo: SerializableType1, bar: SerializableType2) -> SerializableType3:
        ...
```

For subscriptions, I see it as

```
class SomeRemoteObject:
    async def subscribe_to_something(self, identifier: SerializableIdentifier) -> AsyncIterator[SubscriptionUpdate]:
        ...
```

Internal to this, you'd use a nonce to identify subscriptions. Remote messages containing subscription updates could use that subscription nonce to route it to the correct async iterator, and also use that nonce to inform the remote to cancel the subscription.

This makes local use really easy. You get the async iterator with the subscription api, then simply `await` to get items out.

[As I write this, I realize you probably actually want to return an async context manager so you can more affirmatively cancel the subscription with a remote message no matter how you exit locally.]

I realize these comments aren't really about the code. But I've been trying to write code to connect to remote chia nodes and I've found it to be much more complicated than it should be, and how certain changes could allow for better abstractions and smaller code. But maybe you don't want to tackle this now.

Rigidity:
> > We don't have any precedent for doing it this way, and this increases the size of the protocol messages and requires differentiating between the response types just the same. Message responses are identified by an id, so rather than listening for a given response type you just wait until you receive a message with the same nonce as was passed in the request.
> 
> 
> 
> I did notice a `Optional[uint16]` in the wrapper `Message` type. I thought I tried using that and it was `None`, but I will try again. It's definitely the best way I can think of to route messages, by mapping the responses to their requests with a nonce.

It should be the same value as the request id, if that's left as None the response will be too. If it's an update (ie CoinStateUpdate), it will also be None.

chia/full_node/full_node.py
arvidn: yeah, not advertising the capability would be opting out

arvidn:
@richardkiss it sounds like the general features of the protocol you're describing matches what @Rigidity has documented, and is implementing.

Some major differences is that in his protocol you subscribe and request coin states in bulk, with many coind IDs and puzzle hashes. Also, subscriptions aren't given shorter names (but are just the coin ID or puzzle hash). But the updates are still delivered to the client in block-height order, making it possible for the wallet to be interrupted and resuming from the height it left off.

richardkiss:
> > > We don't have any precedent for doing it this way, and this increases the size of the protocol messages and requires differentiating between the response types just the same. Message responses are identified by an id, so rather than listening for a given response type you just wait until you receive a message with the same nonce as was passed in the request.
> > 
> > 
> > I did notice a `Optional[uint16]` in the wrapper `Message` type. I thought I tried using that and it was `None`, but I will try again. It's definitely the best way I can think of to route messages, by mapping the responses to their requests with a nonce.
> 
> It should be the same value as the request id, if that's left as None the response will be too. If it's an update (ie CoinStateUpdate), it will also be None.

Your postman doesn't need to open an envelope to deliver it. Our envelope is `chia.server.outbound_message.Message` with the three fields, `type`, `id` and `data`.  The `type` and `id` work together to determine where this message should be routed. (IMO, it's a bit silly to use both of these fields for this purpose, but this would be pretty tough to change at this point.)

richardkiss:
> Some major differences is that in his protocol you subscribe and request coin states in bulk, with many coin IDs and puzzle hashes. Also, subscriptions aren't given shorter names (but are just the coin ID or puzzle hash). But the updates are still delivered to the client in block-height order, making it possible for the wallet to be interrupted and resuming from the height it left off

What do the updates actually look like? I don't see that in this discussion. Are we using the existing messages to do this?

Rigidity:
> > Some major differences is that in his protocol you subscribe and request coin states in bulk, with many coin IDs and puzzle hashes. Also, subscriptions aren't given shorter names (but are just the coin ID or puzzle hash). But the updates are still delivered to the client in block-height order, making it possible for the wallet to be interrupted and resuming from the height it left off
> 
> What do the updates actually look like? I don't see that in this discussion. Are we using the existing messages to do this?

RespondPuzzleState is used to respond to incremental requests for the initial state. [CoinStateUpdate](https://docs.chia.net/wallet-protocol/#coin_state_update) is sent every block or when a reorg occurs to give you changes for things you're subscribed to. I do not intend to drastically change how the protocol works, but rather to add missing functionality and lift some restrictions. Changing how messages work seems pretty out of scope to me.

Rigidity:
> > > > We don't have any precedent for doing it this way, and this increases the size of the protocol messages and requires differentiating between the response types just the same. Message responses are identified by an id, so rather than listening for a given response type you just wait until you receive a message with the same nonce as was passed in the request.
> > > 
> > > 
> > > I did notice a `Optional[uint16]` in the wrapper `Message` type. I thought I tried using that and it was `None`, but I will try again. It's definitely the best way I can think of to route messages, by mapping the responses to their requests with a nonce.
> > 
> > 
> > It should be the same value as the request id, if that's left as None the response will be too. If it's an update (ie CoinStateUpdate), it will also be None.
> 
> Your postman doesn't need to open an envelope to deliver it. Our envelope is `chia.server.outbound_message.Message` with the three fields, `type`, `id` and `data`. The `type` and `id` work together to determine where this message should be routed. (IMO, it's a bit silly to use both of these fields for this purpose, but this would be pretty tough to change at this point.)

It seems reasonable to me, they mean different things. The unique nonce (id) identifies which message it is, and the type identifies which kind of response you're receiving (ie a success or a failure). If you receive anything you don't expect, you ban the peer.

chia/full_node/full_node.py
Rigidity: This will be moving to another PR later.

chia/full_node/full_node.py
Rigidity: Moving to another PR later.

chia/full_node/full_node.py
Rigidity: Will be another PR

Rigidity:
This is ready for initial reviews, though rate limits have not been set, so it is not ready for merge yet. That is what I will work on next.

chia/full_node/coin_store.py
```diff
@@ -463,21 +469,32 @@ async def batch_coin_states_by_puzzle_hashes(
...
-                    f"{height_filter} "
+                    f"{height_filter} {amount_filter}"
                     f"ORDER BY MAX(cr.confirmed_index, cr.spent_index) ASC "
                     f"LIMIT ?",
-                    puzzle_hashes_db + puzzle_hashes_db + (min_height, min_height, max_items + 1),
+                    (
+                        puzzle_hashes_db
+                        + puzzle_hashes_db
+                        + (min_height, min_height)
+                        + ((min_amount.to_bytes(8, "big"),) if min_amount > 0 else ())
```
arvidn, line 479: did you test that `BLOB`s are correctly compared by `>=` in sqlite?
i.e. byte-by-byte

chia/full_node/full_node_api.py
```diff
@@ -1565,12 +1565,8 @@ async def register_interest_in_puzzle_hash(
...
     ) -> Message:
         trusted = self.is_trusted(peer)
-        if trusted:
-            max_subscriptions = self.full_node.config.get("trusted_max_subscribe_items", 2000000)
-            max_items = self.full_node.config.get("trusted_max_subscribe_response_items", 500000)
-        else:
-            max_subscriptions = self.full_node.config.get("max_subscribe_items", 200000)
-            max_items = self.full_node.config.get("max_subscribe_response_items", 100000)
+        max_items = self.max_subscribe_response_items(peer)
+        max_subscriptions = self.max_subscriptions(peer)
```
arvidn, line 1569: this is a nice refactoring (that could land separately)

chia/full_node/full_node_api.py
```diff
@@ -1720,5 +1712,201 @@ def get_fee_estimates(est: FeeEstimatorInterface, req_times: List[uint64]) -> Li
...
+    ) -> Message:
+        peer_id = peer.peer_node_id
+        max_subscriptions = self.max_subscriptions(peer)
+        subs = self.full_node.subscriptions
+
+        peak_height = self.full_node.blockchain.get_peak_height()
+        assert peak_height is not None
+
+        header_hash = self.full_node.blockchain.height_to_hash(peak_height)
+        assert header_hash is not None
```
arvidn, line 1730: it's not obvious why it shouldn't be possible to subscribe to a node that hasn't synced yet. You could just return the genesis challenge as the previous block hash.
I suppose the peak height would be a bit tricky. But I believe you suggested it could be an `Optional[uint32]`

chia/full_node/full_node_api.py
```diff
@@ -1720,5 +1712,201 @@ def get_fee_estimates(est: FeeEstimatorInterface, req_times: List[uint64]) -> Li
...
+        msg = make_msg(ProtocolMessageTypes.respond_add_puzzle_subscriptions, response)
+        return msg
+
+    @api_request(
+        peer_required=True,
+        reply_types=[ProtocolMessageTypes.respond_add_coin_subscriptions],
+    )
+    async def request_add_coin_subscriptions(
+        self, request: wallet_protocol.RequestAddCoinSubscriptions, peer: WSChiaConnection
+    ) -> Message:
```
arvidn, line 1745: here's a crazy idea; maybe there should be an option to automatically unsubscribe to coin IDs after they've been spent and the spend has been buried under some specified number of confirmations

chia/full_node/full_node_api.py
```diff
@@ -1720,5 +1712,201 @@ def get_fee_estimates(est: FeeEstimatorInterface, req_times: List[uint64]) -> Li
...
+    )
+    async def request_remove_puzzle_subscriptions(
+        self, request: wallet_protocol.RequestRemovePuzzleSubscriptions, peer: WSChiaConnection
+    ) -> Message:
+        peer_id = peer.peer_node_id
+        subs = self.full_node.subscriptions
+
+        if request.puzzle_hashes is None:
+            removed = list(subs.puzzle_subscriptions(peer_id))
+            subs.remove_puzzle_subscriptions(peer_id, removed)
```
arvidn, line 1775: I would expect it to be a little bit more efficient to just tell `subs` to remove all the subscriptions for a peer, rather than asking for them and removing them one at a time. but maybe it's not worth the hassle right now.

I suppose we're returning the ones we removed anyway.

chia/full_node/full_node_api.py
```diff
@@ -1720,5 +1712,201 @@ def get_fee_estimates(est: FeeEstimatorInterface, req_times: List[uint64]) -> Li
...
+
+    @api_request(peer_required=True, reply_types=[ProtocolMessageTypes.respond_puzzle_state])
+    async def request_puzzle_state(
+        self, request: wallet_protocol.RequestPuzzleState, peer: WSChiaConnection
+    ) -> Message:
+        max_items = self.max_subscribe_response_items(peer)
+        max_subscriptions = self.max_subscriptions(peer)
+
+        previous_header_hash = (
+            self.full_node.blockchain.height_to_hash(request.previous_height)
```
arvidn, line 1811: I'm hoping that we can phase out the `height_to_hash` function/class. It's not sustainable to keep this in RAM in the long run, and the DB v2 supports this lookup.

However, the closest thing to this is `prev_block_hash(height)`, but it's not quite what you need. I think it would be good to make an `async` version of `height_to_hash()` that looks it up in the DB instead. But maybe putting that in here is scope-creep

chia/full_node/full_node_api.py
```diff
@@ -1720,5 +1712,201 @@ def get_fee_estimates(est: FeeEstimatorInterface, req_times: List[uint64]) -> Li
...
+            msg = make_msg(ProtocolMessageTypes.reject_coin_state, rejection)
+            return msg
+
+        min_height = uint32(request.previous_height + 1 if request.previous_height is not None else 0)
+
+        coin_states = await self.full_node.coin_store.get_coin_states_by_ids(
+            True,
+            set(request.coin_ids),
+            min_height=min_height,
+            max_items=max_items,
```
arvidn, line 1884: if the response is truncated here, we don't do anything to tell the wallet or anything. This looks like a potential source of corrupt wallet balances.

tests/core/full_node/stores/test_coin_store.py
```diff
@@ -552,12 +552,14 @@ def random_coin_records() -> RandomCoinRecords:
 @pytest.mark.parametrize("include_spent", [True, False])
 @pytest.mark.parametrize("include_unspent", [True, False])
 @pytest.mark.parametrize("include_hinted", [True, False])
+@pytest.mark.parametrize("min_amount", [uint64(0), uint64(30000)])
```
arvidn, line 555: I would use some uncomfortable values here. Like, `0xffff`, `0x7fff`, `0x8000`, or `0x8000000000000000`

arvidn (commented):
Looks really good so far!

My understanding was that there was consensus (or at least some agreement) on altering the protocol to essentially merge the request for subscription with requesting coin states. and the subscription would only kick in if the complete state could be returned. Otherwise the wallet was expected to request the next range of coin states.

The description on this PR appears to not have been updated to reflect this. I missed the meeting where this was discussed, so perhaps I misunderstand something.

chia/full_node/full_node_api.py
Rigidity: The wallet can determine which coin ids were omitted on its own, technically. Perhaps it would make sense though to return only the coin ids that were included?

tests/core/full_node/stores/test_coin_store.py
Rigidity: One of my main concerns with this is that the number of combinations of things to test can blow up exponentially - maybe it's worth it in this case

chia/full_node/full_node_api.py
Rigidity: I assumed this was doing a db lookup as is. Maybe we make this change in a separate PR I can rebase it into this one and switch how it's implemented here. Or we can do it after?

chia/full_node/full_node_api.py
Rigidity: I don't like the idea of keeping track of an additional bit of data associated with each subscription. We do have the `RequestRemoveCoinSubscriptions` message now, so the wallet could unsubscribe from things as needed of its own volition. Do you think that's too roundabout a way to get there, though?

chia/full_node/full_node_api.py
Rigidity: I only did it because I'm tripling the number of places where it's needed - is this important to separate out?

Rigidity:
> My understanding was that there was consensus (or at least some agreement) on altering the protocol to essentially merge the request for subscription with requesting coin states. and the subscription would only kick in if the complete state could be returned. Otherwise the wallet was expected to request the next range of coin states.

`RequestPuzzleState` syncs batch by batch until it reaches the peak and subscribes from there on, and `RequestAddPuzzleSubscriptions` directly adds the subscriptions without requesting initial coin data and returns the current peak as of when you subscribed.

I can get rid of `RequestAddPuzzleSubscriptions`, but I do like the flexibility having it offers you if you want it. And having both `Add`/`Remove` made logical sense to me. Though, I think you could technically get the same behavior by just calling `RequestPuzzleState` with `CoinStateFilters(False, False, False, 0)` to avoid getting any coins and immediately subscribe. So maybe it would just be unnecessary to have it. You could also just call `RegisterForPhUpdates` and ignore the response I suppose.

So I guess it's essentially just fluff?

chia/full_node/full_node_api.py
Rigidity: Yeah I haven't been perfectly happy with this part either, I'll take a look at it again

chia/full_node/full_node_api.py
Rigidity: Seems reasonable, I think there's already a method to remove all puzzle subscriptions, and if not I can add it

chia/full_node/full_node_api.py
arvidn: probably not

chia/full_node/full_node_api.py
arvidn: the only argument in favour of auto-unsubscribe is to let the node clear out subscriptions by itself (and thus saving memory). it's not a strong case.

chia/full_node/full_node_api.py
arvidn: I agree on separate PR. a function performing a db lookup has to be `async`. That's the main hurdle for transitioning.

arvidn:
> I can get rid of RequestAddPuzzleSubscriptions, but I do like the flexibility having it offers you if you want it.

Perhaps I'm missing something. Given that the wallet can't know which block will be the node's peak (by the time it receives the request) I question the usefulness of being able to *just* subscribe. You can't know what you're getting up front. You do see the height and block hash in the subscription response, so you *do* know what you ended up with.

chia/full_node/full_node_api.py
arvidn: the logic here looks a bit fishy. It seems like we end up subscribing, even when we're not "done" sending all the state. It doesn't look like there's any message indicating this (like a subscription response message).

Part of my confusion here is that coin IDs are fundamentally different than puzzle hashes.

chia/full_node/full_node_api.py
```diff
@@ -1720,5 +1712,201 @@ def get_fee_estimates(est: FeeEstimatorInterface, req_times: List[uint64]) -> Li
...
+        )
+
+        if request.subscribe:
+            peer_id = peer.peer_node_id
+            added = list(
+                self.full_node.subscriptions.add_coin_subscriptions(peer_id, request.coin_ids, max_subscriptions)
```
arvidn, line 1890: it would seem reasonable to only subscribe to the coin IDs that were included in the response.

Rigidity: This would be an argument for `RequestAddCoinSubscriptions` then, it would be nice to have a way to subscribe to something which doesn't exist yet.

chia/full_node/full_node_api.py
Rigidity: What do you mean by done? There isn't any batching involved with `RequestCoinState` since you're always going to get a number of coin states at most equal to the number of coin ids you ask for in the response.

chia/full_node/full_node_api.py
Rigidity: I'm not sure it's worth the effort it would take to implement this, and every wallet could have different requirements for how it needs to handle subscriptions. I think using either an in memory list or a database to track subscriptions for each peer on the wallet side could work, or worst case you can now reset the subscriptions with the new protocol.

chia/full_node/full_node_api.py
Rigidity: Implemented

tests/core/full_node/stores/test_coin_store.py
Rigidity: Didn't affect it that much, just added these values

chia/full_node/coin_store.py
Rigidity: The SQLite docs say
> When two BLOB values are compared, the result is determined using `memcmp()`.

https://www.sqlite.org/datatype3.html

chia/full_node/full_node_api.py
arvidn: right. it's the requestor's responsibility to not ask for more coin states than what fit in the response. If you request a coin ID that doesn't exist yet, you still subscribe to it, so you're notified if it appears. Right?

When subscribe means that if there's a reorg, and the coin disappears or re-appears or is spent, you get notified.

It still seems a bit difficult for the requestor to understand what happened if a request hits against `max_items`. You won't get the coin state, even though you requested it. Do you interpret that as if the coin not existing or as the message was truncated?

chia/full_node/full_node_api.py
arvidn: I changed my mind on this. I agree that it would be good to be able to subscribe to coins that don't exist yet

chia/full_node/full_node_api.py
Rigidity: I've removed these messages.

chia/full_node/full_node_api.py
```diff
@@ -1724,5 +1711,192 @@ def get_fee_estimates(est: FeeEstimatorInterface, req_times: List[uint64]) -> Li
...
+        max_subscriptions = self.max_subscriptions(peer)
+        subs = self.full_node.subscriptions
+
+        request_puzzle_hashes = list(dict.fromkeys(request.puzzle_hashes))
+
+        # This is a limit imposed by `batch_coin_states_by_puzzle_hashes`, due to the SQLite variable limit.
+        # It can be increased in the future, and this protocol should be written and tested in a way that
+        # this increase would not break the API.
+        count = 15000
+        puzzle_hashes = request_puzzle_hashes[:count]
```
Rigidity, line 1768: Hard coding this in a couple places isn't great I'll admit. Maybe this should be configurable (though it must be below the SQLite variable limit and the assumption that it's at minimum 32,700).

chia/full_node/full_node_api.py
```diff
@@ -1724,5 +1711,192 @@ def get_fee_estimates(est: FeeEstimatorInterface, req_times: List[uint64]) -> Li
...
+        )
+        assert previous_header_hash is not None
+
+        if request.header_hash != previous_header_hash:
+            rejection = wallet_protocol.RejectPuzzleState(uint8(wallet_protocol.RejectStateReason.REORG))
+            msg = make_msg(ProtocolMessageTypes.reject_puzzle_state, rejection)
+            return msg
+
+        # Check if the request would exceed the subscription limit now.
+        def check_subscription_limit() -> Optional[Message]:
```
Rigidity, line 1783: This check is done twice because it's cheap, just in case there's a race condition of some sort after the await point below. Not sure if this is worth the messy code or even necessary.

chia/full_node/hint_store.py
```diff
@@ -38,6 +39,23 @@ async def get_coin_ids(self, hint: bytes, *, max_items: int = 50000) -> List[byt
             await cursor.close()
         return [bytes32(row[0]) for row in rows]
 
+    async def get_coin_ids_multi(self, hints: Set[bytes], *, max_items: int = 50000) -> List[bytes32]:
```
Rigidity, line 42: I assumed this would be a big optimization from doing a bunch of individual queries, but I don't actually think it's that much faster interestingly. I'm pretty disappointed by our SQLite DB performance overall with the wallet protocol.

chia/protocols/protocol_message_types.py
```diff
@@ -118,4 +118,16 @@ class ProtocolMessageTypes(Enum):
     new_unfinished_block2 = 92
     request_unfinished_block2 = 93
 
+    # New wallet sync protocol
+    request_remove_puzzle_subscriptions = 94
```
Rigidity, line 122: These don't have rate limits set. Working on that now.

chia/protocols/shared_protocol.py
```diff
@@ -7,7 +7,7 @@
 from chia.util.ints import int16, uint8, uint16
 from chia.util.streamable import Streamable, streamable
 
-protocol_version = "0.0.36"
+protocol_version = "0.0.37"
```
Rigidity, line 10: Is this still correct?

Rigidity (commented):
Here's some self comments, for added context to anyone else reviewing

Rigidity:
I think we should either get rid of the duplicated check if it's not important enough, or add `# pragma: no cover` to it. Since it's an almost impossible scenario to be in imo.

chia/full_node/full_node_api.py
arvidn: could you just use the constant instead?

chia/full_node/hint_store.py
arvidn: ideally, we would have a unit test for this function as well

arvidn (commented):
I think we would want rate limits, and if it's not too much trouble, a unit test for the new hint store function

chia/_tests/util/build_network_protocol_files.py
```diff
@@ -87,7 +87,7 @@ def visit_wallet_protocol(visitor: Callable[[Any, str], None]) -> None:
     visitor(respond_header_blocks, "respond_header_blocks")
     visitor(coin_state, "coin_state")
     visitor(register_for_ph_updates, "register_for_ph_updates")
-    visitor(reject_block_headers, "reject_block_headers"),
```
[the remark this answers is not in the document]
Rigidity: Not sure, this was a formatter change, maybe as a result of the new Black update

chia/_tests/core/full_node/stores/test_coin_store.py
```diff
@@ -685,6 +696,15 @@ async def test_batch_many_coin_states(db_version: int, cut_off_middle: bool) ->
         assert len(all_coin_states) == (25001 if cut_off_middle else 50000)
 
 
+@pytest.mark.anyio
+async def test_batch_no_puzzle_hashes(db_version: int) -> None:
+    async with DBConnection(db_version) as db_wrapper:
+        # Initialize coin and hint stores.
+        coin_store = await CoinStore.create(db_wrapper)
+        await HintStore.create(db_wrapper)
+        await coin_store.batch_coin_states_by_puzzle_hashes([])
```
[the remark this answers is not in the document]
Rigidity: Yes I can add that - I initially added the test because I found a bug where if the request was empty the query would fail with my old implementation, so I wanted to make sure that was covered in the future.

chia/full_node/full_node_api.py
Rigidity: This has been fixed - if the request is higher than max items it will fail truncate it, and the returned coin ids will represent everything which was actually queried (deduplicated). If it exceeds the subscription limit the request will fail now.

chia/full_node/coin_store.py
```diff
@@ -432,18 +433,18 @@ async def batch_coin_states_by_puzzle_hashes(
         # It can be changed later without breaking the protocol, but this is a practical limit for now.
         assert len(puzzle_hashes) <= 15000
```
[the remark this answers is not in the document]
Rigidity: Bumped it up to match the variable and use a constant.

chia/full_node/coin_store.py
```diff
@@ -455,33 +456,49 @@ async def batch_coin_states_by_puzzle_hashes(
...
+                    ),
                 )
 
-            for row in await cursor.fetchall():
-                coin_states.append(self.row_to_coin_state(row))
+                for row in await cursor.fetchall():
+                    coin_states.append(self.row_to_coin_state(row))
+
+                coin_states.sort(key=lambda cr: max(cr.created_height or uint32(0), cr.spent_height or uint32(0)))
+                while len(coin_states) > max_items + 1:
```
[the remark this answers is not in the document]
Rigidity: Bumped it up, we can always revisit this later but it's fine to be higher for now.

chia/full_node/full_node_api.py
```diff
@@ -1596,14 +1592,9 @@ async def register_interest_in_puzzle_hash(
...
 
-        hint_coin_ids: Set[bytes32] = set()
-        if max_items > 0:
-            for puzzle_hash in puzzle_hashes:
-                ph_hint_coins = await self.full_node.hint_store.get_coin_ids(puzzle_hash, max_items=max_items)
-                hint_coin_ids.update(ph_hint_coins)
-                max_items -= len(ph_hint_coins)
-                if max_items <= 0:
-                    break
+        hint_coin_ids = await self.full_node.hint_store.get_coin_ids_multi(
```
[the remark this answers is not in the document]
Rigidity: Added the test

chia/full_node/full_node_api.py
Rigidity: Yes, done

chia/full_node/hint_store.py
Rigidity: Just added it

chia/protocols/protocol_message_types.py
Rigidity: Now they do

chia/protocols/shared_protocol.py
Rigidity: Will need to update this once https://github.com/Chia-Network/chia-blockchain/pull/17791 lands.

chia/_tests/wallet/test_new_wallet_protocol.py
```diff
@@ -359,32 +368,14 @@ async def test_request_coin_state_limit(one_node: OneNode, self_hostname: str) -
...
+    assert len(response.coin_states) == len(coin_records) - 20000
 
     for coin_state in response.coin_states:
         coin_record = coin_records[coin_state.coin.name()]
         assert coin_record.coin_state == coin_state
         assert coin_record.confirmed_block_index > 1
 
-    # The expected behavior when the limit is exceeded, is to skip the rest
-    # The order is not guaranteed, so it can't be relied upon here
-    # Just request the remaining coin ids in subsequent request(s)
```
arvidn: did this behavior change? Do we just get an error now, and none of the coins were subscribed to?

arvidn (commented):
looks good!

chia/_tests/wallet/test_new_wallet_protocol.py
Rigidity: Yes

arvidn (approved):

arvidn:
I think it's reasonable to not cover the checks for race condition

PR #17346, opened 2024-01-16
Explore looking beyond mempool items that hit the maximum cost limit

AmineKhaldi:


chia/full_node/mempool.py
```diff
@@ -403,15 +404,30 @@ def create_bundle_from_mempool_items(
                 continue
             try:
                 cost = uint64(0 if item.npc_result.conds is None else item.npc_result.conds.cost)
-                unique_coin_spends, cost_saving, unique_additions = eligible_coin_spends.get_deduplication_info(
-                    bundle_coin_spends=item.bundle_coin_spends, max_cost=cost
-                )
+                if skipped_items >= 3:
```
arvidn, line 407: I think a comment is warranted here, explaining what's going on.
If we've encountered 3 transactions that don't fit in the remaining block size, we want to keep looking for smaller transactions that might fit, but to avoid spending too much time on expensive transactions, we start to take short cuts.

chia/full_node/mempool.py
```diff
@@ -403,15 +404,30 @@ def create_bundle_from_mempool_items(
...
+                    # Process this item straight away (no eligbility consideration)
+                    log.info("Processing mempool item without eligbility consideration: %s", name.hex())
+                    unique_coin_spends = []
+                    unique_additions = []
+                    for spend_data in item.bundle_coin_spends.values():
+                        unique_coin_spends.append(spend_data.coin_spend)
+                        unique_additions.extend(spend_data.additions)
+                    cost_saving = 0
```
arvidn, line 415: I would have expected that if this is a transaction eligible for identical spend deduplication, we have to just skip it. otherwise we'll include a double spend in the block, no?
alternatively, if the only reason we have to perform the expensive operation (running CLVM) is to compute `cost_savings`, we could do everything else but just over-estimate the cost of the transaction.

It's not obvious to me that this is correct. Even if it is, I think these questions ought to be answered in comments here.

chia/full_node/mempool.py
```diff
@@ -403,15 +404,30 @@ def create_bundle_from_mempool_items(
...
+                    )
                 item_cost = cost - cost_saving
                 log.info("Cumulative cost: %d, fee per cost: %0.4f", cost_sum, fee / item_cost)
                 if (
                     item_cost + cost_sum > self.mempool_info.max_block_clvm_cost
                     or fee + fee_sum > DEFAULT_CONSTANTS.MAX_COIN_AMOUNT
                 ):
+                    # Let's skip this one
+                    skipped_items += 1
+                    if skipped_items < 20:
```
arvidn, line 428: I think we should also break out of the loop here if `item_cost` is less than some threshold. That would mean that we're getting very close to the limit anyway, and maybe we won't realistically find a transaction small enough.

i.e.

```
if skipped_items < 20 and item_cost > 6000000:
```

the 6 million cut-of comes from: https://www.chia.net/2024/01/11/getting-to-know-the-mempool-and-transaction-fees/

chia/full_node/mempool.py
```diff
@@ -25,6 +25,11 @@
 
 log = logging.getLogger(__name__)
 
+# Maximum number of items to skip when looking for transactions that fit a block
```
arvidn, line 28: I think this comment should explain what it means for an item to be "skipped".

chia/full_node/mempool.py
```diff
@@ -395,6 +400,8 @@ def create_bundle_from_mempool_items(
         log.info(f"Starting to make block, max cost: {self.mempool_info.max_block_clvm_cost}")
         with self._db_conn:
             cursor = self._db_conn.execute("SELECT name, fee FROM tx ORDER BY fee_per_cost DESC, seq ASC")
+        skipped_items = 0
+        STANDARD_XCH_SPEND_COST = 6_000_000
```
arvidn, line 404: good idea to give this a name. I think it would make sense to put it at the top with the other constants and also give it a comment that this determines how close to the block size limit we're willing to go, basically

chia/full_node/mempool.py
```diff
@@ -403,15 +410,39 @@ def create_bundle_from_mempool_items(
                 continue
             try:
                 cost = uint64(0 if item.npc_result.conds is None else item.npc_result.conds.cost)
-                unique_coin_spends, cost_saving, unique_additions = eligible_coin_spends.get_deduplication_info(
-                    bundle_coin_spends=item.bundle_coin_spends, max_cost=cost
-                )
+                if skipped_items >= ELIGIBILITY_SKIP_THRESHOLD:
```
arvidn, line 413: I think it would be better to call this by the actual feature, e.g. `DEDUP` rather than the fact that items are *eligible* for said feature. It will become especially confusing when we add more features to the mempool that items may be eligible for

chia/full_node/mempool.py
```diff
@@ -403,15 +410,39 @@ def create_bundle_from_mempool_items(
...
+                    # If we've encountered `ELIGIBILITY_SKIP_THRESHOLD`
+                    # transactions that don't fit in the remaining block size,
+                    # we want to keep looking for smaller transactions that
+                    # might fit, but we also want to avoid spending too much
+                    # time on potentially expensive ones, hence this shortcut.
+                    unique_coin_spends = []
+                    unique_additions = []
+                    for spend_data in item.bundle_coin_spends.values():
+                        if spend_data.eligible_for_dedup:
+                            raise Exception(f"Skipping transaction with eligible coin(s): {name.hex()}")
```
arvidn, line 423: just to make sure I understand this correctly; If we don't abort here, we would construct an invalid block, right? We would include a double-spend.

Also, we could just `continue` here, rather than throwing an exception, couldn't we?

Ganbin (commented):
I have done something similar in #17374 but this code seems more efficient I believe. 

According to some benchmark I have done ([see results in gist](https://gist.github.com/Ganbin/beb3734549529a5cce9e35438fc60042#file-results_10_small-txt)) The `MAX_SKIPPED_ITEMS = 20` seems to be a good value.

chia/full_node/mempool.py
```diff
@@ -403,15 +418,39 @@ def create_bundle_from_mempool_items(
...
+                    cost_saving = 0
+                else:
+                    unique_coin_spends, cost_saving, unique_additions = eligible_coin_spends.get_deduplication_info(
+                        bundle_coin_spends=item.bundle_coin_spends, max_cost=cost
+                    )
                 item_cost = cost - cost_saving
                 log.info("Cumulative cost: %d, fee per cost: %0.4f", cost_sum, fee / item_cost)
                 if (
                     item_cost + cost_sum > self.mempool_info.max_block_clvm_cost
                     or fee + fee_sum > DEFAULT_CONSTANTS.MAX_COIN_AMOUNT
```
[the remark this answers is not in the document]
AmineKhaldi: Yes, it's purely defensive. I've separated the condition and added an informative comment.

Ganbin:
> I sent an update adopting your suggestions. Please consider benchmarking it when you have a moment.

I have tested this solution in my benchmark and I can see that it is more efficient to reset the counter each time we find a new transaction to include. It prevent to loop through all the 20 items (`MAX_SKIPPED_ITEMS`).

You can see on the code of the PR #17374 that I have implemented a `max_retry_add_more_items = 10`. Which mean that we only look the next 10 items when one item exceed the max size. If we find 1 item we reset the counter.

On my benchmark I only reach a maximum of 6 retry where the average was `1.14` `[0,3,1,0,1,6,0,0,1,0,1,1,1,1]` I believe we could set the `max_retry_add_more_items = 5` to gain some small efficiency (see result below).

On the number of total skipped items we can see the difference:

With `MAX_SKIPPED_ITEMS = 20` : `[0,20,20,0,20,7,0,0,20,0,1,20,20,1]`
With `max_retry_add_more_items = 10` : `[0,23,12,0,11,7,0,0,12,0,1,12,12,1]`
With `max_retry_add_more_items = 5`: `[0,18,7,0,6,6,0,0,7,0,1,7,7,1]`

In all 3 cases we were able to include the same number of transactions.

Maybe this is going too deep in the optimization, feel free to skip it if you feel it is not necessary.

Also my sample are quite small (only 14 full mempool json) although some were taken when a big spend was preventing the block to be full.

_*Notice that my benchmark are only to see the efficiency of the "algorithm", it doesn't take into account the `eligible_coin_spends.get_deduplication_info` improvements_

chia/full_node/mempool.py
```diff
@@ -403,22 +418,58 @@ def create_bundle_from_mempool_items(
...
-                )
+                if skipped_items >= PRIORITY_TX_THRESHOLD:
+                    # If we've encountered `PRIORITY_TX_THRESHOLD` number of
+                    # transactions that don't fit in the remaining block size,
+                    # we want to keep looking for smaller transactions that
+                    # might fit, but we also want to avoid spending too much
+                    # time on potentially expensive ones, hence this shortcut.
+                    unique_coin_spends = []
+                    unique_additions = []
+                    for spend_data in item.bundle_coin_spends.values():
```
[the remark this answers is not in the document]
AmineKhaldi: We're trying to balance between code complexity and efficiency. For instance, here we'd need to also update the skipped items count, check if we reached the maximum skipped items and branch accordingly... etc.

chia/full_node/mempool.py
```diff
@@ -403,22 +418,58 @@ def create_bundle_from_mempool_items(
...
-                    or fee + fee_sum > DEFAULT_CONSTANTS.MAX_COIN_AMOUNT
-                ):
+                log.info(
+                    "Cumulative cost: %d, fee per cost: %0.4f, item cost: %d", cost_sum, fee / item_cost, item_cost
+                )
+                new_fee_sum = fee_sum + fee
+                if new_fee_sum > DEFAULT_CONSTANTS.MAX_COIN_AMOUNT:
+                    # Such a fee is very unlikely to happen but we're defensively
+                    # accounting for it
+                    break
```
arvidn, line 447: this line doesn't have test coverage. given our restrictions on per-transaction fee, it might be hard to construct a scenario where this is exercised. I would be fine with a `pragma: no cover`

arvidn (commented):
I think this is looking good. if anything, it might be nice to have specific tests that create a mempool with fake spend bundles with configurable costs and fees. To ensure we pick the items we'd expect. with some parameterized test

emlowe:
Looks good to me

chia/full_node/mempool.py
```diff
@@ -403,6 +404,16 @@ def create_bundle_from_mempool_items(
...
+                    # We're trying beyond the skipped item, so do a quick check
+                    # for the cost as-is, and the fees, against the limits
+                    if (
+                        cost + cost_sum > self.mempool_info.max_block_clvm_cost
+                        or fee + fee_sum > DEFAULT_CONSTANTS.MAX_COIN_AMOUNT
+                    ):
+                        # This item is not going to fit either, we're done here
+                        break
+
                 unique_coin_spends, cost_saving, unique_additions = eligible_coin_spends.get_deduplication_info(
```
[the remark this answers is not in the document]
AmineKhaldi: We have logic centered around `PRIORITY_TX_THRESHOLD` for that purpose.

chia/full_node/mempool.py
```diff
@@ -412,7 +423,10 @@ def create_bundle_from_mempool_items(
                     item_cost + cost_sum > self.mempool_info.max_block_clvm_cost
                     or fee + fee_sum > DEFAULT_CONSTANTS.MAX_COIN_AMOUNT
                 ):
-                    break
+                    # We should be done, but let's skip this one and try the
+                    # next one, in case it fits
+                    skipped_an_item = True
```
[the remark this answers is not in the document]
AmineKhaldi: We have some levers around items with eligible spends, items with higher cost than typical spends and overall maximum skipped items.

arvidn (approved):
nice job!

PR #17354, opened 2024-01-18
fix duplicates file and upload name uniqueness

altendky:
### Purpose:

uploaded artifacts must now have unique names.  also, this makes the file names unique for coverage and pytest outputs so they can all make it to the post job without clobbering each other.

### DIFF
--- a/tests/build-job-matrix.py
+++ b/tests/build-job-matrix.py
@@ -145,9 +145,13 @@ for path, index in test_paths_with_index:
 
     max_index_characters = len(str(args.duplicates))
     index_string = f"{index:0{max_index_characters}d}"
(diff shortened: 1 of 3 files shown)


tests/build-job-matrix.py
```diff
@@ -143,6 +143,16 @@ def update_config(parent: Dict[str, Any], child: Dict[str, Any]) -> Dict[str, An
 
         pytest_monitor_enabling_paths.append(path)
 
+    max_index_characters = len(str(args.duplicates))
```
arvidn, line 146: technically, I think this should be `args.duplicates-`. consider the case of `--duplicates 10`. you'll have indices 0-9

arvidn (approved):

PR #17356, opened 2024-01-18
Annotate nft_puzzles.py

AmineKhaldi:


chia/wallet/nft_wallet/nft_puzzles.py
```diff
@@ -58,11 +58,7 @@ def create_nft_layer_puzzle_with_curry_params(
 
 def create_full_puzzle_with_nft_puzzle(singleton_id: bytes32, inner_puzzle: Program) -> Program:
     if log.isEnabledFor(logging.DEBUG):
-        log.debug(
-            "Creating full NFT puzzle with inner puzzle: \n%r\n%r",
-            singleton_id,
-            inner_puzzle.get_tree_hash(),
-        )
+        log.debug("Creating full NFT puzzle with inner puzzle: \n%r\n%r", singleton_id, inner_puzzle.get_tree_hash())
```
arvidn, line 61: this `get_tree_hash()` is probably pretty expensive. I believe we call this regardless of whether debug logging is enabled

arvidn (approved):
I would guess that using a pragma comment to suppress a type warning is better than `cast()` because once the typing is fixed for `clvm`, mypy will warn about the pointless pragma and we can remove it. Will mypy warn about a pointless cast?

AmineKhaldi:
> I would guess that using a pragma comment to suppress a type warning is better than `cast()` because once the typing is fixed for `clvm`, mypy will warn about the pointless pragma and we can remove it. Will mypy warn about a pointless cast?

Yep, `[redundant-cast]`.

chia/wallet/nft_wallet/nft_puzzles.py
AmineKhaldi: We have it under the condition `if log.isEnabledFor(logging.DEBUG):` so that should be fine.

PR #17362, opened 2024-01-18
Annotate test_singleton_lifecycle_fast.py

AmineKhaldi:


tests/wallet/test_singleton_lifecycle_fast.py
```diff
@@ -31,25 +31,25 @@
...
-    def add_puzzle(self, puzzle: Program):
+    def add_puzzle(self, puzzle: Program) -> None:
         self._db[puzzle.get_tree_hash()] = Program.from_bytes(bytes(puzzle))
 
     def puzzle_for_hash(self, puzzle_hash: bytes32) -> Optional[Program]:
         return self._db.get(puzzle_hash)
 
 
-def from_kwargs(kwargs, key, type_info=Any):
+def from_kwargs(kwargs: Dict[str, Any], key: str, type_info: Any = Any) -> Any:
```
arvidn, line 52: this looks fishy:
```
type_info: Any = Any
```

Should it not be `Type[Any]`?

arvidn (approved):

tests/wallet/test_singleton_lifecycle_fast.py
AmineKhaldi: I tried that originally but I got `error: Incompatible default for argument "type_info" (default has type "object", argument has type "type[Any]")  [assignment]`. Please note this is currently not even used (as noted in the function's docstring) so I deemed `Any` to be adequate at this point.

arvidn: could we just delete the parameter instead?

AmineKhaldi: Sure. PR #17375.

PR #17369, opened 2024-01-20
Add support for defining a list of full node peers to connect to

felixbrucker:
### Purpose:
Allow the full node to connect to a list of full nodes configured via `config.yaml` and automatically reconnect should the connection be interrupted/disconnected.

### Current Behavior:
Currently this is not supported, a workaround is to add peers manually via cli, but once not connected anymore they will not automatically reconnect.

### New Behavior:
A list of full nodes can be configured via `full_node_peers` in the `config.yaml`, the same way other peers are already configured.

### DIFF
--- a/chia/server/start_full_node.py
+++ b/chia/server/start_full_node.py
@@ -16,7 +16,7 @@ from chia.server.outbound_message import NodeType
 from chia.server.start_service import RpcInfo, Service, async_run
 from chia.types.aliases import FullNodeService
 from chia.util.chia_logging import initialize_service_logging
-from chia.util.config import load_config, load_config_cli
+from chia.util.config import get_unresolved_peer_infos, load_config, load_config_cli
 from chia.util.default_root import DEFAULT_ROOT_PATH
 from chia.util.ints import uint16
 from chia.util.misc import SignalHandlers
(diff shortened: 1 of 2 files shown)


xearl4:
I am very much looking forward to this! We (spacefarmers.io) operate several meshes of full nodes to support pool needs. At the moment, we use a script to interconnect the nodes and keep them connected. I'll be happy to abandon that script :)

emlowe:
Well, that seems simple enough.

emlowe (approved):

TheLastCicada:
This will also be very useful if running a small private testnet where all peers are known.  Can skip running an introducer potentially.

danieljperry:
What is the expected syntax when adding a full node peer to the config? For example, something like this?
```
full_node_peers:
    host: 12.34.56.78
    port: 8444
```

felixbrucker:
Almost:
```yaml
full_node_peers:
  - host: 12.34.56.78
    port: 8444
```

danieljperry:
I previously tried using a hyphen, but I didn't indent the `port` initially. Now I have it working. Thanks!

LCG22:
Does this wallet support these settings?

felixbrucker:
@LCG22 this PR only affects the full node service, not the wallet service

LCG22:
> 此 PR 仅影响全节点服务，不影响钱包服务

Can you let it support the wallet ？

felixbrucker:
As far as i know this is already possible

Issue #17186, opened 2024-01-02
[Bug] MempoolManager.new_peak() fails in case of a re-org where the new tip is not a transaction block


madMAx43v3r:
### What happened?

In case of a re-org, `MempoolManager.new_peak()` is only called with the new peak, however the function early exits if the given peak is not a transaction block:
https://github.com/Chia-Network/chia-blockchain/blob/cd78dbafdd8e8b5fc109eadcfcb6c2fbbd49ad5f/chia/full_node/mempool_manager.py#L622

As such the mempool is not updated correctly in case of a re-org where the new peak is not a transaction block. Only at the next transaction block it will be corrected.

### Version

master

### What platform are you using?

Linux

### What ui mode are you using?

CLI

wjblanke:
Thanks Arvid will take a look

arvidn:
It's not immediately obvious what the best approach to fixing this is. Maybe we could back-track to find the most recent transaction block to see if it matches the current peak.

Another approach, that might be simpler, would be to update `self.peak` every time we see a block, including non-transaction blocks. That way, a (non-tx) block that doesn't follow in the chain could trigger the mempool reorg

arvidn:
the mempool doesn't have the ability to traverse the blockchain. If it gets a non-tx block that's disconnected from its current peak, it can't step back to find the most recent transaction block.

I think this issue has to be fixed in `full_node.py` and make sure it only passes transaction blocks to `mempool_manager.new_peak()`

arvidn:
I'm addressing this issue here: https://github.com/Chia-Network/chia-blockchain/pull/17370

wjblanke:
Thanks for the report!

PR #17370, opened 2024-01-20
improve mempool reorg logic when the peak is a non-transaction block

arvidn:
## Purpose:

This PR addresses the issue raised by https://github.com/Chia-Network/chia-blockchain/issues/17186.

## Background

In order for the mempool to evict transactions as they become invalid (either because they made it into a block or the coins they spend became spent by some other transaction), it needs to know about new blocks added to the chain.

The "ground truth" about whether a transaction is valid or not is the `CoinStore`, which the mempool has access to via the `get_coin_record` function, passed into it. This function reflects the peak of the chain, the current coin set.

### transaction blocks

The mempool needs to keep a reference to the most recent *transaction* block, because those are what matters for timelock conditions. e.g. height-based timelock conditions compare the argument against the *previous* transaction block's height, and the time-based timelock conditions compare the argument against the *previous* transaction block's timestamp. Only transaction blocks have timestamps. (see [mempool_check_time_locks()](https://github.com/Chia-Network/chia-blockchain/blob/main/chia/full_node/mempool_check_conditions.py#L189))

Since the mempool validates transactions before including them, it needs to know about the most recent transaction block, not necessarily the actual peak.

### Reorgs

When a re-org happens, the fork we switch over to is not replayed one block at a time to the mempool, it will just learn about the most recent peak. If this peak is not a transaction block, there's not much the mempool can do about it (currently). It can't update its set of transactions, because it doesn't know what the most recent transcaction block is.

As pointed out in https://github.com/Chia-Network/chia-blockchain/issues/17186, this delays updating the mempool until we receive a transaction block on the new chain. This in turn risks us failing to create a valid block if we happen to find a valid proof-of-space for that next transaction block.

## Solution

Since the mempool itself can't traverse the blockchain, even if it would detect the reorg, it can't step back to find the most recent transaction block on the new chain. A simpler solution (I believe) is to ensure that the full node only ever passes the most recent transaction block to the mempool.

This patch adds a sibling function to `Blockchain.get_peak()` called `Blockchain.get_tx_peak()`, which returns the most recent transaction block (which may be the peak, if that is a transaction block).

### Current Behavior:

the mempool may be delayed one block in updating after a reorg

### New Behavior:

The mempool updates immediately with a reorg

### Testing Notes:

There are two new tests for the `Blockchain` class, one where blocks are just added linearly and the `get_tx_block()` is tested to always return the most recent tx block. The other test exrcises a reorg where the switch-over happens on a non-transaction block, and ensures it still returns the most recent one.

### DIFF
--- a/chia/consensus/blockchain.py
+++ b/chia/consensus/blockchain.py
@@ -194,6 +194,26 @@ class Blockchain(BlockchainInterface):
             return None
         return self.height_to_block_record(self._peak_height)
 
+    def get_tx_peak(self) -> Optional[BlockRecord]:
+        """
+        Return the most recent transaction block. i.e. closest to the peak of the blockchain
+        Requires the blockchain to be initialized and there to be a peak set
+        """
+
+        if self._peak_height is None:
+            return None
+        tx_height = self._peak_height
+        tx_peak = self.height_to_block_record(tx_height)
+        while not tx_peak.is_transaction_block:
+            # it seems BlockTools only produce chains where the first block is a
+            # transaction block, which makes it hard to test this case
+            if tx_height == 0:  # pragma: no cover
+                return None
+            tx_height = uint32(tx_height - 1)
+            tx_peak = self.height_to_block_record(tx_height)
+
+        return tx_peak
+
     async def get_full_peak(self) -> Optional[FullBlock]:
         if self._peak_height is None:
             return None
--- a/chia/full_node/full_node.py
+++ b/chia/full_node/full_node.py
@@ -316,7 +316,7 @@ class FullNode:
                     f"time taken: {int(time_taken)}s"
                 )
                 async with self.blockchain.priority_mutex.acquire(priority=BlockchainMutexPriority.high):
-                    pending_tx = await self.mempool_manager.new_peak(peak, None)
+                    pending_tx = await self.mempool_manager.new_peak(self.blockchain.get_tx_peak(), None)
                 assert len(pending_tx) == 0  # no pending transactions when starting up
 
                 full_peak: Optional[FullBlock] = await self.blockchain.get_full_peak()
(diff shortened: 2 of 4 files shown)


chia/full_node/full_node.py
```diff
@@ -1544,9 +1544,8 @@ async def peak_post_processing(
 
         # Update the mempool (returns successful pending transactions added to the mempool)
         spent_coins: List[bytes32] = [coin_id for coin_id, _ in state_change_summary.removals]
-        mempool_new_peak_result: List[Tuple[SpendBundle, NPCResult, bytes32]] = await self.mempool_manager.new_peak(
-            self.blockchain.get_peak(), spent_coins
-        )
+        mempool_new_peak_result: List[Tuple[SpendBundle, NPCResult, bytes32]]
+        mempool_new_peak_result = await self.mempool_manager.new_peak(self.blockchain.get_tx_peak(), spent_coins)
```
AmineKhaldi, line 1548:
```suggestion
        mempool_new_peak_result: List[Tuple[SpendBundle, NPCResult, bytes32]] = await self.mempool_manager.new_peak(
            self.blockchain.get_tx_peak(), spent_coins
        )
```

AmineKhaldi (commented):
Looks good! Just one suggestion.

chia/consensus/blockchain.py
```diff
@@ -194,6 +194,24 @@ def get_peak(self) -> Optional[BlockRecord]:
...
+        Requires the blockchain to be initialized and there to be a peak set
+        """
+
+        if self._peak_height is None:
+            return None
+        tx_height = self._peak_height
+        tx_peak = self.height_to_block_record(tx_height)
+        while not tx_peak.is_transaction_block:
+            if tx_height == 0:
+                return None
```
arvidn, line 209: would it make sense to add a `pragma: no cover` for this? as far as I can tell, the chains we generate with block tools all start with a transaction block

chia/full_node/full_node.py
arvidn: I prefer keeping the function call in a single line

chia/consensus/blockchain.py
AmineKhaldi: With a comment to document this assumption, I think that sounds reasonable.

AmineKhaldi (approved):

emlowe (approved):

chia/consensus/blockchain.py
Starttoaster: Closing because the conversation blocks the merge button and it looks like this was resolved anyway..

PR #17375, opened 2024-01-22
Introduce satisfies_hint to check objects against type hints, use it in from_kwargs and fix all the incorrect and missing cases

AmineKhaldi:


arvidn:
coverage report:
```
tests/wallet/test_singleton_lifecycle_fast.py (93.3%): Missing lines 112
```

tests/wallet/test_singleton_lifecycle_fast.py
```diff
@@ -106,7 +109,7 @@ def solve_anyone_can_spend(solver: Solver, puzzle_db: PuzzleDB, args: List[Progr
     This is the anyone-can-spend puzzle `1`. Note that farmers can easily steal this coin, so don't use
     it except for testing.
     """
-    conditions = from_kwargs(kwargs, "conditions", List[Program])
+    conditions = from_kwargs(kwargs, "conditions")
```
arvidn, line 112: do we not have any test for this function?

tests/wallet/test_singleton_lifecycle_fast.py
```diff
@@ -49,9 +49,12 @@ def puzzle_for_hash(self, puzzle_hash: bytes32) -> Optional[Program]:
         return self._db.get(puzzle_hash)
 
 
-def from_kwargs(kwargs: Dict[str, Any], key: str, type_info: Any = Any) -> Any:
-    """Raise an exception if `kwargs[key]` is missing or the wrong type"""
-    """for now, we just check that it's present"""
+def from_kwargs(kwargs: Dict[str, Any], key: str) -> Any:
```
arvidn, line 52: it looks like the `type_info` parameter is meant to determine the return type. maybe construct it from the value in the dict, or cast it.

arvidn (commented):
it looks like not using the type parameter is the problem, not that we pass it in for no reason

tests/util/test_misc.py
```diff
@@ -428,3 +431,24 @@ async def test_recording_web_server_specified_response(
...
+        (uint64(42), uint64, True),
+        ("42", int, False),
+        ([4, 2], List[int], True),
+        ([4, "2"], List[int], False),
+        ((4, 2), Tuple[int, int], True),
+        ((4, "2"), Tuple[int, int], False),
+        ((4, 2), Tuple[int, ...], True),
+        ((4, "2"), Tuple[int, ...], False),
+        ([(4, Program.to([2]))], List[Tuple[int, Program]], True),
+        ([(4, "2")], Tuple[int, str], False),
```
arvidn, line 450: I don't see `Dict` here, even though it looks like `satisfies_hint()` supports it

arvidn (approved):
very nice!

tests/util/test_misc.py
AmineKhaldi: Yep, I added cases that are similar to how we're currently using it, but I can add more if needed.

Starttoaster:
Pinged Arvid to see if the comments here are resolved. There's one posted in the 11th hour after approving, and a couple more earlier on that I'm assuming can be resolved since the approval came after.

